PR #1143, opened 2022-12-09
RemoveArtifactsRecording with amplitude scaling and time jittering

yger:
Hi @alejoe91 @samuelgarcia 

Why my RemoveArtifactsRecording object is now only able to run with 1 core, and not with n_jobs=-1? I mean, I can create the recording, this is working, but if used later by a waveform_extractor objects, only n_jobs = 1 is allowed. 

Maybe I should do a save() of the objects? Is my situation doomed?


yger:
@alejoe91 I made the changes we discussed, and while this is better, this is still not working. I have an error that I don't understand... My pipeline is the following: file in int16, filtering with output in float32. If I use the removeartefact objects after, this is working with 1 core, the data internally are in float32, everything is smooth. Using n_jobs > 1, data internally are in int16 !! and thus the subtraction of the artefacts (in float32) is invalid.

yger:
The issue might be allow related to the fact that I'm slicing my channels, to exclude some. I'll try to make a minimal script, but there is a bug in the propagation of info in the lazy chain, somewhere

alejoe91:
Can you try without slicing? Also check the dtype for each step's get_traces()?

alejoe91:
@yger strange! Let's look more into it next week

yger:
Seems that forcing the artefacts to be of the same type as the recording is solving the problem. Will test more on Monday

yger:
The problem lies in the file waveform_tools.py, and I guess start L247 with the load_extractor function. In fact, when loading a int16 recording, even if the pre-processing chain will involve some dtype conversion (to float32 here), then when reloaded, the dtypes are not propagated and thus the initial dtype is kept and exposed. By adding a line 
print(recording.dtype)
Just after this load_extractor call, we can see that this is int16 with several jobs (wrong), and float32 with only 1 (true, because my pre processing pipeline has filtered the recording and change dtype to float32)

yger:
@samuelgarcia @alejoe91 Let me know if you have an amazing and simple idea to solve such a situation before I'll start patching it on my own. I guess, the idea is that every Preprocessor should have a dtype. And when reloaded, maybe the last object should inherit from the dtype of the lazy cascade instead of just its initial type....

alejoe91:
Where did you add the print line?

yger:
You can add it L248 of the file waveform_tools.py

If n_jobs = 1, the dtype is correct (float32, because a dtype has been changed by downstream filtering object). If this is higher, then dtype is the one of the original file (int16)

alejoe91:
@yger I think I found the issue!!! 

Can you add `dtype` in the filter `self._kwargs` here?
https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/preprocessing/filter.py#L94-L96

yger:
Problem is solved thanks to @alejoe91 . Dtype was the issue

yger:
Everything is fine now @alejoe91 this is good to merge (on my side...) Best and thanks a lot for the help!

alejoe91:
I'll let @samuelgarcia have a final look!

yger:
Would you be fine with an extension of this object such that when mode median or average is selected, we can have the option to scale the template in amplitudes (scale_amplitude), and/or also align it with some optimal time jitter? 
A bit like template-matching, but it turns out this is quite important for fine artefact removal.
Moreover, what about providing a WaveformExtractor itself in the constructor if we want to subtract pre-computed artefacts? Maybe this would be simpler at the end, especially when the Sparsity will be properly handled at the WaveformExtractor level. Because ideally, one would like to be able to define a spatial mask for the artefacts, in order to know how to optimize jitters/amplitudes

alejoe91:
> Would you be fine with an extension of this object such that when mode median or average is selected, we can have the option to scale the template in amplitudes (scale_amplitude), and/or also align it with some optimal time jitter? A bit like template-matching, but it turns out this is quite important for fine artefact removal. Moreover, what about providing a WaveformExtractor itself in the constructor if we want to subtract pre-computed artefacts? Maybe this would be simpler at the end, especially when the Sparsity will be properly handled at the WaveformExtractor level. Because ideally, one would like to be able to define a spatial mask for the artefacts, in order to know how to optimize jitters/amplitudes

Sounds good to me!

yger:
The PR is good to me, an additional time jitter can be given to optimize the fit of the templates. Currently, only the cosine metric is used, but maybe an option could be added to  use L2 norm?

yger:
I guess I should modify it a bit to deal with sparse waveformextractor, I'll let you know when done

yger:
Sounds good to me. Now user can provide a dict for the WaveFormExtractor that will be used to detect the artefacts, so sparsity can be handed via the params there. Clean

alejoe91 (approved):

PR #1145, opened 2022-12-12
Implement tPCA component

### LINKED ISSUE #1123: [implementation] tPCA component
### Current issue

PCA is a common way to reduce waveform dimensionality and extract features. The Paninski lab uses a temporal PCA (tPCA) approach, in which waveforms are first extracted on a subset of the recording (on a sabset of channels close to the peak), they are stacked (from `(num_spikes, num_samples, num_channels)` to `(num_spikes * num_channels, num_samples)`), and a PCA model is trained (the same model is applied to all channels). The same model is then used to reduce dimensionality to `num_pca`, so that input waveforms are transformed from `(num_spikes, num_samples, num_channels)` to `(num_spikes, num_pca, num_channels )` (with `num_pca=8`). 

In spikeinterface, we do have the same mechanism in `postprocessing.compute_principal_components(..., method="by_channel_global")`, but not as a sorting component.

### Proposed solution

Implement a `WaveformTemporalPCA` class which inherits from `PeakPipelineStep`.

The class should have a `fit()` method that looks as follows"
```
def fit(self, recording, **kwargs)
```
This function, internally, should select a portion of the recording (maybe random chunks), detect peaks and extract waveforms, compute the PCA model, and save it to `pkl` (it needs to be saved to allow for parellelization)

The `compute_buffer()` function, will apply the `model.transform()` function to the extracted waveforms in the buffer an return `tpcs_waveforms` with dimensions `(num_spikes, num_pca, num_channels )`

### ETA since in progress

2-3 weeks

### Other requirements

- [ ] Documentation
- [ ] Testing

h-mayorquin:
This shoud eventually close #1123

This is the first draft that I make here to discuss the design of this component. As I am not sure how this will hold together with all the other step I am writing something very rought here to get the discussion going.

There are two important points, the fitting (in the sklearn sense which is training) functionality and the transform functionality which is implemented as an `PeakPipelineStep` instance.

## Fitting
For this, what is needed are the following ingredients: recording -> peaks -> waveforms.
The PCA algorithm will work on the extracted waveforms as a dimensionality reduction technique. That is, for this there are two steps. The first step is a waveform extraction and the second is the model fitting.  The first step, the waveform extraction, can be easily understood as a `PeakPipelineStep` and that's how I implemented it in order to guarantee that it plays nicely with what you have done (that is, the method for finding peaks and post-peak selection can then be treated as modular steps). Maybe your design intent was to have the waveform collecting for training handled differently? What do you guys think?

To go beyond that I think that the whole `DimensionalityReduction` can be implemented as a `PeakPipelineStep`, I am concerned about how the training process will interact with your `ChunkRecordingExecutor` machinery so I will leave this out of the discussion now.

## Transform
This is the easest one as it is only an application of a loaded model. My main question here is where this should be located and what kind of tests do you expect? My plan right now was just to add only test and non-assertion-throwing tests as this is what I have seen you doing but I wonder if you have something else on mind.

Finally, some comments here and there in the code.


spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,96 @@
...
+    pipeline = Pipeline([("stack", temporal_data_stacking), 
+                        ("pca", pca_model), 
+                        ("unstack", channel_and_spike_data_separator),
+                        ])
+    
+    # TODO discuss and implement chunking
+    pipeline.fit(waveforms)
+    
+    # Save - TODO, discuss appropiate place and form
+    dump(pipeline, "pipeline.joblib")
```
h-mayorquin, line 69: Two questions here:
1) Do you use joblib for picking or the pickle library?
2) Do you use temporal naming logic  for storing this or should we allow the user to determine that?

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,96 @@
...
+
+    n_channels = recording.get_num_channels()
+    
+    
+    # TODO discuss parameters to propagate
+    peaks = detect_peaks(recording, method='by_channel',
+                            peak_sign='neg', detect_threshold=5, exclude_sweep_ms=0.1,
+                            **job_kwargs)
+
+    reduced_peaks = peaks  # TODO: We should use a peak_location method here
```
h-mayorquin, line 48: Any peak_location method to handle random selection? I think that is what @alejoe91 wanted.

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,96 @@
...
+
+def stack_channel_with_spikes(waveform):
+    n_spikes, n_samples, n_channels = waveform.shape 
+
+    return waveform.transpose(0, 2, 1).reshape(n_spikes * n_channels, n_samples)
+
+def separate_channel_from_spikes(temporal_staggered_waveform, n_components, n_channels):
+    n_spikes = int(temporal_staggered_waveform.shape[0] / n_channels)
+    reshaped = temporal_staggered_waveform.reshape(n_spikes, n_channels, n_components)
+    return reshaped.transpose(0, 2, 1)
```
alejoe91, line 36: not needed (we can use the `postprocessing` module)

spikeinterface/sortingcomponents/dimensionality_reduction.py
alejoe91: We can just use the `peaks["channel_ind"]` and look for channels in the vicinity of the main peak (see [here](https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/sortingcomponents/features_from_peaks.py#L106-L116) for an example)

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,96 @@
...
+
+def fit_temporal_pca(recording, n_components, job_kwargs, whiten=True):
```
alejoe91, line 38: IMO, we can make full use of the `postprocessing` module here and this should be a `self` function:

```
from spikeinterface.postprocessing import compute_principal_components, get_template_channel_sparsity
from spikeinterface.sortingcomponents.peak_selection import select_peaks

peaks = detect_peaks(recording, method='by_channel',
                                       peak_sign='neg', detect_threshold=5, exclude_sweep_ms=0.1,
                                      **job_kwargs)
# we can either subsample peaks, or cut some shorter recording segments from the recording (let's say we go for the second option)
sub_peaks = select_peaks(peaks) 
# see https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/sortingcomponents/peak_selection.py#L8

# make a sorting object out of it:
sorting = si.NumpySorting.from_peaks(peaks, sampling_frequency=recording.sampling_frequency) 
# see https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/core/numpyextractors.py#L209

# extract waveforms fast and in-memory (provided that we don't have na enormous number of peaks)
we = si.extract_waveforms(recording, sorting, folder=None, mode="memory", **job_kwargs)
# in this case unit_ids are channel_ids

# compute sparsity based on local_radius_um (should be an input)
sparsity = get_template_channel_sparsity(we, method="radius", radius_um=local_radius_um)

# compute PCA by_channel_global (with sparsity)
pc = compute_principal_components(we, n_components=n_components, mode='by_channel_global', sparsity=sparsity)

pca_model = pc.get_pca_model()

# model_folder should be an input
pca_model_path = model_folder / "pca_model.pkl"

with pca_model_path.open("wb") as f:
      pickle.dump(pca_model, f)

self._kwargs["model_path"] = model_path
```

The saved model can be used as is in the `compute_buffer()` function.

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,96 @@
...
+        self.peak_sign = peak_sign
+        self._kwargs.update(dict(all_channels=all_channels, peak_sign=peak_sign))
+        self._dtype = recording.get_dtype()
+
+    def get_dtype(self):
+        return self._dtype
+
+    def fit(self, recording, n_components, job_kwargs, whiten=True):
+        
+        pca_pipeline = fit_temporal_pca(recording, n_components, job_kwargs, whiten=whiten)
```
alejoe91, line 89: I would implement the fitting here directly, without relying on a third function

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,96 @@
...
+        
+        return pca_pipeline
+    
+    def compute_buffer(self, traces, peaks, waveforms):
+        pca_pipeline = load("pipeline.joblib")
```
alejoe91, line 94: here we need to additionally only project the sparse channels (similar to [here](https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/sortingcomponents/features_from_peaks.py#L106-L116))
and run the `pca_model.transform()` only on those

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class TemporalPCAProjection(PeakPipelineStep):
+    need_waveforms = True
+
+    def __init__(self, recording, ms_before=1., ms_after=1.,  peak_sign='neg', all_channels=True):
+        PeakPipelineStep.__init__(self, recording, ms_before=ms_before, ms_after=ms_after)
+        self.all_channels = all_channels
+        self.peak_sign = peak_sign
+        self._kwargs.update(dict(all_channels=all_channels, peak_sign=peak_sign))
```
alejoe91, line 81: In my opinion we would need to add a `model_path` kwargs that allows us to re-instantiate the PCA model from it (which is required for parallelization) and add:

```
if model_path is not None:
    with open(model_path, "rb") as f:
        self.model = pickle.load(f)
else:
    self.model = None
```

h-mayorquin: Scikit-learn advices on using jolbib instead of pickle for model persistence:

https://scikit-learn.org/stable/model_persistence.html

Why deviate from that?

alejoe91: In the principal components we have been using pickle. I would say let's stick to pickle for consistency and update this later ;)

spikeinterface/sortingcomponents/dimensionality_reduction.py
h-mayorquin: All right, I wanted to use standard scikit-learn machinery instead of the ad-hoc concatenation that you have for this method there. I will change it.

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,96 @@
...
+        self._dtype = recording.get_dtype()
+
+    def get_dtype(self):
+        return self._dtype
+
+    
+    def compute_buffer(self, traces, peaks, waveforms):
+        
+        return waveforms
```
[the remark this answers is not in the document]
h-mayorquin: I was wondeirng whether you wanted to keep in between the modules as lowly coupled as possible so I implemented a local solution.

spikeinterface/sortingcomponents/dimensionality_reduction.py
h-mayorquin: How do you the`fit` method to interact with the step here? Are you envisioning that they if there is no model previously stored in the `model_path` then there should be a call to the fit method? 

In general, are you expecting uses to interact with the fit method directly? How do you expect users to utilize the PCA step?

h-mayorquin: So, something like the following:
```python
instance = TemporalPCAProjection(**kwargs)
instance.fit(**some_other_kwargs)
instance.compute_buffer()
```

spikeinterface/sortingcomponents/dimensionality_reduction.py
h-mayorquin: I have something similar now in the last commit.

spikeinterface/sortingcomponents/dimensionality_reduction.py
h-mayorquin: OK, changed in the last commit.

spikeinterface/sortingcomponents/dimensionality_reduction.py
h-mayorquin: Ok, I added a test looking like this in the last commit. While I was working on it, I was wondering, why not `run_peak_pipeline` instead of `detect_peaks`?

spikeinterface/sortingcomponents/dimensionality_reduction.py
h-mayorquin: I have implemented this now. I have a question:

In the issue it is mentioned that the expected output of the projected waveforms should be `(num_spikes, num_pca, num_channels )`, on the other hand, we are projecting here only the sparse channels which means that the outermost , `num_channels`, is going to be have a lot of zeros into it if they are sparse enough. This how this looks in the current implementation.

The question is, should we return  instead something like (`num_spikes, num_channels, num_channels_per_waveform)` to save space?  Where the `num_channels_per_waveform` is the number of effective channels per waveform, that is, the ones that are preserved by the sparsity.   And if so, does this even have a uniform block structure? That is, is the number of channels assigned to each unit constant? Certaintly, it is with the current assumptions: calculating sparsity with using a ratius and assuming that most recording devices have a uniform spatial distribution of their electrodes. But, are those assumptions valid for other sparsity methods or maybe unforseen geometry of recording devices?  The second best would be to return a dict for every wafeform but that would make all the subsequent operations less amenable to vectorization which sounds undesirable.

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,79 @@
...
+        # compute PCA by_channel_global (with sparsity)
+        sparsity = get_template_channel_sparsity(we, method="radius", radius_um=self.local_radius_um)
+        pc = compute_principal_components(we, n_components=n_components,  mode='by_channel_global',
+                                        sparsity=sparsity, whiten=whiten)
+        pca_model  = pc.get_pca_model()
+
+        # model_folder should be an input
+        pca_model_path = self._kwargs["model_path"]
+
+        with open(pca_model_path, "wb") as f:
```
h-mayorquin, line 47: I wish there was a way of pickling the whole `WaveformPrincipalComponent` object. This will allow to use the machinery you have created in the post-processing model to also do the projectins (taking into the account the sparsity already given) but pickling gets me an hdfm error.

spikeinterface/sortingcomponents/dimensionality_reduction.py
alejoe91: Yes, I think it's better to return the sparse projections, but as you said different units might have more or less channels. One options could be to have arrays with the max number of sparse channels and fill in the rest with nans

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,78 @@
...
+        PeakPipelineStep.__init__(self, recording, ms_before=ms_before, ms_after=ms_after, local_radius_um=local_radius_um)
+        self.all_channels = all_channels
+        self.peak_sign = peak_sign
+        self._kwargs.update(dict(all_channels=all_channels, peak_sign=peak_sign, model_path=model_path))
+        self._dtype = recording.get_dtype()
+
+    def get_dtype(self):
+        return self._dtype
+
+    def fit(self, recording, n_components, job_kwargs, whiten=True):
```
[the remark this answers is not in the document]
h-mayorquin: Done in the last commit.

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,79 @@
...
+        
+        peaks = detect_peaks(recording, method='by_channel', peak_sign='neg', detect_threshold=5, exclude_sweep_ms=0.1, **job_kwargs)
+
+        sub_peaks = select_peaks(peaks, method="uniform", select_per_channel=False, n_peaks=1000) # How to select n_peaks
```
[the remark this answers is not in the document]
h-mayorquin: Added a heuristic of around 1000 per channel in the last commit:
`  n_peaks = recording.get_num_channels() * 1e3`

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,79 @@
...
+        self.peak_sign = peak_sign
+        self._kwargs.update(dict(all_channels=all_channels, peak_sign=peak_sign, model_path=model_path))
+        self._dtype = recording.get_dtype()
+
+    def get_dtype(self):
+        return self._dtype
+
+    def fit(self, recording, n_components, job_kwargs, whiten=True):
+        
+        peaks = detect_peaks(recording, method='by_channel', peak_sign='neg', detect_threshold=5, exclude_sweep_ms=0.1, **job_kwargs)
```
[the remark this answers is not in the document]
h-mayorquin: I don't think that you can have two ** (unpacking operator) in the definition of a function call, am I missing something?

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,79 @@
...
+
+        sub_peaks = select_peaks(peaks, method="uniform", select_per_channel=False, n_peaks=1000) # How to select n_peaks
+
+        ms_before = self._kwargs["ms_before"]
+        ms_after = self._kwargs["ms_after"]
+        sorting = NumpySorting.from_peaks(sub_peaks, sampling_frequency=recording.sampling_frequency) 
+        we = extract_waveforms(recording, sorting, ms_before=ms_before, ms_after=ms_after, folder=None, mode="memory", **job_kwargs)
```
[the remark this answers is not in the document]
h-mayorquin: Did in the last commit.

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,99 @@
...
+        # Detect peaks and sub-sample them
+        peaks = detect_peaks(recording, **detect_peaks_params, **job_kwargs)
+        peaks = select_peaks(peaks, **peak_selection_params) # How to select n_peaks
+
+        # Create a waveform extractor
+        ms_before = self._kwargs["ms_before"]
+        ms_after = self._kwargs["ms_after"]
+        sorting = NumpySorting.from_peaks(peaks, sampling_frequency=recording.sampling_frequency) 
+        we = extract_waveforms(recording, sorting, ms_before=ms_before, ms_after=ms_after, folder=None, 
+                               mode="memory", max_spikes_per_unit=None, **job_kwargs)
```
alejoe91, line 56: With the new sparsity object (#1152 ), we could either already extract sparse wavefroms as:

```suggestion
        we = extract_waveforms(recording, sorting, ms_before=ms_before, ms_after=ms_after, folder=None, 
                               mode="memory", max_spikes_per_unit=None, sparse=True, method="radius", radius_um=self.local_radius_um, **job_kwargs)
```

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,99 @@
...
+
+        # compute PCA by_channel_global (with sparsity)
+        sparsity = get_template_channel_sparsity(we, method="radius", radius_um=self.local_radius_um)
```
alejoe91, line 59: Or, we compute sparsity the new way:
```suggestion
        sparsity = ChannelSparsity.from_radius(we, radius_um=self.local_radius_um)
```

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def __init__(self, recording, ms_before=1., ms_after=1.,  peak_sign='neg', all_channels=True):
+        PeakPipelineStep.__init__(self, recording, ms_before=ms_before, ms_after=ms_after)
+        self.all_channels = all_channels
+        self.peak_sign = peak_sign
+        self._kwargs.update(dict(all_channels=all_channels, peak_sign=peak_sign))
+        self._dtype = recording.get_dtype()
+
+    def get_dtype(self):
+        return self._dtype
```
alejoe91, line 25: This could be moved to `PeakPipelineStep` to optionally return waveforms with peaks

h-mayorquin: I removed this from here and did this in another PR #1182

spikeinterface/sortingcomponents/dimensionality_reduction.py
h-mayorquin: Changed to this functionality, I think that using a proper class is better than passing keyword arguments to the waveform in the other option. This option is closer to this.

h-mayorquin:
@alejoe91 I added the functionality and a test so the projections are return in a sparse form. Here this means that the "invalid" channels are filled with `np.nan`.  As discussed, I also vectorized further the projection operation. I think there might be a way of vectorizing it further but I think this should be done once we have converged in a common way of handling sparsity across the pipeline. 

Let me know what you think and if something is missing.

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,81 @@
...
+from spikeinterface.sortingcomponents.peak_pipeline import run_peak_pipeline, PeakPipelineStep
+from spikeinterface.sortingcomponents.peak_detection import detect_peaks
+from spikeinterface.sortingcomponents.peak_selection import select_peaks
+from spikeinterface.postprocessing import compute_principal_components
+from spikeinterface.core.sparsity import ChannelSparsity
+from spikeinterface import extract_waveforms, NumpySorting
+
+
+class TemporalPCA(PeakPipelineStep):
+    need_waveforms = True
```
alejoe91, line 14: Can you add docstrings to the class too?

h-mayorquin: Done.

spikeinterface/sortingcomponents/tests/test_dimensionality_reduction.py
```diff
@@ -0,0 +1,78 @@
...
+    n_components = 3
+    job_kwargs = dict(n_jobs=1, chunk_size=10000, progress_bar=True)
+    n_peaks = recording.get_num_channels() * 1e3 # Heuristic for extracting around 1k waveforms per channel
+    peak_selection_params = dict(method="uniform", select_per_channel=True,  n_peaks=n_peaks)
+    detect_peaks_params = dict(method='by_channel', peak_sign='neg', detect_threshold=5, exclude_sweep_ms=0.1)
+    temporal_pca.fit(recording, n_components, detect_peaks_params, peak_selection_params, job_kwargs)
+
+    steps = [temporal_pca]
+    
+    peaks, projected_waveforms = detect_peaks(recording, pipeline_steps=steps)
```
alejoe91, line 31: Can you add 2 options?

1. fitting with `n_jobs=2`
2. running `detect_peaks` in parallel (`n_jobs=2`)

This makes sure that the serialization is working

h-mayorquin: I added now a new extra test case that does this two things.

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,141 @@
...
+        return self.pca_model
+    
+    def compute_buffer(self, traces, peaks, waveforms):        
+
+        if self.pca_model == None:
+            exception_string = (
+                f"Pca model not found, "
+                f"Train the pca model before running compute_buffer using {self.__name__}.fit(kwargs)"
+            )
+            raise AttributeError(exception_string)
```
samuelgarcia, line 125: This check should be done before.
Here this is call at every chunk.

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,141 @@
...
+        n_waveforms, n_samples , n_channels = waveforms.shape
+        n_components = self.pca_model.n_components
+        
+        # Initializing non-valid channels with nan (sparsity is used to indicate valid channels)
+        projected_waveforms = np.full(shape =(n_waveforms, n_components, n_channels), fill_value=np.nan)
+
+        for waveform_index, main_channel in enumerate(peaks['channel_ind']):
+            channel_indexes_sparsified,  = np.nonzero(self.neighbours_mask[main_channel])
+            sparsified_waveform = waveforms[waveform_index, :, channel_indexes_sparsified]
+            projected_waveforms[waveform_index, :, channel_indexes_sparsified] = self.pca_model.transform(sparsified_waveform)
```
samuelgarcia, line 136: here there is something I do not understand. Sorry.
With the `mode='by_channel_global'` the pca_model is supposed to be applied channel per channel.
Here this is not the case. You transform to a buffer which is (sample, channel). This don't make sens to me.
I need more explanation. Sorry.

spikeinterface/sortingcomponents/dimensionality_reduction.py
```diff
@@ -0,0 +1,141 @@
...
+                f"Pca model not found, "
+                f"Train the pca model before running compute_buffer using {self.__name__}.fit(kwargs)"
+            )
+            raise AttributeError(exception_string)
+        
+        n_waveforms, n_samples , n_channels = waveforms.shape
+        n_components = self.pca_model.n_components
+        
+        # Initializing non-valid channels with nan (sparsity is used to indicate valid channels)
+        projected_waveforms = np.full(shape =(n_waveforms, n_components, n_channels), fill_value=np.nan)
```
samuelgarcia, line 131: Here we have a very sparse 3d array.
I was thinking that we were outputing only a few channels.
With this approach the output will be very very big and 90% of nan.
Could we only have a shape with (n_waveforms, n_components, some_channels) with some_channels=max of sparsity.

alejoe91, line 131:
```suggestion
        projected_waveforms = np.full(shape =(n_waveforms, n_components, n_max_sparse_channels), fill_value=np.nan)
```

spikeinterface/sortingcomponents/dimensionality_reduction.py
alejoe91: I think @samuelgarcia  is right. We probably need `sparsified_waveforms.T`, so we input a `(num_channels, num_samples)` array

spikeinterface/sortingcomponents/dimensionality_reduction.py
h-mayorquin: I don't see where else this could be. You can not do it in the __init__ because you might call `fit` after initialization.

spikeinterface/sortingcomponents/dimensionality_reduction.py
h-mayorquin: We briefly discussed this before. With the curren structure of the waveforms if you do 
`waveforms[:, :, 5]` you will get all the waveforms for channel five and this will be the same channel for every waveform. The structure that you are proposing breaks this structure as not every wave has the same valid channels. 

I don't know if this problem can be solved with arrays, you could do it with another structure such as a dict:
`waveform_dict[wave_index][channel_index]` <- this returns all the timestamps and only has valid `channel_index` keys for active channels.

spikeinterface/sortingcomponents/dimensionality_reduction.py
alejoe91: I agree with Ramon. this only adds the `if self.pca_model == None:` in terms of computation and we need to protect against running the compute buffer without an instantiated pca model (which is possible at init if you instantiate the component for fitting!)

spikeinterface/sortingcomponents/dimensionality_reduction.py
alejoe91: I think what we agreed also with Charlie and Julien is to return the matrix with sixe of the max number of sparse channels and fill in with nan the invalid channels. The valid channels will always be at the beginning of the indices.

So for example, if we have 5 channels as max sparse channels, but one waveform i has only 3 valid channels in its sparsity, it's projections will be:
`projected_waveforms[i, 0, :] = [val1, val2, val3, nan, nan]`

alejoe91: that is true, but the `PeakPipelineStep` will know is own sparsity, so the information about which channels the values correspond to is accessible :)

h-mayorquin: I thogught about this as I was writing the comment above and I could not see it. That is, I am not sure how would you use the mask or the dictionary in sparsity to map the two different structures to their dense form.
```
projected_waveforms[i, 0, :] = [val_i1, val_i2, val_i3, nan, nan]
projected_waveforms[j, 0, :] = [val_j1, val_j2, val_j3, nan, nan]
```

I am thinking as I am writing here. Here you have the channels scrambled, the sparsity tells you which channels are nan and which are not nan but I don't see how to undo the necessary scrambling from this mapping. Maybe there is an easy way? I will think about it.

Independently of this I feel that the choice of the sparse representation for waveforms should be more general and probably a PR on its own. Personally, I think that the `ChannelSaprsity` class should have a method to change from a dense to a sparse representation (either the representation that you propose here or any other that we decide to use). The responability of sparsifying and densifying is then centralized and makes more sense semantically in that place. As you mention, then we can use the sparsity instance (that every `PeakPipelineStep` is aware off) to return sparse representations everywhere if so we decide.

spikeinterface/sortingcomponents/dimensionality_reduction.py
h-mayorquin: The model is trained to transform samples with `n_timestamps` features (where `n_waveforms, n_timestamps, n_channels = all_waveforms.shape`). What matter for a scikit-learn model is the number of features, the first dimension is the number of samples.

In the code that you have in the principal component module:
```python
        elif mode == "by_channel_global":
            shape = (new_waveforms.shape[0], p['n_components'], num_channels)
            projections = np.zeros(shape)
            for wf_ind, chan_ind in enumerate(channel_inds):
                projections[:, :, chan_ind] = pca_model.transform(new_waveforms[:, :, wf_ind])
```
You extract the samples channel by channel so you end up with a vector of size  (`n_waveforms`, `n_timestamps`), that is, `n_waveforms` samples and `n_timestamps` as features. This works. 

In the code here I extract the samples waveform by waveform so I end up with a vector of size (`n_sparse_channels`, `n_timestamps`, that is, `n_sparse_channels` as samples and `n_timestamps` as features so  it works as well.

I decide to go for the second choice because that way you don't need to load large samples in memory. That is, I did not profile this, maybe the other way is better...

h-mayorquin:
Can we move forward with this?

samuelgarcia:
I think we should rebase this with the new peakpipeline.
And also maybe make a common node between this and the denoise because this ywo nodes somehow have a model that depend on params.

Also I would create a new file but move this inside "features_from_peaks.py".
PCA is clearly a featurisation and somehow all features are somehow a dimentionality redution.
As soon as this new feature is not too big I would put it in the same file.

PR #1148, opened 2022-12-14
No problem on my laptop with sc2 tests

yger:
This PR is just a test, as Sam told me that sc2 tests are not passing on github. I want to double check, since on my laptop, everything is working fine. Could there be a version issue?


alejoe91:
Working on my local machine too

samuelgarcia:
This faulty sorter was in fact spkykingcirucs-1

PR #1150, opened 2022-12-14
Add `sorting_info` property to Sorting object

### LINKED ISSUE #1117: [feat-core] add sorting_info to `BaseSorting` object
Currently, when running a spike sorting job, the output Sorting object contains no information on the spike sorting (the Recording gets automatically registered to the sorting object, but only in-memory).

We should annotate the Sorting object with these info that are generated by the `BaseSorter` class:
- `spikeinterface_params`: includes sorter name and parameters (add version?)
- `spikeinterface_recording`: dictionary of the input recording

These info could be stored both as annotations (so that they are automatically saved and loaded), but they could also be exposed as an additional object variable `sorting.sorting_info['params']/['recording']`. Annotation names should be protected (e.g. `__sorting_info_params__`, `__sorting_info_recording__`)

alejoe91:
fixes #1117


spikeinterface/sorters/basesorter.py
```diff
@@ -290,11 +290,17 @@ def get_result_from_folder(cls, output_folder):
...
+        # register recording to Sorting object
         recording = load_extractor(output_folder / 'spikeinterface_recording.json')
         if recording is not None:
             # can be None when not dumpable
             sorting.register_recording(recording)
-        # add sorting info here
+        # set sorting info to Sorting object
+        rec_dict = json.load(open(output_folder /'spikeinterface_recording.json', 'r'))
+        params_dict = json.load(open(output_folder / 'spikeinterface_params.json'))
+        log_dict = json.load(open(output_folder / 'spikeinterface_log.json'))
```
alejoe91, line 301:
```suggestion
        params_dict = json.load(open(output_folder / 'spikeinterface_params.json', 'r'))
        log_dict = json.load(open(output_folder / 'spikeinterface_log.json', 'r'))
```

samuelgarcia (approved):

PR #1153, opened 2022-12-16
[feat-exp] Optimizing Phy-based extractors

DradeAW:
Optimizing Phy-based extractors for faster loading.

There are still 2 steps which take up ~90% of the loading time (I've labeled the steps with comments in the code).
I'm trying to think if I can find good alternatives.

I'm working on this because for a thing I'm doing, I need to load 324 phy folders, which right now takes around 11 minutes.

### DIFF
--- a/spikeinterface/extractors/phykilosortextractors.py
+++ b/spikeinterface/extractors/phykilosortextractors.py
@@ -97,7 +97,8 @@ class BasePhyKilosortSortingExtractor(BaseSorting):
             del cluster_info["id"]
 
         # update spike clusters and times values
-        spike_clusters_clean_idxs = np.isin(spike_clusters, cluster_info["cluster_id"].values)  # Slow
+        bad_clusters = [clust for clust in clust_id if clust not in cluster_info['cluster_id'].values]
+        spike_clusters_clean_idxs = ~np.isin(spike_clusters, bad_clusters)
         spike_clusters_clean = spike_clusters[spike_clusters_clean_idxs]
         spike_times_clean = spike_times[spike_clusters_clean_idxs]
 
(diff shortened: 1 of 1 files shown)


DradeAW:
The changes reduced the time taken from ~11min to ~5min for my datasets.

spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -97,7 +97,8 @@ def __init__(self, folder_path, exclude_cluster_groups=None, keep_good_only=Fals
             del cluster_info["id"]
 
         # update spike clusters and times values
-        spike_clusters_clean_idxs = np.in1d(spike_clusters, cluster_info["cluster_id"].values).nonzero()[0]
+        bad_clusters = [clust for clust in clust_id if clust not in cluster_info['cluster_id'].values]
+        spike_clusters_clean_idxs = ~np.isin(spike_clusters, bad_clusters)
```
DradeAW, line 101: Since there are usually less bad clusters than good ones, it's faster to call `np.isin` giving it the bad clusters, and then flipping the boolean array.

spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -109,17 +110,21 @@ def __init__(self, folder_path, exclude_cluster_groups=None, keep_good_only=Fals
...
                 else:
                     new_si_id = si_id
                 unit_ids[i] = new_si_id
 
-                spike_clusters_new[spike_clusters_clean == phy_id] = new_si_id
+            # Little hack to replace values in spike_clusters_clean to spike_clusters_new very efficiently.
+            from_values = cluster_info['cluster_id'].values
+            sort_idx = np.argsort(from_values)
+            idx = np.searchsorted(from_values, spike_clusters_clean, sorter=sort_idx)
+            spike_clusters_new = unit_ids[sort_idx][idx]
```
DradeAW, line 126: Hack found from https://stackoverflow.com/questions/19666626/replace-all-elements-of-python-numpy-array-that-are-greater-than-some-value

alejoe91 (approved):

PR #1176, opened 2022-12-29
Add global_job_kwargs functions

### LINKED ISSUE #1174: [feat-core] Define global job_kwargs
Since `**job_kwargs` are used extensively throughout the code-base, it would be nice to have a core function that sets the `job_kwargs` globally. Individual function would then merge additionally passed job args with global ones at run-time.

alejoe91:
closes #1174

### DIFF
--- a/spikeinterface/core/job_tools.py
+++ b/spikeinterface/core/job_tools.py
@@ -44,7 +44,9 @@ def split_job_kwargs(mixed_kwargs):
     This can be useful for some function with generic signature
     mixing specific and job kwargs.
     """
-    specific_kwargs, job_kwargs = {}, {}
+    from .globals import get_global_job_kwargs
+    job_kwargs = get_global_job_kwargs()
+    specific_kwargs = {}
     for k, v in mixed_kwargs.items():
         if k in job_keys:
             job_kwargs[k] = v
--- a/spikeinterface/comparison/studytools.py
+++ b/spikeinterface/comparison/studytools.py
@@ -20,6 +20,7 @@ import os
 import pandas as pd
 
 from spikeinterface.core import load_extractor
+from spikeinterface.core.job_tools import split_job_kwargs
 from spikeinterface.extractors import NpzSortingExtractor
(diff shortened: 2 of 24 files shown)


spikeinterface/sorters/si_based_sorters/spyking_circus2.py
```diff
@@ -43,19 +44,17 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
 
         assert HAVE_HDBSCAN, 'spykingcircus2 needs hdbscan to be installed'
     
-        params['job_kwargs']['verbose'] = verbose
-        params['job_kwargs']['progress_bar'] = verbose
-
-        if "n_jobs" in params["job_kwargs"]:
-            if params["job_kwargs"]["n_jobs"] == -1:
-                params["job_kwargs"]["n_jobs"] = os.cpu_count()
-
```
samuelgarcia, line 52: Th is not a good idea to move this to `split_job_kwargs()` because this split function is not always call befoire this.

samuelgarcia: oups sorry this is call after.

spikeinterface/core/job_tools.py
```diff
@@ -52,6 +53,10 @@ def split_job_kwargs(mixed_kwargs):
             job_kwargs[k] = v
         else:
             specific_kwargs[k] = v
+    # if n_jobs is -1, set to os.cpu_count()
+    if "n_jobs" in job_kwargs:
+        if job_kwargs["n_jobs"] == -1:
+            job_kwargs["n_jobs"] = os.cpu_count()
```
samuelgarcia, line 59: I think this is not a good idea to use this function as a checker.
If we need a checker lets make a specific function

alejoe91: I agree. Something like `fix_job_kwargs`, which merges with globals and does the -1 to cpu count operation?

samuelgarcia (approved):

PR #1177, opened 2022-12-29
[feat-qm] Implement missing metrics (amplitude median, drift metrics)

### LINKED ISSUE #1175: [feat-qm]: implement missing quality metrics
Port missing quality metrics (see [docs](https://spikeinterface.readthedocs.io/en/latest/module_qualitymetrics.html)):

- [x] Amplitude median
- [x] Cumulative/Maximum drift
- [x] Sliding refractory period

We'll hold off `noise_cutoff` for now, since it doesn't seem to be optimized for "real" amplitude distributions in uV

alejoe91:
Port missing quality metrics (see [docs](https://spikeinterface.readthedocs.io/en/latest/module_qualitymetrics.html)):

- [x] Amplitude median
- [x] Drift metrics

`noise_cutoff` and `slidingRP` implemented in #1197


alejoe91:
@samuelgarcia @jsiegle this is ready to merge

spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -488,6 +485,207 @@ def compute_amplitudes_cutoff(waveform_extractor, peak_sign='neg',
 )
 
 
+def compute_amplitudes_median(waveform_extractor, peak_sign='neg'):
```
samuelgarcia, line 488: The function name is not intuitive.
I would add the log concept in the title.

alejoe91: The return value is not log

spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -488,6 +485,207 @@ def compute_amplitudes_cutoff(waveform_extractor, peak_sign='neg',
...
+
+    # if a segment is shorter than 2 * interval_s, we set the drift metrics to -1
+    for segment_index in range(recording.get_num_segments()):
+        seg_length = recording.get_num_samples(segment_index)
+        if  seg_length <= 2 * interval_samples:
+            warnings.warn(f"Segment {segment_index} is too short to compute drift metrics "
+                          f"with interval_s of {interval_s}. It needs to be at least {2*interval_s}s long")
+            for unit_id in unit_ids:
+                maximum_drift_segs[unit_id].append(-1)
+                cumulative_drift_segs[unit_id].append(-1)
```
samuelgarcia, line 622: why not a NaN directly ?

spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -488,6 +485,207 @@ def compute_amplitudes_cutoff(waveform_extractor, peak_sign='neg',
...
+        else:
+            num_bin_edges = seg_length // interval_samples + 1
+            bins = np.arange(num_bin_edges) * interval_samples
+            spike_vector = sorting.to_spike_vector()
+
+            # compute median positions (if less than min_spikes_per_interval, median position is 0)
+            median_positions = np.nan * np.zeros((len(unit_ids), num_bin_edges - 1))
```
samuelgarcia, line 629: Here the median positions are computed at every segment.
This is an important choice!!!
If segment are consecutive (because a short pause) in thet case the median positions should global no ?

spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -488,6 +485,207 @@ def compute_amplitudes_cutoff(waveform_extractor, peak_sign='neg',
...
+            for bin_index, (start_frame, end_frame) in enumerate(zip(bins[:-1], bins[1:])):
+                i0 = np.searchsorted(spike_vector['segment_ind'], segment_index)
+                i1 = np.searchsorted(spike_vector['segment_ind'], segment_index + 1)
```
samuelgarcia, line 632: This search sorted should be done the segment loop.
Here you compute this at every period.

spikeinterface/qualitymetrics/misc_metrics.py
alejoe91: We don't have that information. I think it's better to treat each segment as separate

spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -488,6 +485,207 @@ def compute_amplitudes_cutoff(waveform_extractor, peak_sign='neg',
...
+                spike_locations_in_segment = spike_locations[i0:i1]
+                i0 = np.searchsorted(spikes_in_segment['sample_ind'], start_frame)
+                i1 = np.searchsorted(spikes_in_segment['sample_ind'], end_frame)
+                spikes_in_bin = spikes_in_segment[i0:i1]
+                spike_locations_in_bin = spike_locations_in_segment[i0:i1][direction]
+
+                for unit_ind in np.arange(len(unit_ids)):
+                    mask = spikes_in_bin['unit_ind'] == unit_ind
+                    if np.sum(mask) >= min_spikes_per_interval:
+                        median_positions[unit_ind, bin_index] = np.median(spike_locations_in_bin[mask])
```
samuelgarcia, line 643: Should we add a "direction" option to compute only in one direction ?

spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -488,6 +485,207 @@ def compute_amplitudes_cutoff(waveform_extractor, peak_sign='neg',
...
+                    # deal with nans: if more than 50% nans --> set to nan
+                    if np.sum(np.isnan(median_pos)) > min_fraction_valid_intervals * len(median_pos):
+                        max_drift = -1
+                        cum_drift = -1
+                    else:
+                        if np.isnan(median_pos[0]):
+                            # set first value to first non-nan value
+                            median_pos[0] = median_pos[~np.isnan(median_pos)][0]
+                        max_drift = np.nanmax(position_diff) - np.nanmin(position_diff)
+                        cum_drift = np.nansum(position_diff)
```
samuelgarcia, line 657: This is strange "position_diff" is not compuited in that case.
I think I do not understand.

spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -488,6 +485,207 @@ def compute_amplitudes_cutoff(waveform_extractor, peak_sign='neg',
...
+                else:
+                    position_diff = np.diff(median_pos)
```
samuelgarcia, line 659: Should this diff be absolut. So the cumulative sum  would be really cumulative. No ?

spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -49,11 +45,16 @@ def _set_params(self, metric_names=None, qm_params=None, peak_sign=None,
...
+            if self.waveform_extractor.is_extension('principal_components'):
                 # by default 'nearest_neightbor' is removed because too slow
                 pc_metrics = _possible_pc_metric_names.copy()
                 pc_metrics.remove("nearest_neighbor")
                 metric_names += pc_metrics
+            # if spike_locations are not available, drift is removed from the list
+            if not self.waveform_extractor.is_extension('spike_locations'):
+                if "drift" in metric_names:
+                    metric_names.remove("drift")
+
```
samuelgarcia, line 57: I am not super happy with this hard coded. Stuff.
I known we did it for PCA. We should find something more elegant.
We can keep it at the moment.

spikeinterface/qualitymetrics/misc_metrics.py
alejoe91: so I can do a max later...

spikeinterface/qualitymetrics/misc_metrics.py
alejoe91: right

spikeinterface/qualitymetrics/misc_metrics.py
alejoe91: we do have `direction='y'`

spikeinterface/qualitymetrics/misc_metrics.py
alejoe91: Not sure, in my opinion if a unit drifts and then goes back to the original position cumulative drift could be zero...

spikeinterface/qualitymetrics/misc_metrics.py
alejoe91: see https://github.com/alejoe91/spikeinterface/blob/missing-metrics/spikeinterface/qualitymetrics/misc_metrics.py#L638

spikeinterface/qualitymetrics/misc_metrics.py
alejoe91: But you are probably right, added abs

alejoe91:
@oliche @jsiegle 

In porting and discussing about the drift metrics, @samuelgarcia and I realized a couple of issues with max_drift and cumulative drift:
- "max_drift" is calculated with the the peak-to-peak
- "cumulative_drift" sums the absolute value of the drift signal for each interval. This makes it very sensitive to 
   the number of bins (and hence the recording duration)

For these reasons, we refactored the drift metrics to return these 3 metrics:
- The `drift_ptp` is the peak-to-peak of the drift signal for each unit. (same as "max_drift")
- The `drift_std` is the standard deviation of the drift signal for each unit.
- The `drift_mad` is the median absolute deviation of the drift signal for each unit.

The "drift_std" and "drift_mad" are measures of the dispersion of the drift signal and are insensitive to the recording duration. We think these would be better metrics to capture the drift of a unit.

Let us know what you think!

More explanation in the [docs](https://github.com/SpikeInterface/spikeinterface/blob/293a511260e234607fc10936d8a6032bcb122bea/doc/quality_metrics/drift.rst)

jsiegle:
That proposal sounds good! We've never really used cumulative drift in practice, and the issues you pointed out make it sub-optimal as a general metric.

alejoe91:
Thanks Josh! I'll fix the conflicts and we'll merge tomorrow then :)

samuelgarcia (approved):

Issue #619, opened 2022-05-05
Remove artifact function with explicit timestamps


khl02007:
`RemoveArtifactsRecording` takes as input `list_triggers`, which are in frames. On the other hand, `ms_before` and `ms_after` are in milliseconds. This is converted to frames using the sampling rate later. As a result, one cannot make use of `ms_before` and `ms_after` if one uses explicit timestamps (perhaps because there are missing samples in the recording). @alejoe91 @samuelgarcia  what do you think about changing this such that `get_times` is used to infer the conversion from ms to frames? This means the padding must be computed separately for each trigger, so the performance may take a hit. Alternatively, could make the user specify `frames_before` and `frames_after`, which are computed manually before calling remove artifact.

alejoe91:
@khl02007 was this fixed?

khl02007:
will test this out and let you know

khl02007:
oops, sorry I thought this was the one about masking out single samples. I don't think there has been a PR to address this - I had opened it to ask you guys your thoughts. what do you think - specify frames before and after, or keep the ms before and after and just infer the number of frames it corresponds to with the explicit timestamps?

alejoe91:
@khl02007 I would keep `ms_before` and `ms_after` and add optionally `frames_before` and `frames_after`. If the latter are givem, the `ms_*` params are ignored. @khl02007 does that sound ok?

khl02007:
@alejoe91 that sounds like it would work. happy new year!

alejoe91:
Happy new year to you too Kyu!!!

khl02007:
By the way, happy to make these changes and submit a PR if you are OK with that

alejoe91:
I'm super OK with that! :)

h-mayorquin:
@alejoe91 Did this idea never pan out? Does `RemoveArtifactsRecording` now have that functionality?

PR #1180, opened 2023-01-02
Add `frames_before` and `frames_after` to `RemoveArtifactsRecording`

khl02007:
@alejoe91 
#619 
How to pass in these params in 'median' and 'average' modes?

### DIFF
--- a/spikeinterface/preprocessing/remove_artifacts.py
+++ b/spikeinterface/preprocessing/remove_artifacts.py
@@ -155,7 +155,7 @@ class RemoveArtifactsRecording(BasePreprocessor):
                 if (frames_before is not None) and (frames_after is not None):
                     pad = [frames_before, frames_after]
                 else:
-                    pad = []
+                    pad = None
 
         fit_sample_interval = int(fit_sample_spacing * sf / 1000.)
         fit_sample_range = fit_sample_interval * 2 + 1
--- a/spikeinterface/preprocessing/remove_artifacts.py
+++ b/spikeinterface/preprocessing/remove_artifacts.py
@@ -23,11 +23,27 @@ class RemoveArtifactsRecording(BasePreprocessor):
     list_triggers: list of list
         One list per segment of int with the stimulation trigger frames
     ms_before: float or None
-        Time interval in ms to remove before the trigger events. 
-        If None, then also ms_after must be None and a single sample is removed
+        Time interval in ms to remove before the trigger events.
+        Either `ms_before` or `frames_before` (see below) can be given, but not both:
+            if both `ms_before` and `frames_before` are given, throws an error;
+            if both `ms_before` and `frames_before` are None, then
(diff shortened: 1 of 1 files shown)


alejoe91:
Thanks @khl02007 

Probably it will be a good idea to propagate this option to the extract waveforms. I suggest we add it to #1152 and wait for it to be merged before adding it here. Sounds good?

khl02007:
@alejoe91 yeah sounds good

spikeinterface/preprocessing/remove_artifacts.py
```diff
@@ -98,22 +116,52 @@ def __init__(self, recording, list_triggers, ms_before=0.5, ms_after=3.0, mode='
...
+            if (ms_before is not None) or (frames_before is not None):
+                raise ValueError("You must specify either `ms_after` and `frames_after`",
+                                 " (unless you want to remove single samples, in which case",
+                                 " set `ms_before` and `frames_before` to None)")
+        elif (ms_after is not None) and (frames_after is None):
             ms_after = float(ms_after)
+        elif (ms_after is None) and (frames_after is not None):
+            frames_after = int(frames_after)
+        else:
+            raise ValueError("You cannot specify both `ms_after` and `frames_after`; choose one.")
```
alejoe91, line 143: @khl02007 IMO the logic could be simpler. If `frames_before is not None`, then assert that `frames_after` is not None and use it. By default, they will be None, so if they are used they can be prioritized over the ms_before/after.

alejoe91:
@khl02007 after more thinking and discussion we thought that the problem of uneven sampling is not only here, but also, for example, whee extracting waveforms. 

Internally, even though we allow to `set_times()`, SpikeInterface always assumes that buffers are uniformly sampled. In case they are not, some steps of the processing, e.g., filtering, would be wrong.

Therefore, it is fair to leave `ms_before` and `ms_after`, which are only used to estimate the padding assuming samples are uniform (just dividine by sampling rate). These are in our opinion better arguments because they are more intuitive and robust agains sampling rate.

So basically, we would keep the arguments as they are, but extend the docstring with this additional piece of information.

Do you agree?

(Sorry for the dealy in replying to this!!!)

khl02007:
@alejoe91 I suppose this means spikeinterface as a rule won't support any processing of recordings with uneven sampling. Is that right? I agree that this is a challenge. What do you suggest that we do in general? Interpolate the 'holes' before any processing steps?

alejoe91:
Hi @khl02007 

Indeed several processing steps assume uniform sampling. In case of small gaps (say a few tens of samples), I don't think this is a problem. In case there are larger gaps, the correct way to handle it is to split the recording into multiple pseudo-continuous segments. That way each continuous block would be processed correctly. This approach should also work for the `remove_artifacts` function!

khl02007:
@alejoe91 yes OK to close

PR #1182, opened 2023-01-05
Add return waveform functionality to peak pipeline

h-mayorquin:
Added functionality so the peak pipeline in the sorting components module can return the waveforms. That is, a keyword argument is used so besides the operations indicated on each of the steps the waveforms are always returned (always at the beginning). 

While I was working on #1145 I found myself on need of something like this both for testing and debugging process and I think it might be useful in general.

I added a simple test to PR as well.

### DIFF
--- a/spikeinterface/sortingcomponents/peak_pipeline.py
+++ b/spikeinterface/sortingcomponents/peak_pipeline.py
@@ -78,7 +78,7 @@ class PeakPipelineStep:
         raise NotImplementedError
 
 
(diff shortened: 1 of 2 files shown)


samuelgarcia:
@alejoe91 : this is not a good solution I think.

We had discusiion with @h-mayorquin.

The solution of having a class returning this is better.
Lets discuss this later.
**Please do NOT merge this!!!!!**

alejoe91:
@h-mayorquin @samuelgarcia closing this in favor of the Step approach

PR #1183, opened 2023-01-05
[feat-sc] Refactor motion_estimate

samuelgarcia:
* [x] Refactor motion_estimate using class
* [X] Refactor test 
* [x] set better default params
* [x] docstring for params
* [x] unify params across methods


samuelgarcia:
@alejoe91 : ready.

PR #1196, opened 2023-01-11
[bug-postprocessing] Fix template metrics for sparse waveforms

alejoe91:
@samuelgarcia somehow this was not caught by our tests.

Basically in the template metrics we need to retrieve all templates, otherwise it's more difficult to estimate the correct max channel index

### DIFF
--- a/spikeinterface/postprocessing/template_metrics.py
+++ b/spikeinterface/postprocessing/template_metrics.py
@@ -80,12 +80,14 @@ class TemplateMetricsCalculator(BaseWaveformExtractorExtension):
             template_metrics = pd.DataFrame(
                 index=multi_index, columns=metric_names)
 
+        all_templates = self.waveform_extractor.get_all_templates()
         for unit_id in unit_ids:
-            template_all_chans = self.waveform_extractor.get_template(unit_id)
+            unit_index = self.waveform_extractor.sorting.id_to_index(unit_id)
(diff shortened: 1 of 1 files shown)


spikeinterface/postprocessing/template_metrics.py
```diff
@@ -80,12 +80,14 @@ def _run(self):
             template_metrics = pd.DataFrame(
                 index=multi_index, columns=metric_names)
 
+        all_templates = self.waveform_extractor.get_all_templates()
         for unit_id in unit_ids:
```
samuelgarcia, line 84:
```suggestion
        for unit_ind, unit_id in enumerate(unit_ids):
```

spikeinterface/postprocessing/template_metrics.py
```diff
@@ -80,12 +80,14 @@ def _run(self):
...
-            template_all_chans = self.waveform_extractor.get_template(unit_id)
+            unit_index = self.waveform_extractor.sorting.id_to_index(unit_id)
```
samuelgarcia, line 85: with the enumerate we can remove this

alejoe91:
Done camarade! I was not sure that `unit_ids` were all units, but they are! In case they could have been specified externally, we would have needed to get the correct index.

Ready to merge!

samuelgarcia (approved):

PR #1197, opened 2023-01-11
Add sliding RP violations IBL metric

jsiegle:
These changes add two key quality metrics from the IBL (noise cutoff and sliding RP violations). I also refactored some of the other quality metrics (and the correlogram calculation) so they can be used independently from the SpikeInterface objects (e.g. if you just have access to spike times and amplitudes).

The metrics computed by the SpikeInterface implementation are generally consistent with the outputs saved in the [IBL Brain Wide Map](https://int-brain-lab.github.io/iblenv/notebooks_external/data_release_brainwidemap.html), but they don't always match perfectly. I will post some figures shortly.


alejoe91:
**IMPORTANT** there was quite a bit of inconsistency in function/parameter names.

Unified in the following way:
- function names: `compute_{metric_name}s`, e.g. `compute_presence_ratios`, `compute_isi_violations`, `compute_sliding_rp_violations`, `compute_amplitude_cutoffs`
- parameter key: `{metric_name}`, e.g. `isi_violation`, `amplitude_cutoff`

alejoe91:
We removed `noise_cutoff` for now because it gives weird results on the amplitude distributions from SpikeInterface (it was designed on the amplitudes from KS, which have different distributions somehow)

alejoe91 (approved):

PR #1199, opened 2023-01-13
[fix-postprocessing/qm] Quality metrics and postprocessing fixes for sparsity and global job_kwargs

alejoe91:
This PR solves several issues in the postprocessing and quality metrics modules.

### General fixes

- `set_global_job_kwargs()` updates existing global_kwargs with new entries
- sparsity by radius can work with waveform_extractor without recording
- propagated `compute_sparsity` instead of `ChannelSparsity.from_...` for better readability
- the `compute_principal_components` and `compute_metrics` can also use a subset of `job_kwargs`.

### Fixes to quality metrics

- changed the construction of the `pca_metrics_one_unit()` to already pass unit-specific pca scores and labels (including the target unit and its neighbor) and avoid reloading several times the pca files with calls to the `get_all_projections()` function.
- the PCA metrics have been refactored to use the new sparsity approach (either sparse waveforms, or an external `ChannelSparsity` object). In particular, the `nn_isolation` is computed by taking the intersection of the sparsity of different units. Only units with at least 50% shared channels (by default), are considered overlapping and are used to compute nn_isolation.
- added **MISSING** tests for `nearest_neihgbor` (`nn_hit_rate`, `nn_miss_rate`), `nn_isolation`, and `nn_noise_overlap`
- remove `nn_isolation` and `nn_noise_overlap`from default metrics because they can be very slow, since they fit a PCA model to every pair of neighbor units. **NOTE**: the computation becomes much more efficient with sparsity ;)

### DIFF
--- a/spikeinterface/postprocessing/principal_component.py
+++ b/spikeinterface/postprocessing/principal_component.py
@@ -127,8 +127,6 @@ class WaveformPrincipalComponent(BaseWaveformExtractorExtension):
         all_projections: np.array
             The PCA projections (num_all_waveforms, num_components, num_channels)
         """
-        recording = self.waveform_extractor.recording
-
         if unit_ids is None:
             unit_ids = self.waveform_extractor.sorting.unit_ids
 
@@ -137,7 +135,7 @@ class WaveformPrincipalComponent(BaseWaveformExtractorExtension):
         for unit_index, unit_id in enumerate(unit_ids):
             proj = self.get_projections(unit_id)
             if channel_ids is not None:
-                chan_inds = recording.ids_to_indices(channel_ids)
+                chan_inds = self.waveform_extractor.channel_ids_to_indices(channel_ids)
                 proj = proj[:, :, chan_inds]
             n = proj.shape[0]
             if outputs == 'id':
@@ -231,8 +229,8 @@ class WaveformPrincipalComponent(BaseWaveformExtractorExtension):
 
         # update job_kwargs with global ones
         job_kwargs = fix_job_kwargs(job_kwargs)
-        n_jobs = job_kwargs['n_jobs'] if 'n_jobs' in job_kwargs else 1
-        progress_bar = job_kwargs['progress_bar'] if 'progress_bar' in job_kwargs else False
+        n_jobs = job_kwargs['n_jobs']
+        progress_bar = job_kwargs['progress_bar']
 
         # prepare memmap files with npy
         projection_objects = {}
--- a/spikeinterface/postprocessing/principal_component.py
+++ b/spikeinterface/postprocessing/principal_component.py
@@ -686,7 +686,9 @@ def compute_principal_components(waveform_extractor, load_if_exists=False,
         pc = WaveformPrincipalComponent.create(waveform_extractor)
         pc.set_params(n_components=n_components, mode=mode, whiten=whiten, dtype=dtype,
                       sparsity=sparsity)
-        pc.run(n_jobs=n_jobs, progress_bar=progress_bar)
+        # update job_kwargs with global ones
+        job_kwargs = fix_job_kwargs(dict(n_jobs=n_jobs, progress_bar=progress_bar))
+        pc.run(n_jobs=job_kwargs['n_jobs'], progress_bar=job_kwargs['progress_bar'])
 
     return pc
 
--- a/spikeinterface/postprocessing/principal_component.py
+++ b/spikeinterface/postprocessing/principal_component.py
@@ -376,6 +376,8 @@ class WaveformPrincipalComponent(BaseWaveformExtractorExtension):
 
         for unit_ind, unit_id in units_loop:
             wfs, channel_inds = self._get_sparse_waveforms(unit_id)
+            if len(wfs) < p['n_components']:
+                continue
             if n_jobs in (0, 1):
                 for wf_ind, chan_ind in enumerate(channel_inds):
                     pca = pca_models[chan_ind]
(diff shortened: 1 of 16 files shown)


spikeinterface/postprocessing/principal_component.py
```diff
@@ -229,6 +229,11 @@ def _run(self, n_jobs=1, progress_bar=False):
         we = self.waveform_extractor
         num_chans = we.get_num_channels()
 
+        # update job_kwargs with global ones
+        job_kwargs = fix_job_kwargs(job_kwargs)
+        n_jobs = job_kwargs['n_jobs'] if 'n_jobs' in job_kwargs else 1
```
samuelgarcia, line 234: With the fix n_jobs is always in job_kwargs no ?
I would do `n_jobs = job_kwargs['n_jobs']`

spikeinterface/postprocessing/principal_component.py
```diff
@@ -229,6 +229,11 @@ def _run(self, n_jobs=1, progress_bar=False):
...
+        progress_bar = job_kwargs['progress_bar'] if 'progress_bar' in job_kwargs else False
```
samuelgarcia, line 235: Same here.

spikeinterface/postprocessing/principal_component.py
```diff
@@ -444,10 +451,10 @@ def _fit_by_channel_global(self, progress_bar):
         # with 'by_channel_global' we can't parallelize over channels
         for unit_ind, unit_id in units_loop:
             wfs, _ = self._get_sparse_waveforms(unit_id)
-            if wfs.size == 0:
+            shape = wfs.shape
+            if shape[0] * shape[2] < p['n_components']:
```
samuelgarcia, line 455: good catch!

spikeinterface/postprocessing/unit_localization.py
```diff
@@ -285,7 +284,7 @@ def compute_monopolar_triangulation(waveform_extractor, optimizer='minimize_with
     return unit_location
 
 
-def compute_center_of_mass(waveform_extractor, peak_sign='neg', num_channels=10):
+def compute_center_of_mass(waveform_extractor, peak_sign='neg', radius_um=50):
```
samuelgarcia, line 287: excelent!

samuelgarcia: do we have the same parameters for pike localization ?

spikeinterface/qualitymetrics/pca_metrics.py
```diff
@@ -28,8 +27,7 @@
 _default_params = dict(
     nearest_neighbor=dict(
         max_spikes=10000,
-        min_spikes=10,
-        n_neighbors=4,
+        n_neighbors=5,
```
samuelgarcia, line 30: why this 4 to 5 by default ?

spikeinterface/qualitymetrics/pca_metrics.py
```diff
@@ -397,46 +408,52 @@ def nearest_neighbors_isolation(waveform_extractor: si.WaveformExtractor, this_u
...
         # find units whose signal channels (i.e. channels inside some radius around
         # the channel with largest amplitude) overlap with signal channels of the target unit
-        closest_chans_all = get_template_channel_sparsity(waveform_extractor, method='radius',
-                                                          outputs='index', peak_sign='both',
-                                                          radius_um=radius_um)
-        closest_chans_target_unit = closest_chans_all[this_unit_id]
+        if waveform_extractor.is_sparse():
+            sparsity = waveform_extractor.sparsity
+        else:
+            sparsity = compute_sparsity(waveform_extractor, method='radius', peak_sign='both',
```
samuelgarcia, line 419: do we have the peak_sign somewher to propagate here ?

samuelgarcia:
This is globally cool.
The change in pca_metrics.py is big and hard to follow.
I would add a test to compute pca metrics with dense and non dense and check that we have almost the same.
Maybe this is already done.

samuelgarcia:
Test are done in fact.

spikeinterface/qualitymetrics/pca_metrics.py
alejoe91: It was 5 originally and we accidentally changed it

spikeinterface/postprocessing/unit_localization.py
alejoe91: I'll check

spikeinterface/postprocessing/template_metrics.py
```diff
@@ -70,12 +70,12 @@ def _run(self):
...
-            channel_ids = []
+            index_unit_ids = []
+            index_channel_ids = []
             for unit_id, sparse_channels in extremum_channels_ids.items():
-                unit_ids += [unit_id] * len(sparse_channels)
-                channel_ids += list(sparse_channels)
-            multi_index = pd.MultiIndex.from_tuples(list(zip(unit_ids, channel_ids)),
+                index_unit_ids += [unit_id] * len(sparse_channels)
+                index_channel_ids += list(sparse_channels)
+            multi_index = pd.MultiIndex.from_tuples(list(zip(index_unit_ids, index_channel_ids)),
```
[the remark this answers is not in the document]
alejoe91: Yes, because it was overwriting `unit_ids`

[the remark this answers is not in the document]
samuelgarcia: oups ok.

spikeinterface/postprocessing/unit_localization.py
alejoe91: The method_kwargs propagation is a bit obscure in `spike_localization`, since it calls the `localize_peaks`. In that case the arg should be `local_radius_um`. We have to refactor the classes so that we can expose the method kwargs automatically

samuelgarcia: ok

samuelgarcia (approved):

Issue #1124, opened 2022-11-30
[infrastructure] Make `BaseWaveformTransformer` class


alejoe91:
### Current issue

Currently, the [`PeakPipeline`](https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/sortingcomponents/peak_pipeline.py) mechanism uses enables to detect peaks, extract waveforms on-the-fly, and do a series of computations on the waveforms (with `PeakPipelineSteps`).

However, extracted waveforms could be further processed before making the computations (e.g. cleaning)

### Proposed solution

We propose to make a `BaseWaveformTransformer` class with a `transform()` method (and optionally a `fit()`) method. During peak detection, a list of `WaveformTransformer` instances can be passed to the `detect_peaks()` function, so that extracted waveforms are processed on-the-fly before further processing.

The `WaveformDenoiser` will inherit from this base class.

### ETA since in progress

2-3 weeks

### Other requirements

- [ ] Documentation
- [ ] Testing

samuelgarcia:
Note that we need something a bit more flexible because what we also need is to be able to clean the waveform inside a peak detection method itself because we want to itertivelyy detect on the residual after a "clean".

alejoe91:
@samuelgarcia as we discussed the peak detection method could optionally return waveforms, no? So the iterative subtraction + cleaning will be a specific PeakDetection method

samuelgarcia:
Yes it is what I mean this futur brand new peak detection method will also need this waveform cleaner. So it will no be used only as a step in the chain but also (and mainly) inside a method. So a functional prroach will be certainly be needed.

h-mayorquin:
@alejoe91 @samuelgarcia should we close this one?

PR #1203, opened 2023-01-17
[Draft] Implement `WaveformTransformer` as a base class for all the wave transformations.

h-mayorquin:
Should fix [#1124](https://github.com/SpikeInterface/spikeinterface/issues/1124). 

This PR implements a WaveformTransformer base class and a specific instance `MaxValueNormalization` of the former for testing and illustration purposes. Most of this PR focuses on integrating this with the peak pipeline so it works well with the serialization for parallel process of of the pipeline.

As discussed before, this is a work in progress s and we are waiting for @samuelgarcia to give it a try to its graph computation approach. Meanwhile, I am pushing this to acess public feedback from @alejoe91 and as a base for working on the WaveformDenoiser.


samuelgarcia:
@h-mayorquin : Since there is a concensus on #1209 I think we can close this no ?

h-mayorquin:
@samuelgarcia I would like to delay the closing of this until I rebase the same concept to #1209.

samuelgarcia:
Ok sure.

PR #1204, opened 2023-01-17
[feat-extractors] Streaming IBL recordings

CodyCBakerPhD:
Introduces a Recording extractor for streaming data from the IBL release via ONE/brainbox.

Sometimes more convenient that downloading the data in its entirety as in the CBin IBL extractor.

Source documentation: https://int-brain-lab.github.io/iblenv/notebooks_external/loading_raw_ephys_data.html#Loading

Will make some comments in the code

@alejoe91 was there anything else you wanted me to add to the tests?


spikeinterface/extractors/stream_ibl.py
```diff
@@ -0,0 +1,204 @@
...
+
+        return stream_names
+
+    def __init__(
+        self,
+        session: str,
+        stream_name: str,
+        load_sync_channel: bool = False,
+        cache_folder: Optional[Union[Path, str]] = None,
+        remove_cached: bool = True,
```
CodyCBakerPhD, line 103: Note their default for this is `False`, but I find that potentially unsafe to not inform the user of

Otherwise it could quietly fill up disk space in whatever the default cache location is without the individual knowing

spikeinterface/extractors/stream_ibl.py
```diff
@@ -0,0 +1,204 @@
...
+            if ".lf" in raw_suffixes_by_probe:
+                stream_names.append(probe_label + ".lf")
+
+        return stream_names
+
+    def __init__(
+        self,
+        session: str,
+        stream_name: str,
+        load_sync_channel: bool = False,
```
CodyCBakerPhD, line 101: [A note from the IBL docs](https://int-brain-lab.github.io/iblenv/notebooks_external/loading_raw_ephys_data.html#Option-1:-Stream-snippets-of-raw-ephys-data) indicates this might not be exactly the same in behavior as it is for the SpikeGlxRecordingExtractor (where it's just that one extra single channel from the `ap.bin` or `lf.bin` file), they seem to indicate it could be multiple AUX channels leading into a NIDQ/DAQ but I have yet to find example files where this is the case

CodyCBakerPhD, line 101: [A note from the IBL docs](https://int-brain-lab.github.io/iblenv/notebooks_external/loading_raw_ephys_data.html#Option-1:-Stream-snippets-of-raw-ephys-data) indicates this might not be exactly the same in behavior as it is for the SpikeGlxRecordingExtractor (where it's just that one extra single channel from the ap.bin or lf.bin file), they seem to indicate it could be multiple AUX channels leading into a NIDQ/DAQ but I have yet to find example files where this is the case

spikeinterface/extractors/stream_ibl.py
```diff
@@ -0,0 +1,204 @@
...
+        dtype = file_streamer.dtype
+        BaseRecording.__init__(self, channel_ids=channel_ids, sampling_frequency=sampling_frequency, dtype=dtype)
+        self.extra_requirements.append("ONE-api")
+        self.extra_requirements.append("ibllib")
+        self.set_channel_gains(gains)
+        self.set_channel_offsets(offsets)
+
+        # set probe
+        probe = pi.read_spikeglx(meta_file)
+        self.set_probe(probe, in_place=True)
```
samuelgarcia, line 149: I load_sync_channel is True this ìnplace` is not possible because contact size will not the channel size. (this is a requirement for the `inplace`)
We need to protect this with 
```suggestion
        if not load_sync_channels:
            probe = pi.read_spikeglx(meta_file)
            self.set_probe(probe, in_place=True)
```

samuelgarcia:
Hi Cody.
This is super super cool.
I will make some test soon to play with IBL datasets.

spikeinterface/extractors/stream_ibl.py
CodyCBakerPhD: Sure, would you need it [here](https://github.com/catalystneuro/spikeinterface/blob/streaming_ibl_extractor/spikeinterface/extractors/cbin_ibl.py#L86-L87) in the CBin code too then? The meta files ought to be exactly the same

CodyCBakerPhD: I notice the core SpikeGLX framework also does something different as well: https://github.com/catalystneuro/spikeinterface/blob/streaming_ibl_extractor/spikeinterface/extractors/neoextractors/spikeglx.py#L69-L73

spikeinterface/extractors/stream_ibl.py
```diff
@@ -0,0 +1,204 @@
...
+        meta_file = file_streamer.file_meta_data  # streamer downloads uncompressed metadata files on init
+        meta = read_meta_file(meta_file)
+        info = extract_stream_info(meta_file, meta)
+        channel_ids = info["channel_names"]
+        gains = info["channel_gains"]
+        offsets = info["channel_offsets"]
+        if not load_sync_channel:
+            channel_ids = channel_ids[:-1]
+            gains = gains[:-1]
+            offsets = offsets[:-1]
```
weiglszonja, line 136: might not be necessary, not sure more than one sync channels can be expected, but `file_streamer.nsync` would return the number of sync channels so that could be used here. 
```suggestion
            num_sync_channels = file_streamer.nsync
            channel_ids = channel_ids[:-num_sync_channels]
            gains = gains[:-num_sync_channels]
            offsets = offsets[:-num_sync_channels]
```

spikeinterface/extractors/stream_ibl.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def get_traces(self, start_frame, end_frame, channel_indices):
+        if start_frame is None:
+            start_frame = 0
+        if end_frame is None:
+            end_frame = self.get_num_samples()
+
+        traces = self._file_streamer[start_frame:end_frame]
+        if not self._load_sync_channel:
+            traces = traces[:, :-1]
```
weiglszonja, line 199: same comment could be applied here with `file_streamer.nsync`

spikeinterface/extractors/stream_ibl.py
CodyCBakerPhD: I examined 60+ sessions and the value was never greater than 1

CodyCBakerPhD: I suspect this streamer does not load data from the NIDQ and we're instead supposed to use the other loaders for that

spikeinterface/extractors/neoextractors/spikeglx.py
```diff
@@ -70,7 +70,6 @@ def __init__(self, folder_path, load_sync_channel=False, stream_id=None, stream_
                 self.set_probe(probe, in_place=True, group_mode="by_shank")
             else:
                 self.set_probe(probe, in_place=True)
-            self.set_probe(probe, in_place=True)
```
CodyCBakerPhD, line 73: This line seemed redundant? Y/N?

spikeinterface/extractors/stream_ibl.py
CodyCBakerPhD: OK added it here and to the CBIN as well

spikeinterface/core/baserecordingsnippets.py
```diff
@@ -206,7 +206,7 @@ def get_probegroup(self):
             positions = self.get_property('location')
             if positions is None:
                 raise ValueError(
-                    'There is not Probe attached to recording. use set_probe(...)')
+                    'There is no Probe attached to this recording. Use set_probe(...) to attach one.')
```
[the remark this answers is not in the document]
alejoe91: Thanks!!!

spikeinterface/extractors/neoextractors/spikeglx.py
alejoe91: Y

spikeinterface/extractors/stream_ibl.py
```diff
@@ -0,0 +1,225 @@
...
+    def get_num_samples(self):
+        return self._file_streamer.ns
+
+    def get_traces(self, start_frame, end_frame, channel_indices):
+        if start_frame is None:
+            start_frame = 0
+        if end_frame is None:
+            end_frame = self.get_num_samples()
+
+        traces = self._file_streamer[start_frame:end_frame]
```
alejoe91, line 218: @CodyCBakerPhD looking at some traces I suspect that they are already in Volts. 

There are 2 options here:

1. we convert back to int16 by doing `traces * 1e6 / gain_to_uV`
2. we convert to uV and drop the `gain_to_uV` (otherwise we would convert twice). We can still keep the gain property as `original_gain`

alejoe91: I vote for option 1, so we don't do additional operations on the fly. Note that the traces will always be in uV in this case.

alejoe91: I pushed option 1 :)

alejoe91: and opened an issue on the ibl-lib side: https://github.com/int-brain-lab/ibllib/issues/560

CodyCBakerPhD: @alejoe91 It looks as though you pushed option 2 (keeping as `float32`, just adjusting from V to uV)

So I'm intending to write this data to NWB

A quick test comparing writing the data as `int16` vs. `float32` shows about a 33% increase in disk space when writing as `float32` (compression enabled in both cases; higher in the `float32` case but not 2x as good as `int16`)

I'm planning to write an estimated 62 TB worth of this to start with, so I'd rather that not take up more space than needed

It also looks like the chunk that gets downloaded on the fly to the cache is in the original `int16` type, so it's just the API that scales automatically

Could we maybe add a `return_scaled` option to this `get_traces`, and if `True` do what you've implemented, and if `False` (which I would use) actually do option 1, cast and return as `int16`?

alejoe91: Yep sorry I meant option 2. 

Ok for me to do option 1 and option 2 (in uV) when `return_scaled` is True. The optimal solution would be that the `Streamer` gets updated accordingly!

CodyCBakerPhD: > Ok for me to do option 1 and option 2 (in uV) when return_scaled is True. 

Done


> The optimal solution would be that the Streamer gets updated accordingly!

Agreed

CodyCBakerPhD: I also renamed the temporary property to make it clearer what it's supposed to do. If they adjust the return behavior then we can return the class gain/offset properties to normal SpikeInterface behavior

spikeinterface/extractors/tests/test_stream_ibl.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+if __name__ == '__main__':
+    TestDefaultStreamingIblExtractorApBand.setUpClass()
+    test1 = TestDefaultStreamingIblExtractorApBand()
+    test1.setUp()
+    test1.test_get_stream_names()
+    test1.test_representation()
+    test1.test_dtype()
+    test1.test_channel_ids()
```
CodyCBakerPhD, line 165: Also, is there a reason to avoid using `unittest.main()` for things like this?

alejoe91: mainly to allow to run a subset of tests when calling the main function. Also if tests fail in this case things are much easier to debug than running through unittest (you can use the debugger and navigate the stack)

spikeinterface/extractors/stream_ibl.py
```diff
@@ -216,15 +225,18 @@ def __init__(self, file_streamer, load_sync_channel: bool = False):
     def get_num_samples(self):
         return self._file_streamer.ns
 
-    def get_traces(self, start_frame, end_frame, channel_indices):
+    def get_traces(self, start_frame: int, end_frame: int, channel_indices, return_scaled: bool = False):
```
alejoe91, line 228: @CodyCBakerPhD please remove the `return_scaled` argument.

This is handled at the `BaseRecording` level, which automatically applies the `gain_to_uV` and   `offset_to_uV`.
So, for now, this function should just return the `int16` values, and we should set gains and offsets with `channel_gains` and `channel_offsets`.

CodyCBakerPhD: So, just to confirm, what you're suggesting is to

- revert https://github.com/SpikeInterface/spikeinterface/pull/1204/files#diff-5eb6b5a9283986dd8203d4f61aac997f4c2ed6cb0e3fdf2250d87a9dae97fd49R158-R159 back to actually setting the true gains as `gain_to_uV`
- in the segment `get_traces`, use those as I am now to undo the Volt scaling so that what gets returned by each call is the `int16`
- so that if someone wants to go back to microVolts in SI they'd request `return_scaled` to re-apply what the segment initially unscaled?

alejoe91: exactly. At leas until we are able to return int16 directly from the streamer

alejoe91: I know!!! ehehehe

spikeinterface/extractors/stream_ibl.py
alejoe91: Then it probably makes sense to call it `load_sync_channels` and to adopt @weiglszonja 's suggestion of using `file_streamer.nsync`. What do you think @CodyCBakerPhD ?

spikeinterface/extractors/stream_ibl.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def get_traces(self, start_frame: int, end_frame: int, channel_indices):
+        if start_frame is None:
+            start_frame = 0
+        if end_frame is None:
+            end_frame = self.get_num_samples()
+        if channel_indices is None:
+            channel_indices = slice(None)
+
+        traces = np.rint(self._file_streamer[start_frame:end_frame] / self._file_streamer.sample2volts).astype("int16")
```
CodyCBakerPhD, line 232: @alejoe91 I also just noticed some casting errors without calling some sort of rounding before `int16` casting

image

alejoe91: Sounds good! Thanks for catching it

spikeinterface/extractors/stream_ibl.py
CodyCBakerPhD: See my other notes on the matter - despite that note I left here about _their_ documentation, it appears as if the argument always follows the same behavior as it does on every other SpikeGlxExtractor, where there is only a single extra channel for each band

CodyCBakerPhD: I'm running a holistic scan to confirm then we can resolve the conversations

```python
from spikeinterface.extractors import StreamingIblExtractor
from tqdm import tqdm_notebook
from one.api import ONE

one = ONE(base_url='https://openalyx.internationalbrainlab.org', password='international', silent=True)

all_sessions = list(one.alyx.rest(url="sessions", action="list"))

all_sessions_nsync = dict()
for session in tqdm_notebook(all_sessions):
    session_id = session["id"]
    for stream_name in StreamingIblExtractor.get_stream_names(session=session_id):
        recording = StreamingIblExtractor(session=session_id, stream_name=stream_name)
        all_sessions_nsync.update({session_id: recording._file_streamer.nsync})
```

ETA 5 hours

alejoe91: :) nice!

CodyCBakerPhD: Well, that ETA (according to tqdm) is probably off by quite a bit, its starting with the sessions that have ecephys but only a small percentage of the total number of sessions have any so we'll see

CodyCBakerPhD: Yeah, only took like 15 minutes.

Confirmed the usage of `nsync`

```python
set(all_sessions_nsync.values())
>>> {1}
```

alejoe91: ok then! Then let's merge as is. We can always make changes later if things change

alejoe91 (approved):

spikeinterface/extractors/stream_ibl.py
```diff
@@ -0,0 +1,239 @@
...
+
+            each returned value in `sessions` refers to it as the 'id'.
+
+        Returns
+        -------
+        stream_names : list of str
+            List of stream names as expected by the `stream_name` argument for the class initialization.
+        """
+        assert HAVE_BRAINBOX_ONE, cls.installation_mesg
+        one = ONE(base_url="https://openalyx.internationalbrainlab.org", password="international", silent=True, cache_dir=cache_folder)
```
CodyCBakerPhD, line 89: One small change that I propagated after your last approval - I noticed it was storing session info in the default cache dir even when using a specified cache_folder

spikeinterface/extractors/stream_ibl.py
```diff
@@ -0,0 +1,239 @@
...
+            The session ID to extract recordings for.
+            In ONE, this is sometimes referred to as the 'eid'.
+            When doing a session lookup such as
+
+            >>> from one.api import ONE
+            >>> one = ONE(base_url="https://openalyx.internationalbrainlab.org", password="international", silent=True)
+            >>> sessions = one.alyx.rest('sessions', 'list', tag='2022_Q2_IBL_et_al_RepeatedSite')
+
+            each returned value in `sessions` refers to it as the 'id'.
+
```
samuelgarcia, line 82: Here I would add some documentaion on cache folder.
On my machine the cache folder machinery of ONE push everythin in my home linux.
`/home/samuel/Downloads/ONE/openalyx.internationalbrainlab.org/cache`

In my institute home are very very small!! I woudl easily saturate the home by mistake.
Maybe we should have clear doc on this.

Do you know what is cache eactly here ?
Only metadata or on all download data chunks ?

spikeinterface/extractors/stream_ibl.py
```diff
@@ -0,0 +1,239 @@
...
+        return self._file_streamer.ns
+
+    def get_traces(self, start_frame: int, end_frame: int, channel_indices):
+        if start_frame is None:
+            start_frame = 0
+        if end_frame is None:
+            end_frame = self.get_num_samples()
+        if channel_indices is None:
+            channel_indices = slice(None)
+
```
samuelgarcia, line 231:
```suggestion
# This is a temporary fixed to mimic the correct behavior to get the none scaled traces:
```

samuelgarcia:
Hi Cody,
This is really cool.
Thanks a lot for this.
I made some comments on the fly.

I am wondering how long are the test for githib actions here ?

Ready to merge on my side.

spikeinterface/extractors/stream_ibl.py
CodyCBakerPhD: I enhanced the docstring a bit to clarify some of these things.

The cache folder is apparently necessary, all metadata about session IDs, probe names, the AP/LF band meta header files all go here on-demand.

It does indeed seem to tend to choose the 'Downloads' folder on a given system, which you're right, may be quite small.

This is also why I have `remove_cached=True` set as the default, whereas they keep it as `False` by default. That's probably best for performance when repeatedly reusing the same snippets of data multiple times, but when loading a full sessions worth (> 100 GB) that could very easily fill a home volume.

With `remove_cached=True`, each 'chunk' of data (~15 MB x the number of chunks you need to stream to return the specified slice of the traces) gets removed after you read it from the streamer and copy it into RAM, do your operations on it, then load in another slice.

CodyCBakerPhD:
> I am wondering how long are the test for githib actions here ?

Haven't timed the CI but when running them on the DANDI Hub they take less than 20 seconds I would say, they're not heavy at all (only 1 'chunk' as IBL stores them, which is 384 channels for 1 second or ~30000 samples of compressed int16 + header files).

alejoe91:
@samuelgarcia can I merge?

spikeinterface/extractors/stream_ibl.py
```diff
@@ -42,9 +42,11 @@ class StreamingIblExtractor(BaseRecording):
         If not then the probe is loaded.
     cache_folder : str, optional
         The location to temporarily store chunks of data during streaming.
-        The default uses the folder designated by ONE.alyx._par.CACHE_DIR / "cache".
-    remove_cached : bool, default: true
-        Whether or not to remove data from the cache after it is read.
+        The default uses the folder designated by ONE.alyx._par.CACHE_DIR / "cache", which is typically the designated
+        'Downloads' folder on your operating system. As long as `remove_cached` is set to True, the only files that will
+        persist in this folder are the metadata header files and the chunk of data being actively streamed and used in RAM.
```
samuelgarcia, line 47: perfect

PR #1209, opened 2023-01-19
Initial refactoring for node pipeline

samuelgarcia:
@alejoe91 @h-mayorquin  @yger @cwindolf @julienboussard 

This refactor the PipelineStep into PipelineNode that enable to make a simple graph for featurization on peaks.
This enable Node that can denoise the waveform buffer.

Have a look to `test_peak_pipeline.py` which show case some simple toy example.

If we have agreement on this I will propagate to any place needed.


h-mayorquin:
I like this. This being a directed acylical graph provide us with a pipeline with some additional flexibility for the sake of computation efficiency. I dedicated some time today to think about this extra flexibility. These are my thoughts so far and I want to see if I am not missing anything.

The proposed method covers the following scenarios.

2023-01-249(PeakFeatureExtractionPipeline)

i think this covers all the possible scenarios save a combination of them. Am I getting all the scenarios that you were thinking off @samuelgarcia?

Maybe the core of this is that it improves a normal pipeline in a more computational efficient way. That is, the clase number II above avoids the computation / extraction of the waveforms twice. 

Inicidentally and, as case III illustrates, this simplifies the computational expression of features that requires the computation of two features as an input. Are there cases of this? 

So far the examples that I have seen only look like the case I above. 

**Question**: are there specific examples in known pipelines, that use II or III? I looked for this on the code for tridesclous2 and spikecircus2 and could not find them (but maybe they are there).

h-mayorquin:
Comment:
Something to think about is that because we parallelize over chunks of the traces the computation of features that require the state of the features outside of its respective chunk is not possible. Maybe it is not common that a features requires anything outside of their local peaks for computation but is a limitation that might be worth to have on mind if we commit to this approach.

spikeinterface/sortingcomponents/tests/test_peak_pipeline.py
```diff
@@ -62,17 +72,29 @@ def test_run_peak_pipeline():
...
-
+    # 3 nodes two have outputs
+    nodes = [
+        ExtractDenseWaveforms(recording, name='extract_waveforms', ms_before=.5, ms_after=1.,  have_global_output=False),
+        WaveformDenoiser(recording, name='denoiser', parents=['extract_waveforms'], have_global_output=False),
+        MyStep(recording, name='simple_step', param0=5.5),
+        MyStepWithWaveforms(recording, name='step_on_raw_wf', parents=['extract_waveforms']),
+        MyStepWithWaveforms(recording, name='step_on_denoised_wf', parents=['denoiser'])
+    ]
+
```
h-mayorquin, line 89: In the example above, are you envisoning that steps that are not related to the other steps at all like the "simple_step"  will be used in the pipeline?

spikeinterface/sortingcomponents/peak_pipeline.py
```diff
@@ -11,88 +11,136 @@
...
+    def get_trace_margin(self):
+        return max(self.nbefore, self.nafter)
+    
+    def compute(self, traces, peaks):
+        waveforms = traces[peaks['sample_ind'][:, None] + np.arange(-self.nbefore, self.nafter)]
+        return waveforms
 
-def run_peak_pipeline(recording, peaks, steps, job_kwargs, job_name='peak_pipeline', squeeze_output=True):
+
+def check_graph(nodes):
```
h-mayorquin, line 90: I think this method could be separated in two:

`assert_graph_structure` which can be run in the general init and `propagate_node_instances` (or something like that) that is used to propagate the instances in the `worker_ctx` intializer.

This is probably slightly more efficient and more readable as well (the function right now does not have a specific enough name and is performing two roles).

spikeinterface/sortingcomponents/peak_pipeline.py
```diff
@@ -11,88 +11,136 @@
...
-            self.contact_locations = recording.get_channel_locations()
-            self.channel_distance = get_channel_distances(recording)
-            self.neighbours_mask = self.channel_distance < local_radius_um
+    This is a generic object that will make some computation on peak given a buffer
+    of traces.
+    Typically used for extrating features (amplitudes, localization, ...)
+    
+    A Node can optionally connect to other nodes with the parents and receive inputs from others.
+    """
+    def __init__(self, recording, name, have_global_output, parents=None):
```
h-mayorquin, line 29: Suggestion:
I think that `return_ouput` is more clear than `have_global_output`
Maybe there is something better...

h-mayorquin (commented):
I like the change from `compute_buffer` to `compute` and the refactoring of the local_peak calculation. Some minor commments for my own understanding.

samuelgarcia:
Hi @samuelgarcia.
Thank you for this feedback.
Indeed this 3 scenario are possible with this PR.
Somehow I am trying to make example inside /spikeinterface/spikeinterface/sortingcomponents/tests/test_peak_pipeline.py
Not all cases are covered but I can work on this.

At the moment spykingcircus2 are using high level function to extract features `compute_features_from_peaks()`.
No non pipeline like this.

About the comments the parralellisation over chunks, yes you are right but it is related to the `ChunkRecordingExecutor` itself and mainly the global design. This PR will not make drastic chnages about this.
Worker do not communicate between thenself. I think this is ok for the moment to have an embaraingly parralell approach.

Happy to discuss with you about this in call soon.

spikeinterface/sortingcomponents/features_from_peaks.py
```diff
@@ -45,25 +46,25 @@ def compute_features_from_peaks(
     """
     job_kwargs = fix_job_kwargs(job_kwargs)
 
-    steps = []
+    nodes = [
```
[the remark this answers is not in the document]
samuelgarcia: I prefer to remove totally the concept of step and replace clearly by node so it is more obvious that it is a graph.
`need_waveforms` have dissapear. It is the responsability to have the node `ExtractDenseWaveforms` in the graph when needed.

spikeinterface/sortingcomponents/tests/test_peak_pipeline.py
samuelgarcia: Yes we can have node that not depend to each other. In short having parralel graph.

spikeinterface/sortingcomponents/peak_pipeline.py
samuelgarcia: OK thank you. This make sens. I will do that.

spikeinterface/sortingcomponents/peak_pipeline.py
samuelgarcia: yep. More elegant. I will do that.

yger:
Hi guys. I'm remotely following what you are doing, but I like the approach a lot. I guess with such a node approach, it would be easy to write its own custom denoising node such that I can extract waveforms, denoise, and then compute features from these on the fly. Am I right? Circus2 would need such an approach :-)

samuelgarcia:
> I guess with such a node approach, it would be easy to write its own custom denoising node such that I can extract waveforms, denoise, and then compute features from these on the fly. Am I right?

Exactement!

spikeinterface/sortingcomponents/peak_pipeline.py
samuelgarcia: done

spikeinterface/sortingcomponents/peak_pipeline.py
samuelgarcia: Done

h-mayorquin (commented):
I am coming back to the point that @alejoe91 made about passing instances instead of string references at initialization.

I think that the way that we are doing where instances are propagated only at run time instead of at initialization goes against the programming principle that [things should fail as early as possible](https://stackoverflow.com/questions/2807241/what-does-the-expression-fail-early-mean-and-when-would-you-want-to-do-so) with all the problems that it entails.

As an example of the this, when running a waveform denoiser it will be good to know at initialization that the temporal parameters of the selected model are not compatible with the WaveformExtractor that was choosen instead of at run-time.

I remember that you pointed out that you were concerned about serialization @samuelgarcia? I think I can address this in an easy way using some of the work that I did for the other approach. I will put together a short PR in top of this to showcase this. Did you have any other concern?

samuelgarcia:
Hi @h-mayorquin.
1. About parents as an instance : I have the feeling that it will make the declarative way super heavy. We will need to first declare all nodes and then connect then end then create the list. Honnetsly I think that parent by label appraoch is enough for the moment.
2. About naming every outputs (and so going for dict) :  my choice 50/50 between the two options : order args or dict
dict will lead to semantic like `input['A'] + input['B']` instead of `A+ B`. Let me maturate that point a few days more.


Lets keep in mind that this functionality is very internal.
And will be used mainly internally for extrating features efficiently.
Even in sorter class we will use higher level function like `compute_features_from_peaks()`.

spikeinterface/sortingcomponents/peak_pipeline.py
```diff
@@ -11,88 +11,171 @@
...
+    for node in nodes:
+        # give the the instances from all nodes.
+        # Usefull for parameters propagation and checking
+        node.give_other_nodes(dict_nodes)
+
+    for node in nodes:
+        node.post_check()
+    
+
+def run_peak_pipeline(recording, peaks, nodes, job_kwargs, job_name='peak_pipeline', squeeze_output=True):
```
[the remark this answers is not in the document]
samuelgarcia: Not sure to follow you on this. `nodes` is quite self explanatory a list of nodes.

spikeinterface/sortingcomponents/peak_pipeline.py
```diff
@@ -105,94 +188,92 @@ def run_peak_pipeline(recording, peaks, steps, job_kwargs, job_name='peak_pipeli
...
-    return outs
+            # the node need imputs from other nodes
+            inputs = tuple()
+            for parent_name in node.parents:
+                other_out = outputs[parent_name]
+                if not isinstance(other_out, tuple):
+                    other_out = (other_out, )
+                inputs += other_out
 
+            out = node.compute(traces, local_peaks, *inputs)
```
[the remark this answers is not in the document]
samuelgarcia: I made a global answer on this in the main thread.

h-mayorquin:
@samuelgarcia 
> 1. About parents as an instance : I have the feeling that it will make the declarative way super heavy. We will need to first declare all nodes and then connect then end then create the list. Honnetsly I think that parent by label appraoch is enough for the moment.

So it is going to be the current approach
```python
    nodes = [
        ExtractDenseWaveforms(recording, name='extract_dense_waveforms', ms_before=.5, ms_after=1.,  return_ouput=False),
        WaveformDenoiser(recording, name='denoiser', parents=['extract_dense_waveforms'], return_ouput=False),
    ]
```
Against the following:
```python
    extract_dense_waveforms =  ExtractDenseWaveforms(recording, ms_before=.5, ms_after=1.,  return_ouput=False)
    waveform_denoiser = WaveformDenoiser(recording, name='denoiser', parents=extract_dense_waveforms, return_ouput=False)
    
    nodes = [extract_dense_waveforms, waveform_denoiser]
```
Is the extra assignation that bothers you when you say making the declarative super heavy? 
The instance passing approach also relieves the user of the function of having to declare names for every class and us, the mantainers, from having to keep machinery to make references /names unique (which is duplicating the work that python already does). Plus, we get close to the principle that indicates that things should fail early.

> 2. About naming every outputs (and so going for dict) :  my choice 50/50 between the two options : order args or dict
>    dict will lead to semantic like `input['A'] + input['B']` instead of `A+ B`. Let me maturate that point a few days more.

All right, I am curios what will you come up with : )

> Lets keep in mind that this functionality is very internal. And will be used mainly internally for extrating features efficiently. Even in sorter class we will use higher level function like `compute_features_from_peaks()`.

alejoe91 (approved):

alejoe91:
@h-mayorquin merged as is. We can work on updates later. But at least you can start refactoring tPCA and WaveformDenoiser with this :)

samuelgarcia:
@h-mayorquin : I think tPCA and WaveformDenoiser should have the same base class because theses nodes have a fit/transform with an underlying model to save/load.
Is it your plan ?

PR #1025, opened 2022-10-21
Add IBL channel interpolation to preprocessing

JoeZiminski:
Hi All,

Apologies for the delay with this, this PR is to port the [IBL channel interpolation](https://github.com/int-brain-lab/ibl-neuropixel/blob/df0058747bb9c7da380c469e4efa9c841b757472/src/neurodsp/voltage.py#L213). Its close to complete I think but have marked as draft as have a couple of questions.

1) The IBL pipeline is for neuropixels probs and so the offset (euclidean distance in um) is [calculated across x, y coordinates](https://github.com/JoeZiminski/spikeinterface/blob/1b82aa8d21141eaebe5608faa60d0d4f09f7ed90/spikeinterface/preprocessing/interpolate_bad_channels.py#L117). I see that probeinterface can support 3D arrangements, I believe passing the full get_probe().contact_positions matrix and calculating the L2 norm (all channel positions - bad channel position) here will cover the 2D scenario as well as 3D. However, I just wanted to check I hadn't missed anything that might cause problems appying this method in 3D.

2) I couldn't figure out how to get the scaled version of the data from within BasePreprocessorSegment. To match IBL output in test_interpolate_bad_channels I used [ScaleRecording()](https://github.com/JoeZiminski/spikeinterface/blob/1b82aa8d21141eaebe5608faa60d0d4f09f7ed90/spikeinterface/preprocessing/tests/test_interpolate_bad_channels.py#L35), but not sure if there is a better way

3) remove cupy option from original IBL function, happy to put back in if dependency is okay

4) My understanding of the Kriging distance is the distance between contact probes. If kriging distance is not supllied will [estimate from contact positions](https://github.com/JoeZiminski/spikeinterface/blob/1b82aa8d21141eaebe5608faa60d0d4f09f7ed90/spikeinterface/preprocessing/interpolate_bad_channels.py#L45)  (used Y dimension to match IBL default arg). Currently if probe units are not "um" will raise NotImplementedError for now. Are other units supported in ProbeInterface?

5) Highlighting the test currently requies IBL install from github in case this is an issue. Will also write a test on generate_recording() data before converting to ready to review.

Will make new PR soon for bad channel detection and kfiltering,

Cheers,
Joe

PR #1222, opened 2023-01-27
Promote distinctipy color as a second choice because it is slow.

samuelgarcia:
I discoverd that distinctipy which is great is in fact very very slow!
17s to generate 250 colors.
Since the get_unit_colors is call at almost every plots. I put back matplotlib by default.

### DIFF
--- a/spikeinterface/widgets/_legacy_mpl_widgets/__init__.py
+++ b/spikeinterface/widgets/_legacy_mpl_widgets/__init__.py
@@ -1,4 +1,3 @@
-from .utils import get_unit_colors
 
 # basics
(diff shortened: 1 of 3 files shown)


samuelgarcia:
@alejoe91 : if you could merge if you are happy this would help the sigui which is broken until this is merge because of this https://github.com/SpikeInterface/spikeinterface-gui/pull/37

alejoe91:
Great! I'll review it when you review #1025 :)

samuelgarcia:
OK. I think I merge already the #1025

alejoe91 (approved):

Issue #1118, opened 2022-11-29
Avoid random and computation in preprocessing init class


samuelgarcia:
Some classes for preprocessing like `WhitenRecording, `NormalizeByQuantileRecording`,`ZScoreRecording` are using internally `get_random_data_chunks()`.

This make the end user experience easier but this is rather bad for:
  * reproducibility
  * parralel processing

For parralel processing particularly  it is really really bad because:
  * every worker make a differents random and so the noise or covariance matrix is not the same across workers.
     And so the n_jobs>1 make different results for every run.
  * the startup of every worker can be super long when n_jobs is very high because each worker is fighting for CPU ressource 
     for all inversing  a covaraince matrix for instance.

We have a way to store _kwargs but we should have another way (other dict) that would enable to restore the class very quickly in the same state without any random in between.

zm711:
How close are we to this? The issue is two years old and I doubt we really look at it. It is an ongoing effort, but I think we've made huge progress no?

zm711:
@alejoe91 I'll bring you into this :) What do you think? Close this as super stale for now? I don't know if keep this as a reminder is serving a purpose at this point.

alejoe91:
This is fixed :) the random stuff is computed in the main process and added as kwargs

PR #1227, opened 2023-01-29
Serialization for recording and sorting objects

h-mayorquin:
As discussed, it would be nice to have a way of serializing the recording and sorting objects. For example, an obvious advantage of this is that we don't need to define `to_dict` and `_from_dict` methods on objects that have a recording or a sorter as an attribute (whose responsability is to produce representations which can be serialized for parallel implementations as we do for example in the sorting components library). Having pickable extractors simplifies creating new classess and functions that use recording and sorters and in addition it reduces our maintenance burden. Moreover, using the reduce protocol will make our workflow compatible with `pickle`, the standard for serialization library in the python ecosystem.

The PR here implements this feature by relying on the available machinery that's built already in the `to_dict` functions. I am putting this forward as a minimal disruptive change that adds the desired serialization future.

I added tests for all recorders and sorters. Mearec seems to be failing because of a type issue when comparing and the EDF readers seems to have a lock-in mechanism for IO reference. All the other tests are passed in my personal computer. Pushing this here to test the CI.

I think that in the future we would like to make the machinery for serialization a subset of the machinery to json-ify (right know they are the same) in order to improve the performance of the former (and also, to have better semantics, to separate the `kwars` which I think is an overly general name IMHO).

Let me know what you think.


samuelgarcia:
Really cool. That what I wanted to do since long time.
I will make feedback on this soon.
I did not figure out that it was so easy.

I propose to coordinate with you to also remore the `to_dict()` mechanisim every time we run a `ChunkRecordingExecutor`

samuelgarcia:
There is also somethingk important to take in account.

Some etxractor that do preprocessing need some precomputation (noise estimation, whitening matice, ...)
We need a secondary dict used also in the pidckeling this to avoid recomptation in the worker.

Read this #1118

I would propose to also include this in the PR.

h-mayorquin:
I am very confused about why the nwbextractor test is failing with streaming through ffspec. The test run well as a stand alone and it seems that the changes play in a strange way with the Session mechanism of pytest (which I don't fully understand).

Have you guys seen this before?
Also I noticed that while running the extractor tests the global folder points to the preprocessing directory which I guess hints at a bug with the use of the global temporal directory for caching. That said, I don't think this bug comes from there, at least not entirely.

h-mayorquin:
OK, I sorted this out. It is not clear how but the maxwell pipeline was messing up hdf5 when installing its plug-in and that caused the NWBRecordingExtractor to fail only when streaming with fsspec.

I think this is ready to review, I am thinking we can make this PR a small part of large serialization effort described on #1235. Let me know what you think.

alejoe91:
> OK, I sorted this out. It is not clear how but the maxwell pipeline was messing up hdf5 when installing its plug-in and that caused the NWBRecordingExtractor to fail only when streaming with fsspec.
> 
> I think this is ready to review, I am thinking we can make this PR a small part of large serialization effort described on #1235. Let me know what you think.

That sounds like a nightmare to debug!!!

spikeinterface/core/base.py
```diff
@@ -562,10 +562,14 @@ def load(file_path, base_folder=None):
...
 
+    def __reduce__(self):
+        
+        return (self.from_dict, (self.to_dict(), ))
+
     @staticmethod
     def load_from_folder(folder):
         return BaseExtractor.load(folder)
-
+
```
h-mayorquin, line 572:
```suggestion
 
```

spikeinterface/core/base.py
```diff
@@ -562,10 +562,14 @@ def load(file_path, base_folder=None):
...
```
h-mayorquin, line 572:
```suggestion

```

spikeinterface/extractors/neoextractors/maxwell.py
```diff
@@ -30,14 +30,19 @@ class MaxwellRecordingExtractor(NeoBaseRecordingExtractor):
...
     mode = 'file'
     NeoRawIOClass = 'MaxwellRawIO'
     name = "maxwell"
     has_default_locations = True
 
     def __init__(self, file_path, stream_id=None, stream_name=None, block_index=None, 
-                 all_annotations=False, rec_name=None):
+                 all_annotations=False, rec_name=None, install_maxwell_plugin=False):
+        if install_maxwell_plugin:
+            self.install_maxwell_pluging()
```
alejoe91, line 44:
```suggestion
            self.install_maxwell_plugin()
```

spikeinterface/extractors/neoextractors/maxwell.py
```diff
@@ -61,6 +66,12 @@ def map_to_neo_kwargs(cls, file_path, rec_name=None):
         neo_kwargs = {'filename': str(file_path), 'rec_name': rec_name}
         return neo_kwargs
 
+    def install_maxwell_pluging(self, force_download=False):
```
alejoe91, line 69:
```suggestion
    def install_maxwell_plugin(self, force_download=False):
```

alejoe91 (commented):
@h-mayorquin just a couple of small typos

spikeinterface/extractors/neoextractors/maxwell.py
h-mayorquin: Done in the last commit.

spikeinterface/extractors/neoextractors/maxwell.py
h-mayorquin: Done in the last commit.

samuelgarcia (approved):

PR #1233, opened 2023-01-30
Testing `test_s3_nwb_fsspec` caching

h-mayorquin:
I am trying to pin down the errors in #1227

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -22,6 +22,12 @@ def pytest_sessionstart(session):
 
 
 def pytest_collection_modifyitems(config, items):
+    """
+    This function marks (in the pytest sense) the tests according to their name and file_path location
+    Marking them in turn allows the tests to be run by using the pytest -m marker_name option.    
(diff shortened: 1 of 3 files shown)


h-mayorquin:
OK, turns out that the error in #1227 was caused by something else (I was blaming caching by fsspec for the segmentation fault but I was wrong.).

That said, I think we could merge these small additions. They are:
1) Added docstring to `pytest_collection_modifyitems` which should be useful for other developer on the future that want to quickly understand what that function is doing in our test suit.
2) Removed a redudant call to `check_nwb_install` in `NwbRecordingExtractor`.
3) Tes the capabilities of `NwbRecordingExtractor` to allocate a specific caching directory path for streaming. For this, I am using the same `global_test_folder` that I saw in other tests.

What do you think?

alejoe91 (approved):
Awesome @h-mayorquin :)

Thanks!

PR #1239, opened 2023-01-31
[Issue - do not merge] `test_calculate_drift_metrics` is failing on master.

h-mayorquin:
The merge of PR's #1209  and #1143 is causing a test to fail in my own PR (#1227). However, the same error appears on master when I try locally. Looking at the tests of #1143 I think their test was brittle so I am looking at this now.

Edit: confirmation. test_calculate_drift_metrics is failing on master.


h-mayorquin:
Closing now as we have done #1241 and keeping the issue #1240 to address this in the future in

Issue #1240, opened 2023-01-31
`test_calculate_drift_metrics` is failing on master


h-mayorquin:
I noticed this while working on #1227. I think that #1209 broke the test but because it is in a different module did not trigger the CI (this seems like a problem with our CI). Maybe it was even before that (could be #1143 ?).

Anyway, I propose to put the test on pause so we can continue working / merging while we repair this.

When this issue is closed this should also close #1239.

alejoe91:
Thanks Ramon, I'll take a look rn

alejoe91:
I agree we could pause the test for now, can you make a PR so I can merge it?

h-mayorquin:
Done.  I also added another marker so this is ran whenever there is a change in sorting components to keep track of it. 

Probably, we should modify the CI in such a way that if a test imports a module, changing the imported module should run the tests. Maybe this can be done with importlib or by parsing the PATH. I wonder how much of hack that would be.

PR #1241, opened 2023-01-31
Pausing `test_calculate_drift_metrics` test

h-mayorquin:
This is related to #1240. I also added another marker so this is ran whenever there is a change in sorting components to keep track of it.


alejoe91:
I wouldn't mark as `xfail` but just as `skip` for the moment. It shouldn't fail :P

alejoe91 (approved):

PR #1242, opened 2023-02-01
Motion estimation updates

cwindolf:
Opening a pull request for this, but still need to update the tests (+ if you have any review)


spikeinterface/sortingcomponents/motion_estimation.py
```diff
@@ -404,34 +439,41 @@ def get_windows(rigid, bin_um, contact_pos, spatial_bin_edges, margin_um, win_st
...
             elif win_shape == 'rect':
                 win = np.abs(bin_centers - win_center) < (win_sigma_um / 2.)
                 win = win.astype('float64')
+            elif win_shape == 'triangle':
+                center_dist = np.abs(bin_centers - win_center)
+                in_window = center_dist <= (win_sigma_um / 2.)
+                win = -center_dist
+                win[~in_window] = 0
+                win[in_window] -= win[in_window].min()
+                win[in_window] /= win[in_window].max()
```
samuelgarcia, line 472: Finally the trend is on triangle ??

samuelgarcia:
@cwindolf : thanks a lot.
I merge our benchmark motion PR so this PR is now on top of master.

spikeinterface/sortingcomponents/motion_estimation.py
```diff
@@ -180,91 +174,132 @@ class DecentralizedRegistration:
...
-        minimum correlation to estimate a motion shift for pairwise displacement bellow the value not used.
+        Minimum correlation between pair of time bins in order for these to be
+        considered when optimizing a global displacment vector to align with
+        the pairwise displacements.
     time_horizon_s: None or float
         When not None the parwise discplament matrix is computed in a small time horizon.
         In short only pair of bins close in time.
         So the pariwaise matrix is super sparse and have values only the diagonal.
-    convergence_method='lsqr_robust' or 'gradient_descent'
+    convergence_method='lsmr', 'lsqr_robust', 'gradient_descent'
```
samuelgarcia, line 200: Maybe we could add a little explanation/history of theses methods and related paper no ?

spikeinterface/sortingcomponents/motion_estimation.py
```diff
@@ -793,14 +865,136 @@ def jac(p):
...
+        step_size = 0.1
+        round_mode = np.round  # floor?
+        best_ref = np.median(displacement)
+        max_zeros = np.sum(round_mode(displacement - best_ref) == 0)
+        for ref in np.arange(np.floor(displacement.min()), np.ceil(displacement.max()), step_size):
+            n_zeros = np.sum(round_mode(displacement - ref) == 0)
+            if n_zeros > max_zeros:
+                max_zeros = n_zeros
+                best_ref = ref
+        displacement -= best_ref
```
samuelgarcia, line 995: From what I understand the continuity of spatial dim is handle already by spatial_prior.
Here we are handeling the "common" stuff and not by windowd stuff.
Am I correct ? If so I would vote for median by default.
Empirical mode sometimes fails due to the bin size. No ?
Did you ebnchmark this "mode_search" ?

spikeinterface/sortingcomponents/motion_estimation.py
cwindolf: Haha, I like them, and you'll like this -- with `win_shape == 'triangle'` and `win_sigma_um == 2 * win_step_um`, then `upsample_to_histogram_bin` leads to linear interpolation of the motion estimate between bins :)

spikeinterface/sortingcomponents/motion_estimation.py
cwindolf: Mode search is a little silly -- could be sped up with a call to histogram anyway. I'd be happy with median by default.

cwindolf:
Since spatial_prior is slow/memory hungry, and since it's not always necessary (it only helps on recordings with a lot of glitches and blank spots), I added something to ensure spatial continuity well enough with `spatial_prior=False`. So, I think it will be safe to have spatial_prior=False by default.

spikeinterface/sortingcomponents/motion_estimation.py
samuelgarcia: Lets keep then a bit more!

spikeinterface/sortingcomponents/motion_estimation.py
```diff
@@ -180,91 +181,176 @@ class DecentralizedRegistration:
...
+                temporal_prior=temporal_prior,
+                spatial_prior=spatial_prior,
+                soft_weights=soft_weights,
+                progress_bar=False,
+            )
+        elif len(non_rigid_windows) > 1:
+            # if spatial_prior is False, we still want keep the spatial bins
+            # correctly offset from each other
+            for i in range(len(non_rigid_windows) - 1):
+                motion[:, i + 1] -= np.median(motion[:, i + 1] - motion[:, i])
```
samuelgarcia, line 327: Should we also make sure that np.median(motion[:, 0]) == 0 ?
On short the first layer do not move in median. and then the diff is 0 relative to the previous ?

samuelgarcia: Forget it this is done just afte rthis.

JonathanAMichaels:
@cwindolf I was just testing this branch. Works well – except spatial_prior=True seems to fail at compute_global_displacement because of the different structure of all_pairwise_displacement_weights.

samuelgarcia (approved):

PR #1245, opened 2023-02-01
Implement tPCA component with node pipeline

h-mayorquin:
Porting #1145 for using the node pipeline.


spikeinterface/sortingcomponents/waveforms/pca.py
```diff
@@ -0,0 +1,150 @@
...
+        n_waveforms, n_samples , n_channels = waveforms.shape
+        n_components = self.pca_model.n_components
+        
+        # Initializing non-valid channels with nan (sparsity is used to indicate valid channels)
+        projected_waveforms = np.full(shape =(n_waveforms, n_components, n_channels), fill_value=np.nan)
+
+        for waveform_index, main_channel in enumerate(peaks['channel_ind']):
+            channel_indexes_sparsified,  = np.nonzero(self.neighbours_mask[main_channel])
+            sparsified_waveform = waveforms[waveform_index, :, channel_indexes_sparsified]
+            projected_waveforms[waveform_index, :, channel_indexes_sparsified] = self.pca_model.transform(sparsified_waveform)
```
alejoe91, line 148: Did we agree to use `max_sparse_channels` and place all valid channels in the first positions?

h-mayorquin: Yes, that's something I need to do.

alejoe91, line 148:
```suggestion
        projected_waveforms = self.pca_model.transform(waveforms.swapaxes(1, 2).reshape(-1, n_samples))
```

h-mayorquin: OK, I added a test for this when the waveforms are sparse.

h-mayorquin: Done in the last commit.

spikeinterface/sortingcomponents/waveforms/waveform_utils.py
```diff
@@ -0,0 +1,20 @@
+def to_channelless_representation(waveforms):
```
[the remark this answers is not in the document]
h-mayorquin: As discussed in the meeting, I think this is not specific enough (stack how? unstack from where?). I think that is better to name operations that reflect the domain of application. If, temporal representation is more meaningful term in this domain I think that `to_temporal_representation` is better (that is, if it is obvious for everyone that this means "throw the spatial information out and consider every channel an independent sample"). Moreover, this can be explained in the docstring and this is why it is useful to have a common method for the scattered use of this operation.

I think that "channeless_representation" point specifically to what the method is doing but is not a common word in the English language ( google n-grams reports a strange usage pattern)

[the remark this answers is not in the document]
alejoe91: Ok for me!

spikeinterface/sortingcomponents/waveforms/waveform_utils.py
```diff
@@ -0,0 +1,20 @@
...
+    Transform waveforms to channelless representation. Collapses the channel dimension leaving only 
+    temporal information. 
+    """
+    num_waveforms, num_time_samples, num_channels = waveforms.shape
+    num_channelless_waveforms = num_waveforms * num_channels
+    channelless_waveforms = waveforms.swapaxes(1, 2).reshape((num_channelless_waveforms, num_time_samples))        
+
+    return channelless_waveforms
+
+def from_channelless_representation(channelless_waveforms, num_channels):
```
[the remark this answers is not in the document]
h-mayorquin: Done in the last change

h-mayorquin:
I discussed with @samuelgarcia , there are three options concerning the fit method:
1) The way that I have, where you instantiate and then fit.
2) Make it a class method that he favors.
3) Make it a step in the pipelien mode.

Let's discuss this on the next meeting, I will think about pros and cons.

spikeinterface/sortingcomponents/waveforms/pca.py
```diff
@@ -0,0 +1,175 @@
...
+from spikeinterface.postprocessing import compute_principal_components
+from spikeinterface.core import BaseRecording
+from spikeinterface.core.sparsity import ChannelSparsity
+from spikeinterface import extract_waveforms, NumpySorting
+from spikeinterface.core.job_tools import _shared_job_kwargs_doc
+from spikeinterface.core.recording_tools import get_channel_distances
+from .waveform_utils import to_temporal_representation, from_temporal_representation
+
+
+class PCBaseNode(PipelineNode):
```
alejoe91, line 19:
```suggestion
class tPCABaseNode(PipelineNode):
```
Should we rename it to tPCA to keep the semantic of the Paninski group? (same for the file `tpca.py`)

spikeinterface/sortingcomponents/waveforms/pca.py
```diff
@@ -0,0 +1,175 @@
...
+        # model_folder should be an input
+        pca_model_path = self._kwargs["model_path"]
+
+        with open(pca_model_path, "wb") as f:
+            pickle.dump(self.pca_model, f)
+            
+        return self.pca_model
+    
+PCBaseNode.fit.__doc__ = PCBaseNode.fit.__doc__.format(_shared_job_kwargs_doc)
+class PCAProjection(PCBaseNode):
```
alejoe91, line 101:
```suggestion
class tPCAProjection(tPCABaseNode):
```
same here

spikeinterface/sortingcomponents/waveforms/pca.py
```diff
@@ -0,0 +1,175 @@
...
+            )
+            raise AttributeError(exception_string)
+        
+        num_waveforms, num_samples , num_channels = waveforms.shape
+        
+        channeless_waveform = to_temporal_representation(waveforms)
+        projected_chaneless_waveforms = self.pca_model.transform(channeless_waveform)
+        projected_waveforms = from_temporal_representation(projected_chaneless_waveforms, num_channels)
+
+        return projected_waveforms
```
alejoe91, line 175: Should we add a `tPCADenoiser` class to return clean waveforms after backprojection?

alejoe91: we could also add it later :)

alejoe91 (commented):
Thanks @h-mayorquin 

Looks good to me. I have some re-naming suggestions. Also, can you make sure indentations in functions and classes args are correct?

spikeinterface/sortingcomponents/waveforms/pca.py
```diff
@@ -0,0 +1,175 @@
...
+
+        # model_folder should be an input
+        pca_model_path = self._kwargs["model_path"]
+
+        with open(pca_model_path, "wb") as f:
+            pickle.dump(self.pca_model, f)
+            
+        return self.pca_model
+    
+PCBaseNode.fit.__doc__ = PCBaseNode.fit.__doc__.format(_shared_job_kwargs_doc)
```
alejoe91, line 100:
```suggestion
PCBaseNode.fit.__doc__ = PCBaseNode.fit.__doc__.format(_shared_job_kwargs_doc)


```

spikeinterface/sortingcomponents/waveforms/pca.py
```diff
@@ -0,0 +1,175 @@
...
+from spikeinterface.core.recording_tools import get_channel_distances
+from .waveform_utils import to_temporal_representation, from_temporal_representation
+
+
+class PCBaseNode(PipelineNode):
+    
+    def __init__(self, recording: BaseRecording, parents: list = list[PipelineNode], ms_before: float = 1., ms_after: float = 1.,
+            peak_sign: str = 'neg', all_channels: bool = True, model_path: str = None, local_radius_um: float = None,
+            return_ouput=True,
+            ):
```
alejoe91, line 24:
```suggestion
                 peak_sign: str = 'neg', all_channels: bool = True, model_path: str = None, local_radius_um: float = None,
                 return_ouput=True,
            ):
```

spikeinterface/sortingcomponents/waveforms/pca.py
```diff
@@ -0,0 +1,175 @@
...
+        """
+        Base class for PCA projection nodes. Contains the logic of the fit method that should be inherited by all the 
+        child classess. The child should implement a compute method that does a specific operation 
+        (e.g. project, denoise, etc)
+        """
+        
+        PipelineNode.__init__(self, recording=recording, parents=parents, return_ouput=return_ouput)
+        self.all_channels = all_channels
```
alejoe91, line 32: where is this used?

spikeinterface/sortingcomponents/waveforms/pca.py
```diff
@@ -0,0 +1,175 @@
...
+    ms_after : float, optional
+        The number of milliseconds to include after the peak of the spike, by default 1.
+    peak_sign : str, optional
+        The sign of the peak, either 'neg' or 'pos', by default 'neg'.
+    all_channels : bool, optional
+        Whether to extract spikes from all channels or only the channels with spikes, by default True.
+    model_path : str, optional
+        The path to the pca model, by default None.
+    local_radius_um : float, optional
+        The radius (in micrometers) to use for definint sparsity, by default None.
```
alejoe91, line 128: So this is only used for fitting. IMO, it the parent waveforms are already sparse, this should be ignored and sparsity should be inferred from the parent. We should make sure that the docs reflect this. 

A similar argument could be done for `ms_before`, `ms_after`. Even if the cut outs are not exactly the same for the parent waveforms, as long as they are larger than the cutouts the model needs, the `compute_buffer` should take care of futher clipping the waveforms. 
E.g., if the parent waveforms have `ms_before/after=2` and the PCA node has `ms_before/after=1`, at compute time the original waveforms sould be cutout to the correct length before running through the model. What do you think?

spikeinterface/sortingcomponents/waveforms/pca.py
h-mayorquin: I think this is a good idea, I would go for the full Temporal thought.

spikeinterface/sortingcomponents/waveforms/pca.py
h-mayorquin: Yes, @samuelgarcia was suggestig getting rid of all of these parameters at the init. At this moment I agree with this. I will think about it and we can discuss this today later at the meeting.

h-mayorquin:
@alejoe91 @samuelgarcia 
I did the following requested changes:
1) The fit method is now a static method.
2) I added logic to save, load and test a mismatch between the temporal interval  of the waveform extractor (i.e. ms_before, ms_after, sampling_frequency) and the trained PCA model.
3) The parameters characteristing the temporal interval (ms_before, ms_after, sampling_frequency) and others such as sparisty are removed from the classess.
4) All the checks are performed at __init__ now that the fit method is static (achieving two things that @samuelgarcia wanted).
5) The model also checks that one of the parents is a form of waveform extractor. For this, I added a base class for  both waveform extractor that mainly handles the temporal interval properties (ms_before, ms_after, etc).

Let me know how it looks to you.

spikeinterface/sortingcomponents/waveforms/temporal_pca.py
```diff
@@ -0,0 +1,231 @@
...
+                f"model_path folder is not a folder or does not exist. \n"
+                f"A model can be trained by using{self.__class__.__name__}.fit(...)"
+            )
+            raise TypeError(exception_string)
+
+        # Load the model and the time interval dict from the model_folder
+        model_path = Path(model_folder_path) / "pca_model.pkl"
+        with open(model_path, "rb") as f:
+            self.pca_model = pickle.load(f)
+        model_time_interval_path = Path(model_folder_path) / "time_interval.json"
```
alejoe91, line 44:
```suggestion
        model_time_interval_path = Path(model_folder_path) / "params.json"
```

I would call this params to unify across possible models

spikeinterface/sortingcomponents/waveforms/temporal_pca.py
```diff
@@ -0,0 +1,231 @@
...
+        with open(model_time_interval_path, "rb") as f:
+            self.waveform_time_interval_dict = json.load(f)
```
alejoe91, line 46:
```suggestion
            self.params = json.load(f)
```
same here and elsewhere ;)

spikeinterface/sortingcomponents/waveforms/temporal_pca.py
```diff
@@ -0,0 +1,231 @@
...
+
+        # Find waveform extractor in the parents
+        try: 
+            waveform_extractor = next(parent for parent in self.parents if isinstance(parent, WaveformExtractorNode))
```
alejoe91, line 50: Should we make sure that there is only a single `WaveformExtractorNode` parent? If I understand correctly this statement only checks that there is at least one.

IMO, if we want, we should avoid having multiple `WaveformExtractorNode` parents, because the same could be achieved by instantiating two nodes (example):

```
waveforms ----- denoise ----- temporal_pca
             |
             ---------------- temporal_pca
```

in this case one would instantiate two `TemporalPCA` nodes, each taking care of one waveform input.

spikeinterface/sortingcomponents/waveforms/temporal_pca.py
```diff
@@ -0,0 +1,231 @@
...
+            Waveforms extracted from the recording using a WavefomExtractor node.
+
+        Returns
+        -------
+        np.ndarray
+            The projected waveforms.
+
+        """
+
+        num_waveforms, num_samples, num_channels = waveforms.shape
```
alejoe91, line 225:
```suggestion
        num_channels = waveforms.shape[2]
```
Not using the other two

alejoe91 (changes requested):
Thanks @h-mayorquin 

Looks really great! Just a couple of minor comments/renaming variables left  and then it's good to go for me :) Moving on to the other PRs now

spikeinterface/sortingcomponents/waveforms/temporal_pca.py
samuelgarcia: I think I would check that we have only a unique parent. And this parents is WaveformExtractorNode

spikeinterface/sortingcomponents/waveforms/temporal_pca.py
```diff
@@ -0,0 +1,231 @@
...
+        
+        ms_before_mismatch = waveforms_ms_before != model_ms_before
+        ms_after_missmatch = waveforms_ms_after != model_ms_after
+        sampling_frequency_mismatch = waveforms_sampling_frequency != model_sampling_frequency
+        if ms_before_mismatch or ms_after_missmatch or sampling_frequency_mismatch:
+            exception_string = (
+                "PCA model and waveforms mismatch \n"
+                f"{model_ms_before=} and {waveforms_ms_after=} \n"
+                f"{model_ms_after=} and {waveforms_ms_after=} \n"
+                f"{model_sampling_frequency=} and {waveforms_sampling_frequency=} \n"
```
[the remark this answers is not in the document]
samuelgarcia: I guess yes. It display model_ms_before=1.5.

spikeinterface/sortingcomponents/waveforms/temporal_pca.py
```diff
@@ -0,0 +1,231 @@
...
+            )
+            raise ValueError(exception_string)
+    
+    @staticmethod
```
samuelgarcia, line 82: really cool :)

spikeinterface/sortingcomponents/waveforms/temporal_pca.py
```diff
@@ -0,0 +1,231 @@
...
+        n_components: int,
+        model_folder_path: str,
+        detect_peaks_params: dict,
+        peak_selection_params: dict,
+        job_kwargs: dict = None,
+        ms_before: float = 1.0,
+        ms_after: float = 1.0,
+        whiten: bool = True,
+        local_radius_um: float = None,
+    ) -> IncrementalPCA:
```
samuelgarcia, line 94: Could we also add the predetected peaks in params  to skip the detect and slelect ?

samuelgarcia:
Really good PR. Easy and pleasant to read.

Ready to merge on my side.

I am not super happy with the naming "wavforms" as a sub folder in sorting components but we can rename this later.

samuelgarcia (approved):

alejoe91 (approved):

spikeinterface/sortingcomponents/waveforms/waveform_utils.py
```diff
@@ -0,0 +1,20 @@
...
+
+def from_temporal_representation(channelless_waveforms, num_channels):
+    """
+    Transform waveforms from channelless representation. The inverse of to_temporal_representation
+    """
+    num_channelless_waveforms, num_time_samples = channelless_waveforms.shape
+    num_waveforms = num_channelless_waveforms // num_channels
+
+    waveforms = channelless_waveforms.reshape(num_waveforms, num_channels, num_time_samples).swapaxes(2, 1)
+    return waveforms
```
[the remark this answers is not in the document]
h-mayorquin: Thanks for changing the names for me. This makes sense!

PR #1249, opened 2023-02-01
Re-enable test for ros3 streaming 

h-mayorquin:
Separating from #1248 to leave CI working. This is to try to re-encative ROSS streaming.

Edit: Done, this should be working. I am not sure if the problem was scipy changing its version (we don't have it pined for this) or ubuntu-latest upgrading and affecting how OpenBLAS is accessed. Anyway, the problem is solved by installing OpenBLAS directly.

I added two more things:
1) A test on the CI file for this that checks specifically for successful installation of the ros3 driver (so the test going failing because ros3 is not detected does not go silent which is a mistake that I made before)
2) A more helfpul assertion message in the `NWBRecordingExtractor` when the users select streaming ros3 but this is not installed. Plus, I removed what I think was a misleading assertion of when streaming is used.

### DIFF
--- a/.github/workflows/core-test.yml
+++ b/.github/workflows/core-test.yml
@@ -1,29 +1,29 @@
-# name: spikeinterface.core with conda
+name: spikeinterface.core with conda
 
-# on:
-#   pull_request:
-#     branches: [master]
-#     types: [synchronize, opened, reopened, ready_for_review]
+on:
+  pull_request:
+    branches: [master]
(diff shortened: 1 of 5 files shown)


alejoe91 (approved):

alejoe91:
Great! Thanks for looking into this @h-mayorquin 

So in the end the problem was openblas missing?

h-mayorquin:
I think so, it seems the way they build scipy in recent versions:
https://github.com/scipy/scipy/issues/16308

PR #1250, opened 2023-02-01
Fix slicing channels of remove artifact median/avg modes

alejoe91:
The median/average rm artefact needs all traces at the beginning. Slicing channels moved at the end


samuelgarcia:
yes of course!!!

yger (approved):
Good catch

PR #1251, opened 2023-02-01
Small fixes to IblStreaming extractor

CodyCBakerPhD:
I'm planning on eventually contributing a similar interface for the sorting data through their API (the stand-alone AlfSortingExtractor is a bit too limiting to debug for the time being)

So here are some small fixes to the recording extractor I recently added

1) before it gets released and we end up having to do a deprecation cycle, renaming it to follow more closely the SI convention for such classes

`StreamingIblExtractor` -> `IblStreamingRecordingExtractor`

2) small fix to fetching probe suffixes by label (previously should not have been hard-coded to `probe00`)

I need to add a test for #2 still - specific to a session that has multiple probes


spikeinterface/extractors/iblstreamingrecording.py
```diff
@@ -88,7 +88,7 @@ def get_stream_names(cls, session: str, cache_folder: Optional[Union[Path, str]]
             List of stream names as expected by the `stream_name` argument for the class initialization.
         """
         assert HAVE_BRAINBOX_ONE, cls.installation_mesg
-        one = ONE(base_url="https://openalyx.internationalbrainlab.org", password="international", silent=True, cache_dir=cache_folder)
+        one = ONE(cache_dir=cache_folder)
```
CodyCBakerPhD, line 91: Just leaving as a note here, using `cache_dir!=None` here causes a problem sometimes in certain environments

See https://github.com/int-brain-lab/ibllib/issues/564 for more details

CodyCBakerPhD: Also, the CI here really did not like this either, so reverted to previous usage

CodyCBakerPhD:
@samuelgarcia @alejoe91 Ready for review

alejoe91 (approved):

PR #1254, opened 2023-02-02
Silence interval

yger:
Addition of a new pre-processor that can take a list of periods (t_start, t_stop) and blank then with 0, or put some noise. This might be required by people doing micro/macro recordings since they have manually labelled artefacts that do not necessarily saturates (so we can not use blank/clip) but would like to be removed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "spikeinterface"
-version = "0.97.0"
+version = "0.98.0.dev0"
 authors = [
   { name="Alessio Buccino", email="alessiop.buccino@gmail.com" },
   { name="Samuel Garcia", email="sam.garcia.die@gmail.com" },
@@ -123,8 +123,8 @@ test = [
 
     # for github test : probeinterface and neo from master
     # for release we need pypi, so this need to be commented
-    # "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",
-    # "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",
+    "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",
+    "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",
     
 ]
 


alejoe91:
@yger the "silence" mode is the same as the `mode="zeros"` for the `remove_artifact` function, no?
I would just add the `noise` mode to that class instead of making a new preprocessor. What do you think?

yger:
No, because some users want to provide a list of periods (and their sizes can differ), and for now, the remove_artefacts somehow assumes the same ms_before and ms_after for every artefacts, am I wrong?

spikeinterface/preprocessing/silence_periods.py
```diff
@@ -0,0 +1,118 @@
...
+    periods are zeroed-out (mode = 'zeros'). This is only recommended 
+    for traces that are centered around zero (e.g. through a prior highpass
+    filter); if this is not the case, you can also fill the periods with noise.
+
+    Parameters
+    ----------
+    recording: RecordingExtractor
+        The recording extractor to remove artifacts from
+    list of lists/arrays
+        One list per segment of tuples (frame_start, rame_stop) to silence
```
alejoe91, line 21:
```suggestion
        One list per segment of tuples (start_frame, end_frame) to silence
```

spikeinterface/preprocessing/silence_periods.py
```diff
@@ -0,0 +1,118 @@
...
+    Silence user-defined periods from recording extractor traces. By default, 
+    periods are zeroed-out (mode = 'zeros'). This is only recommended 
+    for traces that are centered around zero (e.g. through a prior highpass
+    filter); if this is not the case, you can also fill the periods with noise.
+
+    Parameters
+    ----------
+    recording: RecordingExtractor
+        The recording extractor to remove artifacts from
+    list of lists/arrays
```
alejoe91, line 20:
```suggestion
    list_periods: list of lists/arrays
```

alejoe91:
@khl02007 maybe this approach will also work for you! So you can silence/remove specific frame ranges without relying on the `remove_artifacts` function :)

alejoe91:
fixes #619

Antho2422 (approved):

Antho2422 (approved):

PR #1255, opened 2023-02-02
Fix widget tests of quality metrics 

h-mayorquin:
The widgets tests are failing on master. I think I know where the bug is coming from. Testing here. 

Also, looking at the setup of the widgets tests, it seems that they depend heavily on post-processing and quality metrics. Changes on those modules should trigger the widgets tests. Adding that here to the CI.

### DIFF
--- a/spikeinterface/qualitymetrics/quality_metric_list.py
+++ b/spikeinterface/qualitymetrics/quality_metric_list.py
@@ -32,7 +32,7 @@ _misc_metric_name_to_func = {
     "firing_rate" : compute_firing_rates,
     "presence_ratio" : compute_presence_ratios,
     "snr" : compute_snrs,
-    "isi_violations" : compute_isi_violations,
+    "isi_violation" : compute_isi_violations,
     "rp_violation" : compute_refrac_period_violations,
     "sliding_rp_violation" : compute_sliding_rp_violations,
     "amplitude_cutoff" : compute_amplitude_cutoffs,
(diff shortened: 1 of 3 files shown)


alejoe91:
> The widgets tests are failing on master. I think I know where the bug is coming from. Testing here.
> 
> Also, looking at the setup of the widgets tests, it seems that they depend heavily on post-processing and quality metrics. Changes on those modules should trigger the widgets tests. Adding that here to the CI.

Gooood idea!

spikeinterface/qualitymetrics/quality_metric_list.py
```diff
@@ -32,7 +32,7 @@
     "firing_rate" : compute_firing_rates,
     "presence_ratio" : compute_presence_ratios,
     "snr" : compute_snrs,
-    "isi_violation" : compute_isi_violations,
+    "isi_violations" : compute_isi_violations,
```
alejoe91, line 35:
```suggestion
    "isi_violation" : compute_isi_violations,
```

Can you change instead this line: https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/widgets/tests/test_widgets.py#L51 ?

Names are more consistent now, i.e., QM names are always singular (while functions are plural)

alejoe91 (changes requested):
Please see comment ;)

spikeinterface/qualitymetrics/quality_metric_list.py
h-mayorquin: Done.

alejoe91 (approved):

PR #1258, opened 2023-02-02
Extend WaveformExtractor `rec_attributes` and propagate recordingless mode

### LINKED ISSUE #1217: Allow running metrics from waveform extractor in recording-less mode
Hi Alessio, Samuel and all,

I'm opening this issue to track progress regarding the possibility of running quality metrics from the waveform extractor alone (in mode "with_recording=False")

The rationale is that operations on the sorter output should be as self-contained as possible in general (decoupled from the raw recordings). Metrics might need be recomputed for specific analysis or to keep track with new metrics, and keeping the processed bin files alongside the sorter output (in order to be able to instantiate a recording) is in general too costly in terms of disk space, while relying on the raw data hinders data sharing and would require re-processing the whole recording every time we need to do some postprocessing.  

I see that currently, even `isi_violations` and  `compute_firing_rate` require the recording (only to pull the sampling frequency and total duration). `compute_snr` relies on the noise levels which could also in theory be obtained from the early samples of waveforms.

Amplitudes, locations, PCA components could ideally be computed only on the subset of extracted waveforms.

Best,
Tom

alejoe91:
fixes #1217

This PR extends the `rec_attributes` adding the following: `num_samples`, `is_filtered`, `properties` (excluding `contact_vector`).

The completely be independent of the underlying recording, the `WaveformExtractor` now has the following methods:
- `get_num_samples()`
- `get_total_samples()`
- `get_total_duration()`
- `is_filtered()`
- `get_recording_property()` (and `get_sorting_property`)

This allwo most modules to be compatible with recordingless modes:
- `curation` (which required `MockWavevorm` to extend `WaveformExtractor`)
- `postprocessing` (only `noise_levels`, `spike_locations`, `spike_amplitudes` and PCA `run_for_all_spikes` require recording
- `quality_metrics` (all metrics can be recomputed with previously computed `noise_levels`,  `spike_locations`, ans `spike_amplitudes`  extensions - only `nn_noise_overlap` requires recording)
- `exporters` (one can now export to Phy in recordingless mode - of course without the recording output)
- `widgets`

In addition, I removed some Deprecations and unused tests.

### DIFF
--- a/spikeinterface/core/waveform_extractor.py
+++ b/spikeinterface/core/waveform_extractor.py
@@ -72,6 +72,10 @@ class WaveformExtractor:
             for k in ('channel_ids', 'sampling_frequency', 'num_channels'):
                 if k not in rec_attributes:
                     raise ValueError(f'Missing key in rec_attributes {k}')
+            for k in ('num_samples', 'properties', 'is_filtered'):
+                if k not in rec_attributes:
+                    warn(f'Missing optional key in rec_attributes {k}: '
+                         f'some recordingless functions might not be available')
             self._rec_attributes = rec_attributes
         else:
             assert recording.get_num_segments() == sorting.get_num_segments(), \
@@ -237,6 +241,19 @@ class WaveformExtractor:
     def create(cls, recording, sorting, folder, mode="folder", remove_if_exists=False,
                use_relative_path=False, allow_unfiltered=False, sparsity=None):
         assert mode in ("folder", "memory")
+        # create rec_attributes
+        properties_to_attrs = deepcopy(recording._properties)
+        if 'contact_vector' in properties_to_attrs:
+            del properties_to_attrs['contact_vector']
+        rec_attributes = dict(
+            channel_ids=recording.channel_ids,
+            sampling_frequency=recording.get_sampling_frequency(),
+            num_channels=recording.get_num_channels(),
+            num_samples=[recording.get_num_samples(seg_index)
+                            for seg_index in range(recording.get_num_segments())],
+            is_filtered=recording.is_filtered(),
+            properties=properties_to_attrs
+        )
         if mode == "folder":
             folder = Path(folder)
             if folder.is_dir():
@@ -262,11 +279,6 @@ class WaveformExtractor:
             # dump some attributes of the recording for the mode with_recording=False at next load
             rec_attributes_file = folder / 'recording_info' / 'recording_attributes.json'
             rec_attributes_file.parent.mkdir()
-            rec_attributes = dict(
-                channel_ids=recording.channel_ids,
-                sampling_frequency=recording.get_sampling_frequency(),
-                num_channels=recording.get_num_channels(),
-            )
             rec_attributes_file.write_text(
                 json.dumps(check_json(rec_attributes), indent=4),
                 encoding='utf8'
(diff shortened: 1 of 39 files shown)


spikeinterface/core/waveform_extractor.py
```diff
@@ -237,6 +241,19 @@ def load_from_zarr(cls, folder, with_recording=True, sorting=None):
     def create(cls, recording, sorting, folder, mode="folder", remove_if_exists=False,
                use_relative_path=False, allow_unfiltered=False, sparsity=None):
         assert mode in ("folder", "memory")
+        # create rec_attributes
```
alejoe91, line 244: `rec_attributes` are always set!

spikeinterface/curation/auto_merge.py
```diff
@@ -392,10 +393,11 @@ def compute_templates_diff(sorting, templates, num_channels=5, num_shift=5, pair
...
 
-class MockWaveformExtractor:
-    def __init__(self, recording, sorting):
-        self.recording = recording
-        self.sorting = sorting
+class MockWaveformExtractor(WaveformExtractor):
+    def __init__(self, we, sorting):
+        WaveformExtractor.__init__(self, recording=None, sorting=sorting, folder=None,
+                                   rec_attributes=we._rec_attributes, allow_unfiltered=True,
+                                   sparsity=None)
```
samuelgarcia, line 400: I think this is better to not be a real WaveformExtractor but only to mock!
We should not put inheritance here.
I did this to be faster.
lets keep the MockWaveformExtractor independant until we have an intermediate.

spikeinterface/curation/auto_merge.py
```diff
@@ -430,7 +431,7 @@ def check_improve_contaminations_score(we, pair_mask, contaminations,
         unit_id1, unit_id2 = sorting.unit_ids[ind1], sorting.unit_ids[ind2]
         sorting_merged = MergeUnitsSorting(sorting, [unit_id1, unit_id2], new_unit_id=unit_id1).select_units([unit_id1])
         # make a lazy fake WaveformExtractor to compute contamination and firing rate
-        we_new = MockWaveformExtractor(recording, sorting_merged)
+        we_new = MockWaveformExtractor(we, sorting_merged)
```
samuelgarcia, line 434: This is weird.

spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -171,20 +164,27 @@ def compute_snrs(waveform_extractor, peak_sign: str = 'neg', peak_mode: str = "e
...
+        noise_levels = waveform_extractor.load_extension("noise_levels").get_data()
+    else:
+        assert waveform_extractor.has_recording(), \
+            ("SNR cannot be computed because waveform_extractor is recordingless and "
+             "'noise_levels' extension is not available")
+        if random_chunk_kwargs_dict is None:
+            random_chunk_kwargs_dict = {}
+        noise_levels = get_noise_levels(waveform_extractor.recording,
+                                        return_scaled=waveform_extractor.return_scaled,
+                                        **random_chunk_kwargs_dict)
```
samuelgarcia, line 177: Cool. I wanted to make a separate PR to do this at many palces.
Lets keep this here for now.

samuelgarcia (approved):

Issue #1235, opened 2023-01-31
Serialization project outline


h-mayorquin:
Hi @alejoe91 @samuelgarcia 
I want to propose here an order to work in a series of serialization issues, discuss and get your feedback and direction. I would like to work on this in the following order, each item being a tailored PR to keep the workflow modular and specific.  Let me know what you think of the order and priorities.

- [x] Adding serialization capabilities to recorders and sorters  #1227. 
- [x] Using the serialization capabilites to add initial instantiation capabilties to graph computation pipeline (#1209) that Sam developed #1271
- [x] I: Using the serialization capabilites to remove `to_dict` mechanism in the sorting components module #1272
- [x] II: Using the serialization capabilites to remove `to_dict` mechanism in the sorting components module #1325
- [x] III: Using the serialization capabilites to remove `to_dict` mechanism in the pre-processing module #1446
- [x] IV: Using the serialization capabilites to remove `to_dict` mechanism in the postprocessing module #1455
- [x] V: Using the serialization capabilites to remove `to_dict` mechanism in the curation module #1456
- [x] Using the same capabilities to avoid random and computation in preprocessing init class (solving issue #1118).
- [x] Added a JSON decoder to have a more standard structure #1397.
- [x] Separated flags for whether an extractor objects is dumpable to json and pickable https://github.com/SpikeInterface/spikeinterface/pull/1775
- [x] Separate serialization and json-nification infraestructure.

What do you think? If you agree with this proposal I would like to request a review of #1227 so I can start working on the other items.

alejoe91:
Hi Ramon!

It sounds like a good plan to me and I totally endorse it. @samuelgarcia what do you think?

samuelgarcia:
Cool.
Lets discuss thsi with a call soon.
Remove the to_dict in parralel processing is good.

Here some task are independant. I would make separate small PRs for fast discussion and merging.

h-mayorquin:
Yes, I am thinking on making a PR per point above (or maybe more for the later items). Let's agree on when to do a call at the end of today's meeting : ).

h-mayorquin:
Relevant for this project:
https://github.com/SpikeInterface/spikeinterface/pull/1831

alejoe91:
@h-mayorquin can we close this? I checked the last box after #2027

The last serialization issue (https://github.com/SpikeInterface/spikeinterface/issues/1118) is kind of standalone

h-mayorquin:
[...]

PR #1271, opened 2023-02-06
Refactor graph computation mechanism I

h-mayorquin:
After #1209 therer were still pending issues to solve. Namely:
1) Propagated parents to nodes at initilization (enabling the post_check) mechanism to be run at init and not to wait until the `ChunkRecordingExecutor`
2) Decide what to do about output returning (dict vs tuples vs something else).

This PR addresses the first which is also a point in #1235.
With the serialization machinery of #1227 we can pass the parents instances to a child node without concerns for serialization problems down the line. This is done in the PR and the examples and other functions have been adjusted to define the node pipeline with instances instead of names.  

I also refactored the tests in the `test_peak_pipeline` to have more meaningful names. 

I am thinking on adding a test to showcase the post_init mechanism as a chained PR here as this PR already is too big on itself.


samuelgarcia:
Hi Ramon.
I am OK with this.
This is super cool.
merci beaucoup.

@alejoe91 I propose to merge it before the release, so today. Can you check this ?

samuelgarcia (approved):

PR #1272, opened 2023-02-06
Remove `to_dict` mechanism for parallel processing in sorting components module

h-mayorquin:
As described in #1235 and enabled by #1227 we are planning on getting rid of the `the_dict` mechanism for parallel processing. The advantages of this are that we don't need to build `to_dict` functionality for every new class containing a recording that we define. That is, developing should be faster and less burdensome.

That said, that is a big change so I am here making a smaller PR that only removes this on the sorting components as a testing ground. The idea would be to triage the fuctionality in this module and propagate to the whole repo if everything seems to going all right after a while.

This together with #1271 should make the early stages of the sorting components pipeline easier to develop and mantain.


samuelgarcia:
Hi Ramon,
thanks for this.
My guess is that we have some other place in the code that could be changed with the pickle approach.

Can you search for `ChunkRecordingExecutor` everywhere and look around if some code could be refactor also ?

samuelgarcia (approved):

h-mayorquin:
> 

@samuelgarcia 
Yes, this is on my to-do list.

PR #1278, opened 2023-02-07
Allow unfiltered wvf_extractor in export_to_phy

DradeAW:
Allow export_to_phy to export an unfiltered wvf_extractor if it has the 'allow_unfiltered' attribute set to True.


samuelgarcia:
Salut Aurelien.
I am not sure to like this as an attribute because we should make it dumpable and loadable for binary and zarr format also.

Maybe having the flat in export_to_phy would be easier. No ?

@alejoe91 : what do you think ?

alejoe91:
This only triggers a warning, no?

DradeAW:
> This only triggers a warning, no?

Yes, but I hate having warnings when I know that what's running is what I want ^^


> I am not sure to like this as an attribute because we should make it dumpable and loadable for binary and zarr format also.
> Maybe having the flat in export_to_phy would be easier. No ?

I did it this way to avoid having an extra parameter in `export_to_phy`.
But if you don't think this is the best way to implement it we can discuss it!

I mean since the `WaveformExtractor` already checks if the recording is filtered or not, is it really necessary to also have it in `export_to_phy`?

alejoe91 (approved):

PR #1279, opened 2023-02-07
Add `PlexonSortingExtractor` for plexon spiking data

weiglszonja:
Add extractor for plexon spiking data based on `PlexonRawIO`, needs `file_path` the file path to plexon spiking data (.plx format) and  `sampling_frequency` the sampling frequency for the spiking channels.


CodyCBakerPhD:
@weiglszonja Your class setup is the same as the [NeuralynxSortingExtractor](https://github.com/catalystneuro/spikeinterface/blob/add_plexon_sorting_extractor/spikeinterface/extractors/neoextractors/neuralynx.py#L60-L63) so does it work if you just mimic the [test for that class](https://github.com/catalystneuro/spikeinterface/blob/add_plexon_sorting_extractor/spikeinterface/extractors/tests/test_neoextractors.py#L148-L154)?

weiglszonja:
Without specifying the `sampling_frequency` it will try to auto guess sampling frequency, however the sorting data does not have "signal_channels", so it will error out. The test file for the [NeuralynxSortingExtractor](https://github.com/catalystneuro/spikeinterface/blob/add_plexon_sorting_extractor/spikeinterface/extractors/neoextractors/neuralynx.py#L60-L63) has "signal_channels". 

The plexon file for the Tye Lab is similar when reading with PlexonRawIO, it does not have "signal_channels" either, only "spike_channels". (this is why I chose that test file)

```
../neoextractors/neobaseextractor.py:202: in __init__
    sampling_frequency, stream_id = self._auto_guess_sampling_frequency()
        sig_channels = self.neo_reader.header['signal_channels']
>       assert sig_channels.size > 0, 'sampling_frequency could not be inferred from the signals, set it manually'
E       AssertionError: sampling_frequency could not be inferred from the signals, set it manually
```

samuelgarcia:
Hi Szonja.
Indeed, for some neo sorting extractor that do not have attached traces there is not standard way to auto guess the sampling rate.
But I guess that the information must be somewhere in the neo class so we can make a specific little hack in this case. This would be more convinient. I will check.

alejoe91:
@weiglszonja I added a fix to automatically fetch the sampling rate from NEO :)

Can we proceed to final review or is it still a draft?

weiglszonja:
Ah thank you @samuelgarcia and @alejoe91! I think it's ready for review.

alejoe91 (approved):

PR #1282, opened 2023-02-07
Add `SingleChannelToyDenoiser` with node pipeline

h-mayorquin:
Porting of #1219 to the new pipeline mechanism. This uses the mechanism implemented in #1271 to run a check on intialization for a waveform extraction step on the parents.  Note that this PR has #1271 as its base and requires it to work.


pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
     "threadpoolctl",
     "tqdm",
     "probeinterface>=0.2.16",
+    "huggingface_hub",
```
alejoe91, line 28: Let's move this to the `full` installation. The base installation is only for the base SI classes

h-mayorquin: This makes sense.

spikeinterface/sortingcomponents/waveforms/neural_network_denoiser.py
```diff
@@ -0,0 +1,116 @@
...
+        Asserts that the model and the waveform extractor have the same temporal parameters
+        """
+        # Extract temporal parameters from the waveform extractor
+        waveforms_ms_before = waveform_extractor.ms_before
+        waveforms_ms_after = waveform_extractor.ms_after
+        waveforms_sampling_frequency = waveform_extractor.recording.get_sampling_frequency()
+
+        # Load the model temporal parameters
+        repo_id = "SpikeInterface/test_repo"
+        filename = "toy_model_peak_interval.json"
```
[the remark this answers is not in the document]
h-mayorquin: My thinking was that this is used for the specic temporal parameters that determine whether the model matches with the waveform extractor (sampling_frequency, ms_before, ms_after, etc). In opposition to the structural parameters and weights are determined in the other pickled file. A similar story with the TemporalPCA. That is, it is not all the parameters but the temporal matching ones.

From that perspective I think that `params.json` is not specific enough. If want to introduce, for other models that use other set of parameters (not just structural and temporal matching like PCA and the CNN here) we would create an unecessary ambiguity. 

I think that something minimal like `temporal_parameters.json` or something of the sort is better.

alejoe91:
@h-mayorquin regarding 

IMO it's ok that a processing waveforms node is used as input to the tpca denoising (or NN denoising). What we should do is to make the `SingleChannelToyDenoiser` and `TemporalPCADenoising` extend also from the [`WaveformExtractorNode`](https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/sortingcomponents/peak_pipeline.py#L55). This could also be actually a good approach to force propagation of the `ms_before` and `ms_after` attributes.

h-mayorquin:
> @h-mayorquin regarding
> 
> IMO it's ok that a processing waveforms node is used as input to the tpca denoising (or NN denoising). What we should do is to make the `SingleChannelToyDenoiser` and `TemporalPCADenoising` extend also from the [`WaveformExtractorNode`](https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/sortingcomponents/peak_pipeline.py#L55). This could also be actually a good approach to force propagation of the `ms_before` and `ms_after` attributes.

@alejoe91 
I am not sure if this was pointing to something, maybe to the suggestion on the other thread about a single node vs multiple node.

Anyway, as I am working on this I am thinking we need two clasess: ``WaveformExtractorNode`` and ``WaveformProcessingNode`. The former, is useful here to check if there is an extractor node in the parents (we can check by instance) whereas the second is useful to have the general logic for matching temporal attributes (ms_before, ms_after,e tc) and for inherting as you mention. If we merge these two concepts we can't do the first instance check to assert that waveforms are extracted in the parents.

I kind of don't like the profileration of clasess though but I think this is better right now than disambiguating extractor from processing with internal class parameters.

pyproject.toml
h-mayorquin: Done.

spikeinterface/sortingcomponents/waveforms/neural_network_denoiser.py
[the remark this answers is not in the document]
alejoe91: I disagree here @h-mayorquin 

In this case the params might only have temporal params info, but what if a model is built on some geometry or other parameters? In that case we would extend this params name, e.g. `temporal_geometry_local_radius.json`?

In my opinion it's better to unify all names for models as `params.json`, so each model folder will have a `model.*` (.pt, .pkl) and a `params.json`

spikeinterface/sortingcomponents/waveforms/neural_network_denoiser.py
```diff
@@ -0,0 +1,116 @@
...
+        # Denoise
+        denoised_temporal_waveforms = self.denoiser(temporal_waveforms_tensor).detach().numpy()
+
+        # Reconstruct representation with channels
+        desnoised_waveforms = from_temporal_representation(denoised_temporal_waveforms, num_channels)
+
+        return desnoised_waveforms
+
+
+class PaninskiSingleChannelDenoiser(nn.Module):
```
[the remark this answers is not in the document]
alejoe91: I would say that we can name models as they become available. Maybe something like `SingleChannel1dCNNDenoiser` is ok for now. Note that the `params.json` (or whatever we end up calling it) could also have the model class name, so several models could co-exist.

spikeinterface/sortingcomponents/waveforms/neural_network_denoiser.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def forward(self, x):
+        x = x[:, None]
+        x = self.conv1(x)
+        x = self.conv2(x)
+        x = x.view(x.shape[0], -1)
+        x = self.out(x)
+        return x
+
+    def load(self, device="cpu"):
```
alejoe91, line 113:
```suggestion
    def load(self, device="cuda"):
```
I would make `cuda` default, and check if a GPU is available (in that case we need to cast the waveforms as `torch.tensor`s)

spikeinterface/sortingcomponents/waveforms/neural_network_denoiser.py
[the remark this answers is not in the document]
h-mayorquin: I see how you view it. All the parameters, not only temporal, should go in that file. 

I was not thinking on this because for the neural network models the network parameters are in the pickle file so we already have two parameters files now. From there, I thought we would have models with 3, 4, and so on so I wanted to be specific. Maybe we won't require more than 2 and your approach makes more sense.

So, let's stick with the unification -all parameters in that file- and split later if it makes sense.

spikeinterface/sortingcomponents/waveforms/neural_network_denoiser.py
h-mayorquin: This can be done but I will note the following: 

Our approach of dividing work across chunks guarantess that there is a lot of IO operations which is basically the domain where GPUs are not that great. Moving data is costly. I am not sure this is going to make it better and it implies a little bit of more machinery to the function -which is not that much-.

I guess we should profile this.

pyproject.toml
h-mayorquin: @samuelgarcia Here.

spikeinterface/sortingcomponents/waveforms/neural_network_denoiser.py
h-mayorquin: Done in the last commit.

spikeinterface/sortingcomponents/waveforms/neural_network_denoiser.py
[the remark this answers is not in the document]
h-mayorquin: Done in the last commit.

spikeinterface/sortingcomponents/waveforms/neural_network_denoiser.py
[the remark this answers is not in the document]
h-mayorquin: Done in the last commit.

spikeinterface/sortingcomponents/waveforms/neural_network_denoiser.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def forward(self, x):
+        x = x[:, None]
+        x = self.conv1(x)
+        x = self.conv2(x)
+        x = x.view(x.shape[0], -1)
+        x = self.out(x)
+        return x
+
+    def load(self, device="cuda"):
```
alejoe91, line 115: What happens if cuda is not available? Does it automatically use CPU instead?

alejoe91: Apparently not ;)

https://github.com/SpikeInterface/spikeinterface/actions/runs/4353692677/jobs/7608059335

h-mayorquin: All right. I have a gpu card on my local machine so I did not catch this one. 

I propose that we delay this till we have some robust mechanism to test whether gpu capabilities are available:
https://github.com/SpikeInterface/spikeinterface/pull/1398

What do you think?

alejoe91: Ok for me ;)

h-mayorquin: Cool. Done, this should be ready to go.

alejoe91: @h-mayorquin let's keep it in tests for now, since it's still quite experimental. Is that ok?

alejoe91 (approved):

samuelgarcia (approved):

PR #1297, opened 2023-02-09
Produce timing report for core tests

h-mayorquin:
As @samuelgarcia mentioned our tests take too long to run. I am testing producing a summary so we can keep a public log of the longest tests. The idea is that of "if you want a problem measure it". 

I am testing now this for the core tests and will propagate to the other tests afterwards if it works.


alejoe91:
Awesome Ramon!!! Thank you! :)

alejoe91 (approved):

PR #1300, opened 2023-02-09
High level function for drift : correct_motion()

samuelgarcia:
This implement a high level function to estimate and interpolate dirft.

* [x] correct_motion()
* [x] test
* [x] documentation


yger:
There is a typo in the function name. Plus, I think it would make sense to have a default peak number selected, as we discussed, that would depend on the duration of the recording. Do you really want the default params for select peaks to be None?

samuelgarcia:
Salut Pierre.
oui oui bien sûr. C'est juste un premier jet.

I am also thinkink of having some "preset" option.
the preset would choose default dict that make for the user a combinaison of parameters to mimic KS2.5 or the paninski pipeline.

src/spikeinterface/preprocessing/motion.py
```diff
@@ -0,0 +1,350 @@
+import time
+from pathlib import Path
+
+import numpy as np
+import json
+import copy
+
+from spikeinterface.core import get_noise_levels, fix_job_kwargs
+from spikeinterface.core.core_tools import SIJsonEncoder
+
```
samuelgarcia, line 10: @cwindolf @yger 
I have started some preset of parameters for the a high level function.
Could you have look and make some feedback ?

src/spikeinterface/preprocessing/motion.py
```diff
@@ -0,0 +1,350 @@
...
+        "estimate_motion_kwargs": dict(
+            method="decentralized",
+            direction="y",
+            bin_duration_s=2.0,
+            rigid=False,
+            bin_um=5.0,
+            margin_um=0.0,
+            win_shape="gaussian",
+            win_step_um=50.0,
+            win_sigma_um=150.0,
```
cwindolf, line 39: this seems like a lot of very narrow windows, I would expect very noisy results in real data. I usually use more like win_step_um=400.0, win_sigma_um=800.0 (I think that the window sigma should be divided by 2 in the gaussian case, since gaussian kernels really extend out to 2 or 3 std devs, so this would match the behavior of rect windows more closely)

cwindolf: ^ what I have written would currently lead to an error because window sigma cannot be < window step, but I think these params are the way to go and that that restriction should be removed -- been meaning to make a PR about that

src/spikeinterface/sortingcomponents/motion_interpolation.py
```diff
@@ -125,7 +131,7 @@ def interpolate_motion_on_traces(
     # inperpolation kernel will be the same per temporal bin
     for bin_ind in np.unique(bin_inds):
         # Step 1 : channel motion
-        if spatial_bins is None:
+        if spatial_bins.shape[0] == 0:
```
cwindolf, line 134: nice! happy this is here now

src/spikeinterface/sortingcomponents/motion_interpolation.py
```diff
@@ -17,7 +17,13 @@
 
 
 def correct_motion_on_peaks(
-    peaks, peak_locations, times, motion, temporal_bins, spatial_bins, direction="y", progress_bar=False
+    peaks,
+    peak_locations,
+    times,
+    motion,
+    temporal_bins,
```
cwindolf, line 24: it would be useful to clarify whether these are bin edges or bin centers

src/spikeinterface/sortingcomponents/motion_interpolation.py
cwindolf: same for traces?

src/spikeinterface/sortingcomponents/motion_interpolation.py
cwindolf: I think it would be nice if estimate_motion returned an object so that this function could take fewer arguments, and to avoid replicating the logic about temporal bins and motion interpolation in several places (here in peaks and in traces and anywhere else a user may want to get displacement at a particular depth and time)

maybe a class like this one? https://github.com/evarol/dredge/blob/main/dredge-python/dredge/motion_util.py#L22

src/spikeinterface/sortingcomponents/motion_interpolation.py
samuelgarcia: what do you mean ?

src/spikeinterface/sortingcomponents/motion_interpolation.py
samuelgarcia: This is bin center this is in the docstring no ?
I am agree for the object but this will be in the next round of modification that we will to togoether no ?

src/spikeinterface/preprocessing/motion.py
samuelgarcia: For win_step_um should not change the results lead to some kind of coninuum in space that avoid linear interpolation. I can be win_sigma_um=200 but you are suggesting to put 100 no ?
I am not sure to follow you.

cwindolf: sorry, I was confusing!

yeah, I usually use win_sigma_um=200 or 250 in real data

and then the other thing I was saying, which is a separate issue, is that I think it could be nice to reparameterize the gaussian window size by a parameter which is 2 * win_sigma_um, which would make the width of the windows match the rect windows more closely

alejoe91 (approved):

PR #1303, opened 2023-02-10
Run code coverage daily as a cron job

h-mayorquin:
Looking at the code coverage site it seems that we have too few data points. That is, we don't know how the code coverage is changing across time. To address that, I am making the code coverage workflow a cronjob so we keep regular sampling of the state of code coverage in the repo. 

Note that the action remains dispatchable as it was before. Plus, I added profiling to the full tests as I have done for core in #1297 and for the full normal tests in #1299.


samuelgarcia:
Daily is maybe too much, no ?
there are some power consumption behind!

h-mayorquin:
@samuelgarcia I think that compare to our normal and later use when we run more than 10 times the tests per day this is a small contribution. Plus, my impression is that this is rather a small footprint compared to other CIs. Finaly, I think that the computation here would be more than compensated if we do something along the lines of #1304.

samuelgarcia:
Yes but the codecov is more than one hour no ? This looks enormous for a everyday action.

alejoe91:
maybe weekly?

h-mayorquin:
Up to you guys, my vote is for as close as daily as it can be. Let me know what's your decision and I will change the script. 
The arguments for daily frequency are: 1) to keep a good tracking of code coverage,  2) detecting if there are any errors that the normal tests are not catching up, and 3) having a fine grained historical record of the timing benchmarks for the tests.

In any case, maybe we can run it daily for 2-3 days till I check that it works as expected and then we can revert to whatever period you chose? Just to ensure the thing works as they should and fix them fast if they don't.

Some facts concerning power. 
Relative costs. I think the relationship between the tests that we run in a normal PR and the full tests once daily are not that different. The former is like 10-20 min whereas the latter is like 30-60. We run the former more than 3 times per day in average I reckon.

Absolute costs. Doing some [googling](https://devblogs.microsoft.com/sustainable-software/how-can-i-calculate-co2eq-emissions-for-my-azure-vm/) get us a cost for running virtual machine at around ~ 4.3 / 24 = 0.17 kwH. This is lower than most house appliances and larger than having a second display. 

Note that these facts can be used to push the argument for any side depening on some background assumptions. I would be interesting to hear how do you think on this topic at some point when we (hopefully!) have some more time to discuss.

alejoe91:
> Up to you guys, my vote is for as close as daily as it can be. Let me know what's your decision and I will change the script. The arguments for daily frequency are: 1) to keep a good tracking of code coverage, 2) detecting if there are any errors that the normal tests are not catching up, and 3) having a fine grained historical record of the timing benchmarks for the tests.
> 
> In any case, maybe we can run it daily for 2-3 days till I check that it works as expected and then we can revert to whatever period you chose? Just to ensure the thing works as they should and fix them fast if they don't.
> 
> Some facts concerning power. Relative costs. I think the relationship between the tests that we run in a normal PR and the full tests once daily are not that different. The former is like 10-20 min whereas the latter is like 30-60. We run the former more than 3 times per day in average I reckon.
> 
> Absolute costs. Doing some [googling](https://devblogs.microsoft.com/sustainable-software/how-can-i-calculate-co2eq-emissions-for-my-azure-vm/) get us a cost for running virtual machine at around ~ 4.3 / 24 = 0.17 kwH. This is lower than most house appliances and larger than having a second display.
> 
> Note that these facts can be used to push the argument for any side depening on some background assumptions. I would be interesting to hear how do you think on this topic at some point when we (hopefully!) have some more time to discuss.

Thanks for putting together some numbers here. I agree that this would be a minimal overhead in terms of costs and resources, especially after the new caching mechanism.

Merging!

alejoe91 (approved):

PR #1304, opened 2023-02-10
Cancel previous github actions jobs when pull request is updated

h-mayorquin:
This PR modifies the github actions to save energy / computation. How is this? Right now when you push updates to an open PR and the tests are still running those tests don't stop, they just run on the background. This PR cancels those previuosly running tests using the concurrency functionality of github actions when a new PR is push too quickly.

https://docs.github.com/en/actions/using-jobs/using-concurrency

This should save some energy instead of having the jobs running on the background. 

P.D. Right now, only the full tests have the functionality. I will add the same to core immediately after this just for testing if cancelling is working.

EDIT: great, it seems that is working. I will also propagate this to the NWB streaming test.


alejoe91 (approved):

h-mayorquin:
This is not ready yet, it seems to be cancelling the other workflows. That is, it only allows only workflow to run at a time. Let me fix this.

h-mayorquin:
All right, this is working now with the right grouping level.

PR #1310, opened 2023-02-13
Silence interval

yger:
My PR was removed during the release, so I'll bring it back just not to forget about it...


yger:
Actually @samuelgarcia I know you want to refactor that and merge it with remove_artefact, but I just discussed with people that would like to use such a feature, and not sure it will be doable. They even would like to be able to pass, for every period, the channels that should be blanked/zeros, so I'm afraid that the signature for a single function will start to be a nightmare. Why not simple add this one, and we could easily also add a list of channels to be considered for every period. Up to you

spikeinterface/preprocessing/silence_periods.py
```diff
@@ -0,0 +1,118 @@
...
+                   basis
+
+    Returns
+    -------
+    silence_recording: SilencedPeriodsRecording
+        The recording extractor after silencing some periods    
+    """
+    name = 'silence_periods'
+
+    def __init__(self, recording, list_periods, mode='zeros', **random_chunk_kwargs):
```
alejoe91, line 39: Can you add explicitly the 3 arguments for the random chunk finction? see https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/preprocessing/whiten.py#L18-L23

spikeinterface/preprocessing/silence_periods.py
```diff
@@ -0,0 +1,118 @@
...
+
+    mode: str
+        Determines what periods are replaced by. Can be one of the following:
+            
+        - 'zeros' (default): Artifacts are replaced by zeros.
+
+        - 'noise': The periods are filled with a gaussion noise that has the
+                   same variance that the one in the recordings, on a per channel
+                   basis
+
```
alejoe91, line 31: Add args for random chunk extraction: https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/preprocessing/whiten.py#L18-L23

spikeinterface/preprocessing/silence_periods.py
```diff
@@ -0,0 +1,118 @@
...
+            if isinstance(list_periods, (list, np.ndarray)) and not np.isscalar(list_periods[0]):
+                # when unique segment accept list instead of of list of list/arrays
+                list_periods = [list_periods]
+
+        # some checks
+        assert isinstance(list_periods, list), "'list_periods' must be a list (one per segment)"
+        assert len(list_periods) == num_seg, "'list_periods' must have the same length as the number of segments"
+        assert all(isinstance(list_periods[i], (list, np.ndarray)) for i in range(num_seg)), \
+            "Each element of 'list_periods' must be array-like"
+
```
alejoe91, line 55: Maybe add a check for non-overlapping intervals?

spikeinterface/preprocessing/silence_periods.py
```diff
@@ -0,0 +1,118 @@
...
+        self.periods = np.asarray(periods, dtype='int64')
+        self.periods = np.sort(self.periods, axis=0)
+        self.mode = mode
+        self.noise_levels = noise_levels
+
+    def get_traces(self, start_frame, end_frame, channel_indices):
+
+        traces = self.parent_recording_segment.get_traces(start_frame, end_frame, channel_indices)
+        traces = traces.copy()
+        nb_channels = traces.shape[1]
```
alejoe91, line 88:
```suggestion
        num_channels = traces.shape[1]
```

spikeinterface/preprocessing/tests/test_silence.py
```diff
@@ -0,0 +1,43 @@
...
+def test_silence():
+    rec = generate_recording()
+
+    rec0 = silence_periods(rec, list_periods=[[[0, 1000], [5000, 6000]], []], mode="zeros")
+    rec0.save(verbose=False)
+
+    traces0 = rec0.get_traces(segment_index=0, start_frame=0, end_frame=1000)
+    traces1 = rec0.get_traces(segment_index=0, start_frame=5000, end_frame=6000)
+    assert np.all(traces0 == 0) and np.all(traces1 == 0)
+
```
alejoe91, line 32: Can you add a test that retrieves traces between 900 and 5100?

alejoe91 (changes requested):
Thank @yger 

I'm ok in having this new functionality and I agree that adding this option to the remove artifact might be a bit cumbersome in terms of signature. I have a few comments

spikeinterface/preprocessing/tests/test_silence.py
yger: done

spikeinterface/preprocessing/silence_periods.py
yger: done

spikeinterface/preprocessing/silence_periods.py
yger: done

alejoe91:
@yger looks good to me. @samuelgarcia is it fine for you?

samuelgarcia:
@yger : I think we need a seek functionality for the noise no ?

alejoe91:
Hi @yger 

@samuelgarcia suggested that both this function and the remove_artifact live in the same file and use the same `Segment` object. In both cases, the segment could take the intervals to silence or remove artifacts for (the median/average would only be supported by the `rmeove_artifacts` function, because it ensures that intervals are always the same size). What do you think?

yger:
I could, but I think this will complexify everything. The silence_periods is rather simple and straighforward (for now). As said, a natural extension for such a function would be to also provide a list of channels to silence, for every intervals. On the other hand, the remove_artefacts have a more intense machinery if you want to subtract artefact with an optimal lag/amplitude. Thus, combining the two is I guess adding useless complexity...

alejoe91:
> I could, but I think this will complexify everything. The silence_periods is rather simple and straighforward (for now). As said, a natural extension for such a function would be to also provide a list of channels to silence, for every intervals. On the other hand, the remove_artefacts have a more intense machinery if you want to subtract artefact with an optimal lag/amplitude. Thus, combining the two is I guess adding useless complexity...

That's ok for me. I agree that the median/average subtraction is very specific... @samuelgarcia I'm in favor to keep this as a simple preprocessing step for now. 

We can do the integration (of we want) later. Maybe the best would be to have two segments classes:
1. one for zero/noise/linear/cubic interpolation (which can be shared by the 2 preprocessors)
2. one for median/average subtraction (specific to the `remove_artifacts`)

No?

spikeinterface/preprocessing/silence_periods.py
```diff
@@ -0,0 +1,132 @@
...
+        # some checks
+        assert isinstance(list_periods, list), "'list_periods' must be a list (one per segment)"
+        assert len(list_periods) == num_seg, "'list_periods' must have the same length as the number of segments"
+        assert all(isinstance(list_periods[i], (list, np.ndarray)) for i in range(num_seg)), \
+            "Each element of 'list_periods' must be array-like"
+
+        for periods in list_periods:
+            if len(periods) > 0:
+                assert np.all(np.diff(np.array(periods), axis=1) > 0), "t_stops should be larger than t_starts"
+                assert all(e < s for (_, e), (s, _) in zip(periods, periods[1:])), "Intervals should not overlap"
```
samuelgarcia, line 66: THis should be done in the numpy style. No ?

spikeinterface/preprocessing/silence_periods.py
```diff
@@ -0,0 +1,132 @@
...
+
+                for i in intersection:
+
+                    onset = max(0, i[0] - start_frame)
+                    offset = min(i[1] - start_frame, end_frame)
+
+                    if self.mode == 'zeros':
+                        traces[onset:offset, :] = 0
+                    elif self.mode == 'noise':
+                        traces[onset:offset, :] = self.noise_levels[channel_indices] * np.random.randn(offset-onset, num_channels)
```
samuelgarcia, line 126: I think we need a seed somewhere to be able to reproduce the same signals.
The seed could based a mixing aseed number + the segment index + period index.

spikeinterface/preprocessing/silence_periods.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class SilencedPeriodsRecordingSegment(BasePreprocessorSegment):
+
+    def __init__(self, parent_recording_segment, periods, mode, noise_levels):
+
+        BasePreprocessorSegment.__init__(self, parent_recording_segment)
+
+        self.periods = np.asarray(periods, dtype='int64')
+        self.periods = np.sort(self.periods, axis=0)
```
samuelgarcia, line 94: sort have a high cost when mutiprocessing! This should be externally or check at init.

spikeinterface/preprocessing/silence_periods.py
```diff
@@ -0,0 +1,132 @@
...
+            if len(periods) > 0:
+                assert np.all(np.diff(np.array(periods), axis=1) > 0), "t_stops should be larger than t_starts"
+                assert all(e < s for (_, e), (s, _) in zip(periods, periods[1:])), "Intervals should not overlap"
+
+        assert mode in available_modes, f"mode {mode} is not an available mode: {available_modes}"
+
+        if mode in ['noise']:
+            noise_levels = get_noise_levels(recording, return_scaled=False, 
+                                    num_chunks_per_segment=num_chunks_per_segment,
+                                    chunk_size=chunk_size, concatenated=True, seed=seed)
```
samuelgarcia, line 73: For reproducability we should be able to exterbnally give the noise.

spikeinterface/preprocessing/silence_periods.py
```diff
@@ -0,0 +1,132 @@
...
+    for traces that are centered around zero (e.g. through a prior highpass
+    filter); if this is not the case, you can also fill the periods with noise.
+
+    Parameters
+    ----------
+    recording: RecordingExtractor
+        The recording extractor to remove artifacts from
+    list_periods: list of lists/arrays
+        One list per segment of tuples (start_frame, end_frame) to silence
+
```
samuelgarcia, line 22: I think that period should be a list of numpy arrays always for efficiency.

spikeinterface/preprocessing/silence_periods.py
yger: Done

spikeinterface/preprocessing/silence_periods.py
yger: I reuse the same syntax as in remove_artifacts. It can be arrays or list. You want to have a list of arrays?

spikeinterface/preprocessing/silence_periods.py
yger: Externally give the noise? The seed is not enough?

yger:
@samuelgarcia @alejoe91 I think this should be merged, because at this stage, modifications might be too picky. I don't think that we should control the seed of the noise injected into blank signal (we could, but really, what is the point...). Plus regarding the signatures list/arrays, I just tried to be consistent with remove_artefacts. But just my 2 cents...

yger:
@alejoe91 , you need to approve your review also btw

alejoe91:
> @alejoe91 , you need to approve your review also btw

Fixing errors and going through a final review ;)

spikeinterface/preprocessing/silence_periods.py
```diff
@@ -0,0 +1,132 @@
...
+        if start_frame is None:
+            start_frame = 0
+        if end_frame is None:
+            end_frame = self.get_num_samples()
+
+        if len(self.periods) > 0:
+            new_interval = np.array([start_frame, end_frame])
+            lower_index = np.searchsorted(self.periods[:,1], new_interval[0])
+            upper_index = np.searchsorted(self.periods[:,0], new_interval[1])
+
```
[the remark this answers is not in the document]
alejoe91: no that's ok: if end_frame is included, then the period is used (and margins are taken care of) 

it's also properly tested

spikeinterface/preprocessing/silence_periods.py
alejoe91: done

spikeinterface/preprocessing/silence_periods.py
alejoe91: I unified with other functions by passin `**random_chunk_kwargs`

alejoe91 (approved):

PR #1315, opened 2023-02-14
Sparsity threshold based on expected energy for noise

yger:
Addition of a new sparsity methods, that sparsifies the channels as function of the ratio between the expected energy that should be present, assuming the noise levels and the number of waveforms sampled (if they were all random), and the one measures.


spikeinterface/core/sparsity.py
```diff
@@ -193,6 +195,21 @@ def from_threshold(cls, we, threshold, peak_sign='neg'):
             mask[unit_ind, chan_inds] = True
         return cls(mask, we.unit_ids, we.channel_ids)
 
+    @classmethod
+    def from_energy(cls, we, threshold):
+        """
+        Construct sparsity from a threshold based on per channel energy ratio.
+        Use the 'threshold' argument to specify the SNR threshold.
```
alejoe91, line 202:
```suggestion
        Use the 'threshold' argument to specify the energy threshold.
```

spikeinterface/core/sparsity.py
```diff
@@ -8,8 +8,10 @@
         * "best_channels": N best channels with the largest amplitude. Use the 'num_channels' argument to specify the
                          number of channels.
         * "radius": radius around the best channel. Use the 'radius_um' argument to specify the radius in um
-        * "threshold": thresholds based on template signal-to-noise ratio. Use the 'threshold' argument
+        * "threshold": threshold based on template signal-to-noise ratio. Use the 'threshold' argument
                        to specify the SNR threshold.
+        * "energy": threshold based on the expected energy that should be present on the channels, 
+                    given their noise levels. Use the 'threshold' argument to specify the SNR threshold
```
alejoe91, line 14: @yger is this threshold absolute or relative (similar to SNR)? And what would be a good value? Is it a different way to compute SNR?

yger: The threshold is relative. 1 means that you have the energy expected if all the waveforms on a given channel were all pure gaussian noise at the variance of the one from the channel. 

If you set a threshold slightly higher than 1, it means that you will keep only the channels where you you should have a signal (in energy) that strongly differ from noise. Is it clear? This is an attempt, to be maybe more robust that with just a single peak value.

spikeinterface/core/sparsity.py
```diff
@@ -8,8 +8,10 @@
         * "best_channels": N best channels with the largest amplitude. Use the 'num_channels' argument to specify the
                          number of channels.
         * "radius": radius around the best channel. Use the 'radius_um' argument to specify the radius in um
-        * "threshold": thresholds based on template signal-to-noise ratio. Use the 'threshold' argument
+        * "threshold": threshold based on template signal-to-noise ratio. Use the 'threshold' argument
```
alejoe91, line 11:
```suggestion
        * "snr": threshold based on template signal-to-noise ratio. Use the 'threshold' argument
```

alejoe91 (approved):

PR #1318, opened 2023-02-14
Add optional selection indices to ``select_peaks`` function

### LINKED ISSUE #1283: Add selection indices as an output of select_peaks
Currently `sortingcomponents.peak_selection.select_peaks(input_peaks, ...)` outputs a structured numpy array `peaks` which is a 'smart' subset of the original `input_peaks` object. 

Frequently, we have to perform some costly operations, such as monopolar triangulation, on the originally detected `input_peaks` since we need peak locations for some of the peak selection algorithms.  But we may also need the results of these operations for each of the selected peaks, since for example, we need peak locations for some of the clustering algorithms.  To achieve common use-case, we would currently either have to
A) Recompute the peak locations of the selected peaks, which is costly for a large number of selected peaks
B) Determine the index of each selected peak in the original `input_peaks` object, which is also costly for large numbers of selected peaks and/or input peaks

To minimize unnecessary computation, I propose changing the output of `sortingcomponents.peak_selection.select_peaks` to include the selected indices so users can easily index prior computations.  Ex.
```
peaks = detect_peaks(...)
peak_locations = localize_peaks(peaks, ...)
selected_peaks, selected_indices = select_peaks(peaks, peak_locations, ...)
selected_locations = peak_locations[selected_indices]

assert selected_locations == localize_peaks(selected_peaks, ...)
```

Finally, I want to point out that one could imagine many different operations that could be useful for peak selection AND subsequent clustering (ex. what if you want to sample fast-spiking and regular-spiking neurons evenly, so you compute peak width).  So this improvement would be more general than this specific use case.

pauladkisson:
Fixes #1283

### DIFF
--- a/spikeinterface/sortingcomponents/peak_selection.py
+++ b/spikeinterface/sortingcomponents/peak_selection.py
@@ -6,27 +6,26 @@ from sklearn.preprocessing import QuantileTransformer
 
 
 def select_peaks(peaks, method='uniform', seed=None, return_indices=False, **method_kwargs):
-
     """Method to subsample all the found peaks before clustering
     Parameters
     ----------
     peaks: the peaks that have been found
-    method: 'uniform', 'uniform_locations', 'smart_sampling_amplitudes', 'smart_sampling_locations', 
+    method: 'uniform', 'uniform_locations', 'smart_sampling_amplitudes', 'smart_sampling_locations',
     'smart_sampling_locations_and_time'
         Method to use. Options:
             * 'uniform': a random subset is selected from all the peaks, on a per channel basis by default
-            * 'smart_sampling_amplitudes': peaks are selected via monte-carlo rejection probabilities 
+            * 'smart_sampling_amplitudes': peaks are selected via monte-carlo rejection probabilities
                 based on peak amplitudes, on a per channel basis
-            * 'smart_sampling_locations': peaks are selection via monte-carlo rejections probabilities 
-                based on peak locations, on a per area region basis
-            * 'smart_sampling_locations_and_time': peaks are selection via monte-carlo rejections probabilities 
+            * 'smart_sampling_locations': peaks are selection via monte-carlo rejections probabilities
+                based on peak locations, on a per area region basis-
+            * 'smart_sampling_locations_and_time': peaks are selection via monte-carlo rejections probabilities
                 based on peak locations and time positions, assuming everything is independent
-    
+
     seed: int
         The seed for random generations
     return_indices: bool
         If True, return the indices of selection such that selected_peaks = peaks[selected_indices]
-    
+
     method_kwargs: dict of kwargs method
         Keyword arguments for the chosen method:
             'uniform':
@@ -53,7 +52,7 @@ def select_peaks(peaks, method='uniform', seed=None, return_indices=False, **met
                     Total number of peaks to select
                 * peaks_locations: array
                     The locations of all the peaks, computed via localize_peaks
-    
+
     {}
     Returns
     -------
@@ -61,6 +60,18 @@ def select_peaks(peaks, method='uniform', seed=None, return_indices=False, **met
         Selected peaks.
     """
 
+    selected_indices = select_indices(peaks, method=method, seed=seed, **method_kwargs)
+    selected_peaks = peaks[selected_indices]
+    if return_indices:
+        return selected_peaks, selected_indices
+    else:
+        return selected_peaks
+
+def select_indices(peaks, method, seed, **method_kwargs):
+    """Method to subsample all the found peaks before clustering.  Returns selected_indices.
+    This function is wrapped by select_peaks -- see
+    :func:`spikeinterface.sortingcomponents.peak_selection.select_peaks` for detailed documentation.
+    """
     selected_indices = []
     
     if seed is not None:
@@ -249,9 +260,4 @@ def select_peaks(peaks, method='uniform', seed=None, return_indices=False, **met
 
     selected_indices = np.concatenate(selected_indices)
     selected_indices = selected_indices[np.argsort(peaks[selected_indices]['sample_ind'])]
-    selected_peaks = peaks[selected_indices]
-
-    if return_indices:
-        return selected_peaks, selected_indices
-    else:
-        return selected_peaks
+    return selected_indices
--- a/spikeinterface/sortingcomponents/peak_selection.py
+++ b/spikeinterface/sortingcomponents/peak_selection.py
@@ -5,7 +5,7 @@ import scipy
 from sklearn.preprocessing import QuantileTransformer
 
 
-def select_peaks(peaks, method='uniform', seed=None, **method_kwargs):
+def select_peaks(peaks, method='uniform', seed=None, return_indices=False, **method_kwargs):
 
     """Method to subsample all the found peaks before clustering
     Parameters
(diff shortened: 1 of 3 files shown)


pauladkisson:
I know @h-mayorquin expressed his preference against changing the output size based on a flag (see. https://luzkan.github.io/smells/flag-argument).  But I felt like to keep backwards compatibility it's worth it in this case.

h-mayorquin:
I don't understand what do you mean by backwards compatbility here? 

I was suggestion that most of the current `select_peaks` function could be another method `select_peak_indexes` which would return your desired output.

`select_peak_indexes` (or any other better name) then would be used inside `select_peaks` without changing the current functionality at all (hence no break of backwards compatbility) and to achieve your intending use case.

pauladkisson:
I just refactored the ``peak_selection`` submodule as we discussed with ``select_indices(...)`` doing all the heavy lifting and ``select_peaks(...)`` calling ``select_indices`` and optionally returning the indices.

The only thing I wasn't totally sure about was the documentation.  To avoid rewriting things, I just had ``select_indices`` refer to the ``select_peaks`` docstring.  Also, I couldn't double-check the reference because peak_selection isn't in the API docs on sphinx right now.  

Let me know what you think @h-mayorquin.

spikeinterface/sortingcomponents/peak_selection.py
```diff
@@ -51,15 +52,27 @@ def select_peaks(peaks, method='uniform', seed=None, **method_kwargs):
                     Total number of peaks to select
                 * peaks_locations: array
                     The locations of all the peaks, computed via localize_peaks
-    
+
```
alejoe91, line 55: @pauladkisson can you add `selected_indices` (only returned when `return_indices` is `True`)?

spikeinterface/sortingcomponents/peak_selection.py
```diff
@@ -51,15 +52,27 @@ def select_peaks(peaks, method='uniform', seed=None, **method_kwargs):
...
+    selected_peaks = peaks[selected_indices]
+    if return_indices:
+        return selected_peaks, selected_indices
+    else:
+        return selected_peaks
+
+def select_indices(peaks, method, seed, **method_kwargs):
+    """Method to subsample all the found peaks before clustering.  Returns selected_indices.
+    This function is wrapped by select_peaks -- see
+    :func:`spikeinterface.sortingcomponents.peak_selection.select_peaks` for detailed documentation.
```
alejoe91, line 73:
```suggestion
    :py:func:`spikeinterface.sortingcomponents.peak_selection.select_peaks` for detailed documentation.
```

alejoe91 (commented):
thanks Paul, I added a couple of final suggestions

h-mayorquin:
Thanks for adding more tests to this function. This looks good to me. I have only one requirement: I find the curretnt name "select_indices`  too unspecific.  I think you should add peaks to it  (e.g. `select_peak_indexes`) or find another way to make it more concrete.

All the following points are just comments. Feel free to not address them.
- [joke] Indexes vs indices? you as a native speaker have the higher ground here but what about contributing to the regularization of the English language?
- I agree with what @alejoe91 mentioned in yetserday meeting that `select_peaks` is a bad name. But maybe we can wait for the discussion with @yger to pick a better name for the function.
- We should change the name of the test, right now is the same as the test of the `find_peaks` (see the summary here https://github.com/SpikeInterface/spikeinterface/actions/runs/4197283241)
- We should be mindful that these changes conflict in lines (but not in intention!) with the changes in https://github.com/SpikeInterface/spikeinterface/pull/1330 if we merge that first. 
- I kind of feel that `select_channels` should be a top level argument instead of a method one, maybe @yger have some methods in mind that did not make sense at the channel level but so far all of them have it. This will simplify the nested documentation.

yger:
I am not aware of the exact discussion, but do not hesitate to put me in the loop. You are not happy with the name of the function?

select_per_channel could I guess be a top level argument, at the cost of mabe not be used by some selection methods. But the ones that are not needing it are currently not used, so they could be either removed either discard this param for now.

pauladkisson:
@h-mayorquin
- I thought that since the function is already a part of the `peak_selection` submodule, `peak_selection.select_indices` was sufficiently specific, especially since users are encouraged to just use the `select_peaks` function with appropriate flags anyway.  But if you would prefer it to be more explicit in the function name, I am ok with the change.
- I generally prefer to use proper english and 'indices' is definitely more common in math/science usage.  But 'indexes' is actually searchable with find('index'), which is pretty nice tbh, so I'm happy to switch it over (see [here](https://stackoverflow.com/questions/1378781/proper-terminology-should-i-say-indexes-or-indices)).
- As for naming, personally I think `select_peaks` is fine.  If we want a function that enforces some maximum number of peaks, I would suggest adding a separate function called `set_max_n_peaks` (or something) that can call `select_peaks` as necessary.  But, yes, let's discuss with @yger over zoom sometime soon.
- Will change name of test function --> test_select_peaks
- Conflicting Lines: How should we deal with that?
- I'm going to leave the `select_channels` change for later -- not really related to this PR.

spikeinterface/sortingcomponents/peak_selection.py
pauladkisson: Done!

spikeinterface/sortingcomponents/peak_selection.py
pauladkisson: Done

pauladkisson:
@alejoe91 @h-mayorquin
I responded to all your feedback -- should be ready to merge now.

h-mayorquin:
> @h-mayorquin
> 
> * I thought that since the function is already a part of the `peak_selection` submodule, `peak_selection.select_indices` was sufficiently specific, especially since users are encouraged to just use the `select_peaks` function with appropriate flags anyway.  But if you would prefer it to be more explicit in the function name, I am ok with the change.

I am thinking in two cases:
1) Someone who sees these function in the middle of a script.
2) Someone who encounters the function through a ctrl + f mechanism in the repo.


> * I generally prefer to use proper english and 'indices' is definitely more common in math/science usage.  But 'indexes' is actually searchable with find('index'), which is pretty nice tbh, so I'm happy to switch it over (see [here](https://stackoverflow.com/questions/1378781/proper-terminology-should-i-say-indexes-or-indices)).

Thanks for the link, I was kind of joking because I did not know that plura. You make some good poitns but Alessio points out consistency so maybe let's leave this for another terminology discussion : P

They don't have the import context to let them know that this is the case.

> * Will change name of test function --> test_select_peaks

Thanks!

> * Conflicting Lines: How should we deal with that?

This is up to @alejoe91 , either he merges this first and I make another PR with the changes, or that one is merged first and then you will have some to solve merge conflicts. I am fine with doing the work myself if you guys want.

Thanks for the hard work!

alejoe91:
@h-mayorquin let's merge this first, just following chronological order :)

pauladkisson:
> Thanks @pauladkisson
> 
> I noticed you actually switched to `indexes`, but honestly I would rather go with `indices` (as in several other functions/args) in the API. The rest looks good!

Done

alejoe91 (approved):

h-mayorquin:
Merging this @pauladkisson 
Nice commit organization btw! I thinks this is your first PR here, right? : )

alejoe91:
Right! Congrats on your first PR merged @pauladkisson 👏🎉

pauladkisson:
Thank you! 😁

Issue #1311, opened 2023-02-13
Exceeding storage allocation for CI caches 


h-mayorquin:
In some parts of the repo (right now #1310) we are getting errors like this:

![image](https://user-images.githubusercontent.com/6429509/218422849-0bd8f6d9-5797-437e-83f4-f9d4d559db04.png)

I think they are generated by an error on virtual environment caching. This in turn most likely is coming from our large cache sizes which are over the quota right now: 
https://github.com/SpikeInterface/spikeinterface/actions/caches

This situation causes cache trashing as described here:
https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy

Importantly, note that we have repeated caches with the same hash and date for different pull requests. I think this is happening because of the cache accessing logic of github CI:
https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache

That is, the caches need to be created in main but I think right now we are creating them on branches so they don't access each other even if they have the same key.

What should we do? 
If I am right about the above I see three options:
1) Stop caching the virtual environment (how long does the virtual environment takes to be created, for core it takes around half a minute but I guess full is much longer, how much?).
2) Reduce the footprint of the virtual environment so each is smaller.
3) Create caches automatically from the main / master branch so the branches have access to them.

I also think that we should implement logic in the testing workflow files so we have either an assertion -stop the tests- or a clean install when there is a cache miss event.

What do you think? I am missing something?

Edit: some useful reference / context of someone who encountered the same problem: https://github.com/actions/cache/issues/79

h-mayorquin:
Ok, so after discussion with @samuelgarcia on the 2023-02-15 we agreed that we are implementing the solution number 3 above. I already build a solution to have the caches produce in master (see #1321). However, from the [github documentation ](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache) see the following quote:

> When a cache is created by a workflow run triggered on a pull request, the cache is created for the merge ref (refs/pull/.../merge). Because of this, the cache will have a limited scope and can only be restored by re-runs of the pull request. It cannot be restored by the base branch or other pull requests targeting that base branch.

That is, to finish this issue and test that branches will have reading access to the produce caches I need to have #1321 merged into master. Can we move forward with that?

h-mayorquin:
[...]

PR #1321, opened 2023-02-14
Create caches on master

h-mayorquin:
Related to https://github.com/SpikeInterface/spikeinterface/issues/1311.

This script should allow us to create the caches on master and from there I can propagate them to any created branches (so we don't have the duplication problem that we have now).  I already tested this on #1317 and it works correctly and as I expected. To test that the caches can be propagated to the other branches I need to have this in master. That is, I need this to be merged before I can proceed to address the problems described on #1311. 

This should close #1317 and is just a clean version of that one to avoid polluting the commit history.

### DIFF
--- a/.github/workflows/caches_cron_job.yml
+++ b/.github/workflows/caches_cron_job.yml
@@ -1,0 +1,77 @@
+name: Create caches for gin ecephys data and virtual env
+
+on: 
+  workflow_dispatch:
+  schedule:
+    - cron: "0 12 * * *"  # Daily at noon UTC
+
+jobs:
+  create-virtual-env-cache-if-missing:
+    name: Caching virtual env
+    runs-on: "ubuntu-latest"
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/caches_cron_job.yml
```diff
@@ -0,0 +1,77 @@
...
+          pip install datalad-installer
+          datalad-installer --sudo ok git-annex -m datalad/packages
+          pip install datalad
+      - name: Download dataset
+        if: steps.cache-datasets.outputs.cache-hit != 'true'
+        run: |
+          datalad install -rg https://gin.g-node.org/NeuralEnsemble/ephy_testing_data
+          mv --force ./ephy_testing_data ~/spikeinterface_datasets
+          cd ~
+          du -hs spikeinterface_datasets
```
samuelgarcia, line 77: Here we do not download anything. Only the full tree if I am correct. No ? 
Or is the entire dataset download any way ?
Maybe we should get some file that are needed by our tests.

samuelgarcia (approved):

PR #1325, opened 2023-02-16
Remove the rest of to_dict mechanisms from sorting components

h-mayorquin:
Within the scope of #1235. 

A continuation of #1272 as rquested by @samuelgarcia 

Removing the rest of the `to_dict` mechanisms for the recording extractors


alejoe91:
@h-mayorquin just realized that without this PR the `detect_peaks` + nodes fails if `n_jobs>1`. Could you change `n_jobs` to `2` [](https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/sortingcomponents/tests/test_peak_detection.py#L23) (or repeat the same test with `n_jobs=2`)? I think that multiple jobs are almost always used..

h-mayorquin:
@alejoe91 done in the last commit. Ran well locally.

alejoe91 (approved):

PR #1330, opened 2023-02-16
new numpy features for peak selection

h-mayorquin:
Updated the select peaks function to use the new -and recommended - features of numpy for random generator and container operations.

### DIFF
--- a/spikeinterface/sortingcomponents/peak_selection.py
+++ b/spikeinterface/sortingcomponents/peak_selection.py
@@ -7,7 +7,9 @@ from sklearn.preprocessing import QuantileTransformer
 
 def select_peaks(peaks, method='uniform', seed=None, **method_kwargs):
 
-    """Method to subsample all the found peaks before clustering
+    """
+    Method to select a subset of peaks from a set of peaks.
+    Usually use for reducing computational foorptint of downstream methods.
     Parameters
     ----------
     peaks: the peaks that have been found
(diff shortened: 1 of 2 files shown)


spikeinterface/sortingcomponents/peak_selection.py
```diff
@@ -60,9 +62,8 @@ def select_peaks(peaks, method='uniform', seed=None, **method_kwargs):
     """
 
     selected_peaks = []
-    
-    if seed is not None:
-        np.random.seed(seed)
+    seed = seed if seed else None
+    rng = np.random.default_rng(seed=seed)
```
samuelgarcia, line 66: Cool!

samuelgarcia:
This OK for me.
You do not like `in1d` ?

h-mayorquin:
> This OK for me. You do not like `in1d` ?

@samuelgarcia 
I think `isin` reflects intent better. Plus it is also is used in pandas and alludes to the `in` operator in python. Finally, this is the suggestion in the numpy documentation:
https://numpy.org/doc/stable/reference/generated/numpy.in1d.html

h-mayorquin:
@alejoe91 I solved the merge conflicts with #1318

alejoe91:
Awesome! We should also use the rng approach in other parts of the code..I think that in several places we still use the old method

alejoe91 (approved):

PR #1342, opened 2023-02-18
`black` and `isort` support

caniko:
This PR changes nothing except for code-aesthetics. I ran the tests just in case.

Refactor: Applied black and isort. The code formatting in spikeinterface is very difficult to work with as a developer.
Doc: Add guide to run black and isort, stored in CONTRIBUTE.md.


h-mayorquin:
I am all for using black. I think it reduces a lot of cognitive effort for how to code and is becoming more standard.

This will upend currently open PRs though so we if we decide to move on with it should be done wihout breaking that.
My suggestion is that we do it together with #1344 and activate either a bot or a pre-commig hook so this is automatically solved. 

Something that I feel really strong about though is using 120 as a character limit instead of 80. Even people who program old systems are agreeing with this [joke]:
https://www.reddit.com/r/programming/comments/gt4wgn/linus_torvalds_on_80character_line_limit/

caniko:
@h-mayorquin pre-commit hook is already implemented in this PR.

I prefer to set the character limit to a 100.

I am not sure if it is a good idea to combine PRs. Furthermore, I can resolve the conflicts and run the code formatting again before merging! What do you think @alejoe91?

h-mayorquin:
> @h-mayorquin pre-commit hook is already implemented in this PR. Please look at CONTRIBUTE.md.

I did see it. That relies on users being dilligent. I meant to automate it at the repo level so it does not depend on users.

> I prefer to set the character limit to a 100, and this is the current setting. I can make it 120 if there is a majority for it; personally, I feel like I end up turning my head more when it is 120. The settings are exposed at `pyproject.toml`
 
We can settle it by vote if this goes through. Just to make it transpartent to the others this just impliels picking a number in the following configuration for the `.toml` file:

```
[tool.black]
line-length = the_number_that_we_decided
```

> Is it a good idea to combine these PRs? This PR is ready for merging. Furthermore, I can resolve the conflicts and run the code formatting again before merging! What do you think @alejoe91?

I don't mean combining them in one PR. I meant doing these two things that will have extremely large diffs in a batch if we decide to do them both.

alejoe91:
Thanks guys! I would also vote for 120..we can always switch later! One thing at a time :P

We can make another PR for the automatic pre-commit hook

CodyCBakerPhD:
>  meant doing these two things that will have extremely large diffs in a batch if we decide to do them both. Both of them will upend all the open PRs and we should minimize disturbing the work of others.

Specifically, running or merging black directly to the `main` branch that other people have already checked out from, will then produce a massively inflated changelog until that branch is updated to include the reformatting - however, attempting to update the branch (merging re-formatted main into it) inevitably incurs a lot of merge conflicts that are super annoying to deal with.

The recommended approach from past experience is actually to *not* try to update the branch right away in that case and instead copy the pre-commit config over to the branch you want to update and run pre-commit specifically on that - now the majority of the changelog should deflate and conflicts minimized.


> Edit: how comes the CI testing suite is not running here?

First-time contributors must have their action runs be approved in order to run

caniko:
> First-time contributors must have their action runs be approved in order to run

Note that there is no linting with `flake8` in this PR. I only implemented `black` and `isort`.

> We can make another PR for the automatic pre-commit hook

I think we should implement this with the pre-commit hook. The installation of the hooks are still optional, `pre-commit install` is up to the cloner/forker.

> The recommended approach from past experience is actually to not try to update the branch right away in that case and instead copy the pre-commit config over to the branch you want to update and run pre-commit specifically on that - now the majority of the changelog should deflate and conflicts minimized.

They also have to copy the `pyproject.toml`!

> First-time contributors must have their action runs be approved in order to run

I have made PRs before, but they have always been closed :sweat_smile:

CodyCBakerPhD:
My 2 cents learned from experience over the past couple of years...

> and the code is not hard to read.

This outsider would humbly disagree for many of the more complicated inner workings of the codebase, and

> Being beginner friendly is imporant to me.

When I was a beginner with SI (and sometimes still am...), having easier to read code would have helped me quite a bit.

Here's the thing though, with an auto-bot, no one has to learn anything. No one has to think about it. You don't even have to require flake8 as a PR merge requirement (big annoyance).

If you want, you could simply accept code as people want to contribute it, consistent with

> For simple patch and simple PR for me it is OK to have not rules on the formattting.

and then trigger the reformat upon merge, outside of anyone's awareness

As devs, it's also a big time saver to not have to comment on PRs about code formatting or import ordering

caniko:
> No one has to think about it. You don't even have to require flake8 as a PR merge requirement (big annoyance).

This is why I didn't include flake8. However, I am on the side of flake8 after black and isort formatting.

How many beginners submit PRs to `spikeonterface`? It is too complex for someone who doesn't have at least some experience, IMHO.

**Opinion and research data**
Code formatting is beginner friendly as it matures the beginner to become a better developer. Code standardization is known to make collaboration more efficient there has been research on this topic, this was initially reflected upon in PEP8 written by Guido. Public discussion on [stack exchange](https://softwareengineering.stackexchange.com/q/189274), general sentiment is in favor.

Newer languages like Rust and Dart even have native code formatters that are bundled with the compiler. Python came out before the age of git+FOSS, hence our situation.

h-mayorquin:
@samuelgarcia 
I agree that it is important to keep barriers for entry as low as possible. However, as @CodyCBakerPhD mentions, I think we can have the formatting to be a post-merge action so users don't really need to do anything.

PR #1343, opened 2023-02-18
`download_dataset` function not accessing cached data on CI

h-mayorquin:
For some reason the test for the function `download_dataset`  is taking too long on the full tests (see this [summary](https://github.com/SpikeInterface/spikeinterface/actions/runs/4211526876)). I think it is not accessing the cached data for mearec. I think sorting this is low-hanging fruit for making the tests faster.

I am testing here if this is the case.

### DIFF
--- a/.github/workflows/caches_cron_job.yml
+++ b/.github/workflows/caches_cron_job.yml
@@ -91,6 +91,6 @@ jobs:
           du -hs spikeinterface_datasets
           cd spikeinterface_datasets
           pwd
-          ls -lh  # Should show /ephy_testing_data
-          cd /ephy_testing_data
+          ls -lh  # Should show ephy_testing_data
+          cd ephy_testing_data
           ls -lh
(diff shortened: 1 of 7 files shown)


h-mayorquin:
Ok, this is done. 
The tests were not using the cached files because they were looking at `'spikeinterface_datasets' but they are actually in `'spikeinterface_datasets/ephy_testing_data'. This reduces the testing time for the full tests DRASTICALLY. Now the [full tests with codecoverage ran in ~25 min](https://github.com/SpikeInterface/spikeinterface/actions/runs/4215720809) plus we are getting the right profiling time (as the function that were in the top previously were the ones that were calling download_dataset many times). 

I also did two minor things in this PR:
* Changed the output of the tests to be more verbose and some minor refactoring on the testing files. 
* Added docstring to `download_dataset` function

samuelgarcia:
Bravo!

alejoe91 (approved):

PR #1348, opened 2023-02-20
Fix LSB correction

alejoe91:
Fixed the logic of LSB correction case channels are not centered at 0.


samuelgarcia (approved):

samuelgarcia:
This is OK for me.
Why did you put the abs() in the first implementation ? This is really conter intuitive! No ?

samuelgarcia (approved):

PR #1354, opened 2023-02-22
Create `internal` and `external` sorters sub-modules

alejoe91:
- Test are not run for `external` sorters (avoiding to install unnecessary external dependencies)
- Simplifies code coverage
- Simplifies `sorters` module strucrure (with `internal` and `external` subfolders)
- The full tests now are split between `sorters` (core changes to launchers/utils etc), `internal` (tested if changed), and `external` (not tested)

The idea is to add the testing machinery of [spikeinterface-dockerfiles](https://github.com/SpikeInterface/spikeinterface-dockerfiles) to test using docker/singularity


.github/workflows/full-test.yml
```diff
@@ -118,13 +116,17 @@ jobs:
               echo "Quality metrics changed"
               echo "QUALITYMETRICS_CHANGED=true" >> $GITHUB_OUTPUT
             fi
-            if [[ $file == *"/sorters/"* ]]; then
+            if [[ $file == *"/sorters/"* && $file != *"/sorters/internal/"* && $file != *"/sorters/internal/"* ]]; then
```
h-mayorquin, line 119: Should not one the negative conditions be `$file != *"/sorters/external/"`? I think you are repeating the statement here.

conftest.py
```diff
@@ -37,14 +38,19 @@ def pytest_collection_modifyitems(config, items):
...
                     mark = getattr(pytest.mark, mark_name)
                     item.add_marker(mark)
         else:
-            item.add_marker("si_sorters")
+            if "/internal/" in str(rel_path):
+                item.add_marker("sorters_internal")
+            elif "/external/" in str(rel_path):
+                item.add_marker("sorters_external")
+            else:
+                print(rel_path)
```
h-mayorquin, line 52: I guess this print statement should be removed?

It would be nice if we could already do use https://peps.python.org/pep-0634/ but I guess we want people who use older versions to still be able to run the tests?

spikeinterface/sorters/external/combinato.py
```diff
@@ -153,7 +153,7 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
         tmpdir = sorter_output_folder
 
         if verbose:
-            print(f'Running combinato in {tmpdir}...')
+            print(f'Running combinato in {tmpdir}..')
```
h-mayorquin, line 156: What's up with all the cahnges from `...` to `..` here.

conftest.py
h-mayorquin: I would also do the if with the positive side, I think that:
```
        rel_path = Path(item.fspath).relative_to(rootdir)
        if "sorters" in str(rel_path):
            if "/internal/" in str(rel_path):
                item.add_marker("sorters_internal")
            elif "/external/" in str(rel_path):
                item.add_marker("sorters_external")
            else:
                item.add_marker("sorters")
        else:
            for mark_name in mark_names:
                if f"/{mark_name}/" in str(rel_path):
                    mark = getattr(pytest.mark, mark_name)
                    item.add_marker(mark)

```

h-mayorquin (commented):
This looks good to me concerning the CI changes. I could not figure it out based on the names though, are the ones called external the ones that will run with docker, the internal ones will be the ones based on sortingcomponents and the rest is basic machinery common to both, am I getting this right?

spikeinterface/sorters/external/combinato.py
alejoe91: Hi think I did a find and replace for relative paths (`...` -> `..`)! Will fix it

conftest.py
alejoe91: done

h-mayorquin (approved):
Seems good to me, we will avoid running the costly internal sorters tests unless we change them or sorting components. 

Also, we now have a distinction between internal and external expressed in the structure of the repo.

samuelgarcia:
Hey guys.
I did not have chance to have a look to this!!
For big stuff it is cool to have sevarl eyes on it.

h-mayorquin:
Sorry @samuelgarcia my mistake, I remember that we talked about this before and I thought you gave a green flag to this. I will ping you next time for something this large.

PR #1359, opened 2023-02-24
Pass sparsity argument directly to `extract_waveforms`

alejoe91:


spikeinterface/sortingcomponents/peak_pipeline.py
```diff
@@ -56,7 +56,7 @@ class WaveformExtractorNode(PipelineNode):
     """Base class for waveform extractor"""
 
     def __init__(self, recording: BaseRecording, 
-                 ms_before: float, ms_after: float, parents: Optional[list[PipelineNode]]=None, return_ouput: bool=True):
+                 ms_before: float, ms_after: float, parents: Optional[List[PipelineNode]]=None, return_ouput: bool=True):
```
alejoe91, line 59: This is needed for Python3.8 (and also work for > versions). Using `list[]` directly is not supported in 3.8

h-mayorquin: So they only stared making this available after python 3.9? Good to know.

spikeinterface/core/waveform_extractor.py
```diff
@@ -1424,11 +1427,18 @@ def extract_waveforms(recording, sorting, folder=None,
...
+                                           num_spikes_for_sparsity=num_spikes_for_sparsity,
+                                           **estimate_kwargs, **job_kwargs)
+        else:
+            assert isinstance(sparsity, ChannelSparsity), "'sparsity' must be a ChannelSparsity object"
+            unit_id_to_channel_ids = sparsity.unit_id_to_channel_ids
+            assert all(u in sorting.unit_ids for u in unit_id_to_channel_ids), \
+                    "Invalid unit ids in sparsity"
+            for channels in unit_id_to_channel_ids.values():
+                assert all(ch in recording.channel_ids for ch in channels), \
+                    "Invalid channel ids in sparsity"
```
samuelgarcia, line 1441: I think we should have something more easy.
In short `extract_waveforms(recording, sorting, sparsity=something)` should be possible.
Here we force `extract_waveforms(recording, sorting, sparsity=something, sprase=True)`

In my mind `sprse=True` is "try to make make it sparse with estimation"

alejoe91: Ok, so if you pass sparsity `sparse` is ignored ;) I agree

spikeinterface/sortingcomponents/peak_pipeline.py
alejoe91: yeah, it was failing on 3.8 envs. Using `List` seems to be compatible to both

samuelgarcia (approved):

PR #1360, opened 2023-02-24
Typo and return output False by default for waveform nodes

alejoe91:
- Corrects a typo everywhere in sortingcomponents: `return_ouput` --> `return_output`
- Sets default `return_output=False` for extract waveforms nodes


h-mayorquin (approved):
My bad, this somehow flew over me. 

Also, good, I was meaning to ask about the default for waveform extractors. I guess the idea is that users in general will add this node because they want to use them downstream and not because they want to extract waveforms. Hence, the default is to not return.

samuelgarcia:
I am really wondering who can inject in the master such a big typo!!

PR #1363, opened 2023-02-24
Add a margin parameter to get_random_data_chunks

cwindolf:
Wonder if SI would consider extending the API of `get_random_data_chunks` in some way along these lines? Use case: for instance when calling `si.zscore()`, we can exclude the border to avoid issues at the edge of the recording -- that could be helpful for other functions as well, like estimating noise levels. I also considered parameterizing by `min_frame/max_frame`, but that doesn't generalize as well across segments.

### DIFF
--- a/spikeinterface/core/recording_tools.py
+++ b/spikeinterface/core/recording_tools.py
@@ -1,7 +1,7 @@
 import numpy as np
 
 def get_random_data_chunks(recording, return_scaled=False, num_chunks_per_segment=20, 
-                           chunk_size=10000, concatenated=True, seed=0):
+                           chunk_size=10000, concatenated=True, seed=0, margin_frames=0):
     """
     Exctract random chunks across segments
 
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Hi Charlie.
Normally by design all recording handle themself therer own margin.
For instance, the filter recording you ask for [start, stop] already do some [start-margin, stop+margin] for computing and then only giv ethe user [start, stop]

See this https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/preprocessing/filter.py#L112

So when you `get_random_data_chunks()` some margin are already applied depending the underlying preprocess chain you have.
So for me, this patch is not really usefull and maybe can be misleading. Maybe I am missing something.
Would be happy to discuss more this.
Furthermore, the zscore do not have border effect because it is a by sample computation. No ?

alejoe91 (approved):

samuelgarcia:
> Hi Charlie. Normally by design all recording handle themself therer own margin. For instance, the filter recording you ask for [start, stop] already do some [start-margin, stop+margin] for computing and then only giv ethe user [start, stop]
> 
> See this https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/preprocessing/filter.py#L112
> 
> So when you `get_random_data_chunks()` some margin are already applied depending the underlying preprocess chain you have. So for me, this patch is not really usefull and maybe can be misleading. Maybe I am missing something. Would be happy to discuss more this. Furthermore, the zscore do not have border effect because it is a by sample computation. No ?


Oups I miss read you PR. too tired.
I get your need and merge it.

samuelgarcia (approved):

cwindolf:
Thanks! Sorry for the confusion, I can see how it was confusing with the margin in each chunk

PR #1370, opened 2023-02-27
Displaing testing errors on CI at failure

h-mayorquin:
This is something that @alejoe91 requested on slack. Uses a t pipeline to print to both standard output and a file and then a boolean or to identify pytest error.

I also reduced the verbosity of the traces and outputs to make it more similar to the defaults of pytest out of the box.


alejoe91:
Thanks @h-mayorquin !

h-mayorquin:
I am taking a look at #1354 now. Hopefully we can merge that first so you don't have to deal with conflicts.

alejoe91 (approved):

PR #1376, opened 2023-03-01
Refactor temporal pca tests

h-mayorquin:
This PR is mean to refactor the tests for the recently merged temporal PCA so they run faster. Specifically, the refactor uses package-wise fixture scopes to avoid re-computing the fit method and the peaks that are used across the various temporal PCA tests. This approchs also has the benefit of modularizing the test, centralizes the set-up and avoids repeating code.

I also did other minor changes, namely:
* Added doctstrings to the waveform extractor nodes.
* Remove the no-longer used name from the base class
* Added type-hints to some of the classes.

This also works to decouple this changes from the ongoing #1282 so the diff of the latter becomes easier to read and review.


spikeinterface/comparison/tests/test_templatecomparison.py
```diff
@@ -24,7 +24,6 @@ def setup_module():
 
 
 def test_compare_multiple_templates():
-    test_dir = Path("temp_comp_test")
```
h-mayorquin, line 27: I think someone left this in a previous debugg session and was generating files in the wrong place.

.gitignore
```diff
@@ -184,53 +184,7 @@ examples/modules/toolkit/phy/*
...
-toy_sorting_1seg/
-toy_sorting_2seg/
-toy_waveforms/
-toy_waveforms_1seg/
-toy_waveforms_2seg/
-tridesclous_output/
-# waveforms/
-waveforms_mearec/
-waveforms_rm/
+test_folder/
```
h-mayorquin, line 187: In the current iteration all the files go here actually.

spikeinterface/sortingcomponents/tests/test_peak_detection.py
```diff
@@ -29,7 +29,7 @@ def test_detect_peaks():
         repo=repo, remote_path=remote_path, local_folder=None)
     recording = MEArecRecordingExtractor(local_path)
     
-    job_kwargs = dict(n_jobs=2, chunk_size=10000, progress_bar=True, verbose=True)
+    job_kwargs = dict(n_jobs=-1, chunk_size=10000, progress_bar=True, verbose=True)
```
h-mayorquin, line 32: This should make the test use paralell in the CI and also run in the unlikely case someone runs this from a single core machine.

h-mayorquin (commented):
Some comments on the changes.

spikeinterface/sortingcomponents/peak_pipeline.py
```diff
@@ -18,32 +18,34 @@
...
-    Typically used for exctrating features (amplitudes, localization, ...)
-    
-    A Node can optionally connect to other nodes with the parents and receive inputs from others.
-    """
-    def __init__(self, recording, return_output=True, parents=None, name=None):
+    def __init__(
+        self, recording: BaseRecording, return_output: bool = True, parents: Optional[List[Type["PipelineNode"]]] = None
+    ):
+        """
+        This is a generic object that will make some computation on peak given a buffe of traces.
```
alejoe91, line 25:
```suggestion
        This is a generic object that will make some computation on peaks given a buffer of traces.
```

alejoe91 (changes requested):
Just a small typo @h-mayorquin

alejoe91 (approved):

alejoe91:
@samuelgarcia wabt to take a look?

pyproject.toml
```diff
@@ -77,6 +77,7 @@ full = [
     "networkx",
     "distinctipy",
     "matplotlib",
+    "huggingface_hub",
```
samuelgarcia, line 80: I am not sure that this huggingface_hub should be in full.

.gitignore
samuelgarcia: cool

spikeinterface/sortingcomponents/peak_pipeline.py
```diff
@@ -18,32 +18,34 @@
...
             # only one parents is allowed
             parents = [parents]
         self.parents = parents
         
-        self._kwargs = dict(
-            return_output=return_output,
-        )
-        if parents is not None:
-            self._kwargs['parents'] = parents
+        self._kwargs = dict()
```
samuelgarcia, line 47: I think I likde the idea that parent and return_output where handle in the base class.
So we never forget to push in then in all classes

spikeinterface/sortingcomponents/peak_pipeline.py
```diff
@@ -18,32 +18,34 @@
...
         
-            
-    def post_check(self):
-        # can optionaly be overwritten
-        # this can trigger a check for compatibility with other nodes (typically parents)
-        pass
-
```
samuelgarcia, line 46: cool. We forget it!

spikeinterface/sortingcomponents/tests/test_waveforms/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+
+import spikeinterface as si
+import spikeinterface.extractors as se
+
+from spikeinterface.sortingcomponents.peak_detection import detect_peaks
+
+
+@pytest.fixture(scope="package")
+def chunk_executor_kwargs():
+    job_kwargs = dict(n_jobs=-1, chunk_size=10000, progress_bar=False)
```
samuelgarcia, line 11: n_jobs=-1 for tests is maybe a bit agressive.
I like to have n_job=1 and if we really want mltiprocessing n_job=2 or 4.

samuelgarcia:
Except the "huggingface_hub" in full I am Ok with this clean.

pyproject.toml
h-mayorquin: Where do you think is best?
So I moved it here @alejoe91 did not want it `dependencies`

That said, this can wait, so I am removing it from here and adding it on the SingleChannelDenoiser test.

spikeinterface/sortingcomponents/tests/test_waveforms/conftest.py
h-mayorquin: Why not used all the machine resources during testing? Specially for CI where the runners are dedicated.

spikeinterface/sortingcomponents/peak_pipeline.py
h-mayorquin: There are some children that are still acessing kwars. I will remove that in another refactor.

PR #1378, opened 2023-03-01
Add function to remove excess spikes

alejoe91:
Sometimes KS returns spikes outside the number of samples. 

This PR:
- asserts that this is not the case when extracting waveforms
- implements a curation function `remove_excess_spikes` that filters out these extra spikes

### DIFF
--- a/spikeinterface/core/waveform_extractor.py
+++ b/spikeinterface/core/waveform_extractor.py
@@ -13,7 +13,7 @@ from .base import load_extractor
 from .core_tools import check_json
 from .job_tools import _shared_job_kwargs_doc, split_job_kwargs, fix_job_kwargs
 from .recording_tools import check_probe_do_not_overlap
-from .waveform_tools import extract_waveforms_to_buffers
+from .waveform_tools import extract_waveforms_to_buffers, has_exceeding_spikes
 from .sparsity import ChannelSparsity, compute_sparsity, _sparsity_doc
 
 _possible_template_modes = ('average', 'std', 'median')
@@ -245,6 +245,11 @@ class WaveformExtractor:
         properties_to_attrs = deepcopy(recording._properties)
         if 'contact_vector' in properties_to_attrs:
             del properties_to_attrs['contact_vector']
+        if has_exceeding_spikes(recording, sorting):
+            raise ValueError(
+                    "The sorting object has spikes exceeding the recording duration. You have to remove those spikes "
+                    "with the `spikeinterface.curation.remove_excess_spikes()` function"
+                )
         rec_attributes = dict(
             channel_ids=recording.channel_ids,
             sampling_frequency=recording.get_sampling_frequency(),
--- a/spikeinterface/core/waveform_tools.py
+++ b/spikeinterface/core/waveform_tools.py
@@ -352,3 +352,27 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
                     wfs[pos, :, :] = wf
                 else:
                     wfs[pos, :, :] = wf[:, sparsity_mask[unit_ind]]
+
+
+def has_exceeding_spikes(recording, sorting):
+    """
+    Check if the sorting objects has spikes exceeding the recording number of samples, for all segments
+
+    Parameters
+    ----------
+    recording : BaseRecording
+        The recording object
+    sorting : BaseSorting
+        The sorting object
+
+    Returns
+    -------
+    bool
+        True if exceeding spikes, False otherwise
+    """
+    spike_vector = sorting.to_spike_vector()
+    for segment_index in range(recording.get_num_segments()):
+        spike_vector_seg = spike_vector[spike_vector["segment_ind"] == segment_index]
+        if np.max(spike_vector_seg["sample_ind"]) > recording.get_num_samples(segment_index=segment_index) - 1:
+            return True
+    return False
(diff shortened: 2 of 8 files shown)


spikeinterface/core/waveform_extractor.py
```diff
@@ -245,6 +245,11 @@ def create(cls, recording, sorting, folder, mode="folder", remove_if_exists=Fals
         properties_to_attrs = deepcopy(recording._properties)
         if 'contact_vector' in properties_to_attrs:
             del properties_to_attrs['contact_vector']
+        if has_exceeding_spikes(recording, sorting):
```
samuelgarcia, line 248: This is a high cost because it does a to vector.
But Iplan to refactor waveforms extractor using only vector version.
So lets go for this for now.

spikeinterface/core/waveform_tools.py
```diff
@@ -351,4 +351,29 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+        The sorting object
+
+    Returns
+    -------
+    bool
+        True if exceeding spikes, False otherwise
+    """
+    spike_vector = sorting.to_spike_vector()
+    for segment_index in range(recording.get_num_segments()):
+        spike_vector_seg = spike_vector[spike_vector["segment_ind"] == segment_index]
```
samuelgarcia, line 375: I would do a serachsorter to be faster.

spikeinterface/core/waveform_tools.py
```diff
@@ -351,4 +351,29 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+        if len(spike_vector_seg) > 0:
+            if np.max(spike_vector_seg["sample_ind"]) > recording.get_num_samples(segment_index=segment_index) - 1:
```
samuelgarcia, line 377: Instead of max I would spike_vector_seg["sample_ind"][-1] which is necesaarly the max.

spikeinterface/curation/remove_excess_spikes.py
```diff
@@ -0,0 +1,80 @@
...
+    def __init__(self, parent_segment: BaseSortingSegment, num_samples: int) -> None:
+        super().__init__()
+        self._parent_segment = parent_segment
+        self._num_samples = num_samples
+
+    def get_unit_spike_train(self, unit_id, start_frame: Optional[int] = None,
+                             end_frame: Optional[int] = None) -> np.ndarray:
+        spike_train = self._parent_segment.get_unit_spike_train(unit_id, start_frame=start_frame, end_frame=end_frame)
+
+        return spike_train[spike_train < self._num_samples - 1]
```
samuelgarcia, line 57: spike_train[spike_train <= self._num_samples - 1]
or
spike_train[spike_train < self._num_samples]
but here you also remove thelast possible spike no ?

spikeinterface/curation/remove_excess_spikes.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Returns
+    -------
+    sorting_without_excess_spikes: Sorting
+        The sorting without any excess spikes.
+    """
+    if has_exceeding_spikes(recording=recording, sorting=sorting):
+        return RemoveExcessSpikesSorting(sorting=sorting, recording=recording)
+    else:
+        return sorting
```
samuelgarcia, line 80: I understand the why but no sure this is so good to have th hidden stuff.
Anyway it is a detail, lets keep this.

spikeinterface/curation/tests/test_remove_excess_spikes.py
```diff
@@ -0,0 +1,40 @@
...
+    num_spikes = 100
+    num_excess_spikes_per_segment = 5
+    times = []
+    labels = []
+    for segment_index in range(recording.get_num_segments()):
+        num_samples = recording.get_num_samples(segment_index=segment_index)
+        times_segment = np.array([], dtype=int)
+        labels_segment = np.array([], dtype=int)
+        for unit in range(num_units):
+            spike_times = np.random.randint(0, num_samples - 1, num_spikes)
```
samuelgarcia, line 23: I think that it should be 
spike_times = np.random.randint(0, num_samples, num_spikes) because the high is not included

spikeinterface/curation/tests/test_remove_excess_spikes.py
```diff
@@ -0,0 +1,40 @@
...
+            excess_spikes = np.random.randint(num_samples, num_samples + 100, num_excess_spikes_per_segment)
```
samuelgarcia, line 24: Instead of random I would force hard coded corner case like exactly "num_samples" with a manul list.

spikeinterface/core/waveform_extractor.py
alejoe91: it takes 1ms

spikeinterface/curation/remove_excess_spikes.py
alejoe91: oh right!

spikeinterface/core/waveform_extractor.py
samuelgarcia: I do not believe you of course.
For enormous sorting also ? Or only for a small 10s test file ?

alejoe91:
Thanks @samuelgarcia 

Applied all suggestions

spikeinterface/core/waveform_extractor.py
alejoe91: >1h long neuropixels output ;) 

I think we have to have an internal caching of the spike vector in the sorting object, but that's another story ;)

samuelgarcia (approved):

PR #1386, opened 2023-03-02
Remove CI options from runsorter

alejoe91:
See https://github.com/SpikeInterface/spikeinterface/pull/1373#discussion_r1123116199

### DIFF
--- a/spikeinterface/sorters/runsorter.py
+++ b/spikeinterface/sorters/runsorter.py
@@ -434,13 +434,7 @@ if __name__ == '__main__':
                 'bind': str(recording_folder_unix), 'mode': 'ro'}
     volumes[str(parent_folder)] = {'bind': str(parent_folder_unix), 'mode': 'rw'}
     
-    running_on_github_actions = os.getenv("CI") 
-    si_dev_path = None
-    if running_on_github_actions:
-        print(si_version)
-        si_dev_path = os.getenv('SPIKEINTERFACE_DEV_PATH')
-        print(si_dev_path)
-        assert si_dev_path is not None
+    si_dev_path = os.getenv('SPIKEINTERFACE_DEV_PATH', None)
         
     install_si_from_source = False
     if 'dev' in si_version and si_dev_path is not None:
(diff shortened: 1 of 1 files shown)


spikeinterface/sorters/runsorter.py
```diff
@@ -482,16 +476,18 @@ def run_sorter_container(
     if verbose:
         print('Starting container')
     container_client.start()
-    print("******")
-    print("Container started with the following paths")
-    print(si_dev_path_unix, si_source_folder)
+    if verbose:
```
h-mayorquin, line 479: I don't think this is needed.

spikeinterface/sorters/runsorter.py
```diff
@@ -482,16 +476,18 @@ def run_sorter_container(
...
+        print(si_dev_path_unix, si_source_folder)
+    
     # check if container contains spikeinterface already
     cmd_1 = ['python', '-c', 'import spikeinterface; print(spikeinterface.__version__)']
     cmd_2 = ['python', '-c', 'from spikeinterface.sorters import run_sorter_local']
     res_output = ''
     for cmd in [cmd_1, cmd_2]:
         res_output += str(container_client.run_command(cmd))
-    need_si_install = 'ModuleNotFoundError' in res_output or running_on_github_actions
+    need_si_install = 'ModuleNotFoundError' in res_output
```
h-mayorquin, line 490: This one, some of the sorters already have spikeinterface installed. Would it be useful to know if they still run with the latest version of spikeinterface. The option was kind of forcing the installation of the latest version of spikeinterface.

alejoe91: By construction they don't :) this is because so we make sure that we are running the correct SI version in the container. The only option to have SI installed is to manually extend the container. In that case, since it does require a fair amount of actions, I think that the user wants to use the version s/he installed

h-mayorquin: You mean that by construction they don't have spikeinteface installed?

In that case, the `need_si_install` will always be True? (and therefore the force install is superflous?).

Or am I misreading?

alejoe91: The containers that we ship are without spikeinterface. However a user could make a docker container (say `my-docker-id/my-container` with the following Dockerfile:

```
FROM spikinterface/kilosort2-compiled-base:latest

pip install spikeinterface==0.97
```

If the user runs now `ss.run_sorter("kilosort2", docker_image=my-docker-id/my-container:latest")`, SpikeInterface is installed and explicitly pinned to a certain version, so we don't reinstall. If something fails, it's not our fault ;)

h-mayorquin: OK, thanks for explaining it to me.

But that means that for all the sorters that we are testing here in the CI SHOULD install spikeinterface? I am asking because when I was testing this locally I added that bool because some the sorters were not installing spikeinterface. Maybe it was a problem on my machine only.

After we merge this we can verify this.

h-mayorquin (approved):
This looks good to me, maybe @samuelgarcia wants to check this before?

samuelgarcia (approved):

samuelgarcia:
OK for me.
I merge it.

Maybe one day we should refactor a bit to provide an explicit list of what to be installed in `run_sorter_container()` rather that env variable.
Also we could force some SI version depending the sorter generation (klusta need python3.7 and so old SI version).

Issue #1234, opened 2023-01-30
Save function using up all the memory


ziyuw2:
Hi there,
I don't know if this has been brought to your attention before, but the save function seemed to use up all the memory, resulted in my pc crashing three times (16gb of ram) when I ran this function, and my lab mate @vncntprvst 's pc crashing once (32gb of ram). Is there a way for us to allocate only part of our memory for this function? Please let me know! Thank you for your time!

`if (base_folder / "preprocessed").is_dir():
    recording_saved = si.load_extractor(base_folder / "preprocessed")
else:
    recording_saved = recording_all_cmr.save(folder=base_folder / "preprocessed", **job_kwargs)`

vncntprvst:
Just adding that we tried saving using chunks, with something like `recording_all.save(format='binary', folder=r'C:\data_folder', n_jobs=10, chunk_duration='1s', progress_bar=True)`, but that did not alleviate the problem.

alejoe91:
Hi, the function should only use roughly 1s * number of jobs worth of memory..on Linux I'm 100% that is the case :) is this on windows? Can you give us some specs on the `recording_all` object?

samuelgarcia:
On windows it is more complicated because if the underlying format reader use memmap (which is the case for many format) then the chunk memory are released only if necessary and as soon there  enought ram available then memory chunks are NOT released until the memory is full.
So sometimes you can expect a high memory usage but normally no crash.

yger:
Actually, I think I encountered similar problems a while ago with spyking-circus and memmap on windows. The solution was to force the file to be open/close after each chunk is read, to ensure memory is flushed. This is not slower, and it prevents any memory leak

vncntprvst:
Thanks Pierre. It's great to know because for calcium imaging analysis we're likely to make greater use of memmaps, so this tip might help us across the board :)

samuelgarcia:
@yger : I know but this would make an enormous change in many reader in neo.

vncntprvst:
We may switch some processing to WSL or Linux platforms. In the meantime, as a workaround, is there a format that does not use memmap that we could convert recordings to before saving?

alejoe91:
You can try to use `zarr` (and you can even compress the data):

```
recording_saved = recording_all_cmr.save(folder=base_folder / "preprocessed", format='zarr', **job_kwargs)
```

I don't think that's a long-term solution though..running most of the sorters will require to re-save to binary..

[...]

PR #1391, opened 2023-03-03
Add `GeneratorRecording` for testing and diagnosing memory issues

h-mayorquin:
Recently we had some memory issues: #1234

As @samuelgarcia pointed out, this is more likely the result of [known problems](https://github.com/numpy/numpy/issues/13510) with memmap in numpy and we are discussing long-term solutions to this. 

That said, memory problems come both from reading and writing. We write recorder objects for many reasons including long-term storage, serialization, marshalling and reading from docker. The aim of this PR is to introduce a testing tool for memory related problems while writing `LazyRandomRecording`. 

The `LazyRandomRecording` generates traces samples of any desired size and structure on the fly. That is, it does not read and should not keep a reference to the data. In general, it can be used to stress-test the processing and writing parts of any pipeline in high memory load scenarios. This can be accomplished because it decouples reading (which is required in any other recorder that produces a lot of data) from processing or writing. As an example, a specific use case is to test how the `ChunkRecordingExecutor` handle traces chunks of different sizes in high memory load scenarios.

I added some tests and a convenience wrapper as that seems to be a standard practice that both @alejoe91 and @samuelgarcia. I am not completly satisfied with the name so any suggestions are welcome.

What do you guys think?

### DIFF
--- a/.github/workflows/core-test.yml
+++ b/.github/workflows/core-test.yml
@@ -30,22 +30,17 @@ jobs:
           # zarr is needed to test core
           pip install zarr  
           pip install -e .
-      - name: Test memory
+      - name: Test core with pytest
         run: |
-          pip show numpy
-          pytest -vv -rA spikeinterface/core/tests/test_generate.py 
+          pytest -vv -ra --durations=0 --durations-min=0.001 spikeinterface/core | tee report.txt; test ${PIPESTATUS[0]} -eq 0 || exit 1
         shell: bash  # Necessary for pipeline to work on windows
-      # - name: Test core with pytest
-      #   run: |
-      #     pytest -vv -ra --durations=0 --durations-min=0.001 spikeinterface/core | tee report.txt; test ${PIPESTATUS[0]} -eq 0 || exit 1
-      #   shell: bash  # Necessary for pipeline to work on windows
-      # - name: Build test summary
-      #   run: | 
-      #     pip install pandas
-      #     pip install tabulate
-      #     echo "# Timing profile of core tests in ${{matrix.os}}" >> $GITHUB_STEP_SUMMARY 
-      #     # Outputs markdown summary to standard output
-      #     python ./.github/build_job_summary.py report.txt >> $GITHUB_STEP_SUMMARY 
-      #     cat $GITHUB_STEP_SUMMARY 
-      #     rm report.txt
-      #   shell: bash  # Necessary for pipeline to work on windows
+      - name: Build test summary
+        run: | 
+          pip install pandas
+          pip install tabulate
+          echo "# Timing profile of core tests in ${{matrix.os}}" >> $GITHUB_STEP_SUMMARY 
+          # Outputs markdown summary to standard output
+          python ./.github/build_job_summary.py report.txt >> $GITHUB_STEP_SUMMARY 
+          cat $GITHUB_STEP_SUMMARY 
+          rm report.txt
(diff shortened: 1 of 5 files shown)


samuelgarcia:
Hi Ramon. This is super cool and very usefull.
I will make some feedback for the naming soon.

This random recording will be also usefull for generating fake recording efficiently.
I saw you put a seed but I think this will not generate two times the same random depending how many worker we have and also the chunk_size.
This is because we have one seed at the begining but the number of call to random will change depending the number of chunk and wowrker.
It would be super cool to have a strategy that make the seeding and the results independant of the n_jobs.

h-mayorquin:
> Hi Ramon. This is super cool and very usefull. I will make some feedback for the naming soon.
> 
> This random recording will be also usefull for generating fake recording efficiently. I saw you put a seed but I think this will not generate two times the same random depending how many worker we have and also the chunk_size. This is because we have one seed at the begining but the number of call to random will change depending the number of chunk and wowrker. It would be super cool to have a strategy that make the seeding and the results independant of the n_jobs.

Yes I wrote in the documentation that this is not consistent so it can only be used to test form and not values: 
https://github.com/SpikeInterface/spikeinterface/blob/e401bbeda319adf56e0c61b4e471f47a3e0cd0e7/spikeinterface/core/generate.py#L370-L373

My original intention was to revamp the `generate_recording_function` to make it lazy (see the branch name of the other PR that I made lol) however I soon stumbled upon this problem. I had the idea to map the start_frame:end_frame and channel_indeces (the signature of the `get_traces` function together with segment_index) to a specific seed so the results would be consistent. I tested this and it works BUT...

Then there is the rabbit hole:
`get_traces(star_frame:end_frame)` would generate some results that would be consistent (if you call `get_traces(star_frame:end_frame)` and then `get_traces(star_frame:end_frame)` again you will get the same results).

But if you call `get_traces(star_frame:end_frame + 1)` the results would be totally unrelated to  `get_traces(star_frame:end_frame)` while in principle they should only different in the last frame. 

I was that I was doing early optimization and I decided to leave that problem for later.

samuelgarcia:
I think we could have some thinkg like this.

```python
seed = an external number
rng = np.rnadom.default_rng([segment_index, start_frame])
traces = rng.random((num_sample, num_channel))
```

I have tested but I guess that for a given chunk_size we should have the same results independing the number of n_jobs which is not so bad.
No ?
Would you have time to test ?

h-mayorquin:
@samuelgarcia 
Yes, that is what I was trying to say above:
```python
unique_seed_hash = (start_frame, end_frame, segment_index, fix_class_seed)
rng = np.random.default_rng(seed=unique_seed_hash)
traces = self.rng.random(...)
```
Should work to generate consistent traces, but you still have the problem that I described above:
> But if you call get_traces(star_frame:end_frame + 1) the results would be totally unrelated to get_traces(star_frame:end_frame) while in principle they should only different in the last frame

Check the latest changes, I implemented a deterministic function. It now returns a deterministic generator very similar to the one in the `generator_function` (except that it has a different phase per channel). I think this is better for testing as no matter the chunk size, or how do your slice the frames the results would be the same. 

What do you think?

h-mayorquin:
@alejoe91 @samuelgarcia 
As discussed, this is ready to go and we will wait to implement the rest of the features in https://github.com/SpikeInterface/spikeinterface/issues/1413. 

However, the sorting view tests seems to be failing here, any idea on why this could be?

PR #1394, opened 2023-03-06
Remove to dict_mechanism in Motion Correction

### LINKED ISSUE #1345: 'Object of type CommonReferenceRecording is not JSON serializable' when saving recording
Hi,

After the latest update, when I try to do _recording.save(...)_ for a recording that had been filtered, phase corrected, CAR'd and motion corrected, I now get this error when trying to write it to disk. Is this a bug or am I doing  something wrong? Thanks for your help.

Cheers

Laurenz

h-mayorquin:
Related to #1345 and to #1235. 

This fixes the regresion that was introduced with #1325 and temporarily fixed in #1367. 

It adds a regression tests to motion correction in sorting components and the machinery in `check_json` so the recordings are made completley dumpable in json with that function. 

OK, this is working. This should close #1345. I will try to remove the rest of the `to_dict` mechanisms in a subsequent PR.

### DIFF
--- a/spikeinterface/core/core_tools.py
+++ b/spikeinterface/core/core_tools.py
@@ -1,6 +1,7 @@
 from pathlib import Path
 import os
 import sys
+import warnings
 import datetime
 from copy import deepcopy
 import gc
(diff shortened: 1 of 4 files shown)


spikeinterface/core/core_tools.py
```diff
@@ -11,6 +12,7 @@
 
 from .job_tools import (ensure_chunk_size, ensure_n_jobs, divide_segment_into_chunks, fix_job_kwargs, 
                         ChunkRecordingExecutor, _shared_job_kwargs_doc)
+import spikeinterface.core.base as base
```
samuelgarcia, line 15: When submodule partial import I thin I prefer this 
```suggestion
from spikeinterface.core.base import BaseExtractor
```

Issue #1395, opened 2023-03-06
Installation_tips Problem for Windows


zm711:
Howdy all, 

Since I have a windows for work I tested out the installation_tips for beginners and it errors out due to a Win32 PermissionError. For Windows the file access system doesn't release the raw files during the `check_your_install.py` so `shutil.rm` fails due to another process using the files. The rest of the test works fine, but it leaves all example data behind. I've drafted a pull request to fix this that I'll link to this issue.

PR #1396, opened 2023-03-06
Fix Windows Error in Installation Tips

zm711:
This will solve #1395. I just copied the `_clean` and put it into a separate .py file. Once the initial script ends the process releases the file permissions on Windows so running the `cleanup_for_windows.py` that I added worked fine on my machine (Windows 10). In order to prevent some shutil issues I also added in a check for Windows before running `_clean` in the `check_your_install.py`. 

I updated the readme to link to the additional file I added. 

Finally I added in documentation for Mac users (ie they don't use the Anaconda prompt they use Terminal just like Linux).

### DIFF
--- a/installation_tips/README.md
+++ b/installation_tips/README.md
@@ -13,7 +13,7 @@ This environment will install:
  * herdingspikes (not on windows)
 
 Kilosort, Ironclust and HDSort are MATLAB based and need to be installed from source.
-Klusta does not work anymore with python3.8 you should create a similar environment with python3.6
+Klusta does not work anymore with python3.8 you should create a similar environment with python3.6.
 
 ### Quick installation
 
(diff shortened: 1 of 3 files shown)


samuelgarcia:
Cool.
Does the full_spikeinterface_environment_windows.yml is still working on windows ?
We haven't updated it for a while.

zm711:
Yeah. It works perfectly except two things. Cleanup fails for this (as you now know). And then I linked an issue in the gui repo. Both in Windows and in mac the gui doesn't render quite right. So I attached a picture in my issue there. On Mac when I ran the test herdingspikes indicated it failed, so that might also need to be tweaked.

zm711:
Also forgot it has a bunch of the deprecation warnings, but I think that's been fixed (based on recent merging), just wasn't included when I ran the yaml to create the env. One example of the deprecation is below:

```
__init__.py:2804: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('google')`.
Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
```

alejoe91:
> Also forgot it has a bunch of the deprecation warnings, but I think that's been fixed (based on recent merging), just wasn't included when I ran the yaml to create the env. One example of the deprecation is below:
> 
> ```
> __init__.py:2804: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('google')`.
> Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
> ```

@zm711 this is due to Python3.8 I believe. MAybe we should update the environments to use 3.9?
Would you have time to test this? Otherwise we can do it ;)

alejoe91 (approved):

zm711:
@alejoe91 
Happy to test, but when I created the envi it was 3.9. I've copied some of the main packages installed in the env. Let me know if you want the full package list.

python                    3.9.16          h4de0772_0_cpython    conda-forge
spikeinterface            0.97.0                   pypi_0    pypi
spikeinterface-gui        0.6.0                    pypi_0    pypi

alejoe91:
@h-mayorquin tracked down the problem to the `datalad` package..we can live with that! 

Thanks again! Merging :)

PR #1397, opened 2023-03-06
Add specialized JsonEncoder for writing to json

h-mayorquin:
While working on #1394 and removing the rest of ot the `to_dict` mechanisms from pre and post processing as @samuelgarcia requested, I find myself finding problems with writing the `NumpyRecording` to json [*]. While I think they are solvable, the solution will require yet another modification of the  `check_json` function and will add even more complexity to it in the form of more nested cases.

Instead of continuing modifying our in-house jsonification mechanism `check_json` I would like to propose another solution for json-ification which will solve these problems. In this PR I show how we can easily implement a specific JSON encoder for spikeinterface objects (provisionally named `ExtractorJsonEncoder` open for naming suggestions). This Json encoder takes care of our most common problems when writing to json:

* It transforms the numpy arrays to a list version with the appropiate data-types. 
* It transforms datetime objects to their iso format.
* It transforms extractor objects to their `to_dict` representation.

The advantages of using this instead of using our own in-house `check_json` function are:
1)  tIt eliminates the nested checks for types that we have in `check_json` as the base class of `json.JSONEncoder` already takes care of this (compare the class added in this PR to all the branches in the `check_json` function)
2) The mechanism is already integrated into the json library from the standard libary which probably has some optimizations included. 
3) Provides json-nification of our objects with the standard API of the standard python library (using a json encoder) 

In short, I think this is more readable code that follows the conventions of the standard library and solves some of our problems with less lines of code and complexity. What do you think?

---
Edit:
Previous base issue message:

Related to https://github.com/SpikeInterface/spikeinterface/issues/1345 and to https://github.com/SpikeInterface/spikeinterface/issues/1235.

This fixes the regresion that was introduced with https://github.com/SpikeInterface/spikeinterface/pull/1325 and temporarily fixed in https://github.com/SpikeInterface/spikeinterface/pull/1367.

It adds a regression tests to motion correction in sorting components and the machinery in check_json so the recordings are made completley dumpable in json with that function.

OK, this is working. This should close https://github.com/SpikeInterface/spikeinterface/issues/1345. I will try to remove the rest of the to_dict mechanisms in a subsequent PR.


alejoe91:
Thanks @h-mayorquin, this is super cool. Let's discuss about it in the weekly meeting.

spikeinterface/core/core_tools.py
```diff
@@ -80,81 +81,67 @@ def write_python(path, dict):
...
+        # This should transforms integer, floats and bool to their python counterparts
+        if isinstance(obj, np.generic):
+            return obj.item()       
+                
+        if isinstance(obj, np.ndarray):
+            return obj.tolist()
+
+        if isinstance(obj, base.BaseExtractor):
+            return obj.to_dict()
+
```
samuelgarcia, line 102: Could we also add np.dtype ?

h-mayorquin: Good idea, I guess they should be their strings, right?

h-mayorquin: This is done. Check the last version of the tests.

h-mayorquin:
@samuelgarcia It is the error described in #1444 (which I will now close). Basically, I realized that sometimes the tests for the `GeneratorRecording` fail because memory usage decreased a bit during the test (some other memory is liberated). The test previously assumed that memory would only grow. I changed the assertion so the comparison would be symmetric.

Issue #705, opened 2022-06-14
kilosort3 on dandihub


bendichter:
I’m having trouble running kilosort3 singularity on dandi hub.

```python

!pip install git+https://github.com/SpikeInterface/spikeinterface


import spikeinterface.sorters as ss
from spikeinterface.core.testing_tools import generate_recording

recording = generate_recording(
    num_channels=32,
    sampling_frequency=30000.,  # in Hz
    durations=[120],
    set_probe=True,
    ndim=2,
)

recording2 = recording.save(folder="test_recording2")
```

```pytb
write_binary_recording with n_jobs 1  chunk_size None
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Input In [9], in <module>
      4 recording = generate_recording(
      5     num_channels=32,
      6     sampling_frequency=30000.,  # in Hz
   (...)
      9     ndim=2,
     10 )
     12 recording2 = recording.save(folder="test_recording2")
---> 13 sorting = ss.run_kilosort3(recording2, singularity_image=True, verbose=True)

File /opt/conda/lib/python3.9/site-packages/spikeinterface/sorters/runsorter.py:611, in run_kilosort3(*args, **kwargs)
    610 def run_kilosort3(*args, **kwargs):
--> 611     return run_sorter('kilosort3', *args, **kwargs)

File /opt/conda/lib/python3.9/site-packages/spikeinterface/sorters/runsorter.py:108, in run_sorter(sorter_name, recording, output_folder, remove_existing_folder, delete_output_folder, verbose, raise_error, docker_image, singularity_image, with_output, **sorter_params)
     95 common_kwargs = dict(
     96     sorter_name=sorter_name,
     97     recording=recording,
   (...)
    104     **sorter_params,
    105 )
    107 if docker_image or singularity_image:
--> 108     return run_sorter_container(
    109         container_image=docker_image if isinstance(docker_image, str) else singularity_image,
    110         mode="docker" if docker_image else "singularity",
    111         **common_kwargs,
    112     )
    114 return run_sorter_local(**common_kwargs)

File /opt/conda/lib/python3.9/site-packages/spikeinterface/sorters/runsorter.py:397, in run_sorter_container(sorter_name, recording, mode, container_image, output_folder, remove_existing_folder, delete_output_folder, verbose, raise_error, with_output, extra_requirements, **sorter_params)
    395 if use_gpu:
    396     if gpu_capability == 'nvidia-required':
[...]

PR #1398, opened 2023-03-06
Robust method for checking gpu-nvidia availability

h-mayorquin:
This is related to #705 and once we are done with this it probably should be closed.

This method works on dandi-hub when I tested it a while ago, the only problem is that it requires a dependency:
https://nvidia.github.io/cuda-python/overview.html

I am also not sure how to add this as a dependency.

What do you think?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,7 +113,6 @@ test = [
     # Sorters
     "python-cuda",
 
-
     ## sliding_nn
     "pymde",
     "torch",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,6 +77,7 @@ full = [
     "networkx",
     "distinctipy",
     "matplotlib",
+    "cuda-python",
 ]
 
 widgets = [
@@ -109,9 +110,6 @@ test = [
 
     ## install tridesclous for testing ##
     "tridesclous>=1.6.6.1",
-
-    # Sorters
-    "cuda-python",
 
     ## sliding_nn
     "pymde",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,6 +110,10 @@ test = [
     ## install tridesclous for testing ##
     "tridesclous>=1.6.6.1",
 
+    # Sorters
+    "python-cuda",
+
+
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
> This method works on dandi-hub when I tested it a while ago, 

@h-mayorquin Do you mean, with this fix, it would be possible to run a spike sorter that requires an 'NVIDIA GPU' on the T4 GPU instance from the DANDI Hub?

alejoe91:
I would be in favor of adding it to the `full` install, since it's really useful. @samuelgarcia do you agree?

samuelgarcia:
I am not super warn to had too much package in the full.
Maybe we are nto aligned, but for me "full" is more or less "main" (and maybe should be renamed) = main package super well maintain and that evrybody have more or less already.
So I have the tendency to limit as much as possible what is in the full to avoid dependency blowing
For me testing the GPU ability is at the moment a bit in the corner and mainly for sortingcomponents.
Maybe we could: 
  * rename "full" to "main" and keep it little
  * have a "internal_sorter" section that ould have cude/torch/pymde/... which are more specific.

All external sorter are super super hard to installed in the same env as SI mainly for dependency graph so we have to be carfull. No ?

alejoe91:
@samuelgarcia the way we check for GPU availability right now fails on systems that do not have access to the `nvidia-smi` command (e.g. dandihub). This prevents running GPU-based external sorters on those machines. 

Therefore, I think this needs to be a `full` requirement, since it mainly affects the `sorters` module (which is a `main` module)

alejoe91 (approved):

h-mayorquin:
> > This method works on dandi-hub when I tested it a while ago,
> 
> @h-mayorquin Do you mean, with this fix, it would be possible to run a spike sorter that requires an 'NVIDIA GPU' on the T4 GPU instance from the DANDI Hub?

@CodyCBakerPhD sorry for not answering before, this escaped me. What I meant is that the method correctly told me I had cuda when I tried it a while ago. I have not tested that the sorters can be run there. I will run them and come back to you.

PR #1405, opened 2023-03-08
Add super simple sorter based on hdscan

samuelgarcia:


src/spikeinterface/sorters/internal/simplesorter.py
```diff
@@ -0,0 +1,93 @@
...
+        waveforms_params['max_spikes_per_unit'] = None
+        waveforms_params['sparse'] = False
+        we = extract_waveforms(recording, sorting_temp, sorter_output_folder / "waveforms_temp",
+                            **waveforms_params, **job_kwargs)
+
+        # compute PCs
+        pc = compute_principal_components(we, n_components=3, mode="by_channel_global", sparsity=None, whiten=True)
+        pc_features = pc.get_projections(unit_id=0)
+        pc_features_flat = pc_features.reshape(pc_features.shape[0], -1)
+
```
h-mayorquin, line 82: I am wondering if we could use  the node `TemporalPCAProjection` here and run everything before the clustering as a pipeline.

h-mayorquin (commented):
Hey @samuelgarcia I think this is very useful and looks ready to me. Are you waiting for #1486 before this or why is still a draft?

src/spikeinterface/sorters/internal/simplesorter.py
alejoe91: But still we would need to fit it first, no?

h-mayorquin: Yes, we will avoid using calls to the `extract_waveforms` (as we will be using the corresponding) though. Plus, we will be keeping everyhing withing the sorting components module which I think makes sense for a sorter. Maybe I am missing something.

samuelgarcia: In that sepcific case using extract_waveforms is more efficient because it prepare in advance the memmap files instead of what the piepline do (accumulating  in list an dthen concatenate)
And so the compute_principal_components is straigforward.

samuelgarcia:
Leave this open for a while I will be bakc on this. Unless someone want to continue this.

h-mayorquin:
> Leave this open for a while I will be bakc on this. Unless someone want to continue this.

I could continue it, but what it would take? What steps do you think are missing?

samuelgarcia:
close by #2524

PR #1414, opened 2023-03-10
Adding a savgol denoiser for the waveforms

yger:
I just made a simple (but efficient) savgol denoiser to smooth waveforms on-the-fly. I also fixed some typos in neuralnet denoiser

https://en.wikipedia.org/wiki/Savitzky%E2%80%93Golay_filter

### DIFF
--- a/spikeinterface/sortingcomponents/tests/test_waveforms/test_savgol_denoiser.py
+++ b/spikeinterface/sortingcomponents/tests/test_waveforms/test_savgol_denoiser.py
@@ -1,0 +1,39 @@
+import pytest
+
+
+from spikeinterface.sortingcomponents.waveforms.savgol_denoiser import SavGolDenoiser
+from spikeinterface.sortingcomponents.peak_pipeline import (
+    ExtractDenseWaveforms,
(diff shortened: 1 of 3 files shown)


alejoe91 (approved):

alejoe91:
Thanks @yger 

Looks good to me! @samuelgarcia @h-mayorquin wanna have a look?

samuelgarcia:
OK for me also

h-mayorquin:
Looks good to me. Great that you could use the fixtures over there. As both @alejoe91 and @samuelgarcia approved this I am merging it now.

I am thinking that this does not need the recording extractor as an argument (as the relevant frequency is extracted from the parent waveform extractor). I think this argument also applies to the temporal PCA and the NN denoiser. I will remove them in a future refactor.

samuelgarcia:
Hi Ramon.
I am rather than for keeping the recording as an argument even if it is not need for some node.
Lets discuss this before a PR.

h-mayorquin:
Ah, I see, yeah, let's discuss it!

PR #1415, opened 2023-03-11
Add some preprocessing tools (general and LFP-oriented)

cwindolf:
Hi @samuelgarcia , here is a small draft PR with two preprocessing tools: a simple one (`depth_order`) which will save me a lot of typing, and one for averaging across channels at the same depth (or a dimension other than depth, `average_across_direction`). I'm curious if I have done all of the metadata bookkeeping as you would have done.

Let me know your thoughts -- once I'm going in the right direction I'll add something like a `directional_derivative(recording, direction='y', order=2)` for computing spatial derivatives.

### DIFF
--- a/spikeinterface/core/recording_tools.py
+++ b/spikeinterface/core/recording_tools.py
@@ -247,12 +247,13 @@ def get_chunk_with_margin(
             elif add_reflect_padding:
                 # in this case, we don't want to taper
                 traces_chunk = np.pad(
-                    traces_chunk,
-                    [(left_margin, right_margin), (0, 0)],
+                    traces_chunk.astype(dtype),
+                    [(left_pad, right_pad), (0, 0)],
                     mode="reflect",
                 )
             else:
-                assert False  # unreachable
+                # we need a copy to change the dtype
+                traces_chunk = np.asarray(traces_chunk, dtype=dtype)
 
     return traces_chunk, left_margin, right_margin
 
--- a/spikeinterface/core/recording_tools.py
+++ b/spikeinterface/core/recording_tools.py
@@ -301,7 +301,7 @@ def order_channels_by_depth(
             assert dim < ndim, "Invalid dimensions!"
             locations_to_sort += (locations[:, dim],)
         order_f = np.lexsort(locations_to_sort)
-    order_r = np.argsort(order_f)
+    order_r = np.argsort(order_f, kind="stable")
 
     return order_f, order_r
 
--- a/spikeinterface/core/recording_tools.py
+++ b/spikeinterface/core/recording_tools.py
@@ -150,6 +150,12 @@ def get_chunk_with_margin(
 ):
     """
     Helper to get chunk with margin
+
+    The margin is extracted from the recording when possible. If
+    at the edge of the recording, no margin is used unless one
+    of `add_zeros` or `add_reflect_padding` is True. In the first
+    case zero padding is used, in the second case np.pad is called
+    with mod="reflect".
     """
     length = rec_segment.get_num_samples()
 
--- a/spikeinterface/core/recording_tools.py
+++ b/spikeinterface/core/recording_tools.py
@@ -1,8 +1,15 @@
 import numpy as np
 
 
-def get_random_data_chunks(recording, return_scaled=False, num_chunks_per_segment=20, 
-                           chunk_size=10000, concatenated=True, seed=0, margin_frames=0):
+def get_random_data_chunks(
+    recording,
+    return_scaled=False,
+    num_chunks_per_segment=20,
+    chunk_size=10000,
+    concatenated=True,
+    seed=0,
+    margin_frames=0,
+):
     """
     Exctract random chunks across segments
 
@@ -31,22 +38,30 @@ def get_random_data_chunks(recording, return_scaled=False, num_chunks_per_segmen
     # Should be done by changing kwargs with total_num_chunks=XXX and total_duration=YYYY
     # And randomize the number of chunk per segment weighted by segment duration
 
-    # check chunk size    
+    # check chunk size
     for segment_index in range(recording.get_num_segments()):
-        assert chunk_size < recording.get_num_samples(segment_index), (f"chunk_size is greater than the number "
-                                                                        f"of samples for segment index {segment_index}. "
-                                                                        f"Use a smaller chunk_size!")
+        assert chunk_size < recording.get_num_samples(segment_index), (
+            f"chunk_size is greater than the number "
+            f"of samples for segment index {segment_index}. "
+            f"Use a smaller chunk_size!"
+        )
 
     chunk_list = []
     for segment_index in range(recording.get_num_segments()):
         length = recording.get_num_frames(segment_index)
-        
-        random_starts = np.random.RandomState(seed=seed).randint(margin_frames, length - chunk_size - margin_frames, size=num_chunks_per_segment)
+
+        random_starts = np.random.RandomState(seed=seed).randint(
+            margin_frames,
+            length - chunk_size - margin_frames,
+            size=num_chunks_per_segment,
+        )
         for start_frame in random_starts:
-            chunk = recording.get_traces(start_frame=start_frame,
-                                         end_frame=start_frame + chunk_size,
-                                         segment_index=segment_index,
-                                         return_scaled=return_scaled)
+            chunk = recording.get_traces(
+                start_frame=start_frame,
+                end_frame=start_frame + chunk_size,
+                segment_index=segment_index,
+                return_scaled=return_scaled,
+            )
             chunk_list.append(chunk)
     if concatenated:
         return np.concatenate(chunk_list, axis=0)
@@ -59,7 +74,9 @@ def get_channel_distances(recording):
     Distance between channel pairs
     """
     locations = recording.get_channel_locations()
-    channel_distances = np.linalg.norm(locations[:, np.newaxis] - locations[np.newaxis, :], axis=2)
+    channel_distances = np.linalg.norm(
+        locations[:, np.newaxis] - locations[np.newaxis, :], axis=2
+    )
 
     return channel_distances
 
@@ -95,8 +112,8 @@ def get_closest_channels(recording, channel_ids=None, num_channels=None):
     for i in range(locations.shape[0]):
         distances = np.linalg.norm(locations[i, :] - locations, axis=1)
         order = np.argsort(distances)
-        closest_channels_inds.append(order[1:num_channels + 1])
-        dists.append(distances[order][1:num_channels + 1])
+        closest_channels_inds.append(order[1 : num_channels + 1])
+        dists.append(distances[order][1 : num_channels + 1])
 
     return np.array(closest_channels_inds), np.array(dists)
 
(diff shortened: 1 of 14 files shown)


cwindolf:
There are a couple of `# TODO` which I left in there where I have questions

samuelgarcia:
Salut Charlie.
Merci c'est cool.
I think we should discuss how we handle the new channel_ids, probe, contact shape, and properties after this averaging.

spikeinterface/preprocessing/depth_order.py
```diff
@@ -0,0 +1,42 @@
...
+    installed = True
+
+    def __init__(
+        self, parent_recording, channel_ids=None, dimensions=("x", "y")
+    ):
+        order_f, order_r = order_channels_by_depth(
+            parent_recording, channel_ids=channel_ids, dimensions=dimensions
+        )
+        channel_ids = parent_recording.get_channel_ids()
+        reordered_channel_ids = [channel_ids[o] for o in order_f]
```
alejoe91, line 32:
```suggestion
        reordered_channel_ids = parent_recording.channel_ids[order_f]
```
Channel ids are numpy array :)

cwindolf: Got it, thanks!

cwindolf:
I think this is ready for a review if @samuelgarcia or @alejoe91 want to take a look?

cwindolf:
By the way, not entirely related, I just noticed a behavior which confused me for a while until I figured it out.

I was running something like this, where I'm accidentally reading off the edge of the recording.
```python
rec = NumpyRecording(np.ones((30000, 1)), 30000)
rec = highpass_filter(rec)
rec.get_traces(start_frame=40000, end_frame=50000)
```
This leads to an error coming out of `sosfiltfilt`: `ValueError: The length of the input vector x must be greater than padlen, which is 18.` But, the issue is just that `get_traces` is passing an empty array into `sosfiltfilt`, which is not filtfilt's fault.

I can add some guards against this -- check if the traces are empty before running preprocessors like `filter` that would error. But, I was wondering, maybe there is a better way to deal with this issue, since probably a lot of preprocessors would also raise misleading errors in this case.

spikeinterface/core/recording_tools.py
```diff
@@ -95,30 +112,42 @@ def get_closest_channels(recording, channel_ids=None, num_channels=None):
...
+    rec_segment,
+    start_frame,
+    end_frame,
+    channel_indices,
+    margin,
+    add_zeros=False,
+    add_reflect_padding=False,
+    window_on_margin=False,
+    dtype=None,
+):
```
samuelgarcia, line 150: genial!

spikeinterface/core/recording_tools.py
```diff
@@ -167,41 +201,66 @@ def get_chunk_with_margin(rec_segment, start_frame, end_frame,
...
+                    # apply inplace taper on border
+                    taper = (
+                        1 - np.cos(np.arange(margin) / margin * np.pi)
+                    ) / 2
+                    taper = taper[:, np.newaxis]
+                    traces_chunk2[:margin] *= taper
+                    traces_chunk2[-margin:] *= taper[::-1]
+                traces_chunk = traces_chunk2
+            elif add_reflect_padding:
+                # in this case, we don't want to taper
```
samuelgarcia, line 248: Maybe add a waord in the doc about this, no ?

spikeinterface/core/recording_tools.py
```diff
@@ -167,41 +201,66 @@ def get_chunk_with_margin(rec_segment, start_frame, end_frame,
...
+                # we need a copy to change the dtype
+                traces_chunk = np.asarray(traces_chunk, dtype=dtype)
 
     return traces_chunk, left_margin, right_margin
 
 
-def order_channels_by_depth(recording, channel_ids=None, dimensions=('x', 'y')):
+def order_channels_by_depth(
+    recording, channel_ids=None, dimensions=("x", "y")
+):
```
samuelgarcia, line 263: formidable!

spikeinterface/preprocessing/average_across_direction.py
```diff
@@ -0,0 +1,154 @@
...
+        ----------
+        parent_recording : BaseRecording
+            recording to zero-pad
+        direction : str
+            Channels living at unique positions along this direction
+            will be averaged.
+        dtype : optional numpy dtype
+            If supplied, convert recording to this type before averaging.
+            If unset, if the recording is not a floating type, it is converted
+            to float32 by default, or kept as its floating type otherwise.
```
samuelgarcia, line 33: In almost all preprocessing the most often behavior is to keep the parents dtype when it is None.
What is the deep underlying reaon for forcing to float ?

spikeinterface/preprocessing/average_across_direction.py
```diff
@@ -0,0 +1,154 @@
...
+            dtype=self._dtype,
+        )
+
+        # average channels at each depth without resorting to for loop
+        # first, add in the traces from all of the channels at each position
+        # np.add.at is necessary -- it will add multiple times in the same
+        # position, which + and += do not do
+        np.add.at(
+            traces, (slice(None), self.same_along_dim_chans), parent_traces
+        )
```
samuelgarcia, line 143: smart

spikeinterface/core/recording_tools.py
```diff
@@ -229,18 +288,20 @@ def order_channels_by_depth(recording, channel_ids=None, dimensions=('x', 'y')):
     locations = locations[channel_inds, :]
 
     if isinstance(dimensions, str):
-        dim = ['x', 'y', 'z'].index(dimensions)
+        dim = ["x", "y", "z"].index(dimensions)
```
samuelgarcia, line 291: @alejoe91 
Even it is unrelated to this PR, I think we have an error in this.
What we want by default is to short top to bottom and then left to right.
no ?
this lexsort do not do this because we should flip "x" otherwise we short from bottom to top and then left to right.
Am I wrong ?

spikeinterface/preprocessing/depth_order.py
```diff
@@ -0,0 +1,41 @@
...
+        If tuple, it sorts the locations in two dimensions using lexsort.
+        This approach is recommended since there is less ambiguity, by default ('x', 'y')
+    """
+    name = "depth_order"
+    installed = True
+
+    def __init__(
+        self, parent_recording, channel_ids=None, dimensions=("x", "y")
+    ):
+        order_f, order_r = order_channels_by_depth(
```
samuelgarcia, line 28: can you check that the ordering we have do what you want ?

spikeinterface/preprocessing/directional_derivative.py
```diff
@@ -0,0 +1,155 @@
...
+        dim = ["x", "y", "z"].index(direction)
+        if dim > parent_channel_locations.shape[1]:
+            raise ValueError(
+                f"Direction {direction} not present in this recording."
+            )
+
+        # float32 by default if parent recording is integer
+        dtype_ = dtype
+        if dtype is None and recording.dtype.kind != "f":
+            dtype_ = "float32"
```
samuelgarcia, line 54: same here.

samuelgarcia:
Hi Charlie.
This is ok for me.
Maybe we can discuss the behavior of dtype=None a bit more.

@alejoe91 : I have important refactoring in normalized_scale.py in #1437 maybe we will have conflict.

spikeinterface/preprocessing/normalize_scale.py
```diff
@@ -40,40 +45,54 @@ class NormalizeByQuantileRecording(BasePreprocessor):
...
+        self,
+        recording,
+        scale=1.0,
+        median=0.0,
+        q1=0.01,
+        q2=0.99,
+        mode="by_channel",
+        dtype="float32",
+        **random_chunk_kwargs
+    ):
```
samuelgarcia, line 73: Incredible!

spikeinterface/preprocessing/normalize_scale.py
```diff
@@ -226,29 +275,41 @@ def __init__(self, recording, mode="median+mad",
...
+)
 scale = define_function_from_class(source_class=ScaleRecording, name="scale")
-center = define_function_from_class(source_class=CenterRecording, name="center")
-zscore = define_function_from_class(source_class=ZScoreRecording, name="zscore")
+center = define_function_from_class(
+    source_class=CenterRecording, name="center"
+)
+zscore = define_function_from_class(
+    source_class=ZScoreRecording, name="zscore"
+)
```
samuelgarcia, line 315: truely stunning

spikeinterface/core/recording_tools.py
alejoe91: Lex sort goes backwards ;) https://numpy.org/doc/stable/reference/generated/numpy.lexsort.html

spikeinterface/core/recording_tools.py
cwindolf: merci

spikeinterface/core/recording_tools.py
cwindolf: Yup, it is the way we want, I was confused too and checked it

spikeinterface/core/recording_tools.py
cwindolf: Added something, thanks!

cwindolf:
Thank you for the review! Sorry again for all the formatting changes. Bad form!

spikeinterface/preprocessing/average_across_direction.py
CodyCBakerPhD: To clarify, numpy _calculations_ promote to `float` where it makes sense to do so, but numpy _array storage_ is always fixed in its allocation to disk and will therefore always _overflow_ for `int` types

As I read the docstring (though I could be very wrong), the `dtype` refers here to the storage pattern of the time series recording, right?

cwindolf: There are two dtypes involved here. The first is the dtype of the parent recording (before the preprocessing), and the second is the dtype after preprocessing. The parent dtype could be an integer type, and as you say, numpy will promote this when calculating the mean to a float type. This preprocessor does the same, so the output of the preprocessor in the case of a int parent recording and when passing `dtype=None` will be a recording with float32 type, extracted on the fly from the underlying parent recording.

CodyCBakerPhD: Hmm... I see now

Examples go a long way to conveying meaning

And specific explanation of *why* you chose to do it that particular way (maintaining floating point accuracy vs. saving memory space vs. speed of processing) would also go a long way

cwindolf:
Thanks!

I just tried seeing what happens if the preprocessors keep the parent recording when `dtype=None` (but `dtype="float32"` by default). `directional_derivative` is happy, but the averaging hits an error in the division when dividing by the number of channels:
```
numpy.core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'divide' output from dtype('float64') to dtype('int16') with casting rule 'same_kind'
```
So, in that case some type promotion is necessary. But, I'm happy to leave `dtype="float32"` by default and let users hit this error if they try `dtype=None`. @CodyCBakerPhD @samuelgarcia is that okay with you? This behavior is in the commit I just pushed.

samuelgarcia:
Not to_dict is not needed anymore. This is done only when going to json. But for multiprocessing then the pickle is use instead which is fatser.

alejoe91:
yep :)

alejoe91 (approved):

PR #1416, opened 2023-03-13
Update SV curation mechanism

alejoe91:
- get rid of legacy `JOT`
- use `GH` handle instead

### DIFF
--- a/spikeinterface/curation/tests/test_sortingview_curation.py
+++ b/spikeinterface/curation/tests/test_sortingview_curation.py
@@ -14,6 +14,7 @@ if hasattr(pytest, "global_test_folder"):
 else:
     cache_folder = Path("cache_folder") / "curation"
 
+parent_folder = Path(__file__).parent
 
 ON_GITHUB = bool(os.getenv('GITHUB_ACTIONS'))
 KACHERY_CLOUD_SET = bool(os.getenv('KACHERY_CLOUD_CLIENT_ID')) and bool(os.getenv('KACHERY_CLOUD_PRIVATE_KEY'))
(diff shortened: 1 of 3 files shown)


alejoe91:
@samuelgarcia @h-mayorquin 

this will fix the failing codecov tests: https://github.com/SpikeInterface/spikeinterface/actions/runs/4397330223/jobs/7700390707

samuelgarcia:
Hi camarade.
If I understand we need an extra repo just have a hard saved json file for curation for testing ?
Why do'nt we included this little JSON file in the actual repo ? Will thsi file change often ?

alejoe91:
> Hi camarade. If I understand we need an extra repo just have a hard saved json file for curation for testing ? Why do'nt we included this little JSON file in the actual repo ? Will thsi file change often ?

No it won't change! I can actually point to the same file! Thanks for the suggestion

spikeinterface/curation/tests/test_sortingview_curation.py
```diff
@@ -44,44 +44,47 @@ def test_sortingview_curation():
...
-    assert len(sorting_curated_sha_accepted.unit_ids) == 3
-    assert len(sorting_curated_sha_mua.unit_ids) == 6
-    assert len(sorting_curated_sha_mua1.unit_ids) == 5
+    assert len(sorting_curated_json_accepted.unit_ids) == 3
+    assert len(sorting_curated_json_mua.unit_ids) == 6
+    assert len(sorting_curated_json_mua1.unit_ids) == 5
+
+    # from GH
+    # curated link: https://figurl.org/f?v=gs://figurl/spikesortingview-10&d=sha1://bd53f6b707f8121cadc901562a89b67aec81cc81&label=SpikeInterface%20-%20Sorting%20Summary&s={%22sortingCuration%22:%22gh://SpikeInterface/test-sv-sorting-curation/main/test/curation-test.json%22}
+    gh_uri = "gh://SpikeInterface/spikeinterface/master/spikeinterface/curation/tests/sv-curation-test.json"
```
h-mayorquin, line 71: Isn't this pointing this to something that does not exist yet? (this is not yet in master). Is that the reason why the testing time decreased so drastically? 

Should we still keep the commented code above? If so, could you write a longer comment explaining what it is to future onlookers.

h-mayorquin (approved):
This seems OK to me.

PR #1420, opened 2023-03-15
Add `mean_fr_ratio_thresh` kwarg to presence_ratio metric

TomBugnon:


spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -86,8 +86,8 @@ def compute_firing_rates(waveform_extractor):
     return firing_rates
 
 
-def compute_presence_ratios(waveform_extractor, bin_duration_s=60):
-    """Calculate the presence ratio, representing the fraction of time the unit is firing.
+def compute_presence_ratios(waveform_extractor, bin_duration_s=60.0, mean_fr_ratio_thres=0.0):
```
alejoe91, line 89:
```suggestion
def compute_presence_ratios(waveform_extractor, bin_duration_s=60.0, mean_fr_ratio_thresh=0.0):
```

@TomBugnon I think that in other places in the API we use `thresh` instead of `thres` for abbreviating `threshold`. Could you rename the parameter?

spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -121,11 +121,11 @@ def compute_presence_ratios(waveform_extractor, bin_duration_s=60.0, mean_fr_rat
     num_bin_edges = total_length // bin_duration_samples + 1
     bin_edges = np.arange(num_bin_edges) * bin_duration_samples
 
-    mean_fr_ratio_thres = float(mean_fr_ratio_thres)
-    if mean_fr_ratio_thres < 0:
+    mean_fr_ratio_thresh = float(mean_fr_ratio_thresh)
+    if mean_fr_ratio_thresh < 0:
```
alejoe91, line 125: Should we have an upper bound too? At least `0 <= mean_fr_ratio_thresh <= 1`

alejoe91 (changes requested):
Thanks @TomBugnon 

One last comment/suggestion :)

spikeinterface/qualitymetrics/misc_metrics.py
TomBugnon: Yes why not, I didn't put it in because it's not logically required and I thought it might be useful in some case to set a high threshold (eg to catch units with very asymmetrical firing). But again there's probably another much more mathematically sound way of setting a threshold... As you see fit. Maybe a warning if >1?

samuelgarcia:
OK pour moi.

TomBugnon:
How about this?

alejoe91 (approved):

PR #1425, opened 2023-03-16
Change default value for end_frame in FrameSliceSorting

TomBugnon:
Hi
In FrameSliceSorting:
If None (default), the end_frame is now either:
    - The total number of samples, if a recording is assigned to the sorting
    - The maximum spike frame + 1, otherwise

This ensures that `Sorting.frame_slice(start_frame=None, end_frame=None)` not only 
returns all the spikes (as previously), but also that the sortings' registered recording remains complete.

Note that the following configs are now legal (and return empty sortings)
- `start_frame < end_frame <= min_spike_time`
- `max_spike_time < start_frame < end_frame`

### DIFF
--- a/spikeinterface/core/frameslicesorting.py
+++ b/spikeinterface/core/frameslicesorting.py
@@ -2,6 +2,7 @@ import numpy as np
 import warnings
 
 from .basesorting import BaseSorting, BaseSortingSegment
+from .waveform_tools import has_exceeding_spikes
 
 
 class FrameSliceSorting(BaseSorting):
@@ -53,6 +54,11 @@ class FrameSliceSorting(BaseSorting):
             assert start_frame <= parent_n_samples, (
                 "`start_frame` should be smaller than the sortings total number of samples."
             )
+            if has_exceeding_spikes(parent_sorting._recording, parent_sorting):
+                raise ValueError(
+                    "The sorting object has spikes exceeding the recording duration. You have to remove those spikes "
+                    "with the `spikeinterface.curation.remove_excess_spikes()` function"
+                )
         else:
             # Pull df end_frame from spikes
             if end_frame is None:
--- a/spikeinterface/core/frameslicesorting.py
+++ b/spikeinterface/core/frameslicesorting.py
@@ -11,6 +11,25 @@ class FrameSliceSorting(BaseSorting):
 
     Do not use this class directly but use `sorting.frame_slice(...)`
 
+    When a recording is registered for the parent sorting,
+    a corresponding sliced recording is registered to the sliced sorting.
+
+    Note that the returned sliced sorting may be empty.
+
+    Parameters
+    ----------
+    parent_sorting: BaseSorting
+    start_frame: None or int
+        Earliest included frame in the parent sorting(/recording).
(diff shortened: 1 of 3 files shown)


alejoe91 (approved):

alejoe91:
Thanks @TomBugnon 

Looks great!

h-mayorquin:
Would it be possible to have the tests that correspond to this new behavior separated from `test_FrameSliceSorting`? I think that we should start implementing something that looks more like unit-testing. 

Right now, most of our tests look like scripts where the set-up and the tests are not clearly separated and makes them hard to read. 

@TomBugnon Could you separte the tests of the new functionality so I can refactor them later?

TomBugnon:
Hi @h-mayorquin 
I'm not sure I understand precisely what you would like me to do, could you write the outline of what you'd like the frameslicesorting tests to look like?
This is not really a new functionality but a change in how the default `end_frame` is set in FrameSliceSorting, based on the parent sorting's features. I'm not sure where it would belong if not in test_frameslicesorting.
best

h-mayorquin:
@TomBugnon 
In an ideal unit test framework every assertion should be its own test function. This can be taken too far but is not the case here. Concrelyt, If you look at `test_FrameSliceSorting` right now there is plenty of assertions all over the place (it looks more like a script). What I am asking you as a favor (we are quite grateful for your contribution and I don't want to get in this way of this merging!) is the following:
* Erase any old assertion that corresponds to the old behavior in the current `test_FrameSliceSorting` test.
* Add a new testing function (e.g. `test_FrameSliceSortingEnds` or whatever name your fancy) and wite the assertions that corresponds to the new behavior there.

This will imply some code duplication now. That's fine, I would go and refactor this later and use proper fixtures for the setup.

TomBugnon:
Yes makes sense, thanks. Will do shortly.

spikeinterface/core/frameslicesorting.py
```diff
@@ -11,27 +12,67 @@ class FrameSliceSorting(BaseSorting):
...
+                    "with the `spikeinterface.curation.remove_excess_spikes()` function"
+                )
         else:
-            assert end_frame > start_frame, "'start_frame' must be smaller than 'end_frame'!"
+            # Pull df end_frame from spikes
+            if end_frame is None:
+                max_spike_time = 0
+                for u in parent_sorting.get_unit_ids():
+                    max_spike_time = np.max([max_spike_time, np.max(parent_sorting.get_unit_spike_train(u))])
+                end_frame = max_spike_time + 1
```
samuelgarcia, line 68: This has a high cost at every  init which is not so godd for paralel processing.
If we have 500units and n_jobs=80 then we will make 40k np.max at every multiprocessing!!!
We should find a smarter way.
Maybe defer this computation at the last moement (get_traces).
@alejoe91 @TomBugnon what do you think ?

samuelgarcia: Oups. No forget it.
The en_frame is store in the kwargs

samuelgarcia:
OUps sorry I read this to quickly. This is totally ok for me.

PR #1430, opened 2023-03-19
fix typo in class ClipRecording

oaaij-gnahz:
Hi,

I found a typo while using the `preprocessing.clip`: when setting its `_kwargs`, the `a_min` property is set to `a_max` by accident. This typo has no effect when objects stay in memory; however, if the `ClipRecording` object is dumped to JSON and read from disk again (such as when we [run sorter from folder using MountainSort4](https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/sorters/external/mountainsort4.py#L91)), this causes the return of `get_trace()` to be a constant value.

Thanks!


alejoe91:
Thanks for spotting it @oaaij-gnahz

alejoe91 (approved):

PR #1439, opened 2023-03-21
Add min_spikes to slidingRP_violations (support unit with no spike)

TomBugnon:
Hi,
Another little PR!
All the metrics support units with N=0 spikes (which could happen eg when slicing a sorting), and in general return np.nan, except the slidingRP_violations which errors.
To avoid the error and for consistency with other metrics I added this `min_spikes` parameter
Best


alejoe91 (approved):

alejoe91:
Makes sense, thanks @TomBugnon

PR #1441, opened 2023-03-21
Refactor `test_detect_injected_correlation`

h-mayorquin:
@alejoe91 mentioned that this is a test that constantly fails. I added some determinism to the tests. Plus, I think it is easier to test for the location of the peak (which should match the delta of the injected noise) instead of the maximum of the peak (tbh, I had a hard time interpreting why should the peak_value should be around `np.median(cc_01) * 0.7 + num_injected_spikes`).

What do you think? 
I

### DIFF
--- a/spikeinterface/postprocessing/tests/test_correlograms.py
+++ b/spikeinterface/postprocessing/tests/test_correlograms.py
@@ -197,7 +197,7 @@ def test_detect_injected_correlation():
         peak_location_02_ms = bins[np.argmax(cc_10)]
         
         sampling_period_ms = (1000.0 / sampling_frequency)
-        assert abs(peak_location_01_ms) - 1.4 < sampling_period_ms
+        assert abs(peak_location_01_ms) - injected_delta_ms< sampling_period_ms
         assert abs(peak_location_02_ms) - 1.4 < sampling_period_ms
         
     #     import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


spikeinterface/postprocessing/tests/test_correlograms.py
```diff
@@ -191,41 +190,38 @@ def test_detect_injected_correlation():
...
+        peak_location_02_ms = bins[np.argmax(cc_10)]
         
-        half_bin_ms = np.mean(np.diff(bins)) / 2.
-        
-        # check that we have a peak in the CC 01
-        expected_peak_value = np.median(cc_01) * 0.7 + num_injected
-        assert np.max(cc_01) > expected_peak_value
-        assert np.max(cc_10) > expected_peak_value
+        sampling_period_ms = (1000.0 / sampling_frequency)
+        assert abs(peak_location_01_ms) - 1.4 < sampling_period_ms
```
samuelgarcia, line 200: Maybe even this
```suggestion
        assert abs(peak_location_01_ms) - injected_delta_ms< sampling_period_ms
```

samuelgarcia:
`np.median(cc_01) * 0.7 + num_injected_spikes` is bit kriptic.
The idea is have more than the basal coorelation minus a factor + injected (which is known).
Indeed the new test is really easier to read even if it is a bit diferrents.

h-mayorquin:
Thanks for explaining @samuelgarcia. I guess it is the factor the one that I don't know where it is coming from.

Are you fine with having the new test instead?

samuelgarcia:
The factor was very empiric.
Lets go with this!

samuelgarcia (approved):

PR #1443, opened 2023-03-21
Add min_fr kwarg to `nn_isolation` and `nn_noise_overlap`

TomBugnon:
Hi,
This allows to ignore clusters based on both number of spikes and firing rate in `nn_isolation` and `nn_noise_overlap`.

The rationale is that
- min_spikes: ensures that the units considered in have enough spikes to accurately compute PCs
- min_fr: allows one to include units that were oversplit, when only a small relative number of spikes were wrongly clustered

`min_fr = 0` by default so this doesn't change default behavior.
Best


alejoe91 (approved):

alejoe91:
Thanks @TomBugnon

PR #1444, opened 2023-03-21
Make memory tests for `GeneratorRecording` symmetric

h-mayorquin:
I noticed in #1397 that the memory tests for the `GeneratorRecording` are falling because I am only considering the possibility of the memory being a little bit higher. I am re-writing this to have the correct base.


h-mayorquin:
Closing as the changes were contained in #1397

PR #1457, opened 2023-03-27
[Do not merge] Testing windows memory again

h-mayorquin:
As mentioned in some discussion the version of `lazy_recording` that was used in the previous test is not up to the task. I am trying here again to see if the multiprocessing works on windows as it shoud.

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

### DIFF
--- a/.github/workflows/core-test.yml
+++ b/.github/workflows/core-test.yml
@@ -1,43 +1,43 @@
-name: Testing core
+# name: Testing core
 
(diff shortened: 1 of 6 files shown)


h-mayorquin:
Closing this as it is replaced by https://github.com/SpikeInterface/spikeinterface/pull/1489.

Issue #1458, opened 2023-03-28
Path type check is too restrictive in `recursive_path_modifier` method


miketrumpis:
Sometimes attributes with "path" in the name have a `None` value, but there is no condition for None in this method: `spikeinterface.core.core_tools.recursive_path_modifier`

https://github.com/SpikeInterface/spikeinterface/blob/a5942a1cb540c6abb8876b01136044eca39d4842/spikeinterface/core/core_tools.py#L794-L804

For example, a `NwbRecordingExtractor` has a `stream_cache_path` that defaults to `None`.

Made a quick PR for this

alejoe91:
Fixed by #1459

PR #1459, opened 2023-03-28
Allow a None value for paths

miketrumpis:
PR for #1458

### DIFF
--- a/spikeinterface/core/core_tools.py
+++ b/spikeinterface/core/core_tools.py
@@ -793,8 +793,10 @@ def recursive_path_modifier(d, func, target='path', copy=True):
     else:
         for k, v in d.items():
             if target in k:
-                # paths can be str or list of str
-                if isinstance(v, str):
+                # paths can be str or list of str or None
+                if v is None:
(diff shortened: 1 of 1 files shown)


samuelgarcia (approved):

samuelgarcia:
thank you Mike.

spikeinterface/core/core_tools.py
```diff
@@ -793,8 +793,10 @@ def recursive_path_modifier(d, func, target='path', copy=True):
     else:
         for k, v in d.items():
             if target in k:
-                # paths can be str or list of str
-                if isinstance(v, str):
+                # paths can be str or list of str or None
+                if v is None:
+                    continue
+                if isinstance(v, (str, Path)):
```
h-mayorquin, line 799: Shouldn't this be an `elif`?

alejoe91 (approved):

spikeinterface/core/core_tools.py
h-mayorquin: True!

PR #1460, opened 2023-03-29
Fix sortingview timeseries bug

alejoe91:
This bug is due to a rounding of the traces shape when applying the zoom for the sortingview backend.

Fixed by grabbing the correct shapes after applying the zoom


h-mayorquin (commented):
This  looks good to me.

h-mayorquin (approved):

PR #1468, opened 2023-03-31
Localization from templates

yger:
This is a method to localize the peaks inspired by KS, by convolving the peaks with a large number of templates, at a finer spatial resolution


alejoe91:
That was quick 😂

PR #1475, opened 2023-04-03
Hide Y axis when plot_timeseries without channel labels

JoeZiminski:
Just a small suggestion, when first using plot_timeseries (matplotlib backend) I was a little confused by the y-axis numbers and their relationship to channels, until I saw the `show_channel_ids=True` option.

If I understand correctly, the y-axis values on the plots are arbitrary if channel ids are not shown. If so, it just be nice to just hide them as proposed in this PR (images below)

before PR (map):

image

after PR (map):

image

before PR (line):

image

after PR (line):

image


alejoe91 (approved):
Thanks @JoeZiminski

PR #1485, opened 2023-04-04
Add amplitude_scaling implementation

alejoe91:
Implement `amplitude_scalings` extension to compute scaling of each spike with respect to the template.

Currently using the `slope` of the `scipy.stats.linregress` function.

A possible extension would be to instead of using spike by spike regression, to gather all spikes in a nighborhood and fit colliding spikes.


alejoe91:
@TomBugnon here is a first implementation of the amplitude scalings (note that we are moving things to the `main` branch and the `src/spikeinterface` package organization)

Currently, the implementation is naive, since it doesn't account for spike collisions, but it accounts for multiple channels in a user-defined sparsity.
Nevertheless, it should allow us to test whether the amp cutoff metric is dependent on scaling VS absolute amplitudes.

src/spikeinterface/postprocessing/amplitude_scalings.py
```diff
@@ -0,0 +1,304 @@
...
+            cut_out_start = left + sample_centered - cut_out_before
+            cut_out_end = left + sample_centered + cut_out_after
+            if sample_index - cut_out_before < 0:
+                local_waveform = traces_with_margin[:cut_out_end, sparse_indices]
+                template = template[cut_out_before - sample_index :]
+            elif sample_index + cut_out_after > end_frame + right:
+                local_waveform = traces_with_margin[cut_out_start:, sparse_indices]
+                template = template[: -(sample_index + cut_out_after - end_frame)]
+            else:
+                local_waveform = traces_with_margin[cut_out_start:cut_out_end, sparse_indices]
```
samuelgarcia, line 296: why this when we have margin ?

src/spikeinterface/postprocessing/amplitude_scalings.py
```diff
@@ -0,0 +1,304 @@
...
+    margin = worker_ctx["margin"]
+    return_scaled = worker_ctx["return_scaled"]
+
+    i0 = np.searchsorted(spikes["segment_ind"], segment_index)
+    i1 = np.searchsorted(spikes["segment_ind"], segment_index + 1)
+    spikes_in_segment = spikes[i0:i1]
+
+    i0 = np.searchsorted(spikes["segment_ind"], segment_index)
+    i1 = np.searchsorted(spikes["segment_ind"], segment_index + 1)
+    spikes_in_segment = spikes[i0:i1]
```
samuelgarcia, line 257: duplicate

src/spikeinterface/postprocessing/amplitude_scalings.py
```diff
@@ -0,0 +1,304 @@
...
+    unit_inds_to_channel_indices = worker_ctx["unit_inds_to_channel_indices"]
+    nbefore = worker_ctx["nbefore"]
+    cut_out_before = worker_ctx["cut_out_before"]
+    cut_out_after = worker_ctx["cut_out_after"]
+    margin = worker_ctx["margin"]
+    return_scaled = worker_ctx["return_scaled"]
+
+    i0 = np.searchsorted(spikes["segment_ind"], segment_index)
+    i1 = np.searchsorted(spikes["segment_ind"], segment_index + 1)
+    spikes_in_segment = spikes[i0:i1]
```
samuelgarcia, line 253: The segment slicing could be done in worker init once.

src/spikeinterface/postprocessing/amplitude_scalings.py
```diff
@@ -0,0 +1,304 @@
...
+                template = template[cut_out_before - sample_index :]
+            elif sample_index + cut_out_after > end_frame + right:
+                local_waveform = traces_with_margin[cut_out_start:, sparse_indices]
+                template = template[: -(sample_index + cut_out_after - end_frame)]
+            else:
+                local_waveform = traces_with_margin[cut_out_start:cut_out_end, sparse_indices]
+            assert template.shape == local_waveform.shape
+            local_waveforms.append(local_waveform)
+            templates.append(template)
+            linregress_res = linregress(template.flatten(), local_waveform.flatten())
```
samuelgarcia, line 300: Does using linregress do not make it too slow ?
Why not a simple scalar product to speedup ?

samuelgarcia:
Cool!!!
Did some comments on the fly, will have a deeper look later.

src/spikeinterface/postprocessing/amplitude_scalings.py
alejoe91: This is a smaller cut out to get a local waveform

src/spikeinterface/postprocessing/amplitude_scalings.py
alejoe91: But the workers could span multiple segments no?

src/spikeinterface/postprocessing/amplitude_scalings.py
alejoe91: Not it's actually quite fast :) happy to discuss options

src/spikeinterface/postprocessing/amplitude_scalings.py
samuelgarcia: yes but the first serahsorted can be done once for all.

src/spikeinterface/postprocessing/amplitude_scalings.py
samuelgarcia: yes but the margin ensure always the correct lenght no ?
When you get_chunk_with_margin(add_zeros=True)

alejoe91:
> Cool!!! Did some comments on the fly, will have a deeper look later.

Thanks Camarade ;) let's discuss tomorrow

yger:
Just trying to catch up here. What is the difference between such amplitudes and the ones provided via the matching engine? Is it different?

alejoe91:
The only difference is that this knew makes sure that all spikes are matched. Would there be an option for matching engines to not find new spikes and assign a weight to all existing spikes?

yger:
Not at the moment, but anyway this is likely to be slower user the engines, because of all the burden for collisions. I'm just not really getting the use case for such a functions.
If you have a sorting, it means that you got the spikes via the matching engine, am I right? Except in the case of GT, where maybe some spikes are there but not found via matching because of collisions. But then, the amplitudes for these spikes are not likely to be close to the ones of the templates.

samuelgarcia:
@yger of course this PR do not take care of collision.
The idea is to estimate distribution so even with 10% of spikes in collision the distrib will be easy to estimate.
We want some kind of amplitude-scaling cut-off metric.
Here it is the prostprocessing module so totally agnostic to the sorter and the template matching engine behind.
We want metrics that are sorter independant.

PR #1486, opened 2023-04-04
Refactor detect_peak()

samuelgarcia:
Refactor detect_peak() using a wrapper to transform detct class into a pipeline node.

Use the same function internally `run_pipeline()` for the two previous cases:
  * detect peak and compute node on the fly `detect_peak()`
    For this  `PeakDetectorWrapper` was added
  * compute node on already detected peaks `run_peak_pipeline()`
    For this a new node was added `PeakRetriever`

Some constrain at the moment :
  * the first node is necessarly PeakDetector or PeakRetriever
  * the border for detect peak is handle as an special case
  * propagate changes to localize peaks to make smater
  * propagate changes to peak features

Futur plan: 
  * remove `PeakDetectorWrapper` to go for instance only in peak detection classes
  * propagate changes to template mathcing

test_peak_detection.py and test_peak_piepline.py are passing are passing without changes.

@h-mayorquin @alejoe91 : if you want to read it now and give feedback before it gets too complicated.


src/spikeinterface/sortingcomponents/peak_detection.py
```diff
@@ -63,121 +65,76 @@ def detect_peaks(recording, method='by_channel', pipeline_nodes=None,
...
+    return outs
 
-    if extra_margin > 0:
-        peak_sample_ind += extra_margin
 
-    peak_dtype = base_peak_dtype
-    peak_amplitude = traces[peak_sample_ind, peak_chan_ind]
+class PeakDetectorWrapper(PeakDetector):
+    # transitory class to maintain instance based and class method based
+    # TODO later when in master: refactor in every old detector class:
```
h-mayorquin, line 110:
```suggestion
    # TODO later when in main: refactor in every old detector class:
```
; )

src/spikeinterface/sortingcomponents/peak_detection.py
```diff
@@ -63,121 +65,76 @@ def detect_peaks(recording, method='by_channel', pipeline_nodes=None,
...
+    job_name = f'detect peaks using {method}'
     if pipeline_nodes is None:
         squeeze_output = True
     else:
-        check_graph(pipeline_nodes)
-        extra_margin = max(node.get_trace_margin() for node in pipeline_nodes)
         squeeze_output = False
+        job_name += f'  + {len(pipeline_nodes)} nodes'
+
+        # because node are modified inplace (insert parent) they need to copy incase
```
h-mayorquin, line 91: What kind of error this could lead to? (modifying in place) Do we have a tests for that?

src/spikeinterface/sortingcomponents/peak_pipeline.py
```diff
@@ -211,11 +275,30 @@ def compute(self, traces, peaks):
         return sparse_wfs
 
 
+def find_parent_of_type(node, parent_type, unique=True):
```
h-mayorquin, line 278: I agree with what @alejoe91 said about this yesterday.

The code could simply return a list of the parents with the desired type.

`[node for node in parents if instance(node, parent_type)]`

And is the responsability of the each node class to test for uniquness. That follows the single responsability principle. That said, if you agree with this I can take care of this refactor later together with the changes in the waveform extractor. That is, I don't think this is a critical part of this PR.

src/spikeinterface/sortingcomponents/peak_pipeline.py
```diff
@@ -265,6 +344,108 @@ def run_peak_pipeline(recording, peaks, nodes, job_kwargs, job_name='peak_pipeli
...
+    worker_ctx['max_margin'] = max(node.get_trace_margin() for node in nodes)
+
+    return worker_ctx
+
+def _compute_peak_pipeline_chunk(segment_index, start_frame, end_frame, worker_ctx):
+    recording = worker_ctx['recording']
+    max_margin = worker_ctx['max_margin']
+    nodes = worker_ctx['nodes']
+
+    recording_segment = recording._recording_segments[segment_index]
```
h-mayorquin, line 362: Should we use the non-private method `select_segments`?

h-mayorquin (commented):
This looks good to me. The most difficult part is the handling of the outputs so I will take a look at that later today. 

Meanwhile,  some general comments here and there:
* I agree with @alejoe91 that if we wanted to make it more specific `run_pipeline` could be `run_node_pipeline` but I think both choices are equally good.
* Maybe to keep consistency with `WaveformExtractor` terminology,  `PeakRetriever` could be `PeakExtractor`. On the other hand, I actually prefer to change `WaveformExtractor` to `WaveformRetriever` because the word extractor is already overloaded in our API and makes ctrl+f search-like functionality harder.

src/spikeinterface/sortingcomponents/peak_detection.py
samuelgarcia: If you run two times the same piepline then this new nodes is add many times.

src/spikeinterface/sortingcomponents/peak_pipeline.py
samuelgarcia: Yes! better

samuelgarcia:
> This looks good to me. The most difficult part is the handling of the outputs so I will take a look at that later today.
> 
> Meanwhile, some general comments here and there:
> 
> * I agree with @alejoe91 that if we wanted to make it more specific `run_pipeline` could be `run_node_pipeline` but I think both choices are equally good.
> * Maybe to keep consistency with `WaveformExtractor` terminology,  `PeakRetriever` could be `PeakExtractor`. On the other hand, I actually prefer to change `WaveformExtractor` to `WaveformRetriever` because the word extractor is already overloaded in our API and makes ctrl+f search-like functionality harder.

I totally agree to move `WaveformExtractor` to `WaveformRetriever` in the piepline context because it is too ambiguous with the core class.

run_node_pipeline sounds good

src/spikeinterface/sortingcomponents/peak_detection.py
h-mayorquin: Thanks!

src/spikeinterface/sortingcomponents/peak_pipeline.py
h-mayorquin: Check the code in https://github.com/SpikeInterface/spikeinterface/blob/4cea780f4576ffb920e8afee358ff5ca468bf5a2/src/spikeinterface/sortingcomponents/waveforms/neural_network_denoiser.py#L22-L26

samuelgarcia: I known that it can be one line.
But the rules of finding the correct will be more and more complex (multi parent, one parent sevarl out), so I think it is really important to have a central function somwhere instead of copy/paste a one line logic (`[node for node in parents if instance(node, parent_type)]`) for instance unique or parents.

If you want topropose a refactoring for that function list comprehension based I am OK with it.
I made explicit loop to open the door to the complexcity.

h-mayorquin: I don't feel very strong about this. Your reasoning makes sense. Maybe let's wait for more cases to see what abstraction make sense.

samuelgarcia:
Ready on my side.

h-mayorquin:
The calls to `get_chunk_with_margin` for some reason....

src/spikeinterface/sortingcomponents/peak_pipeline.py
```diff
@@ -211,11 +275,30 @@ def compute(self, traces, peaks):
...
+    for parent in node.parents:
+        if isinstance(parent, parent_type):
+            parents.append(parent)
+    
+    if unique and len(parents) == 1:
+        return parents[0]
+    elif not unique and len(parents) > 1:
+        return parents[0]
+    else:
+        None
```
h-mayorquin, line 292:
```suggestion
        return None
```

samuelgarcia: thanks

src/spikeinterface/sortingcomponents/waveforms/savgol_denoiser.py
```diff
@@ -16,10 +16,10 @@ def __init__(
     ):
         super().__init__(recording, return_output=return_output, parents=parents)
         self.order = order
-        if self.parents is None or not (len(self.parents) == 1 and isinstance(self.parents[0], WaveformExtractorNode)):
-            exception_string = f"SavGolDenoiser should have a single {WaveformExtractorNode.__name__} in its parents"
-            raise TypeError(exception_string)
-        waveform_extractor = self.parents[0]
+
+        waveform_extractor = find_parent_of_type(self, WaveformRetrieverNode)
```
alejoe91, line 20: The parent could also be a `WaveformExtractorNode` if the waveforms are retrieved on the fly with the `run_node_pipeline`, right?

alejoe91: ok no that's fine, I misunderstood

alejoe91, line 20: We need to find another name, this is ambiguous since it seems like waveforms have already been saved and only retrieved (like the new `PeakRetriever`)

@h-mayorquin what about `WaveformsNode`? I think it's general enough, it has the concept of waveforms and of being a node. Ok?

alejoe91: I pushed the change, and extended the docstring for the base class ;)

alejoe91 (approved):

PR #1488, opened 2023-04-04
Set line-length to 120 and reformat

alejoe91:
@h-mayorquin as discussed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,8 @@ include = ["spikeinterface*"]
 namespaces = false
 exclude = ["spikeinterface.*.tests"]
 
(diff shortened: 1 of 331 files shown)


h-mayorquin (approved):
LGTM. 
Tests are passing.

samuelgarcia:
I think this a quite bad idea.
This reformating will make big confilct while there some big open PR that do string refactoring.

samuelgarcia:
We decided to make black later!

Having black in main will make the merging from master to main absolutly paifull.
There are enormous PR in the queue, the timing is not good at all.
Just after a release would be a better timing.

I vote to revert this for now in main.

PR #1491, opened 2023-04-05
First PR with main: WaveformThresholder

yger:
I wanted to try my first PR with main, but not able to add/run tests. Is it normal? This PR brings an extra node for pipelines that can be very useful. I called it a WaveformThresholder. Basically, given a certain feature, some channels can be set to 0 in the Waveforms, before downstreams computations (for localization for example). This is something we discuss before, and it is kind of an adaptive masking. Several points, however, occured to me while writing this Node (and also the previous ones). Maybe I misunderstood something with the pipelines, but how do we get the params for nodes higher in the hierarchy? I guess we should add a function to infer params from nodes before, no? Because imagine the following situation. DenseWaveformExtractor -> SavgolFiltering -> WaveformThresholder. I would like my thresholder to have access to nbefore (in the first node). So maybe we could have a function get_param_from_parents that will iteratively look up in all parents (going up in the chain) to seek for the param

### DIFF
--- a/src/spikeinterface/sortingcomponents/tests/test_waveforms/test_waveform_thresholder.py
+++ b/src/spikeinterface/sortingcomponents/tests/test_waveforms/test_waveform_thresholder.py
@@ -1,0 +1,34 @@
+import pytest
+
+
+from spikeinterface.sortingcomponents.waveforms.waveform_thresholder import WaveformThresholder
+from spikeinterface.sortingcomponents.peak_pipeline import (
+    ExtractDenseWaveforms,
+    ExtractSparseWaveforms,
+    WaveformExtractorNode,
+    PipelineNode,
+    run_peak_pipeline,
+)
+
+
+def test_savgol_denoising(mearec_recording, detected_peaks, chunk_executor_kwargs):
+    recording = mearec_recording
+    peaks = detected_peaks
+
+    # Parameters
+    ms_before = 1.0
+    ms_after = 1.0
+
+    # Node initialization
+    extract_waveforms = ExtractDenseWaveforms(
+        recording=recording, ms_before=ms_before, ms_after=ms_after, return_output=True
+    )
+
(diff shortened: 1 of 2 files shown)


alejoe91:
Tests are running! Btw, after discussion with Sam we'll try another strategy to merge with main. We might ask you to make a PR to a different branch then ;)

yger:
Grrrr.... I think I'll do this PR in master, be warned.... And I'll discuss with @samuelgarcia about my issues/questions to access param in the upstream nodes of a given pipeline

h-mayorquin:
@yger To access parameters on the node parents you can do something like what you are doing

```python
parent_whose_properties_I_want = next(node in parents if isinstance(parent, "your_desired_class"))
property_I_want = parent_whose_properties_I_want.property_I_want
```

Are you thinking about something else?

yger:
Nice syntax. But will this loop iteratively over all the parents. Because in my understanding, if I have a a chain node_0, node_1, node_2, then parents are [], [node_0], [node_1]. Thus, when I am in node_2, the property I am looking for is not directly in my parents, in in the parents of my parents. Or should I list all parents for all nodes, and my understanding of the parents field is not correct? This is what is not clear to me...

yger:
Nope, definitely, there is something I don't get with the parents.... 

Coming back to my example WaveformExtractor -> Denoiser -> LocalizationModule What would be the parents for each of these nodes? My intuition would be, in order, [], [waveformextractor] and [waveformextractor, denoiser] am I right? If I do so, then I agree your patch with the next() looping over parents make sense, because indeed, information is stored by one of the parents. 

But then, running run_pipelines() on such a chain of three items [extractor, denoiser, localization] is not working with an error 

TypeError: compute() takes 4 positional arguments but 5 were given

yger:
I understood my mistakes, and I'll close this PR since @samuelgarcia told me that main will be rebased. All PipelineNodes that should expose waveforms to following nodes should inherit from WaveformExtractorNode instead of PipelineNode I think

PR #1492, opened 2023-04-05
max number of channels given by es

luiztauffer:


spikeinterface/extractors/nwbextractors.py
```diff
@@ -187,7 +187,9 @@ def __init__(
             channel_ids = self._es.electrodes["channel_name"][:].astype("str")
         else:
             channel_ids = [self._es.electrodes.table.id[x] for x in self._es.electrodes.data]
-
+        num_traces = self._es.data.shape[1] 
+        channel_ids = channel_ids[0:num_traces]
```
samuelgarcia, line 191: This seems a very too easy fix! how can we be sure that the data correspond to the biginning of channel map ?

alejoe91: We can't unfortunately..@luiztauffer are there other columns in the electrode tables that could tell the electrodes apart?

luiztauffer: I sent a better one now, that I think solves this

spikeinterface/extractors/nwbextractors.py
```diff
@@ -176,18 +176,17 @@ def __init__(
...
-        num_channels = len(self._es.electrodes.data)
 
         # Extractors channel groups must be integers, but Nwb electrodes group_name can be strings
         if 'group_name' in self._nwbfile.electrodes.colnames:
             unique_grp_names = list(np.unique(self._nwbfile.electrodes['group_name'][:]))
 
         # Fill channel properties dictionary from electrodes table
         if "channel_name" in self._nwbfile.electrodes.colnames:
-            channel_ids = self._es.electrodes["channel_name"][:].astype("str")
+            channel_ids = [self._es.electrodes["channel_name"][i] for i in self._es.electrodes.data]
```
samuelgarcia, line 186: What is "data" exactly here ?

spikeinterface/extractors/nwbextractors.py
samuelgarcia: Cool better otherwise we need a big big warning!
But thother solution look like a better way even I do not understand it.

spikeinterface/extractors/nwbextractors.py
luiztauffer: that's the only way I found to extract what seems to be the ids of the electrodes table associated with this electrical series

alejoe91: I think that is a good solution. @CodyCBakerPhD could you take a look? Is the `es.electrodes`  a link to the `electrodes` table region used by the `ElectricalSeries`?

CodyCBakerPhD: The `data` field of a DynamicTableRegion is supposed to be the subregion of the target table that is subindexed, i.e.,

```python
nwbfile.acquisition[es_key].electrodes.data[:]
```

h-mayorquin: I think you should keep the string. I remember that some round-trip tests failed in neuroconv because of this but I can't remember exactly which. At least try to run neuroconv tests with this branch.

PR #1500, opened 2023-04-05
Return `nn_unit_id` along with `nn_isolation`

TomBugnon:
Hi again,

When computing `nn_isolation`, return the nearest neighbor of each unit used to compute the isolation score
This is useful for sanity checks.
I don't think this should affect the tests.
Best


samuelgarcia (approved):

samuelgarcia:
thnaks

PR #1502, opened 2023-04-06
Adding a WaveformThresholder

yger:
PR redone after synced with main. Several points:
- I can not launch test anymore with pytest, is it normal? It used to work but now it complains about conftest.py
- I think that all PipelineNode that will work on waveforms (and thus transform waveforms) should inherit from WaveformNode instead of PipelineNode, ensuring they have at least one WaveformExtractor in their parents. This make sense, and allow these nodes to inherits from the attributes of their parents (ms_before, ms_after) that might be required by downstream node. This is done here for the SavGol filter, but I think this should also be done for all the denoisers

### DIFF
--- a/src/spikeinterface/sortingcomponents/tests/test_waveforms/test_waveform_thresholder.py
+++ b/src/spikeinterface/sortingcomponents/tests/test_waveforms/test_waveform_thresholder.py
@@ -25,11 +25,36 @@ def test_waveform_thresholder(mearec_recording, detected_peaks, chunk_executor_k
         recording=recording, ms_before=ms_before, ms_after=ms_after, return_output=True
     )
 
-    tresholded_waveforms = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='ptp', threshold=3, return_output=True)
-    pipeline_nodes = [extract_waveforms, tresholded_waveforms]
-
+    tresholded_waveforms_ptp = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='ptp', threshold=3, return_output=True)
+    tresholded_waveforms_mean = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='mean', threshold=3, return_output=True)
+    tresholded_waveforms_energy = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='energy', threshold=3, return_output=True)
+    tresholded_waveforms_peak  = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='v_peak', threshold=-5, return_output=True)
+    
+    pipeline_nodes = [extract_waveforms, tresholded_waveforms_ptp]
     # Extract projected waveforms and compare
     waveforms, tresholded_waveforms = run_peak_pipeline(
         recording, peaks, nodes=pipeline_nodes, job_kwargs=chunk_executor_kwargs
     )
-    assert np.all(tresholded_waveforms.ptp(axis=1) > 3)
+    assert np.all(tresholded_waveforms.ptp(axis=1) >= 3)
+
+    pipeline_nodes = [extract_waveforms, tresholded_waveforms_mean]
+    # Extract projected waveforms and compare
+    waveforms, tresholded_waveforms = run_peak_pipeline(
+        recording, peaks, nodes=pipeline_nodes, job_kwargs=chunk_executor_kwargs
+    )
+    assert np.all(tresholded_waveforms.mean(axis=1) >= 0)
+
+    pipeline_nodes = [extract_waveforms, tresholded_waveforms_energy]
+    # Extract projected waveforms and compare
+    waveforms, tresholded_waveforms = run_peak_pipeline(
+        recording, peaks, nodes=pipeline_nodes, job_kwargs=chunk_executor_kwargs
+    )
+    assert np.all(np.linalg.norm(tresholded_waveforms, axis=1) >= 3)
+
+    pipeline_nodes = [extract_waveforms, tresholded_waveforms_peak]
+    # Extract projected waveforms and compare
+    waveforms, tresholded_waveforms = run_peak_pipeline(
+        recording, peaks, nodes=pipeline_nodes, job_kwargs=chunk_executor_kwargs
+    )
+
+    assert np.all(tresholded_waveforms[:, extract_waveforms.nbefore, :] >= -5)
--- a/src/spikeinterface/sortingcomponents/tests/test_waveforms/test_waveform_thresholder.py
+++ b/src/spikeinterface/sortingcomponents/tests/test_waveforms/test_waveform_thresholder.py
@@ -1,5 +1,6 @@
 import pytest
 import numpy as np
+import operator
 
 
 from spikeinterface.sortingcomponents.waveforms.waveform_thresholder import WaveformThresholder
@@ -12,44 +13,64 @@ from spikeinterface.sortingcomponents.peak_pipeline import (
 )
 
 
-def test_waveform_thresholder(mearec_recording, detected_peaks, chunk_executor_kwargs):
-    recording = mearec_recording
-    peaks = detected_peaks
-
+@pytest.fixture(scope="module")
+def extract_waveforms(mearec_recording):
     # Parameters
     ms_before = 1.0
     ms_after = 1.0
 
     # Node initialization
-    extract_waveforms = ExtractDenseWaveforms(
-        recording=recording, ms_before=ms_before, ms_after=ms_after, return_output=True
+    return ExtractDenseWaveforms(
+        recording=mearec_recording, ms_before=ms_before, ms_after=ms_after, return_output=True
     )
 
+def test_waveform_thresholder_ptp(extract_waveforms, mearec_recording, detected_peaks, chunk_executor_kwargs):
+    recording = mearec_recording
+    peaks = detected_peaks
+
     tresholded_waveforms_ptp = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='ptp', threshold=3, return_output=True)
-    tresholded_waveforms_mean = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='mean', threshold=3, return_output=True)
-    tresholded_waveforms_energy = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='energy', threshold=3, return_output=True)
-    tresholded_waveforms_peak  = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='v_peak', threshold=-5, return_output=True)
-    
+
     pipeline_nodes = [extract_waveforms, tresholded_waveforms_ptp]
     # Extract projected waveforms and compare
     waveforms, tresholded_waveforms = run_peak_pipeline(
         recording, peaks, nodes=pipeline_nodes, job_kwargs=chunk_executor_kwargs
     )
+
     assert np.all(tresholded_waveforms.ptp(axis=1) >= 3)
+
+def test_waveform_thresholder_mean(extract_waveforms, mearec_recording, detected_peaks, chunk_executor_kwargs):
+    recording = mearec_recording
+    peaks = detected_peaks
+
+    tresholded_waveforms_mean = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='mean', threshold=3, return_output=True)
 
     pipeline_nodes = [extract_waveforms, tresholded_waveforms_mean]
     # Extract projected waveforms and compare
     waveforms, tresholded_waveforms = run_peak_pipeline(
         recording, peaks, nodes=pipeline_nodes, job_kwargs=chunk_executor_kwargs
     )
+
     assert np.all(tresholded_waveforms.mean(axis=1) >= 0)
+
+def test_waveform_thresholder_energy(extract_waveforms, mearec_recording, detected_peaks, chunk_executor_kwargs):
+    recording = mearec_recording
+    peaks = detected_peaks
+
+    tresholded_waveforms_energy = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='energy', threshold=3, return_output=True)
 
     pipeline_nodes = [extract_waveforms, tresholded_waveforms_energy]
     # Extract projected waveforms and compare
     waveforms, tresholded_waveforms = run_peak_pipeline(
         recording, peaks, nodes=pipeline_nodes, job_kwargs=chunk_executor_kwargs
     )
+
     assert np.all(np.linalg.norm(tresholded_waveforms, axis=1) >= 3)
+
+def test_waveform_thresholder_peak(extract_waveforms, mearec_recording, detected_peaks, chunk_executor_kwargs):
+    recording = mearec_recording
+    peaks = detected_peaks
+
+    tresholded_waveforms_peak = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='v_peak', threshold=-5, return_output=True)
 
     pipeline_nodes = [extract_waveforms, tresholded_waveforms_peak]
     # Extract projected waveforms and compare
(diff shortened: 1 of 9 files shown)


h-mayorquin:
Hi, @yger. 
Concerning the tests they are working for me. That is, running just pytest in the folder wher the repo is (after installing) collects all the tests without a sus.

What error are you getting?
If you used to install the package as editable you will need to re-install as the location changed with the new structure in main.

yger:
Thanks for the feedback. Errors are

ImportError while loading conftest '/home/pierre/github/spikeinterface/src/spikeinterface/sortingcomponents/tests/test_waveforms/conftest.py'.
conftest.py:6: in <module>
E   ModuleNotFoundError: No module named 'spikeinterface.sortingcomponents.peak_detection'

But maybe this needs to be reinstall, as you said

src/spikeinterface/sortingcomponents/waveforms/waveform_thresholder.py
```diff
@@ -0,0 +1,49 @@
...
+    def compute(self, traces, peaks, waveforms):
+        
+        if self.feature == 'ptp':
+            wf_data = waveforms.ptp(axis=1)
+        elif self.feature == 'mean':
+            wf_data = waveforms.mean(axis=1)
+        elif self.feature == 'energy':
+            wf_data = np.linalg.norm(waveforms, axis=1)
+        elif self.feature == 'v_peak':
+            wf_data = waveforms[idx][:, self.nbefore, :]
```
h-mayorquin, line 43: Where the `idx` is defined?

h-mayorquin (commented):
Looks good to me.

Can you add docstring and types to your parameters?
Can you add a test for every strategy that you use in the compute: mean, ptp, etc.

samuelgarcia:
Pierre you should base this on #1486

yger:
Thanks for the feedback. I mean all nodes that are exposing waveforms as outputs (this is what I meant) should inherit from WaveformExtractor and get the same attributes. So denoisers, my thresholder, ... Indeed, not sure for all nodes, but this is up to the user to decide when creating the node. Just that otherwise, some pipelines can not work (for example if I have dense waveforms, then filter, then want to compute something at the center of the denoised waveforms). I need to get their sizes..

h-mayorquin:
> Thanks for the feedback. I mean all nodes that are exposing waveforms as outputs (this is what I meant) should inherit from WaveformExtractor and get the same attributes. So denoisers, my thresholder, ... Indeed, not sure for all nodes, but this is up to the user to decide when creating the node. Just that otherwise, some pipelines can not work (for example if I have dense waveforms, then filter, then want to compute something at the center of the denoised waveforms). I need to get their sizes..

Makes sense to me.
I don't see any downside to it and it will allow to reference the `ms_before` and `ms_after` attribute more easily.

src/spikeinterface/sortingcomponents/tests/test_waveforms/test_waveform_thresholder.py
h-mayorquin, line 1: Thanks for writing the tests. I have separated them to a unit test format here below. Can you copy it?

```
import numpy as np
import pytest

def extract_waveforms(mearec_recording):
    # Parameters
    ms_before = 1.0
    ms_after = 1.0

    # Node initialization
    return ExtractDenseWaveforms(
        recording=mearec_recording, ms_before=ms_before, ms_after=ms_after, return_output=True
    )

def test_waveform_thresholder_ptp(extract_waveforms, mearec_recording, detected_peaks, chunk_executor_kwargs):
    recording = mearec_recording
    peaks = detected_peaks

    tresholded_waveforms_ptp = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='ptp', threshold=3, return_output=True)

    pipeline_nodes = [extract_waveforms, tresholded_waveforms_ptp]
    # Extract projected waveforms and compare
    waveforms, tresholded_waveforms = run_peak_pipeline(
        recording, peaks, nodes=pipeline_nodes, job_kwargs=chunk_executor_kwargs
    )

    assert np.all(tresholded_waveforms.ptp(axis=1) >= 3)

def test_waveform_thresholder_mean(extract_waveforms, mearec_recording, detected_peaks, chunk_executor_kwargs):
    recording = mearec_recording
    peaks = detected_peaks

    tresholded_waveforms_mean = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='mean', threshold=3, return_output=True)

    pipeline_nodes = [extract_waveforms, tresholded_waveforms_mean]
    # Extract projected waveforms and compare
    waveforms, tresholded_waveforms = run_peak_pipeline(
        recording, peaks, nodes=pipeline_nodes, job_kwargs=chunk_executor_kwargs
    )

    assert np.all(tresholded_waveforms.mean(axis=1) >= 0)

def test_waveform_thresholder_energy(extract_waveforms, mearec_recording, detected_peaks, chunk_executor_kwargs):
    recording = mearec_recording
    peaks = detected_peaks

    tresholded_waveforms_energy = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='energy', threshold=3, return_output=True)

    pipeline_nodes = [extract_waveforms, tresholded_waveforms_energy]
    # Extract projected waveforms and compare
    waveforms, tresholded_waveforms = run_peak_pipeline(
        recording, peaks, nodes=pipeline_nodes, job_kwargs=chunk_executor_kwargs
    )

    assert np.all(np.linalg.norm(tresholded_waveforms, axis=1) >= 3)

def test_waveform_thresholder_peak(extract_waveforms, mearec_recording, detected_peaks, chunk_executor_kwargs):
    recording = mearec_recording
    peaks = detected_peaks

    tresholded_waveforms_peak = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='v_peak', threshold=-5, return_output=True)

    pipeline_nodes = [extract_waveforms, tresholded_waveforms_peak]
    # Extract projected waveforms and compare
    waveforms, tresholded_waveforms = run_peak_pipeline(
        recording, peaks, nodes=pipeline_nodes, job_kwargs=chunk_executor_kwargs
    )

    assert np.all(tresholded_waveforms[:, extract_waveforms.nbefore, :] >= -5)
```

src/spikeinterface/sortingcomponents/waveforms/waveform_thresholder.py
```diff
@@ -0,0 +1,61 @@
...
+    that have a given feature below a given threshold
+
+    Parameters
+    ----------
+    feature: string in ['ptp', 'mean', 'energy', 'v_peak']
+        the feature that should be considered
+    threshold: float
+        the threshold below which channels are set to 0
+    """
+
```
h-mayorquin, line 27: Thanks for adding the typing. My suggestion for improving the docstring:


    """
    A node that" performs waveform thresholding based on a selected feature.

    This node allows you to perform adaptive masking by setting channels to 0
    that have a given feature below a certain threshold. The available features
    to consider are peak-to-peak amplitude ('ptp'), mean amplitude ('mean'),
    energy ('energy'), and voltage peak ('v_peak').

    Parameters
    ----------
    recording: BaseRecording
        The recording extractor object.
    return_output: bool, optional
        Whether to return output from this node (default True).
    parents: list of PipelineNodes, optional
        The parent nodes of this node (default None).
    feature: {'ptp', 'mean', 'energy', 'v_peak'}, optional
        The feature to be considered for thresholding (default 'ptp').
    threshold: float, optional
        The threshold value for the selected feature (default 2).
    """
    
Am I understanding and describing the v-peak method appropiately?

src/spikeinterface/sortingcomponents/waveforms/waveform_thresholder.py
```diff
@@ -0,0 +1,61 @@
...
+    def __init__(
+        self, recording: BaseRecording, return_output: bool = True, 
+        parents: Optional[List[PipelineNode]] = None,
+        feature = 'ptp',
+        threshold = 2
```
h-mayorquin, line 32:
```suggestion
        threshold: float = 2
```

src/spikeinterface/sortingcomponents/waveforms/waveform_thresholder.py
```diff
@@ -0,0 +1,61 @@
...
+    feature: string in ['ptp', 'mean', 'energy', 'v_peak']
+        the feature that should be considered
+    threshold: float
+        the threshold below which channels are set to 0
+    """
+
+    def __init__(
+        self, recording: BaseRecording, return_output: bool = True, 
+        parents: Optional[List[PipelineNode]] = None,
+        feature = 'ptp',
```
h-mayorquin, line 31:
```suggestion
        feature: Literal['ptp', 'mean', 'energy', 'v_peak'] = 'ptp'
```
You need to do from typing import Literal

src/spikeinterface/sortingcomponents/waveforms/waveform_thresholder.py
```diff
@@ -0,0 +1,61 @@
...
+        
+        if self.feature == 'ptp':
+            wf_data = waveforms.ptp(axis=1)
+        elif self.feature == 'mean':
+            wf_data = waveforms.mean(axis=1)
+        elif self.feature == 'energy':
+            wf_data = np.linalg.norm(waveforms, axis=1)
+        elif self.feature == 'v_peak':
+            wf_data = waveforms[:, self.nbefore, :]
+
```
h-mayorquin, line 56: This should throw an assertion in the else.

src/spikeinterface/sortingcomponents/waveforms/waveform_thresholder.py
```diff
@@ -0,0 +1,61 @@
...
+        mask = wf_data < self.threshold
+        mask = np.repeat(mask[:, np.newaxis, :], waveforms.shape[1], axis=1)
```
h-mayorquin, line 58: I think you can do this faster with broadcasting:

```
mask = np.broadcast_to(mask[:, np.newaxis, :], waveforms.shape)
```

Look at some profiling:

image

Maybe @samuelgarcia has something to say about readability though.

h-mayorquin (commented):
Thanks for taking into account my suggestion. Here is my final set of suggestions to make this better.

src/spikeinterface/sortingcomponents/waveforms/waveform_thresholder.py
yger: Thanks for the hint!

src/spikeinterface/sortingcomponents/waveforms/savgol_denoiser.py
```diff
@@ -7,20 +7,42 @@
...
         window_length_ms : float = 0.25
     ):
-        super().__init__(recording, return_output=return_output, parents=parents)
-        self.order = order
-        if self.parents is None or not (len(self.parents) == 1 and isinstance(self.parents[0], WaveformExtractorNode)):
-            exception_string = f"SavGolDenoiser should have a single {WaveformExtractorNode.__name__} in its parents"
+        try:
+            waveform_extractor = next(parent for parent in parents if isinstance(parent, WaveformExtractorNode))
+        except (StopIteration, TypeError):
+            exception_string = f"Model should have a {WaveformExtractorNode.__name__} in its parents"
```
h-mayorquin, line 38:
```suggestion
            exception_string = f"{self.__name__} should have a {WaveformExtractorNode.__name__} in its parents"
```
?

src/spikeinterface/sortingcomponents/waveforms/waveform_thresholder.py
```diff
@@ -0,0 +1,77 @@
...
+        self, recording: BaseRecording, return_output: bool = True, 
+        parents: Optional[List[PipelineNode]] = None,
+        feature: Literal['ptp', 'mean', 'energy', 'v_peak'] = 'ptp',
+        threshold: float = 2,
+        operator: callable = operator.le
+    ):
+        try:
+            waveform_extractor = next(parent for parent in parents if isinstance(parent, WaveformExtractorNode))
+        except (StopIteration, TypeError):
+            exception_string = f"Model should have a {WaveformExtractorNode.__name__} in its parents"
```
h-mayorquin, line 51:
```suggestion
            exception_string = f"{self.__name__} should have a {WaveformExtractorNode.__name__} in its parents"
```

h-mayorquin (approved):
Thanks for bearing with me. This looks good to me. I prefered the previous explicit `<` operator but here you are better positioned to judged as a user what is the right level of abstraction that makes sense.

yger:
I realized that depending on the fact that you are looking for positive or negative peaks, you may want to change the comparison to select the waveforms that has to be thresholded... I could infer sign automatically, but this is hard to explain. So I expose the option, and only advanced users will know how to use it with the operator

yger:
@samuelgarcia @alejoe91 @h-mayorquin I just added the update of the _kwargs dict with the params. Is it something that is mandatory. What is exactly the purpose of such a dict? To be reloaded from file? For parallelism?

src/spikeinterface/sortingcomponents/tests/test_waveforms/test_waveform_thresholder.py
```diff
@@ -0,0 +1,96 @@
...
+
+    tresholded_waveforms_mean = WaveformThresholder(recording=recording, parents=[extract_waveforms], feature='mean', threshold=3, return_output=True)
+
+    pipeline_nodes = [extract_waveforms, tresholded_waveforms_mean]
+    # Extract projected waveforms and compare
+    waveforms, tresholded_waveforms = run_peak_pipeline(
+        recording, peaks, nodes=pipeline_nodes, job_kwargs=chunk_executor_kwargs
+    )
+
+    assert np.all(tresholded_waveforms.mean(axis=1) >= 0)
```
alejoe91, line 53:
```suggestion
    assert np.all(tresholded_waveforms.mean(axis=1) >= 3)
```
?

src/spikeinterface/sortingcomponents/waveforms/savgol_denoiser.py
```diff
@@ -7,22 +7,47 @@
 from spikeinterface.sortingcomponents.peak_pipeline import PipelineNode, WaveformExtractorNode
 
 
-class SavGolDenoiser(PipelineNode):
+class SavGolDenoiser(WaveformExtractorNode):
```
alejoe91, line 10: Should this be propagated to the tPCA_denoiser and NNDenoiser?

yger: I think yes, all nodes that are expected to returns waveforms (and thus denoiser) should behave as WaveformExtractorNodes to inherit params from their parents (nbefore/nafter). This is why the modified init here and the search for a wavefor_extractor in the parents

yger: You want me to do that in the branch?

alejoe91: Since you already added it to the `SavGolDenoiser` I think it makes sense to propagate it to all denoisers ;) yeah let's do it in this branch

yger: I did it for the neural network, but for the TemporalPCA, I think the person that wrote the code should do it. Just because it seems to me that there is some useless complexity that should be cleaned if we adopt such a scheme.

yger: Actually, I also did it for TemporalPCA

yger: Minor caveat... I deactivate one of the test of the TemporalPCA node for now, because I saw that there is, in the detect_peak refactoring, a function called find_parent() that will do the job I'm currently doing. We should reactivate the tests while using this function

h-mayorquin:
> @samuelgarcia @alejoe91 @h-mayorquin I just added the update of the _kwargs dict with the params. Is it something that is mandatory. What is exactly the purpose of such a dict? To be reloaded from file? For parallelism?

For the recorders and sorting objects they are necesesary for serialization. But the objects in this module should not need this. Were you getting errors without it?

src/spikeinterface/sortingcomponents/waveforms/savgol_denoiser.py
h-mayorquin: You can leave the tests deactivated if you want. Now that we have agreed on using a common base class for waveform nodes I will refactor and add a general test.

yger:
No, I just wanted to be sure I was not missing a point

yger:
@alejoe91 @h-mayorquin @samuelgarcia I had to modify the function find_parents_type, because for Nodes that will inherit from WaveformsNode, we want to check *before* calling the init if there is a waveformNode in their parents. So we can so launch this function on self.parent, but we should rather do it on the list of parents

yger:
This should also solves the error in #1510

h-mayorquin:
@yger 
> No, I just wanted to be sure I was not missing a point
Nop. You should removed those. You can access attribute directly from any object. The `dict` was done before general serialization was enabled.

> @alejoe91 @h-mayorquin @samuelgarcia I had to modify the function find_parents_type, because for Nodes that will inherit from WaveformsNode, we want to check _before_ calling the init if there is a waveformNode in their parents. So we can so launch this function on self.parent, but we should rather do it on the list of parents

I don't understand this. 
The check now is done when you instantiate, when do you want to do the check and why?

yger:
A waveformsNode object should also expose ms_before and ms_after. Thus, when you want to create one, you must first check that you have a WaveformsNode in the parents (calling find_parent_type), then, once this has been checked, call the init() method of the WaveformsNodes with parent.ms_before and parent.ms_after

PR #1503, opened 2023-04-06
Add amplitude_scaling implementation

alejoe91:
Implement amplitude_scalings extension to compute scaling of each spike with respect to the template.

Currently using the slope of the `scipy.stats.linregress` function.

A possible extension would be to instead of using spike by spike regression, to gather all spikes in a nighborhood and fit colliding spikes.


samuelgarcia:
Could you push here a figure that compare amplitude distribution and scaling distribution

src/spikeinterface/postprocessing/amplitude_scalings.py
```diff
@@ -0,0 +1,304 @@
...
+    waveform_extractor: WaveformExtractor
+        The waveform extractor object
+    sparsity: ChannelSparsity
+        If waveforms are not sparse, sparsity is required if the number of channels is greater than
+        `max_dense_channels`. If the waveform extractor is sparse, its sparsity is automatically used.
+        By default None
+    max_dense_channels: int, optional
+        Maximum number of channels to allow running without sparsity. To compute amplitude scaling using
+        dense waveforms, set this to None, sparsity to None, and pass dense waveforms as input.
+        By default 16
```
bendichter, line 164:
```suggestion
    max_dense_channels: int, default: 16
        Maximum number of channels to allow running without sparsity. To compute amplitude scaling using
        dense waveforms, set this to None, sparsity to None, and pass dense waveforms as input.
```

this is strict Numpy Docstring format

alejoe91: I see, note that VS code automatically gives the `optional` notation when it generates docstrings

h-mayorquin: That's the reason that this usually escpes me as well.

alejoe91:
@samuelgarcia removed collisions for now. Then it should be ready for a final review

samuelgarcia (approved):

PR #1504, opened 2023-04-06
Version mismatch: `print` --> `warning`

alejoe91:
At least it can be filtered :)


samuelgarcia:
I am OK with the warning, the print is annoying.
But we have to have in mind that this problem is per file.

alejoe91:
Of course!

samuelgarcia (approved):

PR #1506, opened 2023-04-06
[Do not merge] Test many calls to memmap

h-mayorquin:
Test memory error on Windows if opening many memmap files at ones (seems to happen for SpikeGLX)


h-mayorquin:
Let's close this. I lost my zeal for fixing these kind of problems and they have not shown up in a while.

PR #1510, opened 2023-04-06
Choices of Feature for the monopolar triangulation

yger:
Similarly to what have been done with CoM, we should let the user decide which feature should be considered for monopolar triangulation. At a first glimpse (see figure below), it looks like using the norm as a feature might be better than ptp. I'll do more tests on that.

image

### DIFF
--- a/src/spikeinterface/postprocessing/unit_localization.py
+++ b/src/spikeinterface/postprocessing/unit_localization.py
@@ -130,15 +130,15 @@ def compute_unit_locations(waveform_extractor,
     return unit_locations
 
 
-def make_initial_guess_and_bounds(wf_ptp, local_contact_locations, max_distance_um, initial_z=20):
+def make_initial_guess_and_bounds(wf_data, local_contact_locations, max_distance_um, initial_z=20):
 
     # constant for initial guess and bounds
-    ind_max = np.argmax(wf_ptp)
-    max_ptp = wf_ptp[ind_max]
+    ind_max = np.argmax(wf_data)
+    max_ptp = wf_data[ind_max]
     max_alpha = max_ptp * max_distance_um
 
     # initial guess is the center of mass
-    com = np.sum(wf_ptp[:, np.newaxis] * local_contact_locations, axis=0) / np.sum(wf_ptp)
+    com = np.sum(wf_data[:, np.newaxis] * local_contact_locations, axis=0) / np.sum(wf_data)
     x0 = np.zeros(4, dtype='float32')
     x0[:2] = com
     x0[2] = initial_z
@@ -152,11 +152,11 @@ def make_initial_guess_and_bounds(wf_ptp, local_contact_locations, max_distance_
     return x0, bounds
 
 
-def solve_monopolar_triangulation(wf_ptp, local_contact_locations, max_distance_um, optimizer):
-    x0, bounds = make_initial_guess_and_bounds(wf_ptp, local_contact_locations, max_distance_um)
+def solve_monopolar_triangulation(wf_data, local_contact_locations, max_distance_um, optimizer):
+    x0, bounds = make_initial_guess_and_bounds(wf_data, local_contact_locations, max_distance_um)
 
     if optimizer == 'least_square':
-        args = (wf_ptp, local_contact_locations)
+        args = (wf_data, local_contact_locations)
         try:
             output = scipy.optimize.least_squares(estimate_distance_error, x0=x0, bounds=bounds, args=args)
             return tuple(output['x'])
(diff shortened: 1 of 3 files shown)


samuelgarcia:
bien joué!
on est des champions.

src/spikeinterface/postprocessing/unit_localization.py
```diff
@@ -256,27 +265,42 @@ def compute_monopolar_triangulation(waveform_extractor, optimizer='minimize_with
...
-        wf = templates[i, :, :]
-        wf_ptp = wf[:, chan_inds].ptp(axis=0)
-        unit_location[i] = solve_monopolar_triangulation(wf_ptp, local_contact_locations, max_distance_um, optimizer)
+        wf = templates[i, :, :][:, chan_inds]
+        if feature == 'ptp':
+            wf_data = wf.ptp(axis=0)
+        elif feature == 'energy':
+            wf_data = np.linalg.norm(wf, axis=0)
+        #elif feature == 'v_peak':
+        #    wf_data = wf[nbefore]
```
alejoe91, line 293: @yger since you're at it can you add this too?

Biophysically speaking, this is the best option IMO :)

yger: Yes and no, because not sure about this one. At least, I would prevent the use of enforce_decrease, that has the assumption that all values should be decreasing as function of the distance. Or I should consider abs values? Having signed values might disturb everything, but willing to have a try

yger:
v_peak is now also available (I used abs(v_peak)). Note that other metrics could be implemented, looks like both energy and v_peak are reducing the variance 

image

src/spikeinterface/postprocessing/unit_localization.py
```diff
@@ -244,6 +249,10 @@ def compute_monopolar_triangulation(waveform_extractor, optimizer='minimize_with
         to make bounddary in x, y, z and also for alpha
     return_alpha: bool default False
         Return or not the alpha value
+    enforce_decrease : bool (default False)
+        Enforce spatial decreasingness for PTP vectors
+    feature: string in ['ptp', 'energy', 'v_peak']
+        Feature to consider for monopolar approximation (default ptp)
```
alejoe91, line 255: @yger can you explain in one line the different options here?

src/spikeinterface/sortingcomponents/tests/test_peak_localization.py
```diff
@@ -45,6 +45,12 @@ def test_localize_peaks():
                                     enforce_decrease=True, **job_kwargs)
     assert peaks.size == peak_locations.shape[0]
     list_locations.append(('minimize_with_log_penality', peak_locations))
+
+    peak_locations = localize_peaks(recording, peaks, method='monopolar_triangulation', 
+                                    optimizer='minimize_with_log_penality',
+                                    enforce_decrease=True, feature='energy', **job_kwargs)
+    assert peaks.size == peak_locations.shape[0]
+    list_locations.append(('minimize_with_log_penality_energy', peak_locations))
```
alejoe91, line 53: Can you add `v_peak` fo completeness?

alejoe91 (changes requested):
Thanks @yger 

Looks great! Could you make the suggested additions before merging?

src/spikeinterface/sortingcomponents/tests/test_peak_localization.py
yger: done

src/spikeinterface/postprocessing/unit_localization.py
yger: done

yger:
@alejoe91 you must approve your review, and this one is good to me to be merged :-)

src/spikeinterface/postprocessing/unit_localization.py
```diff
@@ -248,8 +249,13 @@ def compute_monopolar_triangulation(waveform_extractor, optimizer='minimize_with
...
         Return or not the alpha value
-    enforce_decrease : False
-        To enforce spatial decreasingness for PTP vectors.
+    enforce_decrease : bool (default False)
+        Enforce spatial decreasingness for PTP vectors
+    feature: string in ['ptp', 'energy', 'v_peak']
+        The available features to consider for estimating the position via
+        monopolar triangulation are peak-to-peak amplitudes (ptp, default), 
+        energy ('energy', as L2 norm) or voltages at the center of the waveform
+        (v_peak)
```
alejoe91, line 258:
```suggestion
        monopolar triangulation are peak-to-peak amplitudes ('ptp', default), 
        energy ('energy', as L2 norm) or voltages at the center of the waveform
        ('v_peak')
```

alejoe91 (approved):

PR #1513, opened 2023-04-10
Refactor nwb extractor I

h-mayorquin:
I added tests for the merged #1492 and did a large refactor of the code base. I separated the code logically, expanded on unecessary abreviations (like `es` for `electrical_series`) and re-formated the whole thing.

The part where the properties are added is still a mess. That will need another refactor and possible a couple of tests for that area specficially. Meanwhile, this should improve the readability of the function.


src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -455,22 +489,23 @@ def __init__(self, nwbfile, sampling_frequency, timestamps):
...
-        times = self._nwbfile.units['spike_times'][list(self._nwbfile.units.id[:]).index(unit_id)][:]
+        times = self._nwbfile.units["spike_times"][list(self._nwbfile.units.id[:]).index(unit_id)][:]
 
         if self._timestamps is not None:
-            frames = np.searchsorted(times, self.timestamps).astype('int64')
+            frames = np.searchsorted(times, self.timestamps).astype("int64")
         else:
-            frames = np.round(times * self._sampling_frequency).astype('int64')
+            frames = np.round(times * self._sampling_frequency).astype("int64")
         return frames[(frames > start_frame) & (frames < end_frame)]
```
alejoe91, line 509:
```suggestion
        return frames[(frames >= start_frame) & (frames < end_frame)]
```

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -116,57 +116,43 @@ def __init__(
...
-            if self._es.timestamps is not None:
-                timestamps = self._es.timestamps
-                t_start = self._es.timestamps[0]
+        if hasattr(electrical_series, "timestamps"):
+            if electrical_series.timestamps is not None:
+                timestamps = electrical_series.timestamps
+                t_start = electrical_series.timestamps[0]
 
-        # if rate is unknown, estimate from timestamps
+        # if rate is unknown, estimate from timestamps TODO: Is this possible?
```
CodyCBakerPhD, line 155: Always possible with neuroconv util or nwb inspector util, let me know if you have trouble finding it

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -176,92 +162,100 @@ def __init__(
...
 
+
         # Set offsets
-        offset = self._es.offset if hasattr(self._es, "offset") else 0
-        if offset == 0 and "offset" in self._nwbfile.electrodes:
-            electrode_table_index = self._es.electrodes.data[:]
-            offset = self._nwbfile.electrodes["offset"].data[electrode_table_index]
+        offset = electrical_series.offset if hasattr(electrical_series, "offset") else 0
+        if offset == 0 and "offset" in electrodes_table:
+            offset = electrodes_table["offset"].data[electrodes_indices]
```
CodyCBakerPhD, line 206: Just planning ahead, how much longer do we want to support the round-about way of storing offsets?

CodyCBakerPhD: Hmm.. I guess there might always be some historical files on DANDI (not sure off the top of my head) if SI health check on DANDI files is the concern... I guess always keep it then?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -176,92 +162,100 @@ def __init__(
...
+        for es_ind, (channel_id, electrode_table_index) in enumerate(zip(channel_ids, electrodes_indices)):
+            if "rel_x" in electrodes_table:
+                properties["location"][es_ind, 0] = electrodes_table["rel_x"][electrode_table_index]
+                if "rel_y" in electrodes_table:
+                    properties["location"][es_ind, 1] = electrodes_table["rel_y"][electrode_table_index]
+                if "rel_z" in electrodes_table:
+                    properties["location"][es_ind, 2] = electrodes_table["rel_z"][electrode_table_index]
+
+        # Extract all the other properties
+        for es_ind, (channel_id, electrode_table_index) in enumerate(zip(channel_ids, electrodes_indices)):
```
CodyCBakerPhD, line 231: Since style and readability seems to be the main purpose of the PR, I'd suggest also renaming `es_ind` to `electrical_series_index`

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -176,92 +162,100 @@ def __init__(
...
         else:
             times_kwargs = dict(sampling_frequency=sampling_frequency, t_start=t_start)
 
-        num_frames = int(self._es.data.shape[0])
-
         # Extractors channel groups must be integers, but Nwb electrodes group_name can be strings
-        if "group_name" in self._nwbfile.electrodes.colnames:
-            unique_grp_names = list(np.unique(self._nwbfile.electrodes["group_name"][:]))
+        if "group_name" in electrodes_table.colnames:
+            unique_grp_names = list(np.unique(electrodes_table["group_name"][:]))
```
CodyCBakerPhD, line 171: Since style and readability seems to be the focus of the PR, I'd suggest expanding `unique_grp_names` to `unique_electrode_group_names`

CodyCBakerPhD: Hm... well, maybe `unique_group_names_from_electrodes_table`, which might not match the actual number of unique names of `ElectrodeGroup` objects

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -176,92 +162,100 @@ def __init__(
...
 
         # Fill channel properties dictionary from electrodes table
-        if "channel_name" in self._nwbfile.electrodes.colnames:
-            channel_ids = [self._es.electrodes["channel_name"][i] for i in self._es.electrodes.data]
+        if "channel_name" in electrodes_table.colnames:
+            channel_ids = [electrical_series.electrodes["channel_name"][i] for i in electrodes_indices]
```
CodyCBakerPhD, line 175: I've also been getting more in the habit of using better list compression names, rather than mathematical variables

```suggestion
            channel_ids = [electrical_series.electrodes["channel_name"][electrodes_index] for electrodes_index in electrodes_indices]
```

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -289,12 +287,72 @@ def __init__(
             "stream_cache_path": stream_cache_path,
         }
 
+    def read_nwb_file(self, file_path: str, stream_mode: Literal[None, "ffspec", "ros3"] = None, stream_cache_path: Optional[str] = None) -> NWBFile:
```
CodyCBakerPhD, line 290: Such a minor point I almost feel bad bringing it up... but we usually combine `nwb_file` into `nwbfile` in code (especially function and argument names)

Could this be `read_nwbfile` instead? 😅

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -289,12 +287,72 @@ def __init__(
...
+        """
+        Read an NWB file and return the NWBFile object.
+
+        Parameters
+        ----------
+        file_path : str
+            The path to the NWB file.
+        stream_mode : str, optional
+            The streaming mode to use. Default is None.
```
CodyCBakerPhD, line 299: This, as well as the matching part of the class definition, aren't true anymore if we start using `Literal`

```suggestion
        stream_mode : "fsspec" or "ros3", optional
            The streaming mode to use. Default assumes the file is on the local disk.
```

src/spikeinterface/extractors/tests/test_nwbextractors.py
```diff
@@ -4,9 +4,13 @@
 from tempfile import mkdtemp
 from datetime import datetime
 
+import pytest
 import numpy as np
 from pynwb import NWBHDF5IO, NWBFile
 from pynwb.ecephys import ElectricalSeries
+from pynwb.testing.mock.file import mock_NWBFile
+from pynwb.testing.mock.device import mock_Device
+from pynwb.testing.mock.ecephys import mock_ElectricalSeries, mock_ElectrodeGroup
```
CodyCBakerPhD, line 13: As much as I like these testing classes, they are technically not available in older PyNWB versions - as such, you may want to specify a lower bound to PyNWB in your [testing requirements](https://github.com/catalystneuro/spikeinterface/blob/refactor_nwb_extractor/pyproject.toml#L110) to avoid the situation where the SI minimal requirements say they can support reading in an environment using an older version, but the SI testing suite itself cannot run in that environment

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: This comment is confusing because of the previous logica if structure. I will remove it. Meanwhile, one question: The TimeSeries in NWB require either the rate or the timestamps, right? In other words, if they don't have the former, they NEED to have the latter, correct?

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Good point.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Yes, this is better.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Yes, I think this was the concern. Previous files created by us when we used to write the offset there.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Yes, please do not feel bad, I appreciate that. Given that we are just now introducing this function. This is a good point.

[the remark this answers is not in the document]
h-mayorquin: I think the second makes more sense if we call this option a `stream_mode`. I am changing this. Thanks.

src/spikeinterface/extractors/tests/test_nwbextractors.py
h-mayorquin: Yep, this makes sense. Added that

src/spikeinterface/extractors/nwbextractors.py
CodyCBakerPhD: > This comment is confusing because of the previous logica if structure.

Oh, I just meant for the TODO thing that was added here.

But yeah, `TimeSeries` and all their descendants must have either `starting_tiime` + `rate` (two floats) or just `timestamps` (vector of floats, length matching first axes of `data`). The two are mutually exclusive; I don't know that I've _tried_ to specify both however, maybe it's possible?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -37,24 +37,116 @@ def check_fsspec_install():
...
             raise ValueError("No acquisitions found in the .nwb file.")
-        es = es_list[0]
-    return es
+        electrical_series = electrical_series_list[0]
+
+    return electrical_series
+
+
+def read_nwbfile(
+    file_path: str, stream_mode: Literal["ffspec", "ros3"] = None, stream_cache_path: Optional[str] = None
```
CodyCBakerPhD, line 92:
```suggestion
    file_path: str, stream_mode: Optional[Literal["ffspec", "ros3"]] = None, stream_cache_path: Optional[str] = None
```

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -116,57 +208,41 @@ def __init__(
         electrical_series_name: str = None,
         load_time_vector: bool = False,
         samples_for_rate_estimation: int = 100000,
-        stream_mode: str = None,
+        stream_mode: Optional[str] = None,
```
CodyCBakerPhD, line 211:
```suggestion
        stream_mode: Optional[Literal["fsspec", "ros3"]] = None,
```

CodyCBakerPhD:
Two last suggestions on annotation typing for consistency, otherwise LGTM!

The code is so much easier to read now, great job ❤

CodyCBakerPhD:
Yep, testing on NC as well: https://github.com/catalystneuro/neuroconv/pull/425

h-mayorquin:
Let's fix the NC test first then. Good catch.

CodyCBakerPhD:
Yeah, looks like some `h5py.Dataset` indexing issue on

```
/opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/site-packages/spikeinterface/extractors/nwbextractors.py:497: in __init__
    units_ids = list(self._nwbfile.units.id[:])
```

Full log: https://github.com/catalystneuro/neuroconv/actions/runs/4799443079/jobs/8539080772?pr=425

h-mayorquin:
@CodyCBakerPhD 
OK, I took a stab at this today. The problem is caused by some magic about the io context manager `NWBHDF5IO`. That is, the references become invalid when the io is called in a different function causing problems. 

As we don't have a lot of test for the SortingExtractor I have decided to revert the changes to it. My goal was to refactor the RecordingExtractor and we don't have a lot of tests for the SortingExtractor. When I have time to proper refactor the latter and add some tests I will then change this behavior.

In summary:
I kept all the modifications for the RecordingExtractor.
Undid a small refactor that I did for reading the nwbfile in the SortingExtractor.
Will re-do in the future with a PR specifically aimed at refactoring and addintg tests for the SortingExtractor.

alejoe91:
@h-mayorquin @CodyCBakerPhD so ready to merge?

h-mayorquin:
This is ready to merge on my side.

CodyCBakerPhD:
Roger, roger

alejoe91 (approved):

h-mayorquin:
@CodyCBakerPhD 
Thanks a lot for your time, the error in neuroconv would have gone unoticed and it would have been way more painful to revert and fix on the spot!

PR #1517, opened 2023-04-11
Add silhouette score quality metric

zm711:
First Tom's PR #1499 will need to be merged before this can be done, so I can update to get rid of the merge conflicts. So I'll leave this as a draft until that's done.

Working on the docs I saw that silhouette score had not been implemented. So I implemented it two different ways and was hoping for feedback (if you guys want it implemented in SI--otherwise I can delete it). First is the simplified silhouette score proposal here: https://ieeexplore.ieee.org/document/1410321, which uses centroids rather than pairwise point calculations
Pros:
It is computationally and space-wise o(kn) rather than o(kn2). 
Cons:
Literature is less developed (and was proposed for genomics purposes) and technically doesn't use all info available

And the original silhouette score as proposed by Rousseeuw.
Pros:
It uses all information provided by doing point by point calculations. ML literature uses it.
Cons:
Computationally and space it is o(kn2) which can quickly lead to memory overflow for large datasets.

I figure only one would be implemented, so I'll just delete the one not wanted for the qualitymetrics.

### DIFF
--- a/doc/modules/qualitymetrics/silhouette_score.rst
+++ b/doc/modules/qualitymetrics/silhouette_score.rst
@@ -1,5 +1,5 @@
-Silhouette score (silhouette_score, simplified_silhouette_score)
-============================================
+Silhouette score (:code:`silhouette`, :code:`silhouette_full`)
+==============================================================
 
 Calculation
 -----------
@@ -8,7 +8,7 @@ Gives the ratio between the cohesiveness of a cluster and its separation from ot
 Values for silhouette score range from -1 to 1. 
 
 For the full method as proposed by [Rouseeuw]_, the pairwise distances between each point 
-and every other point :math:`a(i)` in a cluster :math:`C_i` and then iterating through 
+and every other point :math:`a(i)` in a cluster :math:`C_i` are calculated and then iterating through 
 every other cluster's distances between the points in :math:`C_i` and the points in :math:`C_j` 
 are calculated. The cluster with the minimal mean distance is taken to be :math:`b(i)`. The
 average value of :math:`s(i)` is taken to give the final silhouette score for a cluster with
@@ -46,7 +46,7 @@ SpikeInterface provides access to both implementations of silhouette score.
 
 To reduce complexity the default implementation in SpikeInterface is to use the simplified silhouette score.
 This can be changes by switching the silhouette method to either 'full' (the Rousseeuw implementation) or
-('simplified', 'full') for both methods.
+('simplified', 'full') for both methods when entering the qm_params parameter.
 
 
 Literature
--- a/doc/modules/qualitymetrics/silhouette_score.rst
+++ b/doc/modules/qualitymetrics/silhouette_score.rst
@@ -1,11 +1,21 @@
-Silhouette score (not currently implemented)
+Silhouette score
 ============================================
 
 Calculation
 -----------
 
 Gives the ratio between the cohesiveness of a cluster and its separation from other clusters.
-Values for silhouette score range from -1 to 1.
+Values for silhouette score range from -1 to 1. First the pairwise distances between each
+point and every other point :math:`a(i)` in a cluster :math:`C_i` and then iterating through 
+every other cluster distances between the points in :math:`C_i` and the points in :math:`C_j` 
+are calculated. The cluster with the minimal mean distance is taken to be :math:`b(i)`. The
+average value of :math:`s(i)` is taken to give the final silhouette score for a cluster.
+
+.. math::
+    a(i) = \frac{1}{C_i-1} \sum_{j \in C_i, j \neq i} d(i,j)
+    b(i) = \min {J \neq I} \frac{1}{C_j} \sum_{j \in C_j} d(i, j)
+    s(i) = \frac{a(i)-b(i)}{\max(a(i), b(i))}
+    Silhouette Score = \frac{1}{N} \sum^{N} s(i)
 
 Expectation and use
 -------------------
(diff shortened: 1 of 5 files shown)


h-mayorquin:
@zm711 This is a test that is appearing as of today with the CI unrelated to your problems. 

Re-run the job and it should work:
image


If this persists more than two days I will address it from the CI level but most likely the actions will be fixed and this will go away.

zm711:
I like that. I tried using the full for my own data (admittedly on a computer that is RAM limited) and it failed so I had to switch to simplified for my data. So simplified as default with option to do the full makes sense to me.

src/spikeinterface/qualitymetrics/pca_metrics.py
```diff
@@ -804,5 +898,19 @@ def pca_metrics_one_unit(pcs_flat, labels, metric_names, unit_id,
         except:
             nn_noise_overlap = np.nan
         pc_metrics['nn_noise_overlap'] = nn_noise_overlap
+        
+    if 'silhouette_score' in metrics_names:
```
alejoe91, line 902:
```suggestion
    if 'silhouette_score' in metric_names:
```

src/spikeinterface/qualitymetrics/pca_metrics.py
```diff
@@ -804,5 +898,19 @@ def pca_metrics_one_unit(pcs_flat, labels, metric_names, unit_id,
...
+        try:
+            unit_silhouette_score = silhouette_score(pcs_flat, labels, unit_id)
+        except:
+            unit_silhouette_score = np.nan
+        pc_metrics['silhouette_score'] = unit_silhouette_score
+        
+    if 'simplified_silhouette_score' in metrics_name:
```
alejoe91, line 909:
```suggestion
    if 'simplified_silhouette_score' in metric_name:
```

alejoe91 (changes requested):
couple of small typos!

doc/modules/qualitymetrics/silhouette_score.rst
```diff
@@ -1,20 +1,56 @@
-Silhouette score (not currently implemented)
+Silhouette score (silhouette_score, simplified_silhouette_score)
 ============================================
```
alejoe91, line 2:
```suggestion
Silhouette score (silhouette_score, simplified_silhouette_score)
================================================================
```

src/spikeinterface/qualitymetrics/pca_metrics.py
```diff
@@ -679,6 +682,101 @@ def nearest_neighbors_noise_overlap(waveform_extractor: WaveformExtractor, this_
                                                   n_neighbors)
 
         return nn_noise_overlap
+def simplified_silhouette_score(all_pcs, all_labels, this_unit_id):
```
alejoe91, line 685:
```suggestion


def simplified_silhouette_score(all_pcs, all_labels, this_unit_id):
```

src/spikeinterface/qualitymetrics/pca_metrics.py
```diff
@@ -679,6 +682,101 @@ def nearest_neighbors_noise_overlap(waveform_extractor: WaveformExtractor, this_
...
+                distances_for_minimum_cluster = distances_for_other_cluster
+
+    sil_distances = (
+        distances_for_minimum_cluster - distances_for_this_unit
+    ) / np.maximum(distances_for_minimum_cluster , distances_for_this_unit)
+
+    unit_silhouette_score = np.mean(sil_distances)
+    return unit_silhouette_score
+
+def silhouette_score(all_pcs, all_labels, this_unit_id):
```
alejoe91, line 734:
```suggestion

def silhouette_score(all_pcs, all_labels, this_unit_id):
```

src/spikeinterface/qualitymetrics/pca_metrics.py
```diff
@@ -53,7 +53,10 @@
         n_components=10,
         radius_um=100,
         peak_sign='neg'
-    )
+    ),
+    silhouette=dict(
+        method='simplified'
```
alejoe91, line 58:
```suggestion
        method=('simplified',)
```
Since the code can take a tuple as well, I would keep the tuple by default.

The `if "simplified" in qm_params["silhouette"]["method"` also makes more sense with tuple (it should work with strings too, but in that case a `==` would be easier tot read)

src/spikeinterface/qualitymetrics/tests/test_metrics_functions.py
```diff
@@ -102,6 +102,27 @@ def test_nearest_neighbors_metrics():
...
+
+def test_simplified_silhouette_score_metrics():
+    all_pcs1, all_labels1 = create_ground_truth_pc_distributions([1, -1], [1000, 1000])
+    all_pcs2, all_labels2 = create_ground_truth_pc_distributions([1, -2],
+                                                                 [1000, 1000])  # increase distance between clusters
+
+    sim_sil_score1 = simplified_silhouette_score(all_pcs1, all_labels1, 0)
+    sim_sil_score2 = simplified_silhouette_score(all_pcs2, all_labels2, 0)
+
+    assert sim_sil_score1 < sim_sil_score2
```
alejoe91, line 125: nice!! :)

src/spikeinterface/qualitymetrics/pca_metrics.py
alejoe91: no I think that's ok, but let's stick to the `tuple` in the params, because a tuple is actually accepted.

zm711: Already made `tuple` 👍. This was in addition. But as long as you're okay with that then I'll leave the `if` as is.

alejoe91 (approved):

PR #1518, opened 2023-04-11
Add original_cluster_id to PhySortingExtractor

DradeAW:


DradeAW:
> In our testing files that we have the cluster_id is identical to the indexes you are using. Are they different in your files?

After `sorting = PhySortingExtractor("path/to/phy")`, for some of them the unit ids are not the same as the cluster_id in phy.
This is why I wanted to track the original cluster_id.

@alejoe91 You're right!
I'll make the fix right away :)

h-mayorquin:
> After sorting = PhySortingExtractor("path/to/phy"), for some of them the unit ids are not the same as the cluster_id in phy.
This is why I wanted to track the original cluster_id.

@alejoe91 @DradeAW  I don't understand why would they be different if we have the following line which sets the `units_ids` to the `cluster_id` from the files?

https://github.com/SpikeInterface/spikeinterface/blob/48aa2285a719461586d38e0210c5d2c0e06cee12/src/spikeinterface/extractors/phykilosortextractors.py#L136

DradeAW:
> I don't understand why would they be different if we have the following line which sets the units_ids to the cluster_id from the files?

Good question, let me investigate :)

DradeAW:
Ah it's simply because I have a file `cluster_si_unit_ids.tsv`
But I'd still like to know the original cluster_id in phy.

alejoe91:
Because if you save with SI we add a custom column (`si_unit_id`) to make sure that when we load it back the same SI unit IDs are preserved. See the other part of the if-else statement

h-mayorquin:
> Because if you save with SI we add a custom column (`si_unit_id`) to make sure that when we load it back the same SI unit IDs are preserved. See the other part of the if-else statement

@alejoe91 Thanks. That makes sense.

alejoe91 (approved):

Issue #1512, opened 2023-04-10
no option to pass necessary params for whitening int recording


jguides:
A dtype or integer scale argument is required when whitening a recording with integer datatype (see https://github.com/LorenFrankLab/spyglass/issues/445 and https://github.com/LorenFrankLab/spyglass/issues/448). However, there is currently not a way to pass these arguments if you want to whiten just prior to spikesorting with mountainsort4. In the _run_from_folder method of the MountainSort SorterClass, only the recording is sent into the whiten method, with no option to specify the necessary dtype or int_scale parameters (see code snippet below). 

Could this be addressed? Alternatively, could you provide guidance on how to convert the recording to float16 prior to spikesorting?

Thanks in advance for any help with this.

The relevant part of the _run_from_folder method from spikeinterface/sorters/external/mountainsort4.py:
```
        # Whiten
        if p['whiten']:
            if verbose:
                print('whitening')
            recording = whiten(recording=recording)

```

Original error:
```
Running spike sorting on {'nwb_file_name': 'J1620210614_.nwb', 'sort_group_id': 0, 'sort_interval_name': 'raw data valid times no premaze no home no sleep', 'preproc_params_name': 'franklab_tetrode_hippocampus_min_seg', 'team_name': 'JG_DG', 'sorter': 'mountainsort4', 'sorter_params_name': 'franklab_tetrode_hippocampus_30KHz', 'artifact_removed_interval_list_name': 'J1620210614_.nwb_0_raw data valid times no premaze no home no sleep_franklab_tetrode_hippocampus_min_seg_JG_DG_group_0.25_500_8_2_artifact_removed_valid_times'}...
Traceback (most recent call last):
  File "/home/jguidera/anaconda3/envs/spyglass/lib/python3.9/site-packages/spikeinterface/sorters/basesorter.py", line 226, in run_from_folder
    SorterClass._run_from_folder(sorter_output_folder, sorter_params, verbose)
  File "/home/jguidera/anaconda3/envs/spyglass/lib/python3.9/site-packages/spikeinterface/sorters/external/mountainsort4.py", line 108, in _run_from_folder
    recording = whiten(recording=recording)
  File "/home/jguidera/anaconda3/envs/spyglass/lib/python3.9/site-packages/spikeinterface/core/core_tools.py", line 28, in reader_func
    return source_class(*args, **kwargs)
[...]

PR #1519, opened 2023-04-11
Add mountainsort5 sorter

magland:
Adding a new version of mountainsort spike sorter. See https://github.com/magland/mountainsort5

The structure of the sorting parameters is a bit more complicated than for mountainsort4 because the algorithm is available in three different schemes (scheme1, scheme2, scheme3) -- see the above link. Each scheme has its own set of parameters. So for the SpikeInterface wrapper, I flattened the structure... see mountainsort5.py. Depending on which scheme is selected, different parameters will take effect (and others will be ignored).

### DIFF
--- a/src/spikeinterface/sorters/external/mountainsort4.py
+++ b/src/spikeinterface/sorters/external/mountainsort4.py
@@ -105,7 +105,7 @@ class Mountainsort4Sorter(BaseSorter):
         if p['whiten']:
             if verbose:
                 print('whitening')
-            recording = whiten(recording=recording)
+            recording = whiten(recording=recording, dtype='float32')
 
         print('Mountainsort4 use the OLD spikeextractors mapped with NewToOldRecording')
         old_api_recording = NewToOldRecording(recording)
--- a/src/spikeinterface/sorters/external/tests/test_singularity_containers.py
+++ b/src/spikeinterface/sorters/external/tests/test_singularity_containers.py
@@ -66,6 +66,12 @@ def test_mountainsort4(run_kwargs):
     print(sorting)
 
 
+def test_mountainsort5(run_kwargs):
+    sorting = ss.run_sorter("mountainsort5", output_folder=cache_folder / "mountainsort5", **run_kwargs)
+    print("resulting sorting")
+    print(sorting)
+
+
 def test_tridesclous(run_kwargs):
     sorting = ss.run_sorter("tridesclous", output_folder=cache_folder / "tridesclous", **run_kwargs)
     print("resulting sorting")
--- a/doc/contribute.rst
+++ b/doc/contribute.rst
@@ -138,7 +138,7 @@ Now you can start filling out the required methods:
     def _run_from_folder(cls, output_folder, params, verbose):
 
         # Fill code to run your spike sorter based on the files created in the _setup_recording()
-        # You can run CLI commands (e.g. klusta, spykingcircus, tridesclous), pure Python code (e.g. Mountainsort4,
+        # You can run CLI commands (e.g. klusta, spykingcircus, tridesclous), pure Python code (e.g. Mountainsort5,
         # Herding Spikes), or even MATLAB code (e.g. Kilosort, Kilosort2, Ironclust)
 
     @classmethod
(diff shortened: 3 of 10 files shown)


src/spikeinterface/sorters/external/mountainsort4.py
```diff
@@ -105,7 +105,7 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
         if p['whiten']:
             if verbose:
                 print('whitening')
-            recording = whiten(recording=recording)
+            recording = whiten(recording=recording, dtype='float32')
```
alejoe91, line 108: this should address this issue too #1512

src/spikeinterface/sorters/external/tests/test_singularity_containers.py
```diff
@@ -66,6 +66,12 @@ def test_mountainsort4(run_kwargs):
     print(sorting)
 
 
+def test_mountainsort5(run_kwargs):
+    sorting = ss.run_sorter("mountainsort5", output_folder=cache_folder / "mountainsort5", **run_kwargs)
+    print("resulting sorting")
+    print(sorting)
```
alejoe91, line 72: @magland can you make a PR with a Dockerfile here: https://github.com/SpikeInterface/spikeinterface-dockerfiles ?

So we can run it through this singularity test :)

alejoe91 (approved):
Looks good to me! 

It would be great if you can also make an associated Docker image (which should be as straightforward as [mountainsort4](https://github.com/SpikeInterface/spikeinterface-dockerfiles/blob/main/mountainsort4/Dockerfile))

magland:
@alejoe91 before you merge...

I'm wondering about what will happen if the mountainsort5 parameters or api changes. This would break the SI wrapper. So maybe we should have SI depend on a narrow version range for mountainsort5? Then if I update to a new major version, then I'd make a new PR? What do you think?

alejoe91:
> @alejoe91 before you merge...
> 
> I'm wondering about what will happen if the mountainsort5 parameters or api changes. This would break the SI wrapper. So maybe we should have SI depend on a narrow version range for mountainsort5? Then if I update to a new major version, then I'd make a new PR? What do you think?

Yeah I think that's ok! You can check the version in the [is_installed()](https://github.com/magland/spikeinterface/blob/mountainsort5/src/spikeinterface/sorters/external/mountainsort5.py#L76-L82) and maybe warn the user if they need to downgrade mountainsort5?

I think this is the exact reason why we should have a working Dockerfile with the current version :)

magland:
@alejoe91 I added the check for mountainsort5 version 0.1.x
This requires the semantic_version package which is very standard (probably already installed on user's system). Should we make this explicit in the pyproject.toml, or just let it be?

https://github.com/magland/spikeinterface/blob/42f3c0da8b0c7690ea09a782b6506d8f7cbd6b9d/src/spikeinterface/sorters/external/mountainsort5.py#L76-L87

Next I will work on the singularity container.

magland:
@alejoe91 I have made a PR to spikeinterface-dockerfiles

alejoe91:
> @alejoe91 I added the check for mountainsort5 version 0.1.x This requires the semantic_version package which is very standard (probably already installed on user's system). Should we make this explicit in the pyproject.toml, or just let it be?
> 
> https://github.com/magland/spikeinterface/blob/42f3c0da8b0c7690ea09a782b6506d8f7cbd6b9d/src/spikeinterface/sorters/external/mountainsort5.py#L76-L87
> 
> Next I will work on the singularity container.

We use `from packaging.version import parse` in other places. Could you use that one?
```
from packaging.version import parse

if parse(mountainsort5.__version__) > parse("0.1"):
       print(f'WARNING: This version of SpikeInterface expects Mountainsort5 version 0.1.x. You have version {mountainsort5.__version__}')
       HAVE_MS5 = False
```

magland:
> We use `from packaging.version import parse` in other places. Could you use that one?
> 
> ```
> from packaging.version import parse
> 
> if parse(mountainsort5.__version__) > parse("0.1"):
>        print(f'WARNING: This version of SpikeInterface expects Mountainsort5 version 0.1.x. You have version {mountainsort5.__version__}')
>        HAVE_MS5 = False
> ```

Okay I have made the change.

alejoe91 (approved):

PR #1524, opened 2023-04-12
Warning that `setup.py` is legacy

DradeAW:
Warn the user when using `setup.py` for installation as it is now the legacy way of installing SpikeInterface.


h-mayorquin:
Thanks for changing the documentation. I think I will change the rest. 

In my opinion we should not add the warning about setup. The deprecation has been long in the making and it should not be our responsability to warn people about it at this stage: https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

(altought we were doing something bad by still advicing to use that as a method of install).

Up for @alejoe91 to decide if he wants to leave that warning in the setup.

alejoe91 (approved):

Issue #1514, opened 2023-04-11
WobbleMatch TODOs


pauladkisson:
A list of desired changes in the Wobble template matching algorithm to be implemented after benchmarking suite is established.  See [PR](https://github.com/SpikeInterface/spikeinterface/pull/1448#discussion_r1161608885) for more detailed rationale.

- [ ] - refractory_period_frames --> refractory_period_ms
- [ ] - sparsity --> spikeinterface ChannelSparsity
- [ ] - margin --> margin_ms
- [ ] - benchmark find_peaks vs. spikeinterface detect_peaks methods --> replace find_peaks with one of the existing peak detection methods or add find_peaks to the peak detection submodule
- [ ] - benchmark peak_radius
- [ ] - Fix IndexError that pops up in certain downsampling + parameter conditions

yger:
I would add unify the API with the other matching engines. I don't know if this is the best choice (and thus I'm open to discussion), but we decided with @samuelgarcia  to pass a waveform_extractor as an argument to the engines. I guess wobble could make use of that, it would be better for building pipelines and benchmarks

h-mayorquin:
Hi @yger.
I was the one that reviewed code and as I am less familiar with the signature of the other templates this escaped me. I totally agree with you that they should have the same signature. Can you point out to me to the signature of the other methods (is that their `__init__`  to see how that would look like?

The only possible counter-argument is that @pauladkisson did not want to go too far way from the Paninski lab implementation.   What do you think @pauladkisson ?

If you have time it would be great to get some review on the code. I think it would helpful for @pauladkisson to get a review from  someone that actually has expertise on the subject matter as my comments were only about code organization.

alejoe91:
Closing this, TM will need a refactoring anyways soon

PR #1527, opened 2023-04-12
Addition of a Template objects and refactor of matching engines

yger:
Here are some modifications for the Wobble algorithm, that I'll try to proofread intensively (and maybe steal ideas for the omp :-)). However, when running the algorithm out-of-the-box with default params, on a toy benchmark (32 channels), I do not get results as good as the ones you showed last time (see results below). Is it normal? Should I tweak params? I'll try with Neuropixel-like recordings. For now, all the engine are accepting as input a waveform_extractor. This is convenient, because we could also use the intrinsic sparsity there to factorize this among engines. 
I guess we should also have more benchmarks on times, and more unit tests as mentioned in #1514

image


alejoe91:
That looks ok for me! I think that the main reason why it was not implemented in the first place was to run this implementation against the original ones, which uses template arrays.

Maybe we can make both options available? Either pass a `WaveformExtractor`, or `templates`, `nbefore`, `nafter`

pauladkisson:
The reason why I used templates as an input instead of the waveform extractor is that the wobble algorithm can support multiple templates per unit, which would be hard (impossible?) to define in a waveform extractor depending on how you want the templates to look for each unit.  I'm not totally sure what the use case is to have multiple templates for each unit, but @cwindolf should be able to explain.

I agree that all the template matching algorithms should have the same mandatory inputs (obviously not optional parameters), but I think the more fundamental input is the templates themselves rather than the waveform extractor.  In general, one might want to perform template matching with templates from another recording, with some alterations, smoothing etc.  So, I think we should define template matching algorithms as starting with the templates as a given rather than extracting them in some way from the waveform extractor.

cwindolf:
Yes, there are multiple reasons for not using a waveform extractor
 - Multiple templates per unit (will explain this in detail soon)
 - Templates can be computed using fancy methods other than the default. Taking templates as an input instead of WE decouples template computation and deconvolution, which is important for us
 - What if we want to extract waveforms in a new way which is drift-aware? Current waveform tooling doesn't support this, so again for us it is helpful to decouple

yger:
Ok, as said I can be convinced. The safe option would be to add two options. Either a waveform extractor (and with sparsity methods that could be standardized among engines), or templates with nbefore/nafter. I'll work on that

yger:
Actually, here is what I'll do. For all methods, we will always provide a waveform_extractor (such that nbefore and nafter can be inferred), and a 'sparsity' field, such that if the waveform_extractor is not sparse, it will be sparsified via the core methods of spikeinterface and the params in this 'sparsity' field (this will allow finer comparisons between engines, because currently, each of them has its own ways of sparsifying). 

Meanwhile, methods will also have extra fields called templates and sparsity_mask. If these fields are provided, then the templates will be taken from here (bypassing the waveform_extractor), and same for the sparsity_mask. Will it be fine for everybody?

yger:
Yes, the best would be to have a from_templates() method and thus a single waveform_extractror. With the extra features that such an object should be able to handle several templates per units. But ok for the green light, I'll start with my solution for now

yger:
No i guess you are right, I just wanted to make use of existing objects, and because the waveform_extractor has a lot of  functions/attributes used by the engines, I thought about reusing it. But we'll see. Now the code is working here with the behavior explained before: if templates/sparsity_mask are manually provided, then fine, they are used by default. If a waveform_extractor is passed, otherwise, templates are obtained from it, and eventually sparsified with the same method for all sorters (that could be specified). This is important, because currently all engines are dealing with their own way to sparsify, such that we can not really compare among themselves
I'll be gone tomorrow for holidays, back on the 24th. We could have a call then to discuss the details, no hurry.

cwindolf:
I was reviewing this code and found a small bug -- in `subtract_spike_train`, some `np.subtract.at` statements in our impl have become `-=`, which is a problem when there are overlapping indices (there will be). For example,

```
x1 = 10 * np.ones(11)
x2 = 10 * np.ones(11)

# overlapping indices
inds = [1, 2, 2, 2, 3, 3]

x1[inds] -= 1
x1
# => array([10.,  9.,  9.,  9., 10., 10., 10., 10., 10., 10., 10.])

np.subtract.at(x2, inds, 1)
x2
# => array([10.,  9.,  7.,  8., 10., 10., 10., 10., 10., 10., 10.])
```

I'll keep reviewing and look for any further differences. To fix those, should I open a new PR, or what's the best way?

yger:
This is used in Circus (non omp). In fact, for every centroid, we use several snippets of waveforms to estimate the noise and optimize the range of the allowed amplitudes. But since omp is much better, not sure I want to maintain that, just making the point.
As said, I think the best would be to allow every method to work with either a Templates object (with temporal attributes and that can easily be obtained from a WaveformExtractor). Or maybe directly with the waveform_extractor (that can be more generic, also have precomputed noise levels, ...). Let's discuss that later. For now tests are failing for unknown and unclear reasons

pauladkisson:
> This is used in Circus (non omp).

Thanks.  I see that circus uses the waveform snippets in the `_optimize_amplitudes()` function.  It looks like it returns some kind of optimized amplitudes for each unit.

I still think this function could be pretty easily extracted out of the matching and pass the 'optimal amplitudes' into circus's matching alg.  

Maybe we could have another file in the matching submodule (or its own submodule?) called `extract_templates` or `curate_templates` (or something) to explicitly encode each sorter's preferred way to extract/curate templates from a waveform extractor.  Something to think about.

In any case, I look forward to discussing in more detail on Thursday.

yger:
@alejoe91 @samuelgarcia  I've tried to add a TemplatesDictionary object, feel free to let me know what do you think about it. All the matching methods are now working with it, and the best maybe would be to add to the waveform_extractor object an option to directly export such an object. @pauladkisson the sparsification performed internally by wobble could/should be removed then, for now I've left it there, but the plan would be to perform sparsification in a common framework, before sending templates to the engines

src/spikeinterface/sortingcomponents/matching/main.py
```diff
@@ -1,21 +1,200 @@
...
+
+from threadpoolctl import threadpool_limits
+import numpy as np
+
 from spikeinterface.core.job_tools import ChunkRecordingExecutor, fix_job_kwargs
 from spikeinterface.core import get_chunk_with_margin
 
 
-def find_spikes_from_templates(recording, method="naive", method_kwargs={}, extra_outputs=False, **job_kwargs):
+class TemplatesDictionary(object):
```
alejoe91, line 14: This class should live in core, see #1982

alejoe91:
@yger let's do this step by step and close this now.

First the `Templates` class (#1982) then the rest!

Issue #1409, opened 2023-03-09
New naming for peak data type 


h-mayorquin:
The current base_peak_dtype:
```
base_peak_dtype = [('sample_ind', 'int64'), ('channel_ind', 'int64'),
                   ('amplitude', 'float64'), ('segment_ind', 'int64')]
```
Do I remember correctly that we agreed to change this to have more explicit naming in a meeting but we did not follow through?

If that is the case, I can do it. 

What is the new naming?

- [x] * `sample_ind` -> `sample_index` #1556
- [x] * `channel_ind` -> `channel_index` #1534
- [x] * `amplitude` -> `ampitude`
- [x] * `segment_ind` -> `segment_index #1535
- [x] * `unit_id` -> `unit_index` [for spikevector in sorting objects] #1558
- [ ] * `cluster_ind` -> `cluster_index` #1571

samuelgarcia:
Hi
I am agree for `ind` -> `index`
But I think we can keep `amplitudes` 

I prefer `peaks['ampitude']` over `peaks['peak_ampitude']`

I you go the modification can you also check the changes:
 * inside sorter module
 * in doc

h-mayorquin:
Good point about the redundancy when referencing it from a peaks object!

alejoe91:
`indices` should be the correct plural

h-mayorquin:
@alejoe91 @samuelgarcia 
Note that neo uses `channel_indexes` (e.g. `_get_analogsignal_chunk`), are you guys fine with that discrepancy?

samuelgarcia:
Copy paste from the web:

**Both “indexes” and “indices” are correct English plurals of the singular noun “index.” But there is a preference for indices outside North America because of solid adherence to Latin plurals.**

I guess Alessio sometimes feel a latin guy.

h-mayorquin:
A Roman heritage. 

Outside of that it is easier to edit in bulk when the plurars are different from the singulars but on the other hand if you are looking for places where a term is used it might be a good idea that a simple search (no regular expression) would return both the terms (and that happens with the regular plural). I think that @pauladkisson pointed out something along those lines but can't recall.


No idea what's better. I think it is just better here to pick a standard and stick with it.

pauladkisson:
[...]

PR #1534, opened 2023-04-13
Change `channel_ind` for `channel_index` in `peak_dtype` for sortingcomponents 

h-mayorquin:
Let's try to move forward with #1409 which got sidestep because of the changes to main.

This PR deals with the changes for `channel_ind` to `channel_index` and only in the places where the peak structure is used. I will change the others in a PR and the rest of variables with that name more gradually.


h-mayorquin:
Thanks for getting those missed ones @alejoe91 !

PR #1537, opened 2023-04-14
Temporary supress nwb3 test

h-mayorquin:
This should cancel it while we merge another PR.

### DIFF
--- a/.github/workflows/s3-nwb-test.yml
+++ b/.github/workflows/s3-nwb-test.yml
@@ -1,49 +1,49 @@
-name: S3 NWB Test
-on:
-  pull_request:
-    types: [synchronize, opened, reopened]
-    branches:
-      - master
-      - main
-concurrency:  # Cancel previous workflows on the same pull request
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+# name: S3 NWB Test
+# on:
+#   pull_request:
+#     types: [synchronize, opened, reopened]
+#     branches:
+#       - master
+#       - main
+# concurrency:  # Cancel previous workflows on the same pull request
+#   group: ${{ github.workflow }}-${{ github.ref }}
+#   cancel-in-progress: true
 
-jobs:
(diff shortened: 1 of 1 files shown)


src/spikeinterface/extractors/tests/test_nwb_s3_extractor.py
```diff
@@ -36,7 +37,7 @@ def test_s3_nwb_ros3():
         trace_scaled = rec.get_traces(segment_index=segment_index, return_scaled=True, end_frame=2)
         assert trace_scaled.dtype == 'float32'
 
-
+@pytest.mark.streaming_extractors
```
alejoe91, line 40: @h-mayorquin are markers mutually exclusive? If not, we should make sure that in the `conftest.py` we skip this file when marking `extractors`

.github/workflows/full-test.yml
```diff
@@ -126,7 +126,7 @@ jobs:
         run: ./.github/run_tests.sh core
       - name: Test extractors
         if: ${{ steps.modules-changed.outputs.EXTRACTORS_CHANGED == 'true' || steps.modules-changed.outputs.CORE_CHANGED == 'true' }}
-        run: ./.github/run_tests.sh extractors
+        run: ./.github/run_tests.sh "extractors and not streaming_extractors"
```
h-mayorquin, line 129: @alejoe91 They are not exclusive. That's why this is changed here. 

On that logic, what about the IBI extractors which are also streaming?

src/spikeinterface/extractors/tests/test_nwb_s3_extractor.py
h-mayorquin: Check the other comment.

.github/workflows/full-test.yml
alejoe91: Correct, maybe we should move the IBL streamer tests in the s3 file?

alejoe91 (approved):

PR #1539, opened 2023-04-14
Fix crash in `export_to_phy`

DradeAW:
Fixes a crash in `export_to_phy` when `copy_binary=False`.

### DIFF
--- a/src/spikeinterface/exporters/tests/test_export_to_phy.py
+++ b/src/spikeinterface/exporters/tests/test_export_to_phy.py
@@ -41,6 +41,7 @@ def test_export_to_phy():
                   n_jobs=1, chunk_size=10000,
                   progress_bar=True)
 
+# Test for previous crash when copy_binary=False.
     export_to_phy(waveform_extractor, output_folder2,
                   compute_pc_features=False,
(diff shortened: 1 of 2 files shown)


h-mayorquin:
Is it possible to add a tests for this?

DradeAW:
This test passes with the new fix, but indeed crashes without it.

alejoe91 (approved):

src/spikeinterface/exporters/tests/test_export_to_phy.py
```diff
@@ -23,20 +23,30 @@ def test_export_to_phy():
...
             shutil.rmtree(f)
 
     waveform_extractor = extract_waveforms(recording, sorting, waveform_folder)
 
-    export_to_phy(waveform_extractor, output_folder,
+    export_to_phy(waveform_extractor, output_folder1,
                   compute_pc_features=True,
                   compute_amplitudes=True,
                   n_jobs=1, chunk_size=10000,
                   progress_bar=True)
```
h-mayorquin, line 43:
```suggestion

# Test for previous crash when copy_binary=False.
```

PR #1541, opened 2023-04-14
blank_staturation with a window around saturation signal

JulesLebert:
The preprocessing function that removes saturation just remove the samples where the signal reaches the threshold. This changes allow to also remove the signal around these saturations points (useful to remove parts of a recording where the cable was unplugged for example).

The problem with the modified clip function in pure python is that it is very slow. I have tried a vectorised approach that uses a convolve function from scipy.signal but it only improves performances marginally. Therefore, I have included a numba implementation of the function that increases the speed by ~10 times on my machine and that will only run if numba is installed.

### DIFF
--- a/src/spikeinterface/preprocessing/clip.py
+++ b/src/spikeinterface/preprocessing/clip.py
@@ -1,7 +1,7 @@
 import numpy as np
 
 try:
-    from numba import njit, guvectorize, float64
+    from numba import njit
     HAVE_NUMBA = True
 except ModuleNotFoundError as err:
     HAVE_NUMBA = False
@@ -77,6 +77,10 @@ class BlankSaturationRecording(BasePreprocessor):
     fill_value: float or None
         The value to write instead of the saturating signal. If None, then the value is
         automatically computed as the median signal value
+    ms_before: float (default 0)
+        Time (ms) to replace before the saturation signal
+    ms_after: float (default 0)
+        Time (ms) to replace after the saturation signal
     num_chunks_per_segment: int (default 50)
         The number of chunks per segments to consider to estimate the threshold/fill_values
     chunk_size: int (default 500)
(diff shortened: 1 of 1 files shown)


h-mayorquin (changes requested):
Thanks for your contribution.

My thinking in order of descending importance:
1) I think that the `ClipRecording` segment should not be modified. Right now, it is very straightforward and does its own thing very well. If we really want to have something like this we can just create another `RecordingSegment`. In brief, I feel the price of complexity is worse than the price of a small duplication. Finally, this should have some tests.
2) Are you sure that you need the loop for the numpy implementation? I think that you can calculate where the indexes of `np.were` change by more than one (`np.diff`) an just expand to the left and right accordingly. This will avoid the inner loop which is the most costly part of your function. Am I missing something?
3) I think that numba in general will play bad with our way of parallelizing with the `ChunkRecordingExecutor`. The former because I am not sure if it will require recompling per core which is usually a higher cost than the operation it implements. Have you tested this?

src/spikeinterface/preprocessing/clip.py
```diff
@@ -83,6 +93,13 @@ class BlankSaturationRecording(BasePreprocessor):
...
     """
+    name = 'blank_staturation'
+
+    def __init__(self, recording, abs_threshold=None, quantile_threshold=None,
+                 direction='upper', fill_value=None,
+                 ms_before=0, ms_after=0,
+                 num_chunks_per_segment=50, chunk_size=500, seed=0):
+
 
     name = "blank_staturation"
```
alejoe91, line 104:
```suggestion
```

alejoe91:
@JulesLebert is this still being worked on?

src/spikeinterface/preprocessing/clip.py
```diff
@@ -135,41 +151,106 @@ def __init__(
...
+        traces[max(0, index - frames_before):min(len(traces), index + frames_after + 1), chan] = value_max
+    return traces
+
+if HAVE_NUMBA:
+    # Numba
+    @njit(cache=True)
+    def _replace_slice_max_numba(traces, a_max, frames_before, frames_after, value_max):
+        m, n = traces.shape
+        to_clear = np.zeros(m, dtype=np.bool_)
+        for j in range(n):
```
samuelgarcia, line 227: you are looping channel then time.
I have the intuition that the reverse would be faster to have coalesing memory access.
no ?

samuelgarcia:
@JulesLebert 
This looks good to me.
I have the intuition that when could have ms_before=None ms_after=None by default and if there both None we keep the previous bahavior that used to be fast and simple.
What do you think ?

alejoe91:
Closing because integrated into #4297

PR #1543, opened 2023-04-17
Allow to set the recording of `WaveformExtractor`

DradeAW:
I need to load multiple wvf_extractor with the same `recording` object, to then compute the amplitudes.

For that, I cannot use `with_recording=False`, but if I set it to `True`, then it will load the same recording tens of times (which takes time and RAM).

I'd like the possibility to use `with_recording=recording`.

Thanks :)

### DIFF
--- a/src/spikeinterface/core/waveform_extractor.py
+++ b/src/spikeinterface/core/waveform_extractor.py
@@ -85,7 +85,7 @@ class WaveformExtractor:
                                            sorting.get_sampling_frequency(), decimal=2)
             if not recording.is_filtered() and not allow_unfiltered:
                 raise Exception('The recording is not filtered, you must filter it using `bandpass_filter()`.'
-                           3     'If the recording is already filtered, you can also do '
+                                'If the recording is already filtered, you can also do '
                                 '`recording.annotate(is_filtered=True).\n'
                                 'If you trully want to extract unfiltered waveforms, use `allow_unfiltered=True`.')
             self._rec_attributes = rec_attributes
--- a/src/spikeinterface/core/waveform_extractor.py
+++ b/src/spikeinterface/core/waveform_extractor.py
@@ -1,3 +1,4 @@
+import math
 import pickle
 from pathlib import Path
 import shutil
@@ -595,7 +596,7 @@ class WaveformExtractor:
         else:
             if recording.get_num_segments() != self.get_num_segments():
                 raise ValueError(f"Couldn't set the WaveformExtractor recording: num_segments do not match!\n{self.get_num_segments()} != {recording.get_num_segments()}")
-            if abs(recording.sampling_frequency - self.sampling_frequency) > 1e-2:
+            if not math.isclose(recording.sampling_frequency, self.sampling_frequency, abs_tol=1e-2, rel_tol=1e-5):
                 raise ValueError(f"Couldn't set the WaveformExtractor recording: sampling frequency doesn't match!\n{self.sampling_frequency} != {recording.sampling_frequency}")
             try:  # Will fail if called from __init__ because can't check from previous channel_ids.
                 if np.array_equal(self.channel_ids, recording.channel_ids):
@@ -1477,7 +1478,7 @@ def load_waveforms(folder, with_recording: bool = True, sorting: Optional[BaseSo
     ----------
     folder : str or Path
         The folder / zarr folder where the waveform extractor is stored
-    with_recording : bool | BaseRecording, optional
+    with_recording : bool, optional
         If True, the recording is loaded, by default True.
         If False, the WaveformExtractor object in recordingless mode.
     sorting : BaseSorting, optional
--- a/src/spikeinterface/core/waveform_extractor.py
+++ b/src/spikeinterface/core/waveform_extractor.py
@@ -1,6 +1,7 @@
 import pickle
 from pathlib import Path
 import shutil
+from typing import Optional
 import json
 
 import numpy as np
@@ -11,6 +12,7 @@ import probeinterface
 
 from .base import load_extractor
 from .baserecording import BaseRecording
+from .basesorting import BaseSorting
 from .core_tools import check_json
 from .job_tools import _shared_job_kwargs_doc, split_job_kwargs, fix_job_kwargs
 from .recording_tools import check_probe_do_not_overlap
@@ -63,35 +65,10 @@ class WaveformExtractor:
 
     """
     extensions = []
-    def __init__(self, recording, sorting, folder=None, rec_attributes=None, allow_unfiltered=False,
-                 sparsity=None):
-        if recording is None:
-            # this is for the mode when recording is not accessible anymore
-            if rec_attributes is None:
-                raise ValueError('WaveformExtractor: if recording is None then rec_attributes must be provied')
-            # some check on minimal attributes (probegroup is not mandatory)
-            for k in ('channel_ids', 'sampling_frequency', 'num_channels'):
-                if k not in rec_attributes:
-                    raise ValueError(f'Missing key in rec_attributes {k}')
-            for k in ('num_samples', 'properties', 'is_filtered'):
-                if k not in rec_attributes:
-                    warn(f'Missing optional key in rec_attributes {k}: '
-                         f'some recordingless functions might not be available')
-            self._rec_attributes = rec_attributes
-        else:
-            assert recording.get_num_segments() == sorting.get_num_segments(), \
-                "The recording and sorting objects must have the same number of segments!"
-            np.testing.assert_almost_equal(recording.get_sampling_frequency(),
-                                           sorting.get_sampling_frequency(), decimal=2)
-            if not recording.is_filtered() and not allow_unfiltered:
-                raise Exception('The recording is not filtered, you must filter it using `bandpass_filter()`.'
-                                'If the recording is already filtered, you can also do '
-                                '`recording.annotate(is_filtered=True).\n'
-                                'If you trully want to extract unfiltered waveforms, use `allow_unfiltered=True`.')
-            self._rec_attributes = rec_attributes
-
-        self._recording = recording
+    def __init__(self, recording: BaseRecording, sorting: BaseSorting,
+                 folder=None, rec_attributes=None, allow_unfiltered=False, sparsity=None):
         self.sorting = sorting
+        self.set_recording(recording, rec_attributes, allow_unfiltered)
 
         # cache in memory
         self._waveforms = {}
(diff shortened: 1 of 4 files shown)


src/spikeinterface/core/waveform_extractor.py
```diff
@@ -146,7 +147,10 @@ def load_from_folder(cls, folder, with_recording=True, sorting=None):
         folder = Path(folder)
         assert folder.is_dir(), f'This waveform folder does not exists {folder}'
 
-        if not with_recording:
+        if isinstance(with_recording, BaseRecording):
+            recording = with_recording
+            rec_attributes = None
+        elif not with_recording:
```
alejoe91, line 153: Can you change the docstring of the `extract_waveforms` function to reflect the new behavior?

DradeAW: I'm assuming you mean `load_waveforms`. It's done.

src/spikeinterface/core/waveform_extractor.py
```diff
@@ -1465,8 +1469,10 @@ def load_waveforms(folder, with_recording=True, sorting=None):
     ----------
     folder : str or Path
         The folder / zarr folder where the waveform extractor is stored
-    with_recording : bool, optional
+    with_recording : bool | BaseRecording, optional
         If True, the recording is loaded, by default True
+        If False, the WaveformExtractor object in recordless
+        If BaseRecording, the given recording is used
```
alejoe91, line 1475:
```suggestion
        If True, the recording is loaded, by default True.
        If False, the WaveformExtractor object in recordingless mode.
        If BaseRecording, the given recording is used as the WaveformExtractor.recording
```

DradeAW:
I don't know what this fail is about?

DradeAW:
I didn't quantify the RAM usage, but my recording is a `MEArecRecordingExtractor` with 1,500 units, 64 channels, and is 45-minutes long. So loading it takes ~7s (7s * ~20 = way too much time in my opinion)

I don't see the necessity to have redundant information when I can have all waveform extractor pointing to the same recording (20 pointers to the same object)

h-mayorquin:
The size of the underlying data should not affect the time it takes for a recording to be loaded. I think there is something wrong with the fact that your recording takes seven second too load and that is the problem here that should be addressed.

Let me elaborate. The problem is that a bunch of our code assumes that spikeinterface objects are cheap to load and that costly operations should happen only after (i.e. when you call `get_traces` or do some `run` method). Ideally, they should only be pointers to a file. If initialization is a a bottleneck in your code, that is something that we should address.

h-mayorquin:
You mean that an operation like:
```
recording, sorting = se.read_mearec(local_path)
```
Is taking seven seconds right?

alejoe91:
> You mean that an operation like:
> 
> ```
> import spikeinterface.extractors as se
> recording, sorting = se.read_mearec(local_path)
> ```
> 
> Is taking seven seconds right?

That's MEArec fault...

DradeAW:
`wvf_extractor = si.load_waveforms(folder, with_recording=True)` takes 7s while `with_recording=False` takes less than 0.1s.

Also `recording = se.MEArecRecordingExtractor("/mnt/hdd/data/MEArec/45m_1500cells/recording.h5")` takes 7s (not taking into account the time to import `spikeinterface.extractors`).

DradeAW:
But even with the `MEArec` problem not being 100% lazy, I still think it's a good idea to give the user the opportunity to give the recording object.
I know probably very few people will use it, but it's a few lines of code.

h-mayorquin:
I really think that it would be better to fix the parser in neo to only extract the information instead of creating a patch for something that should not be broken.

The problem is not the number of lines. The problem is increasing the complexity of the function by adding more logic and breaking the types of input parameters. The former creates more cases that need testing as they can go wrong while the latter makes it more difficult to document and reason about it. The code base is already complex, type loosy and under-tested. Small decisions like this will keep snowballing the API even more in that direction.

Don't we have a way of atttaching a recording to a waveform extractor that was loaded `with_recording = False`? That could be a solution that satisfies power and does not require modifying the API.

(Note that the savings gained by this would be dwarfed anyway if users need to do any parallel processing as every chunk will have the same overhead but hidden within the machinery of `ChunkRecordingExecutor`)

h-mayorquin:
> But even with the `MEArec` problem not being 100% lazy, I still think it's a good idea to give the user the opportunity to give the recording object. I know probably very few people will use it, but it's a few lines of code.

Thanks a lot for timing it. I have been thinking for a while about solving the problem of slow imports ...

DradeAW:
If it's possible to load using `with_recording=False` and then setting the recording object, I'm completely fine with that.

DradeAW:
> Don't we have a way of attaching a recording to a waveform extractor that was loaded `with_recording = False`?

Then we should expose the recording object in the `WaveformExtractor` (right now it's `_recording`)

alejoe91:
> > Don't we have a way of attaching a recording to a waveform extractor that was loaded `with_recording = False`?
> 
> Then we should expose the recording object in the `WaveformExtractor` (right now it's `_recording`)

Ok then it's settled: let's add a `set_recording()` function that sets the `_recording` after checking consistency with:
- `we.channel_ids`
- `we.get_channel_locations()`
- `we.sampling_frequency`

DradeAW:
Ok, it's not pretty and probably needs some refactoring, but here's a version that works

DradeAW:
> I think it looks good!

Thanks, but I'm not a fan of the `try, except` to check for channel_ids ^^

src/spikeinterface/core/waveform_extractor.py
```diff
@@ -599,6 +577,40 @@ def _reset(self):
...
+        else:
+            if recording.get_num_segments() != self.get_num_segments():
+                raise ValueError(f"Couldn't set the WaveformExtractor recording: num_segments do not match!\n{self.get_num_segments()} != {recording.get_num_segments()}")
+            if abs(recording.sampling_frequency - self.sampling_frequency) > 1e-2:
+                raise ValueError(f"Couldn't set the WaveformExtractor recording: sampling frequency doesn't match!\n{self.sampling_frequency} != {recording.sampling_frequency}")
+            try:  # Will fail if called from __init__ because can't check from previous channel_ids.
+                if np.array_equal(self.channel_ids, recording.channel_ids):
+                    raise ValueError(f"Couldn't set the WaveformExtractor recording: channel_ids do not match!\n{self.channel_ids} != {recording.channel_ids}")
+            except:
+                pass
```
alejoe91, line 604: Maybe check this only `if rec_attributes is not None`?

src/spikeinterface/core/waveform_extractor.py
```diff
@@ -599,6 +577,40 @@ def _reset(self):
...
+                if k not in rec_attributes:
+                    raise ValueError(f"WaveformExtractor: Missing key '{k}' in rec_attributes")
+            for k in ('num_samples', 'properties', 'is_filtered'):
+                if k not in rec_attributes:
+                    warn(f'Missing optional key in rec_attributes {k}: '
+                         f'some recordingless functions might not be available')
+        else:
+            if recording.get_num_segments() != self.get_num_segments():
+                raise ValueError(f"Couldn't set the WaveformExtractor recording: num_segments do not match!\n{self.get_num_segments()} != {recording.get_num_segments()}")
+            if abs(recording.sampling_frequency - self.sampling_frequency) > 1e-2:
```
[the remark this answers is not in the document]
DradeAW: Isn't that for arrays?

src/spikeinterface/core/waveform_extractor.py
```diff
@@ -1457,16 +1469,17 @@ def extract_waveforms(recording, sorting, folder=None,
...
     """
     Load a waveform extractor object from disk.
 
     Parameters
     ----------
     folder : str or Path
         The folder / zarr folder where the waveform extractor is stored
-    with_recording : bool, optional
-        If True, the recording is loaded, by default True
+    with_recording : bool | BaseRecording, optional
```
h-mayorquin, line 1480: Should't the BaseRecording be removed here?

src/spikeinterface/core/waveform_extractor.py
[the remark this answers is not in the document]
h-mayorquin: Ther is the equivalent in math from the standard library [math.is_close](https://docs.python.org/3/library/math.html#math.isclose)

h-mayorquin (commented):
This is good, it makes the `__init__` look more clean as a side effect.

src/spikeinterface/core/waveform_extractor.py
DradeAW: Done :)

src/spikeinterface/core/waveform_extractor.py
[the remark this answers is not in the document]
DradeAW: Done!

src/spikeinterface/core/waveform_extractor.py
DradeAW: You mean you could set a new recording with new attributes (like new `channel_ids`) that are different from the previous recording?

Wouldn't that break things?

alejoe91: The `__init__` is mainly called by the [`cls.create()`](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/waveform_extractor.py#L302) and the [`load()`](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/waveform_extractor.py#L184) functions. In both cases, the `rec_attributes` are given. As you mentioned, the `self.channel_ids` cannot be accessed if the object is not initialized aready, so it shouldn't be used in the function.

If you call the `set_recording()` after the object has been created with `with_recording=False`, then the `self._rec_attributes` is already set.

So here's what I would do:

In the `__init__`:

```
...
self._rec_attributes = None
self.set_recording(recording, rec_attributes, allow_unfiltered) # this will set `self._rec_attributes`
```

in the `set_recording()`:
```
if self._rec_attributes is not None:
    channel_ids = self._rec_attributes['channel_ids']
elif rec_attributes is not None:
    channel_ids = rec_attributes['channel_ids']
else:
    raise ValueError
if not np.array_equal(channel_ids, recording.channel_ids):
    raise ValueError(...)
```

DradeAW:
@alejoe91 It seems sometimes the `rec_attributes` are `None` (the tests are failing)

alejoe91:
Thanks, I'll check

src/spikeinterface/core/waveform_tools.py
```diff
@@ -376,3 +377,32 @@ def has_exceeding_spikes(recording, sorting):
             if spike_vector_seg["sample_index"][-1] > recording.get_num_samples(segment_index=segment_index) - 1:
                 return True
     return False
+
+
+def get_rec_attributes(recording):
```
alejoe91, line 382: Move to `recording_tools`?

DradeAW:
Btw, there is no test for `get_rec_attributes`

DradeAW:
No sorry, I have to prepare a talk I'm giving at a conference next week!

alejoe91 (approved):

PR #1544, opened 2023-04-17
Fix bug in contamination if unit is empty

DradeAW:


src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -370,8 +370,11 @@ def compute_refrac_period_violations(waveform_extractor, refractory_period_ms: f
     for i, unit_id in enumerate(sorting.unit_ids):
         nb_violations[unit_id] = n_v = nb_rp_violations[i]
         N = num_spikes[unit_id]
-        D = 1 - n_v * (T - 2*N*t_c) / (N**2 * (t_r - t_c))
-        rp_contamination[unit_id] = 1 - math.sqrt(D) if D >= 0 else 1.0
+        if N == 0:
+            rp_contamination[unit_id] = 0.0
```
alejoe91, line 374:
```suggestion
            rp_contamination[unit_id] = np.nan
```
?

alejoe91: IMO if there are no spikes, you can't compute contamination..it's true that it's harder to deal with nans, but I think it's more correct in this case

alejoe91 (approved):

h-mayorquin:
Can we add a tests for this?

alejoe91 (approved):

PR #1545, opened 2023-04-17
Check wvf_extractor with recording for spike amplitudes

DradeAW:
Without this check, there will just be a crash later on (after having created the `spike_amplitudes` folder). Thus, if the user re-runs with a correct `WaveformExtractor`, there will be a crash because the module will try to load an empty folder.

I think the simplest way to avoid problems like this is the check form the beginning if the `WaveformExtractor` object contains a recording object.

### DIFF
--- a/src/spikeinterface/postprocessing/spike_amplitudes.py
+++ b/src/spikeinterface/postprocessing/spike_amplitudes.py
@@ -172,7 +172,7 @@ def compute_spike_amplitudes(waveform_extractor, load_if_exists=False,
             - If 'concatenated' all amplitudes for all spikes and all units are concatenated
             - If 'by_unit', amplitudes are returned as a list (for segments) of dictionaries (for units)
     """
-    if load_if_exists and waveform_extractor.is_extensionSpikeAmplitudesCalculator.extension_name(SpikeAmplitudesCalculator.extension_name):
+    if load_if_exists and waveform_extractor.is_extension(SpikeAmplitudesCalculator.extension_name):
         sac = waveform_extractor.load_extension(SpikeAmplitudesCalculator.extension_name)
     else:
         sac = SpikeAmplitudesCalculator(waveform_extractor)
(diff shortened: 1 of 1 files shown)


DradeAW:
Maybe we should add a test to check that we can still load in recordless mode?

alejoe91:
I'm pretty sure this will fail in recording less mode. We have an assertion in the _run function, because only in that case we actually need the recording. I don't understand why this is needed in the init

DradeAW:
I see 2 problems:

1- The error is not completely clear: `ValueError: WaveformExtractor is used in mode "with_recording=False" this operation needs the recording`
What operation?

2- in the `_run` function, the folder `spike_amplitudes` has already been created in the waveform_extractor folder (but is empty since it did not manage to run). So when you re-run your code with a wvf_extractor that contains a recording object, the line `spikeinterface.postprocessing.compute_spike_amplitudes(wvf_extractor, load_if_exists=True)` fails with the following error message:
```
  File "benchmark.py", line 145, in <module>
    spike_amplitudes = {name: spost.compute_spike_amplitudes(wvf_extractors[name], load_if_exists=True, outputs="by_unit") for (name, sorting) in tqdm(sortings.items())}
  File "benchmark.py", line 145, in <dictcomp>
    spike_amplitudes = {name: spost.compute_spike_amplitudes(wvf_extractors[name], load_if_exists=True, outputs="by_unit") for (name, sorting) in tqdm(sortings.items())}
  File "/export/home1/users/nsr/wyngaard/dev/spikeinterface/spikeinterface/src/spikeinterface/postprocessing/spike_amplitudes.py", line 178, in compute_spike_amplitudes
    amps = sac.get_data(outputs=outputs)
  File "/export/home1/users/nsr/wyngaard/dev/spikeinterface/spikeinterface/src/spikeinterface/postprocessing/spike_amplitudes.py", line 121, in get_data
amps = self._extension_data[f'amplitude_segment_{segment_index}'][mask]
KeyError: 'amplitude_segment_0'
```

DradeAW:
The other solution would be a:

```python
try:
    recording = we.recording
except:
    print("Error message")
    # Delete folder
```

alejoe91:
I prefer the following solution:
```
if not we.has_recording():
    # delete folder
    raise error
```

alejoe91 (approved):

PR #1548, opened 2023-04-18
Allow option to compute noise with std

DradeAW:
In some cases, it can be useful to compute noise with std instead of mad.

### DIFF
--- a/src/spikeinterface/core/recording_tools.py
+++ b/src/spikeinterface/core/recording_tools.py
@@ -1,6 +1,5 @@
 from typing import Literal
 import numpy as np
-from .baserecording import BaseRecording
 
 
 def get_random_data_chunks(
@@ -120,7 +119,7 @@ def get_closest_channels(recording, channel_ids=None, num_channels=None):
     return np.array(closest_channels_inds), np.array(dists)
 
 
-def get_noise_levels(recording: BaseRecording, return_scaled: bool = True, method: Literal['mad', 'std'] = "mad", **random_chunk_kwargs):
+def get_noise_levels(recording: 'BaseRecording', return_scaled: bool = True, method: Literal['mad', 'std'] = "mad", **random_chunk_kwargs):
     """
     Estimate noise for each channel using MAD methods.
     You can use standard deviation with `method='std'`
--- a/src/spikeinterface/core/recording_tools.py
+++ b/src/spikeinterface/core/recording_tools.py
@@ -118,7 +118,7 @@ def get_closest_channels(recording, channel_ids=None, num_channels=None):
     return np.array(closest_channels_inds), np.array(dists)
 
 
-def get_noise_levels(recording, return_scaled=True, **random_chunk_kwargs):
+def get_noise_levels(recording, return_scaled: bool = True, method: str = "mad", **random_chunk_kwargs):
     """
     Estimate noise for each channel using MAD methods.
 
@@ -129,11 +129,16 @@ def get_noise_levels(recording, return_scaled=True, **random_chunk_kwargs):
     random_chunks = get_random_data_chunks(
         recording, return_scaled=return_scaled, **random_chunk_kwargs
     )
-    med = np.median(random_chunks, axis=0, keepdims=True)
-    # hard-coded so that core doesn't depend on scipy
-    noise_levels = (
-        np.median(np.abs(random_chunks - med), axis=0) / 0.6744897501960817
-    )
+    
+    if method == "mad":
+        med = np.median(random_chunks, axis=0, keepdims=True)
+        # hard-coded so that core doesn't depend on scipy
+        noise_levels = (
+            np.median(np.abs(random_chunks - med), axis=0) / 0.6744897501960817
+        )
+    elif method == "std":
+        mean = np.mean(random_chunks, axis=0, keepdims=True)
+        noise_levels = np.mean(random_chunks - mean, axis=0)  # std = E(X - E(X))
     return noise_levels
 
 
--- a/src/spikeinterface/core/recording_tools.py
+++ b/src/spikeinterface/core/recording_tools.py
@@ -138,7 +138,7 @@ def get_noise_levels(recording, return_scaled: bool = True, method: str = "mad",
         )
     elif method == "std":
         mean = np.mean(random_chunks, axis=0, keepdims=True)
-        noise_levels = np.mean(random_chunks - mean, axis=0)  # std = E(X - E(X))
+        noise_levels = np.sqrt(np.mean((random_chunks - mean)**2, axis=0))  # std = sqrt( E( (X - E(X))^2 ) )
     return noise_levels
 
 
--- a/src/spikeinterface/core/recording_tools.py
+++ b/src/spikeinterface/core/recording_tools.py
@@ -1,4 +1,6 @@
+from typing import Literal
 import numpy as np
+from .baserecording import BaseRecording
 
 
 def get_random_data_chunks(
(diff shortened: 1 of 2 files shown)


DradeAW:
Is this enough data by default?

```python
def get_random_data_chunks(
    ...,
    num_chunks_per_segment=20,
    chunk_size=10000,
    ...
):
```

For MAD with NP1.0 it seems enough because of the accuracy (2.34), but for other probes or std it fluctuates a bit ...

DradeAW:
I mean that's probably okay for a 5-min recording.
For a 30-min recording, just 20 chunks seems a bit low for me ...

alejoe91:
> I mean that's probably okay for a 5-min recording. For a 30-min recording, just 20 chunks seems a bit low for me ...

We have to assume stationarity here, so I don't think it will make a big difference. Have you made some tests?

DradeAW:
> Have you made some tests?

I computed the noise levels (with `method="std")` 100 times, and showed the mean and std of this result:

```
>>> np.mean(res, axis=0)
array([19.51544096, 24.3684446 , 22.91798923, 24.20358974, 24.18334831,
       23.74148212, 25.29098692, 21.57289976, 23.14343668, 30.25203733,
       26.04176506, 24.52578871, 25.79722738, 23.96358292, 23.88542953,
       23.86176592, 24.61518469, 22.60433542, 23.75856028, 22.08723293,
       21.28915342, 22.72654383, 22.43020218, 22.75144594, 25.969058  ,
       25.98043869, 23.50014816, 27.69608162, 25.09962038, 32.49848663,
       23.19088758, 24.92593204, 21.41117277, 23.46025547, 21.15608473,
       27.60597191, 20.91847763, 20.66591206, 22.03754534, 22.42749453,
       23.28249779, 24.92094481, 23.82594101, 23.57078457, 27.56730118,
       25.30133888, 29.89781143, 23.48036154, 23.31993656, 23.05903481,
       24.00712648, 22.47873859, 23.27643503, 22.36511431, 22.69200449,
       23.58804417, 23.88655197, 26.57735249, 23.05347614, 24.70152716,
       27.17167376, 25.51239563, 23.03764837, 27.9972732 ])
>>> np.std(res, axis=0)
array([0.36302169, 0.57957596, 0.28107731, 0.27803827, 0.36665211,
       0.49272872, 0.50085993, 0.20381841, 0.26136973, 0.51107587,
       0.4087304 , 0.25292829, 0.69405322, 0.27178649, 0.28866488,
       0.38217422, 0.52497691, 0.35228526, 0.25876439, 0.25192517,
       0.2224251 , 0.42089143, 0.27307765, 0.27516651, 0.3259604 ,
       0.35148434, 0.27715796, 0.82148874, 0.54672281, 0.82671499,
       0.24867518, 0.34700628, 0.6101986 , 0.37241242, 0.22561205,
       0.69657939, 0.2107673 , 0.19711612, 0.27310834, 0.24504242,
       0.40133192, 0.33939249, 0.31760637, 0.2936658 , 0.80130423,
       0.40363801, 0.5438401 , 0.24563333, 0.42686139, 0.33198546,
       0.69694509, 0.29156065, 0.35224721, 0.31635233, 0.25122586,
       0.28231002, 0.30492242, 0.60238389, 0.26313978, 0.40527181,
       0.69469699, 0.32415631, 0.22623678, 0.38090247])
```

Computing the noise levels 100 times took 12s (~200ms/run) with the recording on an HDD.
This means the std is about 1-2% of the mean (with a MEArec recording that is stationary).

alejoe91:
Which is ok, no? I also think that MAD will probably be more robust because it's less sensitive to the presence of spikes in the first place

DradeAW:
I think it's ok under the assumption that the recording is stationary.
But if it's not, then 20 chunks are probably far from enough.

Let me check with one of my dataset from cerebellar cortex (which have little to no drift)

src/spikeinterface/core/recording_tools.py
```diff
@@ -129,11 +129,16 @@ def get_noise_levels(recording, return_scaled=True, **random_chunk_kwargs):
...
+    
+    if method == "mad":
+        med = np.median(random_chunks, axis=0, keepdims=True)
+        # hard-coded so that core doesn't depend on scipy
+        noise_levels = (
+            np.median(np.abs(random_chunks - med), axis=0) / 0.6744897501960817
+        )
+    elif method == "std":
+        mean = np.mean(random_chunks, axis=0, keepdims=True)
+        noise_levels = np.sqrt(np.mean((random_chunks - mean)**2, axis=0))  # std = sqrt( E( (X - E(X))^2 ) )
```
h-mayorquin, line 141: https://numpy.org/doc/stable/reference/generated/numpy.std.html

src/spikeinterface/core/recording_tools.py
```diff
@@ -118,7 +118,7 @@ def get_closest_channels(recording, channel_ids=None, num_channels=None):
     return np.array(closest_channels_inds), np.array(dists)
 
 
-def get_noise_levels(recording, return_scaled=True, **random_chunk_kwargs):
+def get_noise_levels(recording, return_scaled: bool = True, method: str = "mad", **random_chunk_kwargs):
```
h-mayorquin, line 121: Thanks for adding typings, could you add the rest?
```
def get_noise_levels(
    recording: BaseRecording, 
    return_scaled: bool = True, 
    method: Literal['std', 'mad'] = 'mad', 
    **random_chunk_kwargs
) -> np.array
```

h-mayorquin:
Note that for the stationary case you can calculate the number of samples that will make the confidence interval the same length as the effect size that you care about and get the exact size of N that you need:

https://en.wikipedia.org/wiki/Standard_deviation#Confidence_interval_of_a_sampled_standard_deviation

For the non-stationary case there is no a big N that will save you as you are chasing a moving target, right?

src/spikeinterface/core/recording_tools.py
DradeAW: Ah didn't know about `Literal`, that's quite neat!
Will add it in the next commit :)

src/spikeinterface/core/recording_tools.py
```diff
@@ -118,22 +119,27 @@ def get_closest_channels(recording, channel_ids=None, num_channels=None):
     return np.array(closest_channels_inds), np.array(dists)
 
 
-def get_noise_levels(recording, return_scaled=True, **random_chunk_kwargs):
+def get_noise_levels(recording: 'BaseRecording', return_scaled: bool = True, method: Literal['mad', 'std'] = "mad", **random_chunk_kwargs):
```
h-mayorquin, line 122: Why BaseRecording and not the actual class?

src/spikeinterface/core/tests/test_recording_tools.py
```diff
@@ -32,6 +32,9 @@ def test_get_noise_levels():
     noise_levels = get_noise_levels(rec, return_scaled=True)
     print(noise_levels)
 
+    noise_levels = get_noise_levels(rec, return_scaled=True, method="std")
```
h-mayorquin, line 35: Thanks for adding a tests to this. I don't see the point of printing here though (I know the other one is doing it).

I think this test could use the NumpyExtractor using a trace whose statistics we know (because we created) and test that the noise level is the one that was used to generate the traces. 

What do you think?

DradeAW: Yeah I didn't understand printing either, but just kept it for consistency.

Makes sense, we could also check mad that way by simply generating a `NumpyRecording` that follows a gaussian of known standard deviation :)

src/spikeinterface/core/recording_tools.py
h-mayorquin: Why I mean here, is why quotes.
`BaseRecording` vs BaseRecording. You are getting a circular import error?

Does `BaseRecording` by itself just works?

DradeAW: Yes, the actual class causes a circular import error

alejoe91 (approved):

PR #1551, opened 2023-04-18
Remove insecure `eval` for importlib in neo extractors

h-mayorquin:
I have been meaning to do this for a while. This will achieve the same dynamic importing functionality without the vulnerability of someone creatinga class with malicious code in `self.NeoRawIOClass` and executing this somewhere else. 

I also change the naming for something more meaningful and added a docstring.


alejoe91 (approved):
Thanks! Definitely better than my `eval`-hacking

PR #1553, opened 2023-04-18
Refactor detect peak tests

h-mayorquin:
WIth this PR I want to start a re-organization of the peak detection testing battery.  The idea is having recall test for all the methods which they should pass with 100 % for the mearec data. 

I am also doing a refactor to start separting the methods to a more unit-test-like organization. I am envisioning that we can have a unit-test for each specific method that will test the following for the mearec data:
* Recall (this PR).
* No spikes too close in time for the by_channel methods
* No spikes too close in space for `exclusive` methods.

In this PR I am only adding the first point but I am hoping that in the future when we add a new method we can just simply add an extra unit-test where it will be easy to just tests for all the criteria above and possible more.

### DIFF
--- a/src/spikeinterface/sortingcomponents/tests/test_peak_detection.py
+++ b/src/spikeinterface/sortingcomponents/tests/test_peak_detection.py
@@ -4,7 +4,7 @@ from pathlib import Path
 import pytest
 
 from spikeinterface import download_dataset
-from spikeinterface.extractors import MEArecRecordingExtractor
+from spikeinterface.extractors.neoextractors.mearec import MEArecRecordingExtractor, MEArecSortingExtractor
 
 from spikeinterface.sortingcomponents.peak_detection import detect_peaks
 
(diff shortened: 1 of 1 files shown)


alejoe91:
@h-mayorquin I don't think that unit tests should check the correctness of the method in terms of precision. This would be  better suited for a benchmark function and test. IMO it's ok if the tests only check that the method runs correctly. Let's wait to hear @samuelgarcia 's input on this

h-mayorquin:
Fair enough, @alejoe91.
Just note that the right now the only thing that is tested is that there are no assertion being thrown. That is a very wide definition of "runs correctly" in my opinion. That is, you could return wrong shape, wrong type or wrong calculation altogether without an assertion being thrown. Doing a simple output check, when it is possible, condenses covers a lot of wrong. Plus, it creates an anchor that should remain estable if the is optimized in the future. 

But yes, let's wait for @samuelgarcia to be back. It will be great  good to hear from you guys what vision do you have for how ideal testing should look in spikeinterface. Then I can implement your vision more faithfully.

PR #1557, opened 2023-04-19
Added helpful error message if `cuda` missing

DradeAW:
Added an helpful error message for the user in case he tries to run a sorter dockerized without `cuda-python` installed in the environment.

### DIFF
--- a/src/spikeinterface/sorters/utils/misc.py
+++ b/src/spikeinterface/sorters/utils/misc.py
@@ -62,7 +62,11 @@ def has_nvidia():
     """
     Checks if the machine has nvidia capability.
     """
-    from cuda import cuda
+    
+    try:
+        from cuda import cuda
+    except ModuleNotFoundError as err:
+        raise Exception("This container requires cuda, but the package is not installed. You can install it with:\npip install cuda-python") from err
         
     try:
         cu_result_init,  = cuda.cuInit(0)
--- a/src/spikeinterface/sorters/utils/misc.py
+++ b/src/spikeinterface/sorters/utils/misc.py
@@ -66,7 +66,7 @@ def has_nvidia():
     try:
         from cuda import cuda
     except ModuleNotFoundError as err:
-        raise Exception("This container requires cuda, but the package is not installed. You can install it with:\npip install cuda-python") from err
+        raise Exception("This sorter requires cuda, but the package is not installed. You can install it with:\npip install cuda-python") from err
         
     try:
         cu_result_init,  = cuda.cuInit(0)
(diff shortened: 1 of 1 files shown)


src/spikeinterface/sorters/utils/misc.py
```diff
@@ -62,12 +62,16 @@ def has_nvidia():
     """
     Checks if the machine has nvidia capability.
     """
-    from cuda import cuda
+    
+    try:
+        from cuda import cuda
+    except ModuleNotFoundError as err:
+        raise Exception("This container requires cuda, but the package is not installed. You can install it with:\npip install cuda-python") from err
```
alejoe91, line 69: This is strange because `cuda-python` is included in `full` (see https://github.com/SpikeInterface/spikeinterface/blob/main/pyproject.toml#L93) and we install `[full]` in the container: https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/sorters/runsorter.py#L494-L519

I'm assuming you're using a custom image then @DradeAW ?

DradeAW: I install SI using `pip install -e .` which only installs the core dependencies (and not the full). So I always install the required packages one by one if I see one is missing (probably not the best way to do it, but oh well).

For the image I'm using the regular one for `kilosort3`.

DradeAW: What do you mean?

I'm in my standard conda environment, and this env is missing `cuda-python`, so the line `ss.run_sorter("kilosort3", singularity_image=True)` crashes.

alejoe91, line 69:
```suggestion
        raise Exception("This sorter requires cuda, but the package is not installed. You can install it with:\npip install cuda-python") from err
```
this is really dependent on the sorter, and not on running in a container or not.

alejoe91: also, `cuda-python` is only used to check if the GPU is available. Maybe you can use this message:
`"The cuda-python package is required to check if a GPU is available. You can install it with:\npip install cuda-python"`

DradeAW: Yes but it's only checked if the sorter uses cuda, so is it important for the user to go into the details of exactly why this package is required?

alejoe91: then I would do `requires cuda-python`

DradeAW:
@alejoe91 How's that?

alejoe91 (approved):

PR #1562, opened 2023-04-20
Check extractor properties

DradeAW:
Add a test to check if annotations/properties of two extractors are equal.

### DIFF
--- a/src/spikeinterface/core/testing.py
+++ b/src/spikeinterface/core/testing.py
@@ -1,5 +1,8 @@
 import numpy as np
 from numpy.testing import assert_array_equal
+from .base import BaseExtractor
+from .baserecording import BaseRecording
+from .basesorting import BaseSorting
 
 
 def check_sorted_arrays_equal(a1, a2):
@@ -8,7 +11,7 @@ def check_sorted_arrays_equal(a1, a2):
     assert_array_equal(a1, a2)
 
 
-def check_recordings_equal(RX1, RX2, return_scaled=True, force_dtype=None):
+def check_recordings_equal(RX1: BaseRecording, RX2: BaseRecording, return_scaled=True, force_dtype=None, check_properties: bool = False) -> None:
     assert RX1.get_num_segments() == RX2.get_num_segments()
 
     for segment_idx in range(RX1.get_num_segments()):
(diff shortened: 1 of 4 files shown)


DradeAW:
Still in draft stage, I need to test this a little bit.

@alejoe91 @h-mayorquin Do you have any input?

src/spikeinterface/core/testing.py
```diff
@@ -73,22 +68,15 @@ def check_sortings_equal(SX1, SX2):
...
-#                 assert SX1.get_unit_property(u, prop) == SX2.get_unit_property(u, prop)
-#     # check features
-#     print('Features', sorted(SX1.get_shared_unit_spike_feature_names()), sorted(SX2.get_shared_unit_spike_feature_names()))
-#     assert sorted(SX1.get_shared_unit_spike_feature_names()) == sorted(SX2.get_shared_unit_spike_feature_names())
-#     for feat in SX1.get_shared_unit_spike_feature_names():
-#         for u in SX1.get_unit_ids():
-#             assert np.allclose(np.array(SX1.get_unit_spike_features(u, feat)),
-#                                np.array(SX2.get_unit_spike_features(u, feat)))
+    if check_properties:
+        check_extractor_properties_equal(RX1, RX2)
```
alejoe91, line 72:
```suggestion
        check_extractor_properties_equal(SX1, SX2)
```

src/spikeinterface/core/testing.py
```diff
@@ -73,22 +68,15 @@ def check_sortings_equal(SX1, SX2):
...
+
+
+def check_extractor_properties_equal(EX1, EX2) -> None:
+    assert np.array_equal(sorted(EX1.get_annotation_keys()), sorted(EX2.get_annotation_keys()))
+    assert np.array_equal(sorted(EX1.get_property_keys()), sorted(EX2.get_property_keys()))
+
+    for annotation_name in EX1.get_annotation_keys():
+        assert EX1.get_annotation(annotation_name) == EX2.get_annotation(annotation_name)
+    for property_name in EX1.get_property_keys():
+        np.assert_array_equal(EX1.get_property(property_name), EX2.get_property(property_name))
```
alejoe91, line 82: I would split properties and annotations honestly. For example, a `bandpass_filter` will add the `is_filtered=True` annotation, but properties will be the same!

DradeAW: True, done :)

DradeAW:
It seems that saving a recording to `zarr` doesn't save the annotations properly.

+ Is this normal that saving annotations/properties for `dump_to_json` is hardcoded like this?
https://github.com/SpikeInterface/spikeinterface/blob/1bd169fc19673c60b7335d51ff38f758fd1fcc78/src/spikeinterface/core/base.py#L479

h-mayorquin:
> It seems that saving a recording to `zarr` doesn't save the annotations properly.
> 
> + Is this normal that saving annotations/properties for `dump_to_json` is hardcoded like this?
> 
> https://github.com/SpikeInterface/spikeinterface/blob/1bd169fc19673c60b7335d51ff38f758fd1fcc78/src/spikeinterface/core/base.py#L479-L480

IO in base is kind of a mess. I remember asking about this specifically to Sam and he had a reason for that. If you look at `save_to_folder` you will see that the properties are there saved there, after the properties are dump -called provenance- so I think it was made originally so that properties were not duplicated. Also, I think the properties in the probes are saved in a different way because there was a reason to try to keep that in numpy.

We should probably document the current treatment of properties in IO, agree on what is desirable and design a general approach of it that is documented and tested.

alejoe91:
> It seems that saving a recording to `zarr` doesn't save the annotations properly.

I'm making a PR ;)

DradeAW:
Nope it didn't :/

alejoe91:
> Nope it didn't :/

I looked into it and that's because the `ZarrRecording` also adds the `compression_ratio` as annotation. Maybe we can add a kwarg to check that all annotations and properties of EX1 are in EX2?

DradeAW:
> Maybe we can add a kwarg to check that all annotations and properties of EX1 are in EX2?

We can, but we're adding more complexity to such functions,

Wouldn't it be easier to do that in the tests?
```python
assert "compression_ratio" in zarr_recording.get_annotation_keys()
zarr_recording.remove_annotation("compression_ratio")
check_recordings_equal(recording, zarr_recording, check_annotations=True)
```

alejoe91 (approved):

PR #1563, opened 2023-04-20
Add iterative peak detection

h-mayorquin:
This is the iterative peak detection method. 
I added a basic so let's discuss if anything more is necessary.

### DIFF
--- a/src/spikeinterface/sortingcomponents/peak_detection.py
+++ b/src/spikeinterface/sortingcomponents/peak_detection.py
@@ -145,7 +145,7 @@ class IterativePeakDetector(PeakDetector):
         waveforms = self.waveform_extraction_node.compute(traces=traces_chunk, peaks=local_peaks)
         denoised_waveforms = self.waveform_denoising_node.compute(traces=traces_chunk, peaks=local_peaks, waveforms=waveforms)
         
-        denoised_waveforms_as_traces = self.build_trace_chunk_with_waveforms(sample_indices=local_peaks["sample_ind"], traces_chunk=traces_chunk, 
+        denoised_waveforms_as_traces = self.build_trace_chunk_with_waveforms(sample_indices=local_peaks["sample_index"], traces_chunk=traces_chunk, 
                                                                             waveforms=denoised_waveforms, 
                                                                             extract_dense_waveforms=self.waveform_extraction_node)
         traces_chunk_minus_peak_waveforms = traces_chunk - denoised_waveforms_as_traces
--- a/src/spikeinterface/sortingcomponents/peak_detection.py
+++ b/src/spikeinterface/sortingcomponents/peak_detection.py
@@ -9,7 +9,7 @@ from spikeinterface.core.job_tools import (ChunkRecordingExecutor, _shared_job_k
 from spikeinterface.core.recording_tools import get_noise_levels, get_channel_distances
 
 from spikeinterface.core.baserecording import BaseRecording
-from spikeinterface.sortingcomponents.peak_pipeline import PeakDetector, WaveformsNode
+from spikeinterface.sortingcomponents.peak_pipeline import PeakDetector, WaveformsNode, ExtractSparseWaveforms
 
 from ..core import get_chunk_with_margin
 
@@ -259,7 +259,8 @@ class IterativePeakDetector(PeakDetector):
         return local_peaks_expanded
 
     def substract_waveforms_from_traces(
-        self, traces_chunk, local_peaks, denoised_waveforms
+        self, traces_chunk, 
+        local_peaks, denoised_waveforms,
     ) -> Tuple[np.ndarray, np.ndarray]:
         """
         Subtract denoised waveforms from traces.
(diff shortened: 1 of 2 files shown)


src/spikeinterface/sortingcomponents/peak_detection.py
```diff
@@ -105,6 +105,96 @@ def detect_peaks(recording, method='by_channel', pipeline_nodes=None,
...
+            waveforms = self.waveform_extraction_node.compute(traces=traces_chunk, peaks=local_peaks)
+            denoised_waveforms = self.waveform_denoising_node.compute(traces=traces_chunk, peaks=local_peaks, waveforms=waveforms)   
+            traces_chunk_without_waveforms, _ = self.substract_waveforms_from_traces(traces_chunk, local_peaks, denoised_waveforms)            
+            
+            # update traces_chunk to run the algorithm again and store local_peaks
+            traces_chunk = traces_chunk_without_waveforms
+            
+        all_local_peaks = np.concatenate(local_peaks_list, axis=0)
+        
+        return (all_local_peaks, )
```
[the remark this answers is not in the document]
h-mayorquin: Returning something else breaks the pipeline right now.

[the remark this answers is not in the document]
h-mayorquin: That feature does not work at the moment for PeakDetector.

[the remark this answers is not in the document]
alejoe91: Let's keep it for later then ;)

src/spikeinterface/sortingcomponents/peak_detection.py
```diff
@@ -105,6 +109,220 @@ def detect_peaks(recording, method='by_channel', pipeline_nodes=None,
...
+            )
+
+            # update traces_chunk to run the algorithm again and store local_peaks
+            traces_chunk = traces_chunk_minus_waveforms
+
+            all_waveforms.append(waveforms)
+
+        all_local_peaks = np.concatenate(local_peaks_list, axis=0)
+        all_waveforms = np.concatenate(all_waveforms, axis=0) if len(all_waveforms) != 0 else np.empty((0, 0, 0))
+
```
samuelgarcia, line 229: Maybe we should order peak by there sample_index and apply this order also to waveforms.

src/spikeinterface/sortingcomponents/peak_detection.py
```diff
@@ -105,6 +109,220 @@ def detect_peaks(recording, method='by_channel', pipeline_nodes=None,
...
+        return (all_local_peaks,)
```
samuelgarcia, line 230: We don't return waveforms ? I thought it  was the plan

src/spikeinterface/sortingcomponents/peak_detection.py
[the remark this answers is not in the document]
h-mayorquin: Done in the last commit.

src/spikeinterface/sortingcomponents/peak_detection.py
h-mayorquin: This is going to have the same order as the base method (i.e. `self.peak_detector_node`). Why we would order this if the base method is not ordered? Should all the methods be ordered? How are you thinking about?

samuelgarcia: in this case because this is iterative you can find peak on the second that are earlier that the last one.

h-mayorquin: Good point.

h-mayorquin: Done in the last commit. I added a test specifically for this and then changed the code for passing it.

alejoe91:
@h-mayorquin here we are only missing being able to pass a list of `thresholds` (with `len(thresholds) == num_iter`) that are applied at each iteration, right?

src/spikeinterface/sortingcomponents/peak_detection.py
```diff
@@ -105,6 +109,224 @@ def detect_peaks(recording, method='by_channel', pipeline_nodes=None,
...
+        max_margin : int
+            The maximum margin for the traces.
+
+        Returns
+        -------
+        tuple of ndarray
+            A tuple containing a single ndarray with the detected peaks.
+        """
+
+        traces_chunk = traces_chunk.astype("float32")
```
samuelgarcia, line 191: To avoid creating a new array at each iteration I think we could make a copy of this.
And this make a substraction inplace in the loop. No ?

src/spikeinterface/sortingcomponents/peak_detection.py
```diff
@@ -105,6 +109,224 @@ def detect_peaks(recording, method='by_channel', pipeline_nodes=None,
...
+                peaks=local_peaks, 
+                waveforms=waveforms
+            )
+            traces_chunk_minus_waveforms, _ = self.substract_waveforms_from_traces(
+                traces_chunk, local_peaks, 
+                denoised_waveforms
+            )
+
+            # update traces_chunk to run the algorithm again and store local_peaks
+            traces_chunk = traces_chunk_minus_waveforms
```
samuelgarcia, line 223: Here we could do a inplace substraction instead of creating a new array and changing the ref at each iteration. No ?

h-mayorquin:
@samuelgarcia @alejoe91 I have implemented your feedback in the last commits.

h-mayorquin:
I also added a trick so even with fixtures you can run tests with the main approach that you guys like to use!

alejoe91 (approved):

h-mayorquin:
I think this is a great improvement in terms of both readability and performance, @samuelgarcia !


Thanks for the improvement

samuelgarcia:
@h-mayorquin  :I am confused this is too much compliments.

@alejoe91 : This OK for merging on my side

alejoe91 (approved):

PR #1567, opened 2023-04-20
Extend load_extractor to load zarr and fix zarr annotations

alejoe91:
With this PR one can do:
```
recording = si.load_extractor("recording.zarr")
```

Plus, when using the `save(format="zarr")` function the output is reloaded with `read_zarr()` function, so that it gets all properties and annotations right.

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -543,22 +543,26 @@ class BaseExtractor:
             # case from a folder after a calling extractor.save(...)
             folder = file_path
             file = None
-            
-            # the is spikeinterface<=0.94.0
-            # a folder came with 'cached.sjon'
-            for dump_ext in ('json', 'pkl', 'pickle'):
-                f = folder / f'cached.{dump_ext}'
+
+            if folder.suffix == ".zarr":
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/base.py
```diff
@@ -543,22 +543,26 @@ def load(file_path, base_folder=None):
...
+
+                # spikeinterface>=0.95.0
+                f = folder / f'si_folder.json'
                 if f.is_file():
                     file = f
-            
-            # spikeinterface>=0.95.0
-            f = folder / f'si_folder.json'
-            if f.is_file():
-                file = f
```
alejoe91, line 557: Do something like:

```
if (folder / f'si_folder.json').is_file():
    from .binaryfolder import read_binary_folder
    extractor = read_binary_folder(folder)
else:
    # back-compatibility
```

alejoe91: this is actually needed because a folder can be a `BinaryRecording`, `NpzSorting`, or `NumpySnippets`

alejoe91:
@samuelgarcia good for me

samuelgarcia (approved):

PR #1572, opened 2023-04-24
Fix torch detector for case when non peaks are found

h-mayorquin:
I am hunting a bug in torch but this should be changed anyway so I am pushing it now. This fixes a bug where Torch returns numpy arrays (it should be tensors) when no peaks are found at any stage of the peak detection pipeline. It also, adds logic to the `PeakDetectorWrapper` to handle this issue.


h-mayorquin:
This needs #1576 to work.

samuelgarcia (approved):

PR #1574, opened 2023-04-24
Sign consistency test in peak detection

h-mayorquin:
This tests indicates the error for peak detection methods that use torch. Somehow when peak_sign is equal both for this method torch comes empty handed. That is, no peaks are detected and hence the tests fail.

The tests is simple. Itcalculates peaks with positive sign, peaks with negative sign and then all peaks. It asserts that total should be the sume of positive and negative.

### DIFF
--- a/src/spikeinterface/sortingcomponents/peak_detection.py
+++ b/src/spikeinterface/sortingcomponents/peak_detection.py
@@ -463,57 +463,54 @@ if HAVE_TORCH:
         dtype = torch.float32
         empty_return_value = (torch.tensor([], dtype=dtype) , torch.tensor([], dtype=dtype))
 
-        # -- torch argrelmin
+        
+        # The function uses maxpooling to look for maximum
         if peak_sign == "neg":
-            neg_traces = torch.as_tensor(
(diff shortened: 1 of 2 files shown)


src/spikeinterface/sortingcomponents/peak_detection.py
```diff
@@ -463,57 +463,54 @@ def _torch_detect_peaks(traces, peak_sign, abs_thresholds,
...
-            )
+            traces = -traces
         elif peak_sign == "pos":
-            neg_traces = torch.as_tensor(
-                traces, device=device, dtype=torch.float
-            )
+            traces = traces
         elif peak_sign == "both":
-            neg_traces = torch.as_tensor(
-                -np.abs(traces), device=device, dtype=torch.float
```
h-mayorquin, line 477: This was the source of the error.

h-mayorquin:
The tests are passing for the original case but I discovered that sometimes the test produces duplicated peaks. I think this is a margin bug but I will wait to have a more extensive discussion for this.

alejoe91 (approved):

PR #1577, opened 2023-04-25
Add development section to docs

h-mayorquin:
This PR is two big things:
1) First, I changed the sections contribute to development to centralized all things related to contributors and development. This is the name standard that is used in Numpy and Scikit-learn so I think it is just better to piggyback on that for naming.
2) I extended that section with some new information:
* How to run tests locally,
* How to install datalad.
* How to build the documentation.
* Stylistics conventions.

Plus some minor fixes like old references to master.

Let me know what do you think.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -50,7 +50,7 @@ for folder in folders:
 
 # -- Project information -----------------------------------------------------
 
-project = 'spikeinterface'
+project = 'Spikeinterface'
 copyright = '2022, Alessio Paolo Buccino, Samuel Garcia, Cole Hurwitz, Jeremy Magland, Matthias Hennig'
 author = 'Alessio Paolo Buccino, Samuel Garcia, Cole Hurwitz, Jeremy Magland, Matthias Hennig'
 
--- a/doc/development/development.rst
+++ b/doc/development/development.rst
@@ -1,15 +1,141 @@
-Contribute
+Development
 ==========
 
-To contribute to SpikeInterface, a user/developer can help us integrate in a new recorded file
-format, a new sorted file format, or a new spike sorting algorithm or more!
+How to contribute
+-----------------
 
-In order to make a contribution, you first need to **fork** the SpikeInterface repository.
-Then you can clone from your fork, make a new branch (e.g. :code:`git checkout -b my-contribution`),
-make changes to the code, commit and push to your fork.
-Next, you can open a pull request (PR) from the "Pull Requests" tab of your fork to :code:`spikeinterface/master`.
+To contribute to SpikeInterface, a user/developer there are many tasks that you could help us with:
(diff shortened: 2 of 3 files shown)


doc/conf.py
```diff
@@ -50,7 +50,7 @@
 
 # -- Project information -----------------------------------------------------
 
-project = 'spikeinterface'
+project = 'Spikeinterface'
```
samuelgarcia, line 53: I think I prefer lower case.

doc/development/development.rst
```diff
@@ -1,15 +1,142 @@
...
+
+    git config --global filter.annex.process "git-annex filter-process"
+
+Stylistic conventions
+---------------------
+
+The following are some styling conventions that we follow in SpikeInterface:
+
+#. Avoid using abreviations in variable names (e.g., use :code:`recording` instead of :code:`rec`). It is specially important to avoid single letter variables.
+#. Use index as singular and indices for plural following Numpy. Avoid id, idx, ids, etc.
```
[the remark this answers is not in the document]
h-mayorquin: Yes, that is one good reason to have index not being ids is that it creates ambiguity with ids as in channel_ids. I will add, `id and ids` should be used exclusively for ids as in `channel_ids`.

doc/conf.py
h-mayorquin: I did it so it will be capitalized in the docs in the corner:
image

So, it is consistent with the main title.

DradeAW:
Mmhm, I'm not sure I'm the best man to help you :/

I started using SpikeInterface before the paper was published (like march-april 2020) ; the documentation was pretty incomplete back then. So I'm used to searching stuff in the source code directly rather than the documentation ^^'

I have to admit that I never read it :sweat_smile:

pauladkisson:
I think it would be nice to have some specifications for unit tests and documentation.  For example when I was writing tests for wobble, I just had to guess what was expected and then get feedback/correction during code review (ex. seeding all random operations, explaining the wobble algorithm step-by-step, etc.).  If we define a consistent standard up front, I think it will improve the quality of tests/docstrings and decrease the work for reviewers.  This would be especially helpful for more junior devs and people who are new to spikeinterface like myself.

DradeAW:
> or example when I was writing tests for wobble, I just had to guess what was expected

Personally I just copy-pasted from other chunks of code,
Isn't that the very essence of programming? ^^

alejoe91 (approved):

h-mayorquin:
@pauladkisson 
Thanks for the feedback. I think having a standard of what kind of testing we expect would be great to have. That said, I don't think we agree on that standard so pending that discussion I will add this to my to-do list. 

Thank you both for you feedback @DradeAW @pauladkisson.

This is the first iteration so I will be revisiting this.

PR #1580, opened 2023-04-26
Change v_peak name and thresholder behavior

yger:
After discussing with Samuel, we've decided to change the name v_peak to something more explicit (peak_voltage). In addition, I've modified an util function for GridLocalization to remove a deprecation warning, and I've also modified the WaveformThresholder object to not assume data are whitened. 

@samuelgarcia @alejoe91 @h-mayorquin Regarding the noise_levels, this would really be helpful if they could be attached to the recording object, in order to avoid duplicated computations all over the places (in sparsities, in thresholder, in matching engines, ...). I know that know they can be attached to waveform_extractor, and this is a good step, but to me these noise levels are really at the recording level

### DIFF
--- a/src/spikeinterface/sortingcomponents/waveforms/waveform_thresholder.py
+++ b/src/spikeinterface/sortingcomponents/waveforms/waveform_thresholder.py
@@ -6,7 +6,7 @@ import numpy as np
 import operator
 from typing import Literal
 
-from spikeinterface.core import BaseRecording
+from spikeinterface.core import BaseRecording, get_noise_levels
 from spikeinterface.sortingcomponents.peak_pipeline import (
     PipelineNode,
     WaveformsNode,
@@ -21,7 +21,7 @@ class WaveformThresholder(WaveformsNode):
     This node allows you to perform adaptive masking by setting channels to 0
     that have a given feature below a certain threshold. The available features
     to consider are peak-to-peak amplitude ('ptp'), mean amplitude ('mean'),
-    energy ('energy'), and voltage peak ('v_peak').
+    energy ('energy'), and peak voltage ('peak_voltage').
 
     Parameters
     ----------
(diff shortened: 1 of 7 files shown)


alejoe91:
@yger maybe the `get_noise_levels` could automatically set the `noise_level` property to the recording and return that in case it's already computed? We could also add a `force_recompute` arg to allow one to recompute it (e.g. with different params)

samuelgarcia:
Yep I am agree with you Alessio.

alejoe91:
Different PR, should be relatively small ;)

h-mayorquin:
How are they attached?

Right now they could be added as a property of the recoder and then have any method that requires check for a property call noise levels:

Pseudocode:
```
noise_levels = get_noise_levels(recording)
recording.set_property(noise_levels)

def method_that_requieres_noise():
      noise_levels = recording.get_property(key="noise_levels")
```

How they should look like?

```
get_noise_levels(recording)  # Set the noise level as a property as a side effect
def method_that_requieres_noise():
       noise_levels = recording.noise_levels
```

Am I getting this right? 

It would be good if this were to be discussed in a different issue and not in a PR.

src/spikeinterface/sortingcomponents/waveforms/waveform_thresholder.py
```diff
@@ -31,19 +31,25 @@ class WaveformThresholder(WaveformsNode):
         Whether to return output from this node (default True).
     parents: list of PipelineNodes, optional
         The parent nodes of this node (default None).
-    feature: {'ptp', 'mean', 'energy', 'v_peak'}, optional
+    feature: {'ptp', 'mean', 'energy', 'peak_voltage'}, optional
         The feature to be considered for thresholding (default 'ptp').
```
alejoe91, line 35:
```suggestion
    feature: {'ptp', 'mean', 'energy', 'peak_voltage'}, optional
        The feature to be considered for thresholding (default 'ptp'). Features are normalized with the channel noise levels.
```

alejoe91 (commented):
Just one suggestion for the docstring, the resl looks good!

alejoe91 (approved):

Issue #1413, opened 2023-03-10
GeneratorRecording outline


h-mayorquin:
Hi, so I am writing here the outline of the features that the GeneratorRecording should have as discussed on today's meeting. The purpose of this issue is to discuss, describe and agree what we want ant to outline the features that we will be adding to it.

The general idea, as described in #1391, is to have a generator whose memory footpring is fully controlled. This is very useful for testing purposes and specifically decoupling reading from writing bugs on high-memory loads. Plus, if we add more functionality, as discussed below, maybe this function can be used in more general scenarios.

Ther are at least two important properties:
1. Consistency with signature:  `get_traces(start_frame, end_frame, segment_index)` output is totally determined by the parameters on the signature.
2. Array-like behavior:  `get_traces(start_frame, end_frame) == get_traces(start_frame,end_frame + 10)[start_frame:end_frame, :]`

Notablye that the naive solution of:
```python
def _random_traces(self, start_frame: int, end_frame: int) -> np.ndarray:

    unique_seed_hash = (start_frame, end_frame, self.seed)
    local_rng = np.random.default_rng(seed=unique_seed_hash)
    traces = local_rng.random(size=(end_frame - start_frame, self.num_channels), dtype=self.dtype)
    
    return traces
```
Has property 1 but not property 2.

Right now, #1391 has a solution that accomplishes these two properties to this by returning a "deterministic" function with period output that looks like this:
![image](https://user-images.githubusercontent.com/6429509/224374249-c578cc6f-e2e7-4d47-8669-3ca42dbe48a1.png)

The general shape is controlled by a random seed and it produces channels with different period, phase and amplitude. I was thinking in using to test `detect_peaks` scenarios when I wrote that. 

That said, we would like to have more functionality to this function in the form of modes so we can support a wider variety of cases. The following outline describes them.

- [ ] Add capabilities to generate pure noise which also has property 2 above.
- [ ] Add capabilites to combine the "deterministic" output of the current solution with noise on top.
- [ ] Add simply sinusoidal behavior for LFP like scenarios.

h-mayorquin:
For the pure noise scenario the idea is to generate a basic block of purse noise `basic_trace = np.random.rand(some_small_number_of_frame, number_of_channels)` and then use the frames with a modulus operation as a periodic tile to generate a noisy trace that has properties 1 and 2 above. Something like:

`basic_trace[start_frame % size:end_frame % size]` 

[...]

PR #1581, opened 2023-04-26
Add pure noise generation to generator recording

h-mayorquin:
Following on #1413. This adds the capabilities of generating pure noise with the recording above.

I also added tests for the two modalities to assure that the generator behaves as an array in the expected way:
* Running get_traces twice gets the sames results.
* Running get_traces and then slicing is the same as running get_traces in the smaller slice.

### DIFF
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -387,6 +387,9 @@ def synthetize_spike_train_bad_isi(duration, baseline_rate, num_violations, viol
 from typing import Union, Optional, List, Literal
 
 class GeneratorRecording(BaseRecording):
+    
+    available_modes = ['pure_noise', 'random_peaks']
+    
     def __init__(
         self,
         durations: List[float],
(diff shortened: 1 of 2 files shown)


DradeAW:
Could we also add `mode="white_noise"`?
This is easily generated by each point following a Gaussian distribution (rather than a uniform distribution like the `"pure_noise"` method).

I think it could be useful and is rather easy to implement :)

h-mayorquin:
@DradeAW could you add the use case that you are envisioning to issue #1413 so it is not lost?

Add your reasoning behind of why do you think this would be useful to have.

src/spikeinterface/core/generate.py
```diff
@@ -432,23 +444,68 @@ def __init__(
...
+            self.amplitudes = self.rng.normal(loc=70, scale=10.0, size=self.num_channels)  # Amplitudes of 70 +- 10
+            self.amplitudes *= self.rng.choice([-1, 1], size=self.num_channels)    # Both negative and positive peaks
+        
+        elif self.mode == "pure_noise":
+            self.traces_generator = self._random_traces_generator
+
+            # Configuration of mode
+            noise_size_ms = 50.0
+            noise_size_frames = int(noise_size_ms * self.sampling_frequency / 1000)
+            self.basic_noise_block = self.rng.random(size=(noise_size_frames, self.num_channels))
```
alejoe91, line 507: @h-mayorquin I discussed with @samuelgarcia and we agree it shoudl be `normal` instead. Note that the way it is it returns samples between 0 and 1!

h-mayorquin: Yes, makes sense. I was trying to imitate the current `generate_recording`. Great.

h-mayorquin: Done. Now the default mode is to generate `white_noise`.

alejoe91 (approved):

PR #1583, opened 2023-04-26
Caching noise levels

yger:
As we just discussed, here is a proposal to cache noise levels in the recording. I think this makes a lot of sense, because otherwise, we might face some variability in these levels is estimated several times during a pipeline. However, not familiar with the whole architecture of SI. If new recordings are made from a given recordings, I am assuming this property is not inherited. Is it correct?


src/spikeinterface/core/recording_tools.py
```diff
@@ -149,6 +149,38 @@ def get_closest_channels(recording, channel_ids=None, num_channels=None):
...
+
+    Internally it samples some chunk across segment.
+    And then, it use MAD estimator (more robust than STD)
+
+    """
+    
+    if 'noise_levels' in recording._properties and not force_recompute:
+        noise_levels = recording.get_property(key="noise_levels")
+        return noise_levels
+    else:
```
h-mayorquin, line 166: The else nesting here is unecessary after the early return.

src/spikeinterface/core/recording_tools.py
```diff
@@ -149,6 +149,38 @@ def get_closest_channels(recording, channel_ids=None, num_channels=None):
...
+    """
+    Estimate noise for each channel using MAD methods.
+    You can use standard deviation with `method='std'`
+
+    Internally it samples some chunk across segment.
+    And then, it use MAD estimator (more robust than STD)
+
+    """
+    
+    if 'noise_levels' in recording._properties and not force_recompute:
```
alejoe91, line 163:
```suggestion
    if 'noise_levels' in recording.get_property_keys() and not force_recompute:
```

src/spikeinterface/core/recording_tools.py
```diff
@@ -149,6 +149,38 @@ def get_closest_channels(recording, channel_ids=None, num_channels=None):
...
+        
+        if method == "mad":
+            med = np.median(random_chunks, axis=0, keepdims=True)
+            # hard-coded so that core doesn't depend on scipy
+            noise_levels = (
+                np.median(np.abs(random_chunks - med), axis=0) / 0.6744897501960817
+            )
+        elif method == "std":
+            noise_levels = np.std(random_chunks, axis=0)
+        recording.set_property('noise_levels', noise_levels)
```
alejoe91, line 179: I would call the property `noise_level`, no s

alejoe91 (changes requested):
Thanks Pierre, a couple of comments!

src/spikeinterface/core/tests/test_recording_tools.py
```diff
@@ -24,14 +24,25 @@ def test_get_closest_channels():
...
     rec.set_channel_gains(0.1)
     rec.set_channel_offsets(0)
-    noise_levels = get_noise_levels(rec, return_scaled=True)
+    noise_levels = get_noise_levels(rec, return_scaled=True, force_recompute=True)
 
     noise_levels = get_noise_levels(rec, return_scaled=True, method="std")
 
+    # Generate a recording following a gaussian distribution to check the result of get_noise.
+    std = 6.0
+    traces = np.random.normal(loc=10.0, scale=std, size=(1_000_000, 2))
```
h-mayorquin, line 38: Test that use random numbers should be deterministic.

h-mayorquin:
I think it will be better to have a method of `BaseRecording` that calls `get_noise_levels` instead of introducing a new `get_noise_levels` implicit side effect. Besides it being bad practice to have side effects in the abstract, maybe people who were previously using this function were not expecting their recording object to be modified (by name this is a `getter` not a `setter`).

I agree with @yger that when available, the waveform extractor should get its noise level from the recording on it. Not clear what to do in recordingless instances though.

alejoe91:
> As we just discussed, here is a proposal to cache noise levels in the recording. I think this makes a lot of sense, because otherwise, we might face some variability in these levels is estimated several times during a pipeline. However, not familiar with the whole architecture of SI. If new recordings are made from a given recordings, I am assuming this property is not inherited. Is it correct?

Yes that is correct. We can do something like this: we usually use the [`copy_metadata`](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/base.py#L244) to propagate properties and annotations. We could add a `skip_properties` there which is `noise_level` for the `basepreprocessor`, so that pre-processors won't copy that property over

src/spikeinterface/core/recording_tools.py
yger: Done

src/spikeinterface/core/recording_tools.py
yger: indeed

src/spikeinterface/core/tests/test_recording_tools.py
yger: done

h-mayorquin:
If for any reason you would not like to add another method to BaseRecording you can create another method

def set_noise_levels(recording: 'BaseRecording', return_scaled: bool = True, method: Literal['mad', 'std'] = "mad", 
    force_recompute: bool=False, **random_chunk_kwargs):

    if 'noise_level' in recording._properties and not force_recompute:
        noise_levels = recording.get_property(key="noise_level")
    else:
        noise_levels = get_noise_levels(args)
        recording.set_property('noise_level', noise_levels)

samuelgarcia:
We also need to have 2 properties for this noise_levels_raw and nose_levels_scaled.

src/spikeinterface/core/recording_tools.py
```diff
@@ -147,6 +126,43 @@ def get_closest_channels(recording, channel_ids=None, num_channels=None):
...
+    And then, it use MAD estimator (more robust than STD)
+
+    """
+
+    if return_scaled:
+        key = "noise_level_scaled"
+    else:
+        key = "noise_level_raw"
+
+    if key in recording._properties and not force_recompute:
```
alejoe91, line 150:
```suggestion
    if key in recording.get_property_keys() and not force_recompute:
```

alejoe91:
@samuelgarcia @h-mayorquin implemented the skipping mechanism for the preprocessors (for the `noise_level_*` properties). Let me know what you think!

alejoe91:
@h-mayorquin actually I would delay the creation of wrapper functions for later, so we can have a consistent behavior for all `recording_tools` functions.

alejoe91 (approved):

src/spikeinterface/preprocessing/basepreprocessor.py
```diff
@@ -20,7 +20,9 @@ def __init__(self, recording, sampling_frequency=None, channel_ids=None, dtype=N
             dtype = recording.get_dtype()
 
         BaseRecording.__init__(self, sampling_frequency, _channel_ids, dtype)
-        recording.copy_metadata(self, only_main=False, ids=channel_ids)
+        recording.copy_metadata(
+            self, only_main=False, ids=channel_ids, skip_properties=["noise_level_scaled", "noise_level_raw"]
```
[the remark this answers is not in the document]
alejoe91: Note that this is not the case for all "transformations". For example, `frame/channel_slice` will inherit the original noise levels, but we need to enforce skipping for preprocessing (at least)

h-mayorquin:
@alejoe91 makes sense! Let's also discuss that in a general issue!

src/spikeinterface/preprocessing/basepreprocessor.py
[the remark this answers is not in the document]
samuelgarcia: I think this is dangerous. We should reset this field for any tranformation included channel slice (even if in that precise case it is usefull to propagate).
Making the assumption that transform in the core will not affect the noise level and the one in preprocessing can affect is dangerous. I would reset theses 2 properties in all cases.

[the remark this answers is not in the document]
samuelgarcia: For instance AppendSegmentRecording have rec0.copy_metadata(self).
This would be bad to keep this noise level in that case.

src/spikeinterface/core/channelsaggregationrecording.py
```diff
@@ -13,6 +13,8 @@ class ChannelsAggregationRecording(BaseRecording):
 
     """
 
+    _skip_properties = []
+
```
samuelgarcia, line 17: This is too dangerous!
Imagine we happen some other propertis in `Baserecording._skip_properties` ?
We will totally forget about this.
I propose to remove this.

src/spikeinterface/core/channelslice.py
```diff
@@ -14,6 +14,8 @@ class ChannelSliceRecording(BaseRecording):
 
     """
 
+    _skip_properties = []
```
samuelgarcia, line 17: SAme comments here.

alejoe91: I think it's ok! For these recordings the noise levels are exactly the same

src/spikeinterface/core/channelsaggregationrecording.py
alejoe91: I think it's ok! For these recordings the noise levels are exactly the same

src/spikeinterface/core/baserecording.py
```diff
@@ -25,6 +25,8 @@ class BaseRecording(BaseRecordingSnippets):
     _main_properties = ["group", "location", "gain_to_uV", "offset_to_uV"]
     _main_features = []  # recording do not handle features
 
+    _skip_properties = ["noise_level_raw", "noise_level_scaled"]
```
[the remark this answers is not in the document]
alejoe91: They are handled at base

src/spikeinterface/core/channelsaggregationrecording.py
samuelgarcia: This is not my point. Imagine we have more properties to be protected than theses two one.
What will will do is to put then in RecordingBase and without thinking to check this everywhere.
A _skip_properties reset at some places is dangerous.

src/spikeinterface/core/channelslice.py
alejoe91: What about removing the noise ones in the init specifically?

samuelgarcia: It is a class attributes you cannot do this in the init.
My point is that it is better to recompute (even it is useless) the noise level in the very very very ucommon case where you 1) compute noise 2) concatenate channels (or slice) 3) compute the noise again.
Because it avoid deep deep bugs at long term, not bug in the code but bug in the processing workflow,  which very very hard to track because of this kind of small details

alejoe91: done

src/spikeinterface/core/channelsaggregationrecording.py
alejoe91: ok, done

samuelgarcia (approved):

PR #1587, opened 2023-04-27
Adding percentile as an option for grid localization

yger:
After discussing with Sam, and while playing with benchmarks for drift correction, looks like this could be useful to only keep the best scalar products while estimating the position

### DIFF
--- a/src/spikeinterface/sortingcomponents/peak_localization.py
+++ b/src/spikeinterface/sortingcomponents/peak_localization.py
@@ -289,17 +289,22 @@ class LocalizeGridConvolution(PipelineNode):
         The margin for the grid of fake templates
     prototype: np.array
         Fake waveforms for the templates. If None, generated as Gaussian
+    percentile: float (default 10)
+        The percentage in [0, 100] of the best scalar products kept to 
+        estimate the position
     """
     def __init__(self, recording, return_output=True, parents=['extract_waveforms'], local_radius_um=50., upsampling_um=5,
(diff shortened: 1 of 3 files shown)


h-mayorquin:
@alejoe91 
It was a networking problem with the current run of hugging interface. Flaky CI. If this persists on other issues I will take a closer look.

src/spikeinterface/sortingcomponents/peak_localization.py
```diff
@@ -289,17 +289,22 @@ class LocalizeGridConvolution(PipelineNode):
...
+        sigma_um=np.linspace(10, 50., 5), sigma_ms=0.25, margin_um=50., prototype=None, percentile=10):
         PipelineNode.__init__(self, recording, return_output=return_output, parents=parents)
         
         self.local_radius_um = local_radius_um
         self.sigma_um = sigma_um
         self.margin_um = margin_um
         self.upsampling_um = upsampling_um
-        contact_locations = recording.get_channel_locations()
+        self.percentile = 100 - percentile
+        assert 0 <= self.percentile <= 100, "Percentile should be in [0, 100]"
```
alejoe91, line 305: Make sure the mask keeps at least some dot products

yger: I am not worried about that and/or I do not get the point. We are generated hundreds of templates, and thus even a percentile of 1 will ensure enough non_zeros dot products

samuelgarcia:
This is OK for me.

PR #1589, opened 2023-04-27
Match IBL highpass spatial filtering more closely

cwindolf:
Thanks to @chris-langfield and @oliche for finding this fix! https://github.com/int-brain-lab/ibl-neuropixel/issues/8

This PR matches the parameters of neurodsp, except for `n_channel_pad`, which I wasn't sure is generic across probes. But, I added `n_channel_pad=60` in that section of the doc where the spikeinterface version of IBL's destriping pipeline is described.


cwindolf:
Ah, not sure why the numba stuff is in this PR... maybe because I was working off master

samuelgarcia:
THanks Charlie. DO you now have the same results in SI than in ibl framework ?

cwindolf:
Not exactly yet (or very close -- I'd expect some random difference based on what chunks were used to compute stats, but I'm seeing a larger difference than I'd expect from that effect). I'll dig into the rest of the functions and see what's up and keep this PR rolling.

JoeZiminski:
Hi @cwindolf hope things are good, thanks for this PR I helped port this function from IBL so very interested in ensuring a good match. Currently the spatial filter is tested against IBL [here](https://github.com/SpikeInterface/spikeinterface/blob/72b0835cd855cb3fac165a362646a7312c0c73f3/src/spikeinterface/preprocessing/tests/test_highpass_spatial_filter.py#L81) and a note that there are some minor differences in the [scaling on load](https://github.com/int-brain-lab/ibl-neuropixel/blob/d39672537c96f7e205704eae5d317abbef917a00/src/neuropixel.py#L54) that can cause some differences in output. One way to check this is not influencing is to load data in spikeinterface, use `get_traces(return_scaled=True)` and convert the units to feed this array into the IBL routines. Happy to help take a further look next week!

alejoe91:
Thanks @cwindolf  and @chris-langfield!

I vote for making 60 the default, since this will be mainly applied to NP data. A better option moving forward will probably by to indicate the padding in um, but let's keep it for a follow up!

src/spikeinterface/preprocessing/highpass_spatial_filter.py
```diff
@@ -60,8 +60,8 @@ class HighpassSpatialFilterRecording(BasePreprocessor):
     """
     name = 'highpass_spatial_filter'
 
-    def __init__(self, recording, n_channel_pad=None, n_channel_taper=5, direction="y",
-                 apply_agc=True, agc_window_length_s=0.01, highpass_butter_order=3,
+    def __init__(self, recording, n_channel_pad=None, n_channel_taper=0, direction="y",
```
alejoe91, line 63:
```suggestion
    def __init__(self, recording, n_channel_pad=60, n_channel_taper=0, direction="y",
```

src/spikeinterface/preprocessing/highpass_spatial_filter.py
```diff
@@ -27,8 +27,8 @@ class HighpassSpatialFilterRecording(BasePreprocessor):
     recording : BaseRecording
         The parent recording
     n_channel_pad : int
-        Number of channels to pad prior to filtering. 
-        Channels are padded with mirroring. 
+        Number of channels to pad prior to filtering.
+        Channels are padded with mirroring.
         If None, no padding is applied, by default 5
```
alejoe91, line 32:
```suggestion
        If None, no padding is applied, by default 60
```

alejoe91 (approved):

alejoe91:
> Hi @cwindolf hope things are good, thanks for this PR I helped port this function from IBL so very interested in ensuring a good match. Currently the spatial filter is tested against IBL [here](https://github.com/SpikeInterface/spikeinterface/blob/72b0835cd855cb3fac165a362646a7312c0c73f3/src/spikeinterface/preprocessing/tests/test_highpass_spatial_filter.py#L81) and I see a comment that there are some minor differences in the scaling on load that can cause some differences in output (I am 99% sure [this](https://github.com/int-brain-lab/ibl-neuropixel/blob/d39672537c96f7e205704eae5d317abbef917a00/src/neuropixel.py#L54) the scaling I was referring too, apologies if it is not this, my comment was not verbose enough!) .
> 
> One way to check this is not influencing is to load data in spikeinterface, use `get_traces(return_scaled=True)` and convert the units for IBL, then feed this array into the IBL routines. Happy to help take a further look next week!

@JoeZiminski note that those tests are not run on GH. I agree that the default values should match the default IBL implementation, and apparently this was not the case for the padding. We also got weird results in the past at the borders and @oliche confirmed that the default at IBL was 60. So let's merge the new default!

PR #1590, opened 2023-04-27
Remove unecessary ensure_n_jobs

h-mayorquin:
Removing some code that is not useful and asking some questions.

### DIFF
--- a/src/spikeinterface/core/job_tools.py
+++ b/src/spikeinterface/core/job_tools.py
@@ -223,6 +223,9 @@ def ensure_chunk_size(recording, total_memory=None, chunk_size=None, chunk_memor
     else:
         if n_jobs == 1:
             # not chunk computing
+            #TODO Discuss, Sam, is this something that we want to do?
+            # Even in single process mode, we should chunk the data to avoid loading the whole thing into memory I feel
+            # Am I wrong?
             chunk_size = None
         else:
             raise ValueError('For n_jobs >1 you must specify total_memory or chunk_size or chunk_memory')
@@ -330,10 +333,6 @@ class ChunkRecordingExecutor:
             returns = []
         else:
             returns = None
-
-        import sys
-        if self.n_jobs != 1 and not (sys.version_info >= (3, 8)):
-            self.n_jobs = 1
 
         if self.n_jobs == 1:
             if self.progress_bar:
--- a/src/spikeinterface/core/job_tools.py
+++ b/src/spikeinterface/core/job_tools.py
@@ -221,7 +221,14 @@ def ensure_chunk_size(recording, total_memory=None, chunk_size=None, chunk_memor
         else:
             raise ValueError('chunk_duration must be str or float') 
     else:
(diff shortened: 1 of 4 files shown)


src/spikeinterface/core/job_tools.py
```diff
@@ -223,6 +223,9 @@ def ensure_chunk_size(recording, total_memory=None, chunk_size=None, chunk_memor
     else:
         if n_jobs == 1:
             # not chunk computing
+            #TODO Discuss, Sam, is this something that we want to do?
+            # Even in single process mode, we should chunk the data to avoid loading the whole thing into memory I feel
+            # Am I wrong?
             chunk_size = None
```
h-mayorquin, line 229: @samuelgarcia I realize that when n_jobs=1 set chunk_size = None
Whicn inside the function `ensure_chunk_size` ends up causing that the whole traces are extracted per segment:

```
def divide_segment_into_chunks(num_frames, chunk_size):
    if chunk_size is None:
        chunks = [(0, num_frames)]
```

Do we want this behavior? My feelign is that it shouldn't as indicated in the comment ot the right
but ...
If so, I really think this should not happen in such an obscure way (hidden deep in two functions). 
Let me know what do you think.

src/spikeinterface/core/job_tools.py
```diff
@@ -331,10 +334,6 @@ def run(self):
         else:
             returns = None
 
-        import sys
```
h-mayorquin, line 334: We don't support Python 3.7 anymore, this can go.

src/spikeinterface/core/job_tools.py
samuelgarcia: The on purpose chunk_size=None have the meaning of do it with no chunking.
Forcing this when n_job=1 is bad idea! I do not remember why we have this here I guess it come from the previous framework logic to keep some compatibility.

samuelgarcia: Fortunatly this code zone should be reach any more with the recent get_global_job_kwargs() mechanism.

h-mayorquin: OK. I am removing this.

I am not sure this is working everywhere but we will see. For that to happen the function `fix_job_kwargs` should have been called which in turn  fetches the job_kwargs from globals and updates if neede. This is done in most places but I am not sure I did it when writing code that takes job_kwargs....

I find this behavior brittle. I think it would be way easier to have that as default values of the ChunkRecordingExecutor...

h-mayorquin:
This issue is ready to go. It is just removing unused code. I mainly used to get the attention of @samuelgarcia about the point in the discussion above.

src/spikeinterface/core/job_tools.py
alejoe91: Suggestion:
This last case could be covered by:
- `chunk_size=a big number`
- `chunk_duration="full"`

And we should check that at least one of `chunk_size`, `chunk_duration`, `chunk_memory` or `total_memory` is not None.

samuelgarcia (approved):

PR #1593, opened 2023-04-28
fix NWBFile type hints

JoeZiminski:
A [recent commit](https://github.com/SpikeInterface/spikeinterface/commit/72b0835cd855cb3fac165a362646a7312c0c73f3) nicely added type hints but in the case `NWBFile` and `ElectricalSeries` cannot be imported the script crashes. This PR just wraps these type hints in strings, another approach is to use [this kind of syntax](https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports), but I don't think this will work in this case because the module cannot be imported at all. A tricky one to detect because only is exposed if `pynwb.ecephys` is not installed

PR #1597, opened 2023-04-28
Use annotations from future (PEP563) to solve type hint references

### LINKED ISSUE #1596: `import spikeinterface.extractors` fails if NWB not installed
Hey,

The typing here creates a crash if NWB is not installed https://github.com/SpikeInterface/spikeinterface/blob/72b0835cd855cb3fac165a362646a7312c0c73f3/src/spikeinterface/extractors/nwbextractors.py#L40

```
$ pytest
ImportError while loading conftest '/home/aurelien/Documents/lussac2/tests/conftest.py'.
tests/conftest.py:6: in <module>
    from lussac.core import LussacData, LussacPipeline, MonoSortingData, MultiSortingsData
src/lussac/core/__init__.py:1: in <module>
    from .lussac_data import LussacData, MonoSortingData, MultiSortingsData
src/lussac/core/lussac_data.py:12: in <module>
../../dev/spikeinterface/spikeinterface/src/spikeinterface/extractors/__init__.py:1: in <module>
    from .extractorlist import *
../../dev/spikeinterface/spikeinterface/src/spikeinterface/extractors/extractorlist.py:16: in <module>
    from .nwbextractors import (NwbRecordingExtractor, NwbSortingExtractor,
../../dev/spikeinterface/spikeinterface/src/spikeinterface/extractors/nwbextractors.py:40: in <module>
    def retrieve_electrical_series(nwbfile: NWBFile, electrical_series_name: Optional[str] = None) -> ElectricalSeries:
E   NameError: name 'NWBFile' is not defined
```

h-mayorquin:
This PR resolves #1596.

As described in the issue, importing extractors fails when pynwb is not installed. The problem arises from the newly added type-hints, which are currently evaluated at import time.

PR #1593 suggests a solution that employs forward references, where type hints are defined as strings to force deferred evaluation. While this approach works, we can improve upon it.

Starting with Python 3.8, we can use the following statement at the top of a module to defer the evaluation of type annotations until runtime:

```python
```
More details can be found in PEP 563: https://peps.python.org/pep-0563/

This change allows us to maintain the same notation regardless of whether the module is installed or not, simplifying code navigation.

Moreover, deferring type hints by default is expected to enhance import performance, which aligns with the long-term goals of the library.

Finally, I've added a test to the core that verifies all major modules can be imported with only a core installation, without raising errors. This test would have helped prevent the current issue. Furthermore, the test will be used to monitor import times, helping us track and reduce them in the future.

### DIFF
--- a/.github/profile_and_test_imports.py
+++ b/.github/profile_and_test_imports.py
@@ -1,0 +1,45 @@
+import subprocess
+import numpy as np
+
+import_statement_list = [
+    "import spikeinterface",
+    "import spikeinterface.core",
+    "import spikeinterface.extractors",
+    "import spikeinterface.qualitymetrics",
+    "import spikeinterface.preprocessing",
+    "import spikeinterface.comparison",
+    "import spikeinterface.postprocessing",
+    "import spikeinterface.sortingcomponents",
+    "import spikeinterface.curation",
+]
+
+markdown_output = "## Import Profiling\n\n| Module | Time (seconds) | Standard Deviation |\n| ------ | -------------- | ------------------ |\n"
+n_samples = 10
+
+for import_statement in import_statement_list:
+    time_taken_list = []
+    for _ in range(n_samples):
+        code = f"""
+                import timeit
+                import_statement = "{import_statement}"
+                time_taken = timeit.timeit(import_statement, number=1)
+                print(time_taken)
+                """
+
+        result = subprocess.run(["python", "-c", code], capture_output=True, text=True)
+
+        if result.returncode != 0:
+            error_message  = (
+                f"Error when running {import_statement} \n"
+                f"Error in subprocess: {result.stderr.strip()}\n"
+            )
+            raise ImportError(error_message)
+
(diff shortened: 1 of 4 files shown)


DradeAW:
Love the idea of LazyLoader! :)

samuelgarcia:
Oh my god.
I want to do this since one year without finding the time to do it.
You ar emy heroe.

samuelgarcia:
We need to discuss this lines
https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/widgets/__init__.py#L1C1-L29
because the backend in widgets now needs some try import to be registered and also having the correct docs.

samuelgarcia (approved):

samuelgarcia:
This is OK for me.
about the spikeinterface.widgets that still have HAVE_MPL and HAVE_SORTINGVIEW this will chnage when refactoring the widgets api.
I merge.

samuelgarcia:
Note that #1859 should also remove try except in widget module

Issue #1536, opened 2023-04-13
argument plot_templates=True in plot_unit_waveforms_density_map() causes an error


rat-h:
```
>>> import spikeinterface.full as si
>>> from spikeinterface import widgets as spw
>>> si.__version__
'0.97.1'
>>> we  = si.WaveformExtractor.load_from_folder("TestRun-Tr2wPP/waveforms")
>>> spw.plot_unit_waveforms_density_map(we,unit_ids=[0]); show()
<spikeinterface.widgets.matplotlib.unit_waveforms_density_map.UnitWaveformDensityMapPlotter object at 0x7fdf469cf730>
>>> spw.plot_unit_waveforms_density_map(we,unit_ids=[0],plot_templates=True); show()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "~/spikes/lib/python3.10/site-packages/spikeinterface/widgets/base.py", line 116, in widget_func
    W.do_plot(W.backend, **W.backend_kwargs)
  File "~/spikes/lib/python3.10/site-packages/spikeinterface/widgets/base.py", line 52, in do_plot
    self.check_backend_kwargs(plotter, backend, **backend_kwargs)
  File "~/spikes/lib/python3.10/site-packages/spikeinterface/widgets/base.py", line 46, in check_backend_kwargs
    raise Exception(f"{k} is not a valid plot argument or backend keyword argument. "
Exception: plot_templates is not a valid plot argument or backend keyword argument. Possible backend keyword arguments for matplotlib are: ['figure', 'ax', 'axes', 'ncols', 'figsize', 'figtitle']
```

alejoe91:
@rat-h indeed `plot_templates` is not an argument of the function: https://spikeinterface.readthedocs.io/en/latest/api.html#spikeinterface.widgets.plot_unit_waveforms_density_map

Maybe you want to use `sw.plot_unit_waveforms` instead?
See https://spikeinterface.readthedocs.io/en/latest/api.html#spikeinterface.widgets.plot_unit_waveforms

rat-h:
I marked by >
```
>>> from spikeinterface import widgets as spw
>>> help(spw.plot_unit_waveforms_density_map)
Help on function plot_unit_waveforms_density_map in module spikeinterface.widgets.base:

[...]

PR #1598, opened 2023-04-28
Update Docstrings I (Widgets Module)

zm711:
I'm starting to work on updating some of the docstrings (#1536). This will take multiple passes, but for now I tried to delete parameters not accessible and add parameters that can be set. The one thing that needs to be added for widgets is  **backend_kwargs. I couldn't find a list (in docs or in functions) so that can either be added in the docstrings for all the functions or maybe it would be better just to link to a page in the RTD that lists the possible kwargs @samuelgarcia? I'm happy to add them to this PR if you'd prefer that or add them to the docs (now or in a different PR) if you have list of them somewhere.

### DIFF
--- a/src/spikeinterface/widgets/all_amplitudes_distributions.py
+++ b/src/spikeinterface/widgets/all_amplitudes_distributions.py
@@ -9,16 +9,22 @@ from ..core.waveform_extractor import WaveformExtractor
 
 class AllAmplitudesDistributionsWidget(BaseWidget):
     """
-    Plots distributions of amplitudes as violon plot for all or some units.
+    Plots distributions of amplitudes as violin plot for all or some units.
 
     Parameters
     ----------
     waveform_extractor: WaveformExtractor
         The input waveform extractor
     unit_ids: list
-        List of unit ids.
+        List of unit ids, default None
     unit_colors: None or dict
-        Dict of colors
+        Dict of colors with key: unit, value: color, default None
+    backend : None or str
+        Three possible options:
+        * 'matplotlib': uses matplotlib backend
+        * 'ipywidgets': can only be used in Jupyter notebooks/Jupyter lab
+        * 'sortingview': for web-based GUIs
+        Default is None which uses the matplotlib backend
     """
     possible_backends = {}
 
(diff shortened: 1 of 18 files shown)


src/spikeinterface/widgets/template_similarity.py
```diff
@@ -19,12 +19,18 @@ class TemplateSimilarityWidget(BaseWidget):
     display_diagonal_values : bool
         If False, the diagonal is displayed as zeros.
         If True, the similarity values (all 1s) are displayed. Default False
-    cmap : Matplotlib colormap
-        The matplotlib colormap. Default 'viridis'. (matplotlib backend)
+    cmap : str
```
h-mayorquin, line 22: I don't know how the plot_data dict works but can you pass an actual cmap? Does it only accept string?

h-mayorquin (commented):
This is amazing, thanks a lot for doing this work.

src/spikeinterface/widgets/template_similarity.py
alejoe91: Hi guys, I think it's only string!

src/spikeinterface/widgets/all_amplitudes_distributions.py
```diff
@@ -9,16 +9,22 @@
...
+        List of unit ids, default None
     unit_colors: None or dict
-        Dict of colors
+        Dict of colors with key: unit, value: color, default None
+    backend : None or str
+        Three possible options:
+        * 'matplotlib': uses matplotlib backend
+        * 'ipywidgets': can only be used in Jupyter notebooks/Jupyter lab
+        * 'sortingview': for web-based GUIs
+        Default is None which uses the matplotlib backend
```
samuelgarcia, line 27: We have a system to inject automatically more doc that is backend specific here
https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/widgets/widget_list.py#L78
Maybe we should do this there to avoid copy/paste.

zm711: You're right. I'll fix that tomorrow.

zm711:
Not sure why the tests are failing. Seems like a CI thing since I only touched documentation.

One quick additional question. In this line `radius_um` is initialized, but then never used. Wasn't sure if it is used in a different widget or was from a previous version, but I didn't include in the docstring since I can't see where it is being used.

https://github.com/SpikeInterface/spikeinterface/blob/d5a1ba439a28ed3da31775afe34dcf6c1160ff86/src/spikeinterface/widgets/unit_summary.py#L34-L37

alejoe91 (approved):

Issue #1595, opened 2023-04-28
get_waveforms() should work without sparsity


yger:
Currently, when using a sparse waveform_extractor, we have a minor discrepancies in the fact that get_all_templates() are returning the templates in the full space (setting 0 to sparse channels), while get_waveforms() is getting the data in the sparse space, i.e. returning only the channels where you have information. I think it should be good if we could have an option to make get_waveforms() return dense waveforms, even for a sparse extractor. What do you think? This would be much easier for downstreams analysis.

samuelgarcia:
yes we could add an option in get_waveforms() to force it dense. But I would be keep sparse by default to avoid memory copy.

alejoe91:
I agree! Note that the same behavior is for the `get_template()`, but for returning all templates we need a dense structure because sparsity length might not be constant across units

yger:
Ok, I'll do a PR to achieve this

PR #1599, opened 2023-04-28
Force_dense mode for waveforms and templates

yger:
This solves #1595, with the addition of a force_dense flag in both get_waveforms and get_templates.

### DIFF
--- a/src/spikeinterface/core/waveform_extractor.py
+++ b/src/spikeinterface/core/waveform_extractor.py
@@ -900,7 +900,7 @@ class WaveformExtractor:
 
         return new_we
 
-    def get_waveforms(self, unit_id, with_index=False, cache=False, lazy=True, sparsity=None):
+    def get_waveforms(self, unit_id, with_index=False, cache=False, lazy=True, sparsity=None, force_dense=False):
         """
         Return waveforms for the specified unit id.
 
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/waveform_extractor.py
```diff
@@ -959,6 +961,19 @@ def get_waveforms(self, unit_id, with_index=False, cache=False, lazy=True, spars
...
             wfs = wfs[:, :, sparsity.mask[self.sorting.id_to_index(unit_id)]]
 
+        if force_dense:
+            num_channels = self.get_num_channels()
+            dense_wfs = np.zeros((wfs.shape[0], wfs.shape[1], num_channels), dtype=np.float32)
+            unit_ind = self.sorting.id_to_index(unit_id)
+            if sparsity is not None:
+                unit_sparsity = sparsity.mask[unit_ind]
+                dense_wfs[:, :, unit_sparsity] = wfs
+                wfs = dense_wfs
```
samuelgarcia, line 971: I think we should not allow the case sparsity is None and force_dense=True.
This is non sens.
I would handle only internal sparse.
So we could have a checker `assert self.is_sparse()` when `force_dense=True`
no ?

yger: In fact, I think it could make sense. Someone may want to impose some sparsity and still, get the waveforms in a dense format

alejoe91: I agree with Pierre

samuelgarcia:
Camarade. I was agree!
This weird to have function passing sparsity and then asking to get it dense. This make the doc confusing and I do not see use case.

Issue #1384, opened 2023-03-02
External Sorters improvement outline


h-mayorquin:
We have been changing recently the way we handle and test sorters on the repo. Terminology: internal sorters are the ones build with sorting components and external sorters are basically the rest as they require external machinery. 

First, @alejoe91 created an in-code separation between internal and external sorters in . #1354
And then we are testing the external sorters with their docker images (right now only for singularity here) CI #1373

This issues it to discuss and then oultine improvements and to-do for the sorters

Here are some random issues and we can outline a to-do later.

* Discuss a policy for which version of spikeinterface the docker images should pull, see [discussion link](https://github.com/SpikeInterface/spikeinterface/pull/1373#issuecomment-1450260167).
* Right now it seems that the Tridesclous is image does not have the pakcage installed, see the latest ran of the action that was created in #1373. 
* Right now we have a dispatch for testing the sorters with singularity #1373. Maybe that should be a weekly cron job?
* Also I feel that the tests at #1373 should run for the whole matrix of OS. This is one of the main purposes of the sorters so it would be really nice to test them for all OS. 
* There should be tests that use docker instead of singularity. There is an [action](https://github.com/marketplace/actions/setup-docker) for that. I can take care of that after I finish the following sorting component round. Should be quick.
* Finally, the tests right now only test that no error is produced when run but there are not expected values. Do some of the sorters have deterministic algorithms (accepting seeds for example) where we can detect their expected values and actually test for that. It would be nice to avoid silent errors. If this is not an option we could create a summary with plots to at least have a visual confirmation that everything is running all right. This can be published in github summary the way that I published the profile timing of the tests.

Feel free to add anything.

h-mayorquin:
Some of this was done here:
https://github.com/SpikeInterface/spikeinterface/pull/1764

Need to come back and check at some point if this issue still makes sense.

zm711:
Coming back to check to see if this still makes sense :)

h-mayorquin:
Yeah, let's close it, hopefully we can come back to it if we get the grant to improve this and add better testing.

PR #1600, opened 2023-04-28
Use annotations from future (PEP563) to solve type hint references

### LINKED ISSUE #1596: `import spikeinterface.extractors` fails if NWB not installed

h-mayorquin:
OK, so #1597 turn out too big as there were a lot of imports that were exposed naively. I am passing the small patch to fix the import with pynwb here as that one will probably take longer to reivew. 

Copying PR text verbatim from here:

This PR resolves #1596.

As described in the issue, importing extractors fails when pynwb is not installed. The problem arises from the newly added type-hints, which are currently evaluated at import time.

PR #1593 suggests a solution that employs forward references, where type hints are defined as strings to force deferred evaluation. While this approach works, we can improve upon it.

Starting with Python 3.8, we can use the following statement at the top of a module to defer the evaluation of type annotations until runtime:

```python
```
More details can be found in PEP 563: https://peps.python.org/pep-0563/

This change allows us to maintain the same notation regardless of whether the module is installed or not, simplifying code navigation.

Moreover, deferring type hints by default is expected to enhance import performance, which aligns with the long-term goals of the library.

Finally, I've added a test to the core that verifies all major modules can be imported with only a core installation, without raising errors. This test would have helped prevent the current issue. Furthermore, the test will be used to monitor import times, helping us track and reduce them in the future.

The tests and the use of run-time imports are now in #1597

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,7 +77,7 @@ streaming_extractors = [
     "fsspec", 
     "aiohttp",
     "requests",
-    "pynwb",
+    "pynwb>=2.3.0",
 ]
 
 full = [
--- a/src/spikeinterface/extractors/nwbextractors.py
+++ b/src/spikeinterface/extractors/nwbextractors.py
@@ -1,23 +1,25 @@
+from __future__ import annotations
 from pathlib import Path
-from typing import Union, List, Optional, Literal
+from typing import Union, List, Optional, Literal, Dict
 
 import numpy as np
-import h5py
 
 from spikeinterface import get_global_tmp_folder
 from spikeinterface.core import BaseRecording, BaseRecordingSegment, BaseSorting, BaseSortingSegment
 from spikeinterface.core.core_tools import define_function_from_class
 
 try:
-    import pynwb
     from pynwb import NWBHDF5IO, NWBFile
-    from pynwb.ecephys import ElectricalSeries, FilteredEphys, LFP, ElectrodeGroup
-    from hdmf.data_utils import DataChunkIterator
-    from hdmf.backends.hdf5.h5_utils import H5DataIO
-
+    from pynwb.ecephys import ElectricalSeries, ElectrodeGroup
     HAVE_NWB = True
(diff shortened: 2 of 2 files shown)


JoeZiminski:
Thanks @h-mayorquin  this is great solution, it works well now.

It seems like a nice idea for the images to pull the most recent version of spikeinterface from PyPI rather than directly from main (if, I understand correctly, that's what it's doing). I don't know much about images but will open issue on this unless reason not to?

h-mayorquin:
@alejoe91 
Can you pile up your take here:
https://github.com/SpikeInterface/spikeinterface/issues/1384

I think I was proposing someting similar in the first point of the list on there.

PR #1602, opened 2023-04-29
Make write to binary open and close per chunk

h-mayorquin:
I still wanted to add a memory test like I added for the generator. But this is the idea @samuelgarcia


src/spikeinterface/core/core_tools.py
```diff
@@ -249,95 +215,91 @@ def write_binary_recording(recording, file_paths=None, dtype=None, add_file_exte
...
+    # recover variables of the worker
+    recording = worker_ctx['recording']
+    dtype = worker_ctx['dtype']
+    file_path = worker_ctx["file_path_dict"][segment_index]
+    cast_unsigned = worker_ctx['cast_unsigned']
 
-    if chunk_size is not None and time_axis == 1:
-        print("Chunking disabled due to 'time_axis' == 1")
-        chunk_size = None
+    assert Path(file_path).is_file()
```
h-mayorquin, line 277: This is not needed. I will remove it.

src/spikeinterface/core/generate.py
```diff
@@ -396,8 +396,8 @@ def __init__(
         sampling_frequency: float,
         num_channels: int,
         dtype: Optional[Union[np.dtype, str]] = "float32",
-        seed: Optional[int] = None,
-        mode: Literal['white_noise', 'random_peaks'] = 'white_noise'
+        seed: float = 0,
```
samuelgarcia, line 399: I think I prefer to seed None in the signature and set to something when needed only (so almost all tests) to avoid the same siagnls always for a end user.
And I like the idea that user ask themself when a seed the random state.

h-mayorquin: I was testing for a big that I found. I will revert

src/spikeinterface/core/core_tools.py
```diff
@@ -315,48 +284,56 @@ def write_binary_recording(
...
+    memmap_obj = mmap.mmap(file.fileno(), length=mmmap_length, access=mmap.ACCESS_WRITE, offset=mmap_offset)
+
+    array = np.ndarray.__new__(np.ndarray, shape=shape, dtype=dtype, buffer=memmap_obj, order="C", offset=array_offset)
+    # apply function
+    traces = recording.get_traces(
+        start_frame=start_frame, end_frame=end_frame, segment_index=segment_index, cast_unsigned=cast_unsigned
+    )
+    if traces.dtype != dtype:
+        traces = traces.astype(dtype)
+    array[start_frame:end_frame, :] = traces
```
samuelgarcia, line 333: We are loosing the time_axis logic here.
It used to be important for some format to have time_axis=1 ti be able to write the transpose version.
Even this is super uneffcient.

src/spikeinterface/core/core_tools.py
```diff
@@ -315,48 +284,56 @@ def write_binary_recording(
 write_binary_recording.__doc__ = write_binary_recording.__doc__.format(_shared_job_kwargs_doc)
 
 
-def write_binary_recording_file_handle(
-    recording, file_handle=None, time_axis=0, dtype=None, byte_offset=0, verbose=False, **job_kwargs
-):
-    """
-    Old variant version of write_binary_recording with one file handle.
-    Can be useful in some case ???
-    Not used anymore at the moment.
```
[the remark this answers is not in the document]
alejoe91: Yes we should keep the `time_axis=1` for back-compatibility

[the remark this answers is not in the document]
alejoe91: it's ok

h-mayorquin:
All right, @samuelgarcia .The diff should be way simpler now.

src/spikeinterface/core/core_tools.py
```diff
@@ -215,15 +209,36 @@ def _write_binary_chunk(segment_index, start_frame, end_frame, worker_ctx):
...
+    memmap_obj = mmap.mmap(file.fileno(), length=mmmap_length, access=mmap.ACCESS_WRITE, offset=mmap_offset)
+
+    array = np.ndarray.__new__(np.ndarray, shape=shape, dtype=dtype, buffer=memmap_obj, order="C", offset=array_offset)
     # apply function
     traces = recording.get_traces(
         start_frame=start_frame, end_frame=end_frame, segment_index=segment_index, cast_unsigned=cast_unsigned
     )
-    traces = traces.astype(dtype)
-    rec_memmap[start_frame:end_frame, :] = traces
+    if traces.dtype != dtype:
```
h-mayorquin, line 236: This is to avoid a malloc.

h-mayorquin:
Requires merging https://github.com/SpikeInterface/spikeinterface/pull/1706 first. So, the padding in `write_to_binary_file` can be removed.

h-mayorquin:
This should have minimal or non-existent conflicts after  #1750.

alejoe91:
@h-mayorquin can you fix the conflicts?

samuelgarcia (approved):

PR #1604, opened 2023-05-01
Add pairs with decreased score to auto_merge output

DradeAW:
Added the list of pairs removed before there wasn't an increase in score in the `auto_merge` function in the extra_outputs.


DradeAW:
I don't think the test fail has anything to do with my code

h-mayorquin:
Maybe @CodyCBakerPhD knows.

PR #1610, opened 2023-05-04
Peak detection with numba. Compile at run time

h-mayorquin:
I am working on porting the peak detector methods to `PeakDetector` nodes after Sam's changes.

Something that I wanted to do for a while is to avoid the compilation at import time of numba. This PR address this.

Plus, I did some refactoring of the numba function.

### DIFF
--- a/src/spikeinterface/sortingcomponents/peak_detection.py
+++ b/src/spikeinterface/sortingcomponents/peak_detection.py
@@ -279,9 +279,11 @@ class DetectPeakLocallyExclusive(PeakDetectorWrapper):
     local_radius_um: float
         The radius to use to select neighbour channels for locally exclusive detection.
     """
+    
     @classmethod
     def check_params(cls, recording, peak_sign='neg', detect_threshold=5,
                      exclude_sweep_ms=0.1, local_radius_um=50, noise_levels=None, random_chunk_kwargs={}):
+
 
         if not HAVE_NUMBA:
             raise ModuleNotFoundError('"locally_exclusive" needs numba which is not installed')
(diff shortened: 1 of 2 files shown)


src/spikeinterface/sortingcomponents/peak_detection.py
```diff
@@ -504,23 +506,26 @@ def check_params(cls, recording, peak_sign='neg', detect_threshold=5,
     def get_method_margin(cls, *args):
         exclude_sweep_size = args[2]
         return exclude_sweep_size
-
+    
     @classmethod
     def detect_peaks(cls, traces, peak_sign, abs_threholds, exclude_sweep_size, neighbours_mask):
+        
+        numba_detect_peak = numba.jit(parallel=False)(_numba_detect_peak)
```
DradeAW, line 513: When using `numba.jit`, the `nopython=True/False` argument should be specified (#1616)

If not specified, it throws a warning, because as of now, the default value is `False`, whereas starting with numba 0.59.0, the default value will change to `True`.

h-mayorquin: Yes, I am aware of this change. I have not had the chance of thinking what the value SHOULD be. What do you think?

h-mayorquin: All right, I was thinking across those lines. I will take a look at this later or when it crashes.

Sam wants to do something before this PR so this might take a while.

zm711: Not sure if this still matters but was working with some Numba code so I wanted to drop this here as food for thought. I needed to do a factorial in some code and found out that Numba doesn't support `math.factorial` yet, but `math.gamma` is supported. Since `math.factorial(x)` is pretty close to `math.gamma(x+1)` (as I read on Stack Overflow haha). If we use `math.gamma` we get floating point error if we jit with Numba, but not with the native Python:

```python
math.gamma(11) # native python
3628800.0
math.gamma(11) # called inside of a numba jit-ed function
3628799.9999999995
```
So faster yes, but I think reliability needs to be empirically tested and it depends on whether the potential floating error is compounded or just called once. That being said, I also just always set `nopython=True` and wait for Numba compile errors.

PR #1613, opened 2023-05-05
Fix dtype bug in Windows for `spost.compute_refrac_period_violations`

DradeAW:
So as it turns out, `np.int32 * int = np.int64` on Linux to get behind on overflow error, but defaults to `np.int32` in Windows. This leads to an erroneous `D` in Windows with big units.

Changing the dtype to `int64` should fix all problems.


alejoe91:
@DradeAW can you be more specific in PR names? ;)

alejoe91:
> I on't believe the test failing has anything to do with my code?

Nope ;)

they'll be fixed once this is merged: https://github.com/SpikeInterface/spikeinterface/pull/1611

DradeAW:
If it's possible to merge this quickly?

It's causing my unit tests on Windows to fail ^^

alejoe91 (approved):

alejoe91:
Ok, I'd say the fix is enough circumscribed ;)

PR #1620, opened 2023-05-08
Refactor clustering tests

h-mayorquin:
Two things in this PR.

I am separating the tests so there is one test per method. This is useful so you know which one failed in the CI. Moreover, the tests can be run separatley. I added the trick so you guys can keep running your tests for main by selecting the method.

I also refactor some variable naming and the common elements to all the methods to unify this.


yger:
One option there would be to make all Benchmarks inherits from the BaseClass we created for the BenchmarkMotion. This allows to load/save benchmarks, and thus maybe reload/make plots faster. Happy to discuss more in depth these refactoring, since it might be important to properly decide what should be visualized and how

alejoe91 (approved):

Issue #1619, opened 2023-05-08
Loading sorting from pickle takes an absurd amount of time


DradeAW:
Hi,

I'm saving my sorting's provenance using the pickle format, and loading them from pickle.
I noticed that when the sorting object becomes a bit complicated (with multiple parents), the loading takes an incredibly absurd amount of time (4 minutes and 56 seconds) to load (writing is relatively fast).

Here is a sample of the code:

```python
sorting.dump_to_pickle(file_path="ks2_best.pkl", include_properties=True)
# Do stuff
sorting = si.load_extractor("ks2_best.pkl")
```

Here is the result of `sorting.to_dict()`:

```
sorting.to_dict()
{'class': 'spikeinterface.core.unitsaggregationsorting.UnitsAggregationSorting', 'module': 'spikeinterface', 'kwargs': {'sorting_list': [AlignSortingExtractor: 20 units - 1 segments - 30.0kHz, UnitsSelectionSorting: 46 units - 1 segments - 30.0kHz], 'renamed_unit_ids': [4, 6, 7, 12, 14, 16, 17, 19, 22, 28, 30, 31, 34, 36, 37, 38, 40, 53, 58, 60, 8, 13, 21, 23, 24, 49, 54, 55, 56, 61, 63, 64, 0, 1, 2, 3, 5, 9, 10, 11, 15, 20, 25, 26, 27, 29, 32, 33, 35, 39, 41, 42, 43, 44, 45, 46, 47, 48, 50, 51, 52, 57, 59, 62, 65, 66]}, 'dumpable': True, 'version': '0.98.0.dev0', 'relative_paths': False, 'annotations': {}, 'properties': {}}
```

I also did a `sorting.dump_to_json`, but the file is 3,286 lines long. I can send it via Slack if helpful.

I don't understand why it's taking that long?
Loading the same `sorting` from json takes between 0.5-1 s.

alejoe91:
Thanks for the report.Is there a reason why you're using pkl rather than JSON?

DradeAW:
Yep, 2 reasons:

First - I believe with pickle, numpy arrays stays as arrays and not lists. There's no conversion which is convenient.
Second - json creates big files. In this example (which is still a relatively small sorting for my use case), the pickle file is 14.3 kB, whereas the json file is 241.2 kB. For bigger sorting objects, the json file is something like 14 MB ...

DradeAW:
[...]

PR #1622, opened 2023-05-10
Fix Extractor pickle loading being too slow

DradeAW:
This PR attemps to fix the issue talked about in #1619.


DradeAW:
Ah, if I increase in complexity then I crash because it uses too much RAM (I have 128 GB of RAM)

This is an issue. I'm investigating ...

h-mayorquin:
I thought about this. I think this solution works well. These are my suggestions:
1) The logic for recursive can be inside the `to_dict` function. You only need to call `to_dict` from `to_dict` instead of from `_to_dict_iterative` as you are doing right now. While this makes the current `to_dict` more complicated I think that the parameter name and the logic of the function is clear enough. Separating in this case implies less coheisve code as is revelead by your question of where to put this:

```python
# I don't know where to put this
def _to_dict_iterative(obj, **kwargs):
    if isinstance(obj, dict):
        for key, value in obj.items():
            obj[key] = _to_dict_iterative(value)
    elif isinstance(obj, (list, np.ndarray)):
        for i in range(len(obj)):
            obj[i] = _to_dict_iterative(obj[i])
    elif isinstance(obj, BaseExtractor):
        obj = obj.to_dict(**kwargs)
        obj = _to_dict_iterative(obj)

    return obj
```
2) You only need for `BaseExtractor` to expand in `_kwargs` and not in the whole object, right?
3) I think that the argument recursive should be true by default so the internal pickling can profit from this optimization. Producing serializable dictionaries is the goal of this function and I think a less serializable representation should be opt-in instead of opt-out.

Outside of that, your current implementation is modifying the inplace `_kwargs` which maybe is undesirable? I kind of thing that your performance would decrease if you avoid that but I am not sure. I need to think more about this.

You can use the following code to profile any changes and see that your changes do not break as you iterate:

https://gist.github.com/h-mayorquin/72f492275767fa9f2cb77a0b117eea67

What do you think?

DradeAW:
1. I don't know how to implement it inside `to_dict()` without declaring a function inside a function (which I personally don't like). The reason is that some objects are really complex (a dict mapping to a list mapping to `BaseExtractor`), so you need recursion and not a simple for loop. If you have a way to implement it without recursion I'm all ears!
2. This is what I thought also, but I wasn't sure and such a function should be quick anyway?
3. I think so too. I just didn't want to change the default behaviour without your input :)

You mean I should make a deepcopy of `_kwargs` rather than changing the object itself?
This is probably true, an oversight on my part ^^'

I'll look into profiling and changing some code on Monday!
Another thing I need to look into is RAM usage, because last time I tried to use it I had a memory error (with 128 GB of RAM!!)

h-mayorquin:
> 1. I don't know how to implement it inside `to_dict()` without declaring a function inside a function (which I personally don't like). The reason is that some objects are really complex (a dict mapping to a list mapping to `BaseExtractor`), so you need recursion and not a simple for loop. If you have a way to implement it without recursion I'm all ears!

Something like this?
```python
        to_dict_kwargs = dict(include_annotations=include_annotations, include_properties=include_properties,
                            relative_to=relative_to, folder_metadata=folder_metadata)
        for name, value in self._kwargs.items():
            if isinstance(value, list):
                transform_to_dict = lambda x: x.to_dict(**to_dict_kwargs) if isinstance(x, BaseExtractor) else x
                new_list = [transform_to_dict(element) for element in value]
                self._kwargs[name] = new_list
            elif isinstance(value, BaseExtractor):
                self._kwargs[name] = value.to_dict(**to_dict_kwargs)
```

On a second thought this could also be outside, but I think it should be a method of `BaseExtractor` whose role is to recursively  expand the Extractor objects in the kwargs to dicts. Not sure of the name yet? `kwargs_to_dict_recursively`. What do you think is better?

> 2. This is what I thought also, but I wasn't sure and such a function should be quick anyway?
I just tested the above and it is ten times faster for saving

> You mean I should make a deepcopy of `_kwargs` rather than changing the object itself? This is probably true, an oversight on my part ^^'

Just create a new dict where the expanded kwargs and the dumpt_dict should have that one as `kwargs` instead of the original one.

> I'll look into profiling and changing some code on Monday! Another thing I need to look into is RAM usage, because last time I tried to use it I had a memory error (with 128 GB of RAM!!)

The problem with ram is not related with this. I think. If you are still using mearec, the problem with the sorting mearec extractor is that is not lazy so when you "depickle" every extractor inside your reference pipeline will load the data from mearec. For example, if you had the pipeline in the gist that I shared with you with a mearec extractor, the whole meraec data will be loaded 50 times.  You could test this by using a lazy sorting object. I think you could do `your_sorter.save()` at the beginning of the pipeline (before all the agregations and the sub-selections) to test if this is the case.

Of course, the right thing to do would be to make mearec lazy. But that would be more work.

h-mayorquin:
Thanks by looking into this, btw. This is an optimization that will be felt across a bunch of parts of the library as everything that uses `ChunkRecordingExecutor` basically pickled and depikcles on the go.

h-mayorquin:
@DradeAW 
> I prefer outside using recursion, because in your example, if you have for example a dict of list of dict of `BaseExtractor`, then it's not going to work. Whereas recursion will find it even if it's nested somewhere.

The expression above is recursive over list and BaseExtractor objects. I tried it before sharing it with you and it works. The reason I did not added dict logic is because I don't think we have dicts with BaseExtractors inside kwargs but maybe I am wrong?. If so, it should be straighforward as well:

```python
        new_kwargs = dict()
        transform_to_dict = lambda x: x.to_dict(**to_dict_kwargs) if isinstance(x, BaseExtractor) else x
        for name, value in self._kwargs.items():
            if isinstance(value, list):
                new_list = [transform_to_dict(element) for element in value]
                new_kwargs[name] = new_list
            if isinstance(value, dict):
                new_kwargs[name] = {k: transform_to_dict(v) for k, v in value.items()}
            elif isinstance(value, BaseExtractor):
                new_kwargs[name] = value.to_dict(**to_dict_kwargs)
            else:
                new_kwargs[name] = value
```
Am I missing something?


> Those sortings are not from MEArec, they are from `PhySortingExtractor` (which I guess isn't lazy also).

Neither of them are lazy by looking at the code. I added machinery to the test script that I shared with you to measure memory usage if that is useful for you. Peak memory usage will only work on linux.

DradeAW:
> The reason I did not add dict logic is because I don't think we have dicts with BaseExtractors inside kwargs but maybe I am wrong?

I can't think of an example, but I think that code in `core` should be future-proof.
Imagine if in the future, we have something like:
```python
kwargs = {
    'a': {
        'b': [{'c': sorting_object}]
    }
}
```

Your method would not handle such a case, and hence isn't future-proof.

I'm looking at profiling right now :)
Btw, if you are dividing by `1e6`, it's not `MiB` but `MB` (MiB are if you divide by `2e20`)

DradeAW:
I think also one of the main problem is in `curation.RemoveDuplicatedSpikesSorting` that is not lazy (before this sorting it's relatively fast, and after it's pretty slow).

Maybe I should take a look at that first (especially since loading from json crashes when I reach this module in my pipeline!!)

DradeAW:
> futuring proofing more complex arguments for initializing extractors feels wrong to me

As you prefer :)

> I am not sure where you are getting that transforming bytes to MiB would requires dividing by 2**20? Can you share your reasoning with me? Maybe it is bytes and bits confusion?

1k = 1,000 ; 1ki = 1,024 = 2^10 (see Table here: https://en.wikipedia.org/wiki/Binary_prefix)
1M = 10^6 ; 1Mi = 1,024 * 1,024 = 2^20
This is why when you buy a 1 TB drive, it shows up as 931 GiB

>> I think also one of the main problem is in curation.RemoveDuplicatedSpikesSorting that is not lazy (before this sorting it's relatively fast, and after it's pretty slow).

> Looking at that class it seems that it does some computation at the __init__ in the segment level. Maybe that's what is eating your ram

Yeah exactly

h-mayorquin:
> > futuring proofing more complex arguments for initializing extractors feels wrong to me
> 
> As you prefer :)
> 
> > I am not sure where you are getting that transforming bytes to MiB would requires dividing by 2**20? Can you share your reasoning with me? Maybe it is bytes and bits confusion?
> 
> 1k = 1,000 ; 1ki = 1,024 = 2^10 (see Table here: https://en.wikipedia.org/wiki/Binary_prefix) 1M = 10^6 ; 1Mi = 1,024 * 1,024 = 2^20 This is why when you buy a 1 TB drive, it shows up as 931 GiB

You are completly right, I have done this before in the right way:
https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/tests/test_generate.py
And now I did it wrong. 

Thanks. I will correct the script.

> > > I think also one of the main problem is in curation.RemoveDuplicatedSpikesSorting that is not lazy (before this sorting it's relatively fast, and after it's pretty slow).
> 
> > Looking at that class it seems that it does some computation at the **init** in the segment level. Maybe that's what is eating your ram
> 
> Yeah exactly

h-mayorquin:
> > > futuring proofing more complex arguments for initializing extractors feels wrong to me
> > 
> > 
> > As you prefer :)

This is just my opinion. Maybe Sam and Alessio agree with you.  I think they have been very busy so they have not weight in. I will discuss this briefly with them when we see each other and see what vision they have.

> > > I am not sure where you are getting that transforming bytes to MiB would requires dividing by 2**20? Can you share your reasoning with me? Maybe it is bytes and bits confusion?
> > 
> > 
> > 1k = 1,000 ; 1ki = 1,024 = 2^10 (see Table here: https://en.wikipedia.org/wiki/Binary_prefix) 1M = 10^6 ; 1Mi = 1,024 * 1,024 = 2^20 This is why when you buy a 1 TB drive, it shows up as 931 GiB
> 
> You are completly right, I have done this before in the right way: https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/tests/test_generate.py And now I did it wrong.
> 
> Thanks. I will correct the script.
> 
> > > > I think also one of the main problem is in curation.RemoveDuplicatedSpikesSorting that is not lazy (before this sorting it's relatively fast, and after it's pretty slow).
> > 
> > 
> > > Looking at that class it seems that it does some computation at the **init** in the segment level. Maybe that's what is eating your ram
> > 
> > 
> > Yeah exactly

h-mayorquin:
@DradeAW Did you change anything in your repo for the pre-commit boot to work?

DradeAW:
> @DradeAW Did you change anything in your repo for the pre-commit boot to work?

Not to my knowledge.
I just fixed the conflicts and pulled from main, and it ran on its own.

h-mayorquin:
It is strange because I think the bot pushed a commit to your branch.

DradeAW:
Yes, it did to all 3 of my PRs after pulling from main

DradeAW:
> This is just my opinion. Maybe Sam and Alessio agree with you. I think they have been very busy so they have not weight in. I will discuss this briefly with them when we see each other and see what vision they have.

@alejoe91 @samuelgarcia Do you have a preference?

Once we establish the way to go I can implement it.
Ideally I would need this PR to be merged this week :)

src/spikeinterface/core/base.py
```diff
@@ -951,3 +970,18 @@ def parent_extractor(self):
 
     def set_parent_extractor(self, parent_extractor):
         self._parent_extractor = weakref.ref(parent_extractor)
+
+
+# I don't know where to put this
+def _to_dict_iterative(obj, **kwargs):
```
alejoe91, line 976: `recursive_to_dict` into `core_tools`

src/spikeinterface/core/base.py
```diff
@@ -278,7 +278,12 @@ def copy_metadata(self, other, only_main=False, ids=None):
             other._preferred_mp_context = self._preferred_mp_context
 
     def to_dict(
-        self, include_annotations=False, include_properties=False, relative_to=None, folder_metadata=None
+        self,
+        include_annotations: bool = False,
+        include_properties: bool = False,
+        relative_to=None,
+        folder_metadata=None,
+        recursive: bool = False,
```
alejoe91, line 286: `True` by default

DradeAW:
Thanks @alejoe91 !! :heart:

alejoe91:
@h-mayorquin @DradeAW not sure why but tests are taking suuuuper long!!! I suspect it could be due to something in this PR since #1668 is "fast" as usual. What do you think?

DradeAW:
Mmhm it's worth investigating,

Maybe this new method is faster for big complex extractors, but slower for small need to be reconstructed?

I also had an issue where some extractors required a lot of RAM to load, maybe due to the non-lazyness of some extractors?
Do you know how much RAM the machines for running tests have?

I'm gonna try to export both to pickle and json, and do some benchmarking.
But I have to solve another bug before (right now Lussac is broken :cry: )

src/spikeinterface/core/base.py
samuelgarcia: why finally False ?

alejoe91:
Tests became very slow with true, but we are still investigating if it was a relative import..let's keep this open for now!

DradeAW:
Ok so I tried a benchmark:

```python
import os
import shutil
import time
import numpy as np


N = 20
time_json = np.empty(N, dtype=np.float32)
time_pickle = np.empty(N, dtype=np.float32)
time_dict_false = np.empty(N, dtype=np.float32)
time_dict_true = np.empty(N, dtype=np.float32)


# Create a simple sorting with some parents
sorting1 = se.PhySortingExtractor("/mnt/raid0/data/2020_11_25_eyeblink_1/lussac/output/ks2_best")
sorting2 = se.PhySortingExtractor("/mnt/raid0/data/2020_11_25_eyeblink_1/lussac/output/ms4_cs")

sorting1 = sorting1.select_units([0, 4, 5, 6, 13, 14, 15, 16, 22, 26, 28, 29, 31, 32, 33, 41, 42])
sorting2 = sorting2.select_units([1, 2, 5, 8, 12, 16, 20, 21, 23, 49, 52, 53, 58, 60, 61, 63, 67, 69])

sorting = si.UnitsAggregationSorting([sorting1, sorting2])


# Benchmark
for i in range(N):
	os.makedirs("tmp/benchmark_pickle", exist_ok=True)

	t1 = time.perf_counter()
	
	sorting.dump_to_json("tmp/benchmark_pickle/sorting.json")
	si.load_extractor("tmp/benchmark_pickle/sorting.json")
	
	t2 = time.perf_counter()

	sorting.dump_to_pickle("tmp/benchmark_pickle/sorting.pkl")
	si.load_extractor("tmp/benchmark_pickle/sorting.pkl")

	t3 = time.perf_counter()

	a = sorting.to_dict(recursive=False)
	si.base._load_extractor_from_dict(a)

	t4 = time.perf_counter()

	a = sorting.to_dict(recursive=True)
	si.base._load_extractor_from_dict(a)

	t5 = time.perf_counter()

	time_json[i] = t2 - t1
	time_pickle[i] = t3 - t2
	time_dict_false[i] = t4 - t3
	time_dict_true[i] = t5 - t4
	shutil.rmtree("tmp/benchmark_pickle")

print(f"JSON:\t {1e3*np.mean(time_json):.1f} +- {1e3*np.std(time_json):.1f} ms")
print(f"Pickle:\t {1e3*np.mean(time_pickle):.1f} +- {1e3*np.std(time_pickle):.1f} ms")
print(f"Dict f:\t {1e3*np.mean(time_dict_false):.1f} +- {1e3*np.std(time_dict_false):.1f} ms")
print(f"Dict t:\t {1e3*np.mean(time_dict_true):.1f} +- {1e3*np.std(time_dict_true):.1f} ms")
```

Here is the output:
```JSON:	 308.9 +- 6.1 ms
Pickle:	 309.9 +- 7.4 ms
Dict f:	 305.9 +- 3.7 ms
Dict t:	 305.5 +- 4.2 ms
```

It seems that loading the `PhySortingExtractor` is what takes the most time, and that everything else is negligible.

DradeAW:
So here are some interesting results (time in mm:ss run locally on my machine):

```
# recursive = False
pytest -m core  # 03:36
pytest core/tests/test_waveform_extractor.py  # 01:34

# recursive = True
pytest -m core  # 05:38
pytest core/tests/test_waveform_extractor.py  # 03:19
```

So I think we can investigate in `test_waveform_extractor`.
However I'm sorry but I don't have time to look into it further today, and as of tomorrow I'm taking a week and a half of vacation,
Then I'll probably be stuck writing my PhD thesis :/

alejoe91:
What about merging with `recursive=False`? You can always set the `recursive=True` if you need it (and we probably need to propagate it to the `dump_*` functions

DradeAW:
That works for me!

It would be really nice to understand this phenomenon, but unfortunately it seems that it's gonna take too much time ^^

DradeAW:
I can understand loosing a few ms, but whole minutes on this!?

I would be surprised, but it's worth taking a look. Thanks!

h-mayorquin:
This is the non deep recursive True by default: 
https://github.com/SpikeInterface/spikeinterface/actions/runs/5069547999/jobs/9103314766?pr=1672

What's the commit that was slow here to compare?

DradeAW:
> What's the commit that was slow here to compare?

The slowdown is with `recursive=True`, and I believe your commit was with `recursive=False`?

DradeAW:
I meant that the method `BaseExtractor.to_dict()` had a default value of `recursive=False` meaning it wasn't gonna call the function unless it was asked specifically.

When it's on False, we don't have any problem and it solves my problem of loading big objects.
However, when set on `recursive=True` (meaning it does it even outside pickling, but for all instances), the unit tests slow down.
So any testing to understand why this slowdown happens has to be done with a default `recursive=true` in `BaseExtractor.to_dict`

DradeAW:
> Another thing I need to look into is RAM usage, because last time I tried to use it I had a memory error (with 128 GB of RAM!!)

I can't reproduce the error, and now loading is the same speed for small and very complex sorting objects, while not causing a RAM issue.
Don't know why it's not a problem now, but I'll take it!

DradeAW:
> So we might not need dicts, but we do need lists (e.g. for concatenate recordings!)

My method was deep (meaning it could find an extractor if it was in a dict of list of list of dict of dict of ...
@h-mayorquin stated that this is probably not necessary as there isn't any extractor that has this, while I thought it was better to plan for the future and make something that was going to work on all cases.

This is the deep vs fast method that we're talking about here.

samuelgarcia:
Hi all.
I do not have time to read everything now. Please do not merge with recursive=False.
I have the intuition that this is a bad idea. If this is slower this must hide something deeper.
I would like to have a small look.

alejoe91:
Yeeep, let's get to the bottom of it ;)

h-mayorquin:
This is what my [profiling](https://gist.github.com/h-mayorquin/72f492275767fa9f2cb77a0b117eea67) indicates:

```
---------------------
results for branch fix_pickle [This branch]
Time elapsed for saving: 0.14322662353515625
Time elapsed for loading: 8.988275527954102  <- This is slow
Memory used by loading: 63.67578125 MiB
Peak memory usage: 809.765625 MiB

---------------------
results for branch pickling_fix [The branch in https://github.com/SpikeInterface/spikeinterface/pull/1672]
Time elapsed for saving: 0.2201213836669922
Time elapsed for loading: 0.15368962287902832 <- This is not slow
Memory used by loading: 0.25390625 MiB
Peak memory usage: 690.375 MiB
```

The tests become very slow because loading here with `ChunkRecordingExecutor` over the test suit adds up. I am honestly baffled because the difference in saved (my approach being faster is along the lines what I and @DradeAW I think would expect, somewhat faster but not incredible so). However, somehow the approach here creates representations that way slower to unpickle.

h-mayorquin:
@alejoe91 @samuelgarcia 
I suggest merging #1672 which will solve the problem here and propagate the performance improvment to main. We can merge this branch when it does not degrade performance. This has the advantage of solving the problem of @DradeAW for this week (as he stated above he would need it) while also frees him of the time commitment (as he has expressed he is on a crunch). 

Then, when either you guys or @DradeAW has time you can work on writing a more general function that does not degrade performance and merge this.

samuelgarcia:
Thank Ramon for more proposal on this.
This difference in loading is hard to believe/understand because the dict are the same normally.
Or I am missing something. Or there is something deeper.

I think I prefer the #1672 solution. So no reccurssive option.

h-mayorquin:
@DradeAW 
Turns out the solution proposed in the other PR was wrong:
https://github.com/SpikeInterface/spikeinterface/pull/1672/files#r1204810812

I was getting better performance because of a wrong control structure but it is as you say, they should be doing the same. 

Turns out, the problem is on the loading and there was a deep copy that wa repeated recursively. The correct solution to your problem is on:
https://github.com/SpikeInterface/spikeinterface/pull/1674

DradeAW:
> Turns out, the problem is on the loading and there was a deep copy that was repeated recursively

Ah!
That actually makes a lot of sense.
I didn't like the `deepcopy` but didn't know how to get ride of it.

Good find! ;)

DradeAW:
Should we close this PR?

h-mayorquin:
Yes, how is your performance with the recent changes?

PR #1625, opened 2023-05-10
Add New Benchmarks for Matching

pauladkisson:
# Overview

Adds 3 new benchmarks to the benchmark_matching submodule of sortingcomponents.  

## run_matching_num_spikes
Varies the number of spikes extracted for each unit to inject noise into the templates and assess how matching algorithms perform under challenging conditions.  One use case of this benchmark is to assess matching methods suitability for short recordings with low FRs (ex. motion estimation).

## run_matching_misclassed
Randomly misclassifies spikes for each unit (as long as they are similar enough by cosine similarity) to inject noise into templates and assess how matching algorithms perform with errors in clustering.  

## run_matching_missing_units
Randomly omits units from the ground-truth sorting below a snr_threshold to assess how well matching algorithms reject spikes from units present in the recording but omitted upstream in the pipeline.

## Reorganization
To improve the benchmarking module, I have made the following changes
- Pass waveform_extractor at initialization rather than constructing one from recording/sorting
    - This allows multiple sorters to use the same waveform_extractor --> dramatically speeds up benchmarking for large recordings/many matching methods
 - changed job_kwargs parameter to **job_kwargs for consistency with other methods
 - each of the `run_matching` methods are by `run_matching_vary_parameter` with appropriate parameter variations for streamlined code

## Notes
- I also included changes to the wobble alg to speed it up that I profiled earlier.
- At the top of the file is a utility function `running_in_notebook` that checks whether code is currently running in a jupyter notebook to take advantage of ipython `display()` and `clear_output()` functions for pretty progress bars.  This function should probably be in some utils file, but I'm not sure where.
- Continues PR #1561

### DIFF
--- a/src/spikeinterface/sortingcomponents/benchmark/benchmark_matching.py
+++ b/src/spikeinterface/sortingcomponents/benchmark/benchmark_matching.py
@@ -5,9 +5,14 @@ from spikeinterface.sortingcomponents.matching import find_spikes_from_templates
 from spikeinterface.core import NumpySorting
 from spikeinterface.qualitymetrics import compute_quality_metrics
 from spikeinterface.comparison import CollisionGTComparison, compare_sorter_to_ground_truth
-from spikeinterface.widgets import (plot_sorting_performance,
-                                    plot_agreement_matrix, plot_comparison_collision_by_similarity,
-                                    plot_unit_templates, plot_unit_waveforms, plot_gt_performances)
+from spikeinterface.widgets import (
+    plot_sorting_performance,
+    plot_agreement_matrix,
+    plot_comparison_collision_by_similarity,
+    plot_unit_templates,
+    plot_unit_waveforms,
+    plot_gt_performances,
+)
 
 import time
 import os
@@ -35,8 +40,18 @@ def running_in_notebook():
 class BenchmarkMatching:
     """Benchmark a set of template matching methods on a given recording and ground-truth sorting."""
 
-    def __init__(self, recording, gt_sorting, waveform_extractor, methods, methods_kwargs=None, exhaustive_gt=True,
-                 tmp_folder=None, template_mode='median', **job_kwargs):
+    def __init__(
+        self,
+        recording,
+        gt_sorting,
+        waveform_extractor,
+        methods,
+        methods_kwargs=None,
+        exhaustive_gt=True,
+        tmp_folder=None,
+        template_mode="median",
+        **job_kwargs,
+    ):
         self.methods = methods
         if methods_kwargs is None:
             methods_kwargs = {method: {} for method in methods}
@@ -48,19 +63,21 @@ class BenchmarkMatching:
         self.sampling_rate = self.recording.get_sampling_frequency()
 
         if tmp_folder is None:
-            tmp_folder = os.path.join('.', ''.join(random.choices(string.ascii_uppercase + string.digits, k=8)))
+            tmp_folder = os.path.join(".", "".join(random.choices(string.ascii_uppercase + string.digits, k=8)))
         self.tmp_folder = Path(tmp_folder)
         self.sort_folders = []
 
         self.we = waveform_extractor
         for method in self.methods:
-            self.methods_kwargs[method]['waveform_extractor'] = self.we
+            self.methods_kwargs[method]["waveform_extractor"] = self.we
         self.templates = self.we.get_all_templates(mode=template_mode)
-        self.metrics = compute_quality_metrics(self.we, metric_names=['snr'], load_if_exists=True)
+        self.metrics = compute_quality_metrics(self.we, metric_names=["snr"], load_if_exists=True)
         self.similarity = compute_template_similarity(self.we)
-        self.parameter_name2matching_fn = dict(num_spikes=self.run_matching_num_spikes,
-                                               fraction_misclassed=self.run_matching_misclassed,
-                                               fraction_missing=self.run_matching_missing_units)
+        self.parameter_name2matching_fn = dict(
+            num_spikes=self.run_matching_num_spikes,
+            fraction_misclassed=self.run_matching_misclassed,
+            fraction_missing=self.run_matching_missing_units,
+        )
 
     def __enter__(self):
         self.tmp_folder.mkdir(exist_ok=True)
@@ -94,16 +111,17 @@ class BenchmarkMatching:
         sortings, runtimes = {}, {}
         for method in self.methods:
             t0 = time.time()
-            spikes = find_spikes_from_templates(self.recording, method=method,
-                                                method_kwargs=methods_kwargs[method],
-                                                **self.job_kwargs)
+            spikes = find_spikes_from_templates(
+                self.recording, method=method, method_kwargs=methods_kwargs[method], **self.job_kwargs
+            )
             runtimes[method] = time.time() - t0
-            sorting = NumpySorting.from_times_labels(spikes['sample_index'], unit_ids[spikes['cluster_index']],
-                                                     self.sampling_rate)
+            sorting = NumpySorting.from_times_labels(
+                spikes["sample_index"], unit_ids[spikes["cluster_index"]], self.sampling_rate
+            )
             sortings[method] = sorting
         return sortings, runtimes
 
-    def run_matching_num_spikes(self, spike_num, seed=0, we_kwargs=None, template_mode='median'):
+    def run_matching_num_spikes(self, spike_num, seed=0, we_kwargs=None, template_mode="median"):
         """Run template matching with a given number of spikes per unit.
 
         Parameters
@@ -127,8 +145,9 @@ class BenchmarkMatching:
         """
         if we_kwargs is None:
             we_kwargs = {}
-        we_kwargs.update(dict(max_spikes_per_unit=spike_num, seed=seed, overwrite=True, load_if_exists=False,
-                              **self.job_kwargs))
+        we_kwargs.update(
+            dict(max_spikes_per_unit=spike_num, seed=seed, overwrite=True, load_if_exists=False, **self.job_kwargs)
+        )
 
         # Generate New Waveform Extractor with New Spike Numbers
         we = extract_waveforms(self.recording, self.gt_sorting, self.tmp_folder, **we_kwargs)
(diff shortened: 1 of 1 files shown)


h-mayorquin:
I think that for readability it is better to avoid plurs for collection. That is, if you have cars (for a list of cars ) and car in the same code. You need to pay a lot of attention when reading to see if you are referring to the collection or the single entity. Examples that I saw today are benchmarks, colors, etc.

I think it is better to point out the to the container type: `car_list`, `bencharmk_dict`, `color_set` etc.

pauladkisson:
> I think that for readability it is better to avoid plurals for collection. That is, if you have cars (for a list of cars ) and car in the same code. You need to pay a lot of attention when reading to see if you are referring to the collection or the single entity. Examples that I saw today are benchmarks, colors, etc.
> 
> I think it is better to point out the to the container type: `car_list`, `bencharmk_dict`, `color_set` etc instead of `cars`, `benchmarks` and `colors`.

Is this just a preference or is it some kind of standard?  

I'm not a huge fan of referring to the type of a collection because it limits flexibility.  For example, when a user calls ``run_matching_vary_parameter(parameters)`` the ``parameters`` object could be any iterable (np.ndarray, list, etc.).  But if you call it ``parameter_list`` or ``parameter_array`` then you could be misinforming readers that the collection needs to be an array or a list respectively.  One could call the variable ``parameter_collection`` or ``parameter_iterable`` to try to convey the flexibility of the collection type, but I think ``parameters`` is more readable and more succinct.

pauladkisson:
Quick search found [this](https://ahmedmahmoud-eltaher.medium.com/clean-code-best-practice-for-naming-part-1-f67ebe8c0894) Medium article of someone who likes to use plural nouns for collections

h-mayorquin:
No standard. My own view based on the idea that when reading a bunch of stuff is easy to mistake words that only differ in one letter.

But you seem to have a principled take that you would rather keep the duck typing explicit. And while I disagree about the value of succintness I don't really have a strong argument or empirical fact against your position. And I appreciate a principled take.

Thanks for sharing your views with me.

h-mayorquin:
pre-commit.ci autofix

pauladkisson:
Looks like some formatting tool (black?) edited these files.  Please lmk what tool was used and I will apply the formatting to resolve these conflicts.

h-mayorquin:
@pauladkisson I pushed a PR to your repo that should fix the stuff automatically.

Otherwise you can run something like:
```
pip install pre-commit
pre-commit run --all-files
```
And then commit the changes.

pauladkisson:
Well that didn't work...😅😅

samuelgarcia:
> someone who likes to use plural nouns for collections
I am in this category.

src/spikeinterface/sortingcomponents/benchmark/benchmark_matching.py
```diff
@@ -16,101 +17,483 @@
...
+
+
+def running_in_notebook():
+    try:
+        shell = get_ipython().__class__.__name__
+        notebook_shells = {"ZMQInteractiveShell", "TerminalInteractiveShell"}
+        # if a shell is missing from this set just check get_ipython().__class__.__name__ and add it to the set
+        return shell in notebook_shells
+    except NameError:
+        return False
```
samuelgarcia, line 38: I do not like so much the idea that a piece of code is aware itself where it is run.
I think it is beeter to have clear flags to change a display behavior.

src/spikeinterface/sortingcomponents/benchmark/benchmark_matching.py
```diff
@@ -16,101 +17,483 @@
...
+            raise ValueError("'min_similarity' must be between -1 and 1.")
+        if we_kwargs is None:
+            we_kwargs = {}
+        we_kwargs.update(dict(seed=seed, overwrite=True, load_if_exists=False, **self.job_kwargs))
+        rng = np.random.default_rng(seed)
+
+        # Randomly misclass spike trains
+        spike_time_indices, labels = [], []
+        for unit_index, unit_id in enumerate(self.we.unit_ids):
+            unit_sorting = self.gt_sorting.get_unit_spike_train(unit_id=unit_id)
```
samuelgarcia, line 227: unit_sorting in the context of spikeinterface is a bit confusin.
spike_train is more easy to read no ?

src/spikeinterface/sortingcomponents/benchmark/benchmark_matching.py
```diff
@@ -16,101 +17,483 @@
...
+            num_spikes = int(len(unit_sorting) * fraction_misclassed)
+
+            unit_misclass_idx = rng.choice(np.arange(len(unit_sorting)), size=num_spikes, replace=False)
+            for i, spike in enumerate(unit_sorting):
+                spike_time_indices.append(spike)
+                if i in unit_misclass_idx and at_least_one_similar_unit:
+                    alt_id = rng.choice(similar_unit_ids)
+                    labels.append(alt_id)
+                else:
+                    labels.append(unit_id)
```
samuelgarcia, line 241: I have the feeeling that this loop could be numpy vectorized, no ?
looping over spikes can be super long no ?

src/spikeinterface/sortingcomponents/benchmark/benchmark_matching.py
```diff
@@ -16,101 +17,483 @@
...
+        spike_time_indices, spike_cluster_ids = [], []
+        for unit in present_units:
+            spike_train = self.gt_sorting.get_unit_spike_train(unit)
+            for time_index in spike_train:
+                spike_time_indices.append(time_index)
+                spike_cluster_ids.append(unit)
+        spike_time_indices = np.array(spike_time_indices)
+        spike_cluster_ids = np.array(spike_cluster_ids)
+        gt_sorting = NumpySorting.from_times_labels(
+            spike_time_indices, spike_cluster_ids, self.sampling_rate, unit_ids=present_units
```
samuelgarcia, line 315: We have a method that do the same I think.
`sorting_short = sorting.select_units(some_ids)`
can you check ?

src/spikeinterface/sortingcomponents/benchmark/benchmark_matching.py
```diff
@@ -16,101 +17,483 @@
...
+            {
+                "sorting": sortings,
+                "gt_sorting": gt_sortings,
+                "parameter_value": parameter_values,
+                "parameter_name": parameter_names,
+                "iter_num": iter_nums,
+                "method": methods,
+            }
+        )
+        return matching_df
```
samuelgarcia, line 413: Why a DataFrame here ?
Is it for further selection  on one paritcular columns ?

src/spikeinterface/sortingcomponents/benchmark/benchmark_matching.py
```diff
@@ -149,9 +536,10 @@ def plot_errors_matching(benchmark, unit_id, nb_spikes=200, metric="cosine"):
         ax.set_xlabel(metric)
 
         count += 1
+    return fig, axs
 
 
-def plot_errors_matching_all_neurons(benchmark, nb_spikes=200, metric="cosine"):
+def plot_errors_matching_all_neurons(benchmark, comp, nb_spikes=200, metric="cosine"):
```
samuelgarcia, line 542: Here we break a bit the previous behavior : a benchmark was self contaon andready for ploting.
Here the comp object must externaly given.
I guess we can refactor later.

src/spikeinterface/sortingcomponents/tests/test_benchmark_matching.py
```diff
@@ -0,0 +1,179 @@
...
+        benchmark.compare_all_sortings(pd.DataFrame(), ground_truth="invalid")
+
+
+if __name__ == "__main__":
+    test_run_matching_vary_parameter(benchmark_and_kwargs)
+    test_run_matching_vary_parameter_invalid_inputs(benchmark_and_kwargs)
+    test_run_matching_misclassed_invalid_inputs(benchmark_and_kwargs)
+    test_run_matching_missing_units_invalid_inputs(benchmark_and_kwargs)
+    test_compare_all_sortings(benchmark_and_kwargs)
+    test_compare_all_sortings_invalid_inputs(benchmark_and_kwargs)
```
samuelgarcia, line 179: I am OK to keep this file but I would skip this on github for now.
This will avoid deep and complicated errors when refactoring sorting component, I do not think we want to maintain also benchmark at the same time.

src/spikeinterface/sortingcomponents/benchmark/benchmark_matching.py
pauladkisson: What are the issues with this? I think it would be instructive for me if I knew exactly what you are concerned about.

src/spikeinterface/sortingcomponents/benchmark/benchmark_matching.py
pauladkisson: Yeah, but the previous behavior required the benchmark be associated with a particular comparison object, which limits what you can do with a given benchmark object.  I think it's better to have higher-level benchmarks that see multiple methods/comparisons to limit the duplicated work (generating waveform extractors, template similarity, etc.)

src/spikeinterface/sortingcomponents/tests/test_benchmark_matching.py
pauladkisson: Isn't it easier to maintain if you know when you introduce errors?

src/spikeinterface/sortingcomponents/benchmark/benchmark_matching.py
samuelgarcia: Lets leave this for now.
We will make a clean later.

src/spikeinterface/sortingcomponents/benchmark/benchmark_matching.py
samuelgarcia: This is a general problem acrross all benchmar class.
When exploring a grid parameters : some parts are fixed and some parts need recompute.
We should find a general way.
Lets do this later.

src/spikeinterface/sortingcomponents/tests/test_benchmark_matching.py
samuelgarcia: OK.
Lets keep this and I will make a decission on keep or not this tests running later.

samuelgarcia (approved):

PR #1626, opened 2023-05-10
Neo sorting refactor

h-mayorquin:
As discussed today, @samuelgarcia 

1) I've enhanced the documentation of the methods by adding more comprehensive docstrings. Additionally, I introduced a new class attribute, `need_t_start_from_signal_stream`, which should be set to True for those sorting extractors that require this specific logic. I also changed the name of  the `handle_spike_frame_directly` attribute as it did not have a reference to neo before which I found very confusing the first time I encountered.

2) I implemented a test to validate the accurate extraction of frames and timestamps. This test ensures that there are no spikes occurring within the same frame for a given unit - an error that arises if the format returns timestamps that haven't been converted to frames. Furthermore, when dealing with times, the test verifies that no spikes are closer than 1/sampling_rate, which is the minimum difference expected in timestamps.

3) This implementation necessitates a stream_index. Consequently, I've shifted the logic for checking the number of streams to the `__init__` methods of the classes that require it, namely Blackrock and Neuralynx. However, Blackrock is currently failing due to one of the tests involving more than one stream. This suggests that the logic for selecting the stream might need to be placed there. In fact, I'm considering whether methods requiring automatic guesses should perform this during the __init__ phase. @samuelgarcia, what are your thoughts on this? It might lead to code duplication, which is less than ideal. I'll continue to ponder this.

4) Finally, I've converted some methods into class methods and propagated them upwards to make them accessible in the `__init__` methods of the respective classes.


src/spikeinterface/extractors/tests/common_tests.py
```diff
@@ -117,12 +117,26 @@ def test_open(self):
...
+            sampling_frequency = sorting.get_sampling_frequency()
 
-            for segment_index in range(num_seg):
+            for segment_index in range(num_segments):
                 for unit_id in unit_ids:
-                    st = sorting.get_unit_spike_train(segment_index=segment_index, unit_id=unit_id)
+                    # Test that spike train has the propert units
+                    spike_train = sorting.get_unit_spike_train(segment_index=segment_index, unit_id=unit_id)
+                    sample_differences = np.diff(spike_train)
+                    no_spike_firing_in_the_same_frame = np.all(sample_differences > 0)
```
samuelgarcia, line 129: I am not sure that all formats ensure this.
Some sorting could make mistakes and return 2 spikes at the same sample index not ?

src/spikeinterface/extractors/neoextractors/blackrock.py
```diff
@@ -80,12 +82,27 @@ class BlackrockSortingExtractor(NeoBaseSortingExtractor):
 
     mode = "file"
     NeoRawIOClass = "BlackrockRawIO"
-    handle_spike_frame_directly = False
+    neo_returns_timestamps = True
     name = "blackrock"
 
-    def __init__(self, file_path, sampling_frequency=None):
+    def __init__(self, file_path, stream_index: Optional[int] = None):
```
samuelgarcia, line 88: I am not sure to be 100% happy with this because if you have a blackrock- spike files (nev) without any otehr file.
Then you cannot load it. And this stream_index expose to users seams a bit complicated.

src/spikeinterface/extractors/neoextractors/blackrock.py
```diff
@@ -80,12 +82,27 @@ class BlackrockSortingExtractor(NeoBaseSortingExtractor):
...
+            error_msg = (
+                "Black rock requires analog signal streams nsx5 or nsx6 to be present in for inferring spike times"
+            )
+
+            # Return the index of the first stream that is named nsx5 or nsx6 in that order
+            stream_index = next((index for index, name in enumerate(stream_names) if name == "nsx5"), None)
+            if stream_index is None:
+                stream_index = next((index for index, name in enumerate(stream_names) if name == "nsx6"), None)
+            assert stream_index is not None, error_msg
+            stream_index = 0
```
samuelgarcia, line 103: I do not understand this. Why reseting to this.

src/spikeinterface/extractors/neoextractors/blackrock.py
samuelgarcia: Wheras if the sampling_frequency is exposed then you will always be able to load it.

src/spikeinterface/extractors/neoextractors/neuralynx.py
```diff
@@ -55,12 +55,19 @@ class NeuralynxSortingExtractor(NeoBaseSortingExtractor):
 
     mode = "folder"
     NeoRawIOClass = "NeuralynxRawIO"
-    handle_spike_frame_directly = False
+    neo_returns_timestamps = True
+    need_t_start_from_signal_stream = True
     name = "neuralynx"
 
-    def __init__(self, folder_path, sampling_frequency=None):
+    def __init__(self, folder_path: str, stream_index=None):
```
samuelgarcia, line 62: same remarks than blackrock here.

src/spikeinterface/extractors/tests/common_tests.py
h-mayorquin: That could be true, but I am not too concerned. This code is meant only to be used with the data in our testing suite as a heuristic for proper time scaling. The code does not have too be valid for all the universe of possible spike sorting configurations (where your hypotethical is included).  Thi should  only detect changes in the code within our testing suite so we are alerted that the representation was broken.

Maybe there is a better way. I will think about it. Do you have a suggestion of how to better test this?

src/spikeinterface/extractors/neoextractors/blackrock.py
h-mayorquin: I need to check this again, this was a month ago and it is not fresh in my head. I will check on it on the weekend. What you say makes sense at this level though.  Thanks for taking a look,  Sam.

h-mayorquin: We all agree that `sampling_frequency` should be keept and have precedence.

h-mayorquin: OK, what are the errors. Hierarchy:
1) Only one stream use it.
2) Multiple streams same sampling frequency, use that one.
3) ...

src/spikeinterface/extractors/neoextractors/blackrock.py
h-mayorquin: Removed.

src/spikeinterface/extractors/neoextractors/neuralynx.py
h-mayorquin: Removed.

src/spikeinterface/extractors/neoextractors/blackrock.py
h-mayorquin: Implemented in the last commit.

src/spikeinterface/extractors/neoextractors/neobaseextractor.py
```diff
@@ -293,35 +294,125 @@ def __init__(self, block_index=None, sampling_frequency=None, use_natural_unit_i
...
                 t_start=t_start,
                 sampling_frequency=sampling_frequency,
-                handle_spike_frame_directly=self.handle_spike_frame_directly,
+                neo_returns_timestamps=self.neo_returns_timestamps,
             )
 
             self.add_sorting_segment(sorting_segment)
 
-    def _auto_guess_sampling_frequency(self):
+    def _get_t_start_from_signal_stream(self, segment_index):
```
h-mayorquin, line 302: @alejoe91 @samuelgarcia 
So, as per the comment  https://github.com/SpikeInterface/spikeinterface/pull/1626#discussion_r1250566361

The logic of sampling_frequency was changed.

However, getting the t_start from analog signal requires a stream_index.

Previously, this was handled by using the stream from which the sampling frequency was gotten. But now, this is not always possible. 

When this is not possible, what should we do?
@alejoe91 @samuelgarcia

src/spikeinterface/extractors/neoextractors/neobaseextractor.py
```diff
@@ -293,35 +294,125 @@ def __init__(self, block_index=None, sampling_frequency=None, use_natural_unit_i
...
+            stream_id_to_t_start[stream_id] = t_start
+
+        available_t_starts = list(stream_id_to_t_start.values())
+        unique_t_starts = set(available_t_starts)
+
+        if len(unique_t_starts) == 1:
+            t_start = available_t_starts[0]
+
+        else:
+            error_msg = (
```
h-mayorquin, line 335: Here:
@alejoe91 @samuelgarcia 
What if there is more than one t_start corresponding to the chosen sampling frequency?

h-mayorquin (commented):
Implemented the latest changes as requested, however, this left us with still one condition on what happens when the t_start is required.

src/spikeinterface/extractors/neoextractors/neobaseextractor.py
h-mayorquin: OK, after discussing with @samuelgarcia the logic would be to set if there is a single t_start that matches the frequency and otherwise set it to 0.

h-mayorquin:
This is ready.

src/spikeinterface/extractors/neoextractors/neobaseextractor.py
```diff
@@ -251,40 +254,40 @@ def get_traces(
...
-        # Get the stream index corresponding to the extracted frequency
         spike_channels = self.neo_reader.header["spike_channels"]
-
         if use_natural_unit_ids:
             unit_ids = spike_channels["id"]
             assert np.unique(unit_ids).size == unit_ids.size, "unit_ids is have duplications"
         else:
             # use interger based unit_ids
-            unit_ids = np.arange(spike_channels.size, dtype="int64")
+            unit_ids = np.arange(spike_channels.size, dtype="int32")
```
samuelgarcia, line 279: One day when we will have more than 2G units we will regret that choice :)

src/spikeinterface/extractors/neoextractors/neuralynx.py
```diff
@@ -55,13 +55,14 @@ class NeuralynxSortingExtractor(NeoBaseSortingExtractor):
...
+    neo_returns_timestamps = True
+    need_t_start_from_signal_stream = True
     name = "neuralynx"
 
-    def __init__(self, folder_path, sampling_frequency=None):
+    def __init__(self, folder_path: str, sampling_frequency: float | None = None):
         neo_kwargs = self.map_to_neo_kwargs(folder_path)
         NeoBaseSortingExtractor.__init__(self, sampling_frequency=sampling_frequency, **neo_kwargs)
-        self._kwargs.update(dict(folder_path=str(folder_path)))
+        self._kwargs.update(dict(folder_path=str(folder_path), sampling_frequency=self._sampling_frequency))
```
samuelgarcia, line 65: If we do that we can loose the t_start on the next reload from kwargs because setting frequency reset t_start.
Am I wrong ?

h-mayorquin: Yes, that's a good point. Don't know how to solve this in a good way other than exposing either t_start of the stream_index.

src/spikeinterface/extractors/neoextractors/neobaseextractor.py
h-mayorquin: x D

src/spikeinterface/extractors/neoextractors/blackrock.py
```diff
@@ -80,13 +82,14 @@ class BlackrockSortingExtractor(NeoBaseSortingExtractor):
...
+    neo_returns_timestamps = True
     name = "blackrock"
 
-    def __init__(self, file_path, sampling_frequency=None):
+    def __init__(self, file_path, sampling_frequency: float = None):
         neo_kwargs = self.map_to_neo_kwargs(file_path)
         NeoBaseSortingExtractor.__init__(self, sampling_frequency=sampling_frequency, **neo_kwargs)
-        self._kwargs.update({"file_path": str(file_path)})
+
+        self._kwargs.update({"file_path": file_path, "sampling_frequency": self._sampling_frequency})
```
alejoe91, line 92: Same here:
```suggestion
        self._kwargs.update({"file_path": file_path, "sampling_frequency": sampling_frequency})
```

src/spikeinterface/extractors/neoextractors/neuralynx.py
h-mayorquin: Good idea!

src/spikeinterface/extractors/neoextractors/blackrock.py
h-mayorquin: Done

src/spikeinterface/extractors/neoextractors/plexon.py
```diff
@@ -52,16 +52,13 @@ class PlexonSortingExtractor(NeoBaseSortingExtractor):
 
     mode = "file"
     NeoRawIOClass = "PlexonRawIO"
-    handle_spike_frame_directly = True
```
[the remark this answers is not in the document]
h-mayorquin: Done, let me know if I missed anyone.

h-mayorquin:
I added the warnings and exposed stream_index and stream_ids again. Checked the last commit.

alejoe91:
Neuralynx test is failing @h-mayorquin !

h-mayorquin:
OK, it was a kwargs issue. 

I did one last change, I changed `neo_return_timestamps` to `neo_return_frames` the reson being that it should be made very clear that the function `neo_reader.get_spike_timestamps` does NOT always return timestamps. 

This changes keeps all the logic explicit as it was before but also serves to highlight this fact.

src/spikeinterface/extractors/neoextractors/neobaseextractor.py
```diff
@@ -251,77 +330,237 @@ def get_traces(
...
+        stream_id: Optional[str] = None,
+        stream_name: Optional[str] = None,
+        **neo_kwargs,
+    ):
         _NeoBaseExtractor.__init__(self, block_index, **neo_kwargs)
 
-        self.use_natural_unit_ids = use_natural_unit_ids
+        # Get stream_id from stream_name
+        if stream_name:
+            stream_id = self.build_stream_name_to_stream_id_dict()["stream_name"]
```
samuelgarcia, line 359:
```suggestion
            stream_id = self.build_stream_name_to_stream_id_dict()[stream_name]
```

src/spikeinterface/extractors/neoextractors/neobaseextractor.py
```diff
@@ -251,77 +330,237 @@ def get_traces(
 
 
 class NeoBaseSortingExtractor(_NeoBaseExtractor, BaseSorting):
-    # this will depend on each reader
-    handle_spike_frame_directly = True
+    neo_returns_frames = True
```
samuelgarcia, line 333: definitivelly a better name!!!

src/spikeinterface/extractors/neoextractors/neobaseextractor.py
```diff
@@ -330,44 +569,36 @@ def __init__(
...
         self._t_start = t_start
         self._sampling_frequency = sampling_frequency
-        self._natural_ids = None
-        self.handle_spike_frame_directly = handle_spike_frame_directly
+        self.neo_returns_frames = neo_returns_frames
 
-    def get_natural_ids(self):
-        if self._natural_ids is None:
-            self._natural_ids = list(self._parent_extractor().neo_reader.header["spike_channels"]["id"])
-        return self._natural_ids
```
samuelgarcia, line 351: I guess the ida of this function was to delay construction of the dict to make the init faster.
But this is a details. Lets keep this change.

samuelgarcia:
This was long and tedious but now I guess this implementation is really really better than before.
Thanks a a lot for for this!

samuelgarcia (approved):

h-mayorquin:
Thanks @samuelgarcia for all the help. The joke is that the name of the branch is "small_neo_sorting_refactor" x D

PR #1628, opened 2023-05-11
Adapt `CellExplorerSortingExtractor` to new format

h-mayorquin:
Currently we are doing a conversion https://github.com/catalystneuro/buzsaki-lab-to-nwb/issues/55 where we need to adapt this. Basically, there is a new format change where the sampling frequency is in the same files where the spikes are (!). I am adapting here the sorter in consequence.

Plus, I changed the non-standard name of the input so it plays along with the conversion in the repo and delayed the imports to run time. I also managed to eliminate the dependency.

### DIFF
--- a/src/spikeinterface/extractors/cellexplorersortingextractor.py
+++ b/src/spikeinterface/extractors/cellexplorersortingextractor.py
@@ -11,13 +11,13 @@ class CellExplorerSortingExtractor(BaseSorting):
     """
     Extracts spiking information from .mat files stored in the CellExplorer format.
     Spike times are stored in units of seconds so we transform them to units of samples.
-    
+
     The newer version of the format is described here:
     https://cellexplorer.org/data-structure/
-    
+
     Whereas the old format is described here:
     https://github.com/buzsakilab/buzcode/wiki/Data-Formatting-Standards
-    
+
     Parameters
     ----------
     folder_path: str | Path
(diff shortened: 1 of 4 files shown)


src/spikeinterface/extractors/cellexplorersortingextractor.py
```diff
@@ -1,136 +1,163 @@
...
-    installed = HAVE_SCIPY_AND_HDF5STORAGE
     is_writable = True
-    mode = "file"
-    installation_mesg = "To use the CellExplorerSortingExtractor install scipy and hdf5storage: \n\n pip install scipy  hdf5storage"
+    mode = "folder"
+    installation_mesg = "To use the CellExplorerSortingExtractor install scipy and h5py"
+
+    def __init__(
+        self,
+        folder_path: str | Path,
```
CodyCBakerPhD, line 38: Does the `|` operator properly equate to `Union` when using annotations?

CodyCBakerPhD, line 38: Technically breaking back-compatability in switching from file to folder mode; is this actually necessary to accomodate different versions of the associated header file (sessionInfo.mat vs. sessions.mat)?

h-mayorquin: Yep.

src/spikeinterface/extractors/cellexplorersortingextractor.py
```diff
@@ -1,136 +1,163 @@
...
-        ), f"The spikes_matfile_path ({spikes_matfile_path}) must exist!"
+        spikes_cellinfo_path = self.folder_path / f"{self.session_id}.spikes.cellinfo.mat"
+        assert spikes_cellinfo_path.is_file(), f"The spikes.cellinfo.mat file must exist in the {self.folder_path}!"
         
+        read_as_hdf5 = False 
+        try:
+            spikes_mat = scipy.io.loadmat(file_name=str(spikes_cellinfo_path), simplify_cells=True)
+        except NotImplementedError:
+            spikes_mat = h5py.File(name=spikes_cellinfo_path, mode="r")
+            read_as_hdf5 = True
```
CodyCBakerPhD, line 63: A fair bit cleaner this way; there may be some minor syntax changes between `hdf5storage` and `h5py` when accessing array fields however

src/spikeinterface/extractors/cellexplorersortingextractor.py
h-mayorquin: https://peps.python.org/pep-0604/

src/spikeinterface/extractors/cellexplorersortingextractor.py
CodyCBakerPhD: Hoping the 2 cell explorer datasets on GIN are representative of what you might see with those differences

src/spikeinterface/extractors/cellexplorersortingextractor.py
```diff
@@ -1,136 +1,163 @@
...
+        unit_ids_available = "UID" in spikes_mat["spikes"].keys()
+        assert unit_ids_available, "The spikes.cellinfo.mat file must contain field 'UID'!"
 
-        try:
-            spikes_mat = scipy.io.loadmat(file_name=str(spikes_matfile_path))
-            self.read_spikes_info_with_scipy = True
-        except NotImplementedError: 
-            spikes_mat = hdf5storage.loadmat(file_name=str(spikes_matfile_path))
-            self.read_spikes_info_with_scipy = False
+        spike_times_available = "times" in spikes_mat["spikes"].keys()
```
CodyCBakerPhD, line 78: Are the access patterns as dictionaries actually equivalent for both backends? (scipy vs. h5py)

I had thought that was a bigger part of the previous design

src/spikeinterface/extractors/cellexplorersortingextractor.py
h-mayorquin: If they keep changing the format it is easier to get the top folder and modify the internal function to find the necessary files. My currrent take is that the price of backwards compatibility is pay to built a less brittle approach.

src/spikeinterface/extractors/cellexplorersortingextractor.py
```diff
@@ -1,136 +1,163 @@
...
-        assert np.all(
-            np.isin(["UID", "times"], spikes_mat["spikes"].dtype.names)
-        ), "The spikes.cellinfo.mat file must contain a 'spikes' struct with fields 'UID' and 'times'!"
+        unit_ids = spikes_mat["spikes"]["UID"]
+        spike_times = spikes_mat["spikes"]["times"]
+        
+        if read_as_hdf5:
+            unit_ids = unit_ids[:].squeeze().astype("int")
+            spike_references = spike_times[:]  # These are HDF5 references
+            spike_times = [spikes_mat[reference[0]][()].squeeze() for reference in spike_references]
```
CodyCBakerPhD, line 87: Oh I see, conditional reshaping is done after the fact... But the availability of outer keys does not change from scipy vs h5py

src/spikeinterface/extractors/cellexplorersortingextractor.py
h-mayorquin: Yep, I think it is a little bit cleaner. To be honest, when I started doing this earlier this afternoon I had the hope that I would be able to keep the spikes acess lazy for the case when matlab new format is used (something that hdf5 storage does not do). It was a little bit more complicated than I thought. I might come back to this if I have more time when we are doing the full conversion with all the sessions.

src/spikeinterface/extractors/cellexplorersortingextractor.py
h-mayorquin: scipy with the keyword argument `simplify_cells` returns dictionaries. The h5py returns groups who have keys as well.

src/spikeinterface/extractors/cellexplorersortingextractor.py
CodyCBakerPhD: Nice!

src/spikeinterface/extractors/cellexplorersortingextractor.py
h-mayorquin: We have three datasets that cover all the cases for the previous format (sessionInfo in new vs old format and spikes.cellinfo in new vs old format). 

This is thanks to you btw, @CodyCBakerPhD . You made me do it back in the day : P and then you went and reduce the file to upload it to gin.

I also have a file from the repo with the new format that I tested locally.

src/spikeinterface/extractors/cellexplorersortingextractor.py
CodyCBakerPhD: I'd actually say folder formats are way more brittle; they are very easy to break by simply moving the files around, which is something that has been observed previously

Whereas having different arguments for the different required (or optional) file paths enables more degrees of freedom for the user that aren't dependent on following a strict folder pattern

h-mayorquin: You have a point there. I will think more about this.

h-mayorquin: OK, I reverted to file mode. I still want to standarize this with the other sorters so I flaged the old name for the file `spikes_matfile_path` with a deprecation warning for six months. None of this should affect neuroconv though, as the API over there does expose `file_path` instead of the soon to be deprecated `spikes_matfile_path`.

h-mayorquin: Note that you need `from __future__ import annotations` at the top of the file before python 3.10 though.

src/spikeinterface/extractors/cellexplorersortingextractor.py
h-mayorquin: I also added the third datasets to the tests.

src/spikeinterface/extractors/cellexplorersortingextractor.py
```diff
@@ -1,136 +1,206 @@
...
+    https://cellexplorer.org/data-structure/
+
+    Whereas the old format is described here:
+    https://github.com/buzsakilab/buzcode/wiki/Data-Formatting-Standards
 
     Parameters
     ----------
-    spikes_matfile_path : PathType
-        Path to the sorting_id.spikes.cellinfo.mat file.
+    file: str | Path
```
CodyCBakerPhD, line 25:
```suggestion
    file_path: str | Path
```

src/spikeinterface/extractors/cellexplorersortingextractor.py
```diff
@@ -1,136 +1,206 @@
...
-                 session_info_matfile_path: OptionalPathType=None,
-                 sampling_frequency: Optional[float] = None):
-        assert self.installed, self.installation_mesg
+    installation_mesg = "To use the CellExplorerSortingExtractor install scipy and h5py"
+
+    def __init__(
+        self,
+        file_path: str | Path | None = None,
+        sampling_frequency: float | None = None,
+        session_info_matfile_path: str | Path | None = None,
```
CodyCBakerPhD, line 42: Since not shying away from deprecation, maybe

`session_info_matfile_path` -> `session_info_file_path`

for consistency?

src/spikeinterface/extractors/cellexplorersortingextractor.py
h-mayorquin: Thanks.

src/spikeinterface/extractors/cellexplorersortingextractor.py
h-mayorquin: Yes. I think that makes sense.

src/spikeinterface/extractors/shybridextractors.py
```diff
@@ -92,6 +92,15 @@ def write_recording(recording, save_path, initial_sorting_fn, dtype='float32', v
...
         **write_binary_kwargs: keyword arguments for write_to_binary_dat_format() function
         """
+        try:
+            import hybridizer.io as sbio
+            import hybridizer.probes as sbprb
+            import yaml
+
+            HAVE_SBEX = True
+        except ImportError:
+            HAVE_SBEX = False
```
CodyCBakerPhD, line 102: Were these changes supposed to be on a different branch? Not that it matters much

CodyCBakerPhD (approved):

CodyCBakerPhD:
@h-mayorquin This LGTM to me, thanks for working on it!

src/spikeinterface/extractors/shybridextractors.py
h-mayorquin: I am slowly moving imports inwards. Probably should be in a different branch but I don't think it matters that much.

h-mayorquin:
If you agree with this, maybe @alejoe91 can merge it?

h-mayorquin:
pre-commit.ci autofix

CodyCBakerPhD (approved):

alejoe91 (approved):

PR #1632, opened 2023-05-12
Add pre-commit-config file

alejoe91:
This PR adds the `.pre-commit-config.yaml` file to prepare for the pre-commit bot.

PR #1633, opened 2023-05-12
Black formatting

alejoe91:
Reformat `src` files in black style (line length 120)

This should be merged after #1632 and after enabling the pre-commit bot, to make sure it successfully runs (in this case, the bot should not apply any changes!)


h-mayorquin:
Let me know when you enable the pre-commit bot. 

To ease the pill, if you really dislike what black does in one line or block you can opt-out with:

```
Black reformats entire files in place. It doesn’t reformat lines that end with # fmt: skip or blocks that start with # fmt: off and end with # fmt: on. # fmt: on/off
```
Second paragraph here:
https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#code-style

samuelgarcia:
@alejoe91 @h-mayorquin :
before merging could we have a clear view of the plan for all already open PRs ?
does a simple `git pull upsteram main` will be enough or will it be a infinit pain ?
Could you simulate a try a afew PRs that change many files ?

alejoe91:
Hi @samuelgarcia 

What @CodyCBakerPhD suggested is the following: 

1. We update open PRs so that they get #1632
2. We activate `pre-commit CI`, so that we don't even have to manually reformat
3. We merge this

@h-mayorquin what do you think?

h-mayorquin (approved):

PR #1635, opened 2023-05-13
Enable filtering warnings for distutils and imp

h-mayorquin:
Datalad, setuptools and boto are using to-deprecated methods and this is shown whenever someone wants to run a script.  We have some lines before in the `__init__` but:
1) They are not working for me (I think it should be `ignore` instead of `always`)
2) They are too general I feel. We might want to know about other cases.

This should solve this.


h-mayorquin:
pre-commit.ci autofix

h-mayorquin:
pre-commit.ci autofix

alejoe91 (approved):

PR #1638, opened 2023-05-15
Make `RemoveDuplicatedSpikesSortingSegment` lazy.

h-mayorquin:
This comes from discussing with @DradeAW on https://github.com/SpikeInterface/spikeinterface/pull/1622. Just putting here as a suggestion right now.

Right now this curation calculates all the duplicated spikes at `__init__` the PR here changes this to have a dict that will cache the values as they are produced so this can be initialized lazily.


src/spikeinterface/curation/remove_duplicated_spikes.py
```diff
@@ -39,48 +39,60 @@ def __init__(self, sorting: BaseSorting, censored_period_ms: float = 0.3, method
...
-                                       censored_period, method=method, seed=seed)
-
+        self._duplicated_spikes = {unit_id: None for unit_id in unit_ids}
 
-    def get_unit_spike_train(self, unit_id, start_frame: Optional[int] = None,
-                             end_frame: Optional[int] = None) -> np.ndarray:
+    def get_unit_spike_train(
+        self, unit_id, start_frame: Optional[int] = None, end_frame: Optional[int] = None
+    ) -> np.ndarray:
         spike_train = self._parent_segment.get_unit_spike_train(unit_id, start_frame=None, end_frame=None)
```
DradeAW, line 74: Won't that be very heavy in the case of chunking?

h-mayorquin: Good point. Is curation stuff passed to `ChunkRecordingExecutor`, we can use _kwargs to avoid that.

h-mayorquin:
pre-commit.ci autofix

h-mayorquin:
I am closing this. I think this requires a more general solution.

PR #1641, opened 2023-05-15
Run `auto_merge` with un-ran `WaveformExtractor`

DradeAW:
WIP to be able to run `curation.auto_merge` with a `WaveformExtractor` that has not be run.

The idea behind it is that the templates are needed at the end, and by then, usually more then half the units aren't considered for merging.
This results in a lot of time computing templates for nothing.

The idea is to be able to give a un-ran `WaveformExtractor` object to the function, and it will run it on the subset of units if needed.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -141,13 +141,17 @@ spikeinterface.preprocessing
 
 .. automodule:: spikeinterface.preprocessing
 
+    .. autofunction:: astype
+    .. autofunction:: average_across_direction
     .. autofunction:: bandpass_filter
     .. autofunction:: blank_staturation
     .. autofunction:: center
     .. autofunction:: clip
     .. autofunction:: common_reference
     .. autofunction:: correct_lsb
+    .. autofunction:: depth_order
     .. autofunction:: detect_bad_channels
+    .. autofunction:: directional_derivative
     .. autofunction:: filter
     .. autofunction:: gaussian_bandpass_filter
     .. autofunction:: highpass_filter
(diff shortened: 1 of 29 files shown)


DradeAW:
This is done on my side,

Not sure about the `WaveformExtractor.was_run` though, there may be a way to define a function that checks it without making a variable?
Not sure how to make that though since you have to check memory, folder, zarr ...

DradeAW:
Can someone re-run the test please?
It failed because it couldn't connect to a server, and I don't have the rights to re-run a failed test.

Thanks :)

samuelgarcia:
Salut Aurelien,
I am not sure to be agree on all detail of this story.
1. You are in some cases doing an in-place modification of the waveforms extractor( you change the sorting)
2. You are overwritting some parameters : 100 spikes.

I totally see the goal (being) faster.

What I would do is:
  * check at the very begenning that we waveform is is not run. And if need to be be run : cerate a new one and then run on it what you need depending the step and than no sparse hactch at eevry steps.

DradeAW:
This is also what I had in mind in the beginning, but then realized that if we want to make a new WaveformExtractor, then the `mode="folder"` becomes a little tricky.

How do you propose we fix this?

DradeAW:
> This is also what I had in mind in the beginning, but then realized that if we want to make a new WaveformExtractor, then the mode="folder" becomes a little tricky.

@samuelgarcia Any idea?

zm711:
@DradeAW, this can probably be closed now with the move to `SortingAnalyzer` no?

samuelgarcia:
Yes it is.
And this original was one of the reason to make the SortingAnalyzer.

PR #1642, opened 2023-05-15
Minimal change

h-mayorquin:
This should activate the commit bot. Let's test.


CodyCBakerPhD:
FYI this PR will not result in the pre-commit running since it's from an external fork

In that event the pre-commit bot essentially acts like a normal linter if it encounters format errors; the user should run pre-commit manually on their local branch in that event

alejoe91:
Ahhh! So pre-commit only acts on internal branches?? That explains it!

alejoe91:
@h-mayorquin can you make a minimal PR from an internal branch?

PR #1644, opened 2023-05-15
Add typing to `download_dataset`

h-mayorquin:
This is just a small PR where I tested the hook.

### DIFF
--- a/src/spikeinterface/core/datasets.py
+++ b/src/spikeinterface/core/datasets.py
@@ -8,7 +8,7 @@ import warnings
 from .globals import get_global_dataset_folder, is_set_global_dataset_folder
 
 
-def download_dataset(repo=None, remote_path=None, local_folder=None, update_if_exists=False, unlock=False):
+def download_dataset(repo=None, remote_path=None, local_folder=None, update_if_exists=False, unlock=False) -> Path:
     """
     Function to download dataset from a remote repository using datalad.
 


src/spikeinterface/core/datasets.py
```diff
@@ -59,7 +66,7 @@ def download_dataset(repo=None, remote_path=None, local_folder=None, update_if_e
 
     if remote_path is None:
         warnings.warn(message="No remote path provided, returning None")
-        return
+        return None  # TODO: is this the best option? I Think this should not accept remote_path is None.
```
alejoe91, line 69: Not sure, but let's just add typing in this PR (I'll leave the comment in)

h-mayorquin: Yes, I find accepting remote_path as None very strange as the function just kicks you out and throws a warning.

Probably was done as a hack for something else at some point though.

alejoe91: Should we just raise an exception?

h-mayorquin: Yes, I think would be better.

But I think it would be even better if `remote_path` would be a required argument in the signature. 

The `repo` keyword in the signature is set automatically to:

https://github.com/SpikeInterface/spikeinterface/blob/0468095ed194755ebec26fa22e8ec814c271ce64/src/spikeinterface/core/datasets.py#L46-L49

So, it makes sense to have `repo` as an optional argument. 

But the `remote_path` keyword is determines the dataset that you want to download. I don't see why it makes sense to pass no dataset. 

I think we should the warning to a deprecation warning and then at some point make `remote_path` a required argument.

src/spikeinterface/core/datasets.py
```diff
@@ -1,14 +1,21 @@
 """
 Some simple function to retrieve public datasets with datalad
 """
+from __future__ import annotations
 
 from pathlib import Path
 
 import warnings
-from .globals import get_global_dataset_folder, is_set_global_dataset_folder
+from .globals import get_global_dataset_folder
```
alejoe91, line 10:
```suggestion
DEFAULT_REPO = "gin.gnode..."
DEFAULT_REMOTE_PAT= "mearec/mearec_10s.h5"
```

src/spikeinterface/core/datasets.py
```diff
@@ -1,14 +1,21 @@
...
 
 
-def download_dataset(repo=None, remote_path=None, local_folder=None, update_if_exists=False, unlock=False):
+def download_dataset(
+    repo: str | None = None,
+    remote_path: str | None = None,
```
alejoe91, line 14:
```suggestion
    repo: str  = DEFAULT_REPO,
    remote_path: str = DEFAULT_REMOTE_PATH,
```

src/spikeinterface/core/datasets.py
h-mayorquin: Thinking about twice I think having the str as a default argument as you initially proposed is better.

alejoe91 (approved):

samuelgarcia (approved):

PR #1652, opened 2023-05-16
Allow to pass `sorting` to `qualitymetrics.compute_refrac_period_violations`

DradeAW:
Since the recording is not necessary, allow both `wvf_extractor` and `sorting` objects as input


DradeAW:
Ah actually you do need the recording to know the total number of samples

alejoe91:
> Ah actually you do need the recording to know the total number of samples

You should have the get num samples and total samples available directly at the vaveform extractor level even in recording less mode ;)

PR #1655, opened 2023-05-16
Documentation-Fix RST formatting Typos

zm711:
Was looking at docs and noticed a couple places where the rst wasn't formatted. Just fixed.

### DIFF
--- a/doc/development/development.rst
+++ b/doc/development/development.rst
@@ -17,9 +17,9 @@ There are various ways to contribute to SpikeInterface as a user or developer. S
 We use a forking workflow <https://www.atlassian.com/git/tutorials/comparing-workflows/forking-workflow>_ to manage contributions. Here's a summary of the steps involved, with more details available in the provided link:
 
 * Fork the SpikeInterface repository.
-* Create a new branch (e.g., :code:git switch -c my-contribution).
+* Create a new branch (e.g., :code:`git switch -c my-contribution`).
 * Modify the code, commit, and push changes to your fork.
-* Open a pull request from the "Pull Requests" tab of your fork to :code:spikeinterface/main.
+* Open a pull request from the "Pull Requests" tab of your fork to :code:`spikeinterface/main`.
 * By following this process, we can review the code and even make changes as necessary.
 
 While we appreciate all the contributions please be mindful of the cost of reviewing pull requests <https://rgommers.github.io/2019/06/the-cost-of-an-open-source-contribution/>_ .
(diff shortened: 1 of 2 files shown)


alejoe91 (approved):
Thanks!

doc/development/development.rst
```diff
@@ -164,7 +164,7 @@ As described in the `black documentation <https://black.readthedocs.io/en/stable
 
 The following are some styling conventions that we follow in SpikeInterface:
 
-#. Avoid using abreviations in variable names (e.g., use :code:`recording` instead of :code:`rec`). It is specially important to avoid single letter variables.
+#. Avoid using abreviations in variable names (e.g., use :code:`recording` instead of :code:`rec`). It is especially important to avoid single letter variables.
```
h-mayorquin, line 167: This was on me. I was not aware of that distinction. Thanks!

zm711: @h-mayorquin In American English they are actually used pretty interchangeable, but the unofficial rule is specially is used in spoken and especially in written English. It's just convention though.

h-mayorquin: Thanks. Useful to know.

h-mayorquin:
I have a generla question @zm711 , what ares of the documentation do you think are lacking in terms of content? What do you tihnk would be (or would have been) helpful to add?

zm711:
@h-mayorquin That's an interesting question. I think the beginner section for install could actually be moved to the docs. Currently it links to the Github folder and if someone is a beginner the jump to all the files and markdown might already be a scary leap and they are starting with python (or moving from Matlab where the install is all behind the scenes/toolkits). The number of times someone (myself included when I started) installed a scientific program in base instead of an environment is unfortunate. So I would like to have at least conda instructions (since most scientists start with conda/minconda) for creating a virtual environment in the beginner docs.

I think the API docs needs to be updated (ie we need to keep adding docstrings). Since most IDEs will include docstring info it is actually really helpful to know if a function consumes a `Sorting` or `WaveformExtractor`. And if the docstring is unclear a beginner would go to the documentation only to see that the same lack of info is present in the API section. (More advanced would just go to source to fill in the gaps).

Based on the questions/issues. I think an explanation of lazy and what parts of spikeinterface are lazy and when someone should save is important. I know Alessio is adding a piece to the documentation, but scientists not coming from any computer background might not really know disk vs RAM and the benefits of using one or the other. So a fully fleshed out section on some of these "computer science" tricks that si uses behind the scenes might help newbies understand someone of the errors since currently it is hard to know when to switch. 

I think a final lower priority section needs to explain the ProbeInterface and Neo portions. I think someone of the Neo terminology makes sense once you are used to it, but an explanation might be more beneficial. Currently there is just a link to the Neo page. But if any Neo api is being exposed through the si api then documentation will make it less likely for beginners to make mistakes. Same for ProbeInterface, some recording systems have info and some don't and I think one fundamental issue when starting spike sorting is the realization that the numbers on the screen aren't the same as the numbers on the actual electrode due to wiring. The ProbeInterface documentation covers that fact, but I think a small section here explaining that as well would be beneficial. More like an FYI: did you know that without a wiring diagram you might not be sorting what you think? Then a brief explanation what is meant by wiring and then link to ProbeInterface for deeper info.

Which leads to just more examples. The example has one mega example (+Neuropixel example). So having some more troubleshooting examples of each module. There are spiketutorials, but making that clearer in the documentation or including them in the docs might be helpful. Some people don't read the docs they just go to juypter and try to run the examples, but then they use a different recording equipment or they want to use a different sorter and now it doesn't work. So they hop to another example hoping to fill in the pieces. It's just a different learning style, so having a richer example section might help the "learn by doing" people.

Just my two cents though. Hopefully something out of that is helpful.

h-mayorquin:
This is SUPER USEFUL.

I will open and issue  or discussion and expand upon this with some ideas that I have.

Thanks a bunch for taking the time to write @zm711

samuelgarcia:
@zm711 merci beaucoup.
We should organize doc sprint soon.

PR #1658, opened 2023-05-17
Allow `new_unit_id` to be same id in `CurationSorting.merge`

DradeAW:
Right now it's not possible to do `CurationSorting.merge([1, 2], new_unit_id=1)`, and it should be possible

### DIFF
--- a/src/spikeinterface/curation/curationsorting.py
+++ b/src/spikeinterface/curation/curationsorting.py
@@ -80,7 +80,7 @@ class CurationSorting:
         current_sorting = self._sorting_stages[self._sorting_stages_i]
         if new_unit_id is None:
             new_unit_id = self._get_unused_id()[0]
-        else:
+        elif new_unit_id not in units_to_merge:
             assert new_unit_id not in current_sorting.unit_ids, f"new_unit_id already exists!"
         new_sorting = MergeUnitsSorting(
             parent_sorting=current_sorting,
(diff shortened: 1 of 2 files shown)


zm711:
@DradeAW Isn't it a `Phy` convention that the numbers always need to get bigger? So I think if you allow this it would break Phy compatibility if people were going back and forth been SI and Phy.

DradeAW:
But I'm not changing the way it works by default, I'm just giving the user the opportunity to do so if he wishes.

In my code it's important that the unit id stays the same!

zm711:
Totally fair. So the only problem case would be the case where someone wanting to use Phy but not knowing the Phy rules tries to merge to the same number. I don't think there's an easy way to do a warning without bothering users. But if a user just sticks to default it should be fine (so someone not knowing Phy rules would hopefully just use default?) 

I'm understanding that correctly?

DradeAW:
I don't see what the problem is?

You can use `CurationSorting.merge([1, 2], new_unit_id=1)` then `export_to_phy` and it will work perfectly.

I think I'm not understanding something here...

zm711:
Definitely my poor explanation of the edge case I was thinking of. Sorry. 

I was imagining someone who already tried some curation in Phy, then decided to curate with SI and then tried to go back to Phy to visualize how their merges went and did something like 1+2 in Phy equals 10 and then in SI they want to merge 3 into 10 but decide to shrink the number back down to 1, 2, or 3. But maybe that is such an impossible case it doesn’t matter. (I haven’t tested this so really don’t know if it would even matter or whether SI internally fixes the issues).

alejoe91 (approved):

DradeAW:
Thanks!

PR #1659, opened 2023-05-17
Add censorship of correlograms in `scur.auto_merge`

DradeAW:
Some sorters can detect spikes multiple times.
This results in peaks in the auto and/or cross-correlograms that can impact detection of merged units.
This should fix it.


samuelgarcia:
@DradeAW : I understand the goal but this also artificialy break the main idea of the correlogram no ?
for duplicated spikes it would be better to be explicit.

I would not put censor_correlograms_ms=1.5 by default.
Are you targeting a few sample for duplicated  spikes  ?

DradeAW:
> I would not put censor_correlograms_ms=1.5 by default
I put 0.15 by default! (it can be reduced to 0.10)

> I understand the goal but this also artificialy break the main idea of the correlogram no ?
I don't think it really break the idea because there is a gaussian filtering afterwards that takes into consideration a much bigger window that the censorship window.
Now I'm not against putting a value of 0.0 by default if you think it's best :)

samuelgarcia:
>I put 0.15 by default! (it can be reduced to 0.10)
oups sorry I read it to quicly.
make more sens
Lets keep this default
I still think that we should clean duplicated spikes before this.

DradeAW:
Cleaning duplicated spikes will work for auto-correlograms, but not for cross-correlograms!

Some sorters can detect the same spike multiple time and assign it in different units.

samuelgarcia (approved):

PR #1660, opened 2023-05-18
Added some typing in `WaveformExtractor`

DradeAW:
Added some typing in `WaveformExtractor` + fixed a typo

### DIFF
--- a/src/spikeinterface/core/waveform_extractor.py
+++ b/src/spikeinterface/core/waveform_extractor.py
@@ -121,7 +121,7 @@ class WaveformExtractor:
         return txt
 
     @classmethod
-    def load(cls, folder, with_recording=True, sorting=None) -> WaveformExtractor:
+    def load(cls, folder, with_recording=True, sorting=None) -> 'WaveformExtractor':
         folder = Path(folder)
         assert folder.is_dir(), "Waveform folder does not exists"
         if folder.suffix == ".zarr":
@@ -130,7 +130,7 @@ class WaveformExtractor:
             return WaveformExtractor.load_from_folder(folder, with_recording=with_recording, sorting=sorting)
 
     @classmethod
-    def load_from_folder(cls, folder, with_recording=True, sorting=None) -> WaveformExtractor:
+    def load_from_folder(cls, folder, with_recording=True, sorting=None) -> 'WaveformExtractor':
         folder = Path(folder)
         assert folder.is_dir(), f"This waveform folder does not exists {folder}"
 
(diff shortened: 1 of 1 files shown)


alejoe91 (approved):

alejoe91 (approved):

src/spikeinterface/core/waveform_extractor.py
```diff
@@ -74,9 +74,9 @@ def __init__(
         sorting: BaseSorting,
         folder=None,
         rec_attributes=None,
-        allow_unfiltered=False,
+        allow_unfiltered: bool = False,
         sparsity=None,
-    ):
+    ) -> None:
```
h-mayorquin, line 79: I find that is not settled whether`__init__` method returns should be annotated. How do you think about it?

alejoe91: Oups! It slipped! @h-mayorquin if I understand correctly you mean that we don't need this for the init, right?

DradeAW: Yes I don't remember where it is, but it is considered good practice to annotate `__init__() -> None:`

h-mayorquin: I think it does not matter as it is redundant for most purposes as far as a I know. The `__init__` should always return None and static type checkers will ignore it. I was curios if @DradeAW knew something about it or was doing it for a specific purpose.

h-mayorquin: The discussion around it is a bit confusiong though.
https://stackoverflow.com/questions/46779046/correct-type-annotation-for-init

DradeAW: No nothing specific. Just following the PEP 484 "Note that the return type of `__init__` ought to be annotated with `-> None`"

If you'd rather not have it, no problem on my end :)

h-mayorquin: Na, that's fine. It does not take and it does not add as I view it.

As far as a Iunderstand that was done because mipy was not checking `__init__` at some point but then Guido jumped in and they changed that. After that, it is not necessary but it could be that another type checker comes in the future that does not make that assumption.

DradeAW: > The discussion around it is a bit confusiong though.
> https://stackoverflow.com/questions/46779046/correct-type-annotation-for-init

Apparently it is necessary when no augments are annotated, but optional if at least one argument is annotated (based on `mypy`).
Personally I think if it's required in some cases, it's better to have it in all of them for consistency. But this is a personal opinion, and I can live without it ^^

h-mayorquin: Sounds reasonable.

samuelgarcia: very intersting discussion.

Issue #1432, opened 2023-03-20
Force `_main_ids` to be strings


alejoe91:
Related to the serialization thread (#1397), after discussing with @h-mayorquin, we thought that casting `Base._main_ids` as `str` **always** may solve several issues with keys not being json serializable.
Allowing ids to be integers is mainly to maintain a similar behavior with `spikeextractors`, but now that ids and indices are clearly separated I think we can just use strings.

@samuelgarcia thoughts?

magland:
I understand why forcing IDs to be str is desirable. However, I believe this would break many of the recordings/sortings that I work with, that use integer IDs. So not sure what is the best way forward here.

alejoe91:
@magland could you elaborate why they would break? If you use, for instance, channel IDs to index or slice an array, that is not the way it should be because we have a clear api distinction between IDs and indices now. Do you have other breaking examples?

magland:
I have a lot of existing code that uses integers for IDs when creating recordings/sortings. Frank lab also has recordings/sortings stored to disk in custom formats that I believe use integers for IDs.

magland:
For example, here's a line of code that I just wrote:

```
recording_signal = se.NumpyRecording([signal], sampling_frequency=recording_sub_whitened.sampling_frequency, channel_ids=np.arange(1, KK + 1, dtype=np.int32))
```

It would be cumbersome to convert that code to use strings instead of ints.

alejoe91:
No we would do the conversion in the base object and under the hood. So if the downstream code does not depend on the IDs being int there's no problem

magland:
So we'd be able to pass in ints for any of the API functions and they get translated to strings under the hood?

alejoe91:
Well normally you would then the `channel_ids` from the object and pass them around!

alejoe91:
I see! I agree it's better going forward. Probably also better to switch sooner than later!

magland:
> Maybe it should wise to deprecate explcitly in next release (actual master) and remove ids integer in the following one. So anyone could figure out which part of code will be broken in advance. This would be more smart no ?

I think that would be problematic, since you'd get a bunch of deprecation warnings for things that aren't really going to be problems. Like if I set the unit IDs as integers, that is still going to be supported according to @alejoe91. What you can't test for is how the user interprets the output of `.unit_ids` or `.channel_ids`.

alejoe91:
Yep ;)

samuelgarcia:
[...]

PR #1661, opened 2023-05-18
Added some typing in `BaseExtractor`

DradeAW:
I figured since this is kind of the 'root' class, it needed some typing.

I didn't put them all, but there are several.
Feel free to add more if you want.


DradeAW:
I was thinking also: we could make `ID_Type = Union[int, str]`, a type dedicated to what `unit_id` and `channel_id` can be.

That way if we want to add a new accepted type for ids, we can change just `ID_Type` and it will propagate.

What do you think?

DradeAW:
> you have a phd manuscript to write you not waste your time on python typing. Life is short.

I agree it's not the most effective use of my time, but my brain just needs the details to be right or I'll feel like I'm doing bad work ^^'

h-mayorquin:
Hey, @DradeAW.
See this first:
https://github.com/SpikeInterface/spikeinterface/issues/1432

Now on the subject matter. In general, I prefer to not have indirection for such a small collection of types. We have a `FolderPathType = TypeVar("FolderPathType", str, Path)` in another project and I find it annoying. I need to remember where to import this specific type so it gets on my way when I am coding. In this case I find `channel_ids: Union[str, int]` clear enough. 

Finally, I think I prefer the notation introduce in [PEP 604](https://peps.python.org/pep-0604/) which makes this even less verbose `str | int` (this can be enabled pre python 3.10 by using `[from __future__ import annotations]`.

DradeAW:
> Finally, I think I prefer the notation introduce in [PEP 604](https://peps.python.org/pep-0604/) which makes this even less verbose `str | int` (this can be enabled pre python 3.10 by using `[from __future__ import annotations]`.

I prefer it too (and use it in my own code)!
But my understanding was that it was introduced in Python 3.9, and thus not compatible with Python 3.8? (which SpikeInterface uses).

I personally don't like using `channel_ids: str | int` because what if, in the future, we decide that another type can represent channel_ids? We can't change the type everywhere, it would be too much work.

But if the plan is to switch to `str` in the future, then we can just wait until this is done :)

I'm just worried about the "under the hood" conversion to str: then if I give `unit_ids` to the sorting with integers, then `sorting.unit_ids == unit_ids --> False`

h-mayorquin:
Using `from __future__ import annotations.` will allow you tu use the short notation for python 3.8 and 3.9. 

```
I personally don't like using channel_ids: str | int because what if, in the future, we decide that another type can represent channel_ids? We can't change the type everywhere, it would be too much work.
```

This is a good point that it is good to have a centralized defintion. However, good IDEs should make the refactoring of this type of code easy. If the typing were a little bit more complicated or it was more likely to change it would move me in your direction, though.

samuelgarcia (approved):

PR #1665, opened 2023-05-19
Cleaner `WaveformExtractor.run_extract_waveforms`

DradeAW:
There was duplicated code from `BaseSorting.to_spike_vector`

This code should be cleaner like this :)


samuelgarcia:
This is clearly easier to read but I am not sure that this is more efficient to create intermediate dict of list.
When many many units constructing a spike vector directly is for me clearly faster.
Maybe I am wrong

samuelgarcia:
You rigth this is the same.

samuelgarcia (approved):

PR #1668, opened 2023-05-23
Refactor `write_binary_recording` tests

h-mayorquin:
Refactoring tests for #1602. We should merge this one before.

### DIFF
--- a/src/spikeinterface/core/tests/test_core_tools.py
+++ b/src/spikeinterface/core/tests/test_core_tools.py
@@ -1,10 +1,13 @@
 import platform
-import pytest
 from pathlib import Path
 
-from spikeinterface.core import generate_recording
+import pytest
+import numpy as np
 
+from spikeinterface.core import generate_recording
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/tests/test_core_tools.py
```diff
@@ -1,10 +1,13 @@
 import platform
-import pytest
 from pathlib import Path
 
-from spikeinterface.core import generate_recording
+import pytest
+import numpy as np
 
+from spikeinterface.core import generate_recording
```
alejoe91, line 7:
```suggestion
```

not using this anymore right?

alejoe91 (approved):
A small comment Ramon!

src/spikeinterface/core/tests/test_core_tools.py
h-mayorquin: Thanks

h-mayorquin:
> A small comment Ramon!

Turns out it was still needed. But now it is not anymore.

PR #1672, opened 2023-05-24
Test pickling speed without recursive

h-mayorquin:
Checking test durations.


samuelgarcia:
I am OK with this.
I think I prefer this solution over #1622
Lets make reccurssive always.

Why do we need `prepare_for_pickling` ?

src/spikeinterface/core/base.py
```diff
@@ -298,6 +298,23 @@ def to_dict(
         dump_dict: dict
             Serialized dictionary
         """
+        to_dict_kwargs = dict(
+            include_annotations=include_annotations,
+            include_properties=include_properties,
+            relative_to=relative_to,
+            folder_metadata=folder_metadata,
+        )
```
samuelgarcia, line 306: I think that include annotation and properties should be maybe False when propagating recurssively in the lambda function. I think we do not need then.
Only the top level are needed.

The old behavior before refactoring was like this when to_dict() was done everywhere.

h-mayorquin: Makes, sense, what about `folder_metadata`?

h-mayorquin: Wait, on a second thought, would not that beat the purpose that @DradeAW wanted to use this as a provenance. It should be a faithful version all the way up.

DradeAW: I don't mind having `include_annotations/properties` False by default, but I'd like the option to manually set it to true (yes I do want everything about the provenance, including all the annotations and properties).

Thanks everyone for all the hard good work!

src/spikeinterface/core/base.py
```diff
@@ -298,6 +298,23 @@ def to_dict(
...
+
+        new_kwargs = dict()
+        transform_extractors_to_dict = lambda x: x.to_dict(**to_dict_kwargs) if isinstance(x, BaseExtractor) else x
+        for name, value in self._kwargs.items():
+            if isinstance(value, list):
+                new_kwargs[name] = [transform_extractors_to_dict(element) for element in value]
+            if isinstance(value, dict):
```
h-mayorquin, line 313: This second if is wrong. It should be an elif. But if I correct this, the performance degrades to the one of https://github.com/SpikeInterface/spikeinterface/pull/1622.

src/spikeinterface/core/base.py
samuelgarcia: The question is more that we propagate ecussively properties or not.
I think we should not.
This is too heavy and useless.
And was not teh case before.

h-mayorquin: OK. Check the latest comments on the other PR though:
https://github.com/SpikeInterface/spikeinterface/pull/1622#issuecomment-1562000001

The correct solution to pickle taking too long to load is now in :
https://github.com/SpikeInterface/spikeinterface/pull/1674

DradeAW: Annotations aren’t heavy.
Also dumping to json doesn’t propagate properties. I thought the point of pickle was to contain all the information since it’s compressing everything

h-mayorquin:
OK, as @samuelgarcia wanted, we should keep the recursive version for a while. Maybe we will find some use for it and then I guess it is better to have this as a ready-to-use option in our hand if we do.

We also improved the docstring.

Maybe we should add an issue that if we don't find a use for the this in let's say 2 years we should just drop it altogether?

alejoe91 (approved):

h-mayorquin:
Yes!

PR #1673, opened 2023-05-24
More LFP-oriented preprocessing tools

cwindolf:
This PR has a few small changes to preprocessing tools:
 - Adds a flag to `filter` to allow users to ask for reflect-style padding on the boundary
 - Changes `filter` to let scipy compute the filter coefficients rather than doing it manually
 - Adds a new bad channels detection method to `detect_bad_channels`, which is oriented for LFP use cases. Inspired by Olvier's work, this one flags channels when they are not correlated enough with the median of their spatial neighbors, which is a useful way of finding dead or hot LFP channels.
 - Adds `astype`, which just lets you `rec = si.astype(rec, "float32")` or whatever. This is slightly redundant with the dtype arguments in other preprocessors, but not completely and I think it is handy to have around.
 - Changes default `edge_order` to 1 in `directional_derivative`, since it turns out `edge_order=2` looks horrible :) 

I plan to follow this soon with more registration code, but didn't want to drop a gigantic PR on you guys, so thought I would break it up a bit


cwindolf:
Fixed the changes and added a test, but along the way I wanted to make a change to `recording_tools.get_chunk_with_margin`. Previously, it allowed start_frame=None or end_frame=None when not doing any padding (i.e. add_zeros==add_reflect_padding==False), but if doing padding it asserted that these were not None. I  added logic to set start_frame=0 and end_frame=length when adding padding, which matches the behavior when not padding.

src/spikeinterface/preprocessing/astype.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def get_traces(self, start_frame, end_frame, channel_indices):
+        if channel_indices is None:
+            channel_indices = slice(None)
+        traces = self.parent_recording_segment.get_traces(start_frame, end_frame, channel_indices)
+        return traces.astype(self.dtype)
+
+
+# function for API
+astype = define_function_from_class(source_class=AstypeRecording, name="astype")
```
samuelgarcia, line 52: Very cool.
I wanted to do it.
I would also add the logic in `BaseRecording.astype()` with method that make a local import of this.
COuld try this ?

src/spikeinterface/preprocessing/detect_bad_channels.py
```diff
@@ -206,6 +218,53 @@ def detect_bad_channels(
...
+        correlations = []
+        for chunk in random_data:
+            chunk = chunk.astype(np.float32)
+            chunk = chunk - np.median(chunk, axis=0, keepdims=True)
+            padded_chunk = np.pad(chunk, [(0, 0), (0, 1)], constant_values=np.nan)
+            # channels with no neighbors will get a pure-nan median trace here
+            neighbmeans = np.nanmedian(
+                padded_chunk[:, channel_index],
+                axis=2,
+            )
```
samuelgarcia, line 253: This writting is really brilliant!
I am just wondering the cost of this new memory allocation for pading versus a loop over channel and local median. Is really faster ?

src/spikeinterface/preprocessing/filter.py
```diff
@@ -65,24 +65,21 @@ def __init__(
...
-                assert len(band) == 2
-                Wn = [e / sf * 2 for e in band]
-            else:
-                Wn = float(band) / sf * 2
-            N = filter_order
             # self.coeff is 'sos' or 'ab' style
-            filter_coeff = scipy.signal.iirfilter(N, Wn, analog=False, btype=btype, ftype=ftype, output=filter_mode)
+            filter_coeff = scipy.signal.iirfilter(
+                filter_order, band, fs=fs, analog=False, btype=btype, ftype=ftype, output=filter_mode
+            )
```
samuelgarcia, line 82: Fair enough. I was thinking that `fs` was introduced lately but it was 5 yeras ago!

src/spikeinterface/preprocessing/astype.py
cwindolf: Yes, nice idea! will add this

src/spikeinterface/preprocessing/detect_bad_channels.py
cwindolf: It was just easier to code up for me, since this is how we always do this kind of thing. I bet it's faster but it probably doesn't matter too much.

src/spikeinterface/preprocessing/tests/test_filter.py
```diff
@@ -22,36 +22,46 @@ def test_filter():
     rec = generate_recording()
     rec = rec.save()
 
-    rec2 = bandpass_filter(rec, freq_min=300.0, freq_max=6000.0)
+    #     rec2 = bandpass_filter(rec, freq_min=300.0, freq_max=6000.0)
 
-    # compute by chunk
-    rec2_cached0 = rec2.save(chunk_size=100000, verbose=False, progress_bar=True)
+    #     # compute by chunk
```
samuelgarcia, line 27: DO not foget to uncomment this later.

src/spikeinterface/preprocessing/detect_bad_channels.py
samuelgarcia: I bet this fatser on pytorch because the padding is lazy.

samuelgarcia:
Merci Charlie. This almost OK for me.

src/spikeinterface/preprocessing/tests/test_filter.py
cwindolf: Ma foi!

cwindolf:
thanks so much for the reviews @samuelgarcia ! re-added the tests and added the `rec.astype()` method

src/spikeinterface/preprocessing/astype.py
```diff
@@ -0,0 +1,52 @@
...
+        dtype,
+    ):
+        BasePreprocessorSegment.__init__(self, parent_recording_segment)
+        self.dtype = dtype
+
+    def get_traces(self, start_frame, end_frame, channel_indices):
+        if channel_indices is None:
+            channel_indices = slice(None)
+        traces = self.parent_recording_segment.get_traces(start_frame, end_frame, channel_indices)
+        return traces.astype(self.dtype)
```
h-mayorquin, line 48: I am flagging thsi as a point of discssuion. When I was writing on fixin the memory leak of writing I realized that this function always creates a malloc (even if the types) are the same:

https://github.com/SpikeInterface/spikeinterface/pull/1602/files#r1204822550

I am not sure what behavior you guys want here (maybe always returning a copy is the desired behavior) but I wanted to let you know of this in case is important:

https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html

cwindolf: I never knew! I guess we can set `copy=False` to fix that. Will do

h-mayorquin: I think there is some other casting down there to float32 that can be avoided.

src/spikeinterface/preprocessing/detect_bad_channels.py
```diff
@@ -206,6 +218,53 @@ def detect_bad_channels(
...
+        channel_index = np.full((num_channels, max_neighbors), num_channels)
+        for c in range(num_channels):
+            my_neighbors = np.flatnonzero(neighbors_mask[c])
+            channel_index[c, : my_neighbors.size] = my_neighbors
+
+        # get the correlation of each channel with its neighbors' median inside each chunk
+        # note that we did not concatenate the chunks here
+        correlations = []
+        for chunk in random_data:
+            chunk = chunk.astype(np.float32)
```
h-mayorquin, line 246: here

cwindolf: float is needed here since the logic relies on nans. plus, things will need to be float when computing the correlation

cwindolf: (and, this is just for some small chunks, not the whole recording)

h-mayorquin: I was  thinking that you could avoid the malloc if chunk is already np.float32.

cwindolf: ah, got it, sorry, wasn't following. added copy=False here too.

src/spikeinterface/preprocessing/astype.py
cwindolf: Yeah, it is terrible! Anyway, copy=False has been added in the cases here. Thanks @h-mayorquin !

samuelgarcia (approved):

PR #1674, opened 2023-05-24
Fix pickle loading

h-mayorquin:
This was a rabit hole.

OK, so this is the solution that is performing the best now:

```
results for branch fix_loading_pickle [This branch]
Time elapsed for saving: 0.0029435157775878906
Time elapsed for loading: 0.03637957572937012
Memory used by loading: 1.03125 MiB
Peak memory usage: 684.25390625 MiB
```

What was making loading and the [benchmark](https://gist.github.com/h-mayorquin/72f492275767fa9f2cb77a0b117eea67) very slow for loading was the deep copy.

As discussed on the other PR https://github.com/SpikeInterface/spikeinterface/pull/1622#issuecomment-1562000001 the other approaches are terrible in their loading and memory footprint performance. This approach is just better overall, saving, loading and memory usage.

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -948,7 +948,7 @@ def load_extractor(file_or_folder_or_dict, base_folder=None) -> BaseExtractor:
 
 
 def load_extractor_from_dict(d, base_folder=None) -> BaseExtractor:
-    print("Use load_extractor(..) instead")
+    warnings.warn("Use load_extractor(..) instead")
     return BaseExtractor.from_dict(d, base_folder=base_folder)
 
 
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -836,41 +836,60 @@ def _check_if_dumpable(d):
         return d["dumpable"]
 
 
-def _load_extractor_from_dict(dic):
-    cls = None
+def _load_extractor_from_dict(dic) -> BaseExtractor:
+    """
+    Convert a dictionary into an instance of BaseExtractor or its subclass.
+
+    This function takes a dictionary that represents the state of an extractor,
+    and reconstructs the extractor from the dictionary.
+
+    Parameters
+    ----------
+    dic : dict
+        A dictionary representation of the extractor which  must contain the following keys:
+        - "kwargs": A dictionary of keyword arguments to be passed to the extractor class upon instantiation.
+        - "class": The full name of the extractor class, including the module name.
+        - "version": The version of the extractor class used when the dictionary was created.
+        - "annotations": A dictionary of annotations.
+        - "properties": A dictionary of properties.
+
+    Returns
+    -------
+    BaseExtractor
+        An instance of BaseExtractor or its subclass.
+    """
+    extractor_class = None
     class_name = None
 
     if "kwargs" not in dic:
         raise Exception(f"This dict cannot be load into extractor {dic}")
-    kwargs = deepcopy(dic["kwargs"])
 
-    # handle nested
-    for k, v in kwargs.items():
-        if isinstance(v, dict) and is_dict_extractor(v):
-            kwargs[k] = _load_extractor_from_dict(v)
-
-            # handle list of extractors list
-    for k, v in kwargs.items():
-        if isinstance(v, list):
-            if all(is_dict_extractor(e) for e in v):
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/base.py
```diff
@@ -836,41 +836,60 @@ def _check_if_dumpable(d):
...
+        - "version": The version of the extractor class used when the dictionary was created.
+        - "annotations": A dictionary of annotations.
+        - "properties": A dictionary of properties.
+
+    Returns
+    -------
+    BaseExtractor
+        An instance of BaseExtractor or its subclass.
+    """
+    extractor_class = None
```
h-mayorquin, line 861: I changed this from csl because that's a reserved word.

src/spikeinterface/core/base.py
```diff
@@ -836,41 +836,62 @@ def _check_if_dumpable(d):
         return d["dumpable"]
 
 
-def _load_extractor_from_dict(dic):
-    cls = None
```
h-mayorquin, line 840: This is a reserved word in python.

src/spikeinterface/core/base.py
```diff
@@ -926,7 +947,7 @@ def load_extractor(file_or_folder_or_dict, base_folder=None):
         return BaseExtractor.load(file_or_folder_or_dict, base_folder=base_folder)
 
 
-def load_extractor_from_dict(d, base_folder=None):
+def load_extractor_from_dict(d, base_folder=None) -> BaseExtractor:
     print("Use load_extractor(..) instead")
```
DradeAW, line 951: I think it's better to use `warnings.warn` or `logging.warning` rather than `print`

alejoe91: Yep, this is old (not pointint fingers here, but I think it's @samuelgarcia 's fault!)

@h-mayorquin can you make this a `warnings.warn`? 

@DradeAW we don't use logging (but might switch in the future ;)

src/spikeinterface/core/base.py
```diff
@@ -836,41 +836,62 @@ def _check_if_dumpable(d):
...
+    -------
+    BaseExtractor
+        An instance of BaseExtractor or its subclass.
+    """
+    extractor_class = None
     class_name = None
 
     if "kwargs" not in dic:
         raise Exception(f"This dict cannot be load into extractor {dic}")
-    kwargs = deepcopy(dic["kwargs"])
```
samuelgarcia, line 845: This deepcopy was somhow important and not important.

For instance if you want to clone a recording in order to make an inplace change (on properties or probe or whatever). The two unerlying data (list, dict, arrays) must be copied.

I woulda dd an option for this and then check in detail in with case a copy needed or not.

samuelgarcia: In short we should have a copy but not recurssive.

DradeAW: There are some problems with `deepcopy`

When I use it in the Lussac tests, sometimes the annotations dissapear ...
So I wouldn't completely trust it!

DradeAW: I think the solution above re-creates `list` and `dict` right?

Maybe we can add `if not isinstance(obj, BaseExtractor): deepcopy(obj)`, since the problem was deep-copying multiple times the extractors.

h-mayorquin: That said, as @DradeAW mentions above this recreates the kwargs. I actually fail to see how this changes will fail to support the desired use-case. The kwargs are re-instantiated to initialize a class and then become _kwargs of the new class which should not be changed in there.

 The properties were taken directly from `dict` in the previous implementation (the one in main) and that's left as it is (see line 894) so I don't think this use case was ever supported.

samuelgarcia:
Thankyou for the feedback of deepcopy and many option.
Lets leave copy/deepcopy outisude this.
There is already a copy (not deep) by creating a new dict.
The deepcopy was needed before and was done at load time to be sure to not reference any more something from the parent. Something this can be needed. For parralel stuf this is not needed. For loading from json/pickle this is also not needed. Fore sure it was needed in some caeses.


> Recently, in a discussion in neuroconv @alejoe91 wanted to create a copy of an extractor and we ended up doing this:

I think the choice done was not good because it was referencing the parents. It you want a clone the best is to get the dict. Then copy or deepcopy depending the level of detach you want from the parent and then recreate it from dict.

PR #1675, opened 2023-05-25
Make NWB imports lazy

h-mayorquin:
I was profiling something and discussing with @alejoe91 today and we realize that the following imports are not lazy:

* Tridesclous
* Pynwb
* one.api import ONE

I am making them lazy here.

### DIFF
--- a/src/spikeinterface/extractors/nwbextractors.py
+++ b/src/spikeinterface/extractors/nwbextractors.py
@@ -8,37 +8,13 @@ from spikeinterface import get_global_tmp_folder
 from spikeinterface.core import BaseRecording, BaseRecordingSegment, BaseSorting, BaseSortingSegment
 from spikeinterface.core.core_tools import define_function_from_class
 
-try:
-    from pynwb import NWBHDF5IO, NWBFile
-    from pynwb.ecephys import ElectricalSeries, ElectrodeGroup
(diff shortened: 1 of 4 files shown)


alejoe91 (approved):

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -8,37 +8,13 @@
...
 
-
-def check_fsspec_install():
-    assert HAVE_FSSPEC, "To stream NWB data with fsspec, install fsspec: \n\n pip install fsspec aiohttp requests\n\n"
+def import_lazily():
+    "Makes annotations / typing available lazily"
+    global NWBFile, ElectricalSeries, NWBHDF5IO
+    from pynwb import NWBFile
+    from pynwb.ecephys import ElectricalSeries
+    from pynwb import NWBHDF5IO
```
alejoe91, line 17: Wouldn't it be better to just have `-> "NWBFile"` instead?

h-mayorquin: This has the marginal benefit of allowing me to navigate (click) on the types but it does add this ugly function. I am on the edge. I will like it to keep it here as an experiment though if you are fine with that.

PR #1693, opened 2023-06-06
Fix crash with kilosort and `verbose`

DradeAW:


src/spikeinterface/sorters/external/kilosortbase.py
```diff
@@ -120,6 +120,9 @@ def _get_specific_options(cls, ops, params):
     @classmethod
     def _setup_recording(cls, recording, sorter_output_folder, params, verbose):
         cls._generate_channel_map_file(recording, sorter_output_folder)
+        job_kwargs = get_job_kwargs(params, verbose)
+        if "verbose" not in job_kwargs:
```
h-mayorquin, line 124: Do we need this? That is, shuld the job_kwargs contain an entry for this?

PR #1696, opened 2023-06-06
Improve representation of Sorting in memory

samuelgarcia:
This should improve the in memory representation of Sorting using spike vector:
  * faster for multiprocessing
  * faster saving/loading when serialized
  * slower acces to individual to individuala spike train but caching system to compensate

The main important concept is this one:
```python

# this can sometimes have a very slow init (because check, or searchsorted, or mask or dict creation)
sorting = si.read_a_sorting_format()

# we can trnasform to NumpySorting with to_numpy_sorting()
# this have a cost of reading all spikes once.
# but then this sorting is very fast at init and have internally one big vector only
# so easy dumpability
sorting_np = sorting.to_numpy_sorting()

# This is the same but with a sharedmemm buffer
sorting_shm = sorting.to_shared_memory_sorting()

# or also
sorting_np = sorting.to_multiprocessing(n_jobs=1)
sorting_shm = sorting.to_multiprocessing(n_jobs=2)

# sorting_np and sorting_shm are dumpable
d = sorting_shm.to_dict()
d = sorting_np.to_dict()
```

Step 1 :
* [X] Refactor NumpySorting : use internal numpy spike_vector.
* [X] NumpySorting.from_dict() > NumpySorting.from_unit_dict() because was overwriting base.
* [x] deprecate Sorting.get_all_spike_trains() and converge to Sorting.to_spike_vector()
        @alejoe91 @h-mayorquin : this sub step was tedious, please double check that it is OK

Step 2
* [x] implement caching for spike vector
* [x] implement caching for dict of spikes times

Step 3 : 
* [x] Implement sharememNumpySorting
* [x] Add Sorting.to_sharemem()
* [x] Propagate when sorting is used in multiprocessing

Step 4
* [x] Implement new format based on memmap spike vector
* [x] Use this format by default in sorting.save()

This PR also:
  * make numpy sorting dumpable

To be discussed:
  * waveform extension when slicing use a per segment vector slice. I think It should easier to have a global slicing on 
     the full spike_vector. Maybe lets keep in actual state but lets have this in mind for postprocessing refactoring.


h-mayorquin:
Hey, @samuelgarcia. Each of the steps seem like a big change, I think that it would be easier to review and test if each was a PR on its own.

src/spikeinterface/core/numpyextractors.py
```diff
@@ -146,22 +170,31 @@ def from_times_labels(times_list, labels_list, sampling_frequency, unit_ids=None
...
+            spikes.append(spikes_in_seg)
+        spikes = np.concatenate(spikes)
+
+        sorting = NumpySorting(spikes, sampling_frequency, unit_ids)
 
         return sorting
 
     @staticmethod
-    def from_dict(units_dict_list, sampling_frequency) -> "NumpySorting":
+    def from_unit_dict(units_dict_list, sampling_frequency) -> "NumpySorting":
```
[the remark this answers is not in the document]
alejoe91: agreed, we should at least deprecate it...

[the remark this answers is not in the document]
samuelgarcia: No, the to_dict() was a huge mistake because it overwrited the to_dict() to dump the object.
It was use only internally.
So I think it is OK to get it back without transitory process

samuelgarcia:
> Hey, @samuelgarcia. Each of the steps seem like a big change, I think that it would be easier to review and test if each > was a PR on its own.

Of course and we decide with Alessio to make 4 small PRs instead of one big but :
  *b ecause I am lazy
  * because I want to publicaly show that I am stubborn
  * because Iwant to revenge against the pre-commit.ci

then I decided alone to do the exact oposite!
Feel free to hardly critize this choice and hate me for a while.

src/spikeinterface/core/numpyextractors.py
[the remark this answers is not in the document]
samuelgarcia: And also now the NumpySorting is dumpable so we really need the from_dict()/to_dict() to have the true behavior that come from the base.

src/spikeinterface/core/numpyextractors.py
```diff
@@ -238,26 +305,94 @@ def from_peaks(peaks, sampling_frequency) -> "NumpySorting":
...
+        times = self.spikes_in_seg[self.spikes_in_seg["unit_index"] == unit_index]["sample_index"]
+
         if start_frame is not None:
             times = times[times >= start_frame]
         if end_frame is not None:
             times = times[times < end_frame]
         return times
 
 
+class SharedMemmorySorting(BaseSorting):
```
DradeAW, line 348: Shouldn't it be `SharedMemorySorting`?

samuelgarcia: Oups. thank you. Going back to my english class now.

h-mayorquin:
All right, this will make us it harder for us to review. How are you profiling memory and performance? 

I would like to run some tests on windows and if they are similar to what you are doing it would be better.

samuelgarcia:
Hi Ramon, at the moment this is only concepts. I haven't tested performences.
I will do.

For windows, I do not now how well it will work. For sure the dumpability and having one unique vector will help a lot because otherwie we where picking many many arrays (one per units / segmnts).

I add a `Sorting.to_multiprocessing()` facility. If the sharedmemory machinery do not work well on window we can keep the NumpySorting in that function,  this will be centralized.

h-mayorquin:
@samuelgarcia 
This work for me:
```python
import numpy as np
import json

# Define the structured array
data = np.array([(1, "John", 28),
                 (2, "Alice", 32),
                 (3, "Bob", 45)],
                dtype=[('id', int), ('name', 'U10'), ('age', int)])

# Convert the structured array to a nested list
nested_list = data.tolist()

# Dump the nested list to a JSON file
from spikeinterface.core.core_tools import SIJsonEncoder

json_string = json.dumps(data, cls=SIJsonEncoder)
json.loads(json_string)
```

But most likely re-intialization does not work well when loaded from here if your class is expecting an structured array input. You will ned do something like this:

https://github.com/SpikeInterface/spikeinterface/blob/11a22b7728f0604778659f7536728f155cc433d6/src/spikeinterface/core/numpyextractors.py#L37-L44

src/spikeinterface/core/numpyextractors.py
```diff
@@ -93,30 +98,56 @@ def get_traces(self, start_frame, end_frame, channel_indices):
...
+            self.add_sorting_segment(NumpySortingSegment(spikes, segment_index, unit_ids))
+
+        # important trick : the cache is already spikes vector
+        self._cached_spike_vector = spikes
 
-        sorting.copy_metadata(source_sorting)
+        self._kwargs = dict(spikes=spikes, sampling_frequency=sampling_frequency, unit_ids=unit_ids)
 
+    @staticmethod
+    def from_sorting(source_sorting: BaseSorting, with_metadata=False) -> "NumpySorting":
```
h-mayorquin, line 141: This is a good change in naming.

src/spikeinterface/core/basesorting.py
```diff
@@ -190,31 +215,43 @@ def get_times(self, segment_index=None):
         else:
             return None
 
-    def _save(self, format="npz", **save_kwargs):
+    def _save(self, format="numpy_folder", **save_kwargs):
```
h-mayorquin, line 218: I think `npz_folder` and `numpy_folder` are pretty obscure references. Probably it will be a good idea to add clarification of pros and cons in the docstring.

Also, an annoying thing of the pattern we are using right now where extractor.save is called from the base class is that this string will not propagate to that method.

h-mayorquin (commented):
Took the first look at it. I will need to read this a couple of times. Surprised you rae using Folder instead of Directory here.  I thought you were a linux-all-in kind of guy : P

src/spikeinterface/core/basesorting.py
h-mayorquin: Short name is fine if it is documented where it is used. right now we don't have documentation of these two formats anywhere do we?

samuelgarcia:
@alejoe91 : it would be really nice to merge this soon after the release because because this make many changes that could lead to conflict tedious to track if we accept more PR in between

src/spikeinterface/core/tests/test_basesorting.py
```diff
@@ -144,14 +165,15 @@ def test_npy_sorting():
...
-    sorting = NumpySorting.from_dict({}, 30000)
+    sorting = NumpySorting.from_unit_dict({}, 30000)
 
     assert len(sorting.unit_ids) == 0
 
-    spikes = sorting.get_all_spike_trains()
-    assert len(spikes) == 1
-    assert len(spikes[0][0]) == 0
-    assert len(spikes[0][1]) == 0
+    with pytest.warns(DeprecationWarning):
```
h-mayorquin, line 172: To be honest, I am oddly suprised that you decided to test deprecation warnings. I think that I am on the side that is likely to be wrong by making too much testing and this is like on the line for me.

Not that I think is wrong, just pondering about my surprise.

src/spikeinterface/core/numpyextractors.py
```diff
@@ -94,37 +99,63 @@ def get_traces(self, start_frame, end_frame, channel_indices):
 
 
 class NumpySorting(BaseSorting):
+    """
```
h-mayorquin, line 102: Great that you added a docstring!

src/spikeinterface/core/numpyextractors.py
```diff
@@ -94,37 +99,63 @@ def get_traces(self, start_frame, end_frame, channel_indices):
...
+    In memory sorting object.
+    The internal representation is always done with a long "spike vector".
+
+
+    But we have convinient function to instantiate from other sorting object, from time+labels,
```
h-mayorquin, line 107: convenience*

It would be even better if we added the specific methods that you mentioned. I can add them later after we merge as well.

src/spikeinterface/core/numpyextractors.py
```diff
@@ -94,37 +99,63 @@ def get_traces(self, start_frame, end_frame, channel_indices):
...
+    sampling_frequency: float
+        The sampling frequency in Hz
+    channel_ids: list
+        A list of unit_ids.
+    """
+
     name = "numpy"
 
-    def __init__(self, sampling_frequency, unit_ids=[]):
+    def __init__(self, spikes, sampling_frequency, unit_ids):
```
h-mayorquin, line 122: How funny that I neve realized that the intialization was done always done with static method. 

Very elegant that you fit this right into it. As if this is the right datastructure to initialize this object. I like it.

src/spikeinterface/core/numpyextractors.py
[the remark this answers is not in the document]
h-mayorquin: Yeah, I remember I was confused about this a month ago. This is the right call. It is the current method that is overloaded with the other `to_dict` as you were telling us.

src/spikeinterface/core/numpyextractors.py
```diff
@@ -240,26 +310,100 @@ def from_peaks(peaks, sampling_frequency) -> "NumpySorting":
...
+        times = self.spikes_in_seg[self.spikes_in_seg["unit_index"] == unit_index]["sample_index"]
+
         if start_frame is not None:
             times = times[times >= start_frame]
         if end_frame is not None:
             times = times[times < end_frame]
         return times
 
 
+class SharedMemorySorting(BaseSorting):
```
h-mayorquin, line 352: Does this works on windows?

src/spikeinterface/core/basesorting.py
```diff
@@ -347,7 +412,7 @@ def get_all_spike_trains(self, outputs="unit_id"):
             spikes.append((spike_times, spike_labels))
         return spikes
```
h-mayorquin, line 413: I think that `to_spike_vector` can be confusing because we have the sorter methods `to_numpy_sorting`, `to_shared_memory_sorting` and `to_dict` that transform the sorter and output a different representation of the sorter.

`to_spike_vector` though just builds a vector representation of the spikes. I think that something like `build_spike_vector` would be easier to discover when someone is exploring the sorter with auto-completion and will help disambiguate the semantic convention that we have in the `to_something` methods.

h-mayorquin (commented):
This is cool, there are some tests failing but I review some of the core functionality as I guess that would not change that much.

I am still thinking about the profiling about as I think from the big changes here:
* Changing NumpySorting to operate with a vector
* Spike vector caching
* SharedMem

This covers the third one? I need to take a second look, this would have been easier if we separated these things in different PRs : P

Overall this looks good, I hope I can help with some of the profiling. Thanks for this work, Sam.

src/spikeinterface/core/sortingfolder.py
```diff
@@ -0,0 +1,137 @@
...
+
+    extractor_name = "NpzFolder"
+    mode = "folder"
+    name = "npzfolder"
+
+    def __init__(self, folder_path):
+        folder_path = Path(folder_path)
+
+        with open(folder_path / "npz.json", "r") as f:
+            d = json.load(f)
```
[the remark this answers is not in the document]
h-mayorquin: Same in the sorter above.

h-mayorquin:
Something else that I am thinking about.
We have the convention of having `BinaryRecordingExractor` for the bare bones and `BinaryFolderRecording` for loading `BinaryRecordingExractor` with all the metadata and the stuff. The recording object of our internal format.

Now we do the same with `NumpyFolderSorting` which is going to play a similar role as the sorter of our internal format.

When I started exploring the library I found the Folder name confusing in an extractor. I was thinking to myself: why do we have `BinaryFolderRecording` when we already have `BinaryRecordingExractor`, aren't they kind of doing the same? And I think I asked you guys about it.

Now I am thinking you are losing an opportunity for disambiguation branding, @samuelgarcia @alejoe91 !
What if we named this `SpikeInterfaceExtractor` and `SpikeInterfaceSorter` I think this a class name that reflects the intent very clearly and also bears our prideful name.

src/spikeinterface/core/basesorting.py
```diff
@@ -7,6 +7,9 @@
 from .waveform_tools import has_exceeding_spikes
 
 
+minimum_spike_dtype = [("sample_index", "int64"), ("unit_index", "int64"), ("segment_index", "int64")]
```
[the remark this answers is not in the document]
samuelgarcia: I think global variable is OK.

src/spikeinterface/core/basesorting.py
```diff
@@ -190,31 +215,43 @@ def get_times(self, segment_index=None):
...
         """
         This function replaces the old CachesortingExtractor, but enables more engines
-        for caching a results. At the moment only 'npz' is supported.
+        for caching a results.
+
+        Since v0.98.0 'numpy_folder' is used by defult.
+        From v0.96.0 to 0.97.0 'npz_folder' was the default.
+
+
+        At the moment only 'npz' is supported.
```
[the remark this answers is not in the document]
samuelgarcia: oups I forgot to remove this line

src/spikeinterface/core/basesorting.py
```diff
@@ -242,6 +287,17 @@ def get_total_num_spikes(self):
...
+    def count_total_num_spikes(self):
+        """
+        Get total number of spikes summed across segment and units.
+
+        Returns
+        -------
+        total_num_spikes: int
+            The total number of spike
+        """
+        return self.to_spike_vector().size
```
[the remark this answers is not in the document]
samuelgarcia: In many case the time vector is be already there (numpy, sharemem, ...), so clearly faster.
In other case we need to load all spiketrain anyway so the cost is almost the same.

src/spikeinterface/core/basesorting.py
samuelgarcia: good idea. @alejoe91 are you agree with this naming proposal ?
This could break some old code.

src/spikeinterface/core/basesorting.py
```diff
@@ -347,7 +412,7 @@ def get_all_spike_trains(self, outputs="unit_id"):
             spikes.append((spike_times, spike_labels))
         return spikes
 
-    def to_spike_vector(self, extremum_channel_inds=None):
+    def to_spike_vector(self, concatenated=True, extremum_channel_inds=None, use_cache=True):
```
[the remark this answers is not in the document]
samuelgarcia: concatenated=True/False is truely need. because in many place we want to loop over segments without to cut at the limit.

use_cache is also important in case you want to use it once and do not want to keep it in memory.

extremum_channel is need for spike location computation. This could be removed here and add more complexity on compute_spike_olocation side. Not sure if this is relevant or not.

src/spikeinterface/core/basesorting.py
```diff
@@ -370,31 +441,131 @@ def to_spike_vector(self, extremum_channel_inds=None):
...
+            ext_channel_inds = np.array([extremum_channel_inds[unit_id] for unit_id in self.unit_ids])
 
-        pos = 0
-        for segment_index, (spike_times, spike_labels) in enumerate(spikes_):
-            n = spike_times.size
-            spikes[pos : pos + n]["sample_index"] = spike_times
-            spikes[pos : pos + n]["unit_index"] = spike_labels
-            spikes[pos : pos + n]["segment_index"] = segment_index
-            pos += n
+        if use_cache and self._cached_spike_vector is not None:
```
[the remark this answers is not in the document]
samuelgarcia: I do not think I understant your question. There are 2 internal caches.
dict of spike train and spike vector.
get_unit_spike_train cache the dict and to_spike_vector cache the spike vector.

src/spikeinterface/core/basesorting.py
```diff
@@ -370,31 +441,131 @@ def to_spike_vector(self, extremum_channel_inds=None):
...
-            # vector way
-            spikes["channel_index"] = ext_channel_inds[spikes["unit_index"]]
+        else:
+            # the cache not needed or do not exists yet
+            spikes = []
+            for segment_index in range(self.get_num_segments()):
+                sample_indices = []
+                unit_indices = []
+                for u, unit_id in enumerate(self.unit_ids):
+                    spike_times = st = self.get_unit_spike_train(unit_id=unit_id, segment_index=segment_index)
```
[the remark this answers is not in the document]
samuelgarcia: thanks

src/spikeinterface/core/basesorting.py
```diff
@@ -370,31 +441,131 @@ def to_spike_vector(self, extremum_channel_inds=None):
...
+
+            if concatenated:
+                spikes = np.concatenate(spikes)
+
+            if use_cache and self._cached_spike_vector is None and extremum_channel_inds is None:
+                # cache it if necessary but only without "channel_index"
+                if concatenated:
+                    self._cached_spike_vector = spikes
+                else:
+                    self._cached_spike_vector = np.concatenate(spikes)
```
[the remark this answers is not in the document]
samuelgarcia: I am learning from masters.

src/spikeinterface/core/numpyextractors.py
samuelgarcia: :)

src/spikeinterface/core/tests/test_basesorting.py
alejoe91: @samuelgarcia is changing....

src/spikeinterface/core/numpyextractors.py
samuelgarcia: There are already. They are class method. I enhance the doc.

src/spikeinterface/core/numpyextractors.py
samuelgarcia: I think this work. We use to have bugs when using sharedmem stuff but we were not cleaning it properly.
We can make more test for this on windows platform.

src/spikeinterface/core/sortingfolder.py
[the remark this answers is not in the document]
samuelgarcia: I think that using "d" for a dict when the piece of code around 5 lines make the reading easier than "dictionary"
I known that your prefer a longer variable name.

src/spikeinterface/core/basesorting.py
alejoe91: yes, sounds good to me

src/spikeinterface/core/basesorting.py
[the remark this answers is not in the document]
alejoe91: maybe move the new args after `extremum_channel_inds=None`?

src/spikeinterface/core/basesorting.py
```diff
@@ -370,31 +440,131 @@ def to_spike_vector(self, extremum_channel_inds=None):
...
 
         return spikes
 
+    def to_numpy_sorting(self, propagate_cache=True):
+        """
+        Turn any sorting in a NumpySorting.
+        usefull to have it in memory with a unique vector representation.
```
alejoe91, line 511:
```suggestion
        useful to have it in memory with a unique vector representation.
```

src/spikeinterface/core/tests/test_basesorting.py
```diff
@@ -102,6 +119,10 @@ def test_BaseSorting():
     for unit in sorting_clean.get_unit_ids():
         assert unit not in empty_units
 
+    sorting4 = sorting.to_numpy_sorting()
+    sorting5 = sorting.to_multiprocessing(n_jobs=2)
```
alejoe91, line 123: can you add a test with `save` with more than ` job? This should also run on Windows

src/spikeinterface/exporters/to_phy.py
```diff
@@ -158,8 +158,9 @@ def export_to_phy(
 
     # export spike_times/spike_templates/spike_clusters
     # here spike_labels is a remapping to unit_index
-    all_spikes = sorting.get_all_spike_trains(outputs="unit_index")
-    spike_times, spike_labels = all_spikes[0]
+    all_spikes_seg0 = sorting.to_spike_vector(concatenated=False)[0]
```
alejoe91, line 161: Note that Phy only works for 1 segment

alejoe91:
> Something else that I am thinking about. We have the convention of having `BinaryRecordingExractor` for the bare bones and `BinaryFolderRecording` for loading `BinaryRecordingExractor` with all the metadata and the stuff. The recording object of our internal format.
> 
> Now we do the same with `NumpyFolderSorting` which is going to play a similar role as the sorter of our internal format.
> 
> When I started exploring the library I found the Folder name confusing in an extractor. I was thinking to myself: why do we have `BinaryFolderRecording` when we already have `BinaryRecordingExractor`, aren't they kind of doing the same? And I think I asked you guys about it.
> 
> Now I am thinking you are losing an opportunity for disambiguation branding, @samuelgarcia @alejoe91 ! What if we named this `SpikeInterfaceExtractor` and `SpikeInterfaceSorter` I think this a class name that reflects the intent very clearly and also bears our prideful name. Most people access this through the loading methods so I am not afraid about backwards compatibility.

I agree! We can rename the `Folder` classes as `SpikeInterfaceRecordingFolder`/`SpikeInterfaceSortingFolder`/`SpikeInterfaceSnippetsFolder`. We can also do it in a separate PR. @h-mayorquin could you open an issue about it?

samuelgarcia:
> I agree! We can rename the Folder classes as >SpikeInterfaceRecordingFolder/SpikeInterfaceSortingFolder/SpikeInterfaceSnippetsFolder. We can also do it in a separate > PR. @h-mayorquin could you open an issue about it?

In fact this is not a good idea because we will be able to reload folder from previous version and this will prevent us to make later a new (V3) folder version of recording or sorting.
Lets discuss this but I think this is outside this PR.

src/spikeinterface/exporters/to_phy.py
samuelgarcia: I know this make the code more explicit no ?

src/spikeinterface/core/tests/test_basesorting.py
samuelgarcia: yep

src/spikeinterface/core/basesorting.py
samuelgarcia: `get_spike_vector()`

alejoe91: Again, let's keep it for now and change it in a future PR

src/spikeinterface/core/tests/test_basesorting.py
samuelgarcia: NumpyFolder and NpzFolder do not have parralel writting.
The shared mem sorting is already tested in many place because we do `to_multiprocessing()` for waveform extractor

alejoe91 (approved):

alejoe91:
@h-mayorquin @samuelgarcia looks good to me!!

h-mayorquin:
Give me some time to do one last check today in the afternoon.

h-mayorquin (changes requested):
Give me some time to do one last check today in the afternoon.

src/spikeinterface/core/basesorting.py
h-mayorquin: If Sam is fine with `get_spike_vector` I am as well. I think that's an improvement over the existing one that we can release with this big change. If you don't have any reason against I suggest we change it now.

src/spikeinterface/core/numpyextractors.py
h-mayorquin: OK.

h-mayorquin (commented):
OK, Sam, thanks for waiting for me to have one last look. I added a bunch of comments but there is three things that I would really like to see In their order of importance:

1.  Change of  `get_spike_vector` to `get_spike_vector` if Alessio agrees.
2.  Send `extremum_channel_inds` logic to `compute_spike_olocation`.
3. Use another single letter variable instead of d for dictionaries in the added code.

I did not have time to look at the SharedMemory thing but I guess Alessio did.

h-mayorquin (commented):
OK, Sam, thanks for waiting for me to have one last look. I added a bunch of comments but there is three things that I would really like to see In their order of importance:

1.  Change of  `get_spike_vector` to `get_spike_vector` if Alessio agrees.
2.  Send `extremum_channel_inds` logic to `compute_spike_olocation`.
3. Use another single letter variable instead of d for dictionaries in the added code.

I did not have time to look at the SharedMemory thing but I guess Alessio did.

src/spikeinterface/core/sortingfolder.py
```diff
@@ -0,0 +1,136 @@
...
+from .base import _make_paths_absolute
+from .basesorting import BaseSorting, BaseSortingSegment
+from .npzsortingextractor import NpzSortingExtractor
+from .core_tools import define_function_from_class
+from .numpyextractors import NumpySortingSegment
+
+
+class NumpyFolderSorting(BaseSorting):
+    """
+    NumpyFolderSorting is the new internal format used in spikeinterface (>=0.99.0) for caching sorting objects.
```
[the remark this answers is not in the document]
h-mayorquin: Actually done.

src/spikeinterface/core/sortingfolder.py
```diff
@@ -0,0 +1,136 @@
...
+
+    It is a simple folder that contains:
+      * a file "spike.npy" (numpy format) with all flatten spikes (using sorting.to_spike_vector())
+      * a "numpysorting_info.json" containing sampling_frequency, unit_ids and num_segments
```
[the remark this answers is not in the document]
h-mayorquin: Also, apparently got an old  version

samuelgarcia:
Hi Ramon.
Thank for last feedback.


> Change of get_spike_vector to get_spike_vector if Alessio agrees.

OK for this but later in an other PR.
I will not have time to include this now because it needs more check to see if it do not break deeper and old code.
I really need this PR merge before all other ones and I lease for vaccation in a few day and Alessio will be also out for a while.


> Send extremum_channel_inds logic to compute_spike_olocation.

Same anwser. I agree but lets do this outside this PR. We can open an issue to not forget.


> Use another single letter variable instead of d for dictionaries in the added code.

OK I will do.

> I did not have time to look at the SharedMemory thing but I guess Alessio did.

I think yes.

h-mayorquin:
Ok, @samuelgarcia 

> OK for this but later in an other PR.
I will not have time to include this now because it needs more check to see if it do not break deeper and old code.
I really need this PR merge before all other ones and I lease for vaccation in a few day and Alessio will be also out for a while.

Maybe you can just create a the new method and have the old method call the new one with a deprecation warning? That should be quick, work as a reminder and not break any old code. 

I will open an issue about the rest of the stuff that we discussed here as soon as I have time.

DradeAW:
Could you merge the main branch onto this one?
I wanted to check that Lussac works with this new representation of sorting objects, but it's missing a few commits that make my algorithm crash

h-mayorquin:
@DradeAW done
You can do this locallily btw:

```
git switch name_of_same_remote/sharedmem_sorting
git merge upstream/main 
```

DradeAW:
All tests passed on my end!
I had to change some bits of code because of the new nomenclature, but in general I think the new names are better :)

Awesome stuff!

src/spikeinterface/core/sortingfolder.py
```diff
@@ -0,0 +1,136 @@
...
+
+    extractor_name = "NumpyFolderSorting"
+    mode = "folder"
+    name = "NumpyFolder"
+
+    def __init__(self, folder_path):
+        folder_path = Path(folder_path)
+
+        with open(folder_path / "numpysorting_info.json", "r") as f:
+            d = json.load(f)
```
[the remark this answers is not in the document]
samuelgarcia: done

src/spikeinterface/core/basesorting.py
```diff
@@ -190,31 +216,41 @@ def get_times(self, segment_index=None):
         else:
             return None
 
-    def _save(self, format="npz", **save_kwargs):
+    def _save(self, format="numpy_folder", **save_kwargs):
```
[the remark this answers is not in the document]
samuelgarcia: The save_kwargs was already there. It is to be flexible with differents format like we do in the rec._save()
Lets keep thos for now.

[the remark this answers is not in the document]
h-mayorquin: Ok, true that they were already there.

Let's discuss that in other PR issue as you say.

src/spikeinterface/core/basesorting.py
```diff
@@ -356,11 +420,17 @@ def to_spike_vector(self, extremum_channel_inds=None):
 
         Parameters
         ----------
+        concatenated: bool
+            By default the output is one numpy vector with all spikes from all segments
+            With concatenated=False then it is a list of spike vector by segment.
         extremum_channel_inds: None or dict
             If a dictionnary of unit_id to channel_ind is given then an extra field 'channel_index'.
```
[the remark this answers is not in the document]
samuelgarcia: Ok I changed this.

src/spikeinterface/core/basesorting.py
```diff
@@ -356,11 +420,17 @@ def to_spike_vector(self, extremum_channel_inds=None):
...
             This can be convinient for computing spikes postion after sorter.
 
             This dict can be computed with `get_template_extremum_channel(we, outputs="index")`
+        use_cache: bool
+            When True (default) the spikes vector is cache in an attribute of the object.
```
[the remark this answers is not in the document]
samuelgarcia: Done

samuelgarcia:
@h-mayorquin : I did a few changes and then I merge this if this pass tests.

h-mayorquin:
Sounds good to me, @samuelgarcia .

samuelgarcia:
Thanks a  lot.
@DradeAW : do you see some improvement ?
I think we will have more improvement when we will refactor the waveform extractor object I think.
But on some code it should be already faster to have the cache and no copy for parralel processing.

DradeAW:
> @DradeAW : do you see some improvement ?

It's hard to know for sure, I didn't make reproducible experiments.
For sure it's between 0-10% faster (so at worse the same in my case).

> I think we will have more improvement when we will refactor the waveform extractor object I think.

Can't wait! :)

Issue #1695, opened 2023-06-06
stream_id not found in containerized sorting


LaurenzMuessig:
Hi guys,

I recently had to re-install spike interface from scratch. Now when I try to run a containerized sorting (kilosort2) with a concatenated recording I get the error below. There is a copy of the concatenated and processed binary on the local disk. If I run it on a single binary pointing to the raw data folder it runs fine. Is there a possibility that this bug has crept in again in the most recent version. Any help is greatly appreciated. Thanks. This should be similar issue as in #1662.  

full error message:

Starting container
Installing spikeinterface from sources in spikeinterface/kilosort2-compiled-base
Installing dev spikeinterface from remote repository
Installing extra requirements: ['neo']
Running kilosort2 sorter inside spikeinterface/kilosort2-compiled-base
Stopping container

---------------------------------------------------------------------------
SpikeSortingError                         Traceback (most recent call last)
Cell In[3], line 82
     78     rec_Fin.save(folder=os.path.join(outDir,'raw'),verbose=True,progress_bar=True)
     79     # sort
     80     # recTemp = se.SpikeGLXRecordingExtractor(outDir,stream_id='imec0.ap')
     81     # sorting = ss.run_kilosort3(recording=rec_filt,output_folder=outDir,docker_image=True,**KS3Params) 
---> 82     sorting = ss.run_kilosort2(recording=rec_Fin,output_folder=outDir,docker_image=True,verbose=True, **KS2Params)
     83     # this needs work - rename cached traces to ap.bin, plus need to generate the .meta file 
     84     # runCatGT(fileName,outDir,params)
     85     
     86 # print('Yeepie')

File ~/si/spikeinterface/src/spikeinterface/sorters/runsorter.py:693, in run_kilosort2(*args, **kwargs)
    692 def run_kilosort2(*args, **kwargs):
--> 693     return run_sorter("kilosort2", *args, **kwargs)

File ~/si/spikeinterface/src/spikeinterface/sorters/runsorter.py:135, in run_sorter(sorter_name, recording, output_folder, remove_existing_folder, delete_output_folder, verbose, raise_error, docker_image, singularity_image, with_output, **sorter_params)
    133         else:
    134             container_image = singularity_image
--> 135     return run_sorter_container(
    136         container_image=container_image,
    137         mode=mode,
    138         **common_kwargs,
    139     )
    141 return run_sorter_local(**common_kwargs)

[...]

Issue #1689, opened 2023-06-05
Drop recrusive option for `to_dict` if we don't find a use case by 2024


h-mayorquin:
Recently we solved some pickling loading speed with big provenance on #1674.

While profiling, we found that the most performant solution for pickling and unpickling big provenance pipelines was to NOT recursively expand Extractors in the kwargs with to_dict.

That said, we decided to also merged #1672 as @samuelgarcia wanted to keep the functionality as this was how it was before the serialization effort in https://github.com/SpikeInterface/spikeinterface/issues/1235 and in case we found a use case. I am adding this issue here to keep track of the context and discussion. The idea was that maybe there is some use case for the recursive expansion and then it would be easier to use if we have the code already at hand. If this does not come to pass by next year, I think we should get rid of this code in the `to_dict` function as this is a central function, and we should try to keep the complexity down.

DradeAW:
I found another use case for `recursive=True`:

```python
base_folder = Path(...)

npz_sorting.to_dict(relative_to=base_folder)  # Works like a charm
npz_sorting.select_units(...).to_dict(relative_to=base_folder)  # Doesn't work because path is in a parent
npz_sorting.select_units(...).to_dict(relative_to=base_folder, recursive=True)  # Works correctly
```

alejoe91:
@h-mayorquin I think we can close this

h-mayorquin:
Yes!

Issue #1684, opened 2023-05-30
Failure in run_sort_by_property


thomas2cherian:
Hi,

I have implemented spikeinterface (with mountainsort4) on a dataset with large number of channels distributed across individual electrode arrays. I tested the sorting by splitting the recording based on the electrode array ID so that sorting runs only on the channels within each array. I then attempted to use the run_sorter_by_property function (to run the sorting across each unique array) as follows:

```
sortedData = si.run_sorter_by_property("mountainsort4", preprocessedData,
                                       grouping_property = 'fmaID',
                                       working_folder = baseFolder / "sorted",
                                       docker_image = True,
                                       verbose = True,
                                       detect_threshold = 3,
                                       filter = False,
                                       num_workers = 6,
                                       adjacency_radius = 50,
                                       clip_size = 50)`
```
The sorting seems to run fine (based on the verbose outputs) but I am encountering the error as follows (see errorLog.pdf attached for the full code). In the working folder, I see sub folders named with the arrayID that all only contain the folder 'mountainsort4' (with spikeinterface_job.json within).

```
SpikeSortingError: get result error. the folder do not contain spikeinterface_log.json
```

[errorLog.pdf](https://github.com/SpikeInterface/spikeinterface/files/11601573/errorLog.pdf)

 Am I using the function incorrectly? Is there any alternative way to combine sorting output from split recordings?

alejoe91:
Thanks for the report @thomas2cherian 

Let me try to reproduce the issue!

alejoe91:
@thomas2cherian 

I was able to reproduce the issue and looking into it

thomas2cherian:
@alejoe91 Thank you for looking into this issue!

alejoe91:
@thomas2cherian  can you test if this PR #1698  fixes your issue?

In brief you have to:
1. install SpikeInterface from that PR
2. set the following env variable: `SPIKEINTERFACE_DEV_PATH="path-to-cloned-repo` (this uses your local copy of SI in the container

Let me know if you have questions!

alejoe91:
yes, and install from the cloned repo with `pip install -e .`

thomas2cherian:
Hi @alejoe91, I tried what you have suggested but unfortunately the issue persists. I am getting the following error now:
```
[Errno 2] No such file or directory: 'sorted\\6A\\mountainsort4\\spikeinterface_job.json'
```

The verbose output from run_sort_by_property are as follows:
```
Starting container
[...]

PR #1698, opened 2023-06-06
Add recursive=True when caching objects before container

alejoe91:
Related to #1695

In #1674 we set the default behavior of `to_dict()` not to recursively serialize sub-parents. This is totally fine locally, but when running in a container we need the full dictionary to be serialized in order to properly find the volumes to map.

The issue is not totally solved, still WIP

### DIFF
--- a/src/spikeinterface/sorters/runsorter.py
+++ b/src/spikeinterface/sorters/runsorter.py
@@ -377,7 +377,7 @@ def run_sorter_container(
     output_folder.mkdir(parents=True, exist_ok=True)
 
     # find input folder of recording for folder bind
-    rec_dict = recording.to_dict()
+    rec_dict = recording.to_dict(recursive=True)
     recording_input_folders = find_recording_folders(rec_dict)
 
     if platform.system() == "Windows":


h-mayorquin:
Interesting. Do you have a working example?

alejoe91:
Yep!

For context: the `find_recording_folders` is used to find entries of the recording dict with a `path` in the name, so that the parent folder is bound to the container.

```
from spikeinterface.sorters.runsorter import find_recording_folders

rec = se.read_spikeglx("~/spikeinterface_datasets/ephy_testing_data/spikeglx/Noise4Sam_g0/", stream_id="imec0.ap")
rec_concat = si.concatenate_recordings([rec] * 10)

print(f"Parent folders - no recursion {find_recording_folders(rec_concat.to_dict())}")

# >>> Parent folders - no recursion []

print(f"Parent folders - with recursion {find_recording_folders(rec_concat.to_dict(recursive=True))}")

# >>> Parent folders - with recursion [PosixPath('/home/alessio/spikeinterface_datasets/ephy_testing_data/spikeglx')]
```

samuelgarcia:
Good catch camarade.
Does it solve the situation totally ?

alejoe91:
Still investigating

alejoe91:
fixes #1695 and #1684

- For #1684 we have to delay the creation of the sub-folder for the output of each group, otherwise we run into a FolderExists error
- For #1695 we have to explicitly make the dict recursive when using a container

DradeAW:
Does this close #1689 then?

alejoe91:
It is definitely as use case ;)

alejoe91:
Just tested locally for this issue #1684 and it runs fine on my machine:

image

alejoe91:
Fixes also #1695 on my side (tested with NP recordings):

image

alejoe91:
Good to merge on my side!

src/spikeinterface/sorters/runsorter.py
```diff
@@ -545,6 +553,7 @@ def run_sorter_container(
     cmd = ["python", f"{in_container_script_path_unix}"]
     res_output = container_client.run_command(cmd)
     run_sorter_output = res_output
+    run_sorter_output = run_sorter_output[2:-1].replace("\\n", "\n").replace("\\'", "'")
```
samuelgarcia, line 556: Why this ?
Is it system specific or terminal specific ?
Could you add a comment to remember ?

alejoe91: done!

src/spikeinterface/sorters/runsorter.py
```diff
@@ -553,10 +553,6 @@ def run_sorter_container(
     cmd = ["python", f"{in_container_script_path_unix}"]
     res_output = container_client.run_command(cmd)
     run_sorter_output = res_output
-    # this line "cleans" the output from the container by:
-    # - removing the "r" of the raw string
-    # - correcting character for new line and '
-    run_sorter_output = run_sorter_output[2:-1].replace("\\n", "\n").replace("\\'", "'")
```
alejoe91: good catch @samuelgarcia !!! Decoding the direct output makes the trick!

src/spikeinterface/sorters/runsorter.py
```diff
@@ -314,7 +320,7 @@ def stop(self):
     def run_command(self, command):
         if self.mode == "docker":
             res = self.docker_container.exec_run(command)
-            return str(res.output)
+            return res.output.decode("utf-8")
```
samuelgarcia, line 323: I would do res.output.decode("utf-8", errors="ignore") to be safer no ?

alejoe91, line 323:
```suggestion
            return res.output.decode(encoding="utf-8", errors="ignore")
```

alejoe91: done!

samuelgarcia (approved):

PR #1700, opened 2023-06-06
Remove unecessary try-except for shared memory

h-mayorquin:
This only made sense before python 3.8 which now is the minimal version we support.

### DIFF
--- a/src/spikeinterface/core/tests/test_core_tools.py
+++ b/src/spikeinterface/core/tests/test_core_tools.py
@@ -1,4 +1,5 @@
 import platform
+from multiprocessing.shared_memory import SharedMemory
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Hi Ramon.
Also, we have the same try except in core_tools.py make_shared_array().

Officialy the end of life of python 3.7 is 27 Jun 2023.
I propose to be good citizen and wait until this :)

alejoe91:
@samuelgarcia we have 3.8 as a requirement in the pyproject though!

samuelgarcia:
OK. So lets also remove the other try.

h-mayorquin:
All right, removed the other.

samuelgarcia (approved):

PR #1701, opened 2023-06-06
Fix cosine similarity matrix in `TemplateComparison` 

khl02007:
`postprocessing.template_similarity.compute_template_similarity` seems to have changed the way it accepts the second waveform extractor as an argument but `TemplateComparison` was not updated to reflect this. This causes an error when the two waveform extractors have different number of units. There may be other errors like this in the comparison module.


alejoe91 (approved):

alejoe91:
Good catch @khl02007 

thanks!

PR #1706, opened 2023-06-07
Add preprocessing for padding traces

h-mayorquin:
Adding a preprocessing step that pads traces to the left and to the right.

### DIFF
--- a/src/spikeinterface/preprocessing/tests/test_zero_padding.py
+++ b/src/spikeinterface/preprocessing/tests/test_zero_padding.py
@@ -36,10 +36,11 @@ def test_zero_paddin_channel():
     assert np.allclose(tr[:, :num_original_channels], rec.get_traces())
 
 
-def test_zero_padding_trace_full_trace():
+def test_trace_padded_recording_full_trace():
     num_channels = 4
     num_samples = 10
-    traces = np.ones((num_samples, num_channels))
+    rng = np.random.default_rng(seed=0)
+    traces = rng.random(size=(num_samples, num_channels))
     traces_list = [traces]
     recording = NumpyRecording(traces_list=traces_list, sampling_frequency=30_000)
 
@@ -56,10 +57,11 @@ def test_zero_padding_trace_full_trace():
     assert np.allclose(padded_traces[-padding_right:, :], np.zeros((padding_right, num_channels)))
 
 
-def test_zero_padding_trace_retrieve_original_trace():
+def test_trace_padded_recording_retrieve_original_trace():
     num_channels = 4
     num_samples = 10
-    traces = np.ones((num_samples, num_channels))
+    rng = np.random.default_rng(seed=0)
+    traces = rng.random(size=(num_samples, num_channels))
     traces_list = [traces]
     recording = NumpyRecording(traces_list=traces_list, sampling_frequency=30_000)
 
@@ -77,10 +79,11 @@ def test_zero_padding_trace_retrieve_original_trace():
     assert np.allclose(padded_traces, traces)
 
 
-def test_zero_padding_trace_retrieve_traces_with_partial_padding():
+def test_trace_padded_recording_retrieve_traces_with_partial_padding():
     num_channels = 4
     num_samples = 10
-    traces = np.ones((num_samples, num_channels))
+    rng = np.random.default_rng(seed=0)
+    traces = rng.random(size=(num_samples, num_channels))
     traces_list = [traces]
     recording = NumpyRecording(traces_list=traces_list, sampling_frequency=30_000)
 
--- a/src/spikeinterface/preprocessing/tests/test_zero_padding.py
+++ b/src/spikeinterface/preprocessing/tests/test_zero_padding.py
@@ -73,6 +73,33 @@ def test_trace_padded_recording_full_trace(recording, padding_start, padding_end
 
 
 @pytest.mark.parametrize("padding_start, padding_end", [(5, 5), (0, 5), (5, 0), (0, 0)])
+def test_trace_padded_recording_full_trace_with_channel_indices(recording, padding_start, padding_end):
+    num_channels = recording.get_num_channels()
+    num_samples = recording.get_num_samples()
+
+    padded_recording = TracePaddedRecording(
+        parent_recording=recording,
+        padding_start=padding_start,
+        padding_end=padding_end,
+    )
+
+    channel_indices = [0, 2]
+    padded_traces = padded_recording.get_traces(channel_ids=channel_indices)
+
+    # Until padding_start the traces should be filled with zeros
+    assert np.allclose(padded_traces[:padding_start, :], np.zeros((padding_start, num_channels)))
+    last_frame_of_original_trace = num_samples + padding_start
+
+    # Then from padding_start to the number of samples plus padding_start it should be the original traces
+    original_traces = recording.get_traces(channel_indices=channel_indices)
+    original_traces_from_padding = padded_traces[padding_start:last_frame_of_original_trace, :]
+    assert np.allclose(original_traces_from_padding, original_traces)
+
+    # After the original trace is over it should have zeros until the end of the padding right
+    assert np.allclose(padded_traces[last_frame_of_original_trace:, :], np.zeros((padding_end, num_channels)))
+
+
+@pytest.mark.parametrize("padding_start, padding_end", [(5, 5), (0, 5), (5, 0), (0, 0)])
 def test_trace_padded_recording_retrieve_original_trace(recording, padding_start, padding_end):
     num_samples = recording.get_num_samples()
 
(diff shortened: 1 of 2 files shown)


src/spikeinterface/preprocessing/tests/test_zero_padding.py
```diff
@@ -15,7 +17,7 @@
 set_global_tmp_folder(cache_folder)
 
 
-def test_zero_padding():
+def test_zero_paddin_channel():
```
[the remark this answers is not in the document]
h-mayorquin: Thanks.

src/spikeinterface/preprocessing/zero_channel_pad.py
```diff
@@ -1,10 +1,80 @@
...
+    def get_traces(self, start_frame, end_frame, channel_indices):
+        if start_frame is None:
+            start_frame = 0
+        if end_frame is None:
+            end_frame = self.get_num_samples()
+
+        trace_size = end_frame - start_frame
+        output_traces = np.zeros((trace_size, self.num_channels), dtype=self.dtype)
+
+        # If the stat_frame is shorter than padding left then we add the zeros to till start_frame is reached
```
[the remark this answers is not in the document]
h-mayorquin: Thanks a bunch!
I improved the documentation a bit in the last commit.

Not that is your reponsablity but you know you can make suggestions by doing the following:

image

Just passing a tip in case you did not know.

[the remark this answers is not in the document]
DradeAW: Ah so this is how you do the magic suggestions!
And here I thought it was because I wasn't an admin in the repo ^^'

Thanks for the tip!

src/spikeinterface/preprocessing/tests/test_zero_padding.py
```diff
@@ -34,5 +36,119 @@ def test_zero_padding():
...
+
+    # Find that there as many frames padded at the end as expected
+    padded_traces_end = padded_traces[-number_of_paded_frames_at_end:, :]
+    expected_zeros = np.zeros((number_of_paded_frames_at_end, num_channels))
+    assert np.allclose(padded_traces_end, expected_zeros)
+
+
 if __name__ == "__main__":
-    test_zero_padding()
+    test_zero_paddin_channel()
```
alejoe91, line 154:
```suggestion
    test_zero_padding_channel()
```

src/spikeinterface/preprocessing/tests/test_zero_padding.py
```diff
@@ -34,5 +36,119 @@ def test_zero_padding():
...
+
+    padded_recording = TracePaddedRecording(
+        parent_recording=recording,
+        padding_left=padding_left,
+        padding_right=padding_right,
+    )
+
+    # Extract the traces with partial padding in the left and to the right
+    number_of_padded_frames_at_start = 2
+    start_frame = padding_left - number_of_padded_frames_at_start
```
alejoe91, line 129: What if this is negative? (2.g. `0 - 2`)?
```suggestion
    start_frame = max(padding_left - number_of_padded_frames_at_start, 0)
```

src/spikeinterface/preprocessing/zero_channel_pad.py
```diff
@@ -1,10 +1,80 @@
...
+        The parent recording segment from which the traces are to be retrieved.
+    padding_left : int
+        The amount of padding to add to the left of the traces.
+    padding_right : int
+        The amount of padding to add to the right of the traces.
+    """
+
+    def __init__(self, parent_recording: BaseRecording, padding_left: int = 0, padding_right: int = 0):
+        super().__init__(recording=parent_recording)
+
```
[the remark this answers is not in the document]
h-mayorquin: OK, As Alessio gave thumbs up I imagine this is approved. I commited the change now.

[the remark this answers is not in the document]
h-mayorquin: Changed this.

src/spikeinterface/preprocessing/tests/test_zero_padding.py
h-mayorquin: Hey @alejoe91 . Thanks for checking this, it does not have sense to have `number_of_padded_frames_at_start` larger than 0 in the case where `padding_start` is 0 (which was on the parametric test). I changed the test to account for this.

Somehow though, there was a `(-1 ) x (-1 ) =  1` situation because the tests passed when I parametrized,

h-mayorquin:
I added two more tests:
1) Retrieve partial padding original traces patial padding at the end.
2) Retrieve with channels.

Both of them were failing on the previous implementation. The current implementation should be ready.

src/spikeinterface/preprocessing/zero_channel_pad.py
```diff
@@ -1,10 +1,115 @@
...
+        # This contains the padded elements by default and we add the original traces if necessary
+        trace_size = end_frame - start_frame
+        if isinstance(channel_indices, (np.ndarray, list)):
+            num_channels = len(channel_indices)
+        elif channel_indices == slice(None):
+            num_channels = self.num_channels
+        else:
+            raise ValueError(f"Unsupported channel_indices type: {type(channel_indices)} raise an issue on github ")
+
+        output_traces = np.full(shape=(trace_size, num_channels), fill_value=self.fill_value, dtype=self.dtype)
```
samuelgarcia, line 90: I think that if we are not one the border we could avoid this buffer alocation by returning directly the original_traces shifted.
This buffer only make sens on the left/right border.

src/spikeinterface/preprocessing/tests/test_zero_padding.py
```diff
@@ -34,5 +36,227 @@ def test_zero_padding():
...
+
+    # Find that there as many frames padded at the end as expected
+    padded_traces_end = padded_traces[frame_where_original_trace_ends:, :]
+    expected_zeros = np.zeros((number_of_paded_frames_at_end, num_channels))
+    assert np.allclose(padded_traces_end, expected_zeros)
+
+
 if __name__ == "__main__":
-    test_zero_padding()
+    test_zero_padding_channel()
```
samuelgarcia, line 262: I have the feeling that this tests are really really long for a this small task, no ?

samuelgarcia:
Hi Ramon. This is cool. I made 2 comments.

src/spikeinterface/preprocessing/tests/test_zero_padding.py
h-mayorquin: Where are you coming from, @samuelgarcia? It is unusual for people to argue for less testing but the opinion does exist out there. 

How are you thinking about this?

I am going to say something obvious but maybe is useful for establishing context and starting the conversation:
I used TTD here and each of those tests uncovered errors on my original implementation that I had to improve to pass. If someone wants to change -for example for optimization process- the function in the future then they will have this safeguard so they can  optimize away without being concerned about introducing new bugs.

src/spikeinterface/preprocessing/zero_channel_pad.py
h-mayorquin: I thought about that.

Honestly, I find it annoying that I have to reduce sequential readiability for the sake of memory optimizations out here but it probably makes sense. I wish that `get_traces` would have an out method like ufunctions in numpy so we could just pass previously allocated memory.

h-mayorquin: Check this out:

#1735

src/spikeinterface/preprocessing/tests/test_zero_padding.py
samuelgarcia: héhé. I am from France and you know it! :)
Is there ant correlation ? maybe...

src/spikeinterface/preprocessing/zero_channel_pad.py
h-mayorquin: That will a better solution but sadly I don't have time. I implemented you suggestion, check out the last PR.

src/spikeinterface/preprocessing/tests/test_zero_padding.py
h-mayorquin: Just to add something here, the torought thesting gave me the confidence for implementing your suggestion here https://github.com/SpikeInterface/spikeinterface/pull/1706#discussion_r1225883702 without me being concerned that something will broken.

alejoe91 (approved):

alejoe91:
@h-mayorquin @samuelgarcia this looks good to go for me!

h-mayorquin:
It is ready on my side if @samuelgarcia agrees. This will unblock https://github.com/SpikeInterface/spikeinterface/pull/1602.

samuelgarcia (approved):

samuelgarcia:
Merci camarades.
SOrry for the delay of merging.

Issue #1690, opened 2023-06-05
run_extract_waveforms after cluster split curation with Phy


Antho2422:
Hi everyone, 

I'm trying to make a curation using Phy. My current process is the following : 

- Run sorter spykingcircus2
- export waveforms to then export results to phy. My waveforms are saved in a new 'waveforms' folder.
- Do manual curation using Phy. I don't have any issues when I assign groups to clusters ('good', 'mua', 'noise) but I think the issue is coming from Merge/split of clusters. 
- Reading phy and creating a PhySortingExtractor (correctly updated because I have the right number of units after curation)
- I then try to remove my old 'waveforms' folder to create a new one with the updated waveforms from the new clusters. I'm trying to extract_waveforms.

I get an error on this last step when running extract waveforms. I tried to :
- don't remove original 'waveforms' folder but trying to extract curated waveforms from the curated sorting in a new folder.
- Running on my recording and updated sorting the function 'extract_waveforms() (with mode='memory' and mode='memmap')
- Create a WaveformExtractor object, then setting the params and the error is occuring when I run  we.run_extract_waveforms()

Nothing worked unfortunately. Sometimes the merge of units worked but it's not consitent. The split of cluster never worked tho...

This is my error : OSError: [Errno 22] Invalid argument: 'path_to_my_data\\waveforms\\waveforms_1.npy'

Please see attached a SC of my error.

Thank you, 
Anthony

![image](https://github.com/SpikeInterface/spikeinterface/assets/98460480/043f662f-d791-40b5-b149-04a1f2e0c76a)

alejoe91:
Hi Anthony,

The `run_extract_waveforms` is failing because it points to the "old" sorting object (the uncurated one) and you deleted the old waveforms folder.

Instead, you should re-extract waveforms using the recording and the new sorting object:
```
sorting_curated = se.read_phy("path-to-phy-folder")

we_curated = si.extract_waveforms(recording, sorting_curated, "new_folder")
```

Antho2422:
Hi Alessio, 

Thank you for your answer. I tried to review my process and if I'm not mistaken I don't think I'm pointing to the 'old' sorting object. I'm giving the sorting curated to extract_waveforms() and I try to not delete the old waveforms folder and I still get the error. The file waveforms_1.npy exists but I cannot open it in "wb"

Antho2422:
[...]

PR #1718, opened 2023-06-13
Allow PhyExtractor to handle splits from Phy

zm711:
**Problem** 

As seen in #1690 splitting in Phy leads to some problems for SI. Phy has two main actions merges which delete two rows of the `cluster_info` tsv/csv and creates a new row (which should) create a nan for the `si_unit_id`. In the case of splits however it takes one row in the tsv and copies it to two new rows with every column except `cluster_id` and `n_spikes` being the same value as the previous single row. Thus the `si_unit_id` is also duplicated in the case of splits.  This issue will only occur after the user has done any SI work because the only way the `si_unit_id` can be duplicated is if it already exists. So performing sorting/Phy outside of SI and doing splitting wouldn't cause this issue if trying to extract an old dataset.

**Potential Solution**

I've put one potential solution in this pull request. Originally I was thinking of a more complicated logic to handle duplicates and reuse numbers. But the issue with that more complicated logic is that it may still miss edge cases (If-elif-else would likely miss something). For example I know some Phy users who split a cluster multiple times to remove stimulation artifacts, which means rather than simple duplicates there could actually be `n` occurrences of the same `si_unit_id`. I've also tested merging and splitting repeatedly on Phy and the memcache can misbehave in extreme conditions so the merging and splitting results are not 100% consistent. So the simple solution of this PR is to check for `nan` which should indicate a merge (almost all the time) and check for duplicates (anything that is not 1) which should indicate (a) split(s). And if either of these conditions occur just use a new `si_unit_id` number. Simple logic, easier to debug if a new edge case appears in the future and behaves similarly to how Phy itself behaves so if users want to use Phy they won't need to change their intuition too much. If it is easier to read I could also make 
```python
if np.isnan(x)
elif np.count_nonzero(x)
else
```
instead, but I figured the shortcircuiting `or` would be fine since really the implication is in the case of a merge or a split.

### DIFF
--- a/src/spikeinterface/extractors/phykilosortextractors.py
+++ b/src/spikeinterface/extractors/phykilosortextractors.py
@@ -123,7 +123,7 @@ class BasePhyKilosortSortingExtractor(BaseSorting):
             for i, (phy_id, si_id) in enumerate(
                 zip(cluster_info["cluster_id"].values, cluster_info["si_unit_id"].values)
             ):
-                if np.isnan(si_id) or np.count_nonzero(cluster_info["si_unit_id"].values==si_id)!=1:
+                if np.isnan(si_id) or np.count_nonzero(cluster_info["si_unit_id"].values == si_id) != 1:
                     max_si_unit_id += 1
                     new_si_id = int(max_si_unit_id)
                 else:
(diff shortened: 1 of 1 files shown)


alejoe91:
Great solution @zm711 

Thank you!!

Fixes #1690

alejoe91 (approved):

PR #1723, opened 2023-06-15
Add original folder to phy extractor

DradeAW:
Added an annotation with the path to the original phy folder to the phy extractors.


DradeAW:
Another option would be to do something like `self.phy_folder = phy_folder`

It wouldn't propagate to sub sorting if operations are applied, which might or might not be what we want.

alejoe91 (approved):

PR #1733, opened 2023-06-21
Fix bug in `auto_merge` with no units

DradeAW:
Fixed a bug where `auto_merge` would crash if the sorting object has no units. Removed unused import.


alejoe91:
I guess you needed the import!

alejoe91 (changes requested):
some missing imports are causing an issue

DradeAW:
Yes, but it should be `scipy.signal` rather than `scipy.spatial`!!

Fixing now.

alejoe91 (approved):

PR #1736, opened 2023-06-22
Add plexon2 recording, sorting and event support

JuliaSprenger:
This PR is building on top of https://github.com/NeuralEnsemble/python-neo/pull/1214 and can only be merged once the pl2 format is supported by Neo.


JuliaSprenger:
It looks like not all required packages are installed for the CI tests. As far as I can see you are installing the latest version of the neo package, but not `all` dependencies, as this would include the missing package `zugbruecke` [here](https://github.com/NeuralEnsemble/python-neo/blob/6ce00dc9c13e85c2b2cbf6dacb6d66f93b0099c6/pyproject.toml#L121). Is this intended or would you prefer to install the missing package explicitly in the workflow? @CodyCBakerPhD

src/spikeinterface/extractors/neoextractors/plexon2.py
```diff
@@ -0,0 +1,101 @@
...
+
+    Parameters
+    ----------
+    file_path: str
+        The file path to load the recordings from.
+    """
+
+    mode = "file"
+    NeoRawIOClass = "Plexon2RawIO"
+    handle_spike_frame_directly = True
```
samuelgarcia, line 55: Can you have a look to #1626 @h-mayorquin made some variable name changes.

samuelgarcia:
Thanks a lot @JuliaSprenger 
Can you have a look when #1626 will be merge you will need a variable name change

samuelgarcia:
@alejoe91 do we include this also in the relase no ?

alejoe91:
Is it released on neo?

alejoe91:
@JuliaSprenger tests seem to fail because of `zugbruecke` (had to add it in the `extractors` dependencies). Would you have time to look into this?

alejoe91 (approved):

alejoe91:
@samuelgarcia @h-mayorquin good to merge for me!

I included the missing dependencies in CI and updated after #1626

.github/actions/build-test-environment/action.yml
```diff
@@ -40,3 +40,10 @@ runs:
         tar xvzf git-annex-standalone-amd64.tar.gz
         echo "$(pwd)/git-annex.linux" >> $GITHUB_PATH
       shell: bash
+    - name: Install wine (needed for Plexon2)
+      run: |
+        sudo rm -f /etc/apt/sources.list.d/microsoft-prod.list
+        sudo dpkg --add-architecture i386
+        sudo apt-get update -qq
+        sudo apt-get install -yqq --allow-downgrades libc6:i386 libgcc-s1:i386 libstdc++6:i386 wine
+      shell: bash
```
alejoe91, line 49: This can be quite heavy and not needed (unless plexon2 wrapper changes). We could make a manual workflow that installs wine and tests plexon2 only.

Otherwise we can make an `install_wine.yml` action and install + run plexon2 tests only if plexon2 has changed.

alejoe91:
@samuelgarcia done with the suggestions!

src/spikeinterface/extractors/tests/test_neoextractors.py
```diff
@@ -290,6 +290,32 @@ def test_pickling(self):
         pass
 
 
+# We mark plexon2 tests as they require additional dependencies (wine)
+@pytest.mark.plexon2
```
h-mayorquin, line 294: In POSIX the dependencies are wine and https://pypi.org/project/zugbruecke/ right?

In windows what are the dependecies?

I don't like that if I don't have wine or this experimental packages my test will fail if run pytest -m "extractors"

Can we add skipif tests to accomplish this? (this can be another PR if you think this is too much)

I am suggesting:

```python
import subprocess
import platform

def has_dependencies():
    """
    Check if required dependencies are installed.
    """
    
    os_type = platform.system()

    if os_type == "Windows":
        # On Windows, check for dependencies if any (probably just return True)
        return True

    elif os_type == "Linux":
        # Check for 'wine' using dpkg
        try:
            result_wine = subprocess.run(['dpkg', '-l', 'wine'], 
                                         stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
        except subprocess.CalledProcessError:
            return False
        
        # Check for 'zugbruecke' using pip
        try:
            import zugbruecke
            return True
        except subprocess.CalledProcessError:
            return False

    else:
        raise ValueError(f"Unsupported OS: {os_type}")

# Use pytest.mark.skipif to conditionally skip the test class
@pytest.mark.skipif(not has_dependencies(), reason="Required dependencies not installed")
@pytest.mark.plexon2
class Plexon2RecordingTest(RecordingCommonTestSuite, unittest.TestCase):
    ExtractorClass = Plexon2RecordingExtractor
    downloads = ["plexon"]
    entities = [
        ("plexon/4chDemoPL2.pl2", {"stream_id": "3"}),
    ]
        ```

Comment (h-mayorquin) [reply src/spikeinterface/extractors/tests/test_neoextractors.py:294]: I also think that maybe this will help us avoid adding -some- complexity to the testing in the CI but less sure about it.

Comment (alejoe91) [reply src/spikeinterface/extractors/tests/test_neoextractors.py:294]: Good point @h-mayorquin , I'll do something along these lines!

Comment (alejoe91) [reply src/spikeinterface/extractors/tests/test_neoextractors.py:328]: Done!

File: .github/workflows/full-test.yml [hunk 4]
```diff
@@ -122,11 +126,14 @@ jobs:
...
         run: chmod +x .github/run_tests.sh
+      - name: Install Wine (Plexon2)
+        if: ${{ steps.modules-changed.outputs.PLEXON2_CHANGED == 'true' }}
+        uses: ./.github/actions/install-wine
       - name: Test core
         run: ./.github/run_tests.sh core
       - name: Test extractors
         if: ${{ steps.modules-changed.outputs.EXTRACTORS_CHANGED == 'true' || steps.modules-changed.outputs.CORE_CHANGED == 'true' }}
-        run: ./.github/run_tests.sh "extractors and not streaming_extractors"
+        run: ./.github/run_tests.sh "extractors and not streaming_extractors and not plexon2"
```
h-mayorquin, line 136: Do we still need not here with the new condition that you added? Will it not be skipped automatically when wine is not installed and test it with all the other extractor when wine is  installed?

.github/workflows/full-test.yml
```diff
@@ -157,3 +164,6 @@ jobs:
       - name: Test internal sorters
         if: ${{ steps.modules-changed.outputs.SORTERS_INTERNAL_CHANGED == 'true' || steps.modules-changed.outputs.SORTINGCOMPONENTS_CHANGED || steps.modules-changed.outputs.CORE_CHANGED == 'true' }}
         run: ./.github/run_tests.sh sorters_internal
+      - name: Test plexon2
```
h-mayorquin, line 167: That way, we would not need this, right?

h-mayorquin (commented):
This looks good to me. One last question about the CI logic.

.github/workflows/full-test.yml
alejoe91: nope, removed!

PR #1737, opened 2023-06-22
Sorter relative path

miketrumpis:
This set of changes exposes functionality to make sorting runs using a relative path for the recording, by exposing these options

* `relative_to` option for `BaseExtractor.dump_to_json`
* `base_folder` option in `spikeinterface.core.base.load_extractor`

The keyword `recording_relative_path` is added to a few signatures in `BaseSorter`.
Logic here is to use `relative_to` for recording JSON dump in `initialize_folder`, and then `base_folder` for JSON load in `run_from_folder` and `get_results_from_folder`.

The bulk of the diffs propagate the options overloads of `BaseSorter._run_from_folder`.
Finally, I added the option to the `spikeinterface.sorters.runsorter.run_sorter_local` method. Did not get into the container space, as I felt that can be more delicate.

### DIFF
--- a/src/spikeinterface/sorters/basesorter.py
+++ b/src/spikeinterface/sorters/basesorter.py
@@ -237,9 +237,7 @@ class BaseSorter:
         t0 = time.perf_counter()
 
         try:
-            SorterClass._run_from_folder(
-                sorter_output_folder, sorter_params, verbose, relative_to=relative_to
-            )
+            SorterClass._run_from_folder(sorter_output_folder, sorter_params, verbose, relative_to=relative_to)
             t1 = time.perf_counter()
             run_time = float(t1 - t0)
             has_error = False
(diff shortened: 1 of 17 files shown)


alejoe91 (changes requested):
THanks @miketrumpis 

I would "unify" the API by directly using the `relative_to` instead of the `recording_relative_path`. I think this is clear enough :)

What do you think?

miketrumpis:
Happy to search-n-replace if that's your preference, but what about `base_folder`?

alejoe91:
Thanks! `base_folder` is when loading the object, so I think that the meaning is a bit different hence the different name

samuelgarcia:
Hi @miketrumpis : are you ok to close ?

miketrumpis:
Yup!

PR #1741, opened 2023-06-23
Make binary recording memmap efficient

h-mayorquin:
In the context of #1602 and https://github.com/NeuralEnsemble/python-neo/pull/1265. 

This does not go all the way to use native python buffers which is what we have done in those PR but it moves closer in that direction. I added a further test.

### DIFF
--- a/src/spikeinterface/core/binaryrecordingextractor.py
+++ b/src/spikeinterface/core/binaryrecordingextractor.py
@@ -162,6 +162,10 @@ class BinaryRecordingSegment(BaseRecordingSegment):
         self.time_axis = time_axis
         self.datfile = datfile
         self.num_samples = (Path(datfile).stat().st_size - file_offset) // (num_chan * np.dtype(dtype).itemsize)
+        if self.time_axis == 0:
+            self.shape = (self.num_samples, self.num_chan)
+        else:
+            self.shape = (self.num_chan, self.num_samples)
 
     def get_num_samples(self) -> int:
         """Returns the number of samples in this signal block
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/binaryrecordingextractor.py
```diff
@@ -156,25 +156,39 @@ def get_binary_description(self):
...
+        return self.num_samples
 
     def get_traces(
         self,
         start_frame: Union[int, None] = None,
         end_frame: Union[int, None] = None,
         channel_indices: Union[List, None] = None,
     ) -> np.ndarray:
-        traces = self._timeseries[start_frame:end_frame]
+        data = np.memmap(self.datfile, self.dtype, mode="r", offset=self.file_offset, shape=self.shape)
```
samuelgarcia, line 184: Hi Ramon.
We should use the same trick as in neo. Open the file in init an d create the memmap on the fly.
open/close on network drive has a cost!

DradeAW:
I don't understand why opening the memmap on `get_traces` is more efficient?

Opening and closing comes at a cost, but once opened it should only be a pointer to the file?
How is this more efficient?

h-mayorquin:
Long story short, the memmaps are not release and lead to a linear memroy increase. This is one way that we find of avoiding this problem. Opening memmaps and files is not that expensive when compared to most operations saved on the network case which is what Sam is talking about.

src/spikeinterface/core/binaryrecordingextractor.py
h-mayorquin: I made the two of them but I wanted to make smaller PRs:

#1742

DradeAW:
> the memmaps are not released

Interesting, I always thought that calling `memmap.close()` inside `__del__` took care of this issue. I'm gonna have to look into this for my own code!

Thank you :)

h-mayorquin:
You can read here why there is no unified interface:

https://github.com/numpy/numpy/issues/13510

h-mayorquin:
I am thinking on writing a blogpost about it but June is a very busy time for contracts. Maybe next month.

samuelgarcia:
I propose to close this and merge #1742 instead.

h-mayorquin:
Fine by me, I added the close flag on the other PR.

PR #1742, opened 2023-06-24
Make binary recording memmap efficient II (using native memmaps)

h-mayorquin:
Larger change than #1741. Can be merged after or instead of it.

Uses native memmaps instead and adds an extra test.

Similar approach as the one used in https://github.com/SpikeInterface/spikeinterface/pull/1741.

### DIFF
--- a/src/spikeinterface/core/binaryrecordingextractor.py
+++ b/src/spikeinterface/core/binaryrecordingextractor.py
@@ -169,6 +169,12 @@ class BinaryRecordingSegment(BaseRecordingSegment):
         else:
             self.shape = (self.num_chan, self.num_samples)
 
+        byte_offset = self.file_offset
+        dtype_size_bytes = self.dtype.itemsize
+        data_size_bytes = dtype_size_bytes * self.num_samples * self.num_chan
+        self.memmap_offset, self.array_offset = divmod(byte_offset, mmap.ALLOCATIONGRANULARITY)
+        self.memmap_length = data_size_bytes + self.array_offset
+
     def get_num_samples(self) -> int:
         """Returns the number of samples in this signal block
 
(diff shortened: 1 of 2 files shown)


h-mayorquin:
This should close #1741

src/spikeinterface/core/binaryrecordingextractor.py
```diff
@@ -156,25 +157,59 @@ def get_binary_description(self):
...
         start_frame: Union[int, None] = None,
         end_frame: Union[int, None] = None,
         channel_indices: Union[List, None] = None,
     ) -> np.ndarray:
-        traces = self._timeseries[start_frame:end_frame]
+        byte_offset = self.file_offset
+        dtype_size_bytes = self.dtype.itemsize
+        mmap_offset, array_offset = divmod(byte_offset, mmap.ALLOCATIONGRANULARITY)
+        data_size_bytes = dtype_size_bytes * self.num_samples * self.num_chan
+        mmmap_length = data_size_bytes + array_offset
```
samuelgarcia, line 190: Just for the pleasure to be picky : this could be in the init, no ?

h-mayorquin: Yep, makes sense, I moved it in the last commit.

That said, I think we could be even more efficient and only reserve the buffer that corresponds to the start_frame to end_frame. But let's do that in another PR.

samuelgarcia (approved):

src/spikeinterface/core/binaryrecordingextractor.py
samuelgarcia: perfect!

samuelgarcia:
Impeccable!

PR #1746, opened 2023-06-26
Propagate `get_times_kwargs` to BaseRecording

h-mayorquin:
Recording segments have a dictionary characterizing their timing properties:

https://github.com/SpikeInterface/spikeinterface/blob/5369cb0e799a52a1ccb1de6feacf05c41f65dd95/src/spikeinterface/core/baserecording.py#L660-L663

 Recently, we have been working on time alignment in neuroconv. For data ingestion / extraction is very useful to have this information available from the recoder.  This is what this PR does. It creates a new method at the `BaseRecording` level that returns the information witt the usual logic of working directly for single segment and requiring segment specification for multiple segment. 

The idea is that the segments should be an implementation detail and that users should not have to dig into them to extract timing information if so they desire.

I also added more complete docstrings to the segment method.

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -321,6 +321,32 @@ class BaseRecording(BaseRecordingSnippets):
         """
         return self.has_scaled()
 
+    def get_times_kwargs(self, segment_index=None) -> dict:
+        """
+        Retrieves the timing attributes for a given segment index. As with
+        other recorders this method only needs a segment index in the case
+        of multi-segment recordings.
+
+        Returns
+        -------
+        dict
+            A dictionary containing the following key-value pairs:
+
+            - 'sampling_frequency': The sampling frequency of the RecordingSegment.
+            - 't_start': The start time of the RecordingSegment.
+            - 'time_vector': The time vector of the RecordingSegment.
+
+        Notes
+        -----
+        The keys are always present, but the values may be None.
+        """
+
+        segment_index = self._check_segment_index(segment_index)
+        rs = self._recording_segments[segment_index]
+        time_kwargs = rs.get_times_kwargs()
+
+        return time_kwargs
+
     def get_times(self, segment_index=None):
         """Get time vector for a recording segment.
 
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Hi Ramon.
This is OK for me.
`get_times_kwargs` was more or less internal stuff. If we promote this external API maybe we could finde a better name no ?
`get_times_info` ? I do not have specific ideas.

h-mayorquin:
Good idea, @samuelgarcia. `get_times_info` is a better name than the current, I will see if I can come up with a better one.

alejoe91:
Upvote `get_times_info` for the `BaseRecording` function. Fine to keep `get_times_kwargs` for segments though, since it's internal

src/spikeinterface/core/baserecording.py
```diff
@@ -321,6 +321,32 @@ def has_scaled_traces(self):
         """
         return self.has_scaled()
 
+    def get_times_kwargs(self, segment_index=None) -> dict:
```
alejoe91, line 324:
```suggestion
    def get_times_info(self, segment_index=None) -> dict:
```

h-mayorquin: This is fine by me, the only thing that I would like is for it to be:

`get_time_info` without the plural.

Why?

We have a function called `get_times` and the `get_times_info` is not -exclusively- about it. So there is a 

I am thinking on something along the lines of:
`get_temporal_info` 

To totally disambiguate from times which are a non explicit property of the recording.

Maybe 
`get_temporal_parameters` 
?

alejoe91: yeah you are right, `get_time_info` sounds good to me. Also `temporal` is a good name, but it's a bit to abstract!

h-mayorquin: Yeah, I also don't fully like temporal.

The time information of your recorder.
The temporal information of your recorder.
The "word that condenses how your recorder handles time" information of your recorder.

Couldn't come with anything better.

samuelgarcia: get_time_info is ok for me.

h-mayorquin: Let's do that one then.

samuelgarcia (approved):

samuelgarcia:
OK for me.

PR #1753, opened 2023-06-28
Make recording JSON relative_to output_folder default for BaseSorter

### LINKED ISSUE #1739: Support for serializing recordings with relative paths during sorting run
I've noticed that the underlying support for relative paths in the JSON dump & load operators is never exposed at the level of `spikeinterface.sorters.basesorter.BaseSorter`.

Made a PR to address this #1737. I wasn't able to do a simple `python -m pytest ./src/spikeinterface` (lots of `ImportError`s). This is informally tested, and everything was round-tripped correctly.

The kwarg exposure is limited to propagating the changes up to `run_sorter_local`, but I didn't touch the container logic stuff (yet).

alejoe91:
This PR fixes #1739 and with a simpler approach than #1737 

Basically, we always save the `spikeinterface_recording.json` relative to the `output_folder`. This way, we don't have to change any signature in the class!

### DIFF
--- a/src/spikeinterface/sorters/basesorter.py
+++ b/src/spikeinterface/sorters/basesorter.py
@@ -141,7 +141,7 @@ class BaseSorter:
 
         rec_file = output_folder / "spikeinterface_recording.json"
         if recording.is_dumpable:
-            recording.dump_to_json(rec_file)
+            recording.dump_to_json(rec_file, relative_to=output_folder)
         else:
             d = {"warning": "The recording is not dumpable"}
             rec_file.write_text(json.dumps(d, indent=4), encoding="utf8")
@@ -221,9 +221,6 @@ class BaseSorter:
 
         SorterClass = sorter_dict[sorter_name]
 
-        # not needed normally
-        #  recording = load_extractor(output_folder / 'spikeinterface_recording.json')
-
         now = datetime.datetime.now()
         log = {
             "sorter_name": str(SorterClass.sorter_name),
@@ -300,7 +297,7 @@ class BaseSorter:
             sorting = cls._get_result_from_folder(output_folder)
 
         # register recording to Sorting object
-        recording = load_extractor(output_folder / "spikeinterface_recording.json")
+        recording = load_extractor(output_folder / "spikeinterface_recording.json", base_folder=output_folder)
         if recording is not None:
             # can be None when not dumpable
             sorting.register_recording(recording)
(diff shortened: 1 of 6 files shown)


alejoe91:
@samuelgarcia @miketrumpis

alejoe91:
Everything seems to work. Here is the `spikeinterface_recording.json` from an MS4 run:
image

Also for more complex runs (`run_sorters`, `run_sorter_by_property`), the  `spikeinterface_recording.json` are relative to `output_folder`:
image

Here the folder organization:
image


Closes #1737

alejoe91:
Ready for final review! Tested also with containers

miketrumpis:
This is perfect.. what I wanted to do in practice. Thanks @alejoe91!

src/spikeinterface/sorters/basesorter.py
```diff
@@ -271,7 +268,8 @@ def run_from_folder(cls, output_folder, raise_error, verbose):
         if has_error and raise_error:
             print(log["error_trace"])
             raise SpikeSortingError(
-                f"Spike sorting failed. You can inspect the runtime trace in {output_folder}/spikeinterface_log.json"
+                f"Spike sorting failed. You can inspect the runtime trace in {output_folder}/spikeinterface_log.json. "
+                f"Error trace: {log['error_trace']}"
```
samuelgarcia, line 272: Not sure this is a good idea because this error_trace can be enourmous and not well formated.
This is why the error was short and we used a print(log["error_trace"]) on line before

src/spikeinterface/sorters/basesorter.py
```diff
@@ -290,7 +288,8 @@ def get_result_from_folder(cls, output_folder):
 
         if bool(log["error"]):
             raise SpikeSortingError(
-                "Spike sorting failed. You can inspect the runtime trace in spikeinterface_log.json"
+                f"Spike sorting failed. You can inspect the runtime trace in {output_folder}/spikeinterface_log.json. "
+                f"Error trace: {log['error_trace']}"
```
samuelgarcia, line 292: Here also I would put a print(log["error_trace"]) before an dkeep the error tarce small.

samuelgarcia:
Hi @alejoe91.
OK for me except the error traces.

src/spikeinterface/sorters/basesorter.py
alejoe91: Ah, but it's not clear that that's the error! I'll remove the previous print and keep it like this. Usually the error trace is not too long, and if it is it' good to print it!

alejoe91: done

PR #1754, opened 2023-06-28
`num_chan` to `num_channels` in `BinaryRecordingExtractor`

h-mayorquin:
While I was working on #1742 I realized that the `BinaryRecordingExtractor` does not follow the convention of the library of referring to number of channels as `num_channels` (as we do in `BaseRecording.get_num_channels`). Instead, it uses `num_chan`. 

This PR starts adding a new argument `BinaryRecordingExtractor` and will throw a deprecation warning for the old argument.

### DIFF
--- a/src/spikeinterface/core/binaryrecordingextractor.py
+++ b/src/spikeinterface/core/binaryrecordingextractor.py
@@ -1,13 +1,12 @@
 from typing import List, Union
 import mmap
-
-import shutil
+import warnings
 from pathlib import Path
 
 import numpy as np
 
 from .baserecording import BaseRecording, BaseRecordingSegment
-from .core_tools import read_binary_recording, write_binary_recording, define_function_from_class
+from .core_tools import write_binary_recording, define_function_from_class
 from .job_tools import _shared_job_kwargs_doc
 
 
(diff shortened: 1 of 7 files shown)


samuelgarcia:
Good idea. And lets keep the wrning for a while because we have so many code with the old variable name....

src/spikeinterface/core/binaryrecordingextractor.py
```diff
@@ -55,20 +56,26 @@ def __init__(
...
         t_starts=None,
         channel_ids=None,
         time_axis=0,
         file_offset=0,
         gain_to_uV=None,
         offset_to_uV=None,
         is_filtered=None,
+        num_chan=None,
     ):
+        num_channels = num_channels or num_chan
```
alejoe91, line 70: What if they are both given? and maybe different? e.g. `num_channels=16` and `num_chan=32`?

alejoe91 (changes requested):
@h-mayorquin thanks for this! I have a small comment: IMO we should make the init checks even more stringent! Currently, one could pass both arguments and in that case the behavior is not clear

src/spikeinterface/core/binaryrecordingextractor.py
h-mayorquin: This:
https://en.wikipedia.org/wiki/Short-circuit_evaluation

Together with this:
https://docs.python.org/3/library/stdtypes.html#truth-value-testing

It is a common python idiom but maybe I should make the point of avoiding it as it can be confusing.

alejoe91 (approved):

PR #1756, opened 2023-06-29
Fix lazy generator test

h-mayorquin:
I realize that some assertions were missing in the lazy generator tests. 

This PR fixes it.


alejoe91:
@h-mayorquin still WIP or ready to review?

h-mayorquin:
I don't know why the mac test are failing. I need to think about it.

h-mayorquin:
It is ready now, @alejoe91, turns out that mac is a bit better to liberate memory than ubuntu so the test does not pass it because some of the memory was already free. This was couple with my tests being memory_used ~ memory_expected which now has been modified to memory_used <= memory_expected + tolerance

alejoe91 (approved):

PR #1766, opened 2023-07-03
Fix KS padded recording

alejoe91:
See https://github.com/SpikeInterface/spikeinterface/commit/6a6c82f2cbce269f3eaaf47a931dd8bd4fc803bc#r120509954

@h-mayorquin @samuelgarcia


h-mayorquin (approved):
Great, yes, that escaped me. One of the few instances where using positional arguments would have avoided the error.

PR #1772, opened 2023-07-03
Exclude end_frame in PhyKilosort get_unit_spike_train

TomBugnon:
This excludes end_frame in get_unit_spike_train.

This changes the extractor's behavior...!
But is required for a consistent definition of spike_train values (ie: frame of the parent extractor, and therefore exclusive) and in practice: to avoid IndexErrors when using frame_slice sortings.


DradeAW:
My bad, this is my fault!

Thanks for the catch :)

alejoe91 (approved):

PR #1775, opened 2023-07-04
Add _is_json_serializable to extractor

samuelgarcia:
Add _is_json_serializable
is_dumpable > _is_dumpable
clean some is_dumpable at some places

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -58,7 +58,9 @@ class BaseExtractor:
         #  * number of units for sorting
         self._properties = {}
 
-        self.is_dumpable = True
+        self._is_dumpable = True
+        self._is_json_serializable = True
+
 
         # extractor specific list of pip extra requirements
         self.extra_requirements = []
@@ -372,7 +374,7 @@ class BaseExtractor:
             "class": class_name,
             "module": module,
             "kwargs": kwargs,
-            "dumpable": self.is_dumpable,
+            "dumpable": self._is_dumpable,
             "version": version,
             "relative_paths": (relative_to is not None),
         }
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -373,7 +373,6 @@ class BaseExtractor:
             "class": class_name,
             "module": module,
             "kwargs": kwargs,
-            "dumpable": self._is_dumpable,
             "version": version,
             "relative_paths": (relative_to is not None),
         }
@@ -471,9 +470,14 @@ class BaseExtractor:
         return clone
 
     def check_if_dumpable(self):
-        # Find all nested _is_dumpable
-        dict_repr = self.to_dict()
-        kwargs = dict_repr["kwargs"]
+        """Check if the object is dumpable, including nested objects.
+
+        Returns
+        -------
+        bool
+            True if the object is dumpable, False otherwise.
+        """
+        kwargs = self._kwargs
         for value in kwargs.values():
             # here we check if the value is a BaseExtractor, a list of BaseExtractors, or a dict of BaseExtractors
             if isinstance(value, BaseExtractor):
@@ -485,9 +489,15 @@ class BaseExtractor:
         return self._is_dumpable
 
     def check_if_json_serializable(self):
-        # Find all nested _json_serializable
-        dict_repr = self.to_dict()
-        kwargs = dict_repr["kwargs"]
+        """
+        Check if the object is json serializable, including nested objects.
+
+        Returns
+        -------
+        bool
+            True if the object is json serializable, False otherwise.
+        """
+        kwargs = self._kwargs
         for value in kwargs.values():
             # here we check if the value is a BaseExtractor, a list of BaseExtractors, or a dict of BaseExtractors
             if isinstance(value, BaseExtractor):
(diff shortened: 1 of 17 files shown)


src/spikeinterface/core/base.py
```diff
@@ -58,7 +58,9 @@ def __init__(self, main_ids: Sequence) -> None:
         #  * number of units for sorting
         self._properties = {}
 
-        self.is_dumpable = True
+        self._is_dumpable = True
+        self._is_json_serializable = True
```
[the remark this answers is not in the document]
samuelgarcia: I think this is only internal.
property are more when an attribute is expose to the user and protected.
I do not think we want to expose this concept and document it. This is internal no ?

src/spikeinterface/core/baserecording.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, sampling_frequency: float, channel_ids: List, dtype):
             self, channel_ids=channel_ids, sampling_frequency=sampling_frequency, dtype=dtype
         )
 
-        self.is_dumpable = True
+        self._is_json_serializable = True
```
h-mayorquin, line 35: Is this needed? Isn't this inherited from BaseExtractor?

src/spikeinterface/core/basesnippets.py
```diff
@@ -26,7 +26,7 @@ def __init__(
         )
         self._nbefore = nbefore
         self._snippet_len = snippet_len
-        self.is_dumpable = True
+        self._is_json_serializable = True
```
h-mayorquin, line 29: Same here, it should be inherited right?

src/spikeinterface/core/numpyextractors.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, traces_list, sampling_frequency, t_starts=None, channel_ids=N
             assert len(t_starts) == len(traces_list), "t_starts must be a list of same size than traces_list"
             t_starts = [float(t_start) for t_start in t_starts]
 
-        self.is_dumpable = True
+        self._is_json_serializable = False
```
h-mayorquin, line 62:
```suggestion
        self._is_json_serializable = False  # Too large of a file, bad it to convert to json
```

src/spikeinterface/core/tests/test_jsonification.py
```diff
@@ -141,7 +141,10 @@ def __init__(self, attribute, other_extractor=None, extractor_list=None, extract
         self.other_extractor = other_extractor
         self.extractor_list = extractor_list
         self.extractor_dict = extractor_dict
-        self.is_dumpable = True
+
+        # this already the case by default
+        self._is_dumpable = True
```
h-mayorquin, line 146: I don't thinks this is needed, is it? It should be inherited?

src/spikeinterface/core/base.py
h-mayorquin, line 1: I think this line should be changed to use the json check instead of the dumpable check:

https://github.com/catalystneuro/spikeinterface/blob/87f27d87c96b5a30d5136f8e2ee5ccfa803af057/src/spikeinterface/core/base.py#L537

h-mayorquin (commented):
Great,
why not go all the way and call `is_dumpable` to `is_picklable` instead of `is_dumpable`? or `can_be_pickled`

src/spikeinterface/core/base.py
samuelgarcia: Yes I know I also need to rewrite a reccurisive check_json_serializable function and chcek we never use the direct flag fr this but the reccursive function

alejoe91:
- [ ] re-write check_if_dumpable with recursive function
- [ ] write check_if_json_serializable
- [ ] remove instances of `if ... _is_dumpable/is_json_serializable` and use `check_***` functions
- [ ] check where check_if_dumpable should be check_if_json_serializable

src/spikeinterface/core/base.py
```diff
@@ -372,7 +373,7 @@ def to_dict(
             "class": class_name,
             "module": module,
             "kwargs": kwargs,
-            "dumpable": self.is_dumpable,
+            "dumpable": self._is_dumpable,
```
samuelgarcia, line 376: we can remove this now no ?

src/spikeinterface/core/base.py
```diff
@@ -470,7 +471,32 @@ def clone(self) -> "BaseExtractor":
         return clone
 
     def check_if_dumpable(self):
-        return _check_if_dumpable(self.to_dict())
+        # Find all nested _is_dumpable
+        dict_repr = self.to_dict()
```
samuelgarcia, line 475: Lets add explicitly to_dict(recursive=false) to avoid errors later

src/spikeinterface/core/base.py
```diff
@@ -470,7 +471,32 @@ def clone(self) -> "BaseExtractor":
...
+        kwargs = dict_repr["kwargs"]
```
samuelgarcia, line 476: why not kwargs = self._kwargs to avoid the to_dict ?

src/spikeinterface/core/base.py
alejoe91: yep, removed

src/spikeinterface/core/base.py
alejoe91: good point, done

src/spikeinterface/core/base.py
alejoe91: good point, done!

alejoe91 (approved):

PR #1776, opened 2023-07-04
Fixed `num_chan` warning and propagate to remaining instances

DradeAW:
kwargs in `BinaryRecordingExtractor` were causing warnings to be thrown when loading from dict.

### DIFF
--- a/src/spikeinterface/core/binaryfolder.py
+++ b/src/spikeinterface/core/binaryfolder.py
@@ -50,6 +50,9 @@ class BinaryFolderRecording(BinaryRecordingExtractor):
 
         self._kwargs = dict(folder_path=str(folder_path.absolute()))
         self._bin_kwargs = d["kwargs"]
+        if "num_channels" not in self._bin_kwargs:
+            assert "num_chan" in self._bin_kwargs, "Cannot find num_channels or num_chan in binary.json"
+            self._bin_kwargs["num_channels"] = self._bin_kwargs["num_chan"]
 
     def is_binary_compatible(self):
         return True
@@ -58,7 +61,7 @@ class BinaryFolderRecording(BinaryRecordingExtractor):
         d = dict(
             file_paths=self._bin_kwargs["file_paths"],
             dtype=np.dtype(self._bin_kwargs["dtype"]),
-            num_channels=self._bin_kwargs["num_chan"],
+            num_channels=self._bin_kwargs["num_channels"],
             time_axis=self._bin_kwargs["time_axis"],
             file_offset=self._bin_kwargs["file_offset"],
         )
--- a/src/spikeinterface/core/binaryfolder.py
+++ b/src/spikeinterface/core/binaryfolder.py
@@ -53,9 +53,6 @@ class BinaryFolderRecording(BinaryRecordingExtractor):
         self._bin_kwargs = d["kwargs"]
         if "num_channels" not in self._bin_kwargs:
             assert "num_chan" in self._bin_kwargs, "Cannot find num_channels or num_chan in binary.json"
-            warnings.warn(
-                f"`num_chan` contained in folder {folder_path} is deprecated: it will not be redeable starting with version 0.100."
-            )
             self._bin_kwargs["num_channels"] = self._bin_kwargs["num_chan"]
 
     def is_binary_compatible(self):
--- a/src/spikeinterface/core/binaryfolder.py
+++ b/src/spikeinterface/core/binaryfolder.py
@@ -1,6 +1,5 @@
 from pathlib import Path
 import json
-import warnings
 
 import numpy as np
 
--- a/src/spikeinterface/core/binaryfolder.py
+++ b/src/spikeinterface/core/binaryfolder.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 import json
+import warnings
 
 import numpy as np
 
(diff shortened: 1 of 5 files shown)


src/spikeinterface/core/binaryrecordingextractor.py
```diff
@@ -121,7 +121,7 @@ def __init__(
             "sampling_frequency": sampling_frequency,
             "t_starts": t_starts,
             "num_channels": num_channels,
-            "num_chan": num_channels,  # TODO: This should be here at least till version 0.100.0
+            "num_chan": num_chan,  # TODO: This should be here at least till version 0.100.0
```
samuelgarcia, line 124: we should remove this line, I think.
There is no reason to dump this

src/spikeinterface/core/binaryfolder.py
```diff
@@ -50,6 +50,9 @@ def __init__(self, folder_path):
 
         self._kwargs = dict(folder_path=str(folder_path.absolute()))
         self._bin_kwargs = d["kwargs"]
+        if "num_channels" not in self._bin_kwargs:
```
h-mayorquin, line 53: Should this be removed as well?

src/spikeinterface/core/core_tools.py
```diff
@@ -160,15 +160,15 @@ def add_suffix(file_path, possible_suffix):
     return file_path
 
 
-def read_binary_recording(file, num_chan, dtype, time_axis=0, offset=0):
+def read_binary_recording(file, num_channels, dtype, time_axis=0, offset=0):
```
h-mayorquin, line 163: I don't think we are using this function anymore, maybe we can drop it? just a comment though, not important for this PR.

src/spikeinterface/core/binaryfolder.py
h-mayorquin: The assertion under I mean.

alejoe91: either `num_channels` or `num_chan` should be in the dict right?

h-mayorquin: Shouldn't it be only `num_channels` now?

src/spikeinterface/core/binaryfolder.py
```diff
@@ -50,6 +50,9 @@ def __init__(self, folder_path):
...
+            assert "num_chan" in self._bin_kwargs, "Cannot find num_channels or num_chan in binary.json"
+            self._bin_kwargs["num_channels"] = self._bin_kwargs["num_chan"]
```
h-mayorquin, line 55: Here, I think I just quoted the wrong line, @alejoe91

src/spikeinterface/core/binaryfolder.py
alejoe91: yes, but people have been saving in binary format for a while now! So I think this is needed for back-compatibility!

DradeAW: Then we should add a warning that this will not be supported in SI >= 0.100?

h-mayorquin: Ah, fair enough, makes sense!

h-mayorquin: In my view a deprecation warning makes sense when loading old files makes sense.

DradeAW: I mean it's already implemented in the `WaveformExtractor` load

alejoe91: What would a deprecation warning do here? Let's say one saved a recording object with 0.97 and stored the folder ina remote drive/cloud. I don't think we need to require them to overwrite the object with a new version, but we should still be able to load it! 

If one uses a newer version of SI and passed the `num_chan` argument, (s)he will get the warning anyways!

DradeAW: > I don't think we need to require them to overwrite the object with a new version, but we should still be able to load it!

Isn't that what we do with the `WaveformExtractor`?

alejoe91: Where?

src/spikeinterface/core/binaryfolder.py
```diff
@@ -52,6 +53,7 @@ def __init__(self, folder_path):
         self._bin_kwargs = d["kwargs"]
         if "num_channels" not in self._bin_kwargs:
             assert "num_chan" in self._bin_kwargs, "Cannot find num_channels or num_chan in binary.json"
+            warnings.warn(f"`num_chan` contained in folder {folder_path} is deprecated: it will not be redeable starting with version 0.100.")
```
alejoe91: We should maintain back-compatibility even after 0.100!

alejoe91: So @DradeAW @h-mayorquin I would remove the deprecation warning at loading time...

src/spikeinterface/core/binaryfolder.py
DradeAW: Ah I guess it was removed. In the past when the version changed there was a warning with the previously-loaded waveform extractors :)

src/spikeinterface/core/binaryfolder.py
DradeAW: Removed.

alejoe91 (approved):

PR #1777, opened 2023-07-04
Check probe in waveform extractor and add `has_probe` method

### LINKED ISSUE #1759: No Probe attached Error when using extract_waveforms in mode folder.
Hi!
I think I am facing a bug because I try to extract some waveforms from a recording which does not have a probe, and I get this error

image

The cause seems to be that the waveforms extractor checks if there is a probe to save as json, but the `BaseRecording.getprobegroup` method returns the value error if there is no probe.

Besides, it works well with in memory extraction, but this is not what I need.

MWE:
```python
import numpy as np
import spikeinterface as si

# Create synthetic data
num_channels = 4
sampling_frequency = 10000
duration = 1.0
num_samples = int(sampling_frequency * duration)
data = np.random.randn(num_samples, num_channels)

# Create a recording extractor
recording = si.NumpyRecording(traces_list=data, sampling_frequency=sampling_frequency)

# Create synthetic spike sorting results
spike_frames = np.array([100, 300, 500, 700])
spike_labels = np.array([0, 1, 1, 0])

# Create a sorting extractor
sorting = si.NumpySorting.from_times_labels(
    [spike_frames], [spike_labels], sampling_frequency=sampling_frequency
)
# It is not dumpable and does not make sense to save it with extract_waveforms
# but this is not the point of this example

# Extract waveforms for the first sorted spike times
waveforms = si.extract_waveforms(recording, sorting, "test", overwrite=True)
```

Thanks for your help!

alejoe91:
Fixes #1759 by checking for the probe for every `mode`

### DIFF
--- a/src/spikeinterface/core/baserecordingsnippets.py
+++ b/src/spikeinterface/core/baserecordingsnippets.py
@@ -223,10 +223,7 @@ class BaseRecordingSnippets(BaseExtractor):
                 raise ValueError("There is no Probe attached to this recording. Use set_probe(...) to attach one.")
             else:
                 warn("There is no Probe attached to this recording. Creating a dummy one with contact positions")
-                ndim = positions.shape[1]
-                probe = Probe(ndim=ndim)
-                probe.set_contacts(positions=positions, shapes="circle", shape_params={"radius": 5})
-                probe.set_device_channel_indices(np.arange(self.get_num_channels(), dtype="int64"))
+                probe = self.create_dummy_probe_from_locations(positions)
                 #  probe.create_auto_shape()
                 probegroup = ProbeGroup()
                 probegroup.add_probe(probe)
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -275,8 +280,34 @@ def set_dummy_probe_from_locations(self, locations, shape="circle", shape_params
...
+            Electrode shapes, by default "circle"
+        shape_params : dict, optional
+            Shape parameters, by default {"radius": 1}
+        axes : str, optional
+            If ndim is 3, indicates the axes that define the plane of the electrodes, by default "xy"
+        """
+        probe = self.create_dummy_probe_from_locations(locations, shape=shape, shape_params=shape_params, axes=axes)
         self.set_probe(probe, in_place=True)
 
+    def set_dummy_probe(self):
```
samuelgarcia, line 303: What is the use case of this ?
I think having a wrirng probe automatically is dangerous. I think the end have to do it himself.

alejoe91:
@samuelgarcia ready!

samuelgarcia (approved):

PR #1780, opened 2023-07-04
Simplify CellExplorer sorting with pymat reader

h-mayorquin:
When I was working in a conversion I started used this reader for matlab files. Pymat reader uses a consistent interface for matlab files that are in the old format and in hdf5. This allows us to simplify the if conditions code that we have in `CellExplorerSortingExtractor` and have a cleaner code.

Plus, I added some conditions to avoid reading waveform data when possible and a third condition for extracting sampling frequency from the session file when this is available.

### DIFF
--- a/src/spikeinterface/extractors/cellexplorersortingextractor.py
+++ b/src/spikeinterface/extractors/cellexplorersortingextractor.py
@@ -112,7 +112,7 @@ class CellExplorerSortingExtractor(BaseSorting):
             "filtWaveform_all",
             "timeWaveform_all",
             "channels_all",
-        ]
+        ]  # Note that the ignore_fields only works in matlab files that were saved on hdf5 files
         matlab_data = read_mat(self.spikes_cellinfo_path, ignore_fields=ignore_fields)
         spikes_data = matlab_data["spikes"]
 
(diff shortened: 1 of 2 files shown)


src/spikeinterface/extractors/cellexplorersortingextractor.py
```diff
@@ -149,15 +150,29 @@ def __init__(
         sorting_segment = CellExplorerSortingSegment(spiketrains_dict, unit_ids)
         self.add_sorting_segment(sorting_segment)
 
-        self.extra_requirements.append("scipy")
+        self.extra_requirements.append(["scipy", "pymatreader"])
```
alejoe91, line 153: Do we still need `scipy` then?

h-mayorquin: This other library usew scipy under the hood, where do we use this extra requirements btw?

alejoe91: In the containers, to install all requirements to reload the extractor ;) If `scipy` is a requirement of `pymatreader`, then we can drop it!

h-mayorquin: Thanks, let's drop it.

h-mayorquin: Removed.

alejoe91 (approved):

PR #1783, opened 2023-07-05
Handle empty units in quality metrics

### LINKED ISSUE #1760: Error when computing quality metrics for Kilosort 2.5
Hi, I'm currently trying to compute quality metrics for kilosort 2.5 sorting and waveform results. However, I get an odd error (pasted below) when computing the metrics. The function works fine for other sorters (such as Kilosort3). Upon inspection, it seems as though the function returns an error when computing the isi violations. I tried to bypass calculating this metric and export to phy, but I get another error when trying to open the gui. Is there any fix to the original bug when trying to compute the quality metrics? I feel that if I can fix that, my export to phy should work correctly and I should be able to view the results in the gui. 

Error: 
IndexError                                Traceback (most recent call last)
Cell In[13], line 1
----> 1 qc_ks25_hour = compute_quality_metrics(ks25_hour_we_sparse)

File ~\anaconda3\envs\si_env\lib\site-packages\spikeinterface\qualitymetrics\quality_metric_calculator.py:202, in compute_quality_metrics(waveform_extractor, load_if_exists, metric_names, qm_params, peak_sign, seed, sparsity, skip_pc_metrics, verbose, **job_kwargs)
    199     qmc = QualityMetricCalculator(waveform_extractor)
    200     qmc.set_params(metric_names=metric_names, qm_params=qm_params, peak_sign=peak_sign, seed=seed,
    201                    sparsity=sparsity, skip_pc_metrics=skip_pc_metrics)
--> 202     qmc.run(verbose=verbose, **job_kwargs)
    204 metrics = qmc.get_data()
    206 return metrics

File ~\anaconda3\envs\si_env\lib\site-packages\spikeinterface\core\waveform_extractor.py:1691, in BaseWaveformExtractorExtension.run(self, **kwargs)
   1690 def run(self, **kwargs):
-> 1691     self._run(**kwargs)
   1692     self._save(**kwargs)

File ~\anaconda3\envs\si_env\lib\site-packages\spikeinterface\qualitymetrics\quality_metric_calculator.py:110, in QualityMetricCalculator._run(self, verbose, **job_kwargs)
    107 func = _misc_metric_name_to_func[metric_name]
    109 params = qm_params[metric_name] if metric_name in qm_params else {}
--> 110 res = func(self.waveform_extractor, **params)
    111 # QM with uninstall dependencies might return None
    112 if res is not None:

File ~\anaconda3\envs\si_env\lib\site-packages\spikeinterface\qualitymetrics\misc_metrics.py:412, in compute_sliding_rp_violations(waveform_extractor, bin_size_ms, window_size_s, exclude_ref_period_below_ms, max_ref_period_ms, contamination_values)
    409         spike_train = sorting.get_unit_spike_train(unit_id=unit_id, segment_index=segment_index)
    410         spike_train_list.append(spike_train)
--> 412     contamination[unit_id] = slidingRP_violations(spike_train_list, fs, duration, bin_size_ms, window_size_s,
    413                                                   exclude_ref_period_below_ms, max_ref_period_ms,
    414                                                   contamination_values)
    416 return contamination

File ~\anaconda3\envs\si_env\lib\site-packages\spikeinterface\qualitymetrics\misc_metrics.py:943, in slidingRP_violations(spike_samples, sample_rate, duration, bin_size_ms, window_size_s, exclude_ref_period_below_ms, max_ref_period_ms, contamination_values, return_conf_matrix)
    941 correlogram = None
    942 for spike_samples in spike_samples_list:
--> 943     c0 = correlogram_for_one_segment(spike_samples, np.zeros(len(spike_samples), dtype='int8'),
    944                                      bin_size=max(int(bin_size_ms / 1000 * sample_rate), 1), # convert to sample counts
    945                                      window_size=int(window_size_s * sample_rate))[0, 0]
    946     if correlogram is None:
    947         correlogram = c0

IndexError: index 0 is out of bounds for axis 0 with size 0

Thanks!

JoeZiminski:
This PR closes #1760 by handling the empty-unit case in quality metrics's `compute_sliding_rp_violations` and `compute_isi_violations`. 

It also shows a warning if empty units exist when running `compute_quality_metrics` to indicate the values for these units are un-interpretable.

To ease this, two new conveience functions `get_non_empty_units` and `get_empty_units` are introduced on basesorter.

Happy to make any required changes!

### DIFF
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -292,10 +292,10 @@ class BaseSorting(BaseExtractor):
         BaseSorting
             Sorting object with non-empty units
         """
-        non_empty_units = self.get_non_empty_units()
+        non_empty_units = self.get_non_empty_unit_ids()
         return self.select_units(non_empty_units)
 
-    def get_non_empty_units(self):
+    def get_non_empty_unit_ids(self):
         non_empty_units = []
         for segment_index in range(self.get_num_segments()):
             for unit in self.get_unit_ids():
@@ -304,9 +304,9 @@ class BaseSorting(BaseExtractor):
         non_empty_units = np.unique(non_empty_units)
         return non_empty_units
 
-    def get_empty_units(self):
+    def get_empty_unit_ids(self):
         unit_ids = self.get_unit_ids()
-        empty_units = unit_ids[~np.isin(unit_ids, self.get_non_empty_units())]
+        empty_units = unit_ids[~np.isin(unit_ids, self.get_non_empty_unit_ids())]
         return empty_units
 
     def frame_slice(self, start_frame, end_frame, check_spike_frames=True):
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -292,13 +292,22 @@ class BaseSorting(BaseExtractor):
         BaseSorting
             Sorting object with non-empty units
         """
-        units_to_keep = []
+        non_empty_units = self.get_non_empty_units()
+        return self.select_units(non_empty_units)
+
+    def get_non_empty_units(self):
+        non_empty_units = []
         for segment_index in range(self.get_num_segments()):
             for unit in self.get_unit_ids():
                 if len(self.get_unit_spike_train(unit, segment_index=segment_index)) > 0:
-                    units_to_keep.append(unit)
-        units_to_keep = np.unique(units_to_keep)
-        return self.select_units(units_to_keep)
+                    non_empty_units.append(unit)
+        non_empty_units = np.unique(non_empty_units)
+        return non_empty_units
+
+    def get_empty_units(self):
+        unit_ids = self.get_unit_ids()
+        empty_units = unit_ids[~np.isin(unit_ids, self.get_non_empty_units())]
+        return empty_units
 
     def frame_slice(self, start_frame, end_frame, check_spike_frames=True):
         from spikeinterface import FrameSliceSorting
--- a/src/spikeinterface/qualitymetrics/quality_metric_calculator.py
+++ b/src/spikeinterface/qualitymetrics/quality_metric_calculator.py
@@ -1,5 +1,5 @@
 """Classes and functions for computing multiple quality metrics."""
-
+import warnings
 from copy import deepcopy
 
 import numpy as np
@@ -91,6 +91,7 @@ class QualityMetricCalculator(BaseWaveformExtractorExtension):
         progress_bar = job_kwargs["progress_bar"]
 
         unit_ids = self.sorting.unit_ids
+        self._warn_on_empty_units()
         import pandas as pd
 
         metrics = pd.DataFrame(index=unit_ids)
@@ -138,6 +139,14 @@ class QualityMetricCalculator(BaseWaveformExtractorExtension):
                 metrics[col] = pd.Series(values)
 
         self._extension_data["metrics"] = metrics
+
+    def _warn_on_empty_units(self):
+        empty_units = self.sorting.get_empty_units()
+        if np.any(empty_units):
+            warnings.warn(
+                f"Units {empty_units} are empty. Quality metrcs will be non-interpretable "
+                f"for these units.\n To remove empty units, use `sorting.remove_empty_units()`."
+            )
 
     def get_data(self):
         """
--- a/src/spikeinterface/qualitymetrics/quality_metric_calculator.py
+++ b/src/spikeinterface/qualitymetrics/quality_metric_calculator.py
@@ -92,7 +92,7 @@ class QualityMetricCalculator(BaseWaveformExtractorExtension):
 
         unit_ids = self.sorting.unit_ids
         non_empty_unit_ids = self.sorting.get_non_empty_unit_ids()
-        empty_unit_ids = self.sorting.get_empty_unit_ids()
+        empty_unit_ids = unit_ids[~np.isin(unit_ids, non_empty_unit_ids)]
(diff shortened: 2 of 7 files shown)


alejoe91:
@JoeZiminski can you fix the error in the widgets test?

src/spikeinterface/core/basesorting.py
```diff
@@ -292,13 +292,22 @@ def remove_empty_units(self):
         BaseSorting
             Sorting object with non-empty units
         """
-        units_to_keep = []
+        non_empty_units = self.get_non_empty_units()
+        return self.select_units(non_empty_units)
+
+    def get_non_empty_units(self):
```
samuelgarcia, line 298: I think I would prefer:
get_non_empty_unit_ids

src/spikeinterface/core/basesorting.py
```diff
@@ -292,13 +292,22 @@ def remove_empty_units(self):
...
             for unit in self.get_unit_ids():
                 if len(self.get_unit_spike_train(unit, segment_index=segment_index)) > 0:
-                    units_to_keep.append(unit)
-        units_to_keep = np.unique(units_to_keep)
-        return self.select_units(units_to_keep)
+                    non_empty_units.append(unit)
+        non_empty_units = np.unique(non_empty_units)
+        return non_empty_units
+
+    def get_empty_units(self):
```
samuelgarcia, line 307: I think I would prefer:
get_empty_unit_ids

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -91,6 +91,7 @@ def _run(self, verbose, **job_kwargs):
         progress_bar = job_kwargs["progress_bar"]
 
         unit_ids = self.sorting.unit_ids
+        self._warn_on_empty_units()
```
samuelgarcia, line 94: I think here instead of warning we should compute only metrics on non empty and then will the final dataframe with nan.
what do you think ?

alejoe91: Done!

alejoe91:
@JoeZiminski @samuelgarcia I implemented the following

- added `remove_empty_units` to the read Phy/KS (True by defautl for Kilosort, but False for Phy)
- handle QM for empty units when running `compute_quality_metrics` separately, by only computing QM for non-empty units and set all QM for emoty units to NaNs.

Let me know if you have suggestions!

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -91,6 +91,14 @@ def _run(self, verbose, **job_kwargs):
...
 
         unit_ids = self.sorting.unit_ids
+        non_empty_unit_ids = self.sorting.get_non_empty_unit_ids()
+        empty_unit_ids = self.sorting.get_empty_unit_ids()
+        if len(empty_unit_ids) > 0:
+            warnings.warn(
+                f"Units {empty_unit_ids} are empty. Quality metrcs will be set to NaN "
+                f"for these units.\n To remove empty units, use `sorting.remove_empty_units()`."
+            )
+
```
samuelgarcia, line 101: This will be quite annoying I think repeat over time.
Putting nan is enough no ?
Or we can add a column in the dataframe with a flag empty units ? An empty units is also some kind af a "metrics" we can use later for removal!

alejoe91: This will only be printed once! And I think it's good to know! Maybe the user has forgotten to remove them...

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -91,6 +91,14 @@ def _run(self, verbose, **job_kwargs):
         progress_bar = job_kwargs["progress_bar"]
 
         unit_ids = self.sorting.unit_ids
+        non_empty_unit_ids = self.sorting.get_non_empty_unit_ids()
+        empty_unit_ids = self.sorting.get_empty_unit_ids()
```
samuelgarcia, line 95: Counting spikes have a cost.
Here we are doing it twice!
And then later we are doing "compute_num_spikes".
So we are counting how many spikes 3 times in this functions.
I think once should be enough

alejoe91: done!

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
easy-electrophysiology: I think it's handy to have this information available in the log as it might not be initially clear why there are NaN in the quality metrics output, and checking the log will save looking through the code. It will only shown in the case that the automatic removal of empty units is set `False` so should not show in most cases anway.

alejoe91: Agreed! Let's keep it

JoeZiminski:
Looks good I tested on the dataset that was causing problems, default removing empty units works as expected and when keeping them in, the `quality_metrics.csv` is filled with `NaN`. Cheers!

samuelgarcia (approved):

DradeAW:
I really would have liked having the introduction of `unit_ids=None` in the quality metrics at the end.

You guys broke all my code in a bunch of places!!

samuelgarcia:
@DradeAW : what do you mean ? how this PR can break your code ?
when unit_ids=None we have the same behavior as before no ?

alejoe91:
Argument order! You have to use keyword ARGS because unit_ids is at the top. I can change it...but I think it's a good position

PR #1794, opened 2023-07-07
Move unit_ids argument at the end of qm low-level functions

alejoe91:
@DradeAW this should make the change back-compatible! But you SHOULD use keyworkd arguments!!


samuelgarcia (approved):

DradeAW:
Thanks :)
I’ll make the changes to use keyword arguments!

PR #1796, opened 2023-07-07
Fix deploy

samuelgarcia:


alejoe91 (approved):

.github/workflows/publish-to-pypi.yml
```diff
@@ -22,6 +22,7 @@ jobs:
           pip install pytest
           pip install zarr
           pip install setuptools wheel twine build
+          pip install -e .[test_core]
           pip install -e .
```
h-mayorquin, line 26: Isn't this redundant after btw?

alejoe91: no, otherwise it doesn't install psutils

alejoe91: (that was the first failure)

h-mayorquin: I meant the second one. Line 26.

alejoe91: sorry. right!

PR #1799, opened 2023-07-07
Waveform tools speedup

samuelgarcia:
Waveforms extarctor was extracting waveforms into one file (or one sharemem) per unit.

This lead to "too many file open" in many situtation.
Here I propose another implementation into a unique file.

PROS:
  * easier
  * faster
  * no more "too many" files
  * waveforms folder will be more easy to manipulate because few files.

CONS:
  * need to make bool mask allong spike axis 0 to get waveforms for a unit
  * when sparsity channel are not align inter units.

The idea of this PR is only to add the underlying machinery.
This PR do not change the WaveformExtactor internals.
It will be done in a separted PR using this new `extract_waveforms_to_single_buffer()` to fix the  "too many file open"


src/spikeinterface/core/waveform_tools.py
```diff
@@ -257,33 +258,33 @@ def distribute_waveforms_to_buffers(
...
     processor.run()
 
 
 distribute_waveforms_to_buffers.__doc__ = distribute_waveforms_to_buffers.__doc__.format(_shared_job_kwargs_doc)
 
 
 # used by ChunkRecordingExecutor
-def _init_worker_waveform_extractor(
-    recording, unit_ids, spikes, wfs_arrays_info, nbefore, nafter, return_scaled, inds_by_unit, mode, sparsity_mask
+def _init_worker_ditribute_buffers(
```
alejoe91, line 286:
```suggestion
def _init_worker_distribute_buffers(
```

src/spikeinterface/core/waveform_tools.py
```diff
@@ -328,7 +329,7 @@ def _init_worker_waveform_extractor(
 
 
 # used by ChunkRecordingExecutor
-def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx):
+def _worker_ditribute_buffers(segment_index, start_frame, end_frame, worker_ctx):
```
alejoe91, line 332:
```suggestion
def _worker_distribute_buffers(segment_index, start_frame, end_frame, worker_ctx):
```

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,288 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+    sparsity_mask=None,
+    copy=False,
+    job_name=None,
+    **job_kwargs,
+):
+    """
+    Allocate a single buffer (memmap or or shared memory) and then distribute every waveform into it.
+
+    Contrary to extract_waveforms_to_buffers() all waveforms are extracted in the same buffer, so the spike vector is
+    needed to recover waveforms unit by unit. Importantly in case of sparsity, the channel are not aligned across
```
alejoe91, line 420:
```suggestion
    needed to recover waveforms unit by unit. Importantly in case of sparsity, the channels are not aligned across
```

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,288 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+        s0 = np.searchsorted(spikes["segment_index"], segment_index)
+        s1 = np.searchsorted(spikes["segment_index"], segment_index + 1)
+        segment_slices.append((s0, s1))
+    worker_ctx["segment_slices"] = segment_slices
+
+    return worker_ctx
+
+
+# used by ChunkRecordingExecutor
+def _worker_ditribute_single_buffer(segment_index, start_frame, end_frame, worker_ctx):
```
alejoe91, line 579:
```suggestion
def _worker_distribute_single_buffer(segment_index, start_frame, end_frame, worker_ctx):
```

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,288 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+            if shm is not None:
+                # release all sharedmem buffer
+                # empty array have None
+                shm.unlink()
+            return all_waveforms.copy()
+        else:
+            return all_waveforms, wf_array_info
+
+
+def _init_worker_ditribute_single_buffer(
```
alejoe91, line 540:
```suggestion
def _init_worker_distribute_single_buffer(
```

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,288 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+
+    num_spikes = spikes.size
+    if sparsity_mask is None:
+        num_chans = recording.get_num_channels()
+    else:
+        num_chans = max(np.sum(sparsity_mask, axis=1))
+    shape = (num_spikes, nsamples, num_chans)
+
+    if mode == "memmap":
+        filename = str(folder / f"all_waveforms.npy")
```
alejoe91, line 483:
```suggestion
        filename = str(folder / f"waveforms.npy")
```
I would call it just `waveforms` because there is no guarantee that all spikes are used. No?

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,288 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+
+    inds_by_unit = {}
+    for unit_ind, unit_id in enumerate(unit_ids):
+        (inds,) = np.nonzero(spikes["unit_index"] == unit_ind)
+        inds_by_unit[unit_id] = inds
+
+    if num_spikes > 0:
+        # and run
+        func = _worker_ditribute_single_buffer
+        init_func = _init_worker_ditribute_single_buffer
```
alejoe91, line 508:
```suggestion
        func = _worker_distribute_single_buffer
        init_func = _init_worker_distribute_single_buffer
```

src/spikeinterface/core/waveform_tools.py
```diff
@@ -257,33 +258,33 @@ def distribute_waveforms_to_buffers(
         inds_by_unit[unit_id] = inds
 
     # and run
-    func = _waveform_extractor_chunk
-    init_func = _init_worker_waveform_extractor
+    func = _worker_ditribute_buffers
+    init_func = _init_worker_ditribute_buffers
```
alejoe91, line 262:
```suggestion
    func = _worker_distribute_buffers
    init_func = _init_worker_distribute_buffers
```

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,288 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+    l1 = i1 + s0
+
+    if l1 > l0:
+        start = spikes[l0]["sample_index"] - nbefore
+        end = spikes[l1 - 1]["sample_index"] + nafter
+
+        # load trace in memory
+        traces = recording.get_traces(
+            start_frame=start, end_frame=end, segment_index=segment_index, return_scaled=return_scaled
+        )
```
alejoe91, line 620: Or use `get_traces_with_margin` instead

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,288 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+        wf_array_info = (shm, shm_name, dtype.str, shape)
+    else:
+        raise ValueError("allocate_waveforms_buffers bad mode")
+
+    job_kwargs = fix_job_kwargs(job_kwargs)
+
+    inds_by_unit = {}
+    for unit_ind, unit_id in enumerate(unit_ids):
+        (inds,) = np.nonzero(spikes["unit_index"] == unit_ind)
+        inds_by_unit[unit_id] = inds
```
alejoe91, line 503: I think this is not used
```suggestion
```

alejoe91 (changes requested):
Looks great @samuelgarcia 

A couple of small comments:
- `ditribute` -> `distribute`
- I would use `get_traces_with_margin` instead of `get_traces`. I think it makes the margin handling more readable

src/spikeinterface/core/waveform_tools.py
alejoe91: well actually the segment does not handle scaled and margins can be asymmetric...so nevermind!

src/spikeinterface/core/tests/test_waveform_tools.py
```diff
@@ -52,96 +54,88 @@ def test_waveform_tools():
...
+    some_sparsity = [
+        dict(sparsity_mask=None),
+        dict(sparsity_mask=np.random.randint(0, 2, size=(unit_ids.size, recording.channel_ids.size), dtype="bool")),
+    ]
 
     # memmap mode
-    list_wfs = []
+    list_wfs_dense = []
+    list_wfs_sparse = []
     for j, job_kwargs in enumerate(some_job_kwargs):
```
h-mayorquin, line 79: As an another story against single variable names. 

I want to know why are you using j, k, l and here, what do I do?

I have to inspect and look for those three letters in all the linew that you are added. I can't do search on the term because is a single letter. I takes me a couple of minutes to realize that they are used in line 85 to name the file and that seems (but I am not sure) that is the only use.

If you use something explicit like mode_kwargs_index, sparsity_kwargs_index then I just ctrl+f and the term and I am done. 

The benefit of single letter variables? an aesthetic preference

h-mayorquin:
I started reviewing this, I will finish in the following days.

src/spikeinterface/core/tests/test_waveform_tools.py
```diff
@@ -52,96 +54,88 @@ def test_waveform_tools():
     unit_ids = sorting.unit_ids
 
     some_job_kwargs = [
-        {},
         {"n_jobs": 1, "chunk_size": 3000, "progress_bar": True},
         {"n_jobs": 2, "chunk_size": 3000, "progress_bar": True},
     ]
```
h-mayorquin, line 59: Could you use parameterize for this tests instead of the triple loop for the kwargs inside the test? I will make the code simpler to read (not as nested) and it will also make the tests independent and it will be simpler to spot which one is failing?

h-mayorquin: I want to debug and stop in a spcific part of the code for a specific example and I am not able to with the current configuration.

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,288 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+            else:
+                mask = sparsity_mask[unit_index, :]
+                wf = wf[:, mask]
+                all_waveforms[spike_ind, :, : wf.shape[1]] = wf
+
+        if worker_ctx["mode"] == "memmap":
+            all_waveforms.flush()
+
+
+def split_waveforms_by_units(unit_ids, spikes, all_waveforms, sparsity_mask=None, folder=None):
```
h-mayorquin, line 638: Wheret this will be caled from?

src/spikeinterface/core/waveform_tools.py
samuelgarcia: I think we should even give the filename in the signature instead of give folder no ?

src/spikeinterface/core/tests/test_waveform_tools.py
samuelgarcia: I think the test was already like this.
Sometimes it is cools to also have loops when you share some variable this avoid long pytest.fixture chain + pytest.parametrize storie.
I think this is a matter of taste no ?

src/spikeinterface/core/tests/test_waveform_tools.py
samuelgarcia: For the general case you are right.
But nested loop with i, j, k, l is a classic. no ?
I think really like to have single letter variable name for counter.
For this simple test this is OK

src/spikeinterface/core/waveform_tools.py
samuelgarcia: done

src/spikeinterface/core/waveform_tools.py
samuelgarcia: done

src/spikeinterface/core/waveform_tools.py
samuelgarcia: done

src/spikeinterface/core/waveform_tools.py
samuelgarcia: done

src/spikeinterface/core/waveform_tools.py
samuelgarcia: good catch!!

src/spikeinterface/core/waveform_tools.py
samuelgarcia: This is a future need: we will be able to replace in wavforms_extactors:
`extract_waveforms_to_buffers()`
by
`extract_waveforms_to_single_buffer()` + `split_waveforms_by_units()`
At the moment this is only used in the test to compare the two approach.

src/spikeinterface/core/tests/test_waveform_tools.py
samuelgarcia: Finally I remove the folder kwargs and put file_path instead which is even better the name is external.
We will need this in several context.

src/spikeinterface/core/waveform_tools.py
samuelgarcia: Finally I remove the folder kwargs and put file_path instead which is even better the name is external.
We will need this in several context.
I think this is better. Tell me.

samuelgarcia:
@yger : when this will be merge could you rewrite this function here https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/sortingcomponents/tools.py#L21
with the use of this new `extract_waveforms_to_single_buffer()` so that you can explicitly give a sparse mask which is the diagonal instead of a radius.

samuelgarcia:
@alejoe91 I put back the test of sharemem on window and now it is working!!!

src/spikeinterface/core/tests/test_waveform_tools.py
h-mayorquin: To some deegree it is, @samuelgarcia. Specially for using many fixtures I find myself agreeing with some of your taste arguments (it looks like a lot of clutter) and I have a lot of sympathy for your own usability concerns (adds more boilerplate to your way of debugging by using a main call at the end of the testing file, what a pain). For the latter I tried to help you writing that boilerplate myself for some of the functionst that I have modified.

Maybe we can reach a compromise if you can suggest some workarounds for my debuggin workflow? 
I really want to stop and debugg a specific line with some set of parameters. I don't know of a simple way of doing it this way. I would need to add an if block to check that the parameters are the ones that I want to look at and then set the traces of the debugger to that new if line which is a cumbersome, error prone and labrious task.

With parameterization I can just click a line (even from my IDE) and run the specific mode that I want to test or debugg:
image
(see on the left that I can just run the specific test that I want)

If you really dislike parameterize maybe we could have one loop for test or break them appart in some other way? That will make my life easier and it will cost you only some duplication of code (in practice copy paste).  Plus, it will make the tests closer in spirit to the ideal of unit testing. I am open to other suggestions but I would like you to recognize my troubles the way that I recognize yours.

I would also like to point out that my debugging workflow by setting traces is more standard than yours. It is supported in the [standard library](https://docs.python.org/3/library/pdb.html) and, as you can see above, it is already well integrated in IDEs like VScode. Future developers are more likely to be using that type of workflow. 

Plus, using standard tooling of pytest like parameterize opens the doors to certain futures that we don't have available right now. Two examples: 1) The reporting tool that I built for the CI to see which tests take more time is more useful when the tests are clearly separated. Right now, all the time taken by the tests is lumped into one thing and we can't compare paramters choices to get a quick sense if something runs faster and slower in other architectures for example 2) It does not apply here but there are packages like [xdist](https://pypi.org/project/pytest-xdist/) that we could use to improve the running times of the CI and they work better when the tests are more atomic.

src/spikeinterface/core/tests/test_waveform_tools.py
h-mayorquin: Sorry, Sam, I should have written this on my diary and no in your PR.

[You are not alone
](https://www.reddit.com/r/AskProgramming/comments/e9sfds/what_are_the_acceptable_uses_for_single_letter/)
Lots of people love using single letter variables for counter and will die on that hill. 

That said, I have never seen any argument for keeping the single letter convention other than aestethical preference and I was just pointing out costs by means of a specific story. In this case because the scope of the function is a little too large and it was not obvious where they are used.

src/spikeinterface/core/tests/test_waveform_tools.py
samuelgarcia: OK: one point for you.
I will do this next time I promise.
But I need this in the main quite soon and I am quick busy....

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+        while (in_seg_spikes[i0]["sample_index"] - nbefore) < 0 and (i0 != i1):
+            i0 = i0 + 1
+        while (in_seg_spikes[i1 - 1]["sample_index"] + nafter) > seg_size and (i0 != i1):
+            i1 = i1 - 1
+
+    # slice in absolut in spikes vector
+    l0 = i0 + s0
+    l1 = i1 + s0
+
+    if l1 > l0:
```
h-mayorquin, line 605: Isn't this the same if as above? in line 592? You are just translating.  I think that you can create an if for equality in line 592 and do an early return. 

This will avoid the nesting.

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+    all_waveforms = worker_ctx["all_waveforms"]
+
+    seg_size = recording.get_num_samples(segment_index=segment_index)
+
+    s0, s1 = segment_slices[segment_index]
+    in_seg_spikes = spikes[s0:s1]
+
+    # take only spikes in range [start_frame, end_frame]
+    # this is a slice so no copy!!
+    i0 = np.searchsorted(in_seg_spikes["sample_index"], start_frame)
```
h-mayorquin, line 590:
```suggestion
    i0 = np.searchsorted(in_seg_spikes["sample_index"], start_frame, side="right")
```
Shouldn't this be like this?

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+    """
+    Allocate a single buffer (memmap or or shared memory) and then distribute every waveform into it.
+
+    Contrary to extract_waveforms_to_buffers() all waveforms are extracted in the same buffer, so the spike vector is
+    needed to recover waveforms unit by unit. Importantly in case of sparsity, the channels are not aligned across
+    units.
+
+    Important note: for the "shared_memory" mode wf_array_info contains reference to
+    the shared memmory buffer, this variable must be reference as long as arrays as used.
+    And this variable is also returned.
```
h-mayorquin, line 425: Why do you need to keep a reference to the variable? I feel that there is an or missing after "this variable must be referenced* as long as the array is used" because / or " short exaplanation."

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
                     wfs[pos, :, :] = wf[:, sparsity_mask[unit_ind]]
 
 
+def extract_waveforms_to_single_buffer(
+    recording,
+    spikes,
```
h-mayorquin, line 403: As we lack type for this can we name this variable spike_vector. Will be easier to ctrl+f in the library if someone wants to see the struct without having to read the code.

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+        In case of memmap mode, file to save npy file.
+    dtype: numpy.dtype
+        dtype for waveforms buffer
+    sparsity_mask: None or array of bool
+        If not None shape must be must be (len(unit_ids), len(channel_ids))
+    copy: bool
+        If True (default), the output shared memory object is copied to a numpy standard array.
+        If copy=False then arrays_info is also return. Please keep in mind that arrays_info
+        need to be referenced as long as waveforms_by_units will be used otherwise it will be very hard to debug.
+        Also when copy=False the SharedMemory will need to be unlink manually
```
h-mayorquin, line 455: What's the use case for offering this as an option. If there is, maybe you can mention here in the docstring. If there is None, maybe we can eliminate this parameter.

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+    num_spikes = spikes.size
+    if sparsity_mask is None:
+        num_chans = recording.get_num_channels()
+    else:
+        num_chans = max(np.sum(sparsity_mask, axis=1))
+    shape = (num_spikes, nsamples, num_chans)
+
+    if mode == "memmap":
+        all_waveforms = np.lib.format.open_memmap(file_path, mode="w+", dtype=dtype, shape=shape)
+        wf_array_info = str(file_path)
```
h-mayorquin, line 484: You say that `wf_array_info` is a dict above (which I think is a good choice) but then you go and store a string or a tuple on it. I think you should store either the string or the tuple as items of the dict so you deliver in your promise.

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+        N samples before spike
+    nafter: int
+        N samples after spike
+    mode: str
+        Mode to use ('memmap' | 'shared_memory')
+    return_scaled: bool
+        Scale traces before exporting to buffer or not.
+    file_path: str or path
+        In case of memmap mode, file to save npy file.
+    dtype: numpy.dtype
```
h-mayorquin, line 447: To me it makes sense that the buffer array inhertings the type of the recording. Allowing this flexbility without testing seems unecessary to me and increasing the possbility of bugs.  Am I missing something?

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+    unit_ids: list or numpy array
+        List of unit ids
+    spikes: numpy array
+        The spike vector
+    all_waveforms : numpy array
+        Single buffer containing all waveforms
+    sparsity_mask : None or numpy array
+        Optionally the boolean sparsity mask
+    folder : None or str or Path
+        If a folde ri sgiven all
```
h-mayorquin, line 645: Typo here.

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+        folder = Path(folder)
+    waveforms_by_units = {}
+    for unit_index, unit_id in enumerate(unit_ids):
+        mask = spikes["unit_index"] == unit_index
+        if sparsity_mask is not None:
+            chan_mask = sparsity_mask[unit_index, :]
+            num_chans = np.sum(chan_mask)
+            wfs = all_waveforms[mask, :, :][:, :, :num_chans]
+        else:
+            wfs = all_waveforms[mask, :, :]
```
h-mayorquin, line 663: Why do you increase the dimensionality by one if mask is None?

src/spikeinterface/core/waveform_tools.py
h-mayorquin: Actually I am thinking in how this interacts with return_scaled...

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+
+    # take only spikes in range [start_frame, end_frame]
+    # this is a slice so no copy!!
+    i0 = np.searchsorted(in_seg_spikes["sample_index"], start_frame)
+    i1 = np.searchsorted(in_seg_spikes["sample_index"], end_frame)
+    if i0 != i1:
+        # protect from spikes on border :  spike_time<0 or spike_time>seg_size
+        # useful only when max_spikes_per_unit is not None
+        # waveform will not be extracted and a zeros will be left in the memmap file
+        while (in_seg_spikes[i0]["sample_index"] - nbefore) < 0 and (i0 != i1):
```
h-mayorquin, line 596: You can pre-filter your spikes in the segment before running np.searchsorted for this two conditions to avoid this expensive (and hard to read loop)

samuelgarcia: We need the original index for writting into the buffer.
So this is not possible or would need some extra variable.
This while loop in practice never happen unless the very first spike is very near the border. So this is not expenssive at all.
But this is a bit hard to read I admit it.
Please not not change this it was a very old bug and very hard to track.

src/spikeinterface/core/waveform_tools.py
samuelgarcia: mask is the mask on first dim and this is not None this is line 657 not no increase normally.

src/spikeinterface/core/waveform_tools.py
samuelgarcia: This if is super imprtant because it protect the recording.get_traces() to be compute in line 610.
In short if not spikes then lets skip evrything for this buffer.
Fort loading a few waveforms this make a big diff in performances.
I will add a comment for next reading.

src/spikeinterface/core/waveform_tools.py
samuelgarcia: I think it is side="left" so if the spike is exactly start_frame I want it in this buffer
(and if spike is exactly end_frame it is for next buffer)

let me double check. But then the error would be here since very very long time....

src/spikeinterface/core/waveform_tools.py
samuelgarcia: You are right, I was lazy I think.
I will use a dict.

src/spikeinterface/core/waveform_tools.py
samuelgarcia: Sharemmeory object is a dangerous object.
If you use a view if this one while your detroy the original memory you have very hard to debug stuff. (like segfault)
So a copy to a non shared mem buffer (standard numpy) is wise by default.

But in some cases like :  1. gather waveforms 2. make an average (for template) 3. remove the waveforms. Then in that case we can have this copy=False to avoid a new malloc in step 1.
This is more for internal use. (and a bit speculative at the moement, I admit)
The function that pierre need to rewritte should use this copy=False to be faster for instance.
look in the discussion.

src/spikeinterface/core/waveform_tools.py
h-mayorquin: Right, thanks. I was confused with the sparsity mask that in most places in the code you call just mask. But this is the `spikes_in_the_unit` mask.

src/spikeinterface/core/waveform_tools.py
h-mayorquin: Which original index? i0? 

Aren't you losing the original variable anyway when you modify it here? I agree that if this never happens maybe doing the vectorization is a bad idea. 

If this is an old bug that was hard to track we should have a test for this. No point on relying on your memory and analytical capabilities if we can pin down the behavior that we want and automatize its check with a test.

src/spikeinterface/core/waveform_tools.py
h-mayorquin: I think you are correct, Sam. If you want the start_frame of the chunk in the buffer it should be left.
On the other hand, to following slicing convention maybe end_frame should be -1 in i1 just below?

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+
+    num_spikes = spikes.size
+    if sparsity_mask is None:
+        num_chans = recording.get_num_channels()
+    else:
+        num_chans = max(np.sum(sparsity_mask, axis=1))
+    shape = (num_spikes, nsamples, num_chans)
+
+    if mode == "memmap":
+        all_waveforms = np.lib.format.open_memmap(file_path, mode="w+", dtype=dtype, shape=shape)
```
h-mayorquin, line 483: Why do you open this here if you open it again inside the `__init__` of the workers?

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+    unit_ids,
+    nbefore,
+    nafter,
+    mode="memmap",
+    return_scaled=False,
+    file_path=None,
+    dtype=None,
+    sparsity_mask=None,
+    copy=False,
+    job_name=None,
```
h-mayorquin, line 413: Isn't `job_name` a part of `job_kwargs` ?

src/spikeinterface/core/waveform_tools.py
h-mayorquin: Thanks for the explanation, Sam.

I think that you should only return the shared memroy object (`shm`). No need to return the whole `arrays_info` dict.

Here my suggestion including what you mentioned above:

```
        If True (default), the output shared memory object is copied to a numpy standard array and no reference to the internal shared memory object is kept.
        If copy=False then the shared memory object is also returned. Please keep in mind that the shared memory object
        need to be referenced as long as waveforms_by_units will be used otherwise it might produce segmentation faults which are hard to debug.
        Also when copy=False the SharedMemory will need to be unlink manually if proper cleanup of resources is desired.
```

h-mayorquin (commented):
Some more comments.

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+    if l1 > l0:
+        start = spikes[l0]["sample_index"] - nbefore
+        end = spikes[l1 - 1]["sample_index"] + nafter
+
+        # load trace in memory
+        traces = recording.get_traces(
+            start_frame=start, end_frame=end, segment_index=segment_index, return_scaled=return_scaled
+        )
+
+        for spike_ind in range(l0, l1):
```
h-mayorquin, line 614:
```suggestion
        for spike_index in range(l0, l1):
```

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+    s0, s1 = segment_slices[segment_index]
+    in_seg_spikes = spikes[s0:s1]
+
+    # take only spikes in range [start_frame, end_frame]
+    # this is a slice so no copy!!
+    i0 = np.searchsorted(in_seg_spikes["sample_index"], start_frame)
+    i1 = np.searchsorted(in_seg_spikes["sample_index"], end_frame)
+    if i0 != i1:
+        # protect from spikes on border :  spike_time<0 or spike_time>seg_size
+        # useful only when max_spikes_per_unit is not None
```
h-mayorquin, line 594: Another question, Sam. How does max_spikes_per_unit affect this? I was looking for this and as it is far away in code I could not see it. As far as I got it, you only get a limited number of spikes per unit but this is a border effect so I am failing to see how this two things are related.

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+        all_waveforms = np.ndarray(shape=shape, dtype=dtype, buffer=shm.buf)
+        worker_ctx["shm"] = shm
+        worker_ctx["all_waveforms"] = all_waveforms
+
+    # prepare segment slices
+    segment_slices = []
+    for segment_index in range(recording.get_num_segments()):
+        s0 = np.searchsorted(spikes["segment_index"], segment_index)
+        s1 = np.searchsorted(spikes["segment_index"], segment_index + 1)
+        segment_slices.append((s0, s1))
```
h-mayorquin, line 565: I think this is the spot where it will easier to filter for spikes whose waveform does not lie entirely in the traces with the choosen nbefore and nafter.

```python
spikes_in_segment = spikes[s0:s1]
spikes_left_bounds = spikes_in_segment["sample_index"] - nbefore,
num_spikes_with_invalid_left_bound = np.searchsorted(spikes_left_bounds, 0)
s0 += num_spikes_with_invalid_left_bound
spikes_right_bounds = spikes_in_segment["sample_index"] + nafter,
num_spikes_with_invalid_right_bound = np.searchsorted(spikes_right_bounds, num_frames)
s1 -= num_spikes_with_invalid_right_bound
s1 = np.searchsorted(spikes["segment_index"], segment_index + 1)

```

src/spikeinterface/core/tests/test_waveform_tools.py
samuelgarcia: The `_check_all_wf_equal()` at the end the tripple loop prevent to refactor this using pytest_parametrize

src/spikeinterface/core/waveform_tools.py
samuelgarcia: `spikes` has been for this variable in many places since a while now. I would not chand this naming in this PR.
But wa can discuss in another one no ?

src/spikeinterface/core/waveform_tools.py
samuelgarcia: Thank you.
I added your changes.
I think it is good to return the dict to be consistent with the behavior already established of `extract_waveforms_to_buffers()`

src/spikeinterface/core/waveform_tools.py
samuelgarcia: done

src/spikeinterface/core/waveform_tools.py
samuelgarcia: done

src/spikeinterface/core/waveform_tools.py
samuelgarcia: Yes when return_scale=True the user have choice between float32 or float64 this is why we need this dtype.

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+        filename = wf_array_info
+        all_waveforms = np.load(str(filename), mmap_mode="r+")
+        worker_ctx["all_waveforms"] = all_waveforms
+    elif mode == "shared_memory":
+        from multiprocessing.shared_memory import SharedMemory
+
+        shm, shm_name, dtype, shape = wf_array_info
+        shm = SharedMemory(shm_name)
+        all_waveforms = np.ndarray(shape=shape, dtype=dtype, buffer=shm.buf)
+        worker_ctx["shm"] = shm
```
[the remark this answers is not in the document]
samuelgarcia: Yes but this is crutioal to reference the sharmem when you have a array this point to this buffer.

src/spikeinterface/core/waveform_tools.py
samuelgarcia: OK I did a patch as you proposed.

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +398,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+                # release all sharedmem buffer
+                # empty array have None
+                shm.unlink()
+            return all_waveforms.copy()
+        else:
+            return all_waveforms, wf_array_info
+
+
+def _init_worker_distribute_single_buffer(
+    recording, unit_ids, spikes, wf_array_info, nbefore, nafter, return_scaled, mode, sparsity_mask
```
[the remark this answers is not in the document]
samuelgarcia: removed

src/spikeinterface/core/waveform_tools.py
samuelgarcia: done

src/spikeinterface/core/waveform_tools.py
samuelgarcia: I did another trick finally which is similar and easy.

src/spikeinterface/core/waveform_tools.py
samuelgarcia: The max_spikes_per_unit is outside the machinery of waveform_tools.py the spikes must reduce before the call to this function. Of course it is better when selecting spikes it is good to not select spikes on border. But waveforms tools do not have the concept and the output waveforms shape is supposed to be the same shape[0]. I will ad  a not in the doc.

samuelgarcia: done

src/spikeinterface/core/waveform_tools.py
h-mayorquin: That's fine. I find that spikes by itself is not specific enough for someone that is not familar with the library (is this a list? a dict? what kind of spikes?). I think that if you came up with that data structure you should leverage on it. Plus, if we ever want to refactor, it will be easier to locate the use cases of this structure.

src/spikeinterface/core/waveform_tools.py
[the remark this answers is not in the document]
h-mayorquin: Don't you still have this in the `worker_context["wf_array_info"]`,

src/spikeinterface/core/waveform_tools.py
```diff
@@ -397,6 +391,280 @@ def _waveform_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+    seg_size = recording.get_num_samples(segment_index=segment_index)
+
+    s0, s1 = segment_slices[segment_index]
+    in_seg_spikes = spikes[s0:s1]
+
+    # take only spikes in range [start_frame, end_frame]
+    # this is a slice so no copy!!
+    # the border of segment are protected by nbefore on left an nafter on the right
+    i0 = np.searchsorted(in_seg_spikes["sample_index"], max(start_frame, nbefore))
+    i1 = np.searchsorted(in_seg_spikes["sample_index"], min(end_frame, seg_size - nafter))
```
h-mayorquin, line 592: Nice!

src/spikeinterface/core/waveform_tools.py
h-mayorquin: Thanks for explaining.

src/spikeinterface/core/waveform_tools.py
h-mayorquin: Yeah, makes sense, I understood that I was to avoid trace extraction. I was saying that you could do it with the `i0, i1` instead of the `l` but now that the other piece of code was eliminated I think is equally clear.

src/spikeinterface/core/waveform_tools.py
h-mayorquin: OK,

h-mayorquin: I think it should just return what the recorder returns by default. Do you think that would be wrong?

h-mayorquin (approved):
This looks good to me. Thanks for bearing with me, @samuelgarcia. I learned while doing this PR and I know that that's taxing in your time but this is when I can pick your brain appart and learn about the design decisions that you are making. 

Some pending issues for future PR or issues:
* I still think it will be a good idea to refer to spikes as spike_vector whenever they are, well, spike_vectors.
* Isn't the `job_name` already part of `job_kwargs`?
* I wish we had a test for not extracting spikes whose support would go outside of the borders.

src/spikeinterface/core/waveform_tools.py
h-mayorquin: Great!

src/spikeinterface/core/waveform_tools.py
h-mayorquin: Makes sense.

src/spikeinterface/core/waveform_tools.py
h-mayorquin: Which doc?

src/spikeinterface/core/waveform_tools.py
samuelgarcia: The mode w+ is creating the file.

src/spikeinterface/core/waveform_tools.py
samuelgarcia: job_name is for ChunkRecordingExecutor and is not part of job_kwargs.

h-mayorquin: Shouln't it be? 
job_kwargs is basically al the keyword arguments of `ChunkRecordingExecutor`.

src/spikeinterface/core/waveform_tools.py
samuelgarcia: The logic of dtype is in waveform extactor. I think it is good to keep the logic there.
Here
https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/waveform_extractor.py#L1353

samuelgarcia:
> I think it should just return what the recorder returns by default. Do you think that would be wrong?

When we want to return_scale=True we need another dtype so it is good to control this externally.

src/spikeinterface/core/waveform_tools.py
samuelgarcia: yes. certainlly. but not in this PR

src/spikeinterface/core/waveform_tools.py
h-mayorquin: Thanks for answering, yes, those three last points are outside of the scope of the PR.

alejoe91 (approved):

alejoe91 (approved):

PR #1804, opened 2023-07-10
Deepinterpolation revived: training, transfer, and inference with more flexibility

alejoe91:
Refactored and extended the `deepinterpolation` submodule.

The existing code was only compatible with a pre-trained model from Neuropixels1,0 with interleaved zeros (so that the final shape was 384,2).

### Training

With this PR, one can train a DI model from ANY SpikeInterface object using the `train_deepinterpolation` function. The `desired_shape` parameter controls how the input should be reshaped to make the input image.

For example, a NP1 probe could be treated as follows:

- `desired_shape=(192,2)`: the 4 staggered columns are "aligned" into two columns of 192 electrodes
- `desired_shape=(96, 4)`: each "real" column is a different image column (in this shape the channels would need to be reordered first).

To use the previously built model `(384, 2)` , one can use the zero-channel pad to "manually" interleave zeros in the original recording. 

### Transfer

The `existing_model_path` allows one to do transfer learning from an existing model

```
fine_tuned_model = train_deepinterpolation(recording, ..., existing_model_path="path-to-existing")
```

### Inference

Same as before, but the derived shape is inferred from the network:

```
recording_di = spre.deepinterpolate(recording, model_path="path-to-model", **kwargs)
```

***NOTE:*** it is advised to zscore the recording prior to using DeepInterpolation, but this is not currently enforced.


alejoe91:
@khl02007 FYI ;)

alejoe91:
@jeromelecoq

I updated the `train_deepinteprolation` function as wollows:

- You can now pass a single recording or a list of recordings as input. In case of a recording with multiple segments (e.g., several play/pause) or multiple recordings, I modified the [`_calculate_list_samples`](https://github.com/SpikeInterface/spikeinterface/pull/1804/files#diff-34ba9ca4353bc07923598384bc472c97a03f83eb104b1d63e1b449572539f313R112-R149) so that intervals at borders are exluded.
- You can optionally pass `test_recordings` (a single recording or a recording list) which can be a different recording used for the set
- Unified the correct usage of `total_samples` (and removed overridden functions)
- Set `mean_squared_error` as default loss
- Extended the tests 

Would you have time to review it? It's ready to merge on my side :)

src/spikeinterface/preprocessing/deepinterpolation/generators.py
```diff
@@ -0,0 +1,311 @@
...
+        self.recordings = recordings
+        if len(recordings) > 1:
+            assert (
+                r.get_num_channels() == recordings[0].get_num_channels() for r in recordings[1:]
+            ), "All recordings must have the same number of channels"
+
+        total_num_samples = np.sum(r.get_total_samples() for r in recordings)
+        # In case of multiple recordings and/or multiple segments, we calculate the frame periods to be excluded (borders)
+        exclude_intervals = []
+        pre_extended = pre_frame + pre_post_omission
```
jeromelecoq, line 46: I think the SequentialGenerator might have some of this logic already?

src/spikeinterface/preprocessing/deepinterpolation/generators.py
```diff
@@ -0,0 +1,311 @@
...
+            pre_post_omission=pre_post_omission,
+            desired_shape=desired_shape,
+            batch_size=batch_size,
+            steps_per_epoch=steps_per_epoch,
+            start_frame=start_frame,
+            end_frame=end_frame,
+        )
+
+    # this is overridden to exclude samples from borders
+    def _calculate_list_samples(self, total_frame_per_movie, exclude_intervals=[]):
```
jeromelecoq, line 113: I wonder if this is needed given that you are inheriting from the SequentialGenerator

src/spikeinterface/preprocessing/deepinterpolation/train.py
```diff
@@ -0,0 +1,307 @@
...
+
+global train_func
+
+
+def train_deepinterpolation(
+    recordings: BaseRecording | list[BaseRecording],
+    model_folder: str | Path,
+    model_name: str,
+    desired_shape: tuple[int, int],
+    train_start_s: Optional[float] = None,
```
jeromelecoq, line 26: To what extent using seconds is helpful here? I am always worried about edge conditions created by wrapping code.

src/spikeinterface/preprocessing/deepinterpolation/train.py
```diff
@@ -0,0 +1,307 @@
...
+    test_recordings: Optional[BaseRecording | list[BaseRecording]] = None,
+    pre_frame: int = 30,
+    post_frame: int = 30,
+    pre_post_omission: int = 1,
+    existing_model_path: Optional[str | Path] = None,
+    verbose: bool = True,
+    nb_gpus: int = 1,
+    steps_per_epoch: int = 10,
+    period_save: int = 100,
+    apply_learning_decay: int = 0,
```
jeromelecoq, line 187: I would consider only exposing the parameters you care about, not all of them.

jeromelecoq: For example, which parameters will in practice never change?

.github/workflows/deepinterpolation.yml
```diff
@@ -0,0 +1,37 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest"]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
+      - name: Install dependencies
```
alejoe91, line 26: Test only if deepinteprolation files changed.

src/spikeinterface/preprocessing/deepinterpolation/tests/test_deepinterpolation.py
```diff
@@ -0,0 +1,166 @@
...
+    desired_shape = (num_rows, num_cols)
+    return recording, desired_shape
+
+
+@pytest.fixture
+def recording_and_shape_fixture():
+    return recording_and_shape()
+
+
+def test_deepinterpolation_generator_borders(recording_and_shape_fixture):
```
alejoe91, line 40: skip if deepinteprolation is not installed to simplify CI

.github/workflows/deepinterpolation.yml
alejoe91: We could do the same for plexon2 -  a sepratate action only triggered if files changed. Same for pytorch.

src/spikeinterface/preprocessing/deepinterpolation/train.py
alejoe91: you tell me :) In my opinion there's no harm in exposing everything with defaults!

src/spikeinterface/preprocessing/deepinterpolation/generators.py
alejoe91: to exclude borders? I couldn't find it!

src/spikeinterface/preprocessing/deepinterpolation/generators.py
alejoe91: in this case it's needed because of the borders exclusion

src/spikeinterface/preprocessing/deepinterpolation/train.py
alejoe91: I think that using second here is much more convenient. E.g. I'd rather use 600 to indicate 10 minutes, than 18000000 (easy to forget, add a 0 here)

src/spikeinterface/preprocessing/deepinterpolation/tests/test_deepinterpolation.py
alejoe91: done

samuelgarcia:
Enormous and amazing work camarade.
I hope this will be usefull and usable.
As a general comments I think I prefer a unique file for this kind preprocessing even if the file is enormous.
I understand that it was more easy while develpping to have multiple and sub folder file.

samuelgarcia (approved):

alejoe91:
> Enormous and amazing work camarade. I hope this will be usefull and usable. As a general comments I think I prefer a unique file for this kind preprocessing even if the file is enormous. I understand that it was more easy while develpping to have multiple and sub folder file.

Thanks! The multiple files are useful because I can do lazy imports without conditional definitions ;)

PR #1809, opened 2023-07-10
Improve Documentation VI

zm711:
I realized I had some old fixes/clarifications that I forgot to submit a PR for. Mostly just tiny typo fixes. 

I also did a quick read through of the new motion docs and added some typo fixes. 

I did want to clarify one thing and then I'm more than happy to edit that doc again. "Drift" is an uncountable concept so you can mention correcting drift in general. Or you could mention correcting different "types of drift", but as far as I know pluralizing drift would not be standard English. That being said, if you are trying to create a new terminology around correcting types of drift (by saying drifts), then you could include the "drifts" neologism. I corrected drifts to drift in this PR, but if you want me to change it back just let me know. (I assume by drifts you're referring to things like step-drift vs fast drift vs slow drift and needing to potentially correct all of those differently).


doc/modules/motion_correction.rst
```diff
@@ -7,26 +7,26 @@ Motion/drift correction
...
 Long story short, the main idea is the same as the one used for non-rigid image registration, for example with calcium
 imaging. However, because with extracellular recording we do not have a proper image to use as a reference, the main idea
 of the algorithm is create an "image" via the activity profile of the cells during a given time window. Assuming this
 activity profile should be kept constant over time, the motion can be estimated, by blocks, along the probe's insertion axis
-(i.e. depth) so that we can interpolate the traces to compensate this estimated motion.
+(i.e. depth) so that we can interpolate the traces to compensate for this estimated motion.
 Users with a need to handle drift were currently forced to stick to the use of Kilosort2.5 or pyKilosort (see [Pachitariu2023]_). Recently, the Paninski
 group from Columbia University introduced a possibly more accurate method to estimate the drift (see [Varol2021]_
-and [Windolf2023]_), but this new method was not properly integrated in any sorter.
+and [Windolf2023]_), but this new method was not properly integrated into any sorter.
```
h-mayorquin, line 21: in vs into
Interesting.

doc/modules/motion_correction.rst
```diff
@@ -134,11 +134,11 @@ Low-level API
 -------------
 
 All steps (**activity profile**, **motion inference**, **motion interpolation**) can be launched with distinct functions.
-This can be useful to find the good method and finely tune/optimize parameters at every steps.
+This can be useful to find the best method and finely tune/optimize parameters at each step.
```
h-mayorquin, line 137: a good vs the best?

The plural is wrong but what about "The Police" song? Is there a a semantical distinction between each step and every step?

doc/modules/motion_correction.rst
```diff
@@ -134,11 +134,11 @@ Low-level API
...
 All functions are implemented in the :py:mod:`~spikeinterface.sortingcomponents` module.
 They all have a simple API with SpikeInterface objects or numpy arrays as inputs.
 Since motion correction is a hot topic, these functions have many possible methods and also many possible parameters.
-Finding the good combination of method/parameters is not that easy, but it should be doable, assuming the presets are not
+Finding the best combination of method/parameters is not that easy, but it should be doable, assuming the presets are not
```
h-mayorquin, line 141: Same here, "a good" vs "the best", don't know if the intent was the academic humble style.

h-mayorquin (approved):
Again, thanks a lot for doing this!

 I added some small comments but I trust your (native?) English and sense of style way more than mine.

doc/modules/motion_correction.rst
alejoe91: I think best reads better :)

doc/modules/motion_correction.rst
alejoe91: Same here, no need to be humble :P

alejoe91 (approved):

alejoe91:
Thanks @zm711, also for the english lessons :) :)

doc/modules/motion_correction.rst
samuelgarcia: yes we need to be humble!

samuelgarcia:
Thank you @zm711.
Hopefully you fixed the english grammar after Alessio already fixed after me! The first evrsion was even first.
This look perfect now.

doc/modules/motion_correction.rst
zm711: @ h-mayorquin Yeah super small, but into implies entry like motion. We integrated it into something. Whereas as in is a bit more stationary so. That feature was already in the sorter.

doc/modules/motion_correction.rst
zm711: In this case, you would want "best" because you're optimizing something. If you say "the good" it implies moral goodness like this is the religious combination. You could also say "right" combination instead if best feels too strong. But yeah best would be the standard for this type of situation.

doc/modules/motion_correction.rst
zm711: @h-mayorquin You're right about the Police the song. This is the same distinction as todos los dias/cada dia vs tous les jours/chaque jour (I don't actually know the Italian equivalent). It's more stylistic than required. In English at least each step emphasizes the singular step in the process whereas every step doesn't emphasize that. I like the emphasis better, but it is not wrong to say every step. (step would be need to singular either way).

h-mayorquin: Thanks for the explanation. You are completly right that in (my) Spanish I would say "Optimizar los parametros en cada paso" instead of 'Optimizar los parameteros en todos los pasos". Kudos.

PR #1813, opened 2023-07-11
Proposal to separate timestamp extraction and frames extraction logic in neo sorting extractors

h-mayorquin:
Cleaner diff than #1812.

While I was discussing with @samuelgarcia  on https://github.com/SpikeInterface/spikeinterface/pull/1626 he showed concern that for some neo sorting extractors we were transforming the timestamps to frames and then back to timestamps in some scenarios (the most relevant scenarion being neuroconv). 

I told him that my first intuition was to separate the ` get_unit_spike_train` function so there will be a channel directly from the segment to the option that fetches timestamps instead of frames. @samuelgarcia mentioned that he did not like the idea of patching and that @alejoe91 had exactly the same idea that I have.

This PR is a showcase of this idea. I still don't understand fully @samuelgarcia arguments against it and I think it will be useful to discuss this with a proper diff so we can see what it entails.

This change is not large:
1 ) I implemented a `get_unit_spike_times` method for neo segments.  
2) At the BaseSorting level if there is no `get_unit_spike_times` overiding then ` get_unit_spike_train`  reverts to the old logic.  If there is a `get_unit_spike_times` such as in neo, then it uses that channel to get the timestamps. Everything else is left the same: function signature, spike frames extraction, but now there is a specific way of sending timestamps directly from segments to sorting extractor.
3) This does not change the contract. The classes still need a `get_unit_spiketrain` and that remains the core of spikeinterface sorting extractors. It just adds a way for the segments to send timestamps directly when this is needed.

I think this avoids the problem of back and forth computation that worried Sam and also will make it simpler for us in neuroconv to detect timing errors if they come from neo sorting extractor as there is a specific and independent channel that we can isolate.

What are some downsides of this idea?

### DIFF
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -439,7 +439,7 @@ class BaseSortingSegment(BaseSegment):
             unit_id=unit_id,
             start_frame=start_frame,
             end_frame=end_frame,
-        )
+        ).astype("int64", copy=False)
 
         if parent_sorting.has_recording():
             times = parent_sorting.get_times(segment_index=segment_index)
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -109,18 +109,20 @@ class BaseSorting(BaseExtractor):
     ):
         segment_index = self._check_segment_index(segment_index)
         segment = self._sorting_segments[segment_index]
-        spike_frames = segment.get_unit_spike_train(
-            unit_id=unit_id, start_frame=start_frame, end_frame=end_frame
-        ).astype("int64")
         if return_times:
-            if self.has_recording():
-                times = self.get_times(segment_index=segment_index)
-                return times[spike_frames]
-            else:
-                t_start = segment._t_start if segment._t_start is not None else 0
-                spike_times = spike_frames / self.get_sampling_frequency()
-                return t_start + spike_times
+            spike_times = segment.get_unit_spike_times(
+                unit_id=unit_id,
+                start_frame=start_frame,
+                end_frame=end_frame,
+            )
+            return spike_times
         else:
+            spike_frames = segment.get_unit_spike_train(
+                unit_id=unit_id,
+                start_frame=start_frame,
+                end_frame=end_frame,
+            ).astype("int64")
+
             return spike_frames
 
     def register_recording(self, recording, check_spike_frames=True):
@@ -405,6 +407,7 @@ class BaseSortingSegment(BaseSegment):
         self._t_start = t_start
         BaseSegment.__init__(self)
 
+    # This should be an abstract method
     def get_unit_spike_train(
         self,
         unit_id,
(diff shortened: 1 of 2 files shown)


h-mayorquin:
> @h-mayorquin thansk for the draft. Note that in `BaseSorting` we already have the concept of returning spike trains in times: `sorting.get_unit_spike_train(unit_id, return_times=True)`.
> 
> Maybe overriding that function in the `NeoSortingExtractor` would be enough, no?
> 
> EDIT: sotty I read too fast. Makes sense to further split it in two functions for modularity ;)

Ah, OK,
Yes, also @samuelgarcia wants to keep logic of frames centralized as he fears fragmentation if I remember well. Which makes sense.

src/spikeinterface/extractors/neoextractors/neobaseextractor.py
```diff
@@ -613,6 +615,35 @@ def get_unit_spike_train(self, unit_id, start_frame, end_frame):
 
         return spike_frames
 
+    def get_unit_spike_times(self, unit_id, start_frame, end_frame):
+        """Useful for avoiding unecessary back and forth conversions between frames and times"""
+        spike_channel_index = self.map_from_unit_id_to_spike_channel_index(unit_id)
```
alejoe91, line 620:
```suggestion
        if self.has_recording() and self.recording.has_times():
            return super.get_unit_spike_times(...)
```

src/spikeinterface/core/basesorting.py
```diff
@@ -109,18 +109,20 @@ def get_unit_spike_train(
...
-        ).astype("int64")
         if return_times:
-            if self.has_recording():
-                times = self.get_times(segment_index=segment_index)
-                return times[spike_frames]
-            else:
-                t_start = segment._t_start if segment._t_start is not None else 0
-                spike_times = spike_frames / self.get_sampling_frequency()
-                return t_start + spike_times
+            spike_times = segment.get_unit_spike_times(
```
alejoe91, line 113: Make a `get_unit_spike_times(t_start, t_end, ...)` in `BaseSorting`, otherwise the conversion between frames and times could eb ambicuous (e.g. if there are gaps)

khl02007:
@h-mayorquin @alejoe91 just adding my use case with respect to NWB files: I'm trying to load a sorting saved in NWB with `read_nwb_sorting` but am getting neither the actual spike times in seconds stored in NWB file nor the correct frames that line up with the timestamps associated with the recording (we use explicit timestamps instead of inferring them via start time and sampling rate). I think there should be an option to get (1) the original spike times in seconds or (2) the spike train in units of frames in reference to a timestamps array (perhaps supplied externally or loaded from attached recording)

h-mayorquin:
@khl02007 Thanks for bringing this up. I will try to get back to work on this as soon as possible : )

h-mayorquin:
@khl02007 
Is your NWB file in dandi by any chance? It would be good to have is as a testing case. Or is it small enough that we can share or simle enough that I can genereate it with your script?

khl02007:
@h-mayorquin sure, it's not on DANDI but I can share a small NWB file. I will post a link here soon.

h-mayorquin:
Thanks, also post a short descripton of the expected output so if you x and y operations so I can be sure that I will cover your needs as I implement the solution.

PR #1820, opened 2023-07-12
Add option `relative_to=True`

DradeAW:
Add the option to set `relative_to` to `True`, in which case it will be relative to the folder it's in.
As discussed with @alejoe91

### DIFF
--- a/src/spikeinterface/core/tests/test_baserecording.py
+++ b/src/spikeinterface/core/tests/test_baserecording.py
@@ -2,6 +2,7 @@
 test for BaseRecording are done with BinaryRecordingExtractor.
 but check only for BaseRecording general methods.
 """
+import json
 import shutil
 from pathlib import Path
 import pytest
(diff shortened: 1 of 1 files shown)


DradeAW:
Do you think `relative_to=True` should be the default behaviour?

Absolute paths are a pain because if you change computer or even the folder it's in, everything breaks :'(

samuelgarcia:
Salut Aurelien,
I like this idea, we should have done before!
The only stuff I do not like so much which a typical python sweetmeat is that the same variable can be None or str or Path or bool!
To avoid this we need another variable this would be more explicit but more heavy let us think about it.

DradeAW:
> the same variable can be None or str or Path or bool!

Now with the possibility of `True`, we can change `None -> False` I think.
But you're right, we are still left with `str | Path | bool`. I don't personally mind it, but I understand that it's not completely optimal. Although having 2 arguments also has its downsides ...

alejoe91:
I would go with both options in the same argument!

h-mayorquin:
I agree with @samuelgarcia about the typing. 

I think that the docstring could be changed from the current
```            If True, file_path is serialized relative to the folder of file_path ```

To something like "file and folder paths in extractor objects kwarsg are changed to be relative to this path or the parent folder"

Finally, It would be really nice if we started adding testing to the core funtionality. At least for the core.

alejoe91 (approved):

DradeAW:
Do you want the default to be absolute paths (i.e. `relative_to=None`) or relative (i.e. `relative_to=True`)?

alejoe91:
Absolute

h-mayorquin:
The tests that were added just probe that no exception is raised. What I think we should test is if the different types work as intended. That is, if you pass a boolean and check that all the paths are written correctly.  This wil also serve the function of showing how the interface looks like.

samuelgarcia (approved):

PR #1824, opened 2023-07-13
Allow relative_to with recursive=False (in-place on objects)

### LINKED ISSUE #1819: Loading pipelines containing many nested sorting extractors are non-performant
Hi,

In my pipeline (Lussac), I have very complex sorting objects (made with like tens of `UnitsSelectionSorting`, tens of `UnitsAggregationSorting`, and multiple other curation functions.
Handling those objects is fine, fast, and memory-efficient.
Saving such objects to pickle format or JSON format is also fast and memory-efficient (although the json file is tens of thousands of lines long and can weigh more than 100 MB ^^).

The problem comes when loading.
If the file was saved with `recursive=False`, then the loading is fast and efficient.
But if the file was saved with `recursive=True`, then the loading can take minutes and more than 10 GB of RAM, whereas creating the object takes less than a second and almost no RAM.

In my case, I need to use `recursive=True` because I am using `relative_to` when saving, which can only propagate if loaded recursively.

alejoe91:
Might fix #1819 

For very long chains, dumping with `recursive=True` is a performance killer. However, that was required to make paths relative.
This PRs refactors completely the `recursive_path_finder` function to allow to modify the paths _in place_ on `BaseExtractor` objects. Note that this is only possible if `copy=False` and it raises an exception if `copy=True` 

See: https://github.com/alejoe91/spikeinterface/blob/non-recursive-relative-paths/src/spikeinterface/core/core_tools.py#L825-L896

- Also added a helper function `dict_contains_extractor` to check the "status" of a dictionary.
- Removed a bunch of `d = ` that do not play well with debugger..

@DradeAW can you test this?

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -936,8 +936,8 @@ class BaseExtractor:
 
 
 def _make_paths_relative(d, relative, copy=True) -> dict:
-    relative = str(Path(relative).absolute())
-    func = lambda p: os.path.relpath(str(p), start=relative)
+    relative = Path(relative).absolute()
+    func = lambda p: os.path.relpath(p, start=relative)
     return recursive_path_modifier(d, func, target="path", copy=copy)
 
 
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -555,7 +555,9 @@ class BaseExtractor:
         else:
             raise ValueError("Dump: file must .json or .pkl")
 
-    def dump_to_json(self, file_path: Union[str, Path, None] = None, relative_to=None, folder_metadata=None) -> None:
+    def dump_to_json(
+        self, file_path: Union[str, Path, None] = None, relative_to=None, folder_metadata=None, recursive=False
+    ) -> None:
         """
         Dump recording extractor to json file.
         The extractor can be re-loaded with load_extractor_from_json(json_file)
@@ -566,10 +568,18 @@ class BaseExtractor:
             Path of the json file
         relative_to: str, Path, or None
             If not None, file_paths are serialized relative to this path
+        folder_metadata: str, Path, or None
+            Folder with numpy files containing additional information (e.g. probe in BaseRecording) and properties.
+        recursive: bool
+            If True, all dicitionaries in the kwargs are expanded with `to_dict` as well, by default False.
         """
-        assert self.check_if_dumpable()
+        assert self.check_if_json_serializable()
         dump_dict = self.to_dict(
-            include_annotations=True, include_properties=False, relative_to=relative_to, folder_metadata=folder_metadata
+            include_annotations=True,
+            include_properties=False,
+            relative_to=relative_to,
+            folder_metadata=folder_metadata,
+            recursive=recursive,
         )
         file_path = self._get_file_path(file_path, [".json"])
 
@@ -577,6 +587,9 @@ class BaseExtractor:
             json.dumps(dump_dict, indent=4, cls=SIJsonEncoder),
             encoding="utf8",
         )
+        if relative_to and dict_contains_extractors(dump_dict):
+            # Make paths absolute again
+            dump_dict = _make_paths_absolute(dump_dict, relative_to, copy=False)
 
     def dump_to_pickle(
         self,
@@ -612,6 +625,10 @@ class BaseExtractor:
         file_path = self._get_file_path(file_path, [".pkl", ".pickle"])
 
         file_path.write_bytes(pickle.dumps(dump_dict))
+
+        if relative_to and dict_contains_extractors(dump_dict):
+            # Make paths absolute again
+            dump_dict = _make_paths_absolute(dump_dict, relative_to, copy=False)
 
     @staticmethod
     def load(file_path: Union[str, Path], base_folder=None) -> "BaseExtractor":
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -310,6 +310,7 @@ class BaseExtractor:
         relative_to: Union[str, Path, None] = None,
         folder_metadata=None,
         recursive: bool = False,
+        skip_recursive_path_modifier_warning: bool = False,
     ) -> dict:
         """
         Make a nested serialized dictionary out of the extractor. The dictionary produced can be used to re-initialize
@@ -329,6 +330,8 @@ class BaseExtractor:
             Folder with numpy `npy` files containing additional information (e.g. probe in BaseRecording) and properties.
         recursive: bool
             If True, all dicitionaries in the kwargs are expanded with `to_dict` as well, by default False.
+        skip_recursive_path_modifier_warning: bool
+            If True, skip the warning that is raised when `recursive=True` and `relative_to` is not None.
 
         Returns
         -------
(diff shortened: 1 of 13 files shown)


alejoe91:
An option to make it compatible with `copy=True`, would be to `clone` and replace extractor objects in the dictionary...but one thing at a time

src/spikeinterface/core/base.py
```diff
@@ -920,16 +935,16 @@ def save_to_zarr(
         return cached
 
 
-def _make_paths_relative(d, relative) -> dict:
-    relative = str(Path(relative).absolute())
-    func = lambda p: os.path.relpath(str(p), start=relative)
-    return recursive_path_modifier(d, func, target="path", copy=True)
+def _make_paths_relative(d, relative, copy=True) -> dict:
+    relative = Path(relative).absolute()
+    func = lambda p: os.path.relpath(p, start=relative)
```
alejoe91, line 940: `os.path` is compatible with `pathlib` objects

h-mayorquin: There is an equivalent pathlib function as well:
https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to

src/spikeinterface/core/core_tools.py
```diff
@@ -814,43 +851,49 @@ def recursive_path_modifier(d, func, target="path", copy=True) -> dict:
     dict
         Modified dictionary
     """
+    from .base import BaseExtractor
+
     if copy:
-        dc = deepcopy(d)
+        if dict_contains_extractors(dictionary):
+            raise Exception("The copy argument is only available if the input dictionary does not contain objects")
```
h-mayorquin, line 867:
```suggestion
            raise Exception("The copy argument is only available if the input dictionary does not contain extractorobjects")
```
?

src/spikeinterface/core/base.py
alejoe91: It doesn't work if paths are not direct relatives (i.e. it doesn't find .. locations)

DradeAW: I can attest to that: I had a similar problem a few months back, and the `pathlib` equivalent did not work, and I had to use `os.path.relpath`. (link to where I used it: https://github.com/BarbourLab/lussac/blob/9c71ed6ad1303460566bf46bacf22c25ee805fe3/src/lussac/utils/plotting.py#L26)

src/spikeinterface/core/base.py
```diff
@@ -329,6 +330,8 @@ def to_dict(
             Folder with numpy `npy` files containing additional information (e.g. probe in BaseRecording) and properties.
         recursive: bool
             If True, all dicitionaries in the kwargs are expanded with `to_dict` as well, by default False.
+        skip_recursive_path_modifier_warning: bool
+            If True, skip the warning that is raised when `recursive=True` and `relative_to` is not None.
```
samuelgarcia, line 334: In which situation we would have `recursive=True` and `relative_to` not None ?
I think this option is too much detail and to_dict() is more or less internal.
So I think that either we are strict and we raise an error either we do not care but adding this flag is quite a bit heavy for the API no ?

src/spikeinterface/core/base.py
```diff
@@ -376,11 +380,6 @@ def to_dict(
             "relative_paths": (relative_to is not None),
         }
 
-        try:
-            dump_dict["version"] = imported_module.__version__
-        except AttributeError:
-            dump_dict["version"] = "unknown"
-
```
samuelgarcia, line 383: Is it moved elsewhere ?

src/spikeinterface/core/base.py
```diff
@@ -394,9 +393,12 @@ def to_dict(
             dump_dict["properties"] = {k: self._properties.get(k, None) for k in self._main_properties}
 
         if relative_to is not None:
-            relative_to = Path(relative_to).absolute()
+            relative_to = Path(relative_to).resolve().absolute()
             assert relative_to.is_dir(), "'relative_to' must be an existing directory"
-            dump_dict = _make_paths_relative(dump_dict, relative_to)
+            copy = False if dict_contains_extractors(dump_dict) else True
```
samuelgarcia, line 398: I am not sure to follow the logic here. Are we making an inplace modification of extarctors._kwargs ?
If this is the case I think this is not a good idea.
rec.to_dict() should not modify rec itself. Maybe I am missing something.

src/spikeinterface/core/base.py
```diff
@@ -463,9 +466,9 @@ def clone(self) -> "BaseExtractor":
         """
         Clones an existing extractor into a new instance.
         """
-        d = self.to_dict(include_annotations=True, include_properties=True)
-        d = deepcopy(d)
-        clone = BaseExtractor.from_dict(d)
+        dictionary = self.to_dict(include_annotations=True, include_properties=True)
+        dictionary = deepcopy(dictionary)
+        clone = BaseExtractor.from_dict(dictionary)
```
samuelgarcia, line 471: I prefer "d"

src/spikeinterface/core/base.py
```diff
@@ -557,7 +560,9 @@ def dump(self, file_path: Union[str, Path], relative_to=None, folder_metadata=No
         else:
             raise ValueError("Dump: file must .json or .pkl")
 
-    def dump_to_json(self, file_path: Union[str, Path, None] = None, relative_to=None, folder_metadata=None) -> None:
+    def dump_to_json(
+        self, file_path: Union[str, Path, None] = None, relative_to=None, folder_metadata=None, recursive=False
+    ) -> None:
```
samuelgarcia, line 565: I do not understand. I think that json should be always reccursive  no ?
how could it be not reccurssive ?

src/spikeinterface/core/base.py
alejoe91: there were 2 of them, now it's at lines: https://github.com/alejoe91/spikeinterface/blob/non-recursive-relative-paths/src/spikeinterface/core/base.py#L370-L373

src/spikeinterface/core/base.py
alejoe91: we are. that is the only way `relative_to` can work without `recursive=True`

src/spikeinterface/core/base.py
alejoe91: `d` doesn't work with the debugger.

src/spikeinterface/core/base.py
alejoe91: as you said, `to_dict` is internal so we can keep it in my opinion

src/spikeinterface/core/core_tools.py
```diff
@@ -796,61 +832,84 @@ def recursive_path_modifier(d, func, target="path", copy=True) -> dict:
...
+        skip_targets = [skip_targets]
+
+    if dict_contains_extractors(dictionary):
+        if copy:
+            raise Exception("The copy argument is only available if the input dictionary does not contain objects")
+        else:
+            if not skip_warning:
+                warnings.warn(
+                    "The dictionary contains extractors, so the paths will be modified in-place!" "Use with caution"
+                )
```
samuelgarcia, line 872: I am not sure to be agree with inplace modification. I think we should raise error or force a copy always.

src/spikeinterface/core/base.py
h-mayorquin: Thanks.
This is good to know. It is on a footnote on the link the I shared actually:

```
[PurePath.relative_to()](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to) requires self to be the subpath of the argument, but [os.path.relpath()](https://docs.python.org/3/library/os.path.html#os.path.relpath) does not.
```

Double thanks, guys. I learn a couple of things about paths today. This is an enhancement that will be added in python 3:12, see the argument `walk_up` in the docs over there:

https://docs.python.org/3.12/library/pathlib.html#pathlib.PurePath.relative_to

The corresponding PR:
https://github.com/python/cpython/pull/19813

alejoe91:
closing this in favor of a simpler patch

PR #1830, opened 2023-07-17
Handle probe when load_sync_channel=True for SpikeGLX and OpenEphys

### LINKED ISSUE #1826: Loading sync channel using read_openephys
Hey guys, appreciate all of your terrific work!
I am using OE to record NPX and also save the sync channel - #385. I am trying to load the sync channel using read_openephys (in spikeinterface.extractors). When I use:
`new_recording = se.read_openephys(path_to_recording_file, stream_name="Record Node 101#Neuropix-PXI-100.ProbeA-AP", load_sync_channel=True)` 
I get the following error: "set_probe(inplace=True) must have all channel indices"

alejoe91:
Fixes #1826

Unifies the behavior of `load_sync_channel` between Open Ephys and SpikeGLX:

In case `load_sync_channel=True`, a probe group with an additional dummy 1-channel probe for the sync channel is set. @samuelgarcia note that previously spikeglx was just skipping probe loading.

@neuronzoo could you test this on the Open Ephys dataset? It would also be great if you could record and share with us a very short (<0.5s) dataset with sync channel on so we can use it for testing!

### DIFF
--- a/src/spikeinterface/core/baserecordingsnippets.py
+++ b/src/spikeinterface/core/baserecordingsnippets.py
@@ -147,7 +147,7 @@ class BaseRecordingSnippets(BaseExtractor):
         if number_of_device_channel_indices >= self.get_num_channels():
             error_msg = (
                 f"The given Probe have 'device_channel_indices' that do not match channel count \n"
-                f"{number_of_device_channel_indices} vs {self.get_num_channels()} \n"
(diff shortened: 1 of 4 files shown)


alejoe91:
After discussion, we decided that since the recording with the SYNC channel is not sortable, it's better not to load the probe in that case (same as SpikeGLX). The user will need to load the recording without the sync channel to get the probe

PR #1831, opened 2023-07-17
Cleanup dumping/to_dict

alejoe91:
This PR simplifies a bit the dumping and cleans up some to_dict mechanisms.

In brief:

 - the `dump_to_json` is always `recursive` because that allows to modify the paths in the dict. The `SIJsonEncoder` basically does the same but at dump time
 - the `dump_to_pickle` is always not recursive and it doesn't allow `relative_to` to be set (because it's incompatible with the pickle loading and it would require in-place object modifications)

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -338,6 +338,9 @@ class BaseExtractor:
 
         kwargs = self._kwargs
 
+        if relative_to and not recursive:
+            raise Exception("`relative_to` is only posible when `recursive=True`")
+
         if recursive:
             to_dict_kwargs = dict(
                 include_annotations=include_annotations,
@@ -560,7 +563,7 @@ class BaseExtractor:
     def dump_to_json(self, file_path: Union[str, Path, None] = None, relative_to=None, folder_metadata=None) -> None:
         """
         Dump recording extractor to json file.
-        The extractor can be re-loaded with load_extractor_from_json(json_file)
+        The extractor can be re-loaded with load_extractor(json_file)
 
         Parameters
         ----------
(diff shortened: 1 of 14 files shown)


src/spikeinterface/core/base.py
```diff
@@ -338,6 +338,9 @@ def to_dict(
 
         kwargs = self._kwargs
 
+        if relative_to and not recursive:
+            raise Exception("`relative_to` is only posible when `recursive=True`")
```
h-mayorquin, line 342: Yeah, I agree with @DradeAW that it should be a ValueError.

src/spikeinterface/core/base.py
```diff
@@ -561,23 +564,28 @@ def dump(self, file_path: Union[str, Path], relative_to=None, folder_metadata=No
...
-            This means that file and folder paths in extractor objects kwargs are changed to be relative rather than absolute.
+        relative_to: str, Path, or None
+            If not None, file_paths are serialized relative to this path
+        folder_metadata: str, Path, or None
+            Folder with files containing additional information (e.g. probe in BaseRecording) and properties.
         """
-        assert self.check_if_dumpable()
+        assert self.check_if_json_serializable(), "The extractor is not json serializable"
 
         if relative_to is True:
```
h-mayorquin, line 580:
```suggestion
        # Writing paths as relative_to requires recursively expanding the dict
        if relative_to:
           recursive = True
           # We use relative_to == True to encode using the parent_folder
           relative_to = Path(file_path).parent if relative_to is True else relative_to
```

What do you think?

src/spikeinterface/core/base.py
```diff
@@ -561,23 +564,28 @@ def dump(self, file_path: Union[str, Path], relative_to=None, folder_metadata=No
...
+        The extractor can be re-loaded with load_extractor(json_file)
 
         Parameters
         ----------
         file_path: str
             Path of the json file
-        relative_to: str, Path, True or None
-            If not None, files and folders are serialized relative to this path. If True, the relative folder is the parent folder.
-            This means that file and folder paths in extractor objects kwargs are changed to be relative rather than absolute.
+        relative_to: str, Path, or None
```
h-mayorquin, line 573: Why did you modify the docstring to the old typing?

h-mayorquin (approved):
This looks good to me. Minor questions about documentation.

src/spikeinterface/core/base.py
alejoe91: mistake in merging, fixing back

samuelgarcia (approved):

PR #1836, opened 2023-07-17
Fixed `numba.jit` and `binary num_chan` warnings

DradeAW:
The `numba` warnings came from the fact that `nopython` default behaviour will change in the future with numba 0.59.
Setting `nopython=True` solves this problem.

Also changed a couple `num_chan` to `num_channels` since this is the new name.


alejoe91 (approved):

DradeAW:
And here I thought my implementation of checking for `tridesclous` was more elegant ^^'

samuelgarcia:
Merci Aurelien, c'est cool.

PR #1838, opened 2023-07-17
Fix hdbascan installation in read the docs

alejoe91:
Try to fix docs run

### DIFF
--- a/docs_rtd.yml
+++ b/docs_rtd.yml
@@ -5,5 +5,6 @@ dependencies:
   - python=3.10
   - pip
   - datalad
+  - hdbscan
   - pip:
     - -e .[docs]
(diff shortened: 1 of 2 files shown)


h-mayorquin:
Sorry, I misread you in the other thread, you are proposing a higher bound, I think the problem is the lack of a lower bound. Another option is to put an upper bound (ceiling) on hdbscan.

But probably the easier is just to instead send the hdbscan isntall to the conda file and the figure out this later.

alejoe91:
or installing hdbscan with conda ;)

h-mayorquin:
Probably the canary in the mine of the problems that came after.

alejoe91:
> Probably the canary in the mine of the problems that came after.

Definitely! The release rush made us oversee this!

PR #1835, opened 2023-07-17
Fix neo release bug in Mearec

h-mayorquin:
I just remembered about this. Tests were failing on neuroconv because of this.

This eliminates a bug that should disappear with the next release in neo.

PR #1848, opened 2023-07-19
Fix Mearec handling of new arguments before neo release 0.13

h-mayorquin:
This is a mistake that I made out of ignorance.

I did this PR here:
https://github.com/SpikeInterface/spikeinterface/pull/1835

Which assumes that the changes here would be available in neo version 0.12. At time I did not know that neo version 0.12 was already out as I did not see the tags in github (it is only in pipy). 

But the changes here are not in neo version 0.12 but will come online in the yet to be 0.13:
https://github.com/NeuralEnsemble/python-neo/pull/1258

This PR fixes this here in dev but **I am afraid that the current release mearec is broken.**  The current release is fixed to neo >= 0.12:

https://github.com/SpikeInterface/spikeinterface/commit/c43295d6d6e8a63d124dbfafb17b1937ce4a168e

But this means that lines that I am modifying here will fail (it will assume that a keyword argument that exists does not).

I am wondering why this did not break the release tests. Maybe there is something wrong with them?

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/mearec.py
+++ b/src/spikeinterface/extractors/neoextractors/mearec.py
@@ -8,7 +8,7 @@ import probeinterface as pi
 from .neobaseextractor import NeoBaseRecordingExtractor, NeoBaseSortingExtractor
 
 
-def drop_neo_arguments_before_version_0_13_0(neo_kwargs):
+def drop_invalid_neo_arguments_before_version_0_13_0(neo_kwargs):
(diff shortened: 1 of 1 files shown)


alejoe91:
We can merge this and make another micro patch

h-mayorquin:
@alejoe91, that would be great. That is what semantic versioning is for.

alejoe91 (approved):

PR #1851, opened 2023-07-19
Prepare release 0.98.2

alejoe91:


pyproject.toml
```diff
@@ -138,8 +138,8 @@ test = [
 
     # for github test : probeinterface and neo from master
     # for release we need pypi, so this need to be commented
-    "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",
```
h-mayorquin, line 141: I have a question. Why don't we have these lines  in `dependencies`?

Here:
https://github.com/SpikeInterface/spikeinterface/blob/28166989089739610e328a1d30ce17588279a52d/pyproject.toml#L22-L29

alejoe91: because we only want to install from sources in tests, but for package depemndencies we just want to give bounds on released versions. I think we have to keep this "test" distinction

h-mayorquin: Got you, if that's the case, if this only for the CI, I think these linkes keep track of this automatically for us:

https://github.com/SpikeInterface/spikeinterface/blob/2aba94fe9cc544c467bf4790517205ec4f9e6b8a/.github/actions/build-test-environment/action.yml#L27-L34

But maybe you prefer to make this explicit?

h-mayorquin:
Failing because of this:
https://github.com/SpikeInterface/spikeinterface/pull/1863

I think this should solved it, at least it does in my local machine after installing python-neo 0.12

pyproject.toml
alejoe91: That's true! They should not be needed anymore now. 

Do you suggest to remove them now? 

Maybe let's release first (they are commented out anyways) and then we push this upgrade.

Note that we would still need them in the `docs` section, but if we want to unify this and avoid needing to comment/uncomment things we could move them to the `docs_rtd.yml` pip section ;)

h-mayorquin: I think it makes sense to wait after the release.

alejoe91: @samuelgarcia is mentioning that this is also good for an end user installing from source!

doc/releases/0.98.2.rst
```diff
@@ -0,0 +1,18 @@
+.. _release0.98.2:
+
+SpikeInterface 0.98.2 release notes
+-----------------------------------
+
+19th July 2023
```
alejoe91, line 6:
```suggestion
20th July 2023
```

PR #1854, opened 2023-07-19
Add full tests to publish action.

h-mayorquin:
As described here, doing from main branch to test this.

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -12,24 +12,36 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - name: Set up Python 3.9
+      - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.9'
-      - name: Install dependencies for testing
+          python-version: '3.10'
+      - name: Install packages
         run: |
+          sudo apt install git
+          git config --global user.email "CI@example.com"
+          git config --global user.name "CI Almighty"
           python -m pip install -U pip  # Official recommended way
-          pip install pytest
-          pip install zarr
-          pip install setuptools wheel twine build
-          pip install -e .[test_core]
-      - name: Test core with pytest
+          pip install -e .[test,extractors,full]
+      - name: Install git-annex
         run: |
-          pytest -v src/spikeinterface/core
-      - name: Publish on PyPI
-        env:
-          TWINE_USERNAME: __token__
-          TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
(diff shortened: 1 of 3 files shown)


h-mayorquin:
OK, the tets are faling when the version iends in `.dev` as it should.

https://github.com/SpikeInterface/spikeinterface/actions/runs/5599159488/jobs/10239760746?pr=1854

Let's see if the tests are working as they should.

.github/workflows/publish-to-pypi.yml
```diff
@@ -4,32 +4,46 @@ on:
...
+      #     is_dev_version=$(python -c "import importlib.metadata; package_name = 'spikeinterface'; version = importlib.metadata.version(package_name); print(version.endswith('dev0'))")
+      #     if [ $is_dev_version = "True" ]; then
+      #       echo "Dev version detected,remember to fix the version before the release!"
+      #       exit 1
+      #     fi
+      - name: Test everything but external sorters before release
         run: |
-          python -m build --sdist --wheel
-          twine upload dist/*
+          pytest -m "not sorters_external"
```
h-mayorquin, line 42: I am keping this as in the full tests with code coverage but I don't know if we should test EVERYTHING here.

h-mayorquin:
OK, so the full tests are running reasonably:
https://github.com/SpikeInterface/spikeinterface/actions/runs/5599505029/jobs/10240551443?pr=1854

I think this is ready to go. The only thing left is that if it feals when you push the new tag you should fix the tests and then do a manual run of this action.

.github/workflows/publish-to-pypi.yml
alejoe91: fair to me

alejoe91:
I'm ok with this! Let's wait to hear @samuelgarcia thoughts.

Another thing we should do is to run all tests if the `pyproject.toml` is changed (basically consider it as a core change)

samuelgarcia:
I am not sure to be happy with this.
the idea was:
* we prepare a branch to prepare the release. In this branch there not "dev" and no neo+probeinterface githib install 
   because we comment
* this branch should trigger normally all test, this is not the case unfortunatly.
* Then push to pypi is quite fast and make only on core test on one sytsem (just in case).

But for me the entire test are triggered and check in the prepare relase PR not before pushing to pypi.
Otherwise we have to wait too long.
Lets have a call soon baout this.

h-mayorquin:
This is better done with a combination of :
https://github.com/SpikeInterface/spikeinterface/pull/1856
And:
https://github.com/SpikeInterface/spikeinterface/pull/1858

Closing this.

PR #1856, opened 2023-07-19
Treat pyproject-toml changed as core change

alejoe91:
Related to #1854

@h-mayorquin another (easier) option is to keep the publish to pypi as is and treat a change in `pyproject.toml` as a core change.

The `prepare-release` branch, with the updated version, will therefore trigger full tests before release, so we can correct any errors prior to tagging.

I think this is a better option than #1854 (and @samuelgarcia too). What do you think?

### DIFF
--- a/.github/workflows/full-test.yml
+++ b/.github/workflows/full-test.yml
@@ -66,6 +66,10 @@ jobs:
         id: modules-changed
         run: |
           for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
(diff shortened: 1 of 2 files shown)


samuelgarcia (approved):

h-mayorquin:
I think this is a good idea. One thing that you should be mindful about is that the full tests use the cache mechanism. This means that even if you change the `pyproject.toml` to the fix dependency of python-neo (instead of github), the full tests will still run in neo dev and some errors like the ones in the next micro release will not be detected.

PR #1862, opened 2023-07-20
For the cronjob, avoid re-downloading the cache dataset if present

h-mayorquin:
There is a new option in the cache machinery that allows for only checking if the cache is available and not downloading it:

https://github.com/actions/cache#inputs

This will save some unecessary computation and bandwith use and as an important side effect make @samuelgarcia happier.

This is used in the daily cron job to checks if it is necessary to create new caches for both the gin data and pip dependencies.

### DIFF
--- a/.github/workflows/caches_cron_job.yml
+++ b/.github/workflows/caches_cron_job.yml
@@ -65,6 +65,7 @@ jobs:
         with:
           path: ~/spikeinterface_datasets
           key: ${{ runner.os }}-datasets-${{ steps.repo_hash.outputs.dataset_hash }}
+          lookup-only: 'true'
       - name: Cache found?
         run: echo "Cache-hit == ${{steps.cache-datasets.outputs.cache-hit == 'true'}}"
       - name: Installing datalad and git-annex
@@ -89,7 +90,7 @@ jobs:
         run: |
           cd $HOME
           pwd
-          du -hs spikeinterface_datasets
+          du -hs spikeinterface_datasets  # Should show the size of ephy_testing_data
           cd spikeinterface_datasets
           pwd
           ls -lh  # Should show ephy_testing_data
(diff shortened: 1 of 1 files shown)


alejoe91:
Thanks @h-mayorquin! If I understand correctly, this additional argument will skip recreating a cache for datasets if that's already present. Right?

h-mayorquin:
@alejoe91 yes, you are correct, the job of the cron job here is to download the datasets if they are not available. That is, in the cronjobs the dataset is not actually used so we don't need to download it. This new option they recently added implements that. Saving behavior is not affected so if the data is not available it will be created and uploaded as it should.

.github/workflows/caches_cron_job.yml
```diff
@@ -33,6 +33,7 @@ jobs:
         with:
           path: ${{ github.workspace }}/test_env
           key: ${{ runner.os }}-venv-${{ steps.dependencies.outputs.hash }}-${{ steps.date.outputs.date }}
+          lookup-only: 'true'
```
h-mayorquin, line 36:
```suggestion
          lookup-only: 'true'   # Avoids downloading the data, saving behavior is not affected.
```

.github/workflows/caches_cron_job.yml
```diff
@@ -64,6 +65,7 @@ jobs:
         with:
           path: ~/spikeinterface_datasets
           key: ${{ runner.os }}-datasets-${{ steps.repo_hash.outputs.dataset_hash }}
+          lookup-only: 'true'
```
h-mayorquin, line 68:
```suggestion
          lookup-only: 'true'   # Avoids downloading the data, saving behavior is not affected.
```

alejoe91:
Let's merge after the release, since it will only affect cron jobs ;)

h-mayorquin:
No rush here.

alejoe91 (approved):

PR #1834, opened 2023-07-17
Make all paths resolved and absolute

alejoe91:
Unify handling of all paths in kwargs to be `Path(...).absolute()` and add `resolve()` in base

PR #1863, opened 2023-07-20
fixes to neuroscope

h-mayorquin:


alejoe91:
@h-mayorquin since some NEO readers are not updated to handle strings, we decided to always pass strings as neo_kwargs.

See: #1834

Any downside of this? (except for the addiotional casting to `Path` on the NEO side if needed)

alejoe91:
Seeing now that it fixes an issue, sorry!

alejoe91 (approved):

PR #1865, opened 2023-07-20
Harmonize radius_um

yger:
This PR changes the name local_radius_um into radius_um everywhere, in order to be consistent. At the moment, the two names are co-existing, posing some confusions and/or errors in the code (for example when localizing units or spikes)


samuelgarcia:
Super merci!

@alejoe91 
I think this is good stuff this should not break too much stuff because it is mainly sorting component and internal stuff.

alejoe91 (approved):

samuelgarcia (approved):

PR #1872, opened 2023-07-22
Fix missing import pandas in comparison

alejoe91:


h-mayorquin (approved):
Probably did not surface when I refactored because it does not have a test.

alejoe91:
I know! Added a tiny one that just tests the output type ;)

h-mayorquin:
Great!

PR #1875, opened 2023-07-24
Fix NWB streaming: do not convert to Path if ros3 or fsspec!

alejoe91:


h-mayorquin (changes requested):
OK, this makes sense:

```
file_path = (
    "https://dandi-api-staging-dandisets.s3.amazonaws.com/blobs/5f4/b7a/5f4b7a1f-7b95-4ad8-9579-4df6025371cc"
)

Path(file_path).absolute()
```
Ouputs:

``` 
PosixPath('/home/heberto/development/spikeinterface/bin/https:/dandi-api-staging-dandisets.s3.amazonaws.com/blobs/5f4/b7a/5f4b7a1f-7b95-4ad8-9579-4df6025371cc')
```

Which is clearly wrong. That said, I changed this a while ago and the test ran well a couple of times in between. I am hoping that this was ALL the problem but we will see.

Anyway, this change should lbe done.  There is only one request:

One important thing is that we have the same logic in the sorting extractor:

https://github.com/SpikeInterface/spikeinterface/blob/af440033af92e6ac805c2c28d9c1d04a8f5c01bf/src/spikeinterface/extractors/nwbextractors.py#L470-L486

Can you change that there?

We also don't have tests for the sorting extractor with streaming but we can add that in another PR.

h-mayorquin (approved):
Thanks for adding the test. Looks good to me now that the streaming test is passing.

PR #1879, opened 2023-07-26
Addition of a peak centerer to align peaks correctly

yger:
Creation of a PeakCenterer object to be used in localize() and compute_spike_locations(). In fact, after a sorting, using the channel of the extrememum of the template as a reference channel hides the noise due to peak detection, and the fact that peaks might be sometimes off-centered. The PeakCenterer compensates for that. Using a given radius, for each peak, it will look for the channel where the extremum is obtained to accuretly reflect what happened during the peak detection


samuelgarcia:
#1944 and #1950 should replace this, we can close

Issue #1881, opened 2023-07-26
IndexError when writing TracePadded recording to binary


TomBugnon:
Hi @samuelgarcia ,
I think something is wrong somewhere with TracePaddedRecordings handling of start/end frames, or in the interface with chunking

I get an error when running kilosort. The following code should be able to replicate:

```

from spikeinterface.full import NumpyRecording, concatenate_recordings, write_binary_recording
from spikeinterface.preprocessing.zero_channel_pad import TracePaddedRecording

base_traces = np.zeros((1000, 5), dtype="float64")
base_traces[:] = np.arange(1000)[:, None]
sampling_frequency = 1000
rec0 = NumpyRecording([base_traces] * 3, sampling_frequency)
rec1 = NumpyRecording([base_traces] * 2, sampling_frequency)


# concatenate
rec = concatenate_recordings([rec0, rec1])

padded_rec = TracePaddedRecording(rec, 0, 10)

write_binary_recording(padded_rec, "./padded_rec", dtype="int16")
```

> File ~/projects/ecephys_dev/spikeinterface/src/spikeinterface/core/core_tools.py:314, in write_binary_recording(recording, file_paths, dtype, add_file_extension, byte_offset, auto_cast_uint, **job_kwargs)
>     310 init_args = (recording, file_path_dict, dtype, byte_offset, cast_unsigned)
>     311 executor = ChunkRecordingExecutor(
>     312     recording, func, init_func, init_args, job_name="write_binary_recording", **job_kwargs
>     313 )
> --> 314 executor.run()
> 
> File ~/projects/ecephys_dev/spikeinterface/src/spikeinterface/core/job_tools.py:376, in ChunkRecordingExecutor.run(self)
>     374 worker_ctx = self.init_func(*self.init_args)
>     375 for segment_index, frame_start, frame_stop in all_chunks:
> --> 376     res = self.func(segment_index, frame_start, frame_stop, worker_ctx)
>     377     if self.handle_returns:
>     378         returns.append(res)
> 
> File ~/projects/ecephys_dev/spikeinterface/src/spikeinterface/core/core_tools.py:233, in _write_binary_chunk(segment_index, start_frame, end_frame, worker_ctx)
>     231 array = np.ndarray.__new__(np.ndarray, shape=shape, dtype=dtype, buffer=memmap_obj, order="C", offset=array_offset)
>     232 # apply function
> --> 233 traces = recording.get_traces(
[...]

PR #1883, opened 2023-07-26
Handle edge frames in concatenated rec

TomBugnon:
Fixes https://github.com/SpikeInterface/spikeinterface/issues/1881

### DIFF
--- a/src/spikeinterface/core/segmentutils.py
+++ b/src/spikeinterface/core/segmentutils.py
@@ -169,6 +169,11 @@ class ProxyConcatenateRecordingSegment(BaseRecordingSegment):
         if end_frame is None:
             end_frame = self.get_num_samples()
 
+        # # Ensures that we won't request invalid segment indices
+        if (start_frame >= self.get_num_samples()) or (end_frame <= start_frame):
+            # Return (0 * num_channels) array of correct dtype
+            return self.parent_segments[0].get_traces(0, 0, channel_indices)
+
         i0 = np.searchsorted(self.cumsum_length, start_frame, side="right") - 1
         i1 = np.searchsorted(self.cumsum_length, end_frame, side="right") - 1
 


h-mayorquin (approved):
I think this is fine as a patch.  Thanks for spotting this, @TomBugnon.

 If the maintenace push ever comes to be, we should have a generalized test for boundary conditions, array-like behavior and channel_indices behavior for recording, pre-processing and segment utils. Something like what we have for the neo extractors.

src/spikeinterface/core/segmentutils.py
```diff
@@ -169,6 +169,11 @@ def get_traces(self, start_frame, end_frame, channel_indices):
         if end_frame is None:
             end_frame = self.get_num_samples()
 
+        # # Ensures that we won't request invalid segment indices
+        if (start_frame >= self.get_num_samples()) or (end_frame <= start_frame):
+            # Return (0 * num_channels) array of correct dtype
+            return self.parent_segments[0].get_traces(0, 0, channel_indices)
+
```
samuelgarcia, line 176: Why not raising an error ? start_frame should nver be >= num_samples.
Here we are hidding another and deeper bug.
Could we revert this make a test that trigger the bug and then make a concensus fix ?

TomBugnon: 1: Without this fix, when user calls `get_traces` with out-of-range start/end frames they get an error, which is inconsistent with the other extractors' behavior (which all return an empty array)

2: valid  TracePadRecording.get_traces calls will under the hood call get_traces with start_frame==end_frame and 
raise this error as int the issue I filed

samuelgarcia: I am not sure to be OK this kind of hidden patch for other extractor.
start_frame >= self.get_num_samples() should not be allowed in my mind.

TomBugnon: Fair enough but this case is not tested for in any of the other extractors currently (including Neo based I believe)

PR #1884, opened 2023-07-26
Implement split/merge/clean cluster before template matching.

samuelgarcia:
This is a port:
  * tridesclous auto merge au to split rewritten
  * mixed with a free interpretation of [similar code from culumbia team](https://github.com/cwindolf/dartsort/blob/main/spike_psvae/pre_deconv_merge_split.py)

@cwindolf : I would be happy to discuss this with you one day.

* [X] **split** :  prallel multi method split. Working
* [x] **merge**: done but needimprovement
* [ ] **clean**: 
* [X] **tools** : manipulate sparse waveforms or pca buffer to aogne then on demand 

This will need #1799 before I continue.


samuelgarcia:
Close by #1996

PR #1887, opened 2023-07-26
Add missing tic in ks*_master when skipping prepro

TomBugnon:
Avoids `error using toc` in datashift2 

image


alejoe91:
Shouldn't datashift2 be skipped in this case? Anyways, we're definitely missing this [`tic`](https://github.com/MouseLand/Kilosort/blob/main/preProcess/preprocessDataSub.m#L16)

alejoe91 (approved):

TomBugnon:
I guess some people might want to use SI preprocessing and KS motion correction

alejoe91:
Of course ;)

PR #1895, opened 2023-07-28
Use spike_vector in `count_num_spikes_per_unit`

DradeAW:
`count_num_spikes_per_unit` can be very slow for big sorting objects if it needs to re-build all spike trains from a spike vector (in my sorting object, 16min whereas the new implementation takes 5s)

### DIFF
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -291,7 +291,7 @@ class BaseSorting(BaseExtractor):
             unit_indices, counts = np.unique(spike_vector["unit_index"], return_counts=True)
             for unit_index, unit_id in enumerate(self.unit_ids):
                 if unit_index in unit_indices:
-                    idx = np.argmax(unit_indecex == unit_index)
+                    idx = np.argmax(unit_indices == unit_index)
                     num_spikes[unit_id] = counts[idx]
                 else:  # This unit has no spikes, hence it's not in the counts array.
                     num_spikes[unit_id] = 0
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/basesorting.py
```diff
@@ -278,12 +278,24 @@ def count_num_spikes_per_unit(self):
...
+                for segment_index in range(self.get_num_segments()):
+                    st = self.get_unit_spike_train(unit_id=unit_id, segment_index=segment_index)
+                    n += st.size
+                num_spikes[unit_id] = n
+        else:
+            spike_vector = self.to_spike_vector()
+            unit_indices, counts = np.unique(spike_vector["unit_index"], return_counts=True)
+            for unit_index, unit_id in enumerate(self.unit_ids):
+                if unit_index in unit_indices:
+                    idx = np.argmax(unit_indecex == unit_index)
```
samuelgarcia, line 294: typo unit_indecex > unit_indeces

And also
unit_indices = list(unit_indices)
idx = unit_indices.index(unit_index)
  is maybe more readable no ?

DradeAW:
Even with the typo the tests pass, which is a bit worrying …
I think that casting to a list and using 2 lines is a bit ugly, and that the variable names are pretty self-explanatory. That said if you prefer your version, I’m fine with it

h-mayorquin:
> Even with the typo the tests pass, which is a bit worrying …
Maybe there are no tests for it?

samuelgarcia (approved):

PR #1903, opened 2023-08-03
Restore npzfolder.py file to load previously saved sorting objects

alejoe91:
#1696 removed the `core/npzfolder.py` file. This creates an issue when loading sorting objects saved prior to that PR, because the `load_extractor` looks for the `core.npzfolder.NpzFolderSorting` class. 

Added a link between the old and the new location.


src/spikeinterface/core/npzfolder.py
```diff
@@ -0,0 +1,7 @@
+"""
+This file is for backwards compatibility with the old npz folder structure.
+"""
+
+from .sortingfolder import NpzFolderSorting as NewNpzFolderSorting
+
+NpzFolderSorting = NewNpzFolderSorting
```
samuelgarcia, line 7: Why not 
```suggestion
from .sortingfolder import NpzFolderSorting
```

alejoe91: That should work!

samuelgarcia (approved):

PR #1908, opened 2023-08-03
Change the signature on kilosort's delete intermediate files parameters.

### LINKED ISSUE #1896: Cleanup of kilosort temporary files causing issues with Phy
I was planning to raise an issue on the cleanup of intermediate files in Kilosort but I see this has already been done in #1709 which is great. I would suggest some supplementary additions that allow a little more refinement over the selection of which intermediate files to delete. Currently `temp_wh.dat` is deleted by default and (at least in my hands) this causes Phy to not be able to run as it requires `temp_wh.dat`.

The three cases of intermediate files created during Kilosort-ing with spikeinterface:
1) `recording.dat` - binary recording copied into 'sorting_output' by spikeinterface (useful if concatenating across segments)
2) `temp_wh.dat` - recording that has been preprocessed (e.g. drift corrected) in Kilosort. Despite the confusing name this is required by Phy for visualisation (by default) unless the corresponding path in `params.py` is changed. But as discussed [here](https://github.com/MouseLand/Kilosort/issues/311), it makes sense to visualise based on the preprocessed data.
3) intermediate matlab files

I would suggest changing the arguments and defaults `delete_tmp_files=True` and `delete_recording_dat=False` to a tuple argument
`delete_intermediate_files: Tuple[Literal["recording.dat", "temp_wh.dat", "matlab_files"]] = ("recording.dat", "matlab_files")`. A benefit of this is it allows full flexibility without adding a third argument (for the matlab files). I would suggest keeping `temp_wh.dat` by default because it is required by Phy and instead remove `recording.dat`. But maybe there is a good use case for keeping the `recording.dat` by default too and / or this signature is confusing.

JoeZiminski:
This PR extends cleanup after kilosort sorting, closes #1896.

The signature for deleting intermediate files is changed slightly and is now a `Tuple` with optional entries, `recording.dat`, `temp_wh.dat` and `matlab_files`. The reason for this extension is to give a little more granularity over the selection of what to delete, and to not delete `temp_wh.dat` by default as it is used for Phy.

For now, have default to delete only `matlab_files` because my understanding is that `recording.dat` is not deleted for other sorters and so this might be confusing if the behaviour is different across sorters. 

Do you think it is worth pushing up this functionality to the `run_sorter` level, so that for any sorter `recording.dat` can be deleted (in the cases it is created). The issue with that approach would be that some intermerdiate files are sorter-specific. Alternatively, each sorter has it's own `delete_intermediate_files` parameter that takes keywords specific to it's output, and these can be accessed with `run_sorter`'s kwargs? (in which case, happy to extend this functionality to other sorters, I am just not very familiar with the intermediate files they all create).

### DIFF
--- a/src/spikeinterface/sorters/external/kilosort2.py
+++ b/src/spikeinterface/sorters/external/kilosort2.py
@@ -50,7 +50,8 @@ class Kilosort2Sorter(KilosortBase, BaseSorter):
         "skip_kilosort_preprocessing": False,
         "scaleproc": None,
         "save_rez_to_mat": False,
-        "delete_intermediate_files": ("matlab_files",),
+        "delete_tmp_files": ("matlab_files",),
+        "delete_recording_dat": False,
     }
 
     _params_description = {
@@ -72,8 +73,10 @@ class Kilosort2Sorter(KilosortBase, BaseSorter):
         "skip_kilosort_preprocessing": "Can optionaly skip the internal kilosort preprocessing",
         "scaleproc": "int16 scaling of whitened data, if None set to 200.",
         "save_rez_to_mat": "Save the full rez internal struc to mat file",
-        "delete_intermediate_files": "Delete intermediate files created during sorting. Tuple indicating the "
-        "files to delete. Options are: ('recording.dat', 'temp_wh.dat', 'matlab_files')",
+        "delete_tmp_files": "Delete temporary files created during sorting (matlab files and the `temp_wh.dat` file that"
+        "contains kilosort-preprocessed data). Accepts `False` (deletes no files), `True` (deltes all files)"
+        "or a Tuple containing the files to delete. Options are: ('temp_wh.dat', 'matlab_files')",
+        "delete_recording_dat": "Whether to delete the 'recording.dat' file after a " "successful run",
     }
 
     sorter_description = """Kilosort2 is a GPU-accelerated and efficient template-matching spike sorter. On top of its
--- a/src/spikeinterface/sorters/external/kilosort2.py
+++ b/src/spikeinterface/sorters/external/kilosort2.py
@@ -73,7 +73,7 @@ class Kilosort2Sorter(KilosortBase, BaseSorter):
         "scaleproc": "int16 scaling of whitened data, if None set to 200.",
         "save_rez_to_mat": "Save the full rez internal struc to mat file",
         "delete_intermediate_files": "Delete intermediate files created during sorting. Tuple indicating the "
-                                     "files to delete. Options are: ('recording.dat', 'temp_wh.dat', 'matlab_files')"
+        "files to delete. Options are: ('recording.dat', 'temp_wh.dat', 'matlab_files')",
     }
 
     sorter_description = """Kilosort2 is a GPU-accelerated and efficient template-matching spike sorter. On top of its
--- a/src/spikeinterface/sorters/external/kilosort2.py
+++ b/src/spikeinterface/sorters/external/kilosort2.py
@@ -50,8 +50,7 @@ class Kilosort2Sorter(KilosortBase, BaseSorter):
         "skip_kilosort_preprocessing": False,
         "scaleproc": None,
         "save_rez_to_mat": False,
-        "delete_tmp_files": True,
-        "delete_recording_dat": False,
+        "delete_intermediate_files": ("matlab_files",),
     }
 
     _params_description = {
@@ -73,8 +72,8 @@ class Kilosort2Sorter(KilosortBase, BaseSorter):
         "skip_kilosort_preprocessing": "Can optionaly skip the internal kilosort preprocessing",
         "scaleproc": "int16 scaling of whitened data, if None set to 200.",
         "save_rez_to_mat": "Save the full rez internal struc to mat file",
-        "delete_tmp_files": "Whether to delete all temporary files after a successful run",
-        "delete_recording_dat": "Whether to delete the 'recording.dat' file after a successful run",
+        "delete_intermediate_files": "Delete intermediate files created during sorting. Tuple indicating the "
+                                     "files to delete. Options are: ('recording.dat', 'temp_wh.dat', 'matlab_files')"
     }
 
     sorter_description = """Kilosort2 is a GPU-accelerated and efficient template-matching spike sorter. On top of its
--- a/src/spikeinterface/sorters/external/kilosort2.py
+++ b/src/spikeinterface/sorters/external/kilosort2.py
@@ -74,7 +74,7 @@ class Kilosort2Sorter(KilosortBase, BaseSorter):
         "scaleproc": "int16 scaling of whitened data, if None set to 200.",
         "save_rez_to_mat": "Save the full rez internal struc to mat file",
         "delete_tmp_files": "Delete temporary files created during sorting (matlab files and the `temp_wh.dat` file that "
-        "contains kilosort-preprocessed data). Accepts `False` (deletes no files), `True` (deltes all files) "
+        "contains kilosort-preprocessed data). Accepts `False` (deletes no files), `True` (deletes all files) "
         "or a Tuple containing the files to delete. Options are: ('temp_wh.dat', 'matlab_files')",
         "delete_recording_dat": "Whether to delete the 'recording.dat' file after a successful run",
     }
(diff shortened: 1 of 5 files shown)


samuelgarcia:
Hi Joe.
I am not against this but I have the feeling that the the cost of breaking the compatibility is a bit high (and we need to rebuild all the docker images!!).
@alejoe91 what do you think ?

alejoe91:
We can rebuild the docker images, but it would be great if we can make this bakward-compatible :)

JoeZiminski:
That's a good point about backward compatibility. It would also be nice to move the `delete_recording_dat` to the `run_sorter` level, as I guess this applies to every possible sorter, while `delete_tmp_files` is specific to Kilosort? 

If this is correct, I could rework this PR to:

1) Add `delete_recording_dat` option to `run_sorter`. Revert the kilosort `delete_recording_dat` signature to how it was previously. It would be somewhat vestigial but maintained for backwards compatibility

2) Add `delete_tmp_files`  argument back and allow it to accept `True`, `False`, or a `Tuple` of keywords to delete (as in the current PR). This way it will be backwards compatible but also allow more refined deletions. On further consideration I'm not sure if the `temp_wh.dat` (despite it's confusing name) should really be regarded as a temporary file as it is used for Phy / should be used for waveform extraction. But, if the default is `False`, then I don't think this is much of a problem, and the use of `temp_dat.dat` for Phy could be included on the docstring so people are aware before setting to `True`.

alejoe91:
Thanks @JoeZiminski 

I think your proposed solution is perfect. Note that passing a tuple for `delete_tmp_files` will not be compatible with running on containers until a new version of SI is released. 

@samuelgarcia note that the docker images wouldn't need to be rebuilt, since we're only touching the Python wrappers.

JoeZiminski:
Thanks @alejoe91 @samuelgarcia I have made the discussed changes, let me know if anything further is required. Are there are quick sorter-tests where testing these options could be slotted in?

src/spikeinterface/sorters/external/kilosort2.py
```diff
@@ -73,7 +73,9 @@ class Kilosort2Sorter(KilosortBase, BaseSorter):
         "skip_kilosort_preprocessing": "Can optionaly skip the internal kilosort preprocessing",
         "scaleproc": "int16 scaling of whitened data, if None set to 200.",
         "save_rez_to_mat": "Save the full rez internal struc to mat file",
-        "delete_tmp_files": "Whether to delete all temporary files after a successful run",
+        "delete_tmp_files": "Delete temporary files created during sorting (matlab files and the `temp_wh.dat` file that "
+        "contains kilosort-preprocessed data). Accepts `False` (deletes no files), `True` (deltes all files) "
```
alejoe91, line 77:
```suggestion
        "contains kilosort-preprocessed data). Accepts `False` (deletes no files), `True` (deletes all files) "
```

src/spikeinterface/sorters/external/kilosortbase.py
```diff
@@ -215,16 +215,23 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
...
-            for temp_file in sorter_output_folder.glob("*.m"):
-                temp_file.unlink()
-            for temp_file in sorter_output_folder.glob("*.mat"):
-                temp_file.unlink()
-            if (sorter_output_folder / "temp_wh.dat").exists():
-                (sorter_output_folder / "temp_wh.dat").unlink()
         if params["delete_recording_dat"] and (recording_file := sorter_output_folder / "recording.dat").exists():
             recording_file.unlink()
 
+        if params["delete_tmp_files"]:
```
alejoe91, line 221:
```suggestion
        if isinstance(params["delete_tmp_files"], bool):
            if params["delete_tmp_files"]:
                tmp_to_remove = ("matlab_files", "temp_wh.dat")
            else:
                tmp_to_remove = ()
        else:
            assert isinstance(params["delete_tmp_files"], (tuple, list)), "..."
       
       if "temp_wh.dat" in tmp_to_remove: ... 
```
@JoeZiminski I think it would be easier to read if we test for boolean first!

zm711:
@JoeZiminski 

Just as an FYI and to transfer our convo over from #1896. I was just helping someone on the Phy issue tracker (running KS2 + Phy outside of SI). And they tried both the raw file `*.dat` and the `temp_wh.dat` and they said that it looked much better with the raw data rather the whitened data for KS2. I have confirmed this myself, but I wanted a record of at least one person voting on using raw for KS2.

alejoe91:
> @JoeZiminski
> 
> Just as an FYI and to transfer our convo over from #1896. I was just helping someone on the Phy issue tracker (running KS2 + Phy outside of SI). And they tried both the raw file `*.dat` and the `temp_wh.dat` and they said that it looked much better with the raw data rather the whitened data for KS2. I **haven't** confirmed this myself, but I wanted a record of at least one person voting on using raw for KS2.
> 
> edited: bold to make clear I haven't tested

That's a good point. While spatial whitening is needed by Kilosort, it kind of messes up the spatial structure of the waveforms, making them quite less interpretable. I also prefer looking at "real" signals :)

JoeZiminski:
Thanks @zm711 and @alejoe91 for bringing up this very important point. I was operating under the misapprehension that somehow regions of data including spikes were not whitened (in fact, it is just that these regions are not used in estimation of the covariance matrix for whitening in KS).

I wonder if the changes between KS2 and KS2.5, KS3 are related to [this issue](https://github.com/MouseLand/Kilosort/issues/504) and [this discussion](https://github.com/cortex-lab/phy/issues/1040). I can confirm the full whitening matrix is saved for KS2 but simply a scaling diagonal matrix for more recent versions. In phy, templates / waveforms are un-whitened on the fly for display (as far as I can tell, although exactly when this is on / off I am not entirely sure). Also the issues and [this discussion](https://github.com/cortex-lab/KiloSort/issues/35) above indicates that KS unwhitens the spike for computing the final template versions (as the matlab results variable `rez.Wraw` contain the unwhitened templates).

The primary reason for #1954 was to ensure quality metrics are computed on data that is correct with respect to the kilosort output (e.g. spike times,  channels on which spikes are loading. However, if the case of the whitened data, it does not make sense to use this for quality metrics if waveform data must be un-whitened before interpretation. In this case, I can see that may makes more sense to use the spikeinterface pre-processed data for quality metrics calculation (assuming the user has applied bandpass filtering in spikeinterface and not kilosort). 

However, if drift correction is applied, then I have two concerns, 1) is main one relevant for this discussion:

**1)** That the waveforms extracted from the spikeinterface pre-processed data but drift-correction is applied in kilosort, then template calculation will be incorrect if the spikeinterface-preprocessed data is used. This is because all waveforms for a unit are averaged across channels, but if the spikes drifting across different channels, then the average will be incorrect. Maybe I have made an oversight though in how the templates are computed on the SI side.
2) That applying the whitening matrix calculated prior to drift correction will be incorrect after drift correction because the cross-channel covariance will be different after the kriging interpolation applied during channel shifting. This is more a discussion for another day.

This means we have a bit of a problem, if I understand correctly. The `temp_wh.dat` should be used for quality metric when drift-correction is used in Kilosort (which, I think, is true in the majority of cases). However, using this file natively may lead to sub-optimal results because everything is white. Thus, it may be required for this extractor to un-whiten the data prior to waveform extraction (I hope, the inv. whitening matrix can be found somewhere in KS2.5 and KS3).

JoeZiminski:
Thanks @zm711 and @alejoe91! I have moved my last comment and this discussion to #1954.

I added a final assert check please let me know if anything can be improved, and tested manually. Are there any parameter tests on KS sorting a test for this could be slotted into?

src/spikeinterface/sorters/external/kilosort2_5.py
```diff
@@ -83,7 +83,9 @@ class Kilosort2_5Sorter(KilosortBase, BaseSorter):
         "skip_kilosort_preprocessing": "Can optionaly skip the internal kilosort preprocessing",
         "scaleproc": "int16 scaling of whitened data, if None set to 200.",
         "save_rez_to_mat": "Save the full rez internal struc to mat file",
-        "delete_tmp_files": "Whether to delete all temporary files after a successful run",
+        "delete_tmp_files": "Delete temporary files created during sorting (matlab files and the `temp_wh.dat` file that "
+        "contains kilosort-preprocessed data). Accepts `False` (deletes no files), `True` (deltes all files) "
```
zm711, line 87:
```suggestion
        "contains kilosort-preprocessed data). Accepts `False` (deletes no files), `True` (deletes all files) "
```

zm711 (commented):
Just a couple typo fixes in the docstrings for me @JoeZiminski,

but I'll let @alejoe91 and @samuelgarcia let you know about any tests they want added.

JoeZiminski:
thanks @zm711 !

alejoe91 (approved):

alejoe91:
Looking great @JoeZiminski 

Merging :)

Issue #1906, opened 2023-08-03
Issue saving agreement sorting object to disk


ghost:
I want to run extract_waveforms on an AgreementSortingExtractor object. According to issue #972 I need to save the agreement sorting object to disk first. However, this returns an error saying the object is not JSON serializable.

Here's my full code:
```

mcmp = sc.compare_multiple_sorters(
    sorting_list=[klusta,mountainsort5,pykilosort],
    name_list=['klusta','mountainsort5','pykilosort'],
    verbose=True)
agr = mcmp.get_agreement_sorting(minimum_agreement_count=2)
agr = agr.save(folder='agreement_sorting')
```

And here's the full error message:

```
Traceback (most recent call last):
  File "/tmp/ipykernel_2498083/3956134603.py", line 3, in <module>
    agr2 = agr.save(folder='agreement_sorting')
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lmb/home/rbedford/.local/mambaforge/envs/spikeinterface/lib/python3.11/site-packages/spikeinterface/core/base.py", line 749, in save
    loaded_extractor = self.save_to_folder(**kwargs)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lmb/home/rbedford/.local/mambaforge/envs/spikeinterface/lib/python3.11/site-packages/spikeinterface/core/base.py", line 818, in save_to_folder
    self.dump(provenance_file)
  File "/lmb/home/rbedford/.local/mambaforge/envs/spikeinterface/lib/python3.11/site-packages/spikeinterface/core/base.py", line 558, in dump
    self.dump_to_json(file_path, relative_to=relative_to, folder_metadata=folder_metadata)
  File "/lmb/home/rbedford/.local/mambaforge/envs/spikeinterface/lib/python3.11/site-packages/spikeinterface/core/base.py", line 596, in dump_to_json
    json.dumps(dump_dict, indent=4, cls=SIJsonEncoder),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lmb/home/rbedford/.local/mambaforge/envs/spikeinterface/lib/python3.11/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
          ^^^^^^^^^^^
[...]

PR #1909, opened 2023-08-04
Fixed bug that prevents dumping AgreementSortingExtractor

rbedfordwork:
This bug prevents extracting waveforms from an AgreementSortingExtractor as this process requires dumping the AgreementSortingExtractor to disk first.

The issue was that when initialising the AgreementSortingExtractor, BaseSorting.init() is run after self.is_json_serializable = False, which in turn calls BaseExtractor.__init_() which overrides to self._is_json_serializable = True.

The fix is simply to put self.is_json_serializable = False after BaseSorting.init().

Fixes issue #1906


alejoe91:
Nice catch!!! Thanks!

alejoe91 (approved):

PR #1921, opened 2023-08-15
Convert from samples<->times directly on BaseRecordings

cwindolf:
This implements `sample_index_to_time` and reverse on `BaseRecording`, which is something I noticed I wanted the other day. Are you guys open to this change? If so, I can add a test -- `BaseRecordingSegment.sample_index_to_time` didn't have tests anywhere so wasn't sure where to put it, maybe in `test_time_handling.py`?

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -1,18 +1,18 @@
-from typing import Iterable, List, Union
-from pathlib import Path
 import warnings
+from pathlib import Path
+from typing import Iterable, List, Union
+from warnings import warn
 
 import numpy as np
-
-from probeinterface import Probe, ProbeGroup, write_probeinterface, read_probeinterface, select_axes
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Salut Charlie.
I am OK with this. This is strange that it was not there already because recordingsegment is really a high level public API stuff.

src/spikeinterface/core/baserecording.py
```diff
@@ -1,18 +1,22 @@
-from typing import Iterable, List, Union
-from pathlib import Path
 import warnings
+from pathlib import Path
+from typing import Iterable, List, Union
+from warnings import warn
```
samuelgarcia, line 4: Thank you for the re ordering. This is a huge contribution.

cwindolf: 😆 sorry forgot to turn isort off

alejoe91 (approved):

PR #1936, opened 2023-08-25
Fix Doc Display Issue in `getting_started`

zm711:
Using html for the getting started qc display table didn't respect the RTD boundaries. I just rewrote this as a parsed-literal instead, which should respect the boundaries better (ie will add a slider if it is too big rather than cross the rtd borders). I also think it'll be easier to edit in the future if it is not written in html. 

https://spikeinterface.readthedocs.io/en/latest/how_to/get_started.html


samuelgarcia:
Hi Zach. Thanks a lot for all you tedious fix in the doc.

alejoe91 (approved):

PR #1940, opened 2023-08-28
Amplitude scaling - handle collisions

alejoe91:
@samuelgarcia 

I followed your suggestion and simplified the logic for finding overlapping spikes and representing the collisions.

Here's a gist showcasing the usage and the difference with no collision handling:
https://gist.github.com/alejoe91/167544c8ceebc0de7bfc80302a7b51a0

### DIFF
--- a/src/spikeinterface/postprocessing/amplitude_scalings.py
+++ b/src/spikeinterface/postprocessing/amplitude_scalings.py
@@ -21,6 +21,7 @@ class AmplitudeScalingsCalculator(BaseWaveformExtractorExtension):
         self.spikes = self.waveform_extractor.sorting.to_spike_vector(
             extremum_channel_inds=extremum_channel_inds, use_cache=False
         )
+        self.overlapping_mask = None
 
     def _set_params(
         self,
@@ -132,8 +133,30 @@ class AmplitudeScalingsCalculator(BaseWaveformExtractorExtension):
             **job_kwargs,
         )
         out = processor.run()
-        (amp_scalings,) = zip(*out)
+        (amp_scalings, overlapping_mask) = zip(*out)
         amp_scalings = np.concatenate(amp_scalings)
+        if handle_collisions > 0:
+            from ..core.job_tools import divide_recording_into_chunks
+
+            overlapping_mask_corrected = []
+            all_chunks = divide_recording_into_chunks(processor.recording, processor.chunk_size)
+            num_spikes_so_far = 0
+            for i, overlapping in enumerate(overlapping_mask):
+                if i == 0:
+                    continue
+                segment_index = all_chunks[i - 1][0]
+                spikes_in_segment = self.spikes[segment_slices[segment_index]]
+                i0 = np.searchsorted(spikes_in_segment["sample_index"], all_chunks[i - 1][1])
+                i1 = np.searchsorted(spikes_in_segment["sample_index"], all_chunks[i - 1][2])
+                num_spikes_so_far += i1 - i0
+                overlapping_corrected = overlapping.copy()
+                overlapping_corrected[overlapping_corrected >= 0] += num_spikes_so_far
+                overlapping_mask_corrected.append(overlapping_corrected)
+            overlapping_mask = np.concatenate(overlapping_mask_corrected)
+            print(f"Found {len(overlapping_mask)} overlapping spikes")
+            self.overlapping_mask = overlapping_mask
+        else:
+            overlapping_mask = np.concatenate(overlapping_mask)
 
         self._extension_data[f"amplitude_scalings"] = amp_scalings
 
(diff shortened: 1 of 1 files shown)


src/spikeinterface/postprocessing/amplitude_scalings.py
```diff
@@ -291,10 +365,325 @@ def _amplitude_scalings_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+        # deal with borders
+        if sample_centered - cut_out_before < 0:
+            full_template[: sample_centered + cut_out_after] = template_cut[cut_out_before - sample_centered :]
+        elif sample_centered + cut_out_after > end_frame + right:
+            full_template[sample_centered - cut_out_before :] = template_cut[: -cut_out_after - (end_frame + right)]
+        else:
+            full_template[sample_centered - cut_out_before : sample_centered + cut_out_after] = template_cut
+        X[:, i] = full_template.T.flatten()
+
+    reg = LinearRegression().fit(X, y)
```
samuelgarcia, line 566: Should we have `fit_intercept=False` to be sure to have no intercept when fitting.
And maybe `positive=True`also.

src/spikeinterface/postprocessing/amplitude_scalings.py
```diff
@@ -291,10 +365,325 @@ def _amplitude_scalings_chunk(segment_index, start_frame, end_frame, worker_ctx)
             else:
                 local_waveform = traces_with_margin[cut_out_start:cut_out_end, sparse_indices]
             assert template.shape == local_waveform.shape
-            local_waveforms.append(local_waveform)
-            templates.append(template)
+
             linregress_res = linregress(template.flatten(), local_waveform.flatten())
-            scalings.append(linregress_res[0])
```
samuelgarcia, line 297: why having scipy.linregress and sklearn.linear_model.LnearRegression for colliding spikes ?
Does scipy can make the same ?

src/spikeinterface/postprocessing/amplitude_scalings.py
```diff
@@ -291,10 +365,325 @@ def _amplitude_scalings_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+                    nbefore,
+                    all_templates,
+                    unit_inds_to_channel_indices,
+                    cut_out_before,
+                    cut_out_after,
+                )
+                # the scaling for the current spike is at index 0
+                scalings[spike_index] = scaled_amps[0]
+
+                # make collision_dict indices "absolute" by adding i0 and the cumulative number of spikes in previous segments
```
samuelgarcia, line 394: Not sure that `collision_dict` is the correct naming because `collisions` is the same but with local index no ?

src/spikeinterface/postprocessing/amplitude_scalings.py
```diff
@@ -291,10 +365,325 @@ def _amplitude_scalings_chunk(segment_index, start_frame, end_frame, worker_ctx)
...
+        The maximum number of samples between two spikes to consider them as overlapping
+    unit_inds_to_channel_indices: dict
+        A dictionary mapping unit indices to channel indices
+
+    Returns
+    -------
+    collision_spikes_dict: np.array
+        A dictionary with collisions. The key is the index of the spike with collision, the value is an
+        array of overlapping spikes, including the spike itself at position 0.
+    """
```
samuelgarcia, line 448: I am OK with this but the algortithm is a bit bru-force no ?
we are looping on all spikes and we are doing 2 searchsorted.

Algos like this https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/postprocessing/correlograms.py#L237
or this done @DradeAW 
https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/postprocessing/correlograms.py#L294
avoid the full double loop over spikes.

Algos to find collision is really close to computing correlogram

We can keep this for now but I would add a TODO for later: improve the loop.

samuelgarcia:
@alejoe91 : OK for me.
I did some annoying comments but this is OK to not fix them now btu lets put a clear TODO if we want to improve.

src/spikeinterface/postprocessing/amplitude_scalings.py
alejoe91: yes, but we can unify that

src/spikeinterface/postprocessing/amplitude_scalings.py
alejoe91: yes. I'll call it local VS global

src/spikeinterface/postprocessing/amplitude_scalings.py
alejoe91: ok, later for improvement.

src/spikeinterface/postprocessing/amplitude_scalings.py
alejoe91: So scipy is the fastest and it gives the exact same result (to the 6th decimal) of `LinearRegression`. 

image

The `_prms` version here is with `fit_intercept=False, positive=True`. Using `positive=True` is the same as without, but slightly faster (255 ms ± 1.64 ms VS 286 ms ± 3 ms).
In my opinion it's correct to also fit the intercept for a better fit, even to be more robust against some fluctuiations.

What should we do?

samuelgarcia (approved):

PR #1943, opened 2023-08-29
Sparse waveforms for Spyking Circus 2

yger:
PR to provide the support for sparse waveforms in circus 2, and factorized matching engine. The internal machinery now uses the sparsification mecanisms provided by SI, and thus we should be able to benchmark properly that among sorters


src/spikeinterface/sorters/internal/spyking_circus2.py
```diff
@@ -3,7 +3,7 @@
 import os
 import shutil
 import numpy as np
-import os
+import psutil
```
alejoe91, line 6: is this used anywhere? It needs to be added to the requirements full in any case

yger: Nope, this has been removed

src/spikeinterface/sortingcomponents/clustering/random_projections.py
```diff
@@ -160,86 +159,58 @@ def main_function(cls, recording, peaks, params):
...
+
+        if params["shared_memory"]:
+            waveform_folder = None
+            mode = "memory"
+        else:
+            waveform_folder = tmp_folder / "waveforms"
+            mode = "folder"
+
+        sorting_folder = tmp_folder / "sorting"
+        sorting = NumpySorting.from_times_labels(spikes["sample_index"], spikes["unit_index"], fs)
```
samuelgarcia, line 180: This is faster and should be the default when we already have a spike vector.
Note the you need to given also unit_ids
```suggestion
        sorting = NumpySorting(spikes, fs)
```

alejoe91: @yger this is the only change that is required! Can you add the `unit_ids`?

yger: Done. However, I have a question for the new waveform speedup, with single or multi buffers. @samuelgarcia @alejoe91 shouldn't this functionnality be exposed at the extract_waveforms() level?

samuelgarcia: Yes, this is the plan but not now.
And WaveformEtractor will be refactor a lot anyway not even sure this will be done before the big refactoring.

yger:
@samuelgarcia I've checked the behaviour, and this is working fine, so I think you can proceed with merging this PR

samuelgarcia (approved):

PR #1950, opened 2023-09-01
Spike location with true channel

samuelgarcia:
@alejoe91 @yger : there are some question in the code for you

@yger: can you test this for the localization benchmark ?


src/spikeinterface/postprocessing/spike_locations.py
```diff
@@ -44,13 +50,28 @@ def _run(self, **job_kwargs):
...
 
         we = self.waveform_extractor
 
-        spike_locations = localize_peaks(we.recording, self.spikes, **self._params, **job_kwargs)
+        extremum_channel_inds = get_template_extremum_channel(we, peak_sign="neg", outputs="index")
+
+        params = self._params.copy()
+        channel_from_template = params.pop("channel_from_template")
+
+        # @alessio @pierre: where do we expose the parameters of radius for the retriever (this is not the same as the one for locatization it is smaller) ???
```
yger, line 64: I would expose it in the compute_spike_location function, otherwise this has no sense. In addition, to ensure comparison with/witout this extra mecanism, we should make sure that if radius_um is set to 0, a classical PeakRetriever is used here

src/spikeinterface/postprocessing/spike_locations.py
```diff
@@ -25,9 +26,14 @@ def __init__(self, waveform_extractor):
...
-    def _set_params(self, ms_before=0.5, ms_after=0.5, method="center_of_mass", method_kwargs={}):
-        params = dict(ms_before=ms_before, ms_after=ms_after, method=method)
+    def _set_params(
+        self, ms_before=0.5, ms_after=0.5, channel_from_template=True, method="center_of_mass", method_kwargs={}
+    ):
+        params = dict(
+            ms_before=ms_before, ms_after=ms_after, channel_from_template=channel_from_template, method=method
+        )
         params.update(**method_kwargs)
+        print(params)
```
yger, line 36: remove the print here

src/spikeinterface/core/node_pipeline.py
```diff
@@ -0,0 +1,698 @@
...
+
+    def __init__(
+        self, recording, sorting, channel_from_template=True, extremum_channel_inds=None, radius_um=50, peak_sign="neg"
+    ):
+        PipelineNode.__init__(self, recording, return_output=False)
+
+        self.channel_from_template = channel_from_template
+
+        assert extremum_channel_inds is not None, "SpikeRetriever need the dict extremum_channel_inds"
+
```
yger, line 176: Why the SpikeRetriever is not internally, automatically, getting the extremum_channel_inds? This seems like a useless argument no? You have use cases where users would like to provide their own extremum channels?

src/spikeinterface/core/node_pipeline.py
```diff
@@ -0,0 +1,698 @@
...
+            channel_distance = get_channel_distances(recording)
+            self.neighbours_mask = channel_distance < radius_um
+            self.peak_sign = peak_sign
+
+        # precompute segment slice
+        self.segment_slices = []
+        for segment_index in range(recording.get_num_segments()):
+            i0 = np.searchsorted(self.peaks["segment_index"], segment_index)
+            i1 = np.searchsorted(self.peaks["segment_index"], segment_index + 1)
+            self.segment_slices.append(slice(i0, i1))
```
yger, line 189: I think everywhere in SI where you are doing this, you should do
i0, i1 = np.searchsorted(self.peaks["segment_index"], [segment_index, segment_index + 1])

Should be faster

src/spikeinterface/postprocessing/spike_locations.py
samuelgarcia: PeakRetriever will be not used here this is postprocessing.

src/spikeinterface/postprocessing/spike_locations.py
```diff
@@ -95,12 +116,16 @@ def get_extension_function():
 
 WaveformExtractor.register_extension(SpikeLocationsCalculator)
 
+# @alessio @pierre: channel_from_template=True is the old behavior but this is not accurate
+# what do we put by default ?
+
```
yger, line 121: I would go for the new behavior as a default, but we need to think on the impact with the metrics

src/spikeinterface/core/node_pipeline.py
samuelgarcia: no the instanciation has to be fast because it duplicate on several process this is external when building the node graph

src/spikeinterface/core/node_pipeline.py
samuelgarcia: Yes but this is absolutly everywhere.
A external PR for the entire code would be cool

src/spikeinterface/postprocessing/spike_locations.py
```diff
@@ -44,13 +50,28 @@ def _run(self, **job_kwargs):
...
+        channel_from_template = params.pop("channel_from_template")
+
+        # @alessio @pierre: where do we expose the parameters of radius for the retriever (this is not the same as the one for locatization it is smaller) ???
+        spike_retriever = SpikeRetriever(
+            we.recording,
+            we.sorting,
+            channel_from_template=channel_from_template,
+            extremum_channel_inds=extremum_channel_inds,
+            radius_um=50,
+            peak_sign=self._params.get("peaks_sign", "neg"),
```
yger, line 71: This neg should not be hardcoded, and peak_sign should be an argument of compute_spike_locations

src/spikeinterface/postprocessing/spike_locations.py
```diff
@@ -95,12 +116,16 @@ def get_extension_function():
...
 
 def compute_spike_locations(
     waveform_extractor,
     load_if_exists=False,
     ms_before=0.5,
     ms_after=0.5,
+    channel_from_template=True,
     method="center_of_mass",
     method_kwargs={},
     outputs="concatenated",
```
yger, line 131: I think we need to add peak_sign here, and I would also add radius_um

yger:
We need to provide a value for peaks['amplitudes'], otherwise some localizaiotn methods are not working. I propose to add
local_peaks["amplitude"][i] = traces[peak["sample_index"], local_peaks[i]["channel_index"]] L 220 in spike_locations.py

yger:
@samuelgarcia  can we proceed with this PR such that I can wrap up our next paper ?

samuelgarcia:
@yger : I made some update on this. Can you test this ?

yger (approved):

yger:
@samuelgarcia some tests are still failing, waiting for merging

alejoe91 (changes requested):
fix tests

src/spikeinterface/sortingcomponents/tools.py
```diff
@@ -19,6 +19,9 @@ def make_multi_method_doc(methods, ident="    "):
 
 
 def get_prototype_spike(recording, peaks, job_kwargs, nb_peaks=1000, ms_before=0.5, ms_after=0.5):
+    # TODO for Pierre: this function is really unefficient because it runa full pipeline only for a few
+    # spikes, which leans that traces are entirally computed!!!!!
+    # Please find a better way
```
alejoe91, line 24:
```suggestion
    # TODO for Pierre: this function is really inefficient because it runs a full pipeline only for a few
    # spikes, which means that all traces need to be accesses! Please find a better way
```

src/spikeinterface/postprocessing/spike_locations.py
```diff
@@ -119,6 +147,17 @@ def compute_spike_locations(
         The left window, before a peak, in milliseconds.
     ms_after : float
         The right window, after a peak, in milliseconds.
+    spike_retriver_kwargs: dict
+        A dict that contains the behavior for getting the maximum channel for each spike.
+        This contain dict contains:
```
alejoe91, line 152:
```suggestion
        A dictionary to control the behavior for getting the maximum channel for each spike.
        This dictionary contains:
```

alejoe91 (approved):

PR #1954, opened 2023-09-01
Introduce extractor for Kilosort's `temp_wh.dat`

JoeZiminski:
This PR is currently in-progress. The main is to load Kilosort's `temp_wh.dat` file for Kilosort 1-3. 

For kilosort 2.5 and 3, Phy uses the `temp_wh.dat` file, which (as far as I can tell) always contains all available channels.

For kilosort 1 and 2, the `temp_wh.dat` file is not used for Phy (although waveforms, etc. have previoulsy been calculated from it prior to saving in `.npy` files as it is the kilosort-preprocessed data). Kilosort 1 and 2 will remove channels for inactivity prior to sorting, and saves the used channels in `channel_map.npy`.

The `temp_wh.dat` is zero-padded at the end to match size of N * NT. The attached powerpoint indicates the method kilosort 2.5 and 3 uses for batching during preprocessing. The method is slightly different for 1, 2 but as far as I can tell results in the same output.

The main thing the extractor needs to do is handle the missing channels. The recording also contains zero-padding, but I'm hoping these can just be left there and will not effect waveforms as all spiketimes should be outside of the zero-padded region.

[kilosort_preprocessing_batching.pptx](https://github.com/SpikeInterface/spikeinterface/files/12501589/kilosort_preprocessing_batching.pptx)

### DIFF
--- a/src/spikeinterface/extractors/KilosortTempWhExtractor.py
+++ b/src/spikeinterface/extractors/KilosortTempWhExtractor.py
@@ -8,6 +8,7 @@ import importlib.util
 import runpy
 import numpy as np
 from spikeinterface import WaveformExtractor, extract_waveforms
+from spikeinterface.preprocessing import whiten
 
 
 class KilosortTempWhExtractor(BinaryRecordingExtractor):
@@ -155,88 +156,56 @@ class KilosortTempWhExtractor(BinaryRecordingExtractor):
 
 #         if channel_map.ndim == 2:  # kilosort > 2
 # does kilosort > 2 store shanks differently?             channel_indices = channel_map.ravel()
-
-path_ = Path(r"X:\neuroinformatics\scratch\jziminski\ephys\code\sorter_output")  # sorter_output
-recording_new = KilosortTempWhExtractor(path_)
-
 from spikeinterface import extractors
 from spikeinterface import postprocessing
 
-sorting = extractors.read_kilosort(
-    folder_path=(path_ / "sorter_output").as_posix(),
-    keep_good_only=False,
+path_ = Path(r"X:\neuroinformatics\scratch\jziminski\ephys\code\sorter_output")  # sorter_output
+recording = KilosortTempWhExtractor(path_ / "sorter_output")  # TODO: without this extra is failing
+
+original_recording = load_extractor(path_ / "spikeinterface_recording.json", base_folder=path_)
+
+original_recording = whiten(original_recording, dtype=np.int16, mode="local", int_scale=200)
+
+# Okay this works as a test. Steps are to:
+# 1) run sorting (script below)
+# 2) load the input recording and temp_wh with data above
+# 3) whiten the input recording.
+# 4) correlate loaded temp_wh and pp recording - these should be highly correlated if temp_wh was loaded correctly
+# 5) check all properties and scaling.
+
+# Currently the issues discussed in #1908 are a blocker. Once a decision on this is made
+# this PR can be continued with.
+
+x = recording.get_traces(start_frame=0, end_frame=10000, return_scaled=False)  # TODO: figure scaling
+y = original_recording.get_traces(start_frame=0, end_frame=10000, return_scaled=False)
+
+"""
+Generate Test Data
+
+# Load and Preprocess
+data_path = Path(r"/ceph/neuroinformatics/neuroinformatics/scratch/jziminski/ephys/code/test_run")
+output_path = Path(r"/ceph/neuroinformatics/neuroinformatics/scratch/jziminski/ephys/code/sorter_output")
+
+recording = se.read_spikeglx(data_path)
+
+recording = phase_shift(recording)
+recording = bandpass_filter(
+    recording, freq_min=300, freq_max=6000
+)
+recording = common_reference(
+     recording, operator="median", reference="global"
 )
 
-recording_old = load_extractor(path_ / "spikeinterface_recording.json", base_folder=path_)
-folder_old = Path(r"X:\neuroinformatics\scratch\jziminski\ephys\code\waveform_folder_old")
-waveforms_old = extract_waveforms(
-    recording_old,
-    sorting,
-    folder_old,
-    ms_before=1.5,
-    ms_after=2,
-    max_spikes_per_unit=500,
-    allow_unfiltered=True,
-    load_if_exists=True,
-)  # match kilosort
+Kilosort2_5Sorter.set_kilosort2_5_path("/ceph/neuroinformatics/neuroinformatics/scratch/jziminski/ephys/code/Kilosort-2.5")
 
-folder_new = Path(r"X:\neuroinformatics\scratch\jziminski\ephys\code\waveform_folder_new")
-waveforms_new = extract_waveforms(
-    recording_new,
-    sorting,
-    folder_new,
-    ms_before=1.5,
-    ms_after=2,
-    max_spikes_per_unit=500,
-    allow_unfiltered=True,
-    load_if_exists=True,
-)  # match kilosort
-
-breakpoint()
-
-if False:
-    import matplotlib.pyplot as plt
-
-    plt.plot(kilosort_waveform)
-    plt.show()
-
-    plt.plot(test_waveform)
-    plt.show()
-
-
-# if folder.is_dir():
-#   import shutil
-#   shutil.rmtree(folder)
-
-# run sorting without kilosort preprocessing
-# then, the `temp_wh.dat` should match exactly the original file!
-# I think this is a solid way to test. It is not possible to test against
-#
-
-
-if False:
-    original_recording = load_extractor(path_ / "spikeinterface_recording.json", base_folder=path_)
-    waveforms_old = extract_waveforms(
-        original_recording,
-        sorting,
-        folder,
-        ms_before=1.5,
-        ms_after=2.0,
-        max_spikes_per_unit=500,
-        allow_unfiltered=True,
-        load_if_exists=True,
-    )
-
-    original_recording = load_extractor(path_ / "spikeinterface_recording.json", base_folder=path_)
-
-    # TODO: unit locations don't match kilosort very well, at least in the 1-spike case.
-    # But, this could be due to windowing and should average out over many spikes
-    breakpoint()
-
-    unit_locations_old = postprocessing.compute_unit_locations(waveforms, method="center_of_mass", outputs="by_unit")
-    unit_locations_pandas = pd.DataFrame.from_dict(unit_locations, orient="index", columns=["x", "y"])
-    unit_locations_pandas.to_csv(unit_locations_path)
-
-    utils.message_user(f"Unit locations saved to {unit_locations_path}")
-
-    print(we)
+ss.run_sorter("kilosort2_5",
+              recording,
+              output_folder=output_path,
+              delete_tmp_files=False,
+              delete_recording_dat=False,
+              skip_kilosort_preprocessing=False,
+              do_correction=False,
+              scaleproc=200,
+              n_jobs=1
+              )
+"""
--- a/src/spikeinterface/extractors/KilosortTempWhExtractor.py
+++ b/src/spikeinterface/extractors/KilosortTempWhExtractor.py
@@ -9,22 +9,49 @@ import runpy
 import numpy as np
 from spikeinterface import WaveformExtractor, extract_waveforms
 from spikeinterface.preprocessing import whiten
+import probeinterface
+
+# from spikeinterface.core import channel_slice
+
+# Short-term workoarund so setting the probe does not error
+__version__ = "0.0.0"
+# Inheriting BinaryRecordingExtractor does not work due to
+# _kwargs need to be `output_path` for self.clone() but
+# must be the Binary kwargs for internal use. But, now,
+# there are other issues because everything has to be set to the
+# BinaryRecordingExtractor which is confusing.
+# Okay this really doesn't work as we need to set the probe
+# in the init but setting the probe re-initialises the
+# entire recording so the probe cannot be
+# set in the init of the recording itself.
+
+# This is super-hacky. We need to re-overwrite any properties that
+# are binary-specific and initialised on the init because to_dict() will
+# use them when cloning. Maybe this scheme is not a good one (nesting Extractors).
+# The alternative is to hold the BinaryRecordingExtractor as an attribute,
+# but this is confusing because then it doesn't behave like a recording object,
+# unless all fields are overridden
+# also, avoids this:
+
+# but, there is a problem here in that we will have to impeement every
+# function and feed it back down to BinaryRecording. and maybe we don't need to set probe.
 
 
 class KilosortTempWhExtractor(BinaryRecordingExtractor):
-    def __init__(self, output_path: Path) -> None:
-        # TODO: store opts e.g. ntb, Nbatch etc here.
-
+    def __init__(self, output_path: Path, **kwargs) -> None:
         if self.has_spikeinterface(output_path):
             self.sorter_output_path = output_path / "sorter_output"
 
+            # TODO: tidy this up and remove redundancy.
             channel_indices = self.get_channel_indices()
 
+            breakpoint()
+            # TODO: exactly what data is this loading? recording.dat or orig data?
+            # it cannot be orig because it has been pp...
             original_recording = load_extractor(output_path / "spikeinterface_recording.json", base_folder=output_path)
+
             channel_ids = original_recording.get_channel_ids()
 
-            # TODO: check this assumption - does KS change the scale / offset? can check
-            #  by performing no processing...
             if original_recording.has_scaled():
                 gain_to_uV = original_recording.get_property("gain_to_uV")[channel_indices]
                 offset_to_uV = original_recording.get_property("offset_to_uV")[channel_indices]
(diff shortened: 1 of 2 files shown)


src/spikeinterface/extractors/KilosortTempWhExtractor.py
```diff
@@ -0,0 +1,95 @@
...
+import runpy
+import numpy as np
+
+
+class KilosortTempWhExtractor(BinaryRecordingExtractor):
+    def __init__(self, output_path: Path) -> None:
+        self.sorter_output_path = output_path / "sorter_output"
+        # TODO: store opts e.g. ntb, Nbatch etc here.
+
+        params = runpy.run_path(self.sorter_output_path / "params.py")
```
alejoe91, line 16: nice! I didn't know about it!

src/spikeinterface/extractors/KilosortTempWhExtractor.py
```diff
@@ -0,0 +1,95 @@
...
+            gain_to_uV = original_recording.get_property("gain_to_uV")[
+                channel_indices
+            ]  # TODO: check this assumption - does KS change the scale / offset? can check by performing no processing...
+            offset_to_uV = original_recording.get_property("offset_to_uV")[channel_indices]
+        else:
+            gain_to_uV = None
+            offset_to_uV = None
+
+        self.original_recording_num_samples = original_recording.get_num_samples()
+        new_channel_ids = original_channel_ids[channel_indices]  # TODO: check whether this will erroneously re-order
```
alejoe91, line 45: I think this should be optional. In case someone runs KS ourside of SI, this class should still be able to load it and infer the channels etc. somewhere else, no?

alejoe91: Like the `chanMap.mat` ;) (if available)!

JoeZiminski: Hey @alejoe91 thanks for this! sorry was at a conference last week.

That's a good point, it should not be dependent on having previously run sorting with spikeinterface. The passed filepath could be checked and if it contains `spikeinterface_recording.json` + `sorter_output`, I think it is safe to assume it is SI-run sorting. Otherwise, we can check if it has some key files (`temp_wh.dat`, `channel_map.npy`)  and assume it is the sorter output.

I guess the channel information, positions can be read from `channel_map.npy` and `channel_positions.npy`. In this case, the channel IDs could just be the channel index as a string? And I don't think the uV scaling can be recovered.

Otherwise, if it is loaded from an SI recording, we can recover the uV scaling and the channel ids can be called by their proper channel ID (as currently done). Otherwise these two methods (loading with SI metadata vs. loading from kilosort output) should be equivilent.

Does that make sense / sound reasonable?

alejoe91: Perfectly reasonable! That's exactly what I was suggesting ;)

src/spikeinterface/extractors/KilosortTempWhExtractor.py
```diff
@@ -59,37 +75,168 @@ def __init__(self, output_path: Path) -> None:
...
+#             is_filtered=None,  # TODO: need to get from KS provenence?
+
+# In general, do we store the full channel map in channel contacts or do we
+# only save the new subset? My guess is subset for contact_positions, but full probe
+# for probe. Check against slice_channels.
+#             self.set_probe(probe)  # TODO: what does this mean for missing channels?
+
+#         if channel_map.ndim == 2:  # kilosort > 2
+# does kilosort > 2 store shanks differently?             channel_indices = channel_map.ravel()
+
```
zm711, line 158: @JoeZiminski ,

Kilosort  2 no longer uses the shank definition (https://github.com/MouseLand/Kilosort/issues/155), so this is definitely true for Kilosort2. Based on my search of the code base of Kilosort 3 it seems like kcoords (shanks) are still not being used: (https://github.com/search?q=repo%3AMouseLand%2FKilosort%20kcoords&type=code) (ie the code is for loading the kcoords and putting them in rez, but then they aren't actually used.

zm711: Actually not used in KS2.5 either.
https://github.com/MouseLand/Kilosort/issues/262

JoeZiminski:
Moving some discussion from #1908 here. That 

1) The raw file looks much nicer in KS than the temp_wh.dat for KS2, but by default KS uses temp_wh.dat in KS2.5 and KS3.
2) In general the waveforms look worse after whitening. Generally KS / Phy un-whitens the spikes on the fly for presentation / template analysis (see below).

In general this raises some important issues:

Thanks @zm711 and @alejoe91 for bringing up this very important point. I was operating under the misapprehension that somehow regions of data including spikes were not whitened (in fact, it is just that these regions are not used in estimation of the covariance matrix for whitening in KS).

I wonder if the changes between KS2 and KS2.5, KS3 are related to [this issue](https://github.com/MouseLand/Kilosort/issues/504) and [this discussion](https://github.com/cortex-lab/phy/issues/1040). I can confirm the full whitening matrix is saved for KS2 but only a scaling diagonal matrix for more recent versions. In phy, templates / waveforms are un-whitened on the fly for display (as far as I can tell, although exactly when this is on / off I am not entirely sure). Also the issues and [this discussion](https://github.com/cortex-lab/KiloSort/issues/35) above indicates that KS unwhitens the spike for computing the final template versions (as the matlab results variable `rez.Wraw` contain the unwhitened templates).

The primary reason for #1954 was to ensure quality metrics are computed on data that is correct with respect to the kilosort output (e.g. spike times,  channels on which spikes are strongest). However, if the case of the whitened data, it does not make sense to use this for quality metrics if waveform data must be un-whitened before interpretation. In this case, I can see that may makes more sense to use the spikeinterface pre-processed data for quality metrics calculation (assuming the user has applied bandpass filtering in spikeinterface and not kilosort). 

However, if drift correction is applied, then I have two concerns, 1) is main one relevant for this discussion:

**1)** That the waveforms extracted from the spikeinterface pre-processed data but drift-correction is applied in kilosort, then template calculation will be incorrect if the spikeinterface-preprocessed data is used. This is because all waveforms for a unit are averaged across channels, but if the spikes drifting across different channels, then the average will be incorrect. Maybe I have made an oversight though in how the templates are computed on the SI side.
2) That applying the whitening matrix calculated prior to drift correction will be incorrect after drift correction because the cross-channel covariance will be different after the kriging interpolation applied during channel shifting. This is more a discussion for another day.

This means we have a bit of a problem, if I understand correctly. The `temp_wh.dat` should be used for quality metric when drift-correction is used in Kilosort (which, I think, is true in the majority of cases). However, using this file natively may lead to sub-optimal results because everything is white. Thus, it may be required for this extractor to un-whiten the data prior to waveform extraction (I hope, the inv. whitening matrix can be found somewhere in KS2.5 and KS3).

zm711:
@JoeZiminski 

Just a few more comments to keep in  mind. Using KS in SI leads to the generation of both the whitening and inverse whitening matrices being generated, but currently the `export_to_phy` does not provide a whitening matrix. So if someone does kilosort and then SI and then does export to phy the whitening matrix that is left in the folder (if not deleted) may not apply correctly apply to the data any more. 

If phy is not given a whitening matrix it just uses `np.eye` to create a diagonal matrix of 1s. And thus it uses the same matrix to do all of its unwhitening. Since SI `export_to_phy` does not supply a whitening users that do that method will not actually (maybe because  the file management going from KS to SI is a bit tricky with all the delete options) be seeing the correct unwhitened data in Phy if using `temp_wh.dat`.

And as you said the whitening story in KS2.5 and KS3 is a bit confusing since they were trying to do some diagonal matrix (I forget the issue you linked).

Based on all this discussion it seems to me that coming from `SpikeInterface` it may be best to just use the `recording.dat` for visualization unless we can supply an appropriate whitening matrix/unwhitening matrix for Phy to use. If a user just wants to use KS inside of SpikeInterface then it is likely not a problem, but as soon as people start doing pre/post processing etc in SpikeInterface I think the `temp_wh.dat` will no longer be synced up.

What I don't know at all is how the drift correction fits into this. I suppose we could do the drift correction `['kilosort-like']` (I forget the actual argument) to try to approximate the drift correct, but I think that runs into your (2) above which we can save for another time.

Sorry for the stream of consciousness, but wanted to jot down my ideas (as inchoate as they are), before I forgot.

JoeZiminski:
Hi @zm711 thanks a lot, not at all that's very useful. I have not used `export_to_phy`, what are the kind of cases it is used it? I always thought it was to convert non-kilosort sorting outputs for use in Phy, but did not realise you would use it if you had run KS already (because KS output is already phy-compatible). Is this because there are some postprocessing that can be done in SI and then exported back to Phy format?

I agree I think if the whitening matrix is not available, then this `temp-wh.dat` should not be used. I am leaning towards having the `get_traces()` method of this extractor apply the whitening matrix by default. That way, waveforms will be computed similar to that in Phy, although I fear the scaling is going to be a bit of a nightmare especially across kilosort versions.

JoeZiminski:
Thanks @zm711 that makes sense, I think it is nice to cater for as many edge cases as possible. In this case, as I understand it in the case of KS sorting, reading the `temp_wh.dat` after any `export_to_phy` should not be a problem as `export_to_phy` is only concerned with operations at a higher level than the underlying preprocessed data. e.g. if new unit ids, template amplitudes or principal components are computed from the preprocessed data and written to new `.npy` files, that is okay an should not affect the operation of this extractor. As long as `export_to_phy` is not changing the preprocessed data in any way (e.g. the binary itself or scaling information such as the whiten matrix npy), then I believe there should be no conflict. 

That's a good point about using part of the KS workflow. The cases I can think of are:

1) No SI preprocessing - preprocessing is KS, drift correction in KS+ sorting in KS
    - this is best case scenario for this extractor. Because the waveforms based on the SI recording will be completely wrong but using `temp_wh.dat` will be correct.
2) SI preprocessing - preprocessing is KS, drift correction in KS+ sorting in KS
    - Similar to above, the main problem been the units are determined on drift-corrected data but waveforms computed on non-drift corrected data.
3) preprocessing in SI, skip preprocessing in KS, drift correction in KS+ sorting in KS
   - This is a cool case, in this case the `temp_wh.dat` is not written but KS does overwrite the data, I assume this is done on the `recording.dat`. This would be nice as it is an easy way to load the drift-corrected data for computing waveforms (as can just use SI to re-load the `recording.dat`, that has been drift corrected by KS) . In this case, this extractor could not be used.
4) preprocessing in SI, skip preprocessing in KS, skip direct correction in KS + sorting in KS
- This is nice, as `temp_wh.dat` is not written so the normal spikeinterface pipeline could be used without issue.
5) Any of the above, but changing the sorter outputs using `export_to_phy`
    - I believe as above, as long as `export_to_phy` is changing only sorting results and not data preprocessing this should have no additional effect.

JoeZiminski:
Just to get a handle on the actually effect of calculating templates based on drift vs. noncorrected drift data I tested. First preprocessing run in spikeinterface, then KS is run with `skip_kilosort_preprocessing=True` but `do_correction=True` in kilosort2.5. This means the `recording.dat` is written with KS drift correction.

Then the waveforms can be calculated from the preprocessed data, or preprocessed + drift correction. Some example units from a 10 minute recording in the attached PDF. In general the spikes are sharper after drift correction, as you'd expect, 

[dirft_vs_nodrift.pdf](https://github.com/SpikeInterface/spikeinterface/files/12690975/dirft_vs_nodrift.pdf)

zm711:
Thanks! I'll take a look myself this weekend.

JoeZiminski:
Just a reminder to self, one benefit of this will would be to allow visualisation of the KS data prior to sorting, this might shed light on some of the strange KS indexing errors (one possibility is the input data is completely scrambled in some way by the preprocessing).

JoeZiminski:
Hey @zm711 @alejoe91 further to the discussion in #2650 I think it would make sense to close this PR. Realistically I doubt I will ever complete it as I wouldn't be confident it could be robust across KS versions. Also KS versions < 4 seem to be basically unsupported now. 

Effort would probably be better placed in recommend users run preprocessing in spikeinterface and skip KS preprocessing in order to inspect preprocessing outputs. Also if really necessary it is possible to hack the KS code to skip whitening in order to assess the KS motion correction directly.  But I think handling the un-whitening and various offset corrections across versions seems very difficult with relatively little payoff. Thoughts?

zm711:
@JoeZiminski I think that is fair. I know I don't have the bandwidth to pick this up. So far you and rat-h have been the only people to request this so it is desired but not super desired. Not sure if @alejoe91 wants to keep this around though.

alejoe91:
Maybe what we could do is to show that the SI preprocessing and the KS one (the temp_wh.dat) are the same?

JoeZiminski:
> Maybe what we could do is to show that the SI preprocessing and the KS one (the temp_wh.dat) are the same?

This is a nice idea! I think with the existing settings it is possible to run KS through SI with / without preprocessing (filter, CMR, whiten) and drift correction independently? (e.g. so it would be possible to test for example drift correction with / without whitening). 

I wonder if it is also worth having some kind of backwards-compatibility tests that test running KS through SI vs. KS directly (e.g. the KS GUI). In theory these should be identical but I saw some issues in which there was slower performance through SI than KS directly (though maybe this is related only to KS4). I guess this would also protect against any future breaking changes. Do you think this additional layer of tests is worthwhile?

alejoe91:
yes, we have the `do_correction=True/False` to enable/disable drift and `skip_kilosort_preprocessing` parameters

alejoe91:
> I wonder if it is also worth having some kind of backwards-compatibility tests that test running KS through SI vs. KS directly (e.g. the KS GUI). In theory these should be identical but I saw some issues in which there was slower performance through SI than KS directly (though maybe this is related only to KS4). I guess this would also protect against any future breaking changes. Do you think this additional layer of tests is worthwhile?

I think this would be tricky and a bit too kilosort-centric

JoeZiminski:
Thanks! I agree the implementation may be too tricky to make feasiable. Theoretically though I would argue it should be kilosort-centric as it is essentially a wrapper around kilosort and so you would expect the output be 1:1 the same as running in kilosort directly. 

For example a test such as:
1) load spikeglx file in SI, don't do any preprocessing on it in SI.
2) pass it directly to kilosort in SI for sorting
3) compare output with a previously saved output from running in KS directly.

I am in general tentative about such tests to a fixed standard dataset (which is what the saved output from KS would be here) because intended changes elsewhere in the codebase can inadvertantly affect the thing you are testing and it is difficult to debug. In this case though I don't think it is an issue as there is basically no processing on the SI side unrelated to running KS, except for file loading, which should never change.

alejoe91:
I think these are good examples to run, but not as CI tests. Maybe creating some GISTS/Blogs?

JoeZiminski:
but what if a future change to the KS-running machinery in SI (either the master scripts or something in the images) leads to a bug that results in KS in SI diverging from native KS? My understanding is at present this would not be caught in the CI?

alejoe91:
> but what if a future change to the KS-running machinery in SI (either the master scripts or something in the images) leads to a bug that results in KS in SI diverging from native KS? My understanding is at present this would not be caught in the CI?

That is true. On the other hand, I think that having proper unit-tests for the different steps involved is the way to go. 
I would vote not to be too Kilosort-dependent :)

JoeZiminski:
Yes I agree it makes sense to make this as modular as possible. It's something that will come up soon as I look at bit more at the motion correction, will open an issue for discussion down the line. Cheers!

Djoels:
I'm probably going to have to go through the effort of trying to reverse engineer KS2.5/4 preprocessing, in which case the effort that was done here could be helpful. (Luckily we are not performing drift correction as our in vitro setting doesn't require this).

Would there be any reason this can not be done if whitening_mat_inv.npy and whitening_mat.npy files are present?
The extractor could yield a FileNotFoundError if necessary files are not present?

JoeZiminski:
Hey @Djoels, cool let me know how get on this this, it would be of interest to me also. I think if you are happy with the whitened version of the data, this is relatively straight forward. If I recall correctly, the data is written as binary and padded with zeros at the end, but would just require clipping these zeros off. 

On the whitening side, I am not 100% sure what is going on across versions. In earlier KS versions (1-2), the inverse whitening matrix was saved, so unwhitening the preprocessed data should not be an issue. For 2.5-3 the inverse matrix is now identity (well, identity * scaleproc (default 200)). I am not sure why it was no longer saved. At some point, the templates were saved whitened and needed to be unwhited, e.g. [here](https://github.com/cortex-lab/spikes/blob/fcea2b20e736b533e5baf612752b66121a691128/analysis/templatePositionsAmplitudes.m#L26).  I'm not sure what, but somewhere in the pipeline something has changed and so it seems no longer necessary to fully unwhiten the templates. And the inverse whitening matrix just used for scaling these templates. You can see this change was made [here](https://github.com/MouseLand/Kilosort/blob/f2e570c61b6e920d361bcae6e2350bea03a16506/utils/rezToPhy.m#L69).  However, this means it's not possible to unwhiten the raw data for these versions, as far as I know.

So, I think if you are happy with whitened data only, all versions are good. Otherwise, you will have to change [this line](https://github.com/MouseLand/Kilosort/blob/f2e570c61b6e920d361bcae6e2350bea03a16506/utils/rezToPhy.m#L69) in the kilosort source code to make it save the full inverse whitening matrix (I would save this to a file with a different name, so it doesn't mess up other workflows e.g. phy). I'm not sure what the situation is for kilosort 4 but would be interested in finding out. 

I think this should be possible to hack for all versions, but gave up doing in a general way here as there are too many cases to reliably handle, and it seems the [scaling](https://github.com/cortex-lab/phy/issues/1040) is not fully resolved. Another option is to go into the `rez.mat`, there are more data accessible here, but I never delved in.

PR #1955, opened 2023-09-02
Fix the [full] install for Macs

zm711:
Error with trying to install ".[full]" on a Mac.
```
ERROR: Could not find a version that satisfies the requirement cuda-python; extra == "full" (from spikeinterface[full]) (from versions: none)
ERROR: No matching distribution found for cuda-python; extra == "full"
```

I think they are trying to switch from sys_platform to platform_system in general, but I haven't read too deeply, but based on the fact the old way is error on my M- Mac...

Turns out the boolean can be written as I have (using platform instead of sys). It is discussed here:
https://peps.python.org/pep-0508/#environment-markers

and is now the example given by the 
PyPA for setuptools
https://setuptools.pypa.io/en/latest/userguide/dependency_management.html

(go down about 1/3 the page)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -151,9 +151,9 @@ docs = [
     # for notebooks in the gallery
     "MEArec",   # Use as an example
     "datalad==0.16.2",  # Download mearec data, not sure if needed as is installed with conda as well because of git-annex
-    "pandas", # Don't know where this is needed
-    "hdbscan>=0.8.33",   # For sorters, probably spikingcircus
-    "numba", # For sorters, probably spikingcircus
+    "pandas", # in the modules gallery comparison tutorial
+    "hdbscan>=0.8.33",   # For sorters spykingcircus2 + tridesclous
+    "numba", # For many postprocessing functions
     # for release we need pypi, so this needs to be commented
     "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",  # We always build from the latest version
     "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",  # We always build from the latest version
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,7 +91,7 @@ full = [
     "networkx",
     "distinctipy",
     "matplotlib",
-    "cuda-python; sys_platform != 'darwin'",
+    "cuda-python; platform_system != 'Darwin'",
     "numba",
 ]
 


zm711:
This test failure is maybe related to the new `generate.py`? 
```
FAILED src/spikeinterface/comparison/tests/test_hybrid.py::test_hybrid_spikes_recording - Exception: Warning!
Your templates do not go to 0 on the edges in InjectTemplatesRecording.__init__
Please make your window bigger.
========== 1 failed, 23 passed, 639 deselected, 2 warnings in 34.24s ===========
```

Since the PR only affects installation on Macs and not on Ubuntu...

zm711:
At least for the profile imports, the macOS runner is usually Intel instead of M arm architecture. So using the format I added should work for both (although with Mac allowing use of Rosetta terminal-someone on arm can emulate the intel for the install so the Mac stuff is definitely super complicated).

zm711:
Last note:
So just for fun I tried to uninstall and reinstall (all in a conda environment) a few times with ".[full]" on the M1 and after caching all the files from the original install in never errored again with `sys_platform` or `platform_system`, it only errored on the initial attempt to install, but not subsequent uninstall-reinstall. Still probably best to just use `platform_system`, but I wish the bug was a little less flaky.

alejoe91:
Thanks @zm711 

@samuelgarcia can you check the error in the test here? Maybe we have some unset seeds and we didn't spot this before!

alejoe91 (approved):

pyproject.toml
```diff
@@ -151,9 +151,9 @@ docs = [
     # for notebooks in the gallery
     "MEArec",   # Use as an example
     "datalad==0.16.2",  # Download mearec data, not sure if needed as is installed with conda as well because of git-annex
-    "pandas", # Don't know where this is needed
-    "hdbscan>=0.8.33",   # For sorters, probably spikingcircus
-    "numba", # For sorters, probably spikingcircus
+    "pandas", # in the modules gallery comparison tutorial
+    "hdbscan>=0.8.33",   # For sorters spykingcircus2 + tridesclous
+    "numba", # For many postprocessing functions
```
h-mayorquin, line 156: Thanks a bunch for adding those comments. I added the others. 

Concerning numba, are they used to generate the examples on quality metrics?

h-mayorquin (approved):
This is good. Thanks for fixing it and providing links to the relevant documentation.

pyproject.toml
zm711: https://spikeinterface.readthedocs.io/en/latest/how_to/get_started.html
uses the post processing functions. This is just a written out tutorial, but sphinx can be weird at import, so I'm also still not sure where exactly it is required, but I know the post processing module requires it. Quality metrics are a possibility, since numba is used for the sliding rp violation. I didn't see that as a specific example used, but (there are a lot of docs :) ) . Tridesclous2 also uses numba, but again the examples use Tridesclous ( and I didn't check Tridesclous source for this). 

So I think between the two of us we will eventually get the comments right!

h-mayorquin: Great. Thanks for the improvements : )

PR #1957, opened 2023-09-04
Allow to postprocess on read-only waveform folders

alejoe91:
In some cases a waveform extractor folder can be loaded in read-only mode (e.g., precomputed waverorms on the cloud, attached as data).

The waveform extensions always inherit the `format` from the parent `WaveformExtractor`. However, if the parent is read-only all postprocessing crashes because it tries to add folders to the parent folder.

This PR adds the `is_read_only()` method to the `WaveformExtractor`. In case a waveform extractor is read-only, extensions can be computed, but they will be in memory and not persistent to disk.

EDIT:

Added a small fix to load the `similarity` extension in the `export_to_phy` if that's already available!

### DIFF
--- a/src/spikeinterface/core/waveform_extractor.py
+++ b/src/spikeinterface/core/waveform_extractor.py
@@ -1754,7 +1754,7 @@ class BaseWaveformExtractorExtension:
                     if not self.waveform_extractor.is_read_only():
                         self.extension_folder.mkdir()
                     else:
-                        raise Exception(
+                        warn(
                             "WaveformExtractor: cannot save extension in read-only mode. "
                             "Extension will be saved in memory."
                         )
@@ -1770,7 +1770,7 @@ class BaseWaveformExtractorExtension:
                     if not self.waveform_extractor.is_read_only():
                         self.extension_group = zarr_root.create_group(self.extension_name)
                     else:
-                        raise Exception(
+                        warn(
                             "WaveformExtractor: cannot save extension in read-only mode. "
                             "Extension will be saved in memory."
                         )
--- a/src/spikeinterface/core/waveform_extractor.py
+++ b/src/spikeinterface/core/waveform_extractor.py
@@ -1751,9 +1751,7 @@ class BaseWaveformExtractorExtension:
             if self.format == "binary":
                 self.extension_folder = self.folder / self.extension_name
                 if not self.extension_folder.is_dir():
-                    if not self.waveform_extractor.is_read_only():
-                        self.extension_folder.mkdir()
-                    else:
+                    if self.waveform_extractor.is_read_only():
                         warn(
                             "WaveformExtractor: cannot save extension in read-only mode. "
                             "Extension will be saved in memory."
@@ -1761,15 +1759,16 @@ class BaseWaveformExtractorExtension:
                         self.format = "memory"
                         self.extension_folder = None
                         self.folder = None
+                    else:
+                        self.extension_folder.mkdir()
+
             else:
                 import zarr
 
                 mode = "r+" if not self.waveform_extractor.is_read_only() else "r"
                 zarr_root = zarr.open(self.folder, mode=mode)
                 if self.extension_name not in zarr_root.keys():
-                    if not self.waveform_extractor.is_read_only():
-                        self.extension_group = zarr_root.create_group(self.extension_name)
-                    else:
+                    if self.waveform_extractor.is_read_only():
                         warn(
                             "WaveformExtractor: cannot save extension in read-only mode. "
                             "Extension will be saved in memory."
@@ -1777,6 +1776,8 @@ class BaseWaveformExtractorExtension:
                         self.format = "memory"
                         self.extension_folder = None
                         self.folder = None
+                    else:
+                        self.extension_group = zarr_root.create_group(self.extension_name)
                 else:
                     self.extension_group = zarr_root[self.extension_name]
         else:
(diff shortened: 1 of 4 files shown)


src/spikeinterface/core/waveform_extractor.py
```diff
@@ -1863,53 +1893,56 @@ def save(self, **kwargs):
...
-                elif isinstance(ext_data, np.ndarray):
-                    self.extension_group.create_dataset(name=ext_data_name, data=ext_data, compressor=compressor)
-                elif isinstance(ext_data, pd.DataFrame):
-                    ext_data.to_xarray().to_zarr(
-                        store=self.extension_group.store, group=f"{self.extension_group.name}/{ext_data_name}", mode="a"
-                    )
-                    self.extension_group[ext_data_name].attrs["dataframe"] = True
-                else:
-                    try:
+        if not self.waveform_extractor.is_read_only():
```
h-mayorquin, line 1896: The diff of this is very difficult for me to catch up. Is it this a tab  of everything to the right?

If so, 

I think that you can get away by just doing an early return if the program is ready only.

That is:

```python
if self.waveform_extractor.is_read_only()
   return None
```

This will decrease nesting that at this level is very hard to read and keep track.

On the other hand, maybe this should ben assertion? That is, if a user wants to write to a folder that is read only maybe this is the point where we can tell them with a meanintgul assertion message?

What do you think?

alejoe91: Yes it's a conditional load (only a tab separation).

While I agree that the diff iis harder to read, I think that logically it makes more sense to keep it this way. 
The logic is:
**If I can read it, then read it**

No?

src/spikeinterface/core/waveform_extractor.py
```diff
@@ -517,15 +524,19 @@ def is_extension(self, extension_name) -> bool:
...
-            if self.format == "binary":
-                return (self.folder / extension_name).is_dir() and (
-                    self.folder / extension_name / "params.json"
-                ).is_file()
-            elif self.format == "zarr":
-                return (
-                    extension_name in self._waveforms_root.keys()
-                    and "params" in self._waveforms_root[extension_name].attrs.keys()
-                )
+            # Extensions already loaded in memory
```
h-mayorquin, line 527: I also think that this part can benefit from early return after if (if condition: return None) and avoiding the else to decrease nesting.

src/spikeinterface/core/waveform_extractor.py
h-mayorquin: I think this makes sense if the scope of the scope of the function was simpler.

But here, there is like what, six levels of nesting below? 

I think that readability in the sense of removing one if block from your working memory thriumps faithful representation of logic.

A solution that followins from this line of reasoning is that maybe this if should go in in the function `save` and not here in `_save` as there the logic is expressed perfectly without hindering readability.

h-mayorquin (approved):
This seems like an important feature to support. 

I have some comments about nesting that come from clomatic complexityh which is related to code complexity / hader-to-read more bugs:

https://en.wikipedia.org/wiki/Cyclomatic_complexity

This is why I try to decrease this when I see it.

As a more general comment I wonder if there is another way of accomplishing (supporting operating in read only waveforms)  without having to keep state and sacrifice caching.

Could we just store the extension folder somewhere else instead of the folder where we usually do? After all, that is like or own in-house format so we can do this right? Like, we can use the local caching of the user (temp) and save that as state so the user can still benefit from caching and we keep the state as the folder of the extensions instead of a flag to see if the original folder of the waveforms is read only.

Maybe there are obvious downsides to this that I had not considered. Just throwing out some other possible solutions.

src/spikeinterface/core/waveform_extractor.py
```diff
@@ -399,6 +403,9 @@ def return_scaled(self) -> bool:
     def dtype(self):
         return self._params["dtype"]
 
+    def is_read_only(self) -> bool:
```
h-mayorquin, line 406: Another comment.

An advice that I have read is that logical / boolean flags should be designed in such a way that most of the testing you do will test for True. 

Reason being, that putting not in front of flags is a little bit harder to read as you need to keep two things in working memory instead of one (like remembering the sign of an expression that people usually mess up in algebraic high school operations). 

Here most of the tests are for not `is_read_only`. On the other hand, "is_read_only" is the common category  / flag that the OS usually uses so that seem more important here.

Just a comment.

src/spikeinterface/postprocessing/tests/common_extension_tests.py
```diff
@@ -76,6 +76,15 @@ def setUp(self):
             overwrite=True,
         )
         self.we2 = we2
+
+        # make we read-only
+        we_ro_folder = cache_folder / "toy_waveforms_2seg_readonly"
+        if not we_ro_folder.is_dir():
+            shutil.copytree(we2.folder, we_ro_folder)
+            # change permissions (R+X)
+        we_ro_folder.chmod(0o555)
```
h-mayorquin, line 85: I never learned by memory the chmod changing policy. Maybe I should wirte a post-it in my computer with them until I learn them.

src/spikeinterface/core/waveform_extractor.py
alejoe91: We could do `is_not_read_only` instead! That's a good point!

src/spikeinterface/postprocessing/tests/common_extension_tests.py
zm711: I know this is a super future problem, but chmod doesn't fully work with Windows. So you can't change the writeable bit directly using chmod operations (also can't change the executable bit either). If there is ever a common, full testing framework (that Sam approves of) then this may behave unexpected on Windows tests.
https://docs.python.org/3/library/pathlib.html
https://docs.python.org/3/library/os.html#os.chmod

alejoe91: Good point! Maybe skipping this if on windows for now! I'll make the suggested changes on monday

src/spikeinterface/core/waveform_extractor.py
alejoe91: done

src/spikeinterface/core/waveform_extractor.py
alejoe91: done

alejoe91:
@h-mayorquin @zm711 I applied the suggested changes!

src/spikeinterface/core/waveform_extractor.py
```diff
@@ -517,15 +524,19 @@ def is_extension(self, extension_name) -> bool:
...
-                ).is_file()
-            elif self.format == "zarr":
-                return (
-                    extension_name in self._waveforms_root.keys()
-                    and "params" in self._waveforms_root[extension_name].attrs.keys()
-                )
+            # Extensions already loaded in memory
+            if extension_name in self._loaded_extensions:
+                return True
+            else:
```
h-mayorquin, line 530: You don't need this else. This will allow you to get rid of the nesting.

src/spikeinterface/core/waveform_extractor.py
```diff
@@ -517,15 +524,19 @@ def is_extension(self, extension_name) -> bool:
         if self.folder is None:
             return extension_name in self._loaded_extensions
         else:
```
h-mayorquin, line 526: You don't need this else either. This will allow you to get rid of the nesting.

src/spikeinterface/postprocessing/tests/common_extension_tests.py
```diff
@@ -76,6 +77,16 @@ def setUp(self):
             overwrite=True,
         )
         self.we2 = we2
+
+        # make we read-only
+        if platform.system() != "Windows":
```
h-mayorquin, line 82: Does the test works in windows if this statements is not run? I am wondering if what should be skipped is the test itself if we rely chmod. As @zm711 has pointed out that this is not reliable on windows.

h-mayorquin: Altought the core tests are running on windows so...

h-mayorquin (changes requested):
Just one last question about the tests and this should be good to go.

src/spikeinterface/postprocessing/tests/common_extension_tests.py
zm711: My point is not that it won't work just that it may introduce a hard to detect bug in the future since it does not behave as the user intends. So if you ever have a test that address writability, Windows behavior would diverge from Linux/Mac. If all the tests are only testing the readability, then they wouldn't "experience" the difference in the file identities. So that's why I said future problem since it may in the future add a hard to detect bug, but won't definitely introduce one.

src/spikeinterface/core/waveform_extractor.py
alejoe91: Done

alejoe91:
@h-mayorquin updated with the suggestion.

@samuelgarcia only waiting for your review

samuelgarcia (approved):

PR #1958, opened 2023-09-04
Refactor ISI calculation numpy and numba

zm711:
@DradeAW (I know you're busy so just wanted to fyi since I saw you did the initial implementation)

*Changes to Numpy*
In numpy implementation rather than have `np.histogram` calculate the bins each time by giving the `num_bins` and the `range`, I instead just use the pre-calculated bins. Since it is a double for loop just feeding bins in reduces recalculating the bins multiple times. This led to speed up of 1.44x. I also added an assert just to prevent users from picking too small a bin size.

*Changes to Numba*
Numba allows for use of `np.histogram` so rather than switch between python->numba->python->numba I was able to remove a lot of this chain by using just using histogram. (Although if there is a reason you didn't want to use histogram just let me know!). Due to numba being so fast I really saw only ~1% speed up doing this. So the main advantage is just reduced code burden to keep track of in the future.

One thing that did lead to a bit of a speed up though was removing parallelization. Since creating the threads for parallelization has a significant overhead, it is actually faster to just doing things serially until users have high hundreds or thousands of units. Since I think most users are still <500 units (although correct me if users are telling you they are working on >500), it is actually better not to spend time creating threads. So removing the parallel led to between 1.25-1.39x speed up in my tests. Again if we want to cater more toward people with 1000s of units (although to be fair their lives are already too good) then I could switch it back to parallelization on with `numba.prange`.

### DIFF
--- a/src/spikeinterface/postprocessing/isi.py
+++ b/src/spikeinterface/postprocessing/isi.py
@@ -205,6 +205,7 @@ if HAVE_NUMBA:
     def _compute_isi_histograms_numba(ISIs, spike_trains, spike_clusters, bins):
         n_units = ISIs.shape[0]
 
-        for i in range(n_units):
+        units_loop = numba.prange(n_units) if n_units > 300 else range(n_units)
+        for i in units_loop:
             spike_train = spike_trains[spike_clusters == i]
             ISIs[i] += np.histogram(np.diff(spike_train), bins=bins)[0]


DradeAW:
As long as you make sure that it's giving the same results in all cases (meaning with max_t being a multiple of dt or not), I am fine with these changes :)

I think I was probably just lazy and changed the code from correlograms to ISI ^^

zm711:
That's what I'm testing right now :). I didn't change your calculation for that part.

zm711:
@alejoe91 @samuelgarcia 

I think this has to be a seed issue. The failing test isn't anything I touched. The tests for my refactor all passed.

samuelgarcia:
> I think this has to be a seed issue. The failing test isn't anything I touched. The tests for my refactor all passed.

I will try to have a look but this is a big pain.

zm711:
That did it! Thanks @samuelgarcia !

src/spikeinterface/postprocessing/isi.py
```diff
@@ -256,16 +197,14 @@ def compute_isi_histograms_numba(sorting, window_ms: float = 50.0, bin_ms: float
...
         nogil=True,
         cache=True,
-        parallel=True,
     )
-    def _compute_isi_histograms_numba(ISIs, spike_trains, spike_clusters, max_time, bin_size, sampling_f):
+    def _compute_isi_histograms_numba(ISIs, spike_trains, spike_clusters, bins):
         n_units = ISIs.shape[0]
 
-        for i in numba.prange(n_units):
+        for i in range(n_units):
```
alejoe91, line 208:
```suggestion
        units_loop = numba.prange(n_units) if n_units > 300 else range(n_units)
        for i in units_loop:
```
I guess that the actual breakeven point will depend both on the number of units and the length of the spike trains, but as a rule of thumb I think that only using the number of units would work. @zm711 dows this make sense?

zm711: The length of the spike trains are done sequentially regardless so for this optimization we can try to do the units in parallel, but I wouldn't have a clue how to parallelize doing the actually spike train counting. (Long-winded way of saying I agree with you--had to think it out myself :) ). 

To your point though, I haven't actually tested to see where the value of spinning threads actually switches to a net benefit vs a net cost. 300 seems low to me (I would imagine doing 500) but we are both just picking round numbers I think? So 300 sounds works for me!

alejoe91 (approved):

Issue #1870, opened 2023-07-21
Numpy 1.24 breaks a few sorters


jonahpearl:
Hi there -- I'm installing spikeinterface today and noticed a small issue: the removal of the numpy alias `np.bool` breaks `herdingspikes` in addition to `trisdeclous` (as noted in the install cheat sheet ymls). (This is not the only popular repo where this alias has caused problems...) Maybe consider pinning `numpy=1.22` so that new spikeinterface installs don't hit this issue, while spike sorter maintainers have a chance to fix this bug, unless the last two versions of numpy really provide some major / required improvements?

h-mayorquin:
I guess that @samuelgarcia can fix `tridesclous` but the last release of herdingspikes was three years ago.

alejoe91:
Thanks for reporting @jonahpearl 

I think `tridesclous.1.6.8` (release 2 days ago) is updated, since it passes tests on our side too. 

For herdingspikes, as @h-mayorquin says, they should fix it on their end. We don't want to pin numpy because a specific sorter doesn't support a certain version. The user can either handle the upgrades/downgrades or run with docker/singularity (recommended!)

h-mayorquin:
Maybe somebody can convince Mathias to make some updates. I remember other deprecations and problems over there. We needed to import something prebuilding or something like that. But PI life is a busy one (not being sarcastic here!).

h-mayorquin:
@jonahpearl can we close this issue? Is there anything on the documentation you think could be improved to indicate the state of affaris? 

Right now it someone try to use `herdingspikes` what happens? Is the error informative of what is the problem and what needs to be done?

jonahpearl:
Good question, thanks. I encountered the issue when using the [`check_your_install.py`](https://github.com/SpikeInterface/spikeinterface/blob/main/installation_tips/check_your_install.py) script. When it failed, there was no error message other than "failed", so I went and ran it manually to see the error message. From there it was pretty clear that I had to downgrade `numpy` (though `herdingspikes` at this point requires a pretty specific `numpy` version due to [this error](https://github.com/mhhennig/HS2/issues/58#issuecomment-1645986966) which requires *up*grading it past a certain point, so that was a bit frustrating). 

[...]

Issue #1915, opened 2023-08-09
Error using run_sorter 


Sietse-k:
Hi, I'm working on the "get started" spikeInterface guide (https://spikeinterface.readthedocs.io/en/latest/how_to/get_started.html). Using simulated data. However, I run into the following problem:

sorting_TDC = ss.run_sorter(sorter_name="tridesclous", recording=recording_preprocessed, detect_threshold=4)
print(sorting_TDC)

Traceback (most recent call last):
  File "C:\Users\Sietse\anaconda3\envs\si_env\lib\site-packages\IPython\core\interactiveshell.py", line 3508, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-3-777f8a455281>", line 1, in <module>
    sorting_TDC = ss.run_sorter(sorter_name="tridesclous", recording=recording_preprocessed, detect_threshold=4)
  File "C:\Users\Sietse\anaconda3\envs\si_env\lib\site-packages\spikeinterface\sorters\runsorter.py", line 148, in run_sorter
    return run_sorter_local(**common_kwargs)
  File "C:\Users\Sietse\anaconda3\envs\si_env\lib\site-packages\spikeinterface\sorters\runsorter.py", line 174, in run_sorter_local
    SorterClass.run_from_folder(output_folder, raise_error, verbose)
  File "C:\Users\Sietse\anaconda3\envs\si_env\lib\site-packages\spikeinterface\sorters\basesorter.py", line 269, in run_from_folder
    raise SpikeSortingError(
spikeinterface.sorters.utils.misc.SpikeSortingError: Spike sorting error trace:
Traceback (most recent call last):
  File "C:\Users\Sietse\anaconda3\envs\si_env\lib\site-packages\spikeinterface\sorters\basesorter.py", line 234, in run_from_folder
    SorterClass._run_from_folder(sorter_output_folder, sorter_params, verbose)
  File "C:\Users\Sietse\anaconda3\envs\si_env\lib\site-packages\spikeinterface\sorters\external\tridesclous.py", line 149, in _run_from_folder
    tdc.apply_all_catalogue_steps(cc, catalogue_nested_params, verbose=verbose)
  File "C:\Users\Sietse\anaconda3\envs\si_env\lib\site-packages\tridesclous\cataloguetools.py", line 119, in apply_all_catalogue_steps
[...]

Issue #1476, opened 2023-04-03
PathNotFoundError with Zarr recording extractor when extracting waveforms


vncntprvst:
I'm trying to load waveforms, from a recording that was saved in zarr format. 
When calling 
```
we = si.extract_waveforms(recording_saved_zarr, sorting, folder=base_folder / "waveforms", 
                          load_if_exists=True, overwrite=False, **job_kwargs)
```
I get this error:
```---------------------------------------------------------------------------
PathNotFoundError                         Traceback (most recent call last)
File ~/mambaforge/envs/si_env/lib/python3.9/site-packages/spikeinterface/core/waveform_extractor.py:162, in WaveformExtractor.load_from_folder(cls, folder, with_recording, sorting)
    161 try:
--> 162     recording = load_extractor(folder / 'recording.json',
    163                             base_folder=folder)
    164     rec_attributes = None

File ~/mambaforge/envs/si_env/lib/python3.9/site-packages/spikeinterface/core/base.py:911, in load_extractor(file_or_folder_or_dict, base_folder)
    910 else:
--> 911     return BaseExtractor.load(file_or_folder_or_dict, base_folder=base_folder)

File ~/mambaforge/envs/si_env/lib/python3.9/site-packages/spikeinterface/core/base.py:536, in BaseExtractor.load(file_path, base_folder)
    535     return None
--> 536 extractor = BaseExtractor.from_dict(d, base_folder=base_folder)
    537 return extractor

File ~/mambaforge/envs/si_env/lib/python3.9/site-packages/spikeinterface/core/base.py:371, in BaseExtractor.from_dict(d, base_folder)
    370     d = _make_paths_absolute(d, base_folder)
--> 371 extractor = _load_extractor_from_dict(d)
    372 folder_metadata = d.get('folder_metadata', None)

File ~/mambaforge/envs/si_env/lib/python3.9/site-packages/spikeinterface/core/base.py:854, in _load_extractor_from_dict(dic)
    853 # instantiate extrator object
--> 854 extractor = cls(**kwargs)
    856 extractor._annotations.update(dic['annotations'])

File ~/mambaforge/envs/si_env/lib/python3.9/site-packages/spikeinterface/core/zarrrecordingextractor.py:55, in ZarrRecordingExtractor.__init__(self, root_path, storage_options)
     53     root_path_kwarg = root_path_init
---> 55 self._root = zarr.open(root_path_init, mode="r", storage_options=storage_options)
[...]

Issue #1447, opened 2023-03-22
Problem opening Phy after export_to_phy


marinemm-fr:
I used the function export to Phy with the following parameters : 
`export_to_phy(we_clean, base_folder / 'export_to_phy', compute_pc_features=True, compute_amplitudes=True, sparsity=None, copy_binary=True, remove_if_exists=False, peak_sign='neg', template_mode='median', dtype=None, verbose=True, **job_kwargs)`
I have around 540 units and my recording is about 30 min long, the .dat file generated is about 40 GB and I didn't had any error while creating the file.
When running phy template.gui I have the following error : 
```
from phy.apps.template import template_gui
template_gui("recording.dat")
Traceback (most recent call last):

  Cell In[7], line 2
    template_gui("recording.dat")

  File ~\anaconda3\envs\si_env\lib\site-packages\phy\apps\template\gui.py:209 in template_gui
    model = load_model(params_path)

  File ~\anaconda3\envs\si_env\lib\site-packages\phylib\io\model.py:1433 in load_model
    return TemplateModel(**get_template_params(params_path))

  File ~\anaconda3\envs\si_env\lib\site-packages\phylib\io\model.py:1416 in get_template_params
    params = read_python(params_path)

  File ~\anaconda3\envs\si_env\lib\site-packages\phylib\utils\_misc.py:171 in read_python
    contents = path.read_text()

  File ~\anaconda3\envs\si_env\lib\pathlib.py:1267 in read_text
    return f.read()

  File ~\anaconda3\envs\si_env\lib\encodings\cp1252.py:23 in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]

MemoryError
```

Seems like I ran out of memory (I have 16G RAM).

However when I just run Phy directly from the output from Kilosort 2.5 I am able to open the GUI and work on it.
Does anyone have an idea of how I could work around that problem and be able to run Phy on my curated recording ?
Thank you !

alejoe91:
Hi,

The error is weird, because it seems 9t be generated by a `read_text()` call in reading the parameters.
Can you share the `params.py`? (Even a copy paste here would work)

zm711:
@marinemm-fr 
@alejoe91 

I just realized what the problem with this is. `template_gui` needs to be called on the `params.py` not on the binary file, `recording.dat`. The gui was trying to read the whole ~40gb binary file.

h-mayorquin:
@marinemm-fr 
[...]

PR #1962, opened 2023-09-05
Update installation Tips

zm711:
Last PR before I need to prep some school stuff. 

I updated the installation tips since many users with issues have the `si_env` in the traceback they are using the ymls to create their envs. I moved to 3.10 for now (although I've been using 3.11 myself with no issues).  I released the joblib and numpy restrictions and tested creating this env and running tridesclous with no issues. Bumped up tridesclous to the newest version and also bumped up neo to 0.12. Worked fine on windows, though I can test on a mac too later. 

Also deleted `elephant` since it isn't a dependency. Is there a reason why we would want to keep having SI users download it to their env?

Finally, I added a tiny error log so that if someone runs the code and it fails it will at least say why things failed rather than be silent. So if a user has issues during testing they can at least share the error code on issues as semi-requested in #1870.


zm711:
Tested on an M1 Mac now. Mountainsort caused the pip install to fail with this error:
```
ERROR: Failed building wheel for isosplit5
ERROR: Could not build wheels for isosplit5, which is required to install pyproject.toml-based projects
```

So I commented that out for now in the Mac install. 
Running check on install led to the following:
```
OMP: Info #276: omp_set_nested routine deprecated, please use omp_set_max_active_levels instead.
write_binary_recording with n_jobs = 1 and chunk_size = 30000
write_binary_recording: 100%|##############################################################| 200/200 [00:02<00:00, 69.42it/s]
Import spikeinterface ...OK
Import spikeinterface.full ...OK
extract waveforms memmap: 100%|##############################################################| 77/77 [00:04<00:00, 17.26it/s]
Run tridesclous ...OK
Run herdingspikes ...Fail, Error: The sorter herdingspikes is not installed.Please install it with:  

To use HerdingSpikes run:

       >>> pip install herdingspikes
    More information on HerdingSpikes at:
      * https://github.com/mhhennig/hs2
     
Fitting PCA: 100%|█████████████████████████████████████████████████████████████████████████████| 9/9 [00:09<00:00,  1.02s/it]
Projecting waveforms: 100%|████████████████████████████████████████████████████████████████████| 9/9 [00:00<00:00, 77.34it/s]
Force compute_noise_levels() this is needed
qt.qpa.fonts: Populating font family aliases took 211 ms. Replace uses of missing font family "Menlo" with one that exists to avoid this cost. 
Open spikeinterface-gui ...OK
write_binary_recording: 100%|##############################################################| 200/200 [00:03<00:00, 59.66it/s]
Setting 'return_scaled' to False
extract amplitudes: 100%|################################################################| 200/200 [00:00<00:00, 5167.15it/s]
extract PCs: 100%|#########################################################################| 200/200 [00:03<00:00, 60.11it/s]
Export to phy ...OK
```

So I think adding in the exception actually provides pretty useful info for the user as can be seen by herding spikes failing in during the check install. Rather than do it silently it tells the user to pip install if they want it.

zm711:
I think it is a good idea. Maybe just a chron job? This gets out of sync too often.

alejoe91:
Yep!! A chron job running on all three environments sounds great :)

h-mayorquin:
Good. I will get to it.

h-mayorquin:
Maybe we can merge this so we can test the latest changes? This looks good to me as it is.

zm711:
I haven't tested linux, but both Mac (m1) and Windows (10) are confirmed to work with this code.

zm711:
Actually the check_your_install.py opens the spikeinterface-gui. I've never tried doing Github Actions with gui-s. Will that work?

h-mayorquin:
No, it won't work, that's a part that we will probably have to refactor and skip during CI.

zm711:
> @alejoe91 @h-mayorquin : Why not but this is very low priority, I think.
> We have million of stuff more urgent on the tests side. No ?
>
> I use this code when I prepare spike sorting workshop for students so we can quickly check theinstallation.

I agree that there are more urgent tests in general, but I've linked a few of the issues where people use the install tips as their way of creating a spikeinterface environment. Making sure that beginners first experience with spikeinterface is a good one (with their local 'test' passing) is a good way to keep them using the package. 


#1915, #1870, #1476, #1447

zm711:
I updated the linux yaml to release the version restrictions, but can't test locally to make sure it works. 

I finally hacked together a workflow that checks to make sure that the conda environment can be created on a chronic job. That way it'll be easier to add tests later, but at least for now we can see if the environment gets out of sync sooner rather than with a user issue. You want to take a peek at the yml @h-mayorquin (I always mess them up as you can see with the commit history)? It is pretty simple for now. That way a refactor/proper test can be pushed off to a later time for the `check_your_install.py`.

h-mayorquin:
@zm711 
Thanks for collecting all of those issues where the problem of the environment appears. This is very useful for reference.

I also think that having new users face as little difficulties as possible is very important, it will also make our life easier in the long-term when we switch from developing futures to support. In this case, the script that you wrote which should test that the conda environment is working correctly is a step in the right direction and hopefully -as you did it- will alleviate @samuelgarcia  concerns about priorities. 

Long term though I think we should do two things:
1) Run the full testing suite in the matrix of the all the operating systems.
2) Reference the [conda environment to the pip installation ](https://github.com/h-mayorquin/spikeinterface/blob/09394787e4f15c4268bc0ef8d9883e02163348ee/installation_tips/full_spikeinterface_environment_linux_dandi.yml#L38-L39)so we don't scatter around the specification of our dependencies (harder to track). That way, we can instruct users of other operating systems that are facing a bug to just use our testing suite.

alejoe91 (approved):

alejoe91:
Thanks so much @zm711 

I'll have a final round with @samuelgarcia on Wednesday to merge some PRs including this one! Thanks for the patience!

zm711:
No worries @alejoe91! I just didn't want to let things get out of sync, but I'll let you merge main into it whenever you are ready to merge it.

.github/workflows/installation-tips-test.yml
```diff
@@ -0,0 +1,33 @@
+name: Creates Conda Install for Installation Tips
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 12 * * *"  # Daily at noon UTC
```
alejoe91, line 6:
```suggestion
    - cron: "0 12 * * 0"  # Weekly at noon UTC on Sundays
```

alejoe91 (approved):

PR #1968, opened 2023-09-06
Refactor sorter launcher. Deprecated `run_sorters()` and add `run_sorter_jobs()`

### LINKED ISSUE #1969: Refactor `run_sorters()`
The `run_sorters()` requires more flexibility to handle a variety of jobs.

samuelgarcia:
* Deprecated `run_sorters()`
* Add `run_sorter_jobs()` which is more flexible.

Closes #1969


zm711 (commented):
one typo

doc/modules/sorters.rst
```diff
@@ -285,27 +285,26 @@ Running several sorters in parallel
...
-                                 mode_if_folder_exists='overwrite', sorter_params=sorter_params)
+    # here we run 2 sorters on 2 diffrents recording = 4 jobs
+    recording = ...
+    another_recording = ...
+
+    job_list = [
+      {'sorter_name': 'tridesclous', 'recording': recording, 'output_folder': '/folder1','detect_threshold': 5.},
+      {'sorter_name': 'tridesclous', 'recording': another_recording, 'output_folder': '/folder2', 'detect_threshold': 5.},
+      {'sorter_name': 'herdingspikes', 'recording': recording, 'output_folder': '/folder3', 'clustering_bandwidth': 8., 'docker_image': True},
+      {'sorter_name': 'herdingspikes', 'recording': another_recording, 'output_folder': '/folder4', 'clustering_bandwidth': 8., 'docker_image': True},
```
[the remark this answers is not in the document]
alejoe91: Then let's do `folder1/` or `path-to-folder1`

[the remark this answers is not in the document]
samuelgarcia: as you want this is just doc.

samuelgarcia:
@alejoe91 @h-mayorquin: ready for review on my side.

src/spikeinterface/sorters/launcher.py
```diff
@@ -4,69 +4,187 @@
...
+    dask=dict(client=None),
+    slurm=dict(tmp_script_folder=None, cpus_per_task=1, mem="1G"),
+)
+
+
+_implemented_engine = list(_default_engine_kwargs.keys())
+
+def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=False):
+    """
+    Run several :py:func:`run_sorter()` sequencially or in parralel given a list of job.
```
zm711, line 33:
```suggestion
    Run several :py:func:`run_sorter()` sequentially or in parallel given a list of jobs.
```

src/spikeinterface/sorters/launcher.py
```diff
@@ -4,69 +4,187 @@
...
+    For **engine="loop"** this is equivalent to:
+
+    ..code::
+
+        for job in job_list:
+            run_sorter(**job)
+    
+    For some engines, this function is blocking until the results ("loop", "joblib", "multiprocessing", "dask").
+    For some other engine ("slurm") the function return almost immediatly (akak non blocking) and the results
+    must be retrieve by hand when finished with :py:func:`read_sorter_folder()`.
```
zm711, line 44:
```suggestion
    For some engines ("loop", "joblib", "multiprocessing", "dask"), this function is blocking until the results .
    For other engines ("slurm") the function returns almost immediately (aka non-blocking) and the results
    must be retrieved by hand when finished with :py:func:`read_sorter_folder()`.
```

src/spikeinterface/sorters/launcher.py
```diff
@@ -4,69 +4,187 @@
...
+    Run several :py:func:`run_sorter()` sequencially or in parralel given a list of job.
+
+    For **engine="loop"** this is equivalent to:
+
+    ..code::
+
+        for job in job_list:
+            run_sorter(**job)
+    
+    For some engines, this function is blocking until the results ("loop", "joblib", "multiprocessing", "dask").
```
zm711, line 42: I'm also not sure the meaning of this line as written

zm711 (commented):
A few more typo fixes in the docstring and one clarification question.

src/spikeinterface/sorters/launcher.py
samuelgarcia: Oups. Any idea how reformulate this ?
blocking is when the function call wait until the result (asynchronous call)
oposite to asynchronous (return immediatly)

zm711: I see then I would just say: 

For some engines (...) this function is blocking until the results are returned.
For other engines (..) the function returns almost immediately (aka non-blocking), but the results must be retrieved by hand when the :py:func:`read_sorter_folder()` has finished.

Using the 'but' emphasizes that although the call is asynchronous it is up to the user to get the results when they are ready. I think the confusion is that the protected nature of the word `return` can be tricky to users not used to asyn vs syn programming.

h-mayorquin: My suggestions:
1) This should be added the Notes section as the numpy docstring convention:

https://numpydoc.readthedocs.io/en/latest/format.html#notes

2) As for the clarification, I think lists are better in case you want to add notes later. 
```pythoh
"""
The following engines block the I/O:
* Item1
* Item2
* Item3

Where *blocking* means that the results are {you description of what bocking is from the point of view of the user} . This is in opposition to *asynchronous* wher the {your description of what the antonym is from the point of view of the user}.
 """
```
This later formulation has the advantage that it separates the information of which items are blocking from the description / defintion that you give to help users who don't understand the term.  Advanced users can stop reading as they already know the following information but those who don't can keep reading after the list (this is the same convention that mathematical formulas follow where a description is added after a formula is presented). Plus, as @zm711 suggested you define something by what it is (giving a googlable term) and then also by what is not (negative definition) giving some much needed redundancy for first time readers.

src/spikeinterface/sorters/launcher.py
```diff
@@ -221,6 +344,13 @@ def run_sorters(
     results : dict
         The output is nested dict[(rec_name, sorter_name)] of SortingExtractor.
     """
+
+    warnings.warn(
+        "run_sorters()is deprecated please use run_sorter_jobs() instead",
```
h-mayorquin, line 349: Maybe adding the version here when this will be deprecated?

src/spikeinterface/sorters/launcher.py
zm711: That's a great strategy from @h-mayorquin. I think the issue is that since users will directly interface with this function the bilayer approach of giving the advanced person an off-ramp because they understand and the beginner person enough information so they can at least google later is super important.

src/spikeinterface/sorters/launcher.py
samuelgarcia: done

doc/modules/sorters.rst
```diff
@@ -285,27 +285,26 @@ Running several sorters in parallel
...
-                    'herdingspikes': {'clustering_bandwidth' : 8},
-                    'tridesclous': {'detect_threshold' : 5.},
-                    }
-    sorting_output = run_sorters(sorter_list, recordings, working_folder='tmp_some_sorters',
-                                 mode_if_folder_exists='overwrite', sorter_params=sorter_params)
+    # here we run 2 sorters on 2 different recordings = 4 jobs
+    recording = ...
+    another_recording = ...
+
+    job_list = [
```
h-mayorquin, line 297: I was going to suggest using the format that scikt-learn uses for exhaustive grid searh which should save you some typing:

https://scikit-learn.org/stable/modules/grid_search.html#exhaustive-grid-search

But maybe it does not work that well with some arguments like folder.

Nevertheless I suggest to take a look the strategies that they use. It is good not to reinvent the wheel and also to make use of the patterns and formats that our users already know (as scikit-learn is popular within the scientific comunity)

https://scikit-learn.org/stable/modules/grid_search.html#exhaustive-grid-search

samuelgarcia: I think we do not want any pattern here. We want to process a flat list of run_sorter() with deiffernt engine.
I would not add smantic of looping here and combinaison here.
What we implement here is the job dispatch. If someone want a job list creation that have grid combinason semantic we can do it but there is no need for now.

h-mayorquin: Yep, makes sense to me. Just adding this as relevant to the discussion for the future.

src/spikeinterface/sorters/launcher.py
```diff
@@ -4,69 +4,197 @@
...
-
-    if isinstance(recording, dict):
-        recording = load_extractor(recording)
+from .runsorter import run_sorter
+from .basesorter import is_log_ok
+
+_default_engine_kwargs = dict(
+    loop=dict(),
+    joblib=dict(n_jobs=-1, backend="loky"),
+    processpoolexecutor=dict(max_workers=2, mp_context=None),
```
alejoe91, line 23: Why 2 is default here?

src/spikeinterface/sorters/launcher.py
```diff
@@ -4,69 +4,197 @@
...
+
+    assert engine in _implemented_engine, f"engine must be in {_implemented_engine}"
+
+    engine_kwargs_ = dict()
+    engine_kwargs_.update(_default_engine_kwargs[engine])
+    engine_kwargs_.update(engine_kwargs)
+    engine_kwargs = engine_kwargs_
+
+    if return_output:
+        assert engine in ("loop", "joblib", "processpoolexecutor")
```
alejoe91, line 85: Assertion message missing
```suggestion
        assert engine in ("loop", "joblib", "processpoolexecutor"), "Only 'loop', 'joblib', and 'processpoolexecutor' support return_output=True."
```

src/spikeinterface/sorters/launcher.py
```diff
@@ -221,6 +353,13 @@ def run_sorters(
     results : dict
         The output is nested dict[(rec_name, sorter_name)] of SortingExtractor.
     """
+
+    warnings.warn(
+        "run_sorters()is deprecated please use run_sorter_jobs() instead. This will be removed in 0.100",
```
alejoe91, line 358:
```suggestion
        "run_sorters() is deprecated please use run_sorter_jobs() instead. This will be removed in 0.100",
```

src/spikeinterface/sorters/tests/test_launcher.py
```diff
@@ -31,19 +41,106 @@ def setup_module():
...
+def test_run_sorter_jobs_processpoolexecutor(job_list):
+    if base_output.is_dir():
+        shutil.rmtree(base_output)
+    sortings = run_sorter_jobs(
+        job_list, engine="processpoolexecutor", engine_kwargs=dict(max_workers=2), return_output=True
+    )
+    print(sortings)
+
+
+@pytest.mark.skipif(True, reason="This is tested locally")
```
alejoe91, line 91: dask should not be installed in tests. Why not having the usual `HAVE_DASK ...` and `skipif(not HAVE_DASK, reason="dask is not installed")`?

samuelgarcia: It is a big beast and dask is useless in our case without dask.distributed which also needs some slurm like infrastruture to be tested. So too heavy for me.

PR #1948, opened 2023-08-30
Refactor generate.py

samuelgarcia:
This big PR is a strong refactor of generate.py:
  * move InjectTemplatesRecording into generate.py
  * change GeneratorRecording into NoiseGeneratorRecording and improve it
  * move and refactor generate_templates() into generate.py
  * create generate_ground_truth_recording() to generate recording totally **lazy** + sorting 
  * rewrite toy_example() using generate_ground_truth_recording()

TODO add upsampling concept into InjectTemplatesRecording

A major improvement will be to use this for testing api, algos, metrics, visulisation without any disk space and memory use.

PR #1970, opened 2023-09-06
Some additions to generate.py after #1948

h-mayorquin:
Some additiosn after #1948.

* In `generate_recording`. We say that we will deprecate legacy in the next release but we don't have a deprecation warning. Next release is 0.99, so when this is released it should start throwing a deprecation warning, then we can deprecate it in 0.100. I added the deprecation warning. Plus, if we are going to deprecate we should make lazy the default in main so the code that we start suggestin people to use from main remains valid. Otherwise, they we will need to keep an argument with only one possible value (mode="lazy").
* I added the type directly in `rng.standard_normal` as this was generated an expensive computation when I profiled (got 30 % increase in performance for a one hour recorder with "on_the_fly")
* I also modified the default value for the standard deviation in the NoiseGenerator to 1. I think is a better [Schelling Point.](https://en.wikipedia.org/wiki/Focal_point_(game_theory)).
* I also modified the number of channels in `generate_recording_by_size` to be the same as SpikeGLX. I remember that at some point it caused me some problems in testing for it to have such an incovenient value (1024) which I think I set because it will play nice with round number of seconds at the beginning but it does not make a lot of sense in retrospective.

### DIFF
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -2,6 +2,7 @@ import math
 
 import numpy as np
 from typing import Union, Optional, List, Literal
+import warnings
 
 
 from .numpyextractors import NumpyRecording, NumpySorting
@@ -31,7 +32,7 @@ def generate_recording(
     set_probe: Optional[bool] = True,
     ndim: Optional[int] = 2,
     seed: Optional[int] = None,
-    mode: Literal["lazy", "legacy"] = "legacy",
+    mode: Literal["lazy", "legacy"] = "lazy",
 ) -> BaseRecording:
     """
     Generate a recording object.
@@ -51,10 +52,10 @@ def generate_recording(
         The number of dimensions of the probe, by default 2. Set to 3 to make 3 dimensional probes.
     seed : Optional[int]
         A seed for the np.ramdom.default_rng function
-    mode: str ["lazy", "legacy"] Default "legacy".
+    mode: str ["lazy", "legacy"] Default "lazy".
         "legacy": generate a NumpyRecording with white noise.
-                  This mode is kept for backward compatibility and will be deprecated in next release.
-        "lazy": return a NoiseGeneratorRecording
+                This mode is kept for backward compatibility and will be deprecated version 0.100.0.
+        "lazy": return a NoiseGeneratorRecording instance.
 
     Returns
     -------
@@ -64,6 +65,10 @@ def generate_recording(
     seed = _ensure_seed(seed)
 
     if mode == "legacy":
+        warnings.warn(
+            "generate_recording() : mode='legacy' will be deprecated in version 0.100.0. Use mode='lazy' instead.",
+            DeprecationWarning,
+        )
         recording = _generate_recording_legacy(num_channels, sampling_frequency, durations, seed)
     elif mode == "lazy":
         recording = NoiseGeneratorRecording(
@@ -460,7 +465,7 @@ def synthetize_spike_train_bad_isi(duration, baseline_rate, num_violations, viol
 
 class NoiseGeneratorRecording(BaseRecording):
     """
-    A lazy recording that generates random samples if and only if `get_traces` is called.
+    A lazy recording that generates white noise samples if and only if `get_traces` is called.
 
     This done by tiling small noise chunk.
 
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/generate.py
```diff
@@ -586,35 +591,35 @@ def get_traces(
...
         traces = np.empty(shape=(num_samples, self.num_channels), dtype=self.dtype)
 
-        start_block_index = start_frame // self.noise_block_size
-        end_block_index = end_frame // self.noise_block_size
+        first_block_index = start_frame // self.noise_block_size
+        last_block_index = end_frame // self.noise_block_size
 
         pos = 0
-        for block_index in range(start_block_index, end_block_index + 1):
+        for block_index in range(first_block_index, last_block_index + 1):
```
h-mayorquin, line 604: It was very strange that this version is faster than my vectorized one. Really surprised about this. That said, I realized that I had a some bad variable names which I have corrected here.

src/spikeinterface/core/generate.py
```diff
@@ -503,7 +508,7 @@ def __init__(
         num_channels: int,
         sampling_frequency: float,
         durations: List[float],
-        noise_level: float = 5.0,
+        noise_level: float = 1.0,
```
samuelgarcia, line 511: I think 5 uV is more realistic by default no ?

alejoe91: I agree with @samuelgarcia !

h-mayorquin: Color me skeptical aoubt a consensus in the size of noise for electrophysiology across areas, devices, species? (totally curios to be proven wrong). Do we want to make a claim like that?

That said, even if there was a consensus I think it would make more sense to put it in the downstream functions that are  mean to generate realistic recordings like `generate_ground_truth_recording` and keep this as a general generator that can be used to create different noise scenarios. For that, I think it makes more sense to keep this as the standard white noise instead of trying to make it realistically. For example, if useres wanted to generate pink noise I think it will be easier to do it from the base white noise instead of a scaled version of it.

samuelgarcia:
Merci beaucoup.

> Some additiosn after #1948.
> 
> * In `generate_recording`. We say that we will deprecate legacy in the next release but we don't have a deprecation warning. Next release is 0.99, so when this is released it should start throwing a deprecation warning, then we can deprecate it in 0.100. I added the deprecation warning. Plus, if we are going to deprecate we should make lazy the default in main so the code that we start suggestin people to use from main remains valid. Otherwise, they we will need to keep an argument with only one possible value (mode="lazy").

My plan was to do this in 2 steps first keep "legacy" by default and the switch to lazy. I am ok with lazy now.

> * I added the type directly in `rng.standard_normal` as this was generated an expensive computation when I profiled (got 30 % increase in performance for a one hour recorder with "on_the_fly")

Of course this avoid one amalooc! Really cool. thank you!

> * I also modified the default value for the standard deviation in the NoiseGenerator to 1. I think is a better [Schelling Point.](https://en.wikipedia.org/wiki/Focal_point_(game_theory)).

If we do this. I would like to still have 5 in generate_ground_truth_recording but I think this already the case.

> * I also modified the number of channels in `generate_recording_by_size` to be the same as SpikeGLX. I remember that at some point it caused me some problems in testing for it to have such an incovenient value (1024) which I think I set because it will play nice with round number of seconds at the beginning but it does not make a lot of sense in retrospective.

Not sure we have to mimic the neuropixel always. sometimes it is good to change number for testing.
But in fact, I do not care so much.

src/spikeinterface/core/generate.py
samuelgarcia: lets keep this the generate_ground_truth_recording have already 5

src/spikeinterface/core/generate.py
samuelgarcia: I think we can be even faster in we would use the "out" the rng.standard_normal but the code logic between the 2 modes could not be shared. So lets keep this.
For testing the default mode should be always "tile_pregenerated"

h-mayorquin: When I profiled it was showing the `np.concatenate` was the culprit of it being so slow. I will check out on what you say, I remember that I thought about it.

h-mayorquin:
> > * I also modified the default value for the standard deviation in the NoiseGenerator to 1. I think is a better [Schelling Point.](https://en.wikipedia.org/wiki/Focal_point_(game_theory)).
> 
> If we do this. I would like to still have 5 in generate_ground_truth_recording but I think this already the case.

Great. I think this makes sense.

> > * I also modified the number of channels in `generate_recording_by_size` to be the same as SpikeGLX. I remember that at some point it caused me some problems in testing for it to have such an incovenient value (1024) which I think I set because it will play nice with round number of seconds at the beginning but it does not make a lot of sense in retrospective.
> 
> Not sure we have to mimic the neuropixel always. sometimes it is good to change number for testing. But in fact, I do not care so much.

Yeah, I agree with you, it is just a better default than the one I had before which I chosen because of numerology (e.g. 1024 looks nice)!

PR #1437, opened 2023-03-21
Improve whitening and zscore

samuelgarcia:
Add a int_scale params or zscore and whietning.
Preserve gain/offset for zscore in parallel processing

PR #1971, opened 2023-09-06
Pin down z-scale normalized to integer (!) failure by separating the tests

h-mayorquin:
I have this tests failing in #1970 which I have seen failing before. It is not failing in my computer so I want to separate it in its own test to hopefully isolate the failure.

### DIFF
--- a/src/spikeinterface/preprocessing/normalize_scale.py
+++ b/src/spikeinterface/preprocessing/normalize_scale.py
@@ -293,7 +293,7 @@ class ZScoreRecording(BasePreprocessor):
             means = means[None, :]
             stds = np.std(random_data, axis=0)
             stds = stds[None, :]
-            gain = 1 / stds
+            gain = 1.0 / stds
             offset = -means / stds
 
         if int_scale is not None:
--- a/src/spikeinterface/preprocessing/tests/test_normalize_scale.py
+++ b/src/spikeinterface/preprocessing/tests/test_normalize_scale.py
@@ -80,8 +80,8 @@ def test_zscore():
 
 
 def test_zscore_int():
-    seed = 0
-    rec = generate_recording(seed=seed, mode="lazy")
+    seed = 1
+    rec = generate_recording(seed=seed, mode="legacy")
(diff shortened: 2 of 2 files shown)


h-mayorquin:
OK, I pined down an example of failure.

```python
from spikeinterface.preprocessing import normalize_by_quantile, scale, center, zscore

seed = 1
rec = generate_recording(seed=seed, mode="lazy")
rec_int = scale(rec, dtype="int16", gain=100)

zscore_recording = zscore(rec_int, dtype="int16", int_scale=256, mode="mean+std", seed=seed)
traces = zscore_recording.get_traces(segment_index=0)
trace_mean = np.mean(traces, axis=0)
trace_std = np.std(traces, axis=0)
assert np.all(np.abs(trace_mean) < 1)
assert np.all(np.abs(trace_std - 256) < 1)
```

I am  confused about this:

https://github.com/SpikeInterface/spikeinterface/pull/1437

src/spikeinterface/preprocessing/normalize_scale.py
```diff
@@ -293,7 +293,7 @@ def __init__(
             means = means[None, :]
             stds = np.std(random_data, axis=0)
             stds = stds[None, :]
-            gain = 1 / stds
+            gain = 1.0 / stds
```
samuelgarcia, line 296: good catch!

samuelgarcia:
`assert np.all(np.abs(trace_std - 256) < 1)` was too strong maybe 2% tolrance error with this
`assert np.all(np.abs(trace_std - 256) < 5)` should be better

samuelgarcia:
OK for me.
@alejoe91  do you want to check ?

h-mayorquin:
> `assert np.all(np.abs(trace_std - 256) < 1)` was too strong maybe 2% tolrance error with this `assert np.all(np.abs(trace_std - 256) < 5)` should be better

In my simulations the std is fine, is the mean estimate that has a lot of variability.

alejoe91 (approved):

PR #1979, opened 2023-09-11
Fix `zero_channel_pad` case when selected frames are outside of original data region

JoeZiminski:
I was running into an error when using Kilosort's `skip_kilosort_preprocessing`  that would crash when running preprocessing during binary-writing prior to sorting.

The (test) file I was running had a period at the end of the trace which was required to be purely padding (for running in kilosort), and was larger than the chunk size. This meant that calls were made to `zero_channel_pad.get_traces()` which had both a start and end frame larger than the original data size. This led to the below function call:

```
            original_traces = self.get_original_traces_shifted(
                start_frame=start_frame,
                end_frame=end_frame,
                channel_indices=channel_indices,
            )
```

with `start_frame` and `end_frame` that were larger than the original data. 

As a fix, this PR adds another case in which both the `start_frame` and `end_frame` are larger than the data - if this is `True`, then an array that consists of only padding is returned.

I must admit I am not very familiar with the uses of `zero_channel_pad.py` and have not considered deeply the effect of adding this code to other uses of the function beyond fixing the problem I was having. However, I think in the case that:

```
start_frame > self.num_samples_in_original_segment and end_frame > self.num_samples_in_original_segment
```
There is nothing else that you could possibly need to do except return an array full of padding? And this is quite a rare and unique circumstance, so I don't think this conditional will erroneously catch other cases.

### DIFF
--- a/src/spikeinterface/preprocessing/zero_channel_pad.py
+++ b/src/spikeinterface/preprocessing/zero_channel_pad.py
@@ -94,6 +94,10 @@ class TracePaddedRecordingSegment(BasePreprocessorSegment):
         # Else, we start with the full padded traces and allocate the original traces in the middle
         output_traces = np.full(shape=(trace_size, num_channels), fill_value=self.fill_value, dtype=self.dtype)
 
+        # If start and end frame are outside of the original data region (e.g. for Kilosort), return only paddding
+        if start_frame > self.num_samples_in_original_segment and end_frame > self.num_samples_in_original_segment:
+            return output_traces
+
         # After the padding, the original traces are placed in the middle until the end of the original traces
         if end_frame >= self.padding_start:
             original_traces = self.get_original_traces_shifted(
--- a/src/spikeinterface/preprocessing/zero_channel_pad.py
+++ b/src/spikeinterface/preprocessing/zero_channel_pad.py
@@ -95,7 +95,7 @@ class TracePaddedRecordingSegment(BasePreprocessorSegment):
         output_traces = np.full(shape=(trace_size, num_channels), fill_value=self.fill_value, dtype=self.dtype)
 
         # If start and end frame are outside of the original data region (e.g. for Kilosort), return only paddding
-        if start_frame > self.num_samples_in_original_segment and end_frame > self.num_samples_in_original_segment:
+        if start_frame > self.num_samples_in_original_segment + self.padding_start and end_frame > self.num_samples_in_original_segment + self.padding_start:
             return output_traces
 
         # After the padding, the original traces are placed in the middle until the end of the original traces
@@ -123,11 +123,8 @@ class TracePaddedRecordingSegment(BasePreprocessorSegment):
         """
         original_start_frame = max(start_frame - self.padding_start, 0)
         original_end_frame = min(end_frame - self.padding_start, self.num_samples_in_original_segment)
-        original_traces = self.parent_recording_segment.get_traces(
-            start_frame=original_start_frame,
-            end_frame=original_end_frame,
-            channel_indices=channel_indices,
-        )
+
+        original_traces = self.parent_recording_segment.get_traces(start_frame=original_start_frame, end_frame=original_end_frame, channel_indices=channel_indices,)  # BREAKPOINT HERE!!
 
         return original_traces
 
--- a/src/spikeinterface/preprocessing/zero_channel_pad.py
+++ b/src/spikeinterface/preprocessing/zero_channel_pad.py
@@ -95,7 +95,10 @@ class TracePaddedRecordingSegment(BasePreprocessorSegment):
         output_traces = np.full(shape=(trace_size, num_channels), fill_value=self.fill_value, dtype=self.dtype)
 
         # If start and end frame are outside of the original data region (e.g. for Kilosort), return only paddding
-        if start_frame > self.num_samples_in_original_segment + self.padding_start and end_frame > self.num_samples_in_original_segment + self.padding_start:
+        if (
+            start_frame > self.num_samples_in_original_segment + self.padding_start
+            and end_frame > self.num_samples_in_original_segment + self.padding_start
+        ):
             return output_traces
 
         # After the padding, the original traces are placed in the middle until the end of the original traces
@@ -124,7 +127,11 @@ class TracePaddedRecordingSegment(BasePreprocessorSegment):
         original_start_frame = max(start_frame - self.padding_start, 0)
         original_end_frame = min(end_frame - self.padding_start, self.num_samples_in_original_segment)
 
-        original_traces = self.parent_recording_segment.get_traces(start_frame=original_start_frame, end_frame=original_end_frame, channel_indices=channel_indices,)  # BREAKPOINT HERE!!
+        original_traces = self.parent_recording_segment.get_traces(
+            start_frame=original_start_frame,
+            end_frame=original_end_frame,
+            channel_indices=channel_indices,
+        )  # BREAKPOINT HERE!!
 
         return original_traces
 
--- a/src/spikeinterface/preprocessing/tests/test_zero_padding.py
+++ b/src/spikeinterface/preprocessing/tests/test_zero_padding.py
@@ -6,7 +6,7 @@ from spikeinterface import set_global_tmp_folder
 from spikeinterface.core import generate_recording
 from spikeinterface.core.numpyextractors import NumpyRecording
 
-from spikeinterface.preprocessing import zero_channel_pad
+from spikeinterface.preprocessing import zero_channel_pad, bandpass_filter, common_reference
 from spikeinterface.preprocessing.zero_channel_pad import TracePaddedRecording
 
 if hasattr(pytest, "global_test_folder"):
@@ -39,7 +39,7 @@ def test_zero_padding_channel():
 @pytest.fixture
 def recording():
     num_channels = 4
-    num_samples = 10
+    num_samples = 10000
     rng = np.random.default_rng(seed=0)
     traces = rng.random(size=(num_samples, num_channels))
     traces_list = [traces]
@@ -256,6 +256,77 @@ def test_trace_padded_recording_retrieve_traces_with_partial_padding(recording, 
     padded_traces_end = padded_traces[frame_where_original_trace_ends:, :]
     expected_zeros = np.zeros((number_of_paded_frames_at_end, num_channels))
     assert np.allclose(padded_traces_end, expected_zeros)
+
+@pytest.mark.parametrize("padding_start, padding_end", [(5000, 5000), (5000, 0), (0, 5000)])
+def test_trace_padded_recording_retrieve_full_recording_with_preprocessing(recording, padding_start, padding_end):
+
+    num_samples = recording.get_num_samples()
+    num_channels = recording.get_num_channels()
+
+    recording = bandpass_filter(recording, freq_min=300, freq_max=6000)
+
+    padded_recording = TracePaddedRecording(
+        parent_recording=recording,
+        padding_start=padding_start,
+        padding_end=padding_end,
+    )
+
+    # Cycle through the whole recording, using get_traces() to pull chunks of
+    # size `step`. This emulates the processing of writing to a binary file.
+    # Data that lie within the padding region should be fill value only, while
+    # data from original trace should match exactly. Note that the step
+    # size must be chosen to retreieve data that is purely padding or original data
+    step = 1000
+    start_frames = np.arange(padded_recording.get_num_samples(), step=step)
+    end_frames = start_frames + step
+
+    for start_frame, end_frame in zip(start_frames, end_frames):
+
+        padded_trace = padded_recording.get_traces(start_frame=start_frame, end_frame=end_frame)
+
+        end_padding_region_first_idx = padding_start + num_samples
+
+        if padding_start <= start_frame < end_padding_region_first_idx:
+            original_trace = recording.get_traces(start_frame=start_frame - padding_start, end_frame=end_frame - padding_start)
+            assert np.allclose(padded_trace, original_trace, rtol=0, atol=1e-10)
+        else:
+            assert np.all(padded_trace == padded_recording.fill_value)
+
+
+@pytest.mark.parametrize("padding_start, padding_end", [(5000, 5000), (5000, 0), (0, 5000)])
+def test_trace_padded_recording_retrieve_full_recording_with_preprocessing(recording, padding_start, padding_end):
+
+    num_samples = recording.get_num_samples()
+    num_channels = recording.get_num_channels()
+
+    recording = bandpass_filter(recording, freq_min=300, freq_max=6000)
+
+    padded_recording = TracePaddedRecording(
+        parent_recording=recording,
+        padding_start=padding_start,
+        padding_end=padding_end,
+    )
+
+    # Cycle through the whole recording, using get_traces() to pull chunks of
+    # size `step`. This emulates the processing of writing to a binary file.
+    # Data that lie within the padding region should be fill value only, while
+    # data from original trace should match exactly. Note that the step
+    # size must be chosen to retreieve data that is purely padding or original data
+    step = 1000
+    start_frames = np.arange(padded_recording.get_num_samples(), step=step)
+    end_frames = start_frames + step
+
+    for start_frame, end_frame in zip(start_frames, end_frames):
+
+        padded_trace = padded_recording.get_traces(start_frame=start_frame, end_frame=end_frame)
+
+        end_padding_region_first_idx = padding_start + num_samples
+
+        if padding_start <= start_frame < end_padding_region_first_idx:
+            original_trace = recording.get_traces(start_frame=start_frame - padding_start, end_frame=end_frame - padding_start)
+            assert np.allclose(padded_trace, original_trace, rtol=0, atol=1e-10)
+        else:
+            assert np.all(padded_trace == padded_recording.fill_value)
 
 
 if __name__ == "__main__":
--- a/src/spikeinterface/preprocessing/tests/test_zero_padding.py
+++ b/src/spikeinterface/preprocessing/tests/test_zero_padding.py
@@ -257,9 +257,9 @@ def test_trace_padded_recording_retrieve_traces_with_partial_padding(recording, 
     expected_zeros = np.zeros((number_of_paded_frames_at_end, num_channels))
     assert np.allclose(padded_traces_end, expected_zeros)
 
+
 @pytest.mark.parametrize("padding_start, padding_end", [(5000, 5000), (5000, 0), (0, 5000)])
 def test_trace_padded_recording_retrieve_full_recording_with_preprocessing(recording, padding_start, padding_end):
-
     num_samples = recording.get_num_samples()
     num_channels = recording.get_num_channels()
 
@@ -281,13 +281,14 @@ def test_trace_padded_recording_retrieve_full_recording_with_preprocessing(recor
     end_frames = start_frames + step
 
     for start_frame, end_frame in zip(start_frames, end_frames):
-
         padded_trace = padded_recording.get_traces(start_frame=start_frame, end_frame=end_frame)
 
         end_padding_region_first_idx = padding_start + num_samples
 
         if padding_start <= start_frame < end_padding_region_first_idx:
-            original_trace = recording.get_traces(start_frame=start_frame - padding_start, end_frame=end_frame - padding_start)
+            original_trace = recording.get_traces(
+                start_frame=start_frame - padding_start, end_frame=end_frame - padding_start
+            )
             assert np.allclose(padded_trace, original_trace, rtol=0, atol=1e-10)
         else:
             assert np.all(padded_trace == padded_recording.fill_value)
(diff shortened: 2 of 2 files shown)


h-mayorquin:
What is the error that you are getting?
Can you add a test here:

https://github.com/SpikeInterface/spikeinterface/blob/9fe8efe0941c90c7fb3e46cf67f1aa710f5cd292/src/spikeinterface/preprocessing/tests/test_zero_padding.py

If this fails it indicates that there is a problem with this block as well so I would like to fix that.
https://github.com/SpikeInterface/spikeinterface/blob/9fe8efe0941c90c7fb3e46cf67f1aa710f5cd292/src/spikeinterface/preprocessing/zero_channel_pad.py#L102-L111

h-mayorquin:
Or alternatively just generate a minimal example to reproduce the error and I will add the tests.

JoeZiminski:
Hi @h-mayorquin thanks for this, completely agree this needs a test prior to merge, in general that's a good policy to have (that no PR is merged without a corresponding test of new feature). In this case I overlooked this!  As an aside, more generally, it might be nice to have a PR pull request template for spikeinterface (e.g. [example](https://axolo.co/blog/p/part-3-github-pull-request-template)) to serve as a checklist of such things, and help with documentation of changes. If you think this is worthwhile I will open an issue.

I include the stacktrace and code used to run, but will add a proper test ASAP. In this case the code depends on the file which is 75000 samples long and requires 56200 samples padding at the end for kilosort. The specific error dependends on the preprocessing chain, but in general is some variant of having an empty array in a `get_traces()` call due to the `start_frame` being less than the `end_frame`.

I beleive the fundamental problem is that [this line](https://github.com/SpikeInterface/spikeinterface/blob/9fe8efe0941c90c7fb3e46cf67f1aa710f5cd292/src/spikeinterface/preprocessing/zero_channel_pad.py#L103) is triggered. In this case, the `self.padding_start=0`, `self.padding_end=56200`, `self.num_samples_in_original_segment=75000`, and the expected num samples in the file is 131200. It fails when the `start_frame=90000` and `end_frame=120000`. Therefore this conditional is triggered almost as a default, because the case in which there is no data within `start_frame`, `end_frame` is not caught. Then, [this line](https://github.com/SpikeInterface/spikeinterface/blob/9fe8efe0941c90c7fb3e46cf67f1aa710f5cd292/src/spikeinterface/preprocessing/zero_channel_pad.py#L125C1-L125C9) in `get_original_traces_shifted()` sets `original_end_frame` to `self.num_samples_in_original_segment=75000` so the `get_traces()` is called with a `start_frame=90000` and `end_frame=75000`.

```
spikeinterface/src/spikeinterface/sorters/runsorter.py:148: in run_sorter
spikeinterface/src/spikeinterface/sorters/runsorter.py:173: in run_sorter_local
    SorterClass.setup_recording(recording, output_folder, verbose=verbose)
spikeinterface/src/spikeinterface/sorters/basesorter.py:203: in setup_recording
    cls._setup_recording(recording, sorter_output_folder, sorter_params, verbose)
spikeinterface/src/spikeinterface/sorters/external/kilosortbase.py:156: in _setup_recording
    write_binary_recording(
spikeinterface/src/spikeinterface/core/core_tools.py:314: in write_binary_recording
    executor.run()
spikeinterface/src/spikeinterface/core/job_tools.py:376: in run
    res = self.func(segment_index, frame_start, frame_stop, worker_ctx)
spikeinterface/src/spikeinterface/core/core_tools.py:234: in _write_binary_chunk
    traces = recording.get_traces(start_frame=start_frame, end_frame=end_frame, segment_index=segment_index, cast_unsigned=cast_unsigned)
spikeinterface/src/spikeinterface/core/baserecording.py:284: in get_traces
    traces = rs.get_traces(start_frame=start_frame, end_frame=end_frame, channel_indices=channel_indices)
spikeinterface/src/spikeinterface/preprocessing/zero_channel_pad.py:103: in get_traces
    original_traces = self.get_original_traces_shifted(start_frame=start_frame, end_frame=end_frame, channel_indices=channel_indices,)
spikeinterface/src/spikeinterface/preprocessing/zero_channel_pad.py:125: in get_original_traces_shifted
    original_traces = self.parent_recording_segment.get_traces(
spikeinterface/src/spikeinterface/preprocessing/filter.py:150: in get_traces
    filtered_traces = scipy.signal.sosfiltfilt(self.coeff, traces_chunk, axis=0)
/nfs/nhome/live/jziminski/.conda/envs/spikewrap/lib/python3.10/site-packages/scipy/signal/_signaltools.py:4440: in sosfiltfilt
    edge, ext = _validate_pad(padtype, padlen, x, axis,
/nfs/nhome/live/jziminski/.conda/envs/spikewrap/lib/python3.10/site-packages/scipy/signal/_signaltools.py:4209: in _validate_pad
    raise ValueError("The length of the input vector x must be greater "
E   ValueError: The length of the input vector x must be greater than padlen, which is 33.
```

JoeZiminski:
Hey @h-mayorquin I have pushed a new test case (note, this increases the `num_samples` parameter to 10,000 to get this to work, discussed below. This isn't meant to be a final-version on the tests as I think it could be done better (e.g. as this now means all other tests are running at 10k samples when they don't need to), but just to highlight the problem. (sorry, this post got longer than anticipated).

The error is only observed with certain preprocessing steps. Without (e.g. commenting #recording = bandpass_filter(recording, freq_min=300, freq_max=6000)` in the new test), and putting a breakpoint at [this line](https://github.com/SpikeInterface/spikeinterface/blob/31f679f607b19012b32724e405ad4c8bdd4b5a71/src/spikeinterface/preprocessing/zero_channel_pad.py#L99), in some cases the call to original traces returns an empty array, however this is filled into an array of the fill value. I think then works, but is not the intended use case of this code block. For example, if `start_frame=4000`, `end_frame=5000`, and `self.padding_start=5000` then `end_frame >= self.padding_start` is `True`  and original traces are retrieved, returning an empty trace. However, the rest of the block fills it in with the fill value. This is also true for padding after the data, e.g.  if the data has 10000 samples, with 5000 padding at the start and end. Then `start_frame=10000`, `end_frame=11000` will trigged this block, and again an empty array is returned from `get_original_traced_shifted()`. But it does not matter as the array is filled by the rest of that code block.

If you but a breakpoint in [`get_original_traces_shifted()`](https://github.com/SpikeInterface/spikeinterface/blob/31f679f607b19012b32724e405ad4c8bdd4b5a71/src/spikeinterface/preprocessing/zero_channel_pad.py#L111), choosing a `start_frame` that is larger than the original data (e.g. `if start_frame == 16000: breakpoint()` in the current test case, without the implemented fix you see some strange behaviour. For example, `self.parent_recording_segment.get_traces` is called with `original_start_frame` that is larger than `original_end_frame`. However, this is silent because the it returns an empty array which is filled in as above. (This is a discussion for another day, but I think `get_traces()` should error when nonsense indicies are passed, to stop silent strange behaviour such as this).

However, if preprocessing is run (e.g. filtering), then `get_traces()` calls the parent `get_traces()`. Specifically, then the behaviour depenends on [this call in filter.py](https://github.com/SpikeInterface/spikeinterface/blob/31f679f607b19012b32724e405ad4c8bdd4b5a71/src/spikeinterface/preprocessing/filter.py#L130). As far as I can tell, this function will start return data of the length `self.margin` even if `end_frame` < `start_frame`. However, if `end_frame` is many samples before `start_frame`sif the number of samples in the chunk is smaller than the margin, then `get_chunk_with_margin()` will still return some data. However, if the chunk size is much bigger than the margin (I can't find an exact cutoff) then it will return an empty trace.

For example, calling:

```
start_frame=1000
end_frame=999
traces_chunk, left_margin, right_margin = get_chunk_with_margin(self.parent_recording_segment, start_frame, end_frame, channel_indices, self.margin, add_reflect_padding=False,)

traces_chunk.shape # (299, 4)
# with end_frame=850, traces_chunk.shape = (150, 4)
# end_frame=500, traces_chunk.shape = (0, 4)
```

Thus, when filtering is then done in the case of an empty array, there is a crash. This only manifests when `end_frame` is quite a lot smaller that `start_frame` as compared to the margin. Therefore, I had to increase the samples on the test case to see the error. I think it would also make sense to do some indices checks in `get_chunk_with_margin()` along with `get_traces`.

Anyway, so the bug only manifests when:
* certain preprocessing steps (e.g. filtering) are performed.
* a `get_traces()` call is made on data that is completely in the end-padding (i.e. after the original data) region.
* the chunk size is large compared to the margin used in `get_chunk_with_margin`.

I think this padding problem is a difficult one to solve as there are a lot of potential cases. The full set of possible cases (I think) are, considering a trace with padding before and after the original data:
`1) retrieve all-padding region from the first padding section.
2) retrieve some padding, some data from the first padding / data section.
3) retrieve entirely data section.
4) retrieve some data, some padding from the end padding section.
5) retrieve all padding from the end padding section.

I think it might make sense to explicitly handle each case in the `zero_channel_pad` `get_traces()` call. Also, if you agree, it might be worth raising an issue to add checks on `get_traces()` and `get_chunk_with_margin()` to ensure nonsense indicies are caught.

src/spikeinterface/preprocessing/zero_channel_pad.py
```diff
@@ -119,11 +126,12 @@ def get_original_traces_shifted(self, start_frame, end_frame, channel_indices):
         """
         original_start_frame = max(start_frame - self.padding_start, 0)
```
h-mayorquin, line 127:
```suggestion
        original_start_frame = min(max(start_frame - self.padding_start, 0) self.num_samples_in_original_segment), 
```

h-mayorquin: As the `original_end_frame` is capped at `self.num_samples_in_original_segment`, the same cap should be applied to `original_start_frame`. 

Another alternative, is to remove the two caps.

h-mayorquin (commented):
Single comment to highlight an error in my code.

h-mayorquin:
@JoeZiminski 
Thanks a lot. The example that you added as a test helped me pin down this. Let's discuss to see if we agree. First, I pointed an error in the implementation here. It was not detect in my testing because the recording that I used for testing behaves like numpy for slicing. I will clarify what I mean below.

Now, I think this has uncover what I think is the deep source of the error:
I like to think that the source of this error is in `bandpass_filter`:

https://github.com/JoeZiminski/spikeinterface/blob/dce34a8f48592fdf0085e3998d22e2506f62386e/src/spikeinterface/preprocessing/filter.py#L130-L147

if the chunk margin is np.array([]) then there should be an early return that returns np.array([]). Instead you get the error that you received. 

The reason for this is that recorders should behave like numpy array with respect to slicing. In numpy you have the following behavior:

```python

In [1]: x = np.ones(10)[15:13]

In [2]: x[15:13]
Out[2]: array([], dtype=float64)

In [3]: x[5:3]
Out[3]: array([], dtype=float64)

In [4]: x[10:10]
Out[4]: array([], dtype=float64)
```

But look what happens with the bandpass filter:

```python
from spikeinterface.core.generate import generate_recording, generate_sorting
from spikeinterface.preprocessing import bandpass_filter


recording = generate_recording()
recording_filtered = bandpass_filter(recording, freq_min=300, freq_max=6000)

recording_filtered.get_traces(start_frame=1300, end_frame=1000, segment_index=0)

File [~/miniconda3/envs/spikeinterface_env/lib/python3.10/site-packages/scipy/signal/_signaltools.py:4411](https://file+.vscode-resource.vscode-cdn.net/home/heberto/development/spikeinterface/bin/~/miniconda3/envs/spikeinterface_env/lib/python3.10/site-packages/scipy/signal/_signaltools.py:4411), in sosfiltfilt(sos, x, axis, padtype, padlen)
   4409 ntaps = 2 * n_sections + 1
   4410 ntaps -= min((sos[:, 2] == 0).sum(), (sos[:, 5] == 0).sum())
-> 4411 edge, ext = _validate_pad(padtype, padlen, x, axis,
   4412                           ntaps=ntaps)
   4414 # These steps follow the same form as filtfilt with modifications
   4415 zi = sosfilt_zi(sos)  # shape (n_sections, 2) --> (n_sections, ..., 2, ...)

File [~/miniconda3/envs/spikeinterface_env/lib/python3.10/site-packages/scipy/signal/_signaltools.py:4180](https://file+.vscode-resource.vscode-cdn.net/home/heberto/development/spikeinterface/bin/~/miniconda3/envs/spikeinterface_env/lib/python3.10/site-packages/scipy/signal/_signaltools.py:4180), in _validate_pad(padtype, padlen, x, axis, ntaps)
   4178 # x's 'axis' dimension must be bigger than edge.
   4179 if x.shape[axis] <= edge:
-> 4180     raise ValueError("The length of the input vector x must be greater "
   4181                      "than padlen, which is %d." % edge)
   4183 if padtype is not None and edge > 0:
   4184     # Make an extension of length `edge` at each
   4185     # end of the input array.
   4186     if padtype == 'even':

ValueError: The length of the input vector x must be greater than padlen, which is 33.

```

This is the assertion that you were seeing.

h-mayorquin:
@JoeZiminski 
So to summarize, my beliefs right now. Let me know if you disagree and where so we can move this forward.  I think you got a good summary of the test cases above

> retrieve all-padding region from the first padding section.
retrieve some padding, some data from the first padding / data section.
retrieve entirely data section.
retrieve some data, some padding from the end padding section.
retrieve all padding from the end padding section.

If you look at the test that I added previously they tried to cover exactly this variety of cases. The problem is that they did it with a recorder with that follows numpy slicing convention and the `bandpass_filter` does not. 

So, specific points:

* The code of `TracePaddedRecording` works as expected as a long as the original recording that is padded follows the numpy slicing convention.
* There is an internal error when padding at the end of the original recording where the start_frame and the end_frame are inverted  https://github.com/SpikeInterface/spikeinterface/pull/1979#discussion_r1323333927. This however, should still work as long as the original recording follows the numpy convention (e.g. `array[10:10] == array[12:10]` if you `array.size == 10`).
* `bandpass_filter` get_traces should have the slicing behavior of numpy and right now it does not.

src/spikeinterface/preprocessing/zero_channel_pad.py
h-mayorquin: See here:

https://github.com/SpikeInterface/spikeinterface/pull/1984

Where the tests are passed after this modification.

JoeZiminski:
[Currently Editing]

Thanks @h-mayorquin! That makes a lot of sense, I can see that the error is resolved if the numpy-style slicing convention is maintained for recording object. I also saw this error for `phase_shift` (sorry we have a big meeting today but I can write a test this afternoon) so I think this would need to be made slice-like as well.

I think it would make sense to handle explicitly the cases in which we are in a padding-only region (i.e. in the pre-data padding and post-data padding regions) so no call is made to `get_original_traces_shifted()` unless data is definately required to be retrieved. This should make the intention behind the code clearer and this also does not rely on recording objects acting like numpy arrays when slicing. This would just be something like (pseudocode, and need to check whether stict inequalities are used):

```
end_padding_begin_idx =  original_data_size + padding_start

if (start_frame > end_padding_start and end_frame > end_padding_begin_idx )  or
    (start_frame < padding_start and end_frame < padding_start):
    return output_traces
```
That said, it would be good to write a test to ensure all recording objects act consistently when slides out of bounds / with `end_frame < start_frame`. I can write an issue for this and maybe we can open a PR adding tests for this down the line.

The next question is out of scope for this PR and concerns a major design choice of SI so feel free to ignore. But, do you know the benefits of recording objects slicing like numpy arrays? Although it adds flexibility, it is difficult to determine the cases when such slicing is intended vs. when it is a mistake and should be flagged. For example, I cannot off the top of my head think of a reason to index the data with `end_frame > data_size` or `end_frame < start_frame`. I think attempting to do this in most cases (e.g. for non-expert users) would be an error rather than intended.

I will respond to other comments ASAP, thanks!

h-mayorquin:
> I think it would make sense to handle explicitly the cases in which we are in a padding-only region (i.e. in the pre-data padding and post-data padding regions) so no call is made to `get_original_traces_shifted()` unless data is definately required to be retrieved. This should make the intention behind the code clearer and this also does not rely on recording objects acting like numpy arrays when slicing. This would just be something like (pseudocode, and need to check whether stict inequalities are used):
> 

Mmm yeah, I kind of wanted to avoid that because I though the following code is already handling that case:

https://github.com/SpikeInterface/spikeinterface/blob/b24029810bede1b5043810442abe7bc8364bf0fd/src/spikeinterface/preprocessing/zero_channel_pad.py#L96-L107

But that is implicit...  and I think you are right that "explicit is better than implicit". I will check some performance and and I will probably move to your suggestion.

> 
> That said, it would be good to write a test to ensure all recording objects act consistently when slides out of bounds / with `end_frame < start_frame`. I can write an issue for this and maybe we can open a PR adding tests for this down the line.
> 
> In terms of testing this, I guess adding pre-processing to the tests in `test_zero_channel_pad.py` might not be the right place, if this is mainly for unit-tests. Is there an integration test suite this could be added to? For example, it might be worth testing such preprocessing steps parameterised over various combinations and settings.

> The next question is out of scope for this PR and concerns a major design choice of SI so feel free to ignore. But, do you know the benefits of recording objects slicing like numpy arrays? Although it adds flexibility, it is difficult to determine the cases when such slicing is intended vs. when it is a mistake and should be flagged. For example, I cannot off the top of my head think of a reason to index the data with `end_frame > data_size` or `end_frame < start_frame`. I think attempting to do this in most cases (e.g. for non-expert users) would be an error rather than intended.


We wiill be discussing this issues here:
https://github.com/SpikeInterface/spikeinterface/issues/1989

JoeZiminski:
This is a key question with no clear best-answer. I think the decision depends in part on the broader aims of spikeinterface and the intended audience for which it will be used. On one extreme, there is a highly restrictive API which does not allow any indexing procedure that could be interpreted as a mistake. i.e. indexing the recording out of bounds, or with `start_frame >= end_frame`.  On the other extreme is a very flexible API with few checks that could allow unintended operations to pass silently. 

The benefits of a more restrictive API is 1) easier to track errors and detect bugs 2) enforces very explicit handling of all use cases 3) easier for non-expert users to detect mistakes in their code. The downside is that it increases initial overhead when writing code, and breaks with existing data-slicing API, and is less flexible.

At present,  I lean towards a highly restrictive API that will throw an error if any potentially uninded indexing operations are performed. Although this might lead to some initial overhead to explicitly handle all indexing cases, I think it will pay dividends down the line with increase ease of bug-fixing and reduced number of user-reported issues. 

Also, I think that SI should aim to be the de-facto tool for _in vivo_ electrophysiology analysis in Python, inevitably drawing novice Python users with it's easy-to-use API. Often in practice those analysing electrophysiological data little computational training. Having robust checks on common indexing mistakes will make SI more secure for such uses and save a lot of headaches for new users.

Fundamentally, I think the problem of allowing such indexing operations is it is impossible to tell whether this indexing was done on purpose, or is due to a mistake or oversight. The possible options then are to allow both intentional and unintentional indexing of this sort (i.e. flexible API), or not allow either (restrictve API). I think in general it is safer to handle the intended cases at a higher level that `get_traces()`, and catch all uninteded cases, rather than let both intended and unintended cases pass.

src/spikeinterface/preprocessing/zero_channel_pad.py
JoeZiminski: I think this looks good! I believe there are 7 cases to handle here (which are captured already in your tests), and involve combinations of  data with (padding start only, padding end only, padding start and end) x (index start padding + data, index data only, index end padding + data). I believe this code handles all cases well. An extra benefit of handling the padding-only case at the level above would be they do not need to be considered here.

JoeZiminski:
BTW for the inequality on 
```
if start_frame > self.padding_start and end_frame < self.num_samples_in_original_segment + self.padding_start:
```
should the first inequality be `start_frame >= self.padding_start`? If I understand correctly `padding_start` is the upper-bound on the padding and padding goes from `0:padding_start` e.g. if `padding_start=5` the padded sample indices will be `0, 1, 2, 3, 4`. As such the data will start at sample 5, `start_frame >= self.padding_start`.

JoeZiminski:
Hey @h-mayorquin hope things are good! did you have any more thoughts on this in terms of performance? My guess is because its skips the block entirely, this should always be faster than the previous version. Happy to make any required changes!

h-mayorquin:
Sorry, @JoeZiminski . I actually had in my todo list to come back to this today (I was away the last two weeks) but I ran out of time. 

I will come back to you tomorrow and hopefullly we can get this through this week. Thanks for following on this.

JoeZiminski:
Thanks @h-mayorquin sorry to catch you just as you just got back! Sounds good no rush, thanks a lot for your help!

h-mayorquin:
> BTW for the inequality on
> 
> ```
> if start_frame > self.padding_start and end_frame < self.num_samples_in_original_segment + self.padding_start:
> ```
> 
> should the first inequality be `start_frame >= self.padding_start`? If I understand correctly `padding_start` is the upper-bound on the padding and padding goes from `0:padding_start` e.g. if `padding_start=5` the padded sample indices will be `0, 1, 2, 3, 4`. As such the data would start at sample 5, `start_frame >= self.padding_start`.

As for this, I was not thinking on this like bounds but as "numbers" so counting actual units. (which is not clear in the docstring) Let me think more about it but feel free to suggest which semantics are better.

alejoe91 (approved):

h-mayorquin (approved):
@JoeZiminski 
Thanks a bunch for all your work.
 As we want to have a release soon, we would like to add your fix as it is, I will add the tests that I suggested in a small follow up PR.

JoeZiminski:
Thanks @h-mayorquin, sorry for the slow response. Cheers for taking care of this and merging!

PR #1981, opened 2023-09-12
Additional quality metrics

alejoe91:
This PR adds the following new quality metrics, designed by @musall and slightly modified:

### Firing range

A measure of the variation of firing rate across an experiment. Computed as the distance between the 5th-95th percentile of _local_ firing rates computed on 10s intervals.

### Amplitude CV

A measure of the amplitude distribution dispersion. The CV is computed over temporal bins estimated for each unit to contain an `average_num_spikes_per_bin` spikes. Both the medians and the range (percentile distance) are returned.


doc/modules/qualitymetrics/amplitude_spread.rst
```diff
@@ -0,0 +1,48 @@
...
+===========================================
+
+
+Calculation
+-----------
+
+The amplitude spread is a measure of the amplitude variability.
+It is computed the ratio between the standard deviation and the amplitude mean (aka coefficient of variation).
+To obtain a better estimate of this measure, it is first computed separately for several bins of a prefixed number of spikes
+(e.g 100) and then the median of these values is taken.
```
zm711, line 11:
```suggestion
The amplitude spread is a measure of the amplitude variability.
It is computed as the ratio between the standard deviation and the amplitude mean (aka the coefficient of variation).
To obtain a better estimate of this measure, it is first computed separately for several bins of a prefixed number of spikes
(e.g. 100) and then the median of these values is taken.
```

doc/modules/qualitymetrics/amplitude_spread.rst
```diff
@@ -0,0 +1,48 @@
...
+
+The computation requires either spike amplitudes (see :py:func:`~spikeinterface.postprocessing.compute_spike_amplitudes()`)
+or amplitude scalings (see :py:func:`~spikeinterface.postprocessing.compute_amplitude_scalings()`) to be pre-computed.
+
+
+Expectation and use
+-------------------
+
+Very high levels of amplitude_spread ranges, outside of a physiolocigal range, might indicate noise contamination.
```
zm711, line 20:
```suggestion
Very high levels of amplitude_spread ranges, outside of a physiological range, might indicate noise contamination.
```

doc/modules/qualitymetrics/firing_range.rst
```diff
@@ -0,0 +1,40 @@
...
+
+The firing range indicates the dispersion of the firing rate of a unit across the recording. It is computed by
+taking the difference between the 95-th and 5th percentiles firing rates computed over short time bins (e.g. 10 s).
+
+
+
+Expectation and use
+-------------------
+
+Very high levels of firing ranges, outside of a physiolocigal range, might indicate noise contamination.
```
zm711, line 16:
```suggestion
Very high levels of firing ranges, outside of a physiological range, might indicate noise contamination.
```

doc/modules/qualitymetrics/firing_range.rst
```diff
@@ -0,0 +1,40 @@
+Firing range (:code:`firing_range`)
+===================================
+
+
+Calculation
+-----------
+
+The firing range indicates the dispersion of the firing rate of a unit across the recording. It is computed by
+taking the difference between the 95-th and 5th percentiles firing rates computed over short time bins (e.g. 10 s).
```
zm711, line 9:
```suggestion
taking the difference between the 95th percentile's firing rate and the 5th percentile's firing rate computed over short time bins (e.g. 10 s).
```

zm711: I think making this slightly more explicit will read cleaner. Otherwise it technically needs a percentiles' apostrophe for a plural possessive. Which is easier to confuse a reader.

doc/modules/qualitymetrics/firing_range.rst
```diff
@@ -0,0 +1,40 @@
...
+Example code
+------------
+
+.. code-block:: python
+
+    import spikeinterface.qualitymetrics as qm
+
+    # Make recording, sorting and wvf_extractor object for your data.
+    firing_range = qm.compute_firing_ranges(wvf_extractor)
+    # firing_range is a dict containing the units' IDs as keys,
```
zm711, line 28:
```suggestion
    # firing_range is a dict containing the unit IDs as keys,
```

I think the meaning is that each key is a different unit ID, in which case you would say unit IDs. The way it is written means that each unit has multiple IDs that are being used as keys, but maybe I'm misunderstanding what data is being returned?

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -563,7 +563,129 @@ def compute_synchrony_metrics(waveform_extractor, synchrony_sizes=(2, 4, 8), **k
     return synchrony_metrics
 
 
-_default_params["synchrony_metrics"] = dict(synchrony_sizes=(0, 2, 4))
+_default_params["synchrony"] = dict(synchrony_sizes=(0, 2, 4))
+
+
+def compute_firing_ranges(waveform_extractor, bin_size_s=5, quantiles=(0.05, 0.95), unit_ids=None):
+    """Calculate firing range, the range between the 5th and 95th quantiles of the firing rates distribution
```
zm711, line 570:
```suggestion
    """Calculate firing range, the range between the 5th and 95th percentiles of the firing rates distribution
```

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -563,7 +563,129 @@ def compute_synchrony_metrics(waveform_extractor, synchrony_sizes=(2, 4, 8), **k
...
+    computed in non-overlapping time bins.
+
+    Parameters
+    ----------
+    waveform_extractor : WaveformExtractor
+        The waveform extractor object.
+    bin_size_s : float, default: 5
+        The size of the bin in seconds.
+    quantiles : tuple, default: (0.05, 0.95)
```
zm711, line 579: I see why you chose quantile in the doc string, but technically a quantile needs to be an equal division of the data. So you could dive the data into fifths or tenths with percentiles being hundredths. It might be too annoying to change the term and I think most users would just understand, but I assume I could enter `quantiles = (0.02, 0.97)` and it would still be okay, although in this case you're allowing me to change the percentiles. No need to change, but just to keep in mind that the docs mention percentile and then the function uses quantile.

zm711 (commented):
couple typos :)
And two clarifications.

musall:
Thank you for the improved implementation!
Looks good to me and the added normalization should make it more generalizable.

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -563,7 +563,129 @@ def compute_synchrony_metrics(waveform_extractor, synchrony_sizes=(2, 4, 8), **k
...
+    all_unit_ids = list(sorting.unit_ids)
+    amplitude_spreads = {}
+    for unit_id in unit_ids:
+        amps_unit = amps[spikes["unit_index"] == all_unit_ids.index(unit_id)]
+        amp_mean = np.abs(np.mean(amps_unit))
+        if num_spikes[unit_id] < num_spikes_per_bin:
+            amp_spread = np.std(amps_unit) / amp_mean
+        else:
+            amp_spreads = []
+            for i in range(0, num_spikes[unit_id], num_spikes_per_bin):
```
samuelgarcia, line 680: I am not sure this is a good idea to have bins that are number of spikes.
It would be more intuituive to have real temporal bins.
The bin will be really diffrents depending across units the firing rate.

musall: The idea for using an equal number of spikes was to make sure that the variance in the amplitudes is not driven by variability in firing rate. There are sometimes bursts of high activity and then episodes in the recording with lower activity. When we tried a time-based binning approach this variability in the number of spikes per bin was driving the result and we also had to exclude bins with no spikes.

src/spikeinterface/qualitymetrics/misc_metrics.py
alejoe91: let's use percentiles then ;)

src/spikeinterface/qualitymetrics/misc_metrics.py
alejoe91: @musall actually even using a fixed number of spikes would suffer from variability in firing range. Potentially even more tha using temporal bins. 

Let's take an edge case: let's assume that `num_spikes_per_bin=50` and a neuron fires a 50-spikes burst in 100ms. In this case, the amplitude variance over 50 spikes will be **strongly** impacted by the firing rate. Instead, if we used constant temporal bins (say 10s) and the neuron went back to normal firing regime, the variance would be better estimated. What do you think?

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -563,7 +563,129 @@ def compute_synchrony_metrics(waveform_extractor, synchrony_sizes=(2, 4, 8), **k
...
+                amp_spreads.append(np.std(amps_unit[i : i + num_spikes_per_bin]) / amp_mean)
+            amp_spread = np.median(amp_spreads)
```
alejoe91, line 682:
```suggestion
            amp_spread = np.percentile(amplitude_spreads, percentiles[1]) - np.percentile(amplitude_spreads, percentiles[0])
```

I think that using the median here is not doing what we want. If I understand correctly, this metric should spot/quantify amplitudes like the following:

image

where probably the "bursts" of high-amplitude spiking activity are noise events. If the number of bins with normal activity is greater than the ones with high noisy activity, this behavior will not be captured by the median. Mean would be better, or even percentile distance (like the firing range). What do you think?

src/spikeinterface/qualitymetrics/misc_metrics.py
alejoe91: Maybe we could auto-estimate a correct bin size, based on the average firing rate? E.g. if firing rate is 1Hz and the average number of spikes in bins is 50, then the temporal bin size should be 50 seconds.

musall: That is a good idea. If we can infer temporal bins that are appropriate for the firing rate of a given neuron that should be more stable. We can also test different implementations against human labels to see how well they predict different clusters. We just used an AUC for that and could test that easily.
Would you just estimate the temporal bin size to contain a specific number of spikes on average?

src/spikeinterface/qualitymetrics/misc_metrics.py
musall: I think that makes total sense and as we discussed that was the original idea of the metric. The way it is currently implemented with the median would return the variance of the amplitudes but ignore these bursts that are probably triggered by noise events. It seems to me like both of these (median and percentile distance) are useful but describe different cluster features. Would it be possible to return both or would these be two different metrics then?

src/spikeinterface/qualitymetrics/misc_metrics.py
alejoe91: done

src/spikeinterface/qualitymetrics/misc_metrics.py
alejoe91: done

doc/modules/qualitymetrics/firing_range.rst
alejoe91: done for all quality metrics in last commit

alejoe91:
Ready for final revision! there was a bug with the percentile numbers (should be 0-100, not 0-1) which has now been fixed.

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -522,37 +523,41 @@ def compute_synchrony_metrics(waveform_extractor, synchrony_sizes=(2, 4, 8), **k
     Based on concepts described in [Gruen]_
     This code was adapted from `Elephant - Electrophysiology Analysis Toolkit <https://github.com/NeuralEnsemble/elephant/blob/master/elephant/spike_train_synchrony.py#L245>`_
     """
-    assert np.all(s > 1 for s in synchrony_sizes), "Synchrony sizes must be greater than 1"
+    assert np.all([s > 1 for s in synchrony_sizes]), "Synchrony sizes must be greater than 1"
```
zm711, line 526: Super minor but couldn't you just do
assert min(synchrony_sizes) > 1.
```
5.2 µs ± 23.8 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each) # np.all with list comprehension
```
```
135 ns ± 1.18 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each) # min with tuple/list
```
```
795 ns ± 2.93 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each) # min with np.array
```

But maybe I'm missing something?

alejoe91: Yes! Eheheh

alejoe91, line 526:
```suggestion
    assert min(synchrony_sizes) > 1, "Synchrony sizes must be greater than 1"
```

alejoe91: Done!

doc/modules/qualitymetrics/amplitude_cv.rst
```diff
@@ -0,0 +1,55 @@
...
+    amplitude_cv_median, amplitude_cv_range = sqm.compute_amplitude_cv_metrics(wvf_extractor)
+    # amplitude_cv_median and  amplitude_cv_range are dicts containing the unit ids as keys,
+    # and their amplitude_cv metrics as values.
+
+
+
+References
+----------
+
+.. autofunction:: spikeinterface.qualitymetrics.misc_metrics.compute_amplitude_spreads
```
samuelgarcia, line 49:
```suggestion
.. autofunction:: spikeinterface.qualitymetrics.misc_metrics.compute_amplitude_cv_metrics
```

samuelgarcia (approved):

PR #1982, opened 2023-09-12
Add Template class

h-mayorquin:
First draft. I am putting this out early to get feedback.

Right now the class implements two functionalities.

1) Looking around in the repo, it seems that all the uses of templates assume that the object is a numpy array. The new class has functionality so this assumption is stil valid. That is, the new template class behaves like a numpy array: it can be sliced like a numpy array, and you can use any ufunct (.e.g. `np.mean`). For this see the test. 
2) Serialization: both json and pickle are implemented and tested.


src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,60 @@
...
+from spikeinterface.core.sparsity import ChannelSparsity
+
+
+@dataclass
+class Templates:
+    templates_array: np.ndarray
+    sparsity: ChannelSparsity = None
+    num_units: int = field(init=False)
+    num_samples: int = field(init=False)
+    num_channels: int = field(init=False)
```
alejoe91, line 15: Required:
- `nbefore: int`, `sampling_frequency: float`
Optional:
- `channel_ids`, `unit_ids`

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def __post_init__(self):
+        self.num_units, self.num_samples, self.num_channels = self.templates_array.shape
+
+    # Implementing the slicing/indexing behavior as numpy
+    def __getitem__(self, index):
+        return self.templates_array[index]
```
alejoe91, line 22: It may be better to have a `unit_slice` and `sample/frame_slice` two explicitly slice over the first and last dimension that return a `Templates` object (rather than the sliced `templates_array`).

But maybe in a follow-up PR

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,60 @@
...
+        return self.templates_array
+
+    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs) -> np.ndarray:
+        # Replace any Templates instances with their ndarray representation
+        inputs = tuple(inp.templates_array if isinstance(inp, Templates) else inp for inp in inputs)
+
+        # Apply the ufunc on the transformed inputs
+        result = getattr(ufunc, method)(*inputs, **kwargs)
+
+        return result
```
alejoe91, line 34: Rmeove for same reason above

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,60 @@
...
+
+import numpy as np
+
+from spikeinterface.core.sparsity import ChannelSparsity
+
+
+@dataclass
+class Templates:
+    templates_array: np.ndarray
+    sparsity: ChannelSparsity = None
```
alejoe91, line 12:
```suggestion
    sparsity_mask: np.ndarray(bool) = None
```
We can always instantiate a `ChannelSparsity` if unit_ids and channel_ids are given.
```
get_channel_sparsity(self):
    assert self.unit_ids is not None and self.channel_ids is not None
    return ChannelSparsity(self.sparsity_mask, self.unit_ids, self.channel_ids)
```

alejoe91: this could also be constructed at `___post_init__` and stored as `self.sparsity`

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,60 @@
...
+    num_units: int = field(init=False)
+    num_samples: int = field(init=False)
+    num_channels: int = field(init=False)
+
+    def __post_init__(self):
+        self.num_units, self.num_samples, self.num_channels = self.templates_array.shape
```
alejoe91, line 18: How should we handle the templates_array when they are sparse, so that `sel.num_channels` is actually `self.max_num_sparse_channels`:

```
if self.sparsity_mask is None:
    self.num_units, self.num_samples, self.num_channels = self.templates_array.shape
else:
    self.num_units, self.num_samples, self.max_num_sparse_channels = self.templates_array.shape
```

src/spikeinterface/core/template.py
h-mayorquin: Done.

src/spikeinterface/core/template.py
h-mayorquin: Done.

src/spikeinterface/core/template.py
h-mayorquin: Removed.

src/spikeinterface/core/template.py
h-mayorquin: Yep. If the need arises.

src/spikeinterface/core/template.py
h-mayorquin: Added

alejoe91:
> @alejoe91 @samuelgarcia This is ready for another round of review. Let me know if this is lacking any properties that you think should go here.
> 
> The only thorny issue for me is mismatch between passing a templates_array that is dense but a sparsity mask. We have some options:
> 
> * Either throw an error.
> * Sparsity in the post-init if not sparse.
> * Let ir run and handle problems downstream.
> 
> I prefer the first but knowing that you like permisive methods and flexibility we have 3.

@h-mayorquin I agree that in that case we should throw an error in the post_init

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,156 @@
...
+    ms_after: float = field(init=False)
+    sparsity: ChannelSparsity = field(init=False, default=None)
+
+    def __post_init__(self):
+        self.num_units, self.num_samples = self.templates_array.shape[:2]
+        if self.sparsity_mask is None:
+            self.num_channels = self.templates_array.shape[2]
+        else:
+            self.num_channels = self.sparsity_mask.shape[1]
+        self.nafter = self.num_samples - self.nbefore - 1
```
alejoe91, line 32:
```suggestion
        self.nafter = self.num_samples - self.nbefore
```
`num_samples` is defined as `nbefore + nafter`

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,156 @@
...
+        sparse_waveforms = np.zeros(shape=(self.num_units, self.num_samples, max_num_active_channels))
+        for unit_index, unit_id in enumerate(self.unit_ids):
+            waveforms = self.templates_array[unit_index, ...]
+            sparse_waveforms[unit_index, ...] = self.sparsity.sparsify_waveforms(waveforms=waveforms, unit_id=unit_id)
+
+        return sparse_waveforms
+
+    def are_templates_sparse(self) -> bool:
+        if self.sparsity is None:
+            return False
```
alejoe91, line 99:
```suggestion
        return self.sparsity is not None
```
Otherwise it would return `None`, right?

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,156 @@
...
+
+        if self.templates_array.shape[-1] == self.num_channels:
+            return False
```
alejoe91, line 102: wouldn't it be enough to return:
```
return self.sparsity is not None
```
?

alejoe91 (changes requested):
@h-mayorquin a couple of small comments

src/spikeinterface/core/template.py
h-mayorquin: Only if we don't allow dense arrays when passing sparsity.

src/spikeinterface/core/template.py
h-mayorquin: Thanks this makes sense.

src/spikeinterface/core/template.py
h-mayorquin: So, if we throw an error or sparsify in the post-init this criteria can be simplified.

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,156 @@
...
+
+@dataclass(kw_only=True)
+class Templates:
+    templates_array: np.ndarray
+    sampling_frequency: float
+    nbefore: int
+
+    sparsity_mask: np.ndarray = None
+    channel_ids: np.ndarray = None
+    unit_ids: np.ndarray = None
```
samuelgarcia, line 15: Is it finally mandatory ? I was thinking it was optional.
I do not remmerber.

samuelgarcia: OK.
I get the anwser from line 36

src/spikeinterface/core/template.py
samuelgarcia: I think we should have a simple

```
    def are_templates_sparse(self) -> bool:
        return self.sparsity is not None
```
so there is no ambiguity.
self.sparse = None means that we handle dense
self.sparse not None means that we handle sparse

h-mayorquin: @alejoe91 @samuelgarcia 
Ok, this throws an error at initialization if the user passess `sparsity_mask` but the templates are not dense in the sense that the `sparsity_mask` mandates.

src/spikeinterface/core/tests/test_template_class.py
```diff
@@ -94,11 +96,8 @@ def test_get_sparse_templates(template_type):
...
-            template.num_units,
-            template.num_samples,
-            template.sparsity.max_num_active_channels,
-        )
-        assert not template.are_templates_sparse()
+
+
+def test_initialization_fail_with_dense_templates():
+    with pytest.raises(ValueError, match="Sparsity mask passed but the templates are not sparse"):
+        template = generate_test_template(template_type="sparse_with_dense_templates")
```
alejoe91, line 103: @h-mayorquin in this line, would it make sense to have a `Templates.make_sparse(sparsity)` method which returns a `Templates` object with sparse templates using the `sparsity`? (not only the template_array)

src/spikeinterface/core/template.py
h-mayorquin: I also modified some small inefficiency in the ChannelSparsity that Sam pointed out in the other PR.

src/spikeinterface/core/tests/test_template_class.py
h-mayorquin: Seems like an overkill to me. The only thing that is different is the `template_array`. 

If you want to make it simply and unintrusive for users we shouid sparsify ourselves at `__init__` instead of throwing an error. 
If you want to make it simpler but want to let them know we should have a method to sparsify the template array and tell them to use that.

alejoe91: got it. So the current way to do that would be:
```
templates_dense_array = ...
sparsity = ...

templates_sparse_array = sparsity.sparsify_waveforms(templates_dense)

templates_sparse = Templated(templates_sparse_array, sparsity.mask)
```

Indeed, looks simple enough  :)

alejoe91:
@h-mayorquin we reached the deprecation deadline!!

https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/extractors/cellexplorersortingextractor.py#L56-L59

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,197 @@
...
+    ----------
+    templates_array : np.ndarray
+        Array containing the templates data.
+    sampling_frequency : float
+        Sampling frequency of the templates.
+    nbefore : int
+        Number of samples before the spike peak.
+    sparsity_mask : np.ndarray, optional
+        Binary array indicating the sparsity pattern of the templates.
+        If `None`, the templates are considered dense.
```
zm711, line 22: Just a question here @h-mayorquin. Binary array means 1 and 0s and boolean array would mean True/False. I don't know what the term would be for something that can either be 1/0 or True/False. But are you hoping that this accepts the format 1/0. (Maybe this is just super annoying semantics, but I'm curious how you think about this?)

zm711 (commented):
One question for the docstring. Is there a design reason why you wouldn't want `nbefore` and `nafter` next to each other in the docstring? I'm not super used to data classes in general so maybe the docstring is enforced in this way, but at least from an ease of reading for me it would make sense to have the related concepts together.

src/spikeinterface/core/sparsity.py
```diff
@@ -185,16 +182,20 @@ def densify_waveforms(self, waveforms: np.ndarray, unit_id: str | int) -> np.nda
...
-        assert_msg = (
-            "Waveforms do not seem to be be in the sparsity shape of this unit_id. The number of active channels is "
-            f"{len(non_zero_indices)} but the waveform has {waveforms.shape[-1]} active channels."
-        )
-        assert self.are_waveforms_sparse(waveforms=waveforms, unit_id=unit_id), assert_msg
+        if not self.are_waveforms_sparse(waveforms=waveforms, unit_id=unit_id):
+            error_message = (
+                "Waveforms do not seem to be be in the sparsity shape of this unit_id. The number of active channels is "
+                f"{num_active_channels} but the waveform has non zero values outsies of those active channels: \n"
+                f"{waveforms[..., num_active_channels:]}"
```
alejoe91, line 191:
```suggestion
                "Waveforms do not seem to be in the sparsity shape for this unit_id. The number of active channels is "
                f"{num_active_channels}, but the waveform has non-zero values outsies of those active channels: \n"
                f"{waveforms[..., num_active_channels:]}"
```

src/spikeinterface/core/template.py
alejoe91: No, this should be boolean :)

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,197 @@
...
+    Attributes
+    ----------
+    templates_array : np.ndarray
+        Array containing the templates data.
+    sampling_frequency : float
+        Sampling frequency of the templates.
+    nbefore : int
+        Number of samples before the spike peak.
+    sparsity_mask : np.ndarray, optional
+        Binary array indicating the sparsity pattern of the templates.
```
alejoe91, line 21:
```suggestion
        Boolean 2D array indicating the sparsity pattern of the templates.
```

alejoe91 (changes requested):
Very minor docstring changes, then it's good to go for me! Thanks @h-mayorquin

src/spikeinterface/core/sparsity.py
```diff
@@ -204,7 +205,11 @@ def are_waveforms_dense(self, waveforms: np.ndarray) -> bool:
     def are_waveforms_sparse(self, waveforms: np.ndarray, unit_id: str | int) -> bool:
         non_zero_indices = self.unit_id_to_channel_indices[unit_id]
         num_active_channels = len(non_zero_indices)
-        return waveforms.shape[-1] == num_active_channels
+
+        # If any channel is non-zero outside of the active channels, then the waveforms are not sparse
+        excess_zeros = waveforms[..., num_active_channels:].sum()
+
+        return int(excess_zeros) == 0
```
samuelgarcia, line 212: Not sure to be happy with this.
This is computationaly costy. For me the shape is enough.

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,197 @@
...
+            )
+
+            # Test that the templates are sparse if a sparsity mask is passed
+            if not self._are_passed_templates_sparse():
+                raise ValueError("Sparsity mask passed but the templates are not sparse")
+
+    def to_dict(self):
+        return {
+            "templates_array": self.templates_array.tolist(),
+            "sparsity_mask": None if self.sparsity_mask is None else self.sparsity_mask.tolist(),
```
samuelgarcia, line 92: Why tolist() here normally the json encoder do it for us no ?

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,197 @@
...
+            "unit_ids": self.unit_ids.tolist(),
+            "sampling_frequency": self.sampling_frequency,
+            "nbefore": self.nbefore,
+        }
+
+    @classmethod
+    def from_dict(cls, data):
+        return cls(
+            templates_array=np.array(data["templates_array"]),
+            sparsity_mask=None if data["sparsity_mask"] is None else np.array(data["sparsity_mask"]),
```
samuelgarcia, line 103: np.asarray

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,197 @@
...
+        densified_shape = (self.num_units, self.num_samples, self.num_channels)
+        dense_waveforms = np.zeros(shape=densified_shape, dtype=self.templates_array.dtype)
+
+        for unit_index, unit_id in enumerate(self.unit_ids):
+            waveforms = self.templates_array[unit_index, ...]
+            dense_waveforms[unit_index, ...] = self.sparsity.densify_waveforms(waveforms=waveforms, unit_id=unit_id)
+
+        return dense_waveforms
+
+    def get_sparse_templates(self) -> np.ndarray:
```
samuelgarcia, line 124: What is the use case ? the internal is already sparse no ?

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,197 @@
...
+        for unit_index, unit_id in enumerate(self.unit_ids):
+            waveforms = self.templates_array[unit_index, ...]
+            are_templates_sparse = self.sparsity.are_waveforms_sparse(waveforms, unit_id=unit_id)
+            if not are_templates_sparse:
+                return False
+
+        return are_templates_sparse
+
+    def to_json(self):
+        return json.dumps(self.to_dict())
```
samuelgarcia, line 155: Lets use our encoder no ?

src/spikeinterface/core/sparsity.py
alejoe91: ok for now. If this becomes problematic later we'll fix it.

NOTE: this assumes that the extra channels are zero-padded!

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,197 @@
+import numpy as np
+import json
+from dataclasses import dataclass, field, astuple
+from .sparsity import ChannelSparsity
+
+
+@dataclass(kw_only=True)
```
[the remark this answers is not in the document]
alejoe91: @h-mayorquin can you remove it then?

src/spikeinterface/core/template.py
alejoe91: to be removed

src/spikeinterface/core/template.py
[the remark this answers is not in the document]
h-mayorquin: Yes. 
This is another shortcoming of only having tests for python 3.10 only.

src/spikeinterface/core/template.py
h-mayorquin: Yes, good idea, I changed it.

src/spikeinterface/core/template.py
h-mayorquin: Should be removed in the last commit.

src/spikeinterface/core/template.py
h-mayorquin: np.asarray does not create a copy when the thing is an array, changed.

src/spikeinterface/core/template.py
h-mayorquin: Yes, good call. Changed.

src/spikeinterface/core/sparsity.py
h-mayorquin: Added a boolean to deactivate this.

h-mayorquin:
@zm711 
Docstring related.
I don't have a good solution for this. `n_before` is required for the instantiation of the class but `n_after` is not. I separated parameters from the construction from the attributes in the latest docstring. If you have any suggestion or alternative models for how to do docstirng of dataclassess I am all ears.

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,196 @@
...
+        If `None`, the templates are considered dense.
+    channel_ids : np.ndarray, optional (default=None)
+        Array of channel IDs. If `None`, defaults to an array of increasing integers.
+    unit_ids : np.ndarray, optional (default=None)
+        Array of unit IDs. If `None`, defaults to an array of increasing integers.
+    check_for_consistent_sparsity : bool, optional (default=True)
+        When passing a sparsity_mask, this checks that the templates array is also sparse and that it matches the
+        structure fo the sparsity_masl.
+
+    The following attributes are avaialble after construction:
```
zm711, line 31:
```suggestion
    The following attributes are available after construction:
```

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,196 @@
...
+
+    It is constructed with the following parameters:
+    ----------
+    templates_array : np.ndarray
+        Array containing the templates data.
+    sampling_frequency : float
+        Sampling frequency of the templates.
+    nbefore : int
+        Number of samples before the spike peak.
+    sparsity_mask : np.ndarray, optional (default=None)
```
zm711, line 20:
```suggestion
    sparsity_mask : np.ndarray or None, default: None
```

I think the notation decided on was this for the docstring (whereas the current is pure Numpy Docstring format. But I would let @alejoe91 confirm the format before I throw a bunch of edits.

src/spikeinterface/core/template.py
```diff
@@ -0,0 +1,196 @@
...
+from .sparsity import ChannelSparsity
+
+
+@dataclass
+class Templates:
+    """
+    A class to represent spike templates, which can be either dense or sparse.
+
+    It is constructed with the following parameters:
+    ----------
```
zm711, line 13:
```suggestion
    Parameters
    ----------
```
I think you would still just say parameters here right? Just for cleanliness/lack of verbosity. But I'm not super against having the statement there.

zm711 (commented):
I think dividing the parameters like you have looks nice and clears up the confusion of having related parameters so spread about (for me at least). Thanks @h-mayorquin

src/spikeinterface/core/template.py
h-mayorquin: Dude, I keep forgetting this. Thanks.

src/spikeinterface/core/template.py
h-mayorquin: Yes, this is fine for me, thanks.

h-mayorquin:
This is should be ready btw @alejoe91 @samuelgarcia

alejoe91 (approved):

PR #1984, opened 2023-09-12
Do not merge, test for #1979

h-mayorquin:
Just as a showcase that the tests are passed in #1979


h-mayorquin:
@JoeZiminski has a better solutio in #1979

PR #1985, opened 2023-09-13
Add methods to sparsify and densify waveforms to `ChannelSparsity`

h-mayorquin:
I have been wanting to add something like this for a while and it will be useful in the context of https://github.com/SpikeInterface/spikeinterface/pull/1982.

The idea is that if we have channel sparsity, it should be possible to use the class to compute a sparse or dense representation of a waveform or a template. Now the `ChannelSparsity` object is part of WaveformExtractor but it does not have a way of changing the representation at a waveform at will in the spot. 

Instead, we do sparsifying operations all over the place directly with the mask:

https://github.com/h-mayorquin/spikeinterface/blob/aa5248454462084dcd7a05296ef82c0bb717a285/src/spikeinterface/core/waveform_extractor.py#L946-L950

https://github.com/h-mayorquin/spikeinterface/blob/aa5248454462084dcd7a05296ef82c0bb717a285/src/spikeinterface/core/waveform_extractor.py#L893-L909

I think it is better to centralize this operation where it semantically make sense (a sparsity class should sparsify!) and test it appropriately.

### DIFF
--- a/src/spikeinterface/core/sparsity.py
+++ b/src/spikeinterface/core/sparsity.py
@@ -149,7 +149,7 @@ class ChannelSparsity:
 
         assert_msg = (
             "Waveforms must be dense to sparsify them. "
-            f"Their last dimension {waveforms.shape[-1]} must be equal to the number of channels {self.num_chanenls}"
+            f"Their last dimension {waveforms.shape[-1]} must be equal to the number of channels {self.num_channels}"
         )
         assert self.are_waveforms_dense(waveforms=waveforms), assert_msg
 
--- a/src/spikeinterface/core/sparsity.py
+++ b/src/spikeinterface/core/sparsity.py
@@ -156,9 +156,6 @@ class ChannelSparsity:
         assert self.are_waveforms_dense(waveforms=waveforms), assert_msg
 
         non_zero_indices = self.unit_id_to_channel_indices[unit_id]
-        num_active_channels = len(non_zero_indices)
-        sparsified_shape = waveforms.shape[:-1] + (num_active_channels,)
-        sparsified_waveforms = np.zeros(sparsified_shape, dtype=waveforms.dtype)
         sparsified_waveforms = waveforms[..., non_zero_indices]
 
         return sparsified_waveforms
@@ -194,7 +191,7 @@ class ChannelSparsity:
 
         densified_shape = waveforms.shape[:-1] + (self.num_channels,)
         densified_waveforms = np.zeros(densified_shape, dtype=waveforms.dtype)
-        densified_waveforms[..., non_zero_indices] = waveforms[...]
+        densified_waveforms[..., non_zero_indices] = waveforms
 
         return densified_waveforms
 
--- a/src/spikeinterface/core/sparsity.py
+++ b/src/spikeinterface/core/sparsity.py
@@ -136,17 +136,16 @@ class ChannelSparsity:
         Parameters
         ----------
         waveforms : np.array
-            Dense waveforms with shape (num_units, num_samples, num_channels).
+            Dense waveforms with shape (num_waveforms, num_samples, num_channels) or a
+            single dense waveform (template) with shape (num_samples, num_channels).
         unit_id : str
             The unit_id for which to sparsify the waveform.
 
         Returns
         -------
         sparsified_waveforms : np.array
-            Sparse waveforms with shape (num_units, num_samples, num_active_channels).
-
-        Where num_active_channels is the number of channels that are active for this unit and should be
-        equal to the number of non-zero elements in the mask for this unit.
+            Sparse waveforms with shape (num_waveforms, num_samples, num_active_channels)
+            or a single sparsified waveform (template) with shape (num_samples, num_active_channels).
         """
 
         assert_msg = (
@@ -170,14 +169,16 @@ class ChannelSparsity:
         Parameters
         ----------
         waveforms : np.array
-            The sparsified waveforms array of shape (num_units, num_samples, num_active_channels).
+            The sparsified waveforms array of shape (num_waveforms, num_samples, num_active_channels) or a single
+            sparse waveform (template) with shape (num_samples, num_active_channels).
         unit_id : str
             The unit_id that was used to sparsify the waveform.
 
         Returns
         -------
         densified_waveforms : np.array
-            The densified waveforms array of shape (num_units, num_samples, num_channels).
+            The densified waveforms array of shape (num_waveforms, num_samples, num_channels) or a single dense
+            waveform (template) with shape (num_samples, num_channels).
 
         """
 
--- a/src/spikeinterface/core/sparsity.py
+++ b/src/spikeinterface/core/sparsity.py
@@ -103,8 +103,8 @@ class ChannelSparsity:
         self.max_channel_representation = self.mask.sum(axis=1).max()
 
     def __repr__(self):
-        sparsity = 1 - np.mean(self.mask)
-        txt = f"ChannelSparsity - units: {self.num_units} - channels: {self.num_channels} - sparsity, P(x=0): {sparsity:0.2f}"
+        density = np.mean(self.mask)
+        txt = f"ChannelSparsity - units: {self.num_units} - channels: {self.num_channels} - density, P(x=1): {density:0.2f}"
         return txt
 
     @property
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/sparsity.py
```diff
@@ -119,6 +125,65 @@ def unit_id_to_channel_indices(self):
                 self._unit_id_to_channel_indices[unit_id] = channel_inds
         return self._unit_id_to_channel_indices
 
+    def sparsify_waveforms(self, waveforms: np.ndarray, unit_index: int) -> np.ndarray:
```
h-mayorquin, line 128: Add some checks for dimensions.

src/spikeinterface/core/sparsity.py
```diff
@@ -92,13 +94,17 @@ def __init__(self, mask, unit_ids, channel_ids):
...
         self._unit_id_to_channel_indices = None
 
+        self.num_channels = self.channel_ids.size
+        self.num_units = self.unit_ids.size
+        self.max_channel_representation = self.mask.sum(axis=1).max()
+
     def __repr__(self):
-        ratio = np.mean(self.mask)
-        txt = f"ChannelSparsity - units: {self.unit_ids.size} - channels: {self.channel_ids.size} - ratio: {ratio:0.2f}"
+        sparsity = 1 - np.mean(self.mask)
```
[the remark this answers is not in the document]
alejoe91: +1 for me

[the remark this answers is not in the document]
h-mayorquin: Done.

src/spikeinterface/core/sparsity.py
h-mayorquin: Done.

src/spikeinterface/core/sparsity.py
```diff
@@ -119,6 +125,87 @@ def unit_id_to_channel_indices(self):
                 self._unit_id_to_channel_indices[unit_id] = channel_inds
         return self._unit_id_to_channel_indices
 
+    def sparsify_waveforms(self, waveforms: np.ndarray, unit_index: int) -> np.ndarray:
```
alejoe91, line 128: Is there a reason to go for `unit_index`? For high-level functions we usually use `unit_id` (same for the `unit_id_to_channel...` dictionaries).

src/spikeinterface/core/sparsity.py
```diff
@@ -119,6 +125,87 @@ def unit_id_to_channel_indices(self):
...
+            f"Their last dimension {waveforms.shape[-1]} must be equal to the number of channels {self.num_channels}"
+        )
+        assert self.are_waveforms_dense(waveforms=waveforms), assert_msg
+
+        unit_id = self.unit_ids[unit_index]
+        non_zero_indices = self.unit_id_to_channel_indices[unit_id]
+        num_sparse_channels = len(non_zero_indices)
+        sparsified_shape = waveforms.shape[:-1] + (num_sparse_channels,)
+        sparsified_waveforms = np.zeros(sparsified_shape, dtype=waveforms.dtype)
+        sparsified_waveforms[...] = waveforms[..., non_zero_indices]
```
alejoe91, line 161:
```suggestion
        sparsified_waveforms = waveforms[:, :, non_zero_indices]
```
should be the same, right?

src/spikeinterface/core/sparsity.py
h-mayorquin: OK, let me changed this. I was just chosing the simplest type but what you mention makes sense.

src/spikeinterface/core/sparsity.py
h-mayorquin: It should. And you mentioned it before.

h-mayorquin: Got you now, I Have changed this. Thanks.

src/spikeinterface/core/sparsity.py
```diff
@@ -119,6 +125,84 @@ def unit_id_to_channel_indices(self):
...
+        """
+        Densify sparse waveforms that were sparisified according to a unit's channel sparsity.
+
+        Given a unit_id its sparsified waveform, this method places the waveform back
+        into its original form within a dense array.
+
+        Parameters
+        ----------
+        waveforms : np.array
+            The sparsified waveforms array of shape (num_units, num_samples, num_active_channels).
```
alejoe91, line 173: This is `waveforms_or_template`, right? It can be 3d  (waveforms) or 2d (templates) (as in the tests). Maybe we should change the docstring and variable names accordingly. What do you think @h-mayorquin ?

h-mayorquin: I think this is a good suggesiton. I think on templates as central / canonical / representative forms of waveforms so I would like to keep the name of the variable waveforms as it is the general type. But I think that the docstring should clarify this and make it explicit.

h-mayorquin: I made some changes to the docstring. Let me know what you think. 

If you think that templates should be mentioned explitictly maybe it would still prefer to have instead a method called `sparsify_template` and `densify_template` that take care of the 2 dimensional case and separate the functions.

alejoe91: I think it's ok for now! I agree that templates are "mean" waveforms. Thanks for clarifying the docstring!

alejoe91 (approved):

src/spikeinterface/core/sparsity.py
```diff
@@ -119,6 +125,85 @@ def unit_id_to_channel_indices(self):
                 self._unit_id_to_channel_indices[unit_id] = channel_inds
         return self._unit_id_to_channel_indices
 
+    def sparsify_waveforms(self, waveforms: np.ndarray, unit_id: str) -> np.ndarray:
```
samuelgarcia, line 128: unit_id is not always a str can be int

src/spikeinterface/core/sparsity.py
```diff
@@ -119,6 +125,85 @@ def unit_id_to_channel_indices(self):
...
+            f"Their last dimension {waveforms.shape[-1]} must be equal to the number of channels {self.num_channels}"
+        )
+        assert self.are_waveforms_dense(waveforms=waveforms), assert_msg
+
+        non_zero_indices = self.unit_id_to_channel_indices[unit_id]
+        sparsified_waveforms = waveforms[..., non_zero_indices]
+
+        return sparsified_waveforms
+
+    def densify_waveforms(self, waveforms: np.ndarray, unit_id: str) -> np.ndarray:
```
samuelgarcia, line 162: same here

src/spikeinterface/core/sparsity.py
```diff
@@ -119,6 +125,85 @@ def unit_id_to_channel_indices(self):
...
+
+        """
+
+        non_zero_indices = self.unit_id_to_channel_indices[unit_id]
+
+        assert_msg = (
+            "Waveforms do not seem to be be in the sparsity shape of this unit_id. The number of active channels is "
+            f"{len(non_zero_indices)} but the waveform has {waveforms.shape[-1]} active channels."
+        )
+        assert self.are_waveforms_sparse(waveforms=waveforms, unit_id=unit_id), assert_msg
```
samuelgarcia, line 191: very little detail as a genral partice: here you format the message even before you made the assert.
But if the assert pass there is no need to make the text formatting. And so putting the message after the `assert ...,`  is better no ?

src/spikeinterface/core/sparsity.py
alejoe91: good point!

src/spikeinterface/core/sparsity.py
h-mayorquin: I do feel that waste of computation but I prefer this for readability. This should not have any really impact on performance anywhere.

As an aside to your aside, it is usually considered a bad practice to use asserts the way that we do here (but most people, doing this is very popular!) because you can dissable assertions in optimize code:

https://discuss.python.org/t/mismatch-between-asserts-semantics-and-how-its-used-o-oo-disable/29282?page=2

The recommended way to do this would be to do and if statement and then raise the appropiate exception but I think that most people does not like this? What do you think?

src/spikeinterface/core/sparsity.py
h-mayorquin: Yes, I forgot to do this. In fact, mea culpa, I kind of new it when I did it but I ... wished that unit_ids were only string and intentionally forgot : P

I will push a small fix for this.

I would not like to do have a method with two different type of inputs, then 1/4 of the code is used to validate the input with if cases.

src/spikeinterface/core/sparsity.py
zm711: I think this is a science vs industry thing. Every piece of science code I've ever read uses assert instead of if-exception. I think with the assumption that scientists aren't going to shut off assertions to increase speed. And using the asserts does empower the user such that if they don't want the protections designed by the developer they can shut them off to get the speed boost. But just my two cents.

PR #1986, opened 2023-09-13
Add `job_name` to `job_kwargs`

h-mayorquin:
job_kwargs if the keyword arguments of `ChunkRecordingExecutor` but for some reason `job_name` was missing. This PR fixes that.

Based on this discussion https://github.com/SpikeInterface/spikeinterface/pull/1799#discussion_r1324023115


src/spikeinterface/core/node_pipeline.py
```diff
@@ -366,13 +365,14 @@ def run_node_pipeline(
 
     init_args = (recording, nodes)
 
+    job_kwargs["job_name"] = "pipeline"
```
samuelgarcia, line 368: I am not sure we want to modify in place a global job_kwargs dict we have in many script.

samuelgarcia:
Finally I am not sure this is a good idea. (I change my mind often)

1. job_kwargs is not only for ChunkrecordingExecutor but also more stuff.
2. very often we 
```python
job_kwargs = dict(n_jobs=4, chunk_duration='2s')
my_parralel_func(..., **job_kwargs)
```
I do not want my job_kwargs to be modified my `my_parralel_func()` the job_name will be decided by `my_parralel_func` itself

h-mayorquin:
> Finally I am not sure this is a good idea. (I change my mind often)
> 
> 1. job_kwargs is not only for ChunkrecordingExecutor but also more stuff.

Can you point out this more stuff? This is important to know.

> 2. very often we
> 
> ```python
> job_kwargs = dict(n_jobs=4, chunk_duration='2s')
> my_parralel_func(..., **job_kwargs)
> ```
> 
> I do not want my job_kwargs to be modified my `my_parralel_func()` the job_name will be decided by `my_parralel_func` itself

The name is still decided by the `my_paralel_funct` , the job_kwrgs is already being modified here:
https://github.com/h-mayorquin/spikeinterface/blob/87e5a4046f855d631da83e775000edad9c8f60e5/src/spikeinterface/core/job_tools.py#L53-L78

src/spikeinterface/core/node_pipeline.py
h-mayorquin: This does not modify the global job_kwargs here. It is just replicating the behavior that we had before. 

Or maybe I am missing something?

alejoe91:
We agreed that `job_name` should not be changed by the end user, but it is fixed by the functions that use the `ChunkRecordingExecutor`

PR #1988, opened 2023-09-13
Relax check_borders in `InjectTemplatesRecording`

alejoe91:
The `check_borders` should be eventually moved outside (and it's making a bunch of thests failing at the moment)


h-mayorquin (approved):
Let's merge this.

PR #1999, opened 2023-09-15
Creation of a `TransformSorting` object to track modifications and bencharmk

### LINKED ISSUE #2028: Make `TransformSorting` class
Functions that take an existing `BaseSorting` and modify it, e.g., adding synchrony or removing spikes will return a `TransformedSorting` object.

Internally, this new class will have useful references to:
- `added_spikes_mask`: indices or mask of added spikes in the new spike vector
- `removed_spikes_mask`: indices or mask of removed spikes in the original spike vector

yger:
Draft prototype for a TransformedSorting object. I need feedback, because not sure exactly of what we really want. This object should take any sorting, and dict of units to remove/add spikes (dict of spikes (np array with sepcified dtype obtained via to_spike_vector?)), and then be sorting itself, with some additional attribute: 
- the sorting it has been obtained from
- an array of the removed spikes from the original sorting (as a plain binary mask)
- an array of the added spikes (as a plain binary mask)
I have concerns about the input signature of such an objects. Should we provide spike_vectors? Dicts?

Closes #2028


alejoe91:
Thanks for startnig this, @yger 

Just to recap, this object can be used to add or remove spikes from an existing Sorting object. 
Thinking about it a bit more, we could maybe simplify the implementation by making two small classes to greatly simplify the implementation:

1. `AddedSpikesSorting`: only adds spikes. The signature could be: 
```
def __init__(sorting, added_spikes_vector):
```
Internally, this would simply concatenate the existing and added spikes, get the indices of the added spikes and save it as `added_spikes_indices`

2. `RemovedSpikesSorting`: only removes spikes. The signature could be:
```
def __init__(sorting, remove_spikes_indices):
```
Internally, it would instantiate the new sorting object and save the `remove_spikes_indices`.

While the `AddedSpikesSorting` can be used as an output of the `generate_injected_sorting` and `add_synchrony_to_sorting`, the `RemoveSpikesSorting` could be the base class of the `RemoveDuplicateSpikes`.

What do you think? @samuelgarcia @h-mayorquin ?

samuelgarcia:
I think I prefer a unique object that handle every thing with a clear API:
  * add spikes (with an extra spike vector)
  * remove spikes (with a index of removed spikes)
  * add units (with unit ids list and extra spikes vector)
  * remove units ( with a unit ids list)

This will avoid combining a list of object and multiple iheritance or complicated init chain like we have in the curation layer.
I think it is not so complicated to have a unique object.

alejoe91:
> I think I prefer a unique object that handle every thing with a clear API:
> 
> * add spikes (with an extra spike vector)
> * remove spikes (with a index of removed spikes)
> * add units (with unit ids list and extra spikes vector)
> * remove units ( with a unit ids list)
> 
> This will avoid combining a list of object and multiple iheritance or complicated init chain like we have in the curation layer. I think it is not so complicated to have a unique object.

Ok, so something like this:
```
transform_sorting = TransformSorting(existing_sorting)

transform_sorting.remove_spikes(remove_spike_indices)
transform_sorting.add_spikes(new_spikes)
...
```
I agree that internally we should represent it as a spike vector, but let's say you call `add_spikes` twice..what should the behavior be in that case? Concatenate the added spikes mask? What if you do a add spikes -> remove spikes -> add units -> add spikes ? It could become very complicated to keep the internal references to the `existing_sorting` object.

src/spikeinterface/core/generate.py
```diff
@@ -235,6 +235,95 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+            if len(injected_spike_train) > n_injection:
+                injected_spike_train = np.sort(np.random.choice(injected_spike_train, n_injection, replace=False))
+
+            injected_spike_trains[segment_index][unit_id] = injected_spike_train
+
+    return NumpySorting.from_unit_dict(injected_spike_trains, sorting.get_sampling_frequency())
+
+
+class TransformedSorting(NumpySorting):
+    def __init__(self, sorting, added_spikes={}, removed_spikes={}):
```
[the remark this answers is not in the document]
alejoe91:
```
def __init__(self, sorting, added_spikes_vector_existing=None, added_spikes_vector_new=None, removed_spikes_indices=None):
    spike_vecor = soritng.to_spike_vector()
    if removed_spike_indices is not None:
        spike_vector = np.delete(spike_vector, removed_spike_indices)
    if added_spikes..:
        spike_vector = np.concatenate(())
    # handle new and old unit ids

@classmethod
def remove_spikes(sorting)
```

alejoe91:
After our discussion today:

The `TransformSorting` object initially will only add new spikes/units and keep a reference of the added spikes (`self.added_spikes_indices`), which can be used for Hybrid comparison and `InjectTemplates`

src/spikeinterface/core/generate.py
```diff
@@ -230,6 +230,78 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+
+            if len(injected_spike_train) > n_injection:
+                injected_spike_train = np.sort(np.random.choice(injected_spike_train, n_injection, replace=False))
+
+            injected_spike_trains[segment_index][unit_id] = injected_spike_train
+
+    return NumpySorting.from_unit_dict(injected_spike_trains, sorting.get_sampling_frequency())
+
+
+class TransformedSorting(BaseSorting):
```
alejoe91, line 271: Can you call this `TransformSorting`?

src/spikeinterface/core/generate.py
```diff
@@ -230,6 +230,78 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+        fs = sorting.get_sampling_frequency()
+        unit_ids = sorting.get_unit_ids()
+
+        BaseSorting.__init__(self, fs, unit_ids)
+        sorting.precompute_spike_trains()
+        assert added_spikes.dtype == minimum_spike_dtype, "added_spikes should be a spike vector"
+
+        self._cached_spike_vector = np.concatenate((sorting._cached_spike_vector, added_spikes))
+        self.added = np.zeros(len(self._cached_spike_vector), dtype=bool)
+        self.added[len(sorting._cached_spike_vector) :] = True
```
alejoe91, line 282: wait shouldn't `self.added` correspond to the indices of the added spikes?

src/spikeinterface/core/generate.py
yger: I've changed the names

src/spikeinterface/core/generate.py
yger: Could be, I thought added was a boolean mask with True or False for added spikes. But I can save only the indices, we just need to decide

yger:
For now, the function is not dealing with extra units that won't be already present in the main sorting. I'll think more about what do do, because I still to grasp all the logic of your indices. If in the added_spikes (as a spike_vector) I have indices that are not in the main sorting, i need to provide the unit_ids, am i right? So I guess I should provide an add_units arguments as a dict with {units_ids : [spikes]}, if we want to add extra units

src/spikeinterface/core/generate.py
alejoe91: Ahhh sorry I misread the logic. You first concatenate and then set the last Len(added_spikes) to True! Looks good ;)

yger:
Actually, why not do a add_sorting() method, that will take a sorting as an input instead of only a spike_vector ? Because if we do so, we could detect the unit_ids that will be common (and automatically) merge their spikes, and also the ones that would be different and thus updates indices accordingly. On a side note, if we only do union of BaseSortings, then maybe the name TransformSorting is not the best, and maybe UnionSorting would be more appropriate? TransformSorting can be more generic over times, but since we decided to not handle deletion....

yger:
I spoke with Sam, and not sure he is still in line with what we discussed alltogether last time. Here is a modified version to try to incorporates his comments. 
A light __init__() that will takes spike vector for existing ids, and spike vector for new units (with the unit_ids). The refractory period check is gone, because Sam is against, but I think this should still be there. It has to be discussed. 
Then, I made a handy constructor that will take two sorting and merge them, taking care of properly fusing the unit_ids (so you can merge sortings with non overlapping sets of unit_ids). To properly track the unit ids, TransformSorting now has two internal arrays (added_spikes and added_units) with boolean value to know if we have new spikes/units compared to the original sorting

yger:
Ready for review. Some questions that are left unanswered for now:

- what to do with the check for duplicated spikes/units? @sam was against such a check in the init, to keep it as light as possible, I would be encline to put it but it has to be discussed
- Currently, a TransformSorting is the union of two sortings. But do we want this object to be chained? i.e. can we apply TransformSorting to an already existing TransformSorting, and thus keep adding more spikes? If yes, this would not be to hard to implement, as long as we do not track from which parent sorting the spikes/units have been added (we do not keep track of merges)
- No spikes deletion/unit removal, as we discussed.

To make an illustraton, I've modified inject_som_duplicate_units() and add_synchrony_to_sorting() to now return TransformSorting objects, but this could easily be reverted/modified

src/spikeinterface/core/generate.py
```diff
@@ -220,16 +220,236 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+
+        idx_1 = len(sorting_1.unit_ids) + np.arange(len(exclusive_ids))
+        idx_2 = sorting_2.ids_to_indices(exclusive_ids)
+        not_common = spike_vector_2[~from_existing_units].copy()
+
+        # If indices are not the same, we need to remap
+        if not np.all(idx_1 == idx_2):
+            for i, j in zip(idx_1, idx_2):
+                mask = not_common["unit_index"] == j
+                not_common[mask]["unit_index"] = i
```
alejoe91, line 386: @yger can you explain what is this for?

src/spikeinterface/core/generate.py
```diff
@@ -220,16 +220,236 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+        The unit_ids that should be added if spikes for new units are added
+    refractory_period_ms : float, default 5
+        The refractory period violation to prevent duplicates and/or unphysiological addition
+        of spikes. Any spike times in added_spikes violating the refractory period will be
+        discarded
+
+    """
+
+    def __init__(
+        self, sorting: BaseSorting, added_spikes_existing_units=None, added_spikes_new_units=None, new_unit_ids=[]
```
h-mayorquin, line 297: You should not pass an empty list as a default argument. This is a common python gotcha:

https://nedbatchelder.com/blog/202311/say_it_again_values_not_expressions.html

src/spikeinterface/core/generate.py
```diff
@@ -220,16 +220,236 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+    spikes_duplicated = spikes_duplicated[sort_idxs]
 
-    sorting = NumpySorting(spikes=spikes_all, sampling_frequency=sorting.sampling_frequency, unit_ids=unit_ids)
+    synchronous_spikes = NumpySorting(spikes_duplicated, sorting.get_sampling_frequency(), unit_ids)
+    sorting = TransformSorting.add_from_sorting(sorting, synchronous_spikes)
 
     return sorting
 
 
+def generate_injected_sorting(
```
h-mayorquin, line 234: Is Alessio not moving this already in https://github.com/SpikeInterface/spikeinterface/pull/2253?

Maybe we should merge that first to avoid this small conflict. I had requested over there to add the seed to the np.random there but we can do that later if there is no time. 

What do you think, @alejoe91 ?

h-mayorquin (commented):
This looks OK to me. You have two cases for extension of the masks and you extend appropriately on each of them.

[EDIT] After Alessio explanation I think that what I need to understand is clear. I need to get my feet wet in some of the processing parts of the library to understand this. Hopefully playing with the wit hte IBL dataset will help me there.

src/spikeinterface/core/generate.py
```diff
@@ -220,16 +220,236 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+        assert (
+            sorting_1.get_sampling_frequency() == sorting_2.get_sampling_frequency()
+        ), "sampling_frequency should be the same"
+        assert type(sorting_1.unit_ids[0]) == type(sorting_2.unit_ids[0]), "unit_ids should have the same type"
+        # We detect the indices that are shared by the two sortings
+        mask_1 = np.isin(sorting_2.unit_ids, sorting_1.unit_ids)
+        common_ids = sorting_2.unit_ids[mask_1]
+        exclusive_ids = sorting_2.unit_ids[~mask_1]
+
+        # We detect the indicies in the spike_vectors
```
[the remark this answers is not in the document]
h-mayorquin: Also a type though.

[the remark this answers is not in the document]
h-mayorquin: Eye for an eye!

src/spikeinterface/core/tests/test_generate.py
```diff
@@ -452,6 +453,25 @@ def test_inject_templates():
         check_recordings_equal(rec, saved_loaded, return_scaled=False)
 
 
+def test_transformsorting():
+    sorting_1 = generate_sorting(seed=0)
```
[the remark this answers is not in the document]
yger: i've added some more tests

src/spikeinterface/core/generate.py
```diff
@@ -220,16 +220,236 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+
+
+class TransformSorting(BaseSorting):
+    """
+    Generates a sorting object keeping track of added spikes/units from an external spike_vector.
+    More precisely, the TransformSorting objects keeps two internal arrays added_spikes and added_spikes
+    as boolean mask to track (in the representation as a spike vector) where modifications have been
+    made
+
+    Parameters
```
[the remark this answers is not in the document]
alejoe91: Usually we document at the class level. Not sure it's everywhere like that, but in most of the cases it is!

src/spikeinterface/core/generate.py
alejoe91: @yger I can drop my PR to avoid conflicts. Can you add some docstrics for this function?

src/spikeinterface/core/generate.py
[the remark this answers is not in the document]
h-mayorquin: Got you. All the constructors here have their own docstirng so I guess it should not be any interference from the `__init__` constructor with the other ones.

src/spikeinterface/core/generate.py
```diff
@@ -220,16 +220,236 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+    ----------
+    sorting : BaseSorting
+        The sorting object
+    added_spikes_existing_units : np.array (spike_vector)
+        The spikes that should be added to the sorting object, for existing units
+    added_spikes_new_units: np.array (spike_vector)
```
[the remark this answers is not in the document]
h-mayorquin: Thanks for your explanation.

src/spikeinterface/core/generate.py
```diff
@@ -220,16 +220,319 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+
+        self._kwargs = dict(
+            sorting=sorting,
+            added_spikes_existing_units=added_spikes_existing_units,
+            added_spikes_new_units=added_spikes_new_units,
+            new_unit_ids=new_unit_ids,
+            refractory_period_ms=refractory_period_ms,
+        )
+
+    def get_added_spikes_indices(self):
```
alejoe91, line 392:
```suggestion
    def get_added_spike_indices(self):
```

src/spikeinterface/core/generate.py
```diff
@@ -220,16 +220,319 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+        return np.nonzero(self.added_spikes)[0]
+
+    def get_added_units_indices(self):
```
alejoe91, line 395:
```suggestion
    def get_added_unit_indices(self):
```

src/spikeinterface/core/generate.py
```diff
@@ -220,16 +220,319 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+        return np.nonzero(self.added_units)[0]
+
+    def get_added_units_inds(self):
```
alejoe91, line 398:
```suggestion
    def get_added_unit_ids(self):
```

alejoe91 (changes requested):
some tiny naming suggestions

alejoe91 (approved):

src/spikeinterface/core/generate.py
```diff
@@ -220,16 +220,236 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+
+            injected_spike_trains[segment_index][unit_id] = injected_spike_train
+
+    return NumpySorting.from_unit_dict(injected_spike_trains, sorting.get_sampling_frequency())
+
+
+class TransformSorting(BaseSorting):
+    """
+    Generates a sorting object keeping track of added spikes/units from an external spike_vector.
+    More precisely, the TransformSorting objects keeps two internal arrays added_spikes and added_spikes
```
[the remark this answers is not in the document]
h-mayorquin: @yger I don't understand why you marked this as resolved. Is your intention to repeat the word?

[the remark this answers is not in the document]
yger: My bad, you are right, this has been fixed

alejoe91:
@samuelgarcia reminder to have a final look at this

src/spikeinterface/core/generate.py
```diff
@@ -220,16 +221,340 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+        refractory_period_ms: Optional[float] = None,
+    ):
+        sampling_frequency = sorting.get_sampling_frequency()
+        unit_ids = list(sorting.get_unit_ids())
+
+        if new_unit_ids is not None:
+            assert ~np.any(
+                np.isin(new_unit_ids, sorting.unit_ids)
+            ), "some units ids are already present. Consider using added_spikes_existing_units"
+            if len(new_unit_ids) > 0:
```
samuelgarcia, line 346: we need to force new_units_ids as a list because it can be np.array

src/spikeinterface/core/generate.py
```diff
@@ -220,16 +221,340 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+                assert type(unit_ids[0]) == type(new_unit_ids[0]), "unit_ids should have the same type"
+                unit_ids += list(new_unit_ids)
```
samuelgarcia, line 348: We need to avoid the iplace modification here
```suggestion
                unit_ids = unit_ids + list(new_unit_ids)
```

src/spikeinterface/core/generate.py
```diff
@@ -220,16 +221,340 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+        self.refractory_period_ms = refractory_period_ms
+
+        self.added_spikes_from_existing_mask = np.zeros(len(self._cached_spike_vector), dtype=bool)
+        self.added_spikes_from_new_mask = np.zeros(len(self._cached_spike_vector), dtype=bool)
+
+        if added_spikes_existing_units is not None and len(added_spikes_existing_units) > 0:
+            assert (
+                added_spikes_existing_units.dtype == minimum_spike_dtype
+            ), "added_spikes_existing_units should be a spike vector"
+            added_unit_indices = np.unique(added_spikes_existing_units["unit_index"])
```
samuelgarcia, line 362: This is not valid because because we could have a empty unit.
We can infer without unique with  do np.arange(len(parent_unit_ids), len(parent_unit_ids)+len(new_unit_ids).
No ?

src/spikeinterface/core/generate.py
```diff
@@ -220,16 +221,340 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+
+    def get_added_spikes_from_existing_indices(self):
+        return np.nonzero(self.added_spikes_from_existing_mask)[0]
+
+    def get_added_spikes_from_new_indices(self):
+        return np.nonzero(self.added_spikes_from_new_mask)[0]
+
+    def get_added_units_inds(self):
+        indices = self._cached_spike_vector["unit_index"][self.get_added_spikes_from_new_indices()]
+        return np.unique(self.unit_ids[indices])
```
samuelgarcia, line 422: Sam remark as before.
We could have it as an attr

src/spikeinterface/core/tests/test_generate.py
```diff
@@ -452,6 +455,58 @@ def test_inject_templates():
...
+    assert len(transformed.unit_ids) == 3
+    assert np.all(np.array([k for k in transformed.count_num_spikes_per_unit(outputs="array")]) == 2)
+
+    sorting_1 = NumpySorting.from_unit_dict({0: np.array([12], dtype=int)}, sampling_frequency=20000.0)
+    sorting_2 = NumpySorting.from_unit_dict(
+        {0: np.array([150], dtype=int), 3: np.array([12, 150], dtype=int)}, sampling_frequency=20000.0
+    )
+    transformed = TransformSorting.add_from_sorting(sorting_1, sorting_2)
+    assert len(transformed.unit_ids) == 2
+    assert np.all(np.array([k for k in transformed.count_num_spikes_per_unit(outputs="array")]) == [2, 2])
```
samuelgarcia, line 486: use np.array_equal

src/spikeinterface/core/tests/test_generate.py
```diff
@@ -452,6 +455,58 @@ def test_inject_templates():
...
+    transformed = TransformSorting.add_from_sorting(sorting_1, sorting_3)
+    assert len(transformed.unit_ids) == 50
+
+    sorting_1 = NumpySorting.from_unit_dict({46: np.array([0, 150], dtype=int)}, sampling_frequency=20000.0)
+    sorting_2 = NumpySorting.from_unit_dict(
+        {0: np.array([100, 2000], dtype=int), 3: np.array([200, 4000], dtype=int)}, sampling_frequency=20000.0
+    )
+    transformed = TransformSorting.add_from_sorting(sorting_1, sorting_2)
+    assert len(transformed.unit_ids) == 3
+    assert np.all(np.array([k for k in transformed.count_num_spikes_per_unit(outputs="array")]) == 2)
```
samuelgarcia, line 478: use np.array_equal

src/spikeinterface/core/tests/test_generate.py
```diff
@@ -452,6 +455,58 @@ def test_inject_templates():
...
+
+    sorting_1 = NumpySorting.from_unit_dict({0: np.array([12], dtype=int)}, sampling_frequency=20000.0)
+    sorting_2 = NumpySorting.from_unit_dict(
+        {0: np.array([150], dtype=int), 3: np.array([12, 150], dtype=int)}, sampling_frequency=20000.0
+    )
+    transformed = TransformSorting.add_from_sorting(sorting_1, sorting_2)
+    assert len(transformed.unit_ids) == 2
+    assert np.all(np.array([k for k in transformed.count_num_spikes_per_unit(outputs="array")]) == [2, 2])
+
+    assert transformed.get_added_spikes_from_existing_indices().size == 3
```
samuelgarcia, line 488: This should be 1 no ?

src/spikeinterface/core/generate.py
```diff
@@ -220,16 +221,340 @@ def add_synchrony_to_sorting(sorting, sync_event_ratio=0.3, seed=None):
...
+                np.random.uniform(low=0, high=num_samples[segment_index], size=n).astype(np.int64)
+            )
+
+            # Remove spikes that are in the refractory period.
+            violations = np.where(np.diff(injected_spike_train) < t_r)[0]
+            injected_spike_train = np.delete(injected_spike_train, violations)
+
+            # Remove spikes that violate the refractory period of the real spikes.
+            # TODO: Need a better & faster way than this.
+            min_diff = np.min(np.abs(injected_spike_train[:, None] - spike_train[None, :]), axis=1)
```
samuelgarcia, line 291: This is an enormous matrix!!!!
We should have a better way with this kind of approach:
https://github.com/SpikeInterface/spikeinterface/blob/872da6ef16b1c0113da28d62c00dc74fbd5eb556/src/spikeinterface/postprocessing/correlograms.py#L237

In short make a sorted conatenated vector with spikes times + spike labels ordered.
And then with a small on shift we can detect and count how many spikesare violating.

(More or less the old algo for matching even)

What do you think ?

src/spikeinterface/core/tests/test_generate.py
yger: Done

src/spikeinterface/core/generate.py
yger: ok, done, forcing as a list

src/spikeinterface/core/generate.py
yger: Then we can simply do added_spikes_existint_units = np.arange(len(parent_unit_ids), and remove the check. Because we will assume that the spikes added will exactly comes from the same units

src/spikeinterface/core/generate.py
yger: I can make an attr, we I thought you wanted methods (and there are methods everywhere). Plus here this is not the same as before. If you have added spikes from new units, but with units with no spikes, then the fact that these inds are not returned by get_added_units_inds is not shocking to me. You have no spikes for them... How to you want to return them. You can do what you are suggesting by returning self.unit_ids[len(self.parent_sorting.unit_ids):], but it would requires saving parent_sorting in the __init__ as an attribute. This was currently not the case

src/spikeinterface/core/generate.py
yger: I did not touch this function. I guess the modification of this (and I agree this is not optimal) is out of the scope of this PR

src/spikeinterface/core/generate.py
yger: I've added the attribute self.parent_unit_ids as suggested

src/spikeinterface/core/generate.py
yger: Done

yger:
Ready on my side @samuelgarcia , I've integrated all your comments, thanks alot for the feedback! I've left the modification of the generation of the injected sorting, because this should be done in one other PR I guess

alejoe91 (approved):

PR #2000, opened 2023-09-15
Speed up searchsorted calls

yger:
Quite a lot everywhere in spikeinterface, searchsorted is called twice and thus a speedup can be gained by making a single call

### DIFF
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -1109,8 +1109,7 @@ class InjectTemplatesRecording(BaseRecording):
             num_samples = [num_samples]
 
         for segment_index in range(sorting.get_num_segments()):
-            start = np.searchsorted(self.spike_vector["segment_index"], segment_index, side="left")
-            end = np.searchsorted(self.spike_vector["segment_index"], segment_index, side="right")
+            start, end = np.searchsorted(self.spike_vector["segment_index"], [segment_index, segment_index+1], side="left")
             spikes = self.spike_vector[start:end]
             amplitude_vec = amplitude_vector[start:end] if amplitude_vector is not None else None
             upsample_vec = upsample_vector[start:end] if upsample_vector is not None else None
@@ -1208,8 +1207,8 @@ class InjectTemplatesRecordingSegment(BaseRecordingSegment):
         else:
             traces = np.zeros([end_frame - start_frame, n_channels], dtype=self.dtype)
 
-        start = np.searchsorted(self.spike_vector["sample_index"], start_frame - self.templates.shape[1], side="left")
-        end = np.searchsorted(self.spike_vector["sample_index"], end_frame + self.templates.shape[1], side="right")
+        start, end = np.searchsorted(self.spike_vector["sample_index"], [start_frame - self.templates.shape[1], 
+                                                                    end_frame + self.templates.shape[1] + 1], side="left")
 
         for i in range(start, end):
             spike = self.spike_vector[i]
--- a/src/spikeinterface/curation/remove_duplicated_spikes.py
+++ b/src/spikeinterface/curation/remove_duplicated_spikes.py
@@ -82,8 +82,7 @@ class RemoveDuplicatedSpikesSortingSegment(BaseSortingSegment):
         if end_frame == None:
             end_frame = spike_train[-1] if len(spike_train) > 0 else 0
 
-        start = np.searchsorted(spike_train, start_frame, side="left")
-        end = np.searchsorted(spike_train, end_frame, side="right")
+        start, end = np.searchsorted(spike_train, [start_frame, end + 1], side="left")
 
         return spike_train[start:end]
 
--- a/src/spikeinterface/postprocessing/spike_locations.py
+++ b/src/spikeinterface/postprocessing/spike_locations.py
@@ -77,8 +77,7 @@ class SpikeLocationsCalculator(BaseWaveformExtractorExtension):
         elif outputs == "by_unit":
             locations_by_unit = []
             for segment_index in range(self.waveform_extractor.get_num_segments()):
-                i0 = np.searchsorted(self.spikes["segment_index"], segment_index, side="left")
-                i1 = np.searchsorted(self.spikes["segment_index"], segment_index, side="right")
+                i0, i1 = np.searchsorted(self.spikes["segment_index"], [segment_index, segment_index + 1], side="left")
                 spikes = self.spikes[i0:i1]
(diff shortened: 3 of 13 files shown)


alejoe91:
Thanks Pierre! Can you check why tests are failing?

src/spikeinterface/sortingcomponents/motion_interpolation.py
```diff
@@ -155,8 +155,7 @@ def interpolate_motion_on_traces(
             **spatial_interpolation_kwargs,
         )
 
-        i0 = np.searchsorted(bin_inds, bin_ind, side="left")
-        i1 = np.searchsorted(bin_inds, bin_ind, side="right")
+        i0, i1 = np.searchsorted(bin_inds, [bin_ind, bin_ind + 1] side="left")
```
zm711, line 158:
```suggestion
        i0, i1 = np.searchsorted(bin_inds, [bin_ind, bin_ind + 1], side="left")
```

src/spikeinterface/core/generate.py
```diff
@@ -1208,8 +1209,11 @@ def get_traces(
         else:
             traces = np.zeros([end_frame - start_frame, n_channels], dtype=self.dtype)
 
-        start = np.searchsorted(self.spike_vector["sample_index"], start_frame - self.templates.shape[1], side="left")
-        end = np.searchsorted(self.spike_vector["sample_index"], end_frame + self.templates.shape[1], side="right")
+        start, end = np.searchsorted(
+            self.spike_vector["sample_index"],
+            [start_frame - self.templates.shape[1], end_frame + self.templates.shape[1] + 1],
+            side="left",
+        )
```
zm711, line 1217:
```suggestion

         start = np.searchsorted(self.spike_vector["sample_index"], start_frame - self.templates.shape[1], side="left")
         end = np.searchsorted(self.spike_vector["sample_index"], end_frame + self.templates.shape[1], side="right")
        )

```

For this one is a left search and one is a right search. Maybe this is part of the problem?

zm711: But not sure this is causing the exact failure?

yger: The +1 should handle the left/right discrepancy I guess, but I'll try to find why we have failures

zm711: If the +1 is at the edge of the array, search sorted returns length of the array. So this would fail on edge cases where someone wants that edge.

zm711: Just as super simple example: if our end frame has a value of 4 and we add one to switch it to a left search this could happen:

```
np.searchsorted([1,2,4], 5)
>>> 3
```

yger: My fault, I change some line while copy/pasting function calls.

yger: Right. Maybe I was too quick handling these right cases. Let's wait for test first, then otherwise I could revert old behaviour when right insertion are needed

h-mayorquin:
```python
import numpy as np

x = np.arange(int(1e8))
first_frame = 500_000
chunk_size = 100_000
last_frame = first_frame + chunk_size

# First case: Two individual searches
def search_individual():
    left1 = np.searchsorted(x, first_frame)
    right1 = np.searchsorted(x, last_frame)
    return left1, right1

# Second case: A single search with both values
def search_together():
    left2, right2 = np.searchsorted(x, [first_frame, last_frame])
    return left2, right2

# Use the timeit magic command to compare
print("First case:")
%timeit search_individual()

print("\nSecond case:")
%timeit search_together()

```
Output in my machine:

```
First case:
2.11 µs ± 124 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

Second case:
1.93 µs ± 107 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

h-mayorquin:
5 % improvement for every chunk acumulates. This looks good, @yger .

Not relevant in this PR but I wish we used better naming for the left and right bounds than `i0` , `i1` `s0` and `s1`. 

They are comfortable for the person that knows the code (like algebraic expressions are comfortable to operate in a mathematical theory that you already dominate) but are difficult, terse and non-obvious for someone who encounters the expression for the first time.

src/spikeinterface/core/generate.py
samuelgarcia: @yger please go back to the previous behavior do not change the right/left it is sometimes super complicated to debug!

src/spikeinterface/curation/remove_duplicated_spikes.py
```diff
@@ -82,8 +82,7 @@ def get_unit_spike_train(
         if end_frame == None:
             end_frame = spike_train[-1] if len(spike_train) > 0 else 0
 
-        start = np.searchsorted(spike_train, start_frame, side="left")
-        end = np.searchsorted(spike_train, end_frame, side="right")
+        start, end = np.searchsorted(spike_train, [start_frame, end_frame + 1], side="left")
```
samuelgarcia, line 85: same here you change the side for end. please go back to the previous

src/spikeinterface/postprocessing/spike_locations.py
```diff
@@ -77,8 +77,7 @@ def get_data(self, outputs="concatenated"):
         elif outputs == "by_unit":
             locations_by_unit = []
             for segment_index in range(self.waveform_extractor.get_num_segments()):
-                i0 = np.searchsorted(self.spikes["segment_index"], segment_index, side="left")
-                i1 = np.searchsorted(self.spikes["segment_index"], segment_index, side="right")
+                i0, i1 = np.searchsorted(self.spikes["segment_index"], [segment_index, segment_index + 1], side="left")
```
samuelgarcia, line 80: samde here

src/spikeinterface/sortingcomponents/motion_interpolation.py
```diff
@@ -155,8 +155,7 @@ def interpolate_motion_on_traces(
             **spatial_interpolation_kwargs,
         )
 
-        i0 = np.searchsorted(bin_inds, bin_ind, side="left")
-        i1 = np.searchsorted(bin_inds, bin_ind, side="right")
```
samuelgarcia, line 159: same here

src/spikeinterface/curation/remove_duplicated_spikes.py
yger: done

samuelgarcia (approved):

alejoe91 (approved):

PR #2003, opened 2023-09-15
Improve detect_bad_channels 

samuelgarcia:
`detect_bad_channels()`
  * fix when some recording is not depth ordered.
  * Add more chunk default computation for reproducibility

### DIFF
--- a/src/spikeinterface/preprocessing/detect_bad_channels.py
+++ b/src/spikeinterface/preprocessing/detect_bad_channels.py
@@ -17,7 +17,7 @@ def detect_bad_channels(
     n_neighbors=11,
     nyquist_threshold=0.8,
     direction="y",
-    chunk_duration_s=.5,
+    chunk_duration_s=0.5,
     num_random_chunks=100,
     welch_window_ms=10.0,
     highpass_filter_cutoff=300,
--- a/src/spikeinterface/preprocessing/detect_bad_channels.py
+++ b/src/spikeinterface/preprocessing/detect_bad_channels.py
@@ -17,7 +17,7 @@ def detect_bad_channels(
     n_neighbors=11,
     nyquist_threshold=0.8,
     direction="y",
-    chunk_duration_s=0.5,
+    chunk_duration_s=0.3,
     num_random_chunks=100,
     welch_window_ms=10.0,
     highpass_filter_cutoff=300,
--- a/src/spikeinterface/preprocessing/detect_bad_channels.py
+++ b/src/spikeinterface/preprocessing/detect_bad_channels.py
@@ -304,7 +304,7 @@ def detect_bad_channels_ibl(
     n_neighbors : int, optional
         Number of neighbors to compute median fitler, by default 11
     nyquist_threshold : float, optional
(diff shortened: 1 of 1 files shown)


src/spikeinterface/preprocessing/detect_bad_channels.py
```diff
@@ -306,7 +304,7 @@ def detect_bad_channels_ibl(
     n_neighbors : int, optional
         Number of neighbors to compute median fitler, by default 11
     nyquist_threshold : float, optional
-        Threshold on Nyquist frequency to calculate HF noise band, by default 0.8
+        Threshold on Nyquist frequency to calcureclate HF noise band, by default 0.8
```
alejoe91, line 307:
```suggestion
        Threshold on Nyquist frequency to calculate HF noise band, by default 0.8
```

src/spikeinterface/preprocessing/detect_bad_channels.py
```diff
@@ -17,8 +17,8 @@ def detect_bad_channels(
     n_neighbors=11,
     nyquist_threshold=0.8,
     direction="y",
-    chunk_duration_s=0.3,
-    num_random_chunks=10,
+    chunk_duration_s=0.5,
+    num_random_chunks=100,
```
alejoe91, line 21: I would just increase the number of random chunks.

samuelgarcia: The noisy channel are done on PSD. 300ms looks like quite short. I wanted to increase a bit.
where this 300ms comme from ?

JoeZiminski: Hey both, this was the default from the IBL white paper (code [here](https://github.com/int-brain-lab/ibl-neuropixel/blob/8a8cd834877167eee573ce0ae6b3fdfcbcd05bbc/src/neurodsp/voltage.py#L625)). In the context of the IBL pipeline PSD (I am not sure for the other methods in `detect_bad_channels`), I do not think changing this would make too much difference. At 30 kHz sampling rate the FFT frequency bin-width of 300 ms trace would be 30k / 9k = 3.33 Hz vs. 2 Hz (using 500 ms). The main use of the PSD in the IBL function is to calculate the average power above 80% Nyquist. For example, the true 80% of Nyquist at 30 kHz is the range 12-15000 Hz. This range would be something like 11997.99-14998.32 in practice with 3.33 Hz bins. So I don't think in this context at least it should matter much but I might have made an oversight / mistake.

src/spikeinterface/preprocessing/detect_bad_channels.py
```diff
@@ -17,8 +17,8 @@ def detect_bad_channels(
     n_neighbors=11,
     nyquist_threshold=0.8,
     direction="y",
-    chunk_duration_s=0.3,
-    num_random_chunks=10,
+    chunk_duration_s=0.5,
```
samuelgarcia, line 20:
```suggestion
    chunk_duration_s=0.3,
```

alejoe91 (approved):

PR #2005, opened 2023-09-18
Add an option to flip the order by depth

samuelgarcia:


src/spikeinterface/core/recording_tools.py
```diff
@@ -302,7 +302,7 @@ def get_chunk_with_margin(
     return traces_chunk, left_margin, right_margin
 
 
-def order_channels_by_depth(recording, channel_ids=None, dimensions=("x", "y")):
+def order_channels_by_depth(recording, channel_ids=None, dimensions=("x", "y"), flip=False):
```
h-mayorquin, line 305: Maybe using the standard reverse from the python API would be a good idea?

https://docs.python.org/3/library/functions.html#sorted

It is also supported by the list.sort

But maybe you think that flip describes it better?
Just throwing something that I thought.

src/spikeinterface/core/recording_tools.py
```diff
@@ -302,7 +302,7 @@ def get_chunk_with_margin(
...
     """
     Order channels by depth, by first ordering the x-axis, and then the y-axis.
```
h-mayorquin, line 307: Question out of curiosity, do we have a convention that x goes along the width or the recorder and that y goes in the height / depth (the direction where it is usualy moves) or is it the other way around?

I found myself realizing that I did not know what to expect after reading this.

src/spikeinterface/core/recording_tools.py
h-mayorquin: Ah, on the other hand we have:

https://numpy.org/doc/stable/reference/generated/numpy.flip.html

Which seems to be aligned withat what you are doing.

src/spikeinterface/core/recording_tools.py
alejoe91: Hi @h-mayorquin 

Yes the convention is that `x` is width and `y` is depth. The `('x', 'y')` tuple is fed into `lexsort`, which sorts outer to innermost dimensions ;)

h-mayorquin: Thanks, @alejoe91.

src/spikeinterface/core/recording_tools.py
h-mayorquin: Yep. I also think that your docstring is doing the heavy lifting there anyway.

src/spikeinterface/preprocessing/depth_order.py
```diff
@@ -18,13 +18,18 @@ class DepthOrderRecording(ChannelSliceRecording):
         If str, it needs to be 'x', 'y', 'z'.
         If tuple or list, it sorts the locations in two dimensions using lexsort.
         This approach is recommended since there is less ambiguity, by default ('x', 'y')
+    flip: bool, default False
```
samuelgarcia, line 21:
```suggestion
    flip: bool, default: False
```

src/spikeinterface/core/recording_tools.py
```diff
@@ -316,6 +316,9 @@ def order_channels_by_depth(recording, channel_ids=None, dimensions=("x", "y")):
         If str, it needs to be 'x', 'y', 'z'.
         If tuple or list, it sorts the locations in two dimensions using lexsort.
         This approach is recommended since there is less ambiguity, by default ('x', 'y')
+    flip: bool, default False
```
samuelgarcia, line 319:
```suggestion
    flip: bool, default: False
```

alejoe91 (approved):

Issue #2001, opened 2023-09-15
ChunkRecordingExecutor should skip chunks without peaks


yger:
I discussed that to Sam, but I think that currently, in SI, each time we have a node pipeline (or that we use the ChunkRecordingExecutor), this loads all the chunks to apply some internal function. However, if the first node is a PeakRetriever, one could and should first check that there is indeed peaks in a given chunk. Because otherwise, no need to load the traces, this is a waste or IO, and I guess there could be some speedup there

PR #2011, opened 2023-09-19
Patch to skip chunks without peaks

yger:
Could such a solution work @samuelgarcia ? This is a patch to try to address #2001. The idea is that if we have a PeakSource node, and no peaks between start_frame and end_frame, then we should clearly reduce IO by not loading the traces

### DIFF
--- a/src/spikeinterface/core/node_pipeline.py
+++ b/src/spikeinterface/core/node_pipeline.py
@@ -118,6 +118,12 @@ class PeakRetriever(PeakSource):
     def get_trace_margin(self):
         return 0
 
+    def has_peaks(self, start_frame, end_frame, segment_index):
+        sl = self.segment_slices[segment_index]
+        peaks_in_segment = self.peaks[sl]
+        i0, i1 = np.searchsorted(peaks_in_segment["sample_index"], [start_frame, end_frame])
+        return i0 < i1
+
     def get_dtype(self):
         return base_peak_dtype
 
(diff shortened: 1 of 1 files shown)


yger:
Ok, so what should be done is simply checking if the first node is a PeakSource, and it it has spikes, then load traces otherwise skip. I'll give it a try

yger:
Here we go, let me know if this works with the node_pipeline spirit!

samuelgarcia:
Yes. We check the first node and we skip the entire grpah if threre is no peak.
I propose t return None in that case so it is faster.

src/spikeinterface/core/node_pipeline.py
```diff
@@ -486,9 +498,18 @@ def _compute_peak_pipeline_chunk(segment_index, start_frame, end_frame, worker_c
...
-        recording_segment, start_frame, end_frame, None, max_margin, add_zeros=True
-    )
+
+    if isinstance(nodes[0], PeakRetriever) or isinstance(nodes[0], SpikeRetriever):
+        if nodes[0].has_peaks(start_frame, end_frame, segment_index):
+            traces_chunk, left_margin, right_margin = get_chunk_with_margin(
+                recording_segment, start_frame, end_frame, None, max_margin, add_zeros=True
+            )
+        else:
+            traces_chunk, left_margin, right_margin = None, None, None
```
samuelgarcia, line 508: If we do not have traces maybe we can also skip the node loop no ?

yger:
I've extended the PR to cases where we want to stop everything when at least max_peaks have been discovered. This is working, and to allow shuffliung of the chunks, we need to add some seed arguments to the run() function of the ChunkRecordingExecutor. By doing so, chunks can be shuffled, and as soon as max_peaks have been detected, we stop loading traces (or if no peaks are in the chunks)

alejoe91:
@samuelgarcia @yger should we keep this for later?

yger:
Yes, @sam wants to redesign something, so I guess this will be postponed

yger:
This has been done in #3356

PR #2014, opened 2023-09-19
Extend common postprocessing tests to spikes at borders

alejoe91:
Several postprocessors work in chunks and handle spikes at borders. However, this was not fully tested.

This PR adds the option to `generate_sorting` with `add_spikes_on_border=True` and tests all postprocessing and QC function against sorting objects with spikes on borders.

Also added some extra fixes:

- Fix handling of borders for amplitude_scalings
- Add the option to return sparse PCA projections (and extend tests for it)
- Fix QC metrics tests when NaNs are present


alejoe91:
@samuelgarcia done!

src/spikeinterface/postprocessing/amplitude_scalings.py
```diff
@@ -427,13 +431,13 @@ def _are_unit_indices_overlapping(unit_inds_to_channel_indices, i, j):
     bool
         True if the unit indices i and j are overlapping, False otherwise
     """
-    if len(np.intersect1d(unit_inds_to_channel_indices[i], unit_inds_to_channel_indices[j])) > 0:
+    if np.sum(np.logical_and(sparsity_mask[i], sparsity_mask[j])) > 0:
         return True
     else:
         return False
```
zm711, line 438: I thought for sure that doing 
`return np.sum(np.logical_and(sparsity_mask[i], sparsity_mask[j])) >0` was going to be faster, but when I ran `timeit` they were always within 2 microseconds of each other and would take turns being the faster.

src/spikeinterface/postprocessing/amplitude_scalings.py
```diff
@@ -427,13 +431,13 @@ def _are_unit_indices_overlapping(unit_inds_to_channel_indices, i, j):
     bool
         True if the unit indices i and j are overlapping, False otherwise
     """
-    if len(np.intersect1d(unit_inds_to_channel_indices[i], unit_inds_to_channel_indices[j])) > 0:
+    if np.sum(np.logical_and(sparsity_mask[i], sparsity_mask[j])) > 0:
```
samuelgarcia, line 434:
```suggestion
    if np.any(sparsity_mask[i] & sparsity_mask[j]):
```

samuelgarcia: @zm711 : like this ?

zm711: In my test I was still using `np.logical_and`, so `np.any(np.logical_and(x, y))` and that sped it up a little bit. I didn't try the test with the `&` operation. 

@samuelgarcia because the sparsity_mask is `(n_unit, n_chan)`, no? So it would work better with a vectorized operation?

samuelgarcia: & in numpy is suggar syntax for np.logical_and (if I am correct), I prefer it for readability but we can go back to the previous.
sparsity_mask[i] and sparsity_mask[j] are 1d vectors.

alejoe91: I think this is ok! Itàs very readable and it might be a bit faster. Good for me!

alejoe91:
@samuelgarcia ready to merge on my end

samuelgarcia (approved):

PR #2018, opened 2023-09-19
Add "how to" section on how to load matlab data

h-mayorquin:
We have a couple of users that came to ask us how to load data from matlab. I think that this is a good scenario for writing a how to so we can help new users without having to repeat ourselves.

I am hesistant to go all the way to make a specific extractor from matlab as @alejoe91 suggested because:
1) Matlab files might come in two formats: hdf5 and the classical one. The hdf5 can't be opened by scipy but the old one can and they don't have the same API. There are some project in that direction (shotout to https://pypi.org/project/pymatreader/ where I had a really small contribution recently) but there are still a lot of inconsistencies.
2) I don't have matlab and I don't want to build tests where I have to rely on how the users stored their data there. I feel that we will get some users getting erros within matlab and that sounds like annoying to debug and support.

Because of aspects two I feel that is just easier to give users general instructions of how to transform their data in matlab to a general format. In this case, binary should work well and we should then be able to process the data lazily using the `BinaryRecordingExtractor`. This tutorial does that. If we eventually decide to move all the way to build an extractor for matlab, then we can deprecate this section, meanwhile I think it can be useful to offer support to users.


doc/how_to/load_matalb_data.rst
```diff
@@ -0,0 +1,66 @@
...
+Exporting Data from MATLAB
+--------------------------
+
+First, ensure your data is structured correctly. The data matrix should be organized such that the first dimension corresponds to samples/time and the second dimension to channels.
+
+.. code-block:: matlab
+
+   % Define the size of your data
+   num_samples = 1000;
+   num_channels = 384;
```
zm711, line 15: Tiny point, and not official, but most Matlab users use camel case instead of snake case. As you do below.

doc/how_to/load_matalb_data.rst
```diff
@@ -0,0 +1,66 @@
...
+   # Define the path to your binary file
+   file_path = Path("/The/Path/To/Your/Data/your_data_as_a_binary.bin")
+
+   # Ensure the file exists
+   assert file_path.is_file()
+
+   # Specify the parameters of your recording
+   sampling_frequency = 30_000.0  # in Hz, adjust as per your matlab dataset
+   num_channels = 384  # adjust as per your matlab dataset
+   dtype = "float64"
```
zm711, line 51:
```suggestion
   dtype = "float64"  # equivalent of MATLAB double
```
I would just put it as comment here as well as in the pitfalls.

h-mayorquin:
Here is the link to the relevant section:

https://spikeinterface--2018.org.readthedocs.build/en/2018/how_to/load_matalb_data.html

image

doc/how_to/load_matalb_data.rst
```diff
@@ -0,0 +1,66 @@
...
+Once you have your data in a binary format, you can seamlessly load it into SpikeInterface using the following script:
+
+.. code-block:: python
+
+   from spikeinterface.core.binaryrecordingextractor import BinaryRecordingExtractor
+   from pathlib import Path
+
+   # Define the path to your binary file
+   file_path = Path("/The/Path/To/Your/Data/your_data_as_a_binary.bin")
+
```
zm711, line 44:
```suggestion
   file_path = Path("/The/Path/To/Your/Data/your_data_as_a_binary.bin")
   # or for Windows
   # file_path = Path(r"c:\path\to\your\data\your_data_as_a_binary.bin")
```
I tend to recommend just warning windows users ahead of time so that they don't come with a bunch of path issues.

doc/how_to/load_matalb_data.rst
```diff
@@ -0,0 +1,66 @@
...
+   # Ensure the file exists
+   assert file_path.is_file()
+
+   # Specify the parameters of your recording
+   sampling_frequency = 30_000.0  # in Hz, adjust as per your matlab dataset
+   num_channels = 384  # adjust as per your matlab dataset
```
zm711, line 50:
```suggestion
   sampling_frequency = 30_000.0  # in Hz, adjust as per your MATLAB dataset
   num_channels = 384  # adjust as per your MATLAB dataset
```
Just to maintain style

doc/how_to/load_matalb_data.rst
```diff
@@ -0,0 +1,66 @@
...
+   dtype = "float64"
+
+   # Load the data using SpikeInterface
+   recording = BinaryRecordingExtractor(file_path, sampling_frequency=sampling_frequency,
+                                        num_channels=num_channels, dtype=dtype, gain_to_uV=1, offset_to_uV=0)
```
zm711, line 55: I wonder if we need to warn them about `gain_to_uV` and `offset_to_uV`. Maybe just a comment saying that these are used to convert to the actual voltages in case their data came from a reader that returned "proprietary units" that haven't been converted yet?

zm711 (commented):
I think it would be useful to make sure people can transition. Matlab to python can feel like a pretty big jump, so giving some tips is great. Maybe link the numpy matlab to python doc at the bottom too? 
https://numpy.org/doc/stable/user/numpy-for-matlab-users.html

doc/how_to/load_matalb_data.rst
alejoe91: I think it's a very good point and it makes the Python/MATLAB distinction even clearer :)

doc/how_to/load_matalb_data.rst
```diff
@@ -0,0 +1,66 @@
...
+   # Ensure the file exists
+   assert file_path.is_file()
+
+   # Specify the parameters of your recording
+   sampling_frequency = 30_000.0  # in Hz, adjust as per your matlab dataset
+   num_channels = 384  # adjust as per your matlab dataset
+   dtype = "float64"
+
+   # Load the data using SpikeInterface
+   recording = BinaryRecordingExtractor(file_path, sampling_frequency=sampling_frequency,
```
alejoe91, line 54: Can we switch to the function notation?
```
import spikeinterface as si

si.read_binary(...)
```

doc/how_to/load_matalb_data.rst
```diff
@@ -0,0 +1,66 @@
...
+   # Verify the shape of your data
+   assert recording.get_traces().shape == (num_samples, num_channels)
+
+Common Pitfalls & Tips
+----------------------
+
+1. **Data Shape**: Always ensure that your MATLAB data matrix's first dimension corresponds to samples/time and the second to channels.
+2. **File Path**: Double-check the file path in Python to ensure you're pointing to the right directory.
+3. **Data Type**: When moving data between MATLAB and Python, it's crucial to keep the data type consistent. In our example, we used `double` in MATLAB, which corresponds to `float64` in Python.
+4. **Sampling Frequency**: Ensure you set the correct sampling frequency when loading data into SpikeInterface.
```
alejoe91, line 66:
```suggestion
4. **Sampling Frequency**: Ensure you set the correct sampling frequency in Hz when loading data into SpikeInterface.
```

doc/how_to/index.rst
```diff
@@ -7,3 +7,4 @@ How to guides
     get_started
     analyse_neuropixels
     handle_drift
+    load_matalb_data
```
alejoe91, line 10: Note that you have a typo in the file name: 
```suggestion
    load_matlab_data
```

doc/how_to/load_matalb_data.rst
h-mayorquin: Yes.
I do dislike the full import and I think you should decide on those matters.

doc/how_to/load_matalb_data.rst
h-mayorquin: Thanks!

doc/how_to/index.rst
h-mayorquin: Yep, thanks for the correction! Worse place to have a typo.

doc/how_to/load_matalb_data.rst
h-mayorquin: Yes, thanks. I was concerned as well that because of the format it will not be clear that we talking about matlab code. I switched to camelCase and added an extra sentence to clarify this is indeed matlab code.

doc/how_to/load_matalb_data.rst
h-mayorquin: Thanks. Yes, this makes sense.

doc/how_to/load_matalb_data.rst
h-mayorquin: Good point.
What do you think of the "path/to/your/data/" visual trick to let them know that it is a path? In another package we have similar tutorials:

https://neuroconv.readthedocs.io/en/main/conversion_examples_gallery/recording/spikeglx.html

And we opted for something more concise there (just a variable in caps). I would like to hear if you have any ideas on how to make this clearer.

doc/how_to/load_matalb_data.rst
h-mayorquin: Thanks.

doc/how_to/load_matalb_data.rst
alejoe91: Well the use case could be that the user have their data in `int16`. In that case, gains and offsets are needed to correctly convert to uV. I think it's an important concept to spend a couple of words on and it is related to the task!

doc/how_to/load_matalb_data.rst
alejoe91: yes, in my opinion in the tutorials/how-to we should switch to import modules. I also use it my self since I think that it makes it very clear what eahc line does.

h-mayorquin: By switch to import modules you mean `spikeinterface.core` and the like or the style above where everything is on the `si` space?

alejoe91: All submodules except core:
```
import spikeinterface as si
...
```

h-mayorquin: OK!

doc/how_to/load_matalb_data.rst
zm711: I think the path to variable is super tricky. Because some users will actually put `path/to/variable` thinking that it is a hard coded path. And then adding in windows confusion with the opposite slashes. But as far as strategies I still think it is best despite being a bit longer and less concise.

Only issue with the concise variable style is that for other users they might not realize what's behind the variable. So that's why I prefer the `path/to/variable` style although it is not perfect. For Linux users I think they tend to be more comfortable with the path notation. But Windows has a right+click copy path option that will copy with `x\y\z` instead, so even though they could put in a Path(`c:/my/data`) they almost never do. And then since Windows has backslashes with escapes if we don't warn them about using the raw string instead they will get pathway errors due to the escaping and get frustrated.

In summary, I would leave it as you have it, but add in a the comment for windows users so when they copy the path from the computer they know that they likely need the r.

doc/how_to/load_matalb_data.rst
zm711: I like the idea of having a super goal oriented tutorial to get the job done. But in this case those two values are (at least in my opinion) important for the task at hand. As @alejoe91 said in the case the data was stored as an `int16` and needs to converted into a voltage. I think at a minimum a comment saying that not all data are in the correct format for all sorters so `gain_to_uV` and `offset_to_uV` will give you this fine control and then a link to relevant documentation that explains how these options work. Although in this case I tend toward educating (or maybe you'd argue over-educating) the user rather than overwhelming the user.

h-mayorquin: `gain_to_uv` and `offset_to_uv` are an implementation detail of data extraction. My fear is that most users will not be familiar with them and that they already will have the data in the right units (as they are using that in MATLAB probably to analyze). Introducing these concepts to new users -as I was one- is more likely to raise more questions, confuse and derail. I don't think we can get away with just a few comments.

I think we can meet in the middle. I added a specific section at the end dealing with integer typed traces, this separate the information streams and does not get in the way of new users and gives us more space to introduce the necessary context for using gains and offsets. Could you check it @alejoe91 and @zm711 ?

doc/how_to/load_matalb_data.rst
```diff
@@ -0,0 +1,66 @@
...
+   recording = BinaryRecordingExtractor(file_path, sampling_frequency=sampling_frequency,
+                                        num_channels=num_channels, dtype=dtype, gain_to_uV=1, offset_to_uV=0)
+
+   # Verify the shape of your data
+   assert recording.get_traces().shape == (num_samples, num_channels)
+
+Common Pitfalls & Tips
+----------------------
+
+1. **Data Shape**: Always ensure that your MATLAB data matrix's first dimension corresponds to samples/time and the second to channels.
```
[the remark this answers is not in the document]
h-mayorquin: I added this.  

Thinking about this, it makes me feel that the order argument in the get_traces is kind of redundant:

https://github.com/catalystneuro/spikeinterface/blob/1ead6a33e658bf5a0365d21506a90dd9bd32e67c/src/spikeinterface/core/baserecording.py#L260-L261

What do you think?

doc/how_to/load_matlab_data.rst
```diff
@@ -0,0 +1,100 @@
...
+                              gain_to_uV=gain_to_uV, offset_to_uV=offset_to_uV)
+
+   recording.get_traces(return_scaled=True)  # Return traces in micro volts (uV)
+
+
+This will equip your recording object with capabilities to convert the data to float values in uV using the `get_traces()` method with the `return_scaled` parameter set to `True`.
+
+.. note::
+
+   The gain and offset parameters are usually format depend and you will need to find out the correct values for your data format. You can load your data without gain and offset but then the traces will be in integer values and not in uV.
```
zm711, line 100:
```suggestion
   The gain and offset parameters are usually format dependent and you will need to find out the correct values for your data format. You can load your data without gain and offset but then the traces will be in integer values and not in uV.
```

doc/how_to/load_matalb_data.rst
h-mayorquin: Thanks for your input.

doc/how_to/load_matalb_data.rst
zm711: I'm fine with that section. It fits with the design I like (hive off the extra info for those that need it, but give an off-ramp for those who don't). Thanks @h-mayorquin :)

h-mayorquin: Great. Do check the video linked above if you haven't before. I think is very good by the way.

doc/how_to/load_matlab_data.rst
```diff
@@ -0,0 +1,100 @@
...
+   sampling_frequency = 30_000.0  # Adjust according to your MATLAB dataset
+   num_channels = 384  # Adjust according to your MATLAB dataset
+   dtype = "float64"  # MATLAB's double corresponds to Python's float64
+
+   # Load data using SpikeInterface
+   recording = si.read_binary(file_path, sampling_frequency=sampling_frequency,
+                                        num_channels=num_channels, dtype=dtype)
+
+   # Confirm the data shape
+   assert recording.get_traces().shape == (numSamples, num_channels)
```
zm711, line 61: This assert is technically not possible. Python doesn't know the `numSamples`, `num_channels` from Matlab. (I'm just worried someone will copy paste and then get an error). Maybe instead say :

# confirm the data shape (input your numSamples, and numChannels above into the boolean)

zm711:
@h-mayorquin 

Sorry just one more comment and then looks good to me.

doc/how_to/load_matlab_data.rst
h-mayorquin: Thanks a lot for catching this.

h-mayorquin:
@zm711 
> Sorry just one more comment and then looks good to me.


Thanks for all your feedback. No worries if you want to do more suggestions. I think your feedback has improved the PR a lot.

doc/how_to/load_matalb_data.rst
zm711: It is now bookmarked. I'll check it out this afternoon :)

doc/how_to/load_matlab_data.rst
```diff
@@ -0,0 +1,100 @@
...
+   fileID = fopen('your_data_as_a_binary.bin', 'wb');
+   fwrite(fileID, data, 'double');
+   fclose(fileID);
+
+.. note::
+
+   In your own script, replace the random data generation with your actual dataset.
+
+Loading Data in SpikeInterface
+-----------------------------
```
zm711, line 31:
```suggestion
Loading Data in SpikeInterface
----------------------------------
```

Checked the RTD logs and this needs to be lengthened.

zm711: The suggestion view was terrible for trying to line that. I think it is fixed now in my suggestion, but feel free to double check.

doc/how_to/load_matlab_data.rst
```diff
@@ -0,0 +1,100 @@
...
+   offset_to_uV = 0   # Adjust according to your MATLAB dataset
+
+   recording = si.read_binary(file_path, sampling_frequency=sampling_frequency,
+                              num_channels=num_channels, dtype=dtype_int,
+                              gain_to_uV=gain_to_uV, offset_to_uV=offset_to_uV)
+
+   recording.get_traces(return_scaled=True)  # Return traces in micro volts (uV)
+
+
+This will equip your recording object with capabilities to convert the data to float values in uV using the `get_traces()` method with the `return_scaled` parameter set to `True`.
```
alejoe91, line 96:
```suggestion
This will equip your recording object with capabilities to convert the data to float values in uV using the :code:`get_traces()` method with the :code:`return_scaled` parameter set to :code:`True`.
```

samuelgarcia:
I did not have time to read this yet. Sorry.
Lets keep in mind that how page are local "jupytext" file from `examples/how_to/` converted to rst an not direct rst push.

In short slow notebook are done locally in  `examples/how_to/` and exported to `doc/how_to/` manually
And fast notebook are generated by the sphinx-gallery mechanism in `examples/modules_gallery`.

doc/how_to/load_matlab_data.rst
h-mayorquin: Thanks

doc/how_to/load_matlab_data.rst
h-mayorquin: We will see.

zm711:
I'll be honest this is my confusion in general. `how_to` to mean means general guides to various things. So for example how to convert matlab to python. It doesn't require a notebook, but as an end user I would go to the how to folder to see how to do this. Whereas the current organization is that how to is just for big scripts to run locally as notebooks and convert. I would tend to just add. I don't have a solution, but I can say as a reader of documentation, doing it @h-mayorquin 's way of adding this to the how to makes sense for me being able to find it even though it clashes with the developer work flow.

alejoe91:
> I'll be honest this is my confusion in general. `how_to` to me means general guides to various things. So for example how to convert matlab to python. It doesn't require a notebook, but as an end user I would go to the how to folder to see how to do this. Whereas the current organization is that how to is just for big scripts to run locally as notebooks and convert. I would tend to just add. I don't have a solution, but I can say as a reader of documentation, doing it @h-mayorquin 's way of adding this to the how to makes sense for me being able to find it even though it clashes with the developer work flow.

I agree. It's not a rule that all how-tos should be a python script first. Like in this case, there is no point whatsoever to make a python script

zm711:
Well, wherever this ends up in the toctree it looks good to me!

alejoe91:
Same for me! @samuelgarcia wanna take a final look?

samuelgarcia (approved):

PR #2025, opened 2023-09-20
Update a test expression to remove two logical short circuits

munahaf:
In file: remove_artifacts.py, method: `__init__`, a logical equality check operation was performed. The operands are such that the comparison operation always returns either true or false. Such logical short circuits in code lead to unintended behavior. In this case both parts of a test expression were always returning false.

I suggested that the logical operation should be reviewed for correctness. 

Sponsorship and Support:

This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

### DIFF
--- a/src/spikeinterface/preprocessing/remove_artifacts.py
+++ b/src/spikeinterface/preprocessing/remove_artifacts.py
@@ -165,7 +165,9 @@ class RemoveArtifactsRecording(BasePreprocessor):
                 for l in np.unique(labels):
                     assert l in artifacts.keys(), f"Artefacts are provided but label {l} has no value!"
             else:
-                assert "ms_before" is not None and "ms_after" is not None, f"ms_before/after should not be None for mode {mode}"
+                assert (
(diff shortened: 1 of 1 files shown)


alejoe91 (approved):

munahaf:
I updated the originally proposed fix which was wrong. Probably, there was a typo in which instead of `ms_before` the variable, a literal is used.

alejoe91:
> I updated the originally proposed fix which was wrong. Probably, there was a typo in which instead of `ms_before` the variable, a literal is used.

Thanks! Indeed these are var names ;)

alejoe91 (approved):

PR #2027, opened 2023-09-20
Improve serialization concept : memory/json/pickle

samuelgarcia:
Some recording are not json serializable but they be dump with pickle (the new generator for instance)?.
this prevent the waveform extractor to be load back and also to run sorters.

Here a proposal to extend at some place the engine to serializa an object.
This will be super convinient for running sorter on simulated data which are not file based.

This also make rename `is_dumpable()` to a more explicit `is_memory_serializable()`

This also make a better seprartion between:

  * `is_memory_serializable()` / `check_serializablility("memory")`: a object can be dump to memory for multi processing
  * `check_if_json_serializable()` / `check_serializablility("json")`: a object can be serialized to a **file** using json
  * `check_if_pickle_serializable()` / `check_serializablility("pickle")` : a object can be serialized to file using pickle

In this PR:

 * [X] Improve the concept `check_serializablility(type)` with an internal `self._serializablility` dict
 * [X] enable pickle dump to file in waveform extractor
 * [x] enable pickle dump to file base sorter
 * [x] debug the nested check

Note:
  * NumpRecording is not json and not pickle because it is too big : we do not want a pickle/json copy of theses object.


samuelgarcia:
@alejoe91 : I think I would not complexify too much this PR.
I finally leave the pickle for NumpySorting other is too much a ascade of change for now.
this is not perfect but at least we can play with generators recording.

Tell what you think.

src/spikeinterface/core/tests/test_waveform_extractor.py
```diff
@@ -509,11 +509,48 @@ def test_compute_sparsity():
...
+def test_non_json_object():
+    recording, sorting = generate_ground_truth_recording(
+        durations=[30, 40],
+        sampling_frequency=30000.0,
+        num_channels=32,
+        num_units=5,
+        )
+    
+
+    print(recording.check_serializablility("pickle"))
```
h-mayorquin, line 521: Let's not have prints here.

src/spikeinterface/core/base.py
```diff
@@ -483,13 +484,35 @@ def check_if_dumpable(self):
...
+                for v in value:
+                    if isinstance(v, BaseExtractor) and not v.check_if_dumpable():
+                        return False
+            elif isinstance(value, dict):
+                for v in value.values():
+                    if isinstance(v, BaseExtractor) and not v.check_if_dumpable():
+                        return False
         return self._is_dumpable
 
+    def check_serializablility(self, type="json"):
```
h-mayorquin, line 499: I really dislike this idea of using a dictionary as a carrier of arbitrary serilization method. I think that the previous flag method is simple, explicit and does its job well.

Json and pickle are separate concepts with very different purposes. I think it would better to  have some code duplication in methods like instead of coupling this concepts together which I feel will bring pain in the future.

src/spikeinterface/core/generate.py
```diff
@@ -1431,5 +1433,7 @@ def generate_ground_truth_recording(
     )
     recording.annotate(is_filtered=True)
     recording.set_probe(probe, in_place=True)
+    recording.set_property("gain_to_uV", np.ones(num_channels))
```
h-mayorquin, line 1436: Don't we have set_channel_gains for this?

src/spikeinterface/core/numpyextractors.py
```diff
@@ -517,7 +522,8 @@ def __init__(self, snippets_list, spikesframes_list, sampling_frequency, nbefore
         )
 
         self._is_dumpable = False
```
h-mayorquin, line 524: I thought that "dumpable" was mean to be pickable? Can you remind me of how they are different?

src/spikeinterface/core/numpyextractors.py
```diff
@@ -127,7 +128,9 @@ def __init__(self, spikes, sampling_frequency, unit_ids):
         BaseSorting.__init__(self, sampling_frequency, unit_ids)
 
         self._is_dumpable = True
-        self._is_json_serializable = False
+        self._serializablility["json"] = False
+        # theorically this should be False but for simplicity make generators simples we still need this.
```
h-mayorquin, line 132: Yeah, you guys were talking about this today. I am not sure I got it.

You don't wnat stuff to be pickable because they are heavy. You put all the spike_vector into disk. Is that correct?

h-mayorquin:
Is there any case where saving and loading from json is better than pickling?

src/spikeinterface/core/base.py
samuelgarcia: It can help to make loop like:
```python
for type in ('json', 'pickle'):
   if check_serializablility(type):
      rec.dump(f'file.{type}')
```

Which should be convinient in various places.

src/spikeinterface/core/generate.py
samuelgarcia: Yep I forgot.
I do it.

src/spikeinterface/core/numpyextractors.py
samuelgarcia: yes or no.
dumpable is pickle for multiprocessing but absolutly want to avoid to dump silently NumpyRecording when extract waveforms.
`check_serializablility('pickle)` is clearly for the file version.

NumpyRecording is really the corner case we need to have in mind.

src/spikeinterface/core/numpyextractors.py
samuelgarcia: I wanted to avoid NumpySorting to also be pickle into file. But finally it is too hard!!
So lets keep `self._serializablility["pickle"] = True` for NumpySorting so the generated recording can be save into pickle which the most urgent need.

samuelgarcia:
> Is there any case where saving and loading from json is better than pickling?

We used json for human redability and so for debugging.
And also now pickle do not support anymore (and this is a big pain) the relative_to option which is super convinient.
so json still is a good option I think.

src/spikeinterface/core/tests/test_waveform_extractor.py
samuelgarcia: done

src/spikeinterface/core/base.py
samuelgarcia: We need 3 flags "dumpable", "json_serializable", "pickle_to_disk_srializable"
We could also remove self._is_dumpable and put it also this one in the dict.
This avoid duplication of the (actually buggy) function the recursivly find a non dumpable parent.
@h-mayorquin @alejoe91 : what do you think ?

src/spikeinterface/core/generate.py
samuelgarcia: done

samuelgarcia:
@h-mayorquin I made the change you proposed

src/spikeinterface/core/base.py
h-mayorquin: Yes, so I was saying that I prefered the code duplication (so you can't write your helpful loop) instead of merging these flags in a common dict and processing them in a single function. I don't like nesting even more the `check_if_dumpable` function which is already complicated. I think that is easier for you who has a good idea of what everything there means and where is it located and you want to save yourself a couple of lines of typing but I think that the costs are: 1) Is horrendous to read as a beginner, 2) It couples these two concepts (can be saved to json, can be saved to pickled) into one function and I don't think they should be mixed. [Code should have low-coupling](https://en.wikipedia.org/wiki/Coupling_(computer_programming))

I really think that the corner case of Numpy should be addressed as what it is, a corner case.If what you want to avoid is having stuff being dumped into pickle and json when saving in waveform extractor why not just filter them out right there in the waveform extractor instead of propagating these concepts to the whole library by introducing a new flag?

`can_be_pickled` `can_be_jsonfied` are simple flags that I think should transparently mean what they say in their name. If we believe that some specific functionality in the library should not rely on those mechanisms then we should throw a warning or an assertion in the specific place where the happens. 

It seems conceptually onerous and hard to document to create a whole new category for "thinks that will be sub-optimal in a specific function if done this way". We will soon forget why we introduce them in the first place, whereas if we put a filter where things work wrongly then the use case is kind of self-documented.

alejoe91 (approved):

PR #2037, opened 2023-09-24
Fix: Correct unit ID matching in sortingview curation

rkim48:
**Issue:**
In the process of applying sortingview curation for unit ids of dtype int, the function was using an "in" operator to match unit_id with unit_label, which resulted in false positives. For instance, a unit_id of 10 would mistakenly match a unit_label of 1.

**Fix:**
Enhanced the unit ID matching logic to handle both integer and string types properly, ensuring correct unit curation.
Added split logic for merged units to correctly identify individual unit parts. 

**Test introduced:**
Introduced a unit test test_false_positive_curation() in test_sortingview_curation.py to capture the false-positive scenario, ensuring unit_id 10 doesn't get labeled when unit_label 1 is specified. It also ensures correct curation for both int and string type unit IDs. 

**Additional information:**
I am using Mountainsort5.


src/spikeinterface/curation/tests/test_sortingview_curation.py
```diff
@@ -131,8 +133,46 @@ def test_json_curation():
...
+    print(f"Accepted unit IDs: {sorting_curated_ids[accept_idx]}")
+
+    # Check if unit_id 1 has received the "accept" label.
+    assert sorting_curated_json.get_unit_property(unit_id=1, key="accept")
+    # Check if unit_id 10 has received the "accept" label.
+    # If so, test fails since only unit_id 1 received the "accept" label in test_json.
+    assert not sorting_curated_json.get_unit_property(unit_id=10, key="accept")
+    print(sorting_curated_json.unit_ids)
+    # Merging unit_ids of dtype int creates a new unit id
+    assert 21 in sorting_curated_json.unit_ids
```
alejoe91, line 170: @rkim48 

In this case, do you agree that the unit `21` should receive the label `"artifact"`, inherited from `2`?

alejoe91:
Thanks @rkim48 

I have one comment. In case of str units, after a merge, the merged unit (e.g. ,`"8-9")  inherits the labels from individual units. This is because if unit `"8"` has a label, this line will be True:
```
if unit_label in unit_id_parts:

# "8" in ("8", "9"):
```

However, this is not the case for int units. My suggestion would be to convert all unit ids to strings when applying sortingview curation. Note that we have been discussing about forcing unit and channel IDs to always be strings: https://github.com/SpikeInterface/spikeinterface/issues/1432

src/spikeinterface/curation/tests/test_sortingview_curation.py
rkim48: Apologies, I'm not super familiar with intended behavior but I would assume so. In my latest commit, I made it so it can inherit both labels from unit 2 and unit 12.

rkim48:
@alejoe91 

Thanks for the feedback. Can you check my latest commit? 
I think it takes care of that issue but let me know if there is something I overlooked.

src/spikeinterface/curation/tests/test_sortingview_curation.py
```diff
@@ -130,9 +142,186 @@ def test_json_curation():
...
+    assert 9 not in sorting_exclude_noise.get_unit_ids()
+
+    # Assertions for include_labels
+    sorting_include_accept = apply_sortingview_curation(sorting, uri_or_json=json_path, include_labels=["accept"])
+    print(f"Include accept: {sorting_include_accept.get_unit_ids()}")
+    assert 8 not in sorting_include_accept.get_unit_ids()
+    assert 9 not in sorting_include_accept.get_unit_ids()
+    assert 10 in sorting_include_accept.get_unit_ids()
+
+    print("Test for integer unit IDs passed!\n")
```
zm711, line 247:
```suggestion
```
Are the prints necessary. If the tests pass with pytest they won't display the print anyway

src/spikeinterface/curation/tests/test_sortingview_curation.py
```diff
@@ -130,9 +142,186 @@ def test_json_curation():
...
+    assert "c-d" not in sorting_exclude_noise.get_unit_ids()
+
+    # Assertions for include_labels
+    sorting_include_accept = apply_sortingview_curation(sorting, uri_or_json=json_path, include_labels=["accept"])
+    print(f"Include accept: {sorting_include_accept.get_unit_ids()}")
+    assert "a-b" not in sorting_include_accept.get_unit_ids()
+    assert "c-d" not in sorting_include_accept.get_unit_ids()
+    assert "e-f" in sorting_include_accept.get_unit_ids()
+
+    print("Test for string unit IDs passed!\n")
```
zm711, line 317:
```suggestion
```
same here

src/spikeinterface/curation/tests/test_sortingview_curation.py
```diff
@@ -103,15 +110,20 @@ def test_sha1_curation():
     assert len(sorting_curated_sha1_mua.unit_ids) == 6
     assert len(sorting_curated_sha1_art_mua.unit_ids) == 5
 
+    print("Test for sha1 curation passed!\n")
```
zm711, line 113:
```suggestion
```
and here

src/spikeinterface/curation/tests/test_sortingview_curation.py
```diff
@@ -75,9 +77,14 @@ def test_gh_curation():
     assert len(sorting_curated_gh_mua.unit_ids) == 6
     assert len(sorting_curated_gh_art_mua.unit_ids) == 5
 
+    print("Test for GH passed!\n")
```
zm711, line 80:
```suggestion
```

and here.

zm711 (commented):
I don't think it is necessary to have `print('success')` statements in general for testing since in the case of success they will pass silently and in the case of failure they won't ever be triggered.

src/spikeinterface/curation/sortingview_curation.py
```diff
@@ -57,37 +57,55 @@ def apply_sortingview_curation(
...
-                new_unit_id = "-".join(mg)
+                new_unit_id = "-".join(merge_group)
+                curation_sorting.merge(merge_group, new_unit_id=new_unit_id)
             else:
                 # in this case, the CurationSorting takes care of finding a new unused int
-                new_unit_id = None
-            curation_sorting.merge(mg, new_unit_id=new_unit_id)
+                curation_sorting.merge(merge_group, new_unit_id=None)
+                new_unit_id = curation_sorting.max_used_id  # merged unit id
+                labels_dict[str(new_unit_id)] = labels_to_inherit
```
alejoe91, line 80: Probably the logic could be simplified a lot by:

1. Always setting `labels_dict[str(new_unit_id)] = labels_to_inherit` (both in case of int and str)

src/spikeinterface/curation/sortingview_curation.py
```diff
@@ -57,37 +57,55 @@ def apply_sortingview_curation(
...
+        labels_unit = set()
+
+        # Check for exact match first
+        if str(unit_id) in labels_dict:
+            labels_unit.update(labels_dict[str(unit_id)])
+        # If no exact match, check if unit_label is a substring of unit_id (for string unit ID merged unit)
+        else:
+            for unit_label, labels in labels_dict.items():
+                if isinstance(unit_id, str) and unit_label in unit_id:
+                    labels_unit.update(labels)
```
alejoe91, line 105: 2. cast `unit_id` to str here:
```
unit_id_str = str(unit_id)
```
3. Use ar line 99:
```
    if unit_id_str in labels_dict: # which now will also contain string versions of the int new_unit_ids
```

Since we added all merged units to the `labels_dict`, we can get rid of the `else` (L102-105). 
In fact, by construction, the `prop_values` have `len(unit_ids)` size, so when setting the property at L110 we don't care if the unit ids are ints or strings!

@rkim48 do you agree that this could be an easier implementation?

alejoe91 (changes requested):
@rkim48 thanks for the changes. I think that we could simplify a lot the implementation. 
See my comments (I hope they are clear!)

src/spikeinterface/curation/tests/test_sortingview_curation.py
```diff
@@ -130,9 +138,182 @@ def test_json_curation():
...
+    times = np.int_(np.sort(np.random.uniform(0, num_timepoints, num_spikes)))
+    labels = np.random.randint(1, num_units + 1, size=num_spikes)
+
+    sorting = se.NumpySorting.from_times_labels(times, labels, sampling_frequency)
+    print("Sorting: {}".format(sorting.get_unit_ids()))
+
+    # Test curation JSON:
+    test_json = {"labelsByUnit": {"1": ["accept"], "2": ["artifact"], "12": ["artifact"]}, "mergeGroups": [[2, 12]]}
+
+    json_path = "test_data.json"
```
alejoe91, line 163: Can you place this in `src/spikeinterface/curation/tests/` and remove it from gitignore? Otherwise CI tests will fail for sure!

src/spikeinterface/curation/tests/test_sortingview_curation.py
```diff
@@ -130,9 +138,182 @@ def test_json_curation():
...
+            "3": ["reject"],
+            "4": ["noise"],
+            "5": ["accept"],
+            "6": ["accept"],
+            "7": ["accept"],
+        },
+        "mergeGroups": [[1, 2], [3, 4], [5, 6]],
+    }
+
+    json_path = "test_curation_int.json"
```
alejoe91, line 207: Same here: this file will not be found

src/spikeinterface/curation/sortingview_curation.py
rkim48: Yes thank you

rkim48 (commented):
Requested changes have been made

src/spikeinterface/curation/sortingview_curation.py
```diff
@@ -57,37 +57,48 @@ def apply_sortingview_curation(
...
+    properties = {
+        label: np.zeros(len(curation_sorting.current_sorting.unit_ids), dtype=bool)
+        for labels in labels_dict.values()
+        for label in labels
+    }
+
+    # Populate the properties dictionary
+    for unit_index, unit_id in enumerate(curation_sorting.current_sorting.unit_ids):
+        unit_id_str = str(unit_id)
+        # Check for exact match first
```
alejoe91, line 97:
```suggestion
```
Not needed anymore :)

alejoe91 (approved):

Issue #2021, opened 2023-09-19
store path to data in phy export to relative instead of absolute path


Cumol:
When exporting to phy, the `params.py `file contains the path to the recording.dat file as an absolute path. The moment the folder is moved to another machine, phy fails to open as the path is wrong.

https://github.com/SpikeInterface/spikeinterface/blob/843b451282369565ca8e62863b4814c772c479f8/src/spikeinterface/exporters/to_phy.py#L152C18-L152C18

I wonder if it makes sense to turn it into a relative path. I tried it on one session and it resolved the issue. Unless that breaks some other functionality I am not aware of.

alejoe91:
Hi @Cumol 

Yes I think we should give the option to use relative paths with respect to the output folder. We have a similar option for the waveform extractor that we could propagate here. Thanks for the suggestion!

zm711:
Just in case people search here in the future how the `export_to_phy` behaves related to native KS behavior:

KS1 and KS2 supply just the name of the binary file in the `dat_path` 
KS2.5 and KS3 supply the absolute path in the `dat_path` (specifically to the `temp_wh.dat`.

zm711:
@Cumol, the option to use a relative path is now merged with #2041. To use you would need to install spikeinterface from source ([instructions](https://spikeinterface.readthedocs.io/en/latest/installation.html#from-source)) or wait for the next pypi release. The new keyword argument is `use_relative_path`. 

https://github.com/SpikeInterface/spikeinterface/blob/3826a2d8b5a02e53af597f568256fa27b20ee620/src/spikeinterface/exporters/to_phy.py#L68-L70

PR #2041, opened 2023-09-25
Allow relative path for `export_to_phy`

zm711:
This adds another argument to `export_to_phy` to allow for just the `recording.dat` to be given as the `dat_path` (ie the behavior found in KS1 and KS2) with a default of False (current SI behavior as well as the standard behavior of KS2.5 and KS3). 

For #2021.


h-mayorquin:
Will this have the same behavior as our `relative_to` @alejoe91 ?

https://github.com/SpikeInterface/spikeinterface/blob/ff1b8cdb59f0febb21354cb14a3ea2fc7577b202/src/spikeinterface/core/base.py#L306-L327

If that's the case, maybe the name and behavior should be the same?

Or, if it is different, maybe we should add a note in the docstring indicating how it behaves differently from `relative_to`.

zm711:
@alejoe91 just suggested propagating the idea from the `WaveformExtractor` in #2021.

https://github.com/SpikeInterface/spikeinterface/blob/ff1b8cdb59f0febb21354cb14a3ea2fc7f577b202/src/spikeinterface/core/waveform_extractor.py#L241-L251

So that's why I did that, but happy to change. The only change I made is to change the Path for the binary file for Phy to look for. I don't think it should actually affect any SpikeInterface machinery, but definitely correct me I'm happy to add more documentation or change the argument name.

alejoe91:
@zm711 @h-mayorquin let's use `use_relative_path` for functions exposed to the users. Note that it's different than `relative_to`, since the latter expects any Path, while `use_relative_path` is relative to a known folder (waveform folder for `WaveformExtractor`, phy folder for `export_to_phy`)

alejoe91 (approved):

src/spikeinterface/exporters/to_phy.py
```diff
@@ -64,6 +65,8 @@ def export_to_phy(
         Dtype to save binary data
     verbose: bool
         If True, output is verbose
+    use_relative_path : bool, default: False
```
h-mayorquin, line 68: Can you exapnd on this. When you say "saves the `dat_path` as a relative path, ... " 

I am kind of expecting that there is something after, relative to what?

h-mayorquin:
@alejoe91 @zm711 
I was not aware that we had this functionality / keyword in waveform extractor. I basically just was wondering how it was different and I wanted to have that in the documentation so I myself would not be confused about it and know how is different. But maybe the docstring is not the appropriate place for that to be.

A general question I have here is:
Would using `use_relative_path=True` as a default cause any other incovenience. If I understand, the user passes a folder (absolute) and all the stuff that we put in that folder is relative to that folder. If that is the case, I am wonderingi f there is any downside to just use relative by default as the only thing that we expose is the folder.

zm711:
@h-mayorquin 
Here's a copy of the phy folder (a recent running with the new synchrony thanks @alejoe91 :) )
image

So the way phy works is it needs to extract waveforms from a binary file. In this case we have the binary file `recording.dat`. The absolute path is given as (on my Windows :( a `r'c:\user\zm\...\recording.dat'`), but this means that if I go to move this file to our storage server the absolute `dat_path` will no longer work with Phy since it won't find the `recording.dat` in the right place. Relative just means I save it as `dat_path = 'recording.dat'` so when I try to activate Phy it just looks in the same folder as the `params.py` file. Does that make sense? One big annoyance for the relative version of the file is that I need a copy of the `recording.dat` in the folder which means I need to make another copy (another big file eating up storage).

I'll let @alejoe91 answer:
> Would using use_relative_path=True as a default cause any other incovenience. If I understand, the user passes a folder (absolute) and all the stuff that we put in that folder is relative to that folder. If that is the case, I am wonderingi f there is any downside to just use relative by default as the only thing that we expose is the folder.

src/spikeinterface/exporters/to_phy.py
zm711: @h-mayorquin. Great point, that is a lazy docstring. I'll fix that shortly just got to lab so need to set up for the day.

alejoe91:
@h-mayorquin As we move more into server/cloud/remote computing patforms, I think it makes total sense to make `use_relative_path=True` by default.

alejoe91 (changes requested):
@zm711 I noticed one thing. In case `copy_binary` is `False`, but the recording is already binary, we should get the relative path to the phy output folder. Right?

src/spikeinterface/exporters/to_phy.py
```diff
@@ -149,7 +152,10 @@ def export_to_phy(
 
     # write params.py
     with (output_folder / "params.py").open("w") as f:
-        f.write(f"dat_path = r'{str(rec_path)}'\n")
+        if use_relative_path:
+            f.write(f"dat_path = r'recording.dat'\n")
+        else:
+            f.write(f"dat_path = r'{str(rec_path)}'\n")
```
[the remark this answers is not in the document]
zm711:
```suggestion
   if use_relative_path:
            if copy_binary:
                f.write(f"dat_path = r'recording.dat'\n")
            else:
                f.write(f"dat_path = r'{str(Path(rec_path).relative_to(output_folder ))}'\n")
        else:
            f.write(f"dat_path = r'{str(rec_path)}'\n")
```
I was just thinking about that, but I'm not as up to speed with `relative_to`. I amended your fix to use an f-string does that look right?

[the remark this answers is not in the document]
zm711: @alejoe91

src/spikeinterface/exporters/to_phy.py
```diff
@@ -64,6 +65,8 @@ def export_to_phy(
         Dtype to save binary data
     verbose: bool
         If True, output is verbose
+    use_relative_path : bool, default: False
+        If True saves the `dat_path` as a relative path, else an absolute
```
zm711, line 69:
```suggestion
        If True saves the binary file `dat_path` in the `params.py` relative to 'output_folder', if False, uses an absolute path
```

@h-mayorquin is this better? Or do you want something more substantial?

src/spikeinterface/exporters/to_phy.py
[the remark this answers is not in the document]
zm711: Your edit didn't show for me, but I see it now. Cool.

src/spikeinterface/exporters/to_phy.py
h-mayorquin: This makes sense to me. Maybe you could add the actual example:

```python
        If True saves the binary file `dat_path` in the `params.py` relative to 'output_folder' (i.e. "looks like this", if False, uses an absolute path (i.,e. "looks like this").
```

h-mayorquin: Also, maybe mention how it interats with `copy_binary` as it was described above?

zm711:
```suggestion
        If True saves the binary file `dat_path` in the `params.py` relative to 'output_folder' (ie `dat_path=r'recording.dat'`)
        If False, uses an absolute path in the `params.py` (ie `dat_path=r'path/to/the/recording.dat'`)
```

h-mayorquin:
@zm711 Thanks a lot for the explanation about how phy works at that level. This was very useful for me.

h-mayorquin:
Any way we can test this by the way?

src/spikeinterface/exporters/to_phy.py
zm711: Missed your last comment. Will add now.

zm711:
phy does have a command something like `template-describe` that won't activate the gui. So maybe that way. I can do some research?

src/spikeinterface/exporters/to_phy.py
zm711:
```suggestion
        If True and `copy_binary=True` saves the binary file `dat_path` in the `params.py` relative to `output_folder` (ie `dat_path=r'recording.dat'`). If `copy_binary=False`, then uses a path relative to the `output_folder`
        If False, uses an absolute path in the `params.py` (ie `dat_path=r'path/to/the/recording.dat'`)
```

The only thing I'm not sure of is what a good (not complicated) path example for copy_binary=False, and use_relative=True. 
But how does this look for now @h-mayorquin?

zm711:
So here is the test they use on phy for `template-describe`
https://github.com/cortex-lab/phy/blob/8bf8cf751fcbd44210b3c5bbdba8c34a7b2cb1ce/phy/apps/template/tests/test_gui.py#L39-L43

src/spikeinterface/exporters/to_phy.py
h-mayorquin: Yeah, looks better to me. I don't have ideas on how to provide an example path. Thanks for improving this.

zm711:
I think what we could do is actually run the test using phylib. We can read in all the `export_to_phy` into the `model` in phylib and create a test based on that. `phylib` should be all the calculations etc without any issues with the gui. I don't know if I have the bandwidth for that this week, but maybe I could draft something soon and then get your feedback later @h-mayorquin ?

alejoe91:
Yeah I wouldn't start messing with qt dependencies in tests. Let's leave it as is for now and maybe open an issue to add phylib tests later on!

alejoe91 (approved):

src/spikeinterface/exporters/to_phy.py
```diff
@@ -149,7 +153,13 @@ def export_to_phy(
 
     # write params.py
     with (output_folder / "params.py").open("w") as f:
-        f.write(f"dat_path = r'{str(rec_path)}'\n")
+        if use_relative_path:
+            if copy_binary:
+                f.write(f"dat_path = r'recording.dat'\n")
+            else:
+                f.write(f"dat_path = r'{str(Path(rec_path).relative_to(output_folder))}'\n")
```
samuelgarcia, line 160:
```suggestion
            if copy_binary:
                f.write(f"dat_path = r'recording.dat'\n")
            elif rec_path == "None":
                 f.write(f"dat_path = {rec_path}\n")               
            else:
                f.write(f"dat_path = r'{str(Path(rec_path).relative_to(output_folder))}'\n")
```
we need this case.

zm711: But if `rec_path` is None and there is no binary file to extract from what is the point of using Phy? (It's a real question--I've never tried Phy without access to some raw data). Or does Phy interpret that differently and still works? @samuelgarcia

alejoe91: Oh you can still see a ton of cool visuals! Templates, amplitude plots, firing rates, auto/cross-corr, pca scores, similarities, etc.

alejoe91: we do allow `rec_path` to be `"None"` in the params.py [here](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/exporters/to_phy.py#L130-L146), so we should also consider this case.

zm711: > Oh you can still see a ton of cool visuals! Templates, amplitude plots, firing rates, auto/cross-corr, pca scores, similarities, etc.

That's cool! I'll check it out sometime for fun.

> we do allow rec_path to be "None" in the params.py here, so we should also consider this case.

That's true. I'm wondering if there should be some sort of warning then. I think the average user (or maybe I mean the more novice user), would only want to use Phy for curation and not visualization per se and so maybe a warning in the case of `rec_path=None` would let them know that exporting to Phy without a recording is different than with a recording. (I for example only use Phy for manual curation and would never turn it on just to visualize metrics other then to test to see what it looks like based on this discussion).

alejoe91: You can still curate without the traces. You won't be able to see the TraceView and the single waveforms, but all other features are there. Note that phy will throw a warning itself if it doesn't find the recording

zm711: Okay. Sounds good to me. I'll push Sam's change.

zm711:
@alejoe91 The test failure is a problem with `compute_synchrony_metrics`

```
for synchrony_size in synchrony_sizes:
>                   synchrony_counts[synchrony_size][unit_id] += np.count_nonzero(spike_complexity >= synchrony_size)
E                   IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

It's also occurred in `export_report` and not in `export_to_phy`

alejoe91 (approved):

alejoe91:
@zm711 let me fix the failing test real quick! sorry about that

zm711:
No worries. :) Better to find failing tests sooner rather than later. Thanks for the fix.

Issue #2040, opened 2023-09-25
Add `ignore_timestamps_errors` argument to `OpenEphysLegacyRecordingExtractor`


weiglszonja:
Add `ignore_timestamps_errors` argument to `OpenEphysLegacyRecordingExtractor`
This argument was added to `OpenEphysRawIO` in https://github.com/NeuralEnsemble/python-neo/pull/1213 but in order to utilise the fix we need this argument for the extractor as well. 

I can open a PR for this.

alejoe91:
@weiglszonja it would be great if you can make a PR for this!

PR #2043, opened 2023-09-26
Add `ignore_timestamps_errors` to `OpenEphysLegacyRecordingExtractor`

weiglszonja:
Resolve https://github.com/SpikeInterface/spikeinterface/issues/2040
Related to https://github.com/NeuralEnsemble/python-neo/pull/1213


alejoe91:
@weiglszonja note that NEO needs a new release and this is not included in v0.12. 

Can you check NEO version and only map the argument if version >0.12 ?

weiglszonja:
@alejoe91 Thank you, good point. I added a function for this based on https://github.com/SpikeInterface/spikeinterface/blob/cc66afa8a40a3ec782fb4819b071e4f9c02f5025/src/spikeinterface/extractors/neoextractors/mearec.py#L11

PR #2044, opened 2023-09-26
Make match count, add testing

h-mayorquin:
I am pushing as me and @alejoe91 worked on it. Just to keep it for reference. It is on draft.

### DIFF
--- a/src/spikeinterface/comparison/comparisontools.py
+++ b/src/spikeinterface/comparison/comparisontools.py
@@ -29,27 +29,49 @@ def count_matching_events(spikes_frames1, spikes_frames2, delta=10):
     combined_frames = np.concatenate((spikes_frames1, spikes_frames2))
 
     # Assign labels: 1 for spikes_frames1 and 2 for spikes_frames2
-    labels = np.concatenate((np.ones(spikes_frames1.shape) * 1, np.ones(spikes_frames2.shape) * 2))
+    labels = np.concatenate(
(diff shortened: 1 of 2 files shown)


samuelgarcia:
@alejoe91 @h-mayorquin : I think the main speedup is not to improve the `def count_matching_events(times1, times2, delta=10)` function but to rewrite the `make_match_count_matrix` which is quadratic loop on `count_matching_events` to get directly the count matrix without the `count_matching_events`.
In short do a bit like the correlogram numba approach for all the units at once without a double limits algo for all spikes at once. This should be even more easy with the spikes vector in mind I think.

samuelgarcia:
Another important stuff to avoid to spent too much time on corner cases : the delta 0.4ms is lower than the refactory period so for good units the number of match should be unambiguous. For units that have duplicate or many refactory period violation, the matching is anyway super ambiguous.

h-mayorquin:
> Another important stuff to avoid to spent too much time on corner cases : the delta 0.4ms is lower than the refactory period so for good units the number of match should be unambiguous. For units that have duplicate or many refactory period violation, the matching is anyway super ambiguous.

@samuelgarcia 
If you take this position, would you be up to remove the check for duplicates?

https://github.com/SpikeInterface/spikeinterface/blob/243a30c0f21240d370ffd610c5b762a186ecf351/src/spikeinterface/comparison/comparisontools.py#L36-L37C24

PR #2045, opened 2023-09-26
Svd convolutions

yger:
Improve the matching engine of circus2, using an SVD low rank decomposition for computing the convolutions. faster and smotthing the templates more efficiently


yger:
This branch will only affects SC2, and could be merged asap.

alejoe91 (approved):

PR #2048, opened 2023-09-28
CellExplorer: fix tests and deprecation

alejoe91:
We had a deprecation period in the `CellExplorerSorting` which has expired.

This PR updates the tests and raises an exception ins tead of a warning.


h-mayorquin (approved):
Thanks.
Can you drop it from the __init__ as well,

alejoe91:
> Thanks. Can you drop it from the **init** as well,

done

PR #2053, opened 2023-09-29
Fixes Apptainer bug plus small fixes

mhhennig:
Main change: change to `pip install --user` for sorters run in containers so python packages are installed in writable location. This otherwise breaks apptainer.

There are a few small changes, sorry they all end up in the same pull request. There were some type issues in the mda and numpy extractors.


alejoe91:
Thanks @mhhennig !!

Yeah no worries about the small changes this time! They are smalle enough :)

src/spikeinterface/extractors/mdaextractors.py
```diff
@@ -216,10 +216,14 @@ def write_sorting(sorting, save_path, write_primary_channels=False):
         times_list = []
         labels_list = []
         primary_channels_list = []
-        for unit_id in unit_ids:
+        for unit_id_i, unit_id in enumerate(unit_ids):
```
samuelgarcia, line 219: use unit_index to be align with other modules

alejoe91:
> Hi Matthias. What are the issue with numpy extractors ?

Seems only MdaSorting

src/spikeinterface/extractors/mdaextractors.py
h-mayorquin: Thanks Sam!

mhhennig: If unit IDs are non-integer (which I've had), this would fail unless an index is created (type problem later on).

mhhennig:
> > Hi Matthias. What are the issue with numpy extractors ?
> 
> Seems only MdaSorting

NumpyExtractor: `assert times.dtype.kind == 'i', 'numpy array of spike times must be integer'`
We must check for any integer type (failed for me for a sorting I had), so needs to be:
`assert (times.dtype.kind == 'i') or (times.dtype.kind == 'u'), 'numpy array of spike times must be integer'`

src/spikeinterface/extractors/mdaextractors.py
mhhennig: Hang on, this would discard unit_ids if they are set in non-increasing manner. MDA requires integers, so if there are numbers we should use these (HDSort for example produces strange unit numbers, and I like to save stuff intio mda, but want to retain the original ids).

src/spikeinterface/extractors/mdaextractors.py
```diff
@@ -216,10 +216,14 @@ def write_sorting(sorting, save_path, write_primary_channels=False):
...
             times = sorting.get_unit_spike_train(unit_id=unit_id)
             times_list.append(times)
-            labels_list.append(np.ones(times.shape) * unit_id)
+            # unit id may not be numeric
+            if unit_id.dtype.kind in "biufc":
+                labels_list.append(np.ones(times.shape) * unit_id)
```
samuelgarcia, line 224: unit_id is int or str only. float are not allowed

src/spikeinterface/extractors/mdaextractors.py
alejoe91: Have you tried sorting.save(folder=...) 
? That's what we use and we recommend!

src/spikeinterface/extractors/mdaextractors.py
mhhennig: I seem to recall I've had this (float unit ids). How to deal with that, an assert here or just allow it?

alejoe91: No by construction they can only be int or str. Otherwise an error is raised in base

alejoe91:
@mhhennig to keep the repo clean, we all make PRs from forks, and not internal branches. Could you also do it next time? ;)

src/spikeinterface/extractors/mdaextractors.py
alejoe91, line 219:
```suggestion
        for unit_index, unit_id in enumerate(unit_ids):
```

src/spikeinterface/extractors/mdaextractors.py
```diff
@@ -216,10 +216,14 @@ def write_sorting(sorting, save_path, write_primary_channels=False):
...
+            else:
+                labels_list.append(np.ones(times.shape) * unit_id_i)
```
alejoe91, line 226:
```suggestion
                labels_list.append(np.ones(times.shape) * unit_index)
```

alejoe91:
> @mhhennig to keep the repo clean, we all make PRs from forks, and not internal branches. Could you also do it next time? ;)

This is what I mean:
image

src/spikeinterface/extractors/mdaextractors.py
```diff
@@ -216,10 +216,14 @@ def write_sorting(sorting, save_path, write_primary_channels=False):
         times_list = []
         labels_list = []
         primary_channels_list = []
-        for unit_id in unit_ids:
+        for unit_id_i, unit_id in enumerate(unit_ids):
             times = sorting.get_unit_spike_train(unit_id=unit_id)
             times_list.append(times)
-            labels_list.append(np.ones(times.shape) * unit_id)
+            # unit id may not be numeric
+            if unit_id.dtype.kind in "biufc":
```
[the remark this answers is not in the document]
alejoe91: actually you're right, int or str is currently not enforced..but we should!

src/spikeinterface/extractors/mdaextractors.py
alejoe91: this was just a naming issue

alejoe91:
> > > Hi Matthias. What are the issue with numpy extractors ?
> > 
> > 
> > Seems only MdaSorting
> 
> NumpyExtractor: `assert times.dtype.kind == 'i', 'numpy array of spike times must be integer'` We must check for any integer type (failed for me for a sorting I had), so needs to be: `assert (times.dtype.kind == 'i') or (times.dtype.kind == 'u'), 'numpy array of spike times must be integer'`

makes sense, will push the fix here: #2061

alejoe91 (approved):

Issue #2046, opened 2023-09-27
Documentation Proposals


zm711:
As I was looking over the docs I was thinking of two changes that I think would be beneficial, but since they would be a bit of work to implement (I'm happy to do 1 or 2 PRs for this), I wanted to make sure no one was absolutely against them.

1) Switch import statements to follow module aliases, for example in the curation module docs it says:
```python
from spikeinterface.curation import MergeUnitsSorting, get_potential_auto_merge

sorting = run_sorter('kilosort', recording)

we = extract_waveforms(recording, sorting, folder='wf_folder')

# merges is a list of lists, with unit_ids to be merged.
merges = get_potential_auto_merge(we, minimum_spikes=1000,  maximum_distance_um=150.,
                                  peak_sign="neg", bin_ms=0.25, window_ms=100.,
                                  corr_diff_thresh=0.16, template_diff_thresh=0.25,
                                  censored_period_ms=0., refractory_period_ms=1.0,
                                  contamination_threshold=0.2, num_channels=5, num_shift=5,
                                  firing_contamination_balance=1.5)

# here we apply the merges
clean_sorting = MergeUnitsSorting(sorting, merges)
```

I would propose instead to use
```python
import spikeinterface as si
import spikeinterface.curation as scur
import spikeinterface.sorters as ss

sorting = ss.run_sorter('kilosort', recording)
we = si.extract_waveforms(recording, sorting, folder = 'wf_folder')
...
clean_sorting = scur.MergeUnitSorting(sorting, merges)
```
Although an advanced user can import whatever they want, I think helping beginners to use the namespace would be beneficial. And in the docs seeing that `run_sorter` is part of `ss` which is the alias of `sorters` will help them find their way around the documentation by knowing which module to go to if they aren't sure where a function came from.

2) The documentation makes extensive usage of positional arguments rather than keyword arguments. I think adding in keyword arguments would help users understand what they are putting into a function and reduce the risk of putting the wrong data into a function. in the same example I would add

```python
sorting = ss.run_sorter(sorter_name='kilosort', recording=recording)
we = si.extract_waveforms(recording=recording, sorting=sorting, folder='wf_folder')
...
```

This is a bit more verbose, but it encourages good function practices and if something goes wrong it will help them generate issues on the tracker/ search through the source code to see what is going on. 

But do let me know if anyone is completely against adding these features to the docs. Otherwise I'll get started working on some of those updates when I have some free time.

[...]

Issue #1907, opened 2023-08-03
Unexpected keyword 'radius_um' in Neuropixel Guide


nmtimme:
Hello

I'm trying to go through the guide to using spike interface to spike sort neuropixel data (https://spikeinterface.readthedocs.io/en/latest/how_to/analyse_neuropixels.html), but I'm running into an error with detect_peaks. Thank you for quickly addressing my issue with plot_traces  #1899 !

I'm running version 0.98.2 on a Linux cluster with a GPU. I'm running on a virtual desktop, so I can view plots and interact with GUIs. I installed spikeinterface using miniconda 4.12.0 using this .yml file:

```
name: si_env
channels:
  - conda-forge
  - defaults
dependencies:
  - python=3.9
  - pip>=21.0
  - mamba
  # numpy 1.22 break numba which break tridesclous
  - numpy<1.22
  # joblib 1.2 is breaking hdbscan
  - joblib=1.1
  - tqdm
  - matplotlib
  - h5py
  - pandas
  - scikit-learn
  - networkx
  - pybind11
  - loky
  - hdbscan
  - numba
  - jupyter
  - mpi4py
  - cxx-compiler
  - libxcb
  - pyqt=5
  - pyqtgraph
  - htop
  - ipywidgets
  - ipympl
  - pip:
    - ephyviewer
    - neo>=0.11
    - elephant>=0.10.0
    - probeinterface>=0.2.11
    - MEArec>=1.8
    - spikeinterface[full, widgets]
    - spikeinterface-gui
    - tridesclous>=1.6.6.1
    - spyking-circus>=1.1.0
    # - phy==2.0b5
```

I am running the following script when I encounter the error:

```
import spikeinterface.full as si
import spikeinterface.widgets as siw
print(si.__version__)

import numpy as np
import matplotlib.pyplot as plt
from pathlib import Path

# Define the path to the data
base_folder = Path('/N/project/lapishLabWorkspace/NickPilotNeuropixelEphysData')
openephys_folder = base_folder / '2023-03-15_15-18-59'

# Figure out the data streams in this data set
stream_names, stream_ids = si.get_neo_streams('openephys', openephys_folder)
print('stream_names',stream_names)

# Read the data LFP data from probe A (this automatically loads the probe map)
raw_rec = si.read_openephys(openephys_folder, stream_name='Record Node 101#Neuropix-PXI-100.ProbeA-AP')
print('raw_rec',raw_rec)

[...]

Issue #1899, opened 2023-07-31
plot_traces error in neuropixel guide


nmtimme:
Hello

I'm trying to go through the guide to using spike interface to spike sort neuropixel data (https://spikeinterface.readthedocs.io/en/latest/how_to/analyse_neuropixels.html), but I'm running into an error with plot_traces not being an attribute of spikeinterface.

I'm running version 0.98.2 on a Linux cluster with a GPU. I'm running on a virtual desktop, so I can view plots and interact with GUIs. I installed spikeinterface using miniconda 4.12.0 using this .yml file:

```
name: si_env
channels:
  - conda-forge
  - defaults
dependencies:
  - python=3.9
  - pip>=21.0
  - mamba
  # numpy 1.22 break numba which break tridesclous
  - numpy<1.22
  # joblib 1.2 is breaking hdbscan
  - joblib=1.1
  - tqdm
  - matplotlib
  - h5py
  - pandas
  - scikit-learn
  - networkx
  - pybind11
  - loky
  - hdbscan
  - numba
  - jupyter
  - mpi4py
  - cxx-compiler
  - libxcb
  - pyqt=5
  - pyqtgraph
  - htop
  - ipywidgets
  - ipympl
  - pip:
    - ephyviewer
    - neo>=0.11
    - elephant>=0.10.0
    - probeinterface>=0.2.11
    - MEArec>=1.8
    - spikeinterface[full, widgets]
    - spikeinterface-gui
    - tridesclous>=1.6.6.1
    - spyking-circus>=1.1.0
    # - phy==2.0b5
```

I am running the following script when I encounter the error:

```
import spikeinterface.full as si
print(si.__version__)

import numpy as np
import matplotlib.pyplot as plt
from pathlib import Path

# Define the path to the data
base_folder = Path('/N/project/lapishLabWorkspace/NickPilotNeuropixelEphysData')
openephys_folder = base_folder / '2023-03-15_15-18-59'

# Figure out the data streams in this data set
stream_names, stream_ids = si.get_neo_streams('openephys', openephys_folder)
print('stream_names',stream_names)

# Read the data LFP data from probe A (this automatically loads the probe map)
raw_rec = si.read_openephys(openephys_folder, stream_name='Record Node 101#Neuropix-PXI-100.ProbeA-AP')
print('raw_rec',raw_rec)

# Plot the bottom part of the channel map
fig, ax = plt.subplots(figsize=(15, 10))
[...]

PR #2057, opened 2023-09-29
Add Keywords to the Documentation Functions

zm711:
As discussed in #2046. 

I didn't really touch the imports ( I only added imports in a couple spots where it was missing). Once we settle on that I'll just do that in a separate PR.

### DIFF
--- a/doc/modules/curation.rst
+++ b/doc/modules/curation.rst
@@ -24,21 +24,21 @@ The merging and splitting operations are handled by the :py:class:`~spikeinterfa
 
     from spikeinterface.curation import CurationSorting
 
-    sorting = run_sorter('kilosort2', recording)
+    sorting = run_sorter(sorter_name='kilosort2', recording=recording)
 
-    cs = CurationSorting(sorting)
+    cs = CurationSorting(parent_sorting=sorting)
 
     # make a first merge
-    cs.merge(['#1', '#5', '#15'])
+    cs.merge(units_to_merge=['#1', '#5', '#15'])
 
     # make a second merge
-    cs.merge(['#11', '#21'])
+    cs.merge(units_to_merge=['#11', '#21'])
 
     # make a split
     split_index = ... # some criteria on spikes
-    cs.split('#20', split_index)
+    cs.split(split_unit_id='#20', indices_list=split_index)
 
-    # here the final clean sorting
+    # here is the final clean sorting
     clean_sorting = cs.sorting
 
 
@@ -60,12 +60,12 @@ merges. Therefore, it has many parameters and options.
 
     from spikeinterface.curation import MergeUnitsSorting, get_potential_auto_merge
 
-    sorting = run_sorter('kilosort', recording)
+    sorting = run_sorter(sorter_name='kilosort', recording=recording)
 
-    we = extract_waveforms(recording, sorting, folder='wf_folder')
+    we = extract_waveforms(recording=recording, sorting=sorting, folder='wf_folder')
 
     # merges is a list of lists, with unit_ids to be merged.
-    merges = get_potential_auto_merge(we, minimum_spikes=1000,  maximum_distance_um=150.,
+    merges = get_potential_auto_merge(waveform_extractor=we, minimum_spikes=1000,  maximum_distance_um=150.,
                                       peak_sign="neg", bin_ms=0.25, window_ms=100.,
                                       corr_diff_thresh=0.16, template_diff_thresh=0.25,
                                       censored_period_ms=0., refractory_period_ms=1.0,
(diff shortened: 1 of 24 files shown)


doc/modules/curation.rst
```diff
@@ -24,21 +24,21 @@ The merging and splitting operations are handled by the :py:class:`~spikeinterfa
 
     from spikeinterface.curation import CurationSorting
 
-    sorting = run_sorter('kilosort2', recording)
+    sorting = run_sorter(sorter_name='kilosort2', recording=recording)
 
-    cs = CurationSorting(sorting)
+    cs = CurationSorting(parent_sorting=sorting)
```
zm711, line 29: Is there a reason for this? Why do some functions use `parent_sorting` or `parent_recording` and most use `sorting` or `recording`.

doc/modules/exporters.rst
```diff
@@ -28,15 +28,14 @@ The input of the :py:func:`~spikeinterface.exporters.export_to_phy` is a :code:`
...
 
     # the waveforms are sparse so it is faster to export to phy
-    folder = 'waveforms'
-    we = extract_waveforms(recording, sorting, folder, sparse=True)
+    we = extract_waveforms(recording=recording, sorting=sorting, folder='waveforms', sparse=True)
 
     # some computations are done before to control all options
-    compute_spike_amplitudes(we)
-    compute_principal_components(we, n_components=3, mode='by_channel_global')
+    compute_spike_amplitudes(waveform_extractor = we)
```
alejoe91, line 34:
```suggestion
    compute_spike_amplitudes(waveform_extractor=we)
```

doc/modules/extractors.rst
```diff
@@ -27,40 +28,52 @@ Every format can be read with a simple function:
...
+Although most recordings are loaded with the :py:mod:`~spikeinterface.extractors`
+a few file formats are loaded from the :py:mod:`~spikeinterface.core` module
+
+.. code-block:: python
+
+    import spikeinterface as si
+
+    recording_binary = si.read_binary(file_path='binary.bin')
+
+    recording_zarr = si.read_zarr(file_path='zarr_file.zarr')
```
alejoe91, line 60: @zm711 note that recently I exposed the `read_binary` and `read_zarr` functions to the extractors mofule too. (https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/extractors/extractorlist.py#L14-L15)

I think we can remove this paragrapth for sake of clarity, i.e., all `read_*` functions are in `extractors`. What do you think?

doc/modules/sortingcomponents.rst
```diff
@@ -183,15 +184,15 @@ Here is an example with non-rigid motion estimation:
 .. code-block:: python
 
     from spikeinterface.sortingcomponents.peak_detection import detect_peaks
-    peaks = detect_peaks(recording, ...)
+    peaks = detect_peaks(recording=ecording, ...) # as in above example
```
alejoe91, line 187:
```suggestion
    peaks = detect_peaks(recording=recording, ...) # as in above example
```

alejoe91 (changes requested):
Thank you so much @zm711 

Just spotted a couple of small typos and have one comment!

doc/modules/extractors.rst
zm711: I didn't realize that (Thanks :)). The autofunction is connected to the source, so I thought maybe it was in core. I'll delete (it is better if they are all extractor).

zm711, line 60:
```suggestion
```

alejoe91 (approved):

doc/modules/curation.rst
zm711: Yeah, I think the path of least resistance and the most obvious for the user is just to use `recording` and `sorting`. I don't think I would guess that I have a `parent_sorting` vs `child_sorting` at the api level for most functions. (Also less typing).

alejoe91: Agreed. It's also a good time to make the switch. Do you think we should deprecate the old arguments or just break compatibility towards the next release?

zm711: Based on the recent issues with functions and arguments (#1907, #1899, #1678) it might be a better idea to deprecate for at least one release (although tearing off the band-aid would be cleaner like in #1865, which I would prefer), it seems like that change caused a bit of confusion for users.

alejoe91: well this is only changing an argument name that it's probably not really used by most users. I think that most users will use `banpass_filter(recording)` and not `bandpass_filter(parent_recording=recording)` (same for `channel_slice`, other preprocessors, etc..). Keeping back-compatibility would mean adding a temporary argument and setting the main `recording` to None, which would possibly require to make some other args None by default...

Let's just make the switch. I checked and it's not that many files :)

PR #2061, opened 2023-10-02
Check that main_ids are str or int

alejoe91:
Came up in #2053. We always assume channel/unit ids are str or ints, but we have no explicit check for it in main. 

This PR fixes it.


src/spikeinterface/extractors/cellexplorersortingextractor.py
```diff
@@ -118,7 +118,7 @@ def __init__(
         spike_times = spikes_data["times"]
 
         # CellExplorer reports spike times in units seconds; SpikeExtractors uses time units of sampling frames
-        unit_ids = unit_ids[:].tolist()
+        unit_ids = unit_ids[:].astype(int).tolist()
```
h-mayorquin, line 121: Should we do strings to start moving in that direction?
```suggestion
        unit_ids = unit_ids[:].astype(int).tolist()
        unit_ids = [str(unit_id) for unit_id in unit_ids]
```

src/spikeinterface/extractors/cellexplorersortingextractor.py
```diff
@@ -118,7 +118,8 @@ def __init__(
...
+        unit_ids = [str(unit_id) for unit_id in unit_ids]
```
samuelgarcia, line 122: Why are we going to str in this extractor ?
Why do we need int and then str and why not directly str ?

alejoe91: right, we only need the second line

alejoe91: done

samuelgarcia (approved):

PR #2070, opened 2023-10-04
Adjust eps for whitening in case of very small magnitude data

### LINKED ISSUE #2064: in whiten.py, eps is too large for some datasets
When the magnitude of the recording traces is very low (e.g. ~ 1e-6 or lower), the eps used in the whiten preprocessing is too large. This results in a whitened signal that is not normalized properly, and detection for mountainsort (and perhaps other algorithms) then fails.

https://github.com/SpikeInterface/spikeinterface/blob/ca48f6b62610e780861dea428faf817631561055/src/spikeinterface/preprocessing/whiten.py#L70-L72

Here is a dataset for which the raw traces are float and low magnitude
https://dandiarchive.org/dandiset/000463/draft/files?location=sub-BH395

The simplest fix is to change this to 1e-16. Not sure if this will introduce any other issues.

magland:
Resolves #2064 

I explain what is happening in the comments

I ran the following test:

```python
import numpy as np

# toy example
recording, sorting = se.toy_example(num_channels=4, duration=10, seed=0)
recording_whitened = spre.whiten(recording)
traces = recording_whitened.get_traces(segment_index=0)
print(np.var(traces, axis=0))

recording2 = spre.scale(recording, gain=1e-8)
recording2_whitened = spre.whiten(recording2)
traces2 = recording2_whitened.get_traces(segment_index=0)
print(np.var(traces2, axis=0))
```

And got the expected variances close to 1
```bash
[1.0054402  1.0075228  0.9955653  0.96718967]
[0.99692386 0.99832404 0.98711884 0.9610572 ]
```

And without the adjustment the second array is very wrong
```bash
[1.0054402  1.0075228  0.9955653  0.96718967]
[1.1976793e-08 1.1291137e-08 1.2642105e-08 1.6557108e-08]
```

### DIFF
--- a/src/spikeinterface/preprocessing/whiten.py
+++ b/src/spikeinterface/preprocessing/whiten.py
@@ -171,11 +171,11 @@ def compute_whitening_matrix(recording, mode, random_chunk_kwargs, apply_mean, r
     # Typically we can assume that data is in units of
     # microvolts, but this is not always the case. When data
     # is float type and scaled down to very small values, then the
-    # default eps=1e-6 can be too large, resulting in incorrect
+    # default eps=1e-8 can be too large, resulting in incorrect
     # whitening. We therefore check to see if the data is float
     # type and we estimate a more reasonable eps in the case
     # where the data is on a scale less than 1.
-    eps = 1e-6 # the default
+    eps = 1e-8 # the default
     if data.dtype.kind == "f":
         median_data_sqr = np.median(data ** 2) # use the square because cov (and hence S) scales as the square
         if median_data_sqr < 1 and median_data_sqr > 0:
--- a/src/spikeinterface/preprocessing/whiten.py
+++ b/src/spikeinterface/preprocessing/whiten.py
@@ -68,7 +68,7 @@ class WhitenRecording(BasePreprocessor):
                 M = np.asarray(M)
         else:
             W, M = compute_whitening_matrix(
-                recording, mode, random_chunk_kwargs, apply_mean, radius_um=radius_um, eps=1e-8
+                recording, mode, random_chunk_kwargs, apply_mean, radius_um=radius_um
             )
 
         BasePreprocessor.__init__(self, recording, dtype=dtype_)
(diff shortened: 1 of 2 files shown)


src/spikeinterface/preprocessing/whiten.py
```diff
@@ -167,6 +165,20 @@ def compute_whitening_matrix(recording, mode, random_chunk_kwargs, apply_mean, r
     cov = data.T @ data
     cov = cov / data.shape[0]
 
+    # Here we determine eps used below to avoid division by zero.
+    # Typically we can assume that data is in units of
+    # mV, but this is not always the case. When data
```
alejoe91, line 170:
```suggestion
    # Typically we can assume that data is either unscaled integers or in units of
    # uV, but this is not always the case. When data
```

alejoe91 (approved):

magland:
@alejoe91 I have invited you to my fork of spikeinterface.

magland:
@alejoe91 It seems that the implementation does not quite agree with the doc string. It says

> If None, eps is estimated from the data.

But then the code says

if data.dtype.kind == "f" or eps is None:

Notice the "or"

IMO, the estimation of eps should only happen if eps is None.... AND ... the default should be None => effective default is 1e-8. That's somewhat different from what you have. If you do go the route of having default eps is None, which is what I am suggesting, then it needs to be updated throughout the codebase.

alejoe91:
@magland done! Thanks for the feedback!

src/spikeinterface/preprocessing/whiten.py
```diff
@@ -15,23 +15,27 @@ class WhitenRecording(BasePreprocessor):
...
+    int_scale : None or float, default: None
         Apply a scaling factor to fit the integer range.
         This is used when the dtype is an integer, so that the output is scaled.
         For example, a value of `int_scale=200` will scale the traces value to a standard deviation of 200.
-    W : 2d np.array
-        Pre-computed whitening matrix, by default None
-    M : 1d np.array or None
+    eps : float or None, default: None
+        Small epsilon to regularize SVD.
+        If None, eps is default to 1e-8. If the data is float type and scaled down to very small values,
```
alejoe91, line 34:
```suggestion
        If None, eps is defaulted to 1e-8. If the data is float type and scaled down to very small values,
```

magland:
@alejoe91 looks good to me!

PR #2071, opened 2023-10-04
Change some default parameters for better user experience

### LINKED ISSUE #2008: Request to Change Default Behavior of run_sorter() Function to Prevent Data Loss
## Description:

Hello! I would like to request a change regarding the default behavior of the [spikeinterface.sorters.run_sorter()](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/sorters/runsorter.py) function.

As it currently stands, the [`run_sorter()`](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/sorters/runsorter.py#L94) function deletes everything in the output folder by default. While I understand that this may be intended for convenience, it has had a profoundly devastating impact on my work. I've been working with MEArec data and the sorting and extractor classes and wanted to dump the results into my output folder for this project. Unfortunately, this resulted in the immediate deletion of all other outputs, logs, results, and trained models I had carefully developed for this ongoing project, as well as my whole post-processed dataset. The loss has hit very hard, and and I want to make sure this doesn't happen to anyone else.

## Safety Concerns:
Deleting content by default can lead to unintended consequences, especially when working in production environments or with sensitive data. Users may expect the function to be non-destructive by default, and this behavior can easily lead to data loss or operational disruptions. Issues or mistakes with scope could also cause attempted deletion of root directories and entire drives!

The docs mention the parameter in two locations: the [sorters module description](https://spikeinterface.readthedocs.io/en/latest/modules/sorters.html) and the [API reference](https://spikeinterface.readthedocs.io/en/latest/api.html), but **neither** mention that the default behavior is to delete everything. The only clue is the function signature which shows that True for deleting the entire folder. I strongly recommend that there be a warning in the documentation if the default behavior for any function is data destruction for an entire directory.

## Suggested Improvement:
I kindly suggest changing the default parameter for remove_existing_folder to False, thereby requiring an explicit option to perform deletion. This approach aligns with best practices in library design, where destructive actions should require **explicit user consent**.

## Benefits:
* A significant reduction in the risk of accidental data loss.
* Enhanced user experience, providing users with the confidence that their data remains secure by default.
* Consistency with the principle of least surprise, where default behavior minimizes unexpected and harmful outcomes.

I believe that making this change will enhance the usability and safety of the library for all users. In the meantime I was able to recover one trained model, but everything else seems to be permanently gone. I will begin rebuilding from scratch as best I can.

Thank you for your attention to this request.

Environment:

    Python 3.11.2
    SpikeInterface 0.98.2

samuelgarcia:
Change some default parameters for better user experience
* waveform sparse by default
* ms_before=1.0,    ms_after=2.0,
* correlogram with bin_ms= 1.0
*  remove_existing_folder: bool = False for run_sorter()

close #2008

and also small fix (uncorrelated)


alejoe91:
Not sure with sparsity=True by default. I would throw a warning for num_channels > 64

samuelgarcia:
For the user experience sparsity=True make everything faster with less disk space.
Even for 32 channels it is better.
I do not see why we should have the "bad" default option and make a warning to user saying : "hey you should use the other option!"

zm711:
I just want to throw a vote Sam's way on this one. I often forget to set sparsity to True when I'm doing tests on things and so this would be a nice change for me. My logic being : 1) it helps me, 2) so many people are dealing with probes of >32 channels so making the default the better (again my opinion) option and then letting people with smaller probes switch to false if they want seems better. 

But I could be convinced the other direction, just my gut and disk space is with Sam on this one.

src/spikeinterface/core/waveform_extractor.py
```diff
@@ -1507,7 +1507,7 @@ def extract_waveforms(
         If True and recording has gain_to_uV/offset_to_uV properties, waveforms are converted to uV.
     dtype: dtype or None
         Dtype of the output waveforms. If None, the recording dtype is maintained.
-    sparse: bool (default False)
+    sparse: bool (default True)
```
samuelgarcia, line 1510:
```suggestion
    sparse: bool, default: True
```

alejoe91 (approved):

alejoe91 (changes requested):
I think that default sparsity is messing up with all prosptocessing tests. @samuelgarcia can you fix that?

samuelgarcia:
> I just want to throw a vote Sam's way on this one.

I am so happy

> I  think that default sparsity is messing up with all prosptocessing tests. @samuelgarcia can you fix that?

I will try my best.

alejoe91 (approved):

PR #2078, opened 2023-10-05
Improve assert messages (preprocessing & core)

zm711:
My main goal for the next couple PRs will be to 

1) fix typos in asserts/exceptions/etc
2) add assert/exception/etc messages when they are lacking to make sure users can provide that info in GH issues

I started fixing a docstring/typing for one file, but I decided I wanted to keep this PR within a scope of just improving messaging around errors. I also want to do this in a couple PRs so I can get some feedback before I do all files.

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -45,7 +45,7 @@ class BaseExtractor:
         self._kwargs = {}
 
         # 'main_ids' will either be channel_ids or units_ids
-        # They is used for properties
+        # They are used for properties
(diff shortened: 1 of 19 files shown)


samuelgarcia:
Zach, you are amazingly heroic !
If you come to Lyon to will receive a bottle of cote du Rhone for this.
If Alessio do the same with an Italian wine, you will be able compare and rank the two flavor.
The result should make Alessio grumpy.

alejoe91 (approved):

PR #2079, opened 2023-10-06
Fix plot_traces with ipywidgets

samuelgarcia:
Fix plot_traces with ipywidgets:
 * when channel_ids is not None
 * order_channel_by_depth for ipywidgets
 * order_channel_by_depth when channel_ids is given


src/spikeinterface/widgets/traces.py
```diff
@@ -88,26 +88,33 @@ def __init__(
         else:
             raise ValueError("plot_traces recording must be recording or dict or list")
 
+        if "location" in rec0.get_property_keys():
```
alejoe91, line 91: You can use `rec0.has_channel_locations()`

yger (approved):

PR #2081, opened 2023-10-06
small fix in gtstudy

samuelgarcia:
small fix in gtstudy

### DIFF
--- a/src/spikeinterface/comparison/groundtruthstudy.py
+++ b/src/spikeinterface/comparison/groundtruthstudy.py
@@ -165,7 +165,7 @@ class GroundTruthStudy:
             sorting_exists = sorting_folder.exists()
 
             sorter_folder = self.folder / "sorters" / self.key_to_str(key)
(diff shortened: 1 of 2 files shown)


src/spikeinterface/comparison/groundtruthstudy.py
```diff
@@ -185,6 +185,9 @@ def run_sorters(self, case_keys=None, engine="loop", engine_kwargs={}, keep=True
                 if log_file.exists():
                     log_file.unlink()
 
+            if sorter_folder_exists:
```
yger, line 188: this is not correct. You have an issue L168 also, with wrong names

yger: The if L181 should be if sorter_folder_exists no?

samuelgarcia: L181 is OK we are removing the sorting folder

alejoe91 (approved):

PR #2100, opened 2023-10-13
Fix for `module_gallery` with new sparsity default

zm711:
The waveform extractor example is broken (due to change in sparsity default in #2071:

https://spikeinterface.readthedocs.io/en/latest/modules_gallery/core/plot_4_waveform_extractor.html#sphx-glr-modules-gallery-core-plot-4-waveform-extractor-py

We need to force the dense example to be dense rather than let it default.


alejoe91 (approved):

alejoe91:
Thanks Zach!

zm711:
Failures are unrelated. They are occurring in the sorter class tests.

alejoe91:
Yep, @samuelgarcia and @yger broke something in their tools! We shouldn't let them roam too free ;)

PR #2102, opened 2023-10-13
Fix grouping of OpenEphys NPIX

### LINKED ISSUE #2099: Question about SI "group" property for multi-shank probe
Hi All,

I have an OpenEphys recording (NP2.0, 4-shank) loaded with `read_openephys()` that loads the data and probe in a way that seems correct (plot of probe below). Althought, in the probe view only 2 of the shanks are plot, but indeed the activate channel position overlaid over the (missing) other two shanks looks correct.

However, the `recording.get_channel_groups()` is set to all zero. I had assumed behind the scenes this would set the channels to be grouped by shank (if I recall correctly this is the case with SpikeGLX output data). I am happy to set the groups manually but wanted to check if this should happen automagically / I am doing something wrong.

Cheers!
Joe

image

alejoe91:
Fixes #2099 

@JoeZiminski can you test this out?

### DIFF
--- a/examples/modules_gallery/qualitymetrics/plot_4_curation.py
+++ b/examples/modules_gallery/qualitymetrics/plot_4_curation.py
@@ -61,4 +61,4 @@ print(keep_unit_ids)
 
 curated_sorting = sorting.select_units(keep_unit_ids)
 print(curated_sorting)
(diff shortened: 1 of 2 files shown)


samuelgarcia (approved):

JoeZiminski:
Thanks @alejoe91, works perfectly!

PR #2108, opened 2023-10-16
Use versioned kilosort options as default in `skip_preprocessed_data`.

### LINKED ISSUE #1977: `skip_kilosort_preprocessing` sets KS > 2.5 options for KS 2
When skipping kilosort preprocessing, [kilosort2_master.m](https://github.com/samuelgarcia/spikeinterface/blob/96c28e0ffffdcae16377c2274aaa2f86ed3e2ca8/src/spikeinterface/sorters/external/kilosort2_master.m) uses the default options from kilosort 2.5, however these are not the same in the original KS implementation (see kilosort 2  preprocessSubData below). In particular, the way the buffering is done was changed (e.g. `NTbuff      = NT + 4*ops.ntbuff;` in ks2 vs. `ops.NT + 3*ops.ntbuff`).

I think the KS2.5 settings are set for KS2, KS2.5 and KS3 (there are print statements of `fprintf("SKIP kilosort2.5 preprocessing\n");` for all three KS versions). As far as I can tell KS2.5 and KS3 are identical (although I have not checked line-by-line yet). But, I think the KS2 version needs updating for sure.

I'm not sure of the downstream effect this would have, it could well be silent if these options (e.g. `ops.NTbuff`) are not used downstream by KS.

Happy to make a PR but not sure when I'd be able and don't want to forget!

```
function rez = preprocessDataSub(ops)
% this function takes an ops struct, which contains all the Kilosort2 settings and file paths
% and creates a new binary file of preprocessed data, logging new variables into rez.
% The following steps are applied:
% 1) conversion to float32;
% 2) common median subtraction;
% 3) bandpass filtering;
% 4) channel whitening;
% 5) scaling to int16 values

tic;
ops.nt0 	  = getOr(ops, {'nt0'}, 61); % number of time samples for the templates (has to be <=81 due to GPU shared memory)
ops.nt0min  = getOr(ops, 'nt0min', ceil(20 * ops.nt0/61)); % time sample where the negative peak should be aligned

NT       = ops.NT ; % number of timepoints per batch
NchanTOT = ops.NchanTOT; % total number of channels in the raw binary file, including dead, auxiliary etc

bytes       = get_file_size(ops.fbinary); % size in bytes of raw binary
nTimepoints = floor(bytes/NchanTOT/2); % number of total timepoints
ops.tstart  = ceil(ops.trange(1) * ops.fs); % starting timepoint for processing data segment
ops.tend    = min(nTimepoints, ceil(ops.trange(2) * ops.fs)); % ending timepoint
ops.sampsToRead = ops.tend-ops.tstart; % total number of samples to read
ops.twind = ops.tstart * NchanTOT*2; % skip this many bytes at the start

Nbatch      = ceil(ops.sampsToRead /(NT-ops.ntbuff)); % number of data batches
ops.Nbatch = Nbatch;

[chanMap, xc, yc, kcoords, NchanTOTdefault] = loadChanMap(ops.chanMap); % function to load channel map file
ops.NchanTOT = getOr(ops, 'NchanTOT', NchanTOTdefault); % if NchanTOT was left empty, then overwrite with the default

if getOr(ops, 'minfr_goodchannels', .1)>0 % discard channels that have very few spikes
    % determine bad channels
    fprintf('Time %3.0fs. Determining good channels.. \n', toc);
    igood = get_good_channels(ops, chanMap);

    chanMap = chanMap(igood); %it's enough to remove bad channels from the channel map, which treats them as if they are dead

    xc = xc(igood); % removes coordinates of bad channels
    yc = yc(igood);
    kcoords = kcoords(igood);

    ops.igood = igood;
else
    ops.igood = true(size(chanMap));
end

ops.Nchan = numel(chanMap); % total number of good channels that we will spike sort
ops.Nfilt = getOr(ops, 'nfilt_factor', 4) * ops.Nchan; % upper bound on the number of templates we can have

rez.ops         = ops; % memorize ops

rez.xc = xc; % for historical reasons, make all these copies of the channel coordinates
rez.yc = yc;
rez.xcoords = xc;
rez.ycoords = yc;
% rez.connected   = connected;
rez.ops.chanMap = chanMap;
rez.ops.kcoords = kcoords;


NTbuff      = NT + 4*ops.ntbuff; % we need buffers on both sides for filtering

rez.ops.Nbatch = Nbatch;
rez.ops.NTbuff = NTbuff;
rez.ops.chanMap = chanMap;


fprintf('Time %3.0fs. Computing whitening matrix.. \n', toc);

% this requires removing bad channels first
Wrot = get_whitening_matrix(rez); % outputs a rotation matrix (Nchan by Nchan) which whitens the zero-timelag covariance of the data


fprintf('Time %3.0fs. Loading raw data and applying filters... \n', toc);

fid         = fopen(ops.fbinary, 'r'); % open for reading raw data
if fid<3
    error('Could not open %s for reading.',ops.fbinary);
end
fidW        = fopen(ops.fproc,   'w'); % open for writing processed data
if fidW<3
    error('Could not open %s for writing.',ops.fproc);    
end

for ibatch = 1:Nbatch
    % we'll create a binary file of batches of NT samples, which overlap consecutively on ops.ntbuff samples
    % in addition to that, we'll read another ops.ntbuff samples from before and after, to have as buffers for filtering
    offset = max(0, ops.twind + 2*NchanTOT*((NT - ops.ntbuff) * (ibatch-1) - 2*ops.ntbuff)); % number of samples to start reading at.
    if offset==0
        ioffset = 0; % The very first batch has no pre-buffer, and has to be treated separately
    else
        ioffset = ops.ntbuff;
    end
    fseek(fid, offset, 'bof'); % fseek to batch start in raw file

    buff = fread(fid, [NchanTOT NTbuff], '*int16'); % read and reshape. Assumes int16 data (which should perhaps change to an option)
    if isempty(buff)
        break; % this shouldn't really happen, unless we counted data batches wrong
    end
    nsampcurr = size(buff,2); % how many time samples the current batch has
    if nsampcurr<NTbuff
        buff(:, nsampcurr+1:NTbuff) = repmat(buff(:,nsampcurr), 1, NTbuff-nsampcurr); % pad with zeros, if this is the last batch
    end

    datr    = gpufilter(buff, ops, chanMap); % apply filters and median subtraction

    datr    = datr(ioffset + (1:NT),:); % remove timepoints used as buffers

    datr    = datr * Wrot; % whiten the data and scale by 200 for int16 range

    datcpu  = gather(int16(datr)); % convert to int16, and gather on the CPU side
    count = fwrite(fidW, datcpu, 'int16'); % write this batch to binary file
    if count~=numel(datcpu)
        error('Error writing batch %g to %s. Check available disk space.',ibatch,ops.fproc);
    end
end

rez.Wrot    = gather(Wrot); % gather the whitening matrix as a CPU variable

fclose(fidW); % close the files
fclose(fid);

fprintf('Time %3.0fs. Finished preprocessing %d batches. \n', toc, Nbatch);

rez.temp.Nbatch = Nbatch;
```

JoeZiminski:
closes #1977.

When using `skip_preprocessed_data` option in kilosort, default options are set from Kilosort's `preprocessDataSub.m` file. Kilosort2 is subtly different to Kilosort2.5 and Kilosort3 but the previous default values did not reflect this. This PR updates the default configs to match those found in `preprocessDataSub.m` of kilosort 2, 2.5 and 3 respectively. This did not require much, only slightly changing 2 variables for kilosort2.

I had a couple of questions:

1) 
  I saw this code back
  ```
  ops.NchanTOT = getOr(ops, 'NchanTOT', NchanTOTdefault); % if NchanTOT was left empty, then overwrite with the default
  ```
  in all three KS versions but not in the master matlab files. I guess this is because NchanTOT is expected to always be filled in. But I just wanted to check.

2) maybe now is a good time to also fix the `tic`/`toc` issue?

I'll take another look tomorrow just in case I missed anything!

### DIFF
--- a/src/spikeinterface/sorters/external/kilosort2_5_master.m
+++ b/src/spikeinterface/sorters/external/kilosort2_5_master.m
@@ -64,7 +64,7 @@ function kilosort2_5_master(fpath, kilosortPath)
             rez.ops.Nbatch = Nbatch;
             rez.ops.NTbuff = NTbuff;
 
-            tic;  % tocs are coming TODO
+            tic;  % tocs are supplied in other parts of KS code
 
         else
             % preprocess data to create temp_wh.dat
--- a/src/spikeinterface/sorters/external/kilosort2_5_master.m
+++ b/src/spikeinterface/sorters/external/kilosort2_5_master.m
@@ -26,7 +26,7 @@ function kilosort2_5_master(fpath, kilosortPath)
         if skip_kilosort_preprocessing
             % hack to skip the internal preprocessing
             % this mimic the preprocessDataSub() function
-            fprintf("SKIP kilosort2.5 preprocessing\n");
+            fprintf("Skipping kilosort2.5 preprocessing\n");
 
             ops.nt0 	  = getOr(ops, {'nt0'}, 61); % number of time samples for the templates (has to be <=81 due to GPU shared memory)
             ops.nt0min  = getOr(ops, 'nt0min', ceil(20 * ops.nt0/61)); % time sample where the negative peak should be aligned
@@ -49,11 +49,13 @@ function kilosort2_5_master(fpath, kilosortPath)
             rez.xcoords = xc;
             rez.ycoords = yc;
             Nbatch      = ceil(ops.sampsToRead / ops.NT); % number of data batches
+            ops.Nbatch = Nbatch;
             NTbuff      = ops.NT + 3*ops.ntbuff; % we need buffers on both sides for filtering
 
             rez.Wrot    = eye(ops.Nchan); % fake whitenning
             rez.temp.Nbatch = Nbatch;
-            % fproc is the same as the binary
+            % fproc (preprocessed output) is the same as the fbinary (unprocessed input)
+            % because we are skipping preprocessing.
             ops.fproc = ops.fbinary;
 
             rez.ops = ops; % memorize ops
@@ -62,7 +64,7 @@ function kilosort2_5_master(fpath, kilosortPath)
             rez.ops.Nbatch = Nbatch;
             rez.ops.NTbuff = NTbuff;
 
-            tic;  % tocs are coming
+            tic;  % tocs are coming TODO
 
         else
             % preprocess data to create temp_wh.dat
--- a/src/spikeinterface/sorters/external/kilosort2_master.m
+++ b/src/spikeinterface/sorters/external/kilosort2_master.m
@@ -64,7 +64,7 @@ function kilosort2_master(fpath, kilosortPath)
             rez.ops.Nbatch = Nbatch;
             rez.ops.NTbuff = NTbuff;
 
-            tic;  % tocs are coming TODO
+            tic;  % tocs are supplied in other parts of KS code
 
         else
             % preprocess data to create temp_wh.dat
--- a/src/spikeinterface/sorters/external/kilosort2_master.m
+++ b/src/spikeinterface/sorters/external/kilosort2_master.m
@@ -26,7 +26,7 @@ function kilosort2_master(fpath, kilosortPath)
         if skip_kilosort_preprocessing
             % hack to skip the internal preprocessing
             % this mimic the preprocessDataSub() function
-            fprintf("SKIP kilosort2.5 preprocessing\n");
+            fprintf("Skipping kilosort2 preprocessing\n");
 
             ops.nt0 	  = getOr(ops, {'nt0'}, 61); % number of time samples for the templates (has to be <=81 due to GPU shared memory)
             ops.nt0min  = getOr(ops, 'nt0min', ceil(20 * ops.nt0/61)); % time sample where the negative peak should be aligned
@@ -48,12 +48,14 @@ function kilosort2_master(fpath, kilosortPath)
             rez.yc = yc;
             rez.xcoords = xc;
             rez.ycoords = yc;
-            Nbatch      = ceil(ops.sampsToRead / ops.NT); % number of data batches
-            NTbuff      = ops.NT + 3*ops.ntbuff; % we need buffers on both sides for filtering
+            Nbatch      = ceil(ops.sampsToRead /(NT-ops.ntbuff)); % number of data batches
+            ops.Nbatch = Nbatch;
+            NTbuff      = NT + 4*ops.ntbuff; % we need buffers on both sides for filtering
 
             rez.Wrot    = eye(ops.Nchan); % fake whitenning
             rez.temp.Nbatch = Nbatch;
-            % fproc is the same as the binary
+            % fproc (preprocessed output) is the same as the fbinary (unprocessed input)
+            % because we are skipping preprocessing.
             ops.fproc = ops.fbinary;
 
             rez.ops = ops; % memorize ops
@@ -62,7 +64,7 @@ function kilosort2_master(fpath, kilosortPath)
             rez.ops.Nbatch = Nbatch;
             rez.ops.NTbuff = NTbuff;
 
-            tic;  % tocs are coming
+            tic;  % tocs are coming TODO
 
         else
             % preprocess data to create temp_wh.dat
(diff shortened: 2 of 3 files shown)


zm711:
@JoeZiminski 
>I had a couple of questions:

>I saw this code back

>ops.NchanTOT = getOr(ops, 'NchanTOT', NchanTOTdefault); % if NchanTOT was left empty, then overwrite with the default


I think you are pointing out a concern kilosort users have for example [kilosort565](https://github.com/MouseLand/Kilosort/issues/565), that may be an oversight in just using defaults sometimes. I also haven't dug too deep, so I'm super curious to see what you find.

src/spikeinterface/sorters/external/kilosort2_5_master.m
```diff
@@ -62,7 +64,7 @@ function kilosort2_5_master(fpath, kilosortPath)
             rez.ops.Nbatch = Nbatch;
             rez.ops.NTbuff = NTbuff;
 
-            tic;  % tocs are coming
+            tic;  % tocs are coming TODO
```
alejoe91, line 67: The tocs are actually in the KS codebase, see https://github.com/SpikeInterface/spikeinterface/pull/1887

alejoe91:
Thanks @JoeZiminski 
> ```
> ops.NchanTOT = getOr(ops, 'NchanTOT', NchanTOTdefault); % if NchanTOT was left empty, then overwrite with the default
> ```
> 
> in all three KS versions but not in the master matlab files. I guess this is because NchanTOT is expected to always be filled in. But I just wanted to check.

This is automatically filled in `kilosortbase.py`: https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/sorters/external/kilosortbase.py#L90-L91

> 2. maybe now is a good time to also fix the `tic`/`toc` issue?
> 
The toc is not an issue. In the main KS code the `tic` is in the `preprocessDataSub.m` and `toc`s are elsewhere in the code, so it's correct to have a hanging `tic` there (also see here: https://github.com/SpikeInterface/spikeinterface/pull/1887).

Final note: we'll have to rebuild the Docker images after this, since the matlab files changed (just for our reference)

zm711:
@alejoe91 
> The toc is not an issue. In the main KS code the tic is in the preprocessDataSub.m and tocs are elsewhere in the code, so it's correct to have a hanging tic there (also see here: https://github.com/SpikeInterface/spikeinterface/pull/1887).

I think I understand Joe's confusion (because when I read that comment I also was slightly confused). The comment makes it sound like a todo request and not that they are coming later in the code. Maybe it would be better to change the comment to something like 

```
# tocs are supplied in other parts of KS code
```

src/spikeinterface/sorters/external/kilosort2_5_master.m
zm711, line 67:
```suggestion
            tic;  % tocs are supplied in other parts of KS code
```

src/spikeinterface/sorters/external/kilosort2_master.m
```diff
@@ -62,7 +64,7 @@ function kilosort2_master(fpath, kilosortPath)
             rez.ops.Nbatch = Nbatch;
             rez.ops.NTbuff = NTbuff;
 
-            tic;  % tocs are coming
+            tic;  % tocs are coming TODO
```
zm711, line 67:
```suggestion
            tic;  % tocs are supplied in other parts of KS code
```

src/spikeinterface/sorters/external/kilosort3_master.m
```diff
@@ -62,7 +64,7 @@ function kilosort3_master(fpath, kilosortPath)
             rez.ops.Nbatch = Nbatch;
             rez.ops.NTbuff = NTbuff;
 
-            tic;  % tocs are coming
+            tic;  % tocs are coming TODO
```
zm711, line 67:
```suggestion
            tic;  % # tocs are supplied in other parts of KS code
```

zm711 (commented):
@JoeZiminski, see my above comment to Alessio.

PR #2109, opened 2023-10-17
Improve assert messaging (exporters, curation, part of sorters)

zm711:
Just another tiny set of typo/grammar fixes for some asserts and trying to provide the user more information about how to fix certain asserts. 

Also in curation there was a typo in two internal function names so I corrected them just to make them easier to find if needed in the future (I didn't run a test on those, so if I see if fail in the CI I'll fix, but they should be called from other functions where I fixed the call too).

### DIFF
--- a/src/spikeinterface/curation/curation_tools.py
+++ b/src/spikeinterface/curation/curation_tools.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 from typing import Optional
 import numpy as np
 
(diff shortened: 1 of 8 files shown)


alejoe91 (approved):

alejoe91:
Thanks @zm711 

Would you label this kind of PRs documentation? I'm unsure since definitely they improve error messages (which is kinda documentation), but it's not documentation per se!

zm711:
Last time you did enhancement, so I wasn't actually sure which label you preferred, but documentation works for me!

PR #2111, opened 2023-10-17
Fixes typing error in waveforms

h-mayorquin:
Fixes dev tests failing neuroconv. This would also have failed for our users of python 3.8.

I also got another try imported and sent it to the runtime.

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/neuroscope.py
+++ b/src/spikeinterface/extractors/neoextractors/neuroscope.py
@@ -9,12 +9,6 @@ from spikeinterface.core.core_tools import define_function_from_class
 
 from .neobaseextractor import NeoBaseRecordingExtractor
 
-try:
-    from lxml import etree as et
-
-    HAVE_LXML = True
-except ImportError:
-    HAVE_LXML = False
 
 PathType = Union[str, Path]
 OptionalPathType = Optional[PathType]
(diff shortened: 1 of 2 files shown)


src/spikeinterface/extractors/neoextractors/neuroscope.py
```diff
@@ -121,7 +113,7 @@ def __init__(
         exclude_shanks: Optional[list] = None,
         xml_file_path: OptionalPathType = None,
     ):
-        assert self.installed, self.installation_mesg
+        from lxml import etree as et
```
alejoe91, line 116:
```suggestion
        try:
            from lxml import etree as et
        except:
            raise ImportError(self.installation_mesg)
```
@h-mayorquin I thought we were still trying and raising an installation msg, but in the init. Am I wrong? See https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/extractors/cbin_ibl.py#L46-L49

src/spikeinterface/extractors/neoextractors/neuroscope.py
```diff
@@ -108,8 +102,6 @@ class NeuroScopeSortingExtractor(BaseSorting):
     """
 
     extractor_name = "NeuroscopeSortingExtractor"
-    installed = HAVE_LXML
-    installation_mesg = "Please install lxml to use this extractor!"
```
alejoe91, line 112: and keep the `installation_mesg`, or define later

alejoe91 (approved):

PR #2112, opened 2023-10-17
Add actual test for `check_your_install`

zm711:
This is for @h-mayorquin because I know he appreciates the value of good testing :)

I added an optional command line argument that will shut off the gui which will allow us to test `check_your_install.py` in the ci. For the end user I have the command line option default to include testing the gui so the end user won't even know a change occurred.  I know this is a bit of weird design (ie it would make more sense to have a `gui` flag set to `True`), but I wanted to maintain the old behavior so that users don't need to worry about command line arguments.


h-mayorquin (approved):
Thanks, this is fine to me.

PR #2113, opened 2023-10-18
fixed typo in variable name

skeltoh:
Changed assignment of all_temp_files to all_tmp_files to match the use further down. This was breaking the docker based kilosort2 setup.


alejoe91:
Thanks!

PR #2114, opened 2023-10-18
Optimize `make_match_count_matrix` 

h-mayorquin:
OK guys. This was really tedious to test and think about (I did the other Hilbert Space approach out of pure spite) but here it is. 

70 times faster in the following benchmark. It would be interesting to know if you think that some other parameters or regime is more interesting:

(`make_match_count_matrix_si` is the current method in main)
```python
durations=[18000.0], num_units=100
function.__name__='make_match_count_matrix_si'
--------------------------------------------------
Mean time over 5 iterations: 44.444810 seconds
Std over 5 iterations: 1.043592 seconds
Std without compliation over: 1.166766 seconds
Mean without compliation 44.443244 seconds
--------------------------------------------------
function.__name__='make_match_count_matrix'
--------------------------------------------------
Mean time over 5 iterations: 0.714384 seconds
Std over 5 iterations: 0.189878 seconds
Std without compliation over: 0.009023 seconds
Mean without compliation 0.619530 seconds
--------------------------------------------------
Speedup: 71.7369745920062
Speedup including compilation: 62.21417947326248
```

Here the code:
```python
num_units  = 100
#num_units = 3
durations = [5 * 60 * 60.0]
# durations = [60 * 60.0]
sampling_frequency = 30_000.0
seed = 20
sorting = generate_sorting(num_units, durations=durations, sampling_frequency=sampling_frequency, seed=seed)


samples_per_second = sampling_frequency
samples_per_millisecond = samples_per_second / 1000
delta_milliseconds = 3.334
delta_frames = int(delta_milliseconds * samples_per_millisecond)

sorting1  = sorting
sorting2  = sorting

number_it = 5

    functions = [make_match_count_matrix_si,
                 make_match_count_matrix, 
                # make_match_count_matrix_theoretical,
                 ]

    results = {}

    # Iterate through the functions and collect timing information
    for function in functions:
        print(f"{function.__name__=}")
        
        # Run the function multiple times to gather timing data
        times = [timeit.timeit(lambda: function(sorting1, sorting2, delta_frames=delta_frames), number=1) for _ in range(number_it)]
        
        # Calculate mean and std of the times
        mean_time = np.mean(times)
        std_time = np.std(times)
        mean_without_compilation = np.mean(times[1:])
        std_without_compilation = np.std(times[1:])
        
        
        print("-" * 50)
        print(f"Mean time over {number_it} iterations: {mean_time:2.6f} seconds")
        print(f"Std over {number_it} iterations: {std_time:2.6f} seconds")
        print(f"Std without compliation over: {std_without_compilation:2.6f} seconds")
        print(f"Mean without compliation {mean_without_compilation:2.6f} seconds")

        # print(f"Timing results: {times}")
        
        results[function.__name__] = {"iterations": number_it, "times": times, "mean": mean_time, "std": std_time, "mean_without_compilaton": mean_without_compilation}

        print("-" * 50)
        
    # Divide the mean without compilation of the first function by the mean without compilation of the second function
    speedup = results[functions[0].__name__]['mean_without_compilaton'] / results[functions[1].__name__]['mean_without_compilaton']
    speed_up_with_compliation = results[functions[0].__name__]['mean'] / results[functions[1].__name__]['mean']
    print(f"Speedup: {speedup}")
    print(f"Speedup with numba compilation: {speed_up_with_compliation}")
```


samuelgarcia:
Yeah! You are a true heroe.
Thanks to this, I think we will be able to remove this joblib dependencies soon!!!

alejoe91:
@h-mayorquin I think we need here to test `numba` import: 
https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/postprocessing/correlograms.py#L7-L12

and define the functions only `if HAVE_NUMBA`:
https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/postprocessing/correlograms.py#L323-L327

Right?

h-mayorquin:
@alejoe91 
You are totally right, Alessio!

I wanted to avoid the bare try-except (as that still imports numba even when you are not using it) but for numba the type of solution that preserves the performance is a monster. Check the last commit.

I wonder if we should eat the performance importing cost and just have the simple try except.

zm711:
I actually had a semi-related question to the point of this PR, so I'm hoping to just index here so if I forget I can come back. I understand the point of keeping minimal dependencies, but there is so much numba in this code why isn't it a dependency?

> There is always a quality metrics test failing somewhere, any idea about this one @alejoe91?

> It seems unrelated to the change here but I have been wrong about this before...

I've seen that test fail on my PRs before too. I think it is a non-deterministic test that fails on some runs and not others, it probably needs a patch at some point.

samuelgarcia:
About dependency this is OK to more heavy dependency in sub modules pecific.
comparison having numba is OK.
what I do not like is core having numba and scipy.

So if comparison need numba we can use global import for numba I think like do evrywhere else.

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -109,50 +109,136 @@ def count_match_spikes(times1, all_times2, delta_frames):  # , event_counts1, ev
...
 
-    Returns
-    -------
-    match_event_count: array (int64)
-        Matrix of match count spike
-    """
-    import pandas as pd
+    # Nested function
+    @numba.jit(nopython=True, nogil=True)
+    def compute_matching_matrix_inner(
```
samuelgarcia, line 122: what is the "inner" for ? "in segment" ?

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -109,50 +109,136 @@ def count_match_spikes(times1, all_times2, delta_frames):  # , event_counts1, ev
...
+        frames_spike_train1,
+        frames_spike_train2,
+        unit_indices1,
+        unit_indices2,
+        num_units_sorting1,
+        num_units_sorting2,
+        delta_frames,
+    ):
```
samuelgarcia, line 130: Lets add a small doc here on the strategy used in the double loop

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -109,50 +109,136 @@ def count_match_spikes(times1, all_times2, delta_frames):  # , event_counts1, ev
...
+            while index2 < len(frames_spike_train2):
+                frame2 = frames_spike_train2[index2]
+                not_a_match = abs(frame1 - frame2) > delta_frames
+                if not_a_match:
+                    break
+
+                # Map the match to a matrix
+                row, column = unit_indices1[index1], unit_indices2[index2]
+
+                # Jeremy Magland condition, the same unit can't match twice
```
samuelgarcia, line 158: It is not the same unit, it is the same spike.

samuelgarcia: but the logic is OK. lets just change the comments. You can remove Jeremy it was a concensus on the group

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -109,50 +109,136 @@ def count_match_spikes(times1, all_times2, delta_frames):  # , event_counts1, ev
...
+                if frame1 != previous_frame1_match[row, column] and frame2 != previous_frame2_match[row, column]:
+                    previous_frame1_match[row, column] = frame1
+                    previous_frame2_match[row, column] = frame2
+
+                    matching_matrix[row, column] += 1
+
+                index2 += 1
+
+                # Advance the minimal index 2 if not in the last loop iteration
+                if not_in_the_last_loop:
```
samuelgarcia, line 168: I do not think we need this not_in_the_last_loop protection because if it is the last loop moving the index2 do not hurt

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -109,50 +109,136 @@ def count_match_spikes(times1, all_times2, delta_frames):  # , event_counts1, ev
...
+                    not_a_match_with_next = abs(next_frame1 - frame2) > delta_frames
```
samuelgarcia, line 169: Habing the abs() sounds dangerous here. Let think about it.

src/spikeinterface/core/generate.py
```diff
@@ -173,7 +173,7 @@ def generate_sorting(
             duration=durations[segment_index],
             refractory_period_ms=refractory_period_ms,
             firing_rates=firing_rates,
-            seed=seed,
+            seed=seed + segment_index,
```
samuelgarcia, line 176: oups. good catch. maybe we could have something to generate several seeds we several segment. but ouside the scope of this PR

src/spikeinterface/comparison/comparisontools.py
h-mayorquin: What do you mean by dangerous? 

Sam, I don't read your mind! I don't know what I should think about.

src/spikeinterface/comparison/comparisontools.py
h-mayorquin: Yep, you are right is spikes.

src/spikeinterface/comparison/comparisontools.py
h-mayorquin: Just means ia closure, but this is not necessary, I am removing the inner.

src/spikeinterface/comparison/comparisontools.py
h-mayorquin: OK.

src/spikeinterface/comparison/comparisontools.py
h-mayorquin: Yes, I had a different implementation and this should have been removed.

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -109,50 +109,136 @@ def count_match_spikes(times1, all_times2, delta_frames):  # , event_counts1, ev
...
+
+    sample_frames2_accumulator = []
+    unit_indices2_accumulator = []
+
+    for segment_index in range(num_segments_sorting1):
+        spike_trains1 = [sorting1.get_unit_spike_train(unit_id, segment_index) for unit_id in unit1_ids]
+        sample_frames1_accumulator.extend(spike_trains1)
+
+        unit_indices1 = [np.full(len(train), unit) for unit, train in enumerate(spike_trains1)]
+        unit_indices1_accumulator.extend(unit_indices1)
```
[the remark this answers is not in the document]
h-mayorquin: I just tested this:

The current implementation is faster and avoids the extra memory allocation for the segment index that is not used here. In addition, looking at the current implementation of `to_spike_vector` the sorting is done only over the segments (and not across them). I need all the frames sorted, not just within segment.



```
function.__name__='build_current_context'
mean_time=0.75972664649953 - std_time=0.0021718486690029907
[0.758596622999903, 0.7616681849995075, 0.7567235919996165, 0.7619181859990931]
--------------------------------------------------
function.__name__='build_spike_vector'
mean_time=1.015768427749208 - std_time=0.003838008096622767
[1.0212817799983895, 1.0171307839991641, 1.0110996240000532, 1.0135615229992254]
--------------------------------------------------
```

https://gist.github.com/h-mayorquin/c5efc61079c09ca1362addafa3f4ad17

src/spikeinterface/comparison/comparisontools.py
h-mayorquin: Removed in the last commit.

alejoe91:
Thanks Ramon! This looks great to me! @samuelgarcia do you have any more comments?

samuelgarcia:
Let me read this again.

h-mayorquin:
Just had a discussion with Sam. Some points.
1) I was not aware that sorting objects can reset their time. For that and more, spikes from different sorters should not be compared as I am currently doing. I will change this.
2) I will change to `spike_vector`, that will be improved later and I think it will make the code easier to read.
3) We are both weary of the current solution for dynamically importing numba. We should think on a general solution on how to do this.

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -3,7 +3,6 @@
 """
 
 import numpy as np
-from joblib import Parallel, delayed
```
samuelgarcia, line 6: This is what I prefer!

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -109,48 +108,166 @@ def count_match_spikes(times1, all_times2, delta_frames):  # , event_counts1, ev
...
-    """
-    Make the match_event_count matrix.
-    Basically it counts the matching events for all given pairs of spike trains from
-    sorting1 and sorting2.
+def get_optimized_compute_matching_matrix():
+    """
+    This function is to avoid the bare try-except pattern when importing the compute_matching_matrix function
+    which uses numba. I tested using the numba dispatcher programatically to avoids this
+    but the performance improvements were lost. Think you can do better? Don't forget to measure performance against
+    the current implementation!
```
samuelgarcia, line 116:
```suggestion
    the current implementation!
    TODO: unify numba decorator across all modules
```

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -109,48 +108,167 @@ def count_match_spikes(times1, all_times2, delta_frames):  # , event_counts1, ev
...
+
+    spike_vector1_segments = sorting1.to_spike_vector(concatenated=False)
+    spike_vector2_segments = sorting2.to_spike_vector(concatenated=False)
+
+    num_segments_sorting1 = sorting1.get_num_segments()
+    num_segments_sorting2 = sorting2.get_num_segments()
+    max_segment_to_compare = max(num_segments_sorting1, num_segments_sorting2)
+
+    # Segments should be matched one by one
+    for segment_index in range(max_segment_to_compare):
```
samuelgarcia, line 246:
```suggestion
    assert num_segments_sorting1 == num_segments_sorting2, "make_match_count_matrix : sorting1 and sorting must have the same segment number"

    # Segments should be matched one by one
    for segment_index in range(num_segments_sorting1):
```

samuelgarcia (approved):

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -109,48 +108,169 @@ def count_match_spikes(times1, all_times2, delta_frames):  # , event_counts1, ev
...
+    matching_matrix = np.zeros((num_units_sorting1, num_units_sorting2), dtype=np.uint16)
+
+    spike_vector1_segments = sorting1.to_spike_vector(concatenated=False)
+    spike_vector2_segments = sorting2.to_spike_vector(concatenated=False)
+
+    num_segments_sorting1 = sorting1.get_num_segments()
+    num_segments_sorting2 = sorting2.get_num_segments()
+    assert (
+        num_segments_sorting1 == num_segments_sorting2
+    ), "make_match_count_matrix : sorting1 and sorting must have the same segment number"
```
alejoe91, line 245:
```suggestion
    ), "make_match_count_matrix : sorting1 and sorting2 must have the same segment number"
```

h-mayorquin:
great. lets work on the numba thing.

PR #2116, opened 2023-10-19
Remove debug plots from template metrics

alejoe91:
@samuelgarcia I think I forgot to push the removal of `debug_plots` from template metrics. Here it is!


samuelgarcia:
It was OK for me to have then what I did not like was to have then in the signature.
but this is better.

samuelgarcia (approved):

PR #2126, opened 2023-10-24
Use pipeline for postprocessing

alejoe91:
Re-implement as pipeline nodes:

- [x] spike amplitudes
- [x] spike locations
- [x] amplitude scalings

and make a `compute_postprocessing` function to merge the pipeline nodes so that the traces are traversed only once


src/spikeinterface/core/node_pipeline.py
```diff
@@ -129,7 +136,7 @@ def compute(self, traces, start_frame, end_frame, segment_index, max_margin):
 
         # make sample index local to traces
         local_peaks = local_peaks.copy()
-        local_peaks["sample_index"] -= start_frame - max_margin
+        local_peaks["sample_index"] -= start_frame - left_margin
```
alejoe91, line 139: @samuelgarcia this was wrong for the first chunk!!!

src/spikeinterface/core/node_pipeline.py
```diff
@@ -154,16 +161,26 @@ class SpikeRetriever(PeakSource):
         If False, the max channel is computed for each spike given a radius around the template max channel.
     extremum_channel_inds: dict of int
         The extremum channel index dict given from template.
-    radius_um: float (default 50.)
+    radius_um: float, default: 50.
         The radius to find the real max channel.
         Used only when channel_from_template=False
-    peak_sign: str (default "neg")
+    peak_sign: str, default: "neg"
```
zm711, line 167:
```suggestion
    peak_sign: "neg" | "pos" | "both" default: "neg"
```

I would suggest using this notation. Pre-3.10 the preferred notation for this would be `Literal['pos', 'neg']`, because although we input these as strings really there are only a limited number of strings recognized. Another option would be `'pos' or 'neg' or 'both'`, but for users actually using the doc strings it would be clearer to tell them the choice rather than say `str`. I've been slowly trying to update this. But what do you think?

src/spikeinterface/core/node_pipeline.py
samuelgarcia: no because of the zero padding

src/spikeinterface/core/node_pipeline.py
```diff
@@ -504,19 +533,21 @@ def _compute_peak_pipeline_chunk(segment_index, start_frame, end_frame, worker_c
...
+            node_output[0]["sample_index"] += extra_margin_left
         elif isinstance(node, PeakSource):
-            node_output = node.compute(traces_chunk, start_frame, end_frame, segment_index, max_margin)
+            node_output = node.compute(traces_chunk, start_frame, end_frame, segment_index, left_margin, right_margin)
         else:
             # TODO later when in master: change the signature of all nodes (or maybe not!)
-            node_output = node.compute(traces_chunk, *node_input_args)
+            node_output = node.compute(
+                traces_chunk, start_frame, end_frame, segment_index, left_margin, right_margin, *node_input_args
+            )
```
samuelgarcia, line 550: camarade, I am not super with this change.
A main design of nodes and local compute was it was separate from the start/stop and segment index.
In short a bit like a computing kernel. Why do we need to had theses 5 new args for all nodes ?

alejoe91: Not for all nodes, but spike amplitudes need at least the margins!

samuelgarcia:
Done with SortingAnalyzer.

PR #2130, opened 2023-10-25
Fix peak activity example & bugs in `peak_activity`

zm711:
The final function in the example was missing the `peaks` argument causing it to fail. (See link)

https://spikeinterface.readthedocs.io/en/latest/modules_gallery/widgets/plot_4_peaks_gallery.html#sphx-glr-modules-gallery-widgets-plot-4-peaks-gallery-py

I also added keyword arguments to encourage their use and make clear what arguments are going into the examples.

### DIFF
--- a/examples/modules_gallery/widgets/plot_4_peaks_gallery.py
+++ b/examples/modules_gallery/widgets/plot_4_peaks_gallery.py
@@ -48,6 +48,8 @@ print(peaks.dtype.fields.keys())
 
 si.plot_peak_activity(recording=rec_filtred, peaks=peaks)
 
+plt.show()
+
 ##############################################################################
 # can be also animated with bin_duration_s=1.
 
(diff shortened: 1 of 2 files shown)


zm711:
Also very minor, but the example uses British English spelling rather than US spelling. It honestly doesn't matter much, but in your function names you always use US spelling, so do you want me to make sure everything I see is consistent in the future or do you not care?

zm711:
Also discovered a couple bugs in `_plot_one_bin()`. 

1) It was trying to pull the recording from self, when `rec` is an argument from the function.
2) The positional arguments were out of order, so I switched to keyword to prevent this error

I'll see if I find anything else while I read through it all again.

zm711:
Based on my reading getting the dynamic content will be more trouble than it's worth in this PR. This is good as is for now (unless you want the dynamic example deleted since it appears static anyway).

samuelgarcia:
oups. My bad. I did not check examples when I ported theses widgets into new API.
Thnaks for the fix.

samuelgarcia (approved):

zm711:
I tested the dynamic example locally and it looked super super cool @samuelgarcia!

PR #2133, opened 2023-10-26
Add sparsity_temp_folder option to extract_waveforms

### LINKED ISSUE #2077: Pre-compute sparsity should have a `tempfolder` mode
The `precompute_sparsity` is used to compute the sparsity of each unit during the waveform extraction. However, this is always done in memory and if there are many units, multiple passes are done in unit blocks (of 100 units). For many units, it would be faster to save the waveforms in a temporary folder, compute sparsity, and then remove the tmp folder.

alejoe91:
Fixes #2077

### DIFF
--- a/src/spikeinterface/core/waveform_extractor.py
+++ b/src/spikeinterface/core/waveform_extractor.py
@@ -1481,7 +1481,9 @@ def extract_waveforms(
     dtype=None,
     sparse=True,
     sparsity=None,
+    sparsity_temp_folder=None,
     num_spikes_for_sparsity=100,
+    unit_batch_size=200,
     allow_unfiltered=False,
     use_relative_path=False,
     seed=None,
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/waveform_extractor.py
```diff
@@ -1612,6 +1621,8 @@ def extract_waveforms(
             ms_before=ms_before,
             ms_after=ms_after,
             num_spikes_for_sparsity=num_spikes_for_sparsity,
+            unit_batch_size=unit_batch_size,
+            temp_folder=sparsity_temp_folder,
```
samuelgarcia, line 1625: This 2 args should be in estimate_kwargs.

alejoe91: then `num_spikes_for_sparsity` too? Still good to list them in the docstrings though

samuelgarcia (approved):

PR #2135, opened 2023-10-26
Add quality_metrics specific module requirements to `pyproject.toml`

### LINKED ISSUE #1223: Add 'extra requirement' for quality metric sub-module
Issue discovered by adjusting dependencies on https://github.com/catalystneuro/neuroconv/pull/234, an extractor used in the testing suite uses some methods from the quality metrics submodule, and has minimal spikeinterface + scikit-learn installed, but ultimately results in some obscure errors about some level of `compute_quality_metrics` returning `None` instead of an object with a `_fields` attribute

The same code was previously working without issue, the only thing changed on that PR is the order and extend of secondary dependencies so I suspect that's the source of the problem

A simple fix would be to add an additional 'extra_requires' category for using the QM module, since I don't expect to need everything in `[full]` (or do I?)

zm711:
Fixes #1223 

There could always be more submodule specific options in the toml, but since qm is a smaller module and since it was requested should be an easy addition.

Also deleted a couple of unused imports.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,7 +103,7 @@ widgets = [
     "sortingview>=0.11.15",
 ]
 
-quality_metrics = [
+qualitymetrics = [
     "scikit-learn",
     "scipy",
     "pandas",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,6 +103,13 @@ widgets = [
     "sortingview>=0.11.15",
 ]
 
+quality_metrics = [
+    "scikit-learn",
+    "scipy",
+    "pandas",
+    "numba",
+]
+
 test_core = [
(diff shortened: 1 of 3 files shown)


zm711:
Not sure what the extractor test failure is for this, but there's no rush for this PR.

alejoe91 (approved):

alejoe91:
It's a recent update to `probeinterface` (https://github.com/SpikeInterface/probeinterface/pull/206). I'll make a PR to deal with the different versions.

zm711:
Thanks @alejoe91! I know prepping for releases are huge undertakings!

samuelgarcia (approved):

PR #2136, opened 2023-10-27
Fix open ephys probe loading and unify probeinterface import syntax

alejoe91:
Fixes CI errors occurring due to this pribeinterface PR: https://github.com/SpikeInterface/probeinterface/pull/206

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/openephys.py
+++ b/src/spikeinterface/extractors/neoextractors/openephys.py
@@ -13,7 +13,7 @@ from pathlib import Path
 import numpy as np
 import warnings
 
-import probeinterface as pi
+import probeinterface
 
 from .neobaseextractor import NeoBaseRecordingExtractor, NeoBaseSortingExtractor, NeoBaseEventExtractor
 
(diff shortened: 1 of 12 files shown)


src/spikeinterface/extractors/neoextractors/openephys.py
```diff
@@ -187,9 +185,17 @@ def __init__(
...
                     self.set_probe(probe, in_place=True)
-                probe_name = probe.annotations["probe_name"]
+
+                # this try-except handles a breaking change in probeinterface after v0.2.18
+                # in the new version, the Neuropixels model name is stored in the "model_name" attribute,
+                # rather than in the "probe_name" annotation
+                try:
+                    model_name = probe.model_name
+                except Exception as e:
+                    model_name = probe.annotations["probe_name"]
```
samuelgarcia, line 195: Lets do this without try except please.

alejoe91: We can't use the version since we don't have dev in probeinterface...

src/spikeinterface/extractors/shybridextractors.py
```diff
@@ -2,7 +2,7 @@
 
 import numpy as np
 
-from probeinterface import read_prb, write_prb
+import probeinterface as pi
```
samuelgarcia, line 5: I am not really sure that "pi" is a good alias!!!!
why not keeping flat import ?

alejoe91: is the same as `import spikeinterface as si`

alejoe91: plus it's like that in many other files..let's unify this

PR #2134, opened 2023-10-26
Unify 'defaults' and use of quotations across docstrings

alejoe91:
inspired by @zm711  :)

The way we defined defaults was very non-homogeneous (`(default ...)`, `, by default ...`, `default=...`). Made a first pass to unify this as:

```
param_name: type, default: ...
```

Might have missed some, but should have caught most of them!

And still missing a couple of modules (`sorters`, `sortingcomponents`)

PR #2138, opened 2023-10-28
Add `Docstring` section to development docs

zm711:
A companion documentation section to go with @alejoe91 #2134. This way future developers will be adequately warned about what is expected in the docstrings. 

Read through the docstring section and let me know if you want me to include more info.

*Other minor changes*
    *I restylized black to Black since that is their official name whereas black is the package (similar to this repo :) ) 
    * added a note about fmt skip vs fmt off for black
    * couple typo fixes

One question is near the bottom the section originally said edit the travis.yml, but I assumed it was suppose to be edit the pyproject to add the dependencies to the test section, but if I'm wrong correct that section.

### DIFF
--- a/doc/development/development.rst
+++ b/doc/development/development.rst
@@ -50,7 +50,7 @@ If you want to run a specific test in a specific file, you can use the following
 
 .. code-block:: bash
 
-    pytest pytest src/spikeinterface/core/tests/test_baserecording.py::specific_test_in_this_module
+    pytest src/spikeinterface/core/tests/test_baserecording.py::specific_test_in_this_module
 
 We also mantain pytest markers to run specific tests. For example, if you want to run only the tests
 for the :code:`spikeinterface.extractors` module, you can use the following command:
@@ -121,18 +121,18 @@ Stylistic conventions
 ---------------------
 
 
-SpikeInterface maintains a consistent coding style across the project, leveraging the black Python code formatter.
+SpikeInterface maintains a consistent coding style across the project, leveraging the Black Python code formatter.
 This helps to ensure readability and maintainability of the code, making it easier for contributors to collaborate.
 
-To install black, you can use pip, the Python package installer. Run the following command in your terminal:
+To install Black, you can use pip, the Python package installer. Run the following command in your terminal:
 
 .. code-block:: bash
 
     pip install black
 
-This will install black into your current Python environment.
+This will install Black into your current Python environment.
 
-In addition to black, we use pre-commit to manage a suite of code formatting.
+In addition to Black, we use pre-commit to manage a suite of code formatting.
 Pre-commit helps to automate the process of running these tools before every commit,
 ensuring that all code is checked for style.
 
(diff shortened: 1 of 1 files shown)


doc/development/development.rst
```diff
@@ -151,23 +151,70 @@ navigate to your local repository in your terminal and run the following command
...
 .. code-block:: python
 
     # fmt: off
+    code here
+    # fmt: on
 
 As described in the `black documentation <https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#code-style>`_,
 
-The following are some styling conventions that we follow in SpikeInterface:
+
```
h-mayorquin, line 173: Let's make a small section here that is "doctring conventions" or something like this.

doc/development/development.rst
```diff
@@ -151,23 +151,70 @@ navigate to your local repository in your terminal and run the following command
...
+        great_data: dict
+            A dictionary of the processed data
+        """
+
+
+Note that in this example we demonstrate two other docstring conventions followed by SpikeInterface. First, that all string arguments should be presented
+with double quotes. This is the same stylistic convention followed by Black and enforced by the pre-commit for the repo. Second, when a parameter is a
+string with a limited number of values (e.g. :code:`mean` and :code:`median`), rather than give the type a value of :code:`str`, please list the possible strings
+so that the user knows what the options are.
+
```
h-mayorquin, line 211: And we can have another small one here called "miscelleanous conventions" or something like this.

h-mayorquin (approved):
This looks good to me. I did this the first time and this is an improvement upon it. I have been meaning to add @alejoe91 convention for a while so I am thankful that @zm711 decided to do this.

I added two small suggestions that are up to @zm711 to follow up or not.

doc/development/development.rst
zm711: Done :)

doc/development/development.rst
zm711: Done : )

zm711:
While you're checking this out @h-mayorquin. I know you added the datalad instructions but I for one can't get datalad working on Mac ( m-series). I think there is a mismatch between datalad, git-annex, and brew. I'll update those instructions if I can figure it out, but just wanted that on your radar in case you know any magic tips for getting it to work.

h-mayorquin:
@zm711 I was kind of proud that I got datalad working for all the three OSs in the CI at some point in time. This was accomplished  using those instructions. Because of the Sam moratorium on testing we ended up not using that but ... flawed human as I am, I wanted to keep those instructions somewhere and they ended up there.

In retrospective, this is wrong. We are doing duplication by keeping those instructions there and, even worse, making promises that we don't know if we can keep ("the method works").  We should just point out to the documentation in datalad and be done with it.

Plus, now that we have good generators we will be getting rid of Datalad for core testing (which is great!) so this is even less necessary as we will only be needing if for extractor testing.

doc/development/development.rst
h-mayorquin: Thanks!

zm711:
First off I always super appreciate all your hard work getting the ci working @h-mayorquin!

> Plus, now that we have good generators we will be getting rid of Datalad for core testing (which is great!) so this is even less necessary as we will only be needing if for extractor testing.

I agree. I think switching the examples to the generators (in the modules gallery) would be super beneficial. Because right now we allow people to download the scripts, but the scripts will fail if the user doesn't have datalad. 

> We should just point out to the documentation in datalad and be done with it.

I'll change the datalad section right now just so that we don't "over-promise". 

(I was able to get datalad with windows, but I had to specify my git-annex version in the datalad installer and jump through a couple hoops, so yeah I think the packages are just changing at different rates, which makes it hard to have the instructions just work). 

Also just as an fyi (since you worked so hard on the ci). I recently tried to run the whole test-suite on windows (I ended about 7/8 of the way through) and ended up with something like 50-60 test failures. Mostly due to `shutil.rmtree` being blocked by Windows permssion errors.

h-mayorquin:
Thanks for testing this on windows. 

In the last meeting we discussed testing. I am hoping that by the end of this week and next week I will be working in some outstanding* maintenance issues which will include testing. 

 (I just realized that this is synonym with great which I find strange)

PR #2145, opened 2023-10-31
Remove postprocessing `template_tools` 

zm711:
There was a note in the file saying only to keep in version 0.97.0 and to remove after. Since we are at 0.99.0 shouldn't this be removed? Or did you want to keep it for a few more versions?


zm711:
Looks like tests were still relying on it. Let me check.

PR #2146, opened 2023-10-31
Implemented `sd_ratio` as quality metric

DradeAW:
A "new" metric I've been working on.
I hope to implement it fast enough for it to be part of v0.99

- [x] Implement censored period
- [x] Add tests
- [x] Add documentation


src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -1365,3 +1365,57 @@ def _compute_rp_violations_numba(nb_rp_violations, spike_trains, spike_clusters,
             spike_train = spike_trains[spike_clusters == i]
             n_v = _compute_nb_violations_numba(spike_train, t_r)
             nb_rp_violations[i] += n_v
+
+
+def compute_SD_test(
```
alejoe91, line 1370:
```suggestion
def compute_std_test(
```
I'm not super happy with the name. What abut `sd_ratio`?

DradeAW: Yeah I'm fine with that name!

`sd_ratio` or `std_ratio`?

doc/modules/qualitymetrics/sd_ratio.rst
```diff
@@ -0,0 +1,38 @@
+Standard Deviation (SD) ratio (:code:`sd_ratio`)
+==============================================
+
+Calculation
+-----------
+
+All spikes from the same neuron should have the same shape. This means that at the peak of the spike, the standard deviation of the voltage should be the same as that of noise. If spikes from multiple neurons are grouped in a single unit, chances are the standard deviation of spikes amplitude is going to increase.
```
zm711, line 7:
```suggestion
All spikes from the same neuron should have the same shape. This means that at the peak of the spike, the standard deviation of the voltage should be the same as that of noise. If spikes from multiple neurons are grouped into a single unit, the standard deviation of the spike amplitudes would likely be increased.
```

doc/modules/qualitymetrics/sd_ratio.rst
```diff
@@ -0,0 +1,38 @@
...
+
+This metric, first described [Pouzat]_ then adapted by Wyngaard, Llobet & Barbour (paper in writing), returns the ratio between both standard deviations:
```
zm711, line 9:
```suggestion
This metric, first described [Pouzat]_ then adapted by Wyngaard, Llobet & Barbour (in preparation), returns the ratio between both standard deviations:
```
In US academic English we typically use this to signify writing unless yours is the British way in which case it is fine to leave alone.

doc/modules/qualitymetrics/sd_ratio.rst
```diff
@@ -0,0 +1,38 @@
...
+	S = \frac{\sigma_{\mathrm{unit}}}{\sigma_{\mathrm{noise}}}
+
+To remove the effect of drift on spikes amplitude, :math:`\sigma_{\mathrm{unit}}` is computed by subtracting each spike amplitude, and dividing the resulting standard deviation by :math:`\sqrt{2}`.
+Also to remove the effect of bursts (which can have lower amplitudes), you can specify a censored period (by default 4.0 ms) where spikes happening less than this period after another spike will not be considered.
+
+
+Expectation and use
+-------------------
+
+For a unit representing a single neuron, this metric should return a value close to one. However for unit that are contaminated, the value can be significantly higher.
```
zm711, line 21:
```suggestion
For a unit representing a single neuron, this metric should return a value close to one. However for units that are contaminated, the value can be significantly higher.
```

doc/modules/qualitymetrics/sd_ratio.rst
```diff
@@ -0,0 +1,38 @@
...
+
+
+Example code
+------------
+
+.. code-block:: python
+
+	import spikeinterface.qualitymetrics as sqm
```
zm711, line 29: module level import :)

doc/modules/qualitymetrics/sd_ratio.rst
```diff
@@ -0,0 +1,38 @@
...
+
+	sd_ratio = sqm.compute_sd_ratio(wvf_extractor, censored_period_ms=4.0)
+
+
+Literature
+----------
+
+Introduced by [Pouzat]_ (2002).
+Expanded by Wyngaard, Llobet and Barbour (in writing).
```
zm711, line 38:
```suggestion
Expanded by Wyngaard, Llobet and Barbour (in preparation).
```

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -1365,3 +1365,68 @@ def _compute_rp_violations_numba(nb_rp_violations, spike_trains, spike_clusters,
...
+    **kwargs,
+):
+    """
+    Computes the SD (Standard Deviation) of each unit spikes amplitude, and compare it to that of noise.
+
+    Parameters
+    ----------
+    waveform_extractor : WaveformExtractor
+        The waveform extractor object.
+    censored_period_ms : float
```
zm711, line 1384:
```suggestion
    censored_period_ms : float, default: 4.0
```

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -1365,3 +1365,68 @@ def _compute_rp_violations_numba(nb_rp_violations, spike_trains, spike_clusters,
...
+        The censored period in milliseconds. This is to remove any potential bursts that could affect the SD.
+    correct_for_drift: bool
```
zm711, line 1386:
```suggestion
    correct_for_drift: bool, default: True
```

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -1365,3 +1365,68 @@ def _compute_rp_violations_numba(nb_rp_violations, spike_trains, spike_clusters,
...
+        If True, will subtract the amplitudes sequentiially to significantly reduce the impact of drift.
+    unit_ids : list or None
```
zm711, line 1388:
```suggestion
    unit_ids : list or None, default: None
```

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -1365,3 +1365,68 @@ def _compute_rp_violations_numba(nb_rp_violations, spike_trains, spike_clusters,
...
+
+def compute_sd_ratio(
+    wvf_extractor: WaveformExtractor,
+    censored_period_ms: float = 4.0,
+    correct_for_drift: bool = True,
+    unit_ids=None,
+    **kwargs,
+):
+    """
+    Computes the SD (Standard Deviation) of each unit spikes amplitude, and compare it to that of noise.
```
zm711, line 1378:
```suggestion
    Computes the SD (Standard Deviation) of each unit's spike amplitudes, and compare it to the standard deviation of noise.
```

What is noise in this case? Might be helpful to say what noise you are using as the noise in the calculation.

zm711 (commented):
@DradeAW,

I know this is still in draft, but I tagged a couple notes. You can ignore and fix yourself if you want, but just wanted these on your radar.

DradeAW:
@zm711 Thanks for the feedback!
Useful and appreciated :)

I'll make the changes after my lunch!

DradeAW:
Ok I like the state of the PR as it is now. A few things that are not perfect:

- Computing noise outside of spikes from the unit (or subtracting the template)
- Considering jitter
- Here (https://github.com/SpikeInterface/spikeinterface/blob/7cd707d2039f72ad7a593895d10ef40a8c5a17c5/src/spikeinterface/qualitymetrics/misc_metrics.py#L1428) I'm accessing a private variable. Should we add a method `get_param(key)`?

zm711:
Could you also do a big favor with this PR and fix the order of `qualitymetrics.rst`. They are suppose to be in alphabetical order, but they got out of order. I can't comment on lines that you haven't touched, but it would be cool to have the docs fixed in this PR :)

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -1365,3 +1365,90 @@ def _compute_rp_violations_numba(nb_rp_violations, spike_trains, spike_clusters,
...
+
+    sd_ratio = {}
+    for unit_id in unit_ids:
+        spk_amp = []
+
+        for segment_index in range(wvf_extractor.get_num_segments()):
+            spike_train = wvf_extractor.sorting.get_unit_spike_train(unit_id, segment_index=segment_index).astype(
+                np.int64
+            )
+            censored_indices = _find_duplicated_spikes_keep_first_iterative(spike_train, censored_period)
```
samuelgarcia, line 1440: Not sure this is a good idea this cross dependency in this way between curation and qualitymetrics.

samuelgarcia:
Hi Aurelien.
Thanks for this.
If I understand corretly here the sd is computed only at peak amplitude but the 2002 paper is computing the sd on the entire waveform snipet. I think this make a big diffrences no ?
Would it be better to have the sd at waveforms levels with few spikes ( the one already extracted)

DradeAW:
Yes this is different than the 2002 paper (in multiple ways), but I still cited it because it was the first one to write down the idea.

We can talk about it. I found this implementation easier, but you are right there may be more information available by using the whole waveform (although it makes drift correction and censored period trickier).

src/spikeinterface/qualitymetrics/misc_metrics.py
DradeAW: What would you suggest as an alternative?

DradeAW:
The test is failing because of a bug that is solved by #2220

samuelgarcia (approved):

src/spikeinterface/qualitymetrics/misc_metrics.py
alejoe91: that's ok for now, but we can discuss higher-level hierarchy later

alejoe91 (approved):

PR #2148, opened 2023-10-31
Change phy exporter to not save `template_ind` in the case of dense waveform_extractor

### LINKED ISSUE #2023: Dense vs Sparse template file saving for `export_to_phy`
While we are on the thinking about changes to the `export_to_phy`. Phy uses the presence of `template_ind.npy` to assume sparsity whereas for KS2 the file KS2 returns is actually `templates_ind.npy` and so Phy assumes that the templates will be dense. Should spikeinterface use this fact such that if the user provides sparsity it returns an actual `template_ind.npy` so Phy can create the sparse templates and if the user does not provide sparsity/have sparsity the exporter instead will write the `np.ones` to `templates_ind.npy`.

https://github.com/cortex-lab/phylib/blob/2b567cd2e2c382bd5e8926e9abd9a2f9353bfd76/phylib/io/model.py#L691-L727 For how phylib handles sparse vs dense templates.

relevant code lines in spikeinterface
https://github.com/SpikeInterface/spikeinterface/blob/2aabd146b919ac4c907d8fd017115c7c60717486/src/spikeinterface/exporters/to_phy.py#L91-L98
https://github.com/SpikeInterface/spikeinterface/blob/2aabd146b919ac4c907d8fd017115c7c60717486/src/spikeinterface/exporters/to_phy.py#L172-L189

zm711:
Fixes #2023. Just trying to do a cleanup of the issue tracker for 0.99.0. Definitely could go later though.

2 things in this PR
1) As discussed in #2023, there is no point in saving the `template_ind.npy` if the data is dense since it is suppose to be the sparsity for phy. So I just set a boolean to not save `template_ind.npy` if the waveform extractor is dense. Phy will use the lack of `template_ind.npy` to treat it as a dense dataset so Phy's behavior isn't affected (might even be improved since it will treat the data as dense from the get go).

2) with the new default of sparse=True (for `extract_waveforms()`), I'm a little worried that users might not realize their input waveform_extractor is automatically sparse and they might try to layer a sparsity on top. In this case rather than stop the program with an assert I changed it to a warning, so that people know the waveform_extractor is already sparse. But if you want the assert instead that can be reverted.

### DIFF
--- a/src/spikeinterface/exporters/to_phy.py
+++ b/src/spikeinterface/exporters/to_phy.py
@@ -94,13 +94,16 @@ def export_to_phy(
             "argument to enforce sparsity (see compute_sparsity())"
         )
 
+    save_sparse = True
     if waveform_extractor.is_sparse():
         used_sparsity = waveform_extractor.sparsity
-        assert sparsity is None
+        if sparsity is not None:
+            warnings.warn("If the waveform_extractor is sparse the 'sparsity' argument is ignored")
     elif sparsity is not None:
         used_sparsity = sparsity
     else:
         used_sparsity = ChannelSparsity.create_dense(waveform_extractor)
+        save_sparse = False
     # convenient sparsity dict for the 3 cases to retrieve channl_inds
     sparse_dict = used_sparsity.unit_id_to_channel_indices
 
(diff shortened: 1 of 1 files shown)


src/spikeinterface/exporters/to_phy.py
```diff
@@ -94,13 +94,16 @@ def export_to_phy(
             "argument to enforce sparsity (see compute_sparsity())"
         )
 
+    save_sparse = True
     if waveform_extractor.is_sparse():
         used_sparsity = waveform_extractor.sparsity
-        assert sparsity is None
+        if sparsity is not None:
+            warnings.warn("If the waveform_extractor is sparse the 'sparsity' argument is ignored")
```
alejoe91, line 101: Great idea!

alejoe91 (approved):

PR #2149, opened 2023-10-31
Add assert error in the case of unichannel data for sortingview backend + minor clarifications

### LINKED ISSUE #1918: Issue using plot_timeseries function with sortingview backend
Hi, I'm trying to plot my own data using the 'plot_timeseries' function in the widgets module, and have it displayed in the web GUI generated by sortingview. However, when I try to call the function as such:

timeseries = sw.plot_timeseries(recording, backend='sortingview', mode='map')
(I get an assertion error if I call the function without the 'mode="map"' argument)

An empty plot is generated as shown: 
Screenshot 2023-08-11 at 5 22 16 PM

On the other hand, the data plots perfectly normally using the ipywidgets backend:

timeseries = sw.plot_timeseries(recording, backend='ipywidgets')
Screenshot 2023-08-11 at 5 24 56 PM

Additionally, is there any function for plotting principal components that can also be displayed in the sortingview GUI?

Thanks for  the help!

zm711:
Fixes #1918,

I just put an assert to double check that each `traces` doesn't have just one channel (though maybe >1 would be a better assert if my assumption of 0 not being possible is wrong). This should prevent the issue seen in #1918.

Also tried to clarify a few errors asserts and added an assert comment for `clim`.

I think with this I'm done with super tiny PR fixes (for the moment :) ). I'll go back to working on module level assert touch-ups and do another pass of the docs after the release is done.

### DIFF
--- a/src/spikeinterface/widgets/traces.py
+++ b/src/spikeinterface/widgets/traces.py
@@ -85,7 +85,7 @@ class TracesWidget(BaseWidget):
             rec0 = recordings[0]
         else:
             raise ValueError(
-                "plot_traces 'recording' must be recording or dict or list, recording type "
+                "plot_traces 'recording' must be recording or dict or list, recording "
                 f"is currently of type {type(recording)}"
             )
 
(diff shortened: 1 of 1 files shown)


src/spikeinterface/widgets/traces.py
```diff
@@ -473,11 +477,11 @@ def plot_sortingview(self, data_plot, **backend_kwargs):
         try:
             import pyvips
         except ImportError:
-            raise ImportError("To use the timeseries in sorting view you need the pyvips package.")
+            raise ImportError("To use `plot_traces()` in sortingview you need the pyvips package.")
```
zm711, line 480: This is showing up as an unused import? Is this necessary somewhere else?

alejoe91: It is used internally by the `sortingview` backend. I'd leave it here.
Yes it should be `plot_traces` :)

alejoe91 (approved):

src/spikeinterface/widgets/traces.py
zm711: Cool thanks. I've only tested `sortingview` once, so other than Jeremy's talk (& yours) I didn't know too much about it.

PR #2152, opened 2023-11-02
small generation.py doc fix

h-mayorquin:
As discussed in today's meeting.

### DIFF
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -146,6 +146,8 @@ def generate_sorting(
         The refractory period in ms
     add_spikes_on_borders : bool, default: False
         If True, spikes will be added close to the borders of the segments.
+        This is for testing some post-processing functions when they have
+        to deal with border spikes.
     num_spikes_per_border : int, default: 3
         The number of spikes to add close to the borders of the segments.
     border_size_samples : int, default: 20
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/generate.py
```diff
@@ -343,7 +346,7 @@ def synthesize_random_firings(
         The firing rate of each unit (in Hz).
         If float, all units will have the same firing rate.
     add_shift_shuffle: bool, default: False
-        Optionaly add a small shuffle on half spike to autocorrelogram
+        Optionaly add a small shuffle on half of the spikes to make the autocorrelogram less flat.
```
zm711, line 349:
```suggestion
        Optionally add a small shuffle on half of the spikes to make the autocorrelogram less flat.
```
while you're updating :)

zm711 (commented):
One typo that can fixed since you're updating.

src/spikeinterface/core/generate.py
h-mayorquin: Totally escaped us! And it was the three of us x D

src/spikeinterface/core/generate.py
```diff
@@ -186,6 +188,7 @@ def generate_sorting(
         spikes_in_seg["segment_index"] = segment_index
         spikes.append(spikes_in_seg)
 
+        #
```
alejoe91, line 191:
```suggestion
```

alejoe91 (approved):

PR #2153, opened 2023-11-02
Add warning that `fork` for the `mp_context` is not recommended for macOS

zm711:
Actually I was reading up on multiprocessing and learned that as of 3.8 `fork` should really be avoided on macOS since the resource manager might spin threads. Docs [here](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods). 

Although fork can be used on macOS it is not guaranteed to work safely ( and now `spawn` is the macOS and Windows default), so I added a warning into the job_tools to warn users that setting to fork on macOS is no longer considered safe.

I could change Linux to "not macOS POSIX-based systems" if you'd like though.

+ I noticed one typo in the workflow I wrote for `check_your_install` so I fixed that.

### DIFF
--- a/.github/workflows/installation-tips-test.yml
+++ b/.github/workflows/installation-tips-test.yml
@@ -36,4 +36,4 @@ jobs:
       run: python ./installation_tips/check_your_install.py --ci # ci flag turns off gui
     - name: Windows cleanup
       if: ${{ matrix.label == 'windows' }}
-      run: python .installation_tips/cleanup_for_windows.py
+      run: python ./installation_tips/cleanup_for_windows.py
--- a/.github/workflows/installation-tips-test.yml
+++ b/.github/workflows/installation-tips-test.yml
@@ -36,4 +36,4 @@ jobs:
       run: python ./installation_tips/check_your_install.py --ci # ci flag turns off gui
     - name: Windows cleanup
       if: ${{ matrix.label == 'windows' }}
-      run: python ./installation_tips/cleanup_for_windows.py
+      run: python .installation_tips/cleanup_for_windows.py
--- a/doc/modules/core.rst
+++ b/doc/modules/core.rst
@@ -518,9 +518,9 @@ These are a set of keyword arguments which are common to all functions that supp
     A float like 0.5 means half of the availables core.
 * progress_bar: bool
     If True, a progress bar is printed
-* mp_context: str or None
-    Context for multiprocessing. It can be None (default), "fork" or "spawn".
-    Note that "fork" is only available on UNIX systems (not Windows)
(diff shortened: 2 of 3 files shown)


h-mayorquin:
The standard advice is that fork should not be used anywhere. They are changing the default in python 3.13. I have say before that we should change the default in general.

h-mayorquin:
See here:
https://github.com/python/cpython/issues/84559

@alejoe91 @samuelgarcia should we change this? 

I know that fork is faster rand probably faster in some cases but the default should not be something that could be unsafe. We should make the people that want to take the risk opt-in instead of opt-out.

h-mayorquin:
See also:
https://discuss.python.org/t/concerns-regarding-deprecation-of-fork-with-alive-threads/33555

zm711:
We could definitely just add a general warning for all OSs instead, but if mac or linux users want to force it for now they still can. At least for mac the default context from multiprocessing has already switched. But I'm with @h-mayorquin, in that risky behavior should be opt-in instead of opt-out.

alejoe91 (approved):

zm711:
So I fixed my elif which would give the warning even if the user gave 'spawn' for Mac. And added a general warning. I suppose we could just wait for python to switch the mp context itself, although maybe in a separate PR we could discuss switching the default to 'spawn' no matter what?

alejoe91:
Guys, let's discuss more about adding the `warning` on linux in general. I would say to wait until the next release!

zm711:
> Guys, let's discuss more about adding the warning on linux in general. I would say to wait until after the release!

Done

samuelgarcia:
Just to clarify. We need to keep fork in linux. This is more efficient.
The warning is when we already have thread which is not our case at all. So it is safe.
We can optionally have thread starting in the forked process but our default is one thread.
So I think our fork is safe in linux.
I am OK for the darwin warning but not at all when linux.
I also think that even when python will remove "fork" as default we should maintain the "fork" as default on linux on our side as soon as it work because we don't have threads.
In short lets merge this since Zach removed it.

samuelgarcia (approved):

h-mayorquin:
The intention was doing after release, but is not about threads Sam. Read the discussion by the python mantainers that I shared. Quote:

> While that is a great thing for you to do, unfortunately on POSIX doing enough isn’t actually possible. It is impossible to safely code threads that can be used in an application that calls fork() in Python **because the CPython runtime itself is by definition unsafe for use after fork().** 

https://discuss.python.org/t/concerns-regarding-deprecation-of-fork-with-alive-threads/33555


Se this for a recollection of problems.
https://github.com/python/cpython/issues/84559

PR #2157, opened 2023-11-02
Use python methods instead of parsing and eleminate try-except in to_dict

h-mayorquin:
Small improvements

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -316,31 +316,62 @@ class BaseExtractor:
         recursive: bool = False,
     ) -> dict:
         """
-        Make a nested serialized dictionary out of the extractor. The dictionary produced can be used to re-initialize
-        an extractor using load_extractor_from_dict(dump_dict)
+        Construct a nested dictionary representation of the extractor.
+
+        This method facilitates the serialization of the extractor instance by converting it
+        to a dictionary. The resulting dictionary can be used to re-initialize the extractor
+        through the `load_extractor_from_dict` function.
+
+        Examples
+        --------
+        >>> dump_dict = original_extractor.to_dict()
+        >>> reloaded_extractor = load_extractor_from_dict(dump_dict)
 
         Parameters
         ----------
-        include_annotations: bool, default: False
-            If True, all annotations are added to the dict
-        include_properties: bool, default: False
-            If True, all properties are added to the dict
-        relative_to: str, Path, or None, default: None
-            If not None, files and folders are serialized relative to this path
-            Used in waveform extractor to maintain relative paths to binary files even if the
-            containing folder / diretory is moved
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/base.py
```diff
@@ -366,27 +366,24 @@ def to_dict(
...
+        class_name = f"{module_import_path}.{class_name_no_path}"  # e.g. 'spikeinterface.core.generate.AClass'
         module = class_name.split(".")[0]
-        imported_module = importlib.import_module(module)
 
-        try:
-            version = imported_module.__version__
-        except AttributeError:
-            version = "unknown"
+        imported_module = importlib.import_module(module)
+        spike_interface_version = getattr(imported_module, "__version__", "unknown")
```
alejoe91, line 376:
```suggestion
        module_version = getattr(imported_module, "__version__", "unknown")
```
@h-mayorquin this could also be something external/on top of to SpikeInterface, like SpikeForest

h-mayorquin: Ah! good to know.

h-mayorquin: OK, modified this and improved the docstring a wee more.

src/spikeinterface/core/base.py
```diff
@@ -316,35 +316,69 @@ def to_dict(
...
-        relative_to: str, Path, or None, default: None
-            If not None, files and folders are serialized relative to this path
-            Used in waveform extractor to maintain relative paths to binary files even if the
-            containing folder / diretory is moved
-        folder_metadata: str, Path, or None
-            Folder with numpy `npy` files containing additional information (e.g. probe in BaseRecording) and properties.
-        recursive: bool, default: False
-            If True, all dicitionaries in the kwargs are expanded with `to_dict` as well
+        include_annotations : bool, default: False
+            Whether to include all annotations in the dictionary, by default False.
```
alejoe91, line 333:
```suggestion
            Whether to include all annotations in the dictionary
```
These were removed on purpose and moced to the "type" line

alejoe91: I would also vote to end the description of each param without a "."

What do you guys think?

h-mayorquin: I don't have anything against but putting punctuation at the end is too hardwired for me so I will keep making this mistake. At this level I think we should implement a spellcheck like in roiextractors or neuroconv

h-mayorquin:
Actually highly relevant for this:
https://docs.python.org/3/glossary.html#term-qualified-name

I was not aware that python had a very specific terminology for this. This is great.

https://peps.python.org/pep-3155/

samuelgarcia:
OK for me.
thanks for sharing this doc.

alejoe91 (approved):

PR #2158, opened 2023-11-02
`WaveformExtractor.is_extension` --> `has_extension`

DradeAW:
Fixed typo

### DIFF
--- a/src/spikeinterface/core/waveform_extractor.py
+++ b/src/spikeinterface/core/waveform_extractor.py
@@ -557,7 +557,7 @@ class WaveformExtractor:
                 )
 
     def is_extension(self, extension_name) -> bool:
-        warn("WaveformExtractor.is_extension is deprecated! Use `has_extension` instead.")
+        warn("WaveformExtractor.is_extension is deprecated and will be removed in version 0.102.0! Use `has_extension` instead.")
         return self.has_extension(extension_name)
 
     def load_extension(self, extension_name: str):
--- a/src/spikeinterface/core/waveform_extractor.py
+++ b/src/spikeinterface/core/waveform_extractor.py
@@ -558,7 +558,9 @@ class WaveformExtractor:
 
     def is_extension(self, extension_name) -> bool:
         warn(
-            "WaveformExtractor.is_extension is deprecated and will be removed in version 0.102.0! Use `has_extension` instead.", DeprecationWarning, stacklevel=2
+            "WaveformExtractor.is_extension is deprecated and will be removed in version 0.102.0! Use `has_extension` instead.",
+            DeprecationWarning,
+            stacklevel=2,
         )
         return self.has_extension(extension_name)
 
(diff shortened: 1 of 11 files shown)


DradeAW:
This is done,

I added a deprecation warning to `is_extension`

src/spikeinterface/core/waveform_extractor.py
```diff
@@ -556,7 +556,11 @@ def is_extension(self, extension_name) -> bool:
                     and "params" in self._waveforms_root[extension_name].attrs.keys()
                 )
 
-    def load_extension(self, extension_name):
+    def is_extension(self, extension_name) -> bool:
+        warn("WaveformExtractor.is_extension is deprecated! Use `has_extension` instead.")
```
h-mayorquin, line 560: We should say exactly when it will be deprecated.

h-mayorquin (changes requested):
Let's make the timeline of deprecations explicit. I have seen many stuff that says it will be deprecated but I don't know when it was introduced.

I suggest that you mark this has deprecated in two versions down the line.

src/spikeinterface/core/waveform_extractor.py
DradeAW: Well it's deprecated since now, but will be removed in version xx
I thought I would leave the choice of which version to you guys :)

zm711:
```suggestion
        warn("WaveformExtractor.is_extension is deprecated and will be removed in version 0.102.0! Use `has_extension` instead.")
```
Since this will make it into v100: 2 down the line would be 102 if @h-mayorquin wants two versions.  (2 seems fair although I could also see an argument for 1 depending on release speeds). 

I agree overall that the code base does have some deprecation warnings without explicit removal versions which makes it hard to clean up some of the legacy code so I think moving forward having a specific version number would be nice. @alejoe91 do you have a desired number of versions for deprecations like this just so everyone can be on the same page.

h-mayorquin: Let's do two versions from the current one then?

h-mayorquin: Thanks, this is fine for me.

samuelgarcia:
OK for me

h-mayorquin (approved):
OK, for me as well
Thanks @DradeAW

src/spikeinterface/core/waveform_extractor.py
```diff
@@ -556,7 +556,13 @@ def is_extension(self, extension_name) -> bool:
                     and "params" in self._waveforms_root[extension_name].attrs.keys()
                 )
 
-    def load_extension(self, extension_name):
+    def is_extension(self, extension_name) -> bool:
+        warn(
+            "WaveformExtractor.is_extension is deprecated and will be removed in version 0.102.0! Use `has_extension` instead."
+        )
```
alejoe91, line 562:
```suggestion
        warn(
            "WaveformExtractor.is_extension is deprecated and will be removed in version 0.102.0! Use `has_extension` instead.", DeprecationWarning, stacklevel=2
        )
```

alejoe91 (approved):

PR #2161, opened 2023-11-03
(TEST) Fix maxwell tests

alejoe91:


src/spikeinterface/extractors/tests/test_neoextractors.py
```diff
@@ -278,7 +278,6 @@ class CedRecordingTest(RecordingCommonTestSuite, unittest.TestCase):
     ]
 
 
-@pytest.mark.skipif(ON_GITHUB, reason="Maxwell plugin not installed on GitHub")
```
h-mayorquin, line 281: Should we still spike them locally if the user does not have the plugin? Otherwise we will have fails when we run the tests locally.

alejoe91: It auto-installs if needed

h-mayorquin: Ah great!

samuelgarcia (approved):

PR #2163, opened 2023-11-03
Make `dump_to_pickle` great again!

h-mayorquin:
I tesed in core, let's hope it does not break anything else
@DradeAW this should allow you to use `relative_to` with pickle in a fast way. Can you test it for your use case and if it fails show me that json work for that case with a small example?

```python
---------------------
results for branch main
---------------------
---------------------
Pickling with dump_to_pickle
Time elapsed for saving: 0.18831443786621094
Time elapsed for loading: 3.6071527004241943
Memory used by loading: 88.28515625 MiB
---------------------
Pickling directly
Time elapsed for saving: 0.00238800048828125
Time elapsed for loading: 0.14467382431030273
Memory used by loading: 0.625 MiB
Peak memory usage: 285.4609375 MiB

---------------------
results for branch pickle_fix
---------------------
---------------------
Pickling with dump_to_pickle
Time elapsed for saving: 0.06076359748840332
Time elapsed for loading: 0.03839468955993652
Memory used by loading: 0.125 MiB
---------------------
Pickling directly
Time elapsed for saving: 0.0024383068084716797
Time elapsed for loading: 0.03805899620056152
Memory used by loading: 1.125 MiB
Peak memory usage: 194.58203125 MiB

```
Check this script:
https://gist.github.com/h-mayorquin/10fa5b0594d5ce71f0ada1bc7f72f0ec

Main drawback:
The implementation relies on a very weak of metaprograming (we could use global variables as well but I like them even less). Maybe there is a better way of doing this but I just wanted to show a proof of concept.

If this works for @DradeAW I will clean this further.

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -430,14 +430,14 @@ class BaseExtractor:
             The loaded extractor object
         """
         # for pickle dump relative_path was not in the dict, this ensure compatibility
-        if dictionary.get("relative_paths", False):
+        if base_folder is not None:
             assert base_folder is not None, "When  relative_paths=True, need to provide base_folder"
             dictionary = _make_paths_absolute(dictionary, base_folder)
         extractor = _load_extractor_from_dict(dictionary)
         folder_metadata = dictionary.get("folder_metadata", None)
         if folder_metadata is not None:
             folder_metadata = Path(folder_metadata)
-            if dictionary.get("relative_paths", False):
+            if base_folder is not None:
                 folder_metadata = base_folder / folder_metadata
             extractor.load_metadata_from_folder(folder_metadata)
         return extractor
@@ -642,21 +642,41 @@ class BaseExtractor:
         if relative_to:
             relative_to = Path(file_path).parent if relative_to is True else Path(relative_to)
             relative_to = relative_to.resolve().absolute()
-            # if relative_to is used, the dictionaru needs recursive expansion
-            recursive = True
-        else:
-            recursive = False
+            self._kwargs = self._make_paths_in_object_relative(
+                self._kwargs,
+                relative_to,
+            )
 
         dump_dict = self.to_dict(
             include_annotations=True,
             include_properties=include_properties,
             folder_metadata=folder_metadata,
-            relative_to=relative_to,
-            recursive=recursive,
         )
         file_path = self._get_file_path(file_path, [".pkl", ".pickle"])
 
         file_path.write_bytes(pickle.dumps(dump_dict))
+
+    def _make_paths_in_object_relative(self, obj: Any, relative_to: Path) -> Any:
+        """
+        Recursively traverse through the object to find and modify file paths to be relative.
+        """
+        if isinstance(obj, BaseExtractor):
+            # If obj is an instance of the current class, check its _kwargs
+            new_kwargs = {k: self._make_paths_in_object_relative(v, relative_to) for k, v in obj._kwargs.items()}
+            obj._kwargs = new_kwargs
+        elif isinstance(obj, list):
+            # If obj is a list, check each element
+            return [self._make_paths_in_object_relative(item, relative_to) for item in obj]
+        elif isinstance(obj, dict):
+            # If obj is a dict, check each value
+            return {k: self._make_paths_in_object_relative(v, relative_to) for k, v in obj.items()}
+        elif isinstance(obj, (str, Path)):
+            # If obj is a file path, make it relative
+            path = Path(obj)
+            if path.is_absolute():
+                relative_path = os.path.relpath(path, start=relative_to)
+                return relative_path
+        return obj
 
     @staticmethod
     def load(file_path: Union[str, Path], base_folder: Optional[Union[Path, str, bool]] = None) -> "BaseExtractor":
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -403,7 +403,7 @@ class BaseExtractor:
         if relative_to is not None:
             relative_to = Path(relative_to).resolve().absolute()
             assert relative_to.is_dir(), "'relative_to' must be an existing directory"
-            dump_dict = _make_paths_relative(dump_dict, relative_to, copy=False)
+            dump_dict = _make_paths_relative(dump_dict, relative_to)
 
         if folder_metadata is not None:
             if relative_to is not None:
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/base.py
```diff
@@ -403,7 +403,7 @@ def to_dict(
         if relative_to is not None:
             relative_to = Path(relative_to).resolve().absolute()
             assert relative_to.is_dir(), "'relative_to' must be an existing directory"
-            dump_dict = _make_paths_relative(dump_dict, relative_to)
+            dump_dict = _make_paths_relative(dump_dict, relative_to, copy=False)
```
h-mayorquin, line 406:
```suggestion
            dump_dict = _make_paths_relative(dump_dict, relative_to)
```

samuelgarcia:
Cool Ramon.
Impressive.
Really nice solution for saving.
Not sure I like the class attribute for loading but this should be easy to go for similar solution.

src/spikeinterface/core/base.py
```diff
@@ -977,16 +1005,16 @@ def save_to_zarr(
...
-def _make_paths_relative(d, relative) -> dict:
+def _make_paths_relative(d, relative, copy=True) -> dict:
     relative = str(Path(relative).resolve().absolute())
     func = lambda p: os.path.relpath(str(p), start=relative)
-    return recursive_path_modifier(d, func, target="path", copy=True)
+    return recursive_path_modifier(d, func, target="path", copy=copy)
 
 
-def _make_paths_absolute(d, base):
+def _make_paths_absolute(d, base, copy=True):
```
samuelgarcia, line 1014: This is a bad idea to expose at top level.
The first call of the recursive must be a copy always to avoid complicated inplace modification.
I do not see where we need copy=False.

Here th logic was copy the first call always and then internally no copy.

h-mayorquin: I was experimenting with this. Not important to me. Will revert. Sorry for the distraction.

samuelgarcia: ok

DradeAW:
I find the use of a class variable in this case pretty elegant, good find @h-mayorquin!

I'll test if it works on my hand and get back to you :)

samuelgarcia:
@DradeAW : did you made some tests ?

src/spikeinterface/core/base.py
```diff
@@ -985,7 +1013,7 @@ def _make_paths_relative(d, relative) -> dict:
 
 def _make_paths_absolute(d, base):
     base = Path(base)
-    func = lambda p: str((base / p).resolve().absolute())
+    func = lambda p: str((base / p).resolve())
```
samuelgarcia, line 1016: why ?

h-mayorquin: Redundant. Resolve does what absolute does and better:
https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve

DradeAW:
TLDR: It works (can move the folder elsewhere, low RAM usage), but is **extremely slow**!!

Output on `main` with no `relative_to`:
```
**********************************
****** units_categorization ******
**********************************
Loading sortings from previous run...
100%|███████████████████████████████████████████████████████| 9/9 [00:00<00:00, 370.39it/s]


**********************************
******* remove_bad_units_0 *******
**********************************
Loading sortings from previous run...
100%|███████████████████████████████████████████████████████| 9/9 [00:00<00:00, 157.69it/s]


**********************************
******** remove_bad_units ********
**********************************
Loading sortings from previous run...
100%|████████████████████████████████████████████████████████| 9/9 [00:00<00:00, 55.82it/s]


**********************************
********** align_units ***********
**********************************
Loading sortings from previous run...
100%|███████████████████████████████████████████████████████| 9/9 [00:00<00:00, 193.29it/s]


**********************************
**** remove_duplicated_spikes ****
**********************************
Loading sortings from previous run...
100%|████████████████████████████████████████████████████████| 9/9 [00:05<00:00,  1.75it/s]


**********************************
******* remove_bad_units_2 *******
**********************************
Loading sortings from previous run...
100%|████████████████████████████████████████████████████████| 9/9 [00:05<00:00,  1.78it/s]


**********************************
******* remove_bad_units_3 *******
**********************************
Loading sortings from previous run...
100%|████████████████████████████████████████████████████████| 9/9 [00:05<00:00,  1.77it/s]


**********************************
********** merge_units ***********
**********************************
Loading sortings from previous run...
100%|████████████████████████████████████████████████████████| 9/9 [00:06<00:00,  1.49it/s]


**********************************
*** remove_duplicated_spikes_2 ***
**********************************
Loading sortings from previous run...
100%|████████████████████████████████████████████████████████| 9/9 [00:09<00:00,  1.10s/it]


**********************************
******** export_to_phy_0 *********
**********************************
Loading sortings from previous run...
100%|████████████████████████████████████████████████████████| 9/9 [00:09<00:00,  1.10s/it]
```

Output on this PR with `relative_to`:
```
**********************************
****** units_categorization ******
**********************************
Loading sortings from previous run...
100%|███████████████████████████████████████████████████████| 9/9 [00:00<00:00, 179.13it/s]


**********************************
******* remove_bad_units_0 *******
**********************************
Loading sortings from previous run...
100%|████████████████████████████████████████████████████████| 9/9 [00:00<00:00, 45.89it/s]


**********************************
******** remove_bad_units ********
**********************************
Loading sortings from previous run...
100%|████████████████████████████████████████████████████████| 9/9 [00:00<00:00, 17.73it/s]


**********************************
********** align_units ***********
**********************************
Loading sortings from previous run...
100%|████████████████████████████████████████████████████████| 9/9 [01:02<00:00,  6.99s/it]


**********************************
**** remove_duplicated_spikes ****
**********************************
Loading sortings from previous run...
 11%|███████████▊                                                                                              | 1/9 [21:44<2:53:57, 1304.63s/it]
% Interrupted because way to slow
```

h-mayorquin:
@DradeAW slow compared to what? Slower than what it was?

Can you also share what you are profiling?

DradeAW:
I run my algorithm lussac (which generated sorting objects that, at each step, are more and more complicated).
I showed the loading time of each sorting at each step (9 sortings), and showed the time for loading on `main` and on this PR.

h-mayorquin:
@DradeAW Btw if you don't have time and this is not that vital for you I am fine with lettings this go. In fact, I kind of prefer to not support `relative_to` to begin with.

DradeAW:
It's not vital, but it would be really convenient to be able to move the processed analyses around.
I'll try to make a simplified case to reproduce this behaviour!

If you want more details, a call would make things easier :)

h-mayorquin:
Thanks for the simplified case, I will save that call for later once I actually understand what's goin on -or I am confused about the right things-

PR #2165, opened 2023-11-04
fix compute_matching_matrix

samuelgarcia:
@alejoe91 @h-mayorquin 
A quick and very urgent fix before the webinar next Tuesday.

### DIFF
--- a/src/spikeinterface/comparison/comparisontools.py
+++ b/src/spikeinterface/comparison/comparisontools.py
@@ -201,8 +201,11 @@ def get_optimized_compute_matching_matrix():
             while index2 < len(frames_spike_train2):
                 frame2 = frames_spike_train2[index2]
                 not_a_match = abs(frame1 - frame2) > delta_frames
+
                 if not_a_match:
                     # Go to the next frame in the first train
+                    while frames_spike_train2[lower_search_limit_in_second_train] + delta_frames < frame1:
+                        lower_search_limit_in_second_train += 1
                     break
 
                 # Map the match to a matrix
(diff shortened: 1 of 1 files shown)


alejoe91:
Thanks camarade! I'll run the tutorial on my side and we can release a patch on monday

h-mayorquin:
What was failing? Any example so I can build a non-regression test?

samuelgarcia:
Hi Ramon.
I you run this tutorial : https://github.com/SpikeInterface/spiketutorials/tree/master/Official_Tutorial_SI_0.99_Nov23
The section "9. Spike sorting comparison" give a total blank agreement matrix.
You can download the dataset on the link.

It took some time to discover that the index2 have to be moved when no match.
When doing this simple patch the matrix looks like the old one but I am not 100% sure that this is correct.
I think to reproduce this we need a case where the index1 is advancing and the index2 is left behind or something like that.

h-mayorquin:
Let me take a look! I have two hours before the next person comes to see the house.

h-mayorquin:
@samuelgarcia 
I can't run the full tutorial, I don't have the files for running kilosort 2 locally and the oen from docker is telling me - wrongly, I think- that I don't have the required nvidia. 

I think it is easier if you send me the pickled sorters by slack then I can track the nature of the bug.

Also, I have no idea why the full tests are failing but I ran the comparison tools tests locally in this branch and they are passing.


What do you think of just reverting the function for the release to the old version. I left the basic components still there:

https://github.com/SpikeInterface/spikeinterface/blob/f725a406f3cc16452636f5bf0652a5d541571be2/src/spikeinterface/comparison/comparisontools.py#L8-L37

I fixed some of the bugs that the old function had but it seems that I introduced some others. We can then fix the other bugs post-release.

h-mayorquin:
But by the way if you feel confident that this is a good fix feel free to merge. I just think that is less safe than either releasing with the old version or having two people understand the nature of the bug (I can't not really understand it by looking at the diff).

alejoe91:
@h-mayorquin I can send you the sortings

PR #2167, opened 2023-11-04
Fix proper search make matrix match count

h-mayorquin:
Related to #2165.

Simplify the implementation and solve a bug where no proper search was taking place in the second spike train.

This is a mistake for trying to write overly clever code with whiles.

Now the code is simpler as the logic is simple.
For every frame:
   1) Find the first match. Move the minimal search point here
   2) From the first match count as many matches onwards until no more matches are available.


src/spikeinterface/comparison/tests/test_comparisontools.py
```diff
@@ -135,6 +135,23 @@ def test_make_match_count_matrix_repeated_matching_but_no_double_counting():
     assert_array_equal(result.to_numpy(), expected_result)
 
 
+def test_make_match_count_matrix_test_proper_search_in_the_second_train():
+    "Search exhaustively in the second train, but only within the delta_frames window"
+    frames_spike_train1 = [500, 600, 650]
```
h-mayorquin, line 140:
```suggestion
    frames_spike_train1 = [500, 600, 800]
```

src/spikeinterface/comparison/tests/test_comparisontools.py
```diff
@@ -135,6 +135,23 @@ def test_make_match_count_matrix_repeated_matching_but_no_double_counting():
...
+    frames_spike_train2 = [0, 100, 200, 300, 500, 800]
+    unit_indices1 = [0, 0, 0]
+    unit_indices2 = [0, 0, 0, 0, 0, 0]
+    delta_frames = 20
+
+    sorting1, sorting2 = make_sorting(frames_spike_train1, unit_indices1, frames_spike_train2, unit_indices2)
+
+    result = make_match_count_matrix(sorting1, sorting2, delta_frames=delta_frames)
+
+    expected_result = np.array([[1]])
```
h-mayorquin, line 150:
```suggestion
    expected_result = np.array([[2]])
```

PR #2168, opened 2023-11-04
Add import specific documentation

zm711:
Related to #2046. (But doesn't close it). 

This is just a starting pass, so no need to rush to review yet, especially with the webinar coming up :).

I figured before doing a pass at the docs I want to get the import portion squared away. This PR moves import to its own section. Previously it was part of the core module, but if someone wants to look for importing the package it is probably nice for it to have its own searchable tab rather than be buried in one of the modules docs.

### DIFF
--- a/doc/import.rst
+++ b/doc/import.rst
@@ -1,0 +1,81 @@
+Importing SpikeInterface
+========================
+
+SpikeInterface allows for the generation of powerful and reproducible spike sorting pipelines.
+Flexibility is built into the package starting from import to maximize the productivity of
+the developer and the scientist. Thus there are three ways that SpikeInterface and its components
+can be imported:
+
+
+Importing by Module
+-------------------
+
+Since each spike sorting pipeline involves a series of often repeated steps, many of the developers
+working on SpikeInterface recommend importing in a module by module fashion. This will allow you to
+keep track of your processing steps (preprocessing, postprocessing, quality metrics, etc.). This can
+be accomplished by:
+
+.. code-block:: python
+
+    import spikeinterface as si
+
+to import the :code:`core` module followed by:
+
+.. code-block:: python
+
+    import spikeinterface.extractors as se
+    import spikeinterface.preprocessing as spre
+    import spikeinterface.sorters as ss
+    import spikinterface.postprocessing as spost
+    import spikeinterface.qualitymetrics as sqm
+    import spikeinterface.exporters as sexp
+    import spikeinterface.comparsion as scmp
+    import spikeinterface.curation as scur
+    import spikeinterface.sortingcomponents as sc
+    import spikeinterface.widgets as sw
+
+to import any of the other modules you wish to use.
+
+The benefit of this approach is that it is lighter and faster than importing the whole package and allows
+you to choose which of the modules you actually want to use. If you don't plan to export the results out of
+SpikeInterface than you don't have to :code:`import spikeinterface.exporters`. Additionally the documentation
+of the package is set-up in a modular fashion, so if you have a problem with :code:`spikeinterface.curation`,
+you will know to go to the :code:`curation` section of this documention. The disadvantage of this approach is
+that you have more aliases to keep track of.
+
+
+Flat Import
+-----------
+
+A second option is to import the SpikeInterface package in :code:`full` mode. This would be similar to
+what is seen with packages like NumPy (:code:`np`) or Pandas (:code:`pd`). To accomplish this one does:
+
+
+.. code-block:: python
+
+    import spikeinterface.full as si
+
+
+This import statement will import all of SpikeInterface modules as one flattened module.
+Note that importing :code:`spikeinterface.full` will take a few extra seconds, because some modules use
+just-in-time :code:`numba` compilation performed at the time of import.
+We recommend this approach for advanced users, since it requires a deeper knowledge of the API. The advantage
+being that users with advanced API knowledge can access all functions using one alias.
+
+
+Importing Individual Functions
+------------------------------
+
+Finally, some users may find it useful to have extremely light imports and only import the exact functions
+they plan to use. This can easily be accomplished by importing functions directly into the name space.
+
+For example:
+
+.. code-block:: python
+
+    from spikeinterface.preprocessing import bandpass_filter, common_reference
+    from spikeinterface.core import extract_waveforms
+    from spikeinterface.extractors import read_binary
+
+As mentioned this approach only imports exactly what you plan on using so is the most minimalist. It does require
+knowledge of the API to know which module to pull a function from.
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -14,7 +14,7 @@ To install the current release version, you can use:
 
 The :code:`[full]` option installs all the extra dependencies for all the different sub-modules.
 
-Note that if using Z shell (:code:`zsh` - the default shell on mac), you will need to use quotes (:code:`pip install "spikeinterface[full]"`).
+Note that if using Z shell (:code:`zsh` - the default shell on macOS), you will need to use quotes (:code:`pip install "spikeinterface[full]"`).
 
 
 To install all interactive widget backends, you can use:
@@ -63,14 +63,14 @@ as :code:`spikeinterface` strongly relies on these packages to interface with va
 
 
 It is also sometimes useful to have local copies of :code:`neo` and :code:`probeinterface` to make changes to the code. To achieve this, repeat the first set of commands,
-replacing `https://github.com/SpikeInterface/spikeinterface.git` with the appropriate repository in the first code block of this section.
+replacing :code:`https://github.com/SpikeInterface/spikeinterface.git` with the appropriate repository in the first code block of this section.
 
 For beginners
 -------------
 
 We provide some installation tips for beginners in Python here:
 
-https://github.com/SpikeInterface/spikeinterface/tree/master/installation_tips
+https://github.com/SpikeInterface/spikeinterface/tree/main/installation_tips
 
 
 
@@ -89,12 +89,16 @@ Requirements
 Sub-modules have more dependencies, so you should also install:
 
   * zarr
+  * h5py
   * scipy
   * pandas
   * xarray
-  * sklearn
+  * scikit-learn
   * networkx
   * matplotlib
+  * numba
+  * distinctipy
+  * cude-python (for non-macOS users)
 
 
 All external spike sorters can be either run inside containers (Docker or Singularity - see :ref:`containerizedsorters`)
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -40,6 +40,7 @@ SpikeInterface is made of several modules to deal with different aspects of the 
 
     overview
     installation
+    import
     modules/index
     how_to/index
     modules_gallery/index
--- a/doc/modules/core.rst
+++ b/doc/modules/core.rst
@@ -22,37 +22,6 @@ All classes support:
   * multiple segments, where each segment is a contiguous piece of data (recording, sorting, events).
 
 
-Import rules
-------------
-
-Importing the SpikeInterface module
-
-.. code-block:: python
-
-    import spikeinterface as si
-
-will only import the :code:`core` module. Other submodules must be imported separately:
-
-.. code-block:: python
-
-    import spikeinterface.extractors as se
-    import spikeinterface.sorters as ss
-    import spikeinterface.widgets as sw
-
-
-A second option is to import the SpikeInterface package in :code:`full` mode:
-
-.. code-block:: python
-
-    import spikeinterface.full as si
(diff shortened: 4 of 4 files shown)


doc/import.rst
```diff
@@ -56,11 +56,11 @@ what is seen with packages like NumPy (:code:`np`) or Pandas (:code:`pd`). To ac
     import spikeinterface.full as si
 
 
-This import statement will import all of SpikeInterface modules as one flattened module.
+This import statement will import all of the SpikeInterface modules as one flattened module.
 Note that importing :code:`spikeinterface.full` will take a few extra seconds, because some modules use
 just-in-time :code:`numba` compilation performed at the time of import.
 We recommend this approach for advanced users, since it requires a deeper knowledge of the API. The advantage
```
samuelgarcia, line 62:
```suggestion
We recommend this approach for advanced (or lazy) users, since it requires a deeper knowledge of the API. The advantage
```

doc/import.rst
```diff
@@ -56,11 +56,11 @@ what is seen with packages like NumPy (:code:`np`) or Pandas (:code:`pd`). To ac
...
-being that users with advanced API knowledge can access all functions using one alias.
+being that users can access all functions using one alias.
```
samuelgarcia, line 63:
```suggestion
being that users can access all functions using one alias without the need of memorizing all aliases.
```

samuelgarcia:
Hi Zach.
I could not resits in adding my point of view.

zm711:
I think it is fair to add the pros and cons of all views. Needing to memorize a bunch of aliases could be annoying for some users and it is a fair point. :)

doc/import.rst
```diff
@@ -0,0 +1,81 @@
...
+    import spikeinterface.qualitymetrics as sqm
+    import spikeinterface.exporters as sexp
+    import spikeinterface.comparsion as scmp
+    import spikeinterface.curation as scur
+    import spikeinterface.sortingcomponents as sc
+    import spikeinterface.widgets as sw
+
+to import any of the other modules you wish to use.
+
+The benefit of this approach is that it is lighter and faster than importing the whole package and allows
```
h-mayorquin, line 39: I am not sure this is true. The picture is actually quite muddy because of our use of try-except in many parts of the library. Check out the import time profiling in this PR:

https://github.com/SpikeInterface/spikeinterface/actions/runs/6798751083

h-mayorquin: So I think I will stress the mnemonic effects that you were arguing for in the issue. We really don't have performance guarantees yet (but I would love to add that in the future).

doc/import.rst
```diff
@@ -0,0 +1,81 @@
...
+you to choose which of the modules you actually want to use. If you don't plan to export the results out of
+SpikeInterface then you don't have to :code:`import spikeinterface.exporters`. Additionally the documentation
+of the package is set-up in a modular fashion, so if you have a problem with the module :code:`spikeinterface.curation`,
+you will know to go to the :code:`curation` section of this documention. The disadvantage of this approach is
+that you have more aliases to keep track of.
+
+
+Flat Import
```
h-mayorquin, line 47: Did anyone favor this option in the issue? I kind of forgot it. Could we have this to the end?  I am not sure this is what numpy or pandas do btw. Is that the case? They also have `np.linalg` in numpy at least.

doc/import.rst
```diff
@@ -0,0 +1,81 @@
...
+what is seen with packages like NumPy (:code:`np`) or Pandas (:code:`pd`). To accomplish this one does:
+
+
+.. code-block:: python
+
+    import spikeinterface.full as si
+
+
+This import statement will import all of the SpikeInterface modules as one flattened module.
+Note that importing :code:`spikeinterface.full` will take a few extra seconds, because some modules use
```
h-mayorquin, line 60: I also think this is not true. The compilation is performed at import time when you have the package. Because of the try-except block.

doc/import.rst
```diff
@@ -0,0 +1,81 @@
...
+just-in-time :code:`numba` compilation performed at the time of import.
+We recommend this approach for advanced (or lazy) users, since it requires a deeper knowledge of the API. The advantage
```
h-mayorquin, line 62: Is anyone of us using it?
I would recommend the functionality that we are using ourselves as that will be the one that is better tested.

doc/installation.rst
```diff
@@ -89,12 +89,16 @@ Requirements
...
   * scipy
   * pandas
   * xarray
-  * sklearn
+  * scikit-learn
   * networkx
   * matplotlib
+  * numba
+  * distinctipy
+  * cude-python (for non-macOS users)
```
h-mayorquin, line 101:
```suggestion
  * cuda-python (for non-macOS users)
```
?

h-mayorquin (commented):
Thanks for doing this, some comments!

doc/import.rst
zm711: I copied that from your guys old docs. :)

doc/import.rst
zm711: That's fair. It was perhaps an oversimplification to explain a point.

doc/installation.rst
zm711: Thanks. 100% a typo on my port. Haha.

doc/import.rst
zm711: @samuelgarcia , you like this sometimes right?

doc/import.rst
h-mayorquin: I saw that when I was at the end of the PR. Feel free to ignore this and I will correct it later.

doc/import.rst
zm711: In the old docs it went 
Import core
Import full
no mention of by function.

So I just kept the same general flow and added by function at the bottom. Happy to move unless Sam strong wants it above. He added edits to the full paragraph so he definitely took a peek.

doc/import.rst
h-mayorquin: Yes, I mean, I think that your mind is on the right place. For user what matters is that: "all the functions that I could use or want will be available without nesting" so `si.function1`, `si.function2`, `si.function3` will work I think your example is fine from that perspective as it conveys the right idea with a familiar example.

Mine was more a technical caveat by saying that I don't think numpy or pandas are putting ALL their stuff the way that we are doing it:

https://github.com/SpikeInterface/spikeinterface/blob/329197618a9b48ef876d1e8b8e79f07f4abf5e49/src/spikeinterface/full.py#L26

doc/import.rst
h-mayorquin: All right. Makes sense to not make a lot of changes on the same PR. Thanks for explaining yourself.

doc/import.rst
zm711: Very true! They keep more functions more private. I can change the wording slightly so more technically users are put off, but keep the simplicity of the example. Thanks :)

doc/import.rst
h-mayorquin: All right, then the same as below. Feel free to ignore it and I can correct it later. 
I wanted to make you aware that there are actually not great performance differents in import time and show you that we keep a profile of that in the unittests. 

As we become better that difference should increase but it is not very large right now as you can see in the link.

doc/import.rst
zm711: Or you re-write all of spikeinterface so it this one line of documentation is true :P. Pick your path wisely.

h-mayorquin: Long term goal.
I moved try-excepts here and there when I find them.

The biggest thing is the numba import with try-except I believe.

doc/import.rst
zm711: I think this one I'll change. Until we get everything optimized there's no point in being misleading (ie the chance the documentation gets changed is lower just because it drops off people's todo list). Looks like extractors and comparison are the big offenders. Too bad extractors are part of each workflow.

doc/import.rst
zm711: I don't have the docs on my hand, but I read somewhere that try-except is super fast as long as you never trigger the except. Once you trigger the except it becomes terribly slow. But I'll need to read up on it more.

h-mayorquin: It is not the try-except structure. Is the fact that if you have the package in the try part then the package is imported.
We want to delay imports until they are actually used and not load them at import time (googel lazy loading).

For most packages we just put the import in the right place. But for plotting libraries and numba that is difficult. I don't remember what was the deal in the extractor module.

doc/import.rst
```diff
@@ -0,0 +1,85 @@
...
+-------------------
+
+Since each spike sorting pipeline involves a series of often repeated steps, many of the developers
+working on SpikeInterface recommend importing in a module by module fashion. This will allow you to
+keep track of your processing steps (preprocessing, postprocessing, quality metrics, etc.). This can
+be accomplished by:
+
+.. code-block:: python
+
+    import spikeinterface as si
```
zm711, line 20: Based on the speed metrics @h-mayorquin supplied why don't we suggest

`import spikeinterface.core as si`

It has a slight speed advantage directly importing rather than doing without core? And it follows the syntax for all the other submodules?

doc/import.rst
zm711: I see what you mean for this.

doc/import.rst
samuelgarcia: I do not understand the question. Do I like this flat import ? yes I use it almost always.

For me the comparison to numpy/pndas is not the best I think.
I would more compare to scipy many modules that **could be** imported flat.

doc/import.rst
samuelgarcia: I also think that this statement on numba on numba is not true. I do not think numba is making compiling but numba is doing something. Maybe we should remove this numba related story and focus on the main message : "everythink is flat and and bit slower"

doc/import.rst
samuelgarcia: Yes. When I teahc spikeinterface  with tutorial very often I use this spikeinterface.full so participant do not have to focus on the submodules concept and take spikeinterface a a whole.

samuelgarcia:
Merci Zach. This is almost OK for me.

doc/import.rst
alejoe91: When I teach/show tutorials I use the by module approach

doc/import.rst
h-mayorquin: @samuelgarcia 
The question was if that somebody was using it. You have answered it already by saying you do. Thanks.

doc/import.rst
h-mayorquin: I think we should remove the numba story yes

doc/import.rst
```diff
@@ -0,0 +1,85 @@
...
+
+.. code-block:: python
+
+    import spikeinterface.full as si
+
+
+This import statement will import all of the SpikeInterface modules as one flattened module.
+Note that importing :code:`spikeinterface.full` will take a few extra seconds, because some modules use
+just-in-time :code:`numba` compilation performed at the time of import.
+We recommend this approach for advanced (or lazy) users, since it requires a deeper knowledge of the API. The advantage
```
zm711, line 65:
```suggestion
We recommend this approach for advanced (or lazy) users, since it requires a deeper knowledge of the API. The advantage
```

doc/import.rst
```diff
@@ -0,0 +1,85 @@
...
+to go to the :code:`curation` section of this documention. The disadvantage of this approach is that you have
+more aliases to keep track of.
+
+
+Flat Import
+-----------
+
+A second option is to import the SpikeInterface package in :code:`full` mode. This would be similar to
+what is seen with packages like NumPy (:code:`np`) or Pandas (:code:`pd`), which offer the majority of
+their functionality with a single alias and the option to import additional functionality separately.
```
zm711, line 53:
```suggestion
A second option is to import the SpikeInterface package in :code:`full` mode. 
```

doc/import.rst
alejoe91: I think this is perfect!

alejoe91 (approved):
Looks great to me!

doc/import.rst
```diff
@@ -0,0 +1,81 @@
...
+-------------------
+
+Since each spike sorting pipeline involves a series of often repeated steps, many of the developers
+working on SpikeInterface recommend importing in a module by module fashion. This will allow you to
+keep track of your processing steps (preprocessing, postprocessing, quality metrics, etc.). This can
+be accomplished by:
+
+.. code-block:: python
+
+    import spikeinterface as si
```
zm711, line 20:
```suggestion
    import spikeinterface.core as si
```

doc/import.rst
zm711: Done!

samuelgarcia (approved):

alejoe91 (approved):

PR #2171, opened 2023-11-05
Set a more standard  sampling frequency default in ground truth recording

h-mayorquin:
should the default be like this 30_000 Hz which is the more common instead of 25_000?


samuelgarcia:
The 25k was on purpose!
I think this is sane to test the pipeline against a "non standard" freq. Just to check that the 30k we have in mind is not hard coded some where in the code.
30k is not the standard one, this is the dominante one (because of Neuropixel).
Old system are 10k. And some system are 20k.

h-mayorquin:
Should't we just set it like that in our tests then?

Why expose our users to the non-dominant paradigm just for our own-sake?

Maybe it is good to not have the dominant one but I think the decision should be user centric. What is the best defaults for users?

alejoe91:
After discussion, we agree that the recording GT generation is only for testing, so it's ok to have a non-standard frequency

PR #2172, opened 2023-11-06
Grid depth

yger:
Add a 3d mode to the grid_convolution method, such that a putative depth estimate can also be guessed while estimating the position of the cells. Note that only the relative depth of the cells is correct, i.e. this is not an absolute number, but this could still be usefull for registration methods.


src/spikeinterface/postprocessing/unit_localization.py
```diff
@@ -423,21 +424,26 @@ def compute_grid_convolution(
...
+    # print(template_positions.shape)
     templates = waveform_extractor.get_all_templates(mode="average")
 
     peak_channels = get_template_extremum_channel(waveform_extractor, peak_sign, outputs="index")
     unit_ids = waveform_extractor.sorting.unit_ids
 
     weights_sparsity_mask = weights > sparsity_threshold
 
-    unit_location = np.zeros((unit_ids.size, 2), dtype="float64")
+    if mode == "2d":
```
alejoe91, line 441:
```suggestion
    assert mode in ("2d", "3d"), "mode can be '2d' or '3d'"
    if mode == "2d":
```

src/spikeinterface/postprocessing/unit_localization.py
```diff
@@ -451,23 +457,27 @@ def compute_grid_convolution(
...
-            found_positions += np.dot(dot_products[count], template_positions[nearest_templates])
+            found_positions[:2] += np.dot(dot_products[count], template_positions[nearest_templates])
 
         unit_location[i, :] = found_positions / scalar_products.sum()
 
+        if mode == "3d":
+            best_template = np.argmin(np.linalg.norm(template_positions - unit_location[i, :2], axis=1))
+            w = weights[:, :, best_template]
+            dot_products = np.dot(w[:, channel_mask], global_products) / np.sum(w, axis=1)
+            unit_location[i, 2] = (dot_products * sigma_um).sum() / dot_products.sum()
```
alejoe91, line 479: Why is the implementation different to the one in `src/spikeinterface/sortingcomponents/peak_localization.py`?

yger: This is the same, modulo the loop over spikes in sortingcomponents

yger:
Tests have been added. As said, this implementation is rather simple. We look for the best match on the upsampled grid of template, and we estimate a relative depth via the spread of the template (estimated via the sigma_um). While the value are only relative, it has the same information as the z estimate obtained via the monopolar approximation

samuelgarcia (approved):

PR #2173, opened 2023-11-06
Improvement to `get_empty_units()`

DradeAW:
`sorting.remove_empty_units()` can be very slow when the spike trains aren't cached

I propose to call `count_num_spikes_per_unit()` which is already optimized.
The downside is that it will not remove a unit that is empty in one segment but not empty in another. Is this a problem?

### DIFF
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -366,7 +366,7 @@ class BaseSorting(BaseExtractor):
     def get_non_empty_unit_ids(self):
         num_spikes_per_unit = self.count_num_spikes_per_unit()
 
-        return np.array([unit_id for unit_id in self.unit_ids if num_spikes_per_unit[unit_id] == 0])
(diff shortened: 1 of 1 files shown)


samuelgarcia:
I think "empty" means zeros spike in all segments. No ?

alejoe91:
> I think "empty" means zeros spike in all segments. No ?

Agreed

DradeAW:
Then this PR also fixes that :)

zm711:
Would it be beneficial to put that into a docstring then? So that users can see that the definition of empty is zero spikes in all segments. I would agree that makes for a fair definition, but I can imagine a user attempting to use this function and thinking it is an 'any' style and not an 'all' style.

zm711:
Thanks @DradeAW :)

alejoe91 (approved):

alejoe91:
Good to merge for me. @samuelgarcia wanna take a look?

samuelgarcia (approved):

PR #2174, opened 2023-11-06
Add ignore_timestamps_error to OpenEphysLegacyRecording

### LINKED ISSUE #2107: How to sort only a segment of an Open Ephys recording
Hello

I would appreciate any advice on how to sort only a segment of an open ephys recording. A portion of my recordings have a discontinuity in the timestamps which produces an error like #1194 . I've found that this error is caused by a short ~17 ms discontinuity ~34 ms before the end of the recording. The recording time stamps also do not start at 0, but I have other recordings that also do not start at zero and spike sort successfully (so long as they do not possess the discontinuity at the end).

I attempted to use the ignore_timestamps_errors option described in https://github.com/NeuralEnsemble/python-neo/issues/1210 and https://github.com/NeuralEnsemble/python-neo/pull/1213, but this did not work. I received a syntax error, but perhaps I am not using the most up to date version of the code. I'd rather just cut off the end of the recording in case there are other discontinuities in the middle of the recording that I should be aware of.

Is there a way in spike interface to only extract and sort a certain part of the recording, such as from time step 1 to time step end - 30000? I believe this would occur at:

raw_rec = si.read_openephys(openephys_folder,stream_name='Signals CH')

I investigated the block_index argument, but I'm not sure of the correct syntax to use. I would appreciate any advice that people could provide. I've looked around the issues and documentation and I couldn't find anything that directly addressed this point, but if I missed it, I apologize and please direct me to the correct resource. Thank you!

~Nick

alejoe91:
Should fix #2107

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/openephys.py
+++ b/src/spikeinterface/extractors/neoextractors/openephys.py
@@ -82,7 +82,9 @@ class OpenEphysLegacyRecordingExtractor(NeoBaseRecordingExtractor):
             all_annotations=all_annotations,
             **neo_kwargs,
         )
-        self._kwargs.update(dict(folder_path=str(Path(folder_path).absolute())))
+        self._kwargs.update(
+            dict(folder_path=str(Path(folder_path).absolute())), ignore_timestamps_errors=ignore_timestamps_errors
+        )
 
     @classmethod
     def map_to_neo_kwargs(cls, folder_path, ignore_timestamps_errors=False):


src/spikeinterface/extractors/neoextractors/openephys.py
```diff
@@ -82,7 +82,9 @@ def __init__(
             all_annotations=all_annotations,
             **neo_kwargs,
         )
-        self._kwargs.update(dict(folder_path=str(Path(folder_path).absolute())))
+        self._kwargs.update(
+            dict(folder_path=str(Path(folder_path).absolute())), ignore_timestamps_errors=ignore_timestamps_errors
```
samuelgarcia, line 86: This option is very very dangerous because it do not ensure channel alignement!

alejoe91: unrelated though. If it's an argument, it should be in the `self._kwargs`. The misalignment problem is NEO-related

h-mayorquin (approved):
Is there anything stoping this?

By the way, I think we should use `resolve` instead of `absolute`:

https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve

h-mayorquin:
@samuelgarcia 
Thanks for exlpaining Sam. I guess you and @alejoe91 discussed this previously but I did not understand that from the discussion above. You don't want to propagate the option until is fixed on neo.Got you.

alejoe91:
Closing this because we need a proper fix on the NEO side. See this issue: https://github.com/NeuralEnsemble/python-neo/issues/1351

PR #2175, opened 2023-11-06
Allow precomputing spike trains

DradeAW:
Allows for fast pre-computation of spike trains


DradeAW:
Wow, to my surprise this is not faster!
I don't understand why, maybe I made a mistake somewhere ...

```python
import time


t1 = time.perf_counter()

sorting = si.load_extractor("/mnt/raid0/data/MEArec/1h_3000cells/analyses/ks2_5_pj7-3/sorting")
	sorting.get_unit_spike_train(unit_id)

t2 = time.perf_counter()

sorting = si.load_extractor("/mnt/raid0/data/MEArec/1h_3000cells/analyses/ks2_5_pj7-3/sorting")
sorting.precompute_spike_trains()
	sorting.get_unit_spike_train(unit_id)

t3 = time.perf_counter()

print(f"Old way: {t2-t1:.1f} s")
print(f"New way: {t3-t2:.1f} s")
```

```
Old way: 16.5 s
New way: 18.3 s
```

DradeAW:
Ok this is way better with the latest commit!

```python
import time
import numpy as np



sorting1 = si.load_extractor("/mnt/raid0/data/MEArec/1h_3000cells/analyses/ks2_5_pj7-3/sorting")
sorting2 = si.load_extractor("/mnt/raid0/data/MEArec/1h_3000cells/analyses/ks2_5_pj7-3/sorting")

t1 = time.perf_counter()

for unit_id in sorting1.unit_ids:
	sorting1.get_unit_spike_train(unit_id)

t2 = time.perf_counter()

sorting2.precompute_spike_trains()
for unit_id in sorting2.unit_ids:
	sorting2.get_unit_spike_train(unit_id)

t3 = time.perf_counter()

print(f"Old way: {t2-t1:.1f} s")
print(f"New way: {t3-t2:.1f} s")

for unit_id in sorting1.unit_ids:
	assert np.all(sorting1.get_unit_spike_train(unit_id) == sorting2.get_unit_spike_train(unit_id))
```

```
Old way: 16.3 s
New way: 1.4 s
```

DradeAW:
For reference, my sorting dataset contains 531 units for a total of 11,298,565 spikes (1h recording)

src/spikeinterface/core/core_tools.py
```diff
@@ -934,3 +941,40 @@ def convert_bytes_to_str(byte_value: int) -> str:
         byte_value /= 1024
         i += 1
     return f"{byte_value:.2f} {suffixes[i]}"
+
+
+def spike_vector_to_dict(spike_vector: np.ndarray) -> dict:
+    assert HAVE_NUMBA
```
zm711, line 947:
```suggestion
    assert HAVE_NUMBA, ('numba must be installed to use spike_vector_to_dict with "
    "from_spike_vector=True" set "from_spike_vector=False" to use without numba')
```

I think having a more helpful assert for users that try to precompute will be necessary since installing core will not automatically install numba. This is just a first draft assert statement to help us track where the error is coming from.

src/spikeinterface/core/basesorting.py
```diff
@@ -424,6 +427,31 @@ def get_all_spike_trains(self, outputs="unit_id"):
             spikes.append((spike_times, spike_labels))
         return spikes
 
+    def precompute_spike_trains(self, from_spike_vector: bool = True):
```
zm711, line 430: Since this is in core where numba is not automatically installed wouldn't it be safer to have `from_spike_vector` default to False, so that a user who has only installed core doesn't instantly hit an assert error?

zm711 (commented):
Just a question and suggestion @DradeAW

src/spikeinterface/core/basesorting.py
DradeAW: I added the parameter to be complete, but I actually don't know when it would be interesting to precompute spike trains not from the spike vector.
Can you think of a use case?

src/spikeinterface/core/core_tools.py
DradeAW: I don't like the message here (because maybe this function will be used by another method elsewhere?)

I'll make a commit to have a better error message :)

zm711: Cool. and thanks! 

Yeah it is annoying/poor form to reference another function here. It would be better to have one assert in the original function, but we can't survey for numba there and a separate assert here to also make sure numba is useable.

src/spikeinterface/core/basesorting.py
```diff
@@ -6,8 +6,16 @@
...
+from .core_tools import spike_vector_to_dict
 from .waveform_tools import has_exceeding_spikes
 
+try:
+    import numba
+
+    HAVE_NUMBA = True
+except:
+    HAVE_NUMBA = False
+
```
zm711, line 18: I don't know if @samuelgarcia is going to like all the attempts at importing numba....

src/spikeinterface/core/basesorting.py
zm711: Not off the top of my head. But I'll think about it :)

src/spikeinterface/core/basesorting.py
DradeAW: I just did exactly what samuel and I did in postprocessing/correlograms!
I never liked this, but trying to keep the code consistant (and you don't like my elegant but complicated code)

zm711: I meant in core. Last time I asked why SI didn't just make numba a core dependency he said because he didn't want to add more imports to core to keep it light and fast. 

Nothing against the code just warning you before you work too hard on it. :)

samuelgarcia:
Hi Aurelien.
thanks for this.

I will not have time to get feedback on this very soon.
But globaly:
 * thanks a lot for this speed up, this was on my todo list
 * not sure to like the idea of having numba in core... lets see. It can be used optionally if installed, but then we need
    it for  testing.
* not sure to like the semantic in one function. In short many Sorting are already spiketrains centric (most of then) and 
   some of then are spike train first (numpysorting, ...). We need this function for 2 clear distinct case:
   1. when spike train centric to preload in memory
   2. when spike vector centric to make the convertion all at once. The caching used to be spiketrain per spiketrain which 
      also a valid scenario I think.
  The semantic have to be very clear for this 2 cases. This is now done with one function with one option.
* I like the idea to explicitly call this function.

DradeAW:
> not sure to like the semantic in one function [...] We need this function for 2 clear distinct case

@samuelgarcia You would prefer to split this in two functions with no parameter each?

DradeAW:
> not sure to like the idea of having numba in core... lets see. It can be used optionally if installed, but then we need it for testing.

I removed the need for numba (there is now a default if it's not installed).
For the tests, I've added a skip if numba is not installed, so it will not make the test fail.

src/spikeinterface/core/core_tools.py
```diff
@@ -934,3 +941,41 @@ def convert_bytes_to_str(byte_value: int) -> str:
         byte_value /= 1024
         i += 1
     return f"{byte_value:.2f} {suffixes[i]}"
+
+
+def spike_vector_to_dict(spike_vector: np.ndarray) -> dict:
```
h-mayorquin, line 946: This returns a list doesn't?

src/spikeinterface/core/core_tools.py
```diff
@@ -934,3 +941,41 @@ def convert_bytes_to_str(byte_value: int) -> str:
...
+        spike_vector["unit_index"].astype(np.int64),
+        spike_vector["segment_index"].astype(np.int64),
+    )
+
+    return spike_trains
+
+
+if HAVE_NUMBA:
+
+    @numba.jit((numba.int64[::1], numba.int64[::1], numba.int64[::1]), nopython=True, nogil=True, cache=True)
```
h-mayorquin, line 960: Have you tested not having eager compilation? If you really want to use eager compilation (when I tested this myself in the past it yield no benefits) this should support all the integers types.
Let's not use cache as I am unsure how that plays out with multiprocessing and the numba documentation is full of caveasts.

src/spikeinterface/core/core_tools.py
```diff
@@ -934,3 +941,41 @@ def convert_bytes_to_str(byte_value: int) -> str:
...
+    def _vector_to_dict(sample_index, unit_index, segment_index):
+        spike_trains = []
+        n_units = 1 + np.max(unit_index)
+
+        for seg in range(1 + segment_index[-1]):
+            n_spikes = np.zeros(n_units, dtype=np.int64)
+            for i in range(len(sample_index)):
+                n_spikes[unit_index[i]] += 1
+
+            d = numba.typed.Dict()
```
h-mayorquin, line 970: Let's avoid single variable names that crash with the debugger. More explicit is better but if you want to use a single letter variable use something other than d.

src/spikeinterface/core/core_tools.py
h-mayorquin: Also needs documentation.

src/spikeinterface/core/core_tools.py
```diff
@@ -934,3 +941,41 @@ def convert_bytes_to_str(byte_value: int) -> str:
...
+        spike_vector["segment_index"].astype(np.int64),
+    )
+
+    return spike_trains
+
+
+if HAVE_NUMBA:
+
+    @numba.jit((numba.int64[::1], numba.int64[::1], numba.int64[::1]), nopython=True, nogil=True, cache=True)
+    def _vector_to_dict(sample_index, unit_index, segment_index):
```
h-mayorquin, line 961: Needs a docstring as well,

src/spikeinterface/core/basesorting.py
h-mayorquin: The bare try except do adds to the import time though. That is, in the current status numba will be imported when you import anything from spikeinterface which is undesirable.

At the moment, we either add a more complicated logic like this:
https://github.com/SpikeInterface/spikeinterface/blob/329197618a9b48ef876d1e8b8e79f07f4abf5e49/src/spikeinterface/comparison/comparisontools.py#L111-L232

Which is too complex.
or
We eat some import time.

My preference is for the former. 

But it would be good to come up with something simpler that avoids the import time with numba. For most stuff just moving the import inwards to the function solves the problem.

DradeAW:
> How does this interacts plays with the caching option in `to_spike_vector` and `get_unit_spike_train`.

It overrides it (forces a recompute).

> Isn't it duplicating behavior?

No, because it is a fast computation of all spike trains at once, rather than one unit by one unit (which is, as I showed above, much faster).

> How would this play out with multiprocessing, all this pre-computing is lost isn't it?

I'm not sure, but I believe if you get the spike train it should find it in the cache in all cases?

src/spikeinterface/core/core_tools.py
DradeAW: I haven't tried non-eager compilation (I just personally don't believe in not typing when compiling ^^')
I don't even know how it works to be honest, and the spike vector forces int64 anyway (look at `to_spike_vector`).
But feel free to change it if you prefer it another way!

For the cache, isn't it important?
I mean if you need to recompile it everytime you call it, then you lose a big portion of the time saved?

src/spikeinterface/core/core_tools.py
DradeAW: > This returns a list doesn't?

Correct!
Forgot to change that when I implemented multi-segments ^^'

src/spikeinterface/core/core_tools.py
DradeAW: I though Samuel said it was unnecessary for stuff the user doesn't interact with (such as ExtractorSegments or functions starting with `_`)?

h-mayorquin: If that is what @samuelgarcia believes I think he is wrong on that one. 

Documentation is also for other developers. Also for your future self.

For this specific one it makes my reviewing slower and leads to misundersandings. I don't know what this function is doing  by the name or what is the intention for example. There is a lot of implicit knowledge and intent on it that could be solved with some comments.

src/spikeinterface/core/core_tools.py
h-mayorquin: Concerning typing: 
If you can have a more general function for no price in performance then that's better. I am fine with the typing, just add the more general version. We migt change some of the typing of the spike vector. The typing of the units and the segments is too large for what they represent for example. It is not good to not have this implicit coupling between different parts of the library (the implicit coupling here is the typing of the spike vector with the typing that this function takes).

Concerning cache:

Yes, caching will improve speed when calling multiple times. When profiling you should reflect the two cases. The first run of the function and the second. Users of the library will be exposed to the first use and that is important information.

I am concerned about cache invalidation and how this will play out with multiprocessing:
https://numba.readthedocs.io/en/stable/developer/caching.html#cache-clearing

This option works great when using it for your personal scripts when you have access to your system and know the code well so you can troubleshoot. But this is library code and we don't have control of what environments the users will run the code. Cache invalidation is a big problem and I don't know enough about how this works to take the risk.

I prefer to go with the safe option, then when we have more knowledge and ways of testing this we can enable caching globally for numba. This is better than having to solve difficult bug issues of users. Trade-off between maintanability and optimization.

h-mayorquin:
> No, because it is a fast computation of all spike trains at once, rather than one unit by one unit (which is, as I showed above, much faster).

So are the functions introduced here doing a better job at calculating the `spike_vector` than `to_spike_vector`?

src/spikeinterface/core/basesorting.py
DradeAW: I have to say, I personally hate having a function inside a function.
But if you prefer it that way, I can make it.

src/spikeinterface/core/core_tools.py
DradeAW: Alright, I'll make one then!

src/spikeinterface/core/basesorting.py
h-mayorquin: No, the preferred way right now is to use the try-except block. I was experimenting above.

I am saying that the try-except block adds to the import time.

DradeAW:
> So are the functions introduced here doing a better job at calculating the spike_vector than to_spike_vector?

No, I haven't touched that.
This is in converting the `spike_vector` into spike trains.
The method right now loops over all units, and calls `spike_train = spike_vector[ids == unit_id]`
Which is very slow for big vectors.

This new method allows the conversion in a very efficient manner!

src/spikeinterface/core/core_tools.py
```diff
@@ -934,3 +941,72 @@ def convert_bytes_to_str(byte_value: int) -> str:
...
+
+    Returns
+    -------
+    spike_trains: list[dict]:
+        A list containing, for each segment, the spike trains of all units.
+    """
+    assert HAVE_NUMBA, "spike_vector_to_dict() requires `numba`!"
+
+    spike_trains = _vector_to_dict(
+        spike_vector["sample_index"].astype(np.int64),
```
h-mayorquin, line 963: You should do `copy=False` in as type to avoid the copying when is not necessary. Otherwise this is doing copies all the time:

https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html

src/spikeinterface/core/core_tools.py
```diff
@@ -934,3 +941,72 @@ def convert_bytes_to_str(byte_value: int) -> str:
...
+        segment_index: array
+            spike_vector["segment_index"]
+
+        Returns
+        -------
+        spike_trains: list[dict]
+            A list containing, for each segment, the spike trains of all units.
+        """
+
+        spike_trains = []
```
h-mayorquin, line 993: It is very confusing that you have a variable named `spike_trains` and then another named `spk_trains` with different type and then one goes inside the other.

The outer one could be called `spike_trains_per_segment`.

src/spikeinterface/core/core_tools.py
```diff
@@ -934,3 +941,72 @@ def convert_bytes_to_str(byte_value: int) -> str:
...
+            n_spikes = np.zeros(n_units, dtype=np.int32)
+            for i in range(len(sample_index)):
+                n_spikes[unit_index[i]] += 1
+
+            spk_trains = numba.typed.Dict()
+            for i in range(n_units):
+                spk_trains[i] = np.empty(n_spikes[i], dtype=np.int64)
+
+            ind = np.zeros(n_units, dtype=np.int64)
+            for i in range(len(sample_index)):
```
h-mayorquin, line 1006: Can you use enumerate and some better variable name that `ind`. I think something like `current_x` or `counter_x` where x is the purpose would make your intent more transparent and easier to read.

you have indexes, ind, i, all over the place so it is hard to know what is doing what.

src/spikeinterface/core/core_tools.py
```diff
@@ -934,3 +941,72 @@ def convert_bytes_to_str(byte_value: int) -> str:
         byte_value /= 1024
         i += 1
     return f"{byte_value:.2f} {suffixes[i]}"
+
+
+def spike_vector_to_dict(spike_vector: np.ndarray) -> list[dict]:
```
h-mayorquin, line 946: `dict` is a very unspecific name here.

Can you just use `spike_vector_to_spike_trains` or `spike_trains_dict` at the end if you want to be more specific. 

Then in the documentation you can write that the dicts are dictionary where spike_train_dict[unit_index] gives you all the spikes frames for a given unit. 

I know the library to some degree and it took me a while to understand what you are doing.

src/spikeinterface/core/numpyextractors.py
```diff
@@ -347,13 +347,16 @@ def get_unit_spike_train(self, unit_id, start_frame, end_frame):
...
+        start = 0 if start_frame is None else np.searchsorted(self.spikes_in_seg["sample_index"], start_frame)
+        end = (
+            len(self.spikes_in_seg)
+            if end_frame is None
+            else np.searchsorted(self.spikes_in_seg["sample_index"], end_frame)
+        )
+
         unit_index = self.unit_ids.index(unit_id)
-        times = self.spikes_in_seg[self.spikes_in_seg["unit_index"] == unit_index]["sample_index"]
+        times = self.spikes_in_seg[start:end][self.spikes_in_seg[start:end]["unit_index"] == unit_index]["sample_index"]
```
h-mayorquin, line 358: This will be easier if we did not use the concatenated version of `to_spike_vector`

h-mayorquin:
> No, I haven't touched that.
This is in converting the spike_vector into spike trains.
The method right now loops over all units, and calls spike_train = spike_vector[ids == unit_id]
Which is very slow for big vectors.

Got you,thanks, I realized I was not understand what you were doing and went and re-read again.

h-mayorquin (commented):
OK, I understood what you are doing now and I can offer more comments.

If we go on with this, should the `proposed `cached_spike_train` be an attribute of `BaseSorting` like spike_vector is?

src/spikeinterface/core/core_tools.py
zm711: Actually I really agree with this. In the library sometimes I forget what is a dict and what is a np.array and so it is hard to follow indexing efforts in general.

src/spikeinterface/core/core_tools.py
DradeAW: Copying is necessary in some cases (for example, the spike vector of the `NumpySorting` is a `memmap`, which is incompatible in numba if you enforce the type).

I tried not enforcing the type and running with `copy=False`, and this is ~70% slower on my machine with my sorting object.

h-mayorquin: The suggestion above is to avoid a redundant copy if the types are the same. Currently, if your function is already `np.int64` you will generate a copy anyway.

 If copying is necessary use an explicit function for copying. Do not rely on the implicit copying of `astype`

src/spikeinterface/core/numpyextractors.py
DradeAW: I do not understand?

src/spikeinterface/core/core_tools.py
DradeAW: Done :+1:

src/spikeinterface/core/core_tools.py
DradeAW: Done :+1:

src/spikeinterface/core/numpyextractors.py
h-mayorquin: Unrelated to this PR. But the `to_spike_vector` function has a function to return a list of spike_vectors per segment instead of all the spike_vectors concanteted, so you could avoid the first np.searchsorted which look for the segment indexes.

src/spikeinterface/core/core_tools.py
h-mayorquin: This is a suggestion btw. If you feel strongly about having a variable `ind`, that's up to you.

src/spikeinterface/core/numpyextractors.py
h-mayorquin: https://github.com/SpikeInterface/spikeinterface/blob/329197618a9b48ef876d1e8b8e79f07f4abf5e49/src/spikeinterface/core/basesorting.py#L427

`concatenated` keyword.

src/spikeinterface/core/core_tools.py
DradeAW: The memmap is of type `np.int64`, but it needs to be a `np.ndarray`!
That's a two for one special with the `astype` ^^

src/spikeinterface/core/core_tools.py
DradeAW: I feel strongly that I do not know how to name my variables :joy: 
So having someone to double check whether it's undrstandable is a good idea ahah

src/spikeinterface/core/core_tools.py
DradeAW: Done :+1:

src/spikeinterface/core/numpyextractors.py
DradeAW: Ah ok!

I didn't touch that, but I removed the disgusting `times[times > start]` which doesn't take advantage of the fact that it is ordered!

src/spikeinterface/core/core_tools.py
h-mayorquin: > The memmap is of type np.int64, but it needs to be a np.ndarray!

I think that you are equivocating numeric type with data-type. np.int64 is a numeric data type, np.ndarray is a datatype. Numpy memmaps are np.ndarray and can also have different numeric data types.

src/spikeinterface/core/core_tools.py
h-mayorquin: Thanks!

h-mayorquin: The heuristic is to be specific. What is this representing or used for?

There is no gain for conciseness in programming.

DradeAW: > The heuristic is to be specific. What is this representing or used for?
> There is no gain for conciseness in programming.

I half-agree. Sometimes to be specific the variable name takes 30 characters which I find too much and makes it un-readable.
But I agreed with you with the variables you mentioned above :)

src/spikeinterface/core/core_tools.py
h-mayorquin: I am actually curios about what error you are getting when you pass a memmap to numba.
I am also curios how you spike_vector ended up being a memmap.

DradeAW: > I am actually curios about what error you are getting when you pass a memmap to numba.

`TypeError: No matching definition for argument type(s) readonly array(int64, 1d, A), readonly array(int64, 1d, A), readonly array(int64, 1d, A)`

> I am also curios how you spike_vector ended up being a memmap.

I think this is the default for the `NumpySorting`?

h-mayorquin: Thanks for sharing the error. I was not aware of this:
You should do:
```python
array = np.array(array) if isintance(array, np.memmap) else array
```
This will avoid copy when it is not necessary.

It is not the default of NumpySorting. There are many constructors to the object.

DradeAW: > Thanks for sharing the error. I was not aware of this

I also wasn't until today ^^

> You should do:
> `array = np.array(array) if isintance(array, np.memmap) else array`

I can do that. But if the dtype isn't `np.int64`, are you sure that `copy=False` is the way to go?
It will change the vector, which may not be what the user wants?

h-mayorquin: Are you modifying the array inside the funciton?

DradeAW: If you do `.astype(np.int64, copy=False)` then it changes the array no?
Or have I misunderstood something?

DradeAW: Ah I just read the documentation, and you are right!
`copy=False` only disables copying if everything is correct.

I'll fix that :)

h-mayorquin: Glad you read the documentation.

DradeAW:
Here is the benchmark with a bigger sorting object (the biggest one I ever used):
3,000 units, 1 hour, total of 74,638,923 spikes (same code as shown above, but with a different sorting object)

```
Old way: 614.1 s
New way: 7.7 s
```

This is ~80 times faster.

h-mayorquin:
Yeah, I don't really like having numba for something like this in the core, can you profile against this vectorized version for the last example above?

```python
    def precompute_spike_trains(self):
        """
        Pre-computes and caches all spike trains for this sorting. Should be 
        quicker if you already have the spike vector.

        """
    
        unit_ids = self.unit_ids

        spike_trains = {}
        
        spike_vector_per_segment = self.to_spike_vector(concatenated=False)
        for segment_index, spike_vector in enumerate(spike_vector_per_segment):
            spike_trains_segment = dict()
            for unit_id in unit_ids:
                spike_train = spike_vector[spike_vector['unit_index'] == unit_id]['sample_index']
                spike_trains_segment[unit_id] = spike_train

            spike_trains_segment[segment_index] = spike_trains_segment

        self._cached_spike_trains = spike_trains
```

samuelgarcia:
@zm711 @DradeAW @h-mayorquin 
I made a refactor of this that should made happy all of us (included me)

* `create a sorting_tools.py`
* cache from spike vector or not is automatic depending spike_vector is cached or not
* the function import numba internally in a try cache. If numba is not there then have another implementation
* use the none concatenated version of spike vector which is faster.
* have explicit num_units rather than max(unit_index) + 1
* return directly a dict of dict

Tell me what you think.

src/spikeinterface/core/numpyextractors.py
samuelgarcia: "disgusting" is a bit hard but I get the feeling. Note that generally this `start_frame` \  `end_frame` is very often None.

src/spikeinterface/core/basesorting.py
```diff
@@ -449,6 +451,33 @@ def get_all_spike_trains(self, outputs="unit_id"):
             spikes.append((spike_times, spike_labels))
         return spikes
 
+    def precompute_spike_trains(self, from_spike_vector=None):
```
samuelgarcia, line 454: I do not like too much this method name.
Maybe `cache_all_spike_trains()` would be more explicit

src/spikeinterface/core/numpyextractors.py
samuelgarcia: @h-mayorquin 
Now it use to_spike_vector(concatenated=False) which is better.

src/spikeinterface/core/basesorting.py
DradeAW: To me, I would expect a function `cache_all_spike_trains` to not perform any computation, but just move the result of computation to the cache.
But I agree that the name could maybe be improved

DradeAW:
There are probably a few places in `core` that could be improved by calling this function, because they need the spike train of all unit ids (maybe when extracting waveforms?)

src/spikeinterface/core/sorting_tools.py
```diff
@@ -0,0 +1,90 @@
+import numpy as np
+
+
+def spike_vector_to_spike_trains(spike_vector: list[np.array], unit_ids: np.array) -> dict[dict]:
+    """
+    Computes all spike trains for all units/segments from a spike vector list.
+
+    Internally call numba if numba is installed.
```
zm711, line 8:
```suggestion
    Internally calls numba if numba is installed.
```

src/spikeinterface/core/sorting_tools.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Parameters
+    ----------
+    spike_vector: list[np.ndarray]
+        List of spike vector optained with sorting.to_spike_vector(concatenated=False)
```
zm711, line 13:
```suggestion
        List of spike vectors obtained with sorting.to_spike_vector(concatenated=False)
```

src/spikeinterface/core/sorting_tools.py
```diff
@@ -0,0 +1,90 @@
...
+    unit_ids: np.array
+        Unit ids
+
+    Returns
+    -------
+    spike_trains: dict[dict]:
```
zm711, line 19: is this `list[dict]` or `dict[dict]`?
```suggestion
    spike_trains: list[dict]:
```

src/spikeinterface/core/sorting_tools.py
```diff
@@ -0,0 +1,90 @@
...
+def get_numba_vector_to_list_of_spiketrain():
+    if hasattr(get_numba_vector_to_list_of_spiketrain, "_cached_numba_function"):
+        return get_numba_vector_to_list_of_spiketrain._cached_numba_function
+
+    import numba
+
+    @numba.jit((numba.int64[::1], numba.int64[::1], numba.int64), nopython=True, nogil=True, cache=True)
+    def vector_to_list_of_spiketrain_numba(sample_indices, unit_indices, num_units):
+        """
+        Fast implementation of vetor_to_dict using numba loop.
```
zm711, line 68:
```suggestion
        Fast implementation of vector_to_dict using numba loop.
```

zm711 (commented):
cleanup of docstrings

DradeAW:
Latest benchmark: even faster

~200x times faster on a very big sorting:
```
Old way: 634.3 s
New way: 2.9 s
```

~35 times faster for a sorting I typically use:
```
Old way: 24.9 s
New way: 0.7 s
```

h-mayorquin:
Thanks @samuelgarcia . I introduced some machinery to test this:

https://github.com/SpikeInterface/spikeinterface/pull/2227

I will give it a  test as soon as I can.

src/spikeinterface/core/sorting_tools.py
```diff
@@ -0,0 +1,90 @@
...
+        for s in range(num_spikes):
+            num_spike_per_units[unit_indices[s]] += 1
+
+        spike_trains = []
+        for u in range(num_units):
+            spike_trains.append(np.empty(num_spike_per_units[u], dtype=np.int64))
+
+        current_x = np.zeros(num_units, dtype=np.int64)
+        for s in range(num_spikes):
+            spike_trains[unit_indices[s]][current_x[unit_indices[s]]] = sample_indices[s]
```
h-mayorquin, line 82: This line is a monster to read, guys.
Don't... you see that?

DradeAW: This is a line of code that is going to be executed possibly millions of times,
So when speed is of the essence, I always try to minimize creating variables to go as fast as possible.

Maybe I'm wrong and it makes no difference, but that was my way of thinking

h-mayorquin:
I found no difference (as I expected)

```python
import time
import numpy as np


num_units = 1000
durations = [10 * 60 * 60.0]
seed = 25

sorting1 =  SortingGenerator(num_units=num_units, durations=durations, seed=seed)
sorting2 = SortingGenerator(num_units=num_units, durations=durations, seed=seed)

t1 = time.perf_counter()

for unit_id in sorting1.unit_ids:
	sorting1.get_unit_spike_train(unit_id)

t2 = time.perf_counter()

sorting2.precompute_spike_trains()
for unit_id in sorting2.unit_ids:
	sorting2.get_unit_spike_train(unit_id)

t3 = time.perf_counter()


print(f"Good old : {t2-t1:.1f} s")
print(f"Pre-computing: {t3-t2:.1f} s")

for unit_id in sorting1.unit_ids:
	assert np.all(sorting1.get_unit_spike_train(unit_id) == sorting2.get_unit_spike_train(unit_id))

Good old : 1.6 s
Pre-computing: 1.6 s
```

This might be a phenomenon exclusive to mearec, do you guys suspect that it happens for any other format? Can you c-profile your code to se where is the time spent?

It is very strange to me that we calculate the `spike_trains` through the `spike_vector` which is a terrible representation for that: you mix your spike vectors into one long vector (with a sort as a cost) and then you use a numba function to demix because, well, it was not made for that.

I think that a function to transform `spike_vector` to `spike_trains` through numba makes a lot of sense in the `NumpySorting`. That is where it should be most useful and makes thematic sense. Specially if this is a phenomeon exclusive to Mearec I don't think it is worth adding -more- complexity to the core just for this case.

The design principle is to corall complexity away from the core. Implementation details of specific formats should not leak up as a general methods.  All the current functionality can be coralled to NumpySorting and currently baserecording has a `.to_numpy_sorting` method that enables this easily. That is the place to handle functionality that is useful when you have all the data in memory. The core is for lazy operations.

Now, if most formats are like this I think it would make sense but I really want to keep the core simple. I think the bar should be high.

src/spikeinterface/core/sorting_tools.py
h-mayorquin: You should profile : / 

References are really cheap.

h-mayorquin: Check it out:

```python
import timeit

# Measuring the time taken to create a variable in Python
setup_code = ""
test_code = "a = 10"
number_of_iterations = int(1e8)

# Using timeit to measure the execution time for initial assignment
time_taken_assignment = timeit.timeit(stmt=test_code, setup=setup_code, number=number_of_iterations)

# Measuring the time taken for reassignment in Python
reassignment_code = "b = a"

# Using timeit to measure the execution time for reassignment
time_taken_reassignment = timeit.timeit(stmt=reassignment_code, setup=test_code, number=number_of_iterations)

print("Time taken for initial assignment: ", time_taken_assignment)

# Calculating the additional time taken for reassignment
additional_time = time_taken_reassignment - time_taken_assignment

print("Time taken for reassignment: ", time_taken_reassignment)
print("Additional time for reassignment: ", additional_time)

Time taken for initial assignment:  5.179084311999759
Time taken for reassignment:  5.114150272998813
Additional time for reassignment:  -0.06493403900094563
```
Binding by rererence is within the measrument error for the ten of millions.

DradeAW: Less than a tenth of a million multiplied by tens of millions is undefined :)

But I see your point (plus this is a reassignment in compiled code, not python).
I have to make an effort to make lines more readable!

DradeAW:
Yes pre-computing only makes things faster if you come from a spike vector (in other scenarios it's going to be the same speed).

However the `NumpySorting` is not the only sorting using a spike vector (e.g. `PhySortingExtractor`).
Plus Sam wanted to make a focus on making sorting more spike vector centered.
Plus `NumpySorting` is the default save method, so it might be used a lot.

> Currently baserecording has a `.to_numpy_sorting`

I believe there is a typo here ^^

Thanks for the input!
This should not be taken lightly indeed. However I believe that retrieving spike trains is a core functionality of a sorting object, and thus should be optimized (I posted the massive gain for huge sorting objects).

h-mayorquin:
@DradeAW You are correct. My scenario of performance above is wrong. Here is the corrected scenario:

```python
import time
import numpy as np
from spikeinterface.core.generate import SortingGenerator


num_units = 1000
durations = [10 * 60 * 60.0]
seed = 25

sorting1 =  SortingGenerator(num_units=num_units, durations=durations, seed=seed)
sorting2 = SortingGenerator(num_units=num_units, durations=durations, seed=seed)
sorting2.to_spike_vector()

t1 = time.perf_counter()

for unit_id in sorting1.unit_ids:
	sorting1.get_unit_spike_train(unit_id)

t2 = time.perf_counter()
print(f"Good old : {t2-t1:.1f} s")


t3 = time.perf_counter()
sorting2.precompute_spike_trains()
for unit_id in sorting2.unit_ids:
	sorting2.get_unit_spike_train(unit_id)

t4 = time.perf_counter()


print(f"Pre-computing: {t4-t3:.1f} s")

for unit_id in sorting1.unit_ids:
	assert np.all(sorting1.get_unit_spike_train(unit_id) == sorting2.get_unit_spike_train(unit_id))

Good old : 1.6 s
Using numba to compute spike trains
Pre-computing: 1.9 s
After caching:
Good old : 1.6 s
Using numba to compute spike trains
Pre-computing: 1.4 s
``` 

So, it is a bit slower on the first run but then after you cache I it becomes faster. I would label this as no-gain but also no loss even in the case with really fast-io (that is, the case of my generator). So **I think this supports your case for adding this** as even in this extreme case it performs relleatively well (or not bad). So, yes, empiricism works, I am now in your side, this should work OK in most cases, I still don't like the cache option and would like to have a more readable version of the numba implementation but that should be easy. Thanks for bearing with me.

There is an important point though that should not hold this PR. Without numba installed this is terrible slow:

```
not using numba
Good old : 1.8 s
Pre-computing: 76.9 s
```

In the light of this, my current position is the following:
* If we are going to have this we should have numba in core. Otherwise, we are adding a very slow option that will get triggered without the user knowing about this. Given that the point of adding this is to make performance faster I would not like to run the super-slow function silently. Another option is to throw an assertion letting the user know that this method is not available if they don't have numba. But at this point, I don't really see why not just adding numba at the core. It is not heavy, it is fast to import and it  it would also enable other improvements that I have in mind and would avoid all the annoyance about importing numba in a special way. 

As I was writting this I think the numba option is better fit for another issue.

DradeAW:
> Without numba installed this is terrible slow:

Then there is a bug, because it should not happen since it should be doing the same operation (for loop and getting the unit spike train). This should be fixed before merging!

> I still don't like the cache option and would like to have a more readable version of the numba implementation but that should be easy.

I'll add this on my todo list :)

h-mayorquin:
> > Without numba installed this is terrible slow:
> 
> Then there is a bug, because it should not happen since it should be doing the same operation (for loop and getting the unit spike train). This should be fixed before merging!

Don't you believe that numba is just way faster compared to looping? Can you elaborate on why this has to be a bug on the numpy version?

DradeAW:
Numba is faster, but if numba isn't installed, then it should just do it the good old way.

So if numba isn't installed, the behaviour before and after this PR shouldn't change!

src/spikeinterface/core/sorting_tools.py
```diff
@@ -0,0 +1,90 @@
...
+
+        HAVE_NUMBA = True
+    except:
+        HAVE_NUMBA = False
+
+    if HAVE_NUMBA:
+        # the trick here is to have a function getter
+        vector_to_list_of_spiketrain = get_numba_vector_to_list_of_spiketrain()
+    else:
+        vector_to_list_of_spiketrain = vector_to_list_of_spiketrain_numpy
```
h-mayorquin, line 35: Here, if you don't have numba you use the numpy version.

samuelgarcia: This is eaxctly the "old" way of course.

src/spikeinterface/core/sorting_tools.py
```diff
@@ -0,0 +1,90 @@
...
+    for segment_index, spikes in enumerate(spike_vector):
+        sample_indices = np.array(spikes["sample_index"]).astype(np.int64, copy=False)
+        unit_indices = np.array(spikes["unit_index"]).astype(np.int64, copy=False)
+        list_of_spiketrains = vector_to_list_of_spiketrain(sample_indices, unit_indices, num_units)
+        spike_trains[segment_index] = dict(zip(unit_ids, list_of_spiketrains))
+
+    return spike_trains
+
+
+def vector_to_list_of_spiketrain_numpy(sample_indices, unit_indices, num_units):
```
h-mayorquin, line 48: If the "old way" you mean using the numpy version then that is what is super-slow and I think should be avoided.

@samuelgarcia @DradeAW 

If by the "old way" you mean just calling get_unit_spike_train for every unit (which is what I meant) then I think is indeed a bug because is calling this function.

DradeAW: In my version, it was calling `get_unit_spike_train(unit_id)`.

But the function Sam wrote should be doing the same thing the `NumpySorting` does when you call `get_unit_spike_train` ... So it should be the same performance

h-mayorquin: If the sorting is a NumpySorting that would be correct, and it makes sense: NumpySorting is slow at extracting spike trains because, wel, the representation is bad for that. That's what it is. 

But if you sorting is not as bad as Numpy for calling `get_unit_spike_train` then the current implementation is terrible without Numba. 

That is, If you have a `spike_vector` cached in something that is not as bad `NumpySorting` for extracting spikes *and* you don't have Numba then you will be silently calling a terribly slow function.

DradeAW: Ah I see the problem!

You are speaking about a (corner) case where the `get_unit_spike_train` function is very fast, yet for some reason the spike trains are not cached but the spike vector is.
Indeed you are correct, in this case it would be terribly slow.

h-mayorquin: I take issue with you calling that `corner_case`. Why are you so confident than sorting to build the spike_vector and then masking within resulting structured array is a faster operation than just calling `get_unit_spike_train`? 

My original appeal was to restrict this to methods that already have the `spike_vector` pre-calculated like `NumpySorting` and.a as  you were saying, maybe Phy. There, you can be confident than calling this unlikely to be bad. But then I tested this method with numba and it works OK even for those cases with kind of fast-io so that seems OK.

But for functions that that don't have `spike_vector` pre-calculated this seems like a corner case:

1) You already called `get_unit_spike_train` to get your `spike_vector` (so your data should be cached already, not need for this)
2) Then, you already called `get_unit_spike_train` but somehow it is better to mask a very long vector (generating a spike in memory) and extract it from there.

So again, I think this works great for cases that already have `spike_vector` pre-calculated and it works OK even for other cases (if you have numba), but, if you don't have it? or if you don't have numba? Are you sure that `get_unit_spike_train` is that bad in general?

DradeAW: Ah I read again the function, and realized it does not check whether the spike trains cache is already full (this is in another PR). This should be added.

The corner case I was talking about is when `get_unit_spike_train` is very fast, yet the spike vector is cached but not the spike trains (since the spike vector is computed form the spike trains, they should be cached).

DradeAW:
@samuelgarcia I think you are thinking of another PR,

This PR is different!

src/spikeinterface/core/basesorting.py
```diff
@@ -445,6 +447,33 @@ def get_all_spike_trains(self, outputs="unit_id"):
...
+        """
+        Pre-computes and caches all spike trains for this sorting
+
+
+
+        Parameters
+        ----------
+        from_spike_vector: None | bool, default: None
+            If None, then it is automatic dependin
```
alejoe91, line 459:
```suggestion
            If None, then it is automatic depending on whether the spike vector is cached.
```

PR #2176, opened 2023-11-06
Fix crash when default start / end frame arguments on `motion interpolation` are used.

JoeZiminski:
Currently if motion correction is performed then `get_traces()` is called with default arguments (i.e. `start_frame=None` or `end_frame=None`) then the routine crashes. This form of calling `get_traces()` is used in some sorters.

This fixes by setting `start_frame=0` and `end_frame = self.get_num_samples()` if either of these values are `None`, which I saw was done elsewhere in the codebase. Let me know if any changes are needed!

### DIFF
--- a/src/spikeinterface/sortingcomponents/motion_interpolation.py
+++ b/src/spikeinterface/sortingcomponents/motion_interpolation.py
@@ -386,18 +386,18 @@ class InterpolateMotionRecordingSegment(BasePreprocessorSegment):
                 "time_vector for InterpolateMotionRecording do not work because temporal_bins start from 0"
             )
             # times = np.asarray(self.time_vector[start_frame:end_frame])
-        else:
-            if start_frame is None:
-                start_frame = 0
-            if end_frame is None:
-                end_frame = self.get_num_samples()
 
-            times = np.arange((end_frame or self.get_num_samples()) - (start_frame or 0), dtype="float64")
-            times /= self.sampling_frequency
-            t0 = start_frame / self.sampling_frequency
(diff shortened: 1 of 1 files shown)


src/spikeinterface/sortingcomponents/motion_interpolation.py
```diff
@@ -386,13 +386,18 @@ def get_traces(self, start_frame, end_frame, channel_indices):
...
-            # if self.t_start is not None:
-            #     t0 = t0 + self.t_start
-            times += t0
+
+        if start_frame is None:
+            start_frame = 0
+        if end_frame is None:
+            end_frame = self.get_num_samples()
+
+        times = np.arange((end_frame or self.get_num_samples()) - (start_frame or 0), dtype="float64")
```
alejoe91, line 395:
```suggestion
        times = np.arange(end_frame - start_frame, dtype="float64")
```
We don't need this extra protection anymore

JoeZiminski:
Hey @alejoe91 thanks missed those! I don't know if this syntax (i.e. adding `=None` defaults to all arguments) would be prefferred / is more standard across the codebase? Happy with either approach.

alejoe91:
> Hey @alejoe91 thanks missed those! I don't know if this syntax (i.e. adding `=None` defaults to all arguments) would be preferred / is more standard across the codebase? Happy with either approach.

The `segment.get_traces()` is usually not exposed and only used internally, so that's why we don't have defaults. For exposed functions I agree it's generally better to set defaults explicitly!

Thanks for the fix!

PR #2182, opened 2023-11-08
Fix compute matching v3

samuelgarcia:
Hi @h-mayorquin and @alejoe91 
It is hard to explain but I felt uncomfortable with theses fixes #2165 and #2167.
So I did a thrid proposal : I swear this is not a battle!
Can you check this ?

This PR is built on top on the one from @h-mayorquin.


alejoe91:
Can you summarize the differences with respect to #2167?

alejoe91:
Ok, I think I understand your approach: you are skipping the second loop ([this one](https://github.com/h-mayorquin/spikeinterface/blob/fix_proper_search_make_matrix_match_count/src/spikeinterface/comparison/comparisontools.py#L219-L234)) and directly counting matches in the first loop for spikes within `[spike_frame1 - delta_frame, spike_frame1 + delta_frame]`. Correct?

Honestly, this makes total sense to me, also comparing the two implementations side to side. @h-mayorquin can you check on your side?

samuelgarcia:
correct.

h-mayorquin:
Passes the same test so we are just discussing preferences over algorithm writing style. I don't think is a good use of our time to discuss this. 

Measure the speeds, chose the fastest.

h-mayorquin:
You have a speed measure script in the first PR already ready. Should take 10 minutes to measure the two branches.

alejoe91:
I don't think it's only about speed. This implementation is easy to read and it avoids the second loop, no?
Anyways, let's check speed too!

h-mayorquin:
Easy to read is precisely what I don't want to discuss because I am obviously biased to what I wrote. I wrote it like that for a reason! 

I do believe though that this might be faster so that seems like a clean criteria to decide.

samuelgarcia:
Here the speeds, I must have done something wrong:

The V1 (#2165) : 3.25 ms ± 168 µs
The V2 (#2167) : 5.36 s ± 39.2 ms
The V3 (#2182 PR) : 3.6 ms ± 199 µs

Lets go for this one. I will update a bit the doc

h-mayorquin (approved):
Thanks for measuring Sam. The fact that Alessio also thinks this is more readable is also a very strong point..

h-mayorquin:
What do you mean by doing something wrong, Sam

alejoe91 (approved):

alejoe91:
Closes #2165 
Closes #2167

samuelgarcia:
> What do you mean by doing something wrong, Sam

Maybe I am doing wrong the time measurement.

h-mayorquin:
Got you, it seems correct to me. This avoids the second loop so it should be faster? That is what I was expecting. You should have the shared the script though.

Issue #2029, opened 2023-09-20
large `total_memory` when writing binary file leads to memory error if preprocessing step converts to `float32`


JoeZiminski:
I am trying to choose chunk size to use a large percentage of available system memory*. The determination of `chunk_size` in [`job_tools/ensure_chunk_size`](https://github.com/SpikeInterface/spikeinterface/blob/30a66133bd1d5cf33b7814abe09a69312d04095e/src/spikeinterface/core/job_tools.py#L220C5-L220C5) is based on `recording.dtype`. However in some preprocessing steps (e.g. `phase_shift`, I am unsure of others) the data is temporarily converted to `float32`. This then leads to a memory error.

The only fix I can think of is to check the preprocessing steps on the recording and, if any of the steps works in `float32`, then that should be used to compute `chunk_size`. Maybe there is another workaround?

*this is because, as I understand it, larger chunk size will be better to reduce batch edge effects. Please let me know if there are other issues I have not considered.

zm711:
I'm linking a couple semi-related issues since they also have to do with a lack of guard rails currently in spikeinterface for running certain operations.

#1063. In #1063 the issue is that without sufficient storage on disk spikeinterface tries the operation rather than warns the user. 

And more recently this has also come up in #2026 about potentially needing an spikeinterface kwarg optimizer.

h-mayorquin:
Thanks for collecting those issues @zm711.

@JoeZiminski 
Thanks for brining this issue. This has made me change my mind about something in https://github.com/SpikeInterface/spikeinterface/issues/914

I think your fix looks right to me:
> The only fix I can think of is to check the preprocessing steps on the recording and, if any of the steps works in float32, then that should be used to compute chunk_size. Maybe there is another workaround?

When operating with `BasePreprocessor` the dtype that `ensure_chunk_size` uses should be the "largest" within all the parent recordings. This will give you the upper limit as that is what we care about. 

On the topic of an accurate estimation of your memory footprint. Other issues that you might have not considered is that some preprocessing steps have a larger memory footprint that you might think from their dtype and their size (`end_frame` - `start_frame`) alone. When I did the `GeneratorRecording` (now `NoiseGeneratorRecording`) I put a lot of effort avoiding any extra memory allocation (calls to system malloc). It is kind of a craft to avoid them in numpy and I am not sure we have been very careful avoid that.

[...]

Issue #1972, opened 2023-09-06
zscore fails when using whitenoise as an input.


h-mayorquin:
Related to this:

Bug:

```python
from spikeinterface.preprocessing import normalize_by_quantile, scale, center, zscore
from spikeinterface.core import generate_recording

seed = 1
rec = generate_recording(seed=seed, mode="lazy")
rec_int = scale(rec, dtype="int16", gain=100)

zscore_recording = zscore(rec_int, dtype="int16", int_scale=256, mode="mean+std", seed=seed)
traces = zscore_recording.get_traces(segment_index=0)
trace_mean = np.mean(traces, axis=0)
trace_std = np.std(traces, axis=0)
assert np.all(np.abs(trace_mean) < 1)
assert np.all(np.abs(trace_std - 256) < 1)
```

For most seeds with lazy this fails. 

@samuelgarcia you are the one that implemented this range thing for z-scores, any idea on why this might be?

alejoe91:
Interesting! I'll also take a look! Thanks for tracking this down @h-mayorquin

h-mayorquin:
I thought more about this this morning. It is just a statistical problem with the test.  This is not a bug. 
The thing is that if you scale your data by a standadrd deviation you increase the standard error:

$`  SE = \frac{\sigma}{\sqrt{n}}  `$ 

Then you need a larger n. If you calculate the standard error of the mean for the chunks available you get that the standard error is not small enough for the test above to be sound.

Plus, casting to int is not a linear function (it behaves like floor for positives and like ceiling for negatives) which also increases the standard deviation of the samples and will make the standard error larger in practice.

Here, I estimate taking the mean 1_000 times with 100_000 samples. You can see that most of the estimates are farther from the true than 0 and therefore will fail on the test above.


```
estimates = (256 * np.random.randn(int(5e4), 1000).astype("int16")).mean(axis=0)
plt.hist(estimates, bins=100)
plt.xlabel("Estimated mean from the process")
plt.ylabel("Frequency")
```

![image](https://github.com/SpikeInterface/spikeinterface/assets/6429509/a0a17b7d-4467-4176-973f-03a45176379c)

h-mayorquin:
[...]

Issue #914, opened 2022-08-26
Specification of the dtype for Preprocessors


yger:
I saw a discussion concerning the dtype of the whitening pre-processors, but I just want to highlight that there is some inconsistencies between the pre-processors: some can specify an output dtype (filtering, whitening), and some can not (CAR, artefacts, ...)

Maybe there should be something more abstract and general such that all pre-processors can control the dtype output?

h-mayorquin:
Yes, so I can think on two solutions. 

Either throw a warning if the parent recording and the known necessary output does not match. Or be more demanding and make users transform the parent recording with `AstypeRecording` if we believe this is a serious error that they should not do (similar to what we do with  the `is_filtered` flag).

In any case, then, it seems that dtype should not be in any `__init__`.

Let's see if @samuelgarcia wants to chime in.

h-mayorquin:
> For some preprocessors the output will necessarily change. E.g., zscore and whitening will output a float no matter what!

This and https://github.com/SpikeInterface/spikeinterface/issues/2029 have made me change my mind. It does not makes sense that pre-processing takes the input of their parent for some cases. Now I feel that some processors like dtype and phase shift should have their own dtype and we should thrown a warning if there is an increase in memory representation.

I still believe that control of dtype should be done iwth AstypeRecording instead in the `__init__`.

PR #2189, opened 2023-11-10
Fix filtering rounding error

DradeAW:
The `astype(dtype)` always converts int to the lower integer, which in some cases causes real problems.

This PR fixes that.

### DIFF
--- a/doc/modules/preprocessing.rst
+++ b/doc/modules/preprocessing.rst
@@ -74,6 +74,8 @@ dtype (unless specified otherwise):
 
 Some scaling pre-processors, such as :code:`whiten()` or :code:`zscore()`, will force the output to :code:`float32`.
 
+When converting from a :code:`float` to an :code:`int`, the value will first be rounded to the nearest integer.
+
 
 Available preprocessing
 -----------------------
--- a/src/spikeinterface/preprocessing/filter.py
+++ b/src/spikeinterface/preprocessing/filter.py
@@ -153,6 +153,10 @@ class FilterRecordingSegment(BasePreprocessorSegment):
             filtered_traces = filtered_traces[left_margin:-right_margin, :]
         else:
             filtered_traces = filtered_traces[left_margin:, :]
+
+        if np.issubdtype(self.dtype, np.integer):
+            filtered_traces = filtered_traces.round()
+
         return filtered_traces.astype(self.dtype)
 
 
--- a/src/spikeinterface/preprocessing/filter_gaussian.py
+++ b/src/spikeinterface/preprocessing/filter_gaussian.py
@@ -74,6 +74,9 @@ class GaussianFilterRecordingSegment(BasePreprocessorSegment):
         filtered_fft = traces_fft * (gauss_high - gauss_low)[:, None]
         filtered_traces = np.real(np.fft.ifft(filtered_fft, axis=0))
 
+        if np.issubdtype(dtype, np.integer):
+            filtered_traces = filtered_traces.round()
+
         if right_margin > 0:
             return filtered_traces[left_margin:-right_margin, :].astype(dtype)
         else:
--- a/src/spikeinterface/preprocessing/normalize_scale.py
+++ b/src/spikeinterface/preprocessing/normalize_scale.py
@@ -20,6 +20,10 @@ class ScaleRecordingSegment(BasePreprocessorSegment):
     def get_traces(self, start_frame, end_frame, channel_indices):
         traces = self.parent_recording_segment.get_traces(start_frame, end_frame, channel_indices)
         scaled_traces = traces * self.gain[:, channel_indices] + self.offset[:, channel_indices]
+
+        if np.issubdtype(self._dtype, np.integer):
+            scaled_traces = scaled_traces.round()
+
         return scaled_traces.astype(self._dtype)
 
 
--- a/src/spikeinterface/preprocessing/phase_shift.py
+++ b/src/spikeinterface/preprocessing/phase_shift.py
@@ -103,6 +103,8 @@ class PhaseShiftRecordingSegment(BasePreprocessorSegment):
 
         traces_shift = traces_shift[left_margin:-right_margin, :]
         if self.tmp_dtype is not None:
+            if np.issubdtype(self.dtype, np.integer):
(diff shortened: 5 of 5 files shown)


DradeAW:
An example of how I found this error:

I was plotting the distribution of spikes amplitude, and wanted to compare it to the distribution of voltage on the same channel.

The bin 0 was twice bigger than what it should be, because of how `astype(np.int16)` works.
The bin x includes values between [x ; x+1). **Except** for the bin 0 which includes values between (-1; 1).

Rounding before converting to an int fixes this issue, and more globally makes things neater.

zm711:
I take the point that you'd prefer this to behave like a normal round (I think that it is better too for my mental schema), but the standard for casting from float -> int (for example) is just to drop the decimal values. So you don't think that this might lead some users to be confused by the non-standard behavior of casting to an int, but seeing the value round? Just curious how you're thinking about this "round" occurring internal to the function without the user potentially knowing about it?

DradeAW:
Well if I have a value at 12, and I apply a slight filter which reduces the value to 11.98, I expect the value to be 12 after casting to an integer ...

I don't think many users expect to receive 11?

DradeAW:
The thing is that most recording are with int16 today. And by default in SI when you filter, you get back an int16.
And as I stated above, there is a problem with 0.

zm711:
Actually I completely agree for Gaussian. In that case you use the dtype from the traces and so the user has no option. If the recording is an int16 and it gets filter the user is stuck with int16.

But in `filter` you are required to specify the dtype, so the user could choose to return the data with a dtype that maintains the filtered values. So my question is more just for filter where you could warn that user to return in float if they wanted to keep the decimal.

DradeAW:
> But in `filter` you are required to specify the dtype

https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/preprocessing/filter.py#L70

I think a lot of users don't specify the `dtype` ^^
But let's see what @alejoe91 @samuelgarcia and @h-mayorquin think!

samuelgarcia:
Hi.
thank you for this fix. It was a need since long time. I also in some cases enormous bin at zeros in distrib.
I think this is not only filtering but all preprocessing that preserve the input dtype.
Maybe we could make a larger check on this.

About a warning when do not use the decimal will be really painfull. For me this preserving the input dtype is a feature in spikeinterface.preprocessing since long time.

For me it has to be documented well.
We already have this in the doc : https://spikeinterface.readthedocs.io/en/latest/modules/preprocessing.html#impact-on-recording-dtype
For sure, Zach, will find a more adavnced formula for this

DradeAW:
> I think this is not only filtering but all preprocessing that preserve the input dtype.
> Maybe we could make a larger check on this.

I can try to do that if you think that's best!

zm711:
Okay fair enough. If it is occurring everywhere in the same manner and not ad hoc, then it makes sense.

>For me it has to be documented well.
We already have this in the doc : https://spikeinterface.readthedocs.io/en/latest/modules/preprocessing.html#impact-on-recording-dtype
For sure, Zach, will find a more adavnced formula for this

I can double check that part of the docs (I already found a typo reading it just now so I'll do another pass soon). And yeah I think just making it clear in the docs that this is what is happening is super important for users. I will say that I did add the `np.rint` in Neo for selecting frames for analog signals, so I'm definitely not opposed to this :) just curious about what is given as the responsibility/freedom of the user.

> About a warning when do not use the decimal will be really painfull. For me this preserving the input dtype is a feature in spikeinterface.preprocessing since long time.

Fair enough! If it is expected behavior for the library then people can't expect to be warned for the norm.

h-mayorquin:
> The astype(dtype) always converts int to the lower integer, which in some cases causes real problems.

This is not true:
```python
# Random floats between -1 and 1
x = np.random.uniform(-2, 2, 10)
print(x)
x.astype("int")
output:
[-0.01363288  0.42181579  1.25074111  0.19889013  1.11397242 -0.95708403
  1.15950627 -1.3823785   1.43423315 -0.61537481]

```

I actually don't know enough about filtering to have an opinion on this. My own -weakly held- take now is that there should NOT be an option to force the output in a specific format:

https://github.com/SpikeInterface/spikeinterface/issues/914#issuecomment-1734761048

I think that the tricks that we use don't achieve saving memory:
https://github.com/SpikeInterface/spikeinterface/issues/2029
They are also confusing as as some point we built a test that was statistically wrong:
https://github.com/SpikeInterface/spikeinterface/issues/1972

And, the most subjective, tricks like the int scale in the Z-score hurt my mathematical sensibilities:
https://github.com/h-mayorquin/spikeinterface/blob/1ca0655e371a8dbc34c49ffde251b3628c03db8b/src/spikeinterface/preprocessing/normalize_scale.py#L239-L242

Looking forward to learn here, 
What is able to control the output meaning to achieve?

DradeAW:
> This is not true:

```
>>> import numpy as np
>>> 
>>> np.random.uniform(-1, 1, 20).astype(np.int32)
array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
      dtype=int32)
```

By lower I meant "closer to 0"

h-mayorquin:
The correct claim is that casting behaves like floor for positive values and like a ceiling for negative values. There is not a conception of "lower" that I can map to that.

doc/modules/preprocessing.rst
```diff
@@ -74,6 +74,8 @@ dtype (unless specified otherwise):
 
 Some scaling pre-processors, such as :code:`whiten()` or :code:`zscore()`, will force the output to :code:`float32`.
```
zm711, line 75: This is part of your problem @h-mayorquin? The forcing of the output? Just so I can stay on the same page.

h-mayorquin: Yes, I don't know why we have that option but my proble most likely is ignorance. Maybe there are good scenarios that I jus't don't understand.

zm711: I honestly don't know either, so I'm happy to learn too :)

samuelgarcia (approved):

samuelgarcia:
Merci beaucoup. This was important.

PR #2191, opened 2023-11-10
Fix corner case in make_match_count_matrix()

samuelgarcia:
* Fix corner case in make_match_count_matrix()
* Add symetric option and propagate in SymmetricSortingComparison/GroundTruthComparison
* protect with index instead of frame

### DIFF
--- a/src/spikeinterface/comparison/comparisontools.py
+++ b/src/spikeinterface/comparison/comparisontools.py
@@ -124,12 +124,12 @@ def get_optimized_compute_matching_matrix():
 
     @numba.jit(nopython=True, nogil=True)
     def compute_matching_matrix(
-        frames_spike_train1,
-        frames_spike_train2,
+        spike_frames_train1,
+        spike_frames_train2,
         unit_indices1,
         unit_indices2,
-        num_units_sorting1,
-        num_units_sorting2,
+        num_units_train1,
+        num_units_train2,
         delta_frames,
     ):
         """
@@ -137,30 +137,33 @@ def get_optimized_compute_matching_matrix():
 
         Given two spike trains, this function finds matching spikes based on a temporal proximity criterion
         defined by `delta_frames`. The resulting matrix indicates the number of matches between units
-        in `frames_spike_train1` and `frames_spike_train2`.
+        in `spike_frames_train1` and `spike_frames_train2`.
 
         Parameters
         ----------
-        frames_spike_train1 : ndarray
-            Array of frames for the first spike train. Should be ordered in ascending order.
-        frames_spike_train2 : ndarray
-            Array of frames for the second spike train. Should be ordered in ascending order.
+        spike_frames_train1 : ndarray
+            An array of integer frame numbers corresponding to spike times for the first train. Must be in ascending order.
+        spike_frames_train2 : ndarray
+            An array of integer frame numbers corresponding to spike times for the second train. Must be in ascending order.
         unit_indices1 : ndarray
-            Array indicating the unit indices corresponding to each spike in `frames_spike_train1`.
+            An array of integers where `unit_indices1[i]` gives the unit index associated with the spike at `spike_frames_train1[i]`.
         unit_indices2 : ndarray
-            Array indicating the unit indices corresponding to each spike in `frames_spike_train2`.
-        num_units_sorting1 : int
-            Total number of units in the first spike train.
-        num_units_sorting2 : int
-            Total number of units in the second spike train.
+            An array of integers where `unit_indices2[i]` gives the unit index associated with the spike at `spike_frames_train2[i]`.
+        num_units_train1 : int
+            The total count of unique units in the first spike train.
+        num_units_train2 : int
+            The total count of unique units in the second spike train.
         delta_frames : int
-            Maximum difference in frames between two spikes to consider them as a match.
+            The inclusive upper limit on the frame difference for which two spikes are considered matching. That is
+            if `abs(spike_frames_train1[i] - spike_frames_train2[j]) <= delta_frames` then the spikes at `spike_frames_train1[i]`
+            and `spike_frames_train2[j]` are considered matching.
 
         Returns
         -------
         matching_matrix : ndarray
-            A matrix of shape (num_units_sorting1, num_units_sorting2) where each entry [i, j] represents
-            the number of matching spikes between unit i of `frames_spike_train1` and unit j of `frames_spike_train2`.
+            A 2D numpy array of shape `(num_units_train1, num_units_train2)`. Each element `[i, j]` represents
+            the count of matching spike pairs between unit `i` from `spike_frames_train1` and unit `j` from `spike_frames_train2`.
+
 
         Notes
         -----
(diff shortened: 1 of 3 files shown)


src/spikeinterface/comparison/comparisontools.py
```diff
@@ -232,7 +176,61 @@ def compute_matching_matrix(
...
+    1. Iterate through each spike in the first train
+    2. For each spike, find the first match in the second train.
+    3. Save the index of the first match as the new `second_train_search_start `
+    3. For each match, find as many matches as possible from the first match onwards.
+
+    An important condition here is that the same spike is not matched twice. This is managed by keeping track
+    of the last matched frame for each unit pair in `last_match_frame1` and `last_match_frame2`
+    There are corner cases where a spike can be counted twice in the the spiketrain 2 in case of bursting situations
+    (below delta_frames) in the spiketrain 1. To ensure that the number of match do not exceed the number of spike,
+    we applied a final clip.
```
zm711, line 228:
```suggestion
    An important condition here is that the same spike is not matched twice. This is managed by keeping track
    of the last matched frame for each unit pair in `last_match_frame1` and `last_match_frame2`
    There are corner cases where a spike can be counted twice in the spiketrain 2 if there are bouts of bursting activity
    (below delta_frames) in the spiketrain 1. To ensure that the number of matches does not exceed the number of spikes,
    we apply a final clip.
```

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -232,7 +176,61 @@ def compute_matching_matrix(
     return compute_matching_matrix
 
 
-def make_match_count_matrix(sorting1, sorting2, delta_frames):
+def make_match_count_matrix(sorting1, sorting2, delta_frames, symmetric=False):
+    """
+    Compute a matrix representing the matches between two Sorting objects.
```
zm711, line 181:
```suggestion
    Computes a matrix representing the matches between two Sorting objects.
```

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -232,7 +176,61 @@ def compute_matching_matrix(
...
+
+    Given two spike trains, this function finds matching spikes based on a temporal proximity criterion
+    defined by `delta_frames`. The resulting matrix indicates the number of matches between units
+    in `spike_frames_train1` and `spike_frames_train2` for each pair of units.
+
+    Note that this algo is not symmetric and biased toward sorting1 is the ground truth.
```
zm711, line 187:
```suggestion
    Note that this algo is not symmetric and is biased with `sorting1` representing ground truth for the comparison
```

Just trying to clean this a bit. But maybe I made it too strong?

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -232,7 +176,61 @@ def compute_matching_matrix(
...
+    which signifies the minimal index in the second spike train that might match the upcoming spike
+    in the first train.
+
+    The logic can be summarized as follows:
+    1. Iterate through each spike in the first train
+    2. For each spike, find the first match in the second train.
+    3. Save the index of the first match as the new `second_train_search_start `
+    3. For each match, find as many matches as possible from the first match onwards.
+
+    An important condition here is that the same spike is not matched twice. This is managed by keeping track
```
zm711, line 224:
```suggestion
    An important condition is that the same spike is not matched twice. This is managed by keeping track
```

zm711 (commented):
Just a couple docstring suggestions

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -132,62 +132,6 @@ def compute_matching_matrix(
         num_units_train2,
         delta_frames,
     ):
-        """
```
h-mayorquin, line 135: Why did you remove the docstring from here?

Adding the other is great but what is the gain of having less?

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -232,7 +176,61 @@ def compute_matching_matrix(
...
+    ----------
+    sorting1 : Sorting
+        An array of integer frame numbers corresponding to spike times for the first train. Must be in ascending order.
+    sorting2 : Sorting
+        An array of integer frame numbers corresponding to spike times for the second train. Must be in ascending order.
+    delta_frames : int
+        The inclusive upper limit on the frame difference for which two spikes are considered matching. That is
+        if `abs(spike_frames_train1[i] - spike_frames_train2[j]) <= delta_frames` then the spikes at `spike_frames_train1[i]`
+        and `spike_frames_train2[j]` are considered matching.
+    symmetric: bool, dfault False
```
h-mayorquin, line 199: I wish we had a better name for this one.

src/spikeinterface/comparison/comparisontools.py
h-mayorquin, line 135: Why do we need to remove the docstring here? It is great that you added the other (my bad!) it here but is not like we are paying real state for docstrings here.

src/spikeinterface/comparison/comparisontools.py
h-mayorquin, line 199: I wish we could come up with a better name of this but I don't have anything better.

h-mayorquin: Like, it makes me expected the returning matrix to be symmetric but that can't be.

h-mayorquin (approved):
Great that you solved this, @samuelgarcia .?

You can also uncomment the stuff here if you want:

https://github.com/samuelgarcia/spikeinterface/blob/dc55dfc97a66ae8681a501951bbdaa1be8342be9/src/spikeinterface/comparison/tests/test_comparisontools.py#L122-L135

It is passing locally for me.

h-mayorquin (changes requested):
Yeah. Actually at least leave the arguments of the docstring if you are very adamant about real state. 

I put a lot effort writing that docstring

zm711:
> Yeah. Actually at least leave the arguments of the docstring if you are very adamant about real state.

Agreed. It's super helpful to at least have arguments and returns in a docstring. I understand not wanting duplication or wanting to reduce lines, but knowing the ins and outs will help with development, but also with users sending in bug reports. Recently there was an issue with a function not accepting tuples when it's partner function only output tuples so knowing there was that disconnect would have found the issue earlier.

src/spikeinterface/comparison/comparisontools.py
samuelgarcia: Sorry, for me it was better for the higher function. This is why I moved it.
Having docstring here for function nested in another one will be never generated in the doc nor directly accessible. We can put it back maybe maybe not.

src/spikeinterface/comparison/comparisontools.py
samuelgarcia: `force_symmetry` ? `ensure_symmetry` ? sounds better ?

samuelgarcia:
@h-mayorquin : sorry for the removal of the docstring. For me it was an internal function. Maybe we can put doc everywhere.

src/spikeinterface/comparison/comparisontools.py
h-mayorquin: Both are good choices to me actually. The two very good.

src/spikeinterface/comparison/comparisontools.py
h-mayorquin: That's a good point that I did not consider, but I have two points:
* It is still useful for the developer reading the code.
* I am taking your point that we should modify the way we are doing the numba functions here seriously. When we do that, I am expecting that we would be able to expose the docstrings again. And for that time, I would not like to have to urge the commit history to fetch it back. 

Given that is costing nothing, would you mind leaving it just as it is? We can discuss this after the release and I can myself go and create another PR to remove if you want it. But I don't want to get this discussion get in the way of the release and I don't want to lose the docstring to the commit history either.

samuelgarcia: done

samuelgarcia: OK done.

src/spikeinterface/comparison/comparisontools.py
samuelgarcia: done

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -133,11 +129,9 @@ def compute_matching_matrix(
...
     ):
         """
-        Compute a matrix representing the matches between two spike trains.
-
-        Given two spike trains, this function finds matching spikes based on a temporal proximity criterion
-        defined by `delta_frames`. The resulting matrix indicates the number of matches between units
-        in `spike_frames_train1` and `spike_frames_train2`.
+        Internal function used by `make_match_count_matrix()`.
+        This function is for one segment only.
+        The llop over segment is done in `make_match_count_matrix()`
```
alejoe91, line 134:
```suggestion
        The loop over segment is done in `make_match_count_matrix()`
```

h-mayorquin:
```suggestion
        The loop over segment is done in `make_match_count_matrix()`
```

alejoe91 (approved):

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -161,31 +155,8 @@ def compute_matching_matrix(
...
-        Notes
-        -----
-        This algorithm identifies matching spikes between two ordered spike trains.
-        By iterating through each spike in the first train, it compares them against spikes in the second train,
-        determining matches based on the two spikes frames being within `delta_frames` of each other.
-
-        To avoid redundant comparisons the algorithm maintains a reference, `second_train_search_start `,
-        which signifies the minimal index in the second spike train that might match the upcoming spike
-        in the first train.
+            A 2D numpy array of shape `(num_units_train1, num_units_train2)`
```
h-mayorquin, line 158:
```suggestion
            A 2D numpy array of shape `(num_units_train1, num_units_train2)`. Each element `[i, j]` represents
            the count of matching spike pairs between unit `i` from `spike_frames_train1` and unit `j` from `spike_frames_train2`.
```

h-mayorquin (approved):
This is fine by me.

My final comment (feel free to ignore it) is that maybe the `ensure_symmetry` functionality should be in the `make_agreement_scores` as this function can't never really be symmetric and it makes more sense over there which is where you need the scores to be symmetric.

PR #2196, opened 2023-11-13
Use `uint64` for match count

DradeAW:
`uint16` leads to overflow and erroneous count.


h-mayorquin (approved):
This was my mistake for lack of domain knowledge. Thanks @DradeAW . Good intuition about numerical data type sizes.

PR #2207, opened 2023-11-15
Add `rename_units` method in sorting

h-mayorquin:
Added a new `rename_units` method. 

I always forget that the way to do this is to use the `select_units` method with a second argument. 

I think adding a specific method for this is better for discoverability and also is closer to the ideal that "methods should do one thing'" which is also easier for mantainability and code reading.

I also added some missing typing and corrected some docstrings to reflect the fact that some methods don't modify a place but return a new object.


samuelgarcia:
Parfait!

alejoe91 (approved):

Issue #2206, opened 2023-11-15
SpikingCircus2 throwing error with multi-segment recordings


FrancescoNegri:
Hi!
I was trying to run SpykingCircus2 with a multi-segment recording (3 segments for a total of ~18 mins) since, according to my understanding of the [documentation](https://spikeinterface.readthedocs.io/en/latest/modules/sorters.html#handling-multi-segment-recordings), only SpykingCircus2 and Tridesclous support multi-segment data.

However, the `run_sorter(...)` function throws this error:
```
{
	"name": "ValueError",
	"message": "This sorter spykingcircus2 does not handle multi-segment recordings, use si.concatenate_recordings(...)",
	"stack":  ...
}
```

Is such a behaviour expected?
Thank you in advance :)

yger:
SC2 should work with multi segment, I'll investigate since I don't know where and why this message is raised...

samuelgarcia:
Oups. This is a bug. I guess a bad assert somewhere. Pierre you fix this ?

alejoe91:
Problem is here: https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/sorters/basesorter.py#L50

You need to set it to True in the SC2 class

alejoe91:
Note that also TDC2 is missing it!

yger:
This has been fixed in #2208

PR #2208, opened 2023-11-15
Handling segments in SC2 and TDC2

yger:
We should maybe test that more in depth, but at least, the default is now correct


alejoe91 (approved):

alejoe91:
Fixes #2206

PR #2209, opened 2023-11-15
Add an option for count_num_spikes_per_unit

samuelgarcia:
This should be merge after #2198

### DIFF
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -267,16 +267,16 @@ class BaseSorting(BaseExtractor):
             DeprecationWarning,
             stacklevel=2,
         )
-        return self.count_num_spikes_per_unit()
+        return self.count_num_spikes_per_unit(outputs="dict")
 
-    def count_num_spikes_per_unit(self, output="dict"):
+    def count_num_spikes_per_unit(self, outputs="dict"):
         """
         For each unit : get number of spikes  across segments.
 
         Parameters
         ----------
-        output: "dict" | "array", dfault: "dict"
-            Return a dict (key is unit_id) or an numpy array.
+        outputs: "dict" | "array", dfault: "dict"
+            Control the type of the returned object: a dict (keys are unit_ids) or an numpy array.
 
         Returns
         -------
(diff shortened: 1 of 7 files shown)


samuelgarcia:
With this, some part of the code using `count_num_spikes_per_unit()` can be rewritten with array in mind and should be better than looping over a dict. (for instance using np.any, np.all, ...)

should I do this in the same PR ?

h-mayorquin:
@samuelgarcia 
> should I do this in the same PR ?

I vote not, one PR per feature as much as we can.

src/spikeinterface/core/basesorting.py
```diff
@@ -269,35 +269,40 @@ def get_total_num_spikes(self):
         )
         return self.count_num_spikes_per_unit()
 
-    def count_num_spikes_per_unit(self) -> dict:
+    def count_num_spikes_per_unit(self, output="dict"):
```
h-mayorquin, line 272: I think it would be better to just make another function for this behavior.

What is the advantage of doing it with a keyword?
The disavatnages are:
* Harder to discover
* More complicated output type which makes harder to reason about code.
* Adds complexity to the function.

Functions should do one thing and do it well. Specially at the core. Or make two private methods that do the specific task and make this a dispatcher with options.

samuelgarcia:
@DradeAW : can you chekc this ?

src/spikeinterface/core/basesorting.py
```diff
@@ -267,37 +267,60 @@ def get_total_num_spikes(self):
...
+                    all_spiketrain_are_cached = False
+                    break
+        else:
+            all_spiketrain_are_cached = False
 
-        if self._cached_spike_trains is not None:
-            for unit_id in self.unit_ids:
-                n = 0
+        if all_spiketrain_are_cached or self._cached_spike_vector is None:
+            # case one 1 or 3
```
DradeAW, line 305: case 1 or 3 (you have an extra "one")

DradeAW:
I think the way you implemented it is the right way: using spike trains if they are all cached or if the spike_vector isn't cached.

Also love the idea of getting an array or a dict!
I never know the return type and often need to convert it depending on my needs, I like the option :)

I'll make some test to make sure it makes things faster on my code.

src/spikeinterface/core/basesorting.py
zm711: Maybe he meant one case: either 1 or 3, but not both? ;)

zm711 (commented):
Although I do think that separating out functions is likely better, I added docstring/error clarifications with the assumption that this PR will remain as is. Feel free to ignore if you plan to change the structure.

src/spikeinterface/core/basesorting.py
samuelgarcia: Done

src/spikeinterface/core/basesorting.py
samuelgarcia: For me the main advantage is internal use of `sorting.count_num_spikes_per_unit(output="array")`
replacing at several places the uggly  `np.array(list(sorting.count_num_spikes_per_unit().values()))`

h-mayorquin: But you can just have another method that will allow you to do exactly the same:

`sorting.count_num_spikes_per_unit_array' instead of `np.array(list(sorting.count_num_spikes_per_unit().values()))`

I think the method is a good idea.
I think mixing this new logic in the old method is not for the reasons stated above.

 I am asking: why do you want to mix the logic to produce counts as arrays and counts as dict in the same fuction? What's the advantage of it? Is than an aesthethical thing? You dislike having more functions? You personally find keywords easier to discover than functions? I am looking for your reasoning here.

"Functions should do only one thing and do it well" seems like a really good principle to me that is more or less widely accepted. Design rules can -and sometimes should- be broken, but I think we should have a good reason to do it. Which one is here?

alejoe91 (approved):

PR #2210, opened 2023-11-15
Avoid duplicated template and quality metric names

alejoe91:
In principle, the user can set duplicated metrics name. This simple PR makes a unique of the metric names to avoid duplication in the output dataframes

### DIFF
--- a/src/spikeinterface/postprocessing/template_metrics.py
+++ b/src/spikeinterface/postprocessing/template_metrics.py
@@ -60,7 +60,7 @@ class TemplateMetricsCalculator(BaseWaveformExtractorExtension):
             metric_names += get_multi_channel_template_metric_names()
         metrics_kwargs = metrics_kwargs or dict()
         params = dict(
-            metric_names=[str(name) for name in metric_names],
+            metric_names=[str(name) for name in np.unique(metric_names)],
             sparsity=sparsity,
             peak_sign=peak_sign,
             upsampling_factor=int(upsampling_factor),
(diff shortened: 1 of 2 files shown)


src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -60,7 +60,7 @@ def _set_params(
             metric_names += get_multi_channel_template_metric_names()
         metrics_kwargs = metrics_kwargs or dict()
         params = dict(
-            metric_names=[str(name) for name in metric_names],
+            metric_names=[str(name) for name in np.unique(metric_names)],
```
h-mayorquin, line 63: If this is a list maybe you can use set?

h-mayorquin (approved):
Otherwise, looks good to me. Simple change. Was it causing any trouble?

src/spikeinterface/postprocessing/template_metrics.py
h-mayorquin: e.g:  `{str(name) for name in metric_names}`

alejoe91: well it needs to be a list!

alejoe91:
yes, especially if you use `include_multi_channel_metrics=True` and you define some of these metrics in the `metric_names`

src/spikeinterface/postprocessing/template_metrics.py
h-mayorquin: Yeah, it could also be:
[[str(name) for name in set(metric_names)]

But it probably does not matter for this size anyway : )

PR #2218, opened 2023-11-16
Remove joblib in favor of ParallelProcessExecutor

alejoe91:


samuelgarcia:
This is a good news!

src/spikeinterface/postprocessing/principal_component.py
```diff
@@ -411,10 +414,14 @@ def _fit_by_channel_local(self, n_jobs, progress_bar):
...
-                    for wf_ind, chan_ind in enumerate(channel_inds)
-                )
+                # parallel
+                items = [(pca_model_files[chan_ind], wfs[:, :, wf_ind]) for wf_ind, chan_ind in enumerate(channel_inds)]
+                n_jobs = min(n_jobs, len(items))
+
+                with ProcessPoolExecutor(max_workers=n_jobs) as executor:
+                    results = executor.map(partial_fit_one_channel, items)
+                    for res in results:
+                        pass
```
samuelgarcia, line 424: Maybe parralelize over units would be more efficient this would avoid starting and stoping pool many times.

zm711:
I actually had a question about multiprocessing in general. For `n_jobs=-1` you do `os.cpu_count()`, but sometimes I read it is best to leave at least one core behind for the computer itself. I can understand that `os.cpu_count()-1` could lead to 0 if someone is on a single core, but is there really a potential problem with `n_jobs=-1` or do you think that's fine for users to default to using all cores? (I really don't know multiprocessing in the python context :( sorry). But I love the branch name!

samuelgarcia:
In spikeinetrface, you can do `n_jobs=-2`to leave one core free.
You can also do `n_jobs=0.8` you use 80% of cores.

h-mayorquin:
> In spikeinetrface, you can do `n_jobs=-2`to leave one core free. You can also do `n_jobs=0.8` you use 80% of cores.

@CodyCBakerPhD You have some evolutionary convergence here in yours and Dandi's preference:
https://github.com/catalystneuro/neuroconv/blob/1a7ae7a62e502614fa6b52e511310736aed55552/src/neuroconv/tools/nwb_helpers/_models/_zarr_models.py#L156-L163

Sam went further and even made the datatype a float though.

zm711:
Thanks Sam, I guess my question is more out of scope for this PR, but I meant more in the sense that some users have had issues with `n_jobs=-1` and so I wonder if there's a more sensible default (like `n_jobs=-2` to save one core). SI already does an excellent job of letting users tune their multiprocessing, but for the new users who just say "these guys are the experts I'm not going to change any of their defaults", I wonder if -1 is the best.

alejoe91:
> Thanks Sam, I guess my question is more out of scope for this PR, but I meant more in the sense that some users have had issues with `n_jobs=-1` and so I wonder if there's a more sensible default (like `n_jobs=-2` to save one core). SI already does an excellent job of letting users tune their multiprocessing, but for the new users who just say "these guys are the experts I'm not going to change any of their defaults", I wonder if -1 is the best.

Good point. Maybe default should be something like 0.9

src/spikeinterface/postprocessing/principal_component.py
alejoe91: So, after some thougth, I think it's not possible to parallelize over units. In the `by_channel_local` mode, there will be 1 PCA model per channel. If we parallelize over units and have units that share some channels, their process will attempt to partial fit the same model, so we would need some synchronizer of some sort, which sounds very tricky. I think this is why parallelization was implemented at the channel level in the first place. @samuelgarcia makes sense?

h-mayorquin:
@zm711 @alejoe91 
Note that this calculates number of logical cpus, though, what you should have is one physical cpu available in cases as this which are mostly -we expect- computational expensive. Usually you have two logical per physical, we should discuss this in a different issue though.

zm711:
Fair enough. Do we want to just open an overall job_kwargs issue/discussion to brainstorm on this? (the logical processors are an intel thing right (so not amd/Mac), so this might get super complicated, ie I agree a separate thread would be way better). I can open the issue later today!

src/spikeinterface/postprocessing/principal_component.py
```diff
@@ -385,12 +385,15 @@ def _fit_by_channel_local(self, n_jobs, progress_bar):
 
         tmp_folder = p["tmp_folder"]
         if tmp_folder is None:
-            tmp_folder = "tmp"
-        tmp_folder = Path(tmp_folder)
+            if n_jobs > 1:
+                import tempfile
+
+                tmp_folder = tempfile.mkdtemp(prefix="tmp", dir=".")
```
alejoe91, line 391: use system "/tmp" or `spikeinterface_global_tmp_folder()`

src/spikeinterface/postprocessing/principal_component.py
```diff
@@ -424,6 +432,7 @@ def _fit_by_channel_local(self, n_jobs, progress_bar):
                 with open(pca_model_file, "rb") as fid:
                     pca_models.append(pickle.load(fid))
                 pca_model_file.unlink()
+            print(f"Removing {tmp_folder}")
```
h-mayorquin, line 435: Shouldn't we remove this print?

src/spikeinterface/postprocessing/principal_component.py
```diff
@@ -385,12 +386,15 @@ def _fit_by_channel_local(self, n_jobs, progress_bar):
 
         tmp_folder = p["tmp_folder"]
         if tmp_folder is None:
-            tmp_folder = "tmp"
-        tmp_folder = Path(tmp_folder)
+            if n_jobs > 1:
+                import tempfile
+
+                tmp_folder = tempfile.mkdtemp(prefix="tmp", dir=get_global_tmp_folder())
```
h-mayorquin, line 392: Why do we need the prefix tmp? If we are adding a prefix should it be something more informative like principial_component?

src/spikeinterface/postprocessing/principal_component.py
```diff
@@ -385,12 +386,15 @@ def _fit_by_channel_local(self, n_jobs, progress_bar):
 
         tmp_folder = p["tmp_folder"]
         if tmp_folder is None:
-            tmp_folder = "tmp"
-        tmp_folder = Path(tmp_folder)
+            if n_jobs > 1:
+                import tempfile
```
h-mayorquin, line 390: I guess this could go on top as it is part of the standard library?

src/spikeinterface/postprocessing/principal_component.py
alejoe91: Oups! Yes! Let me fix both!

src/spikeinterface/postprocessing/principal_component.py
alejoe91: done

src/spikeinterface/postprocessing/principal_component.py
alejoe91: done, using `pca`

src/spikeinterface/postprocessing/principal_component.py
alejoe91: done

samuelgarcia (approved):

src/spikeinterface/postprocessing/principal_component.py
h-mayorquin: Thanks.

h-mayorquin:
Fine by me. Merging!

Issue #1891, opened 2023-07-27
Error run_sorters() : ValueError: path is on mount 'C:', start on mount


Antho2422:
Hi all, 

I'm currently trying to run spykingcircus2 via the function run_sorters().

I'm running the script on a Windows PC and save results on another drive. I get this error message : 
```python
Failed: [undefined]ValueError: path is on mount 'C:', start on mount '\\\\l2export\\iss02.charpier'
``` 

Here is the full error message : 
```python
..\..\..\AppData\Local\pypoetry\Cache\virtualenvs\priam-5spsvreu-py3.9\lib\site-packages\spikeinterface\core\base.py:942: in <lambda>
    func = lambda p: os.path.relpath(str(p), start=relative)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

path = 'C:\\Users\\anthony.pinto\\Documents\\git\\priam\\data\\raw\\test_model_v0.1\\03046_2021-07-09_12-02\\base_recording'
start = '\\\\l2export\\iss02.charpier\\priam\\test_suite\\outputs\\session_test\\pat_03046_1482\\eeg\\test_sorting\\sorting\\0\\spykingcircus2'

    def relpath(path, start=None):
        """Return a relative version of a path"""
        path = os.fspath(path)
        if isinstance(path, bytes):
            sep = b'\\'
            curdir = b'.'
            pardir = b'..'
        else:
            sep = '\\'
            curdir = '.'
            pardir = '..'
    
        if start is None:
            start = curdir
    
        if not path:
            raise ValueError("no path specified")
    
        start = os.fspath(start)
        try:
            start_abs = abspath(normpath(start))
            path_abs = abspath(normpath(path))
            start_drive, start_rest = splitdrive(start_abs)
            path_drive, path_rest = splitdrive(path_abs)
            if normcase(start_drive) != normcase(path_drive):
>               raise ValueError("path is on mount %r, start on mount %r" % (
                    path_drive, start_drive))
E                   ValueError: path is on mount 'C:', start on mount '\\\\l2export\\iss02.charpier'

C:\Program Files\Python39\lib\ntpath.py:703: ValueError
``` 

Any clue on what do I have this issue ? I was doing the same thing (I believe) on version 0.98.0.dev0 and it was working correctly. 

Thank you, 

Anthony

[...]

PR #2219, opened 2023-11-17
Add Windows specific dump options

### LINKED ISSUE #2186: Automatic Mapping of Network Drives for Windows
## Issue

As discussed in #1891, SI fails for network drives (in Windows) because the mapped version of the drive and the actual path of the drive don't sync up.
For example, 

```
z:\server_folder\data
```
is actually
```
0.0.0.0\\server_folder\data
```

Normally this discrepancy doesn't really matter. But inside the SI machinery the paths do not resolve to be the same.

If I put in my path as 
```python
from pathlib import Path
data_path = Path(r'z:\server_folder\data')
```
this is accepted by SpikeInterface in a data extractor
```python

recording = se.read_mearc(data_path / "mearec_test_10s.h5")
```
But then if I try to save or run sorter it fails
```python
recording.save(folder=data_path/ "new_folder", overwrite=True)
```

```python

  Cell In[4], line 1
    recording.save(folder=rec_path/"new_test", overwrite=True)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:779 in save
    loaded_extractor = self.save_to_folder(**kwargs)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:864 in save_to_folder
    cached = self._save(folder=folder, verbose=verbose, **save_kwargs)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\baserecording.py:480 in _save
    binary_rec.dump(folder / "binary.json", relative_to=folder)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:569 in dump
    self.dump_to_json(file_path, relative_to=relative_to, folder_metadata=folder_metadata)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:602 in dump_to_json
    dump_dict = self.to_dict(

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:406 in to_dict
    dump_dict = _make_paths_relative(dump_dict, relative_to)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:983 in _make_paths_relative
    return recursive_path_modifier(d, func, target="path", copy=True)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\core_tools.py:824 in recursive_path_modifier
    recursive_path_modifier(kwargs, func, copy=False)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\core_tools.py:849 in recursive_path_modifier
    dc[k] = [func(e) for e in v]

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\core_tools.py:849 in <listcomp>
    dc[k] = [func(e) for e in v]

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:982 in <lambda>
    func = lambda p: os.path.relpath(str(p), start=relative)

  File ~\anaconda3\envs\spykes\Lib\ntpath.py:754 in relpath
    raise ValueError("path is on mount %r, start on mount %r" % (
ValueError: path is on mount 'Z:', start on mount '\\\\0.0.0.0\\server_folder'
```

or trying to `run_sorter` we get the same error:

```python
ss.run_sorter('tridesclous', recording, working_folder=data_path/ "output")
```

## Solution

The solution to this issue is to actually put in the path with UNC notation (forward slashes required), ie
```python
data_path = Path(r'0.0.0.0//server_folder/data')
```
Then the two paths will match. So the two possible solutions are:

1) Put it in the documentation to use UNC paths and hope users follow the rules (not ideal, user is sad)
2) Have a helper function in SI that converts mapped paths to UNC paths automatically (user is happy)

For (2) I don't yet know what machinery it would require inside of spikeinterface, so it might not be trivial to distinguish local
vs mapped drives. So it still might be beneficial for the time being to mention (either in a warning or in the docs) to use UNC 
pathways if wanting to interact with network mapped drives. (They are also terribly slow for everything). 

Now sorting locally and moving to storage in network mapped drive would be a reasonable request, but I will put that into
the next Windows issue that I need to open and link here.

## Notes

Also note this is different from the issue of sorting on `c:\` and saving sorting results on `d:\`. ie the `run_sorter` issue is distinct from this issue at the moment. The goal of this is issue is thinking about working on the same network mapped drive, so that the notations match up.

Also this is probably lower priority since sorting on the network is very very slow. So the bigger issue should be sorting locally and then saving the output to a different drive.

Edit to link other issue #2187

### LINKED ISSUE #2187: Problem with Sorting on one Drive (c:) and outputting results to Another Drive (d:) on Windows
## Issue

As discussed in #1891, when using `run_sorter`, it errors if two different drives are specified on Windows. (I'm stealing Heberto's example from the discussion): So saving from one drive to another works fine (as tested).

```python
from spikeinterface.core.datasets import download_dataset
from spikeinterface.extractors.neoextractors.mearec import (
    MEArecRecordingExtractor
)
from spikeinterface.core import load_extractor

local_file_path = download_dataset(remote_path="mearec/mearec_test_10s.h5")
recording = MEArecRecordingExtractor(local_file_path)

file_path = "C:\\Users\\heberto\\spikeinterface_datasets\\ephy_testing_data\\mearec\\mearec_test_10s.h5"
recording = MEArecRecordingExtractor(file_path=file_path)

MEArecRecordingExtractor: 32 channels - 32.0kHz - 1 segments - 320,000 samples - 10.00s 
                          float32 dtype - 39.06 MiB
  file_path: [C:\Users\heberto\spikeinterface_datasets\ephy_testing_data\mearec\mearec_test_10s.h5](file:///C:/Users/heberto/spikeinterface_datasets/ephy_testing_data/mearec/mearec_test_10s.h5)

folder_path = Path("D:\\folder_to_write_stuff")
folder_path.is_dir(), folder_path.anchor, folder_path.root

recording.save(folder=folder_path, overwrite=True)

write_binary_recording with n_jobs = 1 and chunk_size = 32000
write_binary_recording: 100%|##########| 10/10 [00:02<00:00,  3.67it/s]
BinaryFolderRecording: 32 channels - 32.0kHz - 1 segments - 320,000 samples - 10.00s 
                       float32 dtype - 39.06 MiB

import os

folder_path = "D:\\folder_to_write_stuff"
print(os.listdir(folder_path))

['provenance.json', 'probe.json', 'properties', 'traces_cached_seg0.raw', 'binary.json', 'si_folder.json']
```

But if instead we try to run sorter with the same dataset we get:

```python

  Cell In[6], line 1
    ss.run_sorter('tridesclous', recording, output_folder = folder_path)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\sorters\runsorter.py:147 in run_sorter

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\sorters\runsorter.py:170 in run_sorter_local
    output_folder = SorterClass.initialize_folder(recording, output_folder, verbose, remove_existing_folder)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\sorters\basesorter.py:141 in initialize_folder
    recording.dump(rec_file, relative_to=output_folder)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:569 in dump
    self.dump_to_json(file_path, relative_to=relative_to, folder_metadata=folder_metadata)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:602 in dump_to_json
    dump_dict = self.to_dict(

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:406 in to_dict
    dump_dict = _make_paths_relative(dump_dict, relative_to)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:983 in _make_paths_relative
    return recursive_path_modifier(d, func, target="path", copy=True)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\core_tools.py:824 in recursive_path_modifier
    recursive_path_modifier(kwargs, func, copy=False)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\core_tools.py:847 in recursive_path_modifier
    dc[k] = func(v)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:982 in <lambda>
    func = lambda p: os.path.relpath(str(p), start=relative)

  File ~\anaconda3\envs\spykes\Lib\ntpath.py:754 in relpath
    raise ValueError("path is on mount %r, start on mount %r" % (

ValueError: path is on mount 'C:', start on mount 'D:'
```
I will note that the `folder_to_write_stuff` and the `sorter_output` folders are created on the `D:` drive so error is after the folder generation.

## Solution

Not sure, but can move our new thread to discuss here. Maybe this is directly linked #2186 and both can be fixed by a similar global solution. As @h-mayorquin said, the `rec.save` works fine, so the resolving machinery between local drives isn't the main problem (as it is in #2186, which fails save and run_sorter), but maybe the order of the operations is contributing. 

## Notes

I would say this is a bit higher priority since many people would want to sort on C: where they have a fast drive and then move to either a storage drive or a portable drive. 

I marked as bug since from my (windows-centric) perspective this seems like expected behavior to just work (I know it's not trivial to implement), but if you think it falls more in the enhancement, I could also see it labeled that way instead if the guarantee is to work on the (`C:`) for Windows only and everything else is bonus.

zm711:
Fixes #2187 and #1891.

When in doubt follow Sam's intuition. I just do a platform.system check and if windows I don't use `relative_to`. Now it lets me sort from (I've tested this on 3 separate data sets now--but if you want me to do some more specific local tests let me know):
```
c->d
c->network
```

with no problems. I know @h-mayorquin would prefer not to have OS specific behavior, but this works. So for Windows users that do want to sort and switch drives this seems like the best solution for now (maybe we add a place holder test for the future when testing occurs on all os's?)

I also added this to pickle, but can delete if you only want this on the json dump.

Unfortunately to use `run_sorter` Windows users must use mapped drives and not unc paths because the `cmd prompt` where the spikeinterface script is run does not accept uncs paths. So this doesn't fix #2186 (unfortunately--since for some parts of SI we need unc paths and for others we need mapped paths).

### DIFF
--- a/src/spikeinterface/sorters/basesorter.py
+++ b/src/spikeinterface/sorters/basesorter.py
@@ -9,7 +9,7 @@ import json
 import traceback
 import shutil
 import warnings
-
+import platform
 
 from spikeinterface.core import load_extractor, BaseRecordingSnippets
 from spikeinterface.core.core_tools import check_json
@@ -138,9 +138,23 @@ class BaseSorter:
 
         rec_file = output_folder / "spikeinterface_recording.json"
         if recording.check_serializablility("json"):
-            recording.dump(rec_file, relative_to=output_folder)
+            # because Windows has different drives we can't do `relative_to`
+            # unless user is on same drive, which is not guaranteed
+            if platform.system() == "Windows":
+                recording.dump(
+                    rec_file,
+                )
+            else:
+                recording.dump(rec_file, relative_to=output_folder)
         elif recording.check_serializablility("pickle"):
-            recording.dump(output_folder / "spikeinterface_recording.pickle", relative_to=output_folder)
+            # because Windows has different drives we can't do `relative_to`
+            # unless user is on same drive, which is not guaranteed
+            if platform.system() == "Windows":
+                recording.dump(
+                    output_folder / "spikeinterface_recording.pickle",
+                )
+            else:
+                recording.dump(output_folder / "spikeinterface_recording.pickle", relative_to=output_folder)
         else:
             # TODO: deprecate and finally remove this after 0.100
             d = {"warning": "The recording is not serializable to json"}


src/spikeinterface/sorters/basesorter.py
```diff
@@ -138,9 +138,23 @@ def initialize_folder(cls, recording, output_folder, verbose, remove_existing_fo
...
+            # because Windows has different drives we can't do `relative_to`
+            # unless user is on same drive, which is not guaranteed
+            if platform.system() == "Windows":
+                recording.dump(
+                    output_folder / "spikeinterface_recording.pickle",
+                )
+            else:
+                recording.dump(output_folder / "spikeinterface_recording.pickle", relative_to=output_folder)
         else:
             # TODO: deprecate and finally remove this after 0.100
```
zm711, line 159: Did we want to add a deprecation warning in the else in this PR?

h-mayorquin:
@zm711 Sam is fine with doing windows testing it seems. I was against having different OS behavior without testing because neither me Alessio or Sam use Windows. We would get errors that we would not be able to catch or test. But this should be a moot point once we do windows testing in the CI. Probably mid-December seeing how much work we have : /

src/spikeinterface/sorters/basesorter.py
```diff
@@ -138,9 +138,23 @@ def initialize_folder(cls, recording, output_folder, verbose, remove_existing_fo
 
         rec_file = output_folder / "spikeinterface_recording.json"
         if recording.check_serializablility("json"):
-            recording.dump(rec_file, relative_to=output_folder)
+            # because Windows has different drives we can't do `relative_to`
+            # unless user is on same drive, which is not guaranteed
+            if platform.system() == "Windows":
+                recording.dump(
```
h-mayorquin, line 144: You are absically throwing the `relative_to` out in windows.

Wouldn't it better to instead tell users that this option is not supported in windows so they don't run away with the impression that they did indeed save their files with relative paths?

zm711: So add a warning? Or an assert level and make them re-enter without relative_to?

zm711: I'm still reading up on this part of the code. I just implemented Sam's suggestion, but haven't read too deep into the rest of the function. Once I made his change I tested and it worked, but I agree Windows users should know. For instance if I do c-> c I could do relative_to without problem. So maybe there is a more universal change for Windows that would be better.

h-mayorquin: Is that a folder that you are saving or a file?
Are you able to load it again?

zm711: Both. So I have tested this with `run_sorter` and `run_sorter_by_property` with Kilosort2 and Kilosort3 (I am able to read the folder of files with `read_kilosort` and with Mountainsort5 (saves one `.npz` file, which I am able to read with `read_npz_sorting`). I haven't tested any sort of portability yet though. But the provenance files/log files all give absolute paths, so I'm assuming they are not portable, but I can test Monday.

h-mayorquin: Yes, they are not portable. Can you load them with the `base_folder` keyword, this is what I mean,

zm711: Could you give me a code snippet of what you want me to try please? I have never loaded anything with the `base_folder` keyword.

zm711: So testing
`recording.dump_to_pickle(..)`
`recording.dump(..)`

both failed with this stack trace
```
  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:600 in dump
    self.dump_to_json(file_path, relative_to=relative_to, folder_metadata=folder_metadata)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:633 in dump_to_json
    dump_dict = self.to_dict(

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:437 in to_dict
    dump_dict = _make_paths_relative(dump_dict, relative_to)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:1014 in _make_paths_relative
    return recursive_path_modifier(d, func, target="path", copy=True)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\core_tools.py:824 in recursive_path_modifier
    recursive_path_modifier(kwargs, func, copy=False)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\core_tools.py:847 in recursive_path_modifier
    dc[k] = func(v)

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:1013 in <lambda>
    func = lambda p: os.path.relpath(str(p), start=relative)

  File ~\anaconda3\envs\spykes\Lib\ntpath.py:754 in relpath
    raise ValueError("path is on mount %r, start on mount %r" % (

ValueError: path is on mount 'C:', start on mount 'D:'
```
trying `sorting.dump_to_pickle` failed with:
```

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\base.py:659 in dump_to_pickle
    file_path.write_bytes(pickle.dumps(dump_dict))

PicklingError: Can't pickle <function BaseExtractor.from_dict at 0x000001C297972020>: it's not the same object as spikeinterface.core.base.BaseExtractor.from_dict
```

So it seems `relative_to` fails on Windows when moving between drives as a feature. Is there something else I can test? Or do you think this would need to be a deeper change then just at the sorting level in this case?

samuelgarcia:
Hi Zach,
sorry for the delay to give feedbakc on this.
Here you are removing the relative machanism for windows in sorter.

We discuss a bit and finally we will implement sometings in the same direction :
 * at all place we use relative_to dump
 * add a function that check if relative_to is possible or not.
 * if not possible then use the absolute if possible then use the relative.

This function checking could be : I am on windows + path are unc or dos one + path are on same drive

I propose to start this soon.

This would close this PR so. I leave this PR open as a reminder (to avoid procastination).

zm711:
Sounds good all! Thanks.

PR #2223, opened 2023-11-17
Fix #2201

### LINKED ISSUE #2201: Concatenating files to sort together
Hello! I have multiple ~3 minute long recordings that I want to sort together. But when I try this:
image

I get this runtime warning, and the size of the concatenated file is less than the size of any of the individual files:
image

Could you please let me know what I'm doing wrong? Greatly appreciate any help, thanks!

h-mayorquin:
Annoying bug that happens with concatenation but can happen with any recording that is long enough. Here is the problem. To create the file, the size of the file-to-be is esimated with this line:

https://github.com/h-mayorquin/spikeinterface/blob/65e4c178dc3bd0f826ce2b10fcb31a8e7da4698b/src/spikeinterface/core/core_tools.py#L297

But sometimes the number of frames is a numpy scalar (and not a python integer). Numpy scalars behave like numpy objects and they have an overflow.  That means that the file size ends up being smaller than it should and the writting to the buffer fails.

You can see that there is an overflown ins the screenshot that the user shared with us:

image

Plus, you can see that the concatenated recording is smaller than the pieces. 

So, to solves it, we need to be sure that `get_num_samples()`i is not a numpy scalar but a python scalar (they can be as long as as they need to be). I modified the code to do that.

### DIFF
--- a/src/spikeinterface/core/core_tools.py
+++ b/src/spikeinterface/core/core_tools.py
@@ -204,43 +204,6 @@ def _init_binary_worker(recording, file_path_dict, dtype, byte_offest, cast_unsi
     return worker_ctx
 
 
-# used by write_binary_recording + ChunkRecordingExecutor
-def _write_binary_chunk(segment_index, start_frame, end_frame, worker_ctx):
-    # recover variables of the worker
-    recording = worker_ctx["recording"]
-    dtype = worker_ctx["dtype"]
-    byte_offset = worker_ctx["byte_offset"]
-    cast_unsigned = worker_ctx["cast_unsigned"]
-    file = worker_ctx["file_dict"][segment_index]
-
-    # Open the memmap
-    # What we need is the file_path
-    num_channels = recording.get_num_channels()
-    num_frames = recording.get_num_frames(segment_index=segment_index)
-    shape = (num_frames, num_channels)
-    dtype_size_bytes = np.dtype(dtype).itemsize
-    data_size_bytes = dtype_size_bytes * num_frames * num_channels
-
-    # Offset (The offset needs to be multiple of the page size)
-    # The mmap offset is associated to be as big as possible but still a multiple of the page size
-    # The array offset takes care of the reminder
-    mmap_offset, array_offset = divmod(byte_offset, mmap.ALLOCATIONGRANULARITY)
-    mmmap_length = data_size_bytes + array_offset
-    memmap_obj = mmap.mmap(file.fileno(), length=mmmap_length, access=mmap.ACCESS_WRITE, offset=mmap_offset)
-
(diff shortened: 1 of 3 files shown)


src/spikeinterface/core/core_tools.py
```diff
@@ -312,6 +275,43 @@ def write_binary_recording(
     executor.run()
 
 
+# used by write_binary_recording + ChunkRecordingExecutor
```
h-mayorquin, line 278: I moved this so it follows the pattern that all the `ChunkRecordingExector` chunk functions follow. The function is after the caller of `ChunkRecordingExecutor`. Sorry for my OCD.

src/spikeinterface/core/segmentutils.py
```diff
@@ -148,17 +148,17 @@ def __init__(self, recording_list, ignore_times=True, sampling_frequency_max_dif
 class ProxyConcatenateRecordingSegment(BaseRecordingSegment):
     def __init__(self, parent_segments, sampling_frequency, ignore_times=True):
         if ignore_times:
-            d = {}
-            d["t_start"] = None
-            d["time_vector"] = None
-            d["sampling_frequency"] = sampling_frequency
+            time_kwargs = {}
```
h-mayorquin, line 151: This also got in my way when debuggin, sorry @samuelgarcia . If you want to keep single letter, feel free to rever to something other than d.

src/spikeinterface/core/core_tools.py
alejoe91: No worries! Honestly I also found it aannoying

DradeAW:
> But sometimes the number of frames is a numpy scalar (and not a python integer). Numpy scalars behave like numpy objects and they have an overflow

You managed to exceed an `int64`!?
I'm curious to know in what context

h-mayorquin:
@DradeAW 
I am not sure it was int64. But this is the line:

```
        data_size_bytes = dtype_size_bytes * num_frames * num_channels
```

`num_frames` as a numpy scalar, so that means `data_size_bytes` was also a numpy scalar. The product exceed the size of the numpy dtype and there the error. You can check the parent issue for the numbers, they will be reflected on the screenshots and you can calculate from there what datatype was being used.

DradeAW:
For a recording using 8 bytes (int64, float64) lasting a full day at 100 kHz on 10,000 channels, that's `6.9*10^14`
`int64 = 2^63 = 9.2*10^18`

I already had a problem of overflow with the coincidence matrix.
I think we should stop using `int32` and `int16` in cases where it does not matter to stop this problem from happening. `int64` covers almost all cases and is future-proof.

h-mayorquin:
For these kind of thing normal python int is even better. The problem was that we propagated a numpy scalar.

alejoe91 (approved):

PR #2226, opened 2023-11-18
Add posion statistics to generate_sorting

h-mayorquin:
I added a function that generates spikes as if they are a poisson process which, in my understanding, is the most common statistics of firing rate.

This is is also 3 times faster. Generates a ten hour sorting with 1000 units in 7 seconds instead of 20 as the current one. A ~ 3 speedup.

```
--------------------------------------------------
`synthesize_poisson_spike_vector`
Mean time over 3 iterations: 6.94 seconds
Std over 3 iterations: 0.01 seconds
times=['6.92', '6.96', '6.95']
--------------------------------------------------
`synthesize_random_firings`
Mean time over 3 iterations: 21.12 seconds
Std over 3 iterations: 0.09 seconds
times=['21.05', '21.05', '21.25']

Speedup: 3.04
```

Plus, it is more memory efficient as it has around 70 % memory requirements of the current function (I did try to make it less memory hungry but it is  HARD). It also seems that the current implementation has some kind of leak, see the temporal profile of memory utilizaiton in the profile above.

image

The first three are the PR implementation, the last three is the current implementation.

### DIFF
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -263,7 +263,7 @@ class BaseSorting(BaseExtractor):
 
     def get_total_num_spikes(self):
         warnings.warn(
-            "Sorting.get_total_num_spikes() is deprecated, se sorting.count_num_spikes_per_unit()",
+            "Sorting.get_total_num_spikes() is deprecated and will be removed in spikeinterface 0.102, se sorting.count_num_spikes_per_unit()",
             DeprecationWarning,
             stacklevel=2,
         )
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -168,7 +168,7 @@ def generate_sorting(
     spikes = []
     for segment_index in range(num_segments):
         num_samples = int(sampling_frequency * durations[segment_index])
-        times, labels = synthesize_random_firings(
+        times, labels = synthesize_random_firings_poisson(
             num_units=num_units,
             sampling_frequency=sampling_frequency,
             duration=durations[segment_index],
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -386,23 +386,21 @@ def synthesize_random_firings_poisson(
     spike_frames = np.cumsum(inter_spike_frames, axis=1, out=inter_spike_frames)
     spike_frames = spike_frames.ravel()
 
-    # We map units to corresponding spike times and flatten the array
-    unit_indices = np.repeat(np.arange(num_units, dtype="uint16"), num_spikes_max).ravel()
+    # We map un its to corresponding spike times and flatten the array
+    unit_indices = np.repeat(np.arange(num_units, dtype="uint16"), num_spikes_max)
 
     # Eliminate spikes that are beyond the duration
     mask = spike_frames <= max_frames
     num_correct_frames = np.sum(mask)
-    spike_frames[:num_correct_frames] = spike_frames[mask]
-    unit_indices[:num_correct_frames] = unit_indices[mask]
+    spike_frames[:num_correct_frames] = spike_frames[mask]  # Avoids a malloc
+    unit_indices = unit_indices[mask]
 
-    # All of this to avoid a malloc
     spike_frames = spike_frames[:num_correct_frames]
-    unit_indices = unit_indices[:num_correct_frames]
-
     # This should use tim or radix sort which is good for integers and presorted data. I profiled. re-profile in doubt.
     sort_indices = np.argsort(spike_frames, kind="stable")
-    spike_frames = spike_frames[sort_indices]  # This is still generating a malloc, probably the ravel
+
     unit_indices = unit_indices[sort_indices]
+    spike_frames = spike_frames[sort_indices]
 
     return spike_frames, unit_indices
 
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -3,6 +3,7 @@ import warnings
 import numpy as np
 from typing import Union, Optional, List, Literal
 import warnings
+from math import ceil
 
 
 from .numpyextractors import NumpyRecording, NumpySorting
@@ -168,7 +169,7 @@ def generate_sorting(
     spikes = []
     for segment_index in range(num_segments):
         num_samples = int(sampling_frequency * durations[segment_index])
-        times, labels = synthesize_random_firings_poisson(
+        samples, labels = synthesize_random_firings_poisson(
             num_units=num_units,
             sampling_frequency=sampling_frequency,
             duration=durations[segment_index],
@@ -179,11 +180,11 @@ def generate_sorting(
 
         if empty_units is not None:
             keep = ~np.isin(labels, empty_units)
-            times = times[keep]
+            samples = samples[keep]
             labels = labels[keep]
 
-        spikes_in_seg = np.zeros(times.size, dtype=minimum_spike_dtype)
-        spikes_in_seg["sample_index"] = times
+        spikes_in_seg = np.zeros(samples.size, dtype=minimum_spike_dtype)
+        spikes_in_seg["sample_index"] = samples
         spikes_in_seg["unit_index"] = labels
         spikes_in_seg["segment_index"] = segment_index
         spikes.append(spikes_in_seg)
@@ -319,46 +320,93 @@ def generate_snippets(
 ## spiketrain zone ##
 def synthesize_random_firings_poisson(
     num_units=20,
-    sampling_frequency=30000.0,  # Not sure we need this?
+    sampling_frequency=30000.0,
     duration=60,
     refractory_period_ms=4.0,
     firing_rates=3.0,
     seed=0,
 ):
+    """
+    Generate random spike frames for neuronal units based on a Poisson process.
+
+    This function simulates the spike frames of a number of neuronal units, each firing according to
+    a Poisson process. The spike times are discretized to align with a given sampling frequency.
+
+    Parameters
+    ----------
+    num_units : int, optional
+        Number of neuronal units to simulate (default is 20).
+    sampling_frequency : float, optional
+        Sampling frequency in Hz (default is 30000.0).
+    duration : float, optional
+        Duration of the simulation in seconds (default is 60).
+    refractory_period_ms : float, optional
+        Refractory period between spikes in milliseconds (default is 4.0).
+    firing_rates : float or array_like, optional
+        Firing rate(s) in Hz. Can be a single value or an array of firing rates for each unit
+        (default is 3.0).
+    seed : int, optional
+        Seed for random number generator (default is 0).
+
+    Returns
+    -------
+    spike_frames : ndarray
+        1D array of spike frames.
+    unit_indices : ndarray
+        1D array of unit indices corresponding to each spike.
+
+    Notes
+    -----
+    - The function uses a geometric distribution to simulate the discrete inter-spike intervals,
+    based that would be an exponential process.
+    - To ensure all spikes within the duration are captured, it initially generates more spikes
+    than expected and then filters out the excess.
+    - Refractory periods are enforced by adding a fixed number of frames to subsequent spikes.
+    - The output is sorted based on spike timings for easier analysis.
+    """
+
     rng = np.random.default_rng(seed=seed)
 
     if np.isscalar(firing_rates):
         firing_rates = np.full(num_units, firing_rates, dtype="float64")
 
-    mean_inter_spike_time = 1.0 / firing_rates
-    spikes_per_unit = (duration * firing_rates).astype(int)
-    max_spikes = np.max(spikes_per_unit)
-
     refractory_period_seconds = refractory_period_ms / 1000.0
+    refactory_period_frames = int(refractory_period_seconds * sampling_frequency)
 
-    # Generate inter spike times, add the refactor period and accumulate for sorted spike times
-    scale = mean_inter_spike_time[:, np.newaxis] - refractory_period_seconds
-    inter_spike_frames = rng.exponential(scale=scale, size=(num_units, max_spikes))
-    inter_spike_frames += refractory_period_seconds
+    # Equivalence between exponential and geometric distributions scaled to the discrete mean
+    geometric_p = 1 - np.exp(-firing_rates / sampling_frequency)
+
+    # We generate as many spikes as the mean plus two std to be sure we have enough
+    max_frames = duration * sampling_frequency
+    max_p = geometric_p.max()
+    num_spikes_expected = ceil(max_frames * max_p)
+    num_spikes_std = int(np.sqrt(num_spikes_expected * (1 - max_p)))
+    num_spikes_max = num_spikes_expected + 2 * num_spikes_std
+
+    # Generate inter spike frames, add the refactory samples and accumulate for sorted spike frames
+    inter_spike_frames = rng.geometric(p=geometric_p[:, np.newaxis], size=(num_units, num_spikes_max))
+    inter_spike_frames[:, 1:] += refactory_period_frames
 
     spike_frames = np.cumsum(inter_spike_frames, axis=1, out=inter_spike_frames)
     spike_frames = spike_frames.ravel()
 
     # We map units to corresponding spike times and flatten the array
-    unit_indices = np.repeat(np.arange(num_units, dtype="uint16"), max_spikes).ravel()
+    unit_indices = np.repeat(np.arange(num_units, dtype="uint16"), num_spikes_max).ravel()
 
     # Eliminate spikes that are beyond the duration
-    mask = spike_frames < duration
-    num_spikes_correct_duration = np.sum(mask)
-    spike_frames[:num_spikes_correct_duration] = spike_frames[mask]
-    unit_indices[:num_spikes_correct_duration] = unit_indices[mask]
+    mask = spike_frames <= max_frames
+    num_correct_frames = np.sum(mask)
+    spike_frames[:num_correct_frames] = spike_frames[mask]
+    unit_indices[:num_correct_frames] = unit_indices[mask]
+
+    # All of this to avoid a malloc
+    spike_frames = spike_frames[:num_correct_frames]
+    unit_indices = unit_indices[:num_correct_frames]
 
     # This should use tim or radix sort which is good for integers and presorted data. I profiled, re-profile in doubt.
     sort_indices = np.argsort(spike_frames, kind="stable")
     spike_frames = spike_frames[sort_indices]  # This is still generating a malloc, probably the ravel
     unit_indices = unit_indices[sort_indices]
-
-    spike_frames = np.take_along_axis(spike_frames, sort_indices, axis=0)
 
     return spike_frames, unit_indices
 
@@ -403,14 +451,6 @@ def synthesize_random_firings(
     """
 
     rng = np.random.default_rng(seed=seed)
-
-    # unit_seeds = [rng.integers(0, 2 ** 63) for i in range(num_units)]
-
-    # if seed is not None:
-    #     np.random.seed(seed)
-    #     seeds = np.random.RandomState(seed=seed).randint(0, 2147483647, num_units)
-    # else:
-    #     seeds = np.random.randint(0, 2147483647, num_units)
 
     if np.isscalar(firing_rates):
         firing_rates = np.full(num_units, firing_rates, dtype="float64")
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -317,6 +317,51 @@ def generate_snippets(
 
 
 ## spiketrain zone ##
+def synthesize_random_firings_poisson(
+    num_units=20,
+    sampling_frequency=30000.0,  # Not sure we need this?
+    duration=60,
+    refractory_period_ms=4.0,
+    firing_rates=3.0,
+    seed=0,
+):
+    rng = np.random.default_rng(seed=seed)
+
+    if np.isscalar(firing_rates):
+        firing_rates = np.full(num_units, firing_rates, dtype="float64")
+
+    mean_inter_spike_time = 1.0 / firing_rates
+    spikes_per_unit = (duration * firing_rates).astype(int)
+
+    refractory_period_seconds = refractory_period_ms / 1000.0
+
+    unit_indices = np.empty(num_units, dtype="uint16")
+
+    # Generate spike times for each unit
+    max_spikes = np.max(spikes_per_unit)
+    scale = mean_inter_spike_time[:, np.newaxis] - refractory_period_seconds  # Keeps the firing rate as expected
+    inter_spike_frames = rng.exponential(scale=scale, size=(num_units, max_spikes))
+    inter_spike_frames += refractory_period_seconds
+
+    # This already produces pre-sorted spike times and flatten the array
+    spike_frames = np.cumsum(inter_spike_frames, axis=1, out=inter_spike_frames)
+    spike_frames = spike_frames.ravel()
+
+    # We map units to corresponding spike times and flatten the array
+    unit_indices = np.empty(spike_frames.size, dtype="uint16")
+    unit_indices[:] = np.repeat(np.arange(num_units), max_spikes).ravel()
+
+    # Eliminate spikes that are beyond the duration
+    mask = spike_frames < duration
+    num_spikes_correct_duration = np.sum(mask)
+    spike_frames[:num_spikes_correct_duration] = spike_frames[mask]
+    unit_indices[:num_spikes_correct_duration] = unit_indices[mask]
+
+    sort_indices = np.argsort(spike_frames, kind="stable")
+    spike_frames[:] = spike_frames[sort_indices]  # This is still generating a malloc, probably the ravel
+    unit_indices[:] = unit_indices[sort_indices]
+
+    return spike_frames, unit_indices
 
 
 def synthesize_random_firings(
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -396,7 +396,7 @@ def synthesize_random_firings_poisson(
     unit_indices = unit_indices[mask]
 
     spike_frames = spike_frames[:num_correct_frames]
-    # This should use tim or radix sort which is good for integers and presorted data. I profiled. re-profile in doubt.
+    # Stable should use tim or radix sort which is good for integers and presorted data. I profiled. re-profile in doubt.
     sort_indices = np.argsort(spike_frames, kind="stable")
 
     unit_indices = unit_indices[sort_indices]
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -169,7 +169,7 @@ def generate_sorting(
     spikes = []
     for segment_index in range(num_segments):
         num_samples = int(sampling_frequency * durations[segment_index])
-        samples, labels = synthesize_random_firings_poisson(
+        samples, labels = synthesize_poisson_spike_vector(
             num_units=num_units,
             sampling_frequency=sampling_frequency,
             duration=durations[segment_index],
@@ -318,7 +318,7 @@ def generate_snippets(
 
 
 ## spiketrain zone ##
-def synthesize_random_firings_poisson(
+def synthesize_poisson_spike_vector(
     num_units=20,
     sampling_frequency=30000.0,
     duration=60,
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -332,24 +332,20 @@ def synthesize_random_firings_poisson(
 
     mean_inter_spike_time = 1.0 / firing_rates
     spikes_per_unit = (duration * firing_rates).astype(int)
+    max_spikes = np.max(spikes_per_unit)
 
     refractory_period_seconds = refractory_period_ms / 1000.0
 
-    unit_indices = np.empty(num_units, dtype="uint16")
-
-    # Generate spike times for each unit
-    max_spikes = np.max(spikes_per_unit)
-    scale = mean_inter_spike_time[:, np.newaxis] - refractory_period_seconds  # Keeps the firing rate as expected
+    # Generate inter spike times, add the refactor period and accumulate for sorted spike times
+    scale = mean_inter_spike_time[:, np.newaxis] - refractory_period_seconds
     inter_spike_frames = rng.exponential(scale=scale, size=(num_units, max_spikes))
     inter_spike_frames += refractory_period_seconds
 
-    # This already produces pre-sorted spike times and flatten the array
     spike_frames = np.cumsum(inter_spike_frames, axis=1, out=inter_spike_frames)
     spike_frames = spike_frames.ravel()
 
     # We map units to corresponding spike times and flatten the array
-    unit_indices = np.empty(spike_frames.size, dtype="uint16")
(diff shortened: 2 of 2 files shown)


CodyCBakerPhD:
@h-mayorquin Curious what speeds you observe with

```python
import numpy

seed = 0
random_number_generator = numpy.random.default_rng(seed=seed)

number_of_units = 1_000

firing_rates = 10.0  # Hz, scalar usage
# firing_rates = [10.0 for unit_index in range(number_of_units)]  # Hz, vector usage

duration = 60.0  # seconds

# sampling_frequency = None
sampling_frequency = 30_000  # Hz, if specified

# refractory_period = None
refractory_period = 4.0  # milliseconds


def _clean_refractory_period(original_spike_times: numpy.ndarray, refractory_period_seconds: float) -> numpy.ndarray:
    inter_spike_intervals = numpy.diff(original_spike_times, prepend=refractory_period_seconds)
    violations = inter_spike_intervals < refractory_period_seconds  # scale ms to s
    if numpy.any(violations):
        return original_spike_times
    spike_time_shifts = refractory_period_seconds - inter_spike_intervals[violations]
    return original_spike_times[violations] + spike_time_shifts


if numpy.isscalar(firing_rates):
    number_of_spikes_per_unit = random_number_generator.poisson(lam=firing_rates * duration, size=number_of_units)
else:
    number_of_spikes_per_unit = numpy.empty(shape=number_of_units, dtype="uint16")
    for unit_index in range(number_of_units):
        number_of_spikes_per_unit[unit_index] = int(
            random_number_generator.poisson(lam=firing_rates[unit_index] * duration, size=1)
        )

spike_times = list()
if sampling_frequency is None:
    for number_of_spikes in number_of_spikes_per_unit:
        spikes = numpy.sort(random_number_generator.uniform(low=0, high=duration, size=number_of_spikes))
        if refractory_period is not None:
            spikes = _clean_refractory_period(
                original_spike_times=spikes, refractory_period_seconds=refractory_period / 1e3
            )
        spike_times.append(spikes)
else:
    for number_of_spikes in number_of_spikes_per_unit:
        spikes = numpy.sort(
            random_number_generator.integers(
                low=0, high=int(duration * sampling_frequency), size=number_of_spikes, dtype="uint64"
            )
        ) / sampling_frequency
        if refractory_period is not None:
            spikes = _clean_refractory_period(
                original_spike_times=spikes, refractory_period_seconds=refractory_period / 1e3
            )
        spike_times.append(spikes)
```

src/spikeinterface/core/basesorting.py
```diff
@@ -263,7 +263,7 @@ def get_unit_property(self, unit_id, key):
 
     def get_total_num_spikes(self):
         warnings.warn(
-            "Sorting.get_total_num_spikes() is deprecated, se sorting.count_num_spikes_per_unit()",
+            "Sorting.get_total_num_spikes() is deprecated and will be removed in spikeinterface 0.102, se sorting.count_num_spikes_per_unit()",
```
zm711, line 266:
```suggestion
            "Sorting.get_total_num_spikes() is deprecated and will be removed in spikeinterface 0.102, use sorting.count_num_spikes_per_unit()",
```

zm711: Just because se might make people think of sorting extractors instead of use. I know this will be removed, but just to be clear.

src/spikeinterface/core/generate.py
```diff
@@ -317,6 +318,91 @@ def generate_snippets(
...
+        Number of neuronal units to simulate (default is 20).
+    sampling_frequency : float, optional
+        Sampling frequency in Hz (default is 30000.0).
+    duration : float, optional
+        Duration of the simulation in seconds (default is 60).
+    refractory_period_ms : float, optional
+        Refractory period between spikes in milliseconds (default is 4.0).
+    firing_rates : float or array_like, optional
+        Firing rate(s) in Hz. Can be a single value or an array of firing rates for each unit
+        (default is 3.0).
```
zm711, line 347:
```suggestion
    num_units : int, default: 20
        Number of neuronal units to simulate 
    sampling_frequency : float, default: 30000.0
        Sampling frequency in Hz 
    duration : float, default: 60.0
        Duration of the simulation in seconds
    refractory_period_ms : float, default: 4.0
        Refractory period between spikes in milliseconds
    firing_rates : float or array_like, default: 3.0
        Firing rate(s) in Hz. Can be a single value for all units or an array of firing rates with
        each element being the firing rate for one unit
```
Just enforcing docstring style :)

I made the last one a little more verbose (not that I thought yours was unclear, but I sometimes it is nice to spell it out. You can feel free to revert that part if you want.

src/spikeinterface/core/generate.py
```diff
@@ -317,6 +318,91 @@ def generate_snippets(
 
 
 ## spiketrain zone ##
+def synthesize_poisson_spike_vector(
+    num_units=20,
+    sampling_frequency=30000.0,
+    duration=60,
```
zm711, line 324:
```suggestion
    duration=60.0,
```
Since the docstring says it should be a float.

src/spikeinterface/core/generate.py
```diff
@@ -317,6 +318,91 @@ def generate_snippets(
...
+    duration : float, optional
+        Duration of the simulation in seconds (default is 60).
+    refractory_period_ms : float, optional
+        Refractory period between spikes in milliseconds (default is 4.0).
+    firing_rates : float or array_like, optional
+        Firing rate(s) in Hz. Can be a single value or an array of firing rates for each unit
+        (default is 3.0).
+    seed : int, optional
+        Seed for random number generator (default is 0).
+
```
zm711, line 350:
```suggestion
    seed : int, default: 0 
        Seed for random number generator

```

src/spikeinterface/core/generate.py
```diff
@@ -317,6 +318,91 @@ def generate_snippets(
...
+    -------
+    spike_frames : ndarray
+        1D array of spike frames.
+    unit_indices : ndarray
+        1D array of unit indices corresponding to each spike.
+
+    Notes
+    -----
+    - The function uses a geometric distribution to simulate the discrete inter-spike intervals,
+    based that would be an exponential process for continuous time.
```
zm711, line 361: This isn't quite clear. Maybe it is missing a word. I'm not quite sure how to fix.

zm711 (commented):
docstring cleanup :)

src/spikeinterface/core/generate.py
h-mayorquin: You are correct this is not clear. I will expand on this. Thanks. Also for all the other comments. They all make sense and are very helpful as usual.

h-mayorquin: I modified the docstring here, let me know what you think and if it is still unclear to you. Also, any other advice you might have is useful.

h-mayorquin:
@CodyCBakerPhD 
Most of the computation in the function happens here:
https://github.com/catalystneuro/spikeinterface/blob/ac41b95eb31fc76d21fd2e8f0917092d086b481e/src/spikeinterface/core/generate.py#L400-L405

That is the bottleneck. Generating sorted concatenated spikes and their correspoding units. This is how the output should look like (more or less):

```python
spike_frames
[  452 15511 37989 42417 62234 71248 74939]
unit indexes
[1 1 0 0 1 1 0]
```

The concatenated frames have to be sorted but the frames of different units can happen at the same time.

Reading your code it seems that you just concatenated sorted spikes and I can't see the vector of the units. Adding the concatenation and the sorting without the units part the code is already slower than the old implementation with a %%timeit decorator in a notebook.

src/spikeinterface/core/generate.py
```diff
@@ -317,6 +318,93 @@ def generate_snippets(
...
+
+    # We generate as many spikes as the mean plus two std to be sure we have enough
+    max_frames = duration * sampling_frequency
+    max_p = geometric_p.max()
+    num_spikes_expected = ceil(max_frames * max_p)
+    num_spikes_std = int(np.sqrt(num_spikes_expected * (1 - max_p)))
+    num_spikes_max = num_spikes_expected + 2 * num_spikes_std
+
+    # Generate inter spike frames, add the refactory samples and accumulate for sorted spike frames
+    inter_spike_frames = rng.geometric(p=geometric_p[:, np.newaxis], size=(num_units, num_spikes_max))
```
DradeAW, line 385: Why use this complex stuff rather than a simple `refractory_period + np.random.exponential(scale=beta, size=...)`?
Is this faster / more memory efficient?

+ only 2*std will fail around 2.5% of the time. I think 4*std is best :)

h-mayorquin: > There is a memory spikes coverage trade-off here.
I am not sure if you understood me. The longer the cutout (the one that I am making 2 * stds) the more extra memory I have to allocate and then drop. Maybe I am not understanding what you mean by negligible?

I am not sure what you mean by bias, it is surely not in a formal sense, right (like a biased estimator). Can you be more specific about what statistical effect you are talking about? 

> To me it sounds statistically easier to think about it that way, and the code is easier to read (but it might be just me)
You might be right about it. If there are no memory-costs or if there is something wrong statistically I might change to this. I don't think that you can use float allocated space to allocate integers though so I don't know if you avoid memory costs and the argument will have to rely in something being wrong statistically.

DradeAW:
> Maybe I am not understanding what you mean by negligible?

To me, the difference between 10,000 ; 10,200 and 10,400 is negligible.

> I am not sure what you mean by bias, it is surely not in a formal sense

I meant that the resulting ISI might be erroneous.
Let's take the (extreme) example where lambda is significantly smaller than the delta time between two samples.
Then you're only generating zeros. But if you are casting after, you will still add values smaller than 0 and after multiple spikes, it will go to the next sample.

h-mayorquin:
> To me, the difference between 10,000 ; 10,200 and 10,400 is negligible.

Right, it was me who did not understood above. You are correct. Let's switch to 4 stds. I think it makes sense.


> I meant that the resulting ISI might be erroneous.
Let's take the (extreme) example where lambda is significantly smaller than the delta time between two samples.
Then you're only generating zeros. But if you are casting after, you will still add values smaller than 0 and after multiple spikes, it will go to the next sample.



I still don't understand this. Lambda is the firing rate, right?
What would be an erroneous ISI? Maybe can test this, are you saying that the frames generated by this function transformed to times will not have an exponential distribution in some sense?

Also, this is something that the current implementation is doing, generating the frames in integers. Is this being avoided right now? Maybe this will be helpful to me for understanding.

DradeAW:
Let's take an example where the generation would give `[4.4, 3.3, 6.3]`. The cumulative sum is `[4.4, 7.7, 14.0]`, which after casting is `[4, 7/8, 14]` (depending on the rule)

But if you first convert to integers, then you have `[4, 7, 13]` (which is clearly different, and what I was referring to when I said bias).
Of course if the firing rate is low, you won't notice any difference. But with a high firing rate and high refractory period, I think it can be noticible.

h-mayorquin:
@DradeAW 
But I am not sure what I am doing is equivalent. I am generating the integers already, where a spike can be (or not) on each of the  ticks of the sampling rate. I am not converting to integers. Like, the problem would be the other way around, how good a binomial distribution approximates a poisson but the assumptions are very well held here.

> But with a high firing rate and high refractory period, I think it can be noticeable.
Yeah, maybe try to run some experiments? Maybe you are right but "noticeable" is not very actionable to me, do you expect it to fail a fitness of test? The mean would be far from the true mean? It would be more or less skwed than the exponential?

DradeAW:
>  I am generating the integers already, where a spike can be (or not) on each of the ticks of the sampling rate

I sorry I misunderstood!
Then yes, it is fine :)

h-mayorquin:
Ah, OK, still it would be useful to know if this breaks in some limit. Check it out:

image

Blue is empirical.

DradeAW:
You can easily test the extreme case by setting a refractory period at 10.0 ms and a mean firing rate at 99.99 Hz.

The expected behaviour is not that everything is at exactly 10.0ms, because of sampling accuracy

h-mayorquin:
https://github.com/SpikeInterface/spikeinterface/pull/2163

I don't understand the second line completly, but you can see that the bins look all right around the refactory period.

h-mayorquin:
The script:

```python
import numpy as np
from scipy.stats import expon


from spikeinterface.core.generate import generate_sorting


seed = 4
firing_rates = 99.9
refractory_period_ms = 10
sampling_frequency = 30000.0
sorting = generate_sorting(num_units=1, durations=[3600 * 10], sampling_frequency=sampling_frequency, seed=seed, refractory_period_ms=refractory_period_ms, firing_rates=firing_rates)
sorting.get_unit_spike_train(0, return_times=True)

# Get the analytical PDF of the exponential distribution
x = np.linspace(0, 10.0/firing_rates, 1000)
pdf = expon.pdf(x, scale=1/firing_rates)

# Plot the exponential distribution
refractory_period_seconds = refractory_period_ms / 1000.0
plt.plot(x + refractory_period_seconds, pdf, 'r-', label='Exponential Distribution')

# Plot the histogram of inter-spike intervals
times = sorting.get_unit_spike_train(0, return_times=True)
plt.hist(np.diff(times), bins=200, density=True, alpha=0.5, label='Inter-spike Intervals')

plt.xlabel('Inter-spike Interval (ms)')  # Update x-axis label
plt.ylabel('Probability Density')

# Add the refractory period and firing rate as a legend or title
legend_text = f"Refractory Period: {refractory_period_ms} ms\nFiring Rate: {firing_rates} Hz"
plt.legend(title=legend_text)

plt.axvline(refractory_period_seconds, color='k', linestyle='dashed', linewidth=1)
plt.xlim(0, 0.020)

# Update x-axis tick labels to milliseconds
ticks = np.arange(0, 0.021, 0.005)
tick_labels = [f"{int(tick * 1000)}" for tick in ticks]
plt.xticks(ticks, tick_labels)

plt.show()
```

DradeAW:
This looks wrong, can you check that what you get is at 99.99 Hz?

I'll try to use my own code to see what it looks like

h-mayorquin:
Can you elaborate on what's wrong?

DradeAW:
The standard deviation with spikes happening at 20ms interval seems wrong for a spike train with a mean firing rate of 99.99 Hz

h-mayorquin:
I cut the figure for zooming, what do you mean the standard deviation at 20 milliseconds? I am really confused.

DradeAW:
Well 99.99 Hz is basically 100 Hz, meaning an event every 10ms (in average).

So with a refractory period of 10.0ms, I don't expect any spikes happening at 20.0ms or beyond

h-mayorquin:
I think you are fooled by normality, the red is the theoretical curve, the exponential has a lot of kurtosis.

DradeAW:
Here is my code:

```python
import math
import numpy as np
import numpy.typing as npt
import plotly.graph_objects as go


def generate_spike_train(firing_rate: float, t_r: float, t_max: int, sf: float) -> npt.NDArray[np.int64]:
	"""
	firing_rate in Hz, t_r in ms, t_max in number of samples, sf in Hz
	"""

	n_expected_spikes = firing_rate * t_max / sf
	beta = 1/firing_rate - t_r*1e-3

	spike_train = np.cumsum(t_r*1e-3 + np.random.exponential(scale=beta, size=int(n_expected_spikes + 6.5*math.sqrt(n_expected_spikes))))
	end = np.searchsorted(spike_train, t_max/sf, side="left")
	assert end < len(spike_train)

	return (spike_train[:end] * sf)


spike_train = generate_spike_train(99.9, 10.0, 30*60*30_000, 30_000)
ISI = np.diff(spike_train)


fig = go.Figure().set_subplots(rows=2, cols=1)

fig.add_trace(go.Histogram(
	x=ISI / 30,
	xbins=dict(start=10.0, end=10.15, size=0.001)
), row=1, col=1)
fig.add_trace(go.Histogram(
	x=ISI.round().astype(np.int64) / 30,
	xbins=dict(start=10.0, end=10.15, size=1/30)
), row=2, col=1)

fig.show()
```

And the generated plot (on top is without sampling, below is after sampling):
newplot

As you can see, nothing at 20.0 ms :)

h-mayorquin:
I think that if you code does not follow the theoretical curve your code is wrong? You have the pdf for this process, your samples should follow it.

Also, why are you substracting the refactory period from the exponential distribution. It is the times that are shifted, not the rate.

DradeAW:
> why are you substracting the refactory period from the exponential distribution 

Well because if you don't, you generate something for 99.9 Hz, but when you add the refractory period it's no longer 99.9 Hz :)
As I told you, check the length of your generated spike train and if it matches the asked firing rate.

This is something I've worked a lot on, and checked heavily ^^

h-mayorquin:
Your math seems to be failing you, really, you sould shift the spikes in time, not modify the rate of your poisson process. 

If you do, then your curve behaves exactly like mine which is also the theoretical curve:

```python
import math
import numpy as np
import numpy.typing as npt
import plotly.graph_objects as go
from scipy.stats import expon


def generate_spike_train(firing_rate: float, t_r: float, t_max: int, sf: float) -> npt.NDArray[np.int64]:
	"""
	firing_rate in Hz, t_r in ms, t_max in number of samples, sf in Hz
	"""

	n_expected_spikes = firing_rate * t_max / sf
	beta = 1.0/firing_rate 

	spike_train = np.cumsum(t_r*1e-3 + np.random.exponential(scale=beta, size=int(n_expected_spikes + 6.5*math.sqrt(n_expected_spikes))))
	end = np.searchsorted(spike_train, t_max/sf, side="left")
	assert end < len(spike_train)

	return (spike_train[:end] * sf)


firing_rates = 99.9
sampling_frequency = 30_000
refractory_period_ms = 10.0
spike_train = generate_spike_train(firing_rates, refractory_period_ms, 30*60*30_000, sampling_frequency)



# Get the analytical PDF of the exponential distribution
x = np.linspace(0, 5.0/firing_rates, 1000)
pdf = expon.pdf(x, scale=1/firing_rates)

# Plot the exponential distribution
refractory_period_seconds = refractory_period_ms / 1000.0
plt.plot(x + refractory_period_seconds, pdf, 'r-', label='Exponential Distribution')

# Plot the histogram of inter-spike intervals
times = spike_train / sampling_frequency
plt.hist(np.diff(times), bins=200, density=True, alpha=0.5, label='Inter-spike Intervals')

plt.xlabel('Inter-spike Interval (ms)')  # Update x-axis label
plt.ylabel('Probability Density')

# Add the refractory period and firing rate as a legend or title
legend_text = f"Refractory Period: {refractory_period_ms} ms\nFiring Rate: {firing_rates} Hz"
plt.legend(title=legend_text)

plt.axvline(refractory_period_seconds, color='k', linestyle='dashed', linewidth=1)
mean = 1.0 / firing_rates
plt.axvline(mean + refractory_period_seconds, color='green', linestyle='dashed', linewidth=1, label="mean")

# # Calculate quartiles
# quartiles = np.percentile(np.diff(times), [25, 50, 75])

# # Add vertical lines for each quartile
# for quartile in quartiles:
#     plt.axvline(quartile + refractory_period_seconds, color='red', linestyle='dashed', linewidth=1)

plt.xlim(0, 5.0/firing_rates)
plt.show()
```

image

h-mayorquin:
By shifting the parameter of your exponential like that (making it smaller) you are increasing the firing rate like crazy. This concentrates the probability mass of the distribution for lower intervals and of course you don't see anything at all pass the second quartile!

DradeAW:
@h-mayorquin I think you are completely missing the point ...
If you generate an array with `beta = 1 / firing_rate`, then the array you generate will have the correct mean firing rate. But then if you add the refractory period, obviously you change the mean firing rate (and it is no longer what you asked for).

You need to correct lambda/beta by the refractory period you're going to set.

With my example above:
```
beta = 1/firing_rate - t_r*1e-3  # Leads to a mean firing rate of 99.9
beta = 1/firing_rate  # Leads to a mean firing rate of 49.980555555
```

As I stated: call `compute_firing_rate` on your sorting object and check the firing rate of the generated unit!!!

DradeAW:
> Your math seems to be failing you

Here is the mathematical proof then:

The mean of a exponential distribution is `E(X) = 1 / lambda = beta`
So `E(X + refractory_period) = E(X) + refractory_period`

Thus, `E(X) + refractory_period = 1 / firing_rate <=> E(X) = 1 / firing_rate - refractory_period`
QED

CodyCBakerPhD:
Fun fact to let people know - you can type LaTeX in GitHub now 😃

```math
\begin{align}

\mathbb{E}[X] \equiv \frac{1}{\lambda} = \beta

\end{align}
```

```math
\begin{align}

\mathbb{E}[X + c] = \mathbb{E}[X] + c = \frac{1}{\lambda} + c \neq \frac{1}{\lambda}

\end{align}
```

$\forall c > 0$

h-mayorquin:
What do you think if this reframe: Your proposal keeps the total number of spikes in time divided by time equal to the firing rate by sacrificing the instantaneous firing rate. Mine does the opposite and preserves the instantaneous firing rate as given at the expense of the count. I care about the statistics being changed, why should I change my statistics to keep the counting over longer averages like that?

h-mayorquin:
> > Your math seems to be failing you
> 
> Here is the mathematical proof then:
> 
> The mean of a exponential distribution is `E(X) = 1 / lambda = beta` So `E(X + refractory_period) = E(X) + refractory_period`
> 
> Thus, `E(X) + refractory_period = 1 / firing_rate <=> E(X) = 1 / firing_rate - refractory_period` QED

I did not mean this, I meant your expectations about how the exponential distribution should look like. But now that I understandg you are privileging the count divided by long-term time that makes sense.

DradeAW:
I really don't understand what you are doing,

We just want a unit with a given mean firing rate and refractory period, and you method just outputs a wrong firing rate.

DradeAW:
> you are privileging the count divided by long-term time that makes sense

Yes! It is the definition of the mean firing rate!!

DradeAW:
I've not read this anywhere, this is what I do by simply manipulating statistics :)

I may not be the correct answer in some cases, but in the case of firing_rate = 99.9 Hz and refractory_period = 10 ms, this version simply fails

h-mayorquin:
> I've not read this anywhere, this is what I do by simply manipulating statistics :)

x D 
Lol

I will ask Alessio and Sam what they prefer for this case next time we meet as I don't think they will dare to read this super long-thread that you and I did.

alejoe91:
I have been passively enjoyed the passionate exchange ;)

JoeZiminski:
Interesting! Could I confirm that the only outstanding point of discussion is whether to rescale $\lambda$ such that the passed firing rate to the function is the firing rate of the final spike train? Alternatively, the passed firing rate is the firing rate $\lambda$ used for the Poisson process prior to adding refractory effects. Are there any other differences between the implementations? (let me know if I have misunderstood)

On this, I am not familiar with the conventions in the field, some seem to do this rescaling [1](https://elephant.readthedocs.io/en/latest/_modules/elephant/spike_train_generation.html)[2](https://github.com/btel/python-in-neuroscience-tutorials/blob/master/poisson_process.ipynb) whereas in other sources [3](https://www.cns.nyu.edu/~david/handouts/poisson.pdf)[Dayan & Abbot] it is not mentioned.

h-mayorquin:
> Could I confirm that the only outstanding point of discussion is whether to rescale 
 such that the passed firing rate to the function is the firing rate of the final spike train? Alternatively, the passed firing rate is the firing rate  used for the Poisson process prior to adding refractory effects.
 
 @JoeZiminski this is a good summary.
 
 Thanks for the references. So Bartoz does indeed modify the rate no restrains, Andrew does modify the rate but excludes the case of the refatory period being larger than the inter spike period. Plus, Andrew does provide a reference to Stefan paper where they called this Poisson Process with dead time to get away from the fact that this is not a poisson process anymore:
 
 https://link.springer.com/article/10.1007/s10827-011-0362-8

This answers my concerns. If we decide to go for the version in the aferomentioned paper I am fine with it. Thanks, @DradeAW, I never thought carefully before on when do the assumptions of the Poisson renewal process that I learned from Abbot's treatment breaks down. I still think that the bursting model they have over there makes more sense but this seems to just be another equally valid convention.

DradeAW:
> only outstanding point of discussion is whether to rescale such that the passed firing rate to the function is the firing rate of the final spike train?

Maybe this whole conversation is just my bad for not understanding this 😅 
But as a user, if I ask for a spike train with a given firing rate, as expect the output to be that firing rate?

Sorry if I seemed a bit harsh, but criticizing my math when I spent many hours / days on the question tends to trigger me ahah

h-mayorquin:
@DradeAW 
> But as a user, if I ask for a spike train with a given firing rate, I expect the output to be that firing rate?

I think this is a very strong point. In fact, I feel convinced by it. 

The red-hering was that we started discussing about the statistics.  You came to a PR that is called "add poisson statistics" and then told me that the ISI interval should not follow the theoretical exepected distribution for a Poisson. Turns out that you have a good reason for that to be not be poisson but I instead doubled down on "this is a poisson distribution, why do you want me to make my statistics not poissonian, you must be getting something wrong!".  At the end I think we came to the right [crux](https://www.lesswrong.com/tag/double-crux) and we should take pride on that:

I asked you:
>  Your proposal keeps the total number of spikes in time divided by time equal to the firing rate by sacrificing the instantaneous firing rate. Mine does the opposite and preserves the instantaneous firing rate as given at the expense of the count. I care about the statistics being changed, why should I change my statistics to keep the counting over longer averages like that

And I think that you have a great answer for my question there which is:
> But as a user, if I ask for a spike train with a given firing rate, I expect the output to be that firing rate?

I do believe that user-centered decisions trumps mathematical soudness so that argment convinces me.

And looking in retrospective the discussion went to the distribution because I failed to read the following. You said:
> This looks wrong, can you check that what you get is at 99.99 Hz?

I actually did not understand that you were asking me to measure the firing rate there (probably the use of the proposition "at" and the omission of the word firing rate, writing is hard). And then I started discussing about the pdf of the distribution.

But overall the discussion was good to me. I learned something new and I think making this a Poisson distribution with dead times (as Stefan and co call it) is better for maintenance and usability. So thanks for bearing with me, I think the library is better thanks to your efforts.

DradeAW:
Alright, everything turned out well ^^

Sorry I know I'm not always super great at expressing stuff (even in my native language :sweat_smile:), but we both learned something!

Although I'm still curious about something: your implementation is still Poissonian?
I thought (but I might be wrong about this) that a Poisson distribution with a refractory period was not a true Poissonian distribution?
Genuinely asking!

h-mayorquin:
Not strictly, the support is changed (by the shifting) but the ISI is still an exponential and the mean and the std ratio should remain constant in the new support.

That said, I am not sure that is not the case in the poisson process with dead time. I tried to fit to the exponential with a shift and that did not work but maybe the paraemters need to be changed in some other way.

I pushed a version of your suggestion now but the tests are failing. It seems that those are the new metrics. I will take a look later.

src/spikeinterface/qualitymetrics/tests/test_metrics_functions.py
```diff
@@ -384,7 +384,7 @@ def test_calculate_sd_ratio(waveform_extractor_simple):
     sd_ratio = compute_sd_ratio(waveform_extractor_simple)
 
     assert np.all(list(sd_ratio.keys()) == waveform_extractor_simple.unit_ids)
-    assert np.allclose(list(sd_ratio.values()), 1, atol=0.2, rtol=0)
+    assert np.allclose(list(sd_ratio.values()), 1, atol=0.25, rtol=0)
```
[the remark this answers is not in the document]
DradeAW: No it's not meaningful, I think it's due to the recording being relatively short (few spikes). You can increase the tolerance :)

> I don't think you need the list over there, righ?

Maybe not! I'm never sure if the conversion is necessary or not so I tend to always put it because I'm too lazy to check ^^'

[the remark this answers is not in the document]
h-mayorquin: OK, thanks!

h-mayorquin:
And this are some related plots that I did after our discussion. 

## Discretization
This should answer @DradeAW question about how the discretization behaves in the extreme case:
image

The bins accumulate in `1/sampling_rate` because that the smallest time represented by the sorting.

## How much does the method actually modifies the firing rates
For myself I was curios on how much is the instantaneous firing rate modified depending on the refactory period:
image
(note the color map is cut at 10 but it goes very large)

image

image

So, yes, the lesson is quite a lot, for high firing rates a small refactory period can easily double the firing rate.

src/spikeinterface/core/generate.py
```diff
@@ -284,6 +285,114 @@ def generate_snippets(
...
+        raise ValueError(
+            f"The given refractory period {refractory_period_ms} is too long for the firing rates {firing_rates}"
+        )
+
+    # p is the probably of an spike per tick of the sampling frequency
+    binomial_p = firing_rates / sampling_frequency
+    # We estimate how many spikes we will have in the duration
+    max_frames = duration * sampling_frequency
+    max_binomial_p = float(np.max(binomial_p))
+    num_spikes_expected = ceil(max_frames * max_binomial_p)
```
zm711, line 364:
```suggestion
    num_spikes_expected = int(np.ceil(max_frames * max_binomial_p))
```

Any interest in this instead and then you can remove the ceil import from math? Or did you really only want to use math.ceil?

h-mayorquin: What's the advantage of this? 

Last time I checked the math module functions are faster for scalars than numpy function as they avoid the overhead. Speed won't matter that much at this scale though.

zm711: Honestly, the only advantage for this scalar is that you import one less function into the code that is only used once. But reducing imports is not necessarily a good reason. So my comment was more question than hard recommendation.

h-mayorquin: @zm711 I see. Yes,  importing from the standard library at will is my prior until proven otherwise. 

Run the following script:
```python
import pkgutil
import timeit
import sys 

# Get a list of all standard library modules
standard_lib_modules = [module for module in pkgutil.iter_modules() if module.name in sys.stdlib_module_names]

# Dictionary to store import times
import_times = {}

for module in standard_lib_modules:
    # Measure the import time
    time = timeit.timeit(f"import {module.name}", number=1)
    import_times[module.name] = time

# Print or process the import times
for module, time in import_times.items():
    print(f"{module}: {time} seconds")
``` 

You will see that importing from the standard library is at the scale of main memory reference:
https://brenocon.com/dean_perf.html

src/spikeinterface/qualitymetrics/tests/test_metrics_functions.py
```diff
@@ -395,7 +395,7 @@ def test_calculate_sd_ratio(waveform_extractor_simple):
     )
 
     assert np.all(list(sd_ratio.keys()) == waveform_extractor_simple.unit_ids)
-    # assert np.allclose(list(sd_ratio.values()), 1, atol=0.2, rtol=0)
+    assert np.allclose(sd_ratio.values(), 1, atol=0.25, rtol=0)
```
h-mayorquin, line 398:
```suggestion
    assert np.allclose(list(sd_ratio.values()), 1, atol=0.25, rtol=0)
```

src/spikeinterface/core/generate.py
zm711: Thanks @h-mayorquin! Makes sense.

samuelgarcia:
Hi Ramon and Aurelien.
Thanks a lot for this PR and this discusssion.
This is really an excelent discussion and piece of work.

I have to admit (with lot of shame) that I did had time to read it before today.
So lets merge this now.

samuelgarcia (approved):

PR #2227, opened 2023-11-18
Add spike-train based lazy SortingGenerator

h-mayorquin:
While discussing in https://github.com/SpikeInterface/spikeinterface/pull/2175 and https://github.com/SpikeInterface/spikeinterface/pull/2209 and the other PR that is now I realized that we don't have a way of tesing lazy recorders performance.

Currently, `generate_sorting` alredy pre-loads the data as a numpy array which means that we can't test how performancer or other IO operations would behave in different scenarios.

This PR aims to fix that by introduces a new sorter that generates spikes on the fly in a reproducible way. Plus, it behaves consistency across calls and by slicing across spikes.

The max memory it should use is the complete size of its longest spike train.


src/spikeinterface/core/generate.py
```diff
@@ -566,6 +566,116 @@ def synthetize_spike_train_bad_isi(duration, baseline_rate, num_violations, viol
...
+    def __init__(
+        self,
+        num_units: int = 5,
+        sampling_frequency: float = 30000.0,  # in Hz
+        durations: List[float] = [10.325, 3.5],  #  in s for 2 segments
+        firing_rates: float | np.ndarray = 3.0,
+        refractory_period_ms: float | np.ndarray = 4.0,  # in ms
+        seed: Optional[int] = None,
+    ):
+        unit_ids = np.arange(num_units)
```
zm711, line 582: Will we have a docstring for this? For example you have durations as a list of 2 floats for 2 segments, so for future users it might be beneficial for them to know that the len(durations)==n_segments. And what are the implications of giving one firing rate vs an array of firing rates if I as a user try something like this?

h-mayorquin: Yeah, we should. The duration thing is a very pervasive in a bunch of functions we should document it. Less pevasive but not exclusive to this one is the firing rates as an array or float.

alejoe91: @h-mayorquin can you add a docstring here? Maybe explaining the difference with the `generate_sorting`

h-mayorquin: Added.

h-mayorquin: I also corrected a typo of refactory to refractory

h-mayorquin:
Hey guys @alejoe91 @samuelgarcia . I want to do some performance testing and I would like to compare dictionary vs the numpy approach. What do we need to move this forward?

alejoe91:
@h-mayorquin this is ok to merge for me!! It would be great for perfomrmance testing :)

alejoe91 (approved):

src/spikeinterface/core/generate.py
```diff
@@ -1054,6 +1054,176 @@ def synthetize_spike_train_bad_isi(duration, baseline_rate, num_violations, viol
...
+        refractory_period_frames = int(refractory_period * self.sampling_frequency)
+        spike_frames[1:] += refractory_period_frames
+
+        if start_frame is not None:
+            start_index = np.searchsorted(spike_frames, start_frame, side="left")
+        else:
+            start_index = 0
+
+        if end_frame is not None:
+            end_index = np.searchsorted(spike_frames[start_index:], end_frame, side="right")
```
samuelgarcia, line 1219: I think we are using left anywhere else in the package.
In my mind if a spike is exactly end_index then it should not be included in the output
```suggestion
            end_index = np.searchsorted(spike_frames[start_index:], end_frame, side="left")
```

samuelgarcia:
Thank you Ramon, every thing is perfect except a very small detail.
10 month before a merge is good perf. sorry.

alejoe91 (approved):

PR #2228, opened 2023-11-18
Remove legacy mode for `generate_recording`

h-mayorquin:
The time has come.

There is nothing like the joy of deprecation and getting rid of old code.

image
Courtesy of Dall-e

### DIFF
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -35,7 +35,7 @@ def generate_recording(
 ) -> BaseRecording:
     """
     Generate a lazy recording object.
-    Useful for testing for testing API and algos.
+    Useful for testing API and algos.
 
     Parameters
     ----------
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -48,7 +48,7 @@ def generate_recording(
         Note that the number of segments is determined by the length of this list.
     set_probe: bool, default: True
     ndim : int, default: 2
-        The number of dimensions of the probe, default: 2. Set to 3 to make 3 dimensional probes.
+        The number of dimensions of the probe, default: 2. Set to 3 to make 3 dimensional probe.
     seed : Optional[int]
         A seed for the np.ramdom.default_rng function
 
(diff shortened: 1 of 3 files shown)


zm711:
Someone should really explain to Dall-e that it is really bad for the spines of books to lay flat like that.

alejoe91 (approved):

DradeAW:
Here you're not deprecating but removing the function, or have I misunderstood?

alejoe91:
it was already called `legacy`, so it was kind of born deprecated

DradeAW:
I'm talking about the title of the PR, which sounds ambiguous to me.

I might be mistaking, but I thought

- Deprecating = We advise not to use, we will not support / drop this later
- Removing = You can't use it anymore

h-mayorquin:
@DradeAW You are correct. It should be removing. I actually read a discussiong today in the python forums about it, mea culpa.

samuelgarcia:
I am OK with this but I think we decided to keep 2 versions after a deprecation no ?
here it is only one.

h-mayorquin:
@samuelgarcia no, we added a removing version at that point in time. Maybe you are confusing this with the modes of the `GeneratorRecording`? This is about the in-memory version of `generate_recording` the utility function.

src/spikeinterface/core/generate.py
```diff
@@ -32,10 +32,9 @@ def generate_recording(
     set_probe: Optional[bool] = True,
     ndim: Optional[int] = 2,
     seed: Optional[int] = None,
-    mode: Literal["lazy", "legacy"] = "lazy",
 ) -> BaseRecording:
     """
-    Generate a recording object.
+    Generate a lazy recording object.
     Useful for testing for testing API and algos.
```
zm711, line 38:
```suggestion
    Useful for testing API and algos.
```
When one testing is not enough...

src/spikeinterface/core/generate.py
```diff
@@ -52,10 +51,6 @@ def generate_recording(
         The number of dimensions of the probe, default: 2. Set to 3 to make 3 dimensional probes.
```
zm711, line 51:
```suggestion
        The number of dimensions of the probe, default: 2. Set to 3 to make 3 dimensional probe.
```

zm711 (commented):
Since we are here two typos in the docstring that can be fixed.

alejoe91 (approved):

PR #2238, opened 2023-11-22
Fix typo: `_serializablility` -> `_serializability`

alejoe91:


src/spikeinterface/comparison/multicomparisons.py
```diff
@@ -189,7 +189,7 @@ def save_to_folder(self, save_folder):
             stacklevel=2,
         )
         for sorting in self.object_list:
-            assert sorting.check_serializablility(
+            assert sorting.check_serializability(
                 "json"
             ), "MultiSortingComparison.save_to_folder() need json serializable sortings"
```
zm711, line 194:
```suggestion
            ), "MultiSortingComparison.save_to_folder() needs json serializable sortings"
```

zm711 (commented):
just one small assert typo that could be fixed while this is being edited.

samuelgarcia:
very strange that Zach haven't catch this before....

h-mayorquin:
> very strange that Zach haven't catch this before....

I think this is a testament to the difficult spelling of the word : P

zm711:
Haha. It is a very difficult word to spell. But what I'm hearing is that I should turn the proofreading up to 10 and really find every single mistake :P 

Also I tend not to edit the private kwargs etc as much just because I don't know where they all go and since I can't run the full test suite locally it is trickier to make those kinds of changes (like the `has_extension` that Aurelien fixed), but now that I know it's fair game....

h-mayorquin (approved):

PR #2239, opened 2023-11-22
Use context managers on Deep Interpolation

h-mayorquin:
Following on https://github.com/SpikeInterface/spikeinterface/pull/2195

### DIFF
--- a/src/spikeinterface/extractors/mdaextractors.py
+++ b/src/spikeinterface/extractors/mdaextractors.py
@@ -442,18 +442,21 @@ def is_url(path):
 
 
 def _download_bytes_to_tmpfile(url, start, end):
-    try:
-        import requests
-    except:
-        raise Exception("Unable to import module: requests")
-    headers = {"Range": "bytes={}-{}".format(start, end - 1)}
-    r = requests.get(url, headers=headers, stream=True)
-    fd, tmp_fname = tempfile.mkstemp()
-    os.close(fd)
-    with open(tmp_fname, "wb") as f:
-        for chunk in r.iter_content(chunk_size=1024):
-            if chunk:
-                f.write(chunk)
+    import requests
+
+    headers = {"Range": f"bytes={start}-{end - 1}"}
+
+    with requests.get(url, headers=headers, stream=True) as r:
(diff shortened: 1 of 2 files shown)


src/spikeinterface/extractors/mdaextractors.py
```diff
@@ -442,18 +442,21 @@ def is_url(path):
...
-    os.close(fd)
-    with open(tmp_fname, "wb") as f:
-        for chunk in r.iter_content(chunk_size=1024):
-            if chunk:
-                f.write(chunk)
+    import requests
+
+    headers = {"Range": f"bytes={start}-{end - 1}"}
+
+    with requests.get(url, headers=headers, stream=True) as r:
```
h-mayorquin, line 449: Double context because more is clearly better.

src/spikeinterface/preprocessing/deepinterpolation/generators.py
```diff
@@ -86,11 +86,11 @@ def __init__(
         sequential_generator_params["total_samples"] = self.total_samples
         sequential_generator_params["pre_post_omission"] = pre_post_omission
 
-        json_fd, json_path = tempfile.mkstemp(suffix=".json")
-        os.close(json_fd)
-        with open(json_path, "w") as f:
+        with tempfile.NamedTemporaryFile(suffix=".json") as f:
             json.dump(sequential_generator_params, f)
```
alejoe91, line 90: This seems to be failing...

h-mayorquin: New error I have not idea. Any suggestions? It seems is not finding the training weights?

alejoe91: It looks like it's the same problem no? Still failing at dump (check the first failure, all later ones are dependant on it)

h-mayorquin: Yeah, you are right, I was confused by the rest of the stack trace. It was just old binary mode vs text mode in the context manager.

alejoe91 (approved):

PR #2240, opened 2023-11-22
Remove `run_sorters` for version 0.100.0

zm711:
Not sure if you guys were thinking of releasing a 99.2, but `run_sorters` is suppose to be removed in 100 and so it should be time to fully remove this code if you're not planning on more 99 releases. 

I was also thinking as this code base gains traction it might be beneficial to have a small deprecation/removal section in the release notes since things like: `plot_timeseries` have tripped people up and since `run_sorters` will be a bigger removal as well.


zm711:
Let me fix the imports as well.

alejoe91 (approved):

PR #2169, opened 2023-11-04
Support remfile and file-like in nwb rec extractor

magland:
Adds "remfile" stream mode option for NwbRecordingExtractor.

Also allows passing in a file-like object. This is important for dendro for embargoed datasets because we need to pass in a file-like object that is capable of renewing its remote url periodically as the presigned download url expires.

In terms of parameter order, I put the new parameter "file" after electrical_series_name because somebody may be calling this without naming the arguments (since * is not used in the constructor definition).

This is untested (are there unit tests for nwb extractor?)

@alejoe91 @samuelgarcia @luiztauffer

PR #2246, opened 2023-11-23
Add option for no caching option to the `NWBRecordingExtractor` when streaming

h-mayorquin:
Caching is option in fsspec. Sometimes this will be desired by the users and sometimes this will allow us to avoid some errors as in:

https://github.com/SpikeInterface/spikeinterface/pull/2169#issuecomment-1824418183

I am enabling this as an option for the Recording Extractor. I will do the same soon for the sorting but that requires some more unification.


alejoe91 (approved):

alejoe91:
Same here :)

PR #2247, opened 2023-11-23
Fix `rp_violations` when specifying `unit_ids`

DradeAW:
The index `i` in the for loop is incorrect and leads to wrong contamination in the dict when specifying some unit ids,

This PR fixes it :)


src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -389,7 +389,10 @@ def compute_refrac_period_violations(
     nb_violations = {}
     rp_contamination = {}
 
-    for i, unit_id in enumerate(unit_ids):
+    for i, unit_id in enumerate(sorting.unit_ids):
```
alejoe91, line 392: What about computing nb_rp_violations only on `unit_ids` and not on `sorting.unit_ids` and have a
```
if unit_ids is None:
    unit_ids = sorting.unit_ids
```
?

DradeAW: I thought about it, but then this requires doing `sorting.select_units(unit_ids).to_spike_vector()`

And this means recomputing the spike vector, which I worry might be slower.
We could maybe do `sorting.to_spike_vector()` and then remove any indices not in unit_ids?

alejoe91: Makes sense. The most elegant option would be to override the `UnitsSelectionSorting` so that if the parent is cached, it slices it and then go for the `sorting.select_units(unit_ids).to_spike_vector()`.

I'm also fine with this "simpler" implementation. Let me know what you prefer!

DradeAW: I think overriding would be the best option, but every time I mention overriding a method, @samuelgarcia doesn't like it ^^'

zm711:
@DradeAW, since you're editing this could you fix the docstring. Currently it has one of the arguments with `ùs` instead of `ms` 🙏 (it's L337 with the argument: `censored_period_ùs`)

alejoe91 (approved):

alejoe91:
@DradeAW let's keep it this way for now. It serves its purpose :)

src/spikeinterface/qualitymetrics/misc_metrics.py
samuelgarcia: Bad idea.
And also this will not as easy because you will need to move all the unit_index to the new_index!

PR #2253, opened 2023-11-23
Move `comparison.hybrid.generate_injected_sorting` to `core.generate`

### LINKED ISSUE #1976: Move `generate_injected_sorting` to `generate.py` and control seed
The [generate_injected_sorting](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/comparison/hybrid.py#L202) should be moved to the `core.generate` module and we should add the `seed` option.

alejoe91:
Fixes #1976

### DIFF
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -177,6 +177,65 @@ def generate_sorting(
     return sorting
 
 
+def generate_injected_sorting(
+    sorting: BaseSorting,
+    num_samples: List[int],
+    max_injected_per_unit: int = 1000,
+    injected_rate: float = 0.05,
+    refractory_period_ms: float = 1.5,
+) -> NumpySorting:
+    """
+    Generates a sorting object with injected spikes from an existing sorting object.
+
+    Parameters
+    ----------
+    sorting : BaseSorting
+        The sorting object
+    num_samples : list
+        The number of samples in each segment
+    max_injected_per_unit : int, default: 1000
+        The maximum number of injected spikes per unit
+    injected_rate : float, default: 0.05
+        The rate of injected spikes with respect to the total number of spikes
+    refractory_period_ms : float, default: 1.5
+        The refractory period in ms
+
+    Returns
+    -------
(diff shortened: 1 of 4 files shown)


src/spikeinterface/core/generate.py
```diff
@@ -177,6 +177,65 @@ def generate_sorting(
...
+    """
+    injected_spike_trains = [{} for seg_index in range(sorting.get_num_segments())]
+    t_r = int(round(refractory_period_ms * sorting.get_sampling_frequency() * 1e-3))
+
+    for segment_index in range(sorting.get_num_segments()):
+        for unit_id in sorting.unit_ids:
+            spike_train = sorting.get_unit_spike_train(unit_id, segment_index=segment_index)
+            n_injection = min(max_injected_per_unit, int(round(injected_rate * len(spike_train))))
+            # Inject more, then take out all that violate the refractory period.
+            n = int(n_injection + 10 * np.sqrt(n_injection))
```
h-mayorquin, line 216: I think from the issue we are missing the seed, right?

src/spikeinterface/core/generate.py
```diff
@@ -177,6 +177,65 @@ def generate_sorting(
...
 
+def generate_injected_sorting(
+    sorting: BaseSorting,
+    num_samples: List[int],
+    max_injected_per_unit: int = 1000,
+    injected_rate: float = 0.05,
+    refractory_period_ms: float = 1.5,
+) -> NumpySorting:
+    """
+    Generates a sorting object with injected spikes from an existing sorting object.
```
h-mayorquin, line 188: Can we take this cahnce to improve this docstring a little bit. I don't really understand what this object does.

Is it a sorting object that has spikes sampled from another sorting?

h-mayorquin (changes requested):
This looks OK, to me. It is just moving as in the issue. Just let's add a sed as in the issue.

The docstring is a bonus but not necessary for me, we can do later.

alejoe91:
Closing in favor of #1999

PR #2254, opened 2023-11-23
Implement save_to_zarr for BaseSorting

### LINKED ISSUE #2231: Add zarr backend for sorting object
Currently, both `Recording` and `WaveformExtractor` have a zarr backend, but not the `Sorting` object

alejoe91:
Closes #2231


src/spikeinterface/core/zarrextractors.py
```diff
@@ -155,7 +157,138 @@ def get_traces(
...
+        self._unit_ids = list(unit_ids)
+
+    def get_unit_spike_train(
+        self,
+        unit_id,
+        start_frame: Union[int, None] = None,
+        end_frame: Union[int, None] = None,
+    ) -> np.ndarray:
+        sample_indices = self._spikes_dset["sample_index"][self._segment_slice][start_frame:end_frame]
+        unit_indices = self._spikes_dset["unit_index"][self._segment_slice][start_frame:end_frame]
```
samuelgarcia, line 249: To handle start_frame/end_frame you need a searchsorted!!

src/spikeinterface/core/basesorting.py
```diff
@@ -240,6 +240,44 @@ def _save(self, format="numpy_folder", **save_kwargs):
                 warnings.warn("The registered recording will not be persistent on disk, but only available in memory")
                 cached.register_recording(self._recording)
 
+        elif format == "zarr":
```
samuelgarcia, line 243: Lets do `ZarSortingExtrcator.write()` instead having the logic here.

src/spikeinterface/core/zarrextractors.py
```diff
@@ -0,0 +1,424 @@
...
+    """
+
+    extractor_name = "ZarrSorting"
+    installed = HAVE_ZARR  # check at class level if installed or not
+    mode = "file"
+    # error message when not installed
+    installation_mesg = "To use the ZarrSortingExtractor install zarr: \n\n pip install zarr\n\n"
+    name = "zarr"
+
+    def __init__(self, root_path: Union[Path, str], storage_options: dict | None = None):
```
h-mayorquin, line 242: What do you think of unify and just call everything a `file_path` or `folder_path`. That way, if we want to use keyword arguments we don't need to remember specific keywords for every different format. Does that makes sense here?

alejoe91: Yep, nothing against it! Refactoring :)

src/spikeinterface/core/testing.py
```diff
@@ -88,6 +90,19 @@ def check_sortings_equal(
             train1 = np.sort(SX1.get_unit_spike_train(id, segment_index=segment_idx))
             train2 = np.sort(SX2.get_unit_spike_train(id, segment_index=segment_idx))
             assert np.array_equal(train1, train2)
+            train1 = np.sort(SX1.get_unit_spike_train(id, segment_index=segment_idx, start_frame=30))
```
h-mayorquin, line 93: Can you add a comment on why those were included so the future person who sees this can make sense of this? As in, "it is important to tests this because ... ' 


 Maybe testing for these properties could be a different function with a meaningful name that is called from within `check_sortings_equal` or from somewhere else.

h-mayorquin (commented):
Any chance we can add some pickling test somewhere?

src/spikeinterface/core/zarrextractors.py
```diff
@@ -0,0 +1,424 @@
...
+        assert "spikes" in self._root.keys(), "'spikes' dataset not found!"
+        spikes_group = self._root["spikes"]
+        segment_slices_list = spikes_group["segment_slices"][:]
+        segment_slices = [slice(s[0], s[1]) for s in segment_slices_list]
+
+        BaseSorting.__init__(self, sampling_frequency, unit_ids)
+
+        for segment_index in range(num_segments):
+            soring_segment = ZarrSortingSegment(spikes_group, segment_slices[segment_index], unit_ids)
+            self.add_sorting_segment(soring_segment)
```
alejoe91, line 276: Another option is to create the `self._cached_spike_vector` and then use the `NumpySortingSegment` (maybe rename as `SpikeVectorSortingSegment`, in a different PR)

src/spikeinterface/core/zarrextractors.py
```diff
@@ -0,0 +1,424 @@
...
+
+        # load annotations
+        annotations = self._root.attrs.get("annotations", None)
+        if annotations is not None:
+            self.annotate(**annotations)
+
+        self._kwargs = {"root_path": root_path_kwarg, "storage_options": storage_options}
+
+    @staticmethod
+    def write_sorting(sorting: BaseSorting, zarr_path: Union[str, Path], storage_options: dict | None = None, **kwargs):
```
alejoe91, line 293: Add `write_sorting_to_group`/`read_sorting_from_group` to interface with opened zarr objects

alejoe91: done

src/spikeinterface/core/testing.py
alejoe91: done

src/spikeinterface/core/base.py
```diff
@@ -1006,38 +995,18 @@ def save_to_zarr(
...
-            zarr_root.attrs["provenance"] = check_json(self.to_dict())
-        else:
-            zarr_root.attrs["provenance"] = None
-
-        # save data (done the subclass)
-        save_kwargs["zarr_root"] = zarr_root
         save_kwargs["zarr_path"] = zarr_path
         save_kwargs["storage_options"] = storage_options
         save_kwargs["channel_chunk_size"] = channel_chunk_size
         cached = self._save(verbose=verbose, **save_kwargs)
```
samuelgarcia, line 1009: Should be:
```suggestion
        cached = self._save(format="zarr", verbose=verbose, **save_kwargs)
```
no ?

src/spikeinterface/core/basesorting.py
```diff
@@ -240,6 +240,18 @@ def _save(self, format="numpy_folder", **save_kwargs):
...
+            from .zarrextractors import ZarrSortingExtractor
+
+            zarr_path = save_kwargs.pop("zarr_path")
+            storage_options = save_kwargs.pop("storage_options")
+            ZarrSortingExtractor.write_sorting(self, zarr_path, storage_options, **save_kwargs)
+            cached = ZarrSortingExtractor(zarr_path, storage_options)
+
+            if self.has_recording():
+                warnings.warn("The registered recording will not be persistent on disk, but only available in memory")
+                cached.register_recording(self._recording)
```
samuelgarcia, line 253: I am OK for this now. But We will need to find a way for this when we design the SortingResult object.

src/spikeinterface/core/zarrextractors.py
```diff
@@ -0,0 +1,558 @@
...
+from .basesorting import BaseSorting, BaseSortingSegment
+from .core_tools import define_function_from_class, check_json
+from .job_tools import split_job_kwargs
+
+try:
+    import zarr
+
+    HAVE_ZARR = True
+except ImportError:
+    HAVE_ZARR = False
```
samuelgarcia, line 18: Given the fact that:
 * zarr will take a more important place in the spikeinterface with the sortingresult object
 * zarr is small and well maintained
 * zarr have frequent released
we could add zarr in hard dependencies one day no ?

src/spikeinterface/core/zarrextractors.py
```diff
@@ -0,0 +1,558 @@
...
+    Parameters
+    ----------
+    sorting: BaseSorting
+        The sorting extractor object to be added to the zarr group
+    zarr_group: zarr.hierarchy.Group
+        The zarr group
+    kwargs: dict
+        Other arguments passed to the zarr compressor
+    """
+    from numcodecs import Delta
```
samuelgarcia, line 353: Does this delta encoding will be a definitive choice ?
I guess that it has better compression no ?
But if one day we want to change this choice then we have a different version of our format.
For instance does this delta is also good for spike unit_index ?

src/spikeinterface/core/zarrextractors.py
samuelgarcia: I think it is relevant to do this in this PR.
The actual ZarrSortingSegment will be slow.

src/spikeinterface/core/base.py
alejoe91: I think it's already in the `save_kwargs`

src/spikeinterface/core/basesorting.py
alejoe91: later

src/spikeinterface/core/zarrextractors.py
alejoe91: agree, adding to core

src/spikeinterface/core/zarrextractors.py
alejoe91: it's best for spike samples because it just takes the diff between consecutive samples before standard compression.

src/spikeinterface/core/zarrextractors.py
alejoe91: done

src/spikeinterface/core/zarrextractors.py
alejoe91: done

src/spikeinterface/core/zarrextractors.py
alejoe91: done

samuelgarcia (approved):

samuelgarcia (approved):

PR #2271, opened 2023-11-29
run_sorter in container check json or pickle

### LINKED ISSUE #2261: AttributeError when setting docker=True in spikeinterface.sorters.run_sorter
Hello, I get a consistent error when I run the `run_sorter` function with `docker=True`. I have tried this with a number of sorters such as `'ironclust'`, `'kilosort3'`, `'mountainsort5'`, `'herdingspikes'` and I always get the same error. The same code works with local sorters such as `'spykingcircus2'` and `'tridesclous2'`. This is my first time using docker and docker desktop so this may not be a bug; the images seem to have pulled just fine though.

**1)** OS, IDE: windows 11, VS code.

**2)** spikeinterface version:  I have installed spikeinterface using the [windows conda yml file](https://github.com/SpikeInterface/spikeinterface/blob/main/installation_tips/full_spikeinterface_environment_windows.yml). Here is the full list of my environment packages [env.txt](https://github.com/SpikeInterface/spikeinterface/files/13465583/env.txt).

**3)** Python code:
```python
import spikeinterface.sorters as ss

test_recording, _ = se.toy_example(
    duration=30,
    seed=0,
    num_channels=32,
    num_segments=1,
    firing_rate=10,
    num_units=10,
)

sorter = 'kilosort3'
sorting = ss.run_sorter(sorter_name=sorter,
    recording=test_recording,
    output_folder=sorter+"_output",
    verbose=True,
    docker_image=True)
```

**4)** Error:

```python
File ...\site-packages\spikeinterface\sorters\runsorter.py:141, in run_sorter(sorter_name, recording, output_folder, remove_existing_folder, delete_output_folder, verbose, raise_error, docker_image, singularity_image, delete_container_files, with_output, **sorter_params)
    139         else:
    140             container_image = singularity_image
--> 141     return run_sorter_container(
    142         container_image=container_image,
    143         mode=mode,
...
    extractor = extractor_class(**new_kwargs)
File "/root/.local/lib/python3.9/site-packages/spikeinterface/core/numpyextractors.py", line 137, in __init__
    if spikes.size == 0:
AttributeError: 'list' object has no attribute 'size'
```

samuelgarcia:
run_sorter in container check json or pickle.

This close #2261.

### DIFF
--- a/src/spikeinterface/sorters/runsorter.py
+++ b/src/spikeinterface/sorters/runsorter.py
@@ -422,7 +422,7 @@ def run_sorter_container(
     elif recording.check_serializability("pickle"):
         (parent_folder / "in_container_recording.pickle").write_bytes(pickle.dumps(rec_dict))
     else:
-        raise RuntimeError("To use run_sorter with container the recording must serializable")
+        raise RuntimeError("To use run_sorter with container the recording must be serializable")
 
     # need to share specific parameters
     (parent_folder / "in_container_params.json").write_text(
--- a/src/spikeinterface/sorters/runsorter.py
+++ b/src/spikeinterface/sorters/runsorter.py
@@ -2,6 +2,7 @@ import shutil
 import os
 from pathlib import Path
 import json
+import pickle
 import platform
 from warnings import warn
 from typing import Optional, Union
@@ -414,9 +415,15 @@ def run_sorter_container(
 
     # create 3 files for communication with container
     # recording dict inside
-    (parent_folder / "in_container_recording.json").write_text(
-        json.dumps(check_json(rec_dict), indent=4), encoding="utf8"
-    )
+    if recording.check_serializability("json"):
+        (parent_folder / "in_container_recording.json").write_text(
+            json.dumps(check_json(rec_dict), indent=4), encoding="utf8"
+        )
+    elif recording.check_serializability("pickle"):
+        (parent_folder / "in_container_recording.pickle").write_bytes(pickle.dumps(rec_dict))
+    else:
+        raise RuntimeError("To use run_sorter with container the recording must serializable")
+
     # need to share specific parameters
     (parent_folder / "in_container_params.json").write_text(
         json.dumps(check_json(sorter_params), indent=4), encoding="utf8"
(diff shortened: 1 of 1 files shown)


src/spikeinterface/sorters/runsorter.py
```diff
@@ -414,9 +415,15 @@ def run_sorter_container(
...
-        json.dumps(check_json(rec_dict), indent=4), encoding="utf8"
-    )
+    if recording.check_serializability("json"):
+        (parent_folder / "in_container_recording.json").write_text(
+            json.dumps(check_json(rec_dict), indent=4), encoding="utf8"
+        )
+    elif recording.check_serializability("pickle"):
+        (parent_folder / "in_container_recording.pickle").write_bytes(pickle.dumps(rec_dict))
+    else:
+        raise RuntimeError("To use run_sorter with container the recording must serializable")
```
h-mayorquin, line 425:
```suggestion
        raise RuntimeError("To use run_sorter with container the recording must be serializable")
```

src/spikeinterface/sorters/runsorter.py
```diff
@@ -414,9 +415,15 @@ def run_sorter_container(
 
     # create 3 files for communication with container
     # recording dict inside
-    (parent_folder / "in_container_recording.json").write_text(
-        json.dumps(check_json(rec_dict), indent=4), encoding="utf8"
-    )
+    if recording.check_serializability("json"):
```
h-mayorquin, line 418: Not that it matters but there is a reason you are not using `recording.dump()` directly?

src/spikeinterface/sorters/runsorter.py
```diff
@@ -593,7 +606,10 @@ def run_sorter_container(
 
     # clean useless files
     if delete_container_files:
-        os.remove(parent_folder / "in_container_recording.json")
+        if (parent_folder / "in_container_recording.json").exists():
+            os.remove(parent_folder / "in_container_recording.json")
```
h-mayorquin, line 610: Good that this happens inside the container so we don't have to worry about windows immediate delete error.

h-mayorquin (approved):
This looks good to me. It will make testing recording easier.

In fact, given that this dumped recorders are erased at the end would it not be better to just dump them to pickle all the time for the performance gain?

On a second thought, probably not good to mess with was is mostly working without profiling or a good reason.

src/spikeinterface/sorters/runsorter.py
alejoe91: @h-mayorquin this is actually happening on the host machine! So I'm in favor of the additional protection :)

alejoe91 (approved):

alejoe91:
@samuelgarcia this looks good to me!

src/spikeinterface/sorters/runsorter.py
samuelgarcia: @h-mayorquin also want to change this but not in this PR because it imply some path handling and relative_to refactroing.

h-mayorquin: Yes, that makes sense. I was not suggesting that we should do that here. I was asking questions to my benefit and understanding.

This is what I want to convey when I approve a PR but then add comments: "This is good and I think is fine as it is  but what about this or that now that we are on the topic"

In this case it was mere curiosity. Thanks for engaging with me guys.

PR #2276, opened 2023-11-30
Circus2 params + various enhancements

yger:
Enhance the documentation for circus 2, but not really sure how much should be written in the _params_description. I'll create a circus 2 specific documentation in the future. In addition, adds the possibility to only export multi units activity quickly (avoiding clustering+template matching), and to chose the matching engine properly


src/spikeinterface/sorters/internal/spyking_circus2.py
```diff
@@ -30,17 +30,49 @@ class Spykingcircus2Sorter(ComponentsBasedSorter):
...
+        "multi_units_only": "Boolean to get only multi units activity (i.e. one template per electrode)",
+        "job_kwargs": "A dictionnary to specify how many jobs and which parameters they should used",
+        "debug": "Boolean to specify if internal datastructure should be kept for debugging",
+    }
+
+    sorter_description = """Spyking Circus 2 is a rewritting of Spyking Circus, within the SpikeInterface framework
+    It uses a more conservative clustering (compared to Spyking Circus) less prone to hallucinate units and/or find noise.
+    In addition, it also uses a full Orthogonal Matching Pursuit engine to reconstruct the traces, leading to more spikes
+    that can be discovered."""
+
```
zm711, line 75:
```suggestion
    _params_description = {
        "general": "A dictionary to describe how templates should be computed. User can define ms_before and ms_after (in ms) \
                                        and also the radius_um used to be considered during clustering",
        "waveforms": "A dictionary to be passed to all the calls to extract_waveforms that will be performed internally. Default is \
                                        to consider sparse waveforms",
        "filtering": "A dictionary for the high_pass filter to be used during preprocessing",
        "detection": "A dictionary for the peak detection node (locally_exclusive)",
        "selection": "A dictionary for the peak selection node. Default is to use smart_sampling_amplitudes, with a minimum of 20000 peak_sign\
                                         and 5000 peaks per electrode on average.",
        "clustering": "A dictionary to be provided to the clustering method. By default, random_projections is used, but if legacy is set to\
                            True, one other clustering called circus will be used, similar to the one used in Spyking Circus 1",
        "matching": "A dictionary to specify the matching engine used to recover spikes. The method default is circus-omp-svd, but other engines\
                                          can be used",
        "apply_preprocessing": "Boolean to specify whether circus 2 should preprocess the recording or not. If yes, then high_pass filtering + common\
                                                    median reference + zscore",
        "shared_memory": "Boolean to specify if the code should, as much as possible, use an internal data structure in memory (faster)",
        "multi_units_only": "Boolean to get only multi units activity (i.e. one template per electrode)",
        "job_kwargs": "A dictionary to specify how many jobs and which parameters they should used",
        "debug": "Boolean to specify if the internal data structure should be kept for debugging",
    }

    sorter_description = """Spyking Circus 2 is a rewriting of Spyking Circus, within the SpikeInterface framework
    It uses a more conservative clustering algorithm (compared to Spyking Circus), which is less prone to hallucinate units and/or find noise.
    In addition, it also uses a full Orthogonal Matching Pursuit engine to reconstruct the traces, leading to more spikes
    being discovered."""

```

yger:
This PR depends on #2365, with the option to preload the recording in memory, if possible

yger:
This is good to me. The only question I have is that to measure the amount of RAM available to the system before creating a sharedmem in SC2, I would like to use psutil. This is a rather simple package, but not installed by default. Should it be a core dependency? I guess not, but this is also used in test_generate.py, and according to pyproject, it is only installed for testing

zm711:
I'm on your side @yger. I think that install psutil could be useful for a future `jobs_kwarg` optimizer that intelligently looks at number of CPUs/logical processors, RAM, disk usage. So making it a core dependency would make that much easier in the future.

yger:
Yes, I just spoke with @samuelgarcia, he also agrees psutil could go into the core dependencies. Let's wait for @alejoe91 opinion

samuelgarcia (approved):

samuelgarcia:
Merci Pierre!

PR #2277, opened 2023-11-30
Strict inegality for sparsity with radius_um

yger:
Currently, there is an inconsistency in the way we are handling default sparsity (with radius_um). In the sortingcomponents, we are always assuming a strict inegality <, while in the core.sparsity, this was <=. This might be minor, but waveforms detected via node_pipelines are thus not exactly the same as the ones extracted via waveform_extractor. This PR ensures that everywhere, when we are comparing distances w.r.t radius_um, we are using strict inegalities


samuelgarcia:
I agree @DradeAW.
I would go the other way `<=` at the other place

samuelgarcia (approved):

PR #2283, opened 2023-12-01
Make sure sampling frequency is always float

alejoe91:
This makes sure that the `self._sampling_frequency` is always float (as it should be).

Discovered this with a `resample` recording. 

Note that not having a float makes `neuroconv` complain, because `rate` must be float.


h-mayorquin (approved):
OK!
If both the snippets and the sorting set this why not have it on `BaseExtractor`?

alejoe91:
Good point. @samuelgarcia anything wrong in moving `sampling_frequency` in `BaseExtractor`?

Let's anyways do that in a separate PR ;)

h-mayorquin:
> Let's anyways do that in a separate PR ;)
Agree

PR #2285, opened 2023-12-01
Avoid loading `channel_name` property in NwbRecording`

alejoe91:
As in title. Note that this could potentially cause troubles when loading an NWB file saved by SpikeInterface and saving/appending using `neuroconv`.


h-mayorquin (approved):
I see. I will be testing this soon : )

PR #2287, opened 2023-12-01
Add a more robust `delta_x` to `unit_waveforms`

### LINKED ISSUE #2222: `plot_unit_waveforms` and `plot_unit_templates` display incorrectly for probegroup
Here is one dataset:

image

Here's another dataset:

image

The difference between the datasets are that the second has a probegroup attached rather than a probe. But other than that they should be equivalent. 

In fact the x distance for the probe is 25 um and for the probegroup, the distance on that shank is only 20 um so this isn't a distance thing. I've tried setting the figsize myself and that didn't help.

This is currently in the `matplotlib` backend.

zm711:
Fixes #2222 

## The Problem
Basically I go through this in the issue, but what I figured out is that when using a Cambridge neurotech H7 probe which has a slanting and slowly changing x position for the contacts the `delta_x` was being calculated as 0.5 which was leading to the waveforms being super squished.

image

I think the basic logic of this function before (but correct me if I'm wrong), was that in the case of `x_chans.size==1` then the x axis is all stacked so let's force an `delta_x` of 40 to prevent this type of skinny waveforms. If there are multiple unique values in `x_chans`, however, then it will likely give us a reasonable offset of something > 10 (I think 22.5 is standard for a lot of probes). But in the rare case that the deltas are less than 10 (like in the case of the H7) the figures look like above.

## My solution

If the `delta_x < 20` the waveform figure probably will look too squished so in the rare case this delta is < 20 then just force the delta to be 20. I could see an argument to be 10, so I am putting them on top of each other so we can compare:

### setting default to 10
image

### setting default to 20
image

### setting default back up to 40
image

I thought 20 looked like the best default so that's why I put it for this PR, but I could be convinced to something else like 22.5 or something like that if you all felt strongly.


zm711:
Actually I think a stricter cutoff is important some tetrodes are only spread 10-15 away so we wouldn't want to switch those to too big a default, but I think things less than 5 (at least for me) it makes sense to give them a bigger spread similar to what is done in the case of things stacked on top of each other.

alejoe91:
@zm711 another possibility is to check the uniqueness of x-values. If probes have contacts which are aligned, e.g. most of multi-column silicon probes, then `delta_x` is computed correctly as `delta_x = np.min(np.diff(np.unique(channel_locations[:, 0])))`. In my opinion, we could check the counts of the unique values. For example, if the unique values of x are 4 and the counts are 96 (this would be the case for NP1), then computing the delta_x as is makes sense. 

For the probe that you are using, you would get something like 10 values with 5 or so counts. In this case, we could adopt a different strategy knowing that almost all contacts are staggered. What do you think?

alejoe91 (approved):

PR #2294, opened 2023-12-05
Refactor NWB extractor to separate reading hdf5 file from reading nwb

h-mayorquin:
This is a small refactor which is a pre-amble to implement fast mode reading from Dandi.

### DIFF
--- a/src/spikeinterface/extractors/nwbextractors.py
+++ b/src/spikeinterface/extractors/nwbextractors.py
@@ -68,6 +68,61 @@ def retrieve_electrical_series(nwbfile: NWBFile, electrical_series_name: Optiona
     return electrical_series
 
 
+def _read_hdf5_file(
+    *,
+    file_path: str | Path | None,
+    file: BinaryIO | None = None,
+    stream_mode: Literal["ffspec", "ros3", "remfile"] | None = None,
+    cache: bool = False,
+    stream_cache_path: str | Path | None = None,
+):
+    import h5py
+
+    if stream_mode == "fsspec":
+        import fsspec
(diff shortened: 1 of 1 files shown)


src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -68,6 +68,61 @@ def retrieve_electrical_series(nwbfile: NWBFile, electrical_series_name: Optiona
...
+        else:
+            ffspec_file = fsspec_file_system.open(file_path, "rb")
+
+        hdf5_file = h5py.File(name=ffspec_file, mode="r")
+
+    elif stream_mode == "ros3":
+        assert file_path is not None, "file_path must be specified when using stream_mode='ros3'"
+
+        drivers = h5py.registered_drivers()
+        assertion_msg = "ROS3 support not enbabled, use: install -c conda-forge h5py>=3.2 to enable streaming"
```
zm711, line 105: Is ros3 only through conda-forge. I know most people use anaconda, but most of the other assertion messages say pip install instead. I know you just copied this over from the old assert, but was just curious why we had this there originally?

h-mayorquin: Yep, it requires h5py with that driver which I don't know how else to install. We are kind of beating around the bush about deprecating ros3 as we have not found any advantages over fsspec so far and is hard to test and hard to install.

zm711: Makes sense. You seem to be in the camp to just remove. Is there anything in favor of keeping it?

alejoe91 (approved):

PR #2297, opened 2023-12-05
Add `.stream.cbin` compatibility to `CompressedBinaryIblExtractor`

### LINKED ISSUE #2295: add "{stream_name}.stream.cbin" file extension to CompressedBinaryIblExtractor
When downloading IBL data, if you download a snippet using the brainbox.io.spikeglx Streamer object, it will be stored as a `{stream_name}.stream.cbin` file. The metafiles will also be stored as `{stream_name}.stream.ch` and `{stream_name}.stream.meta`. This file extension is currently not compatible with the `CompressedBinaryIblExtractor` which can only load `{stream_name}.cbin`. After some discussions with IBL, we think it would be useful if the CompressedBinaryIblExtractor could also load and use these files. I would be happy to implement this as a PR. :)

colehurwitz:
Fixes #2295

### DIFF
--- a/src/spikeinterface/extractors/cbin_ibl.py
+++ b/src/spikeinterface/extractors/cbin_ibl.py
@@ -57,7 +57,9 @@ class CompressedBinaryIblExtractor(BaseRecording):
         # snippets downloaded from IBL have the .stream.cbin suffix
         cbin_stream_files = list(folder_path.glob(f"*.{stream_name}.stream.cbin"))
         curr_cbin_files = cbin_stream_files if len(cbin_stream_files) > len(cbin_files) else cbin_files
-        assert len(curr_cbin_files) == 1, f"There should only be one `*.cbin` file in the folder, but {print(curr_cbin_files)} have been found"
+        assert (
+            len(curr_cbin_files) == 1
+        ), f"There should only be one `*.cbin` file in the folder, but {print(curr_cbin_files)} have been found"
         cbin_file = curr_cbin_files[0]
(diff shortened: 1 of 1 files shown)


src/spikeinterface/extractors/cbin_ibl.py
```diff
@@ -53,9 +53,12 @@ def __init__(self, folder_path, load_sync_channel=False, stream_name="ap"):
...
 
         # explore files
-        cbin_files = list(folder_path.glob(f"*.{stream_name}.cbin"))
-        assert len(cbin_files) == 1
-        cbin_file = cbin_files[0]
+        cbin_files = list(folder_path.glob(f"*{stream_name}.cbin"))
+        # snippets downloaded from IBL have the .stream.cbin suffix
+        cbin_stream_files = list(folder_path.glob(f"*.{stream_name}.stream.cbin"))
+        curr_cbin_files = cbin_stream_files if len(cbin_stream_files) > len(cbin_files) else cbin_files
+        assert len(curr_cbin_files) == 1
```
zm711, line 60:
```suggestion
        assert len(curr_cbin_files) == 1, f"There should only be one `*.cbin` file in the folder, but {print(curr_cbin_files)} have been found"
```
Maybe something like this just to give the user a sense of what is going wrong if this assert is triggered.

alejoe91 (approved):

alejoe91 (approved):

samuelgarcia:
Thank you Cole.

PR #2298, opened 2023-12-06
Add hdf5 backend support for`NwbRecordingExtractor` 

h-mayorquin:
Depends on #2294. This makes the NWB extraction faster as it avoids the nwb validation.

@magland If you have some time to sparse it would be great to get your pair of eyes.


magland:
@h-mayorquin Overall this looks very good.

I am a bit concerned about the electrical_series_name / electrical_series_location. I understand you are trying to maximize the chance that it will find the correct electrical series object, and provide helpful output when it is not found. But I think it's best to be explicit about the procedure that is used to determine this. Perhaps you could simplify wherever possible? And more importantly spell out the procedure in the docs for the class?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -451,13 +402,385 @@ def __init__(
...
+                load_time_vector=load_time_vector,
+                samples_for_rate_estimation=samples_for_rate_estimation,
+                stream_mode=stream_mode,
+                stream_cache_path=stream_cache_path,
+                file=file,
+                electrical_series_location=electrical_series_location,
+                cache=cache,
+            )
+        else:
+            assert False, f"Invalid io_backend: {io_backend} use 'pynwb' or 'hdf5'"
```
alejoe91, line 774:
```suggestion
            raise ValueError(f"Invalid io_backend: {io_backend} use 'pynwb' or 'hdf5'")
```
Isn't it better?

h-mayorquin:
For the BrainWide IBL dataset:
remfile and hdf5: 8.9 seconds
remfil and pynwb: 1 minute and 28 seconds
fsspec and hdf5: 3 minutes 18 seconds
fsspec and pynwb: 20 minutes and 31 seconds.

image

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -451,13 +402,385 @@ def __init__(
...
+        electrical_series_name: str | None = None,
+        load_time_vector: bool = False,
+        samples_for_rate_estimation: int = 100000,
+        stream_mode: Optional[Literal["fsspec", "ros3", "remfile"]] = None,
+        stream_cache_path: str | Path | None = None,
+        *,
+        file: BinaryIO | None = None,  # file-like - provide either this or file_path
+        electrical_series_location: str | None = None,
+        cache: bool = False,
+        io_backend: Literal["pynwb", "hdf5"] = "pynwb",
```
h-mayorquin, line 748:
```suggestion
        io_backend: Literal["pynwb", "h5py"] = "pynwb",
```

h-mayorquin:
After a discussion with @alejoe91 we decided to change the name to "use_pynwb" and the default should be `False`. When we add support for Zarr we should identify on the fly if the file is Zarr and use either h5py or the zarr library to extract the data and modify the documentation when `use_pynwb=False`. Meanwhile, the reading of zarr files will require `use_pynwb=True`.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -451,13 +402,378 @@ def __init__(
         }
 
 
+class _NWBHDF5RecordingExtractor(BaseRecording):
+    """
+    A RecordingExtractor for NWB files. This uses the hdf5 API to extract the traces and
+    the metadata and is called by the NwbRecordingExtractor factory. This should be faster
```
alejoe91, line 408:
```suggestion
    the metadata and is called by the NwbRecordingExtractor factory. This is faster
```

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -451,13 +402,378 @@ def __init__(
...
+        elif "timestamps" in electrical_series.keys():
+            t_start = electrical_series["timestamps"][0]
+            sampling_frequency = 1 / np.median(np.diff(electrical_series["timestamps"][:1000]))
+
+        # TimeSeries need to have either timestamps or rate
+        if sampling_frequency is None:
+            raise ValueError(
+                "Could not find rate information as both 'rate' and "
+                "'timestamps' are missing from the file. "
+                "Use the 'sampling_frequency' argument."
```
alejoe91, line 487: I don't think the `sampling_frequency` is among the available args (and it should not be).

In this case, we should just say that the electrical series is not valid

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -451,13 +402,378 @@ def __init__(
...
+            )
+
+        times_kwargs = dict(sampling_frequency=sampling_frequency, t_start=t_start)
+
+        # Extractors channel groups must be integers, but Nwb electrodes group_name can be strings
```
alejoe91, line 492: Groups can also be strings. They are just any other property!

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -451,13 +402,378 @@ def __init__(
...
+                    properties["location"][electrical_series_index, 1] = electrodes_table["rel_y"][
+                        electrode_table_index
+                    ]
+                if "rel_z" in electrodes_table:
+                    properties["location"][electrical_series_index, 2] = electrodes_table["rel_z"][
+                        electrode_table_index
+                    ]
+
+        # Extract all the other properties
+        for electrical_series_index, (channel_id, electrode_table_index) in enumerate(
```
alejoe91, line 559: Can't ou just merge these 2 loops anf handle "rel_x/y/z" as a special case (like "channel_name", "location", etc.)?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -451,13 +402,378 @@ def __init__(
...
+        electrodes_table = hdf5_file["/general/extracellular_ephys/electrodes"]
+        electrode_table_columns = electrodes_table.attrs["colnames"]
+
+        # Get sampling frequency
+        if "starting_time" in electrical_series.keys():
+            t_start = electrical_series["starting_time"][()]
+            sampling_frequency = electrical_series["starting_time"].attrs["rate"]
+        elif "timestamps" in electrical_series.keys():
+            t_start = electrical_series["timestamps"][0]
+            sampling_frequency = 1 / np.median(np.diff(electrical_series["timestamps"][:1000]))
```
alejoe91, line 480:
```suggestion
            sampling_frequency = 1 / np.median(np.diff(electrical_series["timestamps"][:samples_for_rate_estimation]))
```

src/spikeinterface/extractors/nwbextractors.py
alejoe91: Same logic as https://github.com/SpikeInterface/spikeinterface/pull/2298/files#diff-94bcb22069dd20ece6a264cf92514e386c25f0851afa2af80cbc7cc5bcba339bR254-R260

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -451,13 +402,378 @@ def __init__(
...
+        representation of the NWB file instead of a file path.
+    load_time_vector: bool, default: False
+        If set to True, the time vector is also loaded into the recording object. Useful for
+        cases where precise timing information is required.
+    samples_for_rate_estimation: int, default: 100000
+        The number of timestamp samples used for estimating the sampling rate. This is relevant
+        when the 'rate' attribute is not available in the ElectricalSeries.
+    stream_mode : "fsspec", "ros3", "remfile", or None, default: None
+        Determines the streaming mode for reading the file. Use this for optimized reading from
+        different sources, such as local disk or remote servers.
```
alejoe91, line 678:
```suggestion
        different sources, such as local disk or remote servers. ("remfile" is recommended)
```

src/spikeinterface/extractors/tests/test_nwbextractors.py
```diff
@@ -185,6 +210,23 @@ def test_nwb_extractor_offset_from_series(path_to_nwbfile, nwbfile_with_ecephys_
...
+        use_pynwb=False,
+    )
+
+    recording_extractor_pynwb = NwbRecordingExtractor(
+        path_to_nwbfile,
+        electrical_series_name=electrical_series_name,
+        use_pynwb=True,
+    )
+
+    check_recordings_equal(recording_extractor_hdf5, recording_extractor_pynwb)
```
alejoe91, line 227: Thanks!!!

alejoe91 (changes requested):
Thansk for this amazing work @h-mayorquin 

I have a couple of very minor final requests!

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: I am fine with fixing this but note that I just copied from main:
https://github.com/h-mayorquin/spikeinterface/blob/3702e9fbc11e8317c815cbdfa44662e22d01f5e2/src/spikeinterface/extractors/nwbextractors.py#L300-L307

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: I would not like to commit so fast to this yet:

https://github.com/NeurodataWithoutBorders/pynwb/issues/1791

I also seen too much variability around my own usage. I prefer to defer this until the nwb team give a recommendation.

src/spikeinterface/extractors/nwbextractors.py
alejoe91: Ah I think that was copied form the Sorting which has the `sampling_frequency` argument! Let's change both messages to:
```
Invalid ElectricalSeries! Could not find rate information as both 'rate' and''timestamps' are missing from the file.
```

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Let's change this.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: I am just eliminate the assertion, it can't happen. Either it has timestamps or rate.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Changed, it was carried out from main.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: I also think that the `samples_for_rate_estimation` default is too large. I think tha 1000 samples should be OK.

alejoe91: agreed. Can you change to 1000?

src/spikeinterface/extractors/nwbextractors.py
alejoe91: Ok

src/spikeinterface/extractors/nwbextractors.py
alejoe91: Probably old ;)

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Should be done.

alejoe91 (approved):

magland:
Thanks @h-mayorquin and @alejoe91 !

Issue #2033, opened 2023-09-21
Functionality to merge recordings after splitting


JoeZiminski:
I would like to provide an option to split recordings before preprocessing so recordings can be pre-processed per shank. There are alternatives (e.g. `groups` in the `common_reference` module) but this is not always provided (e.g. spatial filtering). The problem I am having is that after sorting by group, I had planned to aggregate the sorting results. But, the preprocessed data is still split by group, so I cannot figure out a way to handle the inputs for waveform extraction, because the sorting results can be aggregated but the preprocessed recordings are still split.

One solution would be to re-merge the split, preprocessed (i.e. written to binary) recordings. Is this possible in SI? I would be happy to write a method for this, although I can see why you might not want to include this functionality as it is a little confusing. For example, a user might split the recording, perform lazy pre-processing and then merge, but when preprocessing is actually performed it is no longer split by group. Nonetheless this could be handled with documentation / warnings or other checks.

alejoe91:
Hi @JoeZiminski 

Yes both are possible! You should be able to use the `aggregate_channels` and `aggregate_units` function to merge both. https://spikeinterface.readthedocs.io/en/latest/modules/core.html#manipulating-objects-slicing-aggregating

Basically, you can simply do:
```
recording_aggr = si.aggregate_channels(list_of_split_recordings)
sorting_aggr = si.aggregate_units(list_of_split_recordings)

# optional but convenient:
sorting_aggr.register_recording(recording_aggr)

we = si.extract_waveforms(recording_aggr, sorting_aggr, ...)
```

Let me know if that's what you needed!

alejoe91:
I also think that it would be nice to have a `GroupByRecording` preprocessor, so all the preprocessing by group (or property) and aggregation is done automatically. What do you think? 

An example would look like this:
```
recording_np2_4shanks = se.read_spikeglx(...)

recording_group_by = spre.group_by(recording_np2_4shanks, property="group")

recording_cmr = spre.common_reference(recording_group_by)
```
We could make a mechanism (even at the base preprocessor) to split->apply->merge traces!

@samuelgarcia @h-mayorquin what's your opinion?

JoeZiminski:
Thanks @alejoe91! Completely missed that. I think your suggestion would be really nice, it would be very convenient and reduce the liklihood of people trying this themselves and getting into trouble with the lazy aspect as mentioned above. e.g. (I think this would not work?):

```
split_recording_dict= recording.split_by("group")

[...]

PR #2299, opened 2023-12-06
Add group by preprocessing conveneience function.

JoeZiminski:
This PR creates a convenience function to run all preprocessing by a property (currently just "group") as discussed in #2033. @alejoe91 @h-mayorquin 

In this draft implementation, first all preprocesing steps and associated kwargs are iteratively discovered from the recordings `_kwargs` property. Then these are re-applied to the base recording.

However, testing shows this approach is brittle because often,  the `_kwargs` parameters do not contain only the user-passed kwargs but additional logic based on the recording. For example, in `whiten` the whitening matrix is computed based on the passed `recording` and stored in `kwargs` `W` parameter. Then, trying to apply this to smaller subsets of recordings fails. This kind of pattern occurs a number of times across preprocessors.

A possible solutions to the current implementation is to somehow ensure all preprocessors only store user-passed `kwargs` in their `_kwargs` option, but this seems like a bad idea. Maybe there is another way to get these user-passed kwargs.

Alternatively, the preprocessing steps and associated kwargs could be explicitly passed to the function as suggested in #2033. In this case though if I have understood correctly the user would have to reformat all inputs like:

```
group_by_preprocessing(recording, [PreprocessingClass1, PreprocessingClass2, ...], [preprocessing_kwargs_1, preprocessing_kwargs_2])
```
But personally if this is correct it feels easier to use the existing method of wrap standard preprocessing API in the split for-loop then aggregate rather than having to reformat the pipeline.

To summarise my thoughts on possible approaches for this convenience function / class are:

1) It is difficult to have this as a class / function applied at the end of preprocessing chain due to the issue mentioned above.
2) It does not seem possible to insert this at the start of the preproecssing chain, such that all following preprocessing steps are split. The `get_traces()` of the first-applied class should output as a single array not multiple arrays (i.e. it cannot somehow return 1 array per shank for later preprocessing). My understanding is a class that holds a split version of the recording and applies all subsequent preprocessing steps to each would still need to be handled at the level of each individual preprocessing class.
3) It would be possible to apply a user-defined list of preprocessing steps and parameters, but I'm not sure if it's use is much easier than the existing API.

Now I wonder if adding a bit more documentation to the existing API (e.g. a short tutorial 'preprocessing by shank') would be a better approach (happy to do this).

### An aside

During testing I found two cases where the API of some preprocessors does not match what seems to be standard.
- `ZeroChannelPaddedRecording` saves it's parent recording to a `parent_recording` kwarg not a `recording` kwarg that seems to be used elsewhere.
- `ZeroSilencePeriods` stores the parent recording as a dict not a recording.
Is it worth standardising these? Maybe there are some practicalities to consider that make this not possible.

### DIFF
--- a/src/spikeinterface/preprocessing/group_by_recording.py
+++ b/src/spikeinterface/preprocessing/group_by_recording.py
@@ -3,39 +3,42 @@ from typing import Dict
 from spikeinterface import load_extractor
 import numpy as np
 
-from spikeinterface.core.core_tools import define_function_from_class
 
-from spikeinterface.preprocessing.basepreprocessor import BasePreprocessor, BasePreprocessorSegment
 from spikeinterface.core.channelsaggregationrecording import ChannelsAggregationRecording, aggregate_channels
+import spikeinterface.preprocessing as spre
 
 
-def group_preprocessing_by_property(recording_):  # TODO: rename
-    """ """
+def group_preprocessing_by_property(recording_):
+    """
+    1) Recursively loop through recording ._kwargs storing applied
+       preprocessing steps and _kwargs until the initial recording is found
+    2) Re-apply all steps back onto the base recording.
+    """
     loop_recording = copy.deepcopy(recording_)
     preprocessing_step_list = []
     preprocessing_kwargs_list = []
     while True:
-        # There is a general problem here in that the __init__ sets everything
-        # up according to the full recording but then we split by channel.
-        # However it is not possible to know what kwargs are user-passed and
-        # what are setup during init.
+        # Recursively loop through recordings on kwargs, storing the preprocessing
+        # steps and the associated kwargs ----------------------------------------------
+
         if isinstance(loop_recording, ChannelsAggregationRecording):
             raise ValueError(
                 "Cannot group by recording a recording that" "has already been aggregated with `aggregate_channels`"
             )
 
-        if (
+        if (  # TODO: Handle ZeroChannelPaddedRecording that saves to 'parent_recording rather than 'recording'
             "recording" not in loop_recording._kwargs and "parent_recording" not in loop_recording._kwargs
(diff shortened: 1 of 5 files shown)


zm711:
I will also add @JoeZiminski `CurationSorting`; also used a `parent_sorting` rather than `sorting` (outside of the scope of a preprocessing PR, but a similar problem). I think I tried briefly to regularize with a super naive fix and saw that tests were failing when I did that. Then I dropped trying to regularize and decided to just delete my branch since I was going to need to carefully fix the disconnect for those. But I'm definitely in favor of regularizing parameter naming in general. (I think folder vs output_folder vs working_folder is one of the biggest naming issues I run into bouncing between functions at the argument level). Maybe this is worth an issue to fully discuss?

JoeZiminski:
Thanks @zm711 I agree it is worth an issue as these changes could have some strange unexpected downstream effects!

I will close this PR and open a new one to introduce some additional documentation for split / aggregate recordings.

h-mayorquin:
> To summarise my thoughts on possible approaches for this convenience function / class are:
> 
> 1. It is difficult to have this as a class / function applied at the end of preprocessing chain due to the issue mentioned above.
> 2. It does not seem possible to insert this at the start of the preproecssing chain, such that all following preprocessing steps are split. The `get_traces()` of the first-applied class should output as a single array not multiple arrays (i.e. it cannot somehow return 1 array per shank for later preprocessing). My understanding is a class that holds a split version of the recording and applies all subsequent preprocessing steps to each would still need to be handled at the level of each individual preprocessing class.
> 3. It would be possible to apply a user-defined list of preprocessing steps and parameters, but I'm not sure if it's use is much easier than the existing API.

This summary is very good. Thanks for writing it.

PR #2300, opened 2023-12-06
Use 'path' arg for NWBHDF5IO if file is local

alejoe91:
Fix needed for local NWB files after #2294

### DIFF
--- a/src/spikeinterface/extractors/nwbextractors.py
+++ b/src/spikeinterface/extractors/nwbextractors.py
@@ -112,10 +112,6 @@ def _read_hdf5_file(
         assert file_path is not None, "file_path must be specified when using stream_mode='remfile'"
         rfile = remfile.File(file_path)
         hdf5_file = h5py.File(rfile, "r")
-
-    elif file_path is not None:
-        file_path = str(Path(file_path).resolve())
-        hdf5_file = h5py.File(name=file_path, mode="r")
     else:
         assert file is not None, "Unexpected, file is None"
         hdf5_file = h5py.File(file, "r")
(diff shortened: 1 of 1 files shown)


src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -112,10 +112,6 @@ def _read_hdf5_file(
         assert file_path is not None, "file_path must be specified when using stream_mode='remfile'"
         rfile = remfile.File(file_path)
         hdf5_file = h5py.File(rfile, "r")
-
```
h-mayorquin, line 115: I don't think we need to remove this case as it works for the current file in the CI.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -174,14 +170,20 @@ def read_nwbfile(
...
 
-    hdf5_file = _read_hdf5_file(
-        file_path=file_path,
-        file=file,
-        stream_mode=stream_mode,
-        cache=cache,
-        stream_cache_path=stream_cache_path,
-    )
-    io = NWBHDF5IO(file=hdf5_file, mode="r", load_namespaces=True)
+    if file_path is not None and stream_mode is None:
```
h-mayorquin, line 173: But for your specific file this general fix will determine the behavior as `_read_hdf5_file` will not be called.

h-mayorquin (approved):
As mentioned the file in the CI is working. So there is something about your file or your environment. I would like to find this out but given that this work for both your local file and the CI one this seems like a more robust solution.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: I actually would like to avoid removing this, can we do that?

alejoe91:
Turns out it went away after uninstalling and reinstalling `pynwb`

Issue #2280, opened 2023-12-01
Adding a warning or check for string inputs in ids_to_indices


LeMuellerGuy:
Hello everyone,

Since ids are (mostly?) strings and `ids_to_indices` does not produce an error when entering strings because they are in fact iterable, I think it would be helpful for beginner users to at least add a warning if not an outright error to the `ids_to_indices` function to prevent unwanted behaviour. See the following example:

```
recording = se.MaxwellRecordingExtractor(local_path, stream_name="well000", rec_name="rec0000", install_maxwell_plugin=False)
# produces a faulty input, in this case the index to id 401 is 448, which actually crashes h5py
# ids_to_indices outputs [4,4,8] and not [448] because the string input is split into individual chars before being converted to int
trw = plot_traces(recording, channel_ids=recording.channel_ids[401], mode = "auto", backend="matplotlib", show_channel_ids=True)

# this works fine since the single string is encapsulated in a list
trw = plot_traces(recording, channel_ids=[recording.channel_ids[401]], mode = "auto", backend="matplotlib", show_channel_ids=True)
```

A possible comfort increase for this could look like this (added in the `ids_to_indices` function):
`if isinstance(ids, str): print("string channel ids may produce undesireable output")`
or even
`assert not isinstance(ids, str), "string channel ids must be encapsulated in a list or array"`

I have already looked into just encapsulating the string for the user, however this raises further issues with other functions (in the example's case the plotting)

h-mayorquin:
Agree with @alejoe91 that the input should be strict about the input being a list. I think that the current typing is too permisive and the lack of documentation of the function does not help either.

h-mayorquin:
Was solved by this:
https://github.com/SpikeInterface/spikeinterface/pull/2301

PR #2301, opened 2023-12-06
improve `ids_to_indices` docstring

h-mayorquin:
From the discussion of #2280.

I am not sure I would like to add a check @alejoe91 . If someone passes and int or a string we could coerce that to a list with one element but I dislike that as well.

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -95,7 +95,7 @@ class BaseExtractor:
         Convert a list of IDs into indices, either as an array or a slice.
 
         This function is designed to transform a list of IDs (such as channel or unit IDs) into an array of indices.
-        These indices are useful for data manipulation and accessing properties. When `prefer_slice` is set to `True`,
+        These indices are useful for interacting with data and accessing properties. When `prefer_slice` is set to `True`,
         the function tries to return a slice object if the indices are consecutive, which can be more efficient
         (e.g. with hdf5 files and to avoid copying data in numpy).
 
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -90,17 +90,27 @@ class BaseExtractor:
         else:
             return segment_index
 
-    def ids_to_indices(self, ids: Iterable, prefer_slice: bool = False) -> Union[np.ndarray, slice]:
+    def ids_to_indices(self, ids: list[str, int], prefer_slice: bool = False) -> Union[np.ndarray, slice]:
         """
-        Transform a ids list (aka channel_ids or unit_ids)
-        into a indices array.
-        Useful to manipulate:
-          * data
-          * properties
+        Convert a list of IDs into indices, either as an array or a slice.
 
-        "prefer_slice" is an efficient option that tries to make a slice object
-        when indices are consecutive.
+        This function is designed to transform a list of IDs (such as channel or unit IDs) into an array of indices.
+        These indices are useful for data manipulation and accessing properties. When `prefer_slice` is set to `True`,
+        the function tries to return a slice object if the indices are consecutive, which can be more efficient
+        (e.g. with hdf5 files and to avoid copying data in numpy).
 
+        Parameters
+        ----------
+        ids : list of str or int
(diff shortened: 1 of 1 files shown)


alejoe91 (dismissed):

alejoe91:
@h-mayorquin I would add an 
```
assert isinstance(ids, list), "..."
```
What do you think?

h-mayorquin:
To be honest I was kind of lazy to think about the type. Can it be an array as well or only a list? I suspect some of the code might be passing np.arrays and that assert will break something else. I can deal with that but I would rather do that in another PR immediatley after that just to get the docstring improvement immediatley.

src/spikeinterface/core/base.py
```diff
@@ -90,17 +90,27 @@ def _check_segment_index(self, segment_index: Optional[int] = None) -> int:
...
-        into a indices array.
-        Useful to manipulate:
-          * data
-          * properties
+        Convert a list of IDs into indices, either as an array or a slice.
 
-        "prefer_slice" is an efficient option that tries to make a slice object
-        when indices are consecutive.
+        This function is designed to transform a list of IDs (such as channel or unit IDs) into an array of indices.
+        These indices are useful for data manipulation and accessing properties. When `prefer_slice` is set to `True`,
```
zm711, line 98: I think data manipulation is a little loaded a term for scientific research. Maybe data interaction instead? 

I've been slowly trying to change references to `fake data` in the documentation to `simulated data` for the same reason.

zm711: Or interacting with the data (if you prefer that flow better).

h-mayorquin: x D
That's a good point : )

h-mayorquin, line 98:
```suggestion
        These indices are useful for interacting with data and accessing properties. When `prefer_slice` is set to `True`,
```

alejoe91:
Yeah it could be np.ndarray as well

h-mayorquin:
@alejoe91 let's merge this and I will add add the check in another PR immediatley  just to get the improved docstring in main.

JoeZiminski:
@h-mayorquin this is nice, an important discussion I noticed in some places checks are made against `list` / `np.array` e.g. [here](https://github.com/SpikeInterface/spikeinterface/blob/8e32955e0616b9e399c253aaf0602f3aa646dae4/src/spikeinterface/preprocessing/silence_periods.py#L59) but in that example it could also accept a `tuple`? could this function also accept a `tuple`?

I wonder if it is worth introducing a single function to centralise all of this logic that is currently spread through the codebase. Could be core / utills like:

```
def ensure_array_like(input_):
    
    if instance(str, float, int):  # bit jenky
         array_like = [input_]

    else:
         assert isinstance(input_, (list, tuple, np.ndarray)), ""

    return input_
```
(with specifics up for discussion e.g. is it a good idea to try and coerce to array-like)

samuelgarcia:
Of course np.ndarray can be the input, I think this is the most general case.
We could `ids = np.asarray(ids)` no ?

And so the implementation `indices = np.array([_main_ids.index(id) for id in ids], dtype=int)` could be changed by a np.seachsorted implementation  instead (faster).

h-mayorquin:
@JoeZiminski 
Yes, I am faimliar with the arguments:
> don't repeat yourself (DRY) principle is the number one refactoring that a) reduces bugs b) allows easy extensibiltiy c) actually makes the code more readable (if good function names are used)

Those points makes sense. I think that the problem is that it easily lends itself to programmers eagerly trying to generalize and optimize code (stuff that we like doing) which I think is bad. Here when discussing fixing this we are already discussing adding more features, you see? I think that applying don't repeat yourself to refactor existing code is great (which is what you are proposing) but a bunch of times don't repeat yourself is done in advance of the actual uses cases for the sake of "be prepared" or "future proof" and that tends to be bad (you create an abstraction without use cases which tends to fail because abstracting is hard).

I am not familiar with the specific book recommendation, thanks. Any section of the book that you would recommend specifically?

Best regards,

PR #2304, opened 2023-12-06
Add dev section to pyproject.toml

### LINKED ISSUE #2235: Add `dev` section to `pyproject.toml` ?
I was resetting up a new env on my laptop and realized that there isn't an easy way to get a bunch of the packages for spikeinterface on the developer side. Since this is really only a couple time thing in general maybe this is overkill. But I could imagine something like:

```
dev = [
    "pytest",
    "pytest-cov",
    "black",
    "pre-commit",
    "Sphinx==5.1.1",
    "sphinx_rtd_theme==1.0.0",
    "sphinx-gallery",
    "numpydoc",
    "psutil",
]
```

So for new developers they could do:

```bash
pip install -e .[dev]
```

to get the real meat of the package for testing (for core or add `[full,dev]` for more). Or if they only want to work on one submodule something like:

```bash
pip install -e .[widgets,dev]
```

This way there is quick easy install for most things (datalad/git-annex are a little too OS specific to let us easily get those). This would get the main testing tools (pytest + cov), the linting tools (black + pre-commit), and the basic doc tools (sphinx + numpydoc), so that a user doesn't have to do:

```bash
pip install -e .[full,test_core,test,docs]
pip install black
pip install pre-commit
```

Again making the environment really only happens once (in a while) so maybe not worth the effort, but might make it easier for new developers. I'm curious what people think. Although if this is super contentious this is not a hill I would die on. If totally supported I'd be happy to do a PR for this, but since I just remade my env I wouldn't even need to use this.

zm711:
Closes #2235.

@alejoe91's version of dev section.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -169,6 +169,14 @@ docs = [
 
 ]
 
+dev = [
+    "spikeinterface[test]",
+    "spikeinterface[test_core]",
+    "spikeinterface[docs]",
+    "black",
+    "pre-commit",
+]
+
 [tool.pytest.ini_options]
 markers = [
     "core",


alejoe91 (approved):

alejoe91:
Thansk @zm711 !!

PR #2305, opened 2023-12-06
Add a new empty module for generation of datasets.

samuelgarcia:


doc/modules/generation.rst
```diff
@@ -0,0 +1,12 @@
+Generation module
+=================
```
zm711, line 2:
```suggestion
Generation module (in dev)
==========================
```
Will this be ready or is this a place holder? It seems to me that we should document that this is in development and not fully ready yet since it is a place holding module. But maybe this is a very temporary place holder with a very big code infusion coming soon?

doc/modules/generation.rst
```diff
@@ -0,0 +1,12 @@
...
+
+The :py:mod:`spikeinterface.generation` provide function to generate recording containing spikes.
+This module propose sevral apporach for this : purely synthetic or "hybrid" (with template coming from true datasets).
```
zm711, line 5:
```suggestion
The :py:mod:`spikeinterface.generation` provides functions to generate recordings containing spikes.
This module proposes several approaches for this including purely synthetic recordings as well as "hybrid" recordings (where templates come from true datasets).
```

doc/modules/generation.rst
```diff
@@ -0,0 +1,12 @@
...
+
+
+The :py:mod:`spikeinterface.core.generate` already propose some functions but here there is an extended and more complex
+machinery, for instance generation of drifting recordings.
```
zm711, line 9:
```suggestion
The :py:mod:`spikeinterface.core.generate` already provides functions for generating synthetic data but this module will supply an extended and more complex
machinery, for instance generating recordings that possess various types of drift.
```

zm711 (commented):
Just a few doc comments. Not super crucial now since I can do a bigger edit when the module is more fleshed out.

alejoe91 (approved):

PR #2308, opened 2023-12-07
Fix bug in from_peaks

yger:
Actually, the bug for my empty waveforms come from a deeper bug in NumpySorting.from_peaks(). The function is currently wrong, because no remapping of the channel_index (becoming unit_index) were performed

### DIFF
--- a/src/spikeinterface/core/numpyextractors.py
+++ b/src/spikeinterface/core/numpyextractors.py
@@ -328,6 +328,12 @@ class NumpySorting(BaseSorting):
         if unit_ids is None:
             unit_ids = np.unique(peaks["channel_index"])
 
+        # We need to remap the indices in the spikes
+        old_indices = spikes["unit_index"].copy()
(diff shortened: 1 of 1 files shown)


yger:
No, I think this is not enough. For example I'm detecting peaks, only found on some channel index. I would like to turn that into a sorting, thus using from_peaks. The problem is that unit_ids is the list of channel indices, but we need to then remap into the spike vector channel_indices to arange(len(unit_ids))

yger:
Actually, I think I disagree with the discussion we had. You told me that we could create empty units to avoid the problem. Fine. But then still, if I have N channels and only peaks on a subset of channels, with N < nb_channels, then no way I can infer the total number of channels from the peaks, and create empty units for the channels higher than N. So this does not make sense to me, and remap is the only way to ensure we have one unit per active channel

samuelgarcia:
#2315 is better conceptuablly.
In short the unit_ids must be given otherwise empty units break the unti_index logic.

PR #2313, opened 2023-12-08
Refactor NWBSortingSegment

h-mayorquin:
This is on the same line as #2294 and #2298. I am adding capabilities for the extractors to use the hdf5 backend which should be faster for streaming. The idea is that I will pass the data of the spikes of the unit indexes instead of the whole nwbfile as we are doing right now. As both hdf5 and zarr array follow the numpy API for the purposes of indexing this will allow us to use the `NWBSortingSegment` with both backend.

Also in this PR I am adding a test for extraction of `unit_name` as `channel_ids` when the column exists on the units table and a corresponding test. Also enchancing the testing to test for the correct extraction of properties which will be needed to test the backend in a future PR.


src/spikeinterface/extractors/tests/test_nwbextractors.py
```diff
@@ -190,20 +194,22 @@ def test_sorting_extraction_of_ragged_arrays(tmp_path):
...
+
     spike_times1 = np.array([0.0, 1.0, 2.0])
-    nwbfile.add_unit(spike_times=spike_times1, unit_name="a")
+    nwbfile.add_unit(spike_times=spike_times1, unit_name="a", a_property="a_property_value")
     spike_times2 = np.array([0.0, 1.0, 2.0, 3.0])
-    nwbfile.add_unit(spike_times=spike_times2, unit_name="b")
+    nwbfile.add_unit(spike_times=spike_times2, unit_name="b", a_property="b_property_value")
 
-    ragged_array_bad = [[1, 2, 3], [1, 2, 3, 5]]
+    ragged_array_bad = [[1, 2, 3, 8, 10], [1, 2, 3, 5]]
```
zm711, line 204: Out of curiosity why is this `ragged_array_bad` this is actually a ragged array whereas `ragged_array_good` is technically not ragged right? I understand 'bad' because it is ragged, but then wouldn't it be a little better to call the "bad" array

```
ragged_array = [[1,2,3,8,10],[1,2,3,5]] # ragged array doesn't work with nwb
...
standard_array = [[1,2,3],[4,5,6]] # standard array works fine with nwb.
```

This doesn't affect the tests, but wouldn't it be better naming to not call it `ragged_array_bad` if it is is a good ragged array, which is bad for the structure?

h-mayorquin: The ragged array bad is bad because it does not work with spikeinterface. Not the other way around. Both ragged arrays good and bad work fine in nwb.

We could have better naming though.

zm711: Aw. Cool thanks. My main point was on the naming, but since I don't use nwb too much (yet), I wasn't sure if the problem (with being ragged) was nwb or si. Thanks for the clarification.

h-mayorquin: I mean, you are correct, the namig is ambigious in that way. It could be better. I will think on something.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -571,39 +570,48 @@ def __init__(
...
-        import warnings
-
-        for column in list(self._nwbfile.units.colnames):
-            if column == "spike_times":
-                continue
-
-            # Note that this has a different behavior than self._nwbfile.units[column].data
-            property_values = self._nwbfile.units[column][:]
+        # Add properties
+        properties_to_add = [name for name in name_to_column_data if "index" not in name]
```
h-mayorquin, line 593: check for ending with underscore index to be more precise as a filter.

src/spikeinterface/extractors/tests/test_nwbextractors.py
h-mayorquin: I added more meaningful names, @zm711.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: @alejoe91 this is done, I also added logic to exclude double nested arrays. This could be supported if we want but should be another PR.

src/spikeinterface/extractors/tests/test_nwbextractors.py
```diff
@@ -185,43 +189,64 @@ def test_nwb_extractor_offset_from_series(path_to_nwbfile, nwbfile_with_ecephys_
...
+        name="uniform_ragged_array",
+        description="A uniform ragged array that can be loaded into spikeinterface",
+        data=uniform_ragged_array,
         index=True,
     )
 
+    doubled_ragged_array = [[[0, 1], [2, 3]], [[4, 5], [6, 7]]]
+    nwbfile.add_unit_column(
+        name="doubled_ragged_array",
+        description="A doubled ragged array that can be loaded into spikeinterface",
```
h-mayorquin, line 223:
```suggestion
        description="A doubled ragged array that can not be loaded into spikeinterface",
```

alejoe91 (approved):

src/spikeinterface/extractors/tests/test_nwbextractors.py
zm711: Thanks @h-mayorquin!

PR #2316, opened 2023-12-08
Split by recording first commit.

JoeZiminski:
This PR adds a 'How to' page on preprocessing and sorting when splitting the recording into channel groups.

- I am not sure if a 'How to' is the best place for this though it seems to fit, happy to move
- This duplicates / incorporates the 'Spike sorting by group' section, maybe this existing section could be removed or additional link to the new page.

I also had some general questions

- The example uses a tetrode (16 channels, 4 tetrodes). I don't quite understand this arangement, I understood a tetrode to be 4 wires. Is it possible to have tetorode silicon probes with a number of channels per probe? I also wonder if this is as common as neuropixels shanks, if not maybe the example could use neuropixels shanks, but I am not sure how common different approaches are across the field.
- Does probeinterface have a way to show only certain channel groups? If possible wanted to also include a plot of the different channel groups, highlighted by color or something.
- Sphinx is pinned to an older version, I can see why as sphinx version changes are a nightmare. Nonetheless the newer versions have some cool features (like [dropdowns]*(https://sphinx-design.readthedocs.io/en/latest/dropdowns.html), currently would would be a dropdown is included as a note at the end). At least, I think the version is why this is not working. Is it worth considering unpinning this dependency?

# TODO: can have a tetrode with some channels on it?
# TODO: can probeinterface plot by group?

### DIFF
--- a/doc/how_to/process_by_channel_group.rst
+++ b/doc/how_to/process_by_channel_group.rst
@@ -1,0 +1,128 @@
+Processing a Recording by Channel Group (e.g. Shank)
+===========================================================
+
+In this tutorial, we will walk through how to preprocess and sorting a recording
+per channel group. A channel group is a subset of channels grouped by some
+feature - a typical example is grouping channels by shank in a Neuropixels recording.
+
+First, we can check the channels on our recording are grouped as we expect. For example,
+in the below example we have a X-shank neuropixels recording. We can visualise the
+channel groupings with XXX.
+
+
+On our SpikeInterface recording, we can also access the channel groups per-channel
+
+1) channel groups
+2) index channel ids by groups
+
+Why would you want to preprocess or sort by group?
+
+1)
+2)
+3)
+
+Splitting a Recording by Channel Group
+--------------------------------------
+
+We can split a recording into mutliply recordings, one for each channel group, with the `split_by` method.
+
+Here, we split a recording by channel group to get a `split_recording_dict`. This is a dictionary
+containing the recordings, split by channel group:
+
+```
+split_recording_dict = recording.split_by("group")
+print(split_recording_dict)
+XXXXX
+```
+
+Now, we are ready to preprocess and sort by channel group.
+
+Preprocessing a Recording by Channel Group
+------------------------------------------
+
+The essense of preprocessing by channel group is to first split the recording
+into separate recordings, perform the preprocessing steps, then aggregate
+the channels back together.
+
+Here, we loop over the split recordings, preprocessing each shank group
+individually. At the end, we use the `aggregate_channels` function
+to combine the per-shank recordings back together.
+
+```
+preprocessed_recordings = []
+for recording in split_recordings_dict.values():
+
+
+    shifted_recording = spre.phase_shift(recording)
+
+    filtered_recording = spre.bandpass_filter(shifted_recording)
+
+    referenced_recording = spre.common_reference(filtered_recording)
+
+    preprocessed_recordings.append(referenced_recording)
+
+combined_preprocessed_recording = aggregate_channels(preprocessed_recordings)
+
+```
+
+It is strongly recommended to use the above structure to preprocess by group.
+A discussion of the subtleties of the approach may be found in the below
+Notes section for the interested reader.
+
+# Preprocessing channel in depth
+
+Preprocessing and aggregation of channels is very flexible. Under the hood,
+`aggregate_channels` keeps track of when a recording was split. When `get_traces()`
+is called, the preprocessing is still performed per-group, even though the
+recording is now aggregated.
+
+However, to ensure data is preprocessed by shank, the preprocessing step must be
+applied per-group. For example, the below example will NOT preprocess by shank:
+
+```
+split_recording = recording.split_by("group")
+split_recording_as_list = list(**split_recording.values())
+combined_recording = aggregate_channels(split_recording_as_list)
+
+# will NOT preprocess by channel group.
+filtered_recording = common_reference(combined_recording)
+
+```
+
+Similarly, in the below example the first preprocessing step (bandpass filter)
+would applied by group (although, this would have no effect in practice
+as this preprocessing step is always performed per channel). However,
+common referencing (which is effected by splitting by group) will
+not be applied per group:
+
+```
+split_recording = recording.split_by("group")
+
+filtered_recording = []
+for recording in split_recording.values()
+    filtered_recording.append(spre.bandpass_filtered(recording))
+
+combined_recording = aggregate_channels(filtered_recording)
+
+# As the recording has been combined, common referencing
+# will NOT be applied per channel group.
+referenced_recording = spre.common_reference(combined_recording).
+```
+
+Finally, it is not recommended to apply `aggregate_channels` more than once
+as this will slow down `get_traces()` and may result in unpredictable behaviour.
+
+
+Sorting a Recording by Channel Group
+------------------------------------
+
+Sorting a recording can be performed in two ways. One, is to split the
+recording by group and use `run_sorter` (LINK) separately on each preprocessed
+channel group.
+
+Altearntively, SpikeInterface proves a conveniecne function XXX
+to this.
+
+Example
+
+Done!
--- a/doc/how_to/index.rst
+++ b/doc/how_to/index.rst
@@ -8,3 +8,4 @@ How to guides
     analyse_neuropixels
     handle_drift
     load_matlab_data
+    process_by_channel_group


zm711:
So if the goal of this is to be a full tutorial run locally and converted to `.rst` for the docs then the desired flow is to place [here](https://github.com/SpikeInterface/spikeinterface/tree/main/examples/how_to) as a `.py` file that can then be converted later (typically real examples using NPs etc). This used to be a more strict rule. Super light examples (simulated data or mearec currently) would go to the module gallery in the appropriate section. I honestly am neutral on the location of the tutorials, but just to FYI you as you work on this (in case it is decided it needs to move and be rewritten in .py style).

And thanks! I think the growing number of multishank datasets really requires some documentation for SI users to understand how to do this with the machinery.

JoeZiminski:
Thanks @zm711! It is pretty light with toy example data so I think 'Module Example Gallery' would be the right place, maybe under 'core' as it contains information on both preprocessing and sorting?

alejoe91:
Hi @JoeZiminski 

Thanks so much for this!! @zm711 honestly I think that the gallery is mainly there as legacy and we will remove it in a future refactoring of the docs. The modules documentation is already the place for detailed information about each module, and IMO the gallery is just redundant (plus hard to maintain!). What do you think?

In that regard, I'd keep this in the "How to" section!

zm711:
I was always in support of the `How to`. It just makes mental sense to me whereas module gallery was always confusing to me as a concept. 

I think getting rid of them would be okay (they do break the most--although I will say they did help me catch a bug in one of the widgets one time :) ). My biggest problem with them is that they currently (almost) all require datalad. So if they stick around I really think they need to switch to being a simulated dataset so that anyone can run them rather than just people with the fortitude to get datalad working.

samuelgarcia:
I think we need to keep the gallery which auto generated as long it is fast enought.
The howto section was heavy stuuf that we generate locally but we keep a jupytext file to be able to regenrate when necessary.

alejoe91:
@JoeZiminski did you intend to include this: `doc/sg_execution_times.rst`? What is it?

samuelgarcia:
Hi @JoeZiminski 
When we do a howto doc which is generated from a notebook. (I think this is your case). Then ,  we need to keep the notebook inside the source code in case we when to mody it one day.
In that case we decided to use jupytext to put the notebook in `examples/how_to/*.py` to work on it later.
We push the **py** but not the **ipynb** and finally convert it to rst and copy/paste the rst into the `doc/how_to/*.rst`
This is important to able to re-run the notebook without keeping the ipynb.
Here you only push the converted RST.
Have a look to this https://github.com/SpikeInterface/spikeinterface/tree/main/examples/how_to

Maybe I am wrong and you directly wrote the rst. Then it is OK.

In short, we have 3 types of documentation:
  1. pure RST which is at the moment more or less `modules/*`. If we modify the API this files are not chnaged
  2. generated by `examples/*.py` and rendered automatically as `html/py/ipynb` in the sphinx-gallery
  3. locally run notebook and manually transformed into rst.

For me:
**case 2** is very important even if Alessio not like the actual content. We can improve the content.
**case 3** should be avoid when the generation is super fast and we should put it in **case 2** but when it imply long computation or heavy data we do not have choice.

zm711:
Maybe I should open an issue where we can discuss the rules for example generation and then when we have the official rules laid out I can open a PR to add it to the development portion of the docs so we can have it laid out in stone how we want to organize examples. Does that sound like a plan? I think the structure isn't clear currently so we keep rehashing where examples should go.

samuelgarcia:
I am agree that the structure is not perfect and is should be orthogonal to this case 1 but the sphinx-gallary need to be somehow a section in the main toctree.
lets put this on an issue and maybe lets wirte it in the doc itself!!

JoeZiminski:
Thanks everyone! 

For now I did not make the documentation in a `.py` file because I could not get some cool features like [notes / warnings](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html) to work. Also, [dropdowns](https://sphinx-design.readthedocs.io/en/latest/dropdowns.html) are not working. I think this is because of the sphinx-version but am not sure. I was going to propose updating the sphinx version, however that is outside the scope of this PR.

I will redo this page in the standard format as described above (`.py` > `.ipynb` > `.rst`) and can move discussion of the docs to #2327 and add those features back later.

h-mayorquin:
This looks great. 

I second @alejoe91:
> @JoeZiminski did you intend to include this: `doc/sg_execution_times.rst`? What is it?


Out of curiosity, what do you mean here by unpredictable behavior:
image

Also you say:

> I will redo this page in the standard format as described above (.py > .ipynb > .rst)

Why? I think the rst works better than `.py` and Sam said. 

> Maybe I am wrong and you directly wrote the rst. Then it is OK.

doc/how_to/process_by_channel_group.rst
```diff
@@ -0,0 +1,216 @@
...
+
+        split_recording = recording.split_by("group")
+        split_recording_as_list = list(**split_recording.values())
+        combined_recording = aggregate_channels(split_recording_as_list)
+
+        # will NOT preprocess by channel group.
+        filtered_recording = common_reference(combined_recording)
+
+    In the below example the first preprocessing step (bandpass filter)
+    would applied by channel group (although, in practice this would have no effect
```
h-mayorquin, line 197:
```suggestion
    would be applied by channel group (although, in practice this would have no effect
```
?

doc/how_to/process_by_channel_group.rst
```diff
@@ -0,0 +1,216 @@
...
+
+     aggregate_sorting = run_sorter_by_property(
+        sorter_name='kilosort2',
+        recording=preprocessed_recording,
+        grouping_property='group',
+        working_folder='working_path'
+    )
+
+
+Further notes on preprocessing by channel group
```
h-mayorquin, line 173: I find this section a bit confusing. Is the goal of it eliminating the confusion that people might have of thinking that when they apply a pre-processing step to aggregate recording it will still apply per-group?

If so, I will start the session by stating the possible confusion, then saying that this is not correct and then your illustrative example.

I feel that the first paragraph is trying to add some context but I don't think that's necessary as this is at the end of a tutorial that should have make this all clear.

doc/how_to/process_by_channel_group.rst
```diff
@@ -0,0 +1,216 @@
...
+.. code-block:: python
+
+    import spikeinterface.extractors as se
+    import spikeinterface.preprocessing as spre
+    from spikeinterface import aggregate_channels
+    from probeinterface import generate_tetrode, ProbeGroup
+    import numpy as np
+
+    # Create a toy 384 channel recording with 4 shanks (each shank contain 96 channels)
+    recording, _ = se.toy_example(duration=[1.00], num_segments=1, num_channels=384)
```
h-mayorquin, line 42: Open question:
When do we use `toy_example` and we do we use `generate_recording` or `generate_ground_truth_recording`.  I personally would prefer to only use the latter but maybe it does not matter.

doc/how_to/process_by_channel_group.rst
```diff
@@ -0,0 +1,216 @@
+Processing a Recording by Channel Group
+=======================================
+
+In this tutorial, we will walk through how to preprocess and sort a recording
+separately for *channel groups*. A channel group is a subset of channels grouped by some
+feature - for example a multi-shank Neuropixels recording in which the channels
+are grouped by shank.
+
+**Why preprocess by channel group?**
```
h-mayorquin, line 9: Thanks for adding a why section at the beginning. This is great.

alejoe91:
@JoeZiminski  maybe this was buried in the discussion!

> @JoeZiminski did you intend to include this: `doc/sg_execution_times.rst`? What is it?

Did you include on purpose? If so, should it be referenced somewhere? I would remove it for now!

alejoe91:
@JoeZiminski just pinging you on this! We are planning a release on Friday, so it would be great if you can tackle the remaining points by then :)

doc/how_to/process_by_channel_group.rst
JoeZiminski: Thanks @h-mayorquin this is a very good point, it is strange to have a seciton at the end re-visiting the entire article in more detail. I've dropped the second half of this section and integrated the rest into a note within the page body. Let me know what you think!

doc/how_to/process_by_channel_group.rst
JoeZiminski: I am not sure on this and will defer to @alejoe91 @samuelgarcia

alejoe91: I think that for docs `toy_example` is ok! It's actually easier to intepret

doc/how_to/process_by_channel_group.rst
h-mayorquin: I re-read the current one and it looks great to me!

alejoe91 (approved):

doc/how_to/process_by_channel_group.rst
```diff
@@ -0,0 +1,188 @@
...
+
+        referenced_recording = spre.common_reference(filtered_recording)
+
+        preprocessed_recordings.append(referenced_recording)
+
+    # Combine our preprocessed channel groups back together
+    combined_preprocessed_recording = aggregate_channels(preprocessed_recordings)
+
+Now, when this recording is used in sorting, plotting, or whenever
+calling it's :py:func:`~get_traces` method, the data will have been
```
zm711, line 117:
```suggestion
calling its :py:func:`~get_traces` method, the data will have been
```

zm711 (commented):
one super tiny typo :)

JoeZiminski:
Thanks all for your reviews!

Issue #2319, opened 2023-12-12
problem of amplitude on localy exclusive and by_channel peak detection?


Kayv-cmb:
Hi Spikeinterface,

I have huge 'spike' in my recording more than 1.5mV which for me is higly unprobable considering electrophysiology recording. (Recorded with open_ephys and neuropixel 2.0). I wanted to have a look at those outside of the traditionnal sorter (Kilosort and stuff). However i have issue with the detect_peak function. I wrote a code (see code below) that will plot those peak by channel. However what is plot does not always have the amplitude mentioned in the detect peak (I tried both locally exclusive and by_channel) function. I filtered for <2000 uV peak but what is plotted doesnt have the amplitude expected see figure below
```

import numpy as np
import pandas as pd
import warnings
import tqdm 
def fxn():
    warnings.warn("deprecated", DeprecationWarning)

with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    fxn()
    
from spikeinterface.preprocessing import common_reference,bandpass_filter
import os
import spikeinterface.full as si
import spikeinterface.extractors as se 
import spikeinterface.sorters as ss
import spikeinterface.widgets as sw
import matplotlib.pyplot as plt
# Load the preprocessed data
directory_pre = '/nemo/project/proj-schaefera-ephys/sorted_batch/NP_231120/2023-11-23_12-58-34/recording1/experiment1/'
pre_rec = si.load_extractor(directory_pre+'preprocessing_slice')
from spikeinterface.sortingcomponents.peak_detection import detect_peaks

job_kwargs = dict(chunk_duration='1s', n_jobs=40, progress_bar=True)

peaks_re = detect_peaks(
    recording=pre_rec,
    method='by_channel',
    peak_sign='neg',
    detect_threshold=5,
    exclude_sweep_ms=0.2,
    #radius_um=75,
    noise_levels=None,
    random_chunk_kwargs={},
    **job_kwargs,
)


filtered_peaks = peaks_re[peaks_re['amplitude'] < -2000]
channel_location = pre_rec.get_channel_locations()
list_channel = [] 
for i in tqdm.trange(channel_location.shape[0]):
    channel_peaks = filtered_peaks[filtered_peaks['channel_index']==i]
    plt.figure()
    for p in channel_peaks:
        channel = 'CH'+str(i)
        sample_peak = p['sample_index']
        s = int(sample_peak - 0.002*30000)
        e = int(sample_peak + 0.003*30000)
[...]

PR #2320, opened 2023-12-12
Get traces fixes

zm711:
Based on #2319.

Just wanted to fix the docstring. Basically the docstring says the default for `return_scaled` is None when the default is False.

I also updated the module gallery example that had it listed that return_scaled=True was the default which it isn't.

Finally, I just updated some of the typing/docstring to move to the `|` format since we are moving this way across parts of the code base. I switched the channel_ids from `Iterable` which is technically correct to `list | np.array` because they are the more common typing that users will reach for. I know technically something like a tuple could also work, but since we are not type checking the code base I tend to like to use type hints as recommended types for the user and not strictly accurate types for the developer. If we want to switch it back to Iterable I can, or I could add tuple, but this is my preferred way. 

Related to #2319, I double checked the docs and they do explain already that `return_scaled` is needed in the core section under recording if uV are desired.

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -1,6 +1,6 @@
+from __future__ import anntotations
 import warnings
 from pathlib import Path
-from typing import Iterable, List, Union
 from warnings import warn
 
 import numpy as np
@@ -36,12 +36,12 @@ class BaseRecording(BaseRecordingSnippets):
         "noise_level_mad_scaled",
     ]
 
-    def __init__(self, sampling_frequency: float, channel_ids: List, dtype):
+    def __init__(self, sampling_frequency: float, channel_ids: list, dtype):
         BaseRecordingSnippets.__init__(
             self, channel_ids=channel_ids, sampling_frequency=sampling_frequency, dtype=dtype
         )
 
-        self._recording_segments: List[BaseRecordingSegment] = []
+        self._recording_segments: list[BaseRecordingSegment] = []
 
         # initialize main annotation and properties
         self.annotate(is_filtered=False)
@@ -766,20 +766,20 @@ class BaseRecordingSegment(BaseSegment):
 
     def get_traces(
         self,
-        start_frame: Union[int, None] = None,
-        end_frame: Union[int, None] = None,
-        channel_indices: Union[List, None] = None,
+        start_frame: int | None = None,
+        end_frame: int | None = None,
+        channel_indices: list | np.array | None = None,
     ) -> np.ndarray:
         """
         Return the raw traces, optionally for a subset of samples and/or channels
 
         Parameters
         ----------
-        start_frame: Union[int, None], default: None
+        start_frame: int | None, default: None
             start sample index, or zero if None
-        end_frame: Union[int, None], default: None
+        end_frame: int | None, default: None
             end_sample, or number of samples if None
-        channel_indices: Union[List, None], default: None
+        channel_indices: list | np.array | None, default: None
             Indices of channels to return, or all channels if None
         order: list or None, default: None
             The memory order of the returned array.
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -242,34 +242,34 @@ class BaseRecording(BaseRecordingSnippets):
 
     def get_traces(
         self,
-        segment_index: Union[int, None] = None,
-        start_frame: Union[int, None] = None,
-        end_frame: Union[int, None] = None,
-        channel_ids: Union[Iterable, None] = None,
-        order: Union[str, None] = None,
-        return_scaled=False,
-        cast_unsigned=False,
+        segment_index: int | None = None,
+        start_frame: int | None = None,
+        end_frame: int | None = None,
+        channel_ids: list | np.array | None = None,
+        order: "C" | "F" | None = None,
+        return_scaled: bool = False,
+        cast_unsigned: bool = False,
     ):
         """Returns traces from recording.
 
         Parameters
         ----------
-        segment_index : Union[int, None], default: None
+        segment_index : int | None, default: None
             The segment index to get traces from. If recording is multi-segment, it is required, default: None
-        start_frame : Union[int, None], default: None
+        start_frame : int | None, default: None
             The start frame. If None, 0 is used, default: None
-        end_frame : Union[int, None], default: None
+        end_frame : int | None, default: None
             The end frame. If None, the number of samples in the segment is used, default: None
-        channel_ids : Union[Iterable, None], default: None
+        channel_ids : list | np.array | None, default: None
             The channel ids. If None, all channels are used, default: None
-        order : Union[str, None], default: None
-            The order of the traces ("C" | "F"). If None, traces are returned as they are, default: None
-        return_scaled : bool, default: None
+        order : "C" | "F" | None, default: None
+            The order of the traces ("C" | "F"). If None, traces are returned as they are
+        return_scaled : bool, default: False
             If True and the recording has scaling (gain_to_uV and offset_to_uV properties),
-            traces are scaled to uV, default: False
-        cast_unsigned : bool, default: None
+            traces are scaled to uV
+        cast_unsigned : bool, default: False
             If True and the traces are unsigned, they are cast to integer and centered
-            (an offset of (2**nbits) is subtracted), default: False
+            (an offset of (2**nbits) is subtracted)
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/baserecording.py
```diff
@@ -1,6 +1,6 @@
+from __future__ import anntotations
```
zm711, line 1:
```suggestion
from __futures__ import anntotations
```

zm711, line 1:
```suggestion
from __future__ import annotations
```

src/spikeinterface/core/baserecording.py
```diff
@@ -242,34 +242,34 @@ def get_total_memory_size(self) -> int:
...
-        end_frame : Union[int, None], default: None
+        end_frame : int | None, default: None
             The end frame. If None, the number of samples in the segment is used, default: None
-        channel_ids : Union[Iterable, None], default: None
+        channel_ids : list | np.array | None, default: None
             The channel ids. If None, all channels are used, default: None
-        order : Union[str, None], default: None
-            The order of the traces ("C" | "F"). If None, traces are returned as they are, default: None
-        return_scaled : bool, default: None
+        order : "C" | "F" | None, default: None
```
h-mayorquin, line 265: This should be a literal

src/spikeinterface/core/baserecording.py
```diff
@@ -766,20 +766,20 @@ def get_num_samples(self) -> int:
...
+        start_frame: int | None, default: None
             start sample index, or zero if None
-        end_frame: Union[int, None], default: None
+        end_frame: int | None, default: None
             end_sample, or number of samples if None
-        channel_indices: Union[List, None], default: None
+        channel_indices: list | np.array | None, default: None
             Indices of channels to return, or all channels if None
         order: list or None, default: None
             The memory order of the returned array.
```
zm711, line 785: Why is this here? It is an argument? Was it just copied from the `get_traces` above?

h-mayorquin (commented):
The change from Iterable to something more specific is correct. A single string is also an iterable but channel_ids can't be strings. 

I also would flag https://github.com/SpikeInterface/spikeinterface/issues/2319 as evidence that  `return_scaled` is kind of obscure by itself as a naming. (What scaled? scaled to what?) 

Thanks for doing this.

src/spikeinterface/core/baserecording.py
zm711: It can be a Literal, but as of 3.10 it doesn't have to be a Literal. We can list them as strings, but type checkers will complain. Again I think it looks cleaner without, but if we want to always use Literal we can. As a user I think it is cleaner to see without the surrounding `Literal[a | b]`, but some of these are more personal aesthetics.

h-mayorquin: Interesting. Can you point the documentation of this to me? The change in 3.10 I mean.

I honestly don't care that much about it, if you like it that was it is fine for me. As you mentioned above, our use of typing is an aid to documentation.

zm711:
> I also would flag https://github.com/SpikeInterface/spikeinterface/issues/2319 as evidence that return_scaled is kind of obscure by itself as a naming. (What scaled? scaled to what?)

yeah I guess that's true. Maybe I've worked on Neo so much at this point that I always think about the scaling from the raw data to the voltage, but for new users the fact that most devices save their recordings not in voltage may not be apparent. For someone with domain knowledge I think `return_scaled` makes sense, but for someone without domain knowledge it could be unclear. Do you think highlighting a section (like we did in the Matlab docs you did) where we explain this concept of scaling should go in the main docs under the `Recording` section?

h-mayorquin:
@zm711 that would be good. Most people don't read the documentation though that's why I think there is no cost of having more explicit names. Point for future naming though. We are stuck with the current one.

src/spikeinterface/core/baserecording.py
zm711: Sorry I meant use of the `|` was 3.10. Typing always accepted strings due to how they are parsed. So it used to be that you could do typing with:

```python
def func(a: "int | float " = 5)
```
So the use of strings isn't new to 3.10, but the fact we can use the pipe without needing to in case of strings. So if we want to eventually go to type checking then I think we would need to use `Literal` for the type checker to properly parse. If we are just using this as docs then it doesn't matter. I'll do either. If there is any chance we would consider type checking I would say let's add in the Literal now so it's less burden later, but if we are never going to type check then I like without Literal.

h-mayorquin: I don't undesrstand. Literal means that there is a fixed set of options from which you want could choose: ["C", "F", None] are the set of possible values.

Does "C" | "F" | None now has the same semantics as Literal["C", "D", None]? 
That was the documentation I was looking for. Out of curiosity as I mentioned, if you like it that way, it is fine for me.

zm711:
I added a small note because the docs for that section aren't currently designed for a large section, but feel free to give that small section a read and see if it adds a little bit more emphasis to this scaling feature.

src/spikeinterface/core/baserecording.py
h-mayorquin: Ah, reading the relevant PEPs does not spell it out explictly it seems but I think what you mention is correct.

doc/modules/core.rst
```diff
@@ -92,6 +92,11 @@ with 16 channels:
     timestamps = np.arange(num_samples) / sampling_frequency + 300
     recording.set_times(timestamps, segment_index=0)
 
+**Note**:
+Raw data formats often store data as integer values for memory efficiency. To give these integers meaningful physical units, you can apply a gain and an offset.
```
h-mayorquin, line 96: This is Ok, maybe we can exptlictly say that they are transformed to mV. That is, those are the meaningful physical units.

zm711: Good point, although it's actually uV, which just goes to show how important it is for us to actually specify the units we are converting to!

zm711, line 96:
```suggestion
Raw data formats often store data as integer values for memory efficiency. To give these integers meaningful physical units (uV), you can apply a gain and an offset.
```

doc/modules/core.rst
```diff
@@ -92,6 +92,11 @@ with 16 channels:
...
+Many devices have their own gains and offsets necessary to convert their data and these values are handled by SpikeInterface for its extractors. This
+is triggered by the :code:`return_scaled` parameter in :code:`get_traces()`, (see above example).
```
zm711, line 98:
```suggestion
is triggered by the :code:`return_scaled` parameter in :code:`get_traces()`, (see above example), which will return the traces in uV.
```

src/spikeinterface/core/baserecording.py
samuelgarcia: I guess it was a bad and lazy copy/paste

DradeAW:
Can you expand to `list | np.ndarray | tuple`?

See #2324, I think tuples should be accepted :)

zm711:
> Can you expand to list | np.ndarray | tuple?

Done :)

src/spikeinterface/core/baserecording.py
zm711: Cool. I removed it.

alejoe91 (approved):

samuelgarcia (approved):

samuelgarcia:
Merci Zach

src/spikeinterface/core/baserecording.py
```diff
@@ -766,24 +766,21 @@ def get_num_samples(self) -> int:
 
     def get_traces(
         self,
-        start_frame: Union[int, None] = None,
-        end_frame: Union[int, None] = None,
-        channel_indices: Union[List, None] = None,
+        start_frame: int | None = None,
+        end_frame: int | None = None,
+        channel_indices: list | np.array | None = None,
```
DradeAW, line 771: or tuple

zm711: Thanks. I totally missed that one. I'll push a fix in a second.

zm711, line 771:
```suggestion
        channel_indices: list | np.array | tuple | None = None,
```

src/spikeinterface/core/baserecording.py
```diff
@@ -766,24 +766,21 @@ def get_num_samples(self) -> int:
...
         Parameters
         ----------
-        start_frame: Union[int, None], default: None
+        start_frame: int | None, default: None
             start sample index, or zero if None
-        end_frame: Union[int, None], default: None
+        end_frame: int | None, default: None
             end_sample, or number of samples if None
-        channel_indices: Union[List, None], default: None
+        channel_indices: list | np.array | None, default: None
```
zm711, line 782:
```suggestion
        channel_indices: list | np.array | tuple | None, default: None
```

PR #2322, opened 2023-12-13
Add from `__future__ import annotations` to core base.

JoeZiminski:
#2301 adds improved type hints to `core/base.py` `ids_to_indices()` function. These include the pipe `|` operator as alias for `Union` which requires `from __future__ import annotations` for python < 3.10, which this PR adds.

I have no idea why there is an arrow in the line above :sweat_smile:

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 from pathlib import Path
 import re
 from typing import Any, Iterable, List, Optional, Sequence, Union


h-mayorquin (approved):
Thanks. I should start using python 3.8 on my own machine just to catch this errors : /

h-mayorquin:
@alejoe91 @samuelgarcia I am merging this as core is faulty now for someone with python 3.8 and python 3.9

samuelgarcia:
ok thanks

alejoe91:
Thanks @JoeZiminski and @h-mayorquin 

Yes I also got a report about this!

h-mayorquin:
I would love to but you need to convince @samuelgarcia who is against it for yet to be disclosed reasons.

h-mayorquin:
To be fair to @samuelgarcia here he is not happy with the focus on typing so this is kind of on us though : P

h-mayorquin:
And I think he proposed that maybe the tests should be run  python 3.8 which is an improvement over the status quo (but has other problems).

JoeZiminski:
Thanks Ramon! it will be good to hear from @samuelgarcia  on this. I would suggest to cover all supported python version and operating systems in the CI. I can't see too many downsides and the upside is that all problems are caught before release and saves in maintainanche headache 😅 .

h-mayorquin:
Sam is probably rightfully concern that testing will be slower here on the CI? I run tests locally but I think some people use the CI for that (I think @DradeAW works like that if I remember well). We need to find a solution that does not get on the way of people workflows but at the same time catches as many errors as possible. Still thinking about it.

samuelgarcia:
Hi
I will again look like old and conservative but I am quite in favor of frugality for CI.
Running all python X all numpy is a pretty matrix have a impact an the ernergy.

We now run the core test with an average python version, 3.10 I think.
Maybe a compromise could be to run on the latest + oldest but for core only. That would be 3.8 and 3.12.
And the full would be run with an average version 3.10 or 3.11.

h-mayorquin:
A public reminder to everyone reading this that you can and probably should deactivate testing in your forked version of spikeinterface.

DradeAW:
> I run tests locally but I think some people use the CI for that

If you mean running tests locally VS waiting for GitHub to give the answer, then I do both!
When feeling well, I run the tests locally like a good programmer. And when I'm lazy, I use GitHub's CI ^^'

JoeZiminski:
Thanks @samuelgarcia! I agree it is good to consider the energy implications and think that is a nice comprimise is to run earliest and latest supported. 

I would also advocate for testing on every python version at least once before release. The reason is due to the possibility of very scary and strange effects of version changes that can go missed if not tested in code. This may go beyond simple crashes to changes that result in numerical issues or processes run in completely unexpected ways. In such a case these may effects the results of things like sorting but never be picked up by users because of lack of ground truth and will only never be picked up in numerical tests. 

I agree this is overkill for things like commits to a PR, but maybe can be run on merge into main, or when tags are pushed for new version release?

PR #2323, opened 2023-12-13
Multi-channel template metrics fix

alejoe91:
This PR fixes a few typos and protects against a failure in the `spread` metric if the template is all zeros, so no channels above a relative threshold are found

**EDIT:** the problem was that the normalization of the amplitudes must be done after gaussian smoothing, not before!


alejoe91:
@samuelgarcia @h-mayorquin this is ready to merge on my side!

samuelgarcia (approved):

PR #2324, opened 2023-12-13
ids can be a tuple

DradeAW:
Add `tuple` for `ids_to_indices`

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -91,7 +91,7 @@ class BaseExtractor:
         else:
             return segment_index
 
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/base.py
```diff
@@ -102,7 +104,7 @@ def ids_to_indices(self, ids: list | np.ndarray | None = None, prefer_slice: boo
 
         Parameters
         ----------
-        ids : list or np.ndarray
+        ids : list or np.ndarray or tuple or None
```
zm711, line 107:
```suggestion
        ids : list | np.ndarray | tuple | None, default: None
```

Just so we have the docstring in full compliance. The pipes aren't crucial but we need to add in the default value.

PR #2325, opened 2023-12-13
Fix number overflow for file size when saving recording in core_tools.py

przemyslawj:
This pull request fixes a number overflow issue for saved file size. My concatenated recording threw an error on save due to negative file_size_bytes for the allocated file before the fix.

### DIFF
--- a/src/spikeinterface/core/core_tools.py
+++ b/src/spikeinterface/core/core_tools.py
@@ -257,7 +257,7 @@ def write_binary_recording(
     file_path_dict = {segment_index: file_path for segment_index, file_path in enumerate(file_path_list)}
     for segment_index, file_path in file_path_dict.items():
         num_frames = recording.get_num_frames(segment_index=segment_index)
-        data_size_bytes = dtype_size_bytes * num_frames * num_channels
+        data_size_bytes = np.int64(dtype_size_bytes) * num_frames * num_channels
         file_size_bytes = data_size_bytes + byte_offset
 
         file = open(file_path, "wb+")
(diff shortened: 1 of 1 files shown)


alejoe91:
Thanks for spotting this @przemyslawj

h-mayorquin:
@przemyslawj which format you are using, the overflow is caused by one of your values being a numpy scalar it would be good to see which one so we can fix this issue from the root.

src/spikeinterface/core/core_tools.py
```diff
@@ -257,7 +257,7 @@ def write_binary_recording(
     file_path_dict = {segment_index: file_path for segment_index, file_path in enumerate(file_path_list)}
     for segment_index, file_path in file_path_dict.items():
         num_frames = recording.get_num_frames(segment_index=segment_index)
-        data_size_bytes = dtype_size_bytes * num_frames * num_channels
+        data_size_bytes = np.int64(dtype_size_bytes) * num_frames * num_channels
```
h-mayorquin, line 260: To avoid this type of overflow errors, the general solution is to have num_frames and num_channels to python int which has an infinite size. 

The proble comes with num_frames or num_channels is a numpy scalar.

h-mayorquin:
@przemyslawj can you test?
https://github.com/SpikeInterface/spikeinterface/pull/2326

przemyslawj:
Hi both,
The problem was for a binary recording. I tested https://github.com/SpikeInterface/spikeinterface/pull/2326 and it works, thank you.

h-mayorquin:
You mean reading a binary recording of ours? A `BinaryRecordingExtractor`? 

That is, you mentioned that you were concatenating something, was that a `BinaryRecordingExtractor` that you are concatenating.

Glad that https://github.com/SpikeInterface/spikeinterface/pull/2326 solve your issue, this means that num_frames was a numpy scalar. Thanks for catching it up and reporting it!

PR #2329, opened 2023-12-13
Add an auto adjustment if  `n_jobs` too high on Windows

zm711:
### Problem
Turns out ProcessPoolExecutor has a hard limit in Windows that is not dependent on the number of processors. The limit is 61. So this limit could be breached with a 31 core with 62 logical processors.

### Relevant Docs
https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor

### My Solution
I added an extra check in `ensure_n_jobs` function k to make sure n_jobs is not above 61 and if it is then it cuts off to 61. I would assume most people would move over to a Linux based server at that point, but better safe than sorry no?

### Alternative Solution
We could raise an assert or just let the ProcessPoolExecutor raise the error.


src/spikeinterface/core/job_tools.py
```diff
@@ -142,6 +142,10 @@ def ensure_n_jobs(recording, n_jobs=1):
     elif n_jobs is None:
         n_jobs = 1
 
+    # ProcessPoolExectuor has a hard limit of 61 for Windows
```
zm711, line 145:
```suggestion
    # ProcessPoolExecutor has a hard limit of 61 for Windows
```

alejoe91 (approved):

samuelgarcia (approved):

samuelgarcia:
merci Zach.

PR #2340, opened 2023-12-14
Add `from __future__  import annotations` to core files

zm711:
After #2322, I think it is worth just adding __future__ to all of core. I do everything in 3.11 currently (and test suite is 3.10) so it is easy to forget 3.8 is still being supported. I believe I found a couple type hints that were not 3.8 compliant and were missing from __future__. Once 3.10 is the minimum supported version we can delete these lines easily, but it is safer just to have them until that point. If desired I can do the whole code base next.

Also updated a couple type hints to be more in line with the code base.
Added one docstring.


h-mayorquin:
If the team is OK with this, we then should run testing suite locally with python < 3.10 so we know we did not miss any.

alejoe91:
Let's go for it! But I would do <= 3.10

zm711:
I didn't touch the zarr file on purpose because I @alejoe91 was doing the rewrite and I didn't want there to be an annoying merge conflict if this got merged before that.

alejoe91:
> I didn't touch the zarr file on purpose because I @alejoe91 was doing the rewrite and I didn't want there to be an annoying merge conflict if this got merged before that.

I'll add it on my PR! Thansk for the reminder!

zm711:
Current test failure is the empty generation module. It skipped all tests which is marked as failing I guess. pre-commit issue also unrelated to this PR.

src/spikeinterface/core/node_pipeline.py
```diff
@@ -43,23 +43,27 @@
...
+        self,
+        recording: BaseRecording,
+        return_output: bool | tuple[bool] = True,
+        parents: Optional[list[Type["PipelineNode"]]] = None,
     ):
         """
-        This is a generic object that will make some computation on peaks given a buffer of traces.
-        Typically used for exctrating features (amplitudes, localization, ...)
+                This is a generic object that will make some computation on peaks given a buffer of traces.
+                Typically used for exctrating features (amplitudes, localization, ...)
```
alejoe91, line 53: @zm711 is this extra indentation on purpose? Looks weird to me!

alejoe91: fixed ;)

alejoe91 (approved):

src/spikeinterface/core/node_pipeline.py
zm711: Thanks. It wasn’t intentional. Pre-commit did that to me. I have no clue why…. But thanks!

PR #2348, opened 2023-12-18
Fix RTD warnings

zm711:
There are currently three RTD warnings active:

1) The autodoc is looking for `run_sorters`: this is my bad. I should have fixed that when I removed the function in #2240 
2) SD docs were not underlined correctly fixed
3) the intersphinx mapping for Neo is failing for some reason
```
intersphinx inventory 'https://neo.readthedocs.io/en/stable/objects.inv' not fetchable due to <class 'requests.exceptions.HTTPError'>: 404 Client Error: Not Found for url: https://neo.readthedocs.io/en/stable/objects.inv
building [mo]: targets for 0 po files that are out of date
```

I tried fetching using `latest` instead of `stable` and it seemed to work locally. So this PR is also testing to see if this build works switching the intersphinx. If it does then it can be merged. If it fails then I'll revert the neo mapping and try to read up more on why it could be failing.


zm711:
Okay using neo latest fixed the error for the mapping. Probeinterface is fine with stable... But this is good to go for review.

alejoe91 (approved):

Issue #2335, opened 2023-12-14
Creation of a SharedMemoryRecording


yger:
Actually, it would be great to have such a SharedMemoryRecording object, similarly to what has been done with SharedMemorySorting. With such an object, we could preload in memory at the beginning of a spike-sorting pipeline the entire traces (assuming memory is large enough), and speed up drastically internal sorters. Not everyone has recordings larger than 32Go, so it could be useful, especially for speeding up benchmarks and the generate_recording() traces generated on-the-fly. We could pregenerate everything once, and then quickly access the data

samuelgarcia:
merci beaucoup camarade de rapeller ce que j'ai promis.

yger:
Oui, spécial dédicace :-)

h-mayorquin:
We can use the latest pickle protocol out-of-band buffer to pickle extractors containing no memory maps with shared memory (when memory allows):

https://peps.python.org/pep-0574/

This way, we eliminate the distinction of memory vs memmap from the API which is an implementation detail of our ad-hoc parallelization sheme.

zm711:
Closed by #2365.

PR #2351, opened 2023-12-19
Sharedmem recording

yger:
This is an attempt to address #2335. I know that @samuelgarcia might not be fully happy with the implementation, but despite the fact that no tests have been added yet in the PR, this is working properly and speeding up quite drastically all preprocessing. Basically, you can create a recording, save it to memory and if n_jobs > 1, this is a sharedmemory object (if n_jobs=1, the old beahvior is kept and a NumpyRecording is returned). 
I am open to edit the PR and add tests if needed. I also did not took into account the comments from @h-mayorquin related to https://peps.python.org/pep-0574/, because it seems to me that such a move would require a bigger refactoring. Being not fully experienced with core, I chose to simply mimick what has been done for SharedMemorySorting


h-mayorquin:
@yger 
Thanks for the mention. I think that you are absolutley right. What I was proposing is more of a -possible- long-term refactor. But this should not get in the way of using this technique to solve the problem at hand. Thanks for taking my ideas into account.

samuelgarcia:
Hi Pierre.
Thank you for this.
If you are OK I will make another PR with the same idea but sligtly different implementation. I want the inti to not have the trace_list. and add some construct with classmethod.
I will also move some function to recording_tools at the same time.

yger:
I am super fine with fine, be my guest, feel free to close this PR and make it your way. This was just a proposal/proof of concept to assess performances

yger:
Closing as this will be handled by #2365

PR #2353, opened 2023-12-19
Cache spike_vector from parent

DradeAW:


src/spikeinterface/core/unitsselectionsorting.py
```diff
@@ -46,6 +46,25 @@ def __init__(self, parent_sorting, unit_ids=None, renamed_unit_ids=None):
...
+                return
+
+        parent_spike_vector = self._parent_sorting._cached_spike_vector
+        parent_unit_indices = self._parent_sorting.ids_to_indices(self._unit_ids)
+        mask = np.isin(parent_spike_vector["unit_index"], parent_unit_indices)
+        spike_vector = np.array(
+            parent_spike_vector[mask]
+        )  # np.array() necessary to fix 'read-only' crash with memmaps.
+        spike_vector["unit_index"] = np.searchsorted(
+            parent_unit_indices, spike_vector["unit_index"]
```
samuelgarcia, line 63: The reverse no ?

src/spikeinterface/core/basesorting.py
```diff
@@ -529,6 +535,9 @@ def to_spike_vector(self, concatenated=True, extremum_channel_inds=None, use_cac
             spike_dtype = spike_dtype + [("channel_index", "int64")]
             ext_channel_inds = np.array([extremum_channel_inds[unit_id] for unit_id in self.unit_ids])
 
+        if use_cache and self._cached_spike_vector is None:
+            self._cache_spike_vector_from_parent()
```
samuelgarcia, line 539: I would call this  self._custum_cache_spike_vector() because it can cache this from something else than the parent. No ?

src/spikeinterface/core/unitsselectionsorting.py
DradeAW: When I tried it, it seemed to work:

```python
>>> arr = np.array([0, 1, 0, 0, 7, 1, 1, 0, 1, 7, 0, 1, 9])
>>> np.searchsorted([0, 1, 7, 9], arr)
array([0, 1, 0, 0, 2, 1, 1, 0, 1, 2, 0, 1, 3])
```

src/spikeinterface/core/basesorting.py
DradeAW: Do you have an example in mind?

alejoe91: Maybe a `MergeUnitsSorting`? In this case it's from multiple parents. Custom sounds ok to me

DradeAW:
@h-mayorquin I know you love profiling, so here it is:

```python
# Sortings is a dict with 4 sorting objects
self.sortings = {name: scur.remove_excess_spikes(sorting.remove_empty_units(), recording) for name, sorting in sortings.items()}
```

This single line takes 79.7s on `main`, and 7.5s on this PR

h-mayorquin:
> @h-mayorquin I know you love profiling, so here it is:
> 
> ```python
> # Sortings is a dict with 4 sorting objects
> self.sortings = {name: scur.remove_excess_spikes(sorting.remove_empty_units(), recording) for name, sorting in sortings.items()}
> ```
> 
> This single line takes 79.7s on `main`, and 7.5s on this PR

I do, but some context would be good : )

DradeAW:
> I do, but some context would be good : )

Well I have 4 sorting objects which are `NumpySorting`, all different analyses of the same recording.
I measure the time it takes to run the line above, and as you can see there is quite a speed increase ^^

src/spikeinterface/core/basesorting.py
```diff
@@ -496,6 +496,14 @@ def precompute_spike_trains(self, from_spike_vector=None):
                 for unit_id in unit_ids:
                     self.get_unit_spike_train(unit_id, segment_index=segment_index, use_cache=True)
 
+    def _custom_cache_spike_vector(self) -> None:
+        """
+        Function that can be implemented by some children sorting to quickly
+        cache the spike vector without computing it from spike trains
+        (e.g. computing it from a sorting parent).
+        """
+        pass
```
zm711, line 505: One tiny question: 

If the child class needs to implement why a `pass` and not a `raise NotImplementedError`?

DradeAW: The child class does not need to implement it :)

zm711: So you're relying on the `_` to say it is a private function that people will never accidently call and have nothing happen? 

To better explain my point:
 If someone accidently calls a function I would expect it to say this child class doesn't not support use of this function which is indicated by a `NotImplementedError`. But if this is private then I suppose you are saying that if someone uses they are doing so at their own risk of this failing?

And as a second point what if I as a developer want to add a "failing" test for this so I can implement my own function to have the test pass (TDD), then if it has built in pass I can't use this strategy.

Again, I'm just curious about the strategy and I really appreciate the docstring!

zm711: Okay cool. I don't work at this level with SI that much (and I know you do this stuff all the time) so thanks for the explanation :)

src/spikeinterface/core/basesorting.py
```diff
@@ -496,6 +496,14 @@ def precompute_spike_trains(self, from_spike_vector=None):
                 for unit_id in unit_ids:
                     self.get_unit_spike_train(unit_id, segment_index=segment_index, use_cache=True)
 
+    def _custom_cache_spike_vector(self) -> None:
```
h-mayorquin, line 499: Are you using cache as a verb here? Would not compute instead of cache makes more sense and avoid having something that has 4 verbs? 

But I do remember that you and @samuelgarcia already had a discussion about compute vs cache in some other thread.

src/spikeinterface/core/basesorting.py
```diff
@@ -529,6 +537,9 @@ def to_spike_vector(self, concatenated=True, extremum_channel_inds=None, use_cac
             spike_dtype = spike_dtype + [("channel_index", "int64")]
             ext_channel_inds = np.array([extremum_channel_inds[unit_id] for unit_id in self.unit_ids])
 
+        if use_cache and self._cached_spike_vector is None:
```
h-mayorquin, line 540: I think that after a week I will find this line very confusing for me to read (now it makes sense because I have been in the discusssion and I know the context but I will forget or someone else that did not know will come along).

If I come to this function for the first time in the base class it says "custom_cache_spike_vector" I go to the implementation and is empty. What does it does? When does it do it ? I do ctrl+f in the library and It seems to sometimes (when?) have the side effect of setting a private property (`self._cached_spike_vector`) and then the function continues as it is was before. 

I think it would be more clear if the function were to return the `spike_vector` then you do whatever you need to do with it and do an early return here. That way, it would be very explicit in the code that you are overiding the construction of the spike_vector with custom logic for some cases.

DradeAW: I don't like the idea of returning the spike vector, because then you need to work things out like concatenated vs not concatenated, and you duplicate the parameters and increase the complexity of the code.

Here the goal is simple: creating the cache, and then `to_spike_vector` does what it does with the cache.

src/spikeinterface/core/basesorting.py
```diff
@@ -496,6 +496,14 @@ def precompute_spike_trains(self, from_spike_vector=None):
                 for unit_id in unit_ids:
                     self.get_unit_spike_train(unit_id, segment_index=segment_index, use_cache=True)
 
+    def _custom_cache_spike_vector(self) -> None:
+        """
+        Function that can be implemented by some children sorting to quickly
+        cache the spike vector without computing it from spike trains
+        (e.g. computing it from a sorting parent).
```
alejoe91, line 503:
```suggestion
        Function that can be implemented by some children sorting to quickly
        cache the spike vector without computing it from spike trains
        (e.g. computing it from a sorting parent).
        This function should set the `self._cached_spike_vector`, see for 
        instance the `UnitsSelectionSorting` implementation.
```

src/spikeinterface/core/basesorting.py
alejoe91: I think it's ok!

samuelgarcia (approved):

PR #2354, opened 2023-12-19
Improve CRM docstring

h-mayorquin:
As in the the title. I was thinking on this  yesterday and I thought that I could improve the docstring.
 
I am not modifying the docstring concerning the group parameter for two reasons:
1) Alessio pointed out to me yesterday that once we have the convenience by @JoeZiminski described in https://github.com/SpikeInterface/spikeinterface/issues/2033 we might get rid of this here (which I think is a good idea).
2) I also was finding it hard to describe what the function does when the group option is used. That is, I needed to add to many caveats that I felt substracted from the clarity. If we want we can do that in a second PR.

### DIFF
--- a/src/spikeinterface/preprocessing/common_reference.py
+++ b/src/spikeinterface/preprocessing/common_reference.py
@@ -1,25 +1,47 @@
+from __future__ import annotations
 import numpy as np
+from typing import Optional, Literal
 
 from spikeinterface.core.core_tools import define_function_from_class
 
 from .basepreprocessor import BasePreprocessor, BasePreprocessorSegment
 from ..core import get_closest_channels
+from spikeinterface.core.baserecording import BaseRecording
 
 from .filter import fix_dtype
 
 
 class CommonReferenceRecording(BasePreprocessor):
     """
-    Re-references the recording extractor traces.
+    Re-references the recording extractor traces. That is, the value of the traces are
+    shifted so the there is a new zero (reference).
+
+    The new reference can be estimated either by using common median reference (CMR) or
+    common average reference (CAR).
+
+    The new reference can be set three ways:
+        * "global": the median/average of all channels is set as the new reference.
+        That is, the median/average is subtracted from all channels.
+        * "single": In the simplest case, a single channel from the recorder is set as the new reference.
+        That is, the channel is subtracted from all other channels. To use this option, the `ref_channel_ids` argument
+        is used with a single channel id.
+        A collection of channels can also be used as the new reference. In this case, the median/average of the
+        selected channels is subtracted from all other channels. To use this option, pass the group of channels as
+        a list in `ref_channel_ids`.
+        * "local": the median/average within an annulus is set as the new reference. The parameters
+        of the annulus are specified using the `local_radius` argument. With this option, both close channels and
+        far channels are excluded from the median/average. Note that setting the `local_radius` to (0, exclude_radius)
+        will yield a simple circular local region.
+
 
     Parameters
     ----------
     recording: RecordingExtractor
         The recording extractor to be re-referenced
     reference: "global" | "single" | "local", default: "global"
-        If "global" then CMR/CAR is used either by groups or all channel way.
-        If "single", the selected channel(s) is remove from all channels. operator is no used in that case.
-        If "local", an average CMR/CAR is implemented with only k channels selected the nearest outside of a radius around each channel
+        If "global" the reference is the average or mean across all the channels.
+        If "single", the reference is a single channel or a list of them that need to be set with the `ref_channel_ids`.
+        If "local", the reference are the channels within an annulus that need to be set with the `local_radius`.
     operator: "median" | "average", default: "median"
         If "median", common median reference (CMR) is implemented (the median of
             the selected channels is removed for each timestamp).
@@ -34,9 +56,15 @@ class CommonReferenceRecording(BasePreprocessor):
         list of channels to be applied to each group is expected. If "single" reference, a list of one channel  or an
         int is expected.
     local_radius: tuple(int, int)
-        Use in the local CAR implementation as the selecting annulus (exclude radius, include radius)
-    verbose: bool
-        If True, output is verbose
+        Use in the local CAR implementation as the selecting annulus with the following format:
+
+        `(exclude radius, include radius)`
+
+        Where the exlude radius is the inner radius of the annulus and the include radius is the outer radius of the
+        annulus. The exclude radius is used to exclude channels that are too close to the reference channel and the
+        include radius delintes the outer boundary of the annulus whose role is to exclude channels
+        that are too far away.
+
     dtype: None or dtype
         If None the parent dtype is kept.
 
@@ -44,19 +72,27 @@ class CommonReferenceRecording(BasePreprocessor):
     -------
     referenced_recording: CommonReferenceRecording
         The re-referenced recording extractor object
+
+
+    Notes
+    -----
+        Note that the references are calculated within a `get_traces` context. That is, the reference is calculated
+    for each chunk of traces that is requested. Which means that `get_traces(stat_frame=0, end_frame=10_000)`
+    will be different from `get_traces(start_frame=0, end_frame=20_000)[0:10_000]`. While in most cases the estimation
+    of the mean or median will be the same, it is important to keep this in mind when using this with small chunks.
+
     """
 
     name = "common_reference"
 
     def __init__(
         self,
-        recording,
-        reference="global",
-        operator="median",
+        recording: BaseRecording,
+        reference: Literal["global", "single", "global"] = "global",
+        operator: Literal["median", "average"] = "median",
         groups=None,
         ref_channel_ids=None,
         local_radius=(30, 55),
-        verbose=False,
         dtype=None,
     ):
         num_chans = recording.get_num_channels()
@@ -102,13 +138,13 @@ class CommonReferenceRecording(BasePreprocessor):
         else:
             group_indices = None
         if ref_channel_ids is not None:
-            ref_channel_inds = self.ids_to_indices(ref_channel_ids)
+            ref_channel_indices = self.ids_to_indices(ref_channel_ids)
         else:
-            ref_channel_inds = None
+            ref_channel_indices = None
 
         for parent_segment in recording._recording_segments:
             rec_segment = CommonReferenceRecordingSegment(
-                parent_segment, reference, operator, group_indices, ref_channel_inds, local_radius, neighbors, dtype_
+                parent_segment, reference, operator, group_indices, ref_channel_indices, local_radius, neighbors, dtype_
             )
             self.add_recording_segment(rec_segment)
 
@@ -130,7 +166,7 @@ class CommonReferenceRecordingSegment(BasePreprocessorSegment):
         reference,
         operator,
         group_indices,
-        ref_channel_inds,
+        ref_channel_indices,
         local_radius,
         neighbors,
         dtype,
@@ -140,7 +176,7 @@ class CommonReferenceRecordingSegment(BasePreprocessorSegment):
         self.reference = reference
         self.operator = operator
         self.group_indices = group_indices
-        self.ref_channel_inds = ref_channel_inds
+        self.ref_channel_indices = ref_channel_indices
         self.local_radius = local_radius
         self.neighbors = neighbors
         self.temp = None
@@ -160,22 +196,21 @@ class CommonReferenceRecordingSegment(BasePreprocessorSegment):
         if channel_indices is not None:
             _channel_indices = _channel_indices[channel_indices]
 
+        out_traces = np.zeros((all_traces.shape[0], _channel_indices.size), dtype=self.dtype)
+
         if self.reference == "global":
-            out_traces = np.zeros((all_traces.shape[0], _channel_indices.size), dtype=self.dtype)
             for chan_inds, chan_group_inds in self._groups(_channel_indices):
                 out_inds = np.array([np.where(_channel_indices == i)[0][0] for i in chan_inds])
                 out_traces[:, out_inds] = all_traces[:, chan_inds] - self.operator_func(all_traces[:, chan_group_inds])
 
         elif self.reference == "single":
-            out_traces = np.zeros((all_traces.shape[0], _channel_indices.size), dtype=self.dtype)
             for i, (chan_inds, _) in enumerate(self._groups(_channel_indices)):
                 out_inds = np.array([np.where(_channel_indices == i)[0][0] for i in chan_inds])
                 out_traces[:, out_inds] = all_traces[:, chan_inds] - self.operator_func(
-                    all_traces[:, [self.ref_channel_inds[i]]]
+                    all_traces[:, [self.ref_channel_indices[i]]]
                 )
 
         elif self.reference == "local":
-            out_traces = np.zeros((all_traces.shape[0], _channel_indices.size), dtype=self.dtype)
             for i, chan_ind in enumerate(_channel_indices):
                 out_traces[:, [i]] = all_traces[:, [chan_ind]] - self.operator_func(
                     all_traces[:, self.neighbors[chan_ind]]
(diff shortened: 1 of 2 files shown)


zm711:
Yeah that's a doozy; I think it's definitely headed in a better direction! It's such a complicated function/class under the hood that it is hard to have a docstring that gives enough info without being overwhelming. I'll give it a super careful read through in a couple hours after I finish some morning meetings.

src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -1,25 +1,47 @@
...
+    Re-references the recording extractor traces. That is, the value of the traces are
+    shifted so the there is a new zero (reference).
+
+    The new reference can be estimated either by using common median reference (CMR) or
+    common average reference (CAR).
+
+    The new reference can be set three ways:
+        * "global": the median/average of all channels is set as the new reference.
+        That is, the median/average is subtracted from all channels.
+        * "single": In the simplest case, a single channel from the recorder is set as the new reference.
```
alejoe91, line 25:
```suggestion
        * "single": In the simplest case, a single channel from the recording is set as the new reference.
```

src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -1,25 +1,47 @@
...
+        That is, the channel is subtracted from all other channels. To use this option, the `ref_channel_ids` argument
+        is used with a single channel id.
```
alejoe91, line 27:
```suggestion
        is used with a single channel id. Note that this option will zero out the reference channel.
```

src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -34,29 +56,43 @@ class CommonReferenceRecording(BasePreprocessor):
...
     referenced_recording: CommonReferenceRecording
         The re-referenced recording extractor object
+
+
+    Notes
+    -----
+        Note that the references are calculated within a `get_traces` context. That is, the reference is calculated
+    for each chunk of traces that is requested. Which means that `get_traces(stat_frame=0, end_frame=10_000)`
+    will be different from `get_traces(start_frame=0, end_frame=20_000)[0:10_000]`. While in most cases the estimation
+    of the mean or median will be the same, it is important to keep this in mind when using this with small chunks.
```
alejoe91, line 82: @h-mayorquin I think this is not true. The median/mean etc. is calculated for each sample separately so it doesn't matter how you retrieve the traces

alejoe91 (changes requested):
Thanks @h-mayorquin 

I think that the added `Notes` is not correct, since the subtraction is done sample-wise!

src/spikeinterface/preprocessing/common_reference.py
h-mayorquin: Yes, thanks, the notes are wrong. The average is across the channels not across time right? So it will be affected if you slice channels but I think that right now you slice over all the channels at the beginning to avoid this. 

Thanks!

src/spikeinterface/preprocessing/common_reference.py
h-mayorquin: This is a good addition to make it clear. Thanks.

src/spikeinterface/preprocessing/common_reference.py
h-mayorquin: Thanks : )

src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -1,25 +1,47 @@
...
+        That is, the channel is subtracted from all other channels. To use this option, the `ref_channel_ids` argument
+        is used with a single channel id. Note that this option will zero out the reference channel.
+        A collection of channels can also be used as the new reference. In this case, the median/average of the
+        selected channels is subtracted from all other channels. To use this option, pass the group of channels as
+        a list in `ref_channel_ids`.
+        * "local": the median/average within an annulus is set as the new reference. The parameters
+        of the annulus are specified using the `local_radius` argument. With this option, both close channels and
+        far channels are excluded from the median/average. Note that setting the `local_radius` to (0, exclude_radius)
+        will yield a simple circular local region.
+
```
zm711, line 35:
```suggestion
        * "global": the median/average of all channels is set as the new reference.
            In this case, the 'global' median/average is subtracted from all channels.
        * "single": In the simplest case, a single channel from the recording is set as the new reference.
            This channel is subtracted from all other channels. To use this option, the `ref_channel_ids` argument
            is used with a single channel id. Note that this option will zero out the reference channel.
            A collection of channels can also be used as the new reference. In this case, the median/average of the
            selected channels is subtracted from all other channels. To use this option, pass the group of channels as
            a list in `ref_channel_ids`.
        * "local": the median/average within an annulus is set as the new reference. 
                The parameters of the annulus are specified using the `local_radius` argument. With this option, both 
                channels which are too close and channels which are too far are excluded from the median/average. Note 
                that setting the `local_radius` to (0, exclude_radius)  will yield a simple circular local region.

```

A couple things: This was hard to read visually so I tried to add indents just to make it easier to read the simple explanation (after argument) and the detailed explanation (indented).

Also, (for English we don't typically use 'that is' that often. If you want to sound academic people will use 'ie' which is Latin for that is, but that is would only be used rarely. It's not wrong, it's just not standard.

Finally the local section still isn't quite clear. How can something be too close? (ie the relation between exclude_radius and local_radius is not completely clear just based on this docstring. You mention exclude_radius after you've mentioned that things too close will be exclude which is surprising for a reader (in my opinion).

src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -1,25 +1,47 @@
...
 
     Parameters
     ----------
     recording: RecordingExtractor
         The recording extractor to be re-referenced
     reference: "global" | "single" | "local", default: "global"
-        If "global" then CMR/CAR is used either by groups or all channel way.
-        If "single", the selected channel(s) is remove from all channels. operator is no used in that case.
-        If "local", an average CMR/CAR is implemented with only k channels selected the nearest outside of a radius around each channel
+        If "global" the reference is the average or mean across all the channels.
```
zm711, line 42:
```suggestion
        If "global" the reference is the average or median across all the channels.
```

:)

src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -1,25 +1,47 @@
...
+        If "single", the reference is a single channel or a list of them that need to be set with the `ref_channel_ids`.
```
zm711, line 43:
```suggestion
        If "single", the reference is a single channel or a list of channels that need to be set with the `ref_channel_ids`.
```
explicit better than implicit no? :)

zm711 (commented):
Here are my initial comments. I added in a couple grammar notes since I know you like them @h-mayorquin, but if you ever want me to leave my reasoning off just let me know.

src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -1,25 +1,47 @@
...
 
 
 class CommonReferenceRecording(BasePreprocessor):
     """
-    Re-references the recording extractor traces.
+    Re-references the recording extractor traces. That is, the value of the traces are
+    shifted so the there is a new zero (reference).
+
+    The new reference can be estimated either by using common median reference (CMR) or
+    common average reference (CAR).
```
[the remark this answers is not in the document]
h-mayorquin: I was thinking that CMR and CAR are the methods used to estimate the new reference. As in, I estimate the error using gradient descent (where gradient descent is the method). Would I stll need the article 'a'?

But also, now that I think about this, this is not an estimation, this is the exact value. See Alessio comments for why I was wrong.

src/spikeinterface/preprocessing/common_reference.py
h-mayorquin: Thanks for the suggestions here. I will commit to check how the proposed visual style looks like as the diff is not a good vizualisation for this. Usually I go for this to read code (diff is bad for large PRs I feel):

image

But here it does not work.

> Also, (for English we don't typically use 'that is' that often. 
Appreciate this, definitley a clutch of mine. 

As for the radius, I really think that the parameter is a little bit too general and is hard to explain. But basically, the exclude radius is used to exclude from the reference all the channels that are close to the current channel. So, for every channel you exclude channels that are too close and channels that are too far away. I kind of wish I could put a diagram of this in the notes:

image

Any suggestions are welcomed.

src/spikeinterface/preprocessing/common_reference.py
h-mayorquin: Definitley! In writing avoid pronouns as much as you can.

src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -34,29 +56,35 @@ class CommonReferenceRecording(BasePreprocessor):
...
-        If True, output is verbose
+        Use in the local CAR implementation as the selecting annulus with the following format:
+
+        `(exclude radius, include radius)`
+
+        Where the exlude radius is the inner radius of the annulus and the include radius is the outer radius of the
+        annulus. The exclude radius is used to exclude channels that are too close to the reference channel and the
+        include radius delintes the outer boundary of the annulus whose role is to exclude channels
+        that are too far away.
+
```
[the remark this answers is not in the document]
h-mayorquin: Ah, OK. Yeah, maybe this should be avoid, is just that I felt that the introduction was too long already.

src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -1,25 +1,47 @@
...
     ----------
     recording: RecordingExtractor
         The recording extractor to be re-referenced
     reference: "global" | "single" | "local", default: "global"
-        If "global" then CMR/CAR is used either by groups or all channel way.
-        If "single", the selected channel(s) is remove from all channels. operator is no used in that case.
-        If "local", an average CMR/CAR is implemented with only k channels selected the nearest outside of a radius around each channel
+        If "global" the reference is the average or mean across all the channels.
+        If "single", the reference is a single channel or a list of them that need to be set with the `ref_channel_ids`.
+        If "local", the reference are the channels within an annulus that need to be set with the `local_radius`.
```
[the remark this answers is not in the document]
h-mayorquin: This is way better ,thanks.

src/spikeinterface/preprocessing/common_reference.py
h-mayorquin: Is this what you intended?
image

My auto-corrector did a number on it so if what we have right now it is not feel free to push to my branch with your intended change.

src/spikeinterface/preprocessing/common_reference.py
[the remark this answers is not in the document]
zm711: This is a bit tricky. Descent is actually a noun whereas descend is the verb. So when you perform gradient descent you are descending the gradient. In this case you would use the article for emphasis, I perform a gradient descent ( I am talking about one step in a series of steps). Reference itself is a verb (to reference something), so the default is in general to treat it like a normal verb where we either use the noun form (a reference) or a gerund which can be used as a noun (in this case referencing). But either way you convert this to a noun version. 

to be fair this is an area of argument in English which as a language has a strong vs weak system. For example neologisms should follow a weak system rather than a strong system. So the new word 'to gaslight' should have a past tense of gaslighted, but because light has an irregular past tense of lit, some people would prefer to have the past tense gaslit. 

But above all, if this is a super domain specific terminology (like you know for sure the field uses CAR/CMR without the article) then you would follow the domain even if it is wrong by standard grammar. I understand CMR as a concept, but I won't pretend to be an expert in the field. So I was correcting that based on standard grammar convention (where my gut tells me to say I perform gradient descent, but also tells me to say I used a common median reference to calculate the new reference value).

src/spikeinterface/preprocessing/common_reference.py
zm711: Yes that's what I wanted, except the "local" looks like it did a double set of spacing instead of single set. Tabbing wasn't working for me so I just did four spaces for each section. Not sure if it was your autocorrect or me mistyping.

src/spikeinterface/preprocessing/common_reference.py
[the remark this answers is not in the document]
zm711: To be fair I think people will see this docstring when trying to use the function, which means it's more important. The above docstring is more for the developers right (or people who read source), so I don't know if we need to move it up as long as it is clear in the user facing portion of the docs.

src/spikeinterface/preprocessing/common_reference.py
h-mayorquin: Should be corrected now.

src/spikeinterface/preprocessing/common_reference.py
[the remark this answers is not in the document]
h-mayorquin: Thanks for this explanation. I think I trust your gut in this one as we want to make the reading sound idiomatic if possible.

src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -34,29 +56,35 @@ class CommonReferenceRecording(BasePreprocessor):
...
+        Use in the local CAR implementation as the selecting annulus with the following format:
+
+        `(exclude radius, include radius)`
+
+        Where the exlude radius is the inner radius of the annulus and the include radius is the outer radius of the
+        annulus. The exclude radius is used to exclude channels that are too close to the reference channel and the
+        include radius delineates the outer boundary of the annulus whose role is to exclude channels
+        that are too far away.
+
     dtype: None or dtype
```
zm711, line 68: I figure you might be interested in this debate. On neo we were discussing that dtype is not the real type. It is in fact a 
`numpy.typing.DTypeLike`, but that requires being up to date with your numpy version and it actually super confusing for the end-user although clearer for the developer. I like just calling it dtype because I think most people understand that dtype means whatever can be accepted by the `numpy.dtype()` constructor, but it is something to keep in mind if we ever thing about typing analysis.

h-mayorquin: I think that type of very strict typing at this moment in python is more useful for people with application code.  A lot of the push for typing comes from people who mainly build apps (see fast api for example) and control the user envrionment to the degree that is impossible for us in library code. In the case of apps a typing as a schema / formal specification makes sense.

Library code like spikeinterface is very hard to type to that level. We can move the needled to add some typing so it helps the IDEs of users and works as an aid to documentation and that's kind of the goal to me.  From that perspective adding the more precise (DtypeLike) but at the same time more obscure typing does not make sense yet. 

In two three years one minimal is beyond 3.10 and we have better generics, the typing of array interfaces (e.g. numpy, cuda, dash, etc) is more worked out and spikeinterface data model stabilizes maybe we can start thinking on typing as a specification. Right now, I think we should prioritize clarify over formality.

For that reason I dislike abstracting away types away, that is defining somewhere else and then importig. The user should be able to see when they look at the signature what familiar stuff they can pass an expect.

src/spikeinterface/preprocessing/common_reference.py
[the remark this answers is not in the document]
h-mayorquin: If you call help you will see the whole docstring I think. What do you mean?

samuelgarcia (approved):

samuelgarcia:
Merci Zach and Ramon

alejoe91 (approved):

src/spikeinterface/preprocessing/common_reference.py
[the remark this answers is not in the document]
zm711: A lot of ides will auto display some of the docstring and so they will display this part at base without the full doc string. So I was referring to default behavior for people who rely on not calling help.

src/spikeinterface/preprocessing/common_reference.py
zm711: You’re preaching to the choir. That was the side of the discussion I took and the side we settled on for those exact reasons. But I was curious of your opinions. 

Thanks :)

h-mayorquin: I am happy we agree. Can you point out to the discussion neo out of curosity?

h-mayorquin:
@zm711 thanks again for all the help.

src/spikeinterface/preprocessing/common_reference.py
zm711: Sure [it's here](https://github.com/NeuralEnsemble/python-neo/pull/1334). It's spread every where, but the gist was that Neo doesn't currently have plans for type-checking so instead of being super strict with developer/abstract types that are strictly correct, that user facing helpful types would be better.

Issue #2312, opened 2023-12-07
Overflow problems with CAR (`common_reference`)


JoeZiminski:
I am having some problems with overflow error in `common_reference` when using the mean with neuropixels 2.0 recording. A MRE is below, this is not really a bug as is easily solved by converting the recording `dtype` to `float64` on load. However, it would make sense to perform some internal check if overflow will occur in this instance.

I think the problem is that [here](https://github.com/SpikeInterface/spikeinterface/blob/8e32955e0616b9e399c253aaf0602f3aa646dae4/src/spikeinterface/preprocessing/common_reference.py#L152) the output is self to `self.temp` which is `dtype` of the recording (in this case `int16`). It seems numpy is actually performing the computation is `int16` rather than casting to integer type on return, which is suprising. This was checked by setting `self.temp = np.zeros((all_traces.shape[0],), dtype=np.float64)` and the test passes.

```
import numpy as np
import spikeinterface.preprocessing as spre
from spikeinterface.core.generate import generate_recording

# Generate a recording, int16 with positive Vm (typical for NP recording)
num_channels = 384
recording = generate_recording(num_channels=num_channels, set_probe=True, durations=[0.5])
recording = spre.scale(recording, gain=100, offset=70)  # reduce offset ~ 30 to pass assert
recording = spre.astype(recording, "int16")


# Preprocess in SI
common_ref = spre.common_reference(recording, operator="average")
common_ref_data = common_ref.get_traces()

# Perform CAR with np.mean() casting to float64 under the hood.
recording_data = recording.get_traces()
means = np.mean(recording_data, axis=1).astype(np.int16)  # with dtype=np.int16 assert passes
test_common_ref_data = recording_data - means[:, np.newaxis]

assert np.allclose(common_ref_data, test_common_ref_data, rtol=0, atol=1)

# need to comment out the assert to run this block under failing conditions 
bad_idx = np.where(~np.isclose(common_ref_data, test_common_ref_data, rtol=0, atol=1))
if np.any(bad_idx):
    true_mean = means[bad_idx[0][0]]
    bad_mean = common_ref._recording_segments[0].operator_func(recording_data)[bad_idx[0][0]]
    print(true_mean)
    print(bad_mean)
```

PR #2356, opened 2023-12-19
Use memory efficient numpy for `ScaleRecordingSegment`

h-mayorquin:
I am taking a look at https://github.com/SpikeInterface/spikeinterface/issues/2312 and I realized the scale function is really memory hungry. This should make it better.

5 times reduction in memory usage reduction for the following script:

```python
from spikeinterface.core.generate import generate_recording, generate_recording_by_size

recording = generate_recording_by_size(full_traces_size_GiB=1.0)
recording = spre.scale(recording, gain=100, offset=70)  
recording.get_traces()
```

In main:

```
Total number of allocations: 46166
Total number of frames seen: 2741
Peak memory usage: 5.1 GiB
```

This branch:

```
Total number of allocations: 45810
Total number of frames seen: 2732
Peak memory usage: 1.1 GiB
```

Note that the trace should be 1.0 GiB by design.

### DIFF
--- a/src/spikeinterface/preprocessing/normalize_scale.py
+++ b/src/spikeinterface/preprocessing/normalize_scale.py
@@ -17,14 +17,15 @@ class ScaleRecordingSegment(BasePreprocessorSegment):
         self.offset = offset
         self._dtype = dtype
 
-    def get_traces(self, start_frame, end_frame, channel_indices):
+    def get_traces(self, start_frame, end_frame, channel_indices) -> np.ndarray:
         traces = self.parent_recording_segment.get_traces(start_frame, end_frame, channel_indices)
-        scaled_traces = traces * self.gain[:, channel_indices] + self.offset[:, channel_indices]
+        traces = np.multiply(traces, self.gain[:, channel_indices], out=traces)
+        traces = np.add(traces, self.offset[:, channel_indices], out=traces)
 
         if np.issubdtype(self._dtype, np.integer):
-            scaled_traces = scaled_traces.round()
+            traces = np.round(traces, out=traces)
 
-        return scaled_traces.astype(self._dtype)
+        return traces.astype(self._dtype, copy=False)
 
 
 class NormalizeByQuantileRecording(BasePreprocessor):
--- a/src/spikeinterface/preprocessing/normalize_scale.py
+++ b/src/spikeinterface/preprocessing/normalize_scale.py
@@ -18,7 +18,9 @@ class ScaleRecordingSegment(BasePreprocessorSegment):
         self._dtype = dtype
 
     def get_traces(self, start_frame, end_frame, channel_indices) -> np.ndarray:
-        traces = self.parent_recording_segment.get_traces(start_frame, end_frame, channel_indices)
+        # TODO when we are sure that BaseExtractors get_traces allocate their own buffer instead of just passing
+        # It along we should remove copies in preprocessors including the one in the next line
+        traces = self.parent_recording_segment.get_traces(start_frame, end_frame, channel_indices).copy()
         traces = np.multiply(traces, self.gain[:, channel_indices], out=traces, casting="unsafe")
         traces = np.add(traces, self.offset[:, channel_indices], out=traces, casting="unsafe")
 
--- a/src/spikeinterface/preprocessing/normalize_scale.py
+++ b/src/spikeinterface/preprocessing/normalize_scale.py
@@ -11,6 +11,7 @@ from ..core import get_random_data_chunks
 
 class ScaleRecordingSegment(BasePreprocessorSegment):
     # use by NormalizeByQuantileRecording/ScaleRecording/CenterRecording
+
     def __init__(self, parent_recording_segment, gain, offset, dtype):
         BasePreprocessorSegment.__init__(self, parent_recording_segment)
         self.gain = gain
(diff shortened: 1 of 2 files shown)


src/spikeinterface/preprocessing/normalize_scale.py
```diff
@@ -17,14 +17,15 @@ def __init__(self, parent_recording_segment, gain, offset, dtype):
         self.offset = offset
         self._dtype = dtype
 
-    def get_traces(self, start_frame, end_frame, channel_indices):
+    def get_traces(self, start_frame, end_frame, channel_indices) -> np.ndarray:
         traces = self.parent_recording_segment.get_traces(start_frame, end_frame, channel_indices)
-        scaled_traces = traces * self.gain[:, channel_indices] + self.offset[:, channel_indices]
+        traces = np.multiply(traces, self.gain[:, channel_indices], out=traces)
```
h-mayorquin, line 22: @alejoe91 @samuelgarcia  note that the following will also work in the same way and avoid extra memory allocations:
```python      
      traces *= self.gain[:, channel_indices]
      traces += self.offset[:, channel_indices]
```

If you prefer how that looks but we will need to make a comment like `#Do not move this` whereas the `out` argument makes it more clear (in my view) what is going on. But I don't feel strong about this.

Your call guys

h-mayorquin: Ah forget about it. I actually need the full call to keep backwards compatability with usage.

alejoe91 (approved):

zm711:
In #2189, the `.round()` was added to multiple preprocessing functions/classes. Should this change be made to all of these `.round()` additions @h-mayorquin ? Maybe in a different PR though.

h-mayorquin:
@zm711 I rather do this in a case per case basis. I also don't really like dtype coercing in preprocessing and don't want to think too much about it.I think I should use the library more till I undestand it.

src/spikeinterface/preprocessing/normalize_scale.py
```diff
@@ -17,14 +17,15 @@ def __init__(self, parent_recording_segment, gain, offset, dtype):
         self.offset = offset
         self._dtype = dtype
 
-    def get_traces(self, start_frame, end_frame, channel_indices):
+    def get_traces(self, start_frame, end_frame, channel_indices) -> np.ndarray:
         traces = self.parent_recording_segment.get_traces(start_frame, end_frame, channel_indices)
-        scaled_traces = traces * self.gain[:, channel_indices] + self.offset[:, channel_indices]
+        traces = np.multiply(traces, self.gain[:, channel_indices], out=traces, casting="unsafe")
```
samuelgarcia, line 22: I think we need a copy from the parent no ? If the parent is a memmap this will write to the file.
Am I wrong ?

samuelgarcia: I think I do not like the `out=traces` because if parent is writable then we modify it and if it is not writable it will bug.

samuelgarcia: get_tace() must create a buffer not use the one from the parent

h-mayorquin: All right. Let's create a copy here and discuss the following outside of this PR: 

This behavior -one allocation per `get_traces`- is inconsistent across the database. To the degree that we achieve it it is incidental instead of intentional. We should make it explicit. 

To make this clear, you want: if you have a long chain of say 5 preprocessing steps and the base trace is 1 GiB like this, then spikeinterface should allocate 5 GiB in total (which hopefully would be deallocated outside of the function scope).


My own view is that the `get_trace` of  `BaseExtractors` should allocate always allocate the buffer and not use the memmap as it comes but preprocessors should not as that buffer was already safely allocated before. What do you think?

samuelgarcia: I am 100% agree we should add the traces=None in get_traces() and this would be the unique buffer allocated.

h-mayorquin: OK, let me add a copy here and open an issue to add this to the backlog. Good catch.

alejoe91: @h-mayorquin we are checking this and this returns an `int16` if `traces` is `int16`.

The gain step should instead return floats (maybe 32) and the final step should be recasting. 

Her's a proposal:
```
scaled_traces = self.parent_recording_segment.get_traces(start_frame, end_frame, channel_indices).astype("float32", copy=True)
scaled_traces *= self.gain[:, channel_indices] # in-place
scaled_traces += self.offset[:, channel_indices] # in-place

if np.issubdtype(self._dtype, np.integer):
   scaled_traces = np.round(scaled_traces, out=scaled_traces)
return scaled_traces.astype(self._dtype, copy=False)
```

Can you check the memory consumption of this approach? Note that we have to go to float32 for the gain step!

h-mayorquin:
@alejoe91 
I did the suggested change.

This current version uses 2GiB of peak memory. So it is still a great improvement over the current state of affairs on the main branch and I think we should merge as it is.

I would note this as two future improvements or points of discussion for another PR future issues:
1) When the gain and offset are int we are casting unecessarily to float32. As this is a central class that is used by other pre-processing step as a composition step that's an improvement that I think we could have. Low priority probably as the low-hanging fruit has been picked here.
2) More dangerously, int32 to float32 casting is lossy. The mantissa of the latter is not large enough to represent all the possible values of int32 as far as I am aware. So maybe we want some size dependent casting.

alejoe91 (approved):

samuelgarcia:
Merci beaucoup.

PR #2357, opened 2023-12-19
Correct warning format in neo base extractors

h-mayorquin:
Small typos and format in the warning about t-start and sampling frequencies I found today while working on neuroconv CI.

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/neobaseextractor.py
+++ b/src/spikeinterface/extractors/neoextractors/neobaseextractor.py
@@ -462,7 +462,7 @@ class NeoBaseSortingExtractor(_NeoBaseExtractor, BaseSorting):
                 f"sampling_frequency = {sampling_frequency} \n"
                 "Corresponding to the following stream_id: \n"
                 f"stream_id = {stream_id} \n"
-                "To avoid this warning pass explicitly the sampling frequency or the stream_id"
+                "To avoid this warning pass explicitly the sampling frequency or the stream_id "
                 "when initializing the sorting extractor. \n"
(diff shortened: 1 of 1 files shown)


src/spikeinterface/extractors/neoextractors/neobaseextractor.py
```diff
@@ -462,7 +462,7 @@ def _infer_sampling_frequency_from_analog_signal(self, stream_id: Optional[str]
                 f"sampling_frequency = {sampling_frequency} \n"
```
zm711, line 462: No need to change, but did you know about: 

```python
f"{stream_id=}"
```

This blew my mind when I found out about that.

h-mayorquin: Yeah, I wish I could use it but is not in the oldest version of python that we support.

h-mayorquin: Ah, it seems that it is python 3.8. God to know, I will start to use.

h-mayorquin: https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging

zm711: Sweet. I was thinking I saw it somewhere in the Neo code base (where I discovered it) and Neo also supports 3.8 so I was about to have a freakout needing to track down where I saw it there.

alejoe91 (approved):

PR #2359, opened 2023-12-20
Patch for save_to_memory

yger:
Currently, the following code is leading to a crash

`a = rec.save(format='memory')
b = rec.save_to_memory()
`
a is working fine, but not b, because of an inconsistencies in the way to use the internal _save() function of base. This patch the problem, and make sure that when calling save_to_memory, if no format is specified, the memory one is added


yger:
Closing as this will be handled by #2365

PR #2362, opened 2023-12-20
Fix overflow problems with CAR

h-mayorquin:
This should fix https://github.com/SpikeInterface/spikeinterface/issues/2312.

The following code (reported by @JoeZiminski ) fails on main and works with this branch:

```python
import numpy as np
from spikeinterface.core.generate import generate_recording

# Generate a recording, int16 with positive Vm (typical for NP recording)
num_channels = 384
recording = generate_recording(num_channels=num_channels, set_probe=True, durations=[0.5])
recording = spre.scale(recording, gain=100, offset=70)  # reduce offset ~ 30 to pass assert
recording = spre.astype(recording, "int16")


# Preprocess in SI
common_ref = spre.common_reference(recording, operator="average")
common_ref_data = common_ref.get_traces()

# Perform CAR with np.mean() casting to float64 under the hood.
recording_data = recording.get_traces()
means = np.mean(recording_data, axis=1).astype(np.int16)  # with dtype=np.int16 assert passes
test_common_ref_data = recording_data - means[:, np.newaxis]

assert np.allclose(common_ref_data, test_common_ref_data, rtol=0, atol=1)

# need to comment out the assert to run this block under failing conditions 
bad_idx = np.where(~np.isclose(common_ref_data, test_common_ref_data, rtol=0, atol=1))
if np.any(bad_idx):
    true_mean = means[bad_idx[0][0]]
    bad_mean = common_ref._recording_segments[0].operator_func(recording_data)[bad_idx[0][0]]
    print(true_mean)
    print(bad_mean)
```

The problem I think is some magic with the operators for the grouping part. To solve this issue I separated the implementation for the non-grouping case from the one for grouping. The former is easier to implement and avoids the implicit casting with the current operator application. I also added some tests to see that the slicing with `channel_ids` work as intended. That said, the bug is still not solved for the grouping case but I think we should solve this for the most common use case and then address that when we have more time and the need if that ever comes to be.


src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -180,34 +180,60 @@ def __init__(
...
+            # need input trace
+            all_traces = self.parent_recording_segment.get_traces(start_frame, end_frame, slice(None))
+            all_traces = all_traces.astype(self.dtype)
+
+            self.temp = np.zeros((all_traces.shape[0],), dtype=all_traces.dtype)
+            _channel_indices = np.arange(all_traces.shape[1])
+            if channel_indices is not None:
+                _channel_indices = _channel_indices[channel_indices]
+
+            out_traces = np.zeros((all_traces.shape[0], _channel_indices.size), dtype=self.dtype)
```
alejoe91, line 215: @h-mayorquin 

After some discussion, probably the fix could be very simple and without the if-else `group_indices` with 2 changes:
1. In the `self.operator_func` remove `out=self.temp` and add `keepdims=True`
2. Do not pre allocate `out_traces`
3. Add astype at the end in any case.

I'd be happy to implement this in a separate PR!

h-mayorquin: Up to you guys,

I do believe there is a benefit on separting the non-grouping code to the grouping code.  The former can be made way simpler and will be easier for a future developer to understand, modify and optimize, the later has to be more complicated and will be more difficult. 

I recommend that you add a regression test whatever you decide to do though.

alejoe91: sounds good, then I'll push the same logic to the grouping branch, so it's fixed there too!

alejoe91:
@h-mayorquin I propagated the changes to the `groups` option, cleaned up and simplified that branch and extended tests.

There was a bug in the `local` part that was not caught by current tests ;)

h-mayorquin:
@alejoe91 the separation looks very good. I also think that the implementation you wrote for the new grouping is very clear. I did a small change because the function `in1d` will be deprecated soon:

https://numpy.org/devdocs/reference/generated/numpy.in1d.html

alejoe91 (approved):

PR #2363, opened 2023-12-20
Refactor recording tools

samuelgarcia:
Refactor recording tools : move some function from core_tools to recording_tools.


src/spikeinterface/core/__init__.py
```diff
@@ -79,14 +79,14 @@
...
 from .core_tools import (
-    write_binary_recording,
-    write_to_h5_dataset_format,
-    write_binary_recording,
     read_python,
     write_python,
 )
 from .job_tools import ensure_n_jobs, ensure_chunk_size, ChunkRecordingExecutor, split_job_kwargs, fix_job_kwargs
 from .recording_tools import (
+    write_binary_recording,
```
yger, line 87: twice the same import to write_binary_recording

yger (approved):
Looks good to me (modulo the duplicated import)

src/spikeinterface/core/recording_tools.py
```diff
@@ -2,9 +2,486 @@
...
+    ----------
+    recording: RecordingExtractor
+        The recording extractor object to be saved in .dat format
+    file_path: str or list[str]
+        The path to the file.
+    dtype: dtype or None, default: None
+        Type of the saved data
+        If True, file the ".raw" file extension is added if the file name is not a "raw", "bin", or "dat"
+    byte_offset: int, default: 0
+        Offset in bytes to for the binary file (e.g. to write a header)
```
zm711, line 95:
```suggestion
        Offset in bytes for the binary file (e.g. to write a header)
```

src/spikeinterface/core/recording_tools.py
```diff
@@ -2,9 +2,486 @@
...
+        Type of the saved data
+    verbose: bool, default: False
+        If True, output is verbose (when chunks are used)
+    auto_cast_uint: bool, default: True
+        If True, unsigned integers are automatically cast to int if the specified dtype is signed
+    {}
+
+    Returns
+    ---------
+    arrays: one arrays per segment
```
zm711, line 291:
```suggestion
    arrays: one array per segment
```

src/spikeinterface/core/recording_tools.py
```diff
@@ -2,9 +2,486 @@
...
+):
+    """
+    Save the traces of a recording extractor in an h5 dataset.
+
+    Parameters
+    ----------
+    recording: RecordingExtractor
+        The recording extractor object to be saved in .dat format
+    dataset_path: str
+        Path to dataset in h5 file (e.g. "/dataset")
```
zm711, line 362:
```suggestion
        Path to dataset in h5 file (e.g. "/dataset")
```

I don't know this format as well. This should be `for` unless you are trying to say that the path exists inside the name?

src/spikeinterface/core/recording_tools.py
```diff
@@ -2,9 +2,486 @@
...
+        the file is NOT closed after writing the binary data.
+    time_axis: 0 or 1, default: 0
+        If 0 then traces are transposed to ensure (nb_sample, nb_channel) in the file.
+        If 1, the traces shape (nb_channel, nb_sample) is kept in the file.
+    single_axis: bool, default: False
+        If True, a single-channel recording is saved as a one dimensional array
+    dtype: dtype, default: None
+        Type of the saved data
+    chunk_size: None or int, default: None
+        Number of chunks to save the file in. This avoid to much memory consumption for big files.
```
zm711, line 378:
```suggestion
        Number of chunks to save the file in. This avoids too much memory consumption for big files.
```

src/spikeinterface/core/recording_tools.py
```diff
@@ -2,9 +2,486 @@
...
+        If None and "chunk_memory" is given, the file is saved in chunks of "chunk_memory" MB
+    chunk_memory: None or str, default: "500M"
+        Chunk size in bytes must endswith "k", "M" or "G"
```
zm711, line 381:
```suggestion
        Chunk size in bytes must end with "k", "M" or "G"
```

zm711 (commented):
docstring cleanup while things are being moved

src/spikeinterface/core/__init__.py
```diff
@@ -79,14 +79,14 @@
...
+    write_to_h5_dataset_format,
+    write_binary_recording,
```
alejoe91, line 89:
```suggestion
```

src/spikeinterface/core/recording_tools.py
alejoe91: Hi Zach, I don't understand the comment! Ot seems you are not changing anything in the suggesiton

alejoe91 (approved):

src/spikeinterface/core/recording_tools.py
zm711: Hey Alessio (and Happy New Year by the way!) I was saying that the standard way to say this is:

Path to dataset __for__ the h5 file (e.g. "/dataset"), 

but that since I don't know this format maybe you meant that the name was inside of the file path like:

(e.g."nameofdata/dataset") in which case the using the word `in` is okay. Since I never use this format I didn't want to correct the preposition if that preposition was actually import for the use of the parameter. 

Does that make sense? (ie typically the preposition used is wrong unless specifically meaning that the Path is inside of the h5 file format somewhere.

alejoe91: Happy new year to you too! 

Got it! I think that in this case "Path to dataset **in** the h5 file (e.g. "/dataset")" would be the best :)

src/spikeinterface/core/recording_tools.py
```diff
@@ -2,9 +2,486 @@
...
+):
+    """
+    Save the traces of a recording extractor in an h5 dataset.
+
+    Parameters
+    ----------
+    recording: RecordingExtractor
+        The recording extractor object to be saved in .dat format
+    dataset_path: str
+        Path to dataset in h5 file (e.g. "/dataset")
```
alejoe91, line 362:
```suggestion
        Path to dataset in the h5 file (e.g. "/dataset")
```

src/spikeinterface/core/recording_tools.py
alejoe91: done :)

PR #2364, opened 2023-12-20
Add `rename_channels` method to recording extractors

h-mayorquin:
Easier for users to discover this method with tab completion instead of having to find out by themselves that they need to use channel slice of which this is a particular case.

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -548,6 +548,23 @@ class BaseRecording(BaseRecordingSnippets):
             if time_vector is not None:
                 np.save(folder / f"times_cached_seg{segment_index}.npy", time_vector)
 
+    def rename_channels(self, new_channel_ids: list | np.array | tuple):
+        """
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/baserecording.py
```diff
@@ -548,6 +548,23 @@ def _extra_metadata_to_folder(self, folder):
...
                 np.save(folder / f"times_cached_seg{segment_index}.npy", time_vector)
 
+    def rename_channels(self, new_channel_ids: list | np.array | tuple):
+        """
+        Returns a new recording object with renamed channel ids.
+
+        Parameters
+        ----------
+        new_channel_ids : list or np.array or tuple
+            The new channel ids. They are mapped possitionally to the old channel ids.
```
zm711, line 558:
```suggestion
            The new channel ids. They are mapped positionally to the old channel ids.
```

h-mayorquin: Cheers!

alejoe91 (approved):

Issue #2281, opened 2023-12-01
add option 'with_recording' to waveform extractor select_units 


LiesDec:
It would be useful to have the option 'with_recording' also in the method 'select_units' for the waveformextractor. 

On top of adding the parameter to [L766](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/waveform_extractor.py#L766) and [L827](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/waveform_extractor.py#L827) you would also need to copy the recording_info folder. 
For example modify lines [L801-L802 ](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/waveform_extractor.py#L801-L802) to:

```
if self.has_recording():
    self.recording.dump(new_folder / "recording.json", relative_to=relative_to)
else:
    shutil.copytree(self.folder / "recording_info", new_folder / "recording_info")
```

There might be an extra check needed here as I understand that the 'recording_info' is not created in all spike interface versions.

LiesDec:
@alejoe91  I see you labeled the issue as enhancement, but actually in my case it's more like a bug.

If I use the `select_units` method on a `we` object that has no recording attached (but has the 'recording_info' folder) it fails as it will try to load the recording which is not there. 

```
we_clean = we.select_units(keep_unit_ids, new_folder=datapath / 'raw' / animal / session_dt/ sorting_folder_name /'waveforms_clean')
```

```
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
Input In [7], in <module>
----> 1 we_clean = we.select_units(keep_unit_ids, new_folder=datapath / 'raw' / animal / session_dt/ sorting_folder_name /'waveforms_clean')

File ~/spikeinterface/src/spikeinterface/core/waveform_extractor.py:834, in WaveformExtractor.select_units(self, unit_ids, new_folder, use_relative_path)
    831             json.dump(check_json(new_sparsity.to_dict()), f)
    833     # we = WaveformExtractor.load(new_folder, with_recording = False)
--> 834     we = WaveformExtractor.load(new_folder)
    835 elif self.format == "zarr":
    836     raise NotImplementedError(
    837         "For zarr format, `select_units()` to a folder is not supported yet. "
    838         "You can select units in two steps:\n"
    839         "1. `we_new = select_units(unit_ids, new_folder=None)`\n"
[...]

PR #2368, opened 2023-12-21
WaveformExtractor.select_units also functional if  we.has_recording()=False

LiesDec:
This is a pull request following up on issue #2281.
The proposed changes enable the method WaveformExtractor.select_units to also work if the WaveformExtractor had not recording.

### DIFF
--- a/src/spikeinterface/core/waveform_extractor.py
+++ b/src/spikeinterface/core/waveform_extractor.py
@@ -1995,7 +1995,6 @@ class BaseWaveformExtractorExtension:
                 elif ext_data_file.suffix == ".pkl":
                     ext_data = pickle.load(ext_data_file.open("rb"))
                 else:
-                    print(f"{ext_data_file.name} not created by spikeinterface, skipping this file")
                     continue
                 self._extension_data[ext_data_name] = ext_data
         elif self.format == "zarr":
--- a/src/spikeinterface/core/waveform_extractor.py
+++ b/src/spikeinterface/core/waveform_extractor.py
@@ -1994,6 +1994,9 @@ class BaseWaveformExtractorExtension:
                     ext_data = pd.read_csv(ext_data_file, index_col=0)
                 elif ext_data_file.suffix == ".pkl":
                     ext_data = pickle.load(ext_data_file.open("rb"))
+                else:
+                    print({} "not created by spikeinterface, skipping this file".format(ext_data_file.name))
+                    continue
                 self._extension_data[ext_data_name] = ext_data
         elif self.format == "zarr":
             for ext_data_name in self.extension_group.keys():
--- a/src/spikeinterface/core/waveform_extractor.py
+++ b/src/spikeinterface/core/waveform_extractor.py
@@ -800,8 +800,8 @@ class WaveformExtractor:
 
                 if self.has_recording():
                     self.recording.dump(new_folder / "recording.json", relative_to=relative_to)
-                else:
-                    shutil.copytree(self.folder / "recording_info", new_folder / "recording_info")
+
+                shutil.copytree(self.folder / "recording_info", new_folder / "recording_info")
 
                 sorting.dump(new_folder / "sorting.json", relative_to=relative_to)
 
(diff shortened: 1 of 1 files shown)


LiesDec:
The fix in the second commit was necessary for me (and potentially other users) because in the lab we store our data on a NAS sinology data server. These create @eaDir files to handle user specific views. These files need to be skipped when walking through a directory like what happens in [ _load_extension_data](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/waveform_extractor.py#L1973).

src/spikeinterface/core/waveform_extractor.py
```diff
@@ -1990,6 +1994,9 @@ def _load_extension_data(self):
                     ext_data = pd.read_csv(ext_data_file, index_col=0)
                 elif ext_data_file.suffix == ".pkl":
                     ext_data = pickle.load(ext_data_file.open("rb"))
+                else:
+                    print("{} not created by spikeinterface, skipping this file".format(ext_data_file.name))
```
alejoe91, line 1998:
```suggestion
                    print(f"{ext_data_file.name} not created by spikeinterface, skipping this file")
```
We prefer to use f strings throughout the code!

src/spikeinterface/core/waveform_extractor.py
```diff
@@ -800,6 +800,9 @@ def select_units(self, unit_ids, new_folder=None, use_relative_path: bool = Fals
 
                 if self.has_recording():
                     self.recording.dump(new_folder / "recording.json", relative_to=relative_to)
+                else:
+                    shutil.copytree(self.folder / "recording_info", new_folder / "recording_info")
```
samuelgarcia, line 804: I would remove the else. We should copy this file for all situation no ?

src/spikeinterface/core/waveform_extractor.py
```diff
@@ -1990,6 +1994,9 @@ def _load_extension_data(self):
                     ext_data = pd.read_csv(ext_data_file, index_col=0)
                 elif ext_data_file.suffix == ".pkl":
                     ext_data = pickle.load(ext_data_file.open("rb"))
+                else:
+                    print(f"{ext_data_file.name} not created by spikeinterface, skipping this file")
+                    continue
```
samuelgarcia, line 1999: Not sure to see the use case.
Why would there be a file not created by spikeinterface is this subfolder ?

samuelgarcia:
Hi @LiesDec.
Merci beaucoup.
I made some coments.

alejoe91:
> Hi @LiesDec. Merci beaucoup. I made some coments.

BEcause of this: https://github.com/SpikeInterface/spikeinterface/pull/2368#issuecomment-1866194157

src/spikeinterface/core/waveform_extractor.py
samuelgarcia: oups sorry. I haven't read in detail the comments.

src/spikeinterface/core/waveform_extractor.py
LiesDec: @samuelgarcia I thought the file is copied when you dump the recording?

alejoe91: I agree with @samuelgarcia 

Please remove the else and copy recording info in both cases (with and without recording)

LiesDec: @samuelgarcia @alejoe91  Allright, done :)

src/spikeinterface/core/waveform_extractor.py
```diff
@@ -1990,6 +1994,9 @@ def _load_extension_data(self):
                     ext_data = pd.read_csv(ext_data_file, index_col=0)
                 elif ext_data_file.suffix == ".pkl":
                     ext_data = pickle.load(ext_data_file.open("rb"))
+                else:
+                    print(f"{ext_data_file.name} not created by spikeinterface, skipping this file")
```
alejoe91, line 1998:
```suggestion
```
Removing the print but keeping the logic!

alejoe91 (approved):

samuelgarcia (approved):

PR #2369, opened 2023-12-21
Add stream_folders path to OpenEphysBinaryRecordingExtractor

LiesDec:
For other analyses I require the direct data path of the raw data of on Binary Open Ephys recording (continuous.dat). 
For easy acces I propose to add it to the _kwargs property of the OpenEphysBinaryRecordingExtractor. 
Thanks! :)

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/openephys.py
+++ b/src/spikeinterface/extractors/neoextractors/openephys.py
@@ -217,12 +217,14 @@ class OpenEphysBinaryRecordingExtractor(NeoBaseRecordingExtractor):
                 self.set_property("inter_sample_shift", sample_shifts)
 
         # load synchronized timestamps and set_times to recording
-        if load_sync_timestamps:
-            recording_folder = Path(folder_path) / record_node
-            for segment_index in range(self.get_num_segments()):
-                stream_folder = (
-                    recording_folder / f"experiment{exp_id}" / f"recording{segment_index+1}" / "continuous" / oe_stream
-                )
+        recording_folder = Path(folder_path) / record_node
+        stream_folders = []
+        for segment_index in range(self.get_num_segments()):
+            stream_folder = (
+                recording_folder / f"experiment{exp_id}" / f"recording{segment_index+1}" / "continuous" / oe_stream
+            )
+            stream_folders.append(stream_folder)
+            if load_sync_timestamps:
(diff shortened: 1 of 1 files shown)


src/spikeinterface/extractors/neoextractors/openephys.py
```diff
@@ -242,6 +242,7 @@ def __init__(
                 load_sync_channel=load_sync_channel,
                 load_sync_timestamps=load_sync_timestamps,
                 experiment_names=experiment_names,
+                stream_folder=stream_folder,
```
alejoe91, line 245:
```suggestion
```
Hi @LiesDec 

The `self._kwargs` contains only the argument that are passed to the init. If you want to stor the `stream_folder`, you can store it in a `self._stream_folder` variable instead

alejoe91 (changes requested):
See comment ;)

src/spikeinterface/extractors/neoextractors/openephys.py
LiesDec: I changed it in a new commit :)

src/spikeinterface/extractors/neoextractors/openephys.py
```diff
@@ -244,6 +244,7 @@ def __init__(
                 experiment_names=experiment_names,
             )
         )
+        self._stream_folder = stream_folder
```
alejoe91, line 247: Can you move this before that `self._kwargs` logic? We try to keep that at the end of the init so we know where to look for it!

LiesDec: Of course :) Good to learn these conventions, thanks for your patience.

alejoe91 (changes requested):
@LiesDec 

I noticed that `stream_folder` is not defined if `load_timestamps=False`.
See: https://github.com/SpikeInterface/spikeinterface/blob/41a75e2502a00f1f8b0f958b3433ad1787c40b87/src/spikeinterface/extractors/neoextractors/openephys.py#L223

Can you move that line before the if?

alejoe91:
@LiesDec actually the stream folders should be a list, because there can be multiple recording segments (`recording*` folders in Open Ephys) associated to the same stream. 

I'll push a patch

alejoe91 (approved):

Issue #2373, opened 2023-12-24
SilencePeriodsRecording throws an error 


nadirbt:
Everytime I run a recording that was processed by the silence period function, it throws this error: 

get_random_data_chunks() got an unexpected keyword argument 'noise_levels'

I think this is because the definition of the class initializer is different from the definition of the self._kwargs argument.

yger:
Can you share a minimal example to reproduce the error? I'll have a look asap

nadirbt:
I was able to replicate the bug using this code:

import spikeinterface.full as si
from pathlib import Path

global_job_kwargs = dict(n_jobs=4, chunk_duration="1s")
si.set_global_job_kwargs(**global_job_kwargs)

local_path = Path(__file__).parent / './mearec_test_10s.h5'
recording, _ = si.read_mearec(local_path)

rec_preprocessed = si.silence_periods(recording, list_periods = [(0, 10)], mode='noise')

sorting_MS5 = si.run_sorter(sorter_name='mountainsort5', recording = rec_preprocessed, remove_existing_folder=True)

yger:
This should be fix with #2375. Indeed, noise levels were not saved as internal attributes, sorry for that

nadirbt:
Thank you for fixing this. Has the change been released yet?

yger:
No, this is only in the main branch (that you can install via git), not yet in a release

PR #2375, opened 2023-12-27
Fix for Silence periods (saving noise levels)

yger:
This is a fix for #2373

### DIFF
--- a/src/spikeinterface/preprocessing/silence_periods.py
+++ b/src/spikeinterface/preprocessing/silence_periods.py
@@ -66,7 +66,8 @@ class SilencedPeriodsRecording(BasePreprocessor):
                 ), "Intervals should not overlap"
 
         if mode in ["noise"]:
-            noise_levels = get_noise_levels(recording, return_scaled=False, concatenated=True, **random_chunk_kwargs)
+            if noise_levels is None:
+                noise_levels = get_noise_levels(recording, return_scaled=False, concatenated=True, **random_chunk_kwargs)
         else:
             noise_levels = None
 
--- a/src/spikeinterface/preprocessing/silence_periods.py
+++ b/src/spikeinterface/preprocessing/silence_periods.py
@@ -74,7 +74,6 @@ class SilencedPeriodsRecording(BasePreprocessor):
         else:
             noise_levels = None
 
-        self.noise_levels = noise_levels
         BasePreprocessor.__init__(self, recording)
         for seg_index, parent_segment in enumerate(recording._recording_segments):
             periods = list_periods[seg_index]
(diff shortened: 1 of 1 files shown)


alejoe91 (approved):

src/spikeinterface/preprocessing/silence_periods.py
```diff
@@ -79,9 +78,7 @@ def __init__(self, recording, list_periods, mode="zeros", **random_chunk_kwargs)
             rec_segment = SilencedPeriodsRecordingSegment(parent_segment, periods, mode, noise_levels)
             self.add_recording_segment(rec_segment)
 
-        self._kwargs = dict(
-            recording=recording.to_dict(), list_periods=list_periods, mode=mode, noise_levels=noise_levels
-        )
+        self._kwargs = dict(recording=recording, list_periods=list_periods, mode=mode, noise_levels=noise_levels)
```
alejoe91, line 81:
```suggestion
        self._kwargs = dict(recording=recording, list_periods=list_periods, mode=mode, noise_levels=noise_levels)
```
@yger I don't think this fixes the issue. Actually the `self.noise_levels` in the `SilenceRecording` is not used.
I think that you should add `noise_levels=None` to the `__init__`, so that if it's already computed it is not recomputed.

Accordingly, this should change:
```
if mode in ["noise"]:
    noise_levels = get_noise_levels(recording, return_scaled=False, concatenated=True, **random_chunk_kwargs)
else:
    noise_levels = None
```
to:
```
if mode in ["noise"]:
    if noise_levels is None:
        noise_levels = get_noise_levels(recording, return_scaled=False, concatenated=True, **random_chunk_kwargs)
else:
    noise_levels = None
```

alejoe91 (changes requested):
@yger see my comment. I believe that your change doesn't fix the issue

yger:
I'll have a look later, but with my fix, the big is gone. Maybe this does not cover everything, but at least the reported issue...

src/spikeinterface/preprocessing/silence_periods.py
yger: I  made the changes requested, but the code was already working, even with multiple jobs. Note that I thought the noise_levels are cached now. So if they have been already computed, they might be reloaded no? Aren't they saved as properties of the recording?

alejoe91: That is true! Then we should have remove it from the `self._kwargs`, since caching/reloading would be done directly by the `get_noise_levels`. No?

yger: Yes, but I think that numerous function are requesting noise_levels in the init/kwargs, no? Should it be a more general refactoring? This could be discussed

alejoe91 (approved):

PR #2377, opened 2023-12-29
Add zarrrecordingextractor.py for backward compatibility

alejoe91:
Plus some default values in common ref docstrings

### DIFF
--- a/src/spikeinterface/core/zarrrecordingextractor.py
+++ b/src/spikeinterface/core/zarrrecordingextractor.py
@@ -1,0 +1,5 @@
+"""
+This file is for backwards compatibility with the old zarr extractor file structure.
+"""
+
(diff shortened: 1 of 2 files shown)


h-mayorquin (approved):
Is this for someone importing zarr recording extractor directly from the old file:

```
from spikeinterface.core.zarrrecordingextractor import ZarrRecordingExtractor
```
?

alejoe91:
Indeed. This is a case when a user saved a JSON or pickle file before the refactoring and then uses `load_extractor`, which tries to instantiate the `spikeinterface.core.zarrrecordingextractor.ZarrRecordingExtractor`

h-mayorquin:
Thanks for the info.

samuelgarcia:
ok.

Issue #2380, opened 2024-01-02
Optimization of get_noise_levels (and thus get_random_data_chunks)


yger:
Actually, get_noise_levels makes internal use of get_random_data_chunks that is not parallel. There is thus room for improvements, via parallelization. This could be achieved at low cost by using the run_node_pipeline machinery, assuming some patch like #2011 would be accepted. If this is the case (or we can discuss somehting in line and similar), one could simply provide a list of fake "peak" times, some temporal window, and use the WaveformDense object of the pipeline to get data in parallel. Adding an extra job_kwargs to get_noise_levels would then speed up everything. @samuelgarcia what do you think?

PR #2382, opened 2024-01-02
Extracting traces in parallel to speed up get_noise_levels (or any other traces related functions)

yger:
We can extent the pipeline machinery as explained in #2380 in order to get data chunks in parallel. The list of chunks passed to the ChunkRecordingExecutor can be customized accordingly, to avoid looping over unnecessary chunks


yger:
Is it normal than the behavior of get_random_data_chunks, in main, can allow overlapping chunks? This seems weird to me, and thus the tests are not passing because of some behavior I would not expect. I mean, intuitively when I wan to select N random chunks in a small recording, I expect them to be non-overlapping (to avoid biasing the stats) @alejoe91 @samuelgarcia . But currently, this is not the case....

alejoe91:
> Is it normal than the behavior of get_random_data_chunks, in main, can allow overlapping chunks? This seems weird to me, and thus the tests are not passing because of some behavior I would not expect. I mean, intuitively when I wan to select N random chunks in a small recording, I expect them to be non-overlapping (to avoid biasing the stats) @alejoe91 @samuelgarcia . But currently, this is not the case....

I think that when possible the chunks should be non overlapping indeed

src/spikeinterface/core/recording_tools.py
```diff
@@ -14,6 +14,7 @@ def get_random_data_chunks(
     concatenated=True,
     seed=0,
     margin_frames=0,
+    job_kwargs={},
```
alejoe91, line 17: I would prefer having `**job_kwargs` and automatically populate the dosctring and propagate the arguments.. no?

h-mayorquin: Also, remember that setting default values with mutable objects is a bad practice that lends itself to hard to debug errors:
https://nedbatchelder.com/blog/202311/say_it_again_values_not_expressions.html

yger: Thanks for the feedback, that I'll consider asap. For now this is just a draft, because I realized that several options of the current API (no concatenation / margin frames) will not be straightforward with this current implementation. However, if there is an interest this is doable.

src/spikeinterface/core/recording_tools.py
```diff
@@ -46,41 +47,39 @@ def get_random_data_chunks(
...
+    data = run_traces_pipeline(
+        recording, job_kwargs, all_chunks=all_chunks, return_scaled=return_scaled, squeeze_output=True
+    )
     if concatenated:
-        return np.concatenate(chunk_list, axis=0)
+        return data
     else:
-        return chunk_list
+        num_chunks = recording.get_num_segments() * num_chunks_per_segment
+        return np.split(data, chunk_size * np.arange(1, num_chunks))
```
alejoe91, line 82: @yger you should add `{}` in the docstring and reformat the docs using `_shared_job_kwargs_doc`

See for example the [`extract_waveforms`](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/waveform_extractor.py#L1690) implementation

samuelgarcia:
Salut Pierre.
I am OK with the idea but I am not sure to like the implementation.
The new `run_traces_pipeline` is more or less a `ChunkRecordingExecutor` that return traces.
The is not a super good idea because traces of pickle a transimited to main process which make lot a memory bandwith.
I will try to make another implementation with sharemem and without the pipeline mechanism.
The pipeline module is for peak or spikes adding functionnality for random chunk getter make the module more fuzy. no ?

yger:
I am open to any suggestion. I just wanted to highlight the potential speedup there, especially for slow/remote I/O. Thanks a lot for having a look into that !

yger:
This has been done in #3359 . Can be closed

Issue #1989, opened 2023-09-13
Unify slicing behavior in SpikeInterface


alejoe91:
Related to #1881

Let's discuss how we should handle the behavior of `start_frame`/`end_frame`/`channel_ids` in the `get_traces()`.

Some proposals:
1. Be restrictive: 
- `end_frame` > `start_frame`: passing `start_frame == end_frame` will raise an Exception
- `end_frame` < `num_samples`
2. Numpy-like
...

h-mayorquin:
OK, so initially I wanted to follow numpy as much as possible. I had two cases in mind. Out of bounds slicing where we should do clipping and `start_frame=end_frame` slicing wher we should return an empty array.

```python

In [1]: x = np.ones(10)[15:13]

In [2]: x[15:13]
Out[2]: array([], dtype=float64)

In [3]: x[5:3]
Out[3]: array([], dtype=float64)

In [4]: x[10:10]
Out[4]: array([], dtype=float64)
```

Note that standard lists `[ ]` in python behave in exactly the same way 

Using two arguments: 
1) We already load most of our data as numpy memmaps (so we don't need checks) and then we also interact with scipy and numpy functions (and therefore should folllow their specification as they probably expect input with numpy like qualities).
2) We don't need to come up with a standard ourselves. "Let's just do what numpy does" is easy to follow and avoids the cost of coordinating /agreeing / discssuing between us if we disagree about a specific rule. We can't oursource policing to someone else.

But there is actually a standard that I think we should follow:
https://data-apis.org/array-api/latest/

And this is what it has to say about indexing (quoting from the relevant sections):
https://data-apis.org/array-api/latest/API_specification/indexing.html

1) Out of bounds check: This specification does not require bounds checking. The behavior for out-of-bounds integer indices is left unspecified.
> Rationale: this is consistent with bounds checking for integer indexing; the behavior of out-of-bounds indices is left unspecified. Implementations may choose to clip (consistent with Python list slicing semantics), raise an exception, return junk values, or some other behavior depending on device requirements and performance considerations.

2) `start_frame=end_frame` slicing: the specification says the following:
> If i equals j, a slice must return an empty array, whose axis (dimension) size along the indexed axis is 0.

I think we should just follow that. We can start second guessing standards because we think we know better but I think we increase costs for interaction with the ecosystem and we increase costs of discussing between us if a non-standard rule were we disgagree comes in the future.

JoeZiminski:
[...]

PR #2385, opened 2024-01-05
Add DecimateRecording

TomBugnon:
Hi,

I needed this because I want to downsample while keeping track of the actual underlying frame ids. In my case I don't care about aliasing etc because I will apply it on a smoothed signal, but it might be a good idea to add some safeguards... Maybe changing the name to avoid confusion with the scipy.decimate function?

Best
Tom

### DIFF
--- a/src/spikeinterface/preprocessing/decimate.py
+++ b/src/spikeinterface/preprocessing/decimate.py
@@ -121,6 +121,8 @@ class DecimateRecordingSegment(BaseRecordingSegment):
             start_frame = 0
         if end_frame is None:
             end_frame = self.get_num_samples()
+        end_frame = min(end_frame, self.get_num_samples())
+        start_frame = min(start_frame, self.get_num_samples())
 
         # Account for offset and end when querying parent traces
         parent_start_frame = self._decimation_offset + start_frame * self._decimation_factor
--- a/src/spikeinterface/preprocessing/decimate.py
+++ b/src/spikeinterface/preprocessing/decimate.py
@@ -112,7 +112,9 @@ class DecimateRecordingSegment(BaseRecordingSegment):
         self._dtype = dtype
 
     def get_num_samples(self):
-        return len(range(self._decimation_offset, self._parent_segment.get_num_samples(), self._decimation_factor))
+        parent_n_samp = self._parent_segment.get_num_samples()
+        assert self._decimation_offset < parent_n_samp  # Sanity check (already enforced). Formula changes otherwise
+        return np.ceil((parent_n_samp - self._decimation_offset) / self._decimation_factor)
 
     def get_traces(self, start_frame, end_frame, channel_indices):
         if start_frame is None:
--- a/src/spikeinterface/preprocessing/decimate.py
+++ b/src/spikeinterface/preprocessing/decimate.py
@@ -19,9 +19,11 @@ class DecimateRecording(BasePreprocessor):
     Parameters
     ----------
     recording : Recording
-        The recording extractor to be re-referenced
+        The recording extractor to be decimated.
     decimation_frame_step : int
-    decimation_frame_start : int, default 0
+        Step between successive frames sampled from the parent recording.
+    decimation_frame_start : int, default: 0
+        Index of first frame sampled from the parent recording.
 
     Returns
     -------
--- a/src/spikeinterface/preprocessing/decimate.py
+++ b/src/spikeinterface/preprocessing/decimate.py
@@ -19,20 +19,22 @@ class DecimateRecording(BasePreprocessor):
     Parameters
     ----------
     recording : Recording
-        The recording extractor to be decimated.
+        The recording extractor to be decimated. Each segment is decimated independently.
     decimation_factor : int
         Step between successive frames sampled from the parent recording.
+        The same decimation factor is applied to all segments from the parent recording.
     decimation_offset : int, default: 0
         Index of first frame sampled from the parent recording.
         Expecting `decimation_offset` < `decimation_factor`, and `decimation_offset` < parent_recording.get_num_samples()
         to ensure that the decimated recording has at least one frame. Consider combining DecimateRecording
         with FrameSliceRecording for fine control on the recording start and end frames.
+        The same decimation offset is applied to all segments from the parent recording.
 
     Returns
     -------
     decimate_recording: DecimateRecording
-        The decimated recording extractor object. The full traces of the child recording correspond
-        to the parent traces as follows:
+        The decimated recording extractor object. The full traces of the child recording segment
+        correspond to the traces of the parent segment as follows:
             ```<decimated_traces> = <parent_traces>[<decimation_offset>::<decimation_factor>]```
 
     """
@@ -47,15 +49,17 @@ class DecimateRecording(BasePreprocessor):
     ):
         # Original sampling frequency
         self._orig_samp_freq = recording.get_sampling_frequency()
-        parent_nsamp = recording.get_num_samples()
         if not isinstance(decimation_factor, int) or decimation_factor <= 0:
             raise ValueError(f"Expecting strictly positive integer for `decimation_factor` arg")
         self._decimation_factor = decimation_factor
         if not isinstance(decimation_offset, int) or decimation_factor < 0:
             raise ValueError(f"Expecting positive integer for `decimation_factor` arg")
-        if decimation_offset >= decimation_factor or decimation_offset >= parent_nsamp:
+        parent_min_n_samp = min(
+            [recording.get_num_samples(segment_index) for segment_index in range(recording.get_num_segments())]
+        )
+        if decimation_offset >= decimation_factor or decimation_offset >= parent_min_n_samp:
             raise ValueError(
-                f"Expecting `decimation_offset` < `decimation_factor` and `decimation_offset` < recording.get_num_samples(). "
+                f"Expecting `decimation_offset` < `decimation_factor` and `decimation_offset` < parent_segment.get_num_samples() for all segments. "
                 f"Consider combining DecimateRecording with FrameSliceRecording for fine control on the recording start/end frames."
             )
         self._decimation_offset = decimation_offset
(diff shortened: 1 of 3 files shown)


src/spikeinterface/preprocessing/decimate.py
```diff
@@ -0,0 +1,131 @@
...
+    Important: This uses simple array slicing for decimation rather than eg scipy.decimate.
+    This might introduce aliasing, or skip across signal of interest.
+    Consider  spikeinterface.preprocessing.ResampleRecording for safe resampling.
+
+    Parameters
+    ----------
+    recording : Recording
+        The recording extractor to be re-referenced
+    decimation_frame_step : int
+    decimation_frame_start : int, default 0
```
alejoe91, line 24: Can you add descriptions?

alejoe91, line 24:
```suggestion
    decimation_frame_step : int
    decimation_frame_start : int, default: 0
```
We use `default: ...`  for default values

src/spikeinterface/preprocessing/tests/test_decimate.py
```diff
@@ -0,0 +1,36 @@
...
+def test_decimate():
+    rec = generate_recording()
+
+    N = 101
+    rec = NumpyRecording([np.arange(N).reshape(N,1)], 1)
+    parent_traces = rec.get_traces()
+
+    for decimation_frame_start, decimation_frame_step in itertools.product(
+        [0, 1, 9, 10, 11, 200],
+        [1, 9, 10, 11, 200]
```
alejoe91, line 20: It would be great to use `@pytest.mark.parametrize` to expand tests, see for example [here](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/preprocessing/tests/test_interpolate_bad_channels.py#L84-L88)

alejoe91 (changes requested):
Thanks @TomBugnon 

A few small comments!

src/spikeinterface/preprocessing/tests/test_decimate.py
TomBugnon: got lazy sorry

alejoe91:
Do you mind also running pre-commit (or just black) locally? Since the PR is from an organization, the bot cannot push to it

samuelgarcia:
Super. Merci Tom.
Did you test corner case like get_trace on borders with one sample at the end ?

src/spikeinterface/preprocessing/decimate.py
```diff
@@ -0,0 +1,133 @@
...
+        )
+        self._parent_segment = parent_recording_segment
+        self._decimation_frame_step = decimation_frame_step
+        self._decimation_frame_start = decimation_frame_start
+        self._dtype = dtype
+
+    def get_num_samples(self):
+        return len(
+            range(self._decimation_frame_start, self._parent_segment.get_num_samples(), self._decimation_frame_step)
+        )
```
samuelgarcia, line 111: This is smart but maybe something more explicit with rounding or ceiling integer could be more usefull for debugging no ?

src/spikeinterface/preprocessing/decimate.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def get_traces(self, start_frame, end_frame, channel_indices):
+        if start_frame is None:
+            start_frame = 0
+        if end_frame is None:
+            end_frame = self.get_num_samples()
+
+        # Account for offset and end when querying parent traces
+        parent_start_frame = self._decimation_frame_start + start_frame * self._decimation_frame_step
+        parent_end_frame = parent_start_frame + (end_frame - start_frame) * self._decimation_frame_step
```
samuelgarcia, line 121: How can we be sure that this is not longer than the parent size ?
this is also related to the get_nump_samples() to be sure of ceiling vs rouding.

src/spikeinterface/preprocessing/decimate.py
```diff
@@ -0,0 +1,133 @@
...
+        # Original sampling frequency
+        self._orig_samp_freq = recording.get_sampling_frequency()
+        if not isinstance(decimation_frame_step, int) or decimation_frame_step <= 0:
+            raise ValueError(f"Expecting strictly positive integer for `decimation_frame_step` arg")
+        self._decimation_frame_step = decimation_frame_step
+        if not isinstance(decimation_frame_start, int) or decimation_frame_step < 0:
+            raise ValueError(f"Expecting positive integer for `decimation_frame_step` arg")
+        self._decimation_frame_start = decimation_frame_start
+        resample_rate = self._orig_samp_freq / self._decimation_frame_step
+        self._sampling_frequency = resample_rate
```
alejoe91, line 54:
```suggestion
```

src/spikeinterface/preprocessing/decimate.py
```diff
@@ -0,0 +1,133 @@
...
+        # fix_dtype not always returns the str, make sure it does
+        dtype = fix_dtype(recording, None).str
```
alejoe91, line 56: You can remove this too, since it's inherited from the parent by default in the `BasePreprocessor`
```suggestion
```

src/spikeinterface/preprocessing/decimate.py
```diff
@@ -0,0 +1,133 @@
...
+
+        BasePreprocessor.__init__(self, recording, sampling_frequency=resample_rate, dtype=dtype)
```
alejoe91, line 58:
```suggestion
        BasePreprocessor.__init__(self, recording, sampling_frequency=resample_rate, dtype=None)
```

src/spikeinterface/preprocessing/tests/test_decimate.py
```diff
@@ -0,0 +1,32 @@
...
+
+import itertools
+from spikeinterface import NumpyRecording
+from spikeinterface.core import generate_recording
+from spikeinterface.preprocessing.decimate import DecimateRecording
+import numpy as np
+
+
+@pytest.mark.parametrize("decimation_frame_start", [0, 1, 9, 10, 11, 200])
+@pytest.mark.parametrize("decimation_frame_step", [1, 9, 10, 11, 200])
```
alejoe91, line 12: A couple of comments:

- shouldn't `decimation_frame_start` < `decimation_frame_step` be enforced?
- shouldn't we assert that `decimation_frame_start`  < `num_samples`?

src/spikeinterface/preprocessing/decimate.py
TomBugnon: It can be longer than the parent size if `end_frame > self.get_num_samples()`, but that will still return the correct output I think. 

This is related to the discussion about enforcing `end_frame <= num_samples` some time ago, not sure what was decided then?

src/spikeinterface/preprocessing/decimate.py
TomBugnon: ok ok no laziness allowed that's fine

src/spikeinterface/preprocessing/decimate.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def __init__(
+        self,
+        recording,
+        decimation_factor,
+        decimation_offset=0,
+    ):
+        # Original sampling frequency
+        self._orig_samp_freq = recording.get_sampling_frequency()
+        parent_nsamp = recording.get_num_samples()
```
alejoe91, line 50:
```suggestion
        parent_nsamp = min([recording.get_num_samples(segment_index) for segment_index in range(recording.get_num_segments()])
```
Parent can be multi-segment! ;)

src/spikeinterface/preprocessing/decimate.py
alejoe91: yes, so the majority vote was to be strict about this.

I think it is ok in this case to have end_frame larger than the parent num samples (it shouldn't throuw an error), but capping at parent segment num samples won't hurt

src/spikeinterface/preprocessing/tests/test_decimate.py
```diff
@@ -0,0 +1,40 @@
...
+from pathlib import Path
+
+import itertools
+from spikeinterface import NumpyRecording
+from spikeinterface.core import generate_recording
+from spikeinterface.preprocessing.decimate import DecimateRecording
+import numpy as np
+
+
+@pytest.mark.parametrize("decimation_offset", [0, 1, 9, 10, 11, 100, 101])
```
alejoe91, line 11: Thanks! On the same line, shouldn't `decimation_offset=101` fail?

In #1989 we (kinda) agreed that we shouldn't allow for empty arrays returns. So if you have 101 samples and you ask for `parent_traces[101::*]` it will return an empty array.

src/spikeinterface/preprocessing/decimate.py
TomBugnon: I clipped start_frame/end_frame to num_samp.
Current behavior does not seem to be standardized in other preprocessors so I'm not sure exactly what to enforce here... Should I check that start_frame/end_frame are positive integers?

src/spikeinterface/preprocessing/tests/test_decimate.py
TomBugnon: Not sure what you mean... I check that some parameter combinations raise ValueErrors [here](https://github.com/SpikeInterface/spikeinterface/blob/6b07f9423d7954fb3d49c6f4b4b859c2e3c031de/src/spikeinterface/preprocessing/tests/test_decimate.py#L22) (couldn't find an elegant way to do it with `parameterize`

src/spikeinterface/preprocessing/decimate.py
TomBugnon: OK so should what should we do for multi-segments...
Same decimation factor for all segments, and accept decimation_offset to be either an int or an array of ints ?

alejoe91: I think same for both is fine

src/spikeinterface/preprocessing/tests/test_decimate.py
alejoe91: totally missed that! Looks great!

src/spikeinterface/preprocessing/decimate.py
TomBugnon: ok done

src/spikeinterface/preprocessing/decimate.py
TomBugnon: This is the last point (hopefully)

alejoe91 (approved):

Issue #2387, opened 2024-01-08
get_chunk_with_margin() should accept return_scaled


yger:
As mentionned in #2346, there is slight problem with the sortingcomponents. Currently, because of the internal function get_chunk_with_margin that do not accept return_scaled, all internal operations are performed under the hood on raw data (no scaling). This is quite often not a major issue, but if some fileformat have a negative gain, then the peaks are all upside down internally for the components. We could simply extend this function (and therefore all the node_pipelines) to accept a return_scaled argument, and leave it as default to True, as much as possible?

alejoe91:
@yger this is very tricky. The reason why `return_scaled` is only handled by the `Recording` is that if you have a processing chain you only want to apply scaling once! 

In my opinion, for the `Neuralynx` issue, the user should set the sign to positive, or sorters could do that automatically if they detect a negative gain. What do you think?

yger:
Hum... I did not consider such a problem. I'll keep thinking about it, but if no better solution, then what you are suggesting is indeed the way to proceed. If a negative gain is detected, then we should flip peak signs internally.

yger:
Actually, new proposal before discussion. Since the problem is only in the run_node_pipeline, mostly used in components, then maybe in the node pipeline, when we want to access the (final, i.e. the end of preprocessing chain) traces for the nodes, we could use a different function (get_traces_with_margin) that will make an internal call to get_traces (and thus get the scaling right). So we keep get_chunk_with_margin as a low-level function, not making any rescaling, and we have a new one (only used in the node pipeline) get scaled data. To be discussed

Issue #2346, opened 2023-12-18
peak_sign in detect peaks


Antho2422:
Hi all, 

I have a question about the peak detection with the detect peak function. I was trying to plot the detection threshold on my data and visualize the detected peaks and that was my result:

![image](https://github.com/SpikeInterface/spikeinterface/assets/98460480/8fbd4318-f038-4269-af32-c2ec83061c63)

When I put detect_sign on 'pos' I get only negative peaks and vice-versa. Am I doing something wrong or is there something I did not completly understand ? 

Thanks, 

Anthony

alejoe91:
Hi Anthony,

Your assumption is correct. The peaks should be positive! Can you check the distribution of `peaks["amplitude"]`?

yger:
This is strange, I'll have a look. But not that method='locally_exlusive' with a radius_um = 0 is equivalent to "by_channel" but much faster (using a numba kernel instead of numpy)

yger:
I can not reproduce the problem with main. Note that to visulize peaks, once you get peaks, you can turn them to spikes via si.NumpySorting.from_peaks(peaks, rec.get_sampling_frequency(), unit_ids=recording.channel_ids). And then, once you have a sorting object, you can use the WaveformExtractor to extract snippets and/or si.plot_spikes_on_traces to visualize the peaks on the data

Antho2422:
Note that in the histogram I plot the amplitudes of all the detected peaks (even the peaks out of the range of the 2nd plot)

Antho2422:
> I can not reproduce the problem with main. Note that to visualize peaks, once you get peaks, you can turn them to spikes via NumpySorting.from_peaks(peaks, rec.get_sampling_frequency(), unit_ids=recording.channel_ids). And then, once you have a sorting object, you can use the WaveformExtractor to extract snippets and/or plot_spikes_on_traces to visualize the peaks on the data

Okay I did not know that but I will try :) Thank you

Antho2422:
Hi @alejoe91 @yger , 

I would like to know a bit more about the peak sign detection on my data that seems to be reversed. I experienced the same issue during the sorting. To add more context I'm working with Neuralynx data. When I plot the data everything looks normal but during the peak detection I have to put the peak sign detection on positive in order to detect negative peaks values.

Another thing that I did not understand is that when I retrieve the peaks["amplitudes"] I get positive values but the values does not match the real peaks amplitude values. 

My questions are :

[...]

PR #2388, opened 2024-01-08
get_chunk_with_margin handling return_scaled

yger:
As mentionned in #2387 this PR fixes the problem of return_scaled within the node pipeline. Now, everything will be handled with the scaling, and while it might be slightly slower (marginally), it will avoid issues as observed in #2346 . Plus, all the templates obtained via internal spike sorting pipelines will be in appropriate units

### DIFF
--- a/src/spikeinterface/core/recording_tools.py
+++ b/src/spikeinterface/core/recording_tools.py
@@ -186,15 +186,17 @@ def get_noise_levels(
 
 
 def get_chunk_with_margin(
-    rec_segment,
+    recording,
+    segment_index,
     start_frame,
     end_frame,
-    channel_indices,
+    channel_inds,
     margin,
     add_zeros=False,
     add_reflect_padding=False,
     window_on_margin=False,
     dtype=None,
+    return_scaled=False
 ):
     """
     Helper to get chunk with margin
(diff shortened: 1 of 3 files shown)


yger:
Tests are failing because the problem is more general than I thought, and need some discussion. I do not get why the signature for recording and recordingsegment of the get_traces() should be different. This is at the root of all issues, since some low-level functions are only relying on traces from segments. However, the discrepancy here prevents us to use scaling consistently. We should discuss if this is a built-in features, or if behaviour should be harmonized @samuelgarcia

yger:
I made a solution keeping intact get_chunk_with_margin(), and simply adds an internal get_traces_with_margin() used in node_pipeline. This allows pipelines to deal with scaled data (better for peak amplitudes)

src/spikeinterface/core/recording_tools.py
```diff
@@ -306,6 +306,38 @@ def get_chunk_with_margin(
...
+    start_frame,
+    end_frame,
+    channel_indices,
+    margin,
+    add_zeros=False,
+    return_scaled=False,
+    add_reflect_padding=False,
+    window_on_margin=False,
+    dtype=None,
+):
```
zm711, line 321: Maybe a docstring, even if it is similar to get_traces?

yger: You are absolutly right, some helper/docs have been added

PR #2389, opened 2024-01-08
Add `margin_sd` to gaussian filtering

DradeAW:
Add the option to change the margins for the gaussian filter.

### DIFF
--- a/src/spikeinterface/preprocessing/filter_gaussian.py
+++ b/src/spikeinterface/preprocessing/filter_gaussian.py
@@ -34,7 +34,9 @@ class GaussianBandpassFilterRecording(BasePreprocessor):
 
     name = "gaussian_bandpass_filter"
 
-    def __init__(self, recording: BaseRecording, freq_min: float = 300.0, freq_max: float = 5000.0, margin_sd: float = 5.0):
+    def __init__(
+        self, recording: BaseRecording, freq_min: float = 300.0, freq_max: float = 5000.0, margin_sd: float = 5.0
+    ):
         sf = recording.sampling_frequency
         BasePreprocessor.__init__(self, recording)
         self.annotate(is_filtered=True)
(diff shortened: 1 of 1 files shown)


src/spikeinterface/preprocessing/filter_gaussian.py
```diff
@@ -32,19 +34,23 @@ class GaussianBandpassFilterRecording(BasePreprocessor):
 
     name = "gaussian_bandpass_filter"
 
-    def __init__(self, recording: BaseRecording, freq_min: float = 300.0, freq_max: float = 5000.0):
+    def __init__(
+        self, recording: BaseRecording, freq_min: float = 300.0, freq_max: float = 5000.0, margin_sd: float = 5.0
```
alejoe91, line 38:
```suggestion
        self, recording: BaseRecording, freq_min: float = 300.0, freq_max: float = 5000.0, margin_sd: float = 6.0
```
?

DradeAW: I think 6 by default is too high. Even 5 might be too high and 4 should be completely fine (there's nothing after 4 sd in a Gaussian distribution).

This should help speed  things up a bit.

DradeAW: For reference, at 30kHz, a gaussian filter at 300 Hz would take `5*15.92 = 80` ms margins (compared to the 5ms of the "regular filter")

alejoe91: Just wanted to point out that there was a mismatch between the default in the init and in the doc! ;)

src/spikeinterface/preprocessing/filter_gaussian.py
```diff
@@ -23,6 +23,8 @@ class GaussianBandpassFilterRecording(BasePreprocessor):
         The lower frequency cutoff for the bandpass filter.
     freq_max: float
         The higher frequency cutoff for the bandpass filter.
+    margin_sd: float, default: 6.0
```
alejoe91, line 26:
```suggestion
    margin_sd: float, default: 5.0
```

DradeAW: Ah right, good catch!

alejoe91 (approved):

samuelgarcia (approved):

Issue #2394, opened 2024-01-10
Smoothing using filter_gaussian


TomBugnon:
Hi @DradeAW @alejoe91 

I would like to be able to smooth or low-pass filter a recording. Bandpass recording with near-0 freq_min doesn't work in my case because I don't want to normalize across distant epochs.

I understand low-pass filtering is not supported because it is incompatible with chunking, and there is unfortunately no SmoothRecording (or the like) allowing to apply a rolling mean or median.

However if I understand properly, the GaussianFilterRecording applies convolution with a difference of gaussians to act as a bandpass filter. Currently, `freq_min` is enforced to be strictly positive, but it seems like it could be easily modified to support simple gaussian smoothing with `freq_min=None`. Do you think that would work @DradeAW ?
Best

DradeAW:
Yes, it would require some changes to the code but we could make `freq_min` and `freq_max` optional!

Then on this line (https://github.com/SpikeInterface/spikeinterface/blob/03126a3abbd95621bac0d6f76190326d9081dba1/src/spikeinterface/preprocessing/filter_gaussian.py#L80), we simply have to put only one of them if the other one is `None` :)

TomBugnon:
great thanks, I'll give it a shot!

Also from what I'm seeing so far, get_traces seems much much slower than for the regular bandpassfilterrecording... is that really what it takes?

DradeAW:
Yes Gaussian filtering is inherently slower than butterworth or bessel filtering.

My implementation is actually a lot faster than the scipy implementation from the testing I did.

zm711:
Since the GaussianFilter was merged I'll close this as completed by #2397.

PR #2397, opened 2024-01-10
GaussianBandpassFilter -> GaussianFilter

TomBugnon:
per https://github.com/SpikeInterface/spikeinterface/issues/2394

This makes freq_min OR freq_max optional
and refactors to GaussianFilter rather than GaussianBandpassFilter.

So there is now an actual lowpass filter in spikeinterface !

I think it is correct but please double check..!
@DradeAW  I didn't check the details of the normalization of the gaussian, are the resulting traces guaranteed to be positive in all cases if the input signal is positive?

Examples (on nice OFF periods)
highpass
image

lowpass
image

bandpass
image


src/spikeinterface/preprocessing/filter_gaussian.py
```diff
@@ -74,10 +87,17 @@ def get_traces(
...
 
         traces_fft = np.fft.fft(traces, axis=0)
-        gauss_low = self._create_gaussian(traces.shape[0], self.freq_min)
-        gauss_high = self._create_gaussian(traces.shape[0], self.freq_max)
 
-        filtered_fft = traces_fft * (gauss_high - gauss_low)[:, None]
+        if self.freq_max is not None:
+            pos_factor = self._create_gaussian(traces.shape[0], self.freq_max)
+        else:
+            pos_factor = np.ones((traces.shape[0],))
```
DradeAW, line 94: Shouldn't that be `np.zeros`?

src/spikeinterface/preprocessing/tests/test_filter_gaussian.py
```diff
@@ -21,7 +21,7 @@ def test_filter_gaussian():
     recording.annotate(is_filtered=True)
     recording = recording.save(folder=cache_folder / "recording")
 
-    rec_filtered = gaussian_bandpass_filter(recording)
+    rec_filtered = gaussian_filter(recording)
```
DradeAW, line 25: Maybe we can add tests for lowpass and highpass filters?

DradeAW (commented):
Nice work, thank you!

I noted a couple of things :)

DradeAW:
> I think it is correct but please double check..! @DradeAW I didn't check the details of the normalization of the gaussian, are the resulting traces guaranteed to be positive in all cases if the input signal is positive?

For the lowpass filter, yes it will always be positive. For the highpass and bandpass it's not guaranteed.

src/spikeinterface/preprocessing/filter_gaussian.py
TomBugnon: With np.zeros the resulting traces are negative when applying highpass. 
If the idea for high pass is to remove the filtered from the original signal then I think we want (1 - filtered) rather than (0 - filtered), no?

src/spikeinterface/preprocessing/tests/test_filter_gaussian.py
TomBugnon: yes.. but I'm not sure what the tests should look like

src/spikeinterface/preprocessing/filter_gaussian.py
DradeAW: Ah yes you're right my bad,
I guess I shouldn't be reviewing PRs late in the day ^^'

src/spikeinterface/preprocessing/tests/test_filter_gaussian.py
DradeAW: I guess you can create white noise (each point in time is independent and drawn from a Gaussian), and we can check that after a lowpass/highpass filter, the high/low frequencies are removed (while low/high are unaffected)?

DradeAW:
> For the lowpass filter, yes it will always be positive. For the highpass and bandpass it's not guaranteed.

Expanding on that: any highpass filter will lead to negative values,
Simple proof : since you are removing the frequency '0 Hz', then the mean of your signal is going to be 0 (under the hypothesis that the recording is infinite in length).
To have a mean of 0, you need both negative and positive values :)

alejoe91:
Hi @TomBugnon @DradeAW 

What's the status of this? Any additional changes needed on your side?

alejoe91:
@TomBugnon pinging on this ;)

TomBugnon:
Hi @alejoe91  @DradeAW 

- I modified current code to use reflect padding because padding with zeros leads to enormous edge effects

- I added some sort of testing having a look at the power density with this type of filtering.

Basically the low and high pass filters (the parts I added) seem to work alright, except the transition bands are huge (and much wider for larger cutoff frequencies), so I had to make the testing really really lenient 
image

When used as a bandpass filter, the traces look kinda alright,
(for instance this is 10-50Hz bandpass, compared with original and butterworth)
image
... But the power spectrum is largely attenuated in the passband (I think due to the wide and variable transition bands)
image

So I'm not testing that frequency content is conserved in the passband. 

I'm not sure what filter response is expected exactly, but in any case (after the reflect-padding change) the low-pass filtered signal is identical to what you get with the scipy function.

Sorry but I really need to stop working on this now, so I won't be doing the chunking test or go more in depth with this. Happy to send you a notebook or include some futher changes if you like

Cheers!

DradeAW:
Hey Tom!

Thanks for the update,

First of all, I don't think I agree with the latest commit (enforcing freq min < freq max).
I mean it's possible to want to apply a high-pass filter with a cutoff higher than the low-pass filter (I don't know why someone would want that, but why remove the possibility?)

Also, the response (from what I'm seeing) looks normal. The Gaussian filter is less steep than a high-order Butterworth/Bessel, but this is the reason why it introduces no overshoot / ringing / rebound.

src/spikeinterface/preprocessing/filter_gaussian.py
```diff
@@ -7,40 +7,50 @@
...
         self, recording: BaseRecording, freq_min: float = 300.0, freq_max: float = 5000.0, margin_sd: float = 5.0
     ):
-        sf = recording.sampling_frequency
         BasePreprocessor.__init__(self, recording)
         self.annotate(is_filtered=True)
 
+        if freq_min is None and freq_max is None:
+            raise ValueError("At least one of `freq_min`,`freq_max` should be specified.")
+        if freq_min is not None and freq_max is not None and freq_max <= freq_min:
+            raise ValueError("Expecting `freq_min`<`freq_max`.")
```
alejoe91, line 52:
```suggestion
```
I agree with @DradeAW. Actually specifying `freq_min`>`freq_max` allows to use this as a "band-stop" filter

alejoe91 (approved):

samuelgarcia (approved):

DradeAW:
Interesting, I should take a look at that when I have the time

DradeAW:
I think I may have made a mistake,

Applying two filters means multiplying them in the Fourier domain, so I believe it should be `traces_fft * pos_factor * (1 - neg_factor)`.
Since usually `freq_min` and `freq_max` are far apart, I didn't see it.

Let me make some check to be absolutely sure, and I'll make a PR to fix this!

PR #2404, opened 2024-01-12
NwbSortingExtractor: specify unit table name

alejoe91:
Similarly to the case with multiple `ElectricalSeries`, this PR allows users to specify a unit table to load if multiple unit tables are present (either as main units or in processing modules)

**EDIT**

This PR also refactors the usage of `electrical_series_name`: instead of using the `ElectricalSeries.name` (which could be ambiguous in case of multiple electrical series with the same name at different locations), the `electrical_series_name` is now assumed to be a "path" in the NWB tree, e.g. `acquisition/MyElectricalSeries`. The same goes for `unit_table_name` (which by default is set to `units`)

- [x] Add test with LFP (Processing module)


src/spikeinterface/extractors/tests/test_nwbextractors.py
```diff
@@ -388,5 +388,75 @@ def test_sorting_extraction_start_time_from_series(tmp_path, use_pynwb):
...
+    )
+    spike_times_a1 = np.array([0.0, 1.0, 2.0, 3.0])
+    second_units_table.add_unit(spike_times=spike_times_a1, unit_name="a1", a_second_property="a1_property_value")
+    spike_times_b1 = np.array([0.0, 1.0, 2.0])
+    second_units_table.add_unit(spike_times=spike_times_b1, unit_name="b1", a_second_property="b1_property_value")
+    processing = nwbfile.create_processing_module(name="ecephys", description="test processing module")
+    processing.add(second_units_table)
+
+    file_path = tmp_path / "test.nwb"
+    print(file_path)
```
h-mayorquin, line 424:
```suggestion
```

h-mayorquin (commented):
I think it would be better if the default behavior is to fetch the canonical units table. That is, only do the search if a specific path is passed. Otherwise we are adding computational overhead for the most common case and that will show up on the streaming case.

Note that this behavior would be different from the `ElectricalSeries` because whereas there is no canonical `ElectricalSeries` there is a canonical UnitsTable.

What do you think?

alejoe91:
> I think it would be better if the default behavior is to fetch the canonical units table. That is, only do the search if a specific path is passed. Otherwise we are adding computational overhead for the most common case and that will show up on the streaming case.
> 
> Note that this behavior would be different from the `ElectricalSeries` because whereas there is no canonical `ElectricalSeries` there is a canonical UnitsTable.
> 
> What do you think?

Hi @h-mayorquin 

Yep that makes sense! I like using paths rather thane names. I'm thinking about refactoring the electrical series part too, so that one could have ES with same names but diff locations. E.g. `acquisition/ElectricalSeries` vs `processing/ecephys/ElectricalSeries`. What do you think?

h-mayorquin:
@alejoe91 
I think that makes sense. Thinking about the future, does the access by path approach makes sense for Zarr as well?

By the way, for the NWB backend you will probably need something like this `nwbfile.get_read_io()._file["acquisition/ElectricalSeriesAp"]` to get the object by path.

alejoe91:
I'll give it a try!

alejoe91:
@h-mayorquin this is ready for review! I ended up simplifying the loading so that `electrical_series_name` and `electrical_series_location` are the same (so I removed the latter)

alejoe91:
@h-mayorquin added tests with electrical series at multiple locations. Ready for final review

h-mayorquin (commented):
I read this. 
1) For the unit table features, I still believe that if no name or path of the unit series is passed then we should revert to the default behavior of fetching the canonical units table without searching the entire object. For the reasons stated above concerning streaming.
2)  For the ElectricalSeries feature I think we should not overload the concept of name which has a precise name in NWB. That is, in NWB name does not look like a path and is always a string. This overloading leads to some inconsistencies on the documentation that now equivocates between path/location and name and will be confusing for people who know nwb. Besides, ther are no gains from using the old name. The change still breaks old code even if you keep the signature.

alejoe91:
Good points Ramon! 

I'll switch name to paths and make `units` the default `unit_table_path`

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1108,6 +1194,7 @@ def __new__(
         electrical_series_name: str | None = None,
         sampling_frequency: float | None = None,
         samples_for_rate_estimation: int = 1000,
+        unit_table_name: str = "units",
```
alejoe91, line 1197: @h-mayorquin actually that is default already (note that this is the main `NwbSorting`) class

h-mayorquin: Yeah, but you still call the `retrieve_unit_table` which actually does two things (my fault). Performs a search which is the costly part and then fetches.  I am saying we should avoid calling this function as the search is the costly part.


https://github.com/alejoe91/spikeinterface/blob/33a7babfe03cc5c1a4df3ff3434b72c660558d86/src/spikeinterface/extractors/nwbextractors.py#L75-L109

alejoe91: Got it!

alejoe91:
@h-mayorquin thansk for the feedback. I ported all these changes (and more) in #2441

Should be much easier to read! Closing

PR #250, opened 2021-09-03
Path for run_sorters with docker images

yger:
We would like to be able to use run_sorters with docker images for some of the sorters. This should deal with that. The run_sorters command also accept docker_images = {sorter_name : docker_image}. If provided, adequate docker image are used instead of local sorters

PR #2423, opened 2024-01-21
zarr IO for templates object

h-mayorquin:
Related to the hybrid project as requested.

### DIFF
--- a/src/spikeinterface/core/template.py
+++ b/src/spikeinterface/core/template.py
@@ -165,6 +165,58 @@ class Templates:
             probe=data["probe"] if data["probe"] is None else Probe.from_dict(data["probe"]),
         )
 
+    def save_to_zarr(self, folder_path):
+        import zarr
+
+        zarr_group = zarr.open_group(folder_path, mode="w")
+
+        # Saves one chunk per unit
+        arrays_chunk = (1, None, None)
+        zarr_group.create_dataset("templates_array", data=self.templates_array, chunks=arrays_chunk)
+        zarr_group.create_dataset("channel_ids", data=self.channel_ids)
+        zarr_group.create_dataset("unit_ids", data=self.unit_ids)
+
+        zarr_group.attrs["sampling_frequency"] = self.sampling_frequency
+        zarr_group.attrs["nbefore"] = self.nbefore
+
+        if self.sparsity_mask is not None:
+            zarr_group.create_dataset("sparsity_mask", data=self.sparsity_mask)
+
+        if self.probe is not None:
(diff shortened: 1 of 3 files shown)


yger:
Thanks a lot! Just because I'm also playing with the Templates, with such a PR, will be able to load/save templates as we can currently do with sorting/recordings ? Can we have a save() function that will keep the same syntax as for other objects?

src/spikeinterface/core/template.py
```diff
@@ -165,6 +165,67 @@ def from_dict(cls, data):
...
+
+        if self.sparsity_mask is not None:
+            zarr_group.create_dataset("sparsity_mask", data=self.sparsity_mask)
+
+        if self.probe is not None:
+            probe_dict = self.probe.to_dict(array_as_list=True)
+            zarr_group.attrs["probe"] = probe_dict
+
+    @classmethod
+    def load_from_open_zarr_storage(cls, base_root):
```
alejoe91, line 190:
```suggestion
    def load_from_zarr_group(cls, base_root):
```
A zarr open storage is a zarr group, no?

alejoe91:
@yger for now let's keep the `to_zarr`/`from_zarr`. We can refactor the calls later

src/spikeinterface/core/template.py
alejoe91: It doesn't need to be root. For example, the new sorting result could save this in the `results/templates` group!

h-mayorquin: OK, let's go for the suggested naming and say that the structure of the templates as defined by the writing function should anchored in that group.

h-mayorquin: Done.

alejoe91:
@samuelgarcia ready to merge on my side. Wanna take a look?

h-mayorquin:
Related to #248 and #250

samuelgarcia:
Thanks a lot for this this will be usefull.
Amybe we could add also the possibility to give an existing zarr folder and append this temapltes to specific sub group.
We have this in zarr sorting:
https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/zarrextractors.py#L176
https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/zarrextractors.py#L324

This can be convinient to save templates + other suff.

h-mayorquin:
@samuelgarcia I added the requested method. 

This should be ready save using the zarr machinery developed on the probeinterface side:
https://github.com/SpikeInterface/probeinterface/pull/250

alejoe91 (approved):

PR #2431, opened 2024-01-23
Faster unpickling of ZScoreRecording

cwindolf:
Previously, random data chunks were loaded in the `ZScoreRecording` constructor, even if `gain` and `offset` were supplied. That meant that re-loading the extractor from a pickle, for example, was slower than it needed to be (I noticed this on a network drive where loading chunks can take a while). This branch calls `get_random_data_chunks` only when needed.


h-mayorquin (approved):
Seems good to me.

alejoe91 (approved):

PR #2432, opened 2024-01-24
Fix bug in `get_repolarization_slope` with wrong index type

zm711:
In the low-level implementation of `get_repolarization_slope` line 471/3:

```python
       trough_idx = get_trough_and_peak_idx(template_single)
```

This will return a tuple instead of an int, which will cause the slicing to break later in the function. 

Since the high level call for template metrics supplies `trough_idx` and `peak_idx` this bug would only appear if someone tried to call the low level function by itself.

Also edited some docstrings.


alejoe91:
Thanks Zach!

alejoe91 (approved):

PR #2410, opened 2024-01-17
Improvement to compute sparsity without WaveformsExtractor

samuelgarcia:
This PR:
  * make template_tools.py function working without both `WaveformExtractor` (as before) and `Templates` objects
  * implements a `estimate_templates()` which is fast and do not need `WaveformExtractor`
  * implements `random_spikes_selection()`
  * implements `estimate_sparsity()` function. This is also fast. This could replace `precompute_sparsity()` This will be used by `SortingResult`
 

This PR do not break any signature or behavior. Just improve or enhance previous functions.

PR #2436, opened 2024-01-24
Tools for Generation of Hybrid recordings

yger:
This PR adds to the generation module the possibility to create hybrid recordings. These start from an existing recording and inject known units for benchmarking.

In particular, the user has can retrieve templates from a template database or estimate them from an initial spike sorting output.

The templates can be manipulated with selection, rescalilng and relocation to generate hybrid recordings of jincreasing complexity.

Importantly, a pre-eetimated motion can be used to inject spikes following the existing drift in the recording.

A how to page to showcase all these options is available here: https://spikeinterface--2436.org.readthedocs.build/en/2436/how_to/benchmark_with_hybrid_recordings.html


yger:
This PR depends on #2410

yger:
With this PR, we can get templates quickly from a recording, and also we have a high-level function to create quickly an hybrid recording, given templates and motion. Check if this is fine for you @samuelgarcia @alejoe91 . Note that this will need to be adapted with real templates from @h-mayorquin, because I will need to estimate their putative positions in order to move them around. Is there any plan to make the localization methods also work with templates directly?

src/spikeinterface/sorters/internal/spyking_circus2.py
```diff
@@ -229,31 +229,30 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
...
-            ## And this is it! We have a spyking circus
-            sorting = np.zeros(spikes.size, dtype=minimum_spike_dtype)
-            sorting["sample_index"] = spikes["sample_index"]
-            sorting["unit_index"] = spikes["cluster_index"]
-            sorting["segment_index"] = spikes["segment_index"]
-            sorting = NumpySorting(sorting, sampling_frequency, unit_ids)
+                if verbose:
+                    print("We found %d spikes" % len(spikes))
+
+                ## And this is it! We have a spyking circus
```
alejoe91, line 250: :)

alejoe91:
Thanks @yger 

Looking great! For the localization code, it's currently living in the `unit_localizations.py`, which I don't think is appropriate.

I would suggest to make a `localization_tools.py` in the `postprocessing` folder and migrate all the [localization functions](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/postprocessing/unit_localization.py#L97-L683)
In doing so, we could also add support for a `Templates` object

src/spikeinterface/generation/tests/test_hybrid_tools.py
```diff
@@ -0,0 +1,38 @@
...
+
+if hasattr(pytest, "global_test_folder"):
+    cache_folder = pytest.global_test_folder / "generation"
+else:
+    cache_folder = Path("cache_folder") / "generation"
+
+
+def test_generate_hybrid_no_motion():
+    rec, sorting = generate_ground_truth_recording(sampling_frequency=20000)
+    hybrid, sorting = generate_hybrid_recording(rec)
```
alejoe91, line 19: @yger can you add also a test for passing an external `sorting` and external `templates`?

For testing everything with and without drift, you can make a fixture to create the recording and estimate motion and use parameterized to either set `folder=None/motion_folder` :)

yger:
I've moved all utility functions into localization_tools.py in postprocessing, as suggested by @alejoe91 . Will start to expand the compute positions functions to also work with a Template objects, as this will be requested for hybrid

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,252 @@
...
+from spikeinterface.postprocessing.unit_localization import compute_monopolar_triangulation
+
+
+def estimate_templates_from_recording(
+    recording, ms_before=2, ms_after=2, sorter_name="spykingcircus2", **run_sorter_kwargs
+):
+    """
+     Get templates from a recording. Internally, SpyKING CIRCUS 2 is used (see parameters)
+    with the only twist that the template matching step is not launch. Instead, a Template
+    object is returned based on the results of the clutering.
```
alejoe91, line 29:
```suggestion
    Get templates from a recording. Internally, SpyKING CIRCUS 2 is used by default
    with the only twist that the template matching step is not launched. Instead, a Template
    object is returned based on the results of the clustering. Other sorters can be invoked 
    with the `sorter_name` and `run_sorter_kwargs` parameters.
```

src/spikeinterface/generation/tests/test_hybrid_tools.py
```diff
@@ -0,0 +1,45 @@
...
+from spikeinterface.core.generate import generate_ground_truth_recording, generate_sorting
+from spikeinterface.core import Templates, BaseRecording
+from spikeinterface.preprocessing.motion import correct_motion, load_motion_info
+from spikeinterface.generation.hybrid_tools import estimate_templates_from_recording, generate_hybrid_recording
+
+
+if hasattr(pytest, "global_test_folder"):
+    cache_folder = pytest.global_test_folder / "generation"
+else:
+    cache_folder = Path("cache_folder") / "generation"
```
alejoe91, line 14:
```suggestion
```
Let's start getting rid of these

src/spikeinterface/generation/tests/test_hybrid_tools.py
```diff
@@ -0,0 +1,45 @@
...
+    rec, sorting = generate_ground_truth_recording(sampling_frequency=20000)
+    correct_motion(rec, folder=cache_folder / "motion")
+    motion = load_motion_info(cache_folder / "motion")
+    hybrid, sorting = generate_hybrid_recording(rec, motion)
+
+
+def test_estimate_templates():
+    rec, sorting = generate_ground_truth_recording(num_units=10, sampling_frequency=20000)
+    templates = estimate_templates_from_recording(rec, output_folder=cache_folder / "sc", remove_existing_folder=True)
+    assert len(templates.templates_array) > 0
```
alejoe91, line 38:
```suggestion
def test_estimate_templates(tmp_path):
    rec, sorting = generate_ground_truth_recording(num_units=10, sampling_frequency=20000)
    templates = estimate_templates_from_recording(rec, output_folder=tmp_path / "sc", remove_existing_folder=True)
    assert len(templates.templates_array) > 0
```

src/spikeinterface/generation/tests/test_hybrid_tools.py
```diff
@@ -0,0 +1,45 @@
...
+def test_generate_hybrid_no_motion():
+    rec, sorting = generate_ground_truth_recording(sampling_frequency=20000)
+    hybrid, sorting = generate_hybrid_recording(rec)
+
+
+def test_generate_hybrid_motion():
+    rec, sorting = generate_ground_truth_recording(sampling_frequency=20000)
+    correct_motion(rec, folder=cache_folder / "motion")
+    motion = load_motion_info(cache_folder / "motion")
+    hybrid, sorting = generate_hybrid_recording(rec, motion)
```
alejoe91, line 32:
```suggestion
def test_generate_hybrid_motion(tmp_path):
    rec, sorting = generate_ground_truth_recording(sampling_frequency=20000)
    correct_motion(rec, folder=tmp_path / "motion")
    motion = load_motion_info(tmp_path / "motion")
    hybrid, sorting = generate_hybrid_recording(rec, motion)
```

alejoe91 (changes requested):
@yger looking great!

Just added a few small comments

src/spikeinterface/widgets/unit_waveforms.py
```diff
@@ -1,12 +1,13 @@
 from __future__ import annotations
 
+from coverage import data
```
alejoe91, line 3:
```suggestion
```

alejoe91:
Added additional functionalities for hybrid generations:

- filter a `Templates` object based on amplitudes and depths
- scale `Templates` between min/max user selected amps
- shift `Templates` to move them away from their original position

Also extended the `plot_unit_templates` to work with a `sorting_analyzer_or_templates` object

alejoe91:
@samuelgarcia @yger since I was working on it, I also added a Widen/Narrow option and scalebars for `plot_unit_waveforms/templates`. Here's how it looks:

[Screencast from 14-05-2024 12:39:14.webm](https://github.com/SpikeInterface/spikeinterface/assets/17097257/56759d0d-3d23-43b4-8a45-74dca54df4c5)

This is ready to merge on my side!

src/spikeinterface/generation/template_database.py
```diff
@@ -1,12 +1,133 @@
...
+    """
+    import pandas as pd
+
+    templates_array = []
+    if isinstance(template_df_or_indices, pd.DataFrame):
+        template_info = template_df_or_indices
+    else:
+        template_info = fetch_templates_info().iloc[template_df_or_indices]
+    requested_datasets = np.unique(template_info["dataset"]).tolist()
+    print(f"Fetching templates from {len(requested_datasets)} datasets")
```
h-mayorquin, line 85: a dreaded verbose option here?

src/spikeinterface/generation/template_database.py
```diff
@@ -1,12 +1,133 @@
...
+    """
     s3_path = f"s3://spikeinterface-template-database/{dataset}/"
     zarr_group = zarr.open_consolidated(s3_path, storage_options={"anon": True})
 
     templates_object = Templates.from_zarr_group(zarr_group)
 
     return templates_object
+
+
+def fetch_templates_info() -> "pandas.DataFrame":
```
h-mayorquin, line 31:
```suggestion
import functools
@functools.cache
def fetch_templates_info() -> "pandas.DataFrame":
```

https://docs.python.org/3/library/functools.html#functools.cache

src/spikeinterface/generation/template_database.py
```diff
@@ -1,12 +1,133 @@
-from spikeinterface.core.template import Templates
 import zarr
+import numpy as np
+
+from spikeinterface.core.template import Templates
 
 
-def fetch_templates_from_database(dataset="test_templates.zarr") -> Templates:
+def fetch_template_dataset(dataset="test_templates.zarr") -> Templates:
```
h-mayorquin, line 7:
```suggestion
@functools.cache
def fetch_template_dataset(dataset="test_templates.zarr") -> Templates:
```

src/spikeinterface/generation/template_database.py
```diff
@@ -1,12 +1,133 @@
-from spikeinterface.core.template import Templates
 import zarr
+import numpy as np
+
```
h-mayorquin, line 3:
```suggestion
import functools
```

h-mayorquin (commented):
We can add those things after the PR is merged as well : )

src/spikeinterface/generation/template_database.py
alejoe91: Yep! should be safe here :)

src/spikeinterface/generation/template_database.py
h-mayorquin: Ah, this will fail if people still use python 3.8 : /

We can use `functools.lru_cache(maxsize=1)` instead if you want to support that but I would rather move forward to the bright future of python 3. 9 as minimal.

alejoe91: yes let's do that for the sorting analyzer release

cwindolf:
Hey! Me and @chris-langfield are interested in using this code in our benchmarks. I was wondering how ready / stable it is? Also will the [`Motion` object](https://github.com/SpikeInterface/spikeinterface/pull/2915 ) change the API here?

Lastly, if I bring my own `(n_drift_bins, n_samples, n_channels)` array of templates, can I manually plug that in here? (I couldn't tell if this code assumes that it computes the drifting templates via the `DriftingTemplates.precompute_displacements()`). Would you be open to adding a `DriftingTemplates.from_precomputed(templates_array, drift_bin_um=5, **other_kwargs)` constructor to support that kind of workflow? I'd be super happy to contribute code along those lines.

Also, if there are other to-do items that I could help out with here, let me know!

alejoe91:
Hi @cwindolf 

That is great! :) Sure we'd be happy to add the `from_precomputed` option. This would also make it compatible with MEArec drifting generation.

We are currently splitting this PR into several smaller PRs, so that this one will only have the main generator functions.
Can you make a separate PR for the `from_precomputed`?

alejoe91:
@yger @samuelgarcia @h-mayorquin 

I would suggest that we merge this one. In a follow-up PR I'll add a documentation/How-to page to show how to construct drift-aware hybrid recordings (probably starting from Nick's data).

What do you guys think?

samuelgarcia:
I want to read it first please.

alejoe91:
Let's go through this tomorrow

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+
+    Parameters
+    ----------
+    ms_before: float
+        The time before peaks of templates
+    ms_after: float
+        The time after peaks of templates
+    sorter_name: str
+        The sorter to be used in order to get some fast clustering
+    run_sorter_kwargs: dict
```
alejoe91, line 46: Let's fix colons here

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+        The recording to inject units in
+    motion_info: All the information about the motion
+        The motion datastructure of the recording
+    num_units: int, default: 10
+        Number of units,  not used when sorting is given.
+    sorting: Sorting or None
+        An external sorting object. If not provide, one is genrated.
+    templates: Templates or None, default: None
+        The templates of units.
+        If None they are generated.
```
alejoe91, line 348: colons

src/spikeinterface/core/template.py
```diff
@@ -294,9 +294,9 @@ def from_zarr_group(cls, zarr_group: "zarr.Group") -> "Templates":
         the `add_templates_to_zarr_group` method.
 
         """
-        templates_array = zarr_group["templates_array"]
-        channel_ids = zarr_group["channel_ids"]
-        unit_ids = zarr_group["unit_ids"]
+        templates_array = zarr_group["templates_array"][:]
```
h-mayorquin, line 297: I think there is something to watch out when having this behavior by default. If we are serializing something in multiple places then we will keep making the full read even if the full function is not used. This might be important for reducing the number of calls to S3.

On the other hand this might be better that someone doing multiple requests without the copy...

Myabe instead of going all the way manual as in https://github.com/SpikeInterface/spikeinterface/pull/2853/files we can have a method called get_templates that does this when it is called?

Just thinking aloud.

src/spikeinterface/core/template.py
```diff
@@ -352,6 +352,55 @@ def to_json(self):
     def from_json(cls, json_str):
         return cls.from_dict(json.loads(json_str))
 
+    def select_units(self, unit_ids):
```
h-mayorquin, line 355: Added this and test in https://github.com/SpikeInterface/spikeinterface/pull/2989

src/spikeinterface/generation/drift_tools.py
```diff
@@ -125,13 +125,13 @@ def __init__(self, **kwargs):
 
     @classmethod
     def from_static(cls, templates):
```
h-mayorquin, line 127:
```suggestion
    def from_static_templates(cls, templates):
```
?

h-mayorquin (commented):
Some comments.

src/spikeinterface/core/generate.py
```diff
@@ -1862,7 +1861,7 @@ def get_traces(
             wf = template[start_template:end_template]
             if self.amplitude_vector is not None:
                 wf = wf * self.amplitude_vector[i]
-            traces[start_traces:end_traces] += wf
+            traces[start_traces:end_traces] += wf.astype(traces.dtype, copy=False)
```
samuelgarcia, line 1864: OK but the += should do that automatically for us no ?

src/spikeinterface/core/recording_tools.py
```diff
@@ -709,7 +709,7 @@ def get_chunk_with_margin(
     case zero padding is used, in the second case np.pad is called
     with mod="reflect".
     """
-    length = rec_segment.get_num_samples()
+    length = int(rec_segment.get_num_samples())
```
samuelgarcia, line 712: would be better to do this in def get_num_sumples no ?

src/spikeinterface/core/template.py
samuelgarcia: @h-mayorquin  I am agree but here this is a class method and we really want a in memory loading no ?

src/spikeinterface/core/template.py
samuelgarcia: @yger : could you update from main so ?

src/spikeinterface/generation/drift_tools.py
samuelgarcia: I am OK with this.
@yger could you propagate this change at some places ?

src/spikeinterface/generation/drift_tools.py
```diff
@@ -264,9 +264,9 @@ def __init__(
     ):
         import scipy.spatial
 
-        assert isinstance(
-            drifting_templates, DriftingTemplates
-        ), "drifting_templates must be a DriftingTemplates object"
+        # assert isinstance(
+        #     drifting_templates, DriftingTemplates
+        # ), "drifting_templates must be a DriftingTemplates object"
```
samuelgarcia, line 269: Why ?

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+        The time before peaks of templates
+    ms_after : float
+        The time after peaks of templates
+    sorter_name : str
+        The sorter to be used in order to get some fast clustering
+    run_sorter_kwargs : dict
+        The parameters to provide to the run_sorter function of spikeinterface
+
+
+    sorter_params: keyword arguments for `spyking_circus2` function
```
samuelgarcia, line 50: ??

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+
+    Returns
+    -------
+    templates: Templates
+        The found templates
+    """
+    from spikeinterface.sorters.runsorter import run_sorter
+    from spikeinterface.core.template import Templates
+    from spikeinterface.core.waveform_tools import estimate_templates
```
samuelgarcia, line 59: This could be at modules levels no ?

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+
+    if sorter_name == "spykingcircus2":
+        if "matching" not in run_sorter_kwargs:
+            run_sorter_kwargs["matching"] = {"method": None}
```
samuelgarcia, line 63: Why having sorter specifics here ?

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+
+    sorting = run_sorter(sorter_name, recording, **run_sorter_kwargs)
+
+    from spikeinterface.core.waveform_tools import estimate_templates
```
samuelgarcia, line 67: at module level no ?

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+
+    spikes = sorting.to_spike_vector()
+    unit_ids = np.unique(spikes["unit_index"])
+    sampling_frequency = recording.get_sampling_frequency()
+    nbefore = int(ms_before * sampling_frequency / 1000.0)
+    nafter = int(ms_after * sampling_frequency / 1000.0)
+
+    _, job_kwargs = split_job_kwargs(run_sorter_kwargs)
```
samuelgarcia, line 75: No. This is sorters specific.
Some sorters do not have job_kwargs.
Lets explicitily define job_kwargs in the signature instead of hacking some nested one.

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+
+def estimate_templates_from_recording(
+    recording, ms_before=2, ms_after=2, sorter_name="spykingcircus2", **run_sorter_kwargs
+):
+    """
+    Get templates from a recording. Internally, SpyKING CIRCUS 2 is used by default
+    with the only twist that the template matching step is not launched. Instead, a Template
+    object is returned based on the results of the clustering. Other sorters can be invoked
+    with the `sorter_name` and `run_sorter_kwargs` parameters.
+
```
samuelgarcia, line 37: We should specified that templates is dense.

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+    return Templates(
+        templates_array=scaled_templates_array,
+        sampling_frequency=templates.sampling_frequency,
+        nbefore=templates.nbefore,
+        sparsity_mask=templates.sparsity_mask,
+        channel_ids=templates.channel_ids,
+        unit_ids=templates.unit_ids,
+        probe=templates.probe,
+    )
```
samuelgarcia, line 219: And then we could could call the `scale_template(templates, gain)` here

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+
+
+def shift_templates(
```
samuelgarcia, line 222: shift is a bit fuzy no ? For me shift is more somethign shift in time.
Why not move_templates() ?

samuelgarcia: And also I would call it random_move_templates() this is more explcit.

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+    upsample_factor=None,
+    upsample_vector=None,
+    amplitude_std=0.05,
+    generate_sorting_kwargs=dict(firing_rates=15, refractory_period_ms=4.0, seed=2205),
+    generate_unit_locations_kwargs=dict(margin_um=10.0, minimum_z=5.0, maximum_z=50.0, minimum_distance=20),
+    generate_templates_kwargs=dict(),
+    seed=None,
+):
+    """
+    Generate an hybrid recording with spike given sorting+templates.
```
samuelgarcia, line 334: This cool function deserve more explanation no ?

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+        probe=templates.probe,
+    )
+
+
+def generate_hybrid_recording(
+    recording,
+    motion_info=None,
+    num_units=10,
+    sorting=None,
+    templates=None,
```
samuelgarcia, line 320: I would push this kwargs in second position.

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+        channel_ids=templates.channel_ids,
+        unit_ids=templates.unit_ids,
+        probe=templates.probe,
+    )
+
+
+def generate_hybrid_recording(
+    recording,
+    motion_info=None,
+    num_units=10,
```
samuelgarcia, line 318: Do we need this when templates is given ?

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+    templates : Templates or None, default: None
+        The templates of units.
+        If None they are generated.
+    templates_in_uV : bool, default: True
+        If True, the templates are in uV, otherwise they are in the same unit as the recording.
+        In case the recording has scaling, the templates are "unscaled" before injection.
+    ms_before : float, default: 1.5
+        Cut out in ms before spike peak.
+    ms_after : float, default: 3
+        Cut out in ms after spike peak.
```
samuelgarcia, line 355: This should go in generate_template_kwargs no ?

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+    motion_info : All the information about the motion
+        The motion datastructure of the recording
+    num_units : int, default: 10
+        Number of units,  not used when sorting is given.
+    sorting : Sorting or None
+        An external sorting object. If not provide, one is genrated.
+    templates : Templates or None, default: None
+        The templates of units.
+        If None they are generated.
+    templates_in_uV : bool, default: True
```
samuelgarcia, line 349: Should have something more in line with return_scaled_traces ?

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+        In case the recording has scaling, the templates are "unscaled" before injection.
+    ms_before : float, default: 1.5
+        Cut out in ms before spike peak.
+    ms_after : float, default: 3
+        Cut out in ms after spike peak.
+    unit_locations : np.array, default: None
+        The locations at which the templates should be injected. If not provided, generated (see
+        generate_unit_location_kwargs)
+    upsample_factor : None or int, default: None
+        A upsampling factor used only when templates are not provided.
```
samuelgarcia, line 360: This should go in generate_template_kwargs no ?

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+        generate_sorting_kwargs = generate_sorting_kwargs.copy()
+        generate_sorting_kwargs["durations"] = durations
+        generate_sorting_kwargs["sampling_frequency"] = sampling_frequency
+        generate_sorting_kwargs["seed"] = seed
+        generate_sorting_kwargs["num_units"] = num_units
+        sorting = generate_sorting(**generate_sorting_kwargs)
+    else:
+        num_units = sorting.get_num_units()
+        assert sorting.sampling_frequency == sampling_frequency
+
```
samuelgarcia, line 457: Are we checking that num_units from the sorting is the same as the number of templates ?

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+    if amplitude_std is not None:
+        amplitude_factor = rng.normal(loc=1, scale=amplitude_std, size=num_spikes)
+    else:
+        amplitude_factor = None
+
+    if motion_info is not None:
+        start = np.array([0, np.min(motion_info["motion"])])
+        stop = np.array([0, np.max(motion_info["motion"])])
+        displacements = make_linear_displacement(start, stop, num_step=int((stop - start)[1]))
```
samuelgarcia, line 480: [1] is dangerous this should use a "direction" x or y.
The num step should be expose here we are forcing to one um.

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+        if upsample_vector is None:
+            upsample_factor = templates_array.shape[3]
+            upsample_vector = rng.integers(0, upsample_factor, size=num_spikes)
+
+    if amplitude_std is not None:
+        amplitude_factor = rng.normal(loc=1, scale=amplitude_std, size=num_spikes)
+    else:
+        amplitude_factor = None
+
+    if motion_info is not None:
```
samuelgarcia, line 477: I think we should merge the Motion object first here because it will force to many plainfull changes after.

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+            displacement_unit_factor[count] = a / a.sum()
+
+        hybrid_recording = InjectDriftingTemplatesRecording(
+            sorting=sorting,
+            parent_recording=recording,
+            drifting_templates=drifting_templates,
+            displacement_vectors=[displacement_vectors],
+            displacement_sampling_frequency=displacement_sampling_frequency,
+            displacement_unit_factor=displacement_unit_factor,
+            num_samples=durations * sampling_frequency,
```
samuelgarcia, line 505:
```suggestion
            num_samples=(durations * sampling_frequency).astype("int64"),
```

src/spikeinterface/generation/tests/test_hybrid_tools.py
samuelgarcia: Yes I am agree. We should have stronger tests.

samuelgarcia: we should use the generate_drifting function everywhere in theses tests instead of the core one.

src/spikeinterface/generation/tests/test_mock.py
```diff
@@ -1,3 +0,0 @@
-def test_mock():
-    # TODO: Add test logic here
-    pass
```
samuelgarcia, line 3: do we need this file ?

src/spikeinterface/postprocessing/localization_tools.py
```diff
@@ -0,0 +1,338 @@
...
+
+import numpy as np
+
+try:
+    import numba
+
+    HAVE_NUMBA = True
+except ImportError:
+    HAVE_NUMBA = False
+
```
samuelgarcia, line 13: really cool to have a separate file for all the machinery.

src/spikeinterface/postprocessing/unit_localization.py
```diff
@@ -92,95 +87,8 @@ def get_data(self, outputs="numpy"):
...
-    err = (
-        np.square(wf_data / max_data - data_at(x, y, z, alpha, local_contact_locations)).mean()
-        - np.log1p(10.0 * z) / 10000.0
-    )
-    return err
-
-
 def compute_monopolar_triangulation(
-    sorting_analyzer,
+    sorting_analyzer_or_templates,
```
samuelgarcia, line 91: super cool

src/spikeinterface/preprocessing/motion.py
```diff
@@ -418,6 +420,28 @@ def correct_motion(
 correct_motion.__doc__ = correct_motion.__doc__.format(_doc_presets, _shared_job_kwargs_doc)
 
 
+def save_motion_info(motion_info, folder, overwrite=False):
```
samuelgarcia, line 423: OK for this but then correct_motion() should use it internally to avoid code duplication.

src/spikeinterface/sorters/internal/spyking_circus2.py
```diff
@@ -281,29 +281,30 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
...
 
-            for value in ["chunk_size", "chunk_memory", "total_memory", "chunk_duration"]:
-                if value in matching_job_params:
-                    matching_job_params[value] = None
-            matching_job_params["chunk_duration"] = "100ms"
+            if matching_method is not None:
+                for value in ["chunk_size", "chunk_memory", "total_memory", "chunk_duration"]:
+                    if value in matching_job_params:
+                        matching_job_params[value] = None
+                matching_job_params["chunk_duration"] = "100ms"
```
samuelgarcia, line 288: Is it compatible with latests change in SC2 in main ?

src/spikeinterface/generation/hybrid_tools.py
yger: Because we don't need to launch the template macthing step to get the Templates. If some sorters can have the option to get templates without a proper sorting, then we should use it

src/spikeinterface/generation/hybrid_tools.py
alejoe91: move this to the top

alejoe91: except run sorter, because very specific to this function

src/spikeinterface/generation/hybrid_tools.py
alejoe91: not in this PR, but we could make a `hybrid_sorter` which is designed for this:
- high defautk peak detection threshold
- conservative clustering
- no TM

alejoe91: so we don't need to make custom adjustments to existing end-to-end sorters

src/spikeinterface/generation/hybrid_tools.py
alejoe91: `relocate_templates`

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+
+def generate_hybrid_recording(
+    recording,
+    motion_info=None,
+    num_units=10,
+    sorting=None,
+    templates=None,
+    templates_in_uV=True,
+    ms_before=1.0,
+    ms_after=3.0,
```
alejoe91, line 323: move to generate_templates_kwargs

src/spikeinterface/generation/hybrid_tools.py
alejoe91: move to generate_sorting_kwargs

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+    seed=None,
+):
+    """
+    Generate an hybrid recording with spike given sorting+templates.
+
+    Parameters
+    ----------
+    recording : BaseRecording
+        The recording to inject units in
+    motion_info : All the information about the motion
```
alejoe91, line 340: should be motion object

src/spikeinterface/generation/hybrid_tools.py
alejoe91: done

src/spikeinterface/generation/hybrid_tools.py
alejoe91: done

src/spikeinterface/generation/hybrid_tools.py
alejoe91: done

src/spikeinterface/preprocessing/motion.py
alejoe91: done

src/spikeinterface/generation/hybrid_tools.py
alejoe91: done

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,518 @@
...
+        drifting_templates = DriftingTemplates.from_static(templates_)
+        drifting_templates.precompute_displacements(displacements)
+
+        displacement_sampling_frequency = 1.0 / np.diff(motion_info["temporal_bins"])[0]
+        displacement_vectors = np.zeros((len(motion_info["temporal_bins"]), 2, len(motion_info["spatial_bins"])))
+        displacement_unit_factor = np.zeros((num_units, len(motion_info["spatial_bins"])))
+
+        for count, i in enumerate(motion_info["spatial_bins"]):
+            local_motion = motion_info["motion"][:, count]
+            displacement_vectors[:, 1, count] = local_motion
```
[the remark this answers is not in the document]
alejoe91: done (also below)

alejoe91:
@samuelgarcia @yger @h-mayorquin @cwindolf 

Everything is merged and all suggestions have been incorporated. 
Also extended tests and moved more "tools" functions to the new `localization_tools.py`.

This is ready to merged on my side!

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,562 @@
...
+
+            for count, i in enumerate(spatial_bins_um):
+                local_motion = displacement_segment[:, count]
+                displacement_vector[:, motion.dim, count] = local_motion
+            displacement_vectors.append(displacement_vector)
+
+        # calculate displacement unit factor
+        displacement_unit_factor = np.zeros((num_units, len(spatial_bins_um)))
+        for count in range(num_units):
+            a = 1 / np.abs((unit_locations[count, motion.dim] - spatial_bins_um))
```
cwindolf, line 539: I'm a bit confused about this -- why not just have each unit follow the motion at its location? (I.e., compute one displacement vector for each unit and `displacement_unit_factor = np.eye(num_units)`.) This seems less intuitive and more like a specific choice (why 1/r weighting?)

alejoe91: that's a good point @cwindolf !!!

JoeZiminski:
I am very fascinated by this PR and the somewhat mysterious 'Hybrid' recordings, sounds cool! Could you extend the PR message with an explanation to the naive (me) 😄 ?

src/spikeinterface/generation/hybrid_tools.py
alejoe91: @samuelgarcia @yger can you comment on this?

alejoe91:
Ready for final review here. See How To docs here: https://spikeinterface--2436.org.readthedocs.build/en/2436/how_to/benchmark_with_hybrid_recordings.html

src/spikeinterface/generation/hybrid_tools.py
alejoe91: I think this is ok, or maybe `are_templates_in_uV` would be even better @zm711

src/spikeinterface/generation/hybrid_tools.py
alejoe91: @cwindolf in pratice we don't have the motion at the unit depth, so we need to interpolate/weigh the estimated motions depending on the unit location. This is one way to do that, so that if you have a motion estimated at a certain depth it will be predominant (1/r).

examples/how_to/benchmark_with_hybrid_recordings.py
```diff
@@ -0,0 +1,258 @@
...
+import matplotlib.pyplot as plt
+from pathlib import Path
+
+# -
+
+# %matplotlib inline
+
+si.set_global_job_kwargs(n_jobs=16)
+
+# To make this notebook self-contained, we will simulate a drifting recording similar to the one acquired by Nick Steinmetz and available [here](https://doi.org/10.6084/m9.figshare.14024495.v1), where an triangular motion was imposed to the recording by moving the probe up and down with a micro-manipulator.
```
samuelgarcia, line 48: Why not using a true dataset ?
The goal of the hybrid is really to play with real dataset no ?

examples/how_to/benchmark_with_hybrid_recordings.py
```diff
@@ -0,0 +1,258 @@
...
+        {
+            "drift_mode": "zigzag",
+            "non_rigid_gradient": None,
+            "t_start_drift": 30.0,
+            "t_end_drift": 210,
+            "period_s": 60,
+        }
+    ],
+    "drift_start_um": [0, 60],
+    "drift_stop_um": [0, -60],
```
samuelgarcia, line 62: it is a big one

examples/how_to/benchmark_with_hybrid_recordings.py
```diff
@@ -0,0 +1,258 @@
...
+
+sparsity_plot = si.compute_sparsity(templates_relocated)
+fig = plt.figure(figsize=(10, 10))
+w = sw.plot_unit_templates(templates_relocated, sparsity=sparsity_plot, ncols=4, figure=fig)
+w.figure.subplots_adjust(wspace=0.5, hspace=0.7)
+
+# ## Constructing hybrid recordings
+
+recording_hybrid_no_drift, sorting_hybrid = sgen.generate_hybrid_recording(
+    recording=drifting_recording, templates=templates_relocated, seed=2308
```
samuelgarcia, line 145: should be the none drifting no ?

src/spikeinterface/generation/drift_tools.py
```diff
@@ -117,21 +119,85 @@ class DriftingTemplates(Templates):
     """
 
     def __init__(self, **kwargs):
+        has_templates_moved = "templates_array_moved" in kwargs
+        has_displacements = "displacements" in kwargs
```
samuelgarcia, line 123: Lets put this in the signature to avoid hidden kwargs manipulation

src/spikeinterface/generation/drift_tools.py
```diff
@@ -117,21 +119,85 @@ class DriftingTemplates(Templates):
     """
 
     def __init__(self, **kwargs):
```
samuelgarcia, line 121:
```suggestion
    def __init__(self, **static_template_kwargs):
```

src/spikeinterface/generation/drift_tools.py
```diff
@@ -117,21 +119,85 @@ class DriftingTemplates(Templates):
...
+        precomputed = has_templates_moved or has_displacements
+        if precomputed and not (has_templates_moved and has_displacements):
+            raise ValueError(
+                "Please pass both template_array_moved and displacements to DriftingTemplates "
+                "if you are using precomputed displaceed templates."
+            )
+        templates_array_moved = kwargs.pop("templates_array_moved", None)
+        displacements = kwargs.pop("displacements", None)
+
         Templates.__init__(self, **kwargs)
```
samuelgarcia, line 133:
```suggestion
        Templates.__init__(self, **static_template_kwargs)
```

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,561 @@
...
+        unit_ids=templates.unit_ids,
+        probe=templates.probe,
+    )
+
+
+def generate_hybrid_recording(
+    recording: BaseRecording,
+    sorting: BaseSorting | None = None,
+    templates: Templates | None = None,
+    motion: Motion | None = None,
```
samuelgarcia, line 321: What is the behavior when motion is None ?
Should we raised an error a warning ?
Because generating hybrid without motion do not make sens no ?

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -0,0 +1,561 @@
...
+        motion_array_concat = np.concatenate(motion.displacement)
+        if dim == 0:
+            start = np.array([np.min(motion_array_concat), 0])
+            stop = np.array([np.max(motion_array_concat), 0])
+        elif dim == 1:
+            start = np.array([0, np.min(motion_array_concat)])
+            stop = np.array([0, np.max(motion_array_concat)])
+        elif dim == 2:
+            raise NotImplementedError("3D motion not implemented yet")
+        displacements = make_linear_displacement(start, stop, num_step=int((stop - start)[dim]))
```
samuelgarcia, line 515: Should we expose an option for the num_step ?
here we are forcing a 1um step, this could be hudge, am i wrong ?

examples/how_to/benchmark_with_hybrid_recordings.py
alejoe91: We decided using simulated with @yger 

I think it's good for reproducibility and how to purposes

examples/how_to/benchmark_with_hybrid_recordings.py
alejoe91: on purpose

examples/how_to/benchmark_with_hybrid_recordings.py
alejoe91: no this is the drifting recording without accounting for drift. Good point though, maybe it needs a name change

src/spikeinterface/generation/hybrid_tools.py
alejoe91: maybe a warning. But sometimes recordings are super stable, so one could choose not to interpolate

src/spikeinterface/generation/hybrid_tools.py
samuelgarcia: @cwindolf : I think this could be done but the displacement vector will be enormous and I think this would slow down the generator. Here alessio is implementing one displacement per spatial bin.
@alejoe91 : I would use a linear interpolation between the 2 nearest spatial bins no ?

src/spikeinterface/generation/tests/test_hybrid_tools.py
```diff
@@ -0,0 +1,85 @@
...
+    assert rec.get_num_segments() == hybrid.get_num_segments()
+    assert np.array_equal(rec.get_channel_locations(), hybrid.get_channel_locations())
+    assert sorting_hybrid.get_num_units() == len(hybrid.templates)
+
+
+def test_generate_hybrid_motion(create_cache_folder):
+    cache_folder = create_cache_folder
+    rec, _ = generate_ground_truth_recording(sampling_frequency=20000, durations=[10], seed=0)
+    correct_motion(rec, folder=cache_folder / "motion")
+    motion_info = load_motion_info(cache_folder / "motion")
```
samuelgarcia, line 41: You can us coorect_motion(output_motion_info=True) to avoid a disk write here.
No ?

src/spikeinterface/generation/hybrid_tools.py
alejoe91: I think @yger did this

src/spikeinterface/generation/tests/test_hybrid_tools.py
alejoe91: yes

zm711:
I was tagged, but I can't find where. Happy to comment on something if my opinion was desired. Just got an alert this morning, but now looking I can't find it :)

src/spikeinterface/generation/hybrid_tools.py
cwindolf: Yeah, the `Motion` object knows how to do interpolation already, and it uses linear interpolation rather than this distance-weighted one -- it could be nice to use the Motion object internally, but it's nice to be able to control drift per-unit like this allows. Maybe a middle ground could be to generate a linear interpolation matrix here, so that the output matches what you would get when using the motion object?

alejoe91: Done in the last commits @cwindolf !!!

alejoe91:
> I was tagged, but I can't find where. Happy to comment on something if my opinion was desired. Just got an alert this morning, but now looking I can't find it :)

It was about naming. Now one of the parameter is `templates_in_uV`, but I think it should be `are_templates_in_uV`? Not sure though since it's not a function

zm711:
Cool, I think it is confusing right now because we have two arguments `templates` and `templates_in_uV` which seems to imply that we need to give both (templates and the same templates scaled). The docstring tells us one is a bool, but I do think something like `scale_templates` or `return_scaled` (if we need to be consistent) would be clear. `are_templates_in_uV` could also work with the assumption that we are forcing the user to answer the question, but I think the `scale` term would be more consistent to the library.

alejoe91:
> Cool, I think it is confusing right now because we have two arguments `templates` and `templates_in_uV` which seems to imply that we need to give both (templates and the same templates scaled). The docstring tells us one is a bool, but I do think something like `scale_templates` or `return_scaled` (if we need to be consistent) would be clear. `are_templates_in_uV` could also work with the assumption that we are forcing the user to answer the question, but I think the `scale` term would be more consistent to the library.

@zm711 we need to know if the provided templates are already scaled or not, so we don't want a verb IMO, so I'm leaning towards `are_templates_in_uV`, because it's indeed a question we need the user to provide info for :)

zm711:
> @zm711 we need to know if the provided templates are already scaled or not, so we don't want a verb IMO, so I'm leaning towards are_templates_in_uV, because it's indeed a question we need the user to provide info for :)

Then yes I think that `are_templates_in_uV` is better. I guess my point would be something like `are_templates_scaled` because our library has an equivalence between `uV` and `scaled`.

alejoe91:
> Then yes I think that `are_templates_in_uV` is better. I guess my point would be something like `are_templates_scaled` because our library has an equivalence between `uV` and `scaled`.

I like that!

examples/how_to/benchmark_with_hybrid_recordings.py
alejoe91: Substituted with Nick's data

alejoe91:
Thank you all for the input! I think this will super important for future benchmarks.

@JoeZiminski see updated PR description :)

Merging!

JoeZiminski:
Cheers, that's awesome! Look forward to trying it

PR #2438, opened 2024-01-25
Add bit depth compensation to unsigned_to_signed

BptGrm:
Circling back to NeuralEnsemble/python-neo#1348 : 

This PR, as is, returns correct traces from BioCAM recordings when processed through 
`signed_rec = spre.unsigned_to_signed(recording,bit_depth=12)`.

Looking at traces with : 
```
traces = recording.get_traces(channel_ids=['35','26','15','23','8'],
                               end_frame=4096) # Picking a few random channels
plt.plot(traces)
```
Raw  : 
raw

Processed : 
processed

A few questions : 

 - I'm not sure which version of numpy was around when this function was written, but on line 23 : 
 https://github.com/SpikeInterface/spikeinterface/blob/c2f18bd536437e9f500a4e24f7a0f3773bd79d3a/src/spikeinterface/preprocessing/unsigned_to_signed.py#L23
`dtype.str` now returns the data type as byte order + kind + item size, which means that `.replace("uint", "int")` does nothing and `dtype_signed` ends up being the same as `dtype`, thus traces get returned with the same data type as the raw recording.
I believe `dtype.name` behaves as expected.

 - Also might be related to numpy version, but `.astype()` does not seem to change the value of an array when doubling bit depth as in here : 
 https://github.com/SpikeInterface/spikeinterface/blob/c2f18bd536437e9f500a4e24f7a0f3773bd79d3a/src/spikeinterface/preprocessing/unsigned_to_signed.py#L47-L51

I'm not sure why substracting an offset is necessary in this situation. In fact, when looking at processed traces with the correct data type but the current offset calculations, the traces look like this : 
offset
Which is about the calculated offset minus the 12 bits offset (32768 - 2048 = 30720).

 - Should the 12 bits offset compensation be on by default when open recordings through the BioCAM extractor ? Or at least some kind of warning when opening non-compensated recordings ?

 - I'm not sure in which use case this function is meant to be used, but are there any situations where having a signed data type might cause issues ?

### DIFF
--- a/src/spikeinterface/preprocessing/unsigned_to_signed.py
+++ b/src/spikeinterface/preprocessing/unsigned_to_signed.py
@@ -8,6 +8,7 @@ from .filter import fix_dtype
 class UnsignedToSignedRecording(BasePreprocessor):
     """
     Converts a recording with unsigned traces to a signed one.
+    
     Parameters
     ----------
     recording: Recording
--- a/src/spikeinterface/preprocessing/unsigned_to_signed.py
+++ b/src/spikeinterface/preprocessing/unsigned_to_signed.py
@@ -15,17 +15,18 @@ class UnsignedToSignedRecording(BasePreprocessor):
     def __init__(
         self,
         recording,
+        bit_depth=None,
     ):
         dtype = np.dtype(recording.dtype)
         assert dtype.kind == "u", "Recording is not unsigned!"
         itemsize = dtype.itemsize
         assert itemsize < 8, "Cannot convert uint64 to int64."
-        dtype_signed = dtype.str.replace("uint", "int")
+        dtype_signed = dtype.name.replace("uint", "int")
 
         BasePreprocessor.__init__(self, recording, dtype=dtype_signed)
 
         for parent_segment in recording._recording_segments:
-            rec_segment = UnsignedToSignedRecordingSegment(parent_segment, dtype_signed)
+            rec_segment = UnsignedToSignedRecordingSegment(parent_segment, dtype_signed, bit_depth)
             self.add_recording_segment(rec_segment)
 
         self._kwargs = dict(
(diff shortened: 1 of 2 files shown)


src/spikeinterface/preprocessing/unsigned_to_signed.py
```diff
@@ -15,17 +15,18 @@ class UnsignedToSignedRecording(BasePreprocessor):
     def __init__(
         self,
         recording,
+        bit_depth=None,
```
alejoe91, line 18: If None, it should automatically use the full range IMO:

e.g., `int16` --> `bit_dept=16`

What do you think @BptGrm ?

alejoe91, line 18: Can you add docstrings as well below the class definition to explain this new behavior?

src/spikeinterface/preprocessing/unsigned_to_signed.py
```diff
@@ -46,7 +48,11 @@ def get_traces(self, start_frame, end_frame, channel_indices):
         traces_dtype = traces.dtype
         nbits = traces_dtype.itemsize * 8
         signed_dtype = f"int{2 * (traces_dtype.itemsize) * 8}"
-        offset = 2 ** (nbits - 1)
+        if self.bit_depth is not None:
+            offset = 2 ** (self.bit_depth - 1)
+        else:
+            # offset = 2 ** (nbits - 1)
+            offset = 0
```
alejoe91, line 55:
```suggestion
            offset = 2 ** (nbits - 1)
```
The old behavior should be default. In general, we can assume ADCs are used full range

alejoe91 (changes requested):
See comments

BptGrm:
Sorry, I didn't fully understand the default behavior before. Should be clearer now.

alejoe91:
Thansk for the changes! Last request: could you add a small test here that uses the new `bit_depth` argument?

https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/preprocessing/tests/test_unsigned_to_signed.py

BptGrm:
I hope it's enough, tests are passing for me locally.

alejoe91 (approved):

alejoe91 (approved):

PR #2441, opened 2024-01-26
Add support for NWB-Zarr

alejoe91:
Replaces #2404 

Refactor NWBExtractors to:

- Add support for NWB-Zarr (and tests)
- Remove composite design with `_NwbHDF5`/`_NwbPynwb` desing
- Simplified NwbExtractors implementations at several places

Changes inherited from #2404 
- Specify `unit_table_path` (default `units`)
- Changed `electrical_series_name` to `electrical_series_path`


src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -166,135 +249,453 @@ def read_nwbfile(
...
+    else:
+        raise RuntimeError(f"File {file_path} not found or not readable.")
+    return backend
+
+
+def find_neurodata_type_from_backend(group, path="", result=None, neurodata_type="ElectricalSeries", backend="hdf5"):
     """
-    A RecordingExtractor for NWB files. This uses the NWB API to extract the traces and
-    the metadata and is called by the NwbRecordingExtractor factory.
+    Recursively searches for groups with neurodata_type 'ElectricalSeries' in the HDF5 group or file,
```
alejoe91, line 319:
```suggestion
    Recursively searches for groups with the specified neurodata_type in the HDF5 group or file,
```

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import annotations
 from pathlib import Path
+from turtle import st
```
h-mayorquin, line 3: ?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -97,40 +160,60 @@ def _read_hdf5_file(
...
-    elif file_path is not None:
+    elif stream_mode == "zarr":
+        import zarr
+
+        open_file = zarr.open(file_path, mode="r", storage_options=storage_options)
+
+    elif file_path is not None:  # local
         file_path = str(Path(file_path).resolve())
-        hdf5_file = h5py.File(name=file_path, mode="r")
+        backend = get_backend_from_local_file(file_path)
```
h-mayorquin, line 190: This is strange, you have a backend parameter that you don't use but determine from this other function.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -97,40 +160,60 @@ def _read_hdf5_file(
...
         import remfile
+        import h5py
 
         assert file_path is not None, "file_path must be specified when using stream_mode='remfile'"
         rfile = remfile.File(file_path)
-        hdf5_file = h5py.File(rfile, "r")
+        open_file = h5py.File(rfile, "r")
 
-    elif file_path is not None:
+    elif stream_mode == "zarr":
```
h-mayorquin, line 183: I find it a bit confusiong how zarr is a stream mode and a backend at the same time. How does zarr works as  a stream mode?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -166,102 +249,345 @@ def read_nwbfile(
...
+        from hdmf_zarr import NWBZarrIO
+
+        io = NWBZarrIO(path=open_file.store, mode="r", load_namespaces=True)
+
     nwbfile = io.read()
     return nwbfile
 
 
-class _NwbPynwbRecordingExtractor(BaseRecording):
+def get_backend_from_local_file(file_path: str | Path) -> str:
```
h-mayorquin, line 281: This can be done with magic numbers instead of try-except:

```python
def is_hdf5_file(filename):
    # Source for magic numbers https://www.loc.gov/preservation/digital/formats/fdd/fdd000229.shtml
    # We should find a better one though
    with open(filename, 'rb') as f:
        file_signature = f.read(8)
    return file_signature == b'\x89HDF\r\n\x1a\n'
```

h-mayorquin (commented):
First batch of comments to take advantage that Alessio is on now. Will keep adding more.

src/spikeinterface/extractors/nwbextractors.py
alejoe91, line 3:
```suggestion
```
stupid vs code extensions

src/spikeinterface/extractors/nwbextractors.py
alejoe91: zarr can read directly from cloud buckets! But I see your point. I'll think of a better solution

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -590,173 +874,6 @@ def __init__(
...
-    """
-
-    extractor_name = "NwbRecording"
-    mode = "file"
-    name = "nwb"
-
-    def __new__(
-        cls,
-        file_path: str | Path | None = None,  # provide either this or file
-        electrical_series_name: str | None = None,
```
h-mayorquin, line 723: We probably should  deprecate  ` electrical_series_name`  instead of removing it outright.

alejoe91: agreed

alejoe91: and in case it's used we can automatically set `electrical_series_path=f"acquisition/{electrical_series_name}"`

h-mayorquin: is that consistent with the old behavior? The old behavior looks for the name everywhere not just in acquisition I think.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -12,21 +12,22 @@
...
-    global NWBFile, ElectricalSeries, NWBHDF5IO
+    global NWBFile, ElectricalSeries, Units, NWBHDF5IO
     from pynwb import NWBFile
     from pynwb.ecephys import ElectricalSeries
+    from pynwb.misc import Units
     from pynwb import NWBHDF5IO
 
 
-def retrieve_electrical_series(nwbfile: NWBFile, electrical_series_name: Optional[str] = None) -> ElectricalSeries:
+def retrieve_electrical_series(nwbfile: NWBFile, electrical_series_path: Optional[str] = None) -> ElectricalSeries:
```
h-mayorquin, line 22: I like the convention that you have on the other functions of using _pynwb or _backend to divided the modes. I also think that we should probably make all of these methods private

src/spikeinterface/extractors/tests/test_nwbextractors_streaming.py
```diff
@@ -254,6 +254,45 @@ def test_sorting_s3_nwb_remfile(tmp_path):
     check_sortings_equal(reloaded_sorting, sorting)
 
 
+@pytest.mark.streaming_extractors
+@pytest.mark.skip("Too slow")
+def test_sorting_s3_nwb_zarr(tmp_path):
+    file_path = (
```
h-mayorquin, line 260: Where are these? As in who ise the owner of those files?

src/spikeinterface/extractors/tests/test_nwbextractors.py
```diff
@@ -110,29 +110,80 @@ def nwbfile_with_ecephys_content():
...
+        io_class = NWBHDF5IO
+    elif backend == "zarr":
+        io_class = NWBZarrIO
+    with io_class(str(file_path), mode="w") as io:
+        io.write(nwbfile)
+    return file_path, nwbfile
 
-    return nwbfile_path
+
+@pytest.fixture(scope="module", params=["hdf5", "zarr"])
```
h-mayorquin, line 166: Nice!

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1002,16 +1101,107 @@ def __init__(
...
+                self.internal_sampling_frequency = 1.0 / np.median(np.diff(timestamps[:samples_for_rate_estimation]))
+
+        assert (
+            self.internal_sampling_frequency is not None
+        ), "Couldn't load sampling frequency. Please provide it with the 'sampling_frequency' argument"
+        assert (
+            self.t_start is not None
+        ), "Couldn't load a starting time for the sorting. Please provide it with the 't_start' argument"
+
+        if unit_table_path is None:
```
h-mayorquin, line 1162: This is a nice pattern. You add computation when an error to make the life of the user easier.

src/spikeinterface/extractors/nwbextractors.py
alejoe91: probably not, but it was ambiguous. I think this is a better solution. If it fails, it will print the available paths and users can use the paths instead

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -798,201 +915,183 @@ def get_traces(self, start_frame, end_frame, channel_indices):
...
-        file_path: str | Path,
-        electrical_series_name: str | None = None,
-        sampling_frequency: float | None = None,
-        samples_for_rate_estimation: int = 1000,
-        stream_mode: str | None = None,
-        stream_cache_path: str | Path | None = None,
-        *,
-        t_start: float | None = None,
-        cache: bool = False,
+    def __del__(self):
```
h-mayorquin, line 1045: Good idea to have these methods.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -798,201 +915,183 @@ def get_traces(self, start_frame, end_frame, channel_indices):
...
+        cache: bool = False,
+        storage_options: dict | None = None,
+        use_pynwb: bool = False,
+    ):
+        self.stream_mode = stream_mode
+        self.stream_cache_path = stream_cache_path
+        self.electrical_series_path = electrical_series_path
+        self.file_path = file_path
+        self.t_start = t_start
+        self.internal_sampling_frequency = sampling_frequency
```
h-mayorquin, line 989: why internal sampling frequency? in opposition to external?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -798,201 +915,183 @@ def get_traces(self, start_frame, end_frame, channel_indices):
...
+            self.backend = get_backend_from_local_file(file_path)
+        else:
+            if self.stream_mode == "zarr":
+                self.backend = "zarr"
+            else:
+                self.backend = "hdf5"
 
-        if "unit_name" in units_table:
-            unit_ids = units_table["unit_name"]
+        if use_pynwb:
```
h-mayorquin, line 1001: Like Sam I think that the normal import errors are enough and we don't require special messages for these ones.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -798,201 +915,183 @@ def get_traces(self, start_frame, end_frame, channel_indices):
...
-        assert (
-            self.t_start is not None
-        ), "Couldn't load a starting time for the sorting. Please provide it with the 't_start' argument"
+    def __init__(
+        self,
+        file_path: str | Path,
+        electrical_series_path: str | None = None,
+        sampling_frequency: float | None = None,
+        samples_for_rate_estimation: int = 1_000,
+        unit_table_path: str = "units",
```
h-mayorquin, line 975: I would like to have these path after the *; to make it keyword only but do I remember well from the discussion with Cody that you don't like them?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -166,102 +248,345 @@ def read_nwbfile(
...
+
+    if load_time_vector and timestamps is not None:
+        times_kwargs = dict(time_vector=electrical_series.timestamps)
+    else:
+        times_kwargs = dict(sampling_frequency=sampling_frequency, t_start=t_start)
+
+    return sampling_frequency, times_kwargs
+
+
+def retrieve_electrodes_indices_from_electrical_series(open_file, electrical_series, backend="hdf5"):
```
h-mayorquin, line 376: This seems to be the only place where zarr and hdf5 backend really diffe, right? I would add backend as a suffix to the name and make it a private method as well.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -38,48 +39,109 @@ def retrieve_electrical_series(nwbfile: NWBFile, electrical_series_name: Optiona
...
+        If no unit tables are found in the NWBFile or if multiple unit tables are found but no unit_table_path
+        is provided.
+    AssertionError
+        If the specified unit_table_path is not present in the NWBFile.
+    """
+    from pynwb.misc import Units
+
+    unit_table_dict: Dict[str:Units] = {}
+
+    for item in nwbfile.all_children():
```
h-mayorquin, line 105: This does the search even if the units_table_path is passed.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -798,201 +915,183 @@ def get_traces(self, start_frame, end_frame, channel_indices):
...
         assert (
-            sampling_frequency is not None
+            self.internal_sampling_frequency is not None
         ), "Couldn't load sampling frequency. Please provide it with the 'sampling_frequency' argument"
         assert (
             self.t_start is not None
         ), "Couldn't load a starting time for the sorting. Please provide it with the 't_start' argument"
 
-        units_table = self._nwbfile.units
+        units_table = retrieve_unit_table(self._nwbfile, unit_table_path=unit_table_path)
```
h-mayorquin, line 1094: The pynwb double route still performs the search even if the units table path is passed I think.

h-mayorquin (commented):
Another batch of comments.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: The downside of this is that you might break working code. If you are willing to do that I would rather just thrown an error when someone passess the argument, do a serach and give them the possible paths they can enter.

h-mayorquin: Mmm now that I think about it your proposed solution does avoids breaking some working code (not all of it) which seems better.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: The problem right now is that the streaming for rem and fsspec seem to be coupled to the hdf5 backend which does not need to the case. We should thrown an error when someone passess those streaming methods with zarr and try to fix that later as this PR is very big for this and we probably don't have files for testing this?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -271,49 +596,130 @@ def __init__(
...
+
+        else:
+            channel_ids = [electrodes_table["id"][x] for x in electrodes_indices]
+
+        dtype = electrical_series["data"].dtype
+        electrical_series_data = electrical_series["data"]
+
+        # need this for later
+        self.electrical_series = electrical_series
+        self._file = open_file
```
h-mayorquin, line 682: Good to make this a private attribute as it is only available on these path. Altought I think we could get the file from the nwbfile in the other path and exposed it as well but probably not worth the effort.

h-mayorquin (changes requested):
OK, I think I am done.

This is a lot of work, thanks for that. It is a nice improvement that we now support zarr!. I only have two request and the rest, as all the comments. are suggestions of style that  we can work later if we want.

Requests:
* Add backwards compatability for the `ElectricalSereisName`. Note that the fact that it was a positional argument is problematic. Note as well that adding the path as a positional argument breaks code that was using the other parameters positionally so the path argument should not be added where it currently is. I suggest keyword but the other option is to add it as a position parameter at the end of the position parameters.
* I would like to break the `add_extractor_info_pynwb` into two parts. The first part should be in charge of the gains and offsets and the second part should be in charge of the properties. The main reason I want this is that for streaming it it useful to have the possibility of not extracting the properties and it would be way easier if that step is in a function of its own. Besides, I think the name of those functions would be way more percise than the current one: `fetch_gain_and_offests_pynw/backend` and `fetch_extractor_properties_pynwb/backend` (way more specific than `add_extractor_info` ). Same for the analogous on the sorting case.


Suggestions of style, feel free to ignore them or we can address them in a later PR:
* The name of the function `extract_info_backend` is a bit unclear. What you seem to be doing is that you are extracting the segment_kwargs so maybe you can name it like that. Another option is to break the function into smaller pieces such as "fetch_channel_ids", "fetch_time_info" and "fetch_electrical_series_data" from which you can get the dtype.
* I would have the functions extract the information and then do the setting in the main loop. What do I mean? The functions `add_extractor_info_backend` and `add_extractor_info_pynwb` have duplicated code at the end. I think it is better if they return the properties and then you set the properties in the main `__init__`. These avoids duplicating code and also separates fetching from setting which I think is a good idea.
* The comment above of making all the functions that we are adding private .I made the mistakes of not doing this and as we are changin this quite often we don't want anyone relying on this.

alejoe91:
Thank you so much Ramon! These are all super valid points. I'll implement them first thing on Monday ;)

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Note that this will be way faster specially in the streaming case.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: What does this one do? : O

zm711: I believe that is the command to show a turtle. Haha. What an Easter egg. Some goes to read an NWB file and a turtle pops up on their screen....

src/spikeinterface/extractors/nwbextractors.py
alejoe91: but how do you know if the remote location is hdf5 or zarr?

src/spikeinterface/extractors/nwbextractors.py
alejoe91: Thanks!

alejoe91: Note that this is only run in local mode. For streaming, the streaming mode will "know" its backend

src/spikeinterface/extractors/nwbextractors.py
alejoe91: might be useful externally as well IMO

src/spikeinterface/extractors/tests/test_nwbextractors_streaming.py
alejoe91: The are in a public bucket, so tests should run fine!

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: What do you mean? If it is fsspec how do you know if you are streaming a h5 or zarr or something else?

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: They might but I don't want to mantain or keep the promise of not changing their signature so if people use them they should be aware that we might change them. Hence private.

src/spikeinterface/extractors/nwbextractors.py
alejoe91: `fsspec` is only available for HDF5. The `zarr` streaming mode makes use of `fsspec` too, but this is handled by the Zarr API directly

h-mayorquin: Fsspec is not related at all to hdf5 is it? It is just a way to establis a file system in a remote location with a consistent API.

src/spikeinterface/extractors/nwbextractors.py
alejoe91: initially it was `self._sampling_frequency` (or `self.sampling_frequency`), but these are used by the `BaseSorting` object, so we need another name.

src/spikeinterface/extractors/tests/test_nwbextractors_streaming.py
h-mayorquin: I was just curios if they were yours or where did you find them.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Ah, got you. It is for having it as an attribute. What about `electrical_series_sampling_frequency`?

src/spikeinterface/extractors/tests/test_nwbextractors_streaming.py
alejoe91: yes, they are from a workshop we made in May with the Allen Institute :)

src/spikeinterface/extractors/nwbextractors.py
alejoe91: In our usage, fsspec is used to open a remote h5 file, so we don't provide a general API to read any remote file with fsspec, but only to read remote NWB h5 files.

h-mayorquin: Yeah, I did not generalize that function enough earlier when I made it. Ideally the file should be fetched and then either open with zarr or hdf5. But that's a matter for another PR I feel.

alejoe91:
@h-mayorquin implemented most of the changes! Thank you so much for the deep review :)

We only have to figure out the `zarr` streaming mode and then this is good to go on my side!

src/spikeinterface/extractors/nwbextractors.py
alejoe91: I think there is no need to do that with zarr. `file = zarr.open("s3:/...")` will automatically stream using fsspec, so we shouldn't reinvent the wheel there :)

h-mayorquin (approved):
I moved the new arguments to the end as positional and changed the code so search is not performed for the units table default case in the pynwb pathway.

h-mayorquin:
It seems that the failure of the test is unrelated.

h-mayorquin:
Should we merge this? I can use this a bit with the streaming code before the release just to add some field test : )

alejoe91:
@samuelgarcia promised to approve tomorrow morning. I'd also like to merge this ASAP

h-mayorquin:
Ah, I was not aware that @samuelgarcia intended to take a look. That's fine.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -166,209 +146,453 @@ def read_nwbfile(
...
         samples_for_rate_estimation: int = 1_000,
-        stream_mode: Optional[Literal["fsspec", "ros3", "remfile"]] = None,
+        stream_mode: Optional[Literal["fsspec", "ros3", "remfile", "zarr"]] = None,
         stream_cache_path: str | Path | None = None,
+        electrical_series_path: str | None = None,
+        load_channel_properties: bool = True,
         *,
         file: BinaryIO | None = None,  # file-like - provide either this or file_path
         cache: bool = False,
+        storage_options: dict | None = None,
```
samuelgarcia, line 504: Can this be propagated to kawrgs and the  dumped ?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -377,16 +601,14 @@ def __init__(
...
-
         # set serializability bools
         if file is not None:
             # not json serializable if file arg is provided
             self._serializability["json"] = False
 
         self._kwargs = {
             "file_path": file_path,
-            "electrical_series_name": self._electrical_series_name,
+            "electrical_series_path": self.electrical_series_path,
```
samuelgarcia, line 611: This is super annoying because we will no be able to open nwb recording dump to json from actual relase in next relase.
Could you make some manual tests (automatic test is too complicated)  on cross version compatibility ?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -166,209 +146,453 @@ def read_nwbfile(
...
         self,
         file_path: str | Path | None = None,  # provide either this or file
-        electrical_series_name: str | None = None,
+        electrical_series_name: str | None = None,  # deprecated
         load_time_vector: bool = False,
         samples_for_rate_estimation: int = 1_000,
-        stream_mode: Optional[Literal["fsspec", "ros3", "remfile"]] = None,
+        stream_mode: Optional[Literal["fsspec", "ros3", "remfile", "zarr"]] = None,
         stream_cache_path: str | Path | None = None,
+        electrical_series_path: str | None = None,
```
samuelgarcia, line 499: to kwargs also.

src/spikeinterface/extractors/nwbextractors.py
samuelgarcia: Forget it. The kwargs is left in the signautre.

src/spikeinterface/extractors/nwbextractors.py
samuelgarcia: I would move this at the same place as `electrical_series_name`

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -395,74 +617,100 @@ def __init__(
             "file": file,
         }
 
+    def __del__(self):
+        # backend mode
+        if hasattr(self, "_file"):
+            if hasattr(self._file, "store"):
+                self._file.store.close()
+            else:
+                self._file.close()
```
samuelgarcia, line 626: This is not done automatically when un referencing ?

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: You mean in the signature? I argued the opposite because if someone was relying on the order of position arguments it will break their code.

alejoe91:
@h-mayorquin after discussion with Sam, we decided to be cautious about propagating `storage_options`, since they could contain credentials. In that case, the object will not be JSON/pickle serializable.

samuelgarcia (approved):

PR #2443, opened 2024-01-29
Fix bug in `mergeunits`

DradeAW:
`ind` is not defined if you are in the else ....

Really should have been picked up in the tests.


alejoe91 (approved):

alejoe91:
Thanks @DradeAW

PR #2445, opened 2024-01-29
[IO] added exclude_filename argument to NeuralynxRecordingExtractor

Antho2422:
Added the possibility when loading from neuralynx to give a list of files to ignore to match Neo's arguments.

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/neuralynx.py
+++ b/src/spikeinterface/extractors/neoextractors/neuralynx.py
@@ -24,6 +24,7 @@ class NeuralynxRecordingExtractor(NeoBaseRecordingExtractor):
         Load exhaustively all annotations from neo.
     exlude_filename: list[str], default: None
         List of filename to exclude from the loading.
+        For example, use `exclude_filename=["events.nev"]` to skip loading the event file.
     """
 
     mode = "folder"
--- a/src/spikeinterface/extractors/neoextractors/neuralynx.py
+++ b/src/spikeinterface/extractors/neoextractors/neuralynx.py
@@ -22,22 +22,24 @@ class NeuralynxRecordingExtractor(NeoBaseRecordingExtractor):
         If there are several streams, specify the stream name you want to load.
     all_annotations: bool, default: False
         Load exhaustively all annotations from neo.
+    exlude_filename: list[str], default: None
+        List of filename to exclude from the loading.
     """
 
     mode = "folder"
     NeoRawIOClass = "NeuralynxRawIO"
     name = "neuralynx"
 
-    def __init__(self, folder_path, stream_id=None, stream_name=None, all_annotations=False):
-        neo_kwargs = self.map_to_neo_kwargs(folder_path)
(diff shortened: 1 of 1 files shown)


Antho2422:
#2445

src/spikeinterface/extractors/neoextractors/neuralynx.py
```diff
@@ -22,22 +22,24 @@ class NeuralynxRecordingExtractor(NeoBaseRecordingExtractor):
...
     name = "neuralynx"
 
-    def __init__(self, folder_path, stream_id=None, stream_name=None, all_annotations=False):
-        neo_kwargs = self.map_to_neo_kwargs(folder_path)
+    def __init__(self, folder_path, stream_id=None, stream_name=None, all_annotations=False, exclude_filename=None):
+        neo_kwargs = self.map_to_neo_kwargs(folder_path, exclude_filename)
         NeoBaseRecordingExtractor.__init__(
             self, stream_id=stream_id, stream_name=stream_name, all_annotations=all_annotations, **neo_kwargs
         )
         self._kwargs.update(dict(folder_path=str(Path(folder_path).absolute())))
```
alejoe91, line 38:
```suggestion
        self._kwargs.update(dict(folder_path=str(Path(folder_path).absolute()), exclude_filename=exclude_filename))
```
This is not a "default" neo argument (like `stream_name` etc.), so it should be appended to the `self._kwargs`

src/spikeinterface/extractors/neoextractors/neuralynx.py
```diff
@@ -22,22 +22,24 @@ class NeuralynxRecordingExtractor(NeoBaseRecordingExtractor):
         If there are several streams, specify the stream name you want to load.
     all_annotations: bool, default: False
         Load exhaustively all annotations from neo.
+    exlude_filename: list[str], default: None
```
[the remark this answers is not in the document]
Antho2422: Yes ofc, where do you want me to do that ?

src/spikeinterface/extractors/neoextractors/neuralynx.py
```diff
@@ -22,22 +22,24 @@ class NeuralynxRecordingExtractor(NeoBaseRecordingExtractor):
...
+        List of filename to exclude from the loading.
```
alejoe91, line 26:
```suggestion
        List of filename to exclude from the loading.
        For example, use `exclude_filename=[".nev"]` to skip loading the event files.
```
Not sure this is accurate, but something along these lines ;)

Antho2422: I am not sure this is working I think you need to provide a full filename but a will test it asap thank you

h-mayorquin: Yes, Alessio nailed it. Something like that. It can be a docstring example if you prefer that.

I think this will make it easier for users to identify how and why to use this parameter : )

samuelgarcia (approved):

PR #2458, opened 2024-02-01
Use core packages on core tests

h-mayorquin:
After some failing tests I realized that we are not using the latest dev of probeinterface and neo on the core tests. I guess the reason was that neo does not make much sense but probeinterface does as it is tightly integrated to the core.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,6 +113,11 @@ qualitymetrics = [
 test_core = [
     "pytest",
     "psutil",
+
+    # for github test : probeinterface and neo from master
+    # for release we need pypi, so this need to be commented
+    "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",
+    "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",
 ]
 
 test = [


alejoe91 (approved):

pyproject.toml
```diff
@@ -113,6 +113,11 @@ qualitymetrics = [
 test_core = [
     "pytest",
     "psutil",
+
+    # for github test : probeinterface and neo from master
+    # for release we need pypi, so this need to be commented
+    "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",
+    "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",
```
samuelgarcia, line 120: I am OK with this but core test are not using neo extractor at all.

pyproject.toml
```diff
@@ -113,6 +113,11 @@ qualitymetrics = [
 test_core = [
     "pytest",
     "psutil",
+
+    # for github test : probeinterface and neo from master
+    # for release we need pypi, so this need to be commented
+    "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",
```
samuelgarcia, line 119: OK

samuelgarcia (approved):

PR #2459, opened 2024-02-01
avoid double parsing in Plexon

h-mayorquin:
A small improvement for Plexon to avoid double parsing. This inefficiency was pointed out by @weiglszonja.

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/plexon.py
+++ b/src/spikeinterface/extractors/neoextractors/plexon.py
@@ -32,7 +32,7 @@ class PlexonRecordingExtractor(NeoBaseRecordingExtractor):
         NeoBaseRecordingExtractor.__init__(
             self, stream_id=stream_id, stream_name=stream_name, all_annotations=all_annotations, **neo_kwargs
         )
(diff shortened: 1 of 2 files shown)


src/spikeinterface/extractors/neoextractors/plexon.py
```diff
@@ -60,10 +60,9 @@ class PlexonSortingExtractor(NeoBaseSortingExtractor):
     def __init__(self, file_path):
         neo_kwargs = self.map_to_neo_kwargs(file_path)
         self.neo_reader = NeoBaseSortingExtractor.get_neo_io_reader(self.NeoRawIOClass, **neo_kwargs)
-        self.neo_reader.parse_header()
```
h-mayorquin, line 63: This is already pasred in the method above (62) that gets the neo reader.

alejoe91 (approved):

src/spikeinterface/extractors/neoextractors/plexon.py
samuelgarcia: thanks for the catch. Very strange why we have this here.
I think this is me sorry.

samuelgarcia (approved):

PR #2468, opened 2024-02-05
Add from __future__ import annotations to all src/**.py files

alejoe91:
@DradeAW brute force approach here. Can you test on your side?


DradeAW:
I can confirm it fixes my bug!

Thanks Alessio <3

samuelgarcia (approved):

PR #2475, opened 2024-02-07
Use `has_channel_location` instead of `get_property("contact_vector")`

alejoe91:
The `set_channel_locations` function DOES NOT set a probe.

Many functions downstream look for the `contact_vector`, and this would not be set.


alejoe91:
@samuelgarcia ready tp merge. Should I change the target to `main` then?

samuelgarcia (approved):

samuelgarcia:
yep

PR #2480, opened 2024-02-12
Analyzer

samuelgarcia:


samuelgarcia:
@alejoe91 : I merge this now because @yger  need it in dev to remove WaveformsEtractor from template matching.

@yger : compute_sparsity now also works with Templates for snr when noise_levels is given.

PR #2485, opened 2024-02-13
Remove WaveformExtractor from Templates matching

yger:
Enforcing all objects to be Templates for the matching engines in the sortingcomponents. Note that this is a draft since sparsity has to be better handled when @samuelgarcia will expose all sparsity methods for Templates


src/spikeinterface/core/template.py
```diff
@@ -108,6 +108,11 @@ def __post_init__(self):
                 if not self._are_passed_templates_sparse():
                     raise ValueError("Sparsity mask passed but the templates are not sparse")
 
+    def set_sparsity(self, sparsity):
```
samuelgarcia, line 111: We should ensure that the template is dense before this no ?

src/spikeinterface/core/sparsity.py
```diff
@@ -220,13 +220,15 @@ def are_waveforms_sparse(self, waveforms: np.ndarray, unit_id: str | int) -> boo
 
         return int(excess_zeros) == 0
 
-    def sparisfy_templates(self, templates_array: np.ndarray) -> np.ndarray:
+    def sparsify_templates(self, templates_array: np.ndarray) -> np.ndarray:
         max_num_active_channels = self.max_num_active_channels
-        sparisfied_shape = (self.num_units, self.num_samples, max_num_active_channels)
-        sparse_templates = np.zeros(shape=sparisfied_shape, dtype=templates_array.dtype)
+        num_samples = templates_array.shape[1]
+        sparsified_shape = (self.num_units, num_samples, max_num_active_channels)
```
samuelgarcia, line 226:
```suggestion
        sparsified_shape = (self.num_units, self.num_samples, max_num_active_channels)
```

samuelgarcia (approved):

PR #2488, opened 2024-02-14
Templates matching

yger:
The sparsification in dev is not working, and/or maybe did I missed something? num_samples is not defined here, not sure what happened in your branch @samuelgarcia


samuelgarcia:
@yger : this is fixed in my branch I will merge it now

PR #2513, opened 2024-02-26
Added `round` option to `recording.astype`

DradeAW:
By default `False` to mimic numpy's `astype`


src/spikeinterface/preprocessing/astype.py
```diff
@@ -19,6 +21,7 @@ def __init__(
         self,
         recording,
         dtype=None,
+        round: bool = False,
```
samuelgarcia, line 24: I would put True by default. no ?

DradeAW: I also think it makes more sense for spike-sorting, but @h-mayorquin made a good point that users might expect the same behaviour as numpy.

I'm fine with both

src/spikeinterface/preprocessing/astype.py
```diff
@@ -19,20 +21,26 @@ def __init__(
         self,
         recording,
         dtype=None,
+        round: bool | None = None,
```
samuelgarcia, line 24: Can you add a small doc please ?

DradeAW: Done :)

src/spikeinterface/preprocessing/astype.py
alejoe91: Yes, because rounding is what we normally want in this case. Even though that deviates from numpy behavior, that's better for ephys processing :)

h-mayorquin: I saw that you decided that you wanted to round by default. I am asking why the default ended up being None instead of `True` as Sam requested.

We ended up with three cases intead of two and I don't understad why.

The three cases are:
* `round=True` : rounds everything
* `round=False` : numpy default behavior
* `round=None` : rounds only with int.

Why not just True and False and and True by default as Sam requested. I don't understand why this complication was added and that's what I am asking about.

alejoe91: Because what do you do if the input is float and round is True?

h-mayorquin: I think you answering with a question instead of an statement makes it more difficult for me to understand.  I am thinking as I write but I feel that something is missing.

Let's make a list to see if I can get you.
* recording dtype is int and you cast to float -> rounding by defaut should not make a difference.
* recording dtype is float and you cast to int -> runding by default was the reason this was added
* recording dtype is int and you cast to int -> rounding by default should not make a difference.
* recording dtype is float and you cast to float -> this would be problematic is someone wanted to cast float 64 to float 32.

Is that last case the problematic one? Am I understanding correctly?

alejoe91: Yes. In that case the logic would return a `float64` but rounded. Do you agree?

h-mayorquin: Right. Thanks for explaining.

I propose the following:

I think we can simplify this by making round (preferably called `round_float_to_int` to be fully descriptive) only True and False.

it should be `True` bye default as you guys wanted and it can only take two values:
* When `True` and the input is float, we round before casting to int. If the input is int or the output is float we don't do anything and the function behaes as numpy astype.
* When False it defaults to numpy astype behavior for all the cases.

This fullfils the original purpose and it has some advantages over the current approach:

Advantages:
* Easier to document. The behavior is fully described by what I wrote above. We can enchance this by adding examples of why we did this (i.e. inclusive of 0 in np.hisotgram).
* Avoids having the third unecessary branch of `round=True` that we have now. 
* Simplifies the type of `round`  / `round_float_to_int` by making it boolean only.

DradeAW: If you change it to `round_float_to_int`, then you remove the possibility that someone, for some reason, would want to round but keep it as a float.

I don't see what the problem would be when converting from `float64` to `float32`?*
By default there would be no rounding (as there is no need), but the user ca explicitly ask for it if he wants to

h-mayorquin: Yeah, I want to reduce the generality of the code to make it easier. Less general code is good. Easier to understand, mantain docstrings (which are lacking in this case) and write tests if needed. Plus, in this case I can't really see a use case for the generalization, that is, it does not buy us anything other than generality for the sake of generality. If the use case, then we will be better position to write a general implementation with concrete use case in mind.

The cost of generality on the other hand? it took me some time to understand what each of the options does and where and why I would want them. Not all cases are documented which illustrates my point that is more work We can save a future developer time by making this more concrete and save ourselves the work to document cases that are unlikely to be used. Plus, it is is way easier to document something that is more concrete. Case in point: try writing a docstring for the more general case of the CAR preprocessor, no one has done it for a reason.

PR #2522, opened 2024-02-28
Relax fname requirements for IBL cbin

alejoe91:
The IBL cbin doesn't always comply with the SpikeGLX naming convention


samuelgarcia:
This will integrated in  #2512

PR #2526, opened 2024-02-29
Fix sparsity check for old waveform extractor

chrishalcrow:
Fix bug when reading old dense waveforms.

Only sparse waveforms have a `sparsity.json` file so this should be checked to decide sparsity, rather than the parameter file.


samuelgarcia (approved):

samuelgarcia:
Merci!

PR #2528, opened 2024-02-29
Refactor, test and optimize compute_synchrony_metrics

chrishalcrow:
Planning to standardise, correctness test and optimise the quality metrics module. First one: the `compute_synchrony_metrics function`.

I'm pretty new to this, so all feedback will be listened to and much appreciated! I think @samuelgarcia agreed to give me some detailed feedback in the future, so I'll put this on draft until then.

Two main things in this:
- The generating function `synthesize_random_firings` and `synthesize_poisson_spike_vectors` were adjusted so that you can insert specific spikes, useful for testing. These are used in the new tests.
- A new function `compute_synchrony_counts` is introduced, which computes the synchrony counts.

Some things I'm not sure about:
1. The documentation said that `compute_synchrony_metrics` returns a dict, but it previously outputted a named tuple. We can either change the documentation or the output. I changed the output, as most quality metrics functions return a dict.
2. I flattened any multi-segment spike trains into a single array, which avoids a for loop down the line. Is there a reason not to do this?
3. The algorithm assumes the spike train is sorted in time. Can this be assumed?

I adjusted the algorithm in `compute_synchrony_counts`, which should be easier to understand and is significantly faster. I benchmarked using the following code:

```
import numpy as np
from time import perf_counter

from spikeinterface.core import synthesize_random_firings
from spikeinterface.qualitymetrics import get_synchrony_counts

unit_ids = np.arange(100)

Ns = 10**np.array(np.arange(3,5.1,0.1))

times_by_size=[]
how_many = 100

for train_length in Ns:

    times_list = []

    for _ in range(0,how_many):

        spike_train = synthesize_random_firings(num_units=100, duration=train_length/1000, firing_rates=10.0)

        start = perf_counter()
        get_synchrony_counts(spike_train, np.array((2,4,8)), list(unit_ids))        
        end = perf_counter()

        times_list.append(end - start)

    times_by_size.append( np.median( times_list ) )
```

The results are below. I tabulated (half of) the actual times, and create a log-log plot to check the algorithm's complexity.

benchmarch_sync

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
 -   repo: https://github.com/psf/black
-    rev: 24.2.0
+    rev: 24.3.0
     hooks:
     -   id: black
         files: ^src/
--- a/doc/how_to/analyse_neuropixels.rst
+++ b/doc/how_to/analyse_neuropixels.rst
@@ -628,7 +628,7 @@ Note that our object is not persistent to disk because we use
 .. parsed-literal::
 
     SortingAnalyzer: 383 channels - 31 units - 1 segments - memory - sparse - has recording
-    Loaded 0 extenstions:
+    Loaded 0 extensions:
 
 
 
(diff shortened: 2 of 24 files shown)


chrishalcrow:
> @chrishalcrow not sure what happened here with the conflicts..maybe easier to make a new PR directly to main with the changes to the metrics?

Yes, I tried to merge with the main after the SortingAnalyzer change and have somehow created this mess. Will start again...

PR #2529, opened 2024-03-01
Kilosort4 Wrapper

### LINKED ISSUE #2527: Implement Kilosort4 wrapper
Kilosort4 is out!

- [x] make wrapper in `sorters` (#2529)
- [x] make Dockerfile and image (https://github.com/SpikeInterface/spikeinterface-dockerfiles/pull/70)

alejoe91:
Fixes #2527

Kilosort now has a SpikeInterface Wrapper (`RecordingExtractorAsArray`), so we don't need to save to bin!

I had to port subsections of the [`run_kilosort`](https://github.com/MouseLand/Kilosort/blob/main/kilosort/run_kilosort.py#L21) function to allow to:

- do/skip drift correction
- do/skip preprocessing

Added tests for all cases. 

The parameters and descriptions were ported from their [parameters](https://github.com/MouseLand/Kilosort/blob/main/kilosort/parameters.py) definition.

One last comment:
The `RecordingExtractorAsArray` concatenates multiple segment. I think we should handle this externally to "unconcatenate" the sorting output, so I set `handle_multi_segment` to False.

### DIFF
--- a/src/spikeinterface/sorters/external/kilosort4.py
+++ b/src/spikeinterface/sorters/external/kilosort4.py
@@ -17,7 +17,6 @@ class Kilosort4Sorter(BaseSorter):
     requires_locations = True
 
     _default_params = {
-        "n_chan_bin": 385,
         "nblocks": 1,
         "Th_universal": 9,
         "Th_learned": 8,
@@ -53,7 +52,6 @@ class Kilosort4Sorter(BaseSorter):
     }
 
     _params_description = {
-        "batch_size": "Number of samples included in each batch of data. Default value: 60000.",
         "nblocks": "Number of non-overlapping blocks for drift correction (additional nblocks-1 blocks are created in the overlaps). Default value: 1.",
         "Th_universal": "Spike detection threshold for universal templates. Th(1) in previous versions of Kilosort. Default value: 9.",
         "Th_learned": "Spike detection threshold for learned templates. Th(2) in previous versions of Kilosort. Default value: 8.",
@@ -95,11 +93,9 @@ class Kilosort4Sorter(BaseSorter):
     For more information see https://github.com/MouseLand/Kilosort"""
 
     installation_mesg = """\nTo use Kilosort4 run:\n
-        >>> git clone https://github.com/MouseLand/Kilosort
-        >>> cd Kilosort
-        >>> pip install .
+        >>> pip install kilosort==4.0
 
-    More information on Kilosort4 and its installation procedure at:
+    More information on Kilosort4 at:
         https://github.com/MouseLand/Kilosort
     """
 
--- a/src/spikeinterface/sorters/external/kilosort4.py
+++ b/src/spikeinterface/sorters/external/kilosort4.py
@@ -88,14 +88,10 @@ class Kilosort4Sorter(BaseSorter):
         "scaleproc": "int16 scaling of whitened data, if None set to 200.",
     }
 
-    sorter_description = """Kilosort4 is a GPU-accelerated and efficient template-matching spike sorter. On top of its
-    predecessor Kilosort, it implements a drift-correction strategy. Kilosort3 improves on Kilosort2 primarily in the
-    type of drift correction we use. Where Kilosort2 modified templates as a function of time/drift (a drift tracking
-    approach), Kilosort3 corrects the raw data directly via a sub-pixel registration process (a drift correction
-    approach). Kilosort3 has not been as broadly tested as Kilosort2, but is expected to work out of the box on
-    Neuropixels 1.0 and 2.0 probes, as well as other probes with vertical pitch <=40um. For other recording methods,
-    like tetrodes or single-channel recordings, you should test empirically if v3 or v2.0 works better for you (use
-    the "releases" on the github page to download older versions).
+    sorter_description = """Kilosort4 is a Python package for spike sorting on GPUs with template matching.
+    The software uses new graph-based approaches to clustering that improve performance compared to previous versions.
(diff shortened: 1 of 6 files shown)


src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -0,0 +1,307 @@
...
+from typing import Union
+
+from ..basesorter import BaseSorter
+from .kilosortbase import KilosortBase
+
+PathType = Union[str, Path]
+
+
+class Kilosort4Sorter(BaseSorter):
+    """Kilosort3 Sorter object."""
```
zm711, line 14:
```suggestion
    """Kilosort4 Sorter object."""
```

:)

src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -0,0 +1,307 @@
...
+
+    sorter_description = """Kilosort4 is a GPU-accelerated and efficient template-matching spike sorter. On top of its
+    predecessor Kilosort, it implements a drift-correction strategy. Kilosort3 improves on Kilosort2 primarily in the
+    type of drift correction we use. Where Kilosort2 modified templates as a function of time/drift (a drift tracking
+    approach), Kilosort3 corrects the raw data directly via a sub-pixel registration process (a drift correction
+    approach). Kilosort3 has not been as broadly tested as Kilosort2, but is expected to work out of the box on
+    Neuropixels 1.0 and 2.0 probes, as well as other probes with vertical pitch <=40um. For other recording methods,
+    like tetrodes or single-channel recordings, you should test empirically if v3 or v2.0 works better for you (use
+    the "releases" on the github page to download older versions).
+    For more information see https://github.com/MouseLand/Kilosort"""
```
zm711, line 99:
```suggestion
    sorter_description = """Kilosort4 is a GPU-accelerated and efficient template-matching spike sorter. On top of its
    predecessor Kilosort, it implements a drift-correction strategy. Kilosort4 is a Python package for spike sorting on 
    GPUs with template matching. The software uses new graph-based approaches to clustering that improves 
    performance compared to previous versions. For other recording methods,
    like tetrodes or single-channel recordings, you should test empirically if v3 or v2.0 works better for you (use
    the "releases" on the github page to download older versions).
    For more information see https://github.com/MouseLand/Kilosort"""
```

Less of a hard suggestion and more just what they included themselves [here](https://kilosort.readthedocs.io/en/latest/). Don't know if the description here should include all the 3 vs 2 stuff... ? We could also have the link to their bioarxiv that explains all methods for all versions no?

alejoe91: Thanks Zach. Yes I didn't touch the sorter description yet but it's definitely worth including the preprint here. I'll add it!

zm711:
One other point about the save to binary that might be interesting is that Mountaintsort5 had a speed up by switching to using the cached binary ([see this](https://github.com/flatironinstitute/mountainsort5/issues/12)). Kilosort also recommends using a binary file as well (but more so that you can use their gui and phy which are not really our concerns per se). I'm wondering if it might make sense to give the user an option to do this with write to binary to make this completely compatible with a Kilosort4 workflow for people or not. I could see arguments either way (and SI already has the `export_to_phy` so it doesn't really matter. Just curious what you think about this Alessio.

alejoe91:
Thanks Zach, that's a good point especially regarding the phy output. In my view, we should be as agnostic as possible to the sorter and just use the spike times outputs, so we can better compare and unify pipelines regardless of the sorter. In this spirit, I would rather not add this option and "force" the user to postprocess and export in SI. What do you think?

zm711:
I think that's completely fair for the library. :) I guess if there are a lot of issues requesting something like that then there would be an impetus to do it and it could be reconsidered, but I agree this is the better initial choice. I think keeping the library agnostic is a good guiding principle, so thanks for framing it that way. Makes a lot of sense.

(And I already do all my postprocessing here anyway :) ).

mhhennig:
Looks like the docker image is private right now - but with a pip installed KS4 this works as intended.

alejoe91:
> Looks like the docker image is private right now - but with a pip installed KS4 this works as intended.

It's not pushed to docker hub yet, so it would need to be built locally ;) I'll test it a bit more and publish it

samuelgarcia:
Well done camarade.
Did they keep the same ouput format ?

alejoe91:
> Well done camarade. Did they keep the same ouput format ?

Yes, same old ;)

alejoe91:
Right! Will do that on Friday @samuelgarcia

alejoe91:
@samuelgarcia ready to merge on my side. Added installation docs

src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -0,0 +1,299 @@
...
+                artifact_threshold=artifact,
+                file_object=file_object,
+            )
+            ops["preprocessing"] = dict(hp_filter=None, whiten_mat=None)
+            ops["Wrot"] = torch.as_tensor(np.eye(recording.get_num_channels()))
+            ops["Nbatches"] = bfile.n_batches
+
+        np.random.seed(1)
+        torch.cuda.manual_seed_all(1)
+        torch.random.manual_seed(1)
```
samuelgarcia, line 220: Not a great way!

samuelgarcia:
I merge this now in the main.
We will improve this in futuir PR.
So the bug fix 0.100.1 release will be able to have this fisrt KS4.

samuelgarcia (approved):

PR #2541, opened 2024-03-05
Handle negative gain for neo rawio to make smarter user eperience.

samuelgarcia:
This handle negative gain for neo rawio to make smarter user eperience.

For instance in neuralynx very often the gain is negative for all channels which visualization and peak detection comlicated for user.
This simple patch make evrything easier at the cost of of reverting the traces early.

@h-mayorquin @CodyCBakerPhD : could check how  this would impact neuroconv ?

@yger @alejoe91 : if you are ok with this I need it very very soon (for the scool in Paris)


CodyCBakerPhD:
> @h-mayorquin @CodyCBakerPhD : could check how this would impact neuroconv ?

We don't have special consideration for it

What does it mean to have negative gain?

To us, as long as the value from `data_after_scaling = data_stored_as_int * gain_to_uv * uv_to_volts + offset_in_volts` truly results in `data_after_scaling` being in `Volts`, that's all that matters

alejoe91 (approved):

samuelgarcia:
OK
I merge this now.
Thanks.

PR #2545, opened 2024-03-06
Fixes bug in SortingAnalyzer

DradeAW:
Fixes a bug where `self.folder / "something"` crashes if the given folder was a str to begin with.
Need to cast it to a Path.


samuelgarcia:
Merci.

PR #2549, opened 2024-03-06
Allow `export_to_phy` to work with `fast_templates`

DradeAW:


src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -520,6 +520,34 @@ def _select_extension_data(self, unit_ids):
...
+        Returns
+        -------
+        templates: np.array
+            The returned templates (num_units, num_samples, num_channels)
+        """
+
+        templates = self.data["average"]
+
+        if save:
+            self.save()
```
samuelgarcia, line 543: This is strange beacause "average" is already in templates.
It was the case in TemplateExtension because operator where computed on the fly.

samuelgarcia:
Salut Aurelien.
Ok for me.
I made a few comments.

src/spikeinterface/core/analyzer_extension_core.py
DradeAW: Indeed, this should be removed!

DradeAW:
Done!

I also implemented a function `get_unit_template` because I find it very useful :)

src/spikeinterface/exporters/to_phy.py
```diff
@@ -181,9 +181,17 @@ def export_to_phy(
...
+    if templates_ext is None:
+        templates_ext = sorting_analyzer.get_extension("fast_templates")
+        if templates_ext is not None and template_mode != "average":
+            raise ValueError("export_to_phy with SortingAnalyzer with extension 'fast_templates' can only work with template_mode='average'")
+        dense_templates = templates_ext.get_templates(unit_ids=unit_ids)
+    else:
+        dense_templates = templates_ext.get_templates(unit_ids=unit_ids, operator=template_mode)
+    assert (
+        templates_ext is not None
+    ), "export_to_phy need SortingAnalyzer with extension 'templates' or 'fast_templates'"
```
zm711, line 195:
```suggestion
    ), "export_to_phy requires SortingAnalyzer with either extension 'templates' or 'fast_templates'"
```

could also be needs, but I like the sound of requires better.

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -520,6 +520,49 @@ def _select_extension_data(self, unit_ids):
 
         return new_data
 
+    def get_templates(self, unit_ids=None):
+        """
+        Return average templates for multiple units.
+
+        Parameters
+        ----------
+        unit_ids: list or None
```
zm711, line 529:
```suggestion
        unit_ids: list or None, default: None
```

samuelgarcia:
This is OK for me.
@alejoe91  : do you want to have a look ?

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -520,6 +520,49 @@ def _select_extension_data(self, unit_ids):
...
+
+        return np.array(templates)
+
+    def get_unit_template(self, unit_id):
+        """
+        Return average template for a single unit.
+
+        Parameters
+        ----------
+        unit_id:
```
zm711, line 552: is this an int or a string?

DradeAW: Well it depends,

There is currently no defined type for `unit_id` in SpikeInterface

zm711: Aren't they strings | int. Right?

DradeAW: I believe they are, but who says that won't change in the future?

zm711: You. Right now. Lock it in and take control.

src/spikeinterface/exporters/to_phy.py
```diff
@@ -181,9 +181,19 @@ def export_to_phy(
...
+        if templates_ext is not None and template_mode != "average":
+            raise ValueError(
+                "export_to_phy with SortingAnalyzer with extension 'fast_templates' can only work with template_mode='average'"
+            )
+        dense_templates = templates_ext.get_templates(unit_ids=unit_ids)
+    else:
+        dense_templates = templates_ext.get_templates(unit_ids=unit_ids, operator=template_mode)
+    assert (
+        templates_ext is not None
+    ), "export_to_phy requires SortingAnalyzer with either extension 'templates' or 'fast_templates'"
```
zm711, line 195: I was thinking about this and templates_ext cannot be None, because if it is None above the templates_ext.get_emplates will fail? Or am I misreading this? It seems to me that it would be better to do a check for none after line 185. Something like
```python
if templates_ext is None:
    raise ValueError('Must have either calculated templates or fast_templates')
```

That way regardless of if the `|` notation is accepted you have the check in place.

zm711 (commented):
Quick question @DradeAW for the assert check in this PR :)

src/spikeinterface/exporters/to_phy.py
DradeAW: True, nice catch!

But as you said, I'm waiting for the `|` PR to be merged, to re-write this more cleanly :)

DradeAW:
Ok I updated the implementation to be cleaner, and follow what we agreed to follow in #2570 

This is ready on my end :)

DradeAW:
@alejoe91 The test failed for some weird reason (not my fault), and I don't have the permissions to re-launch the test :/

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -522,6 +522,55 @@ def _select_extension_data(self, unit_ids):
...
+        ), f"Analyzer extension `fast_templates` only works with 'average' templates. Given operator = {operator}"
+        templates = self.data["average"]
+
+        if unit_ids is not None:
+            unit_indices = self.sorting_analyzer.sorting.ids_to_indices(unit_ids)
+            templates = templates[unit_indices, :, :]
+
+        return np.array(templates)
+
+    def get_unit_template(self, unit_id):
```
alejoe91, line 554: @DradeAW the `ComputeTemplates` doesn't have this method. Can you either add it there too or remove it?

src/spikeinterface/exporters/to_phy.py
```diff
@@ -180,8 +180,10 @@ def export_to_phy(
 
     # export templates/templates_ind/similar_templates
     # shape (num_units, num_samples, max_num_channels)
-    templates_ext = sorting_analyzer.get_extension("templates")
-    templates_ext is not None, "export_to_phy need SortingAnalyzer with extension 'templates'"
+    templates_ext = sorting_analyzer.get_extension("templates") or sorting_analyzer.get_extension("fast_templates")
```
alejoe91, line 183: Nice

src/spikeinterface/core/analyzer_extension_core.py
DradeAW: Done!

alejoe91 (approved):

PR #2551, opened 2024-03-06
Allow `auto_merge` to work with `fast_templates`

DradeAW:
Depends on #2549 (which needs to be merged before)


DradeAW:
Now cleaner, in the way we agreed in #2570 

Ready to merge on my end (but only after #2549)

samuelgarcia (approved):

PR #2552, opened 2024-03-06
Cleanup of modules gallery docs (post SorterAnalyzer)

zm711:
@samuelgarcia since you've added zarr into core for SorterAnalyzer we need it to be in base dependencies.... The documentation update broke one of the example code because when you did a zarr with SorterAnalyzer the doc environment didn't have xarray. We could also revert that change and change the example, but at this point it is probably best just to add xarray to core no? 

(Also I see for the examples you're using the list input for the SorterAnalyzer.compute()... Nice)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,6 @@ dependencies = [
     "threadpoolctl>=3.0.0",
     "tqdm",
     "zarr>=0.2.16",
-    "xarray",
     "neo>=0.13.0",
     "probeinterface>=0.2.21",
 ]
@@ -90,6 +89,7 @@ streaming_extractors = [
 full = [
     "h5py",
     "pandas",
+    "xarray",
     "scipy",
     "scikit-learn",
     "networkx",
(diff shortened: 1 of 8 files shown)


samuelgarcia:
xarray is a big stuff for core dep. I think I would prefer avoid it.
It is needed only when using Dataframe and so only quality mertics at the moment. Lets think about this.

pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
     "threadpoolctl>=3.0.0",
     "tqdm",
     "zarr>=0.2.16",
+    "xarray",
```
samuelgarcia, line 27: Honestly I would prefer to avoid this.
We can add it for doc building.

zm711:
But you have it in SorterAnalyzer for zarr format. So it's already in core if someone wants to use zarr format. But for this PR I can switch it back to just be in the docs building.

samuelgarcia:
Thanks for the grammar and typos. I can't beleive I did so much mistake.

pyproject.toml
zm711: Done. Removed from core. added back to full and to docs.

zm711:
It's the curse of someone who types so fast, the examples can't keep up with your mind :) .

samuelgarcia (approved):

Issue #2560, opened 2024-03-09
`SortingAnalyzer.Extension` Feedback


zm711:
So after using the extensions some more I have a couple ideas that I wanted to jot down just to think about.

## 1
One the error messaging isn't super helpful right now for two reasons. One I think we are all so use to the design of the package that when I try to run `compute('quality_metrics')` and it gives me an error I just import quality metrics because I know that's where it lives (ie I know the codebase). But I can imagine a new user doing `compute('spike_locations')` getting an error and not know which submodule to import. This would be super frustrating. So I'm wondering if we could add an attribute to extensions or maybe just a boolean when checking the extensions such that (in pseudocode)
```
if `quality_metrics` in ext_name:
    error_msg += 'import spikeinterface.qualitymetrics'
else:
    error_msg += 'import spikeinterface.postprocessing'
```

## 2
Problem 2 related to this issue is. If I type `.compute('quality_metrics')` without import qualitymetrics or I try `.compute('fake_extension')` I get the same error which is super confusing. I don't know what the solution is here since you're trying to allow people to make their own extensions. But it makes it hard to know what is actually possible or not with the package. (What *a priori* knowledge should people have of what is possible... I guess the docs could help with this a little bit). I'm really not sure what the solution is, but I discovered this when I mistyped `compute('quallity_metrics')` and it failed to register even though my typo clearly doesn't exist I was super confused until I read carefully.

## Finally,
when reloading I was trying to be super light so I only imported core. But I had extensions: [spike_locations, spike_amplitudes, quality_metrics]. Now these didn't load (because I hadn't imported the correct submodules yet). At first I was confused because I knew I had just calculated these values and (had accidentally restarted my computer) so I knew they were there. I checked my extensions folders and they were there. Then it dawned on my that I need to import the modules to be allowed to load my old data. Once I'm use to the new workflow that'll be no big deal, but I'm worried that new users will have issues with "my data is there but I can't get it to load." So is there some way we could scan the extensions folder and warn them that not all data is loaded because not all submodules were imported? (In this case if the extension exists and was saved the user may want to reload it and not realize they need the submodule in order to be allowed to reload it). And If I then import the new submodules can I refresh my `sorting_analyzer` so that it loads the missing extensions? (That's an honest question--I could look in the source code myself, but since I'm already writing a novel I might as well finish). 

[...]

Issue #2327, opened 2023-12-13
Desired structure for examples in the code base


zm711:
Inspired by #2316,

There are two main places to put actual examples in the documentation: 

The `modules_gallery` or the `how_to`. It is currently unclear what the separate goals for these two sections are/ought to be. Either way both are stored in [examples](https://github.com/SpikeInterface/spikeinterface/tree/main/examples), but `modules_gallery` are run by `sphinx-gallery` whereas `how_to` are run locally (by Sam or Alessio) and added to the documentation later.

## Topics of Discussion
What are the purposes for these two separate example streams?
Are they currently appropriately discoverable (ie do they need better names)?
Do they need to be improved (ie even we are keeping the `modules_gallery` what improvements need to be made to make it as useful as possible)?
What rules (if any) do developers need to know to provide examples for the documentation code base?

## Outputs
If there is a clear consensus/set of rules we can add this to the developer docs so we can keep track of this and try to ensure our examples documentation follows a rough set of guidelines (even if they are not hard rules)

samuelgarcia:
Thanks for this clear summary.

Clearly the organisation is not good and the main purpose was technical because the sphinx-gallery is making a separate folder.

JoeZiminski:
Thanks! Overall I think the docs are very comprehensive and contain a lot of important information. Mostly I would suggest structural changes so that users can quickly find relevant information related to their use case. As a relatively naive user of the documentation I would suggest the following changes, mostly based on the [https://diataxis.fr/] approach

1) Keep the 'How to' section and a 'Tutorials' section and fold some existing pages into these. I think for the most part the module gallery examples can be renamed and form additional 'How To's. Conceptually, tutorials and 'how to's' are distinct. 'How To' are short, recipes to tell you how to do specific things (e.g. the current 'append and or / concatenate segments' becomes 'How to analyse multi-segment recordings'). Tutorials are longer, covering a complete workflow from start to finish (I believe SI already has some of these in a repo). For example there may be 2-3 main tutorials that go into a lot of detail. Then the 'How to' sections can be very targetted but not go into a lot of detail / background.

[...]

PR #2563, opened 2024-03-11
`SortingAnalyzer` core documentation

alejoe91:
Some comments and suggestions:

- we need a way to expose and retrieve default parameters for the extensions, e.g.:
```
analyzer.get_default_extension_params("random_spikes")
```
- similarly, and related to #2560, we need a way to be able to list available extensions and automatically load the modules needed


doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
 SortingAnalyzer
 ---------------
 
-The :py:class:`~spikeinterface.core.SortingAnalyzer` is the class which connects a :code:`Recording` and a :code:`Sorting`.
+The :py:class:`~spikeinterface.core.SortingAnalyzer` class is the core object to combine a
+:py:class:`~spikeinterface.core.BaseRecording` and a :py:class:`~spikeinterface.core.BaseSorting` object.
+This is the first step for additional analyses, and the basis of several postprocessing and quality metrics
+computations.
+
+The :py:class:`~spikeinterface.core.SortingAnalyzer` is provides a convenient API to access the underlying
```
zm711, line 174:
```suggestion
The :py:class:`~spikeinterface.core.SortingAnalyzer` provides a convenient API to access the underlying
```

doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
...
+:py:class:`~spikeinterface.core.BaseSorting` and :py:class:`~spikeinterface.core.BaseRecording` information,
+and it supports several extensions (derived from the :py:class:`~spikeinterface.core.AnalyzerExtension` class)
+to perform further analysis.
+
+Importantly, the :py:class:`~spikeinterface.core.SortingAnalyzer` handles the *sparsity* (see [Sparsity]_ section), i.e.,
+the channels on which waveforms and templates are defined on, for example, based on a physical distance from the
+channel with the largest peak amplitude.
+
+Most of the extensions live in the :code:`postprocessing` module, but there are some *core* extensions to:
```
zm711, line 183:
```suggestion
Most of the extensions live in the :code:`postprocessing` module, but there are some *core* extensions too:
```

doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
...
+
+* select random spikes for downstream analysis (e.g., extracting waveforms or fit a PCA model)
```
zm711, line 185:
```suggestion
* select random spikes for downstream analysis (e.g., extracting waveforms or fitting a PCA model)
```

doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
...
+* estimate templates
+* extract waveforms for single spikes
+* compute channel-wise noise levels
+
+.. note::
+
+    Some extensions depend on others, which must be pre-computed. For example:code:`waveforms`, the :code:`waveforms` extension depends on
```
zm711, line 192:
```suggestion
    Some extensions depend on others, which must be pre-computed. For example :code:`waveforms`, the :code:`waveforms` extension depends on
```

doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
...
+        sparse=True, # default
+        format="memory", # default
+    )
+
+    print(sorting_analyzer)
+
+.. code-block:: bash
+
+    >>> SortingAnalyzer: 4 channels - 10 units - 1 segments - memory - sparse - has recording
+    >>> Loaded 0 extenstions:
```
zm711, line 212: We need to fix this typo in the source code. the repr/str should say extensions and not extenstions....

doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
...
+
+The :code:`sorting_analyzer` object implements convenient functions to access the underlying :code:`recording` and
+:code:`sorting` objects information:
```
zm711, line 215:
```suggestion
:code:`sorting` objects' information:
```

doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
...
+    >>> dict_keys(['average'])
+
+# TODO: extension __repr__
+
+.. code-block:: python
+    # arguments can be passed directly to the compute function
+    # note that re-computing an extension will overwrite the existing one
+    waveform_extension_2 = sorting_analyzer.compute("waveforms", ms_before=2, ms_after=5)
+
+    # multiple extensions can be computedwithin the same `compute` call
```
zm711, line 257:
```suggestion
    # multiple extensions can be computed within the same `compute` call
```

doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
...
+    sorting_analyzer.compute(
+        ["random_spikes", "waveforms", "templates", "noise_levels"]
+    )
+
+The :py:class:`~spikeinterface.core.SortingAnalyzer` by default is defined *in memory*, but it can be saved at any time
+(or upon instantiation) to one of the following backends:
+
+* | :code:`zarr`: the sorting analyzer is saved to a [Zarr]() folder, and each extension is a Zarr group. This is the
+  | recommended backend, since Zarr files can be writter to/read from the cloud and compression is applied.
```
zm711, line 266:
```suggestion
  | recommended backend, since Zarr files can be written to/read from the cloud and compression is applied.
```

doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
...
+* | :code:`binary_folder`: the sorting analyzer is saved to a folder, and each extension creates a sub-folder. The extension
+  | data are saved to either :code:`npy` (for arrays), `csv` (for dataframes), or `pickle` (for everything else).
```
zm711, line 268:
```suggestion
  | data are saved to either :code:`npy` (for arrays), :code:`csv` (for dataframes), or :code:`pickle` (for everything else).
```

doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
...
+    )
+    sorting_analyzer_binary = sorting_analyzer.save_as(
+        folder=processed_folder / "sorting_analyzer_bin",
+        format="binary_folder"
+    )
+    # sorting_analyzer_zarr and sorting_analyzer_binary are valid SortingAnalyzers,
+    # now associated to a Zarr storage / binary folder backend
+
+    # create sorting analyzer associated to a backend upon instantiation
+    # create in-memory sorting analyzer object
```
zm711, line 289: Am I misreading this. Why is it backend and in-memory for this comment?

doc/modules/core.rst
zm711: L192 in sortinganalyzer.py to be fixed. It won't let me put a comment there to mark it out.

zm711 (commented):
First pass of comments :)

doc/modules/core.rst
alejoe91: Fixed in last commit

doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
...
+The :py:class:`~spikeinterface.core.SortingAnalyzer` provides a convenient API to access the underlying
+:py:class:`~spikeinterface.core.BaseSorting` and :py:class:`~spikeinterface.core.BaseRecording` information,
+and it supports several extensions (derived from the :py:class:`~spikeinterface.core.AnalyzerExtension` class)
+to perform further analysis.
+
+Importantly, the :py:class:`~spikeinterface.core.SortingAnalyzer` handles the *sparsity* (see [Sparsity]_ section), i.e.,
+the channels on which waveforms and templates are defined on, for example, based on a physical distance from the
+channel with the largest peak amplitude.
+
+Most of the extensions live in the :code:`postprocessing` module, but there are some *core* extensions too:
```
zm711, line 183:
```suggestion
Most of the extensions live in the :code:`postprocessing` module (with the :code:`quality_metrics` extension in the :code:`qualitymetrics` module) , but there are some *core* extensions too:
```

I think we need to be upfront here about the quality metrics since the design choice was made to separate qc from postprocessing we probably want to highlight multiple times that to get all functionality we need to import both modules. But I'm fine with my wording being changed if we want to rewrite this sentence.

doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
...
+* compute channel-wise noise levels
+
+.. note::
+
+    Some extensions depend on others, which must be pre-computed. For example :code:`waveforms`, the :code:`waveforms` extension depends on
+    the :code:`random_spikes` one. If the latter is not computed, computing :code:`waveforms` will throw an error.
+
+
+.. code-block:: python
+    from spikeinterface import create_sorting_analyzer
```
zm711, line 197:
```suggestion

    from spikeinterface import create_sorting_analyzer
```

I think for sphynx you need one return right between the block and the code.

doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
...
+    >>> Loaded 0 extensions:
+
+The :code:`sorting_analyzer` object implements convenient functions to access the underlying :code:`recording` and
+:code:`sorting` objects' information:
+
+.. code-block:: python
+
+    num_channels = sorting_analyzer.get_num_channels()
+    num_units = sorting_analyzer.get_num_units()
+    sampling_frequency = sorting_analyzer.sampling_frequency
```
zm711, line 221:
```suggestion
    sampling_frequency = sorting_analyzer.sampling_frequency()
```

We could do this as the function instead of the attribute right, just to make everything be a function for the docs?

doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
...
+    >>> dict_keys(['average'])
+
+# TODO: extension __repr__
+
+.. code-block:: python
+    # arguments can be passed directly to the compute function
+    # note that re-computing an extension will overwrite the existing one
+    waveform_extension_2 = sorting_analyzer.compute("waveforms", ms_before=2, ms_after=5)
+
+    # multiple extensions can be computed within the same `compute` call
```
zm711, line 257: Do we want a note that there isn't an optimizer and so if doing this one must put them in the correct order (ie random_spikes must go first)?

doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
...
+
+.. code-block:: python
+
+    sorting_analyzer = si.load_sorting_analyzer(folder="my-sorting-analyzer.zarr")
+
+
+.. note::
+
+    When saved to disk, the :code:`SortingAnalyzer` will store a copy of the :code:`Sorting`` object,
+    because that is relatively small and needed for most (if not all!) operations. The same is not
```
zm711, line 309:
```suggestion
    because it is relatively small and needed for most (if not all!) operations. The same is not
```

doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
...
+    true for the :code:`Recording` object, for which only the main properties will be stored (e.g,
+    :code:`sampling_frequency`, :code:`channel_ids`, :code:`channel_locations`, etc.) and
+    a provenance to reload the :code:`Recording`. When loading a :code:`SortingAnalyzer` from disk,
+    an attempt is made to re-instantiate the :code:`Recording` object from the provenance. In case
```
zm711, line 313:
```suggestion
    an attempt is made to re-instantiate the :code:`Recording` object from the provenance. In cases
```

Either in cases of failure...
or in the case of failure.

zm711 (commented):
While we wait for Sam to review I just did one more set of comments.

doc/modules/core.rst
samuelgarcia: we also have get_sampling_frequncy() but I think it is OK to show attributes also no ?

doc/modules/core.rst
samuelgarcia: I think at this level I would avoid to explain this complexity no ?
Maybe we can open a section on depencies after no ?

doc/modules/core.rst
samuelgarcia: Yes I think it is a mistake

doc/modules/core.rst
```diff
@@ -166,10 +166,153 @@ Internally, any sorting object can construct 2 internal caches:
...
+    )
+    sorting_analyzer_binary = sorting_analyzer.save_as(
+        folder=processed_folder / "sorting_analyzer_bin",
+        format="binary_folder"
+    )
+    # sorting_analyzer_zarr and sorting_analyzer_binary are valid SortingAnalyzers,
+    # now associated to a Zarr storage / binary folder backend
+
+    # create sorting analyzer associated to a backend upon instantiation
+    # create in-memory sorting analyzer object
```
samuelgarcia, line 289:
```suggestion
    # We can also create directly a  SortingAnalyzer associated to a backend upon instantiation
    # for instance, this create "zarr" SortingAnalyzer object
```

samuelgarcia (approved):

samuelgarcia:
This is perfect for me.
You can merge when you want.
Thank you Zach also pushing doc towards perfection.

doc/modules/core.rst
zm711: It's just a pedagogical difference. For a beginner they have to learn both the library and the difference between a function and an attribute. So my style is more beginner-friendly in that we remove them having to learn a bunch of python features by saying look everything has a function to grab a piece of data. Right now if I read this I would ask myself my sampling_frequency is special and doesn't have a function whereas all the others have functions and that might distract me from the rest of the docs.

Doing a mix is better for the intermediate learning because they already know enough basics that they focus on the breadth of techniques. So they see that they can both grab the attribute or use a function and can chose their style. 

So for me it would be do we want this documentation to be more beginner or more intermediate friendly. My inclination is that people that read the docs need a bit more beginner friendly. So we give them one good solution and then once they improve their own python skills the'll realize pretty quickly that they can use the attribute if they want.

But this is a super small point.

doc/modules/core.rst
zm711: I think that is a good idea. Separate section so that we don't get too complex right! I do think we need to mention it because it will be frustration and not clear why the function fails if they do the wrong order. So it is important but definitely separate section.

doc/modules/core.rst
```diff
@@ -166,10 +166,154 @@ Internally, any sorting object can construct 2 internal caches:
...
+The :py:class:`~spikeinterface.core.SortingAnalyzer` by default is defined *in memory*, but it can be saved at any time
+(or upon instantiation) to one of the following backends:
+
+* | :code:`zarr`: the sorting analyzer is saved to a [Zarr]() folder, and each extension is a Zarr group. This is the
+  | recommended backend, since Zarr files can be written to/read from the cloud and compression is applied.
+* | :code:`binary_folder`: the sorting analyzer is saved to a folder, and each extension creates a sub-folder. The extension
+  | data are saved to either :code:`npy` (for arrays), :code:`csv` (for dataframes), or :code:`pickle` (for everything else).
+
+
+The :code:`SortingAnalyzer.save_as` function will save the object **and all its extension** to disk.
```
h-mayorquin, line 272: Missing and s for the plural on the bold format.

doc/modules/core.rst
h-mayorquin: I am not convinced that functions are more beginner friendly than attributes.   Consistency I think is a good argument though.

zm711: My argument was completely for consistency. I don't care if one does functions or attributes just be consistent and since most of them were already functions I chose the route of making all functions, but absolutely would be fine to just grab them all as attributes too (from my perspective).

zm711:
We don't have compatibility docs in here for the converting the old waveform extractor to a mock waveform extractor here. Do we want to add a small section now or in a separate PR. I just don't want to forget this since we are going to have a lot of users who need to convert.

alejoe91:
@chrishalcrow is working on a dedicated page that explains how to convert a pipeline from the waveform extractor to the sorting analyzer. I think that is the good place to show how to deal with waveform extractor folders?
We could point to that page here!

doc/modules/core.rst
alejoe91: I normally use the attribute, which I think is easier. Maybe we can show both?

doc/modules/core.rst
alejoe91, line 272:
```suggestion
The :code:`SortingAnalyzer.save_as` function will save the object **and all its extensions** to disk.
```

doc/modules/core.rst
```diff
@@ -166,10 +166,154 @@ Internally, any sorting object can construct 2 internal caches:
...
+    >>> Loaded 0 extensions:
+
+The :code:`sorting_analyzer` object implements convenient functions to access the underlying :code:`recording` and
+:code:`sorting` objects' information:
+
+.. code-block:: python
+
+    num_channels = sorting_analyzer.get_num_channels()
+    num_units = sorting_analyzer.get_num_units()
+    sampling_frequency = sorting_analyzer.sampling_frequency
```
alejoe91, line 222:
```suggestion
    sampling_frequency = sorting_analyzer.get_sampling_frequency()
    # or: sampling_frequency = sorting_analyzer.sampling_frequency
```

alejoe91:
Thanks guys, let's merge :)

doc/modules/core.rst
zm711: Our argument (Heberto and mine since I think we fall in the same camp for documentation in general) is that if we include both we are making new learners/users learn both that Python has attributes + functions and the spikeinterface schema (why is it sampling_frequency and not samping_rate or why frequency and not period) so we are teaching them two things at once rather than teach them one thing (ie spikeinterface stuff). In our schema for docs you teach them one thing super well and then they can learn the other cool Python tricks themselves (function vs attribute). I think you and Sam like showing how many features the project has which is great for more experienced users/developers but I have a feeling that they don't read the docs (for example Aurélien once said he has never looked at the docs). 

My example would be as you learn English you could learn "need", "must", "require", which all have slightly different flavors so I could teach you that they all accomplish the same thing + teach you their different connotations or I could just teach you the word "need" and make sure you really understand that and even if there is a slightly better word "need" would still get the job done.

alejoe91: Then in this case I would rather use `sorting_analyzer.sampling_frequency`, since this is the preferred use.

zm711: Again it's a super small point and not a big deal for this PR :) Just something to keep in mind. We had a big discussion of some of this stuff here https://github.com/SpikeInterface/spikeinterface/issues/2327

PR #2564, opened 2024-03-11
Add note about recommendation not to change batch size in Kilosort1-3

### LINKED ISSUE #2562: KS2.5 / KS3 'spike holes' patched release
Hi Everyone,

A patch was recently released in [KS2.5](https://github.com/MouseLand/Kilosort/compare/kilosort25) and [KS3](https://github.com/MouseLand/Kilosort/commit/d2b5c27465dae22b861e2a5183b6830e61ec4297) to fix the 'spike holes' bug. There were pushed as commits to the relevant branches in the KS repository. 

For the next version of SI I makes sense to re-build the [docker images](https://github.com/SpikeInterface/spikeinterface-dockerfiles/tree/main/kilosort-compiled/kilosort2_5-compiled) from the KS main. However, if I understand correctly the versioning might be a bit tricky. Once the docker files are rebuilt, will all spikeinterface versions will begin pulling the most recent build? As I presume in some circumstances this change might lead to quite different KS outputs, so would be nice to pin to an SI version.

Also just to confirm, when building the images it is necessary to 1) download the Kilosort repository, then switch to the relevant KS version branch before building?

Cheers!

zm711:
Based on the KS readme [here](https://github.com/MouseLand/Kilosort/tree/main). It is not advised for users to change the batch size `NT`. This just adds a note saying that the SpikeInterface default is the KS recommended value as long as the user doesn't change the ntbuff either. Related to #2562.

### DIFF
--- a/src/spikeinterface/sorters/external/kilosort.py
+++ b/src/spikeinterface/sorters/external/kilosort.py
@@ -56,7 +56,7 @@ class KilosortSorter(KilosortBase, BaseSorter):
         "freq_max": "Low-pass filter cutoff frequency",
         "ntbuff": "Samples of symmetrical buffer for whitening and spike detection",
         "Nfilt": "Number of clusters to use (if None it is automatically computed)",
-        "NT": "Batch size (if None it is automatically computed)",
+        "NT": "Batch size (if None it is automatically computed--recommended Kilosort behavior if ntbuff also not changed)",
         "wave_length": "size of the waveform extracted around each detected peak, (Default 61, maximum 81)",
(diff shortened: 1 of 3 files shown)


zm711:
I think in the past for some issues people have suggested (here and on the KS repo) to change the batch size to help with something (maybe the eigenvector issue) so I think the warning serves the purpose of also saying that hacking solutions for that may introduce other bugs. 

That being said warning or not is actually fine by me. :)

alejoe91 (approved):

alejoe91 (approved):

PR #2571, opened 2024-03-13
Proposal for auto import extensions module.

samuelgarcia:
Hi @zm711 @alejoe91.
This is proposal to auto load extensions : when extension is builin and not registered because not impotred yet : then we have an auto load mechanism. This should help the end user.
Tell me what you think.

This is hard to be tested in core tests because it auload submodules which will fail on core testing because of dependencies.
So I made local tests and it works.

This should fix partially #2560.

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -1077,13 +1077,13 @@ def get_extension_class(extension_name: str, autoload=True):
 
     if extension_name not in extensions_dict:
         if extension_name in _builtin_extensions:
+            module = _builtin_extensions[extension_name]
             if autoload:
-                module = _builtin_extensions[extension_name]
                 print(f"module '{module}' is imported automatically for extension '{extension_name}'")
                 imported_module = importlib.import_module(module)
                 extensions_dict = {ext.extension_name: ext for ext in _possible_extensions}
             else:
-                raise ValueError(f"Extension '{extension_name}' is not registered, please import related module before use")
+                raise ValueError(f"Extension '{extension_name}' is not registered, please import related module before use: 'import {module}'")
         else:
             raise ValueError(f"Extension '{extension_name}' is unkown maybe this is an external extension or a typo.")
         
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -909,36 +909,30 @@ class SortingAnalyzer:
 
     def get_saved_extension_names(self):
         """
-        Get extension saved in folder or zarr that can be loaded.
+        Get extension names saved in folder or zarr that can be loaded.
+        This do not load data, this only explores the directory.
         """
-        assert self.format != "memory"
-        global _possible_extensions
+        saved_extension_names = []
+        if self.format == "binary_folder":
+            ext_folder = self.folder / "extensions"
+            if ext_folder.is_dir():
+                for extension_folder in ext_folder.iterdir():
+                    is_saved = extension_folder.is_dir() and (extension_folder / "params.json").is_file()
+                    if not is_saved:
+                        continue
+                    saved_extension_names.append(extension_folder.stem)
 
-        if self.format == "zarr":
+        elif self.format == "zarr":
             zarr_root = self._get_zarr_root(mode="r")
             if "extensions" in zarr_root.keys():
                 extension_group = zarr_root["extensions"]
-            else:
-                extension_group = None
+                for extension_name in extension_group.keys():
+                    if "params" in extension_group[extension_name].attrs.keys():
+                        saved_extension_names.append(extension_folder.stem)
 
-        saved_extension_names = []
-        for extension_class in _possible_extensions:
-            extension_name = extension_class.extension_name
-
-            if self.format == "binary_folder":
-                extension_folder = self.folder / "extensions" / extension_name
-                is_saved = extension_folder.is_dir() and (extension_folder / "params.json").is_file()
-            elif self.format == "zarr":
-                if extension_group is not None:
-                    is_saved = (
-                        extension_name in extension_group.keys()
-                        and "params" in extension_group[extension_name].attrs.keys()
-                    )
-                else:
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1073,9 +1074,21 @@ def get_extension_class(extension_name: str):
...
     extensions_dict = {ext.extension_name: ext for ext in _possible_extensions}
-    assert (
-        extension_name in extensions_dict
-    ), f"Extension '{extension_name}' is not registered, please import related module before use"
+
+    if extension_name not in extensions_dict:
+        if extension_name in _builtin_extensions:
+            module = _builtin_extensions[extension_name]
+            if autoload:
+                print(f"module '{module}' is imported automatically for extension '{extension_name}'")
```
alejoe91, line 1082:
```suggestion
```
should we remove the print?

samuelgarcia: I think it is good to have a message for auto loading no ?
Maybe a warning

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1471,3 +1484,27 @@ def get_pipeline_nodes(self):
     def get_data(self, *args, **kwargs):
         assert len(self.data) > 0, f"You must run the extension {self.extension_name} before retrieving data"
         return self._get_data(*args, **kwargs)
+
+
+# this is a hardcoded list to to improve error message and autoload mechanism
+# this is important because extension are register when the submodule is imported
```
alejoe91, line 1490:
```suggestion
# this is important because extension are registered when the submodule is imported
```

src/spikeinterface/core/sortinganalyzer.py
alejoe91: not sure

zm711: I'm torn here... If someone requests this functionality they should expect us to import what is needed to do the operation right? I guess the concern would be some sort of namespace clash? Like if we run an extractor the user expects a read operation from a file, we don't explicitly say we are creating a memmap which may lead to RAM exhaustion. On the other hand people complain about things failing silently for the job_kwargs.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1073,9 +1074,21 @@ def get_extension_class(extension_name: str):
...
+            if autoload:
+                print(f"module '{module}' is imported automatically for extension '{extension_name}'")
+                imported_module = importlib.import_module(module)
+                extensions_dict = {ext.extension_name: ext for ext in _possible_extensions}
+            else:
+                raise ValueError(
+                    f"Extension '{extension_name}' is not registered, please import related module before use: 'import {module}'"
+                )
+        else:
+            raise ValueError(f"Extension '{extension_name}' is unkown maybe this is an external extension or a typo.")
```
zm711, line 1090:
```suggestion
            raise ValueError(f"Extension '{extension_name}' is unknown maybe this is an external extension or a typo.")
```

This I really like since I have a tendency to type too fast in my terminal. Haha.

zm711 (commented):
As long as you're okay with autoload then this seems good to me. I can also test locally this afternoon. Otherwise, I think just the additional error messaging is much improved.

src/spikeinterface/core/sortinganalyzer.py
alejoe91: The extractors example is a very good point. If someone saves a JSON file with `dump_to_json`, all the modules from `extractors`, `preprocessing`, etc. will be loaded. 
In that case, we store the path to the class so imports are minimal.

Maybe we could do the same here? e.g.

```
_builtin_extensions = {
    "spike_amplitudes": "spikeinterface.postprocessing.spike_amplitudes.SpikeAmplitudeExtension",
    ...
}
```

h-mayorquin: I think it should be removed. What is the user supposed to do with this information?

This sounds more like logging and when it is output by default the convention is to print to standard error as far as I am aware. So a warning. But I personally think that warnings should be actionable so not even sure about that one.

src/spikeinterface/core/sortinganalyzer.py
h-mayorquin: We should get a spell check action. We are spending a lot of energy on these type of errors.

zm711:
So testing this isn't quite working locally for me. I have a binary_folder with following extensions.
image
upon loading I get:
image
and if I try to grab an extension I get
image


I believe there are a few issues:

`get_extension` makes a call to `has_extension` which I believe is failing.
And during loading we also make all call to `get_save_extension_names` which relies on `_possible_extensions` which I believe is not being populated.

What else do you want me to try locally Sam? Were you using zarr backend for this or did you get this to work with binary?

zm711:
Also I never get to the print statement we were discussing above (L1082).

src/spikeinterface/core/sortinganalyzer.py
zm711: Haha, I would love to hear the convo that convinces Sam to add another linter/formatter.

samuelgarcia:
Oups. My bad I tested the compute() let me finish this correctly.

src/spikeinterface/core/sortinganalyzer.py
h-mayorquin: I am not thinking on a linter or formater, there are CI actions that just point out the spell mistakes. We can just check them there instead of fishing for them and relying on our nose.

zm711:
Happy to test `compute()` tomorrow (that was actually the one thing I didn't test-haha). But yeah for me I think `load` is a nicer target since I always import everything when I'm doing computations, but when I'm reloading sometimes I just do core to check things so it might miss some of the extensions.

zm711:
Tested both `compute` and `load_sorting_analyzer` and both worked for me locally.

alejoe91 (approved):

Issue #2569, opened 2024-03-12
Cannot get KS4 to utilize my GPU


mohebial:
Hi,
    I am trying to get KS4 to work on spikeInterface to sort long NPX1.0 recording sessions. Whether I use the docker image or a local KS4 installation, my GPU is not engaged, and processing takes forever to complete. It shouldn't be my CUDA settings, since I can run standalone KS4 just fine using the GPU (RTX 3060). Do you have any suggestions on what might be wrong with my spikeInterface setup? I am running the following script in Jupyterlab on a Windows machine.

Thanks!
Ali
```

# Define recording object using the local file address
raw_rec = si.read_openephys(base_folder,stream_id='0')
rec1 = si.highpass_filter(raw_rec, freq_min=400.)
bad_channel_ids, channel_labels = si.detect_bad_channels(rec1)
rec2 = rec1.remove_channels(bad_channel_ids)
print('bad_channel_ids', bad_channel_ids)
rec3 = si.phase_shift(rec2)
rec4 = si.common_reference(rec3, operator="median", reference="global")
processed_recording = si.concatenate_recordings([rec4])
processed_recording

# Perform Sorting (this will take forever)
sorting = si.run_sorter(sorting_algorithm, processed_recording, output_folder=base_folder / sorting_algorithm / 'sorting_output',
                        docker_image=False, verbose=True, remove_existing_folder= True, delete_container_files=False, **sortingParams)
```

alejoe91:
Hi Ali, 

Thanks for the feedback!

Do you use the same environment? 
To make sure Ks4 uses the GPU you can try this from the notebook:
```
import torch

print(torch.cuda.is_available())
```

If the print is false, then you probably need to reinstall torch with vida support in the env, as explained in the Kilosort README.

I'll investigate what's wrong with the container. Haven't tested it on Windows, but it works on my side on Linux.

mohebial:
Thanks much for looking into this! I am running them both in the same environment (python=3.9, spikeinterface=1.0.1, KS=4.0.1). And the CUDA is available in the notebook. You will see below that it passes that test, and the docker image is running, but my GPU is not in use. The same happens for docker_image=False. But when I use straight KS4 to sort my binary data directly, it's very fast, and my GPU usage is 100%

Best,
Ali

<img width="1826" alt="image" src="https://github.com/SpikeInterface/spikeinterface/assets/4812219/c15e9d18-bab3-4ca2-909f-aa6ab7b51b99">

[...]

PR #2577, opened 2024-03-14
Use GPU in `Kilosort4Sorter`

bagibence:
I was trying to run Kilosort4 with docker but it didn't use my GPU.

After some digging I found that `Kilosort4Sorter` only inherits from `BaseSorter` which by default doesn't use the GPU.
I'm not sure if this is the best solution, but adding `gpu_compatibility = "nvidia-required"` seems to have fixed the issue for me.

Might help with #2569

### DIFF
--- a/src/spikeinterface/sorters/external/kilosort4.py
+++ b/src/spikeinterface/sorters/external/kilosort4.py
@@ -15,6 +15,7 @@ class Kilosort4Sorter(BaseSorter):
 
     sorter_name: str = "kilosort4"
     requires_locations = True
+    gpu_compatibility = "nvidia-required"
 
     _default_params = {
         "nblocks": 1,
--- a/src/spikeinterface/sorters/external/kilosort4.py
+++ b/src/spikeinterface/sorters/external/kilosort4.py
@@ -15,7 +15,7 @@ class Kilosort4Sorter(BaseSorter):
 
     sorter_name: str = "kilosort4"
     requires_locations = True
-    gpu_compatibility = "nvidia-optional"
+    gpu_capability = "nvidia-optional"
 
     _default_params = {
         "nblocks": 1,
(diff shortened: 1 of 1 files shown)


src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -15,6 +15,7 @@ class Kilosort4Sorter(BaseSorter):
 
     sorter_name: str = "kilosort4"
     requires_locations = True
+    gpu_compatibility = "nvidia-required"
```
samuelgarcia, line 18: In theory nvidia is not require because torch also with cpu which slower but good for for testing on laptop wihtout nvidia for instance. So maybe we should add "nvidia-recomended" case that propagate the nvidia docker option only if nvidia is present.

alejoe91:
> I was trying to run Kilosort4 with docker but it didn't use my GPU.
> 
> After some digging I found that `Kilosort4Sorter` only inherits from `BaseSorter` which by default doesn't use the GPU. I'm not sure if this is the best solution, but adding `gpu_compatibility = "nvidia-required"` seems to have fixed the issue for me.
> 
> Might help with #2569

Thanks so much for spotting this @bagibence 

I agree with Sam that we should use `nvidia-optional`. This will not throw an error in case the GPU is missing (e.g., for short tests using CPU only is ok), but it will make the GPU available to the container if present.

See here:
https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/sorters/runsorter.py#L433-L448

src/spikeinterface/sorters/external/kilosort4.py
alejoe91, line 18:
```suggestion
    gpu_compatibility = "nvidia-optional"
```

alejoe91:
@bagibence can you try on your side if the docker run engages the GPU?

bagibence:
Sorry, I was testing `nvidia-optional` and noticed that the PR was wrong. I'll send a fix in a minute.

bagibence:
Yes, now the correct version of `gpu_capability = "nvidia-optional"` does use the GPU with docker.

alejoe91:
> Yes, now the correct version of `gpu_capability = "nvidia-optional"` does use the GPU with docker.

Awesome!!! @samuelgarcia we should make a patch release for this

alejoe91 (approved):

PR #2580, opened 2024-03-15
Clarify status of main and Sortinganalyzer

alejoe91:
It could be very confusing for a user to understand the status of the `SortingAnalyzer` in the docs.

Added a comment in the README and docs, so that landing on `latest` will explain briefly what's going on and where to find the stable docs.


samuelgarcia (approved):

samuelgarcia:
thanks.
that was important and urgent

Issue #2193, opened 2023-11-12
`set_probe` behavior for `RecordingExtractors`


h-mayorquin:
We were discussing on slack about `set_probe` behavior. I am recaping the points here in case we want to make a decision about it. 

Context:
Me and @magland were mentioning that we have been confused about this method. Basically, it returns a new recording with the recording with the attached probe instead of setting the probe in the current recording. For the latter, you need to use `in_place`. 

Now, @alejoe91 mentioned some technical difficulties that are relevant. Basically, for most probe structures, the method can't modify state unless the following condition is met:

https://github.com/SpikeInterface/spikeinterface/blob/329197618a9b48ef876d1e8b8e79f07f4abf5e49/src/spikeinterface/core/baserecordingsnippets.py#L176-L185

So, a `true` setter method would not be as general. 

Then @samuelgarcia mentioned that, in his view, `set_probe` is following the logic of `set_index` in pandas with the `in_place` logic. He believes this is just a matter of documentation and not semantics.

I disagree with that argument for three reasons:
* Even if pandas uses set it is still a bad name. Set methods are the fundamental mutator method, I don't know it should work in a fluen interface way. https://en.wikipedia.org/wiki/Mutator_method
* Pandas has some problems on their own on that regard. See the `SettingWithCopyWarning` in stackoverflow and tutorials.
* `in_place` works in whacky ways in Pandas, mantainers have expressed their [desire to deprecate the keyword](https://www.youtube.com/watch?v=7JHqxODJG9k&t=45m35s) and I don't think we should build on that.

My own proposal would be to make the current `set_probe` method works like a true setter and fails when this is not possible. We could have another method then that works to build a probe independent of the structure so this always return a new recorder object. @magland proposed `recording.create_new_recording_with_probe(probe=probe)` which works very well to me. 

Any other suggestions or experiences?
It could be that this is very clear than most people and mine and Jeremy's troubles are statistical flukes.

alejoe91:
Hi @h-mayorquin 

Thanks for writing this up. Fine for me to have the `set_probe/probegroup` throw an error in case it can't set in place and define another function to return a new object with probe attached. I agree this would be way clearer for the end user!

magland:
Thanks @h-mayorquin 

I would revise and simplify my recommendation to a static method on BaseRecording

```
BaseRecording.create_with_probe(recording, probe=probe)
```

magland:
And one other idea...
[...]

PR #2582, opened 2024-03-15
Put SortingAnalyzer in get started

chrishalcrow:
Updated the get_started how to guide to contain SortingAnalyzer rather than WaveformExtractor.

Feedback welcome!

Also updated some bits and pieces for clarity, and added links to the Docker and ProbeInterface guides.

@samuelgarcia Is my understand of get_saved_extension_names and get_loaded_extension_names correct?


examples/how_to/get_started.py
```diff
@@ -59,9 +59,9 @@
 import spikeinterface.curation as scur
 import spikeinterface.widgets as sw
 
-#  We can also import all submodules at once with this
-#  this internally imports core+extractors+preprocessing+sorters+postprocessing+
-#  qualitymetrics+comparison+widgets+exporters
+# Alternatively, we can import all submodules at once which
+# internally imports core+extractors+preprocessing+sorters+postprocessing+
+# qualitymetrics+comparison+widgets+exporters
```
zm711, line 64:
```suggestion
# Alternatively, we can import all submodules at once with `import spikeinterface.full as si` which
# internally imports core+extractors+preprocessing+sorters+postprocessing+
# qualitymetrics+comparison+widgets+exporters. In this case all aliases in the following tutorial
# would be `si`
```

We don't actually mention how to do that here.... If we are going to mention the alternative we might want to show it. I think the `with this` in the older version was meant to be replaced with `import spikeinterface.full as si`. But they never got around to it.

examples/how_to/get_started.py
```diff
@@ -196,50 +199,91 @@
...
 sorting_KS2 = ss.run_sorter(sorter_name="kilosort2", recording=recording_preprocessed, docker_image=True, verbose=True)
 print(sorting_KS2)
 
-# SpikeInterface provides a efficient way to extract waveforms from paired recording/sorting objects.
-# The `extract_waveforms` function samples some spikes (by default `max_spikes_per_unit=500`)
-# for each unit, extracts their waveforms, and stores them to disk. These waveforms are helpful to compute the average waveform, or "template", for each unit and then to compute, for example, quality metrics.
+# For postprocessing SpikeInterface pairs recording and sorting objects into a `SortingAnalyzer` object.
+# The `SortingAnalyzer` can be loaded in memory or saved in a folder. Here, we save it in binary format.
+
+sa_TDC = si.create_sorting_analyzer(sorting_TDC, recording_preprocessed, format='binary_folder', folder='sa_TDC_binary')
```
zm711, line 212:
```suggestion
sa_TDC = si.create_sorting_analyzer(sorting=sorting_TDC, recording=recording_preprocessed, format='binary_folder', folder='sa_TDC_binary')
```
Everyone decided that for tutorials we would always use keyword arguments to make things as explicit as possible  moving forward :)

examples/how_to/get_started.py
```diff
@@ -196,50 +199,91 @@
...
+# This folder is where all the postprocessing data will be saved such as waveforms and templates. Let's calculate
+# some waveforms. When doing this, the function samples some spikes (by default `max_spikes_per_unit=500`)
+# for each unit, extracts their waveforms, and stores them to disk in `extensions/waveforms`.
+# These waveforms are helpful to compute the average waveform, or "template", for each unit and then to compute, for example, quality metrics.
+# Computations with the `SortingAnalyzer` object are done using the `compute` method:
 
 # +
-we_TDC = si.extract_waveforms(recording_preprocessed, sorting_TDC, "waveforms_folder", overwrite=True)
-print(we_TDC)
+sa_TDC.compute("random_spikes")
```
zm711, line 221: Maybe we need at least one example showing how you can tune parameters (max_spikes_per_unit=300) or something like that, but not sure since this is suppose to be our simplest tutorial.... Would be curious to see others opinions....

examples/how_to/get_started.py
```diff
@@ -251,18 +295,19 @@
 qm_params["drift"]["interval_s"] = 2
 qm_params["drift"]["min_spikes_per_interval"] = 2
 
-qm = sqm.compute_quality_metrics(we_TDC, qm_params=qm_params)
+qm = sqm.compute_quality_metrics(sa_TDC, qm_params=qm_params)
```
zm711, line 298: Do we want to make quality metrics seem like this different thing or would it be better to do something like

sa_TDC.compute('quality_metrics').... and add in the additional arguments?

zm711 (commented):
Line35 has  typo that I can't comment on "the end user". If you would fix that that would be great :). Otherwise I think we just need to decide if we want to change a couple things with this rewrite. But looks pretty good to me otherwise.

alejoe91:
Thanks @chrishalcrow 

Can you rename `sa` to `analyzer`? I think it helps with clarity :)

examples/how_to/get_started.py
chrishalcrow: Yes, much better for it to be analyzer.compute() notation! Done below

examples/how_to/get_started.py
chrishalcrow: Yes, I don't think it's confusing to add some parameters. I've got for computing a few things with default options, then a few things with some random parameters.

chrishalcrow:
Ok, I've incorporate your feedback @zm711 and @alejoe91 , thanks!

doc/how_to/get_started.rst
```diff
@@ -5,37 +5,37 @@
...
-In this introductory example, you will see how to use the SpikeInterface
-to perform a full electrophysiology analysis. We will download simulated
+In this introductory example, you will see how to use SpikeInterface to
+perform a full electrophysiology analysis. We will download a simulated
 dataset, and we will then perform some pre-processing, run a spike
 sorting algorithm, post-process the spike sorting output, perform
 curation (manual and automatic), and compare spike sorting results.
 
-.. code:: ipython
+.. code:: ipython3
```
h-mayorquin, line 14: I would suggest against this. What systems use python by default nowadays?

Python 2 is way behind now, let's hide this shameful past from the users.

doc/how_to/get_started.rst
```diff
@@ -189,64 +194,83 @@ a ``Probe`` and we don’t need to set it *manually*.
 
 
 
-.. image:: get_started_files/get_started_20_1.png
+.. image:: get_started_files/get_started_22_1.png
+
 
+If your recording does not have a ``Probe``, you can set it using
+``set_probe``. There is more information
```
h-mayorquin, line 201: Maybe we should add a comment that this does not sets the probe of the current recorder but creates a copy?

doc/how_to/get_started.rst
```diff
@@ -337,109 +368,221 @@ can print the units found using:
...
 
-SpikeInterface provides a efficient way to extract waveforms from paired
-recording/sorting objects. The ``extract_waveforms`` function samples
-some spikes (by default ``max_spikes_per_unit=500``) for each unit,
-extracts, their waveforms, and stores them to disk. These waveforms are
+This folder is where all the postprocessing data will be saved such as
+waveforms and templates. Let’s calculate some waveforms. When doing
+this, the function samples some spikes (by default
+``max_spikes_per_unit=500``) for each unit, extracts their waveforms,
+and stores them to disk in ``extensions/waveforms``. These waveforms are
```
h-mayorquin, line 414: Is the `analyzer_TDC_binary/extensions/waveforms` the full (relative) path here?

doc/how_to/get_started.rst
```diff
@@ -337,109 +368,221 @@ can print the units found using:
...
+
+    analyzer_TDC.delete_extension("spike_amplitudes")
 
+This deletes the extension’s data in the ``SortingAnalyzer`` folder.
 
-Importantly, waveform extractors (and all extensions) can be reloaded at
-later times:
+Importantly, ``SortingAnalyzers`` (and all extensions) can be reloaded
+at later times: (Here, spike_amplitudes is not loaded since we just
+deleted it)
```
h-mayorquin, line 547: Maybe a mention that this needs to be run from the same folder or form the same placer or othrewise use an absolute path? 

What do you think?

doc/how_to/get_started.rst
```diff
@@ -451,99 +594,411 @@ extensions - e.g., drift metrics resuire ``spike_locations``):
...
+    analyzer_TDC.compute("quality_metrics", qm_params)
+    analyzer_TDC.get_extension("quality_metrics").get_data()
 
 
-Quality metrics are also extensions (and become part of the waveform
-folder):
+.. parsed-literal::
 
-Next, we can use some of the powerful tools for spike sorting
+    /home/nolanlab/Chris/Developing/spikeinterface/src/spikeinterface/qualitymetrics/misc_metrics.py:846: UserWarning: Some units have too few spikes : amplitude_cutoff is set to NaN
```
h-mayorquin, line 657: I think it would look better if we supressed this warnings for the tutorial..

doc/how_to/get_started.rst
```diff
@@ -451,99 +594,411 @@ extensions - e.g., drift metrics resuire ``spike_locations``):
...
+      warnings.warn(f"Some units have too few spikes : amplitude_cutoff is set to NaN")
+    /home/nolanlab/Chris/Developing/spikeinterface/src/spikeinterface/qualitymetrics/misc_metrics.py:999: UserWarning: The recording is too short given the specified 'interval_s' and 'min_num_bins'. Drift metrics will be set to NaN
+      warnings.warn(
+    /home/nolanlab/Chris/Developing/spikeinterface/src/spikeinterface/qualitymetrics/misc_metrics.py:147: UserWarning: Bin duration of 60s is larger than recording duration. Presence ratios are set to NaN.
+      warnings.warn(
+
+
+
+
+.. raw:: html
```
h-mayorquin, line 667: What is this html? : O

h-mayorquin (commented):
Thanks for doing this. This is very useful to me.

examples/how_to/get_started.py
h-mayorquin: This is alsot lost in some of the loading functions that have no arguments specified.

examples/how_to/get_started.py
h-mayorquin: I would rather build some tunning tutorials in the "how to section" and then link them here once we have them.

doc/how_to/get_started.rst
zm711: This is due to the conversion from .py -> .rst. I fixed this in our old files but every time we regenerate the new one this switches back to ipython3. The pygment lexer that rtd uses doesn't like this designation which is why I switched it but this is an annoying mess. Sam mentioned we should look into tuning the .rst converter to try to fix this...

doc/how_to/get_started.rst
zm711: Agreed, but this appears in our example gallery too. It seems like no matter what we do some units trigger this. In my real data it happens too all the time!

doc/how_to/get_started.rst
h-mayorquin: Ah, OK, if it was not an intentional thing then feel free to ignore this @chrishalcrow . What a pain.

doc/how_to/get_started.rst
h-mayorquin: Yeah, we have a lot of non-actionable warnings like that. Maybe we should gather some in an issue and label them "good first issue" : P

doc/how_to/get_started.rst
zm711: Another converter issue :/ . It doesn't even respect the rtd boundaries so it looks super ugly, but this how the converter renders tables......

doc/how_to/get_started.rst
zm711: Yes. The tree is `whatever-you-named-your-analyzer/extensions/waveforms/waveform-stuff`. For `spike_locations` it would be `whatever-you-named-your-analyzer/extensions/spike_locations/spike_locations-stuff`. Or did I misunderstand your question?

doc/how_to/get_started.rst
zm711: I think this is important. Did you finally get rid of `inplace`?

doc/how_to/get_started.rst
h-mayorquin: Yeah, it was just confusing to me where `extensions/waveforms`.

So the full path would be. "path_from_where_i_run_the_file/how_i_named_the_analyzer/extensions/something"

I am not sure what would be more clear that just having the relative path with respect to the `analyzer_TDC_binary`. Maybe we can just write relative to `analyzer_TDC_binary` ?

doc/how_to/get_started.rst
h-mayorquin: Na, the issue is still there. 
https://github.com/SpikeInterface/spikeinterface/issues/2193

We never got Sam take on it.

doc/how_to/get_started.rst
chrishalcrow: Maybe it's slightly clearer if we write the code as two lines?

Importantly, ``SortingAnalyzers`` (and all extensions) can be reloaded
at later times from their saved folders:

sorting_analyzer_path = 'analyzer_TDC_binary'
analyzer_loaded = si.load_sorting_analyzer(sorting_analyzer_path)

How about something like this?

doc/how_to/get_started.rst
chrishalcrow: I've added a comment

doc/how_to/get_started.rst
zm711: I'm wondering if the part of the question is where is `analyzer_TDC_binary`. We could emphasize that relativity by doing
```
sorting_analyzer_path = './analyzer_TDC_binary'
```

Although using the `./` is not completely beginner friendly. But for people that know it adds emphasis to location....

In our other docs that don't come from a `.py` conversion we often do 
```
path = 'where-i-want-my-data/analyzer_TDC_binary'
```
or some other descriptive path, but that doesn't really work here.

examples/how_to/get_started.py
```diff
@@ -59,10 +59,12 @@
...
-#  We can also import all submodules at once with this
-#  this internally imports core+extractors+preprocessing+sorters+postprocessing+
-#  qualitymetrics+comparison+widgets+exporters
-#
+# Alternatively, we can import all submodules at once with `import spikeinterface.full as si` with `import spikeinterface.full as si` which
+# internally imports core+extractors+preprocessing+sorters+postprocessing+
+# qualitymetrics+comparison+widgets+exporters. In this case all aliases in the following tutorial
+# would be `si`. In this case all aliases in the following tutorial
+# would be `si`.
+
```
zm711, line 67: I think it duplicated the lines, but I don't know why :)

doc/how_to/get_started.rst
h-mayorquin: Both adding a second line and using the linux notation for current directoy seems like an improvement to me.

doc/how_to/get_started.rst
h-mayorquin: Yes, that makes sense to me. Maybe also adding the linux dot for current directory `./` as @zm711 suggests?

doc/how_to/get_started.rst
zm711: It works for Windows too with `Path('r.\abc\')` :) or even `Path('./abc/'), but I think you're right in that I could easily google what does the period at the beginning of a file path mean and find out. So adding this is clearer and even for a beginner it should be relatively easy to google to find out.

doc/how_to/get_started.rst
chrishalcrow: Sounds good. I've changed to './analyzer_TDC_binary/extension/waveforms"

doc/how_to/get_started.rst
chrishalcrow: Perfect, I've added both

zm711:
Looks good to me now. Thanks for dealing with all the comments :)

alejoe91 (approved):

samuelgarcia:
merci

PR #2584, opened 2024-03-15
Fix to impose sorting of the peaks in from_peaks()

yger:
As discussed with Sam, we should impose a sorting of the peaks while creating a sorting, to avoid posible bugs downstream

### DIFF
--- a/src/spikeinterface/core/numpyextractors.py
+++ b/src/spikeinterface/core/numpyextractors.py
@@ -457,7 +457,8 @@ class NumpySorting(BaseSorting):
         spikes["sample_index"] = peaks["sample_index"]
         spikes["unit_index"] = peaks["channel_index"]
         spikes["segment_index"] = peaks["segment_index"]
-
+        order = np.argsort(spikes["sample_index"])
+        spikes = spikes[order]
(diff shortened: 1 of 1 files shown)


alejoe91 (approved):

src/spikeinterface/core/numpyextractors.py
```diff
@@ -457,7 +457,8 @@ def from_peaks(peaks, sampling_frequency, unit_ids) -> "NumpySorting":
         spikes["sample_index"] = peaks["sample_index"]
         spikes["unit_index"] = peaks["channel_index"]
         spikes["segment_index"] = peaks["segment_index"]
-
+        order = np.argsort(spikes["sample_index"])
+        spikes = spikes[order]
```
samuelgarcia, line 461: This is wrong!!!!!!!
You have to sort by segment index and then sample index.

samuelgarcia:
This patch is wrong!
Pierre can you submit another PR taking in account the segment index also ?

src/spikeinterface/core/numpyextractors.py
yger: Good catch, always forgetting the segment. This should now work with lexsort

PR #2585, opened 2024-03-15
Save extension class info.

samuelgarcia:


src/spikeinterface/core/core_tools.py
```diff
@@ -476,3 +477,24 @@ def normal_pdf(x, mu: float = 0.0, sigma: float = 1.0):
     """
 
     return 1 / (sigma * np.sqrt(2 * np.pi)) * np.exp(-((x - mu) ** 2) / (2 * sigma**2))
+
+
+def get_class_info(a_class):
```
h-mayorquin, line 482: I strongly suggest against something as vage as `get_class_info` for a function. If I remember well the point of storing this attributes is to store information about how to instantiate the class. From that perspective I think that something like:

* `get_importing_provenance`
* `get_class_origin` 
* `retrieve_class_path` 

or a combination of them. I like `retrieve_importing_provenance` because I think it describes well what it does and it also does not add more suggestions to the get methods that are more used for user relevant information.

src/spikeinterface/core/core_tools.py
```diff
@@ -476,3 +477,24 @@ def normal_pdf(x, mu: float = 0.0, sigma: float = 1.0):
...
+    """
```
h-mayorquin, line 483: I think this docstring is doing too little.
```suggestion
    """
    Retrieve the import provenance of a class, including its an import name that consists of the class name and the module, the top-level module, and the module version.

    Parameters
    ----------
    a_class : type
        The class object for which to retrieve the import provenance.

    Returns
    -------
    dict
        A dictionary containing:
        - 'class':  The module path and the name of the class concatenated (e.g., 'package.subpackage.ClassName').
        - 'module': The top-level module name where the class is defined.
        - 'version': The version of the module if available, otherwise 'unknown'.
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1465,6 +1468,21 @@ def _save_params(self):
             extension_group = self._get_zarr_extension_group(mode="r+")
             extension_group.attrs["params"] = check_json(params_to_save)
 
+    def _save_info(self):
```
h-mayorquin, line 1471: Same comment as for `get_class_info`, this is too vague. Aligned with the above I suggest `save_importing_provenance`.

src/spikeinterface/core/core_tools.py
samuelgarcia: Totally agree.
THanks you for the feedback. lets go for retrieve_importing_provenance

src/spikeinterface/core/sortinganalyzer.py
samuelgarcia: yes

src/spikeinterface/core/core_tools.py
```diff
@@ -476,3 +477,38 @@ def normal_pdf(x, mu: float = 0.0, sigma: float = 1.0):
     """
 
     return 1 / (sigma * np.sqrt(2 * np.pi)) * np.exp(-((x - mu) ** 2) / (2 * sigma**2))
+
+
+def retrieve_importing_provenance(a_class):
+    """
+    Retrieve the import provenance of a class, including its an import name that consists of the class name and the module, the top-level module, and the module version.
```
alejoe91, line 484:
```suggestion
    Retrieve the import provenance of a class, including its import name (that consists of the class name and the module), the top-level module, and the module version.
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1465,6 +1468,21 @@ def _save_params(self):
...
+        # this save class info, this is not uselfull at the moment but this could be usefull in futur
+        # if some class change the data model and if we need to make backwards compatibility
+        # we have the same machanism in base.py for recording and sorting
```
alejoe91, line 1474:
```suggestion
        # this saves the class info, this is not uselful at the moment but could be useful in future
        # if some class changes the data model and if we need to make backwards compatibility
        # we have the same machanism in base.py for recording and sorting
```

src/spikeinterface/sortingcomponents/clustering/split.py
```diff
@@ -231,6 +231,17 @@ def split(
...
+            unique_feat, counts_feat = np.unique(final_features[:, 0], return_counts=True)
+            import matplotlib.pyplot as plt
+
+            fig, ax = plt.subplots()
+            ax.hist(final_features[:, 0])
+            fig, ax = plt.subplots()
+            ax.plot(unique_feat, counts_feat)
+            plt.show()
+
+            print(np.sum(counts_feat > 1))
```
alejoe91, line 244: remove?

samuelgarcia: oups! unrelated.

h-mayorquin (approved):
I think this is a good idea. This looks good to me.

PR #2594, opened 2024-03-18
Fix for sorting from_peaks with lexsort (handling multi segments)

yger:


src/spikeinterface/core/numpyextractors.py
```diff
@@ -457,7 +457,7 @@ def from_peaks(peaks, sampling_frequency, unit_ids) -> "NumpySorting":
         spikes["sample_index"] = peaks["sample_index"]
         spikes["unit_index"] = peaks["channel_index"]
         spikes["segment_index"] = peaks["segment_index"]
-        order = np.argsort(spikes["sample_index"])
+        order = np.lexsort((spikes["segment_index"], spikes["sample_index"]))
```
alejoe91, line 460:
```suggestion
        order = np.lexsort((spikes["sample_index"], spikes["segment_index"]))
```
@yger lexsort works from outer to inner!

try this:
```
rec, sort = si.generate_ground_truth_recording(durations=[10, 10])
spikes = sort.to_spike_vector()
spikes = spikes[np.random.permutation(len(spikes))]
order = np.lexsort((spikes["segment_index"], spikes["sample_index"]))
print(spikes[order][:20])
```

alejoe91 (changes requested):
See comment!

yger:
Oups, indeed, good catch

src/spikeinterface/core/numpyextractors.py
yger: done

alejoe91 (approved):

samuelgarcia (approved):

samuelgarcia:
merci camarades.

Issue #2544, opened 2024-03-05
How to get started


mniazy1:
I am a beginner in using this, and I have no prior knowledge in this field. Currently, I just installed Spike Interface into Anaconda, and I want to move into testing and getting familiar with the interface. However, I am not sure how to get started with "How to 'get started'." Furthermore, I'm not sure if everything is installed fully since I followed the beginners Installation tips, (the cleanup_for_windows.py is not working / invalid). 

I have the Anaconda command prompt open into (si_env) cd'd into Documents. What I want to know is where do I run the commands / code from the How to get started tutorials, such as an Import for example. Do I use Spyder from Anaconda? Or through the Anaconda command prompt (si_env)? Or is it somewhere else to run and operate everything.

zm711:
Howdy @mniazy1,

What do you mean it is not working? What specific error is it giving? We can better diagnose what's going of if we know exactly what is happening. The `check_your_install` can just be run in the command prompt. Ie you should do

```bash
conda activate si_env
python check_your_install.py
```

Then it should run through some testing. If this fails let us know. When it finishes you should do:

```bash
python cleanup_for_windows.py
```


If you want to then run code with the installation_tips env, we have it install Jupyter and so you can create a Jupyter notebook and run any of the tutorials from there. If you want to use Spyder then you would either need to include it in the env (more beginner friendly less storage friendly) or work with ipython kernels (more storage friendly less beginner friendly). 

Just so we can help you the best possible are you a python beginner or a spike sorting beginner? Or just a conda beginner?

mniazy1:
Hello @zm711, sorry for the quite late reply!

For the "cleanup_for_windows.py" in the installation process (step 5), the file itself shows this when clicked on "400: Invalid request." And it is not downloadable as per the steps, so running the second command you showed does not work.

So you're saying after installing Spike Interface and activating the environment, I can then perform the "how to get started" steps through Jupyter notebook from Anaconda?

I have a bit of python experience but I would not say I am an intermediate. As for spike sorting I have 0 previous knowledge and have not worked with it at all.

Thank you!

zm711:
No worries @mniazy1 :) 

I just fixed the bad link. Thanks for finding that. If you want you can clone the whole repo with git or you can just open that file and copy the contents into a `.py` file yourself and then run it locally or you can wait until we merge that fix and then you can download the fixed version.

[...]

PR #2595, opened 2024-03-18
Fix `installation_tips` links

zm711:
in #2544 we discovered that one of the files had a broken link (I think it's my bad sorry!), so I fixed the link. I also switch all the links over to main from master. Although github autoconverted them it's probably better just to link them to main just in case :)


samuelgarcia (approved):

samuelgarcia:
merci.

PR #2600, opened 2024-03-19
Proposal to inject anisotropy while creating fake templates

yger:
This PR refines the way fake templates are currently generated. An ellipsoid centered on the positions of the fake neurons is generated with random dimensions (a,b,c) and rotations angles (Rx, Ry, Rz). Then, the distances of this 3D volume are projected onto the 2D space of the electrodes, and this should create some heterogeneity while creating the spatio-temporal waveforms. By doing so, we are ensuring that the amplitudes are decaying as function of distances in a non-uniform manner


samuelgarcia:
Cool. I check this soon.

samuelgarcia (approved):

Issue #2599, opened 2024-03-18
compare_multiple_templates error


AOONIH:
Hi,

I'm trying to compare sorter outputs over days and tried using the compare_multiple_templates function.  
I get this error:
ValueError: Shape of passed values is (191, 191), indices imply (191, 132)
which comes from: `self.agreement_scores = pd.DataFrame(agreement_scores, index=self.unit_ids[0], columns=self.unit_ids[1])`
I'm not sure why agreement_scores is a square matrix if the number of units is different across the 2 sorters?

AOONIH:
I think I found the issue:
Line 115 in template_similarity.py specifies the function postion arguments incorrectly:
`_compute_template_similarity(waveform_extractor, waveform_extractor_other, method)`

zm711:
Fixed by #2601.

PR #2601, opened 2024-03-19
Fix template similarity when there are multiple sizes of units

chrishalcrow:
Hopefully fixing issue #2599

And added a test to check that the agreement_scores are the expected size.

### DIFF
--- a/src/spikeinterface/postprocessing/template_similarity.py
+++ b/src/spikeinterface/postprocessing/template_similarity.py
@@ -112,7 +112,9 @@ def compute_template_similarity(
         similarity = tmc.get_data()
         return similarity
     else:
-        return _compute_template_similarity(waveform_extractor, waveform_extractor_other, method)
+        return _compute_template_similarity(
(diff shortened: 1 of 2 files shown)


src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -112,7 +112,9 @@ def compute_template_similarity(
         similarity = tmc.get_data()
         return similarity
     else:
-        return _compute_template_similarity(waveform_extractor, waveform_extractor_other, method)
+        return _compute_template_similarity(
+            waveform_extractor=waveform_extractor, waveform_extractor_other=waveform_extractor_other, method=method
```
zm711, line 116: Another win for keyword arguments :)

alejoe91 (approved):

samuelgarcia:
Is it also done on the main branch ?

src/spikeinterface/postprocessing/template_similarity.py
h-mayorquin: Indeed!

PR #2605, opened 2024-03-20
compute_synchrony_metrics update

chrishalcrow:
(Another go of #2528, after the SortingAnalyzer refactor)

Planning to standardise, correctness test and optimise the quality metrics module. First one: the `compute_synchrony_metrics` function.

Two main things in this:
- The generating function `synthesize_random_firings` and `synthesize_poisson_spike_vectors` were adjusted so that you can insert specific spikes, useful for testing. These are used in the new tests.
- A new function `compute_synchrony_counts` is introduced, which computes the synchrony counts.

One thing I'm not sure about: The documentation said that `compute_synchrony_metrics` returns a dict, but I think we should export a namedtuple, which matches most other quality metrics.

I adjusted the algorithm in `compute_synchrony_counts`, which should be easier to understand and is significantly faster. Here are some benchmarking results, with times and create a log-log plot to check the algorithm's complexity. About 30x faster!

benchmarch_sync

Benchmarked using the following code:

```
import numpy as np
from time import perf_counter

from spikeinterface.core import synthesize_random_firings
from spikeinterface.qualitymetrics import get_synchrony_counts

unit_ids = np.arange(100)

Ns = 10**np.array(np.arange(3,5.1,0.1))

times_by_size=[]
how_many = 100

for train_length in Ns:

    times_list = []

    for _ in range(0,how_many):

        spike_train = synthesize_random_firings(num_units=100, duration=train_length/1000, firing_rates=10.0)

        start = perf_counter()
        get_synchrony_counts(spike_train, np.array((2,4,8)), list(unit_ids))        
        end = perf_counter()

        times_list.append(end - start)

    times_by_size.append( np.median( times_list ) )
```


alejoe91:
Thanks @chrishalcrow 

Can you heck why tests are failing?

src/spikeinterface/core/generate.py
```diff
@@ -709,14 +714,21 @@ def synthesize_poisson_spike_vector(
...
-    unit_indices = unit_indices[sort_indices]
-    spike_frames = spike_frames[sort_indices]
+    spike_train.sort(order=["sample_index"])
 
-    return spike_frames, unit_indices
+    segment_length = train_length // segments
+    for i in range(train_length):
+        spike_train["segment_index"][i] = min(i // segment_length, segments - 1)
+
+    return spike_train
```
h-mayorquin, line 731: Please don't change the output of a function in an unrelated PR. This might be a good idea but is a big change as we are violating backwards compatbility so it should be discussed in more detail. 

In particular here, I optimized a lot this function to work very well with that and did not intend it to produce structured output. If you want to push for that we should discuss it outside but you can achieve the same result doing that post-processing outside.

src/spikeinterface/core/generate.py
```diff
@@ -801,11 +821,84 @@ def synthesize_random_firings(
...
+
+        structured_insertions["sample_index"][i] = insert[0]
+        structured_insertions["unit_index"][i] = insert[1]
+        if len(insert) == 3:
+            structured_insertions["segment_index"][i] = insert[2]
+        else:
+            structured_insertions["segment_index"][i] = 0
 
-    return (times, labels)
+    return structured_insertions
```
h-mayorquin, line 901: Same coment here about the output. Input and output are your contracts with your users. We should be careful with changing them (and careful to define them in the first place)

src/spikeinterface/core/generate.py
```diff
@@ -801,11 +821,84 @@ def synthesize_random_firings(
...
+    else:
+        new_spike_train = np.append(spike_train, insertions)
+
+    return new_spike_train
+
+
+def get_structured_insertions(insertions):
+
+    structured_insertions = np.zeros(
+        len(insertions), dtype=[("sample_index", "int64"), ("unit_index", "int64"), ("segment_index", "int64")]
```
h-mayorquin, line 888: I suggest using the `minimum_spike_dtype` so  this is properly coupled or defining it and adding a comment exlpaining why this is mean to be different.

h-mayorquin (changes requested):

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -496,7 +496,51 @@ def compute_sliding_rp_violations(
...
 
-def compute_synchrony_metrics(sorting_analyzer, synchrony_sizes=(2, 4, 8), unit_ids=None, **kwargs):
+def get_synchrony_counts(spikes, synchrony_sizes, all_unit_ids):
+    """Compute synchrony counts, the number of simultaneous spikes with sizes `synchrony_sizes`
+
+    Parameters
+    ----------
+    spikes : memmap
+        The spike train
+    synchrony_sizes : numpy array
```
h-mayorquin, line 506: Is this given in seconds, samples, milliseconds?

h-mayorquin:
Tip:
You have https://numpy.org/doc/stable/reference/generated/numpy.logspace.html to generate your samples. 

Benchamark comments:
*  I think you should fix your seed in your benchmark explictly.
* I find the `spike_length` variable hard to think about. I suggest using time and show us the units. Does it scale well to the relevant domain of hour long sortings?
* Finally, I would be curios to see how this scales with the number of units as well.

src/spikeinterface/core/generate.py
chrishalcrow: Sorry @h-mayorquin - I was only thinking about where this function was used in spikeinterface itself. Wasn't thinking about it being exposed to the user. Will change this :)

src/spikeinterface/core/generate.py
chrishalcrow: Indeed, I'll add some documentation. I think this (and the add_insertions) functions don't have much use outside of testing. Should I "hide" them (i.e. make them _add_insertions etc)?

chrishalcrow:
> Tip: You have https://numpy.org/doc/stable/reference/generated/numpy.logspace.html to generate your samples.
> 
> Benchamark comments:
> 
> * I think you should fix your seed in your benchmark explictly.
> * I find the `spike_length` variable hard to think about. I suggest using time and show us the units. Does it scale well to the relevant domain of hour long sortings?
> * Finally, I would be curios to see how this scales with the number of units as well.

Oooh, thanks for the logspace function.

Maybe it would be helpful to say: the new algorithm is expected to be faster. I was able to replace the use of np.complexity, isin and count_nonzero (I think complexity is the heavy function of these) with some simple selections (like `unique_spike_index[counts >= 2]`).

Spike_length is just the number of spikes in the spike vector. So for generic data, this increases linearly with time. 

I’ve been using a 30 minute 16 channel recording from my lab, which contains 70,000 spikes after sorting. Is this realistic? I’ve also been using this to make sure my algorithm is consistent with the old results. My tests go up to 100,000 spikes. What would a good number to go up to?

I'm testing with unit number now, going up to 1000 units. Is that high enough? The complexity function changes it’s method at some point, and I’m trying to figure out how and when and will report back soon! But the new code is still substantially faster at e.g. 1k units with 100k spikes.

src/spikeinterface/core/generate.py
h-mayorquin: If you think that they will only be used for internal testing then the best is to have them in the corresponding testing module I feel.

I think hiding by default is a good option and then you make it public once you feel confident about the interface and purpose. That minimizes the risk if you want to change the function later. But this is my opinion, I don't think that anyone else would find any problems if you want to add some functions.

src/spikeinterface/core/generate.py
h-mayorquin: Thanks, you can do all this processing outside of `synthesize_poisson_spike_vector`  where you needed. 

My goal for `synthesize_poisson_spike_vector` is to keep a simple function that can be used in many places. For that purpose, I prefer a simple signature that returns a mostly bare bones object so the function can be easily understood, modified and optimized if necessary. Adding more parameters, options and complexity goes against that purpose and I don't think is really necessary at that level.

h-mayorquin:
> Spike_length is just the number of spikes in the spike vector. So for generic data, this increases linearly with time.

On a second thought, this seems more amenable to control and more relevant than duration. It is just probably me that does not have a  good intuitive grasp of this metric while duration I understand better. How are you controlling this though, given that you are passing this parameter in duration divided by 1000?

> I'm testing with unit number now, going up to 1000 units. Is that high enough? The complexity function changes it’s method at some point, and I’m trying to figure out how and when and will report back soon! But the new code is still substantially faster at e.g. 1k units with 100k spikes.

1000 I have been told than 1000 units is like a very high number. 

Bear in mind that the comments I made about the benchmark is just me being curios and suggestions. As long as Sam and Alessio are OK with the rest I would not bother if you don't care about it. The thing that I really feel strongly about is for not changing the signature and output of  `synthesize_poisson_spike_vector`.

chrishalcrow:
> > Spike_length is just the number of spikes in the spike vector. So for generic data, this increases linearly with time.
> 
> On a second thought, this seems more amenable to control and more relevant than duration. It is just probably me that does not have a good intuitive grasp of this metric while duration I understand better. How are you controlling this though, given that you are passing this parameter in duration divided by 1000?
> 
> > I'm testing with unit number now, going up to 1000 units. Is that high enough? The complexity function changes it’s method at some point, and I’m trying to figure out how and when and will report back soon! But the new code is still substantially faster at e.g. 1k units with 100k spikes.
> 
> 1000 I have been told than 1000 units is like a very high number.
> 
> Bear in mind that the comments I made about the benchmark is just me being curios and suggestions. As long as Sam and Alessio are OK with the rest I would not bother if you don't care about it. The thing that I really feel strongly about is for not changing the signature and output of `synthesize_poisson_spike_vector`.

Hello,

I learned a few things from your comments and trying to recreate my previous benchmarking. Thanks! Before: I refactored the code, creating a new function compute_synchrony_count. Then I did a speed test on the new function using the old algorithm against an optimised version of it. This is hard for others to reproduce since you can’t easily run compute_synchrony_count. So now I’m benchmarking against the full compute_synchrony_metrics function. This should be closer to “real world” use than before and easier for anyone to check. And I’ve hopefully made the wording/details of the plots more obvious.

The old benchmarking results are basically unchanged.

And I’ve now done more benchmarking, with changing units. First, here’s a plot of timings with increasing units N but fixed spike train length (of about 50k: in our lab, this is the number of spikes in a typical 20 minute recording). The new algorithm grows linearly with the number of spikes. The old algorithm, which uses more complex functions, changes tactic at ~600 units (I think it’s np.unique that changes its tactic…). Both its tactics look like O(N) algorithms. But even when it changes, it is still about 10x slower than the new one.

benchmarch_sync_N


I also tried keeping the duration and firing rates fixed. So here, as the number of units N increases the length of the spike train increases at the same rate. But the density of spikes per time unit stays constant. Both algorithms grow as N^2, as expected (since units and train length are both increasing with N). Again, the old algorithm changes tactic  but the new code is still significantly faster.

benchmarch_sync_N_sq

The code for both is below:

```
times_by_size=[]
sizes = []

how_many = 100

duration = 10**1.7
# duration = 10**1.5 for the first plot
nums_units = np.arange(100,1201,100)

for num_units in nums_units:

    recording, sorting = generate_ground_truth_recording(
            durations=[
               # duration # for the first plot
               # duration/num_units*nums_units[0] # for the second plot
            ],
            sampling_frequency=30_000.0,
            num_channels=6,
            num_units=num_units,
            generate_sorting_kwargs=dict(firing_rates=10.0, refractory_period_ms=4.0),
            noise_kwargs=dict(noise_level=5.0, strategy="tile_pregenerated"),
            seed=2205,
        )

    sorting_analyzer = create_sorting_analyzer(sorting, recording, format="memory", sparse=False)

    times_list = []
    for _ in range(0,how_many):

        start = perf_counter()
        compute_synchrony_metrics(sorting_analyzer)        
        end = perf_counter()

        times_list.append(end - start)

    times_by_size.append( np.median( times_list ) )
    sizes.append( len(sorting_analyzer.sorting.to_spike_vector()) )

```

src/spikeinterface/core/generate.py
chrishalcrow: Hello, should all be reverted now :)

src/spikeinterface/core/generate.py
chrishalcrow: I'd prefer to keep them close to the synthesize_stuff functions, since they are closely related. I'm a fan of the "related code should be close together" mindset. Maybe they should all be in the testing module? But maybe for another time...

Great, I've hidden it and also added some documentation. Best of all worlds.

src/spikeinterface/qualitymetrics/misc_metrics.py
chrishalcrow: I've made the description more accruate (since it's actually a structured numpy array). One of the fields is "sample_index" which clarifies the unit.

src/spikeinterface/core/generate.py
```diff
@@ -801,13 +821,63 @@ def synthesize_random_firings(
...
+        times, labels = _add_insertions([times, labels], insertions=insertions, insertions_replace=insertions_replace)
+
     sort_inds = np.argsort(times)
     times = times[sort_inds]
     labels = labels[sort_inds]
 
     return (times, labels)
 
 
+def _add_insertions(spike_train, insertions=None, insertions_replace=False, rng=None, seed=None):
```
h-mayorquin, line 834: Passing the seed is easier than passing the rng I feel.

src/spikeinterface/qualitymetrics/misc_metrics.py
h-mayorquin: I was talking about the synchrony_sizes

src/spikeinterface/core/generate.py
```diff
@@ -709,6 +716,12 @@ def synthesize_poisson_spike_vector(
     spike_frames[:num_correct_frames] = spike_frames[mask]  # Avoids a malloc
     unit_indices = unit_indices[mask]
 
+    # Add insertions
```
h-mayorquin, line 719: As I mentioned before, I would like to keep this function simple. I don't want to add extra parameters and deal with the complexity that they bring.

I was suggestion to add these lines in your tests instead of adding an extra parameter to a low level function:

```python
        spike_frames, unit_indices = synthesize_poisson_spike_vector(**)
        spike_frames, unit_indices = _add_insertions(
            [spike_frames, unit_indices], insertions=insertions, insertions_replace=insertions_replace
        )
```

Or create another function called `synthesize_poisson_spike_vector_with_insertions` that calls this one and wraps your argument if you think that the function will be used a lot.

Am I missing something? Is there something important to gain by adding this? 
Fuctions with less parameters are easier to test, document and mantain and it seems that the win here will be saving one line of code writing.

h-mayorquin (commented):
Hey, thanks for removing the change of the output and doing benchmark with units.

I sitll would like to keep the function for generating spikes as simple as possible so I request a removal or the arguments.

Unfortunately, I can't review the synchrony metrics changes as I am not really well acquainted with the method. Maybe @DradeAW or @zm711 can take a look?

zm711:
> Unfortunately, I can't review the synchrony metrics changes as I am not really well acquainted with the method. Maybe @DradeAW or @zm711 can take a look?

@alejoe91 actually ported this from Elephant so he would be the best to review unless @DradeAW knows this super well. I never actually use this metric :)

chrishalcrow:
> > Unfortunately, I can't review the synchrony metrics changes as I am not really well acquainted with the method. Maybe @DradeAW or @zm711 can take a look?
> 
> @alejoe91 actually ported this from Elephant so he would be the best to review unless @DradeAW knows this super well. I never actually use this metric :)

An important part of the update are the four `test_synchrony_counts_blah` in test_metrics_functions.py . These are correctness tests for the main part of the algorithm. I make a small spike_train with certain spikes (e.g. two spikes which fire at the same time) and make sure the metric gives the right answer.  If anyone can think of any good edge cases to test here, please comment!

src/spikeinterface/qualitymetrics/misc_metrics.py
chrishalcrow: Oh, sorry. These are the number of synchronous events you want to count. So if you want to see when two or four spikes fire at the same time you use synchrony_sizes = (2,4). So it's an integer, and I think it will be clear for anyone who knows enough about the metric that their using it.

h-mayorquin: Thanks. I was my lack of familarity is evident then. I somehow imagined that it was the windows in which an event would be count as syncronouse. Thanks for explaining.

h-mayorquin (commented):
Thanks for bearing with me and for adding thorough testing.

src/spikeinterface/core/generate.py
h-mayorquin: I think this is a good decision. If we ever use the function more and we are decided on the signature, then we can make it public.

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -521,49 +565,39 @@ def compute_synchrony_metrics(sorting_analyzer, synchrony_sizes=(2, 4, 8), unit_
     This code was adapted from `Elephant - Electrophysiology Analysis Toolkit <https://github.com/NeuralEnsemble/elephant/blob/master/elephant/spike_train_synchrony.py#L245>`_
     """
     assert min(synchrony_sizes) > 1, "Synchrony sizes must be greater than 1"
-    spike_counts = sorting_analyzer.sorting.count_num_spikes_per_unit(outputs="dict")
-    sorting = sorting_analyzer.sorting
-    spikes = sorting.to_spike_vector(concatenated=False)
+    # Sort the synchrony times so we can slice numpy arrays, instead of using dicts
+    synchrony_sizes_np = np.array(synchrony_sizes, dtype=np.int64)
```
alejoe91, line 569:
```suggestion
    synchrony_sizes_np = np.array(synchrony_sizes, dtype=np.int16)
```
I think that int16 is way more than enough!

src/spikeinterface/core/generate.py
```diff
@@ -831,6 +831,52 @@ def clean_refractory_period(times, refractory_period):
     return times
 
 
+def _add_insertions(spike_train, insertions=None, insertions_replace=False, seed=None):
```
alejoe91, line 834:
```suggestion
def add_spikes_to_spiketrain(spike_indices, spike_labels, segment_indices=None, added_spike_indices=None, added_spike_labels=None, added_segment_indices=None, replace=False, seed=None):
```

sorting.to_spike_vector()

src/spikeinterface/core/generate.py
```diff
@@ -831,6 +831,52 @@ def clean_refractory_period(times, refractory_period):
...
+    insertions_replace: bool, default: False
+        If True, randomly replace generated spikes. If False, add to generated spikes.
+    rng: numpy.random.Generator
+        A random number generator
+    seed: int, default: None
+        seed for the generator
+
+    Returns
+    -------
+        spike_train: np.ndarray
```
alejoe91, line 854:
```suggestion
    spike_train: np.ndarray
```

src/spikeinterface/core/generate.py
```diff
@@ -845,6 +845,85 @@ def clean_refractory_period(times, refractory_period):
     return times
 
 
+def _add_spikes_to_spiketrain(
+    spike_indices,
+    spike_labels,
+    segment_indices=[],
```
alejoe91, line 851:
```suggestion
    segment_indices=None,
```
See https://github.com/SpikeInterface/spikeinterface/pull/2345

src/spikeinterface/core/generate.py
```diff
@@ -845,6 +845,85 @@ def clean_refractory_period(times, refractory_period):
...
+    added_spikes_indices=None,
+    added_spikes_labels=None,
+    added_segment_indices=[],
```
alejoe91, line 854:
```suggestion
    added_segment_indices=None,
```

src/spikeinterface/core/generate.py
chrishalcrow: To quote Joe: "Thanks I did not know about that, I am learning so much working on the SI codebase 😍!"

chrishalcrow:
Hello, realised that the `_add_spikes_to_spiketrain` function was basically just checking a bunch of array lengths then concatenating some arrays together, that were then used for testing. So it wasn't really adding much, and it was actually much simpler to get rid of it and concatenate in the tests directly. Got rid of 200 lines of code, and simplified the pull request a lot!

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -521,49 +565,39 @@ def compute_synchrony_metrics(sorting_analyzer, synchrony_sizes=(2, 4, 8), unit_
...
 
-        # we compute just by counting the occurrence of each sample_index
-        unique_spike_index, complexity = np.unique(spikes_in_segment["sample_index"], return_counts=True)
+    spikes = sorting.to_spike_vector()
+    all_unit_ids = sorting.unit_ids
+    synchrony_counts = get_synchrony_counts(spikes, synchrony_sizes_np, all_unit_ids)
+
+    synchrony_metrics_dict = {}
+    for sync_idx, synchrony_size in enumerate(synchrony_sizes_np):
+        synch_id_metrics_dict = {}
```
alejoe91, line 584:
```suggestion
        sync_id_metrics_dict = {}
```
Can we stick to the same naming? :)

Sorry for being overly annoying @chrishalcrow !!!

chrishalcrow: You can name it anything you want if I can stop working on this ;) Just you wait: my next quality_metrics update is gonna be so careful!

alejoe91 (approved):

samuelgarcia:
Hi.
When we will have time I would be happy to discuss this function.
The synchrony here if I understand correctly is the exact same timestamp but we could have synchrony with very small lags. A lag can be due to small noise and small time shift.
So my main question is how robut is that mertics regarding theses lags ?

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -496,7 +496,51 @@ def compute_sliding_rp_violations(
 )
 
 
-def compute_synchrony_metrics(sorting_analyzer, synchrony_sizes=(2, 4, 8), unit_ids=None, **kwargs):
+def get_synchrony_counts(spikes, synchrony_sizes, all_unit_ids):
```
samuelgarcia, line 499: I would excpet a delta in the signature no ?

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -496,7 +496,51 @@ def compute_sliding_rp_violations(
...
+    synchrony_counts : dict
+        The synchrony counts for the synchrony sizes.
+
+    References
+    ----------
+    Based on concepts described in [Gruen]_
+    This code was adapted from `Elephant - Electrophysiology Analysis Toolkit <https://github.com/NeuralEnsemble/elephant/blob/master/elephant/spike_train_synchrony.py#L245>`_
+    """
+
+    synchrony_counts = np.zeros((np.size(synchrony_sizes), len(all_unit_ids)), dtype=np.int64)
```
samuelgarcia, line 522: In this the synhcony is higly dependant on the sampling rate no ?

alejoe91:
Let's keep it like this for now, since this was the original design. :)

PR #2608, opened 2024-03-20
Analyzer core extension: improve docstrings

alejoe91:
Also changed:
- `compute_select_random_spikes` to `compute_random_spikes` 
- `some_spikes()` function to `random_spikes()`

### DIFF
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -23,7 +23,7 @@ class SelectRandomSpikes(AnalyzerExtension):
     """
     AnalyzerExtension that select some random spikes.
 
-    This will be used by the `waveforms`/`templates`/`fast_templates` extension.
+    This will be used by the `waveforms`/`templates`/`fast_templates` extensions.
 
     This internally use `random_spikes_selection()` parameters are the same.
 
@@ -465,7 +465,7 @@ class ComputeFastTemplates(AnalyzerExtension):
     """
     AnalyzerExtension which is similar to the extension "templates" (ComputeTemplates)
     **but only for average**.
-    This is way faster because it do not need "waveforms" to be computed first.
+    This is way faster because it does not need "waveforms" to be computed first.
 
     Parameters
     ----------
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -1,6 +1,6 @@
 """
 Implement AnalyzerExtension that are essential and imported in core
-  * SelectRandomSpikes
+  * ComputeRandomSpikes
   * ComputeWaveforms
   * ComputeTemplates
 Theses two classes replace the WaveformExtractor
@@ -19,7 +19,7 @@ from .template import Templates
 from .sorting_tools import random_spikes_selection
 
 
-class SelectRandomSpikes(AnalyzerExtension):
+class ComputeRandomSpikes(AnalyzerExtension):
     """
     AnalyzerExtension that select some random spikes.
 
@@ -81,7 +81,7 @@ class SelectRandomSpikes(AnalyzerExtension):
     def _get_data(self):
         return self.data["random_spikes_indices"]
 
-    def random_spikes(self):
+    def get_random_spikes(self):
         # utils to get the some_spikes vector
         # use internal cache
         if not hasattr(self, "_some_spikes"):
@@ -107,8 +107,8 @@ class SelectRandomSpikes(AnalyzerExtension):
         return selected_spikes_in_spike_train
 
 
-compute_random_spikes = SelectRandomSpikes.function_factory()
-register_result_extension(SelectRandomSpikes)
+compute_random_spikes = ComputeRandomSpikes.function_factory()
+register_result_extension(ComputeRandomSpikes)
 
 
 class ComputeWaveforms(AnalyzerExtension):
@@ -156,7 +156,7 @@ class ComputeWaveforms(AnalyzerExtension):
         unit_ids = sorting.unit_ids
 
         # retrieve spike vector and the sampling
-        some_spikes = self.sorting_analyzer.get_extension("random_spikes").random_spikes()
+        some_spikes = self.sorting_analyzer.get_extension("random_spikes").get_random_spikes()
 
         if self.format == "binary_folder":
             # in that case waveforms are extacted directly in files
@@ -223,7 +223,7 @@ class ComputeWaveforms(AnalyzerExtension):
 
     def _select_extension_data(self, unit_ids):
         # random_spikes_indices = self.sorting_analyzer.get_extension("random_spikes").get_data()
-        some_spikes = self.sorting_analyzer.get_extension("random_spikes").random_spikes()
+        some_spikes = self.sorting_analyzer.get_extension("random_spikes").get_random_spikes()
 
         keep_unit_indices = np.flatnonzero(np.isin(self.sorting_analyzer.unit_ids, unit_ids))
         spikes = self.sorting_analyzer.sorting.to_spike_vector()
@@ -244,7 +244,7 @@ class ComputeWaveforms(AnalyzerExtension):
         unit_index = sorting.id_to_index(unit_id)
         # spikes = sorting.to_spike_vector()
         # some_spikes = spikes[self.sorting_analyzer.random_spikes_indices]
-        some_spikes = self.sorting_analyzer.get_extension("random_spikes").random_spikes()
+        some_spikes = self.sorting_analyzer.get_extension("random_spikes").get_random_spikes()
         spike_mask = some_spikes["unit_index"] == unit_index
         wfs = self.data["waveforms"][spike_mask, :, :]
 
@@ -340,7 +340,7 @@ class ComputeTemplates(AnalyzerExtension):
 
         # spikes = self.sorting_analyzer.sorting.to_spike_vector()
         # some_spikes = spikes[self.sorting_analyzer.random_spikes_indices]
-        some_spikes = self.sorting_analyzer.get_extension("random_spikes").random_spikes()
+        some_spikes = self.sorting_analyzer.get_extension("random_spikes").get_random_spikes()
         for unit_index, unit_id in enumerate(unit_ids):
             spike_mask = some_spikes["unit_index"] == unit_index
             wfs = waveforms[spike_mask, :, :]
@@ -504,7 +504,7 @@ class ComputeFastTemplates(AnalyzerExtension):
         unit_ids = sorting.unit_ids
 
         # retrieve spike vector and the sampling
-        some_spikes = self.sorting_analyzer.get_extension("random_spikes").random_spikes()
+        some_spikes = self.sorting_analyzer.get_extension("random_spikes").get_random_spikes()
 
         return_scaled = self.params["return_scaled"]
 
@@ -558,7 +558,7 @@ register_result_extension(ComputeFastTemplates)
 
 class ComputeNoiseLevels(AnalyzerExtension):
     """
-    Computes the noise level associated to each recording channel.
+    Computes the noise level associated with each recording channel.
 
     This function will wraps the `get_noise_levels(recording)` to make the noise levels persistent
     on disk (folder or zarr) as a `WaveformExtension`.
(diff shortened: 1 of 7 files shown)


src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -259,16 +269,25 @@ def _get_data(self):
...
-    Note that when "waveforms" is already done, then the recording is not needed anymore for this extension.
+    When the "waveforms" extension is already computed, then the recording is not needed anymore for this extension.
 
-    Note: by default only the average is computed. Other operator (std, median, percentile) can be computed on demand
+    Note: by default only the average and std are computed. Other operator (std, median, percentile) can be computed on demand
     after the SortingAnalyzer.compute("templates") and then the data dict is updated on demand.
 
+    Parameters
+    ----------
+    operators: list of str | list of (str, float)
```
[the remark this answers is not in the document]
alejoe91: what about `list[str] | list[(str, float)]`?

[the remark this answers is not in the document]
zm711: It's not classy python but what if we went a step further and did
`list[str] | list[(str, float)] (for percentile)`

h-mayorquin:
Instead of `random_spikes` what about `sample_spikes`?

In line with these three big boys:
https://numpy.org/doc/stable/reference/random/generated/numpy.random.sample.html
https://docs.python.org/3/library/random.html#random.sample
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sample.html

zm711:
> Instead of random_spikes what about sample_spikes?

What about the fact that sample can mean something very specific for ephys (ie the sample_rate)? 

Either name works for me to be honest, but just curious what you think. Does consistency with other libraries make the most sense or trying to avoid using the same word in multiple different ways?

h-mayorquin:
> What about the fact that sample can mean something very specific for ephys (ie the sample_rate)?

Mmm I always think on function names as verbs so this connection did not make sense to me. Also, don't we usually use `sampling` instead of `sample` for the other concept?

zm711:
I think that's a fair enough argument. Functions usually are verbs. (This library uses sampling, but in my head it is sample_rate--so maybe I'm just bringing my own experience to it :)

h-mayorquin:
You mght be right. I think I expressed myself wrongly about. I meant that I did not see the connection before. 

I like  `sample_spikes_indices `  because:
1) It is a verb (compared to random)
2) It communicate that it returns indices (not spikes, times or anything else).
3) The fact that is a sampling method makes naturally that we can also add sampling methods other than uniform as you see planned it here https://github.com/SpikeInterface/spikeinterface/blob/9d8f462e306c40a225a446c78cd45f89864b6472/src/spikeinterface/core/sorting_tools.py#L140

The last point is because people usually say random when they mean "uniform random distribution".

(In discrete cases people are indeed more careful and do say fair as in "consider a fair dice").

zm711:
I guess you're raising the good point that `random_spikes` (the public api) is relying on the fact that we use spike to mean the time (in samples or seconds) as well as the waveform, as well as the "object" that contains all of those attributes and in this case spike refers to the index to a "spike".

In reality `random_spikes` is a function that is doing `random_spike_indices` and then the private function is `sample_spike_indices` for any function that relies on these. But I think `random_spikes` is a better name even though it is less precise because at a public level we just want people to know they need to sample some spikes. 

So this makes me wonder if it is better to have api consistency (we chose `random_spikes` so we say `sample_random_spikes` or we choose the more descriptive function name and say `sample_spike_indices`.

> 3. The fact that is a sampling method makes naturally that we can also add sampling methods other than uniform as you see planned it here

Strong point. I think you've convinced me on the merit of "sample" in this case. But I think the API consistency case could still go either way (to be clear I mean `sample_random_spikes` vs `sample_spike_indices` in my mind).

h-mayorquin:
> But I think random_spikes is a better name even though it is less precise because at a public level we just want people to know they need to sample some spikes.

I think I am missing something on this sentence. How does `random_spikes` achieves this better than `sample_spike_indices`?

zm711:
I think for beginners in spike sorting they might actually struggle to understand the idea of spike_indices. 

To be clear I’m saying the public API is `sorting_analyzer.compute(‘random_spikes)` with the internal function for me would be `sample_random_spikes` because we are sampling this pool of random spikes that the user requested. 

Does that make sense? I agree that sample should be included in the name, but I think the public part of the api should be simple/even if using an overloaded term to make sense to beginning users.

h-mayorquin:
> Does that make sense? I agree that sample should be included in the name, but I think the public part of the api should be simple/even if using an overloaded term to make sense to beginning users.

OK, I think I get you. You are saying that, for beginners, being more explicit by including indices at the extension name might confuse more than clarify, correct? That could be...I need to think about it.

> To be clear I’m saying the public API is sorting_analyzer.compute(‘random_spikes)

You writing it like this is useful because the word compute is already acting as a verb when used like this. How does it  look as a parte of a bigger pipeline? If you do, `sorting_analyzer.get_data("random_spikes")`, what does it* return?

* I know this might not exist yet but I know something is on the works.

zm711:
> OK, I think I get you. You are saying that, for beginners, being more explicit by including indices at the extension name might confuse more than clarify, correct? That could be...I need to think about it

Exactly!

Yeah the current workflow would be something like

```python
# create analyzer...
random_spikes_ext = analyzer.compute('random_spikes', kwargs)
random_spikes = random_spike_ext.get_data()
# or what I actually do
analyzer.compute('random_spikes', kwargs)
random_spikes = analyzer.get_extension('random_spikes').get_data()
```
basically each Extension is created using the `compute`. so for example `analyzer.compute('spike_amplitudes')` or `analyzer.compute('quality_metrics')` which will return the `SortingAnalyzer.Extension`. And then there is a `get_data` that only works on that Extension. Sam was saying he wanted to add a `get_extension_data` to combine get_extension and get_data

so something like:

`random_spikes = analyzer.get_extension_data('random_spikes')`

So for another example

```python
analyzer.compute(['random_spikes', 'waveforms', 'templates', 'noise_levels', 'spike_amplitudes', 'spike_locations']).
```
This uses default arguments and just iterates through and calculates each one, but because I'm doing multiple it returns None instead so the only way to get the data is by `get_extension` + `get_data`. 

You could also do something like:
```python
analyzer.compute(dict(random_spikes=dict( max_number_of_spikes=300)), principal_components=dict(n_components=3)))
```

Basically compute is our verb and then we tell what (so the object/noun) that we want to compute.

src/spikeinterface/core/analyzer_extension_core.py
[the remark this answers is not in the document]
alejoe91: done!

alejoe91:
I agree with @zm711 that `spike_indices` is an "advanced" concept which will make the API over complicated. Fine to use it internally, but at the higher level `compute_random_spikes` or `sample_random_spikes` (it doesn't need to be "compute") would be a better choice.

@samuelgarcia what do you think of `sample_random_spikes` for the function? The extension will still be called `random_spikes`

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -80,7 +81,7 @@ def _select_extension_data(self, unit_ids):
     def _get_data(self):
         return self.data["random_spikes_indices"]
 
-    def some_spikes(self):
+    def random_spikes(self):
```
samuelgarcia, line 84: The name some_spikes was very fuzzy now this `random_spikes()` is becoming total cryptic.
This sounds like "make the randomization".
The idea of this function is "give and cache it the sub spikes vector that correspond to the sampling."
So the is function for me is even worst that the previous one which was already bad.

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -106,7 +107,7 @@ def get_selected_indices_in_spike_train(self, unit_id, segment_index):
         return selected_spikes_in_spike_train
 
 
-compute_select_random_spikes = SelectRandomSpikes.function_factory()
+compute_random_spikes = SelectRandomSpikes.function_factory()
```
samuelgarcia, line 110: OK

samuelgarcia: then we change the class name also no ?

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -134,19 +151,12 @@ def nafter(self):
...
-
         recording = self.sorting_analyzer.recording
         sorting = self.sorting_analyzer.sorting
         unit_ids = sorting.unit_ids
 
         # retrieve spike vector and the sampling
-        # spikes = sorting.to_spike_vector()
-        # some_spikes = spikes[random_spikes_indices]
-        some_spikes = self.sorting_analyzer.get_extension("random_spikes").some_spikes()
+        some_spikes = self.sorting_analyzer.get_extension("random_spikes").random_spikes()
```
samuelgarcia, line 159: weird.

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -78,8 +78,8 @@ def compute_similarity_with_templates_array(templates_array, other_templates_arr
 def compute_template_similarity_by_pair(sorting_analyzer_1, sorting_analyzer_2, method="cosine_similarity"):
     templates_array_1 = _get_dense_templates_array(sorting_analyzer_1, return_scaled=True)
     templates_array_2 = _get_dense_templates_array(sorting_analyzer_2, return_scaled=True)
-    similmarity = compute_similarity_with_templates_array(templates_array_1, templates_array_2, method)
-    return similmarity
+    similarity = compute_similarity_with_templates_array(templates_array_1, templates_array_2, method)
+    return similarity
```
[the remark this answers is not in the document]
samuelgarcia: but less poetic

src/spikeinterface/core/analyzer_extension_core.py
alejoe91: `some_spikes` is less weird?

samuelgarcia:
ciao camarade.
thanks for the doc update.
as you guess by the comments the `random_spike()` do not make me happy. lets find another better name

src/spikeinterface/core/analyzer_extension_core.py
alejoe91: What about `get_sampled_spikes`?

samuelgarcia: Maybe yes.

zm711: Or `sample_random_spikes` because you are "sampling" your pool of random spikes. I think this is what Heberto convinced me of. Unless we change the overall function to `sample_random_spikes`, but again my mental schema is that each class is an object (we don't call it "CalculateSpikeAmplitudes" we call it "SpikeAmplitudes" so the object is `RandomSpikes` for me and what we do is compute them. Then other functions sample those random spikes later. Again that's just my mental model of this.

samuelgarcia: Here this method is not making the sampling which is done in the   `_run()`.
This method is making the slice of the spike vector and caching it!!

zm711: In that case since this is just a normal extension shouldn't it just be the `get_data` then? So the whole api is consistent?

samuelgarcia: The get_data() give the indicies this method give the spike vector sliced by this indicies but cached because we acces this slice a lot

zm711: I finally (I hope) understand what's going on. In that case this is really
`get_cached_sampled_spikes` or `get_cached_random_spikes`.  I think saying this is going to the cached spikes is clearer because if we recompute the random spikes we should remake the cache so it should be clear that we are grabbing from the cache and that no additional calculation should be occurring.

alejoe91: What about just `get_sampled_spikes()`? Caching is something we do internally to speed things up, so I don't think it makes sense to have it in the name

zm711: I think I like that name (get_sampled_spikes) the best, but would you then want to change the class/extension to `SampledSpikes` to be consistent? Again for me at least `RandomSpikes` has an intuitive flavor even if "sampling" is the more rigorous term.

alejoe91: I like random too. I'll co for `get_random_spikes()` then for consistency

samuelgarcia (approved):

PR #2610, opened 2024-03-21
Implement `get_parent()` function

### LINKED ISSUE #2596: Is it possible to access original data through preprocesed `BaseRecording` instance
The pipeline
```
Binary File -> Preprocessing -(1)-> Sorting -(2)-> Wave Extractor -(3)-> Neuron feature assessment. 
```
The last class gets `BaseRecording` after (1) and uses `get_traces` to compute a spike "fingerprint". Unfortunately, preprocessing removes something we are looking for. Would it be possible to `get_traces` from the original Binary File without reloading it and creating an additional `BaseRecording` object?

In other words, is there something like that
```python
# preproc is BaseRecording (1)
# which is pointing to the filtered and zeroed data
fingerprint = array([
        preproc.original_data.get_traces(
            start_frame   = sp+frmstart,
            end_frame     = sp+frmend).T
        for sp in sp_train
        ])
```

alejoe91:
Fixes #2596

This is super useful for example to plot different processing layers!

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -82,6 +82,10 @@ class BaseExtractor:
         # This is implemented in BaseRecording or BaseSorting
         raise NotImplementedError
 
+    def get_parent(self) -> Optional[BaseExtractor]:
+        """Returns parent object if it exists, otherwise None"""
+        return getattr(self, "_parent", None)
+
     def _check_segment_index(self, segment_index: Optional[int] = None) -> int:
         if segment_index is None:
             if self.get_num_segments() == 1:
--- a/src/spikeinterface/preprocessing/zero_channel_pad.py
+++ b/src/spikeinterface/preprocessing/zero_channel_pad.py
@@ -188,6 +188,7 @@ class ZeroChannelPaddedRecording(BaseRecording):
 
         # only copy relevant metadata and properties
         parent_recording.copy_metadata(self, only_main=True)
+        self._parent = parent_recording
         prop_keys = parent_recording.get_property_keys()
 
         for k in prop_keys:
--- a/src/spikeinterface/core/channelsaggregationrecording.py
+++ b/src/spikeinterface/core/channelsaggregationrecording.py
@@ -93,6 +93,10 @@ class ChannelsAggregationRecording(BaseRecording):
         self._recordings = recording_list
         self._kwargs = {"recording_list": [rec for rec in recording_list], "renamed_channel_ids": renamed_channel_ids}
 
+    @property
+    def recordings(self):
+        return self._recordings
+
 
 class ChannelsAggregationRecordingSegment(BaseRecordingSegment):
     """
--- a/src/spikeinterface/core/channelslice.py
+++ b/src/spikeinterface/core/channelslice.py
@@ -60,6 +60,7 @@ class ChannelSliceRecording(BaseRecording):
 
         # copy annotation and properties
         parent_recording.copy_metadata(self, only_main=False, ids=self._channel_ids)
+        self._parent = parent_recording
 
         # change the wiring of the probe
         contact_vector = self.get_property("contact_vector")
--- a/src/spikeinterface/core/frameslicerecording.py
+++ b/src/spikeinterface/core/frameslicerecording.py
@@ -59,6 +59,7 @@ class FrameSliceRecording(BaseRecording):
 
         # copy properties and annotations
         parent_recording.copy_metadata(self)
+        self._parent = parent_recording
 
         # update dump dict
         self._kwargs = {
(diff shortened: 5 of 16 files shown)


DradeAW:
Quick (maybe dumb) question: what about when there are multiple parents? (e.g. `UnitsAggregationSorting`)

src/spikeinterface/preprocessing/zero_channel_pad.py
h-mayorquin, line 1: Interesting that this is not a BaseProcessing object : O

alejoe91:
> Quick (maybe dumb) question: what about when there are multiple parents? (e.g. `UnitsAggregationSorting`)

It returns `None`, but we have other ways to return the parent list (`property sortings/recordings`)

src/spikeinterface/preprocessing/zero_channel_pad.py
alejoe91: it cannot be because it changes the number of channels!

h-mayorquin: Ah, I did not know that was a constrain of the object. I guess that applies for other `post-processing-like` objects that we have on sort like channel_slice, etc : O

h-mayorquin (commented):
This looks OK to me.  I also like, on a first thought, the semantics of returning None which can be used  as a check.

Thinking if it is missing from somewhere... but I don't think so. Do we need to do anything to add it to the API?  

Two points, only the first one is relevant and the second one is a suggestion,

1) For the segments we have the weakref mechanism. Would it make sense to have smoething like this for this? Does the context were the other decision was made makes sense her? Probably for @samuelgarcia to weight in on this one.
2) I would go fully verbose and say `get_parent_extractor` and `_parent_extractor`.

zm711:
>Does the context were the other decision was made makes sense her? 

I believe that was largely done to allow mapping between network drives for Windows because the cleanup etc doesn't work on Windows otherwise (due to WinError 32 amongst other classic Windows issues. ) . Neo did have a discussion of circular non-closeable references though that may be of interest in relation to this [here](https://github.com/NeuralEnsemble/python-neo/issues/684)

alejoe91:
> 1. For the segments we have the weakref mechanism. Would it make sense to have smoething like this for this? Does the context were the other decision was made makes sense her? Probably for @samuelgarcia to weight in on this one.

I don't think it does, because in this case you're exposing an object which is already referenced (`self._parent_recording`). Instead, when registering a recording to a sorting, you're "attaching" a complete different recording.

> 2. I would go fully verbose and say `get_parent_extractor` and `_parent_extractor`.

I actually don't really like extractor :) I know that we have a `BaseExtractor` class, but the core objects are `BaseRecording` adn `BaseSorting`

alejoe91:
> > 1. For the segments we have the weakref mechanism. Would it make sense to have smoething like this for this? Does the context were the other decision was made makes sense her? Probably for @samuelgarcia to weight in on this one.
> 
> I don't think it does, because in this case you're exposing an object which is already referenced (`self._parent_recording`). Instead, when registering a recording to a sorting, you're "attaching" a complete different recording.
> 
> > 2. I would go fully verbose and say `get_parent_extractor` and `_parent_extractor`.
> 
> I actually don't really like extractor :) I know that we have a `BaseExtractor` class, but the core objects are `BaseRecording` and `BaseSorting`

samuelgarcia:
I have the intuition that we should make a weak reference instance of a hard link.
Lets maturate a bit this before merging.

alejoe91:
> I have the intuition that we should make a weak reference instance of a hard link. Lets maturate a bit this before merging.

I think it doesn't make sense, since the parent object is also usually stored as `self._something`

src/spikeinterface/core/base.py
```diff
@@ -82,6 +82,10 @@ def get_num_segments(self) -> int:
         # This is implemented in BaseRecording or BaseSorting
         raise NotImplementedError
 
+    def get_parent(self) -> Optional[BaseExtractor]:
+        """Returns parent object if it exists, otherwise None"""
```
h-mayorquin, line 86: @alejoe91 
I see, I vaguely remember that Sam does not like the word extractor either.

```suggestion
        """Returns the parent recording or sorter object if it exists, otherwise None.
        """
```

I have two suggestions to make this more specific less ambiguios: 
1) We can enchance the docstring here  as in my suggestion to disambiguate from parent class in the OOP sense. Maybe adding an example with recording and one with sorting.
2) Make this method private and then have a `parent_recording` and a `parent_sorting` that use `_get_parent` (maybe they could be attributes).

h-mayorquin:
> I don't think it does, because in this case you're exposing an object which is already referenced (self._parent_recording). Instead, when registering a recording to a sorting, you're "attaching" a complete different recording.

I see. Yeah, then maybe let's do it as you have it here and then, if some problem appears, we could add weakref?

zm711:
Since Windows is the biggest culprit for problems is there a test I can run locally to probe this?

alejoe91:
> Since Windows is the biggest culprit for problems is there a test I can run locally to probe this?

I don't think there is! IMO let's merge and see if it causes problems

samuelgarcia (approved):

PR #2611, opened 2024-03-21
Seeding silence periods

yger:
This PR adds the seeding property of the NoiseGeneratorRecording to the remove_silence function. To be more precise, I slightly extended the possibility of the NoiseGeneratorRecording to handle multiple noise_levels, such that internally, the remove_silence can make use of it to fill the portions of the signals where noise should be injected. By doing so, the noise is properly seeded and we are avoiding code duplication.


yger:
Lots of file are changed, but this is only to extend the syntax of the NoiseGeneratorRecording, with noise_levelS instead of noise_level. This makes more sense to me I think

src/spikeinterface/core/generate.py
```diff
@@ -984,8 +984,8 @@ class NoiseGeneratorRecording(BaseRecording):
         The sampling frequency of the recorder.
     durations : List[float]
         The durations of each segment in seconds. Note that the length of this list is the number of segments.
-    noise_level: float, default: 1
-        Std of the white noise
+    noise_levels: float or array, default: 1
+        Std of the white noise (if an array, defined by per channels)
```
alejoe91, line 988: Can you add a test for this new behavior?

src/spikeinterface/preprocessing/silence_periods.py
```diff
@@ -71,28 +72,39 @@ def __init__(self, recording, list_periods, mode="zeros", noise_levels=None, **r
...
+            noise_generator = NoiseGeneratorRecording(
+                num_channels=recording.get_num_channels(),
+                sampling_frequency=recording.sampling_frequency,
+                durations=[recording.select_segments(i).get_duration() for i in range(recording.get_num_segments())],
+                dtype=recording.dtype,
+                seed=seed,
+                noise_levels=noise_levels,
+                strategy="on_the_fly",
+                noise_block_size=int(recording.sampling_frequency),
+            )
```
alejoe91, line 86: same here. Can you add a test in `test_silence`?

src/spikeinterface/core/generate.py
yger: A test has been added in test_generate.py

src/spikeinterface/preprocessing/silence_periods.py
yger: A small test checking that noise is properly seeded has been added

alejoe91 (approved):

src/spikeinterface/core/generate.py
```diff
@@ -1010,18 +1010,24 @@ def __init__(
...
         channel_ids = np.arange(num_channels)
         dtype = np.dtype(dtype).name  # Cast to string for serialization
         if dtype not in ("float32", "float64"):
             raise ValueError(f"'dtype' must be 'float32' or 'float64' but is {dtype}")
         assert strategy in ("tile_pregenerated", "on_the_fly"), "'strategy' must be 'tile_pregenerated' or 'on_the_fly'"
 
+        if np.isscalar(noise_levels):
+            noise_levels = np.ones((1, num_channels)) * noise_levels
+        else:
+            noise_levels = np.asarray(noise_levels)
```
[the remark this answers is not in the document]
yger: I did it

samuelgarcia (approved):

PR #2616, opened 2024-03-22
SortingAnalyzer: `get_extension_default_params`

alejoe91:
and `get_computable_extensions()`

@samuelgarcia @h-mayorquin @zm711 couldn't come up with a better name for the latter. Any suggestion?

At one point, we will make the switch and define proper pydantic models (at least for extension, sorters, quality metrics, template metrics params)

**NOTE:** I changed the `unit_locations` `**method_kwargs` to a `method_kwargs` dict, which is better for inspection

### DIFF
--- a/src/spikeinterface/core/__init__.py
+++ b/src/spikeinterface/core/__init__.py
@@ -144,7 +144,15 @@ from .sparsity import ChannelSparsity, compute_sparsity, estimate_sparsity
 from .template import Templates
 
 # SortingAnalyzer and AnalyzerExtension
-from .sortinganalyzer import SortingAnalyzer, AnalyzerExtension, create_sorting_analyzer, load_sorting_analyzer
+from .sortinganalyzer import (
+    SortingAnalyzer,
+    AnalyzerExtension,
+    create_sorting_analyzer,
+    load_sorting_analyzer,
+    get_available_analyzer_extensions,
+    get_default_anlyzer_extension_params,
+)
+
 from .analyzer_extension_core import (
     SelectRandomSpikes,
     compute_select_random_spikes,
--- a/src/spikeinterface/core/__init__.py
+++ b/src/spikeinterface/core/__init__.py
@@ -150,7 +150,7 @@ from .sortinganalyzer import (
     create_sorting_analyzer,
     load_sorting_analyzer,
     get_available_analyzer_extensions,
-    get_default_anlyzer_extension_params,
+    get_default_analyzer_extension_params,
 )
 
 from .analyzer_extension_core import (
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -1025,6 +1025,37 @@ class SortingAnalyzer:
         else:
             return False
 
+    def get_computable_extensions(self):
+        """
+        Get all extensions that can be computed by the analyzer.
+        """
+        return list(_builtin_extensions.keys())
+
+    def get_default_extension_params(self, extension_name: str):
+        """
+        Get the default params for an extension.
+
+        Parameters
+        ----------
+        extension_name: str
+            The extension name
+
+        Returns
+        -------
+        default_params: dict
+            The default parameters for the extension
+        """
+        import inspect
+
+        extension_class = get_extension_class(extension_name)
+
+        sig = inspect.signature(extension_class._set_params)
+        default_params = {
+            k: v.default for k, v in sig.parameters.items() if k != "self" and v.default != inspect.Parameter.empty
+        }
+
+        return default_params
+
 
 global _possible_extensions
 _possible_extensions = []
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -1045,7 +1045,7 @@ class SortingAnalyzer:
         default_params: dict
             The default parameters for the extension
         """
-        return get_default_anlyzer_extension_params(extension_name)
+        return get_default_analyzer_extension_params(extension_name)
 
 
 global _possible_extensions
@@ -1120,7 +1120,7 @@ def get_available_analyzer_extensions():
     return list(_builtin_extensions.keys())
 
 
-def get_default_anlyzer_extension_params(extension_name: str):
+def get_default_analyzer_extension_params(extension_name: str):
     """
     Get the default params for an extension.
 
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -1029,7 +1029,7 @@ class SortingAnalyzer:
         """
         Get all extensions that can be computed by the analyzer.
         """
-        return list(_builtin_extensions.keys())
+        return get_available_analyzer_extensions()
 
     def get_default_extension_params(self, extension_name: str):
         """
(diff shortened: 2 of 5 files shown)


zm711:
> and get_computable_extensions()

I'm not opposed to `get_computable_extensions`. Since the main syntax of the analyzer is `analyzer.compute(...)` having a get what is computable makes sense. `computable` is not the most discoverable of words, though so I can understand not liking that. I think the more common wording would be `get_possible_extensions` in that because they can be computed they are possible. But we already use `possible_extensions` elsewhere so I don't know if we want to overload.

So for me it comes back to choose an easy intuitive word for beginning users (possible) or choose a more precise accurate word for advanced/dev users (computable). Interested in what the others think!

h-mayorquin (approved):
This looks OK to me.
It is a verb, it is specific and it does what is says. But for some suggestions:

"get _possible_extenssions" works as well the same as  "get_available_extensions". It could also be an attribute instead of a getter. 

Random suggestions that could be useful just because I thought about them:
* What about an attribute?
* Should it be a static method so the user can query which extensions are available without need to instantiate the class? (mabye there is another way to do this that I just don't know about?)
* In the same vein as the last, should there be an attribute that can be imported from top level called "available_sorting_analyzer_extensions`?

Not very well though off, just throwing some random suggetions because sometimes is useful.

alejoe91:
I like the idea of moving the two functions at the module level, as one might want to explore extensions and params before instantiating the analyzer.

What do you think abuot `si.get_available_analyzer_extensions()` and `si.get_analyzer_extension_params()` (I would drop the default here)?

zm711:
> and si.get_analyzer_extension_params() (I would drop the default here)?

For this one you would take in the `extension_name` and return params for just one? I think returning all default params would be confusing, but I like the idea of allowing someone to easily grab the defaults for an extension at a time.

alejoe91:
> > and si.get_analyzer_extension_params() (I would drop the default here)?
> 
> For this one you would take in the `extension_name` and return params for just one? I think returning all default params would be confusing, but I like the idea of allowing someone to easily grab the defaults for an extension at a time.

Yep, same as It Is now

zm711:
> What do you think abuot si.get_available_analyzer_extensions() and si.get_analyzer_extension_params() (I would drop the default here)?

Then this sounds good to me :)

alejoe91:
Done!

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1091,6 +1113,39 @@ def get_extension_class(extension_name: str, auto_import=True):
...
 
 
+def get_available_analyzer_extensions():
+    """
+    Get all extensions that can be computed by the analyzer.
+    """
+    return list(_builtin_extensions.keys())
+
+
+def get_default_anlyzer_extension_params(extension_name: str):
```
zm711, line 1123:
```suggestion
def get_default_analyzer_extension_params(extension_name: str):
```

src/spikeinterface/core/__init__.py
```diff
@@ -144,7 +144,15 @@
...
 
 # SortingAnalyzer and AnalyzerExtension
-from .sortinganalyzer import SortingAnalyzer, AnalyzerExtension, create_sorting_analyzer, load_sorting_analyzer
+from .sortinganalyzer import (
+    SortingAnalyzer,
+    AnalyzerExtension,
+    create_sorting_analyzer,
+    load_sorting_analyzer,
+    get_available_analyzer_extensions,
+    get_default_anlyzer_extension_params,
```
zm711, line 153:
```suggestion
    get_default_analyzer_extension_params,
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1025,6 +1025,28 @@ def has_extension(self, extension_name: str) -> bool:
...
+        ----------
+        extension_name: str
+            The extension name
+
+        Returns
+        -------
+        default_params: dict
+            The default parameters for the extension
+        """
+        return get_default_anlyzer_extension_params(extension_name)
```
zm711, line 1048:
```suggestion
        return get_default_analyzer_extension_params(extension_name)
```

src/spikeinterface/core/tests/test_sortinganalyzer.py
```diff
@@ -143,6 +148,24 @@ def _check_sorting_analyzers(sorting_analyzer, original_sorting):
...
 
+def test_extension_params():
+    from spikeinterface.core.sortinganalyzer import _builtin_extensions
+
+    computable_extension = get_available_analyzer_extensions()
+
+    for ext, mod in _builtin_extensions.items():
+        assert ext in computable_extension
+        if mod == "spikeinterface.core":
+            default_params = get_default_anlyzer_extension_params(ext)
```
zm711, line 159:
```suggestion
            default_params = get_default_analyzer_extension_params(ext)
```

src/spikeinterface/core/tests/test_sortinganalyzer.py
```diff
@@ -143,6 +148,24 @@ def _check_sorting_analyzers(sorting_analyzer, original_sorting):
...
+            print(ext, default_params)
+        else:
+            try:
+                default_params = get_default_anlyzer_extension_params(ext)
```
zm711, line 163:
```suggestion
                default_params = get_default_analyzer_extension_params(ext)
```

zm711 (commented):
I think I got them all :) but anlyzer -> analyzer.

zm711:
Based on failing tests I must've missed one sorry!

src/spikeinterface/core/tests/test_sortinganalyzer.py
```diff
@@ -4,7 +4,12 @@
 import shutil
 
 from spikeinterface.core import generate_ground_truth_recording
-from spikeinterface.core import SortingAnalyzer, create_sorting_analyzer, load_sorting_analyzer
+from spikeinterface.core import (
+    create_sorting_analyzer,
+    load_sorting_analyzer,
+    get_available_analyzer_extensions,
+    get_default_anlyzer_extension_params,
```
zm711, line 11:
```suggestion
    get_default_analyzer_extension_params,
```

src/spikeinterface/postprocessing/unit_localization.py
```diff
@@ -59,7 +57,7 @@ class ComputeUnitLocations(AnalyzerExtension):
     def __init__(self, sorting_analyzer):
         AnalyzerExtension.__init__(self, sorting_analyzer)
 
-    def _set_params(self, method="monopolar_triangulation", **method_kwargs):
+    def _set_params(self, method="monopolar_triangulation", method_kwargs={}):
```
samuelgarcia, line 60: I am not agree with this.

samuelgarcia: Can we revert this ?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1091,6 +1113,39 @@ def get_extension_class(extension_name: str, auto_import=True):
     return ext_class
 
 
+def get_available_analyzer_extensions():
+    """
+    Get all extensions that can be computed by the analyzer.
+    """
+    return list(_builtin_extensions.keys())
```
samuelgarcia, line 1120: This should this + the one that are register. We can have extension outside!

alejoe91:
> I am OK with the new method but I am not agree with the signature change inunit_locaization. This make non backward compatible witg previous versions. We should have discuss this before this super fast merge, I think. No ?

That unifies the params with the spike locations. In general, having **method_kwargs in the set params is bad IMO.

Not that it doesn't really break back-compatibility. For results that are already saved as a `WaveformExtractor`, the params don't do anything since we can't recompute stuff with it.

PR #2617, opened 2024-03-22
IBL extractors revived

alejoe91:
Following #2512 

Implements and updates IBL extractors (streaming):

- Use IBL-like semantics (`eid`, `pid`, etc.). 
- ALFextractor uses `samples` instead of `times`

@oliche can you take a look?


CodyCBakerPhD:
> Removed IblStreamingRecordingExtractor in favor of IblRecordingExtractor (@CodyCBakerPhD any changes on neuroconv required?)

Nothing was ever added to core NeuroConv since it was such a specific usage for the first IBL conversion

src/spikeinterface/extractors/iblextractors.py
```diff
@@ -0,0 +1,355 @@
...
+        self.add_recording_segment(recording_segment)
+
+        self._kwargs = {
+            "eid": eid,
+            "pid": pid,
+            "stream_name": stream_name,
+            "load_sync_channel": load_sync_channel,
+            "cache_folder": cache_folder,
+            "remove_cached": remove_cached,
+            "stream": stream,
```
alejoe91, line 258: `one` is not cached because not dumpable!

src/spikeinterface/extractors/iblextractors.py
```diff
@@ -0,0 +1,355 @@
...
+    -------
+    extractor : IBLSortingExtractor
+        The loaded data.
+    """
+
+    extractor_name = "IBLSorting"
+    name = "ibl"
+    installation_mesg = "IBL extractors require ibllib as a dependency." " To install, run: \n\n pip install ibllib\n\n"
+
+    def __init__(self, pid, good_clusters_only=False, load_unit_properties=True, one=None):
```
alejoe91, line 316: @h-mayorquin added the `good_clusters_only` to only get units passing IBL QC. Also, with `load_unit_properties=True` the brain location is loaded as the `acronym` property. 

We should use `good_clusters_only=True` to build a first round of _potentially_ high-quality templates

pyproject.toml
```diff
@@ -65,9 +65,8 @@ extractors = [
     "sonpy;python_version<'3.10'",
     "lxml", # lxml for neuroscope
     "scipy",
-    # ONE-api and ibllib for streaming IBL
-    "ONE-api>=1.19.1",
-    "ibllib>=2.21.0",
+    "ONE-api>=1.19.1", # alf sorter and streaming IBL
```
h-mayorquin, line 68: Let's upgrade this to their latest versions. I will push this if you don't mind.

src/spikeinterface/extractors/alfsortingextractor.py
h-mayorquin, line 1: This is a very good reduction on lines of code. Great!

src/spikeinterface/extractors/cbin_ibl.py
```diff
@@ -117,6 +121,7 @@ def __init__(self, folder_path, load_sync_channel=False, stream_name="ap"):
         self._kwargs = {
             "folder_path": str(Path(folder_path).absolute()),
             "load_sync_channel": load_sync_channel,
+            "cbin_file": str(Path(cbin_file).absolute()),
```
h-mayorquin, line 124: resolve is better than absolute but probably does not matter. Resolve gets symbolic links and makes the path absolute.

src/spikeinterface/extractors/extractorlist.py
```diff
@@ -29,7 +29,7 @@
 from .nwbextractors import NwbRecordingExtractor, NwbSortingExtractor, read_nwb, read_nwb_recording, read_nwb_sorting
 
 from .cbin_ibl import CompressedBinaryIblExtractor, read_cbin_ibl
-from .iblstreamingrecording import IblStreamingRecordingExtractor, read_ibl_streaming_recording
+from .iblextractors import IblRecordingExtractor, IblSortingExtractor, read_ibl_recording, read_ibl_sorting
```
h-mayorquin, line 32: Maybe we should keep the old with a deprecation warning?

h-mayorquin (commented):
This looks good to me. I think that maybe we should keep the old class for a bit with a deprecation warning, what do you think @alejoe91 ?

Also, I will update the requirements.

src/spikeinterface/extractors/iblextractors.py
h-mayorquin: One question here: Could we store the brain_location as a property named brain_location instead of "acronym"?

src/spikeinterface/extractors/extractorlist.py
alejoe91: I think it was only used internally and for testing. Fine to remove :)

src/spikeinterface/extractors/cbin_ibl.py
alejoe91: done

src/spikeinterface/extractors/iblextractors.py
alejoe91: yup, pushed a patch to use `brain_area`

alejoe91:
@h-mayorquin ready to merge on my end ;)

h-mayorquin (approved):
All right. I think this is ready as well!

Issue #1871, opened 2023-07-21
Error running run_sorter in py script from command prompt


oghenand:
Hi! 

I'm trying to run Ironclust using Spike Interface on a .py file that I want to directly run from the command prompt. However, when I try to do so, I get an odd error likely due to saving of the data to the output folder. When I run the code from jupyter notebook, it runs completely fine, but the error seems to stem from using a .py file. I've attached the output of my error message below. If anyone has encountered a similar error before and would know a fix, that'd be great, thanks!

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\ProgramData\miniconda3\lib\multiprocessing\spawn.py", line 116, in spawn_main
    exitcode = _main(fd, parent_sentinel)
  File "C:\ProgramData\miniconda3\lib\multiprocessing\spawn.py", line 125, in _main
    prepare(preparation_data)
  File "C:\ProgramData\miniconda3\lib\multiprocessing\spawn.py", line 236, in prepare
    _fixup_main_from_path(data['init_main_from_path'])
  File "C:\ProgramData\miniconda3\lib\multiprocessing\spawn.py", line 287, in _fixup_main_from_path
    main_content = runpy.run_path(main_path,
  File "C:\ProgramData\miniconda3\lib\runpy.py", line 289, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "C:\ProgramData\miniconda3\lib\runpy.py", line 96, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "C:\ProgramData\miniconda3\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\Users\Allison\source\repos\Spike_Sorting_Automation_1\Spike_Sorting_Automation_1\Spike_Sorting_Automation_1.py", line 149, in <module>
    IC_si = ss.run_sorter('ironclust', recording = recording_prb, output_folder = r'C:\Users\Allison\Desktop\auto_pilot_721', verbose=False)
  File "C:\Users\Allison\AppData\Roaming\Python\Python310\site-packages\spikeinterface\sorters\runsorter.py", line 143, in run_sorter
    return run_sorter_local(**common_kwargs)
  File "C:\Users\Allison\AppData\Roaming\Python\Python310\site-packages\spikeinterface\sorters\runsorter.py", line 158, in run_sorter_local
    output_folder = SorterClass.initialize_folder(
[...]

PR #2621, opened 2024-03-22
Avoid `np.prod` in `make_shared_array`

h-mayorquin:
np.prod produces a numpy scalars (not integers or floats!). 

Numpy scalars behave like numpy values and can overflow. Example:

```python
In [1]: a_billion_int = 1_000_000_000

In [2]: np.prod((a_billion_int, a_billion_int, a_billion_int, a_billion_int))
Out[2]: -5527149226598858752
```

Python integers on the other hand do not overflow. 

This PR changes this so overflow errors do not happen.

This error appear to a user here:
https://github.com/SpikeInterface/spikeinterface/issues/1871#issuecomment-1947957281

### DIFF
--- a/src/spikeinterface/core/core_tools.py
+++ b/src/spikeinterface/core/core_tools.py
@@ -163,6 +163,7 @@ def make_shared_array(shape, dtype):
     from multiprocessing.shared_memory import SharedMemory
 
     dtype = np.dtype(dtype)
+    shape = (int(x) for x in shape)  # We need to be sure that shape comes in int and not number scalars
     nbytes = prod(shape) * dtype.itemsize
     shm = SharedMemory(name=None, create=True, size=nbytes)
     arr = np.ndarray(shape=shape, dtype=dtype, buffer=shm.buf)
--- a/src/spikeinterface/core/core_tools.py
+++ b/src/spikeinterface/core/core_tools.py
@@ -163,7 +163,7 @@ def make_shared_array(shape, dtype):
     from multiprocessing.shared_memory import SharedMemory
 
     dtype = np.dtype(dtype)
-    shape = tuple(int(x) for x in shape)  # We need to be sure that shape comes in int and numpy number scalars
+    shape = tuple(int(x) for x in shape)  # We need to be sure that shape comes in int and numpy scalars
     nbytes = prod(shape) * dtype.itemsize
     shm = SharedMemory(name=None, create=True, size=nbytes)
     arr = np.ndarray(shape=shape, dtype=dtype, buffer=shm.buf)
--- a/src/spikeinterface/core/core_tools.py
+++ b/src/spikeinterface/core/core_tools.py
@@ -163,7 +163,7 @@ def make_shared_array(shape, dtype):
     from multiprocessing.shared_memory import SharedMemory
 
     dtype = np.dtype(dtype)
-    shape = (int(x) for x in shape)  # We need to be sure that shape comes in int and not number scalars
+    shape = tuple(int(x) for x in shape)  # We need to be sure that shape comes in int and not number scalars
     nbytes = prod(shape) * dtype.itemsize
     shm = SharedMemory(name=None, create=True, size=nbytes)
     arr = np.ndarray(shape=shape, dtype=dtype, buffer=shm.buf)
--- a/src/spikeinterface/core/core_tools.py
+++ b/src/spikeinterface/core/core_tools.py
@@ -163,7 +163,7 @@ def make_shared_array(shape, dtype):
     from multiprocessing.shared_memory import SharedMemory
 
     dtype = np.dtype(dtype)
-    shape = tuple(int(x) for x in shape)  # We need to be sure that shape comes in int and numpy scalars
+    shape = tuple(int(x) for x in shape)  # We need to be sure that shape comes in int instead of numpy scalars
     nbytes = prod(shape) * dtype.itemsize
     shm = SharedMemory(name=None, create=True, size=nbytes)
     arr = np.ndarray(shape=shape, dtype=dtype, buffer=shm.buf)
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/core_tools.py
```diff
@@ -163,7 +163,7 @@ def make_shared_array(shape, dtype):
     from multiprocessing.shared_memory import SharedMemory
 
     dtype = np.dtype(dtype)
-    nbytes = int(np.prod(shape) * dtype.itemsize)
+    nbytes = prod(shape) * dtype.itemsize
```
zm711, line 166: This still needs to be wrapped as an int no?

zm711: Related to Windows failing.

h-mayorquin: itemsize is int, shape is int, that's unecessary. Am I wrong?

zm711: When I check they are coming up as ints, but based on the fact the tests are changing just for Windows and all you did was remove the int over the operation maybe something was being converted.... Not sure.

h-mayorquin: The thing that converts is the `np.prod`. because of an overflow. Let me show you:

```python
In [1]: a_billion_int = 1_000_000_000

In [2]: np.prod((a_billion_int, a_billion_int, a_billion_int, a_billion_int))
Out[2]: -5527149226598858752
```

zm711: Agreed. np.prod was causing an overflow. Absolutely agree, but now tests are failing.

So why do you thing nbytes are coming up as np.int for Windows testing only?

h-mayorquin: I have no idea, my friend. There must be a casting somewhere else.

h-mayorquin: Yes, so TO MY SURPRISE `np.prod` does preserve casting:

```
In [3]: from math import prod

In [4]: x = np.arange(9, dtype="int")[8]

In [5]: prod((x, x))
Out[5]: 64

In [6]: type(prod((x, x)))
Out[6]: numpy.int64
```

(The beauty and dangers of duck-type)!

To realy solve the bug we need to figure out which shape comes as numpy scalars. Otherwise, even this solution will break at some ponint.

zm711: That's wild.

h-mayorquin: Which points to the solution. There is a place where a numpy functions were used to determine the number of channels so we ended up with a numpy scalar on shape. I changed that but I guess the most robus solution is to force casting **inside** the shape. That avoids the error from the root whereas casting the final product (already overflown and hence negative) would not.

Does this makes sense?

src/spikeinterface/core/core_tools.py
```diff
@@ -163,7 +163,8 @@ def make_shared_array(shape, dtype):
     from multiprocessing.shared_memory import SharedMemory
 
     dtype = np.dtype(dtype)
-    nbytes = int(np.prod(shape) * dtype.itemsize)
+    shape = (int(x) for x in shape)  # We need to be sure that shape comes in int and not number scalars
```
zm711:
```suggestion
    shape = tuple(int(x) for x in shape)  # We need to be sure that shape comes in int and not numy scalars
```

zm711: Cool you saw it :)

h-mayorquin: Yes, thanks, I realized it just now. 

You know this generator notation has fucked me up a couple of times, check this out:

```python
is_one_equal_to_0 = np.all((1 == 0 for _ in range(10)))
if is_one_equal_to_0:
    print("This is great, 1 is equal to 0")
``` 

Syntactic suggar is nice until it is not.

zm711: That's a cool example.  The tuple/generator for () and (even sometimes I forget)  {} for set in addition to dictionary. Those mess with me. But yeah the accident generator is way too often.

src/spikeinterface/core/core_tools.py
h-mayorquin: I considered the problem solved but I still don't know why the first iteraiton of the solution did fail on windows and not on unbutu : O

zm711: It's a good thing it did fail because it found another potential point of overflow. I know numpy optimization is different between different OSs so maybe that did it on Windows? But I don't think the different OSs should return different types.... Another one of the great mysteries.

zm711: Windows and linux numpy handle overflow differently. [See this](https://github.com/numpy/numpy/issues/8433). Based on the discussion the behavior will likely be changed in Numpy 2.0, but in short
```
np.full((2, 2), np.iinfo(np.int32).max, dtype=np.int32).sum(axis=0)
shows similar behavior: array([4294967294, 4294967294]) on ubuntu and array([-2, -2]) on windows.
```
> However, this function returns an array instead of a scalar, so we can inspect the dtype. On Ubuntu the dtype is int64, but on Windows the dtype is int32.

So maybe this was due to the difference in how certain cases of overflow occur.

h-mayorquin: Thanks a bunch for sharing the issue!

h-mayorquin: So numpy 2.0 should align this.

zm711: Anything that makes the debugging easier among the operating systems. Much less surface area if one fixes works for all of them. :) (although the final message wasn't completely confident sounding...)

alejoe91 (approved):

samuelgarcia:
bien joué!

PR #2630, opened 2024-03-26
Add supportfor `IntanRecordingExtractor` "one-file-per-signal" and "one-file-per-channel" formats

h-mayorquin:
As stated on the title. This is possible because we merged the following neo:

https://github.com/NeuralEnsemble/python-neo/pull/1402

As this translated cleanly the only thing that I am adding here are the tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "spikeinterface"
-version = "0.101.0"
+version = "0.101.0.dev0"
 authors = [
   { name="Alessio Buccino", email="alessiop.buccino@gmail.com" },
   { name="Samuel Garcia", email="sam.garcia.die@gmail.com" },
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "spikeinterface"
-version = "0.101.0.dev0"
+version = "0.101.0"
 authors = [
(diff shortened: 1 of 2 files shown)


h-mayorquin:
Interesting this fails here. It works locally for me but it is hard to know which of the tests is actually failing because of the structure of the tests with entities.

h-mayorquin:
Ok, the error was solved. The reason was that the tests were not using the latests versions of neo and probeinterface. That is, it seems that these lines are not working for the full tests:

https://github.com/SpikeInterface/spikeinterface/blob/2983463b655a6f10130b8017913f2ca90d4fd9fe/pyproject.toml#L153-L156

Probably something cache related. It is good for you to be aware of this @samuelgarcia @alejoe91

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "spikeinterface"
-version = "0.101.0"
+version = "0.101.0.dev0"
```
alejoe91, line 3: Can you revert this? See other PR 
```suggestion
version = "0.101.0"
```

alejoe91 (changes requested):
Thanks @h-mayorquin 

Looks good, except for the change in version

alejoe91 (approved):

alejoe91:
> Thanks @h-mayorquin
> 
> Looks good, except for the change in version

Changed

PR #2631, opened 2024-03-26
Set version to dev in pyproject for 0.101.0

h-mayorquin:
Shouldn't this be like this?


alejoe91:
@h-mayorquin not anymore! Since 0.100, there is a `DEV_MODE` flag in the main `__init__`

Issue #2629, opened 2024-03-26
Questions about the sorting analyzer


Antho2422:
Hi all, 

I am performing the switch on my pipeline to the **SortingAnalyzer object** instead of WaveformsExtractor and I have some questions on it.

1. Does all the methods related to the waveforms/templates disapeared ? In other terms there is no way to retrieve a template from a specific unit by doing `waveforms.get_unit_template(unit_id=x)` ? 
2. Same question for the mode of the template ("std", "mean"...)
3. I also encountered a bug when computing isi_histograms (I can of course provide more context if needed but I am trying here just in case if it is something known). I get this numba error : `TypeError: No matching definition for argument type(s) array(int64, 2d, C), array(int64, 1d, C), array(int32, 1d, C), array(int32, 1d, C)` on line 149 of `isi.py` file. 
4. I also have questions on the fast_templates and it's difference with the "classic" templates

Sorry for all the questions, 

Hope it is clear enough
Cheers, 
Anthony

samuelgarcia:
Salut Anthony,
1. and 2.
```python
ext = analyzer.get_extension("templates")
templates_average = ext.get_templates(unit_ids=..., operator="average")
templates_std = ext.get_templates(unit_ids=..., operator="std")
```

3. oups @DradeAW @chrishalcrow : could you have a look ?

4. The fast template do not extract waveforms but directly compute the average (and soon the std) on the fly which is a bit faster and use less memory. But this prevent to make percentile and median estimation.

h-mayorquin:
@Antho2422 
For number 3, this seems like a bug. Can you open another issue with your setup (the code that you run) and the full trace?

alejoe91:
> 4. The fast template do not extract waveforms but directly compute the average (and soon the std) on the fly which is a bit faster and use less memory. But this prevent to make percentile and median estimation.

https://github.com/SpikeInterface/spikeinterface/pull/2623

chrishalcrow:
> 3\. I also encountered a bug when computing isi_histograms (I can of course provide more context if needed but I am trying here just in case if it is something known). I get this numba error : `TypeError: No matching definition for argument type(s) array(int64, 2d, C), array(int64, 1d, C), array(int32, 1d, C), array(int32, 1d, C)` on line 149 of `isi.py` file.

There was a similar bug for the correlogram computation, so I think I know what's up. I'll look into it properly tomorrow morning.

samuelgarcia:
merci à tous.

[...]

PR #2632, opened 2024-03-27
Fix isi_histograms when using numba

chrishalcrow:
As mentioned in #2629 

Type error when passing some arrays to numba in the `compute_isi_histograms` function. Fix is to explicitly type `bins`. Hard to test, since the default typing can depend on many things => hard to write something that _doesn't_ work.


samuelgarcia:
merci beaucoup.

samuelgarcia (approved):

Antho2422:
Thank you !!!

PR #2644, opened 2024-03-29
Prepare 0.100.4

alejoe91:
Minor release with this PR:

- #2643

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -972,7 +972,7 @@ class BaseExtractor:
         filters: list[numcodecs.Codec] or None, default: None
             Global filters for zarr (global)
         compressor_by_dataset: dict or None, default: None
-            Optional compressor per dataset.:
+            Optional compressor per dataset:
                 - traces
                 - times
             If None, the global compressor is used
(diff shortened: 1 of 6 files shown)


src/spikeinterface/core/base.py
```diff
@@ -967,9 +967,24 @@ def save_to_zarr(
             For cloud storage locations, this should not be None (in case of default values, use an empty dict)
         channel_chunk_size: int or None, default: None
             Channels per chunk (only for BaseRecording)
+        compressor: numcodecs.Codec or None, default: None
+            Global compressor. If None, Blosc-zstd level 5 is used.
+        filters: list[numcodecs.Codec] or None, default: None
+            Global filters for zarr (global)
+        compressor_by_dataset: dict or None, default: None
+            Optional compressor per dataset.:
```
zm711, line 975:
```suggestion
            Optional compressor per dataset:
```
:)

zm711 (commented):
Did you also just want to point the docs on the readme to 0.100.4 just so that it stays in sync (no real doc changes, but might be good in case people click on it :))

src/spikeinterface/core/base.py
alejoe91: Thanks @zm711 

Will do it here #2645

PR #2649, opened 2024-04-01
Fix get_traces for a local common reference

chrishalcrow:
Fixes bug in `get_traces` for common reference recordings with a `local` reference, where you couldn't select a subset of channels. Noticed it when plotting traces.

Here's minimal code which reproduces the bug

```
from spikeinterface.preprocessing import common_reference

recording = generate_recording(durations=[5.0], num_channels=4)
recording_cmr_segment = common_reference(recording, reference="local")._recording_segments[0]

recording_cmr_segment.get_traces(start_frame=0, end_frame=10, channel_indices=[1,2])
```
This should raises an error.

Also added a test which selects a subset of channels ids, and makes sure it's selected the right one.

### DIFF
--- a/src/spikeinterface/preprocessing/common_reference.py
+++ b/src/spikeinterface/preprocessing/common_reference.py
@@ -189,8 +189,8 @@ class CommonReferenceRecordingSegment(BasePreprocessorSegment):
                 shift = traces[:, self.ref_channel_indices]
                 re_referenced_traces = traces[:, channel_indices] - shift
             else:  # then it must be local
-                re_referenced_traces = np.zeros_like(traces[:, channel_indices])
                 channel_indices_array = np.arange(traces.shape[1])[channel_indices]
+                re_referenced_traces = np.zeros((traces.shape[0], len(channel_indices_array)), dtype="float32")
                 for i, channel_index in enumerate(channel_indices_array):
                     channel_neighborhood = self.neighbors[channel_index]
                     channel_shift = self.operator_func(traces[:, channel_neighborhood], axis=1)
--- a/src/spikeinterface/preprocessing/common_reference.py
+++ b/src/spikeinterface/preprocessing/common_reference.py
@@ -191,10 +191,10 @@ class CommonReferenceRecordingSegment(BasePreprocessorSegment):
             else:  # then it must be local
                 re_referenced_traces = np.zeros_like(traces[:, channel_indices])
                 channel_indices_array = np.arange(traces.shape[1])[channel_indices]
-                for channel_index in channel_indices_array:
+                for i, channel_index in enumerate(channel_indices_array):
                     channel_neighborhood = self.neighbors[channel_index]
                     channel_shift = self.operator_func(traces[:, channel_neighborhood], axis=1)
-                    re_referenced_traces[:, channel_index] = traces[:, channel_index] - channel_shift
+                    re_referenced_traces[:, i] = traces[:, channel_index] - channel_shift
 
             return re_referenced_traces.astype(self.dtype, copy=False)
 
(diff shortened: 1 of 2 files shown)


alejoe91:
Hi Chris, does it fail if you call the `recording.get_traces()` as well? Or only the segment?

chrishalcrow:
> Hi Chris, does it fail if you call the `recording.get_traces()` as well? Or only the segment?

Morning! It works for `get_traces` on the recording. So it's only a problem if some functions (or users) use `get_traces` on the segments directly (e.g. in `plot_traces`).

alejoe91 (approved):

src/spikeinterface/preprocessing/tests/test_common_reference.py
```diff
@@ -71,6 +71,17 @@ def test_common_reference_channel_slicing(recording):
...
 
+def test_common_reference_select_channels_local(recording):
+
+    recording_cmr = common_reference(recording, reference="local")
+    recording_segment = recording_cmr._recording_segments[0]
+
+    traces_all = recording_segment.get_traces(start_frame=0, end_frame=10, channel_indices=[0, 1, 2, 3])
+    traces_sub = recording_segment.get_traces(start_frame=0, end_frame=10, channel_indices=[1, 3])
+
+    assert np.all(traces_all[:, [1, 3]] == traces_sub)
```
alejoe91, line 82: @chrishalcrow instead of this "segment" test, I guess you can trigger the same issue by adding this line:
```
local_trace = recording_local_car.get_traces(channel_ids=["c", "d"])
```
after L71

I would actually prefer to extend that test instead of adding this new one, since it was designed (badly) to test the slicing behavior. Could you do that?

alejoe91 (changes requested):
Thansk @chrishalcrow 

Check comment!

src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -191,10 +191,10 @@ def get_traces(self, start_frame, end_frame, channel_indices):
             else:  # then it must be local
                 re_referenced_traces = np.zeros_like(traces[:, channel_indices])
                 channel_indices_array = np.arange(traces.shape[1])[channel_indices]
```
samuelgarcia, line 193: even theses 2 lines looks very very bad. no ?

alejoe91: this is because `channel_indices` can be a slice and we need an iterator

chrishalcrow: Maybe they look bad, but they're not breaking anything! Is this ultimately due to the difference between `channel_ids` and `channel_indices`?

src/spikeinterface/preprocessing/tests/test_common_reference.py
chrishalcrow: I have extended the test to include what you say, and also slicing the segments for each type of common reference. Also made the tests a little harder by choosing ["b", "d"] rather than ["a", "b"]. I also shorted the generated recording from `5s` to `1s`: I can't see any reason here _not_ to do this and it doubles the speed of the test for me locally.

alejoe91 (approved):

src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -191,10 +191,10 @@ def get_traces(self, start_frame, end_frame, channel_indices):
             else:  # then it must be local
                 re_referenced_traces = np.zeros_like(traces[:, channel_indices])
```
samuelgarcia, line 192: Lets do this fater
```suggestion
                re_referenced_traces = np.zeros((traces.shape[0], len(channel_indices_array.size), dtype=traces.dtype)
```

src/spikeinterface/preprocessing/common_reference.py
alejoe91, line 193:
```suggestion
                channel_indices_array = np.arange(traces.shape[1])[channel_indices]
                re_referenced_traces = np.zeros((traces.shape[0], len(channel_indices_array)), dtype="float32")
```

PR #2651, opened 2024-04-02
Fix typo in fast templates test

alejoe91:
@DradeAW there was a typo in the tests, but still fast and normal templates give the same stds down to the 4th decimal (which I think it's ok)

### DIFF
--- a/src/spikeinterface/core/waveform_tools.py
+++ b/src/spikeinterface/core/waveform_tools.py
@@ -864,7 +864,7 @@ def estimate_templates_with_accumulator(
 
     # average
     waveforms_sum = np.sum(waveform_accumulator_per_worker, axis=0)
-    template_means = waveforms_sum
+    template_means = waveforms_sum.copy()
     unit_indices, spike_count = np.unique(spikes["unit_index"], return_counts=True)
     template_means[unit_indices, :, :] /= spike_count[:, np.newaxis, np.newaxis]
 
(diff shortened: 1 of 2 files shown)


DradeAW:
Yeah, I'm getting -5943.21 sometimes, so I don't think it's a rounding issue ^^

I'll go digging to understand what's going on!

DradeAW:
This is good for me,

Thanks Alessio :)

src/spikeinterface/core/waveform_tools.py
```diff
@@ -864,20 +864,20 @@ def estimate_templates_with_accumulator(
 
     # average
     waveforms_sum = np.sum(waveform_accumulator_per_worker, axis=0)
-    template_means = waveforms_sum
+    template_means = waveforms_sum.copy()
```
samuelgarcia, line 867: We need a copy only if computing STD otherwise the copy is useless.

samuelgarcia (approved):

DradeAW:
I don't like this,
Because now, if `return_std=False`, then `waveforms_sum` is changed (pointer problem)

Of course it's not used in this case, but I think it's bad practice (what if someone implements something in 8 months from now? This is a potential bug)

alejoe91:
> I don't like this, Because now, if `return_std=False`, then `waveforms_sum` is changed (pointer problem)
> 
> Of course it's not used in this case, but I think it's bad practice (what if someone implements something in 8 months from now? This is a potential bug)

This was a request from @samuelgarcia, since copying the buffer can be large for many units...

samuelgarcia (approved):

PR #2652, opened 2024-04-02
Populate generation module and fix hybrid classes

alejoe91:
Moved all functions from `core/generate` and `comparison/hybrid` to the `generation` module

Also updated the `HybridUnits` and `HybridSpikes` to use the `SortingAnalyzer`


samuelgarcia:
You moved almost evrything in generation no ?
I remember we decided to keep minimal generation in core for testing and move in generation complicated cases like drift or collision.
Am I wrong ?

src/spikeinterface/core/__init__.py
```diff
@@ -169,3 +149,20 @@
...
+    generate_snippets,
+    synthesize_random_firings,
+    inject_some_duplicate_units,
+    inject_some_split_units,
+    synthetize_spike_train_bad_isi,
+    generate_templates,
+    noise_generator_recording,
+    generate_recording_by_size,
+    generate_ground_truth_recording,
+)
```
samuelgarcia, line 168: really a bad choice in my mind.

PR #2661, opened 2024-04-03
Fix channel gains in  `NwbRecordingExtractor`  with backend

h-mayorquin:
Currently the reading of the `channel_conversion` to gains in the backend is not working. The reason is that this is written as a dataset and not an attribute the way that the current function assumes.

This PR adds a test that fails with the current state of the repo and also the changes that fix this. 

I detected this while working on a neuroconv issue:
https://github.com/catalystneuro/neuroconv/pull/803


h-mayorquin:
This requires #2665

alejoe91 (approved):

zm711:
Could we try limiting pytest to < 8 and see if that fixes this PR? (based on @h-mayorquin and my discussion I think that is worth a try to see if this is a pytest collection issue). If that fixes this test then the decision would be to permanently limit pytest or rewrite the action and see if that fixes this.

h-mayorquin:
@zm711 the problem was that scipy did not have an upper bound and ibllib was throwing this error when importing.  This illustrate a problem with the non-specificty of the try-except block. It took me a while to discover what was causing it exactly .

zm711:
Cool! As long as you got it!

h-mayorquin:
Yeah, it is good to be aware that there might be something broken in pytest. Let's keep an eye on that.

PR #2662, opened 2024-04-03
Deprecate ros3 as a streaming method

h-mayorquin:
We have discussed about this for a while. It is hard to mantain and test and it does not seem to buy anything over `ffspec`.

I will do it now as I am getting some import errors on the stream extractors test on https://github.com/SpikeInterface/spikeinterface/pull/2661 and I want to get rid of this complexity before digging into this.


h-mayorquin:
This requires:
#2665

samuelgarcia:
merci

PR #2665, opened 2024-04-03
Fix full tests

h-mayorquin:
Full tests are failling. I think this is related to this issue here:

https://github.com/int-brain-lab/ibllib/issues/753

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,7 +88,7 @@ full = [
     "h5py",
     "pandas",
     "xarray",
-    "scipy",
+    "scipy<1.13",
     "scikit-learn",
     "networkx",
     "distinctipy",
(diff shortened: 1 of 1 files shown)


alejoe91 (approved):
Thanks for looking into this @h-mayorquin

pyproject.toml
```diff
@@ -86,7 +88,7 @@ full = [
     "h5py",
     "pandas",
     "xarray",
-    "scipy",
+    "scipy<1.13",
```
samuelgarcia, line 91: maybe for full we could remove this boundarie no ?
is it only for ibl stuff ?

h-mayorquin: I guess we could move the boundary to the testing section. 

We gain:
* Some people will be able to use the latest scipy with other packages that don't use the wrong import.
We lost:
Some people will install full and then it won't work with IBL.

Your call guys, @samuelgarcia @alejoe91

samuelgarcia: upper boundaries is sometimes anoying.
For me, the balance is that very few people use the ibllib reader vs more poeple are making a fresh install with recent scipy.
lets discuss this in a call

PR #2673, opened 2024-04-05
Allowing to bypass the random_spikes_selection to estimate sparsity

yger:
Currently, the random_spikes_selection() function is really slow for large number of spikes/units. Therefore, the sorting analyzer is not operational when you have lots of neurons. This PR allows to bypass such a selection if num_spikes_for_sparsity is set to None. This is a temp fix, while @samuelgarcia will optimize the random_spikes_selection() function


yger:
this has been fixed by #2700

Issue #2213, opened 2023-11-15
Peaks (node_pipeline) detected in the margin of the chunks


maxjuv:
Hi,
I wanted to try the efficiency of a chained peak detection and peak localization. With the same parameters, on the same recording I tried two pipelines (chained vs separated detection/localisation).  The chained pipeline crashes at waveform extraction. In fact, some peaks are too close to border.  
My code for chained detection/localization, SI version 0.100.dev0 :
  
```
extract_dense_waveforms = ExtractDenseWaveforms(recording, ms_before=0.5, ms_after=0.5, return_output=False)
pipeline_nodes = [extract_dense_waveforms,
                 LocalizeMonopolarTriangulation(recording,parents=[extract_dense_waveforms],  **loc_kwargs)]
peaks, peaks_location = detect_peaks(recording,
                                     method='locally_exclusive_torch',
                                     pipeline_nodes = pipeline_nodes ,
                                    **kwargs)
```
My code for separated detection/localization:

```
peaks = detect_peaks(recording, method='locally_exclusive_torch', **kwargs)
peaks_location = localize_peaks(recording, peaks,  method='monopolar_triangulation' , **loc_kwargs )
```

To be noted, for the peak detection I used exclude_sweep_ms=0.1

I tracked the workflow of the node_pipeline and I think the issue comes from the line below. 
https://github.com/SpikeInterface/spikeinterface/blob/e4c5d11fd58259f7fcb0af41f029a5cac9291e82/src/spikeinterface/core/node_pipeline.py#L508
For peak detection node.get_trace_margin() returns exclude_sweep_ms in samples. If an exclusion sweep is defined, then the margin (called extra_margin) for peak detection in refined. It appeared that peaks can be detected in this extra_margin, which is not a problem if detection and localization are not chained. However, if chained, the peak in the In the case peak localization is chained, the waveforms of the peaks in the extra_margin can no longer be extracted. Finally, if exclude_sweep_ms is set to 0, the chained pipeline workes. With the following observation, I think the issue is located L508 in core.node_pipeline.py

Cheers
Maxime

maxjuv:
Update: In fact with my last run, the separated pipeline also failed, during the peak localization (using torch - cuda). So in the end, I don't know where the bug is.
the error is 
```
  File "/home/users/j/juventin/spike_sorters/spikeinterface/src/spikeinterface/core/node_pipeline.py", line 320, in compute
[...]

PR #2674, opened 2024-04-05
Changing the generation of fake template with a decaying exponential and fix torch locally exclusive

yger:
In this PR, we change the way fake templates are generated. Instead of a powerlaw for the decay of the amplitudes as function of the distances, we use a decaying exponential with a spatial_constant. This seems to be more biophysically plausible, and it introduces less artefacts

### DIFF
--- a/src/spikeinterface/core/tests/test_analyzer_extension_core.py
+++ b/src/spikeinterface/core/tests/test_analyzer_extension_core.py
@@ -111,7 +111,7 @@ def test_ComputeWaveforms(format, sparse):
 def test_ComputeTemplates(format, sparse):
     sorting_analyzer = get_sorting_analyzer(format=format, sparse=sparse)
 
-    sorting_analyzer.compute("random_spikes", max_spikes_per_unit=500, seed=2205)
+    sorting_analyzer.compute("random_spikes", max_spikes_per_unit=50, seed=2205)
 
     job_kwargs = dict(n_jobs=2, chunk_duration="1s", progress_bar=True)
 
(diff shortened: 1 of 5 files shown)


alejoe91:
@yger can you fix the failing tests

yger:
Yes, I will, but I have no idea why my modification in the fake templates is triggering a negative std for the template. This make no sense, I'll have a look

alejoe91:
Can you try with this? https://github.com/SpikeInterface/spikeinterface/pull/2651

yger:
found it, it was some default alpha range. Tests should pass now

yger:
Looks like this is triggering an error in a peak detection function with torch... @samuelgarcia , did you made this functio?

alejoe91:
I ported it, I'll take a look

alejoe91:
@yger should be fixed!

Plus it fixes #2213

src/spikeinterface/core/tests/test_analyzer_extension_core.py
```diff
@@ -111,7 +111,7 @@ def test_ComputeWaveforms(format, sparse):
 def test_ComputeTemplates(format, sparse):
     sorting_analyzer = get_sorting_analyzer(format=format, sparse=sparse)
 
-    sorting_analyzer.compute("random_spikes", max_spikes_per_unit=20, seed=2205)
+    sorting_analyzer.compute("random_spikes", max_spikes_per_unit=500, seed=2205)
```
samuelgarcia, line 114: This slow down a lot tests. I prefer a small number!!

alejoe91: It forsnt

alejoe91 (approved):

PR #2675, opened 2024-04-05
Allow covariance matrix for noise generation

yger:
We want to make the noise generation more realistic, and to do so, we can allow a covariance matrix in order to make the noise structure more realistic. However, I know that after discussion with @samuelgarcia , the goal is to keep the generation file as simple as possible, so maybe a more complicated object could be put elsewhere. Open to discussion

### DIFF
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -27,6 +27,11 @@ def _ensure_seed(seed):
         seed = np.random.default_rng(seed=None).integers(0, 2**63)
     return seed
 
+def get_distances_from_probe(probe):
+    locations = probe.contact_positions
+    channel_distances = np.linalg.norm(locations[:, np.newaxis] - locations[np.newaxis, :], axis=2)
+    return channel_distances
+
 
 def generate_recording(
     num_channels: Optional[int] = 2,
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/generate.py
```diff
@@ -2006,6 +2040,19 @@ def generate_ground_truth_recording(
...
+        cov_matrix = np.zeros((num_channels, num_channels), dtype=np.float32)
+        for i in range(num_channels):
+            for j in range(num_channels):
+                if i != j:
+                    cov_matrix[i, j] = (0.5 * correlated_noise) / distances[i, j]
+                else:
+                    cov_matrix[i, j] = 1
+
+    else:
+        cov_matrix = None
```
alejoe91, line 2054: Looks familar :P

alejoe91 (approved):

yger:
Indeed, MEArec is back! After discussing with @DradeAW we realized the noise structure might be crucial for benchmarking. How did you calibrate the noise in mearec? We could discuss that more in-depth, because what would be the key components to make it more plausible. But then as Sam told me, the idea is to keep the generation as simple as possible...

alejoe91:
My suggestion is the following:

- we keep a minimal generate.py in core, with very simple generation stuff (no ellipsoid, complex noises, etc)
- we move the complex machinery to the generation module and make ismy as complex as we want!

How does that sound?

yger:
This sounds right, I'm open to anything. I'll check with @samuelgarcia how to organize such move.

samuelgarcia:
Done in #2683.
So close.

PR #2680, opened 2024-04-08
Death of "fast_templates" extention

samuelgarcia:
"templates" and "fast_templates" extension have been merge together.
Done with Alessio.

### DIFF
--- a/src/spikeinterface/qualitymetrics/pca_metrics.py
+++ b/src/spikeinterface/qualitymetrics/pca_metrics.py
@@ -643,10 +643,7 @@ def nearest_neighbors_noise_overlap(
 
     templates_ext = sorting_analyzer.get_extension("templates")
     assert templates_ext is not None, "nearest_neighbors_isolation() need extension 'templates'"
-    if sorting_analyzer.has_extension("waveforms"):
-        template_operator = "median"
-    else:
-        template_operator = "average"
+
     if n_spikes_all_units is None:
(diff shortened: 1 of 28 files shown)


src/spikeinterface/qualitymetrics/pca_metrics.py
```diff
@@ -643,7 +643,10 @@ def nearest_neighbors_noise_overlap(
 
     templates_ext = sorting_analyzer.get_extension("templates")
     assert templates_ext is not None, "nearest_neighbors_isolation() need extension 'templates'"
-
+    if sorting_analyzer.has_extension("waveforms"):
+        template_operator = "median"
+    else:
+        template_operator = "mean"
```
samuelgarcia, line 649: Not sure to like this. Because the results depend on the sate of analyzer and not on explicit parameters.

alejoe91: plus this should be `average`

alejoe91: actually the `QualityMetrics` extension depends on `waveforms`, so I'll just revert

alejoe91 (approved):

src/spikeinterface/qualitymetrics/pca_metrics.py
samuelgarcia: done

PR #2683, opened 2024-04-09
Generate drift

samuelgarcia:
@alejoe91 : this is a important piece that @yger  will need during my vaccation.

This refactor a bit some parameters in core.generate.

But this mainly add the function `generate_drifting_recording()` with a nice and simple API to generate drift recordings.

Do you think we could merge this very soon ?


src/spikeinterface/generation/drifting_generator.py
```diff
@@ -0,0 +1,401 @@
...
+        bump_interval_s=(30, 90.),
+        seed=None
+    ):
+    """
+    Generator a toy discplacement vector like ziagzag or bumps.
+
+    
+    Parameters
+    ----------
+    drift_mode: "zigzag" | "bumps"
```
zm711, line 56: I'm assuming these are the technical terms for drift :P

samuelgarcia: This is the naing we chosse for the paper in eneuro. See https://www.eneuro.org/content/11/2/ENEURO.0229-23.2023

src/spikeinterface/core/generate.py
```diff
@@ -609,6 +609,25 @@ def generate_snippets(
...
 ## spiketrain zone ##
+
+
+def _ensure_firing_rates(firing_rates, num_units, seed):
+    rng = np.random.default_rng(seed=seed)
+
+    if isinstance(firing_rates, tuple):
+        rng = np.random.default_rng(seed=seed)
+        lim0, lim1 = firing_rates
+        firing_rates = rng.random(num_units) * (lim1 - lim0) + lim0
```
h-mayorquin, line 620: rng.uniform:

https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.uniform.html

src/spikeinterface/core/generate.py
```diff
@@ -635,7 +654,7 @@ def synthesize_poisson_spike_vector(
         Duration of the simulation in seconds
     refractory_period_ms : float, default: 4.0
         Refractory period between spikes in milliseconds
-    firing_rates : float or array_like, default: 3.0
+    firing_rates : float or array_like or tuple, default: 3.0
         Firing rate(s) in Hz. Can be a single value for all units or an array of firing rates with
         each element being the firing rate for one unit
```
h-mayorquin, line 659: Let's add documentation for the tuple.

src/spikeinterface/core/generate.py
```diff
@@ -1079,15 +1116,24 @@ def __init__(
...
         self.seed = seed
         self.strategy = strategy
 
         if self.strategy == "tile_pregenerated":
             rng = np.random.default_rng(seed=self.seed)
-            self.noise_block = (
-                rng.standard_normal(size=(self.noise_block_size, self.num_channels), dtype=self.dtype) * noise_levels
-            )
+
+            if self.cov_matrix is None:
```
h-mayorquin, line 1127: Another option is that the cov is the identity by default and then you don't need this if. But I am not sure if this is less performant. 

Maybe we can do that later.

src/spikeinterface/core/generate.py
```diff
@@ -609,6 +609,25 @@ def generate_snippets(
 
 
 ## spiketrain zone ##
+
+
+def _ensure_firing_rates(firing_rates, num_units, seed):
+    rng = np.random.default_rng(seed=seed)
+
+    if isinstance(firing_rates, tuple):
+        rng = np.random.default_rng(seed=seed)
```
h-mayorquin, line 618: Thisi s repeated from line 615. You can also just pass the rng instead of the seed to this function and avoid another instantiation.

src/spikeinterface/core/generate.py
```diff
@@ -667,8 +686,7 @@ def synthesize_poisson_spike_vector(
 
     rng = np.random.default_rng(seed=seed)
 
-    if np.isscalar(firing_rates):
-        firing_rates = np.full(num_units, firing_rates, dtype="float64")
+    firing_rates = _ensure_firing_rates(firing_rates, num_units, seed)
```
h-mayorquin, line 689: You can pass the rng as well as mentioned above?

src/spikeinterface/generation/drifting_generator.py
```diff
@@ -0,0 +1,505 @@
...
+                factors = 1 - factors
+            f = np.abs(non_rigid_gradient)
+            displacement_unit_factor[:, m] = factors * (1 - f) + f
+
+    displacement_vectors = np.concatenate(displacement_vectors, axis=2)
+
+    return displacement_vectors, displacement_unit_factor, displacement_sampling_frequency, displacements_steps
+
+
+def generate_noise(
```
alejoe91, line 250: Please make a `noise_tools.py` file

alejoe91 (changes requested):
See comment

src/spikeinterface/core/generate.py
samuelgarcia: I was thinking that random was uniform no ?
Lets use uniform it is more explicit maybe.

src/spikeinterface/core/generate.py
samuelgarcia: This internally make a np.dot I think so handling None is good no ?

src/spikeinterface/core/generate.py
samuelgarcia: oups thank you.

src/spikeinterface/core/generate.py
samuelgarcia: The thing is that if the rng is used before or not depending other global option then we do not have the same.
With this seed  given we somehow reset the generator whatever happen.

samuelgarcia:
@h-mayorquin.
I know that you had in mind the memory foot print when making the design but I had mind also this generator for simulation!! héhé!

If the object would become too complex, ok for several classes. But here it is really a minimal change and I think having only one class is ok no ?

src/spikeinterface/core/generate.py
h-mayorquin: Yeah, but only 0 and 1. With uniform you don't need to do the math yourself as you can specificy high and low.

src/spikeinterface/core/generate.py
h-mayorquin: I meant that if cov=None then cov=np.eye(number_of_channels) and then you don't have the if.

But I think this version is more memory efficient when cov is None as it avoids the creation of a large matrix.

src/spikeinterface/core/generate.py
h-mayorquin: You are correct. This is a very good point for reproducibility. Passing the rng means you are passing state which is wrong.

h-mayorquin:
> @h-mayorquin. I know that you had in mind the memory foot print when making the design but I had mind also this generator for simulation!! héhé!
> 
> If the object would become too complex, ok for several classes. But here it is really a minimal change and I think having only one class is ok no ?

Yeah, the complexity is not there yet but it always accumulates sloly. The memory foot-print is still the same for the `cov=None` case I think.

I was mentioning this as something to keep an eye on but this should not stop this PR from being merged. Once the two different purposes drift enough appart we should separate.

src/spikeinterface/core/generate.py
alejoe91: I also vote for `uniform(lim0, lim1)`

alejoe91 (approved):

Issue #2507, opened 2024-02-21
`SortingAnalyzer.compute(save=True)` is a little unclear


zm711:
So when using the `compute` function to do postprocessing, qc etc, I can give save=True or save=False. For testing I just created a `SortingAnalyzer`  in memory. Although it makes sense that save=True does nothing since I'm in memory we will definitely need to document that this is how it works. 

I can imagine a workflow where I created my `SortingAnalyzer` in memory and just assume that by setting `save=True` it automatically saves the result without actually doing the `save_as`. This isn't super clear or user friendly to understand the distinction right now. 

Solutions could be:
* Better warning messaging so people understand how extensions work with memory mode
* Better documentation (not done yet I know :) , but an important section that would need to be added)

Otherwise, the extensions all seem to be working pretty well. I think the error messaging to say "you are missing this extension run comput(extension) " is super useful. That was a really nice touch :) 

I'm going to be testing some plotting this afternoon.

samuelgarcia:
Yes we need to docuemtn, this is one of the best new stuff we have in the new design 2 apporaches:
  * everything in memory (and then save=True/False is ignored) and then save_as()
  * everything on disk (zarr or npy) and then save on the fly (or not) every computation. The save=False make a lot of sens in spikeinterface-gui  because you may want to view crosscorrelogram with new bins without changing the folder itself. See https://github.com/SpikeInterface/spikeinterface-gui/pull/53

zm711:
I agree it's great. I'm constantly wishing I could tweak one tiny thing without changing everything. I think the concern from a mental schema perspective is 

`save` and `save_as` are semantically loaded in that they do the same thing except save_as you change the name of your file.  So I as a user may (erroneously in this case) assume just by running `save` it does the same thing as `save` as just without changing the format. I'm trying to think of a good way to make sure users don't make this incorrect mental leap when it is so easy based on using Office or Apple's docs. (I assume LibreOffice is the same semantically).

Basically I'm saying that people will bring in their own preconceived notions of `save` vs `save_as`.

chrishalcrow:
Hello, just working on docs, and now also think this is quite confusing. Just to make Zach's point more explicit, consider the following code:

```
recording, sorting = generate_ground_truth_recording(
    durations=[
        5.0,
    ],
    sampling_frequency=30_000.0,
    num_channels=6,
    num_units=10,
    generate_sorting_kwargs=dict(firing_rates=6.0, refractory_period_ms=4.0),
[...]

PR #2693, opened 2024-04-10
Extend compute for extension kwargs and pipeline nodes

alejoe91:
@zm711 I remember you requested this feature, and I also think it's super nice!

This enables to run `compute` with:

```
analyzer.compute(
    ["waveforms", "correlograms"], 
   extension_params=dict(
        waveforms=dict(ms_before=1, ms_after=2), 
        correlograms=dict(bin_ms=1),
    )
)
```


src/spikeinterface/core/tests/test_sortinganalyzer.py
```diff
@@ -147,6 +142,9 @@ def _check_sorting_analyzers(sorting_analyzer, original_sorting):
         # unit 1, 3, ... should be removed
         assert np.all(~np.isin(data["result_two"], [1, 3]))
 
+    # test compute with extension-specific params
+    sorting_analyzer.compute(input=["dummy"], dummy=dict(param1=5.5))
```
samuelgarcia, line 146: I do not like this both list + dict writting.
The list was done for a shortcut.
Lets discuss this behavior I think this is too confusing.

For me it was:
 * dict
 * str
 * list
 
Here you are introducing list + dict

alejoe91: I would also prefer and use this. Or even better, having `extension_params` as a sepratate dictionary, that could be also propagated to the `compute_several_extensions(ext_list, ext_params_dict)`

zm711: Honestly I like that flow better. list, dict makes more sense rather than nested dict. So if we remove the option to do it with a dict and force the user to either do: 

str: dict
list: dicts

then I'm okay with that flow.

zm711:
Sounds good :)

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -925,27 +941,28 @@ def compute_several_extensions(self, extensions, save=True, **job_kwargs):
...
+            else:
+                extensions_without_pipeline[extension_name] = extension_params
 
+        # First extensions without pipeline
+        for extension_name, extension_params in extensions_without_pipeline.items():
+            extension_class = get_extension_class(extension_name)
+            if extension_class.need_job_kwargs:
+                self.compute_one_extension(extension_name, save=save, **extension_params)
+            else:
+                self.compute_one_extension(extension_name, save=save, **extension_params)
```
zm711, line 959: How are these different?

zm711 (commented):
Just a quick question @alejoe91

src/spikeinterface/core/sortinganalyzer.py
alejoe91: good question! It was already there :P

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -925,27 +941,28 @@ def compute_several_extensions(self, extensions, save=True, **job_kwargs):
...
+            if extension_class.use_nodepipeline:
+                extensions_with_pipeline[extension_name] = extension_params
+            else:
+                extensions_without_pipeline[extension_name] = extension_params
 
+        # First extensions without pipeline
+        for extension_name, extension_params in extensions_without_pipeline.items():
+            extension_class = get_extension_class(extension_name)
+            if extension_class.need_job_kwargs:
+                self.compute_one_extension(extension_name, save=save, **extension_params)
```
alejoe91, line 957:
```suggestion
                self.compute_one_extension(extension_name, save=save, **extension_params, **job_kwargs)
```

alejoe91: I think this looks more correct

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -802,17 +802,27 @@ def get_num_units(self) -> int:
...
         ----------
         input: str or dict or list
-            If the input is a string then computes one extension with compute_one_extension(extension_name=input, ...)
-            If the input is a dict then compute several extensions with compute_several_extensions(extensions=input)
+            The extesions to compute, which can be passed as:
+
+            * a string: compute one extension. Additional parameters can be passed as key word arguments.
+            * a dict: compute several extensions. The keys are the extension names and the values are the extension parameters.
+            * a list: compute several extensions. The list contains the extension names. Additional parameters can be passed with the extension_params
+              argument.
```
zm711, line 818: I'm not sure if this should go here or not, but how does one pass default params? If I'm the user reading the docstring I'm not sure if it is clear. Maybe it is, but for example if I'm doing a dict: what if I do {'random_spikes': []} instead of {'random_spikes': {}}? or for the list I might assume that if I specify for one I have to specify for all. Maybe we need this in the actual docs rather than here, but something we should think about.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -802,17 +802,27 @@ def get_num_units(self) -> int:
...
+        save: bool, default: True
+            If True the extension is saved to disk
```
zm711, line 820: Also this is confusing. Because save=True only works if format != 'memory'. I have an issue open on this currently #2507. It doesn't have to be in this PR, but I think we need to find a way to be more explicit about how `save` works so that people don't think that in memory mode they can just set `save=True` and things will save. Again I think this needs to probably be in the docstring as well as in our docs.

zm711 (commented):
Code looks good to me now (but I don't fully follow how the node pipeline works in general so don't trust me on that :) 

Just a couple docstring comments. More points of discussion rather than hard recs for this PR.

samuelgarcia:
OK for me but I would improve the doc string and propagate this to the .rst doc. no ?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -802,17 +802,27 @@ def get_num_units(self) -> int:
         return self.sorting.get_num_units()
 
     ## extensions zone
-
-    def compute(self, input, save=True, **kwargs):
+    def compute(self, input, save=True, extension_params=None, **kwargs):
         """
         Compute one extension or several extensiosn.
         Internally calls compute_one_extension() or compute_several_extensions() depending on the input type.
```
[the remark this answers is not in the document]
alejoe91: done!

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -802,17 +802,52 @@ def get_num_units(self) -> int:
...
 
         Parameters
         ----------
         input: str or dict or list
-            If the input is a string then computes one extension with compute_one_extension(extension_name=input, ...)
-            If the input is a dict then compute several extensions with compute_several_extensions(extensions=input)
+            The extesions to compute, which can be passed as:
```
zm711, line 813:
```suggestion
            The extensions to compute, which can be passed as:
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -802,17 +802,52 @@ def get_num_units(self) -> int:
...
+
+            * a string: compute one extension. Additional parameters can be passed as key word arguments.
+            * a dict: compute several extensions. The keys are the extension names and the values are dictiopnaries with the extension parameters.
+            * a list: compute several extensions. The list contains the extension names. Additional parameters can be passed with the extension_params
+              argument.
+        save: bool, default: True
+            If True the extension is saved to disk (only of sorting analyzer format is not "memory")
```
zm711, line 820:
```suggestion
            If True the extension is saved to disk (only if sorting analyzer format is not "memory")
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -802,17 +802,52 @@ def get_num_units(self) -> int:
...
+
+        Compute two extensions with dict as input, one dict per extension
+        >>> analyzer.compute({"random_spikes":{}, "waveforms":{"ms_before":1.5, "ms_after", "2.5"}})
+
+        Compute two extensions with an input list specifying custom parameters for one
+        (the other will use default parameters):
+        >>> analyzer.compute(
+            ["random_spikes", "waveforms"],
+            extension_params={"waveforms":{"ms_before":1.5, "ms_after", "2.5"}}
+        )
```
zm711, line 850: This example looks great to me!

I'm sure Chris or I can update the docs themselves soon. (I think we need a section of compute, deleting extensions during recompute) amongst other things. So I think the scope for the docs is a new PR!

samuelgarcia:
I merge this now.

@chrishalcrow : would you have time to update the doc accordingly to this ?
See example in analyzer.compute().
And also we should remove the "fast_templates" in docs.

samuelgarcia (approved):

zm711:
I'll open an issue and try to list the SortingAnalyzer updates we need when I get to lab!

Issue #2694, opened 2024-04-10
Split function and `new_unit_ids` definition


OlivierPeron:
Hi !

I was looking into the curation module and didn't quite get how to use the split function. Let's say I want to split my unit into 3. To do so I just have to properly define my `indices_list` and then I could just do as so:
`cs.split(unit_id, indices_list)`

The class `SplitUnitSorting` is able to handle this, thanks to `tot_splits = max([v.max() for v in indices_list]) + 1`. 
But then we are asserting `assert len(new_unit_ids) <= tot_splits`, and `new_unit_ids` is defined in the `CurationSorting` class :
```
new_unit_ids = self._get_unused_id(2)
        new_sorting = SplitUnitSorting(
            current_sorting,
            split_unit_id=split_unit_id,
            indices_list=indices_list,
            new_unit_ids=new_unit_ids,
            properties_policy=self._properties_policy,
        )
```
Therefore even if do properly my list, there will always be only 2 `new_unit_ids`.

Would solve it doing `new_unit_ids = self._get_unused_id(max(np.unique(indices_list)))` but I am not sure weither this should be changed or if I am not doing it the proper way.

Thanks !

OlivierPeron:
Ok ! 
I don't think using the len of `indices_list` is the way to go as its len is the number of segments in the recording, and for each segment we have an array as big as the unit's spike train.

In the `SplitUnitSorting` class, the proper calculation is done I think:
`tot_splits = max([v.max() for v in indices_list]) + 1`

Then we would just have to do :
`new_unit_ids = self._get_unused_id(tot_splits)`

and this way we could get rid of the assertion line 52  of the `SplitUnitSorting` class:
`assert len(new_unit_ids) <= tot_splits, "indices_list has more id indices than the length of new_unit_ids"`
as this way we would have `len(new_unit_ids) == tot_splits`.

PR #2695, opened 2024-04-10
Split function and new_unit_ids definition #2694

OlivierPeron:
Change discussed in issue #2694

### DIFF
--- a/src/spikeinterface/curation/curationsorting.py
+++ b/src/spikeinterface/curation/curationsorting.py
@@ -61,7 +61,9 @@ class CurationSorting:
 
     def split(self, split_unit_id, indices_list):
         current_sorting = self._sorting_stages[self._sorting_stages_i]
-        new_unit_ids = self._get_unused_id(2)
+        if type(indices_list) is not list:
+            indices_list = [indices_list]
+        new_unit_ids = self._get_unused_id(max([v.max() for v in indices_list]) + 1)
         new_sorting = SplitUnitSorting(
             current_sorting,
             split_unit_id=split_unit_id,
--- a/src/spikeinterface/curation/curationsorting.py
+++ b/src/spikeinterface/curation/curationsorting.py
@@ -61,7 +61,7 @@ class CurationSorting:
 
     def split(self, split_unit_id, indices_list):
         current_sorting = self._sorting_stages[self._sorting_stages_i]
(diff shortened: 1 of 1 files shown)


src/spikeinterface/curation/curationsorting.py
```diff
@@ -61,7 +61,9 @@ def _get_unused_id(self, n=1):
 
     def split(self, split_unit_id, indices_list):
         current_sorting = self._sorting_stages[self._sorting_stages_i]
-        new_unit_ids = self._get_unused_id(2)
+        if type(indices_list) is not list:
+            indices_list = [indices_list]
```
zm711, line 65:
```suggestion
            indices_list = list(indices_list)
```

Is this the right solution? for example if I do:
```
indices_list = np.array([1,2,3])
```
this will turn this into indices_list = [array([1,2,3])] instead of indices_list = [1,2,3]

whereas using the list constructor will work correctly.

OlivierPeron: What we went is something like `indices_list = [array([1,2,3])]`, with one array per recording' segment. If you have `indices_list = [1,2,3]` then this `max([v.max() for v in indices_list]) + 1` could raise an `AttributeError: 'int' object has no attribute 'max'` if v is int instead of np.int

alejoe91: why not just doing `np.max(v)` instead ov `v.max()`? the former works both for lists and array.

Regarding the checks, we should also check that `indices_list[0]` is an iterable, so I would do this:
```
from collections.abc import Iterable

...

if not isinstance(indices_list, list):
    indices_list = [indices_list]
if not isinstance(indices_list[0], Iterable):
    raise ValueError(...)
new_unit_ids = self._get_unused_id(np.max([np.max(v) for v in indices_list]) + 1)
```

zm711: > This protects from passing indices_list=[1,2,3], which should fail in my opinion. Instead, passing np.array([1,2,3]) or [[1,2,3]] should work.

Exactly. Because we were just checking list I was confused. I also think we need a docstring in this case. Needing a list of a list or an array is not intuitive for a variable named indices_list. With a docstring at least the user can be warned what is expected.

alejoe91:
Closing in favor of #2775

PR #2708, opened 2024-04-11
Change "idw" to "kriging" in spatial_interpolation_kernel parameters

samuelgarcia:
change "idw" to "krigging" in spatial_interpolation_kernel in parameters preset because  really depend on the geometry.

### DIFF
--- a/src/spikeinterface/preprocessing/motion.py
+++ b/src/spikeinterface/preprocessing/motion.py
@@ -69,11 +69,7 @@ motion_options_preset = {
             weight_with_amplitude=False,
         ),
         "interpolate_motion_kwargs": dict(
-            direction=1,
-            border_mode="remove_channels",
-            spatial_interpolation_method="kriging",
-            sigma_um=20.0,
-            p=2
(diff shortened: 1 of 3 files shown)


src/spikeinterface/preprocessing/motion.py
```diff
@@ -71,8 +71,8 @@
         "interpolate_motion_kwargs": dict(
             direction=1,
             border_mode="remove_channels",
-            spatial_interpolation_method="idw",
-            num_closest=3,
+            spatial_interpolation_method="kriging",
+            sigma_um=[20.0, 30.0],
```
yger, line 75: missing p=1 to be constistent

src/spikeinterface/preprocessing/motion.py
```diff
@@ -133,8 +133,8 @@
...
```
yger, line 137: same here

yger:
This PR should also changes the detaulf kriging parameters to 20 (isotropic) and p=2. This gives better results for NP probes

PR #2712, opened 2024-04-12
Remove separate default job_kwarg `n_jobs` for sorters

### LINKED ISSUE #2696: `n_jobs` not being respected between set_global_kwargs and running sorter
As discussed in #2690,

image

zm711:
Fixes #2696 

Right now we aren't respecting the global `n_jobs` that user sets so the only way to override use all cores is to specifically pass `n_jobs` to the sorter. This is unintuitive if we allow the user to set this globally except in the case of sorters where it has to be set locally. 

The current global default appears to be `n_jobs=1` so we discussed whether that should be changed for people who don't know to set `n_jobs` so let me know if we want to do that in this PR too.

### DIFF
--- a/src/spikeinterface/sorters/basesorter.py
+++ b/src/spikeinterface/sorters/basesorter.py
@@ -21,11 +21,11 @@ from spikeinterface.core.job_tools import fix_job_kwargs, split_job_kwargs
 from .utils import SpikeSortingError, ShellScript
 
 
-default_job_kwargs = {"n_jobs": -1}
+default_job_kwargs = {}
 
 default_job_kwargs_description = {
-    "n_jobs": "Number of jobs (when saving ti binary) - default -1 (all cores)",
-    "chunk_size": "Number of samples per chunk (when saving ti binary) - default global",
+    "n_jobs": "Number of jobs (when saving to binary) - default global",
+    "chunk_size": "Number of samples per chunk (when saving to binary) - default global",
     "chunk_memory": "Memory usage for each job (e.g. '100M', '1G') (when saving to binary) - default global",
     "total_memory": "Total memory usage (e.g. '500M', '2G') (when saving to binary) - default global",
     "chunk_duration": "Chunk duration in s if float or with units if str (e.g. '1s', '500ms') (when saving to binary)"


alejoe91:
> The current global default appears to be `n_jobs=1` so we discussed whether that should be changed for people who don't know to set `n_jobs` so let me know if we want to do that in this PR too.

I vote for `n_jobs=0.8` :)

@samuelgarcia @h-mayorquin @chrishalcrow what do you guys think?

zm711:
I guess the one concern with changing to a default of multiprocessing is that we may we see more issues whereas with opt-in not everyone is using those. But if we decide to do a default then I think `0.7` or `0.8` seems safe enough!

src/spikeinterface/sorters/basesorter.py
```diff
@@ -21,11 +21,11 @@
 from .utils import SpikeSortingError, ShellScript
 
 
-default_job_kwargs = {"n_jobs": -1}
+default_job_kwargs = {}
```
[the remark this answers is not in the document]
zm711: We can. Once we decide on the global I get do both changes in a new commit.

h-mayorquin:
I think that 0.8 cores sounds good as a default. My intuition is that we should leave one physical core free as most people multitask x D.

zm711:
@samuelgarcia now that you're back opinions on `n_jobs` default? I think we hard code a default for now as we wait for a `jobs_kwargs` optimizer in the future.

samuelgarcia:
If this is for the sorter 0.8 sounds OK (because it is only to write this binary).

But we need to have in mind that on some computing nodes you can have many many cores that would lead to an enormous ram demand.

So for the main n_jobs I prefer to keep 1 so it force the user to think about it to improve speed.

alejoe91 (approved):

zm711:
Based on #2826 and #2820. I think even for sorters we should default to n_jobs=1 because some computers aren't working with multiprocessing at all. Currently not all sorters accept n_jobs as a parameter so maybe we add that to base sorting instead?

alejoe91:
I agree. Maybe a warning at the `fix_job_kwargs` level that "warns" that using `n_jobs=1` is not optimal?

h-mayorquin:
I am slightly against the warning but I can't think on a better way to let them know : / 

Two suggestions if we go with the warning:
* It should not be in a very general place as some sorters don't accept that and then it becomes and irrelevant warning.
* There should be a way to deactivate it that does not require you to do multiprocessing. As in, yes, I am aware but no, I don't want to use multiprocesing option.

alejoe91:
@samuelgarcia suggested a global flag `global_kwargs_set` that by session is set to True if the user calls the `set_global_job_kwargs`. I think it's a good idea :)

src/spikeinterface/sorters/basesorter.py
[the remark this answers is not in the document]
zm711: Cool. I'm in the recording suite for the next couple days. If you are trying to push this with a deadline definitely just close this (or take over this one), otherwise I will do that in a couple days/if one of my experiments ends early.

zm711:
Thanks Alessio!

src/spikeinterface/core/job_tools.py
```diff
@@ -28,8 +27,9 @@
                     Total memory usage (e.g. "500M", "2G")
                 - chunk_duration : str or float or None
                     Chunk duration in s if float or with units if str (e.g. "1s", "500ms")
-            * n_jobs: int
-                Number of jobs to use. With -1 the number of jobs is the same as number of cores
+            * n_jobs: int | float
+                Number of jobs to use. With -1 the number of jobs is the same as number of cores.
+                Using a float between 0 and 1 will use that fraction of the total cores.
```
zm711, line 32: I'm wondering if we should also make clear that technically someone can use n_jobs > number of cores which can actually degrade performance by splitting off too many processes? I didn't actually realize this until I looked at #2817.

src/spikeinterface/core/job_tools.py
```diff
@@ -60,7 +60,7 @@
 
 
 def fix_job_kwargs(runtime_job_kwargs):
-    from .globals import get_global_job_kwargs
+    from .globals import get_global_job_kwargs, global_job_kwargs_set
```
samuelgarcia, line 63:
```suggestion
    from .globals import get_global_job_kwargs, is_set_global_job_kwargs_set
```

src/spikeinterface/core/job_tools.py
```diff
@@ -99,6 +99,15 @@ def fix_job_kwargs(runtime_job_kwargs):
 
     job_kwargs["n_jobs"] = max(n_jobs, 1)
 
+    if "n_jobs" not in runtime_job_kwargs and job_kwargs["n_jobs"] == 1 and not global_job_kwargs_set:
```
samuelgarcia, line 102:
```suggestion
    if "n_jobs" not in runtime_job_kwargs and job_kwargs["n_jobs"] == 1 and not is_set_global_job_kwargs_set():
```

src/spikeinterface/core/job_tools.py
samuelgarcia: The use of the function better because it ensure that it is a global.

samuelgarcia:
I am OK with the logic. I made a proposal.

src/spikeinterface/core/job_tools.py
zm711: Why? I would love to learn the difference here between checking with the function and just importing the boolean. I don't really use `global` myself.

alejoe91: I agree with Zach here. It's the same and it's super internal

samuelgarcia: because it ensure that this import is not masked in between by another variable with the same name.
globals in python are not super garanty unless you put global before.

samuelgarcia: globals corss moules is evill in python having a function interface is more secure.

zm711: Good to know. Thanks :)

alejoe91: done in last commit

src/spikeinterface/core/job_tools.py
alejoe91: done

src/spikeinterface/core/globals.py
```diff
@@ -130,6 +130,14 @@ def set_global_job_kwargs(**job_kwargs):
     global_job_kwargs_set = True
 
 
+def is_global_job_kwargs_set() -> bool:
+    """
+    Check is the global job kwargs have been manually set.
+    """
+    global global_job_kwargs_set
+    return global_job_kwargs_set
```
samuelgarcia, line 138: This was already existing a fiew line afte rthis one.

zm711: Fixed @samuelgarcia. But I added type hints!

src/spikeinterface/core/job_tools.py
h-mayorquin: Can you open an issue about this?

zm711: Done!

src/spikeinterface/core/job_tools.py
```diff
@@ -99,6 +99,15 @@ def fix_job_kwargs(runtime_job_kwargs):
...
 
+    if "n_jobs" not in runtime_job_kwargs and job_kwargs["n_jobs"] == 1 and not is_set_global_job_kwargs_set():
+        warnings.warn(
+            "`n_jobs` is not set so parallel processing is disabled! "
+            "To speed up computations, it is recommended to set n_jobs either "
+            "globally (with the `spikeinterface.set_global_job_kwargs()` function) or "
+            "locally (with the `n_jobs` argument). Use `spikeinterface.set_global_job_kwargs?` "
+            "for more information about job_kwargs."
+        )
+
```
samuelgarcia, line 110: et voilà.

samuelgarcia (approved):

PR #2684, opened 2024-04-09
Move "return_scaled" options global to SortingAnalyzer.

samuelgarcia:
Move "return_scaled" options global to SortingAnalyzer.

Having a unique and global option for this is less dangerous, for instance to compute snr which depends on 2 options.

PR #2717, opened 2024-04-12
Fix example broken by #2684

zm711:
Moving `returned_scaled` in #2684 broke the sorting_analyzer example file. 

Datalad is broken on my computer so I need to test this here really quick to make sure my fix worked.


zm711:
Well goodness is this fixed the example. Bad news it looks like there might be something down with datalad overall at the moment.....

alejoe91 (approved):

PR #2736, opened 2024-04-19
Fix performance issue for `aggregate_channels`

### LINKED ISSUE #2625: Sorting takes extremely long when sorting a four shank probe by property
Hi there! I'm trying to sort a four-shank Neuropixel recording with Kilosort4 but when I try to use `run_sorter_by_property` it takes 8 hours to sort one of the four shanks and an estimated 160 hours to recompute the spike templates. When I split the recording by shank and do `run_sorter` on a single shank it runs normally. I'm using a local installation of Kilosort4 which has access to the GPU (I checked). Any idea what might be going on? I had a related issue on the Kilosort4 github but it doesn't seem to be a Kilosort issue: https://github.com/MouseLand/Kilosort/issues/631

Setup:
NVIDIA RTX 4080
64 GB RAM
Ubuntu 20.04
kilosort 4.0.2
spikeinterface 0.100.2

Code:
```
            # Apply high-pass filter
            print('Applying high-pass filter.. ')
            rec_filtered = spre.highpass_filter(rec)
            
            # Correct for inter-sample phase shift
            print('Correcting for phase shift.. ')
            rec_shifted = spre.phase_shift(rec_filtered)
            
            # Detect and interpolate over bad channels
            print('Detecting and interpolating over bad channels.. ')
            bad_channel_ids, all_channels = spre.detect_bad_channels(rec_shifted)
            
            # If there are too many bad channels, skip the interpolation step
            prec_bad_ch = np.sum(all_channels == 'noise') / all_channels.shape[0]
            if prec_bad_ch < (1/3):
                rec_interpolated = spre.interpolate_bad_channels(rec_shifted, bad_channel_ids)
                print(f'{np.sum(all_channels == "noise")} ({prec_bad_ch*100:.0f}%) bad channels')
            else:
                rec_interpolated = rec_shifted
                print(f'{np.sum(all_channels == "noise")} ({prec_bad_ch*100:.0f}%) bad channels,',
                      'skipping the interpolation step')
            
            # If there are multiple shanks, do destriping per shank
            print('Destriping.. ')
            if np.unique(rec_interpolated.get_property('group')).shape[0] > 1:
                
                # Loop over shanks and do preprocessing per shank
                rec_split = rec_interpolated.split_by(property='group')
                rec_destriped = []
                for sh in range(len(rec_split)):
                    rec_destriped.append(spre.highpass_spatial_filter(rec_split[sh]))
                
                # Merge back together
                rec_final = si.aggregate_channels(rec_destriped)
                
                # Run spike sorting per shank 
                try:
                  print(f'\nStarting {split(probe_path)[-1]} spike sorting at {datetime.now().strftime("%H:%M")}')
                  sort = run_sorter_by_property(
                      sorter_name=settings_dict['SPIKE_SORTER'],
                      recording=rec_final,
                      grouping_property='group',
                      working_folder=join(probe_path, settings_dict['SPIKE_SORTER'] + id_str),
                      verbose=True,
                      docker_image=settings_dict['USE_DOCKER'],
                      **sorter_params)
                
                except Exception as err:
                    # Log error to disk
                    print(err)
                    logf = open(os.path.join(probe_path, 'error_log.txt'), 'w')
                    logf.write(str(err))
                    logf.close()
                    
                    # Continue with next recording
                    continue
```

alejoe91:
Fixes #2625

### DIFF
--- a/src/spikeinterface/core/channelsaggregationrecording.py
+++ b/src/spikeinterface/core/channelsaggregationrecording.py
@@ -127,7 +127,7 @@ class ChannelsAggregationRecordingSegment(BaseRecordingSegment):
         self,
         start_frame: int | None = None,
         end_frame: int | None = None,
-        channel_indices: list | None = None,
+        channel_indices: list | slice | None = None,
     ) -> np.ndarray:
         return_all_channels = False
         if channel_indices is None:
@@ -142,11 +142,17 @@ class ChannelsAggregationRecordingSegment(BaseRecordingSegment):
                 # in case channel_indices is slice, it has step 1
                 step = channel_indices.step if channel_indices.step is not None else 1
                 channel_indices = list(range(channel_indices.start, channel_indices.stop, step))
+            recording_id_channels_map = {}
             for channel_idx in channel_indices:
-                segment = self._parent_segments[self._channel_map[channel_idx]["recording_id"]]
+                recording_id = self._channel_map[channel_idx]["recording_id"]
                 channel_index_recording = self._channel_map[channel_idx]["channel_index"]
+                if recording_id not in recording_id_channels_map:
+                    recording_id_channels_map[recording_id] = []
+                recording_id_channels_map[recording_id].append(channel_index_recording)
+            for recording_id, channel_indices_recording in recording_id_channels_map.items():
+                segment = self._parent_segments[recording_id]
                 traces_recording = segment.get_traces(
-                    channel_indices=[channel_index_recording], start_frame=start_frame, end_frame=end_frame
+                    channel_indices=channel_indices_recording, start_frame=start_frame, end_frame=end_frame
                 )
                 traces.append(traces_recording)
         else:


src/spikeinterface/core/channelsaggregationrecording.py
```diff
@@ -142,11 +142,17 @@ def get_traces(
                 # in case channel_indices is slice, it has step 1
                 step = channel_indices.step if channel_indices.step is not None else 1
```
zm711, line 143: I'm confused by this. The if statement says if `channel_indices` is None then we check for a step if it is not None, but it has to be None, no?

zm711 (commented):
Looks good to me. Just a couple quick questions, but if you want to ignore them let me know and I can just approve it and we can think about them later. See my comment about the `channel_indices=None`.

My second question is in the else statement we iterate through all the segments (L158), but if we are using the recording_id above wouldn't it make sense that even if the user supplies channels we make sure those channels are in that segment or does that not generate a big enough slowdown to check? 

The actual changes in this PR make sense though. We just collect up all the channels we need and make one pass rather than go channel-by-channel.

src/spikeinterface/core/channelsaggregationrecording.py
zm711: Never mind I was missing the upper portion of the code. I was looking at the diff wrong: )

alejoe91:
> My second question is in the else statement we iterate through all the segments (L158), but if we are using the recording_id above wouldn't it make sense that even if the user supplies channels we make sure those channels are in that segment or does that not generate a big enough slowdown to check?

Yes this can be a bit misleading! In the else statement we just grab all channels because we check previously that we need to return all of them in the correct order, so no need to use the mapping.

PR #2737, opened 2024-04-19
Extend capabilities of plot_traces and spikes_on_traces

alejoe91:
- use `times` when present
- option to plot `events` on top of the traces (super useful to check artifacts, LFP responses, etc.)
- add widget for event navigation
- add checkbox to show/hide colorbar


src/spikeinterface/widgets/traces.py
```diff
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
+import time
```
alejoe91, line 3:
```suggestion
```

alejoe91:
Here's a screencast of the new features :)

[Screencast from 14-05-2024 17:03:22.webm](https://github.com/SpikeInterface/spikeinterface/assets/17097257/e5e1f938-1945-4ae3-91b9-13397993ba8d)

@samuelgarcia good to merge on my side!

src/spikeinterface/widgets/traces.py
```diff
@@ -114,6 +120,15 @@ def __init__(
...
             channel_locations = None
 
+        if not rec0.has_time_vector():
+            self.times = None
+            t_start = 0
+            t_end = rec0.get_duration(segment_index=segment_index)
+        else:
+            self.times = rec0.get_times(segment_index=segment_index)
+            t_start = self.times[0]
+            t_end = self.times[-1]
```
samuelgarcia, line 130: This will not work when you have multiple segment and you use ipywidget because you can switch segment with a dropdone menu
We should have a list instead.
Am I wrong ?
Did you try with multi segment and several t_start and or times vector ?

samuelgarcia: And more importantly we should not have attribute setting sin the init. Everything should be done with the data_plot.

src/spikeinterface/widgets/traces.py
```diff
@@ -203,6 +217,8 @@ def __init__(
             else:
                 raise TypeError(f"'clim' can be None, tuple, or dict! Unsupported type {type(clim)}")
 
+        self.cb = None
```
samuelgarcia, line 220: To support multiple backend we have to avoid self in the init.
Put this in the data_plot.

src/spikeinterface/widgets/traces.py
```diff
@@ -298,29 +323,33 @@ def plot_matplotlib(self, data_plot, **backend_kwargs):
...
-        from .utils_ipywidgets import (
-            check_ipywidget_backend,
-            # make_timeseries_controller,
-            # make_channel_controller,
-            # make_scale_controller,
-            TimeSlider,
-            ChannelSelector,
-            ScaleWidget,
-        )
+        from .utils_ipywidgets import check_ipywidget_backend, TimeSlider, ChannelSelector, ScaleWidget, EventSelector
```
samuelgarcia, line 346: Having attributes settings here is OK.

src/spikeinterface/widgets/traces.py
```diff
@@ -422,20 +472,31 @@ def _mode_changed(self, change=None):
         else:
             self._update_plot()
 
+    def _event_changed(self, change=None):
```
samuelgarcia, line 475: You should handle the segment here.

src/spikeinterface/widgets/traces.py
```diff
@@ -422,20 +472,31 @@ def _mode_changed(self, change=None):
...
-        #     order, _ = order_channels_by_depth(self.rec0, channel_ids)
-        # else:
-        #     order = None
-
         start_frame, end_frame, segment_index = self.time_slider.value
-        time_range = np.array([start_frame, end_frame]) / self.rec0.sampling_frequency
+        if self.times is not None:
+            time_range = np.array([self.times[start_frame], self.times[end_frame]])
+        else:
+            time_range = np.array([start_frame, end_frame]) / self.rec0.sampling_frequency
```
samuelgarcia, line 490: Same here you should handle the segment

samuelgarcia:
This is cool.
I made a few comments camarade.
We need to handle the multisegment times.
And also we must totally avoid the self.attr = value in the init.
Every thing must go into the data_plot.

alejoe91:
@samuelgarcia fixed now for multi-segments!

[Screencast from 16-05-2024 15:40:44.webm](https://github.com/SpikeInterface/spikeinterface/assets/17097257/19dbdefd-b14c-43c3-a4d1-19e70ff9912d)

src/spikeinterface/widgets/traces.py
```diff
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
+from math import e
```
samuelgarcia, line 3:
```suggestion

```

PR #2744, opened 2024-04-23
Regularize whitening

yger:
Add an option to use regularized whitening


DradeAW:
@samuelgarcia @alejoe91 It would be great if this could be merged before the 'rc' release :)

src/spikeinterface/preprocessing/whiten.py
```diff
@@ -162,7 +187,7 @@ def compute_whitening_matrix(recording, mode, random_chunk_kwargs, apply_mean, r
 
     """
     random_data = get_random_data_chunks(recording, concatenated=True, return_scaled=False, **random_chunk_kwargs)
-    random_data = random_data.astype("float32")
+    random_data = random_data
```
samuelgarcia, line 190: Why did you removed this ? When data is integer we we float no ?

src/spikeinterface/preprocessing/whiten.py
```diff
@@ -40,6 +41,12 @@ class WhitenRecording(BasePreprocessor):
     M : 1d np.array or None, default: None
         Pre-computed means.
         M can be None when previously computed with apply_mean=False
+    regularize: bool, default False
+        Boolean to decide if we want to regularize the covariance matrix, using the GraphicalLassoCV() method
+        of sklearn
+    regularize_kwargs: None or dict
+        Dictionary of the parameters that could be provided to the GraphicalLassoCV() method of sklearn, if
+        the covariance matrix needs to be regularized
```
alejoe91, line 49:
```suggestion
    regularize : bool, default: False
        Boolean to decide if we want to regularize the covariance matrix, using the GraphicalLassoCV() method
        of sklearn
    regularize_kwargs : None or dict
        Dictionary of the parameters that could be provided to the GraphicalLassoCV() method of sklearn, if
        the covariance matrix needs to be regularized
```

Let's stick to the correct docstring formatting! Otherwise @chrishalcrow will be very mad!

src/spikeinterface/preprocessing/whiten.py
```diff
@@ -172,8 +197,26 @@ def compute_whitening_matrix(recording, mode, random_chunk_kwargs, apply_mean, r
...
+        if regularize_kwargs is None:
+            regularize_kwargs = {}
+        regularize_kwargs["assume_centered"] = True
+        job_kwargs = get_global_job_kwargs()
+        if "n_jobs" in job_kwargs and "n_jobs" not in regularize_kwargs:
+            n_jobs = job_kwargs["n_jobs"]
+            if isinstance(n_jobs, float) and 0 < n_jobs <= 1:
+                import os
+
+                n_jobs = int(n_jobs * os.cpu_count())
```
samuelgarcia, line 215: I am not sure to like this.
The n_jobs semantics is for the spikeinetrfac by chunk not for any subpackage that would use parralisation.
So for me lets the user control this with regularize_kwargs.

src/spikeinterface/preprocessing/whiten.py
```diff
@@ -55,6 +62,8 @@ def __init__(
         recording,
         dtype=None,
         apply_mean=False,
+        regularize=False,
+        regularize_kwargs={},
```
alejoe91, line 66:
```suggestion
        regularize_kwargs=None,
```
And in the init use:
```
regularize_kwargs = regularize_kwargs is regularize_kwargs is not None else {}
```

src/spikeinterface/preprocessing/whiten.py
```diff
@@ -172,8 +197,26 @@ def compute_whitening_matrix(recording, mode, random_chunk_kwargs, apply_mean, r
...
+        regularize_kwargs["assume_centered"] = True
+        job_kwargs = get_global_job_kwargs()
+        if "n_jobs" in job_kwargs and "n_jobs" not in regularize_kwargs:
+            n_jobs = job_kwargs["n_jobs"]
+            if isinstance(n_jobs, float) and 0 < n_jobs <= 1:
+                import os
+
+                n_jobs = int(n_jobs * os.cpu_count())
+            regularize_kwargs["n_jobs"] = n_jobs
+        estimator = sklearn.covariance.GraphicalLassoCV(**regularize_kwargs)
```
samuelgarcia, line 217: maybe we could have a regularize method that can use lasso but maybe others no ?

src/spikeinterface/preprocessing/whiten.py
alejoe91: I think it's ok, but we can then remove the line:
```suggestion
```

src/spikeinterface/preprocessing/whiten.py
yger: ok, I'll remove this

src/spikeinterface/preprocessing/whiten.py
yger: We could, but I would say let's start with one...

yger: I've tried something... We can pass method: "GraphicalLassoCV" or something else, and the appropriate node is created

src/spikeinterface/preprocessing/tests/test_whiten.py
```diff
@@ -41,6 +41,14 @@ def test_whiten(create_cache_folder):
     assert rec4.get_dtype() == "int16"
     assert rec4._kwargs["M"] is None
 
+    # test regularization
+    with pytest.raises(AssertionError):
```
samuelgarcia, line 45: I do not see in the code wich assert is raised for this.
We should have better error than assert in this no ?

src/spikeinterface/preprocessing/tests/test_whiten.py
```diff
@@ -41,6 +41,14 @@ def test_whiten(create_cache_folder):
...
+        W, M = compute_whitening_matrix(
+            rec, "local", random_chunk_kwargs, apply_mean=False, radius_um=None, regularize=True
+        )
+    # W must be sparse
```
samuelgarcia, line 49: which W are you testing here ?

src/spikeinterface/preprocessing/tests/test_whiten.py
yger: I've used what was already done, open to suggestions

src/spikeinterface/preprocessing/tests/test_whiten.py
yger: I'm just computing W with the default regularization method

samuelgarcia (approved):

samuelgarcia:
Merci.

PR #2750, opened 2024-04-24
Template amplitude should not be returned as absolute value

yger:
It seems more logical to me that get_template_amplitude returns the real value, and not an absolute value.


yger:
The default behaviour is kept, and the new one (more logical ?) is only activated with an extra boolean abs_values

alejoe91 (approved):

PR #2757, opened 2024-04-25
Remove randomness of order in _get_children_dependencies

### LINKED ISSUE #2745: Delete_extension on recompute
So we were testing recomputing a full set of analysis. I wonder if this a windows error. Basically to recreate:

```python

analyzer = si.create_sorting_analyzer(sorting, rec_cmr, format='binary_folder', folder='path')
analyzer.compute(['random_spikes', 'waveforms', 'templates', 'noise_levels', 'spike_amplitudes', 'spike_locations'])
analyzer.compute(['random_spikes', 'waveforms', 'templates', 'noise_levels', 'spike_amplitudes', 'spike_locations'])
```

Error:

```python
ValueError: SortingAnalyzer need extension 'templates' to be computed
```

The extensions that are left behind based on the repr is

```
SortingAnalyzer: 128 channels - 4 units - 1 segments - binary_folder - sparse - has recording
Loaded 6 extensions: random_spikes, noise_levels, correlograms, spike_amplitudes, spike_locations, principal_components
```

I wonder if it the delete extension is missing some of the children because it is deleting in the wrong order? Or maybe because templates can depend on either random_spikes or waveforms the delete_extension on recompute is now not working as expected. Or of course It is a Windows error.....

full stack trace in case we need it

```python
  Cell In[36], line 1
    analyzer.compute('random_spikes')
  File ~\User\GitHub\spikeinterface\src\spikeinterface\core\sortinganalyzer.py:853 in compute
    return self.compute_one_extension(extension_name=input, save=save, **kwargs)
  File ~\UserH\Documents\GitHub\spikeinterface\src\spikeinterface\core\sortinganalyzer.py:909 in compute_one_extension
    self.delete_extension(child)
  File ~\User\GitHub\spikeinterface\src\spikeinterface\core\sortinganalyzer.py:1113 in delete_extension
    ext = self.load_extension(extension_name)
  File ~\Users\GitHub\spikeinterface\src\spikeinterface\core\sortinganalyzer.py:1090 in load_extension
    extension_instance = extension_class(self)
  File ~\User\GitHub\spikeinterface\src\spikeinterface\postprocessing\spike_locations.py:62 in __init__
    extremum_channel_inds = get_template_extremum_channel(self.sorting_analyzer, outputs="index")
  File ~\User\GitHub\spikeinterface\src\spikeinterface\core\template_tools.py:150 in get_template_extremum_channel
    peak_values = get_template_amplitudes(templates_or_sorting_analyzer, peak_sign=peak_sign, mode=mode)
  File ~\User\Documents\GitHub\spikeinterface\src\spikeinterface\core\template_tools.py:86 in get_template_amplitudes
    before = _get_nbefore(templates_or_sorting_analyzer)
  File ~\User\Documents\GitHub\spikeinterface\src\spikeinterface\core\template_tools.py:50 in _get_nbefore
    raise ValueError("SortingAnalyzer need extension 'templates' to be computed")
```

We are 11 commits behind main for this analysis, but there weren't any analyzer changes in those commits so this should be a current issue on Main.

chrishalcrow:
Fixes #2745 , bug discovered and solution discovered by @zm711 

The function `_get_children_dependencies` randomised the list of children, through the `set` function. This is a problem when using this function to delete extensions, since this has an ordering to it. Solution is simply to delete `set`, which didn't do anything (as far as I can tell, from playing with more complicated dependencies in the postprocessing module).

Added test to check that ordering is correct for the core extensions. Note: the test randomly passes or fails randomly for past versions. Test always passes now 😎

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -1187,7 +1187,7 @@ def _get_children_dependencies(extension_name):
             names.append(child)
         grand_children = _get_children_dependencies(child)
         names.extend(grand_children)
-    return list(set(names))
+    return list(names)
 
 
 def register_result_extension(extension_class):
--- a/src/spikeinterface/core/tests/test_analyzer_extension_core.py
+++ b/src/spikeinterface/core/tests/test_analyzer_extension_core.py
@@ -205,9 +205,13 @@ def test_ComputeNoiseLevels(format, sparse):
 def test_get_children_dependencies():
     assert "waveforms" in _extension_children["random_spikes"]
 
-    children = _get_children_dependencies("random_spikes")
-    assert "waveforms" in children
-    assert "templates" in children
+    rs_children = _get_children_dependencies("random_spikes")
+    assert "waveforms" in rs_children
+    assert "templates" in rs_children
+
+    assert rs_children.index('waveforms') < rs_children.index('templates')
+
+
 
 
 def test_delete_on_recompute():
--- a/src/spikeinterface/core/tests/test_analyzer_extension_core.py
+++ b/src/spikeinterface/core/tests/test_analyzer_extension_core.py
@@ -209,9 +209,7 @@ def test_get_children_dependencies():
     assert "waveforms" in rs_children
     assert "templates" in rs_children
 
-    assert rs_children.index('waveforms') < rs_children.index('templates')
-
-
+    assert rs_children.index("waveforms") < rs_children.index("templates")
 
 
 def test_delete_on_recompute():


alejoe91 (approved):

samuelgarcia:
Hi.
Thanks you for looking at this.
For me the set was to remove when a extension name was added by several time by several path ways.
It was more clean.
I did not figure out that the order was destroyed. thank a lot for catching this.
Maybe th np.unique() could make it better.

zm711:
@samuelgarcia , `np.unique` as far as my reading also doesn't preserve order so if we wanted to do this we would have to add a sort function after unique and then profile to see if getting the unique + sorting is faster than just doing the boolean check for the extra times each extension is listed.

PR #2767, opened 2024-04-26
Speed up test_sorting_s3_nwb_zarr

chrishalcrow:
Speed up a very slow test, by only checking some units, cutting the total test run from 33->28 minutes!

A preview of the Great-Test-Speed-Up of 2024: https://github.com/SpikeInterface/SpikeInteface-Hackathon-Edinburgh-May24/issues/8
(This seemed too easy and impactful to wait until the hackathon)

The function `test_sorting_s3_nwb_zarr` downloads a sorting object from s3 and checks some things about it. This has traditionally taken a while because the object is pretty big and has ~450 units. Most of the testing time was spent looping over these 450 units and calculating their spike trains. This is probably overkill, so I propose to only check a random selection of 3/450 units.

You can see the overall speed-up here, from my github page (I realise these are failing, but they are attempting all the tests!!):

Screenshot 2024-04-26 at 12 51 09
(from https://github.com/chrishalcrow/spikeinterface/actions )

### DIFF
--- a/src/spikeinterface/extractors/tests/test_nwbextractors_streaming.py
+++ b/src/spikeinterface/extractors/tests/test_nwbextractors_streaming.py
@@ -290,7 +290,15 @@ def test_sorting_s3_nwb_zarr(tmp_path):
 
     # test to/from dict
     sorting_loaded = load_extractor(sorting.to_dict())
-    check_sortings_equal(sorting, sorting_loaded)
+
+    # just take 3 random units to test
+    rng = np.random.default_rng(seed=2205)
+    three_unit_ids = rng.choice(sorting.unit_ids, size=3)
+    sorting_sub = sorting.select_units(unit_ids = three_unit_ids)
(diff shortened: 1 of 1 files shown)


src/spikeinterface/extractors/tests/test_nwbextractors_streaming.py
```diff
@@ -290,7 +290,14 @@ def test_sorting_s3_nwb_zarr(tmp_path):
 
     # test to/from dict
     sorting_loaded = load_extractor(sorting.to_dict())
-    check_sortings_equal(sorting, sorting_loaded)
+
+    # just take 3 random units to test
+    rng = np.random.default_rng(seed=2205)
```
samuelgarcia, line 295: This seed is my brithday :)
You can put your own one if you want!

chrishalcrow: I'll celebrate your birthday every time I do a randomised test ;)

PR #2769, opened 2024-04-26
Add template `generate_recording_from_template_database`

h-mayorquin:
Thisi s a request for the hybrid project to fetch templates from the s3 database and generate recordings. I addded a pure function that allows to fetch teh template data and a convinence function that wraps this and generates a recording.

How to use:

```python
from spikeinterface.generation import generate_recording_from_template_database

selected_unit_inidces = [0, 23, 44 , 55]
durations = [1.0]

#from spikeinterface.hybrid import generate_recording_from_template_database
selected_unit_inidces  = [0, 1, 2, 3] # Here value from selectedUnitIndices
durations = [1.0] # To be determined by the user
dataset = "test_templates" # The dataset name
recording = generate_recording_from_template_database(selected_unit_inidces, durations=durations, dataset=dataset)
recording
```

Right now I only have `test_templates` as a dataset. This comes from real data but I have not done the whole batch because I am getting a memory leak. I need to test this.

### DIFF
--- a/src/spikeinterface/generation/__init__.py
+++ b/src/spikeinterface/generation/__init__.py
@@ -11,3 +11,8 @@ from .drifting_generator import (
     generate_displacement_vector,
     generate_drifting_recording,
 )
+
+from ._template_database import (
+    fetch_templates_from_database,
+    generate_recording_from_template_database,
+)
--- a/src/spikeinterface/generation/_template_database.py
+++ b/src/spikeinterface/generation/_template_database.py
@@ -1,0 +1,47 @@
+from spikeinterface.core.template import Templates
+from spikeinterface.core import generate_sorting, InjectTemplatesRecording
+import zarr
+
+
+def fetch_templates_from_database(dataset="test_templates"):
+
+    import s3fs
+
+    s3 = s3fs.S3FileSystem(anon=False, client_kwargs={"region_name": "us-east-2"})
+
+    # Specify the S3 bucket and path where your Zarr dataset is stored
+    store = s3fs.S3Map(root=f"spikeinterface-template-database/{dataset}", s3=s3)
+
+    # Load the Zarr group from S3
+    zarr_group = zarr.open(store, mode="r")
+
+    templates_object = Templates.from_zarr_group(zarr_group)
+
+    return templates_object
+
+
+def generate_recording_from_template_database(selected_unit_inidces=None, dataset="test_templates", durations=None):
+
+    durations = durations or [10.0]
+
+    templates_object = fetch_templates_from_database(dataset=dataset)
+
+    if selected_unit_inidces:
+        selected_templates = templates_object.templates_array[selected_unit_inidces, :, :]
+    else:
+        selected_templates = templates_object.templates_array
+
+    num_units = selected_templates.shape[0]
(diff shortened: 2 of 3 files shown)


samuelgarcia:
Hi Ramon. Cool thanks.
I will do some comments in the code.

src/spikeinterface/generation/__init__.py
```diff
@@ -11,3 +11,8 @@
     generate_displacement_vector,
     generate_drifting_recording,
 )
+
+from ._template_database import (
```
samuelgarcia, line 15: I would remove the "_" from the file name.

src/spikeinterface/generation/_template_database.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # Load the Zarr group from S3
+    zarr_group = zarr.open(store, mode="r")
+
+    templates_object = Templates.from_zarr_group(zarr_group)
+
+    return templates_object
+
+
+def generate_recording_from_template_database(selected_unit_inidces=None, dataset="test_templates", durations=None):
```
samuelgarcia, line 23: For we should be coordinate with a call.

I started a drift generator with synthetic templates here : https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/generation/drifting_generator.py#L249

In my mind this function should be extended or refactored to accept real templates then the sorting code and the drift code could be shared.
Then this function could be a simple wrapper on top on the other function that would download template and call the other one.

samuelgarcia:
We should also make a small doc to explain the template library, its URL and make some screen shot.
No ?

src/spikeinterface/generation/_template_database.py
h-mayorquin: Sure, maybe tomorrow after our meeting?

src/spikeinterface/generation/__init__.py
h-mayorquin: OK, are you fine with the location of the file?

h-mayorquin:
> We should also make a small doc to explain the template library, its URL and make some screen shot. No ?

Sure. But where and of what do you want the screenshoot? I migrated all the code to the organization btw:

https://github.com/SpikeInterface/hybrid_template_library

src/spikeinterface/generation/__init__.py
h-mayorquin: The changed.

src/spikeinterface/generation/template_database.py
```diff
@@ -0,0 +1,12 @@
+from spikeinterface.core.template import Templates
+import zarr
+
+
+def fetch_templates_from_database(dataset="test_templates") -> Templates:
+
+    s3_path = f"s3://spikeinterface-template-database/{dataset}/"
+    zarr_group = zarr.open_consolidated(s3_path, storage_options={"anon": False})
```
alejoe91, line 8:
```suggestion
    zarr_group = zarr.open_consolidated(s3_path, storage_options={"anon": True})
```

h-mayorquin: Thanks!

alejoe91 (approved):

PR #2773, opened 2024-04-29
Compute quality metrics after pipeline nodes

### LINKED ISSUE #2772: SortingAnalyzer doesn't respect order of compute in case of nodepipeline + non-nodepipeline code
## MRE

```python
analyzer.compute(['random_spikes', 'waveforms', 'templates' , 'noise_levels', 'spike_locations', 'spike_amplitudes, 'principal_components', 'quality_metrics'])
```
## Problem
It will calculate the non-node pipeline first despite the fact that to do all quality_metrics requires things like `spike_amplitudes`. 

## Solution
I know in the code it separates out the code into the two streams of pipeline and nonpipeline, so maybe we just need to document that this will happen? I mean we could also try to have something more intelligent that sorts this out, but with the current state I don't know think there is an easy solution.

alejoe91:
Fixes #2772

@zm711 this is a temporary solution that works for the time being. We can look for a more flexible and long-term solution later

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -968,7 +968,13 @@ class SortingAnalyzer:
 
         extensions_with_pipeline = {}
         extensions_without_pipeline = {}
+        extensions_post_pipeline = {}
         for extension_name, extension_params in extensions.items():
+            if extension_name == "quality_metrics":
+                # PATCH: the quality metric is computed after the pipeline, since some of the metrics optionally require
+                # the output of the pipeline extensions (e.g., spike_amplitudes, spike_locations).
+                extensions_post_pipeline[extension_name] = extension_params
+                continue
             extension_class = get_extension_class(extension_name)
             if extension_class.use_nodepipeline:
                 extensions_with_pipeline[extension_name] = extension_params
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1020,6 +1026,17 @@ def compute_several_extensions(self, extensions, save=True, **job_kwargs):
                 if save:
                     extension_instance.save()
 
+        # PATCH: the quality metric is computed after the pipeline, since some of the metrics optionally require
+        # the output of the pipeline extensions (e.g., spike_amplitudes, spike_locations).
+        # An alternative could be to extend the "depend_on" attribute to use optional and to check if an extension
+        # depends on the output of the pipeline nodes (e.g. depend_on=["spike_amplitudes[optional]"])
```
samuelgarcia, line 1032: perfect

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1173,7 +1190,7 @@ def _get_children_dependencies(extension_name):
     This function is making the reverse way : get all children that depend of a
     particular extension.
 
-    This is recurssive so this includes : children and so grand children and grand grand children
+    This is recursive so this includes : children and so grand children and grand grand children
```
zm711, line 1193:
```suggestion
    This is recursive so this includes : children and so grand children and great grand children
```

If we are fixing for some reason in English we switch to great after the first grand. So then you just add more greats rather than more grands. Don't know know why :)

zm711 (commented):
Looks good to me. I'll run one quick test when I get to lab :)

zm711:
Tested and this worked for me locally.

samuelgarcia (approved):

PR #2774, opened 2024-04-29
Fix KS4 for v>=4.0.5 and simplify skip correction

### LINKED ISSUE #2742: Fix Kilosort4Sorter
In spikeinterface/sorters/external/kilosort4.py, the function save_sorting is imported from kilosort.run_kilosort.
The function returns the variables 

> ops, similar_templates, is_ref, est_contam_rate

 on line 271.
This worked fine up to kilosort 4.0.4, and in kilosort 4.0.5 the function was changed to return an extra variable:

> ops, similar_templates, is_ref, est_contam_rate, kept_spikes

Which will raise a ValueError on line 271 when save_sorting is called.

alejoe91:
Fixes #2742

### DIFF
--- a/src/spikeinterface/sorters/external/kilosort4.py
+++ b/src/spikeinterface/sorters/external/kilosort4.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from pathlib import Path
 from typing import Union
-from packaging.version import parse
 
 from ..basesorter import BaseSorter
 from .kilosortbase import KilosortBase
--- a/src/spikeinterface/sorters/external/kilosort4.py
+++ b/src/spikeinterface/sorters/external/kilosort4.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
 from pathlib import Path
-import os
 from typing import Union
+from packaging.version import parse
 
 from ..basesorter import BaseSorter
 from .kilosortbase import KilosortBase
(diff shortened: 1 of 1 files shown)


src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
 from pathlib import Path
-import os
 from typing import Union
+from packaging.version import parse
```
zm711, line 5: Where are we using parse? I don't see it in the diff. Maybe the original idea was to treat different versions differently?

src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -227,39 +227,14 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
...
-                device=device,
-                do_CAR=do_CAR,
-                invert_sign=invert,
-                dtype=dtype,
-                tmin=tmin,
-                tmax=tmax,
-                artifact_threshold=artifact,
-                file_object=file_object,
-            )
+            ops["nblocks"] = 0
```
zm711, line 232: This is the KS based solution for skipping drift correction right? 

I think it does make things simpler to just rely on their machinery. Maybe this will speed things up too from the wrapper?

src/spikeinterface/sorters/external/kilosort4.py
zm711, line 5:
```suggestion
```

zm711 (approved):
And I tested this locally both with and without drift correction ran fine on KS 4.0.6

Removal of parse maybe? But otherwise looks good to me.

src/spikeinterface/sorters/external/kilosort4.py
alejoe91: yep indeed that was the original idea, removing :)

src/spikeinterface/sorters/external/kilosort4.py
alejoe91: yes, let's see :)

PR #2775, opened 2024-04-29
Fix split in more than 2 units and extend curation docs and tests

### LINKED ISSUE #2694: Split function and `new_unit_ids` definition

alejoe91:
Fixes #2694 and replaces #2695 (by @OlivierPeron)

@OlivierPeron when making aPR it's better to create a branch different than main so we can directly contribute to it ;)

### DIFF
--- a/src/spikeinterface/curation/curationsorting.py
+++ b/src/spikeinterface/curation/curationsorting.py
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
 from collections import namedtuple
+from collections.abc import Iterable
+from hmac import new
+
 import numpy as np
 
 from .mergeunitssorting import MergeUnitsSorting
@@ -59,9 +62,29 @@ class CurationSorting:
             ids = [str(i) for i in ids]
         return ids
 
-    def split(self, split_unit_id, indices_list):
+    def split(self, split_unit_id, indices_list, new_unit_ids=None):
+        """
+        Split a unit into multiple units.
+
+        Parameters
+        ----------
+        split_unit_id: int
+            The unit to split
+        indices_list: list or np.array
+            A list of index arrays selecting the spikes to split in each segment.
+            Each array can contain more than 2 indices (e.g. for splitting in 3 or more units) and it should
+            be the same length as the spike train (for each segment).
+            If the sorting has only one segment, indices_list can be a single array
+        new_unit_ids: list ot None
+            List of new unit ids. If None, a new unit id is automatically selected
+        """
         current_sorting = self._sorting_stages[self._sorting_stages_i]
-        new_unit_ids = self._get_unused_id(2)
+        if not isinstance(indices_list, list):
+            indices_list = [indices_list]
+        if not isinstance(indices_list[0], Iterable):
+            raise ValueError("indices_list must be a list of iterable arrays")
+        if new_unit_ids is None:
+            new_unit_ids = self._get_unused_id(np.max([len(np.unique(v)) for v in indices_list]))
         new_sorting = SplitUnitSorting(
             current_sorting,
             split_unit_id=split_unit_id,
@@ -81,6 +104,18 @@ class CurationSorting:
         self._add_new_stage(new_sorting, edges)
 
     def merge(self, units_to_merge, new_unit_id=None, delta_time_ms=0.4):
+        """
+        Merge a list of units into a new unit.
+
+        Parameters
+        ----------
+        units_to_merge: list
+            List of unit ids to merge
+        new_unit_id: int
+            The new unit id. If None, a new unit id is automatically selected
+        delta_time_ms: float
+            Number of ms to consider for duplicated spikes. None won't check for duplications
+        """
         current_sorting = self._sorting_stages[self._sorting_stages_i]
         if new_unit_id is None:
             new_unit_id = self._get_unused_id()[0]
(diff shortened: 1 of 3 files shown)


src/spikeinterface/curation/curationsorting.py
```diff
@@ -81,6 +104,18 @@ def split(self, split_unit_id, indices_list):
...
 
     def merge(self, units_to_merge, new_unit_id=None, delta_time_ms=0.4):
+        """
+        Merge a list of units into a new unit.
+
+        Parameters
+        ----------
+        units_to_merge: list
+            List of unit ids to merge
+        new_unit_id: int
```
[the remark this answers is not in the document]
alejoe91: yep, they can! Thanks for catchig this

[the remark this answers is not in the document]
alejoe91: yep, they can! Thanks for catchig this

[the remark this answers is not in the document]
alejoe91: done

src/spikeinterface/curation/curationsorting.py
```diff
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
 from collections import namedtuple
+from collections.abc import Iterable
+from hmac import new
```
zm711, line 5: What is hmac? I've never heard of this?

alejoe91:
@zm711 is this good to merge? I think that the docstring improvement might fix #2816

src/spikeinterface/curation/curationsorting.py
```diff
@@ -59,9 +61,29 @@ def _get_unused_id(self, n=1):
...
         return ids
 
-    def split(self, split_unit_id, indices_list):
+    def split(self, split_unit_id, indices_list, new_unit_ids=None):
+        """
+        Split a unit into multiple units.
+
+        Parameters
+        ----------
+        split_unit_id: int
```
zm711, line 70: can this also be string? since the ids can be strings or ints?

zm711 (approved):
Yep looks good with just one more question @alejoe91,  and then style-wise we could do 
list[str | int] but I don't think that is absolutely necessary for the docstrings, but might make it a bit clearer that ids can still be ints or strings.

src/spikeinterface/curation/curationsorting.py
alejoe91, line 70:
```suggestion
        split_unit_id: int or str
```

alejoe91: done, but this is not a list so Iused `int or str`

alejoe91: done

src/spikeinterface/curation/splitunitsorting.py
```diff
@@ -36,10 +37,20 @@ class SplitUnitSorting(BaseSorting):
...
+        split_unit_indices = np.unique([np.unique(v) for v in indices_list])
+        tot_splits = len(split_unit_indices)
         unchanged_units = parents_unit_ids[parents_unit_ids != split_unit_id]
 
+        # make sure indices list is between 0 and tot_splits - 1
+        indices_zero_based = [np.zeros_like(indices) for indices in indices_list]
+        for segment_index in range(len(indices_list)):
+            for i, split_unit_idx in enumerate(split_unit_indices):
+                indices_zero_based[segment_index][indices_list[segment_index] == split_unit_idx] = i
+
```
zm711, line 53: Could we potentially give `i` a better name. In the future I think it will be better to have the counter with the name of what it is rather than just being the counter. To be honest this double for is hard for me to parse because of the i.

src/spikeinterface/curation/splitunitsorting.py
```diff
@@ -36,10 +37,20 @@ class SplitUnitSorting(BaseSorting):
...
+        ), "The length of indices_list must be the same as parent_sorting.get_num_segments"
+        split_unit_indices = np.unique([np.unique(v) for v in indices_list])
+        tot_splits = len(split_unit_indices)
         unchanged_units = parents_unit_ids[parents_unit_ids != split_unit_id]
 
+        # make sure indices list is between 0 and tot_splits - 1
+        indices_zero_based = [np.zeros_like(indices) for indices in indices_list]
+        for segment_index in range(len(indices_list)):
+            for i, split_unit_idx in enumerate(split_unit_indices):
+                indices_zero_based[segment_index][indices_list[segment_index] == split_unit_idx] = i
```
alejoe91, line 52:
```suggestion
            for zero_based_index, split_unit_idx in enumerate(split_unit_indices):
                indices_zero_based[segment_index][indices_list[segment_index] == split_unit_idx] = zero_based_index
```
@zm711 better?

zm711: Yeah I think that explains the point of doing the counter! So It will be clear in the future. Thanks @alejoe91 !

PR #2777, opened 2024-04-29
Fix codecov testing

alejoe91:
guys this took a lot to debug....

Another reason to advocate not to use mutable objects in functions!!! @samuelgarcia @chrishalcrow @h-mayorquin @zm711 @JoeZiminski 

The problem was that the `ComputeTemplates.set_params` took `operators=["average", "std"]` as defaults, and these were propagated to the params dict without copy.
The `export_reports` then calls the `plot_unit_templates`, which asks for percentiles (which are present there) and somehow these get globally extended to the default kwargs...

So no more mutable arguments from now on!!!

### DIFF
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -286,8 +286,7 @@ class ComputeTemplates(AnalyzerExtension):
     need_job_kwargs = True
 
     def _set_params(self, ms_before: float = 1.0, ms_after: float = 2.0, operators=None):
-        if operators is None:
-            operators = ["average", "std"]
+        operators = operators or ["average", "std"]
         assert isinstance(operators, list)
         for operator in operators:
             if isinstance(operator, str):


src/spikeinterface/widgets/unit_waveforms.py
```diff
@@ -95,7 +95,7 @@ def __init__(
         set_title=True,
         same_axis=False,
         shade_templates=True,
-        templates_percentile_shading=[1, 25, 75, 98],
+        templates_percentile_shading=[1, 25, 75, 99],
```
zm711, line 98: Wouldn't it be better to have this as a tuple then for the default. The end-user can give a list, but we have the default be immutable.

zm711 (commented):
Sounds good to me. I think Heberto started moving some of this over no? Was that PR ever merged....

src/spikeinterface/widgets/unit_waveforms.py
alejoe91: done

chrishalcrow:
Hello,
Just trying to make sure I understand (being a python newbie). A basic example which does something similar would be

```
def a_function(parms=['a','b']):
    a_dict = dict( parms = parms)
    return a_dict

a_dict = a_function()
a_dict['parms'] += ['c']   **
b_dict = a_function()
print(b_dict)
>>> {'parms': ['a', 'b', 'c']}
```

Python only evaluates parms the first time `a_fuction` is called. Then the line with two **s accesses parms directly and mutates it. So when we run the function again the mutated parms are taken instead of ['a', 'b'] (which we may naively expect).

I think the place in our code where we mutate must be here: https://github.com/SpikeInterface/spikeinterface/blob/256757c628c3c8027453f13358067436254e01cb/src/spikeinterface/core/analyzer_extension_core.py#L468 ?? Does that make sense? And do these lines set off any python-anti-pattern alarm bells?

chrishalcrow (approved):
Excellent detective work Alessio! And I think I understand what's going on... 👍

chrishalcrow:
> In general is is a bad idea to change state or rely on the order of execution as an implicit way of defining the flow. You see the guys from scientific python defending this idea here:
> https://learn.scientific-python.org/development/principles/design/#avoid-changing-state

Thanks for the explanation and this link -- that whole site is very informative!

samuelgarcia:
> Regarding how to handle it, I have seen three ways to handle this:
> 
> 1. Use an expression inside the code like that correction of Sam above.
> 2. Make local copies of all the mutable options at the beginning of the function.
> 3. Use non-mutable options like the Zach comment above (but difficult for other structures like dict).

I prefer the version 2. over the 1.

The 3. is the best but only works with tuple that replace list for dict it is more tricky.

Issue #2766, opened 2024-04-26
detect_bad_channels multi probe recording


RobertoDF:
Hi,

I run detect_bad_channels first on a recording that includes two probes: 
`bad_channel_ids, channel_labels = detect_bad_channels(raw_rec, dead_channel_threshold=0.0001)`

and after on the two split recordings:
```
split_preprocessed_recording = raw_rec.split_by("group")
bad_channel_ids_1, channel_labels_1 = detect_bad_channels(split_preprocessed_recording[0], dead_channel_threshold=0.0001)
bad_channel_ids_2, channel_labels_2 = detect_bad_channels(split_preprocessed_recording[1],    dead_channel_threshold=0.0001)
```
the results are different. So it looks the function expects one probe at the time. Is that right? If yes I would add a disclaimer in the documentation.

Thanks,

PR #2783, opened 2024-04-29
Update detect_bad_channels docs

RobertoDF:
Clarify that detect_bad_channels doesn´t automatically deal with multi probe recordings  #2766

### DIFF
--- a/doc/modules/preprocessing.rst
+++ b/doc/modules/preprocessing.rst
@@ -230,7 +230,7 @@ The :code:`detect_bad_channels()` can be used to detect bad channels with severa
 approach to detect bad channels with abnormally high power and the :code:`coherence+psd` method (introduced by [IBL_spikesorting]_),
 which detects bad channels looking at both coherence with other channels and PSD power in the high-frequency range.
 
-Note: This function must be run one probe at a time if multiple probes are present in a recording (see `Processing a Recording by Channel Group <https://spikeinterface.readthedocs.io/en/latest/how_to/process_by_channel_group.html?highlight=split_by>`_).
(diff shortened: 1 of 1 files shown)


zm711:
I want to double check if this should really be on a per-shank basis (some probes are multishank) otherwise I think we may want to mention that if you have multi probes (e.g. in a probegroup). @alejoe91 @samuelgarcia ? 

Once I know which is more correct I'm happy to point out where exactly I would add the extra clarification. Otherwise I think this is a nice addition :)

zm711:
Fixes #2766
Also one tiny thing @RobertoDF if you put fixes xx like I did here it will auto link the issue and close it when the PR is merged. It helps with housekeeping :)

doc/modules/preprocessing.rst
```diff
@@ -230,6 +230,8 @@ The :code:`detect_bad_channels()` can be used to detect bad channels with severa
 approach to detect bad channels with abnormally high power and the :code:`coherence+psd` method (introduced by [IBL_spikesorting]_),
 which detects bad channels looking at both coherence with other channels and PSD power in the high-frequency range.
 
+Note: This function must be run one probe at a time if multiple probes are present in a recording (see `Processing a Recording by Channel Group <https://spikeinterface.readthedocs.io/en/latest/how_to/process_by_channel_group.html?highlight=split_by>`_).
```
alejoe91, line 233:
```suggestion
Note: The :code:`coherence+psd` method must be run on individual probes/shanks separately since it uses the coherence of the signal across the depth of the probe. See `Processing a Recording by Channel Group <https://spikeinterface.readthedocs.io/en/latest/how_to/process_by_channel_group.html?highlight=split_by>`_ for more information.
```

alejoe91: @zm711 @RobertoDF I think this is more accurate :)

alejoe91 (approved):

PR #2785, opened 2024-04-30
Analyzer set recording

alejoe91:
This is useful for example if the analyzer was set up and ran with a tmp cached recording, but later on one wants to set a lazy recording to perform additional operations and computations


src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -572,6 +572,11 @@ def load_from_zarr(cls, folder, recording=None):
 
         return sorting_analyzer
 
+    def set_recording(self, recording):
+        if self._recording is not None:
+            raise ValueError("Recording is already set")
```
alejoe91, line 577:
```suggestion
    def set_temporary_recording(self, recording):
        """
        Add docstrings
        """
        if self._recording is not None:
            raise warnings.warn("Recording is already set ...")
```

zm711:
I really like this idea because we often do the analysis and then move everything over to our server for long-term storage which means the recording-analyzer link gets broken. My only concern (and this is now from experience helping others) is that people forget what goes with what. So is there any way we can have a safety check? This is honestly the end-users responsibility not to mess up, but are there any attributes on the analyzer that could say help make sure the recording being temporarily used is reasonable?

alejoe91:
> I really like this idea because we often do the analysis and then move everything over to our server for long-term storage which means the recording-analyzer link gets broken. My only concern (and this is now from experience helping others) is that people forget what goes with what. So is there any way we can have a safety check? This is honestly the end-users responsibility not to mess up, but are there any attributes on the analyzer that could say help make sure the recording being temporarily used is reasonable?

I think that we could check that the recording has the correct recording attributes, i.e., channel_ids, locations, etc

src/spikeinterface/core/recording_tools.py
```diff
@@ -910,3 +910,35 @@ def get_rec_attributes(recording):
...
+
+def check_recording_attributes_match(recording1, recording2_attributes, skip_properties=True):
+    """
+    Check if two recordings have the same attributes
+
+    Parameters
+    ----------
+    recording1 : BaseRecording
+        The first recording object
+    recording2 : BaseRecording
```
zm711, line 923: docstring says recording2 but you actually just want the attributes right?

src/spikeinterface/core/recording_tools.py
```diff
@@ -910,3 +910,35 @@ def get_rec_attributes(recording):
         dtype=recording.get_dtype(),
     )
     return rec_attributes
+
+
+def check_recording_attributes_match(recording1, recording2_attributes, skip_properties=True):
```
zm711, line 915:
```suggestion
def check_recording_attributes_match(recording1, recording2_attributes, skip_properties=True) -> bool:
```
For a function like this it isn't clear to me that this returns a bool. So I prefer knowing that this just returns a bool. I say this because there are a couple checks in Neo that return Nothing to indicate passing and raise an error to indicate a problem. This returns a bool to say everything is okay. You have it in the docstring, but also seeing it in the type hint is nice (I know we are inconsistent with type hints so we can ignore this if we don't want the type hint).

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -588,6 +588,35 @@ def load_from_zarr(cls, folder, recording=None):
...
+        """
+        # check that recording is compatible
+        assert check_recording_attributes_match(
+            recording, self.rec_attributes, skip_properties=True
+        ), "Recording attributes do not match."
+        assert np.array_equal(
+            recording.get_channel_locations(), self.get_channel_locations()
+        ), "Recording channel locations do not match."
+        if self._recording is not None:
+            warnings.warn("SortingAnalyzer recording is already set. This will overwrite the current recording.")
```
zm711, line 617:
```suggestion
            warnings.warn("SortingAnalyzer recording is already set. Current recording is being overwritten.")
```

I switched from future tense to present tense to try to emphasize it is happening. Future tense gives it a vibe that one could prevent this from happening but by the warning it is too late. 

Also maybe do we want to say that if save_as it used this new recording will be written to the metadata so use caution?

src/spikeinterface/core/recording_tools.py
zm711, line 923:
```suggestion
    recording2_attributes : dict
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -588,6 +588,35 @@ def load_from_zarr(cls, folder, recording=None):
...
+
+        Parameters
+        ----------
+        recording : BaseRecording
+            The recording object to set as temporary recording.
+
+        Raises
+        ------
+        ValueError
+            _description_
```
zm711, line 607: This doesn't raise anything (technically AssertionError), but maybe we just delete this section. We often don't do the raises section.

src/spikeinterface/core/recording_tools.py
```diff
@@ -910,3 +910,35 @@ def get_rec_attributes(recording):
...
+    recording2 : BaseRecording
+        The second recording object
+
+    Returns
+    -------
+    bool
+        True if the recordings have the same attributes
+    """
+    recording1_attributes = get_rec_attributes(recording1)
+    recording1_attributes["probegroup"] = recording1.get_probegroup()
```
zm711, line 932: Why do we need this line? It is never used that I can see?

zm711: Or are we mutating in place? That seems dangerous.

zm711 (commented):
A couple comments and questions. But this looks pretty good to me.

src/spikeinterface/core/sortinganalyzer.py
zm711: Actually reading both our warnings they sound a little frightening. We are not overwriting the recording, but we are point to a new recording. Maybe we think a bit more how to say this without being scary.

alejoe91: `temporarily replaced`?

src/spikeinterface/core/recording_tools.py
alejoe91: we don't need it anymore :)

src/spikeinterface/core/recording_tools.py
```diff
@@ -910,3 +910,34 @@ def get_rec_attributes(recording):
...
+    Returns
+    -------
+    bool
+        True if the recordings have the same attributes
+    """
+    recording1_attributes = get_rec_attributes(recording1)
+    recording2_attributes = deepcopy(recording2_attributes)
+    if skip_properties:
+        recording1_attributes.pop("properties")
+        recording2_attributes.pop("properties")
```
samuelgarcia, line 935: Are upi handling the case skip_properties:=false after this ?
I would remove skip_properties option because we do not use.
No ?

src/spikeinterface/core/tests/test_sortinganalyzer.py
```diff
@@ -28,8 +28,13 @@ def get_dataset():
     return recording, sorting
 
 
-def test_SortingAnalyzer_memory(tmp_path):
-    recording, sorting = get_dataset()
+@pytest.fixture(scope="module")
+def get_dataset():
+    return _get_dataset()
```
samuelgarcia, line 33:
```suggestion
@pytest.fixture(scope="module")
def dataset():
    return get_dataset()
```

src/spikeinterface/core/tests/test_sortinganalyzer.py
```diff
@@ -15,7 +15,7 @@
 import numpy as np
 
 
-def get_dataset():
+def _get_dataset():
```
samuelgarcia, line 18:
```suggestion
def get_dataset():
```

samuelgarcia: This smater no ?
why private ?

src/spikeinterface/core/tests/test_sortinganalyzer.py
```diff
@@ -28,8 +28,13 @@ def get_dataset():
...
 
 
-def test_SortingAnalyzer_memory(tmp_path):
-    recording, sorting = get_dataset()
+@pytest.fixture(scope="module")
+def get_dataset():
+    return _get_dataset()
+
+
+def test_SortingAnalyzer_memory(tmp_path, get_dataset):
```
samuelgarcia, line 36:
```suggestion
def test_SortingAnalyzer_memory(tmp_path, dataset):
```

samuelgarcia: The get_dataset sounds like a function but here it the variable.

src/spikeinterface/core/tests/test_sortinganalyzer.py
```diff
@@ -257,8 +273,10 @@ def test_extension():
 
 if __name__ == "__main__":
     tmp_path = Path("test_SortingAnalyzer")
-    test_SortingAnalyzer_memory(tmp_path)
-    test_SortingAnalyzer_binary_folder(tmp_path)
-    test_SortingAnalyzer_zarr(tmp_path)
+    dataset = _get_dataset()
```
samuelgarcia, line 276:
```suggestion
    dataset = get_dataset()
```

src/spikeinterface/core/sortinganalyzer.py
zm711: I think that sounds better. I'm still curious what would happen if I do:

```python'
analyzer.set_temp_recording(my_new_recording)
analyzer.save_as('memory')
do stuff
analyzer.save_as('zarr', xx)
```
Will that then write the temp_recording to the file instead?

alejoe91: Yes, that's a really good point. Maybe keeping a flag would be good then

alejoe91: @zm711 I added a `self._tmp_recording` variable, which is not used in the `save_as` if the `self._recording` is available, but the `self.recording` property returns the `tmp_recording` over the original recording if it's set, so it's used in computations

zm711: That sounds good to me!

src/spikeinterface/core/recording_tools.py
```diff
@@ -917,3 +917,34 @@ def get_rec_attributes(recording):
         dtype=recording.get_dtype(),
     )
     return rec_attributes
+
+
+def check_recording_attributes_match(recording1, recording2_attributes) -> bool:
```
zm711, line 922: Sorry based on our previous discussions should we name this

do_recording_attributes_match

so it is a question and the bool is the answer?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -747,6 +773,9 @@ def unit_ids(self) -> np.ndarray:
     def has_recording(self) -> bool:
         return self._recording is not None
 
+    def has_temporary_recording(self) -> bool:
+        return self._tmp_recording is not None
```
zm711, line 777: should we make this function and variable have the same name?

so `has_tmp_recording` or `self._temporary_recording`?

I don't think this is crucial but might help us keep these straight...

zm711 (commented):
Just two very minor questions. We could just say let's not worry about this. But just thinking about naming since we've been having a bunch of discussions about this lately.

src/spikeinterface/core/sortinganalyzer.py

src/spikeinterface/core/recording_tools.py
alejoe91: done

zm711:
This is good by me once tests pass . Was @samuelgarcia going to do a final readthrough as well?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -730,7 +756,7 @@ def is_read_only(self) -> bool:
     def recording(self) -> BaseRecording:
         if not self.has_recording():
             raise ValueError("SortingAnalyzer could not load the recording")
-        return self._recording
+        return self._temporary_recording or self._recording
```
samuelgarcia, line 759: We should use  has_temporary_recording here also.
no ?

samuelgarcia (approved):

Issue #2782, opened 2024-04-29
sortingview attribute error


mtehrani-code:
Not having luck with any of the viewers at the moment :/

When I type:

w = si.plot_sorting_summary(curated_sorting, backend="sortingview", curation=True)

I get the error:

AttributeError: 'UnitsSelectionSorting' object has no attribute 'has_extension'

Not sure where the error is coming from. I installed sortingview and did the kachery authentication (I think).

Thanks in advance.

M

alejoe91:
Hi,

As mentioned in the docs, the function needs a SortingAnyzer, not a Sorting object.

zm711:
Maybe we should add a warning here: https://github.com/SpikeInterface/spikeinterface/blob/b6c2c91ecfc71dfa44213eaa30bf8dacf2da72a7/src/spikeinterface/widgets/base.py#L117-L118

Or maybe in the individual widgets themselves to make it clearer to the end-user when a sorting is acceptable and when you must have a sorting_analyzer. Although the argument names try to make this clear `sorting_analyzer` vs `sorting_analyzer_or_sorting` you wouldn't know that pattern if you 1) only use positional args or 2) you only use one plotting function so you don't know the convention.

alejoe91:
but yeah that final else whould be a `ValueError`!!!

mtehrani-code:
@alejoe91 Yes, that worked! Thank you.

I did check the link to the docs for sortingview that was in the tutorial (below). But it looks like it is no longer available:

https://spikeinterface.readthedocs.io/en/latest/module_widgets.html

PR #2791, opened 2024-05-01
Add errors to `ensure` functions  in `Base` of widgets

zm711:
This came up during talking in #2782, but I think `TypeError` instead of `ValueError` no? @alejoe91 ? 

And was there a reason you didn't want an error here originally @samuelgarcia @alejoe91 ?

### DIFF
--- a/src/spikeinterface/widgets/crosscorrelograms.py
+++ b/src/spikeinterface/widgets/crosscorrelograms.py
@@ -46,7 +46,9 @@ class CrossCorrelogramsWidget(BaseWidget):
         backend=None,
         **backend_kwargs,
     ):
-        sorting_analyzer_or_sorting = self.ensure_sorting_analyzer(sorting_analyzer_or_sorting)
+
+        if not isinstance(sorting_analyzer_or_sorting, BaseSorting):
+            sorting_analyzer_or_sorting = self.ensure_sorting_analyzer(sorting_analyzer_or_sorting)
 
         if min_similarity_for_correlograms is None:
             min_similarity_for_correlograms = 0
--- a/src/spikeinterface/widgets/base.py
+++ b/src/spikeinterface/widgets/base.py
@@ -109,17 +109,17 @@ class BaseWidget:
 
     @classmethod
     def ensure_sorting_analyzer(cls, input):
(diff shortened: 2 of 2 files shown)


alejoe91 (approved):

zm711:
I see what the error was. The correlograms were the exception that needed to allow the sorting to pass through. I think this solution is a bit better since it checks for sorting and if not sorting then we do the ensure which will raise an error if someone passes something random. This should be more robust to end-users potentially passing the wrong thing (in my opinion :) )

src/spikeinterface/widgets/crosscorrelograms.py
```diff
@@ -46,7 +46,9 @@ def __init__(
         backend=None,
         **backend_kwargs,
     ):
-        sorting_analyzer_or_sorting = self.ensure_sorting_analyzer(sorting_analyzer_or_sorting)
+
+        if not isinstance(sorting_analyzer_or_sorting, BaseSorting):
+            sorting_analyzer_or_sorting = self.ensure_sorting_analyzer(sorting_analyzer_or_sorting)
```
samuelgarcia, line 51: Here you break the fact that sorting can be an input no ?

samuelgarcia:
Theses function were more here for automatic convertion (with MockWaveform) than for a script type checking.
Sometimes a hack or the type happen after.
The idea of not having an error was to be able to both Sorting or Analyzer for instance.
I think you broke this for correlogram.
I am not sure to valide 100%. Let me have a deeper look.

src/spikeinterface/widgets/crosscorrelograms.py
zm711: Actually tests are passing now. 

The logic for the if statement is 

if base sorting => continue as base sorting
if not base sorting => ensure sorting analyzer (for sorting analyzer or mock waveform)

So both cases are addressed.

alejoe91:
This looks good to me

src/spikeinterface/widgets/crosscorrelograms.py
samuelgarcia: ok
then it is ok.

samuelgarcia:
lets merge and see.

PR #2792, opened 2024-05-01
Add write settings.json for back compatibility for early-adopters of SortingAnalyzer

### LINKED ISSUE #2788: `setting.json` file not written before `return_scaled` was global for a sortinganalyzer and is now a required file for reloading
Basically now that we set the global state of `return_scaled` for `SortingAnalyzer` an extra file is being written at the root of the analyzer folder called `setting.json`. It contains:

```json
{
    "return_scaled": true
}
```

But this means that if I try to call
```python
analyzer = load_sorting_analyzer(....)
```
I get a file not found because it is looking for this setting file. The easy solution is that I'm just copying the `setting.json` from a more recent analyzer, but do we want to make a full solution or just let people know that they need to do this?

zm711:
Fixes #2788 

See issue for full discussion but basically before `return_scaled` was global the `settings.json` file was not written and is now required to reload a SortingAnalyzer. This just writes the file to allow for compatibility for early adopters and can be removed at some point in the future.

@samuelgarcia is this what you had in mind based on our discussion in the issue?

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -416,9 +416,21 @@ class SortingAnalyzer:
         else:
             sparsity = None
 
+        # PATCH: Because SortingAnalyzer added this json during the development of 0.101.0 we need to save
+        # this as a bridge for early adopters. The else branch can be removed in version 0.102.0/0.103.0
+        # so that this can be simplified in the future
+        # See https://github.com/SpikeInterface/spikeinterface/issues/2788
+
         settings_file = folder / f"settings.json"
-        with open(settings_file, "r") as f:
-            settings = json.load(f)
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Hi Zach.
Totaly perfect.

samuelgarcia (approved):

samuelgarcia:
Note that we should do the same for the zarr case no ? Maybe not. Lets do it on demand.

alejoe91 (approved):

zm711:
> Note that we should do the same for the zarr case no ? Maybe not. Let's do it on demand.

Yeah based on our issue tracker so far everyone (my group + myself included) are using `binary_folder`. But that was only because I already had an install of SI without the necessary compression packages and was too lazy to download so I started using binary folder. :)

PR #2794, opened 2024-05-02
Add typing to `write_binary_recording`

h-mayorquin:
Working on the tutorial. Small PR with improvement.

### DIFF
--- a/src/spikeinterface/core/recording_tools.py
+++ b/src/spikeinterface/core/recording_tools.py
@@ -68,12 +68,12 @@ def _init_binary_worker(recording, file_path_dict, dtype, byte_offest, cast_unsi
 
 
 def write_binary_recording(
-    recording,
-    file_paths,
-    dtype=None,
-    add_file_extension=True,
-    byte_offset=0,
-    auto_cast_uint=True,
+    recording: "BaseRecording",
+    file_paths: list[Path | str] | Path | str,
+    dtype: np.ndtype = None,
+    add_file_extension: bool = True,
+    byte_offset: int = 0,
+    auto_cast_uint: bool = True,
     **job_kwargs,
 ):
     """
@@ -94,7 +94,7 @@ def write_binary_recording(
         If True, file the ".raw" file extension is added if the file name is not a "raw", "bin", or "dat"
     byte_offset: int, default: 0
         Offset in bytes for the binary file (e.g. to write a header)
-    auto_cast_uint: bool, default: True
+    auto_cast_uint: bool, default: True  
         If True, unsigned integers are automatically cast to int if the specified dtype is signed
     {}
     """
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/recording_tools.py
```diff
@@ -68,12 +68,12 @@ def _init_binary_worker(recording, file_path_dict, dtype, byte_offest, cast_unsi
...
 def write_binary_recording(
-    recording,
-    file_paths,
-    dtype=None,
-    add_file_extension=True,
-    byte_offset=0,
-    auto_cast_uint=True,
+    recording: "BaseRecording",
+    file_paths: list[Path | str] | Path | str,
+    dtype: np.ndtype = None,
```
zm711, line 73: I'm curious where did you find this type hint? dtype is one of the hard ones because I thought the official type was 

`numpy.typing.`DTypeLike` see docs [here](https://numpy.org/doc/stable/reference/typing.html#numpy.typing.DTypeLike).

But I think that degree of typing is too extreme, so in Neo we settle on `np.dtype` which is actually the construct to make a dtype object of whatever type you need. But maybe you've got some documentation of numpy typing that I could read?

h-mayorquin: Basically went through the same reasoning as you do there.  I am using the dtype as a non-formal hint to the code user (this would not work with a static analyser I think) Plus, the official numpy dtype fails with some of the tests in neuroconv because it does not work with the inspect library in the standard library.

I am hoping that numpy dtype will stabilize as the python typing ecossytem matures and we are able to drop support of older python versions. Then we can switch comfortably, meanwhile this is better than nothing.


All of that reasoning was like a year or two ago and I have been operating on auto-pilot with that assumption. If things have changed and you find out there is a better way I am all ears.

zm711: Works for me. I was hoping you had found something new :) 

But yeah I guess we just have to wait for things to stabilize. Thanks!

src/spikeinterface/core/recording_tools.py
```diff
@@ -110,8 +110,7 @@ def write_binary_recording(
         file_path_list = [add_suffix(file_path, ["raw", "bin", "dat"]) for file_path in file_path_list]
 
     dtype = dtype if dtype is not None else recording.get_dtype()
-    cast_unsigned = False
-    if auto_cast_uint:
+    if auto_cast_uint:  # TODO should we deprecate this given that we have `unsigned_to_signed`?
```
alejoe91, line 113: Yes!

alejoe91: @h-mayorquin can you add it to this PR?

h-mayorquin: Done.

h-mayorquin: I also improved the docstrings a little bit.

zm711: The docstring looks nicer to me :)

src/spikeinterface/core/recording_tools.py
```diff
@@ -110,9 +112,12 @@ def write_binary_recording(
         file_path_list = [add_suffix(file_path, ["raw", "bin", "dat"]) for file_path in file_path_list]
 
     dtype = dtype if dtype is not None else recording.get_dtype()
-    cast_unsigned = False
     if auto_cast_uint:
         cast_unsigned = determine_cast_unsigned(recording, dtype)
+        warning_message = (
+            "auto_cast_uint is deprecated and will be removed in 0.103. Use the `unsigned_to_signed` function instead."
+        )
+        warnings.warn(warning_message, DeprecationWarning)
```
alejoe91, line 120:
```suggestion
        warnings.warn(warning_message, DeprecationWarning, stacklevel=2)
```

alejoe91 (approved):

PR #1781, opened 2023-07-04
Make binary recording memmap efficient III (Avoiding spikes by only reserving memory equal to the trace)

h-mayorquin:
This is a follow up from #1742, it only allocates opens a memmap pointing to the specific trace and not to the whole file.

PR #2796, opened 2024-05-02
Only assign memmap within boundaries for `write_binary`

h-mayorquin:
This is like #1781 but for writing. I am writing the tutorial for how to use the job_kwargs and the memory heap reports crazy allocations for the following code:

```python

from pathlib import Path
from spikeinterface.core import write_binary_recording



minutes = 30
recording = generate_recording(num_channels=384, durations=[60.0 * minutes], seed=0)


job_kwargs = dict(n_jobs=1, total_memory="1G", progress_bar=False, verbose=False)
file_path = Path(__file__).parent / "recording_test_n_jobs.dat"
write_binary_recording(recording=recording, file_paths=[file_path], **job_kwargs)
```

image

Reason being that ta memmap for the whole array (80 GiB) is reserved even if it not used. A possible consequence of this massive heap allocation is that maybe the system will overswap. With the current PR this heap allocations disappears.

image

This will make my life easier for tracking the efficiency of processing and I think is a small increase in complexity.


h-mayorquin:
Performance:

this branch:

image

The main branch:

image

Changes are within the standard deviation of each other.

samuelgarcia:
thanks. This is really cool.

src/spikeinterface/core/recording_tools.py
```diff
@@ -152,33 +152,40 @@ def _write_binary_chunk(segment_index, start_frame, end_frame, worker_ctx):
...
     if traces.dtype != dtype:
         traces = traces.astype(dtype, copy=False)
-    array[start_frame:end_frame, :] = traces
 
-    # Close the memmap
+    memmap_array[...] = traces
+
     memmap_obj.flush()
 
+    memmap_obj.close()
```
samuelgarcia, line 187: Do we need this ?
the memmap_ojob is deleted no ?
This is a bit confusing because the file is stil open. Are we sure that this line do not close the underlying file ?

samuelgarcia:
Merci beaucoup beaucoup!
I ask for a small question.
We have lot of work to do on the neo side as well!!!

src/spikeinterface/core/recording_tools.py
h-mayorquin: I am not sure but the file is not closed, otherwise the tests would not work.

samuelgarcia (approved):

Issue #2789, opened 2024-04-30
Should aggregate channels preserve the channel_ids?


h-mayorquin:
The following code fails:

```python
from spikeinterface.core import aggregate_channels

recording1 = generate_recording(num_channels=3, durations=[10], set_probe=False)  # To avoid location check
recording1 = recording1.rename_channels(new_channel_ids=["a", "b", "c"])
recording2 = generate_recording(num_channels=2, durations=[10], set_probe=False)  
recording2 = recording2.rename_channels(new_channel_ids=["d", "e"])

aggregated_recording = aggregate_channels([recording1, recording2])  
assert aggregated_recording.get_num_channels() == 5
assert list(aggregated_recording.get_channel_ids()) == ['a', 'b', 'c', 'd', 'e']

```

The `channel_ids` of the aggregated recording are:

```
[0, 1, 2, 3, 4]
```

I think that the `channel_ids` should be preserved? Thoughts? Unforseen complexities?

zm711:
Could there ever be a case where the channel_ids are ['a', 'b', 'c'] and ['a', 'b'] such that by keeping them you would mess up  the indexing post-aggregating? I'm not sure why this would happen (combining two recordings that shouldn't be combined...?) but that is the only thing I could think of.

alejoe91:
Yes, in that case we should rename channels. IMO, if the channel ids are unique, then we can preserve them

samuelgarcia:
And also the same dtype.

PR #2797, opened 2024-05-03
Channel aggregation preserves names

h-mayorquin:
This should fix https://github.com/SpikeInterface/spikeinterface/issues/2789

This added a tests for what to do when the channels_ids are not unique or not shared type as discussed in the issue.

The only contentious issue is that it takes a stance on the issue #1432 by making the default strings.


src/spikeinterface/core/channelsaggregationrecording.py
```diff
@@ -49,14 +62,16 @@ def __init__(self, recording_list, renamed_channel_ids=None):
                 break
 
         if not (ok1 and ok2 and ok3 and ok4):
-            raise ValueError("Sortings don't have the same sampling_frequency/num_segments/dtype/num samples")
+            raise ValueError(
+                "Recordings do not have consistent sampling frequency, number of segments, data type, or number of samples."
```
zm711, line 66: I understand the check so feel free to ignore this comment, but the error doesn't really help the user as is. It says "something is wrong", but the user doesn't know. Maybe I have a recording I converted to a different dtype and so I assume it is okay, but don't realize. To make better errors would be to be clear what is wrong, but then that makes the check into a series of if

if not ok1:
raise error for sampling
if not ok2:
raise error for num seg.

But this isn't in the scope of this particular PR, so feel free to ignore.

h-mayorquin: I wanted to modify this one but Sam might like it that way and I did not wanted to increase the scope of the discussion. In my opinion it would be better to have an assertion per check.

zm711: Sounds good. Mental note for a future PR :)

src/spikeinterface/core/channelsaggregationrecording.py
```diff
@@ -16,16 +16,29 @@ class ChannelsAggregationRecording(BaseRecording):
...
-                "'renamed_channel_ids' doesn't have the " "right size or has duplicates!"
-            )
+            assert (
+                len(np.unique(renamed_channel_ids)) == num_all_channels
+            ), "'renamed_channel_ids' doesn't have the right size or has duplicates!"
             channel_ids = list(renamed_channel_ids)
         else:
-            channel_ids = list(np.arange(num_all_channels))
+            # Collect channel IDs from all recordings
+            combined_ids = sum((list(rec.get_channel_ids()) for rec in recording_list), [])
```
alejoe91, line 31: @h-mayorquin channel IDs are vector, so we can simply do:
```
# 1. check dtypes first
are_all_channels_same_type = np.unique([rec.channel_ids.dtype for rec in recording_list]).size == 1
if are_all_channels_same_type:
    combined_ids = np.concatenate([rec.channel_ids for rec in recording_list])
    # check that they are unique
    are_all_channels_unique = len(set(combined_ids)) == num_all_channels
    if are_all_channels_unique:
        channel_ids = combined_ids
    else:
        # If IDs are not unique or not of the same type, use default numeric IDs
        channel_ids = default_channel_ids
```

src/spikeinterface/core/channelsaggregationrecording.py
```diff
@@ -67,31 +82,20 @@ def __init__(self, recording_list, renamed_channel_ids=None):
...
-                # remap device channel indices correctly
-                prop_values["device_channel_indices"] = np.arange(self.get_num_channels())
             self.set_property(key=prop_name, values=prop_values)
 
-        # if locations are present, check that they are all different!
-        if "location" in self.get_property_keys():
-            location_tuple = [tuple(loc) for loc in self.get_property("location")]
-            assert len(set(location_tuple)) == self.get_num_channels(), (
-                "Locations are not unique! " "Cannot aggregate recordings!"
-            )
```
alejoe91, line 85: Why removing this?

src/spikeinterface/core/tests/test_channelsaggregationrecording.py
```diff
@@ -81,5 +80,41 @@ def test_channelsaggregationrecording():
     print(recording_agg_prop.get_property("brain_area"))
 
 
+def test_channel_agregation_preserve_ids():
```
alejoe91, line 83:
```suggestion
def test_channel_aggregation_preserve_ids():
```

src/spikeinterface/core/tests/test_channelsaggregationrecording.py
```diff
@@ -81,5 +80,41 @@ def test_channelsaggregationrecording():
...
+    recording1 = recording1.rename_channels(new_channel_ids=["a", "b", "c"])
+    recording2 = generate_recording(num_channels=2, durations=[10], set_probe=False)
+    recording2 = recording2.rename_channels(new_channel_ids=["a", "b"])
+
+    aggregated_recording = aggregate_channels([recording1, recording2])
+    assert aggregated_recording.get_num_channels() == 5
+    assert list(aggregated_recording.get_channel_ids()) == ["0", "1", "2", "3", "4"]
+
+
+def test_channel_agregation_does_not_preserve_ids_not_the_same_type():
```
alejoe91, line 107:
```suggestion
def test_channel_aggregation_does_not_preserve_ids_not_the_same_type():
```

src/spikeinterface/core/channelsaggregationrecording.py
h-mayorquin: A mistake or god knows what the heck I was thinking. Reverted.

src/spikeinterface/core/channelsaggregationrecording.py
h-mayorquin: Ok. I used numpy operators in the latest version.

alejoe91 (approved):

PR #2798, opened 2024-05-03
Move towards making set_probe in place by default

### LINKED ISSUE #2193: `set_probe` behavior for `RecordingExtractors`

h-mayorquin:
Should fix https://github.com/SpikeInterface/spikeinterface/issues/2193

### DIFF
--- a/src/spikeinterface/core/baserecordingsnippets.py
+++ b/src/spikeinterface/core/baserecordingsnippets.py
@@ -2,11 +2,11 @@ from __future__ import annotations
 from pathlib import Path
 
 import numpy as np
+import warnings
 
 from probeinterface import Probe, ProbeGroup, write_probeinterface, read_probeinterface, select_axes
 
 from .base import BaseExtractor
-from .core_tools import check_json
 from .recording_tools import check_probe_do_not_overlap
 
 from warnings import warn
@@ -178,7 +178,7 @@ class BaseRecordingSnippets(BaseExtractor):
             "in_place will change its default to True in version 0.103 and stop returning a recording"
             " use recording.create_with_probes(probe_or_probegroup) instead to replicate the old functionality"
         )
-        warn.warn(
+        warnings.warn(
             deprecation_msg,
             DeprecationWarning,
             stacklevel=2,
--- a/src/spikeinterface/core/baserecordingsnippets.py
+++ b/src/spikeinterface/core/baserecordingsnippets.py
@@ -76,7 +76,7 @@ class BaseRecordingSnippets(BaseExtractor):
 
         Parameters
         ----------
-        probe_or_probegroup: Probe, list of Probe, or ProbeGroup
+        probe: a Probe object
             The probe(s) to be attached to the recording
         group_mode: "by_probe" | "by_shank", default: "by_probe
             "by_probe" or "by_shank". Adds grouping property to the recording based on the probes ("by_probe")
@@ -90,6 +90,7 @@ class BaseRecordingSnippets(BaseExtractor):
         sub_recording: BaseRecording
             A view of the recording (ChannelSlice or clone or itself)
         """
+
         assert isinstance(probe, Probe), "must give Probe"
         probegroup = ProbeGroup()
         probegroup.add_probe(probe)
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -225,6 +234,27 @@ def set_probes(self, probe_or_probegroup, group_mode="by_probe", in_place=False)
 
         return sub_recording
 
+    def create_with_probes(self, probe_or_probegroup, group_mode="by_probe"):
```
zm711, line 237: I wonder if it should be `create_with_probe` vs `probes`.

zm711 (commented):
For me I liked the always return (i.e. in_place always False), but from the debate it looks like it went the other direction. (That's what I get for being silent). 

So my only comment would more be discoverability/understandability. We distinguish between `set_probe` and `set_probegroup` at the user-level, but now this function does both. Why? shouldn't we have a `create_with_probe` and a separate `create_with_probegroup` to maintain the previous api for users. We don't really have the concept of `probes` right? Instead you choose between A probe and A probegroup neither being plural for the api. 

EDIT: just to be clear I do support making this a one option function in general. The rest of the api always returns a new object so having this function be the one function that gives users the choice is super confusing. For me I was leaning toward return just for api consistency but otherwise I don't care which way it works.

JoeZiminski:
Personally I don't mind if it returns a recording (for API consistency) or sets in place. But if it returns a recording I think it should be renamed from `set_probe` because that implies (to me at least) an in-place operation.

src/spikeinterface/core/baserecordingsnippets.py
h-mayorquin: Ther are two functions, the one with the plural name is the most general as it takes both probe a list or a probe group, so that's where I want the the warnin to be.  I wish there was only one function and if you see the diff in the changes it seems that that was the intention of one point.

Simplifying probe setting could be a issue of its own, I am not making any changes in that direction in this PR. It is just to get the ball rolling to make the behaivor of the function consistent with setter semantics.

h-mayorquin:
As mentioned above, there are two functions `set_probe` and `set_probes`, it seems that `set_probes` is more general. I suggest opening another issue if we want to discuss the probe setting API. The aim of this is just to change the functions to be in place.

Am I missing something?

zm711:
I was referring to the new function you made called `create_with_probes`, which is part of this api. But let me see if I can find the function `set_probes`. My point was actually that I would prefer 

`create_with_probe`
`create_with_probegroup`

to mimic the functions I normally use `set_probe` and `set_probegroup`. I have never seen the `set_probes` function.

h-mayorquin:
All the functions are public so I think you would have to ask @samuelgarcia and @alejoe91 what is the "right way" (intendend) that they envisioned or want. Usually they prefer general functions that handle a bunch of cases (see that they prefer to use `recording.save()` vs `recording.save_to_* ()` and in general I see them adding keyword to functions instead of creating a function for a different behavior)  so I did in this way.

zm711:
Agreed. I don't think I've ever seen us document `set_probes` which is what I meant by "right way", but as you said it is a public function so people could use. Happy to wait to see what they say.

zm711:
I just searched the docs and set_probe is mentioned like 5 times. set_probegroup is mentioned once and set_probes is never mentioned (again based on the search function).

h-mayorquin:
Which search function are you using?

It is here:

https://github.com/h-mayorquin/spikeinterface/blob/4f4fafb96382676c63949647b172ee013f35a290/src/spikeinterface/extractors/neoextractors/spikegadgets.py#L43-L44

zm711:
The search function on readthedocs. I'm talking about rtd documentation.

h-mayorquin:
Done in #4465 by @alejoe91  set_probe/set_probegroup are now in place by default, with in_place deprecated (removal in 0.106.0), which also closes #2193. Closing this.

PR #2828, opened 2024-05-10
Zarr: extract time vector once and for all!

alejoe91:
If the zarr folder has timestamps, they were re-extracted and uncompressed every time the `get_times()` was called. 

This PR loads the zarr array as a numpy array at instantiation to speed things up

### DIFF
--- a/src/spikeinterface/core/zarrextractors.py
+++ b/src/spikeinterface/core/zarrextractors.py
@@ -72,7 +72,7 @@ class ZarrRecordingExtractor(BaseRecording):
             time_kwargs = {}
             time_vector = self._root.get(f"times_seg{segment_index}", None)
             if time_vector is not None:
-                time_kwargs["time_vector"] = time_vector
+                time_kwargs["time_vector"] = time_vector[:]
             else:
                 if t_starts is None:
                     t_start = None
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/zarrextractors.py
```diff
@@ -72,7 +72,7 @@ def __init__(self, folder_path: Path | str, storage_options: dict | None = None)
             time_kwargs = {}
             time_vector = self._root.get(f"times_seg{segment_index}", None)
             if time_vector is not None:
-                time_kwargs["time_vector"] = time_vector
+                time_kwargs["time_vector"] = time_vector[:]
```
zm711, line 75: Maybe I'm misunderstanding the compression-decompression, but how does switching to a view change this?

alejoe91: Without the `[:]` the time_vector is set to a `zarr.Array` object.

This object performs a decompression/retrieval from file upon slicing, so if I call `get_times()` 100 times, the data will be accessed from the file (and optionally decompressed) 100 times.

This simple change slices the array, so that `time_kwargs["time_vector"]` is now a numpy array and not a zarr object anymore.

Hope this is clear :)

zm711: Oh yeah that makes sense. Cool.

zm711 (approved):

samuelgarcia:
As discuss with Alessio this morning.
Reading big vectors in the init is not a very good idea.
We must have a system that delay and cache the time vector reading. In short read the time vector only on demand.

PR #2830, opened 2024-05-10
Switch `Visualise` to `Visualize`

zm711:
Related to the decision to use US spelling.

Last fix we need to make is analyse_neuropixel to analyze_neuropixel. But I don't have access to the raw data to fix that one.

@chrishalcrow  :)

### DIFF
--- a/doc/how_to/viewers.rst
+++ b/doc/how_to/viewers.rst
@@ -1,4 +1,4 @@
-Visualise Data
+Visualize Data
 ==============
 
 There are several ways to plot signals (raw, preprocessed) and spikes.


alejoe91 (approved):

chrishalcrow:
Oh goodness. I'll need a good extension to help me with this - thanks Zach!

JoeZiminski:
[codespell](https://github.com/codespell-project/codespell) is a handy tool for spell checking that does well at ignoring scientific jargon, I'm not sure how it handles UK vs. US but might be helpful!

PR #2834, opened 2024-05-12
Set correct package version in pyproject.toml

h-mayorquin:
You are pointing the version of the next version that will release but this is causing troubles for dependency resolution at neuroconv. More concrently, if I install the package from git for testing, like here:

https://github.com/catalystneuro/neuroconv/blob/fa9285b5a09e6c80e8c3011f08b2cd418fb39aff/.github/workflows/dev-testing.yml#L57

  Then I get an incorrect version of spikeinterface that is incompatible with another bounds.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "spikeinterface"
-version = "0.101.0"
+version = "0.100.6"
 authors = [
   { name="Alessio Buccino", email="alessiop.buccino@gmail.com" },
   { name="Samuel Garcia", email="sam.garcia.die@gmail.com" },


zm711:
I thought this was suppose to be based on the DEV flag in the init?

h-mayorquin:
> I thought this was suppose to be based on the DEV flag in the init?

I don't understand what you mean by **this**.

alejoe91:
There is a DEV_MODE flag in the init that indicates whether it's a dev or release version. 

I think that the version needs to stay 0.101.0, since it has breaking changes with 0.100.x

If this is causing troubles on neuroconv, we should address it there :)

Closing

h-mayorquin:
You are getting the wrong version of if you do `spikeinterface.__version__`, that's the problem. This is not solved by using the dev flag, I am unable to use standard versioning tools like in pip.

I honestly don't understand why you are taking such as strong stance against standard tooling.

alejoe91:
@h-mayorquin what version are you getting? The versionin `main` should be `0.101.0`, the released version `0.100.6`

h-mayorquin:
Yes, which means that pip can't deal with both installing the main version and the restriction of the version being smaller than 0.100.6.  This is not a neuroconv problem, the problem is that pip gets the package version from pyproject.toml and can't execute code dynamically.

PR #2841, opened 2024-05-13
Add error message and tests for `plot_motion`.

### LINKED ISSUE #2725: `plot_motion` does not accept `axes` or `ax` argument
The `widgets.motion.plot_motion()` function asserts that `ax` and `axes` is `None` [here](https://github.com/SpikeInterface/spikeinterface/blob/fa57feeca6b403ae8595204f18313f038f7dc5ac/src/spikeinterface/widgets/motion.py#L76) but the [documentation](https://spikeinterface.readthedocs.io/en/latest/api.html#spikeinterface.widgets.plot_motion) says these should be accepted inputs.

Is there a reason for this? Happy to open a PR to make the function support `ax` / `axes` arguments if not.

Also the latest docs have a number of references to `si.load_motion_info` and `si.plot_motion` but it appears these have since moved, also happy to open PR ASAP but leaving here as don't want to forget.

JoeZiminski:
This PR closes #2725 by adding an error message when `ax` or `axes` arguments are passed, these are `backend_kwargs` which are not supported for this function.

This is slightly suboptimal as the documentation still says they are accepted, but at least it is clear what to do when it does go wrong. I looked into making the function support passing `axes` but it would make things very complicated for (I don't think?) not too much benefit. 

I also added tests for the function and made a couple of minor changes:
- Added a little more explanation to the docstrings..
- renamed the variables `x` and `y` to make slightly more descriptive.
- renamed x-axis label `Times [s]` to `Time [s]`.

I had a couple of questions:
- The `motion_lim` is documented as `Tuple` or `None` but it actually expects an `int`.  I guess this could be changed to accept a `Tuple` similar to `depth_lim`?
- at the moment the widget object `.ax` is set to an empty `Axes` object. Maybe it would be clearer to set this to `None` as it is unused?

Also a more general question, the `matplotlib` and `scipy` imports are not at the top of the file, and I see the tests do not expect `scipy` to be installed. I guess the logic is the same for `matplotlib` which is why it is not top of file? Personally I would add any 'core' python modules to the general dependencies and then there is less maintenance required to have to workaround users potentially not having them installed. I don't think anyone will have trouble installing `matplotlib` or `scipy` unlike heavier modules such as `numba`. However may there are other considerations (e.g. docker images?).

### DIFF
--- a/src/spikeinterface/widgets/motion.py
+++ b/src/spikeinterface/widgets/motion.py
@@ -178,13 +178,14 @@ class MotionWidget(BaseWidget):
         # Plot the Motion Vectors ---------------------------------------------
 
         if dp.motion_lim is None:
-            motion_lim = np.max(np.abs(dp.motion)) * 1.05
+            max_motion_lim = np.max(np.abs(dp.motion)) * 1.05
+            motion_lim = (-max_motion_lim, max_motion_lim)
         else:
             motion_lim = dp.motion_lim
 
         ax2.plot(temporal_bins_plot, dp.motion, alpha=0.2, color="black")
         ax2.plot(temporal_bins_plot, np.mean(dp.motion, axis=1), color="C0")
-        ax2.set_ylim(-motion_lim, motion_lim)
+        ax2.set_ylim(motion_lim)
         ax2.set_xlabel("Time [s]")
         ax2.set_ylabel("Motion [μm]")
         ax2.set_title("Motion vectors")
@@ -204,7 +205,7 @@ class MotionWidget(BaseWidget):
                     dp.spatial_bins[-1],
                 ),
             )
-            im.set_clim(-motion_lim, motion_lim)
+            im.set_clim(motion_lim)
             cbar = fig.colorbar(im)
             cbar.ax.set_xlabel("motion [μm]")
             ax3.set_xlabel("Time [s]")
--- a/src/spikeinterface/widgets/motion.py
+++ b/src/spikeinterface/widgets/motion.py
@@ -97,7 +97,8 @@ class MotionWidget(BaseWidget):
 
         # Setup figures and axes ----------------------------------------------
 
-        self.figure, self.axes, self.ax = make_mpl_figure(**backend_kwargs)
+        self.ax = None
+        self.figure, self.axes, _ = make_mpl_figure(**backend_kwargs)
         fig = self.figure
         fig.clear()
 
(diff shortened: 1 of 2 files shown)


alejoe91:
> I had a couple of questions:
> 
> * The `motion_lim` is documented as `Tuple` or `None` but it actually expects an `int`.  I guess this could be changed to accept a `Tuple` similar to `depth_lim`?

Yeah I think that sounds super reasonable

> * at the moment the widget object `.ax` is set to an empty `Axes` object. Maybe it would be clearer to set this to `None` as it is unused?

Agreed

> Also a more general question, the `matplotlib` and `scipy` imports are not at the top of the file, and I see the tests do not expect `scipy` to be installed. I guess the logic is the same for `matplotlib` which is why it is not top of file? Personally I would add any 'core' python modules to the general dependencies and then there is less maintenance required to have to workaround users potentially not having them installed. I don't think anyone will have trouble installing `matplotlib` or `scipy` unlike heavier modules such as `numba`. However may there are other considerations (e.g. docker images?).

The idea behind `core` VS `full` is that `core` can be used as a super lightweight module only to use the SI data model, e.g., from a third party package. Adding main Python modules to core could result in dependencies issues for these kind of 3rd party minimal usage.

JoeZiminski:
Thanks @alejoe91! I've made the related changes, this is now ready for review.

Out of interest what kind of third-party apps need low-overhead installations? This would be useful for me to know in general when thinking about dependencies on other packages.

src/spikeinterface/widgets/motion.py
```diff
@@ -7,7 +7,16 @@
...
 class MotionWidget(BaseWidget):
     """
-    Plot unit depths
+    Plot a summary of motion correction outputs. The plots shown are:
+
+    1) 'Peak depth' plot displaying the peak location for all detected spikes over time
+       before motion correction.
+    2) 'Corrected peak depth' displayed the peak location for all detected spikes
+       over time after motion correction.
+    3) 'Motion vectors' plot shows the estimated displacement per channel
```
JoeZiminski, line 16: I actually think I got this wrong, it is displacement across binned depths, not channels? In general I can fix these docstring to better use the terminology from the drift correction tutorial.

JoeZiminski: Okay I think I have fixed it now, by replacing "per channel" with "per spatial bin".

JoeZiminski:
Hey @samuelgarcia would you be happy with merging the changes to the `motion.py` in this PR (without the tests)? Assuming the changes to `motion.py` will not make the refactoring harder. If so I can split the changes to `motion.py` and the tests and add the tests back in later after refactoring.

JoeZiminski:
@samuelgarcia  please ignore the above comment, superseded by #3068, and we can add tests later.

PR #2842, opened 2024-05-13
Add `is_scaled` to `Templates` object

h-mayorquin:
Shouod solve https://github.com/SpikeInterface/spikeinterface/issues/2831

Draft because I don't know what will fail : )

PR #2848, opened 2024-05-14
Use the `is_scaled` attribute inside of `template_tools` functions

zm711:
Now that #2842 adds an attribute which indicates that the Templates object is scaled we should use that for the functions inside of template_tools no? If the Templates aren't scaled then we can't return scaled.


samuelgarcia:
Yes and maybe some other places like computing SNR or sparsity.
Thank you.

#2842 was merged too quickly for my taste, I did not read it!

zm711:
Added sparsity. Checked SNR and it only accepts SortingAnalyzer for now.

samuelgarcia:
merci beaucoup.

samuelgarcia (approved):

PR #2853, opened 2024-05-15
[Discussion] Add method to force loading Templates into memory

h-mayorquin:
Discussion PR.  @samuelgarcia @alejoe91 

Following on https://github.com/SpikeInterface/spikeinterface/pull/2828 and as alternative to the code in #2436. 

With this, if you are in a scenario (I guess like plotting?) where you are going to be making repeated calls to the functions then you can call this function. We can even add an argument to the signature and call this at post_init. 

The advantage of this vs loading by default is that the `__init__` of the object becomes lighter at the cost of some complexity.

What do you think?

### DIFF
--- a/src/spikeinterface/core/template.py
+++ b/src/spikeinterface/core/template.py
@@ -188,6 +188,19 @@ class Templates:
 
         return are_templates_sparse
 
+    def get_channel_locations(self) -> np.ndarray:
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Is it still a draft ?

h-mayorquin:
I don't think this should go in the next release. It is a minor detail that is discussion more than anything. I would remove the tag.

h-mayorquin:
Closed as stale.

PR #2856, opened 2024-05-15
Extend plot waveforms/templates to Templates object

alejoe91:
As discussed, starting to break #2436 down into small pieces :)

- [x] Add tests


src/spikeinterface/widgets/tests/test_widgets.py
```diff
@@ -1,4 +1,5 @@
 import unittest
+from numba.cuda import Out
```
samuelgarcia, line 2:
```suggestion
```

samuelgarcia (approved):

PR #2857, opened 2024-05-15
Template fetching methods

h-mayorquin:
Peeled from #2436

There was some discussion about method naming today. So to contrast Alessio more laconic approach I turned up the volume to my usual over-verbose crap. We can trim from here : )

Let me know if I missed something @alejoe91

### DIFF
--- a/src/spikeinterface/generation/template_database.py
+++ b/src/spikeinterface/generation/template_database.py
@@ -1,12 +1,140 @@
-from spikeinterface.core.template import Templates
 import zarr
+import functools
+import numpy as np
+
+from spikeinterface.core.template import Templates
 
 
-def fetch_templates_from_database(dataset="test_templates.zarr") -> Templates:
+@functools.cache
+def fetch_template_object_from_database(dataset="test_templates.zarr") -> Templates:
+    """
+    Fetch a template dataset from the spikeinterface template database.
+    A dataset is a collection of templates with associated metadata for one specific recording.
 
+    Parameters
+    ----------
+    dataset : str, default: "test_templates"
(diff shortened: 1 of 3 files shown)


src/spikeinterface/generation/template_database.py
```diff
@@ -1,12 +1,140 @@
...
+    Templates
+        The templates object.
+    """
+    import pandas as pd
+
+    templates_array = []
+    if isinstance(template_df_or_indices, pd.DataFrame):
+        template_info = template_df_or_indices
+    else:
+        template_info = fetch_templates_database_info().iloc[template_df_or_indices]
```
h-mayorquin, line 88: I think we should removed this option as it changes when reconstructing the database. Or have a global id as @samuelgarcia suggested. Actually that would probably be the most stable thing.

src/spikeinterface/generation/tests/test_template_fetch.py
```diff
@@ -1,13 +1,51 @@
-import pytest
```
samuelgarcia, line 1: Could we change the name to test_template_database.py

h-mayorquin: I think this makes sense. Yes.

h-mayorquin: Done.

src/spikeinterface/generation/template_database.py
alejoe91: let's remove this option

alejoe91 (approved):

PR #2858, opened 2024-05-15
Add binary sufixes to job tools memory specification (e.g. 1GiB vs 1GB)

h-mayorquin:
This is so primarly the ChunkRecordingExecturo can accept binary prefixes like 1 GiB, 500 MiB:

https://en.wikipedia.org/wiki/Binary_prefix

I think this is more useful for working with ram related things.

I am struggling with some memory errors while uploading data to the templates database and I am trying a more fine control so I thought this would be useful.

### DIFF
--- a/src/spikeinterface/core/job_tools.py
+++ b/src/spikeinterface/core/job_tools.py
@@ -8,9 +8,9 @@ import numpy as np
 import platform
 import os
 import warnings
+from spikeinterface.core.core_tools import convert_string_to_bytes
 
 import sys
-import contextlib
 from tqdm.auto import tqdm
 
 from concurrent.futures import ProcessPoolExecutor
@@ -23,7 +23,7 @@ _shared_job_kwargs_doc = """**job_kwargs: keyword arguments for parallel process
                 - chunk_size: int
                     Number of samples per chunk
                 - chunk_memory: str
-                    Memory usage for each job (e.g. "100M", "1G")
+                    Memory usage for each job (e.g. "100M", "1G", "500MiB", "2GiB")
                 - total_memory: str
                     Total memory usage (e.g. "500M", "2G")
                 - chunk_duration : str or float or None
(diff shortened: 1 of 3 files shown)


src/spikeinterface/core/job_tools.py
```diff
@@ -206,9 +196,11 @@ def ensure_chunk_size(
     chunk_size: int or None
         size for one chunk per job
     chunk_memory: str or None
-        must end with "k", "M" or "G"
+        must end with "k", "M", "G", etc for decimal units and "ki", "Mi", "Gi", etc for
+        binary units. (e.g. "1k", "500M", "2G", "1ki", "500Mi", "2Gi")
```
alejoe91, line 200: @h-mayorquin I think that this might make things extra complicated. If I understand correctly, 

1KB = 1000 bytes
1KiB = 1024 bytes

Is that correct? I had to look up the difference between decimal/binary for memory representation, so I think that an end user could be a bit puzzled by this. 

Happy to discuss!

h-mayorquin: I think that having the two is confusing indeed.

But the current setup is confusing as well. If somebody allocates 1G using the current SpikeInterface machinery  (and SpikeInterface was perfect in its handling of memory ...) the memory reported by the system won't match 1G as the user specified.

The reason for this is that RAM memory is reported in binary base. Both when you buy RAM and when you use utilities like system resource, the numbers you see are in binary base. .

Given that memory allocation here is a RAM issue, I think we should follow the convention of RAM specification (binary). However, we have to do things carefully because the decimal base has been there for so long.

Finally, given that using the job_kwargs is a relatively advanced feature, I am far less concerned about this type of confusion and I am fine with having the user specify which one they want to use.

(Hard drive storage, on the other hand, is reported in base 10, both when you calculate the size of an object in the system and when you buy a hard drive)

samuelgarcia (approved):

Issue #2164, opened 2023-11-03
Windows Permission Errors in SI


zm711:
## Motivation ##
As discussed in #2156, this is just a little summary of some hiccups that Windows users have bumped into dealing with the management of tempfiles or deleting files while running SpikeInterface. The total count of user complaints 5 across 3 issues. The 6th  user/4th issue is only semi-related to SpikeInterface since the recording holds a reference to the file while its open so it shouldn't be expected to be delete-able while the recording object exists. I will say I tested `overwrite` with a waveform extractor today and ran into no issues doing a folder overwrite. So my potential concern there was actually not correct. 

## Errors caused by SpikeInterface ##

#1871--3 users ran into WinError32 Error: maybe related to SC2 deleting temp files while running?

#1396 `check_your_install.py` fails due to PermissionError: [WinError 32] The process cannot access the file because it is being used by another process. I fixed by creating a separate cleanup script for Windows and skipping clean up in the check_your_install on Windows

#1225 WinError32 annoyed a user so much they switched to Linux (maybe that's the secret agenda of SI after all :) ). Error was related to `n_jobs` and docker.

## Errors only semi-related to SpikeInterface (ie not internal to package but happening while interacting with the package) ##

#1790: trying to clean up file while SpikeInterface was still holding a reference to them in Windows (WinError32)

## Testing ##

As mentioned in #2138, I tried running the testing suite locally on my Windows machine and was met with tons of WinError32 Errors.  I also tried the suggestion of changing the read-only bit on Windows as suggested in the [python documentation](https://docs.python.org/3/library/shutil.html) and it did not work. Maybe my reading was wrong, but I thought they were getting rid of the onerror and switching to onexec instead so for older python we would need to provide the read-only func to onerror and in future python version it would need to go through onexec. But maybe I was wrong. Either way my current environment for spikeinterface is python 3.11 so my test was using the onerror parameter. And the failure was still WinError32.

## Suggestions ##

In #2156, I had suggested at least adding some explicit testing of the folder generation and deletion into the core tests to ensure that the core writing/removing functions work on Mac/Linux/Windows and I still think that is a good idea since I, for one, don't know when the WinError 32 will exactly occur (and they are super annoying when they do occur).  Or to be more precise I know they occur when things are happening in the same process (like all of pytest), but I never know how to predict when SpikeInterface will generate new processes or use the same process for an operation.

@h-mayorquin, tagging you since the issue linking was so you could re-read over them.

[...]

PR #2861, opened 2024-05-16
Unpin sphinx and add networkx dependency.

JoeZiminski:
This PR unpins the Sphinx version dependency from 5.1.1 as well as read the docs from 1.0.0. It also adds `networkx` which from the logs seems to be a dependency on the docs as an build warning was raised.

I can't detect any issues from the upgrade, although it's quite hard to test. I looked through the logs and they look similar, Number of warnings differs by 1. I went through every page and compared with current docs, everything looks the same and seems to have build without issue, but I did skim pages just randomly checking features like 1) links working 2) images rendering 3) generated values the same. The only differences I saw is that in the new version, the API docs typing is written a little neater using the new shortcuts (e.g. "|") and there is a slight stylistic change (the grey boxes around certain text is now gone). It would be great if someone could double check this in case I missed anything. I'm surprised it worked out the box 😅 seems too good to be true...

BTW I am getting a lot of warnings (~143) when building on Windows, does anyone else? If so we can look to address these when we look at further docs revisions. I am also getting random errors when building some of the sphinx galleries (for main also), it seems that when trying to delete some files they are still in use. I guess this is a platform filesystem difference issue, but would be good to hear if anyone else on windows has had this problem.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -159,8 +159,8 @@ test = [
 ]
 
 docs = [
-    "Sphinx==5.1.1",
-    "sphinx_rtd_theme==1.0.0",
+    "Sphinx",
+    "sphinx_rtd_theme",
     "sphinx-gallery",
     "sphinx-design",
     "numpydoc",
@@ -169,6 +169,7 @@ docs = [
     # for notebooks in the gallery
     "MEArec",   # Use as an example
     "datalad==0.16.2",  # Download mearec data, not sure if needed as is installed with conda as well because of git-annex
+    "networkx",
     "pandas", # in the modules gallery comparison tutorial
     "hdbscan>=0.8.33",   # For sorters spykingcircus2 + tridesclous
     "numba", # For many postprocessing functions
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -159,8 +159,8 @@ test = [
 ]
 
 docs = [
-    "Sphinx",
-    "sphinx_rtd_theme",
+    "Sphinx==5.1.1",
+    "sphinx_rtd_theme==1.0.0",
     "sphinx-gallery",
     "sphinx-design",
     "numpydoc",
@@ -173,7 +173,6 @@ docs = [
     "hdbscan>=0.8.33",   # For sorters spykingcircus2 + tridesclous
     "numba", # For many postprocessing functions
     "xarray", # For use of SortingAnalyzer zarr format
-    "networkx",
     # for release we need pypi, so this needs to be commented
     "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",  # We always build from the latest version
     "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",  # We always build from the latest version
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -173,6 +173,7 @@ docs = [
     "hdbscan>=0.8.33",   # For sorters spykingcircus2 + tridesclous
     "numba", # For many postprocessing functions
     "xarray", # For use of SortingAnalyzer zarr format
+    "networkx",
     # for release we need pypi, so this needs to be commented
     "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",  # We always build from the latest version
     "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",  # We always build from the latest version
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -159,8 +159,8 @@ test = [
 ]
 
 docs = [
-    "Sphinx==5.1.1",
-    "sphinx_rtd_theme==1.0.0",
+    "Sphinx",
+    "sphinx_rtd_theme",
     "sphinx-gallery",
     "sphinx-design",
     "numpydoc",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -169,7 +169,6 @@ docs = [
     # for notebooks in the gallery
     "MEArec",   # Use as an example
     "datalad==0.16.2",  # Download mearec data, not sure if needed as is installed with conda as well because of git-annex
-    "networkx",
     "pandas", # in the modules gallery comparison tutorial
     "hdbscan>=0.8.33",   # For sorters spykingcircus2 + tridesclous
     "numba", # For many postprocessing functions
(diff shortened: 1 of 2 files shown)


zm711:
@JoeZiminski,

One of my goals is to work on the Windows friendliness, so I have an issue floating around about issues still present on Windows. Feel free to add the errors you're seeing there. Basically a lot of the cleanup during testing and doc building don't work on Windows due to permission differences between Windows and Unix-like systems. Alessio and I got the warnings down to 0 previously so it's wild that it's back to 143. Haha. 

But I'm happy to look over the docs from this build later today :)

chrishalcrow:
Hello, there seems to be some style inconsistencies on the API. A good example is the difference between `plot_unit_waveforms_density_map` and `plot_unit_waveforms` (https://spikeinterface--2861.org.readthedocs.build/en/2861/api.html#spikeinterface.widgets.plot_unit_waveforms_density_map)
The first has white boxes, the second has grey. This doesn't happen in the current format (https://spikeinterface.readthedocs.io/en/latest/api.html#spikeinterface.widgets.plot_unit_waveforms_density_map)

I would guess it's to do with our docstrings not being up to scratch? More reason to join me here: https://github.com/SpikeInterface/SpikeInteface-Hackathon-Edinburgh-May24/issues/29 !!

JoeZiminski:
That's great thanks @zm711, that explains why tests are not passing on my Windows machine 😅 Happy to help with this #2164!
I hope you can't replicate the warnings and it is something to do with my setup!

Thanks a lot @chrishalcrow I just blindly assumed this was due to some RTD stylistic change but never actually checked. I think the reason is that for classes the parameters are in grey boxes but for functions the arguments are not. As far as I can tell, on the `rtd-sphinx-theme` website the example that is a function does also not have these grey boxes [here](https://sphinx-rtd-theme.readthedocs.io/en/latest/demo/api.html) (just above "5.2 C++ API"). In version [1.0.0](https://sphinx-rtd-theme.readthedocs.io/en/1.0.0/demo/api.html) these examples are missing, they were introduced in 1.1.0. So my guess is that between 1.0.0 and 1.1.0 they have changed the styling to make a distinction between class and function parameters.

Thats great macOS build was okay, did you get warnings? BTW how do you generate the sharable link of your build, thats really cool!

JoeZiminski:
BTW @zm711 I am having go mess with this due to a weird test failure, if you pull now there are 0 changes!

zm711:
I'm actually going to ping @h-mayorquin too. That fact that you're testing with 0 changes and importing on ubuntu took 17 minutes is weird no?

JoeZiminski:
Thanks that is strange, I pushed again but [this is the run](https://github.com/SpikeInterface/spikeinterface/actions/runs/9111761651/job/25049570635)

h-mayorquin:
Yes, so confusing why the import times are so long. They did not used to be like this but this is not exclusive to this PR:

Here are the times here
https://github.com/SpikeInterface/spikeinterface/actions/runs/9111761651

And this is another PR:

https://github.com/SpikeInterface/spikeinterface/actions/runs/9104003224?pr=2858

I wonder when they became that bad for ubuntu and windows. As you can see is comparison, curation and exporters.

h-mayorquin:
```python

E       AttributeError: module 'matplotlib.cm' has no attribute 'get_cmap'

```

We need to pin matplotlib, right now it is not clear that it should be high or low. I was not even aware that 3.0 is out! (damn : /).

Somebody wants to go and check with the release notes when the change was introduced? This should be a different PR issue as it is not related to the docs per se.

JoeZiminski:
Thanks @h-mayorquin good catch! That was driving me insane, I could not figure out why adding one simple dependency (that is already installed in `full`), but the change seems to trigger re-installing all packages rather than using some kind of cache. 

Seems to be this change [here](https://github.com/SpikeInterface/spikeinterface/actions/runs/9112219162/job/25051018131), agree opening a new PR. Shall we try fixing that specific error and see if tests pass on 3.9.0?

JoeZiminski:
Great I see @h-mayorquin has opened issue #2863. I've restored all changes on this PR and will wait for a decision over there and PR before tests pass. I think @zm711 if you wanted to test the new build it's still worth doing as this matplotlib test error should (?) not affect it.

chrishalcrow:
> Thanks a lot @chrishalcrow I just blindly assumed this was due to some RTD stylistic change but never actually checked. I think the reason is that for classes the parameters are in grey boxes but for functions the arguments are not. As far as I can tell, on the `rtd-sphinx-theme` website the example that is a function does also not have these grey boxes [here](https://sphinx-rtd-theme.readthedocs.io/en/latest/demo/api.html) (just above "5.2 C++ API"). In version [1.0.0](https://sphinx-rtd-theme.readthedocs.io/en/1.0.0/demo/api.html) these examples are missing, they were introduced in 1.1.0. So my guess is that between 1.0.0 and 1.1.0 they have changed the styling to make a distinction between class and function parameters.

The problem is that both `plot_unit_waveforms_density_map` and `plot_unit_waveforms` are classes. But sphinx is treating `plot_unit_waveforms_density_map` as a function. I'm not sure why.

> Thats great macOS build was okay, did you get warnings? BTW how do you generate the sharable link of your build, thats really cool!

Around 15 warnings, depending on whether networkx is there or not, which is roughly the same as running the old version.

It's actually your build! I believe that all pull requests auto-generate a docs build, hosted at e.g. https://spikeinterface--2861.org.readthedocs.build/en/2861/index.html

JoeZiminski:
I *think* it is because SpikeInterface is going a class-to-function conversion (e.g. used as function [here](https://github.com/SpikeInterface/spikeinterface/blob/024c24afaec2d0a0476017c77a75f7c8a1dcb4b3/examples/tutorials/widgets/plot_3_waveforms_gallery.py#L78) and converted [here](https://github.com/SpikeInterface/spikeinterface/blob/024c24afaec2d0a0476017c77a75f7c8a1dcb4b3/src/spikeinterface/widgets/widget_list.py#L137)). Because they are intended to be used as functions they are requested to be documented as functions [here](https://github.com/SpikeInterface/spikeinterface/blob/024c24afaec2d0a0476017c77a75f7c8a1dcb4b3/doc/api.rst?plain=1#L302). So (I believe) that is the intended behaviour.

Great thanks, interesting about the warning number, I still get ~170 on macOS but they are all [this warning](https://stackoverflow.com/questions/62631362/get-rid-of-duplicate-label-warning-in-sphinx) which seems easy to solve. But I wonder still if it's something with my setup probably so if you only see 15 I will see what others report! 

> It's actually your build! I

That's awesome! I'm going to copy this machinery for our repos! thanks

samuelgarcia (approved):

PR #2864, opened 2024-05-16
Fix highpass-spatial-filter return dtype

### LINKED ISSUE #2311: `highpass_spatial_filter` returns as `float32`
I just realised `highpass_spatial_filter` is returning as `float32`. [here](https://github.com/SpikeInterface/spikeinterface/blob/8e32955e0616b9e399c253aaf0602f3aa646dae4/src/spikeinterface/preprocessing/highpass_spatial_filter.py#L235). (cast during scipy filtering I think). Should this be cast to `self.dtype` similar to [filter.py](https://github.com/SpikeInterface/spikeinterface/blob/8e32955e0616b9e399c253aaf0602f3aa646dae4/src/spikeinterface/preprocessing/filter.py#L160)?

I also see the `uint` is forced to float32 in filtering [here](https://github.com/SpikeInterface/spikeinterface/blob/8e32955e0616b9e399c253aaf0602f3aa646dae4/src/spikeinterface/preprocessing/filter.py#L140). Should this also be a consideration for `highpass_spatial_filter`?

alejoe91:
Fixes #2311

@JoeZiminski @h-mayorquin

### DIFF
--- a/src/spikeinterface/preprocessing/tests/test_highpass_spatial_filter.py
+++ b/src/spikeinterface/preprocessing/tests/test_highpass_spatial_filter.py
@@ -125,7 +125,7 @@ def get_ibl_si_data():
     ibl_data = ibl_recording.read(slice(None), slice(None), sync=False)[:, :-1].T  # cut sync channel
 
     si_recording = se.read_spikeglx(local_path, stream_id="imec0.ap")
-    si_recording = spre.scale(si_recording, dtype="float32")
+    si_recording = spre.astype(si_recording, dtype="float32")
 
     return ibl_data, si_recording
 
--- a/src/spikeinterface/preprocessing/highpass_spatial_filter.py
+++ b/src/spikeinterface/preprocessing/highpass_spatial_filter.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import numpy as np
 
 from .basepreprocessor import BasePreprocessor, BasePreprocessorSegment
+from .filter import fix_dtype
 from ..core import order_channels_by_depth, get_chunk_with_margin
 from ..core.core_tools import define_function_from_class
 
@@ -117,6 +118,8 @@ class HighpassSpatialFilterRecording(BasePreprocessor):
         butter_kwargs = dict(btype="highpass", N=highpass_butter_order, Wn=highpass_butter_wn)
         sos_filter = scipy.signal.butter(**butter_kwargs, output="sos")
 
+        dtype = fix_dtype(recording, dtype)
+
         for parent_segment in recording._recording_segments:
             rec_segment = HighPassSpatialFilterSegment(
                 parent_segment,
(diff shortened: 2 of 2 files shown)


JoeZiminski:
This looks great, I'll quickly test it now

JoeZiminski:
Hey Alessio I couldn't push to your fork, how are such issues usually managed? 

The test I added in `test_highpass_spatial_file.py` is below. I had to add the properties so the data could be returned scaled, but I think maybe there is a better way. It would be cool to generalise this test to all preprocessing steps, but my guess is this would be tricky!

```
@pytest.mark.parametrize("dtype", [np.int16, np.float32, np.float64])
def test_dtype_stability(dtype):
    """
    Check that the dtype of the recording and
    output data is as expected, as data is cast to float32
    during filtering.
    """
    num_chan = 32
    si_recording = generate_recording(num_channels=num_chan, durations=[2])
    si_recording.set_property("gain_to_uV", np.ones(num_chan))
    si_recording.set_property("offset_to_uV", np.ones(num_chan))
    si_recording = spre.astype(si_recording, dtype)

    assert si_recording.dtype == dtype

    highpass_spatial_filter = spre.highpass_spatial_filter(si_recording, n_channel_pad=2)

    assert highpass_spatial_filter.dtype == dtype

    filtered_data_unscaled = highpass_spatial_filter.get_traces(return_scaled=False)

    assert filtered_data_unscaled.dtype == dtype

    filtered_data_scaled = highpass_spatial_filter.get_traces(return_scaled=True)

    assert filtered_data_scaled.dtype == np.float32
```

alejoe91:
I think that only maintainers can push! I'll add the suggested test :)

JoeZiminski:
Great thanks, LGTM!

src/spikeinterface/preprocessing/tests/test_highpass_spatial_filter.py
```diff
@@ -109,6 +109,34 @@ def test_highpass_spatial_filter_synthetic_data(num_channels, ntr_pad, ntr_tap,
...
+
+    assert highpass_spatial_filter.dtype == dtype
+
+    filtered_data_unscaled = highpass_spatial_filter.get_traces(return_scaled=False)
+
+    assert filtered_data_unscaled.dtype == dtype
+
+    filtered_data_scaled = highpass_spatial_filter.get_traces(return_scaled=True)
+
+    assert filtered_data_scaled.dtype == np.float32
```
h-mayorquin, line 137: Shouldn't scaled return np.float64 if the non-scaled is float64?

alejoe91: no because scaling casts to float: https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/baserecording.py#L373

h-mayorquin (approved):
Looks good to me. Let's open an issue for the get_traces when the dtype is already float.

PR #2873, opened 2024-05-17
Improve `save_to_folder` docstring

h-mayorquin:
The old one is confusing and has references to `extractor_name` which can be confusing as some classess do have that attribute (at least for now).

### DIFF
--- a/src/spikeinterface/extractors/nwbextractors.py
+++ b/src/spikeinterface/extractors/nwbextractors.py
@@ -18,7 +18,6 @@ def read_file_from_backend(
     cache: bool = False,
     stream_cache_path: str | Path | None = None,
     storage_options: dict | None = None,
-    backend: Literal["hdf5", "zarr"] | None = None,
 ):
     """
     Reads a file from a hdf5 or zarr backend.
@@ -87,7 +86,7 @@ def read_file_from_backend(
     else:
         import h5py
 
-        assert file is not None, "Unexpected, file is None"
+        assert file is not None, "Both file_path and file are None"
         open_file = h5py.File(file, "r")
 
     return open_file
@@ -124,11 +123,6 @@ def read_nwbfile(
     -------
     nwbfile : NWBFile
         The NWBFile object.
-
-    Raises
-    ------
-    AssertionError
-        If ROS3 support is not enabled.
 
     Notes
     -----
@@ -277,14 +271,17 @@ def _retrieve_unit_table_pynwb(nwbfile: "NWBFile", unit_table_path: Optional[str
 
 
 def _is_hdf5_file(filename_or_file):
-    # Source for magic numbers https://www.loc.gov/preservation/digital/formats/fdd/fdd000229.shtml
-    # We should find a better one though
     if isinstance(filename_or_file, (str, Path)):
-        with open(filename_or_file, "rb") as f:
-            file_signature = f.read(8)
+        import h5py
+
+        filename = str(filename_or_file)
+        is_hdf5 = h5py.h5f.is_hdf5(filename.encode("utf-8"))
     else:
         file_signature = filename_or_file.read(8)
-    return file_signature == b"\x89HDF\r\n\x1a\n"
+        # Source of the magic number https://docs.hdfgroup.org/hdf5/develop/_f_m_t3.html
+        is_hdf5 = file_signature == b"\x89HDF\r\n\x1a\n"
+
+    return is_hdf5
 
 
 def _get_backend_from_local_file(file_path: str | Path) -> str:
@@ -680,7 +677,6 @@ class NwbRecordingExtractor(BaseRecording):
 
     def _fetch_recording_segment_info_backend(self, file, cache, load_time_vector, samples_for_rate_estimation):
         open_file = read_file_from_backend(
-            backend=self.backend,
             file_path=self.file_path,
             file=file,
             stream_mode=self.stream_mode,
@@ -862,6 +858,57 @@ class NwbRecordingExtractor(BaseRecording):
         locations, groups = self._fetch_locations_and_groups(electrodes_table, electrodes_indices)
 
         return gains, offsets, locations, groups
+
+    @staticmethod
+    def fetch_available_electrical_series_paths(
+        file_path: str | Path, stream_mode: Optional[Literal["fsspec", "remfile", "zarr"]] = None
+    ) -> List[str]:
+        """
+        Retrieves the paths to all ElectricalSeries objects within a neurodata file.
+
+        Parameters
+        ----------
+        file_path : str | Path
+            The path to the neurodata file to be analyzed.
+        stream_mode : "fsspec" | "remfile" | "zarr" | None, optional
+            Determines the streaming mode for reading the file. Use this for optimized reading from
+            different sources, such as local disk or remote servers.
+
+        Returns
+        -------
+        list of str
+            A list of paths to all ElectricalSeries objects found in the file.
+
+
+        Notes
+        -----
+        The paths are returned as strings, and can be used to load the desired ElectricalSeries object.
+        Examples of paths are:
+            - "acquisition/ElectricalSeries1"
+            - "acquisition/ElectricalSeries2"
+            - "processing/ecephys/LFP/ElectricalSeries1"
+            - "processing/my_custom_module/MyContainer/ElectricalSeries2"
+        """
+
+        if stream_mode is None:
+            backend = _get_backend_from_local_file(file_path)
+        else:
+            if stream_mode == "zarr":
+                backend = "zarr"
+            else:
+                backend = "hdf5"
+
+        file_handle = read_file_from_backend(
+            file_path=file_path,
+            stream_mode=stream_mode,
+        )
+
+        electrical_series_paths = _find_neurodata_type_from_backend(
+            file_handle,
+            neurodata_type="ElectricalSeries",
+            backend=backend,
+        )
+        return electrical_series_paths
 
 
 class NwbRecordingSegment(BaseRecordingSegment):
@@ -1132,7 +1179,6 @@ class NwbSortingExtractor(BaseSorting):
         self, unit_table_path: str = None, samples_for_rate_estimation: int = 1000, cache: bool = False
     ):
         open_file = read_file_from_backend(
-            backend=self.backend,
             file_path=self.file_path,
             stream_mode=self.stream_mode,
             cache=cache,
--- a/src/spikeinterface/extractors/tests/test_nwbextractors.py
+++ b/src/spikeinterface/extractors/tests/test_nwbextractors.py
@@ -319,6 +319,22 @@ def test_retrieving_from_processing(generate_nwbfile, use_pynwb):
     assert np.array_equal(electrical_series_custom.data[:], recording_extractor_custom.get_traces())
 
 
+def test_fetch_available_electrical_series_paths(generate_nwbfile):
+    path_to_nwbfile, _ = generate_nwbfile
+    available_electrical_series = NwbRecordingExtractor.fetch_available_electrical_series_paths(
+        file_path=path_to_nwbfile
+    )
+
+    expected_paths = [
+        "acquisition/ElectricalSeries1",
+        "acquisition/ElectricalSeries2",
+        "processing/ecephys/LFP/ElectricalSeries1",
+        "processing/my_custom_module/MyContainer/ElectricalSeries2",
+    ]
+
+    assert available_electrical_series == expected_paths
+
+
 @pytest.mark.parametrize("electrical_series_path", ["acquisition/ElectricalSeries1", "acquisition/ElectricalSeries2"])
 def test_recording_equality_with_pynwb_and_backend(generate_nwbfile, electrical_series_path):
     path_to_nwbfile, _ = generate_nwbfile
@@ -338,7 +354,7 @@ def test_recording_equality_with_pynwb_and_backend(generate_nwbfile, electrical_
 
 
 @pytest.mark.parametrize("use_pynwb", [True, False])
-def test_failure_with_wrong_electrical_series_name(generate_nwbfile, use_pynwb):
+def test_failure_with_wrong_electrical_series_path(generate_nwbfile, use_pynwb):
     """Test that the extractor raises an error if the electrical series name is not found."""
     path_to_nwbfile, _ = generate_nwbfile
     with pytest.raises(ValueError):


src/spikeinterface/core/base.py
```diff
@@ -846,43 +846,63 @@ def save_to_memory(self, sharedmem=True, **save_kwargs) -> "BaseExtractor":
...
-          * extracting traces by calling get_trace() method in chunks
-          * saving data into file (memmap with BinaryRecordingExtractor)
-          * dumping to json/pickle the original extractor for provenance
-          * dumping to json/pickle the cached extractor (memmap with BinaryRecordingExtractor)
+        This method extracts trace data, saves it to a file (using a memory-mapped format
+        with BinaryRecordingExtractor), and stores both the original extractor's provenance
+        and the cached extractor's metadata in JSON format.
 
-        This replaces the use of the old CacheRecordingExtractor and CacheSortingExtractor.
+        The folder final location and name can be specified in a couple of ways ways:
```
zm711, line 864:
```suggestion
        The folder's final location and name can be specified in a couple of ways:
```

src/spikeinterface/core/base.py
```diff
@@ -846,43 +846,63 @@ def save_to_memory(self, sharedmem=True, **save_kwargs) -> "BaseExtractor":
         return cached
 
     # TODO rename to saveto_binary_folder
-    def save_to_folder(self, name=None, folder=None, overwrite=False, verbose=True, **save_kwargs):
+    def save_to_folder(
+        self,
+        name: str = None,
```
zm711, line 851:
```suggestion
        name: str | Path | None = None,
```

Or you don't like adding None as a potential type hint? And below you say Path so it needs Path at minimum.

src/spikeinterface/core/base.py
```diff
@@ -846,43 +846,63 @@ def save_to_memory(self, sharedmem=True, **save_kwargs) -> "BaseExtractor":
...
+        Save the extractor and its data to a folder.
 
-        The save consist of:
-          * extracting traces by calling get_trace() method in chunks
-          * saving data into file (memmap with BinaryRecordingExtractor)
-          * dumping to json/pickle the original extractor for provenance
-          * dumping to json/pickle the cached extractor (memmap with BinaryRecordingExtractor)
+        This method extracts trace data, saves it to a file (using a memory-mapped format
+        with BinaryRecordingExtractor), and stores both the original extractor's provenance
+        and the cached extractor's metadata in JSON format.
```
zm711, line 862:
```suggestion
        and the cached extractor's metadata in JSON format.
```

Is pickle no longer possible at all?

src/spikeinterface/core/base.py
```diff
@@ -846,43 +846,63 @@ def save_to_memory(self, sharedmem=True, **save_kwargs) -> "BaseExtractor":
...
-        Save extractor to folder.
+        Save the extractor and its data to a folder.
 
-        The save consist of:
-          * extracting traces by calling get_trace() method in chunks
-          * saving data into file (memmap with BinaryRecordingExtractor)
-          * dumping to json/pickle the original extractor for provenance
-          * dumping to json/pickle the cached extractor (memmap with BinaryRecordingExtractor)
+        This method extracts trace data, saves it to a file (using a memory-mapped format
+        with BinaryRecordingExtractor), and stores both the original extractor's provenance
```
zm711, line 861:
```suggestion
        This method extracts trace data, saves it to a file (using the memory-mapped
        BinaryRecordingExtractor format), and stores both the original extractor's provenance
```
would probably be the slightly more standard short way to write this.

src/spikeinterface/core/base.py
zm711: Didn't mean this one as a suggestion. Oops!

zm711 (commented):
A few comments + typo fixes :)

src/spikeinterface/core/base.py
h-mayorquin: I think that's good. But I don't think name can take a Path. Am I wrong?

Side comment, talking about pet peeves this is one of mine. I actually prefer Optional[str] instead of str | None.  But  I think this is inconsequential and I have been following your standard in this one.

src/spikeinterface/core/base.py
h-mayorquin: Thanks, once we decide on the Path name I will batch all of these suggestions.

src/spikeinterface/core/base.py
h-mayorquin: Not for this function I think. Looking a the body of the function it does not seem like it is an option.

src/spikeinterface/core/base.py
zm711: So my comment was originally because the docstring below said str or Path. But looking at the code it has to be a path because of this line

```python
folder = cache_folder / name
```
`/` can't be used with string. Or we need to do name = Path(name) to make sure it is a path first.

zm711: > Side comment, talking about pet peeves this is one of mine. I actually prefer Optional[str] instead of str | None. But I think this is inconsequential and I have been following your standard in this one.

I'm okay with Optional[str], but I don't like when someone does Optional[list[float|int']] because it gets too nested. I don't like the numpy default of doing argument, optional to indicate that a default argument is supplied.

src/spikeinterface/core/base.py
```diff
@@ -846,43 +846,63 @@ def save_to_memory(self, sharedmem=True, **save_kwargs) -> "BaseExtractor":
...
         ----------
-        name: None str or Path
-            Name of the subfolder in get_global_tmp_folder()
-            If "name" is given, "folder" must be None.
-        folder: None str or Path
-            Name of the folder.
-            If "folder" is given, "name" must be None.
-        overwrite: bool, default: False
-            If True, the folder is removed if it already exists
+        name : str or Path, optional
```
zm711, line 884:
```suggestion
        name : str or Path, optional
```

currently this is wrong, because only a path would work.

And this type of optional is my pet peeve :)

src/spikeinterface/core/base.py
zm711: Cool sorry I didn't look at the function I was just looking at the diff! We can skip this one.

src/spikeinterface/core/base.py
h-mayorquin: > / can't be used with string. Or we need to do name = Path(name) to make sure it is a path first.

You totally can:

```python
In [1]: from pathlib import Path

In [2]: Path.cwd() / "here"
Out[2]: PosixPath('/home/heberto/development/neuroconv/here')
```

Concerning the typing discussion, this is just a small preference. I would rather keep doing it consistently with what you have been doing : )

src/spikeinterface/core/base.py
h-mayorquin: Good.

src/spikeinterface/core/base.py
zm711: Sorry you're right. I wasn't thinking that folder could be a string. But you can also do:
```
first_path = Path('.')
second_path = Path('folder')
full_path = first_path / second_path
```

So it can be a string or Path.

h-mayorquin: Oh fair enough. 
I guess that you are right that it **can** be a path and that would work. I don't think this is what it was intended and I would rather not give a type hint for this.

zm711: Then I would delete it from the docstring as well.

h-mayorquin: Oh, yes, sure! 

I think that the intention is for you to pass the name of the file. As you mentioned you could pass Path("name") and that would work. But the reason that I think is a bad idea is because if someone is relying on a linter and then they pass Path("C:/something_here/here_something_else/") then that would throw an error but the linter would accept it. 

To be honest, my intuition is that folder is sufficient as it can do the two things. But I think that probably at some point it was found convenient to save to the tmp folder but still name it.

src/spikeinterface/core/base.py
h-mayorquin: What do you mean? what are the preferences with option and what is your preference?

zm711: Sorry so it is confusing, but there are two uses of optional--one from std library meaning `value | None`

```python
from typing import Optional
Optional[float]
```

And one from numpy docstring style, which is 

```python
def function(my_arg=1)
 """ my_arg: int, optional
```

notice the difference between Optional (from typing) and optional (lowercase numpy docstring convention)

The difference being Optional says an argument is actually optional (for example you choose one between folder or name so you have the option to pick). optional means that a default is already supplied to the argument of the function so you can optionally change the default value. 

I don't like the numpy style because I prefer to be more explicit in the docstring so instead I prefer
```
my_arg: int, default =1 
```

instead of, 
```
my_arg: int, optional
```

So I've worked toward removing all mentions of argument, default_value, optional from the library.
If someone prefers Optional[x] to mean a default of None that doesn't really bother me.

src/spikeinterface/core/base.py
h-mayorquin: Should be done.

src/spikeinterface/core/base.py
zm711: technically Optional[float], optional would be an appropriate numpy docstring because a default is supplied in the arguments of the function. Which I don't like.

h-mayorquin: I don't think that you are right about the meaning of `typing.Optional`:

https://docs.python.org/3/library/typing.html#typing.Optional

> Note that this is not the same concept as an optional argument, which is one that has a default. An optional argument with a default does not require the Optional qualifier on its type annotation just because it is optional. For example:

> def foo(arg: int = 0) -> None:
    ...
> On the other hand, if an explicit value of None is allowed, the use of Optional is appropriate, whether the argument is optional or not. For example:

Optional means that it can be None.

h-mayorquin: But if anything that makes the need for your disctinction stronger,  if there is a default value, you should say what the default values is.

h-mayorquin: I am getting you right if I think that the current docstring is working well with your preference?

src/spikeinterface/core/base.py
```diff
@@ -846,43 +846,63 @@ def save_to_memory(self, sharedmem=True, **save_kwargs) -> "BaseExtractor":
...
-            If "name" is given, "folder" must be None.
-        folder: None str or Path
-            Name of the folder.
-            If "folder" is given, "name" must be None.
-        overwrite: bool, default: False
-            If True, the folder is removed if it already exists
+        name : str , optional
+            The name of the subfolder within the global temporary folder. If `folder`
+            is provided, this argument must be None.
+        folder : str or Path, optional
```
zm711, line 887: folder : Optional[str | Path], default: None

or

folder: str | Path | None , default: None

would be my two preferences. Using `,optional` isn't the `Optional` I'm okay with.

sorry this is tricky. I think my own confusion came from the fact the way I was trying to describe Optional how I thought it would work:

I never expected a function to do `(argument: Optional[str]= 'home_folder')` instead of `(argument: Optional[str] = None)`. But technically you use Optional for both. It just seems like poor design to accept None if it isn't the default argument for me.

Basically numpy optional is to indicated required vs optional arguments, which to me already seems clear based on the function parameters.

```
func(x, y=1)
"""x: float
     y, optional
"""
```

h-mayorquin:
```
I never expected a function to do (argument: Optional[str]= 'home_folder') instead of (argument: Optional[str] = None). But technically you use Optional for both. It just seems like poor design to accept None if it isn't the default argument for me.
```

I agree with this, I would not like to use optional in that case.

I am more confused about numpy convention, can you link it?
My own preference would be to use default=value if there is a default that is not None  and optional to mean that you don't have to specify it.

zm711: https://numpydoc.readthedocs.io/en/latest/format.html

Here's their style guide. They don't try to force you to use optional, but if you use vscode numpydoc style then it autofills in the optional. So it is a recommendation that I think the auto-docstring complete takes too far.

zm711: So I shouldn't blame numpydoc, but instead I should blame our overreliance on tooling :)

h-mayorquin: OK, so numpy doc optional is more expansive than the one in typing.

So, what do you think of using the following option:
* optional only, when is optional in the typing sense (argument=None)
* default, when there is a default value that is not None.

It seems consistent with their docs as they say
> If the default value would not be used as a value, optional is preferred. These are all equivalent:

So, the docstring here is like this:
values that are optional in the typing sense are marked as optional
values that have a default are marksed as default.

What do you think?

src/spikeinterface/core/base.py
```diff
@@ -846,43 +846,63 @@ def save_to_memory(self, sharedmem=True, **save_kwargs) -> "BaseExtractor":
...
-        Save extractor to folder.
+        Save the extractor and its data to a folder.
 
-        The save consist of:
-          * extracting traces by calling get_trace() method in chunks
-          * saving data into file (memmap with BinaryRecordingExtractor)
-          * dumping to json/pickle the original extractor for provenance
-          * dumping to json/pickle the cached extractor (memmap with BinaryRecordingExtractor)
+        This method extracts trace data, saves it to a file (using the memory-mapped
+        BinaryRecordingExtractor format), and stores both the original extractor's provenance
```
alejoe91, line 861: This can be used also to save a `Sorting` object!

src/spikeinterface/core/base.py
```diff
@@ -846,43 +846,63 @@ def save_to_memory(self, sharedmem=True, **save_kwargs) -> "BaseExtractor":
...
+        verbose : bool, default: True
+            If True, print information about the cache folder being used.
+        **save_kwargs
+            Additional keyword arguments to be passed to the underlying save method.
 
         Returns
         -------
-        cached: saved copy of the extractor.
+        cached_extractor
+            A saved copy of the extractor in the specified format (e.g., BinaryRecordingExtractor).
```
alejoe91, line 900: Actually this function returns a `BinaryFolderRecording`, but it's different for the `Sorting` object

src/spikeinterface/core/base.py
alejoe91: So it can be a:
- `BinaryFolderRecording` for the recording
- `NumpySortingFolder` for the sorting

h-mayorquin: Yeah, sure.

src/spikeinterface/core/base.py
h-mayorquin: Yep. I rather make no reference to the type though. I don't know if they are helpful.

src/spikeinterface/core/base.py
h-mayorquin: Actually, I would rather remove mentioned of the specific type. I don't think those details are useful.

src/spikeinterface/core/base.py
h-mayorquin: Can I mark this as resolved, @zm711? I moved the discussion to https://github.com/SpikeInterface/SpikeInterface-Hackathon-Edinburgh-May24/issues/29

src/spikeinterface/core/base.py
h-mayorquin: Removed.

src/spikeinterface/core/base.py
h-mayorquin: Removed.

src/spikeinterface/core/base.py
h-mayorquin: @zm711 ?

h-mayorquin: Thanks, man.

h-mayorquin:
This is ready.

PR #2874, opened 2024-05-17
Remove unused extractor attribute `extractor_name`

h-mayorquin:
Ok, this is within the context of the discussion in https://github.com/SpikeInterface/spikeinterface/pull/2784

This is an unused attribute as far as my search skills in the repo point out. There is a mention to it ne neo utils but reading the function it actually does not use `extractor_name` it but uses the `name` attribute:

https://github.com/h-mayorquin/spikeinterface/blob/5d743469765f60ed5fbfaa819e45a5ae457d7722/src/spikeinterface/extractors/neoextractors/neo_utils.py#L6-L29

In general I think that there are some equivocations between `class.__name__` and the `name` attribute on the package. I did a small change in the repr to disambiguate this.


h-mayorquin:
Requires https://github.com/SpikeInterface/spikeinterface/pull/2876 for some reason : O

samuelgarcia (approved):

samuelgarcia (approved):

PR #2876, opened 2024-05-17
Protect hdbscan import in `TdCClustering`

h-mayorquin:
This is causing some problems on import tests. 

Would that work like this?


alejoe91:
LGTM!

samuelgarcia:
merci

PR #2878, opened 2024-05-17
Handle ms_before and ms_after difference between waveforms and template extensions

h-mayorquin:
They might be different, we can either raise an error like here or change to a warning and priviledge the ones of the waveform as we are doing right now.

What do you guys think?

### DIFF
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -298,6 +298,12 @@ class ComputeTemplates(AnalyzerExtension):
 
         waveforms_extension = self.sorting_analyzer.get_extension("waveforms")
         if waveforms_extension is not None:
+
+            if ms_before != waveforms_extension.ms_before or ms_after != waveforms_extension.ms_after:
+                raise ValueError(
+                    "ComputeTemplates: ms_before and ms_after must be the same as the ones used for the 'waveforms' extension"
+                )
+
             nbefore = waveforms_extension.nbefore
             nafter = waveforms_extension.nafter
         else:


src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -298,6 +298,17 @@ def _set_params(self, ms_before: float = 1.0, ms_after: float = 2.0, operators=N
...
+            ms_before_waveforms = waveforms_extension.nbefore / self.sorting_analyzer.sampling_frequency * 1000
+            ms_after_waveforms = waveforms_extension.nafter / self.sorting_analyzer.sampling_frequency * 1000
+
+            if ms_before != ms_before_waveforms or ms_after != ms_after_waveforms:
+                raise ValueError(
+                    "templates extension parameters are inconsistent with the waveforms extension. \n"
+                    f"waveform extension ms_before={ms_before_waveforms}  ms_after={ms_after_waveforms} \n"
+                    f"templates extension ms_before={ms_before}  ms_after={ms_after}."
+                )
+
```
samuelgarcia, line 311: I think I prefered the previous behavior:
  * use parameters when compute directly
  * ignore parameters and hinertis from waveforms extension when this exists.

alejoe91: Agree with @samuelgarcia here

alejoe91:
@h-mayorquin after discussing with Sam, we think it's ok that waveforms control the ms_before/after, let's just be clearer with this in the docs without too many warnings

h-mayorquin:
All right. Where should this behavior be written?

alejoe91:
@h-mayorquin adding a comment [here](https://spikeinterface.readthedocs.io/en/latest/modules/core.html#sortinganalyzer) in the `Templates` section should be enough

alejoe91:
Made a PR with a doc clarification on the inheritance of ms_before/after. Let's close this one
https://github.com/SpikeInterface/spikeinterface/pull/3663

h-mayorquin:
Thanks @alejoe91

PR #2879, opened 2024-05-17
Try "Handle motion/drift" documentation as a sphinx gallery

### LINKED ISSUE #2881: Handling documentation that takes a long time to build
At present there are a few examples the documentation pages that take a long time to build or require files not available to user.  These are located in the [examples/howto](https://github.com/SpikeInterface/spikeinterface/tree/main/examples/how_to) folder. Ideally, these would use an easily obtainable dataset so the user could run the scripts alongside the example. But this a relatively easy thing to address (e.g. using SI generate recordings machinery or making specific datasets publicly available for `si.download_dataset`) and not the purpose of this issue.

In part because these pages (at least the handle drift page) take a long time to run they need to be handled differently to the other sphinx-gallery pages. It can take (in this PR, swapping out the file for SI generate recordings machinery) around 25 minutes, so too long for everyday use.  At present, they must be built manually and the output moved to the `doc/howto` folder. The significant benefit of this is that this part of the documentation is only built when it needs to be. However, it would be nice to extend this method such that:

1) We can use the sphinx-gallery machinery that the rest of the docs are built with 
2) avoid having to build manually

but still:
3) build this longer-part of the docs only when required.

There are two methods suggested below, the first achieves all goals above but might slow down CI too much day-to-day, I am not familiar enough on this so would be good to get feedback. The second method acheives (1), (3) and simplifies the building to a single command, but it must still be run manually and the resulting page cannot be easily included in a sphinx-gallery.

### Method 1

This way is exemplified in #2879 it would be great to get feedback on, using the 'Handle drift' how-to as an example. The long-to-build docs are written as a `.py` file in sphinx-gallery style and stored with the rest of the tutorials. However, the filename is `long_plot_handle_drift.py` and so by default, sphinx-gallery does not run the script and build with plots. Instead it only converts it to RST and builds very quickly, there are code examples but no plots. To build with plots, a 'tag' is added to the`sphinx-build` command that is handled in the `conf.py` (e.g. `sphinx-build -b html source _build -t handle_drift`). When the tag is provided `"long_plot_handle_drift"` is added to the regexp that finds docs to fully build and the page is build with plots.

```
if tags.has("handle_drift") or tags.has("all_long_plot"):
     sphinx_gallery_conf["filename_pattern"] += '|/long_plot_handle_drift'
```

This way, for day-to-day use the docs can be built such that the long-form tutorials are not fully run. However, when building for release or working specifically on these pages, you can use the tag to build them. The problem is that in the CI, I am not sure exactly when the docs are built. If the built time for the docs goes up to around 25 minutes, , if the full build is run only on merge, or version release, is it a problem? 

### Method 2

Method 2 is similar to method 1, it basically replicates the current way of doing it but makes it easier to run. In this case, you store long-to-build pages in a separate `examples/long_tutorials` folder. These will be output to another folder (`docs/long_tutorials`) when build with sphinx-gallery. Unlike `docs/tutorials`, `docs/long_tutorials` is not added to the `.gitignore` (just like at present, the `docs/howto` is not in the gitignore). Now, instead of using the regexp to fully build the pages, the `conf.py` contains: 

```
if tags.has("handle_drift") or tags.has("all_long_plot"):
     sphinx_gallery_conf['examples_dirs'].append('../examples/long_tutorials/handle_drift'])
     sphinx_gallery_conf['gallery_dirs'].append('long_tutorials/handle_drift')
```

Now when building with `sphinx-build -b html source _build -t handle_drift` these pages are rebuilt as a sphinx-gallery  and the outputs in `docs/long_tutorials` is overwritten (unless there are no changes since last build). The outputs permentantly sit in `long_tutorials` (just like they do now in `how to`) and can be rebuilt and pushed at will. 

A downside is that they are a separate gallery, so you need to link to these pages manually (e.g. from How To). Alternatively, we could have two tutorials gallery, one is quick-to-run and the other are slower to run (as on the user end, even though they can now follow along with the tutorial it takes 25 mins to run). 

I think I am slightly in favour of 'method `1' because you can integrate it fully with existing pages, it is a bit simpler and fully automated. However, maybe it will cause problems with the CI. Will be great to hear what people think.

JoeZiminski:
This PR adds a sphinx-gallery page version of the current "[Handle motion/drift with spikeinterface](https://spikeinterface.readthedocs.io/en/latest/how_to/handle_drift.html)".  It is an implementation of 'method 2' as described in further detail in #2881 . Very keen to get feedback on the general long-docs build approach + the actual content of the page. Currently the new and old version are both there (this version is renamed and appended with NEW) for comparisons sake).

This PR explores how the current 'How To' files that are build manually could be automated, using the [Handle Drift](https://spikeinterface.readthedocs.io/en/latest/how_to/handle_drift.html) page as an example.

The changes are:

1) The handle drift page was rewritten as a sphinx-gallery. It is now linked to, from the 'How To' page, similar to the existing  page on which it is based, but with 'NEW' at the end of the name.
2) Some code is added to the `conf.py` so that the page is only built with plots when a tag `-t handle_drift` is added. It outputs to a 'long_tutorials/handle_drift' folder.

So similar to the previous way of doing it, you build these long-to-run pages, it outputs to a specific folder and the images etc. are shared on github. However, now rather than manually rebuilding they are generated with an optional tag on the `sphinx-build` command. To take a look at this documentation you can pull the PR branch and build the docs in the usual way (without the `-t handle_drift`) and the page should be there. If you wanted to make a change to the documentation, you would build with `-t handle_drift` and push the results to `git`.

## A note on reviewing

This PR keeps the old paragraph on interpreting the motion correction outputs from the previous iteration of this docs. This was based on the kilosort outputs and will not make sense here. However, I thought it made sense to leave this and then figure out the best parameters to use, before re-runnign and writing a new paragraph to interpret the outputs. The section I am referring to is `# A few comments on the figures:`

@samuelgarcia @cwindolf this PR swaps out the kilosort recording for the `generate_drifting_recording()` in this example. You can see the generated plots in the PR changelog. It would be great if you could advise on the best settings to use for the synthetic recording.


JoeZiminski:
OK this is ready for review now.

doc/conf.py
```diff
@@ -130,9 +130,19 @@
     'within_subsection_order': FileNameSortKey,
     'ignore_pattern': '/generate_',
     'nested_sections': False,
-    'copyfile_regex': r'.*\.rst|.*\.png|.*\.svg'
+    'copyfile_regex': r'.*\.rst|.*\.png|.*\.svg',
+    'filename_pattern': '/plot_',
 }
 
+if tags.has("handle_drift") or tags.has("all_long_plot"):
```
chrishalcrow, line 137: Nice! Worth generalising this? Something like:

```suggestion
long_tutorial_names = ["handle_drift"]

for long_tutorial_name in long_tutorial_names:
    if tags.has(long_tutorial_name) or tags.has("all_long_plot"):

        if (tutorial_path := (Path('long_tutorials/'+long_tutorial_name))).is_dir():
            shutil.rmtree(tutorial_path)

        sphinx_gallery_conf['examples_dirs'].append('../examples/long_tutorials/'+long_tutorial_name)
        sphinx_gallery_conf["gallery_dirs"].append(tutorial_path.as_posix())
```

chrishalcrow:
I _love_ this! The local workflow is great: by default you do what you always do, then only need to add the tag when you're working on that doc. Perfect! I also think that sphinx_gallery files are easier to read/edit than the rst files (i.e. plot_handle_drift.py is much easier to edit than handle_drift.rst).

By default the CI won't build the long_tutorials, right? And from the discussion in #2881, the plan would be to run all the long tutorials occasionally, maybe before a release? Is that right?

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+    )
+    results[preset]["motion_info"] = motion_info
+
+# %%
+#.. seealso::
+#   It is often very useful to save ``motion_info`` to
+#   file, so it can be loaded and visualised later. This can be done by setting
+#   the ``folder`` argument of
+#   :py:func:`~spikeinterface.preprocessing.correct_motion()` to a path to write
+#   the motion output to. The ``motion_info`` can be loaded back with
```
zm711, line 185: Maybe we give the file suffix. the motion output to a `.npy` file (I forget the format to be honest) what do you think?

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+#   ``si.load_motion_info``.
+
+# %%
+# -------------------
```
zm711, line 189:
```suggestion
# --------------------
```

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+# Plotting the results
+# -------------------
```
zm711, line 191:
```suggestion
# --------------------
```

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+#
+# Next, let's plot the results of our motion estimation using the ``plot_motion_info()``
+# function. The plot contains 4 panels, on the x-axis of all plots we have
+# the (binned time). The plots display:
```
zm711, line 195: parentheses are strange here? where you going to put time (binned time)?

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+#   * **top left:** The estimated peak depth for every detected peak.
+#   * **top right:** The estimated peak depths after motion correction.
+#   * **bottom left:** The average motion vector across depths and all motion across spatial depths (for non-rigid estimation).
+#   * **bottom right:** if motion correction is non rigid, the motion vector across depths is plotted as a map, with the color code representing the motion in micrometers.
```
zm711, line 199:
```suggestion
#   * **bottom right:** if motion correction is non-rigid, the motion vector across depths is plotted as a map, with the color code representing the motion in micrometers.
```

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+    fig = plt.figure(figsize=(7, 7))
+
+    si.plot_motion_info(
+        results[preset]["motion_info"],
+        recording=corrected_recording,  # the recording is only used to get the real times
+        figure=fig,
+        depth_lim=(400, 600),
+        color_amplitude=True,
+        amplitude_cmap="inferno",
+        scatter_decimate=10,            # Only plot every 10th peak
```
zm711, line 212: Nice. This is another "term" that I think biologists/basic neuroscientists likely won't know so I like this comment a lot!

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+    fig.suptitle(f"{preset=}")
+
+# %%
+# These plots are quite complicated, so it is worth covering them in detail.
+# For every detected spike in our recording, we first estimate
+# its depth (first panel) using a method from
+# :py:func:`~spikeinterface.postprocessing.compute_unit_locations()`.
+#
+# Then, the probe motion is estimated the location of the detected peaks are
+# adjusted to account for this (second panel).
```
zm711, line 223: Not sure how to fix. But missing a word/words

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+# more detailed representation of the motion vectors. We can see the motion plotted
+# as a heatmap at each binned spatial depth across all time bins. It captures
+# the zigzag pattern (alternating light and dark colors) of the injected motion.
+
+# %%
+# A few comments on the figures:
+#   * The preset **'rigid_fast'** has only one motion vector for the entire probe because it is a 'rigid' case.
+#     The motion amplitude is globally underestimated because it averages across depths.
+#     However, the corrected peaks are flatter than the non-corrected ones, so the job is partially done.
+#     The big jump at=600s when the probe start moving is recovered quite well.
```
zm711, line 241:
```suggestion
#     The big jump at 600s when the probe start moving is recovered quite well.
```

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+#   * The preset **kilosort_like** gives better results because it is a non-rigid case.
+#     The motion vector is computed for different depths.
+#     The corrected peak locations are flatter than the rigid case.
+#     The motion vector map is still be a bit noisy at some depths (e.g around 1000um).
```
zm711, line 245:
```suggestion
#     The motion vector map is still a bit noisy at some depths (e.g around 1000um).
```

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+#   * The preset **nonrigid_accurate** seems to give the best results on this recording.
+#     The motion vector seems less noisy globally, but it is not 'perfect' (see at the top of the probe 3200um to 3800um).
+#     Also note that in the first part of the recording before the imposed motion (0-600s) we clearly have a non-rigid motion:
+#     the upper part of the probe (2000-3000um) experience some drifts, but the lower part (0-1000um) is relatively stable.
```
zm711, line 249: I don't like "drifts" since I would say this is an uncountable concept, but it may be too late and the field has settled on this.

zm711 (commented):
Got to run to a meeting, but I did some edits on the first 200 lines. :)

doc/long_tutorials/handle_drift/sg_execution_times.rst
```diff
@@ -0,0 +1,37 @@
+
+:orphan:
```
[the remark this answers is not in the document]
alejoe91: this should be captured by the gitignore, but I think it should be `*sg_execution_times.rst` (instead of `doc/sg_execution_times.rst`)

@JoeZiminski

doc/conf.py
JoeZiminski: I think this is very nice, for now if it works for you I will wait for the upcoming PR tackling the Neuropixels how-to to implement this, just to make checking it's working for multiple cases a bit easier. I'll link to this suggestion in our issue to not forget!

doc/long_tutorials/handle_drift/sg_execution_times.rst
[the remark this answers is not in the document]
JoeZiminski: Great thanks good point, have added here but opened a PR on main #3097

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+For more information on the theory and implementation of these methods,
+see the :ref:`motion_correction` section of the documentation.
+
+---------------------
+What is probe drift?
+---------------------
+
+The inserted probe can move from side-to-side (*'x' direction*),
+up-or-down (*'y' direction*) or forwards-or-backwards (*'z' direction*).
+Movement in the 'x' and 'y' direction is harder to model than vertical
```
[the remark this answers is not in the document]
JoeZiminski: !! good catch!

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+non-rigid drift is instead caused by local movements brain along the
+probe, and can selectively affect subsets of channels..
+
+--------------------------
+The drift correction steps
+--------------------------
+
+The easiest way to run drift correction in SpikeInterface is with the
+high-level :py:func:`~spikeinterface.preprocessing.correct_motion()` function.
+This function takes a preprocessed recording as input and returns a motion-corrected
```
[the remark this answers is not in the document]
JoeZiminski: Good point, canonically it should take a recording that has been at least filtered. But the type of preprocessing is unspecific, and it will work on any recording, so I removed 'preprocessed' here. There is a more detailed discussion on the preprocessing later anyway!

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+| **4.** ``interpolate_motion()`` (perform interpolation on the raw data to account for the estimated drift).
+
+All these sub-steps have many parameters which dictate the
+speed and effectiveness of motion correction. As such, ``correct_motion``
+provides three setting 'presets' which configure the motion correct
+to proceed either as:
+
+* **rigid_fast** - a fast, not particularly accurate correction assuming ridigt drift.
+* **kilosort-like** - Mimics what is done in Kilosort.
+* **nonrigid_accurate** - A decentralised drift correction (DREDGE), introduced by the Paninski group.
```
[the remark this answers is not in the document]
JoeZiminski: :D 🇺🇸

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+from spikeinterface.preprocessing.motion import motion_options_preset
+from spikeinterface.sortingcomponents.motion_interpolation import correct_motion_on_peaks
+from spikeinterface.widgets import plot_peaks_on_probe
+
+# %% read the file
+# Next, we will generate a synthetic, drifting recording. This recording will
+# have 100 separate units with firing rates randomly distributed between
+# 15 and 25 Hz.
+#
+# We will create a zigzag drift pattern on the recording, starting at
```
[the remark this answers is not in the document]
JoeZiminski: This is actually something I wanted to discuss with @samuelgarcia and @cwindolf. At the moment the drift is generated with:

```
_, raw_recording, _ = generate_drifting_recording(
    num_units=num_units,
    duration=duration,
    generate_sorting_kwargs=dict(firing_rates=(15, 25), refractory_period_ms=4.0),
    seed=42,
    generate_displacement_vector_kwargs=dict(motion_list=[
            dict(
                drift_mode="zigzag",
                non_rigid_gradient=None, # 0.1,
                t_start_drift=int(duration/10),
                t_end_drift=None,
                period_s=int(duration/10),
            ),
        ],
    )
)
```
I'm not sure how t o use the `non_rigid_gradient` parameter. Presumably when `None` the drift is rigid? Would be great to get suggestions on the best parameters to use here for a nice example.

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+# 100 seconds and with a peak-to-peak period of 100 seconds (so we will
+# have 9 zigzags through our recording). We also add some non-linearity
```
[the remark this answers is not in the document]
JoeZiminski: Yes thanks this was lazy shorthand for non-rigid drift, I'm not even sure it makes sense! I'll wait for discussion on above point re: rigid drift then update.

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+# with a bandpass filter and a common median reference.
+
+filtered_recording = si.bandpass_filter(raw_recording, freq_min=300.0, freq_max=6000.0)
+preprocessed_recording = si.common_reference(filtered_recording, reference="global", operator="median")
+
+# %%
+#.. warning::
+#    It is better to not whiten the recording before motion estimation, as this
+#    will give a better estimate of the peak locations. Whitening should
+#    be performed after motion correction.
```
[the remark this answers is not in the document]
JoeZiminski: Maybe instead can add a bit more text here on how motion correction fits into the preprocessing pipeline, and mention it should be done after filtering, CAR but before whitening, rather than a warning box? maybe too extreme for something that is probably an edge case.

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+
+    corrected_recording, motion_info = si.correct_motion(
+        preprocessed_recording, preset=preset,  output_motion_info=True, **job_kwargs
+    )
+    results[preset]["motion_info"] = motion_info
+
+# %%
+#.. seealso::
+#   It is often very useful to save ``motion_info`` to
+#   file, so it can be loaded and visualised later. This can be done by setting
```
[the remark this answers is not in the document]
JoeZiminski: 🇺🇸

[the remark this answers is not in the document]
JoeZiminski: This is super useful though, if we don't already should this go in developer docs style guide? This consistency is much better approach to my usual of switching wildly and inconsistently between spelling types across whatever doc I am writing.

examples/long_tutorials/handle_drift/plot_handle_drift.py
JoeZiminski: I had a quick look at it seems to write a lot of output to a folder, including parameters `.json` and mostly `.npy` files. So it is not as simple as the text originally implied. In this case I updated to `all motion outputs to` to avoid going too specific?

examples/long_tutorials/handle_drift/plot_handle_drift.py
JoeZiminski: Oh yes, I think I was going for 'The (binned) time'. Maybe this is too much glossing over the binning as I didn't want to get into the weeds with it. That being said it is quite critical to understanding how the motion algorithms are working. Maybe more information on time and spatial bins in a dropdown?

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
+"""
+===========================================
+Handle probe drift with spikeinterface NEW
```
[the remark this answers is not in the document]
JoeZiminski: Great, cheers for review @zm711!

h-mayorquin:
@JoeZiminski it is in my to-do list to review this. Just a ping

JoeZiminski:
Thanks @h-mayorquin, no rush!

Cheers @chrishalcrow this is a great point, I think an additional section 'when to apply motion correction' would be useful before the summary. For example, I had a case recently (Cambridge Neurotech shank with only 16 channels, which I guess these algorithms were not really designed for) where the motion correction was introducing a lot of artefacts. Also, there was not much motion to begin with as all recording sessions were close together in time and not too long, so it was not really necessary. The output looked like this:

Screenshot 2024-06-28 at 10 44 55

On the top left, the peaks look quite stable. But on the top right after correction, they have been smeared across the probe. Also, there are corresponding jumps in the motion vector where something is obviously going wrong.

A case of real data with motion in is the pictures of the neuropixels data used in the original version of this tutorial (below). So maybe we can include that too to show cases of real data where there is and isn't drift. And this will also show what the motion correction outputs look like when things work / things go wrong. 

Screenshot 2024-06-28 at 10 46 17

Good question about the probes! I'm not sure on this, my guess is that for the most part if you have peaks drifting in the data you probably could use some kind of motion correction. But, these algorithms as currently implemented may fail for probes with lower channel number as above? In the KS4 docs I saw ''For probes with fewer channels (around 64 or less) or with sparser spacing (around 50um or more between contacts), drift estimates are not likely to be accurate, so drift correction should be skipped'. ([here](https://kilosort.readthedocs.io/en/latest/parameters.html#nblocks).

Actually, I just properly looked at their [drift correction section](https://kilosort.readthedocs.io/en/latest/drift.html) which has a very nice explainer, we can link to it here I think.

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,342 @@
...
+# %%
+# ------------------------
+# Summary
+# ------------------------
+#
+# That's it for our tour of motion correction in
+# SpikeInterface. Remember that correcting motion makes some
+# assumptions on your data (e.g. number of channels, noise in the recording)—always
+# plot the motion correction information for your
+# recordings, to make sure the correction is behaviour as expected!
```
zm711, line 342:
```suggestion
# recordings, to make sure the correction is behaving as expected!
```

chrishalcrow:
> I think an additional section 'when to apply motion correction' would be useful before the summary....

Thanks @JoeZiminski . Great. The main point I'd missed was: to see if you have drift, you need to apply the drift correction and take a look at the results. And then give some suggestions for what "bad" results would look like: lots of discrete spikes in the motion vectors and a corrected peak depth plot that makes less sense than the original. Kilosort's page is great! Their discussion on timescales also shows why your motion vector shouldn't be trusted, since those spikes are not on timescales of seconds. I don't think you need to include lots of different datasets, just some advice would be good.

So Kilosort suggests that only high-density high-channel-count probes should be motion corrected. I would mention this right at the start.

examples/long_tutorials/handle_drift/plot_handle_drift.py
JoeZiminski: Yes I wanted to fix to 'motion', but its had to use in some senses e.g. 'upper part of the probe may experience some motion' doesn't quite read properly for me. I guess motion may be too general and makes you think that probe is moving (or maybe the probe is moving because the animal is moving! there is 'motion' in the system) whereas 'drift' is a bit more specific and particularly refers to the phenomenon of the signal drifting across the probe (maybe the probe or may be the brain).

doc/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,367 @@
...
+#    This tutorial can take a long time to run with the default arguments.
+#    If you would like to run this locally, you may want to edit ``num_units``
+#    and ``duration`` to smaller values (e.g. 25 and 100 respectively).
+#
+#    Also note, the below code uses multiprocessing. If you are on Windows, you may
+#    need to place the code within a  ``if __name__ == "__main__":`` block.
+
+
+num_units = 10# 200  # 250 still too many I think!
+duration = 50  # 1000
```
chrishalcrow, line 118: Guessing this was for your local build - should uncomment 200 and 1000.

JoeZiminski: Good spot! have updated

JoeZiminski:
Thanks @chrishalcrow great points have updated, let me know what you think (I'm not super happy with what I've written not sure if its clear, all feedback welcome!)

h-mayorquin (commented):
OK, let me see if I understand. We have three types of documents in the docs:
1) We have things that are static files and code like some of the how-to
2) We have things that are fully built in the gallery because they run in a reasonable time.
3) We have  gallery items that take too long to build so we just built them manually.

We currently use approach 3 for some documents like `get_started` the only thing this approach does is creating a way of calling this automatically through the tag machinery of sphinx. This makes sense to me. I will do a second pass to check on the content.

examples/long_tutorials/handle_drift/plot_handle_drift.py
```diff
@@ -0,0 +1,368 @@
...
+                drift_mode="zigzag",
+                non_rigid_gradient=0.01,
+                t_start_drift=int(duration/10),
+                t_end_drift=None,
+                period_s=int(duration/10),
+            ),
+        ],
+    )
+)
+print(raw_recording)
```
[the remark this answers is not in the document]
h-mayorquin: https://sphinx-gallery.github.io/stable/auto_examples/plot_3_capture_repr.html#capture-repr-examples

doc/conf.py
```diff
@@ -130,9 +130,19 @@
     'within_subsection_order': FileNameSortKey,
     'ignore_pattern': '/generate_',
     'nested_sections': False,
-    'copyfile_regex': r'.*\.rst|.*\.png|.*\.svg'
+    'copyfile_regex': r'.*\.rst|.*\.png|.*\.svg',
```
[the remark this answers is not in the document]
chrishalcrow: Hello, sorry I missed this.

This copies any rst, png or svg files from the source `sphinx-gallery` directory to the output directory. In this case the source directory is `examples/tutorials/` and we would like it to copy `waveform_extractor_to_sorting_analyzer.rst` from there to where the gallery is outputted (`doc/tutorials`). 

This is done so that the tutorial pages (https://spikeinterface.readthedocs.io/en/latest/tutorials/index.html), which is a sphinx-gallery, can include a non-sphinx gallery page (the waveforms page).

The `.svg`s were included to copy over the svg of the extension parent/child figure, but this is not kept in `doc/_images` since it's repeated in a few places.

PR #2884, opened 2024-05-20
Protect zarr properties of type object

alejoe91:
When loading phy/kilosort folders, string array properties are cast to python `Object`. This results in a failure when writing to zarr.

This simple PR protects against Object dtypes and tries to cast the array as str.

### DIFF
--- a/src/spikeinterface/core/zarrextractors.py
+++ b/src/spikeinterface/core/zarrextractors.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import warnings
 from pathlib import Path
 import numpy as np
 import zarr
@@ -315,6 +316,12 @@ def add_properties_and_annotations(
     prop_group = zarr_group.create_group("properties")
     for key in recording_or_sorting.get_property_keys():
         values = recording_or_sorting.get_property(key)
+        if values.dtype.kind == "O":
+            try:
+                values = values.astype("str")
+            except Exception as e:
+                warnings.warn(f"Property {key} not saved because it is a python Object type")
+                continue
         prop_group.create_dataset(name=key, data=values, compressor=None)
 
     # save annotations
--- a/src/spikeinterface/core/zarrextractors.py
+++ b/src/spikeinterface/core/zarrextractors.py
@@ -309,9 +309,7 @@ def get_default_zarr_compressor(clevel: int = 5):
     return Blosc(cname="zstd", clevel=clevel, shuffle=Blosc.BITSHUFFLE)
 
 
-def add_properties_and_annotations(
-    zarr_group: zarr.hierarchy.Group, recording_or_sorting: BaseRecording | BaseSorting
-):
+def add_properties_and_annotations(zarr_group: zarr.hierarchy.Group, recording_or_sorting: BaseRecording | BaseSorting):
     # save properties
     prop_group = zarr_group.create_group("properties")
     for key in recording_or_sorting.get_property_keys():
--- a/src/spikeinterface/core/zarrextractors.py
+++ b/src/spikeinterface/core/zarrextractors.py
@@ -310,18 +310,15 @@ def get_default_zarr_compressor(clevel: int = 5):
 
 
 def add_properties_and_annotations(
-    zarr_group: zarr.hierarchy.Group, recording_or_sorting: Union[BaseRecording, BaseSorting]
+    zarr_group: zarr.hierarchy.Group, recording_or_sorting: BaseRecording | BaseSorting
 ):
     # save properties
     prop_group = zarr_group.create_group("properties")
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/zarrextractors.py
```diff
@@ -315,6 +316,12 @@ def add_properties_and_annotations(
     prop_group = zarr_group.create_group("properties")
     for key in recording_or_sorting.get_property_keys():
         values = recording_or_sorting.get_property(key)
+        if values.dtype.kind == "O":
+            try:
+                values = values.astype("str")
+            except Exception as e:
```
h-mayorquin, line 322: I would encode objects with msgpack.

https://zarr.readthedocs.io/en/stable/tutorial.html#string-arrays

Of the protocols of zarr is the one that is easily available on other langugaes.

h-mayorquin: Or json if you prefer, probably the performance of msgpack does not mater in this case.

alejoe91: I think we should not encode all objects. Say you have a `Sorting` object as a property. I think we should skip saving those to zarr. Ok to use the `msgpack` codec, but I'm not sure why this simple fix is working then, since the error came from the `KSLabel`, which can be `good`, `mua` or `noise` (so it's a variable string). Probably, using `astype(str)` converts it to an `S` type, with is handled differently than `U` dtype. Any thoughts?

h-mayorquin: What was the error you were getting?

I think that zarr in python should handle the old fix width unicode just fine.

alejoe91: can't remember the exact error, but it was about missing an `object_codec`

h-mayorquin: If this is solving ithe problem your original object was neither byte string or unicode:

https://numpy.org/doc/stable/reference/generated/numpy.dtype.kind.html

I would check why KSLabel is not transformed to string when you load it but this code here would solve your problem. Want to merge? I would go and put the specific exception that you are getting in your code though.

alejoe91: No rush! I think that string arrays are automatically loaded as Objects by `pandas`, and KS output is loaded with pandas from the generated csv files. I'll post the error I was getting tomorrow.

h-mayorquin: Dude, this shit works:

```python
import numpy as np



recording1 = generate_recording()
recording2 = generate_recording()

array = np.array([recording1, recording2]) 

array.astype("str")
```

No exception thrown : /

alejoe91: Ouch! Maybe because of the `__repr__` function?

h-mayorquin: Yes, I was just reading the spec. Apparently they use `__str__` by default. So it is difficult to make this fail. Another example:

```python


recording1 = generate_recording()
recording2 = generate_recording()
import random

class RandomAttributes:
    def __init__(self):
        self.attribute1 = random.randint(1, 10)
        self.attribute2 = random.choice(['red', 'blue', 'green'])
        self.attribute3 = random.uniform(0.0, 1.0)

# Create an instance of the class
obj1 = RandomAttributes()
obj2 = RandomAttributes()   

array = np.array([obj1, obj2]) 

new_array =array.astype("str")
new_array
```

 I was hoping we could use casting safe rules but unfortunatley this fails:
```python 
original_string_array = np.array(["this", "that", ])
original_string_array.astype("object").astype("str", casting="safe")
 ```

h-mayorquin: Yeah, I think the quicker error would be to indicate in the respective `pd.load_csv` that the type of that column is string.

alejoe91: Good idea and probably a better fix!

alejoe91: Done, should be better now. Let me check if that solves the issue I was having :)

h-mayorquin (approved):
This works but what I was suggestion is to to use the dtype argument of pd.read_csv here to indicate that that column specifically is string. This will also reduce the memory footprint but I don't think it matters that much in this case.

https://github.com/SpikeInterface/spikeinterface/blob/e0eb995b817a744f9afcb12ef4a69625893d12df/src/spikeinterface/extractors/phykilosortextractors.py#L94

src/spikeinterface/core/zarrextractors.py
h-mayorquin: Great. Approved.

samuelgarcia (approved):

PR #2896, opened 2024-05-22
Important fixes

samuelgarcia:
@alejoe91 
Various but important that prevent demo to be run:
  * zarr in analyzer
  * matplotlib

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -586,7 +586,9 @@ class SortingAnalyzer:
 
         # sparsity
         if "sparsity_mask" in zarr_root:
-            sparsity = ChannelSparsity(np.array(zarr_root["sparsity_mask"]), sorting.unit_ids, rec_attributes["channel_ids"])
+            sparsity = ChannelSparsity(
+                np.array(zarr_root["sparsity_mask"]), sorting.unit_ids, rec_attributes["channel_ids"]
+            )
         else:
             sparsity = None
 
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -585,9 +585,8 @@ class SortingAnalyzer:
             rec_attributes["probegroup"] = None
 
         # sparsity
-        if "sparsity_mask" in zarr_root.attrs:
-            # sparsity = zarr_root.attrs["sparsity"]
-            sparsity = ChannelSparsity(zarr_root["sparsity_mask"], cls.unit_ids, rec_attributes["channel_ids"])
+        if "sparsity_mask" in zarr_root:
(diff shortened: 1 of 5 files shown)


src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -585,9 +585,10 @@ def load_from_zarr(cls, folder, recording=None):
             rec_attributes["probegroup"] = None
 
         # sparsity
-        if "sparsity_mask" in zarr_root.attrs:
-            # sparsity = zarr_root.attrs["sparsity"]
-            sparsity = ChannelSparsity(zarr_root["sparsity_mask"], cls.unit_ids, rec_attributes["channel_ids"])
+        if "sparsity_mask" in zarr_root:
+            sparsity = ChannelSparsity(
+                np.array(zarr_root["sparsity_mask"]), sorting.unit_ids, rec_attributes["channel_ids"]
```
alejoe91, line 590:
```suggestion
                zarr_root["sparsity_mask"][:], sorting.unit_ids, rec_attributes["channel_ids"]
```

src/spikeinterface/widgets/utils_matplotlib.py
```diff
@@ -15,7 +15,7 @@ def make_mpl_figure(figure=None, ax=None, axes=None, ncols=None, num_axes=None,
             if "ipympl" not in matplotlib.get_backend():
                 ax = figure.add_subplot(111)
             else:
-                ax = figure.add_subplot(111, layout="constrained")
+                ax = figure.add_subplot(111)
```
samuelgarcia, line 18: @alejoe91 this do not exists in the signature.
It exists only in figure() or subplots()

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1615,7 +1612,8 @@ def load_data(self):
                 elif "object" in ext_data_.attrs:
                     ext_data = ext_data_[0]
                 else:
-                    ext_data = ext_data_
+                    # this load in memmory
+                    ext_data = np.array(ext_data_)
```
samuelgarcia, line 1616: @alejoe91 we need memory load otherwise the slicing will be buggy

src/spikeinterface/core/sortinganalyzer.py
samuelgarcia: This is less explicit no ?

src/spikeinterface/widgets/utils_matplotlib.py
alejoe91: Good catch! The documentation is a bit ambiguous

src/spikeinterface/core/sortinganalyzer.py
alejoe91: Ok, let's keep it as is at least for now

src/spikeinterface/core/sortinganalyzer.py
alejoe91: Ok fine like this

alejoe91 (approved):

PR #2898, opened 2024-05-23
remove verbose from job_kwargs

samuelgarcia:
Lets see what new bug this could trigger!

@h-mayorquin @zm711 : this has been discuss at many places.
Lets make a list of link here.

### DIFF
--- a/src/spikeinterface/core/job_tools.py
+++ b/src/spikeinterface/core/job_tools.py
@@ -47,7 +47,6 @@ job_keys = (
     "chunk_duration",
     "progress_bar",
     "mp_context",
-    "verbose",
     "max_threads_per_process",
 )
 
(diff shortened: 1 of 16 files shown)


alejoe91:
I agree with this. `verbose` shoudl have nothing to do with `job_kwargs` and we'll have a common and global way to set verbosity once we move move to `logging`

zm711:
Definitely agree too. The verbosity of a library should be separate from the parallelization. 

I do love the tqdm (probably because my personal verbosity is always set to True :P) so looking forward to the work on the logger.

alejoe91 (approved):

h-mayorquin:
Damn guys you went too quick with this. This does not solve the problem.

PR #2901, opened 2024-05-23
Fix remaining verbose clashing

### LINKED ISSUE #2847: Verbose clashing in `recording.save_to_folder` and `job_kwargs`
As in the title, 

https://github.com/SpikeInterface/spikeinterface/blob/3864afde8aad572041b4218ebb1d50b2e7a58d6b/src/spikeinterface/core/base.py#L849-L850

There is a verbose in `save_to_folder` but it also clashes with the `save_kwargs` which should contain `job_kwargs`. 

Suggestion: drop the verbose in `save_to_folder` or rename it to `print_used_folder` if someone wants that functionality.

I swear I read about a similar issue but could not find it.
[Edit] Yes, it was a PR not an issue:
https://github.com/SpikeInterface/spikeinterface/pull/2602

Also, minimal example:

```python

recording = generate_recording(num_channels=3, durations=[10])

from pathlib import Path

folder_path = Path("./test_recording")

job_kwargs={'progress_bar': True, "verbose":True, "n_jobs":2}
recording.save_to_folder(folder=folder_path, verbose=False, overwrite=True, **job_kwargs)   
```

h-mayorquin:
Should close #2847. Following on https://github.com/SpikeInterface/spikeinterface/pull/2898/files. Also removed some comments from https://github.com/SpikeInterface/spikeinterface/pull/2728.

I still think this approach of catching verbose at every point of the call stack is very labor intensive. I mentioned to Sam that we could have more specific verbosity terms:

For the ChunkRecordingExectuor: `display_chunk_info` or `verbose_chunk_executor` 
For the save_to_folder `display_saved_folder`

But he pointed out -correctly I think- that usually verbose is managed in linux like a global option which I think it makes sense. 

Maybe the real work around will be the genera logger as described in https://github.com/SpikeInterface/spikeinterface/pull/2602

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -491,7 +491,7 @@ class BaseRecording(BaseRecordingSnippets):
         rs = self._recording_segments[segment_index]
         return rs.time_to_sample_index(time_s)
 
-    def _save(self, format="binary", **save_kwargs):
+    def _save(self, format="binary", verbose: bool = False, **save_kwargs):
         # handle t_starts
         t_starts = []
         has_time_vectors = []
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/baserecording.py
```diff
@@ -540,14 +540,18 @@ def _save(self, format="binary", **save_kwargs):
 
                 cached = SharedMemoryRecording.from_recording(self, **job_kwargs)
             else:
+                from spikeinterface.core import NumpyRecording
```
h-mayorquin, line 543: This import was missing, I wonder if it means that we are not testing this.

samuelgarcia (approved):

samuelgarcia:
Thanks a lot.
This is OK for me.
Alessio do you want to have a look ?

alejoe91:
Ok for me

PR #2902, opened 2024-05-23
Making `chunk_size=None` behavior for n_jobs=1 explicit

h-mayorquin:
This is necessary for #2892.

As you can see from my previous comment one year ago in the diff I was baffled by this edge case. Today while working on #2892 I finally understood (I hope!) what @samuelgarcia intended. 

I am making this behavior explicit and adding a test.


samuelgarcia:
This is OK for me.

alejoe91 (approved):

alejoe91 (approved):

PR #2905, opened 2024-05-24
Fix some bugs in matched filtering

yger:
This PR fixes two issues with the matched filtering:
- asymetric prototypes were not properly handled
- too many spikes were detected because of issues in the numba kernel


yger:
@samuelgarcia  This solves all the problem with the matched filtering. Figures have been redone in the presentation, much better. I added the ms_before argument to the function, which is now working with asymzetric waveforms (tested)

samuelgarcia (approved):

PR #2907, opened 2024-05-24
Fix more verbosity propagation

samuelgarcia:
@h-mayorquin

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -925,7 +925,7 @@ class SortingAnalyzer:
         Returns
         -------
         result_extension: AnalyzerExtension
-            Return the extension insdef computance.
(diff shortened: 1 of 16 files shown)


src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -961,13 +961,16 @@ def compute_one_extension(self, extension_name, save=True, **kwargs):
 
         extension_instance = extension_class(self)
         extension_instance.set_params(save=save, **params)
-        extension_instance.run(save=save, **job_kwargs)
+        if extension_class.need_job_kwargs:
+            extension_instance.run(save=save, verbose=verbose, **job_kwargs)
+        else:
+            extension_instance.run(save=save)
```
zm711, line 967: Even if it is not parallel an extension could have a verbose setting no? Should we prep for that future?

PR #2908, opened 2024-05-24
hdbscan is awfully slow on CPU side. I added hdbscan-gpu option to us…

rat-h:
A tiny addition to simple sorter. hdbscan is awfully slow on CPU side. This commit adds hdbscan-gpu option to use hdbscan from NVidea cuml.

### DIFF
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -49,9 +49,8 @@ class ComputeRandomSpikes(AnalyzerExtension):
     use_nodepipeline = False
     need_job_kwargs = False
 
-    def _run(
-        self,
-    ):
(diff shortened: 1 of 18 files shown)


zm711:
@rat-h

What is your use case for this? I think the idea behind this sorter was originally just for doing demos, teaching, and for quick checks of data.

h-mayorquin:
> @h-mayorquin Something of 1000 - 1400 time faster. I think, the simple sorter has a problem. hdbscan uses only one CPU doesn't matter what. So acceleration is from overnight + a little bit to 10-20 seconds.

This is great!

zm711:
@rat-h and it works super well for your purposes? (After the speed-up?) That would be super cool.

samuelgarcia (approved):

samuelgarcia:
merci.

PR #2914, opened 2024-05-29
Matched filtering with both peak signs simultaneously

yger:
This PR should allow matched filtering to work with both positive and negative peak sign detection. If both is selected, then templates are duplicated with a sign inversion, and added to the dictionary of templates.
I've also added a drastic speed up of the function, by simplification of the convolution


yger:
This PR should not be merged as it is, because even if this is working without error, the behavior is strange. I suspect that, in the numba kernel, there is a corner case as mentioned by @samuelgarcia  that wrote it with inequalities, such that if templates are at the same spot, we are keeping all peaks. Currently, if only neg or pos is selected for the peak type, this is working fine. But with both, we have way too many peaks detected....

yger:
@samuelgarcia I've found the bug and I've patched it. This is maybe too late for the release, but since it was tagged at some point. Now this branch is working as expected!

samuelgarcia (approved):

samuelgarcia:
Bien joué.

PR #2915, opened 2024-05-29
Internal motion API, aka `Motion` object

cwindolf:
Copy of @samuelgarcia 's https://github.com/SpikeInterface/spikeinterface/pull/2805 which I can write to

From discussion today:
 - [x] Check TODOs in code
 - [x] Handle the sample chunking in the interpolation:
   - (a) interpolation chunks need not be same as motion estimation chunks
   - (b) deal with `_get_closest_ind`
   - [x] add test for different chunk sizes in interpolation
 - [x] Propagate the object around all over the place
 - [x] Time handling agreement: everything goes through the middlemen (`sample_index_to_time`, `time_to_sample_index`), but we assert `not has_time_vector()` for now, since it would be expensive to deal with all of the time vectors in parallel stuff.

I propose that eventually the time_vector is sharedmem or memmapped to make things work in parallel. Alternatively, `sample_index_to_time()` could use `np.fromfile()` with an appropriate offset into the time vector file to read the relevant piece (which is very fast). But then I'm not sure how `time_to_sample_index()` would work. TBD!


src/spikeinterface/sortingcomponents/motion_interpolation.py
```diff
@@ -106,53 +109,45 @@ def interpolate_motion_on_traces(
...
-        # # channel_locations_moved[:, direction] -= channel_motions
-
         if channel_inds is not None:
             channel_locations_moved = channel_locations_moved[channel_inds]
 
         drift_kernel = get_spatial_interpolation_kernel(
             channel_locations,
             channel_locations_moved,
-            dtype="float32",
+            dtype=traces.dtype,
```
samuelgarcia, line 150: dangerous no if the dtype is int16 ? The kriging is mostly between 0 and 1.

cwindolf: Good point! revised the dtype handling now -- I was trying to support other floating dtypes than just float32, and hopefully now that's handled in a better way. I'm on a plane and the wifi blocks my ssh remote but will push tomorrow !

cwindolf:
Btw, all of the `cdist()` calls in `get_spatial_interpolation_kernel()` are probably expensive, should we revise these by caching a k-d tree of the original channel locations?

For instance, the `"idw"` could use [`KDTree.query()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.KDTree.query.html#scipy.spatial.KDTree.query])with `n_neighbors=num_closest` to efficiently get the distances to the num_closest neighbors. We could even set `distance_upper_bound` to a precomputed max distance to the `num_closest`-nearest neighbor in the original geometry.  Interpolation mode `"nearest"` can just get 1 neighbor. For kriging, alas no such optimization can be made.

cwindolf:
@samuelgarcia made most of the changes I'm supposed to, I think... how does it look to you? Hope Edi's been nice!

cwindolf:
Seems like something is going on with the benchmark tests. @samuelgarcia I think you mentioned that you preferred to handle the benchmark parts, but I'm happy to take a look if helpful

alejoe91:
@cwindolf @samuelgarcia I fixed a couple of failing tests. Can we close #2805 ?

cwindolf:
I think so, yeah? This PR is on top of that branch

src/spikeinterface/sortingcomponents/motion_interpolation.py
```diff
@@ -98,6 +101,11 @@ def interpolate_motion_on_traces(
     # assert HAVE_NUMBA
     assert times.shape[0] == traces.shape[0]
 
+    if dtype is None:
+        dtype = traces.dtype
+        if dtype.kind != "f":
+            raise ValueError(f"Can't interpolate traces of dtype {traces.dtype}.")
```
samuelgarcia: We should allow interpolate with dtype int but with going back to int after float computation no ?

cwindolf: I'm kind of confused about ints... if your dtype is int16, doesn't that usually mean you would really want to scale your recording and convert to float before interpolating? is there any use case for interpolating int data?

I'm not thrilled by the idea of a user with int16 traces who naively inputs them into the interpolation and then the data comes out looking horrible -- I think it's at least worth raising a warning in that case?

cwindolf: A user with `rec.dtype == np.int16` traces could also easily `interpolate_motion(rec.astype(np.float32))` if they wanted to keep their data values the same, but I expect most people would probably be in the situation where `interpolate_motion(spre.scale(rec, gain=200.0, dtype=np.float32))` makes more sense?

cwindolf: Wait... maybe those things are the same... haha. sorry. I guess I'm thinking out loud a bit here!

src/spikeinterface/sortingcomponents/motion_utils.py
```diff
@@ -0,0 +1,238 @@
...
+#  * propagate to motion_interpolation.py: DONE
+#  * propagate to preprocessing/correct_motion(): DONE
+#  * generate drifting signals for test estimate_motion and interpolate_motion: SIMPLE ONE DONE?
+#  * uncomment assert in test_estimate_motion (aka debug torch vs numpy diff): DONE
+#  * delegate times to recording object in
+#       * estimate motion: DONE
+#       * correct_motion_on_peaks(): DONE
+#       * interpolate_motion_on_traces(): DONE
+# propagate to benchmark estimate motion
+# update plot_motion() dans widget
```
alejoe91, line 24: only this one is missing right?

cwindolf:
Thanks -- if @samuelgarcia is OK with the int/float thing above, then sounds good!

PR #2917, opened 2024-05-29
Regularize folder argument for user in run_sorter options

zm711:
Working on switching all run_sorter flavors to folder rather than working_folder/output_folder/ folder.


samuelgarcia (approved):

h-mayorquin:
One possible improvement would be to specify that is a path `folder_path` like we do in the input side in extractors.

Issue #2479, opened 2024-02-12
Integration of decoder for automated noise, mua, and sua curation


anoushkajain:
Hello Spikeinterface team,

We’ve been working on extracting single-neuron activity from Neuropixels data, which often requires extensive manual evaluation of spike clusters. 

To streamline this process, we have developed a machine-learning pipeline that employs quality metrics and human labels to classify units into Single-Unit Activity (SUA), Multi-Unit Activity (MUA), and Noise, to reduce curation time and improve the reproducibility of the results.

I have attached a sorting view that shows the output of the decoder: https://figurl.org/f?v=gs://figurl/sortingview-11&d=sha1://f709f74add515e23062e3952076602e40d3e86ce

 As previously discussed with Alessio, we want to integrate this approach into the curation module of SpikeInterface. I am initiating this issue to start our discussion on the next steps towards this integration.

anoushkajain:
@Antho2422 
Hi, we are currently working on putting it in a preprint and we will release it soon.

alejoe91:
A possible solution could be to have in the `spikeinterface.curation` a `ModelBasedCuration` abstract class that would look as follows:

```
class ModelBasedCuration:
    requirements = None

    def __init__(required_metrics,  model_path):
        self.model_path = model_path
        self.required_metrics = required_metrics
        self.instantiated_models = dict()

    def instantiate_model(self):
        raise NotImplementedError

    def apply(self, sorting_analyzer, return_probabilities=True):
        raise NotImplementedError

    def compute_missing_metrics(self):
        raise NotImplementedError

    @staticmethod
    def check_installed():
        if self.requirements is not None:
            for req in requirements: # check installation
        

    # optional
    @staticmethod
    def train(...):
        raise NotImplementedError
```

Then the pretrained model could just inherit from this:
```
class PreTrainedSklearnCuration(ModelBasedCuration):

    requirements = ["sklearn", "xgboost", ...]

    def __init__(required_metrics,  model_path):
        ModelBasedCuration.__init__(required_metrics,  model_path)

    def load_model(self):
        self.instantiated_models["noise-neuron"] = pickle.load(model_path / "noise-neuron.pkl")
        self.instantiated_models["sua-mua"] = pickle.load(model_path / "sua-mua.pkl")

    def apply(self, sorting_analyzer, return_probabilities=True):
        compute_missing_metrics(sorting_analyzer)
        metrics = ...
[...]

PR #2918, opened 2024-05-29
Automatic curation with metrics 

anoushkajain:
Solves #2479

This PR introduces model-based metric curation in the curation module.

The core functionality is implemented by the `auto_label_units`, which used scikit-learn pre-trained model in `skops` format to predict labels based on metrics (could be quality, template, or both).

The PR also adds the possibility to train your own model and to automatically access and apply pre-trained models available on hugging face.


src/spikeinterface/curation/tests/test_model_based_curation.py
```diff
@@ -0,0 +1,130 @@
+import pytest
+from pathlib import Path
+import numpy as np
+import joblib
```
alejoe91, line 4: joblib is not a dependency. Can't you load with `pickle` instead?

jakeswann1:
Scikit-learn is not a dependency either - how do we want to handle this? It would be needed for both the prediction and training steps, if we implement that too

alejoe91:
You should move the imports into the functions that use it ;)

src/spikeinterface/curation/train_manual_curation.py
```diff
@@ -0,0 +1,372 @@
...
+from lightgbm import LGBMClassifier
+from catboost import CatBoostClassifier
+from xgboost import XGBClassifier
+from sklearn.pipeline import Pipeline
+
+# Configure logging
+logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+seed = 42
```
jakeswann1, line 23: Do we want to enforce static seed setting? Might be better to force the user to pass a seed

src/spikeinterface/curation/train_manual_curation.py
```diff
@@ -0,0 +1,375 @@
...
+        GradientBoostingClassifier,
+        SVC,
+        LogisticRegression,
+        XGBClassifier,
+        CatBoostClassifier,
+        LGBMClassifier,
+        MLPClassifier,
+    ]
+
+    metrics_list = [
```
jakeswann1, line 330: This might be better passed to the function by the user as an optional argument?
```suggestion
    def train_model(metrics_path, output_folder, objective_label, metrics_list=None):
        metrics_list = metrics_list if metrics_list is not None else ["num_spikes", ...
```

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,241 @@
+"""
+Model Based Curation Tutorial
```
alejoe91, line 2:
```suggestion
Model-based curation tutorial
```

src/spikeinterface/curation/train_manual_curation.py
```diff
@@ -0,0 +1,659 @@
...
+    labels : list of list | None, default: None
+        List of curated labels for each unit; must be in the same order as the metrics data.
+    metrics_path : str or None, default: None
+        The path to the CSV file containing the metrics data if using 'csv' mode.
+    output_folder : str | None, default: None
+        The folder where outputs such as models and evaluation metrics will be saved.
+    metric_names : list of str | None, default: None
+        A list of metrics to use for training. If None, default metrics will be used.
+    imputation_strategies : list of str | None, default: None
+        A list of imputation strategies to apply. If None, default strategies will be used.
```
alejoe91, line 581: What is the default imputation strategy? What are the options?

src/spikeinterface/curation/train_manual_curation.py
```diff
@@ -0,0 +1,659 @@
...
+    scaling_techniques : list of str | None, default: None
+        A list of scaling techniques to apply. If None, default techniques will be used.
```
alejoe91, line 583: ditto

src/spikeinterface/curation/train_manual_curation.py
```diff
@@ -0,0 +1,659 @@
...
+
+class CurationModelTrainer:
+    """
+    Used to train and evaluate machine learning models for spike sorting curation.
+
+    Parameters
+    ----------
+    labels : list of lists, default: None
+        List of curated labels for each unit; must be in the same order as the metrics data.
+    output_folder : str, default: None
```
alejoe91, line 68:
```suggestion
    folder : str, default: None
```
We recently unified the API to use a straight `folder` instead of `output_folder`. Can you modify it in this new curation functions too?

doc/how_to/auto_curation_prediction.rst
```diff
@@ -0,0 +1,40 @@
...
+        model_folder = "my_folder_with_a_model_in_it",
+    )
+
+The returned labels are a dictionary of model's predictions and it's confidence. These
+are also saved as a property of your ``sorting_analyzer`` and can be accessed like so:
+
+.. code::
+
+    labels = sorting_analyzer.get_property("classifier_label")
+    probabilities = sorting_analyzer.get_property("classifier_probability")
```
alejoe91, line 40:
```suggestion
    labels = sorting_analyzer.sorting.get_property("classifier_label")
    probabilities = sorting_analyzer.sorting.get_property("classifier_probability")
```

examples/tutorials/curation/plot_2_train_a_model.py
```diff
@@ -0,0 +1,121 @@
+"""
+Training a model for automated curation
+=============================
+
+If the pretrained models do not give satisfactory performance on your data, it is easy to train your own classifier using SpikeInterface.
+"""
+import warnings
+warnings.filterwarnings("ignore")
+from pathlib import Path
+import spikeinterface.full as si
```
alejoe91, line 10: I would import by module here, since it wil clearly show where the different functions are coming from

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,241 @@
...
+curating spike sorted output. We'll download a toy model and use it to label our sorted data by using
+Spikeinterface. We start by importing some packages
+"""
+
+import warnings
+warnings.filterwarnings("ignore")
+import numpy as np
+import pandas as pd
+import matplotlib.pyplot as plt
+import spikeinterface.full as si
```
alejoe91, line 15: I would import by module here, since it wil clearly show where the different functions are coming from

alejoe91:
@chrishalcrow just noticed this in the curation model based tutorial:
image
Possibly an ordering issue? maybe using set would be better there

examples/tutorials/curation/plot_3_upload_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+#
+# SpikeInterface doesn't require you to keep this folder structure, we just advise it as
+# best practice.
+#
+# If you've used SpikeInterface to train your model, you have already created such a folder,
+# containing everything except the ``metadata.json`` file. In this file, we suggest saving
+# any information which shows when a model is applicable (and when it is *not*). Taking
+# a model trained on mouse data and applying it to a primate is likely a bad idea (or a
+# great research paper!). And a model trained in tetrode data will have limited application
+# on a silconehigh density probe. Hence we suggest the following dictionary as a minimal
```
alejoe91, line 32:
```suggestion
# on a silcone high-density probe. Hence we suggest the following dictionary as a minimal
```

examples/tutorials/curation/plot_3_upload_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+# amount of information needed. Note that we format the metadata so that the information
+# in common with the NWB data format is consistent with it,
+#
+# .. code-block::
+#
+#     model_metadata = {
+#         "subject_species": "Mus musculus",
+#         "brain_area": "CA1",
+#         "probe":
```
alejoe91, line 41: Maybe this should be `probes` with a list? I imagine one could build a model using data from multiple probes, e.g. NP1/NP2

examples/tutorials/curation/plot_3_upload_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+#
+#     This can be used to automatically label a sorting in spikeinterface. Provided you have a `sorting_analyzer`, it is used as follows
+#
+#     ` ` ` python (NOTE: you should remove the spaces between each backtick. This is just formatting for the notebook you are reading)
+#
+#         from spikeinterface.curation import auto_label_units
+#         labels = auto_label_units(
+#             sorting_analyzer = sorting_analyzer,
+#             repo_id = "SpikeInterface/toy_tetrode_model",
+#             trusted = ['numpy.dtype']
```
alejoe91, line 105:
```suggestion
#             trust_model=True
```
? I think it's easier for users

doc/how_to/auto_curation_training.rst
```diff
@@ -0,0 +1,58 @@
+How to train a model to predict curation labels
+===============================================
+
+A full tutorial for model-based curation can be found `here <https://spikeinterface.readthedocs.io/en/latest/tutorials/curation/plot_2_train_a_model.html>`_.
+
+Here, we assume that you have:
+
+* Two SortingAnalyzers called ``analyzer_1`` and
+  ``analyzer_2``, and have calculated some template and quality metrics for both
+* Manutally curated labels for the units in each analyzer, in lists called
```
alejoe91, line 10:
```suggestion
* Manually curated labels for the units in each analyzer, in lists called
```

doc/how_to/auto_curation_training.rst
```diff
@@ -0,0 +1,58 @@
+How to train a model to predict curation labels
```
alejoe91, line 1: What do you think of merging the "How to train/How to use" into a "How to train and use ..."

alejoe91 (changes requested):
Some docs comments

examples/tutorials/curation/plot_3_upload_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+#
+# .. code-block::
+#
+#     model_metadata = {
+#         "subject_species": "Mus musculus",
+#         "brain_area": "CA1",
+#         "probe":
+#             {
+#              "manufacturer": "Neuropixels",
+#              "name": "Neuropixles 2.0"
```
alejoe91, line 44:
```suggestion
#              "manufacturer": "IMEc",
#              "name": "Neuropixels 2.0"
```

examples/tutorials/curation/plot_3_upload_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+# a model trained on mouse data and applying it to a primate is likely a bad idea (or a
+# great research paper!). And a model trained in tetrode data will have limited application
+# on a silconehigh density probe. Hence we suggest the following dictionary as a minimal
+# amount of information needed. Note that we format the metadata so that the information
+# in common with the NWB data format is consistent with it,
+#
+# .. code-block::
+#
+#     model_metadata = {
+#         "subject_species": "Mus musculus",
```
alejoe91, line 39:
```suggestion
#         "subject_species": ["Mus musculus"],
```

src/spikeinterface/curation/model_based_curation.py
```diff
@@ -0,0 +1,395 @@
...
+    trust_model : bool, default: False
+        Whether to trust the model. If True, the `trusted` parameter that is passed to `skops.load` to load the model will be
+        automatically inferred. If False, the `trusted` parameter must be provided to indicate the trusted objects.
+    trusted : list of str, default: None
+        Passed to skops.load. The object will be loaded only if there are only trusted objects and objects of types listed in trusted in the dumped file.
+
+    Returns
+    -------
+    classified_units : dict
+        A dictionary containing the classified units, where the keys are the unit IDs and the values are a tuple of labels and confidence.
```
alejoe91, line 250: return a `pandas.Dataframe`

alejoe91:
TODO items:

### API

- [ ] (Chris) Add consistency check for metric params:
    - training: with multiple analyzer, if params a metric from different analyzers are not the same, warn and skip metric from params
    - predicting: add `enforce_metric_params=False` in `auto_label_units`. If params are different for a metric, warn the user but do not raise. In case `enforce_metric_params==True`, throw an error

- [ ] (Chris) `auto_label_units` should return a `pandas.Dataframe`

### Models

- [ ] (Anoushka) Update model on HF:
    - [ ] `cascaded_classifier.skops` -> `noise_sua_mua_model.skops`
    - [ ] `meta_data.json` -> `metadata.json`
    - [ ] add `model_info.json`

### Docs

- [ ] (Chris) Favore `trust_model=True` instead of `trusted=...` in tutorials and how to
- [ ] (Chris) Update  `A more realistic example: Neuropixels data` with the ``noise_sua_mua_model.skops` and run on the existing `sorting_analyzer`
- [ ] (Chris) model_info.json: all entries should be lists

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,248 @@
...
+recording, sorting = si.generate_ground_truth_recording(num_channels=4, seed=4, num_units=10)
+sorting_analyzer = si.create_sorting_analyzer(sorting=sorting, recording=recording)
+sorting_analyzer.compute(['noise_levels','random_spikes','waveforms','templates','spike_locations','spike_amplitudes','correlograms','principal_components','quality_metrics','template_metrics'])
+
+##############################################################################
+# This sorting_analyzer now contains the required quality metrics and template metrics.
+# We can check that this is true by accessing the extension data.
+
+all_metric_names = list(sorting_analyzer.get_extension('quality_metrics').get_data().keys()) + list(sorting_analyzer.get_extension('template_metrics').get_data().keys())
+print(np.all(all_metric_names == model.feature_names_in_))
```
alejoe91, line 66: use `set` here

src/spikeinterface/curation/model_based_curation.py
```diff
@@ -0,0 +1,424 @@
...
+
+        quality_metrics_extension = self.sorting_analyzer.get_extension("quality_metrics")
+        template_metrics_extension = self.sorting_analyzer.get_extension("template_metrics")
+
+        if quality_metrics_extension is not None:
+
+            model_quality_metrics_params = model_info["metric_params"]["quality_metric_params"]["qm_params"]
+            quality_metrics_params = quality_metrics_extension.params["qm_params"]
+
+            if model_quality_metrics_params == []:
```
alejoe91, line 175: shouldn't this be a dict?

src/spikeinterface/curation/model_based_curation.py
```diff
@@ -0,0 +1,424 @@
...
+                    raise Exception(warning_message)
+                else:
+                    warnings.warn(warning_message)
+
+        if template_metrics_extension is not None:
+
+            model_template_metrics_params = model_info["metric_params"]["template_metric_params"]["metrics_kwargs"]
+            template_metrics_params = template_metrics_extension.params["metrics_kwargs"]
+
+            if template_metrics_params == []:
```
alejoe91, line 195: ditto

src/spikeinterface/curation/model_based_curation.py
```diff
@@ -169,20 +169,26 @@ def _check_params_for_classification(self, enforce_params=False, model_info=None
...
-                else:
-                    warnings.warn(warning_message)
-
-            # need to make sure both dicts are in json format, so that lists are equal
-            if json.dumps(quality_metrics_params) != json.dumps(model_quality_metrics_params):
-                warning_message = "Quality metrics params do not match those used to train model. Check these in the 'model_info.json' file."
-                if enforce_params is True:
+            inconsistent_metrics = []
+            for metric in model_quality_metrics_params["metric_names"]:
+                if metric in model_quality_metrics_params["qm_params"]:
```
alejoe91, line 177: Shouldn't this be a `not in`?

RobertoDF:
Hi, I was wondering if the classification here is different compared to the one implemented at the [Allen Institute](https://github.com/AllenNeuralDynamics/aind-ephys-unit-classifier). I see that @alejoe91 is involved in both. I am interested especially regarding Neuropixel data (both 1 & 2).
Thanks!

alejoe91:
> Hi, I was wondering if the classification here is different compared to the one implemented at the [Allen Institute](https://github.com/AllenNeuralDynamics/aind-ephys-unit-classifier). I see that @alejoe91 is involved in both. I am interested especially regarding Neuropixel data (both 1 & 2). Thanks!

Hi @RobertoDF 

They are essentially the same models, but this PR adds some backbone machinery and should minimize the amount of custom steps to apply the models :) Once this is merged and release, I plan to update the Allen institute classifier to use this SpikeInterface wrapper.

We apply the same model to both NP1 and NP2 data and it gives quite good results on both, especially in identifying noisy units.

RobertoDF:
Thanks!

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,248 @@
...
+plt.show()
+
+
+##############################################################################
+# Here, there are several false positives (if we consider the human labels to be "the truth").
+#
+# Next, we can also see how the model's confidence relates to the probability that the model
+# label matches the human label
+#
+# This could be used to set a threshold above which you accept the model's classification,
```
JoeZiminski, line 133: This is really cool, even though how it is written is quite clear, for me I had a hard time visualizing it initially.  Maybe a quick example, and motivation of the below plot? It took me a while to figure out what the plot was showing and why, a short sentence at the start of this section could make this clearer.

e.g. For example, we might accept any unit as 'good' that the model predicts as 'good' with accuracy over 80%. Otherwise, if an accuracy is lower than this, we will manually check the unit to make a decision.

Below, we will create a plot that shows ....

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,248 @@
...
+sw.plot_unit_templates(sorting_analyzer, unit_ids=[7,9])
+
+##############################################################################
+# Nice - we see that unit 9 does look a lot more spikey than unit 7. You might think that unit
+# 7 is a real unit. If so, this model isn't good for you.
+#
+# Assess the model performance
+# ----------------------------
+#
+# To assess the performance of the model relative to human labels, we can load or generate some
```
JoeZiminski, line 96: maybe 'labels assigned by a human curator'? (just the first time the concept is mentioned)

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,248 @@
...
+plt.axhline(y=1/len(np.unique(predictions)), color='black', linestyle='--', label='Chance')
+plt.xlabel('Confidence'); #plt.xlim(0.5, 1)
+plt.ylabel('Proportion Agreement with Human Label'); plt.ylim(0, 1)
+plt.title('Agreement vs Confidence (Moving Average)')
+plt.legend(); plt.grid(True); plt.show()
+
+##############################################################################
+# In this case, you might decide to only trust labels which had confidence over above 0.86.
+#
+# A more realistic example: Neuropixels data
```
JoeZiminski, line 177: I found this section quote confusing, I would suggest to rename it like 'Using other models' or something, and move it about the assessment vs. human labels section. Then the tutorial would read like:

1) introduction
2) how to run on tetrode model
3) explain what models are available, example of using a different type of model
4) comparing outputs vs human curated data

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,248 @@
...
+# We can choose which to load in the ``load_model`` function as follows:
+#
+# .. code-block::
+#    import spikeinterface.curation as sc
+#    model, model_info = sc.load_model(
+#        repo_id = "AnoushkaJain3/curation_machine_learning_models",
+#        model_name= 'noise_neuron_model.skops',
+#    )
+#
+# If you run this locally you will receive an error:
```
JoeZiminski, line 197: This section is a little long / technical. I would suggest highlighting the error and fix, and including details in a [dropdown](https://sphinx-design.readthedocs.io/en/latest/dropdowns.html).

e.g.
```
# If you run this locally you will receive an error:
#     UntrustedTypesFoundException: Untrusted types found in the file: ['sklearn.metrics._classification.balanced_accuracy_score', 'sklearn.metrics._scorer._Scorer', 'sklearn.model_selection._search_successive_halving.HalvingGridSearchCV', 'sklearn.model_selection._split.StratifiedKFold']
#
# This error is raised because it is possible to inject malicious code into the model format. If you unequivocally trust the model (e.g. if you have created it), you can bypass this security
# step by passing ``trust_model = True`` to the ``load_model`` function.
#
# In general, you should be cautious when downloading ``.skops`` files and ``.pkl`` files from repos,
# especially from unknown sources. For more information, see the below dropdown.
```

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,248 @@
...
+# step by passing ``trust_model = True`` to the ``load_model`` function.
+#
+# In general, you should be cautious when downloading ``.skops`` files and ``.pkl`` files from repos,
+# especially from unknown sources.
+#
+#
+# Directly applying a sklearn Pipeline
+# ------------------------------------
+#
+# Instead of using ``HuggingFace`` and ``skops``, you might have another way of receiving a sklearn
```
JoeZiminski, line 236: I was a bit confused by this, is 'might have another way' being a somone directly sends to you, or you download from github?

In th below code example, `model` is not defined. Can you remind here how to load `model` from file?

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,248 @@
...
+# pipeline, and want to apply it to your sorted data.
+
+from spikeinterface.curation.model_based_curation import ModelBasedClassification
+
+model_based_classification = ModelBasedClassification(sorting_analyzer, model)
```
JoeZiminski, line 241: What is this step doing? I don't quite understand the class name vs. what we were doing before, isn't everything model based classification?

JoeZiminski: is this another way of doing `auto_label_units()`? If so, is it feasible / good to combine into one API providing a model (either a model object or string to huggingface model?

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,248 @@
...
+labels = model_based_classification.predict_labels()
+labels
+
+##############################################################################
+# Using this, you lose the advantages of the model metadata: the quality metric parameters
+# are not checked and the labels are not converted their original human readable names (like
```
JoeZiminski, line 247: Could go into a bit more detail here, what does 'quality metric parameters are not checked' mean, what are the implications? For the labels, this doesn't seem like much of a problem for the good/bad case. Are there times it is very confusing e.g. 4-5 possible labels good bad mua noise etc?

examples/tutorials/curation/plot_3_upload_a_model.py
```diff
@@ -0,0 +1,126 @@
+"""
```
JoeZiminski, line 1: Not sure about this, but could this be a subsection of the above 'how to train a model' tutorial, as both are reasonably short? It was a little confusing to see three tutorials on the automated curation stuff, though compartmentalization is also very nice.

examples/tutorials/curation/plot_2_train_a_model.py
```diff
@@ -0,0 +1,125 @@
+"""
+Training a model for automated curation
+=============================
+
+If the pretrained models do not give satisfactory performance on your data, it is easy to train your own classifier using SpikeInterface.
+"""
```
JoeZiminski, line 6: Just for readability, I would consider moving the import section under the first header for Step 1 and just saying 'first we will import our dependencies'

examples/tutorials/curation/plot_2_train_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+# Note, you can set the number of cores you use using e.g.
+# si.set_global_job_kwargs(n_jobs = 8)
+
+##############################################################################
+# Step 1: Generate and label data
+# -------------------------------
+#
+# For this tutorial, we will use simulated data to create ``recording`` ``sorting`` objects. We'll
+# create two sorting objects: :code:`sorting_1` is coupled to the real recording, so will contain good
+# units; :code:`sorting_2` is uncoupled, so should produce noise. We'll combine the two into one sorting
```
JoeZiminski, line 28: I get this and it's very nice idea, but could be quite mysterious / confusing to less experienced users. Can a sentence be added to make this very explicit? i.e. in sorting_1 the  times will match exactly the spike in the recording. For sorting_2 the spike times will not be matched with the recording so will mostly be random noise.

examples/tutorials/curation/plot_2_train_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+# object using :code:`si.aggregate_units`.
+#
+# (When making your own model, you should
```
JoeZiminski, line 31: This is a really cool section but I got a bit confused during it. I would reorder slightly and add some examples to make it super clear how you could use it in real life. For example:

1) We will create the test dataset as above. Then, `plot_unit_templates` to confirm our test data is setup properly.
2) In real life, we would have a recording and sorting that is our data. We would then compute quality metrics and manually curate a set of labels. Now, create the sorting analyzer and compute metrics
3) mention again in real life we would now manually curate the data. and tada! here's some labels we got from this.

examples/tutorials/curation/plot_2_train_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+# `load your own recording <https://spikeinterface.readthedocs.io/en/latest/modules/extractors.html>`_
+# and `do a sorting <https://spikeinterface.readthedocs.io/en/latest/modules/sorters.html>`_ on your data.)
+
+recording, sorting_1 = si.generate_ground_truth_recording(num_channels=4, seed=1, num_units=5)
+_, sorting_2 =si.generate_ground_truth_recording(num_channels=4, seed=2, num_units=5)
+
+both_sortings = si.aggregate_units([sorting_1, sorting_2])
+
+##############################################################################
+# The models are based on `quality metrics <https://spikeinterface.readthedocs.io/en/latest/modules/qualitymetrics.html>`_
```
JoeZiminski, line 41: At this stage,  I would not lead with the models. I would say we can select the quality metrics and postprocessing we will use to manually curate the data. Ultimately, the models will be trained on the quality metrics and postprocessing steps that are computed here.

examples/tutorials/curation/plot_2_train_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+
+##############################################################################
+# Let's plot the templates for the first and fifth units. The first (unit id 0) belongs to
+# :code:`sorting_1` so should look like a real unit; the sixth (unit id 5) belongs to :code:`sorting_2`
+# so should look like noise.
+
+sw.plot_unit_templates(analyzer, unit_ids=[0,5])
+
+##############################################################################
+# This is as expected: great! Find out more about plotting using widgets `here <https://spikeinterface.readthedocs.io/en/latest/modules/widgets.html>`_. The labels
```
JoeZiminski, line 57: I would skip the link to the widgets at this stage it was slightly distracting

examples/tutorials/curation/plot_2_train_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+# for our units are then easy to put in a list:
```
JoeZiminski, line 58: Make clear these are generated from a manual curation step that is hidden from the reader

JoeZiminski: and even though obvious, note that the first 5 (sorting 1) are good and second 5 (sorting 2) are bad as expected from the test data.

examples/tutorials/curation/plot_2_train_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+# With the labelled data, we can train the model using the :code:`train_model` function.
+# Here, the idea is that the trainer will try several classifiers, imputation strategies and
+# scaling techniques then save the most accurate. To save time, we'll only try one classifier
+# (Random Forest), imputation strategy (median) and scaling technique (standard scaler).
+
+folder = "my_model"
+
+# We will use a list of one analyzer here, we would strongly advise using more than one to
+# improve model performance
+trainer = sc.train_model(
```
JoeZiminski, line 75: again really nice API

examples/tutorials/curation/plot_2_train_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+    mode = "analyzers", # You can supply a labelled csv file instead of an analyzer
+    labels = [labels],
+    analyzers = [analyzer],
+    folder = folder, # Where to save the model and model_info.json file
+    metric_names = None, # Specify which metrics to use for training: by default uses those already calculted
+    imputation_strategies = ["median"], # Defaults to all
+    scaling_techniques = ["standard_scaler"], # Defaults to all
+    classifiers = None, # Default to Random Forest only. Other classifiers you can try [ "AdaBoostClassifier","GradientBoostingClassifier","LogisticRegression","MLPClassifier"]
+    overwrite = True, # Whether or not to overwrite `folder` if it already exists. Default is False.
+)
```
JoeZiminski, line 85: this was showing an error on the build docs for this PR

examples/tutorials/curation/plot_2_train_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+
+# We will use a list of one analyzer here, we would strongly advise using more than one to
+# improve model performance
+trainer = sc.train_model(
+    mode = "analyzers", # You can supply a labelled csv file instead of an analyzer
+    labels = [labels],
+    analyzers = [analyzer],
+    folder = folder, # Where to save the model and model_info.json file
+    metric_names = None, # Specify which metrics to use for training: by default uses those already calculted
+    imputation_strategies = ["median"], # Defaults to all
```
JoeZiminski, line 81: nice, are there API docs for this outlining all possible inputs?

examples/tutorials/curation/plot_2_train_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+# (Random Forest), imputation strategy (median) and scaling technique (standard scaler).
+
+folder = "my_model"
+
+# We will use a list of one analyzer here, we would strongly advise using more than one to
+# improve model performance
+trainer = sc.train_model(
+    mode = "analyzers", # You can supply a labelled csv file instead of an analyzer
+    labels = [labels],
+    analyzers = [analyzer],
```
JoeZiminski, line 78: I think it might be worth explaining in detail what happens with multiple analyzers, I had to think about it a bit. I guess in reality, you always want to do this. Maybe a short code example underneath?

'In reality, we will usually train a model using multiple analyzers from an experiment. Recall each analyzer contains a set of quality metrics per unit, and we provide the ground-truth (human curated) label for each unit to use in training. Therefore, we have a set of sorting analyzer and corresponding label for each recording in our dataset.

For example, we might load previously generated analyzers and labels from file like:

```
all_labels = []
all_analyzers = []
for name in [some list of animal names]
    analzyer = si.load...
   labels = si.load...
   append..    


```

examples/tutorials/curation/plot_2_train_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+
+##############################################################################
+# The above code saves the model in ``my_model/model.skops``, some metadata in
+# ``my_model/model_info.json`` and the model accuracies in ``model_accuracies.csv``
+# in the specified ``folder``.
+#
+# ``skops`` is a file format: you can think of it as a more-secture pkl file. `Read more <https://skops.readthedocs.io/en/stable/index.html>`_.
+#
+# The ``model_accuracies.csv`` file contains the accuracy, precision and recall of the
+# tested models. Let's take a look
```
JoeZiminski, line 97:
```suggestion
# tested models. Let's take a look:
```

examples/tutorials/curation/plot_2_train_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+# Note, you can set the number of cores you use using e.g.
+# si.set_global_job_kwargs(n_jobs = 8)
+
+##############################################################################
+# Step 1: Generate and label data
+# -------------------------------
+#
+# For this tutorial, we will use simulated data to create ``recording`` ``sorting`` objects. We'll
```
JoeZiminski, line 26:
```suggestion
# For this tutorial, we will use simulated data to create ``recording`` and ``sorting`` objects. We'll
```

examples/tutorials/curation/plot_2_train_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+# tested models. Let's take a look
+
+accuracies = pd.read_csv(Path(folder) / "model_accuracies.csv", index_col = 0)
+accuracies.head()
+
+##############################################################################
+# Our model is perfect!! This is because the task was *very* easy. We had 10 units; where
+# half were pure noise and half were not.
+#
+# The model also contains some more information, such as which features are importantly.
```
JoeZiminski, line 106:
```suggestion
# The model also contains some more information, such as which features are important.
```

JoeZiminski, line 106: Could go into a bit more detail here, what is meant by 'important'? Presumably this is how much each quality metric contributes to the prediction. What do we do with QM that don't contribute much? Shall we drop them?

examples/tutorials/curation/plot_2_train_a_model.py
```diff
@@ -0,0 +1,125 @@
...
+indices = np.argsort(importances)[::-1]
+features = best_model.feature_names_in_
+n_features = best_model.n_features_in_
+
+plt.figure(figsize=(12, 6))
+plt.title("Feature Importances")
+plt.bar(range(n_features), importances[indices], align="center")
+plt.xticks(range(n_features), features, rotation=90)
+plt.xlim([-1, n_features])
+plt.show()
```
JoeZiminski, line 120: This plot is not showing but i guess it's because of the above error

JoeZiminski (commented):
Hey I reviewed the docs first, very nice! I really like the API, in general it seems very intuitive to use. I think this will be a very nice addition. 

I read through the docs as someone unfamiliar with this and just highlighted anything that was not immediately clear to me in a blanket way, please feel free to ignore any suggestions you don't think are useful!

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,248 @@
...
+# model and some metadata about the model.
+
+model, model_info = sc.load_model(
+    repo_id = "SpikeInterface/toy_tetrode_model",
+    trusted = ['numpy.dtype']
+)
+
+
+##############################################################################
+# This model was trained on artifically generated tetrode data. The model object has a nice html
```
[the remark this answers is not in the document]
JoeZiminski: I see this comes a bit later, I have a suggestion there on reordering

examples/tutorials/curation/plot_2_train_a_model.py
JoeZiminski: I guess there would need to be a step to ensure all anazyers have the same QMs computed?

src/spikeinterface/curation/train_manual_curation.py
```diff
@@ -0,0 +1,796 @@
...
+                        qm_params_1 = analyzer_1.extensions["quality_metrics"].params["qm_params"]
+                    if analyzer_2.has_extension("quality_metrics") is True:
+                        qm_params_2 = analyzer_2.extensions["quality_metrics"].params["qm_params"]
+                    if analyzer_1.has_extension("template_metrics") is True:
+                        tm_params_1 = analyzer_1.extensions["template_metrics"].params["metrics_kwargs"]
+                    if analyzer_2.has_extension("template_metrics") is True:
+                        tm_params_2 = analyzer_2.extensions["template_metrics"].params["metrics_kwargs"]
+
+                    conflicting_metrics_between_1_2 = []
+                    # check quality metrics params
```
JoeZiminski, line 281: How similar is this function to `check_metric_names_are_the_same`?

src/spikeinterface/curation/train_manual_curation.py
```diff
@@ -0,0 +1,796 @@
...
+
+        # Extract features
+        try:
+            if (set(self.metric_names) - set(self.testing_metrics.columns) != set()) and self.verbose is True:
+                print(
+                    f"Dropped metrics (calculated but not included in metric_names): {set(self.testing_metrics.columns) - set(self.metric_names)}"
+                )
+            self.X = self.testing_metrics[self.metric_names]
+        except KeyError as e:
+            print("metrics_list contains invalid metric names")
```
JoeZiminski, line 340: Not sure but this message will end up right at the top of the stack trace and may not be easily seen. If possible you could append it to `str(e)` something like `raise KeyError(f"{str(e)}, metrics list ...!")`

src/spikeinterface/curation/tests/test_train_manual_curation.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+def test_load_and_preprocess_full(trainer):
+    temp_file_path = make_temp_training_csv()
+
+    # Load and preprocess the data from the temporary CSV file
+    trainer.load_and_preprocess_csv([temp_file_path])
+
+    # Assert that the data is loaded and preprocessed correctly
+    assert trainer.X is not None
```
JoeZiminski, line 48: is it easy to check that the values are exactly as expected?

src/spikeinterface/curation/tests/test_train_manual_curation.py
```diff
@@ -0,0 +1,206 @@
...
+    assert trainer.y is not None
+    assert trainer.testing_metrics is not None
+
+
+def test_apply_scaling_imputation(trainer):
```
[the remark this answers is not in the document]
JoeZiminski: same with imputer

src/spikeinterface/curation/tests/test_train_manual_curation.py
```diff
@@ -0,0 +1,206 @@
...
+    assert y_val is not None
+    assert imputer is not None
+    assert scaler is not None
+
+
+def test_get_classifier_search_space(trainer):
+
+    classifier = "LogisticRegression"
+    model, param_space = trainer.get_classifier_search_space(classifier)
+    assert model is not None
```
JoeZiminski, line 76: is it possible to check some metadata on the model to ensure it is as expected e.g. of "LogisticRegression" type? Is it worth testing more of the model types?

src/spikeinterface/curation/tests/test_train_manual_curation.py
```diff
@@ -0,0 +1,206 @@
...
+
+    trainer.X = np.ones((10, 3))
+    trainer.y = np.append(np.ones(5), np.zeros(5))
+
+    trainer.evaluate_model_config()
+    trainer_folder = Path(trainer.folder)
+    assert trainer_folder.is_dir()
+    assert (trainer_folder / "best_model.skops").is_file()
+    assert (trainer_folder / "model_accuracies.csv").is_file()
+    assert (trainer_folder / "model_info.json").is_file()
```
JoeZiminski, line 105: possible to check the outputs more thoroughly to ensure they are as expected for this test case?

src/spikeinterface/curation/tests/test_train_manual_curation.py
```diff
@@ -0,0 +1,206 @@
...
+        metric_names=metric_names,
+        imputation_strategies=["median"],
+        scaling_techniques=["standard_scaler"],
+        classifiers=["LogisticRegression"],
+        overwrite=True,
+    )
+    assert isinstance(trainer, CurationModelTrainer)
+
+
+def test_train_model():
```
JoeZiminski, line 130: There are a lot of important parameters for this function, I think the majority are tested above. However it would be nice to check all of these are doing what is expected when this function is called (either that outputs of some test case match as expected, or they are propagated to functions that are tested elsewhere correctly using monkeypatching. This type of deeper testing is quite laborious to setup so might be too much to ask, but a benefit is you can be 100% sure your functions are doing what you expect going forward, and are not broken in some future refactoring / other code changes.

src/spikeinterface/curation/train_manual_curation.py
```diff
@@ -0,0 +1,796 @@
...
+        best_pipeline = Pipeline(
+            [("imputer", best_imputer), ("scaler", best_scaler), ("classifier", best_model.best_estimator_)]
+        )
+
+        self.best_pipeline = best_pipeline
+
+        if self.folder is not None:
+            self._save()
+
+    def _save(self):
```
JoeZiminski, line 549: is this tested in `test_train_manual_curation.py`? Maybe it is implicitly tested through testing other functionality

JoeZiminski (commented):
Please see comments on the code below, I just wrote anything that came to mind, especially for tests, I mentioned a lot of possibilities that would serve to regression-proof this going forward, but it is a lot, so feel free to ignore. Overall this is really nice addition, will be very useful, and the code is super clean / well structured and easy to review.

If possible it would be nice to expand the PR message, I think down the line this is the kind of very useful module that will be a key part of SpikeInterface and extended, so would be useful to have some detail on the PR message for provenance e.g. 1) what does this PR do 2) what approach does it take (high-level) 3) maybe a bit on the background, as I guess this was previously used by the Allen. 

Overall really nice I look forward to using it!

examples/tutorials/curation/plot_3_upload_a_model.py
```diff
@@ -0,0 +1,126 @@
...
+Upload a pipeline to Hugging Face Hub
+=====================================
+"""
+##############################################################################
+# In this tutorial we will upload a pipeline, trained in SpikeInterface, to the
```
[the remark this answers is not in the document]
alejoe91: good idea!

doc/how_to/auto_curation_prediction.rst
```diff
@@ -0,0 +1,40 @@
+How to use a trained model to predict the curation labels
```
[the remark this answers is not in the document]
chrishalcrow: Idea is to give a short, narrow answer. While the tutorials give more context and discuss other options/kwargs, following: https://numpy.org/doc/stable/user/how-to-how-to.html 
Will add a link to the tutorial

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,248 @@
...
+# note: you can use more cores using e.g.
+# si.set_global_jobs_kwargs(n_jobs = 8)
+
+##############################################################################
+# Download a pretrained model
+# ---------------------------
+#
+# Let's download a pretrained model from `Hugging Face <https://huggingface.co/>`_ (HF). The
+# ``load_model`` function allows us to download directly from HF, or use a model in a local
+# folder. The function downloads the model and saves it in a temporary folder and returns a
```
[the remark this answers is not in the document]
chrishalcrow: Mostly covered in new intro; and there's a discussion about using similar models later.

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,248 @@
...
+# model and some metadata about the model.
+
+model, model_info = sc.load_model(
+    repo_id = "SpikeInterface/toy_tetrode_model",
```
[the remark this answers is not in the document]
chrishalcrow: Will mention this when we look at the HTML representation.

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,248 @@
...
+# This model was trained on artifically generated tetrode data. The model object has a nice html
+# representation, which will appear if you're using a Jupyter notebook.
+
+model
+
+##############################################################################
+# The model object (an sklearn Pipeline) contains information about which metrics
+# were used to compute the model. We can access it from the model (or from the model_info)
+
+print(model.feature_names_in_)
```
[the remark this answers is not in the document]
chrishalcrow: Unsure - it's from from `sklearn`

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,248 @@
...
+
+##############################################################################
+# Hence, to use this model we need to create a ``sorting_analyzer`` with all these metrics computed.
+# We'll do this by generating a recording and sorting, creating a sorting analyzer and computing a
+# bunch of extensions. Follow these links for more info on `recordings <https://spikeinterface.readthedocs.io/en/latest/modules/extractors.html>`_, `sortings <https://spikeinterface.readthedocs.io/en/latest/modules/sorters.html>`_, `sorting analyzers <https://spikeinterface.readthedocs.io/en/latest/tutorials/core/plot_4_sorting_analyzer.html#sphx-glr-tutorials-core-plot-4-sorting-analyzer-py>`_
+# and `extensions <https://spikeinterface.readthedocs.io/en/latest/modules/postprocessing.html>`_.
+
+recording, sorting = si.generate_ground_truth_recording(num_channels=4, seed=4, num_units=10)
+sorting_analyzer = si.create_sorting_analyzer(sorting=sorting, recording=recording)
+sorting_analyzer.compute(['noise_levels','random_spikes','waveforms','templates','spike_locations','spike_amplitudes','correlograms','principal_components','quality_metrics','template_metrics'])
```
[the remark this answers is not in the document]
chrishalcrow: No, the `feature_names_in` are just the individual quality and template metrics

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,248 @@
...
+# we'll be a conservative human, who has labelled several units with small amplitudes as 'bad'.
+
+human_labels = ['bad', 'good', 'good', 'bad', 'good', 'bad', 'good', 'bad', 'good', 'good']
+
+# Note: if you labelled using phy, you can load the labels using:
+# human_labels = sorting_analyzer.sorting.get_property('quality')
+
+from sklearn.metrics import confusion_matrix, balanced_accuracy_score
+
+label_conversion = model_info['label_conversion']
```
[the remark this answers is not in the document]
chrishalcrow: Added a description

examples/tutorials/curation/plot_1_automated_curation.py
chrishalcrow: I like having the model performance after the toy_model stuff, because it uses the same dataset. I think we should just link to this section above, and try and improve it a bit

examples/tutorials/curation/plot_1_automated_curation.py
chrishalcrow: Yup yup, we can do this using auto_label_units

examples/tutorials/curation/plot_1_automated_curation.py
chrishalcrow: The model we're using here is being updated. So I'll do these updates once that's finished.

examples/tutorials/curation/plot_2_train_a_model.py
chrishalcrow: Yeah, the `train_model` function checks that the same QMs are computed.
I agree about explaining in more detail, but I think the example of loading a list of analyzers is a bit overkill/distracting.

examples/tutorials/curation/plot_2_train_a_model.py
chrishalcrow: Yeah, that's a good question... it's basically anything that sklearn can take, which is an overwhelming list. @anoushkajain @jakeswann1 do you know a good resource?

examples/tutorials/curation/plot_2_train_a_model.py
chrishalcrow: Thanks - we added PC metrics and some of them had `None` default_params, which we weren't expecting.

examples/tutorials/curation/plot_2_train_a_model.py
chrishalcrow: Currently we don't do anything, which is reasonable. There are some issues with dropping features: suppose you have two highly correlated features. Only one is needed for classification and you'll likely end up with one of these features being "important" and one being unimportant. But which one will depend on the random seed of your training. So it requires thought/analysis from the model creator to decide what to do.

examples/tutorials/curation/plot_2_train_a_model.py
chrishalcrow: I'm pro the widget link. I've put it in some brackets??

examples/tutorials/curation/plot_3_upload_a_model.py
```diff
@@ -0,0 +1,126 @@
...
+# `Hugging Face Hub <https://huggingface.co/>`_ (HFH). To do this, you need a HFH account.
+# If you do not want to make an account, you can simply share the model folder with colleagues.
+# There are also several ways to interaction with HFH: the way we propose here doesn't use
+# many of the tools ``skops`` and hugging face have developed such as the ``Card`` and
+# ``hub_utils``. Feel free to check those out `here <https://skops.readthedocs.io/en/stable/examples.html>`_.
+#
+# The plan is to make a folder with the following file structure
```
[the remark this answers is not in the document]
chrishalcrow: Do you not think the comment below clarifies this?

examples/tutorials/curation/plot_3_upload_a_model.py
```diff
@@ -0,0 +1,126 @@
...
+#         "subject_species": ["Mus musculus"],
+#         "brain_areas": ["CA1"],
+#         "probes":
+#             [{
+#              "manufacturer": "IMEc",
+#              "name": "Neuropixels 2.0"
+#             }]
+#         }
+#
+#     import json
```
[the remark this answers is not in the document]
chrishalcrow: I'll move it to the top of this codeblock

examples/tutorials/curation/plot_3_upload_a_model.py
```diff
@@ -0,0 +1,126 @@
...
+#     with open("my_model_folder/metadata.json", "w") as file:
+#         json.dump(model_metadata, file)
+#
+# You could now share this folder with a colleague, or upload it to github. Or if you'd
```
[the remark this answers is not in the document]
chrishalcrow: My thought was: if the user doesn't want to use HFH, this is where they can exit the tutorial, and they don't have to go any further. I'll maybe just reorder this paragraph a little...

src/spikeinterface/curation/model_based_curation.py
```diff
@@ -0,0 +1,430 @@
...
+
+    def _get_metrics_for_classification(self):
+        """Check if all required metrics are present and return a DataFrame of metrics for classification"""
+
+        import pandas as pd
+
+        quality_metrics, template_metrics = try_to_get_metrics_from_analyzer(self.sorting_analyzer)
+
+        # Create DataFrame of all metrics and reorder columns to match the model
+        calculated_metrics = pd.concat([quality_metrics, template_metrics], axis=1)
```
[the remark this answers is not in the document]
chrishalcrow: Nice spot! We removed this as it's not needed. I then panicked that it is needed; so I've added a new test checking for independence of metric order

src/spikeinterface/curation/model_based_curation.py
```diff
@@ -0,0 +1,430 @@
...
+            A dataframe containing the classified units and their corresponding predictions and probabilities,
+            indexed by their `unit_ids`.
+        """
+        import pandas as pd
+
+        # Get metrics DataFrame for classification
+        if input_data is None:
+            input_data = self._get_metrics_for_classification()
+        else:
+            if not isinstance(input_data, pd.DataFrame):
```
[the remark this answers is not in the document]
chrishalcrow: Yes! Wonderful idea!

src/spikeinterface/curation/model_based_curation.py
```diff
@@ -0,0 +1,430 @@
...
+                raise ValueError("Input data must be a pandas DataFrame")
+
+        if model_info is not None:
```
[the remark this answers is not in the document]
chrishalcrow: We want a bit more information than is kept in self.pipeline. The bigget thing is that `model_info` also contains the metric params, and the spikeinterface version.

src/spikeinterface/curation/model_based_curation.py
```diff
@@ -0,0 +1,430 @@
...
+            except AssertionError:
+                raise ValueError("Labels in predictions do not match those in label_conversion")
+            predictions = [label_conversion[label] for label in predictions]
+
+        classified_units = pd.DataFrame(
+            zip(predictions, probabilities), columns=["prediction", "probability"], index=self.sorting_analyzer.unit_ids
+        )
+
+        # Set predictions and probability as sorting properties
+        self.sorting_analyzer.sorting.set_property("classifier_label", predictions)
```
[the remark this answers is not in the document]
chrishalcrow: Done!

src/spikeinterface/curation/model_based_curation.py
```diff
@@ -0,0 +1,430 @@
...
+
+        import pandas as pd
+
+        quality_metrics, template_metrics = try_to_get_metrics_from_analyzer(self.sorting_analyzer)
+
+        # Create DataFrame of all metrics and reorder columns to match the model
+        calculated_metrics = pd.concat([quality_metrics, template_metrics], axis=1)
+
+        # Remove any metrics for non-existent units, raise error if no units are present
+        calculated_metrics = calculated_metrics.loc[
```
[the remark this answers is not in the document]
chrishalcrow: Any idea how to make a 0 unit sorter easily? `generate_ground_truth_recording` doesn't work with 0 units.

src/spikeinterface/curation/model_based_curation.py
```diff
@@ -0,0 +1,430 @@
...
+            If True and the parameters used to compute the metrics in `sorting_analyzer` are different than the parmeters
+            used to compute the metrics used to train the model, this function will raise an error. Otherwise, a warning is raised.
+        model_info_path : str or Path, default: None
+            Path to model_info.json provenance file
+        """
+
+        quality_metrics_extension = self.sorting_analyzer.get_extension("quality_metrics")
+        template_metrics_extension = self.sorting_analyzer.get_extension("template_metrics")
+
+        if quality_metrics_extension is not None:
```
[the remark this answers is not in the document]
chrishalcrow: Here we're checking the *parameters* used to compute the metrics. We do this seperately because people might want to be strict (we will use the exact same parameters as were used when training the model) or be more free.

Note: if we unify the template and quality metric extensions, this could be refactored nicely.

src/spikeinterface/curation/model_based_curation.py
```diff
@@ -0,0 +1,430 @@
...
+                    raise Exception(warning_message)
+                else:
+                    warnings.warn(warning_message)
+
+            if template_metrics_params != model_template_metrics_params:
+                warning_message = "Template metrics params do not match those used to train model. Parameters can be found in the 'model_info.json' file."
+                if enforce_metric_params is True:
+                    raise Exception(warning_message)
+                else:
+                    warnings.warn(warning_message)
```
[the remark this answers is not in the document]
chrishalcrow: No, we keep all the columns. The columns aren't necessarily bad, just computed using different parameters.

src/spikeinterface/curation/model_based_curation.py
```diff
@@ -0,0 +1,430 @@
...
+            input_data = calculated_metrics[self.required_metrics]
+        else:
+            raise ValueError(
+                "Input data does not contain all required metrics for classification",
+                f"Missing metrics: {required_metrics.difference(calculated_metrics)}",
+            )
+
+        return input_data
+
+    def _check_params_for_classification(self, enforce_metric_params=False, model_info=None):
```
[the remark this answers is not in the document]
chrishalcrow: Yeah, you're right. And when we use this function, it's under an `if model_info is not None`. I don't really know what the pythonic thing to do is. Should we do the `None` check in this function instead, and raise an error if it is?

src/spikeinterface/curation/model_based_curation.py
```diff
@@ -0,0 +1,430 @@
...
+        label_conversion=label_conversion,
+        export_to_phy=export_to_phy,
+        model_info=model_info,
+        enforce_metric_params=enforce_metric_params,
+    )
+
+    return classified_units
+
+
+def load_model(model_folder=None, repo_id=None, model_name=None, trust_model=False, trusted=None):
```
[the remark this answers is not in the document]
chrishalcrow: How do you test this? Put a test model on hugging face, get it, and check that it's what we expect?

src/spikeinterface/curation/train_manual_curation.py
```diff
@@ -0,0 +1,796 @@
...
+            consistent_metric_params = {
+                metric: analyzers[0].extensions["quality_metrics"].params["qm_params"][metric]
+                for metric in consistent_metrics
+            }
+            self.metrics_params["quality_metric_params"]["qm_params"] = consistent_metric_params
+
+        if analyzers[0].has_extension("template_metrics") is True:
+            self.metrics_params["template_metric_params"] = analyzers[0].extensions["template_metrics"].params
+            if "template_metrics" in conflicting_metrics:
+                self.metrics_params["template_metric_params"] = analyzers[0].extensions["template_metrics"].params = {}
```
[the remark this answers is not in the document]
chrishalcrow: Not intended - thanks!

src/spikeinterface/curation/model_based_curation.py
```diff
@@ -0,0 +1,430 @@
...
+    elif len(skops_files) == 1:
+        skops_file = skops_files[0]
+
+    if trust_model and trusted is None:
+        try:
+            model = skio.load(skops_file)
+        except UntrustedTypesFoundException as e:
+            exception_msg = str(e)
+            # the exception message contains the list of untrusted objects after a colon and enswith a period
+            trusted = eval(exception_msg.split(":")[1][:-1])
```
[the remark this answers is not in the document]
chrishalcrow: Done! (Once I push...)

src/spikeinterface/curation/train_manual_curation.py
chrishalcrow: Not very similar, because this is checking the parameters used to compute the metrics.

src/spikeinterface/curation/train_manual_curation.py
```diff
@@ -0,0 +1,796 @@
...
+                    f"Dropped metrics (calculated but not included in metric_names): {set(self.testing_metrics.columns) - set(self.metric_names)}"
+                )
+            self.X = self.testing_metrics[self.metric_names]
+        except KeyError as e:
+            print("metrics_list contains invalid metric names")
+            raise e
+        self.X = self.testing_metrics.reindex(columns=self.metric_names)
+        self.X = self.X.map(lambda x: np.nan if np.isinf(x) else x)
+        self.X = self.X.astype("float32")
+        self.X.fillna(0, inplace=True)
```
chrishalcrow, line 345: Not sure we should be doing this

src/spikeinterface/curation/train_manual_curation.py
```diff
@@ -0,0 +1,796 @@
...
+        # Apply SMOTE for class imbalance
+        if self.smote:
+            try:
+                from imblearn.over_sampling import SMOTE
+            except ModuleNotFoundError:
+                raise ModuleNotFoundError("Please install imbalanced-learn package to use SMOTE")
+            smote = SMOTE(random_state=self.seed)
+            X_train_scaled, y_train = smote.fit_resample(X_train_scaled, y_train)
+
+        return X_train_scaled, X_val_scaled, y_train, y_val, imputer, scaler
```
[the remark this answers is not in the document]
chrishalcrow: Yes, for sure. That also matches the notation we use when we actually apply the funciton.

Could someone check my edits on this function, to make sure my name changes haven't changed it, please?

src/spikeinterface/curation/train_manual_curation.py
```diff
@@ -0,0 +1,796 @@
...
+        The method converts the classifier names to actual classifier instances before evaluating them.
+        The evaluation results, including the best model and its parameters, are saved to the output folder.
+        """
+        from sklearn.model_selection import train_test_split
+
+        X_train, X_test, y_train, y_test = train_test_split(
+            self.X, self.y, test_size=self.test_size, random_state=self.seed, stratify=self.y
+        )
+        classifier_instances = [self.get_classifier_instance(clf) for clf in self.classifiers]
+        self._evaluate(
```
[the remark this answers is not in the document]
chrishalcrow: There might be. It already does quite a lot: there are lots of classifier instances because it tries out different hyperparameters of the classifiers. Someone who knows more might be able to comment @anoushkajain @alejoe91

src/spikeinterface/curation/train_manual_curation.py
chrishalcrow: Yeah, this is used implicitly during the testing e.g. in the trainer() fixture in `test_train_manual_curation`.

src/spikeinterface/curation/train_manual_curation.py
```diff
@@ -0,0 +1,796 @@
...
+    if labels is None:
+        raise Exception("You must supply a list of lists of curated labels using `labels = [[...],[...],...]`")
+
+    if mode not in ["analyzers", "csv"]:
+        raise Exception("`mode` must be equal to 'analyzers' or 'csv'.")
+
+    if (test_size > 1.0) or (0.0 > test_size):
+        raise Exception("`test_size` must be between 0.0 and 1.0")
+
+    trainer = CurationModelTrainer(
```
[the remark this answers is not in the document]
chrishalcrow: I've never written a monkeypatch test; sounds like a good thing to learn to do! Watch this space...

src/spikeinterface/curation/train_manual_curation.py
```diff
@@ -0,0 +1,796 @@
...
+
+    for i, metrics_for_analyzer_1 in enumerate(metrics_for_each_analyzer):
+        for j, metrics_for_analyzer_2 in enumerate(metrics_for_each_analyzer):
+            if i > j:
+                metric_names_1 = set(metrics_for_analyzer_1.keys())
+                metric_names_2 = set(metrics_for_analyzer_2.keys())
+                if metric_names_1 != metric_names_2:
+                    metrics_in_1_but_not_2 = metric_names_1.difference(metric_names_2)
+                    metrics_in_2_but_not_1 = metric_names_2.difference(metric_names_1)
+                    print(metrics_in_1_but_not_2)
```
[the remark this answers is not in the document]
chrishalcrow: Yess, thanks

src/spikeinterface/curation/train_manual_curation.py
```diff
@@ -0,0 +1,796 @@
...
+            self.scaling_techniques,
+            classifier_instances,
+            X_train,
+            X_test,
+            y_train,
+            y_test,
+            self.search_kwargs,
+        )
+
+    def _get_metrics_for_classification(self, analyzer):
```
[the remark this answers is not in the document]
chrishalcrow: Yes - done!

examples/tutorials/curation/plot_3_upload_a_model.py
```diff
@@ -0,0 +1,126 @@
+"""
+Upload a pipeline to Hugging Face Hub
+=====================================
+"""
+##############################################################################
+# In this tutorial we will upload a pipeline, trained in SpikeInterface, to the
+# `Hugging Face Hub <https://huggingface.co/>`_ (HFH). To do this, you need a HFH account.
+# If you do not want to make an account, you can simply share the model folder with colleagues.
+# There are also several ways to interaction with HFH: the way we propose here doesn't use
```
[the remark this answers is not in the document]
chrishalcrow: I want to mention this up-front since, as a user, I put off by having to create an account to use a feature. So I want people to still keep interested, even if they're put off by the account thing. To try and isolate the information, I've subdivided the page in three subsections, hope this helps!

examples/tutorials/curation/plot_3_upload_a_model.py
```diff
@@ -0,0 +1,126 @@
...
+#         model_info.json
+#         training_data.csv
+#         labels.csv
+#         metadata.json
+#
+# SpikeInterface doesn't require you to keep this folder structure, we just advise it as
+# best practice.
+#
+# If you've used SpikeInterface to train your model, you have already created such a folder,
+# containing everything except the ``metadata.json`` file. In this file, we suggest saving
```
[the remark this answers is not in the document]
chrishalcrow: Tried to clarify

examples/tutorials/curation/plot_3_upload_a_model.py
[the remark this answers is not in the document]
chrishalcrow: I again think it's useful to show that you can share stuff without HFH. I've put the comment in brackets to try and not break the flow.

examples/tutorials/curation/plot_1_automated_curation.py
chrishalcrow: Tried to clarify - it is a bit of a tough graph...

examples/tutorials/curation/plot_2_train_a_model.py
chrishalcrow: Adding some resources...

examples/tutorials/curation/plot_2_train_a_model.py
chrishalcrow: Added a small additional sentence.

src/spikeinterface/curation/model_based_curation.py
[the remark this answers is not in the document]
chrishalcrow: Figured it out: `sort = sort.select_units(unit_ids=[])` !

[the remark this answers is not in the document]
chrishalcrow: Ok, you can't make a sorting analyzer with no units. I don't think we need to test this...

examples/tutorials/curation/plot_1_automated_curation.py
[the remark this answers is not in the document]
chrishalcrow: Note to self: if we unify the quality and template metrics it would be easy to make a funciton `compute_all_metrics(metric_names=model.feature_names_in_)`

examples/tutorials/curation/plot_3_upload_a_model.py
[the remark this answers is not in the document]
chrishalcrow: Hello - we discussed during the maintenance meeting yesterday and we agreed to keep it as two, but with a couple extra links between the pages.

examples/tutorials/curation/plot_1_automated_curation.py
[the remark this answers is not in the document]
chrishalcrow: I've added a comment and link here, to the realistic model below. Hopefully that helps!

examples/tutorials/curation/plot_1_automated_curation.py
chrishalcrow: done!

examples/tutorials/curation/plot_1_automated_curation.py
chrishalcrow: Model is updated, so this section is now significantly updated.

examples/tutorials/curation/plot_3_upload_a_model.py
chrishalcrow: Discussed above

src/spikeinterface/curation/tests/test_model_based_curation.py
```diff
@@ -0,0 +1,140 @@
...
+
+    model_info = {"metric_params": {}}
+    model_info["metric_params"]["quality_metric_params"] = sorting_analyzer_for_curation.get_extension(
+        "quality_metrics"
+    ).params
+    model_info["metric_params"]["template_metric_params"] = sorting_analyzer_for_curation.get_extension(
+        "template_metrics"
+    ).params
+
+    model_based_classification._check_params_for_classification(model_info=model_info)
```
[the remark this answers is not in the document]
chrishalcrow: Done in `test_exception_raised_when_metricparams_not_equal`

src/spikeinterface/curation/tests/test_model_based_curation.py
```diff
@@ -0,0 +1,140 @@
...
+    # Function should fail here
+    with pytest.raises(ValueError):
+        model_based_classification._export_to_phy(classified_units)
+    # Make temp output folder and set as phy_folder
+    phy_folder = cache_folder / "phy_folder"
+    phy_folder.mkdir(parents=True, exist_ok=True)
+
+    model_based_classification.sorting_analyzer.sorting.annotate(phy_folder=phy_folder)
+    model_based_classification._export_to_phy(classified_units)
+    assert (phy_folder / "cluster_prediction.tsv").exists()
```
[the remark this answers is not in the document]
chrishalcrow: I don't know much about this since I don't use Phy. Any advice @jakeswann1 @anoushkajain ?

src/spikeinterface/curation/tests/test_model_based_curation.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+@pytest.fixture
+def required_metrics():
+
+    return ["num_spikes", "snr", "half_width"]
+
+
+def test_model_based_classification_init(sorting_analyzer_for_curation, model):
+    # Test the initialization of ModelBasedClassification
```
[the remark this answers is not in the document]
chrishalcrow: All tests (except the phy one) now have docstrings

src/spikeinterface/curation/train_manual_curation.py
```diff
@@ -0,0 +1,796 @@
...
+            if (set(self.metric_names) - set(self.testing_metrics.columns) != set()) and self.verbose is True:
+                print(
+                    f"Dropped metrics (calculated but not included in metric_names): {set(self.testing_metrics.columns) - set(self.metric_names)}"
+                )
+            self.X = self.testing_metrics[self.metric_names]
+        except KeyError as e:
+            print("metrics_list contains invalid metric names")
+            raise e
+        self.X = self.testing_metrics.reindex(columns=self.metric_names)
+        self.X = self.X.map(lambda x: np.nan if np.isinf(x) else x)
```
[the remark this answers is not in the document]
chrishalcrow: Done!

src/spikeinterface/curation/tests/test_train_manual_curation.py
```diff
@@ -0,0 +1,206 @@
...
+    scaling_technique = "standard_scaler"
+    X_train = np.array([[1, 2, 3], [4, 5, 6]])
+    X_val = np.array([[7, 8, 9], [10, 11, 12]])
+    y_train = np.array([0, 1])
+    y_val = np.array([2, 3])
+    X_train_scaled, X_val_scaled, y_train, y_val, imputer, scaler = trainer.apply_scaling_imputation(
+        imputation_strategy, scaling_technique, X_train, X_val, y_train, y_val
+    )
+    assert X_train_scaled is not None
+    assert X_val_scaled is not None
```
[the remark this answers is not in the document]
chrishalcrow: Tested the actual values, but not checked `SMOTE` directly.

src/spikeinterface/curation/tests/test_train_manual_curation.py
chrishalcrow: Yup, added some checks on stuff like this

alejoe91:
@chrishalcrow @anoushkajain let's have a final push on this.

Just reviewing the generated docs, there are some failures:
1. In the [Training a model for automated curation[¶](https://spikeinterface--2918.org.readthedocs.build/en/2918/tutorials/curation/plot_2_train_a_model.html#training-a-model-for-automated-curation)](https://spikeinterface--2918.org.readthedocs.build/en/2918/tutorials/curation/plot_2_train_a_model.html#sphx-glr-tutorials-curation-plot-2-train-a-model-py)
image

2. At the end of [Model Based Curation]()

- `si.auto_label_units` should be `sc.auto_label_units`
image
- Some rendering/formatting issue at the end:
image

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,285 @@
...
+sua_mua_labels = si.auto_label_units(
+    sorting_analyzer,
+    repo_id = "AnoushkaJain3/sua_mua_classifier",
+    trust_model=True,
+)
+
+all_labels = pd.concat([sua_mua_labels, noise_units]).sort_index()
+print(all_labels)
+
+# If you run this without the ``trust_model=True`` parameter, you will receive an error:
```
alejoe91, line 241:
```suggestion
##############################################################################
# If you run this without the ``trust_model=True`` parameter, you will receive an error:
```

examples/tutorials/curation/plot_1_automated_curation.py
```diff
@@ -0,0 +1,285 @@
...
+analyzer_neural = analyzer.remove_units(noise_units.index)
+
+# Apply the sua/mua model
+sua_mua_labels = si.auto_label_units(
+    sorting_analyzer,
+    repo_id = "AnoushkaJain3/sua_mua_classifier",
+    trust_model=True,
+)
+
+all_labels = pd.concat([sua_mua_labels, noise_units]).sort_index()
```
alejoe91, line 238: Are we sure this doesn't mess up the unit ids order of the original analyzer?

chrishalcrow: The indices of the dataframe are the `unit_id`s (ultimately coming from `sorting_analyzer.unit_ids`) So`sua_mua_labels` and `noise_units` have the correct unit_ids as their indices. Then `sort_index` just re-orders the dataframe (https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.sort_index.html). So I think it's all kosher.

alejoe91: Right! I forgot indices are unit ids!

src/spikeinterface/curation/train_manual_curation.py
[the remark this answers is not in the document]
chrishalcrow: Hello - I know understand the internals a bit better! We actually do do the evaluation over a few random test-train splits by default, which is controlled using the `cv` parameter in the `BayesSearchCV` or `RandomizedSearchCV` classes.

alejoe91:
Question for @anoushkajain and @chrishalcrow 

Shouldn't [this](https://huggingface.co/AnoushkaJain3/sua_mua_classifier/blob/main/model_info.json#L8-L9) be "sua"/"mua" and [this](https://huggingface.co/AnoushkaJain3/noise_neural_classifier/blob/main/model_info.json#L8-L9) "noise"/"neuron"?

alejoe91 (approved):

alejoe91:
@chrishalcrow fixed a couple of broken refs in the docs due to the mertics -> metrics correction in the file name.

Merging this! thank you all for working on this! 

It's a great contribution for the community :)

PR #2922, opened 2024-05-29
Update curation module to use `sorting` argument rather than `parent_sorting`

zm711:


zm711:
@DradeAW for discussion!

Fix _kwargs

zm711:
After discussion we are good to do the hard break still need to fix the _kwargs.

alejoe91 (approved):

alejoe91:
@zm711 ready to review?

zm711:
Oops. I opened all my PRs as drafts because the button was on draft mode. Haha. 

Yeah as long as you're okay with a breaking change this is good to go. Sam and Aurelien said they were okay with it, but the alternative would be to make sorting optional and then do a deprecation instead.

alejoe91:
> Oops. I opened all my PRs as drafts because the button was on draft mode. Haha.
> 
> Yeah as long as you're okay with a breaking change this is good to go. Sam and Aurelien said they were okay with it, but the alternative would be to make sorting optional and then do a deprecation instead.

Let's break! 1.0 here we come :)

zm711:
We have conflicts now. I can work on this on Friday unless you want to just do this before than.

alejoe91:
I can fix it ;)

alejoe91 (approved):

alejoe91:
Grazie Zach!

samuelgarcia:
Merci Zach.

PR #2923, opened 2024-05-29
Regularised the API from parent_recording to recording in zero_channel_pad

AnujVader:
https://github.com/SpikeInterface/SpikeInterface-Hackathon-Edinburgh-May24/issues/6

Changed variable name from parent_recording to recording


zm711:
Sam and I discussed about making a breaking change by also switching the `self._kwargs` so we need to think about that. I'll need to look through this and see if you've made that change.

alejoe91 (approved):

alejoe91:
Thanks @AnujVader !

PR #2924, opened 2024-05-29
API: switch `has_scaled` -> `has_scaleable_traces`

zm711:
This is for developer and end-user clarity. `has_scaled` is not clear because what we really have is the ability to scale the traces so this object is `has_scaleable_traces` which clearly states what the boolean is indicating.

+ appropriate deprecations.

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -381,6 +381,11 @@ class BaseRecording(BaseRecordingSnippets):
         bool
             True if the recording has scaled traces, False otherwise
         """
+        warnings.warn(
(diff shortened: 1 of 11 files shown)


src/spikeinterface/core/baserecording.py
```diff
@@ -381,6 +381,11 @@ def has_scaled_traces(self) -> bool:
         bool
             True if the recording has scaled traces, False otherwise
         """
+        warnings.warn(
+            "`has_scaled_traces` is deprecated and will be removed in 0.103.0. Use has_scaled() instead",
```
zm711:
```suggestion
            "`has_scaled_traces` is deprecated and will be removed in 0.103.0. Use has_scaleable_traces() instead",
```

samuelgarcia (approved):

PR #2931, opened 2024-05-30
Add Mac skip for plexon2 Neo test

zm711:
This brings us into line with how Neo does this check which works for both Mac and linux for the wine stuff.


h-mayorquin (approved):
We tested this for mac with @zm711 computer. This works.

alejoe91 (approved):

PR #2933, opened 2024-05-30
WIP: Proposal of format to hold the manual curation information

remi-pr:
We propose a format to contain the  information relative to manual curation.

### DIFF
--- a/src/spikeinterface/curation/curation_format.py
+++ b/src/spikeinterface/curation/curation_format.py
@@ -1,0 +1,34 @@
+from itertools import combinations
+
+
+def validate_curation_dict(curation_dict):
+    """
+    Validate that the curation dictionary given as parameter complies with the format
+
+    Parameters
+    ----------
+    curation_dict : dict
+
+
+    Returns
+    -------
+
+    """
+
+    unit_set = set(curation_dict['unit_ids'])
+    labeled_unit_set = set([lbl['unit_id'] for lbl in curation_dict['manual_labels']])
+    merged_units_set = set(sum(curation_dict['merged_unit_groups'], []))
+    removed_units_set = set(curation_dict['removed_units'])
+    if not labeled_unit_set.issubset(unit_set):
+        raise ValueError("Some labeled units are not in the unit list")
+    if not merged_units_set.issubset(unit_set):
+        raise ValueError("Some merged units are not in the unit list")
+    if not removed_units_set.issubset(unit_set):
+        raise ValueError("Some removed units are not in the unit list")
+    all_merging_groups = [set(group) for group in curation_dict['merged_unit_groups']]
+    for gp_1, gp_2 in combinations(all_merging_groups, 2):
+        if len(gp_1.intersection(gp_2)) != 0:
+            raise ValueError("Some units belong to multiple merge groups")
+    if len(removed_units_set.intersection(merged_units_set)) != 0:
+        raise ValueError("Some units were merged and deleted")
+    return True
--- a/src/spikeinterface/curation/tests/test_curation_format.py
+++ b/src/spikeinterface/curation/tests/test_curation_format.py
@@ -1,0 +1,225 @@
+from spikeinterface.curation.curation_format import validate_curation_dict
+import pytest
+
+
+"""example = {
+    'unit_ids': List[str, int],
+    'labels_definition': {
+        'category_key1':
+        {'name': str,
+         'labels': List[str],
+         'auto_eclusive': bool}
+    },
+    'manual_labels': [
+        {'unit_id': str or int,
+         'label_category_key': str,
+         'label_category_value': list or str
+         }
+    ],
+    'merged_unit_groups': List[List[unit_ids]],  # one cell goes into at most one list
+    'removed_units': List[unit_ids]  # Can not be  in the merged_units
+}
+"""
+
+valid_int = {
+    'unit_ids': [1, 2, 3, 6, 10, 14, 20, 31, 42],
+    'labels_definition': {
+        'quality':
+        {'name': 'quality',
+         'labels': ['good', 'noise', 'MUA', 'artifact'],
+         'auto_eclusive': True},
+        'experimental':
+        {'name': 'experimental',
+         'labels': ['acute', 'chronic', 'headfixed', 'freelymoving'],
+         'auto_eclusive': False}
+    },
+    'manual_labels': [
+        {'unit_id': 1,
+         'label_category_key': 'quality',
+         'label_category_value': 'good'
+         },
+        {'unit_id': 2,
+         'label_category_key': 'quality',
+         'label_category_value': 'noise'
+         },
+        {'unit_id': 2,
+         'label_category_key': 'experimental',
+         'label_category_value': ['chronic', 'headfixed']
+         },
+    ],
+    'merged_unit_groups': [[3, 6], [10, 14, 20]],  # one cell goes into at most one list
+    'removed_units': [31, 42]  # Can not be  in the merged_units
+}
+
+
+valid_str = {
+    'unit_ids': ["u1", "u2", "u3", "u6", "u10", "u14", "u20", "u31", "u42"],
+    'labels_definition': {
+        'quality':
+        {'name': 'quality',
+         'labels': ['good', 'noise', 'MUA', 'artifact'],
+         'auto_eclusive': True},
+        'experimental':
+        {'name': 'experimental',
+         'labels': ['acute', 'chronic', 'headfixed', 'freelymoving'],
+         'auto_eclusive': False}
+    },
+    'manual_labels': [
+        {'unit_id': "u1",
+         'label_category_key': 'quality',
+         'label_category_value': 'good'
+         },
+        {'unit_id': "u2",
+         'label_category_key': 'quality',
+         'label_category_value': 'noise'
+         },
+        {'unit_id': "u2",
+         'label_category_key': 'experimental',
+         'label_category_value': ['chronic', 'headfixed']
+         },
+    ],
+    'merged_unit_groups': [["u3", "u6"], ["u10", "u14", "u20"]],  # one cell goes into at most one list
+    'removed_units': ["u31", "u42"]  # Can not be  in the merged_units
+}
+
+# This is a failure example
+duplicate_merge = {
+    'unit_ids': [1, 2, 3, 6, 10, 14, 20, 31, 42],
+    'labels_definition': {
+        'quality':
+        {'name': 'quality',
+         'labels': ['good', 'noise', 'MUA', 'artifact'],
+         'auto_eclusive': True},
+        'experimental':
+        {'name': 'experimental',
+         'labels': ['acute', 'chronic', 'headfixed', 'freelymoving'],
+         'auto_eclusive': False}
+    },
+    'manual_labels': [
+        {'unit_id': 1,
+         'label_category_key': 'quality',
+         'label_category_value': 'good'
+         },
+        {'unit_id': 2,
+         'label_category_key': 'quality',
+         'label_category_value': 'noise'
+         },
+        {'unit_id': 2,
+         'label_category_key': 'experimental',
+         'label_category_value': ['chronic', 'headfixed']
+         },
+    ],
+    'merged_unit_groups': [[3, 6, 10], [10, 14, 20]],  # one cell goes into at most one list
+    'removed_units': [31, 42]  # Can not be  in the merged_units
+}
+
+
+# This is a failure example
+merged_and_removed = {
+    'unit_ids': [1, 2, 3, 6, 10, 14, 20, 31, 42],
+    'labels_definition': {
+        'quality':
+        {'name': 'quality',
+         'labels': ['good', 'noise', 'MUA', 'artifact'],
+         'auto_eclusive': True},
+        'experimental':
+        {'name': 'experimental',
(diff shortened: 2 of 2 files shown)


magland:
Thanks @remi-pr and @samuelgarcia 

I have some small suggestions. Let's look at this example from your tests

```json
merged_and_removed = {
    "unit_ids": [1, 2, 3, 6, 10, 14, 20, 31, 42],
    "labels_definition": {
        "quality": {"name": "quality", "labels": ["good", "noise", "MUA", "artifact"], "auto_eclusive": True},
        "experimental": {
            "name": "experimental",
            "labels": ["acute", "chronic", "headfixed", "freelymoving"],
            "auto_eclusive": False,
        },
    },
    "manual_labels": [
        {"unit_id": 1, "label_category_key": "quality", "label_category_value": "good"},
        {"unit_id": 2, "label_category_key": "quality", "label_category_value": "noise"},
        {"unit_id": 2, "label_category_key": "experimental", "label_category_value": ["chronic", "headfixed"]},
    ],
    "merged_unit_groups": [[3, 6], [10, 14, 20]],  # one cell goes into at most one list
    "removed_units": [3, 31, 42],  # Can not be  in the merged_units
}
```

labels_definition => label_definitions

auto_eclusive => auto_exclusive

label_definitions.quality.labels => label_definitions.quality.label_options

What exactly does "auto_exclusive" mean?

label_category_key => label_category

label_category_value => labels (make it a list and allow more than one)

Let me know what you think

remi-pr:
Great comments, I will implement quickly.

Just two notes, which are actually just one:

- `auto_exclusive` means that you can only pick one from the list of options . For example a unit can not be 'good' and 'noise'. It could be renamed to just `exclusive`. I would let that up to you.
- _label_category_value => labels (make it a list and allow more than one)_ : This is linked to the auto_exclusive parameter. If only one is allowed -> str if more than one -> list as you can see in the example `"label_category_value": ["chronic", "headfixed"]`. I added a check for that in the validation function.

zm711:
I think I would be with Jeremy here in that `auto_exclusive` isn't super clear. `exclusive` by itself basically says the same thing. I would personally prefer more explicit like `single_label` or `single_label_only`, but I guess `exclusive isn't too bad. 

Also I weakly fall on Jeremy's side for that too. You could have values always be a list and then if there is exclusivity you just validate that the list only contains one thing rather than a string. That way you are only working with the list and not switching between list and string. What is the benefit of your current implementation over doing just the list?

magland:
> * `auto_exclusive` means that you can only pick one from the list of options . For example a unit can not be 'good' and 'noise'. It could be renamed to just `exclusive`. I would let that up to you.

I think I prefer exclusive

> * _label_category_value => labels (make it a list and allow more than one)_ : This is linked to the auto_exclusive parameter. If only one is allowed -> str if more than one -> list as you can see in the example `"label_category_value": ["chronic", "headfixed"]`. I added a check for that in the validation function.

I'm concerned about having two different possible types here. I'm thinking about code that parses this... you'd need a lot of if statements... `if exclusive...` or `if is str`. And converting between exclusive=False and exclusive=True would be a pain. It also makes validation simpler to always have list. Curious about opinion from others.

remi-pr:
I think `exclusive` is a better name you are right. 

labels being a list or a string: We had this conversation with @samuelgarcia and the conclusion was that it is likely that most labels will be exclusive and that sometimes they won't be. Thus by having a string will avoid having labels[0] everywhere it is exclusive. On the other hand we will have to handle the list case separately. I have no strong opinion on this and I am OK with both options.

zm711:
> EDIT: Ooops I just saw reply from @zm711

When someone is agreeing with me I'm always happy to reread the opinions :)

zm711:
Sam and I talked about this more last night and I think his idea was that he would really prefer it to just accept string and make everything mutually exclusive, but that would require a slight reworking on the sotringview side. Then string only would make the most sense.

magland:
I'm a bit concerned about the complexity of this format. Wouldn't this be more straightforward?

```python
... label definitions goes here ...
'labels': [
  {
    'unit_id': '1',
    'manual': {'quality': 'good', 'experimental': 'chronic'}
  },
  ...
]
```

samuelgarcia:
Hi Jeremy,
thanks for the feeback.
I am OK we can simplify we what you propose.
The ultimate stuff to discuss is the value being "list" (always) or scalar or both.
Note that Remy and what for "both" but having list with size one when mode `exclusive` is OK for me if this make your implementation easier.

I will update this PR this week.

samuelgarcia:
@remi-pr @zm711 @magland 
I push some more change with what we decided.
I also did a small doc in the rst : maybe some american eye to check the gramar could be  nice :)

I need to do some more test for converting old format to this new one.

doc/modules/curation.rst
```diff
@@ -41,6 +41,111 @@ The merging and splitting operations are handled by the :py:class:`~spikeinterfa
     # here is the final clean sorting
     clean_sorting = cs.sorting
 
+Manual curation format
+----------------------
+
+SpikeInterface internally support a manual curation format JSON based.
```
zm711, line 47:
```suggestion
SpikeInterface internally supports a JSON-based manual curation format.
```

doc/modules/curation.rst
```diff
@@ -41,6 +41,111 @@ The merging and splitting operations are handled by the :py:class:`~spikeinterfa
...
+When a mnual curation is necessary, modifying in place a dataset is a bad practice.
```
zm711, line 48:
```suggestion
When manual curation is necessary, modifying a dataset in place is a bad practice.
```

doc/modules/curation.rst
```diff
@@ -41,6 +41,111 @@ The merging and splitting operations are handled by the :py:class:`~spikeinterfa
...
+Instead, to keep the reproducibility in the spike sorting piepline, we introduce a manual curation format,
```
zm711, line 49:
```suggestion
Instead, to ensure the reproducibility of the spike sorting pipelines, we have introduced a simple and JSON-based manual curation format.
```

doc/modules/curation.rst
```diff
@@ -41,6 +41,111 @@ The merging and splitting operations are handled by the :py:class:`~spikeinterfa
...
+simple and JSON based. This format defines at the moment : merges + deletions + manual tags.
```
zm711, line 50:
```suggestion
This format defines at the moment : merges + deletions + manual tags.
```

doc/modules/curation.rst
```diff
@@ -41,6 +41,111 @@ The merging and splitting operations are handled by the :py:class:`~spikeinterfa
...
+The simple file can be kept along side the output of a sorter and applied on the result to have a "clean" result.
+
+This format has two part:
+
+  * **definition** with the folowing keys:
+
+    * "format_version" : format specification
+    * "unit_ids" : give the list of unit_ds
```
zm711, line 58:
```suggestion
    * "unit_ids" : the list of unit_ds
```

doc/modules/curation.rst
```diff
@@ -41,6 +41,111 @@ The merging and splitting operations are handled by the :py:class:`~spikeinterfa
...
+    * "label_definitions" : list of label category and possible labels per category.
```
zm711, line 59:
```suggestion
    * "label_definitions" : list of label categories and possible labels per category.
```

doc/modules/curation.rst
```diff
@@ -41,6 +41,111 @@ The merging and splitting operations are handled by the :py:class:`~spikeinterfa
...
+                            Every category can be *exclusive=True* onely one label or *exclusive=False* several labels possible
+
+  * **manual output** curation with the folowing keys:
+
+    * "manual_labels"
+    * "merged_unit_groups"
+    * "removed_units"
+
+Here the description of the format with a simple example:
```
zm711, line 68:
```suggestion
Here is the description of the format with a simple example:
```

zm711 (commented):
Just some comments on the rst to start with.

magland:
@samuelgarcia I think this looks better.

But isn't this redundant in that "quality" is the key and also the name? Same for "putative_type"?

```python
"label_definitions": {
        "quality": {"name": "quality", "label_options": ["good", "noise", "MUA", "artifact"], "exclusive": True},
        "putative_type": {
            "name": "putative_type",
            "label_options": ["excitatory", "inhibitory", "pyramidal", "mitral"],
            "exclusive": False,
        },
    },
```

Maybe "label_definitions" should be a list instead.

samuelgarcia:
The "name" can be be removed, it was more here for title like "Putative cell type".
Sure in our example it is annoyingly redundant.
I propose to remove at the moment.
We choose dict instead instead list because it ensure unique keys.
Is this OK ?

samuelgarcia:
OK done
And I also did a `curation_label_to_dataframe()` to illustrate the exclusive=True/False.

Exclusive=False return boolean columns whereas exclusive=True are unique.

```
   quality excitatory inhibitory pyramidal mitral
1     good      False      False     False  False
2    noise       True      False      True  False
3               False       True     False  False
6               False      False     False  False
10              False      False     False  False
14              False      False     False  False
20              False      False     False  False
31              False      False     False  False
42              False      False     False  False
```

samuelgarcia:
@alejoe91 : this is ready to review on my side.
The implementation of applying the format to a sorting will be done in another PR.
This PR include only machinery for handling/checking/converting the curtaion format.

`apply_sortingview_curation()` still need to be rewrite with this in mind using the convertion.

src/spikeinterface/curation/curation_format.py
```diff
@@ -0,0 +1,169 @@
...
+    Parameters
+    ----------
+    curation_dict : dict
+
+
+    Returns
+    -------
+    Nothing.
+
+
```
alejoe91, line 22:
```suggestion
```

src/spikeinterface/curation/tests/test_curation_format.py
```diff
@@ -0,0 +1,161 @@
...
+
+def test_curation_format_validation():
+    validate_curation_dict(curation_ids_int)
+    validate_curation_dict(curation_ids_str)
+
+    with pytest.raises(ValueError):
+        # Raised because duplicated merged units
+        validate_curation_dict(duplicate_merge)
+    with pytest.raises(ValueError):
+        # Raised because Some units belong to multiple merge groups"
```
alejoe91, line 109:
```suggestion
        # Raised because some units belong to merged and removed unit groups
```

src/spikeinterface/curation/tests/test_curation_format.py
```diff
@@ -0,0 +1,161 @@
...
+        validate_curation_dict(merged_and_removed)
+    with pytest.raises(ValueError):
+        # Some merged units are not in the unit list
+        validate_curation_dict(unknown_merged_unit)
+    with pytest.raises(ValueError):
+        # Raise beecause Some removed units are not in the unit list
```
alejoe91, line 115:
```suggestion
        # Raise because some removed units are not in the unit list
```

src/spikeinterface/curation/tests/test_curation_format.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+def test_convert_from_sortingview_curation_format_v0():
+
+    parent_folder = Path(__file__).parent
+    for filename in (
+        "sv-sorting-curation.json",
+        "sv-sorting-curation-int.json",
+        "sv-sorting-curation-str.json",
+        "sv-sorting-curation-false-positive.json",
```
alejoe91, line 132: Where are these files???

alejoe91 (changes requested):
A few minor comments.

Aren't those 4 files missing?

src/spikeinterface/curation/tests/test_curation_format.py
remi-pr: I think those are from the original curation format (sortingview) so they are there` src/spikeinterface/curation/tests`

alejoe91: Right I forgot about them :)

samuelgarcia:
Merci beacoup Rémi.

remi-pr:
Great job guys. Let me know if you ever want help implementing the curation to a sorting.

alejoe91:
Thanks @remi-pr 

I think that the next level for curation (also automated) would be to add a curation layer to the `SortingAnalyzer`.

We now have the `analyzer.select_units` and we can easily make a `remove_units` to discard rejected units.

The tricky but interesting part would be to implement an `analyzer.merge(merge_unit_ids)`. This function will need to be implemented in each extension to apply/recompute the merge.

For example:
- the `templates` extension could simply recompute the merged template as a weighted sum of the merged unit templates
- the `correlograms` will only recompute correlograms for the merge
- the quality metrics is more tricky: some of them will just need to be recomputed on the merged unit (e.g. all the spike train based), but others (like PCA metrics) will need a recompute on the updated unit neighborhood (e.g., isolation_distance, l-ratio, etc.)

This is a big change, but I think that it's currently the main bottleneck to a clean workflow since now one needs to recompute the analyzer and all extensions after a merge (or a split).

IMO, it makes sense to start with this now since it will be a super useful feature!

yger:
Actually, I have a branch and i'll make a PR soon, to implement such a sorting_analyzer.merge_units() function in a proper manner, similarly to select_units(). Don't start anything, I should push that today hopefully to start the discussion

remi-pr:
Perfect. I am looking forward to this. I am really looking forward to this new feature.

yger:
See #3043

PR #2937, opened 2024-05-30
Add tests for windows and mac

h-mayorquin:
Full testing suit, right now is set as a push for debugging. Will be a cron job per week once we are sure it works.


alejoe91:
Were tests canceled because they took too long?

zm711:
Not a clue. When we were watching them at the hackathon sometimes they would just decide too cancel. Maybe @h-mayorquin knows more? One problem it looks like with Windows is that we are failing at the Neo level. We test core, but not the readers. Maybe I should put that on my to do list on the Neo side...

h-mayorquin:
I do thing that is a too long to download problem. Datalad related I think.

JoeZiminski:
Thanks this is a great idea! I had a look at the neoextractors, it seems to be working for some files but others datalad is failing to download the file completely. It just downloads a 1KB link, the .h5 file with:

`../.git/annex/objects/47/80/MD5E-s62317832--89ebc09e507d9f659836667b08d36cce.h5/MD5E-s62317832--89ebc09e507d9f659836667b08d36cce.h5`

but never actually fills it. Unfortunately, it seems that datalad is [not fully supported](https://handbook.datalad.org/en/latest/intro/windows.html) and strange bugs like this are usual when using on Windows. I've always had a lot of trouble with it as a Windows user.

I'm not certain how best to proceed, in our team people are using [pooch](https://howto.neuroinformatics.dev/open_science/GIN-repositories.html#download-a-gin-dataset-with-python) which has less overhead than datalad and generally works better. I could use it in this case to download the data and make the tests pass. However, switching to `pooch` would probably be quite a big ask.

A temporary workaround could be to introduce `pooch` only when handling downloads for tests, I was planning on refactoring these tests anyway to remove `unittest`. Alternatively, it might be possible to get the datalad download working on windows, but that linked article ([here](https://handbook.datalad.org/en/latest/intro/windows.html)) doesn't give me much hope!

h-mayorquin:
@JoeZiminski the last test indicates that datalad is indeed working on windows. See the latest changes. I need to see why it hangs though. I will test windows now on my machine to see why the sorter test are failing anyway. 

I will give pooch a try though. That maybe will make our life easier.

JoeZiminski:
Thanks @h-mayorquin, I believe it is superficially working but some of the files it downloads are not the real files but only 1KB placeholder files, that for some reason are not properly filled when `dataset.get()` is called. One indication of this is that the tests that are passing [here](https://github.com/SpikeInterface/spikeinterface/actions/runs/9330552919/job/25780919865) are all not using datalad, and the datalad test only actually tests the machinery doesn't crash, but not the validity of the result. 

That been said, _both_ `test_cellexplorerextractor.py` tests and _all_ `test_neoexplorer` tests are failing at the moment on my windows PC where in the CI it is more sproadic. But I have seen cases where some files are downloaded and others not with datalad. So my working hypothesis is that the download is failing sporadically causing those errors, but it's just a guess. It will be great to see what you think when you take a look on the windows PC. I wonder if it possible to see the error messages on the CI? ATM on my machine I get errors related to opening `.h5` files indicating the file is corrupted (which I guess it is because it only contains a string).

h-mayorquin:
> Thanks @h-mayorquin, I believe it is superficially working but some of the files it downloads are not the real files but only 1KB placeholder files, that for some reason are not properly filled when dataset.get() is called. One indication of this is that the tests that are passing [here](https://github.com/SpikeInterface/spikeinterface/actions/runs/9330552919/job/25780919865) are all not using datalad, and the datalad test only actually tests the machinery doesn't crash, but not the validity of the result.

Yes you are correct about this.

I tried implementing a quick sratch with pooch but as I mentioned we have a folder base setup for testing that does not lend well itself to the file_path approach that poch -at least at a first glance- seems to favour. 

We could try pooch if there was a function that could be used to get the contents of the folder without downloading.

zm711:
Philosophically most correctness testing for the neoextractors should be (is I hope) done in our test suite at Neo. The spikeinterface testing for a lot of those files focuses on making sure the wrapper works? I haven't dug into the extractor tests at all? What is the point of them? (the non-neo I totally understand for testing).

h-mayorquin:
There are things like the pickling tests and the correct distribution of attributes and annotations that needs to be tested on spikeinterface side.

For this problem what we ended up doing on neuroconv side is to use s3 buckets to download the data from gin preriodically and download the testing from the buckets.

h-mayorquin:
This is ready to go, let's discuss on the meeting tomorrow how to handle this.

alejoe91 (approved):

src/spikeinterface/extractors/tests/test_datalad_downloading.py
```diff
@@ -1,15 +1,12 @@
...
 
-    HAVE_DATALAD = True
-except:
-    HAVE_DATALAD = False
-
-
-@pytest.mark.skipif(not HAVE_DATALAD, reason="No datalad")
+@pytest.mark.skipif(
+    importlib.util.find_spec("pooch") is None or importlib.util.find_spec("datalad") is None,
+    reason="Etither pooch or datalad is not installed",
```
zm711, line 8:
```suggestion
    reason="Either pooch or datalad is not installed",
```

src/spikeinterface/core/datasets.py
```diff
@@ -14,56 +14,80 @@ def download_dataset(
...
 
     Returns
     -------
     Path
         The local path to the downloaded dataset
+
+    Notes
+    -----
+    The reason we use pooch is because have had problems with datalad not being able to download
+    data on windows machines. Specially in the CI.
```
zm711, line 45:
```suggestion
    data on windows machines. Especially in the CI.
```

zm711 (commented):
Two typo fixes.

alejoe91:
@h-mayorquin this is ready to merge?

h-mayorquin:
Let me actually fix the caches first in another PR.

h-mayorquin:
Ok @alejoe91 the caches are working:

image


This is ready to go. I will add now in another PR the logic for only running tests in modules that change or if core changes.

alejoe91:
Should we squash and merge?

h-mayorquin:
> Should we squash and merge?

Yes, please. I will make them a bit faster with #3084 and once we feel comfortably after a while we can remove the current ubuntu specific workflow I hope.

samuelgarcia:
Thanks a lot for this monstruous effort.

PR #2941, opened 2024-05-30
Template similarity lags

yger:
Add the possibility to specify a max_lag_ms while computing the distance between templates.


alejoe91:
@yger can we take the chance and add the L1 and L2-norms as metrics for similarity? I think this is done elsewhere already in the sortingcomponents

yger:
@alejoe91 This has been added. The next question here (also for @samuelgarcia) is that do we want to rely on such an extension for the curation metrics ? Currently, all the methods ported from lussac are somehow duplicating code to compute distances between templates, maybe we should (if already computed) stick to the metrics of the extension? If so, I can simply add normalized_l1 as a new metric (to replace the one from Lussac)

yger:
All distances can now be used, and the cosine_similarity has been rename to cosine (but could be discussed). Note that various metrics are also used all over the place in the curation/lussac modules. Up to us to decide if we want to harmonize that and to make them available at this extension level. The problem is that:
- one might want to compute several distances (and thus handle several extension of the same type)
- the metric are rather home made and appropriate names need to be found, that will be explicit enough. For example in lussac, l1_normalized need to be explained in the doc. 
Let's discuss if this is an overkill and/or worth developing. These metrics are playing a rather large role for meta merging, so this should not be overlooked.

yger:
This is good to go, and I've just added the metrics used in the curation modules. If this is fine for you, I could now make the curation module depends on these metrics. Note that I think that indeed, the union method is way better than the intersection, currently used in the curation. I've added docs. The only question that remains is that now, all metrics tends to return 0 when templates are similar, except the cosine, where this is 1 when they are similar. Should we consider harmonizing it?

alejoe91:
Thanks @yger !! Yes I think that we should.harmonize towards similarity, which is a more common concept. So equal templates would have a similarity of 1, and orthogonal of 0. For l1/l2 norm, since they are distances, it should be simply 1-dist, right?

yger:
Ok, I'll harmonize toward similarity, ensuring that we have always 1 if template are identical. Then I'll integrate this PR into the one from meta merging and we could start playing with that

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -32,8 +45,8 @@ class ComputeTemplateSimilarity(AnalyzerExtension):
     def __init__(self, sorting_analyzer):
         AnalyzerExtension.__init__(self, sorting_analyzer)
 
-    def _set_params(self, method="cosine_similarity"):
-        params = dict(method=method)
+    def _set_params(self, method="cosine", max_lag_ms=0, support="dense"):
```
samuelgarcia, line 48: do we plan a backcompatible stuff when "cosine_similarity" is given ?

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -9,13 +9,26 @@
...
 
     Parameters
     ----------
-    sorting_analyzer: SortingAnalyzer
+    sorting_analyzer : SortingAnalyzer
         The SortingAnalyzer object
-    method: str, default: "cosine_similarity"
-        The method to compute the similarity
+    method : str, default: "cosine"
+        The method to compute the similarity. Can be in ["l2", "l1", "cosine"]
```
samuelgarcia, line 20: Put cosine first if this is the default

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -60,25 +85,102 @@ def _get_data(self):
...
+                    norms_2 = np.linalg.norm(tgt_templates, ord=2, axis=1)
+                    denominator = norms_1[:, None] + norms_2[None, :]
+                    distances[count] = sklearn.metrics.pairwise.pairwise_distances(
+                        src_templates, tgt_templates, metric="l2"
+                    )
+                    distances[count] /= denominator
+                else:
+                    distances[count] = sklearn.metrics.pairwise.pairwise_distances(
+                        src_templates, tgt_templates, metric=method
+                    )
```
samuelgarcia, line 134: I thik I would prefer explicitly method = "cosine" instead of relying of method from sklearn

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -60,25 +85,102 @@ def _get_data(self):
...
+            units_overlaps = np.sum(mask, axis=2) > 0
+            overlapping_templates = {}
+            for i in range(nb_templates):
+                overlapping_templates[i] = np.flatnonzero(units_overlaps[i])
+
+        # We can use the fact that dist[i,j] at lag t is equal to dist[j,i] at time -t
+        # So the matrix can be computed only for negative lags and be transposed
+
+        for count, shift in enumerate(range(-n_shifts, 1)):
+            if mask is None:
```
samuelgarcia, line 112: We should not duplicate the code twice when mask is None but create a mask that contain a full intersection. No ?

src/spikeinterface/postprocessing/template_similarity.py
alejoe91: yep, let's do it

src/spikeinterface/postprocessing/template_similarity.py
alejoe91: you mean just having: `sklearn.metrics.pairwise.pairwise_distances(..., method="cosine")`?

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -9,13 +9,26 @@
...
     Parameters
     ----------
-    sorting_analyzer: SortingAnalyzer
+    sorting_analyzer : SortingAnalyzer
         The SortingAnalyzer object
-    method: str, default: "cosine_similarity"
-        The method to compute the similarity
+    method : str, default: "cosine"
+        The method to compute the similarity. Can be in ["cosine", "l2", "l1"]
+    max_lag_ms : float, default 0
```
alejoe91, line 21:
```suggestion
    max_lag_ms : float, default: 0
```

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -9,13 +9,26 @@
...
+        If specified, the best distance for all given lag within max_lag_ms is kept, for every template
+    support : str, default "union"
```
alejoe91, line 23:
```suggestion
    support : "dense" | "union" | "intersection", default: "union"
```

src/spikeinterface/postprocessing/template_similarity.py
alejoe91: done

src/spikeinterface/postprocessing/template_similarity.py
alejoe91: done

alejoe91:
@yger @samuelgarcia I moved the support/sparsity logic to the `compute_similarity_with_templates_array` function and extended the template comparison to accept the new arguments (see `compute_template_similarity_by_pair`)

alejoe91 (approved):

samuelgarcia:
let me read this before merging.

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -60,25 +83,109 @@ def _get_data(self):
...
+
     import sklearn.metrics.pairwise
 
     if method == "cosine_similarity":
-        assert templates_array.shape[0] == other_templates_array.shape[0]
-        templates_flat = templates_array.reshape(templates_array.shape[0], -1)
-        other_templates_flat = templates_array.reshape(other_templates_array.shape[0], -1)
-        similarity = sklearn.metrics.pairwise.cosine_similarity(templates_flat, other_templates_flat)
-
+        method = "cosine"
```
samuelgarcia, line 93: maybe a warning ?

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -60,25 +83,109 @@ def _get_data(self):
 compute_template_similarity = ComputeTemplateSimilarity.function_factory()
 
 
-def compute_similarity_with_templates_array(templates_array, other_templates_array, method):
+def compute_similarity_with_templates_array(
+    templates_array, other_templates_array, method, support="union", n_shifts=0, sparsity=None, other_sparsity=None
```
samuelgarcia, line 87: I would use num_shifts here instead of n_shifts for consistency with other places.

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -60,25 +83,109 @@ def _get_data(self):
...
+        for i in range(num_templates):
+            overlapping_templates[i] = np.flatnonzero(units_overlaps[i])
     else:
-        raise ValueError(f"compute_template_similarity(method {method}) not exists")
+        # here we make a dense mask and overlapping templates
+        overlapping_templates = {i: np.arange(other_num_templates) for i in range(num_templates)}
+        mask = np.ones((num_templates, other_num_templates, num_channels), dtype=bool)
+
+    assert n_shifts < num_samples, "max_lag is too large"
+    num_shifts = 2 * n_shifts + 1
```
samuelgarcia, line 131: And here the reverse

src/spikeinterface/postprocessing/tests/test_template_similarity.py
```diff
@@ -8,7 +8,9 @@
 class TestSimilarityExtension(AnalyzerExtensionCommonTestSuite):
 
     def test_extension(self):
-        self.run_extension_tests(ComputeTemplateSimilarity, params=dict(method="cosine_similarity"))
+        self.run_extension_tests(ComputeTemplateSimilarity, params=dict(method="cosine"))
+        self.run_extension_tests(ComputeTemplateSimilarity, params=dict(method="l2"))
+        self.run_extension_tests(ComputeTemplateSimilarity, params=dict(method="l1", max_lag_ms=0.2))
```
samuelgarcia, line 13: You did not adopt the beautifull pytest parametrize ?

alejoe91: you're changing camarade

PR #2947, opened 2024-05-31
Fix marker collection to work on windows

h-mayorquin:
If you look at the diff you will realize that there are unix slahses (e.g. `/`) hardcoded in the conftest to collect the tests for the markers. This PR changes this and also changes the core test to use the markers to test that this works correctly. If it does, the core test should runc orrectly as they are tested for all OSs.

### DIFF
--- a/.github/workflows/core-test.yml
+++ b/.github/workflows/core-test.yml
@@ -31,7 +31,7 @@ jobs:
           pip install -e .[test_core]
       - name: Test core with pytest
         run: |
-          pytest -m "core" -vv -ra --durations=0 --durations-min=0.001 tee report.txt; test ${PIPESTATUS[0]} -eq 0 || exit 1
+          pytest -m "core" -vv -ra --durations=0 --durations-min=0.001 | tee report.txt; test $? -eq 0 || exit 1
         shell: bash  # Necessary for pipeline to work on windows
       - name: Build test summary
         run: |
--- a/.github/workflows/core-test.yml
+++ b/.github/workflows/core-test.yml
@@ -31,7 +31,7 @@ jobs:
           pip install -e .[test_core]
       - name: Test core with pytest
         run: |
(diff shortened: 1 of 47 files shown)


src/spikeinterface/extractors/tests/test_nwbextractors_streaming.py
```diff
@@ -1,54 +1,15 @@
 from pathlib import Path
 import pickle
-from tabnanny import check
```
zm711, line 3: One of the projects should have been removing all the unused imports. Haha.

conftest.py
h-mayorquin, line 1: This is the core of the change.


everything else is imports out of place that get trigerred by pytest imports on the core tests.

Instead of assuming unit slashes on the pathname it assumes that the src/spikeinterface/module is our structure so when writing relative paths to src/spikeinterface the first part of the path is the module.

.github/workflows/core-test.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           pip install -e .[test_core]
       - name: Test core with pytest
         run: |
-          pytest -vv -ra --durations=0 --durations-min=0.001 src/spikeinterface/core | tee report.txt; test ${PIPESTATUS[0]} -eq 0 || exit 1
+          pytest -m "core" -vv -ra --durations=0 --durations-min=0.001 | tee report.txt; test $? -eq 0 || exit 1
```
h-mayorquin, line 34: I changed this as a test that the marker collection works on windows.

This also triggered a lot of imports when only basic dependencies are used which made me change all the imports below.

I think this is good for two reasons:
1) Everyone should be able to run the core tests with the markers even if they don't have all the dependencies.
2) This works good as a test that we are not putting imports on the wrong places.

alejoe91 (approved):

PR #2949, opened 2024-05-31
Make _set_probes private

zm711:
Currently without a deprecation. Do we want a deprecation just to be safe for this one? Again it is not documented but someone could be using it...?

### DIFF
--- a/src/spikeinterface/core/baserecordingsnippets.py
+++ b/src/spikeinterface/core/baserecordingsnippets.py
@@ -101,12 +101,12 @@ class BaseRecordingSnippets(BaseExtractor):
         assert isinstance(probe, Probe), "must give Probe"
         probegroup = ProbeGroup()
         probegroup.add_probe(probe)
-        return self.set_probes(probegroup, group_mode=group_mode, in_place=in_place)
+        return self._set_probes(probegroup, group_mode=group_mode, in_place=in_place)
 
     def set_probegroup(self, probegroup, group_mode="by_probe", in_place=False):
-        return self.set_probes(probegroup, group_mode=group_mode, in_place=in_place)
+        return self._set_probes(probegroup, group_mode=group_mode, in_place=in_place)
 
-    def set_probes(self, probe_or_probegroup, group_mode="by_probe", in_place=False):
+    def _set_probes(self, probe_or_probegroup, group_mode="by_probe", in_place=False):
         """
         Attach a list of Probe objects to a recording.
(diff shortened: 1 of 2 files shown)


src/spikeinterface/extractors/neoextractors/spikegadgets.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, file_path, stream_id=None, stream_name=None, block_index=None
             probegroup = probeinterface.read_spikegadgets(file_path, raise_error=False)
 
         if probegroup is not None:
-            self.set_probes(probegroup, in_place=True)
+            self._set_probes(probegroup, in_place=True)
```
h-mayorquin, line 47: We should change this to use set_probe_group, right?

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -233,6 +233,21 @@ def set_probes(self, probe_or_probegroup, group_mode="by_probe", in_place=False)
 
         return sub_recording
 
+    def set_probes(self, probe_or_probegroup, group_mode="by_probe", in_place=False):
+
+        warning_msg = (
+            "`set_probes` is now a private function and the public function will be "
+            "removed in 0.103.0. Please use `set_probe` or `set_probegroups`"
```
h-mayorquin, line 240:
```suggestion
            "removed in 0.103.0. Please use `set_probe` or `set_probegroups` instead"
```

h-mayorquin (approved):
Looks good to me.

src/spikeinterface/extractors/neoextractors/spikegadgets.py
zm711: Not necessarily. This is a private call to the function so we are entitled to use the private or public api, no?

h-mayorquin: I am just thinking that the code will be more specific.

zm711: I guess I don't know gadgets well enough. We call it a probegroup, but I don't know if it could ever return a probe instead. So set_probes protects us from that case that the variable was named poorly. I guess I should look up the actual read_spikegadgets and see if it is forced to truly be a probegroup.

alejoe91 (approved):

samuelgarcia (approved):

PR #2950, opened 2024-05-31
Standardize colon spacing in docstrings

chrishalcrow:
Standardize colon spacings in docstrings.

All public functions now use the numpydoc colon spacing (https://numpydoc.readthedocs.io/en/latest/format.html#parameters). Meaning that they look like
```
    The recording extractor object to be saved in .dat format
```
rather than
```
recording: RecordingExtractor
    The recording extractor object to be saved in .dat format
```

This means we can use the `numpydoc_validation_checks = {"PR01"}` sphinx parameter to easily check for conflicts between function/class arguments and their documented names.

The format is also consistent with the standard numpydoc autodoc generator (https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring).

### DIFF
--- a/src/spikeinterface/comparison/paircomparisons.py
+++ b/src/spikeinterface/comparison/paircomparisons.py
@@ -664,24 +664,24 @@ class GroundTruthComparison(BasePairSorterComparison):
 
 _template_txt_performance = """PERFORMANCE ({method})
 -----------
-ACCURACY : {accuracy}
-RECALL : {recall}
-PRECISION : {precision}
-FALSE DISCOVERY RATE : {false_discovery_rate}
-MISS RATE : {miss_rate}
+ACCURACY: {accuracy}
+RECALL: {recall}
+PRECISION: {precision}
+FALSE DISCOVERY RATE: {false_discovery_rate}
+MISS RATE: {miss_rate}
 """
 
 _template_summary_part1 = """SUMMARY
 -------
-GT num_units : {num_gt}
-TESTED num_units : {num_tested}
-num_well_detected : {num_well_detected}
-num_redundant : {num_redundant}
-num_overmerged : {num_overmerged}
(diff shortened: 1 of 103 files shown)


zm711 (commented):
Fix the default `:`

src/spikeinterface/comparison/paircomparisons.py
```diff
@@ -664,24 +664,24 @@ def count_units_categories(
...
+GT num_units : {num_gt}
+TESTED num_units : {num_tested}
+num_well_detected : {num_well_detected}
+num_redundant : {num_redundant}
+num_overmerged : {num_overmerged}
 """
 
 _template_summary_part2 = """num_false_positive_units {num_false_positive_units}
-num_bad: {num_bad}
+num_bad : {num_bad}
```
alejoe91, line 684: @chrishalcrow these are for printing the performance, so I think we can revert

alejoe91 (changes requested):
Thanks @chrishalcrow 

Just a little comment!

src/spikeinterface/core/sparsity.py
```diff
@@ -11,29 +11,29 @@
 
 
 _sparsity_doc = """
-    method: str
-        * "best_channels": N best channels with the largest amplitude. Use the "num_channels" argument to specify the
+    method : str
```
zm711, line 14: Since this is an injected docstring does the numpydoc validation still work the same?

alejoe91 (approved):

zm711 (approved):
This looks great to me. Chris could you remember to also change our docstring contributing guide to account for this?

alejoe91:
Thanks @chrishalcrow ! Let's merge :)

samuelgarcia:
Thanks a lot Chris for this super tedious but usefull fight.

PR #2951, opened 2024-05-31
Make easier env vars for local testing

zm711:
As discussed the link between KACHERY and Sortingview is not super clear so this provides an easier to discover env-var for skipping sortingview tests for local development of widgets.

### DIFF
--- a/src/spikeinterface/sorters/external/tests/test_docker_containers.py
+++ b/src/spikeinterface/sorters/external/tests/test_docker_containers.py
@@ -15,7 +15,7 @@ else:
     cache_folder = Path("cache_folder") / "sorters"
 
 
-ON_GITHUB = os.getenv("CI")
+ON_GITHUB = bool(os.getenv("GITHUB_ACTIONS"))
 
 
 def check_gh_settings():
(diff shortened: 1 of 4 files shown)


h-mayorquin:
Can we use pytest markers for this?

zm711:
Only if they are well documented I guess. I like a quick env var for my own local testing. I think we need to document our markers better somewhere. We mention they exist in the docs maybe we should point people to the pyproject.toml for further markers ( I can't remember if we say that or not at the moment).

h-mayorquin:
For you it would be easier to have an environment variable than to have marker "pytest -m "not sorting_view_tests"?

If so, that is good to me. I was asking because I was curios if there were any blockers to just add a marker to those tests.

alejoe91:
The problem is that currently the same test functions loop through several backends, so markers won't work because they skip tests entirely

h-mayorquin:
Good to know.  

Thanks for the info Alessio.

zm711:
That's a really great point Alessio. I hadn't even thought of that.

h-mayorquin (approved):
I see very little cost to this and it improves Zach worflow which is very imporant for us. I say good!

alejoe91 (approved):

PR #2952, opened 2024-05-31
Add extractors for SiNAPS Research Platform

NinelK:
Contains extractors from binary format (raw + filtered + aux channels) as well as an additional extractor from *.h5 format (only filtered channels)


alejoe91:
@NinelK can you allow edits from maintainers? The pre-commit bot failed to refactor the code according to black

NinelK:
> @NinelK can you allow edits from maintainers? The pre-commit bot failed to refactor the code according to black

it was allowed in the settings on the side, not sure why pre-comit.ci can not push.
I ran black locally, and it seems like this resolves the issue.

src/spikeinterface/extractors/sinapsrecordingh5extractor.py
```diff
@@ -0,0 +1,149 @@
...
+import numpy as np
+
+from probeinterface import get_probe
+
+from ..core.core_tools import define_function_from_class
+from ..core import BaseRecording, BaseRecordingSegment
+from ..preprocessing import UnsignedToSignedRecording
+
+
+class SinapsResearchPlatformH5RecordingExtractor_Unsigned(BaseRecording):
```
alejoe91, line 11: Maybe we can just apply the unsigned_to_signed in the get_traces, which is simpler to read and will get rid of the complex class interplay. I can implement that

alejoe91: done

alejoe91:
@NinelK after discussing with @samuelgarcia, we decided that in this case it's worth adding the unsigned to signed scaling directly to the SinapsH5Extractors (avoiding that class trick, which makes it less readable).

Also moved everything to a single file and cleaned docs and helper functions. Ready to merge on my end

alejoe91 (approved):

alejoe91:
@NinelK can you test it one more time on your side? I don't have h5 test data

NinelK:
All works now, just fixed 1 issue with a missing argument

PR #2955, opened 2024-06-01
Improve profile imports test

h-mayorquin:
I am adding two things here:
* An extra line that displays all the sampled times to calculate the average import times. This will be useful for discovering if there is any cache issue affecting the first import in a new machine.
* An actual assertion if the import times goes above 2.5 seconds.


h-mayorquin:
Ah, apparently exporters takes too long to import:

https://github.com/SpikeInterface/spikeinterface/actions/runs/9329949107/job/25682977784?pr=2955

Will open an issue once I confirm and delay adding an import test.

h-mayorquin:
Yes, on windows and mac importing exporters for the full time is very costly:

image


This is ready, we can fix that other issue later and enable upper bound on import times when we have that.

.github/workflows/test_imports.yml
```diff
@@ -38,8 +38,7 @@ jobs:
         shell: bash  # Necessary for pipeline to work on windows
       - name: Install in full mode
         run: |
-          python -m pip install -U pip  # Official recommended way
-          pip install -e .[full]
+          pip install .[full]
```
zm711, line 41: what's the logic for removing -e?

h-mayorquin: There is no reason to install editable here. Plus, the setup tools editable install is in a transition phase right now and I don't want to partake on that.

https://setuptools.pypa.io/en/latest/userguide/development_mode.html

zm711: I see. Cool thanks.

alejoe91 (approved):

samuelgarcia (approved):

samuelgarcia:
Merci.

PR #2969, opened 2024-06-04
Add argument to `IntanRecordingExtractor`  for opening files with discontinous timestamps

h-mayorquin:
Following on the corresponding neo work:

https://github.com/NeuralEnsemble/python-neo/pull/1470

Comes with a check so the argument is only propagated for the upcoming version of neo and does not break current version.

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/intan.py
+++ b/src/spikeinterface/extractors/neoextractors/intan.py
@@ -52,7 +52,15 @@ class IntanRecordingExtractor(NeoBaseRecordingExtractor):
 
     @classmethod
     def map_to_neo_kwargs(cls, file_path, ignore_integrity_checks: bool = False):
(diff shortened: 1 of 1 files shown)


src/spikeinterface/extractors/neoextractors/intan.py
```diff
@@ -23,14 +23,26 @@ class IntanRecordingExtractor(NeoBaseRecordingExtractor):
         If there are several streams, specify the stream name you want to load.
     all_annotations: bool, default: False
         Load exhaustively all annotations from neo.
+    ignore_integrity_checks, bool, default: False.
```
alejoe91, line 26:
```suggestion
    ignore_integrity_checks : bool, default: False.
```

samuelgarcia (approved):

PR #2970, opened 2024-06-04
Fix bug with nan values

DradeAW:
When the first value was a `nan` (because of no value), we ended up with the wrong dtype (e.g. should have been `str`)

### DIFF
--- a/src/spikeinterface/extractors/phykilosortextractors.py
+++ b/src/spikeinterface/extractors/phykilosortextractors.py
@@ -173,9 +173,11 @@ class BasePhyKilosortSortingExtractor(BaseSorting):
                     # pandas loads strings as objects
                     if cluster_info[prop_name].values.dtype.kind == "O":
                         for value in cluster_info[prop_name].values:
-                            if isinstance(value, (np.floating, float)) and np.isnan(value):  # Blank values are encoded as 'NaN'.
+                            if isinstance(value, (np.floating, float)) and np.isnan(
+                                value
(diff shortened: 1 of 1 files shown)


src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -172,7 +172,14 @@ def __init__(
...
                     if cluster_info[prop_name].values.dtype.kind == "O":
-                        prop_dtype = type(cluster_info[prop_name].values[0])
+                        for value in cluster_info[prop_name].values:
+                            if isinstance(value, (np.floating, float)) and np.isnan(
+                                value
+                            ):  # Blank values are encoded as 'NaN'.
+                                continue
+
+                            prop_dtype = type(value)
+                            break
```
alejoe91, line 182: what if they're all NaN?

DradeAW: Then the dtype wouldn't be object, but float?
So you never reach this code (I think)

zm711: I think I'm being dense. Why do you need both checks in the if? Why not just check if `isnan`? Or flip the order of the if so that if it is not nan we short-circuit? Small point but I'm just trying to figure out why we need both.

zm711: Perfect thanks!

alejoe91 (approved):

samuelgarcia (approved):

PR #2927, opened 2024-05-29
Remove cache folder

paulrignanese:
removed the setting of the cache folder as global variable in test_phase_shift in /preprocessing/tests, to see if it survives the CI tests

PR #2971, opened 2024-06-04
Fix deepinterpolation tests

alejoe91:
Updated deepinterpolation code to use v0.2.0

Should fix failing tests in https://github.com/SpikeInterface/spikeinterface/pull/2927

### DIFF
--- a/src/spikeinterface/preprocessing/deepinterpolation/generators.py
+++ b/src/spikeinterface/preprocessing/deepinterpolation/generators.py
@@ -1,15 +1,8 @@
 from __future__ import annotations
-import tempfile
-import json
 from typing import Optional
 import numpy as np
-import os
 
-from ...core import load_extractor, concatenate_recordings, BaseRecording, BaseRecordingSegment
-
-from .tf_utils import has_tf, import_tf
-
-from ...core import load_extractor
+from ...core import concatenate_recordings, BaseRecording, BaseRecordingSegment
 
 from deepinterpolation.generator_collection import SequentialGenerator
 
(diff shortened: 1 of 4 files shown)


src/spikeinterface/preprocessing/deepinterpolation/generators.py
```diff
@@ -41,16 +34,16 @@ def __init__(
                 r.get_num_channels() == recordings[0].get_num_channels() for r in recordings[1:]
             ), "All recordings must have the same number of channels"
 
-        total_num_samples = np.sum(r.get_total_samples() for r in recordings)
+        total_num_samples = np.sum([r.get_total_samples() for r in recordings])
```
zm711, line 37: Is there any chance this could overflow with numpy operations?  total sample number can get to be pretty big.

Same comment for the other instances.

zm711: This is good to know though. I think the issue is data gets bigger and bigger so it's more and more likely to have overflows. I think we should be safe too. These are hard to debug when they go wrong.

alejoe91: @h-mayorquin @zm711 so what do you recommend to do? I think you have a better overview of this problem than me :)

h-mayorquin, line 37:
```suggestion
        total_num_samples = sum([r.get_total_samples() for r in recordings])
```

Sorry, too much talking and too few suggestions : P

I think this is the easiest!

samuelgarcia (approved):

PR #2973, opened 2024-06-04
Add test to check unit structure in quality metric calculator output

chrishalcrow:
When editing quality metric calculators it's easy to get your `unit_id`s mixed up: this PR adds a test which checks that your `unit_ids` are labelled as expected.

This test is mostly important for keeping calm when editing quality metric calculators. It's especially important because there's a lot of potential speed-up to be had (and potential for things to go wrong) by replacing `for` loops over units with numpy slicing. The test checks that the ouputted dictionaries are using the `unit_ids` and not the unit indices.

Had to add a new sorting analyzer with more "interesting" unit_ids. I've made it very short and am planning to port the other tests over to it (unless the long recording is needed), which will speed up testing.

Adding the test revealed something you could almost call a bug: `compute_refrac_period_violations` returned the unit_ids in the order of the sorting object, rather than the order they were given in the `unit_id` argument. There are two reasons this doesn't matter:
 - the output is a dictionary, so the order almost never matters.
 - the `compute` method applied to an extension always uses all non-empty units, and can't accept a user-inputted `unit_ids`, so their order is always the same as the sorting object when used in this way.

This type of "bug" is not what the test is designed to look for - just a little surprise when it didn't immediately pass!

### DIFF
--- a/src/spikeinterface/qualitymetrics/misc_metrics.py
+++ b/src/spikeinterface/qualitymetrics/misc_metrics.py
@@ -1085,10 +1085,10 @@ def compute_drift_metrics(
             spikes_in_bin = spikes_in_segment[i0:i1]
             spike_locations_in_bin = spike_locations_in_segment[i0:i1][direction]
 
-            for unit_index, unit_id in enumerate(unit_ids):
+            for i, unit_id in enumerate(unit_ids):
                 mask = spikes_in_bin["unit_index"] == sorting.id_to_index(unit_id)
                 if np.sum(mask) >= min_spikes_per_interval:
-                    median_positions[unit_index, bin_index] = np.median(spike_locations_in_bin[mask])
+                    median_positions[i, bin_index] = np.median(spike_locations_in_bin[mask])
         if median_position_segments is None:
             median_position_segments = median_positions
         else:


zm711:
A few quality metrics rely on having enough spikes and some rely onbin size, so I think you'd just have to make sure that the recording being used is long enough for those tests to not fail.

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -388,9 +388,7 @@ def compute_refrac_period_violations(
     nb_violations = {}
     rp_contamination = {}
 
-    for i, unit_id in enumerate(sorting.unit_ids):
-        if unit_id not in unit_ids:
-            continue
+    for i, unit_id in enumerate(unit_ids):
```
samuelgarcia, line 391: I think that this is dangerous because vector acces will be wrng no ?

chrishalcrow: We need to make sure that the `i` used in `nb_rp_violations`, which is a numpy array, is synced with the `unit_id`. I think it is. 

I've added another test which takes two almost-identical `sorting_analyzer`s. Their only difference is that their units are in a different order, with different labelings. The test checks that the results are independent of the ordering and labelling. So the new test should catch if anything like this is wrong in any of the quality metrics (it's very easy to mess up!).

alejoe91: @chrishalcrow you see [here](https://github.com/SpikeInterface/spikeinterface/blob/11ae9aa6227e01fd19908d510d142651b8da1c1f/src/spikeinterface/qualitymetrics/misc_metrics.py#L379) that the `nb_rp_violations` have len `num_units`, which is `len(sorting_analyzer.unit_ids)`.

So this is worng. Either we update `num_units` as `len(unit_ids)` or we revert this to `enumerate(sorting.unit_ids)`

alejoe91, line 391: let's use `unit_index` instead of `i`

chrishalcrow: Super. I'm improving the tests so that they will catch this and will then update. While doing so I've found that the `drift` metrics don't handle passed sub-units correctly.

src/spikeinterface/qualitymetrics/tests/test_metrics_functions.py
```diff
@@ -47,6 +52,108 @@
...
+    Takes two almost-identical sorting_analyzers, whose unit_ids are in different orders and have different labels,
+    and checks that their calculated quality metrics are independent of the ordering and labelling.
+    """
+
+    recording, sorting = generate_ground_truth_recording(
+        durations=[2.0],
+        num_units=4,
+        seed=1205,
+    )
+    sorting = sorting.select_units([0, 2, 3])
```
alejoe91, line 118: Since the goal is to make a second analyzer with re-sorted units, why not:
```suggestion
    recording = small_sorting_analyzer.recording
    sorting = small_sorting_analyzer.sorting.select_units([0, 2, 3])
```

chrishalcrow: Oh yes, much better!

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -1085,10 +1085,10 @@ def compute_drift_metrics(
             spikes_in_bin = spikes_in_segment[i0:i1]
             spike_locations_in_bin = spike_locations_in_segment[i0:i1][direction]
 
-            for unit_ind in np.arange(len(unit_ids)):
-                mask = spikes_in_bin["unit_index"] == unit_ind
+            for unit_index, unit_id in enumerate(unit_ids):
+                mask = spikes_in_bin["unit_index"] == sorting.id_to_index(unit_id)
```
alejoe91, line 1089: good catch!!!

alejoe91 (approved):

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -1085,10 +1085,10 @@ def compute_drift_metrics(
             spikes_in_bin = spikes_in_segment[i0:i1]
             spike_locations_in_bin = spike_locations_in_segment[i0:i1][direction]
 
-            for unit_ind in np.arange(len(unit_ids)):
-                mask = spikes_in_bin["unit_index"] == unit_ind
+            for unit_index, unit_id in enumerate(unit_ids):
```
samuelgarcia, line 1088: If unit_ids is not the full list the `unit_index` variable is nt a good choici of name and explain the previous mistake.
I would use `i` or `ind`

src/spikeinterface/qualitymetrics/misc_metrics.py
alejoe91, line 1089:
```suggestion
            for i, unit_id in enumerate(unit_ids):
                mask = spikes_in_bin["unit_index"] == sorting.id_to_index(unit_id)
```

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -1085,10 +1085,10 @@ def compute_drift_metrics(
...
+                mask = spikes_in_bin["unit_index"] == sorting.id_to_index(unit_id)
                 if np.sum(mask) >= min_spikes_per_interval:
-                    median_positions[unit_ind, bin_index] = np.median(spike_locations_in_bin[mask])
+                    median_positions[unit_index, bin_index] = np.median(spike_locations_in_bin[mask])
```
alejoe91, line 1091:
```suggestion
                    median_positions[i, bin_index] = np.median(spike_locations_in_bin[mask])
```

samuelgarcia:
Hi Chris thanks for the catch.

I have the feeling that this kinds of mistake could be at several places in metrics when unit_ids is not None in the signature.
because `for unit_index, unit_id enurmate(unit_ids)` is wrong when unit_ids is not the full list.

We propose to merge this now. But could you make an even stronger tests using unit_ids not None and also check this everywhere enumerate is not wrong in the code ?

PR #2976, opened 2024-06-05
Fix amplitude and location multisegment masks

### LINKED ISSUE #2975: Spike amplitude data for multi-segments is incorrect
When you `get_data` by unit for amplitudes, the sampling is incorrect for multi-segment recordings.

Consider the following code:
```

recording, sorting = si.generate_ground_truth_recording(durations=[10,10], seed=1205)

analyzer = si.create_sorting_analyzer(recording=recording, sorting=sorting)
analyzer.compute(['random_spikes', 'templates', 'spike_amplitudes'])

si.plot_amplitudes(analyzer, segment_index=0)
si.plot_amplitudes(analyzer, segment_index=1)
```
The output for the first plot is
Screenshot 2024-06-05 at 10 29 02
This looks good: each unit has amplitudes based around a common mean.

The second plot looks like this
Screenshot 2024-06-05 at 10 29 07
This doesn't look good. The amplitudes for each unit are scattered randomly.

Issue is in `_get_data` function for `ComputeSpikeAmplitudes`. I'm working on a fix.

chrishalcrow:
Fixes #2975

When retrieving the data from the amplitudes and locations extension, using `get_data`,`all_amplitudes` and `all_locations` were flattened (with respect to segments) while the masks used to extract the amplitudes and locations were not. Hence the extensions will retrieving incorrect data. Another way to say it: the masks to select the amplitudes would have worked on `all_amplitude` data which shaped into (# num segments) x (# num spikes).

@samuelgarcia these functions originally went through `spike_vector_to_indices`, so that NUMBA could be used. Is there a good reason why these must be very fast? I think the new implementation will be much faster than the old non-NUMBA method, but not sure how it compares to NUMBA. Can benchmark if needed. 

I need to add tests, but want some feedback on the question above.

Note: `spike_vector_to_indices` is now only used in `get_random_spikes` and is about 100 lines of code. Could be tempted to get rid of it

### DIFF
--- a/src/spikeinterface/postprocessing/spike_amplitudes.py
+++ b/src/spikeinterface/postprocessing/spike_amplitudes.py
@@ -9,7 +9,6 @@ from spikeinterface.core.template_tools import get_template_extremum_channel, ge
 
 from spikeinterface.core.sortinganalyzer import register_result_extension, AnalyzerExtension
 from spikeinterface.core.node_pipeline import SpikeRetriever, PipelineNode, run_node_pipeline, find_parent_of_type
-from spikeinterface.core.sorting_tools import spike_vector_to_indices
 
 
 class ComputeSpikeAmplitudes(AnalyzerExtension):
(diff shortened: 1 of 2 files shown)


chrishalcrow:
> We used to have a lot of this boolean mask implementtation everywhere. But this is very very slow for high number of units and spikes. You can check. You need to go through all spikes for all units. The numba story is a preindexing per units which go to spikes only 2 times.
> 
> Lets discuss later but this is very very important detail.

I see. I think we can get as fast as numba (and keep things simple) by using some in-built numpy functions. I'll have a go!

samuelgarcia:
Oups no. This could return_inverse sorry.

chrishalcrow:
Had a lot of fun, and got something working where I made an array of masks for each unit, then applied an np.argwhere. Was much faster, but for more units NUMBA absolutely dominated it. Now realised there's a two line solution, arriving in a new PR in a minute...

PR #2977, opened 2024-06-05
Fixed `select_channels`

DradeAW:


src/spikeinterface/core/baserecording.py
```diff
@@ -605,7 +605,7 @@ def _extra_metadata_to_folder(self, folder):
             if time_vector is not None:
                 np.save(folder / f"times_cached_seg{segment_index}.npy", time_vector)
 
-    def select_channels(self, channel_ids: list | np.array | tuple) -> "BaseRecording":
+    def _select_channels(self, channel_ids: list | np.array | tuple) -> "BaseRecording":
```
h-mayorquin, line 608:
```suggestion
    def select_channels(self, channel_ids: list | np.array | tuple) -> "BaseRecording":
```

What do you think on leaving this like this to avoid the navigation problem of clicking an recording object and being sent to snippets?

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -72,6 +72,9 @@ def is_filtered(self):
         # the is_filtered is handle with annotation
         return self._annotations.get("is_filtered", False)
 
+    def _select_channels(self, channel_ids: list | np.array | tuple) -> "BaseRecordingSnippets":
```
h-mayorquin, line 75: Question: why do we need this?

h-mayorquin (commented):
I was not awared that there is a basesnippets which inhertis from baserecordingsnippets so I have some questions.

src/spikeinterface/core/baserecording.py
DradeAW: I think it's better to have a function and overwrite it using inheritance,
I just followed the examples from other functions because I know that @samuelgarcia doesn't like inheritance and abstract classes :)

src/spikeinterface/core/baserecordingsnippets.py
DradeAW: Same as above, ask Sam!!

src/spikeinterface/core/baserecording.py
h-mayorquin: OK, out of the scope then. I agree with you. I always wondered why we don't use abstact classes instead of this template pattern with private functions. Let's discuss this as a separate issue.

h-mayorquin (approved):
Thanks for the fix.

alejoe91 (approved):

samuelgarcia:
@h-mayorquin : I think I agree that the "_" is weird we will refactor it later.

PR #2980, opened 2024-06-05
get_spike_prototype can have NaN sometimes

yger:
The get_prototype function can sometimes trigger warning of Nan. This is because we do not control exactly that the peaks times used to gather the waveforms are within appropriate ranges (wrt nbefore/nafter). We could check, but maybe this is an overkill. At least here we avoid the nbefore case (easy if we have only one segment), and try to catch the warnings


src/spikeinterface/sortingcomponents/peak_selection.py
```diff
@@ -66,8 +71,24 @@ def select_peaks(peaks, method="uniform", seed=None, return_indices=False, **met
         return_indices is True.
     """
 
+    if margin is not None:
+        assert recording is not None, "recording should be provided if margin is not None"
+
     selected_indices = select_peak_indices(peaks, method=method, seed=seed, **method_kwargs)
     selected_peaks = peaks[selected_indices]
+
+    if margin is not None:
```
samuelgarcia, line 80: I have the intuition that we could make a simple while loop directly on selectied_indices to avoid this full spikes vector masking twice and then boolean indexing again.
No ?

PR #2981, opened 2024-06-05
Correct segment counting in SpikeAmpltiudes and SpikeLocations get_data

### LINKED ISSUE #2975: Spike amplitude data for multi-segments is incorrect

chrishalcrow:
Fixes #2975, without losing speed.

The `all_amplitudes` and `all_locations` vectors are flattened, so don't take account of segment index. This PR adds a cumulative sum which keeps track of where each segments begins.

In draft until tests are added.

### DIFF
--- a/src/spikeinterface/postprocessing/spike_amplitudes.py
+++ b/src/spikeinterface/postprocessing/spike_amplitudes.py
@@ -127,13 +127,20 @@ class ComputeSpikeAmplitudes(AnalyzerExtension):
         elif outputs == "by_unit":
             unit_ids = self.sorting_analyzer.unit_ids
             spike_vector = self.sorting_analyzer.sorting.to_spike_vector(concatenated=False)
+            segment_cum_count = np.insert(
+                np.cumsum([len(spike_vector[a]) for a in range(0, self.sorting_analyzer.sorting.get_num_segments())]),
+                0,
+                0,
+            )
             spike_indices = spike_vector_to_indices(spike_vector, unit_ids)
             amplitudes_by_units = {}
             for segment_index in range(self.sorting_analyzer.sorting.get_num_segments()):
                 amplitudes_by_units[segment_index] = {}
                 for unit_id in unit_ids:
                     inds = spike_indices[segment_index][unit_id]
-                    amplitudes_by_units[segment_index][unit_id] = all_amplitudes[inds]
+                    amplitudes_by_units[segment_index][unit_id] = all_amplitudes[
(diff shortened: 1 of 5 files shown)


chrishalcrow:
Note: need to update tests to use fixtures after #2984 is merged.

src/spikeinterface/postprocessing/spike_amplitudes.py
```diff
@@ -127,13 +127,20 @@ def _get_data(self, outputs="numpy"):
         elif outputs == "by_unit":
             unit_ids = self.sorting_analyzer.unit_ids
             spike_vector = self.sorting_analyzer.sorting.to_spike_vector(concatenated=False)
+            segment_cum_count = np.insert(
+                np.cumsum([len(spike_vector[a]) for a in range(0, self.sorting_analyzer.sorting.get_num_segments())]),
+                0,
+                0,
+            )
```
alejoe91, line 134: @chrishalcrow I think this should be handled in the `spike_vector_to_indices`, so it's used by all extensions that do this `by_unit`

alejoe91: @samuelgarcia will fix it because it's his fault!

samuelgarcia:
Hi Chris.
I did another proposal instead of this fix see #3048.
The  spike_vector_to_indices() function was ambiguous with relative or absolut. Now I added an option.
We need the two cases depending on the situation.

alejoe91:
This will be replaced by #3048 

@chrishalcrow can we close?

chrishalcrow:
Yes, closing now

PR #2984, opened 2024-06-05
Remove unittest from postprocessing

JoeZiminski:
This PR removes `unittest` from the tests in the `postprocessing` module. Overall it:

1) Removes `unittest` and changes the `setUpClass` (the running of which was handlded by `unittest`) to a pytest fixture. This still is setup only once-per-subclass instantiation and the run-time of the `postprocessing` tests is within a few seconds of the old runtime.
2) Slightly reworks how `test_extension` is called. Now each subclass has its own `test_extension` function that calls the base class `run_extension_tests`, and the parameters are parameterised which means `pytest` runs 1 test per parameter set. This gives a bit more granulatiry to the test output.
3) In a couple of places add `sparsity` as a test parameterization rather than a loop, for the benefit for separating the tests in the output.
3) This now breaks the existing `if __name__ == "__main__"` blocks which can no longer be run with python. However, all the equivilent functionality can be gained through making plots / interacting with tests from inside test functions, so I removed for now. But please let me know if this will mess with your workflows.
4) In some small places, extended tests by checking for direct equivilency with an expected value rather than over / under an expected bounds.
5) Extended docstrings, removed a few stray `print()` functions.
6) I removed some (but not all) of the commented code that was used for debugging. In this case I tried to remove only the simple cases that I roughly thought would be quicker to rewrite on the fly than look through the commented code and figure out what it is doing. But let me know if it's a problem!

### DIFF
--- a/src/spikeinterface/postprocessing/tests/common_extension_tests.py
+++ b/src/spikeinterface/postprocessing/tests/common_extension_tests.py
@@ -136,7 +136,7 @@ class AnalyzerExtensionCommonTestSuite:
 
         ext = sorting_analyzer.compute(extension_class.extension_name, **params, **job_kwargs)
         assert len(ext.data) > 0
-        main_data = ext.get_data()
+        assert len(main_data) > 0
 
         ext = sorting_analyzer.get_extension(extension_class.extension_name)
         assert ext is not None
--- a/src/spikeinterface/postprocessing/tests/common_extension_tests.py
+++ b/src/spikeinterface/postprocessing/tests/common_extension_tests.py
@@ -73,11 +73,24 @@ class AnalyzerExtensionCommonTestSuite:
     extension_class = None
     extension_function_params_list = None
 
-    @classmethod
-    def setUpClass(cls):
-        cls.recording, cls.sorting = get_dataset()
+    @pytest.fixture(autouse=True, scope="class")
+    def setUpClass(self):
+        """
+        This method sets up the class once at the start of testing. It is
+        in scope for the lifetime of te class and is reused across all
+        tests that inherit from this base class to save processing time and
+        force a small radius.
+
+        When setting attributes on `self` in `scope="class"` a new
+        class instance is used for each. In this case, we have to set
+        from the base object `__class__` to ensure the attributes
+        are available to all subclass instances.
+        """
+        self.__class__.recording, self.__class__.sorting = get_dataset()
         # sparsity is computed once for all cases to save processing time and force a small radius
-        cls.sparsity = estimate_sparsity(cls.recording, cls.sorting, method="radius", radius_um=20)
+        self.__class__.sparsity = estimate_sparsity(
+            self.__class__.recording, self.__class__.sorting, method="radius", radius_um=20
+        )
 
     @property
     def extension_name(self):
@@ -114,12 +127,10 @@ class AnalyzerExtensionCommonTestSuite:
         some_unit_ids = sorting_analyzer.unit_ids[::2]
         sliced = sorting_analyzer.select_units(some_unit_ids, format="memory")
         assert np.array_equal(sliced.unit_ids, sorting_analyzer.unit_ids[::2])
-        # print(sliced)
 
     def test_extension(self):
         for sparse in (True, False):
             for format in ("memory", "binary_folder", "zarr"):
-                print()
                 print("sparse", sparse, format)
                 sorting_analyzer = self._prepare_sorting_analyzer(format, sparse)
                 self._check_one(sorting_analyzer)
--- a/src/spikeinterface/postprocessing/tests/common_extension_tests.py
+++ b/src/spikeinterface/postprocessing/tests/common_extension_tests.py
@@ -61,13 +61,27 @@ def get_sorting_analyzer(recording, sorting, format="memory", sparsity=None, nam
 
 class AnalyzerExtensionCommonTestSuite:
     """
-    Common tests with class approach to compute extension on several cases (3 format x 2 sparsity)
+    Common tests with class approach to compute extension on several cases,
+    format ("memory", "binary_folder", "zarr") and sparsity (True, False).
+    Extensions refer to the extension classes that handle the postprocessing,
+    for example extracting principal components or amplitude scalings.
 
-    This is done a a list of differents parameters (extension_function_params_list).
+    This base class provides a fixture which sets a recording
+    and sorting object onto itself, which are set up once each time
+    the base class is subclassed in a test environment. The recording
+    and sorting object are used in the creation of the `sorting_analyzer`
+    object used to run postprocessing routines.
 
-    This automatically precompute extension dependencies with default params before running computation.
+    When subclassed, a test function that parametrises arguments
+    that are passed to the `sorting_analyzer.compute()` can be setup.
+    This must call `run_extension_tests()`  which sets up a `sorting_analyzer`
+    with the relevant format and sparsity. This also automatically precomputes
+    extension dependencies with default params, Then, `check_one()` is called
+    which runs the compute function with the passed params and tests that:
 
-    This also test the select_units() ability.
+    1) the returned extractor object has data on it
+    2) check `sorting_analyzer.get_extension()` does not return None
+    3) the correct units are sliced with the `select_units()` function.
     """
 
     @pytest.fixture(autouse=True, scope="class")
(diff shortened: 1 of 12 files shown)


src/spikeinterface/postprocessing/tests/common_extension_tests.py
```diff
@@ -61,65 +61,98 @@ def get_sorting_analyzer(recording, sorting, format="memory", sparsity=None, nam
...
             job_kwargs = dict()
 
-        for params in self.extension_function_params_list:
-            print("  params", params)
-            ext = sorting_analyzer.compute(self.extension_name, **params, **job_kwargs)
-            assert len(ext.data) > 0
-            main_data = ext.get_data()
+        ext = sorting_analyzer.compute(extension_class.extension_name, **params, **job_kwargs)
+        assert len(ext.data) > 0
+        main_data = ext.get_data()
```
chrishalcrow, line 139: Could add a test to check that main_data is non-empty? At the moment we do nothing with it.
```suggestion
        assert len(main_data) > 0
```

alejoe91:
@JoeZiminski I'll fix conflicts here since they're my fault ;)

src/spikeinterface/postprocessing/tests/test_correlograms.py
```diff
@@ -14,32 +12,39 @@
...
-    if HAVE_NUMBA:
-        extension_function_params_list.append(dict(method="numba"))
+class TestComputeCorrelograms(AnalyzerExtensionCommonTestSuite):
+
+    @pytest.mark.parametrize(
+        "params",
+        [
+            dict(method="numpy"),
+            dict(method="auto"),
+            pytest.param(dict(method="numba"), marks=pytest.mark.skipif("not HAVE_NUMBA")),
```
chrishalcrow, line 25:
```suggestion
            pytest.param(dict(method="numba"), marks=pytest.mark.skipif(not HAVE_NUMBA, reason="Numba not available")),
```

I hadn't seen condition strings before, so looked them up. Pytest very mildly advises against their use (https://docs.pytest.org/en/7.1.x/historical-notes.html#conditions-as-strings-instead-of-booleans).

src/spikeinterface/postprocessing/tests/test_isi.py
```diff
@@ -16,38 +15,40 @@
...
-    if HAVE_NUMBA:
-        extension_function_params_list.append(dict(method="numba"))
+class TestComputeISIHistograms(AnalyzerExtensionCommonTestSuite):
+
+    @pytest.mark.parametrize(
+        "params",
+        [
+            dict(method="numpy"),
+            dict(method="auto"),
+            pytest.param(dict(method="numba"), marks=pytest.mark.skipif("not HAVE_NUMBA")),
```
chrishalcrow, line 25: Same comment as above.

alejoe91:
@JoeZiminski removed the last traces of cache folders in favor of the module level `create_cache_folder` fixture

chrishalcrow (commented):
Looks good to me. Is running fine for me locally with and without numba, with the right number of expected tests.

There's some commented out plotting stuff in `test_correlograms.py`, line 57 that you might consider deleting. Maybe you've already considered and decided that it's worth keeping.

I won't approve, in case someone has a good reason to keep the `if __name__ == "__main__"` block format.

src/spikeinterface/postprocessing/tests/test_correlograms.py
alejoe91: yes let's use this sintax!

alejoe91:
> Looks good to me. Is running fine for me locally with and without numba, with the right number of expected tests.
> 
> There's some commented out plotting stuff in `test_correlograms.py`, line 57 that you might consider deleting. Maybe you've already considered and decided that it's worth keeping.

Nah let's remove them, it's nice to make a proper cleanup :)

> I won't approve, in case someone has a good reason to keep the if __name__ == "__main__" block format.
That's ok for me. @samuelgarcia ?

src/spikeinterface/postprocessing/tests/test_correlograms.py
JoeZiminski: cool I did not know about that, cheers!

JoeZiminski:
Thanks both! I've made the suggested changes and raised #2990 to follow up on the commented code that indicates numpy and numba did not give (exactly) the same results. It may be they are close to some small tolerance, will look into #2990 soon.

alejoe91 (approved):

JoeZiminski:
Hey @alejoe91 thanks for this! do you think it might be possible to merge it now? It will help with avoiding conflicts with follow up PR to optimise the numba correlogram function. The follow up PR should fix all discrepancies and remove the need for the debugging code, if @samuelgarcia would like some added back I'd be happy to do in the follow up PR.

samuelgarcia:
Hi joe.
I will read it today. I promise!

src/spikeinterface/postprocessing/tests/test_align_sorting.py
```diff
@@ -12,29 +7,44 @@
...
+        unit_id: sorting.get_unit_spike_train(unit_id) + unit_peak_shifts[unit_id] for unit_id in sorting.unit_ids
+    }
+    sorting_unaligned = NumpySorting.from_unit_dict(
+        shifted_unit_dict, sampling_frequency=sorting.get_sampling_frequency()
+    )
 
     sorting_aligned = align_sorting(sorting_unaligned, unit_peak_shifts)
-    print(sorting_aligned)
-
-    for start_frame, end_frame in [(None, None), (10000, 50000)]:
```
samuelgarcia, line 32: the start/end being not none was here to tests corner cases with the shift.
for removing then ?

src/spikeinterface/postprocessing/tests/common_extension_tests.py
```diff
@@ -34,34 +34,65 @@ def get_dataset():
...
-    def setUpClass(cls):
-        cls.recording, cls.sorting = get_dataset()
-        # sparsity is computed once for all cases to save processing time and force a small radius
-        cls.sparsity = estimate_sparsity(cls.recording, cls.sorting, method="radius", radius_um=20)
-
-    @property
-    def extension_name(self):
-        return self.extension_class.extension_name
+    @pytest.fixture(autouse=True, scope="class")
+    def setUpClass(self, create_cache_folder):
```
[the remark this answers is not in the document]
alejoe91: We added a global fixture in the confpytest

src/spikeinterface/postprocessing/tests/test_correlograms.py
```diff
@@ -14,54 +12,47 @@
...
         else:
-            # ~ import matplotlib.pyplot as plt
-            # ~ for i in range(ref_correlograms.shape[1]):
-            # ~ for j in range(ref_correlograms.shape[1]):
-            # ~ fig, ax = plt.subplots()
-            # ~ ax.plot(bins[:-1], ref_correlograms[i, j, :], color='green', label='numpy')
-            # ~ ax.plot(bins[:-1], correlograms[i, j, :], color='red', label=method)
-            # ~ ax.legend()
-            # ~ ax.set_title(f'{i} {j}')
-            # ~ plt.show()
```
samuelgarcia, line 60: Having mpl commented for debugging is sometimes convenient.
But it is Ok I will but then again next time I debug theses :)

src/spikeinterface/postprocessing/tests/test_correlograms.py
```diff
@@ -14,54 +12,47 @@
 from spikeinterface.postprocessing.tests.common_extension_tests import AnalyzerExtensionCommonTestSuite
 from spikeinterface.postprocessing import ComputeCorrelograms
 from spikeinterface.postprocessing.correlograms import compute_correlograms_on_sorting, _make_bins
+import pytest
 
 
-class ComputeCorrelogramsTest(AnalyzerExtensionCommonTestSuite, unittest.TestCase):
-    extension_class = ComputeCorrelograms
```
samuelgarcia, line 20: I am a bit sad.
 I liked the idea that the commontest were aware the extension being tested.
This has could being used in test_extension no ?

src/spikeinterface/postprocessing/tests/test_align_sorting.py
JoeZiminski: I think these values [were not currently used](https://github.com/SpikeInterface/spikeinterface/blob/17ee78552ddff147e1aaefcac70364c4a8520d99/src/spikeinterface/postprocessing/tests/test_align_sorting.py#L32) so removed, but definitely happy to add these back to extend the tests. Their use should be to slice both spike-trains before testing equality?

src/spikeinterface/postprocessing/tests/test_correlograms.py
JoeZiminski: It's okay :D The tested extension is still passed to the `CommonTest`, it could be set as an attribute if required. I think this way made the pytest fixture generation easier, and allows better parameterizing over all the tests (so each tests displays on the output individually, rather than been run in a single test, in a loop in the original `test_extension()`). I think it makes the logic-flow easier to follow as a code reader. However if needed in future the extension could still be set as an attribute on the `CommonTest`.

```
    @pytest.mark.parametrize(
         "params",
         [
             dict(method="numpy"),
             dict(method="auto"),
             pytest.param(dict(method="numba"), marks=pytest.mark.skipif(not HAVE_NUMBA, reason="Numba not available")),
         ],
     )
     def test_extension(self, params):
         self.run_extension_tests(ComputeCorrelograms, params)
```

PR #2987, opened 2024-06-06
Patch unit location

yger:
Currently, unit_locations can not be properly computed for return_scaled=False, because the argument (inherited from the analyzer) is not propage to the get_dense_template() local call


alejoe91 (approved):

alejoe91:
What do you guys feel about renaming `unit_localization.py` to `unit_location.py` (same as the extension name)?

yger:
I agree, I can do that in a separate PR

PR #2988, opened 2024-06-06
Mcsh5 offsets and proper scaling in uV for return_scaled

yger:
Adding the offsets from the mcs5h header in order to have a recording with offset_to_UV and gain_to_UV. In addition, the gain was entered in V, and thus the scaling was not really making sense. As far as I understand, we want to have return_scaled returning values in uV, am I right? If yes, this is doing the job here, given the documentation that can be found online

https://www.multichannelsystems.com/sites/multichannelsystems.com/files/documents/manuals/HDF5%20MCS%20Raw%20Data%20Definition.pdf

### DIFF
--- a/src/spikeinterface/extractors/mcsh5extractors.py
+++ b/src/spikeinterface/extractors/mcsh5extractors.py
@@ -108,9 +108,9 @@ def openMCSH5File(filename, stream_id):
     Specs can be found online
     https://www.multichannelsystems.com/downloads/documentation?page=3
     """
-   
+
     import h5py
-    
+
     rf = h5py.File(filename, "r")
 
     stream_name = "Stream_" + str(stream_id)
--- a/src/spikeinterface/extractors/mcsh5extractors.py
+++ b/src/spikeinterface/extractors/mcsh5extractors.py
@@ -124,8 +124,8 @@ def openMCSH5File(filename, stream_id):
     Tick = info["Tick"][0] / 1e6
     exponent = info["Exponent"][0]
     convFact = info["ConversionFactor"][0]
-    gain_uV = 1e6*(convFact.astype(float) * (10.0**exponent))
-    offset_uV = 1e6*(info['ADZero'].astype(float) * (10.0**exponent))
+    gain_uV = 1e6 * (convFact.astype(float) * (10.0**exponent))
+    offset_uV = 1e6 * (info["ADZero"].astype(float) * (10.0**exponent))
 
     nRecCh, nFrames = data.shape
     channel_ids = [f"Ch{ch}" for ch in info["ChannelID"]]
@@ -155,7 +155,7 @@ def openMCSH5File(filename, stream_id):
         "electrode_labels": electrodeLabels,
         "gain": gain_uV,
         "dtype": dtype,
-        "offset" : offset_uV
+        "offset": offset_uV,
     }
 
     return mcs_info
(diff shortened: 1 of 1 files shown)


src/spikeinterface/extractors/mcsh5extractors.py
```diff
@@ -121,7 +124,8 @@ def openMCSH5File(filename, stream_id):
     Tick = info["Tick"][0] / 1e6
     exponent = info["Exponent"][0]
     convFact = info["ConversionFactor"][0]
-    gain = convFact.astype(float) * (10.0**exponent)
+    gain_uV = 1e6 * (convFact.astype(float) * (10.0**exponent))
+    offset_uV = 1e6 * (info["ADZero"].astype(float) * (10.0**exponent))
```
h-mayorquin, line 128: Given this from the specification:
image


Shouldn't the offset be negative?

h-mayorquin (approved):
This makes sense to me save the question  about the offset. Could we add a link to the specification in the docstring now that we are a it?

https://www.multichannelsystems.com/downloads/documentation?page=3

This link or something better if you know.

src/spikeinterface/extractors/mcsh5extractors.py
h-mayorquin: Great.

samuelgarcia: In our implementation the offset is after the gain.
So the offset should be divided by gain I think. Can you check ?

samuelgarcia:
Does the dtype is always a uint ? why is there this offset ?

src/spikeinterface/extractors/mcsh5extractors.py
```diff
@@ -122,7 +129,8 @@ def openMCSH5File(filename, stream_id):
     Tick = info["Tick"][0] / 1e6
     exponent = info["Exponent"][0]
     convFact = info["ConversionFactor"][0]
-    gain = convFact.astype(float) * (10.0**exponent)
+    gain_uV = 1e6 * (convFact.astype(float) * (10.0**exponent))
+    offset_uV = -1e6 * (info["ADZero"].astype(float) * (10.0**exponent))
```
alejoe91, line 133:
```suggestion
    offset_uV = -1e6 * (info["ADZero"].astype(float) * (10.0**exponent)) / gain_uV
```

alejoe91: See @samuelgarcia commeent

src/spikeinterface/extractors/mcsh5extractors.py
```diff
@@ -122,7 +129,8 @@ def openMCSH5File(filename, stream_id):
     Tick = info["Tick"][0] / 1e6
     exponent = info["Exponent"][0]
     convFact = info["ConversionFactor"][0]
-    gain = convFact.astype(float) * (10.0**exponent)
+    gain_uV = 1e6 * (convFact.astype(float) * (10.0**exponent))
+    offset_uV = -1e6 * (info["ADZero"].astype(float) * (10.0**exponent)) / gain_uV
```
samuelgarcia, line 133:
```suggestion
    offset_uV = -1e6 * (info["ADZero"].astype(float) * (10.0**exponent)) * gain_uV
```

samuelgarcia (approved):

PR #2992, opened 2024-06-06
Moving unit_localization into unit_locations for consistency

yger:
As suggested by @alejoe91 , this PR moves all the functions related to unit_localizations into unit_locations. I think this make sense, and is more coherent with the fact that we are refering to spike_locations


yger:
Note that currently, in the components, this is a bit of a jungle, and localization/locations are not fully harmonized. This is not crucial for now I think, but this is worth keeping in mind for later

alejoe91 (approved):

PR #2993, opened 2024-06-06
Fix timestamps access in NWB extractor

khl02007:
I'm getting `AttributeError: 'Group' object has no attribute 'timestamps'` error when accessing an NWB recording with explicit timestamps. This seems to resolve the error.


h-mayorquin (approved):
This path makes sense, we should add a test for this when we have time.

PR #2997, opened 2024-06-07
whiteningRange added as Kilosort 2.5 parameter

Djoels:
previously this parameter was set to 32, by adding it to default parameters this can now be set to a different value.


alejoe91:
Would it make sense to propagate to KS2 and KS3? Not sure if this is also an available params to KS4

Djoels:
> Would it make sense to propagate to KS2 and KS3? Not sure if this is also an available params to KS4

yes, I just added another commit for KS2 and KS3, same parameter also exists for KS4 but was added already previously

alejoe91 (approved):

alejoe91 (approved):

PR #2827, opened 2024-05-10
Prepare 0.100.7 release

alejoe91:
Propagated the following bug-fix PRs:

- #2649 
- #2810
- #2828
- #2889
- #2891

PR #2998, opened 2024-06-07
Updated release notes and pyproject for 0.101.0rc0

alejoe91:
@samuelgarcia 

I would merge this (#2827) first and release 0.100.7 bug fixes.

then this is ready to follow!

### DIFF
--- a/doc/releases/0.100.5.rst
+++ b/doc/releases/0.100.5.rst
@@ -1,0 +1,12 @@
+.. _release0.100.5:
+
+SpikeInterface 0.100.5 release notes
+------------------------------------
+
+6th April 2024
+
+Minor release with bug fixes
+
+* Open Ephys: Use discovered recording ids to load sync timestamps (#2655)
+* Fix channel gains in NwbRecordingExtractor with backend (#2661)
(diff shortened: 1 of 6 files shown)


zm711:
Like Heberto I don't fully understand the DEV flag? So in this type of release would the dev be False? Or you don't need to change that?

alejoe91:
@samuelgarcia this is an edge case, what do you think?

zm711:
Also super tiny, but if we decide to do some sort of automated changelog in the future this is 0.101.0rc0, no? In the title of the PR :)

samuelgarcia:
The DEV must be False in pypi release. It tell the code that for sure the install not in editable mode and then for installing SI inside conatiners then we use the pypi.
When DEV=True then can copy the sources inside the dockers.

So yes DEV=False for every release and then True just after.

alejoe91:
@samuelgarcia DEV=False is merged. Can someone approve this so we cut the pre-release for @DradeAW ?

@zm711 @h-mayorquin

yger (approved):

zm711 (approved):

alejoe91:
Thanks guys!

PR #3008, opened 2024-06-09
Use "available" for memory caching

DradeAW:
As explained in the doc (https://psutil.readthedocs.io/en/latest/#processes), "available" works better cross-platform. Additionally, "free" means that if a lot of RAM is used by the cache, then it won't pre-process, which is annoying


yger (approved):

yger:
Thanks for the hint, good catch!

PR #3015, opened 2024-06-11
Remove get_template_extremum_channel from spikelocations init

### LINKED ISSUE #3010: Cannot delete/load spike_locations when templates is missing
The extension `spike_locations` relies on template information in its `init`. But spikeinterface initializes extensions when we delete them (when they're saved in a folder). So it's impossible to delete `spike_locations`, meaning it's impossible to do any recomputes on sorting analyzers with `spike_locations`.

Hence the following errors...

```
recording, sorting = si.generate_ground_truth_recording()
sorting_analyzer = si.create_sorting_analyzer(
    recording=recording, 
    sorting=sorting, 
    format="binary_folder", 
    folder="temp_folder",
    overwrite=True
)
sorting_analyzer.compute(["random_spikes", "templates", "spike_locations"])
sorting_analyzer.compute(["random_spikes"])
```

...since the recompute of `random_spikes` first deletes `templates`, then tries to delete `spike_locations`

The problem only occurs for the `spike_locations` extension.

chrishalcrow:
Fixes #3010

Removes `get_template_extremum_channel`, and hence the dependency on the `templates` extension, from the `SpikeLocations` init. It looks like we don't need to saves `spikes` at all in the extensions, I completely removed this line (this is similar to `ComputeSpikeAmplitudes`). @samuelgarcia can you double check this?

Note for #2984, on how to test this. The problem here was that `SpikeLocations` couldn't be initialised without `Templates` being present. To test this we could add something to the `AnalyzerExtensionCommonTestSuite`. Perhaps at the end, we could delete all extensions except the one we're interested in, then save it and load it. This would test that each extension can be saved loaded independent of what other extensions exist? @JoeZiminski thoughts?

### DIFF
--- a/src/spikeinterface/postprocessing/spike_locations.py
+++ b/src/spikeinterface/postprocessing/spike_locations.py
@@ -59,9 +59,6 @@ class ComputeSpikeLocations(AnalyzerExtension):
     def __init__(self, sorting_analyzer):
         AnalyzerExtension.__init__(self, sorting_analyzer)
 
-        extremum_channel_inds = get_template_extremum_channel(self.sorting_analyzer, outputs="index")
-        self.spikes = self.sorting_analyzer.sorting.to_spike_vector(extremum_channel_inds=extremum_channel_inds)
-
     def _set_params(
         self,
         ms_before=0.5,
@@ -89,8 +86,9 @@ class ComputeSpikeLocations(AnalyzerExtension):
     def _select_extension_data(self, unit_ids):
         old_unit_ids = self.sorting_analyzer.unit_ids
         unit_inds = np.flatnonzero(np.isin(old_unit_ids, unit_ids))
+        spikes = self.sorting_analyzer.sorting.to_spike_vector()
 
-        spike_mask = np.isin(self.spikes["unit_index"], unit_inds)
+        spike_mask = np.isin(spikes["unit_index"], unit_inds)
         new_spike_locations = self.data["spike_locations"][spike_mask]
         return dict(spike_locations=new_spike_locations)
 


src/spikeinterface/postprocessing/spike_locations.py
```diff
@@ -59,9 +59,6 @@ class ComputeSpikeLocations(AnalyzerExtension):
     def __init__(self, sorting_analyzer):
         AnalyzerExtension.__init__(self, sorting_analyzer)
 
-        extremum_channel_inds = get_template_extremum_channel(self.sorting_analyzer, outputs="index")
-        self.spikes = self.sorting_analyzer.sorting.to_spike_vector(extremum_channel_inds=extremum_channel_inds)
```
samuelgarcia, line 63: Unfortunatly this is a need!
The spike locations algo need the channel_index per spikes which is not in the "standard/basic" spike_vector.
We need the extremum_channel_inds to assign a channel_index per spikes which is needed exactly here
https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/postprocessing/spike_amplitudes.py#L184

chrishalcrow: Merci.

We get rid of those lines, but the extremum_channel_inds are still calculated here:
https://github.com/SpikeInterface/spikeinterface/blob/17ee78552ddff147e1aaefcac70364c4a8520d99/src/spikeinterface/postprocessing/spike_locations.py#L103
and then sent to thepipeline on the line after that one. It's just that now they aren't calculated in the __init__, they've only calculated when the calculation happens.

I don't think spike_locations uses SpikeAmplitudeNode, which you've linked to.

alejoe91 (approved):

alejoe91:
@samuelgarcia can we merge this?

JoeZiminski:
@chrishalcrow that seems like a good test! For explicit tests on 'correctness' for `SortingAnalzyer`, in #3017 the approach (awaiting feedback) was to test everything thoroughly against the function that takes the `Sorting` object, then quickly check `SortingAnalzyer` and `Sorting` objects match. I'm not sure for `ComputeSpikeLocations` if it also accepts the sorting object? I'm interested in extending postprocessing tests so happy to discuss further! Maybe I'll play around and open a PR next week.

Looking at the main functions for `[ComputeSpikeLocations](https://github.com/SpikeInterface/spikeinterface/blob/e431bd6f7320753a32e357422add300180be17c1/src/spikeinterface/postprocessing/spike_locations.py#L14)` ((`set_params`, `_run_pipeline_node`, `_get_data`) this code is non-trivial and could easily be broken in future refactoring's. As such I'm not really happy with my above approach of testing against `Sorting` once, and would advocate a full test that all expected arguments are passed to the subfunction. This could be achieved by monkeypatching the subfunction and checking the correct arguments are passed. However this is probably a consideration for a little further in the future.

samuelgarcia (approved):

Issue #2800, opened 2024-05-03
Changes to 'Modules Documentation' section of documentation


JoeZiminski:
This is a repost from a discussion originally on #2778 on possible reorganisation of the 'Modules Documentation' section. This issue is a partner to #2327 which tackles organising what is a 'How To' vs. 'Tutorial' and how it is presented with sphinx gallery.

I agree the distinction between In-Depth and other documentation sections might not be clear. I think 'Module Documentation' is okay as a name but it relies on the reader to know what a 'Module' is.

I've had a look through and quick attempt and catalogue these sections (below). The information in there is super-valuable but contains a mix of different categories of documentation. I think this is in part why it is difficult to name and navigate as a user. Because of this I think it matters less what it's called now and we could have a follow-up plan to reorganise it soon. But happy to keep as 'Module Documentation' or maybe 'API Explanation'  (mirroring [diataxis](https://diataxis.fr/) explanation section) or 'Detailed Reference'.

Based on the below I think this section contains a mix of various documentation categories:
- General API details / explanations. These are things like the idea of the [preprocessing chain concept](https://spikeinterface--2778.org.readthedocs.build/en/2778/in_depth/preprocessing.html#overview-the-chain-concept), [Quality metrics module](https://spikeinterface--2778.org.readthedocs.build/en/2778/in_depth/qualitymetrics.html), [sorting components](https://spikeinterface--2778.org.readthedocs.build/en/2778/in_depth/sortingcomponents.html) philosophy and API overview. I think all of this could be kept here in a 'API Explanation'  / 'More Details' / 'Advanced Documentation' / 'User Guide' section. Its all extremely useful and doesn't fall into another existing category. 
-  Targeted API details / explanations. These are related to specific public-facing functions, in particular in the [preprocessing](https://spikeinterface--2778.org.readthedocs.build/en/2778/in_depth/preprocessing.html#available-preprocessing) section. These are close to the API docs, just with a little more info, and possibly risk duplicating the API section. For these I might suggest moving them to the actual function docstring, similar to a Numpy/Scipy-style approach where the public-facing docstrings contain a lot of detail / explanation. Then everything is in one place and linked closely to the code.
- There are some things that could be moved / merged with existing how-to guides and tutorials

General overview:

## Core module

I would suggest the below subsections are API reference / developer docs that cover how the API is structured but aren’t linked to any specific functionality or goal. These could be kind of deep API reference / explanation (mostly for developers):

- Recording
- Sorting
[...]

PR #3017, opened 2024-06-11
Optimize numba cross-correlation and extend `correlograms.py` docstrings and tests

### LINKED ISSUE #2990: Numpy and Numba do not give exactly the same output in unit test environment
There was a test in `test_correlogram.py` that indicates the output of numpy and numba are slightly different when building correlograms.

```
def _test_correlograms(sorting, window_ms, bin_ms, methods):
    for method in methods:
        correlograms, bins = compute_correlograms_on_sorting(sorting, window_ms=window_ms, bin_ms=bin_ms, method=method)
        if method == "numpy":
            ref_correlograms = correlograms
            ref_bins = bins
        else:
            # ~ import matplotlib.pyplot as plt
            # ~ for i in range(ref_correlograms.shape[1]):
            # ~ for j in range(ref_correlograms.shape[1]):
            # ~ fig, ax = plt.subplots()
            # ~ ax.plot(bins[:-1], ref_correlograms[i, j, :], color='green', label='numpy')
            # ~ ax.plot(bins[:-1], correlograms[i, j, :], color='red', label=method)
            # ~ ax.legend()
            # ~ ax.set_title(f'{i} {j}')
            # ~ plt.show()

            # numba and numyp do not have exactly the same output
            # assert np.all(correlograms == ref_correlograms), f"Failed with method={method}"

            assert np.allclose(bins, ref_bins, atol=1e-10), f"Failed with method={method}"
```
The commented section was removed in #2984 but should be looked into, I'm happy to do this going to work a bit on `postprocessing` tests soon.

JoeZiminski:
This PR (closes #2990):

1) optimizes the numba cross-correlation implementation, thanks to @DradeAW for advice on how to best optimise it.
2) extends docstrings in correlogram.py and test_correlogram.py
3) Makes a fix to the numpy implementation of correlogram computation to handle a left-bin edge case that was giving wrong results.
4) Adds some additional tests to `test_correlogram.py` and reworks some existing tests now that numba / numpy implementations match.

I would like to hear @samuelgarcia opinion on the `test_sortinganalyzer_correlograms()` that is supplemental to `run_extension_tests()`. It again creates the sortinganalyzer so I think the implementation could be handled better to avoid repeating sortinganalyzer generation. A benefit is that it directly tests the output of sortinganalyzer output, with a lightweight approach simply against the `sorting` implementation. The `sorting` object is tested in more detail in the rest of the tests.

This PR also makes some potentially contentious API changes that I'm happy to reverse:
- Remove `compute_autocorrelogram_from_spiketrain`. These were only working for the numba implementation and were untested. My thoughts are it is better to force users to use the central API, to restrict the surface API and help with maintenance. If people want to use correlation functions outside of the central SpikeInterface API they can delve into the private functions, but its easier only to support a small amount of entry functions.
- Makes every function except for `compute_correlograms` private (except, `correlogram_for_one_segment` which is called elsewhere from within the codebase).

**TODO**: As far as I can tell the docs are giving the wrong default argument values for `compute_correlograms()`, could someone help me track down where these are set?

More details can be found in the dropdowns below.

<details><summary><b>Details on Numba optimisation</b></summary>

## Numba Optimisation

The existing numba correlogram version had not yet been optimised with the new spike vector approach. It did not scale well when increasing the number of units.

numpy-v-numba-changing-units
numy-v-numba-increase-duration

With the new implementation it is much faster

numpy_vs_numba_new-durations
numpy_vs_numba_new_units
and with a very large recording for fun
numpy_vs_numbanew_long_dur_units

</details>

<details><summary><b>Details on fixing the tests</b></summary>

Below there are some specific details on the steps taken to fix the tests (i.e. address the left-bin problem in the numpy implementation and make the numpy and numba implantations match. It's quite fiddly and only really put here for posterity but is not necessary reading.

## Addressing why the previous numba and numpy version did not match

The previous `numba` implementation did not also match exactly the numpy version, in part due to a difference in handling an edge case. When there is a different between two spike times which divides perfectly into the `bin_size`, the two methods handled the case differently. I don't think there is a correct way to handle this, technically the spike can go in the bin to the left or the right (e.g. if the diff is 5 ms and the `bin_size`is 5 ms, it could be convention to go in 0-5 or 5-10). In the numpy implementation, all spike differences are positive, then the `sign` is iterated over to assign to mirrored unit comparisons (e.g. unit 1 vs. 0 and unit 0 vs. 1). Say we have 15 bins, and `diff//bin_size` is `12`. The numpy version will assign this to the bins `num_bins-12=3` and `num_bins + 12=27`. In contrast, the numba version calculated everything on one direction (e.g. unit 1 vs 0) and then flips the bins. This means that the difference is negative, the bin would be`num_bins-12=3` and stored in the 4th position of the correlogram. When flipped, this becomes the 26th position (instead of 27th).

In the new implementation, due to the spike vector, the handing of mirrored bins is just brute-forced through (i.e. both 1 vs 0 and 0 vs 1 are looped over), leading to the result handling this case the same as the numpy version. There are probably more optimisations that could be done on the numba version to avoid this duplication, but as a) this version is fast b) it matches numpy c) it is simple to read and understand, I don't think this is necessary. Below are some plots showing the improvement as compared to the numpy function.

## Adressing the numpy first-bin problem

Another reason the numba implementations were not matching the numpy implementation is because the numpy implementation was also missing some counts in the very first bin on a certain edge case. Imagine that you have two units and there is a spike time in one unit that is the same time-difference from both another spike on the unit as well as a spike on a different unit. Also, this time difference happens to be exactly the window edge. e.g. `spike_times=[10, 15, 15]`, `spike_labels=[0, 0, 1]` with `num_half_bins=5` (total bins is `10`). At the start, the algorithm shifts the spike times 1 index so that the spike diff is `15-10 = 5`. The way the bins are assigned means that when `sign==-1`, this case should go in the first bin i.e. `diff*sign=-5` which equals `-num_half_bins` so in the `np.ravel_multi_index` this is assigned to the first bin, index zero. 
In contrast, for `sign==1`,  `diff=4` (rather than `diff=5`) goes into the last bin (`diff + num_half_bin=9` i.e. the last index of the bins).  When `sign==1` all spike times with `diff == num_half_bins` are masked and no longer counted. However, this means that for the next spike time in our example, spike time 10 (unit 0) is never compared with spike time 15 (unit 1). As such, this edge case is missed from the first bin (but not from the last bin). To summarize: the first bin misses repeated counts of delays when `delay == num_half_bin`. Unmasking this case at the end of the `sign==1` loop prior to the next `sign==-1` iteration recovers these lost bins.

</details>


src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -295,19 +416,19 @@ def compute_correlograms_numba(sorting, window_size, bin_size):
...
     spikes = sorting.to_spike_vector(concatenated=False)
     correlograms = np.zeros((num_units, num_units, num_bins), dtype=np.int64)
 
     for seg_index in range(sorting.get_num_segments()):
         spike_times = spikes[seg_index]["sample_index"]
         spike_labels = spikes[seg_index]["unit_index"]
 
-        _compute_correlograms_numba(
+        _compute_correlograms_numba_new(
             correlograms, spike_times.astype(np.int64), spike_labels.astype(np.int32), window_size, bin_size
```
DradeAW, line 432: We should use `arr.astype(dtype, copy=False)`

JoeZiminski: Thanks!

src/spikeinterface/postprocessing/tests/common_extension_tests.py
```diff
@@ -77,6 +77,7 @@ class instance is used for each. In this case, we have to set
         )
         self.__class__.cache_folder = create_cache_folder
 
+    # TODO: can delete this!!!
```
JoeZiminski, line 80: TODO: open another PR for this, duplicate function name

JoeZiminski: #3058

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -18,34 +18,48 @@ class ComputeCorrelograms(AnalyzerExtension):
...
+    time bin (rather than a computer correlation or covariance).
+
+    In the present implementation, a 'window' around spikes is first
+    specified. For example, if a window of 100 ms is taken, we will
+    take the correlation at lags from -50 ms to +50 ms around the spike peak.
+    In theory, we can have as many lags as we have samples. Often, this
+    visualisation is too high resolution and instead the lags are binned
+    (e.g. -50 to -45 ms, ..., -5 to 0 ms, 0 to 5 ms, ...., 45 to 50 ms).
+    When using counts as output, binning the lags involves adding up all counts across
+    a range of lags.
```
zm711, line 37: We should discuss this. style:

I think numpydoc recommends 1-2 sentence quick explanation here then go straight to parameters because docstring renderers shouldn't push all the parameter stuff down. This type of info would go in a Notes section at the end for people that want to go deep into this type of information.

I like the style, but depending on what others (@chrishalcrow @h-mayorquin ) think maybe we use a different style with the information at top instead of at bottom?

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -230,11 +267,40 @@ def compute_correlograms_numpy(sorting, window_size, bin_size):
...
+    (skipping the zero-lag case). Next, the differences between
+    spikes times in samples are converted into units relative to
+    bin_size ('binarized'). Spikes in which the binarized difference to
+    their closest neighbouring spike is greater than half the bin-size are
+    masked.
+
+    Finally, the indices of the (num_units, num_units, num_bins) correlogram
+    that need incrementing are done so with `ravel_multi_index()`. This repeats
+    for all shifts along the spike_train until no spikes have a corresponding
+    match within the window size.
```
zm711, line 288: So this would be another example where all this info could be here or go in Notes.

src/spikeinterface/postprocessing/correlograms.py
chrishalcrow: Hello, thanks Zach I hadn't thought about this. I would vote to follow numpy (but I always say this). They say the extended summary should be "A few sentences giving an extended description. This section should be used to clarify functionality, not to discuss implementation detail or background theory, which should rather be explored in the Notes section"

from https://numpydoc.readthedocs.io/en/latest/format.html#extended-summary, cited by https://numpy.org/doc/1.20/docs/howto_document.html. 

Then most of this (very nice!) explanation would go into a Notes section (and would be great to have something like these descriptions in the docs).

Happy to hear arguments against!

JoeZiminski: Nice! I agree this sounds good, I do feel uncomfortable with such long docstrings. It's useful information to have but definitely clutters. I think Notes is a good place for it!

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -107,6 +128,33 @@ def compute_correlograms(
...
+    ----------
+    See ComputeCorrelograms() for parameters.
+
+    Returns
+    -------
+
+    bins : np.ndarray
+        The bins edges in ms
+    window_size : int
+        The window size in samples
```
zm711, line 153: One thing we should discuss is that we have `frame_slice` and now we are saying `samples`. I personally prefer sample terminology, but what do thing about frames/samples vs mixed terminology vs just using frames since that is already built deeply into the library and then we explain the equivalence in the future magic glossary?

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -120,62 +168,56 @@ def _make_bins(sorting, window_ms, bin_ms):
...
+    num_bins = int(2 * num_half_bins)
 
+    return num_bins, num_half_bins
 
-def compute_crosscorrelogram_from_spiketrain(spike_times1, spike_times2, window_size, bin_size):
+
+def _compute_correlograms_on_sorting(sorting, window_ms, bin_ms, method="auto"):
     """
-    Computes the cros-correlogram between two given spike trains.
+    Computes several cross-correlogram in one course from several clusters.
```
zm711, line 195: Do we want cluster or units for our terms?

zm711 (commented):
Did you also profile just so we can see speed differences since you changed both the numpy and numba? I think fixing the edge case is great, but it would be nice to see how speed has changed too.

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -16,35 +16,53 @@
...
     -------
-    ccgs : np.array
+    correlogram : np.array
         Correlograms with shape (num_units, num_units, num_bins)
-        The diagonal of ccgs is the auto correlogram.
-        ccgs[A, B, :] is the symetrie of ccgs[B, A, :]
-        ccgs[A, B, :] have to be read as the histogram of spiketimesA - spiketimesB
+        The diagonal of correlogram is the auto correlogram. The output
+        is in bin counts.
+        correlogram[A, B, :] is the symmetry of correlogram[B, A, :]
```
chrishalcrow, line 41: I don't understand what "the symmetry of" means. Just that `correlogram[A, B, :] = correlogram[B, A, :]`?

DradeAW: `correlogram[A, B, :] = correlogram[B, A, ::-1]`

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -280,111 +358,138 @@ def correlogram_for_one_segment(spike_times, spike_labels, window_size, bin_size
...
-    to accelerate the computation.
+    to accelerate the computation. See
+    `_compute_correlograms_one_segment_numba()` for details.
+
+    Parameters
+    ----------
+    sorting : Sorting
+        A SpikeInterface Sorting object
+    window_size : int
+            The wi  ndow size over which to perform the cross-correlation, in samples
```
chrishalcrow, line 390:
```suggestion
            The wi ndow size over which to perform the cross-correlation, in samples
```

DradeAW:
Since `correlogram[A, B, :] = correlogram[B, A, ::-1]` and since the numba method now computes from all pairs at once, I think there is a way to make it even faster (around twice faster).

Right now we `start_j`, and we are going through all pairs twice.
We could make the second loop start at `i+1`, and update the correlograms of `[A, B]` and `[B, A]` at the same time.

Wouldn't that work?

JoeZiminski:
Thank @DradeAW, I agree this would be a nicer method and speed things up. I had a play around it now but it leads to a lot or finicky indexing and fencepost problems that I can't quite figure out in good time. It's kind of related to the binning issue below but also to the edge-effects issue mentioned in the dropdown:

When trying to flip the bins, the bin-edge behaviour becomes mirrored in the positive vs. negative case rather than consistent across bins. Lets say by convention you are binning with `np.floor` so all spikes are binned to the left-edge. Say you have a bins of 5 ms, and a spike time 17.5 ms difference that needs to go in the 15-20 ms bin. In the negative case`diff/bin_size=-3.5` Then you take the floor and it goes to bin `-4` (indexed relative to the middle, 0th bin). So you say your left bind edge is -20 so you are in the -20 to - 15 bin. If you are in positive-land the floor should be `3` and your left bin edge is 15 (bin is 15-20). If however you flip the bin or you round in the same way you get positive bin 4 which is left-bin edge 20, bin (20-25). 

I think this is largely solvable by consistently using floor. But then, there is a related problem with the left-most and right-most bin edge, when the diff is exactly the `bin_size`, and I think these problems interact somewhat, making it hard to solve neatly. I benchmarked a verion starting the second loop at `i` and indexing twice-per-loop into `correlograms` and the speedup is about 20% (0.0837 vs 0.104 and 3.427 vs 2.722 seconds in the slow and fast cases above), so I guess the loop itself might be not dominating the runtime, and we are not missing out on too much speedup with this double-looping. What do you think?

DradeAW:
Can't you just do this?
```
bin1 = diff // bin_size
bin2 = (-diff) // bin_size
correlograms[spike_labels[i], spike_labels[j], num_half_bins + bin1] += 1
correlograms[spike_labels[j], spike_labels[i], num_half_bins + bin2] += 1
```
There might be a more optimal method without doing the division twice, but I don't think that the division is that computationally expensive :)

src/spikeinterface/postprocessing/correlograms.py
JoeZiminski: Good Q (another one for the glossery!), I prefer `unit` I think WDYT?

JoeZiminski:
That fixing the floor issue described, but not the some other observed issues that may or may not be related. I think where I was stuck on was the middle-bin for cross-correlation that not consistently filled between forward and backward direction in the case that `diff == bin_size`. Attempting to fix this led to some left-and-right bin issues (i.e. left or right bin is empty or has too many counts). Please feel free to take a look, it would be cool to resolve this and implement an even faster option. But I don't think I can take any more of trying to fix these weird bin edge cases for now 😄

src/spikeinterface/postprocessing/correlograms.py
JoeZiminski: I think further to the discussion below on strange bin-edge behaviour, it is easier to not make an explicit statement on the nature of the bin symmetry. How about:

```
The diagonal of the correlogram (e.g. correlogram[A, A, :]) holds the unit auto correlograms. The off-diagonal elements are the cross-correlograms between units, where correlogram[A, B, :] and correlogram[B, A, :] represent cross-correlation between the same pair of units, applied in opposite directions.
```

DradeAW:
> But I don't think I can take any more of trying to fix these weird bin edge cases for now

I understand ahah,
If I find the time I might try things,
But conceptually, my code above should do EXACTLY the same as what the function did previously

src/spikeinterface/postprocessing/correlograms.py
chrishalcrow: I think this is still a bit ambiguous, you could read it quickly and think "Oh yes, A, B is in the opposite order as B, A, that makes sense." So I'd also vote for the equation. You could just add `so that correlogram[A, B, :] = correlogram[B, A, ::-1]` at the end of that description?

src/spikeinterface/postprocessing/correlograms.py
samuelgarcia: I also prefer sample.
frame is totally an historical stuff done by Alessio or Jeremy I guess.
If day we go to version one we need to stop using "frame" I guess but Alessio will be unhappy with this.

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -280,111 +358,138 @@ def correlogram_for_one_segment(spike_times, spike_labels, window_size, bin_size
...
+        ---------
+
+        correlograms: np.array
+            A (num_units, num_units, num_bins) array of correlograms
+            between all units at each lag time bin. This is passed
+            as counts for all segments are added to it.
+        spike_times : np.ndarray
+            An array of spike times (in samples, not seconds).
+            This contains spikes from all units.
+        spike_labels : np.ndarray
```
samuelgarcia, line 456: Now label is the unit_index I would change spike_labels to spike_unit_indices or something similar to avoid confusion no ?

samuelgarcia:
Thanks a lot for the excellent reimplementation.
This was a real need since the spike_vector story.

A main (an historical) reason of the slow implementation was the confusion between unit_index and unit_id in implentation. Now the implementaion use unit_index and this is of course crazy faster.

I would modify variable names to take this in account to avoid confusion  for a futur reader of this code.
What do you think ?

src/spikeinterface/postprocessing/correlograms.py
zm711: I'm firmly on unit. I think cluster is a hold out from thinking about this in mathy terms rather than neuroscience-y terms. So I think unit is clearer for most people.

src/spikeinterface/postprocessing/correlograms.py
JoeZiminski: This is interesting,I am not very familiar with this [distinction between samples and frames](https://www.mathworks.com/help/dsp/ug/sample-and-frame-based-concepts.html#:~:text=and%20Frame%20Rate-,Samples%20and%20Frames,of%20consecutive%20times%20stacked%20together.). I did not know this difference until I just looked it up. If I have understood correctly, I _think_ in this instances 'samples' would be correct, for example we have a window size of 10 samples (that together would make up 1 frame). If you had a window size of 10 frames I think this would imply 10 sets of timeseries? It makes sense for `frame_index` I guess as it reads 'index of the frame you want to slice'? Although I also prefer sample as I think `frame` is less well known.

src/spikeinterface/postprocessing/correlograms.py
JoeZiminski: Agree thanks both have added, initially I thought maybe the formula was not true in all edge cases, but looks like it is!

src/spikeinterface/postprocessing/correlograms.py
JoeZiminski: I've linked to this #2800 for futher discussion

JoeZiminski:
Thanks @samuelgarcia! That makes sense I've renamed `spike_labels` -> `spike_unit_indices`, the new `spike_vector` approach is fantastic!

samuelgarcia (approved):

samuelgarcia:
A bottle of wine is ready for you in Lyon.

h-mayorquin:
Thanks for working on this @JoeZiminski !

PR #3022, opened 2024-06-12
Add peaks_on_probe widget.

### LINKED ISSUE #2986: Is there a widget for plotting the 'peak localisation'?
In the handle motion / drift tutorial, [there is a nice peak localisation plot](https://spikeinterface.readthedocs.io/en/stable/how_to/handle_drift.html#plot-peak-localization) that is coded up in the tutorial. I think I also saw this at the workshop. Is this implemented in a widget? If not and people agree it's a good addition, happy to open a PR!

JoeZiminski:
This PR adds a 'Peaks on Probe' widget that plots the peak locations overlaid on a plot of the probe (I am not a huge fan of the name myself, feel free to suggest others!). This PR closes #2986.

image

This plot has come up a few times in [documentation](https://www.google.com/search?q=handle+motion%2Fdrift+spikeinterface&oq=handle+motion%2Fdrift+spikeinterface&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIGCAEQRRg8MgYIAhBFGDwyBggDEEUYPNIBCDg3NTBqMGo3qAIAsAIA&sourceid=chrome&ie=UTF-8) and [tutorials](https://spikeinterface.github.io/blog/) so I think could be a handy plotting function to have. A nice future addition would be to color-by-unit but if understand correctly from @samuelgarcia this is not currently possible as peaks do not hold unit. For me it is okay to include as-is and refine later if desired, the main reason I would like it is to import and call it from the motion/drift tutorial in #2879.

## Overview

The function works by taking two main inputs, pre-calculated 'peaks' and corresponding 'peak_locations'. This can be either a single one of each or a list of corresponding entries. If multiple peaks / peak locations are passed, they are plot on different column axes of the same plot.

It is optional to select to only plot peaks from a certain segment, or between two times, to restrict the y-axis limits and to handle how peaks are decimated before plotting.

## Questions
I think I might be using `make_mpl_figure()` wrong, I am performing 1 index operation to move the `Axes` objects to a 1d array [here](https://github.com/JoeZiminski/spikeinterface/blob/4640ad47f9ea0f1f4c2fa2058ef854a10212cce4/src/spikeinterface/widgets/peaks_on_probe.py#L97). Is this okay?


src/spikeinterface/widgets/peaks_on_probe.py
```diff
@@ -0,0 +1,161 @@
...
+        an array of length num_peaks with entries: (x, y)
+    segment_index : XXXX
+        If set, only peaks from this recording segment will be used.
+    times : XXX
+        The time period over which to include peaks. If `None`, peaks
+        across the entire recording will be shown.
+    ylimits : XXX
+        The y-axis limits (i.e. the probe height). If `None`, the entire
+        probe will be displayed.
+    drop_peaks : int XXX
```
alejoe91, line 40: I would call this `decimate`

src/spikeinterface/widgets/peaks_on_probe.py
```diff
@@ -0,0 +1,170 @@
...
+            (sample_index, channel_index, amplitude, segment_index)
+    peak_locations : np.array
+        A SpikeInterface 'peak_locations' array created with `localize_peaks()`.
+        an array of length num_peaks with entries: (x, y)
+    segment_index : None | int, default: None
+        If set, only peaks from this recording segment will be used.
+    times_s : None | Tuple, default: None
+        The time period over which to include peaks. If `None`, peaks
+        across the entire recording will be shown.
+    ylimits : None | Tuple, default: None
```
alejoe91, line 37: Maybe `ylim` to use the mpl semantics?

src/spikeinterface/widgets/peaks_on_probe.py
```diff
@@ -0,0 +1,170 @@
...
+        # Plot each passed peaks / peak_locations over the probe on a separate subplot
+        for ax_idx, (peaks_to_plot, peak_locs_to_plot) in enumerate(zip(peaks, peak_locations)):
+
+            ax = self.axes[ax_idx]
+            plot_probe_map(dp.recording, ax=ax)
+
+            # Only keep peaks within the specified time window
+            time_mask = (peaks_to_plot["sample_index"] > int(fs * times_s[0])) & (
+                peaks_to_plot["sample_index"] < int(fs * times_s[1])
+            )
```
alejoe91, line 108: If the `recording.has_time_vector`, should we use that instead?

src/spikeinterface/widgets/peaks_on_probe.py
```diff
@@ -0,0 +1,170 @@
...
+    Plots of different sets of peaks can be created on subplots, by
+    passing a list of peaks and corresponding peak locations.
+
+    Parameters
+    ----------
+    recording : Recording
+        A SpikeInterface recording object.
+    peaks : np.array
+        SpikeInterface 'peaks' array created with `detect_peaks()`,
+         an array of length num_peaks with entries:
```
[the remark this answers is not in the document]
JoeZiminski: Thanks, completely forgot about that!

src/spikeinterface/widgets/peaks_on_probe.py
JoeZiminski: Good point! I added some code but will look over carefully to check. Actually, I think I should be using the `time_to_sample_index()` in this latest commit to abstract away handling of `has_time_vector`, `t_start` etc?

More generally, is there a need for the `time_vector` construct to be held at all? Could the user / us only focus on providing / handling the segments `t_start` with all time vectors constructed only the fly from `t_start`, `sampling_frequency` and segment `num_samples` when required?

alejoe91: yes, I forgot we habe that function! :)

So for some systems it's important to kee a time vector, because there might be tiny misalignments in time (e.g. Neuropixels). This is important when it comes to retrieving spike times from the associated registered recording (with `sorting.get_unit_spike_train(unit_id, return_times=True)`) and correlate them with other behavioral variables on an external plot.

src/spikeinterface/widgets/peaks_on_probe.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+class PeaksOnProbeWidget(BaseWidget):
+    """
+    Generate a plot of spike peaks showing their location on a plot
+    of the probe. Color scaling represents spike amplitude.
+
+    The generated plot overlays the estimated position of a spike peak
+    (as a single point for each peak) onto a plot of the probe. The
+    dimensions of the plot are x axis: probe width, y axis: probe height.
```
zm711, line 16: We may need to think about this. I usually hear probe width and probe depth. We should think about the terminology we are using so that we can be consistent across the codebase + probeinterface!

src/spikeinterface/widgets/peaks_on_probe.py
```diff
@@ -0,0 +1,176 @@
...
+        an array of length num_peaks with entries: (x, y)
+        To plot multiple peaks in subplots, pass a list of `peak_locations`
+        here with each entry having a corresponding `peaks`.
+    segment_index : None | int, default: None
+        If set, only peaks from this recording segment will be used.
+    time_range : None | Tuple, default: None
+        The time period over which to include peaks. If `None`, peaks
+        across the entire recording will be shown.
+    ylimits : None | Tuple, default: None
+        The y-axis limits (i.e. the probe height). If `None`, the entire
```
zm711, line 42: same here height vs depth semantically

src/spikeinterface/widgets/peaks_on_probe.py
```diff
@@ -0,0 +1,176 @@
...
+        probe will be displayed.
+    decimate : int
```
zm711, line 44:
```suggestion
    decimate : int, default: 5
```

zm711 (commented):
One tiny discussion point and one docstring patch.

src/spikeinterface/widgets/peaks_on_probe.py
JoeZiminski: Thanks @zm711! I have changed here to probe depth, do you know if we have an issue for such things? At some point I think we discussed it might be nice to add a glossary to the docs, maybe this kind of thing could go there?

src/spikeinterface/widgets/peaks_on_probe.py
JoeZiminski: Thanks @alejoe91 that makes sense!

JoeZiminski:
@alejoe91 @samuelgarcia I squashed all commits into a single one that adds everything, then added a second one that removes the tests. If its still works to add in this widget without tests for now, the tests can be resurrected after the big refactoring. If happy to proceed, if this can be normal merged (not squash-merged) I'll refer to the commit that removes the test in issue #2986 to keep it tracked for posterity (although this is not strictly necessary if its a problem as I have a backup).

src/spikeinterface/widgets/peaks_on_probe.py
zm711: No we don't, but I really like the idea for a glossary. @h-mayorquin recently pointed me to the numpy glossary which is useful (I didn't realize it existed before). We should do that as part of our doc to-do list.

alejoe91 (approved):

PR #3024, opened 2024-06-12
Add more container dependency checks in `run_sorter`

### LINKED ISSUE #2855: Additional checks on GPU when running containerised sorting
I often have a hard time debugging GPU issues when running containerised sorting. There are already a lot of useful checks in SI and performing checks is not always straightforward e.g. #1398 as can't always rely on commands like `nvidia-smi` being accessible. However I think there might be room for a couple more useful checks, I've divided this into 'easy checks' and 'harder checks'.

### Easier Checks

If would be good to check that docker or singularity is installed at all on the system at all (e.g. `docker --version` returns nonzero code). At the moment if not installed it gives a unclear error. 

If you are running singularity `spython` is a dependency and if docker then `docker` (downloaded from PyPi). However these can't be installed easily by default as they don't all work cross platform. You could do something like the below in the `pyproject.toml`

```
"docker; platform_system=='Windows'",
"docker; platform_system=='Darwin'",
"spython; platform_system=='Linux'",  # I think missing from SI?
"cuda-python; platform_system != 'Darwin'",
```

Alternatively, it would be nice to raise an error if trying to run with docker and `docker` is not installed or running with singularity and `spython` is not installed. 

### Harder Checks

Today we had an issue that was quite hard to debug where  `nvidia-docker` was not installed but was required. It would be great to check if running on docker `nvidia-docker` is installed (some details [here](https://www.howtogeek.com/devops/how-to-use-an-nvidia-gpu-with-docker-containers/)). However, it sounds like this is a bit of a nightmare, AFAIK it's not cross-platform and [nvidia-docker](https://github.com/NVIDIA/nvidia-docker?tab=readme-ov-file) is now superseded by [nvidia-container-toolkit](https://github.com/NVIDIA/nvidia-container-toolkit). Nonetheless it might be worth checking that if trying to use docker, and on Linux, that either `nvidia-docker` or `nvidia-docker2` or `nvidia-container-toolkit` is installed (some more detailed explanation between their differences [here](https://github.com/NVIDIA/nvidia-docker/issues/1268). The best would be to check this through docker directly but I don't think it's possible.

JoeZiminski:
This PR closes #2855 by adding in additional checks on container and GPU dependencies when running `run_sorter()`. 

It adds checks that `docker` or `singularity` are installed (if running sorter with `docker` or `singularity` respectively) and also that the corresponding python packages `docker` and `spython` are installed. It also show a warning if, on Linux, some nvidia-docker tools are not available, which  are usually necessary. See #2855 for a disucssion, but these nvidia-docker dependencies are a little complex so instead of raising an error, only a warning is shown.

I tested all locally and they work okay. It is not easy to test these nvidia-docker checks in code and so have left these, they raise only a warning anyway. The tests are a little bit 'extra', in particulary testing `has_docker_python()` and `has_spython()`. This patches `sys.module` to pretend like these modules are not available, then checks these `has_<dependency>`  functions return `False`. However, these tests are a little bit complex and add a reasonable amount of code to check what is basically a try/except function so although it's nice to test as much as possible, would be happy to remove if the maintenance burden is too high.

The other tests check that at runtime, `run_sorter` raises the appropriate error when the `has_<dependency>` functions return `False`. This is achieved by 'monkeypatching' these `has_<dependency>` functions so they return `False` at runtime. Then it can be checked that `run_sorter` is raising the correct error. I think these tests are more valuable, are less complex that the tests discussed above.

### DIFF
--- a/src/spikeinterface/sorters/tests/test_runsorter_dependency_checks.py
+++ b/src/spikeinterface/sorters/tests/test_runsorter_dependency_checks.py
@@ -13,7 +13,7 @@ import copy
 
 def _monkeypatch_return_false():
     """
-    A function to monkeypatch the `has_<dependecy>` functions,
+    A function to monkeypatch the `has_<dependency>` functions,
     ensuring the always return `False` at runtime.
     """
     return False
@@ -61,12 +61,12 @@ class TestRunersorterDependencyChecks:
     @pytest.fixture(scope="function")
     def uninstall_python_dependency(self, request):
         """
-        This python fixture mocks python modules not been importable
+        This python fixture mocks python modules not being importable
         by setting the relevant `sys.modules` dict entry to `None`.
-        It uses `yeild` so that the function can tear-down the test
+        It uses `yield` so that the function can tear-down the test
         (even if it failed) and replace the patched `sys.module` entry.
 
-        This function uses an `indirect` parameterisation, meaning the
+        This function uses an `indirect` parameterization, meaning the
         `request.param` is passed to the fixture at the start of the
         test function. This is used to reuse code for nearly identical
         `spython` and `docker` python dependency tests.
--- a/src/spikeinterface/sorters/tests/test_runsorter_dependency_checks.py
+++ b/src/spikeinterface/sorters/tests/test_runsorter_dependency_checks.py
@@ -1,0 +1,144 @@
+import os
+import pytest
+from pathlib import Path
+import shutil
+import platform
+from spikeinterface import generate_ground_truth_recording
+from spikeinterface.sorters.utils import has_spython, has_docker_python
+from spikeinterface.sorters import run_sorter
+import subprocess
+import sys
+import copy
+
+
+def _monkeypatch_return_false():
+    return False
+
+
+class TestRunersorterDependencyChecks:
+    """
+    This class performs tests to check whether expected
+    dependency checks prior to sorting are run. The
+    run_sorter function should raise an error if:
+        - singularity is not installed
+        - spython is not installed (python package)
+        - docker is not installed
+        - docker is not installed (python package)
+    when running singularity / docker respectively.
+
+    Two separate checks should be run. First, that the
+    relevant `has_<dependency>` function (indicating if the dependency
+    is installed) is working. Unfortunately it is not possible to
+    easily test this core singularity and docker installs, so this is not done.
+    `uninstall_python_dependency()` allows a test to check if the
+    `has_spython()` and `has_docker_dependency()` return `False` as expected
+    when these python modules are not installed.
+
+    Second, the `run_sorters()` function should return the appropriate error
+    when these functions return that the dependency is not available. This is
+    easier to test as these `has_<dependency>` reporting functions can be
+    monkeypatched to return False at runtime. This is done for these 4
+    dependency checks, and tests check the expected error is raised.
+
+    Notes
+    ----
+    `has_nvidia()` and `has_docker_nvidia_installed()` are not tested
+    as these are complex GPU-related dependencies which are difficult to mock.
+    """
+
+    @pytest.fixture(scope="function")
+    def uninstall_python_dependency(self, request):
+        """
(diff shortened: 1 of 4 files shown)


src/spikeinterface/sorters/tests/test_runsorter_dependency_checks.py
```diff
@@ -0,0 +1,170 @@
...
+from spikeinterface.sorters.utils import has_spython, has_docker_python, has_docker, has_singularity
+from spikeinterface.sorters import run_sorter
+import subprocess
+import sys
+import copy
+
+
+def _monkeypatch_return_false():
+    """
+    A function to monkeypatch the `has_<dependecy>` functions,
```
alejoe91, line 16:
```suggestion
    A function to monkeypatch the `has_<dependency>` functions,
```

src/spikeinterface/sorters/utils/misc.py
```diff
@@ -80,3 +81,71 @@ def has_nvidia():
...
+    output = subprocess.run(command, shell=True, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+    return output
+
+
+def has_docker():
+    return _run_subprocess_silently("docker --version").returncode == 0
+
+
+def has_singularity():
+    return _run_subprocess_silently("singularity --version").returncode == 0
```
alejoe91, line 99:
```suggestion
    return _run_subprocess_silently("singularity --version").returncode == 0 or _run_subprocess_silently("apptainer --version").returncode == 0
```
Apptainer is equivalent to singularity

src/spikeinterface/sorters/tests/test_runsorter_dependency_checks.py
```diff
@@ -0,0 +1,170 @@
...
+    `has_nvidia()` and `has_docker_nvidia_installed()` are not tested
+    as these are complex GPU-related dependencies which are difficult to mock.
+    """
+
+    @pytest.fixture(scope="function")
+    def uninstall_python_dependency(self, request):
+        """
+        This python fixture mocks python modules not been importable
+        by setting the relevant `sys.modules` dict entry to `None`.
+        It uses `yeild` so that the function can tear-down the test
```
zm711, line 66:
```suggestion
        It uses `yield` so that the function can tear-down the test
```

src/spikeinterface/sorters/tests/test_runsorter_dependency_checks.py
```diff
@@ -0,0 +1,170 @@
...
+    Notes
+    ----
+    `has_nvidia()` and `has_docker_nvidia_installed()` are not tested
+    as these are complex GPU-related dependencies which are difficult to mock.
+    """
+
+    @pytest.fixture(scope="function")
+    def uninstall_python_dependency(self, request):
+        """
+        This python fixture mocks python modules not been importable
```
zm711, line 64:
```suggestion
        This python fixture mocks python modules not being importable
```

src/spikeinterface/sorters/tests/test_runsorter_dependency_checks.py
```diff
@@ -0,0 +1,170 @@
...
+        by setting the relevant `sys.modules` dict entry to `None`.
+        It uses `yeild` so that the function can tear-down the test
+        (even if it failed) and replace the patched `sys.module` entry.
+
+        This function uses an `indirect` parameterisation, meaning the
```
zm711, line 69:
```suggestion
        This function uses an `indirect` parameterization, meaning the
```

just style-guiding it :)

zm711 (commented):
couple typos.

zm711 (commented):
This looks good to me, but since I almost never use the containers I will leave it to @alejoe91 for the review for correctness.

samuelgarcia (approved):

samuelgarcia:
This looks perfect to me.
Thanks a lot for sanitazing this zone, this will be extremenlly helpfull for user fightinh with installation.

PR #3034, opened 2024-06-17
add recording.time_slice like recording.frame_slice

h-mayorquin:
Many times when I use spikeinterface I wish had this facility. recording.time_slice() to avoid having to do the conversion myself. This PR adds the functionality and a test.

That said, I am not sure I really like name, any suggestions @zm711, @JoeZiminski? 

I am thinking on discoverability.
I think that if the methods started by slice it would be nice to get the auto-completition to either complete it to time or frame but  because they are not

Maybe there is a better name?

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -679,6 +679,30 @@ class BaseRecording(BaseRecordingSnippets):
         sub_recording = FrameSliceRecording(self, start_frame=start_frame, end_frame=end_frame)
         return sub_recording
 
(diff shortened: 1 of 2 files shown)


alejoe91:
I like `time_slice` too! Thanks for this @h-mayorquin

h-mayorquin:
All right. Then the name as it is is good.

samuelgarcia (approved):

alejoe91 (approved):

samuelgarcia:
merci

PR #3038, opened 2024-06-18
Update cron job for `installation_tips`

zm711:
The mac install has been failing for the past few months.... Based on the instructions it seems like the solution is to 
provide miniconda version so it is downloaded for safety on Mac.

https://github.com/conda-incubator/setup-miniconda

Also updated to v3. Not sure if we want me to switch to push to test or if we just want to merge and then fix after the next cron job if there is a problem?

### DIFF
--- a/.github/workflows/installation-tips-test.yml
+++ b/.github/workflows/installation-tips-test.yml
@@ -28,8 +28,9 @@ jobs:
       with:
         python-version: '3.10'
     - name: Test Conda Environment Creation
-      uses: conda-incubator/setup-miniconda@v2.2.0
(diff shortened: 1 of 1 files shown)


zm711:
I'm not sure why Mac would randomly need miniconda-latest but based on my reading of the error this is the fix. As the CI wizard could you doublecheck my reading of the testing error in the Actions and let me know if you think I'm wrong :)

h-mayorquin (approved):
LGTM

Question:
will we need this if we get to have https://github.com/SpikeInterface/spikeinterface/pull/2937

?

alejoe91 (approved):

PR #3043, opened 2024-06-19
Proposal to implement a sorting_analyzer.merge_units() syntax

yger:
This is WIP, and it can be discussed, but the goal of this PR is to implement a sorting_analyzer.merge_units() call that will work as the select_units() one. 

Currently, this is only working for some extension, but i'll finish all of them while the PR remains in draft mode. This could be super useful for the GUI and/or for meta merging at the end, since ideally we want to be able to merge units with as few recomputations as possible, given all the precomputed extensions of a sorting_analyzer.

Currently, merges should be given as dict {unit_id : list_unit_ids_to_be_merged}, but this could be disscussed. In this current view, templates are then given as means of merged templates, and i'll do the same for positions. Cross-Corr can only be recomputed for the merged units, ...

- [x] random_spikes
- [x] waveforms
- [x] templates (as weighted sum of merged units)
- [x] unit_locations (as weighted sum of merged units)
- [x] spike_locations (should be discussed, since we need in theory to check the peak channel of the templates)
- [x] amplitude_scalings 
- [x] principal components
- [x] template similarity
- [x] quality metrics
- [x] cross-correlograms
- [x] isi
- [x] noise levels

In all these functions, no censor period is applied, but after discussing with @alejoe91 it was noted that maybe this would be worth considering. Let's discuss that @samuelgarcia @alejoe91


yger:
Some questions remains. When we do a merge, do we want to add the possibility to remove spikes that would violate some refractory period? If this, this might make evertyhing a bit more complicated. Also, for some extensions (such as template, unit_locations) I made the choice to use the mean average merges, but this could be discussed. For the template_similarity merge, should we recompute, or just take a sliced value? Minor details that need to be discussed

alejoe91:
> Some questions remain. When we do a merge, do we want to add the possibility to remove spikes that would violate some refractory period? If this, this might make evertyhing a bit more complicated. Also, for some extensions (such as template, unit_locations) I made the choice to use the mean average merges, but this could be discussed. For the template_similarity merge, should we recompute, or just take a sliced value? Minor details that need to be discussed

I think we should discuss about it

zm711:
The current `MergeUnitsSorting` has that option just so we keep that in mind.

https://github.com/SpikeInterface/spikeinterface/blob/4c95a2efbede1d2acbb8d9f63573180c410b26e5/src/spikeinterface/curation/mergeunitssorting.py#L35

Super excited about this!

yger:
Indeed, I could have use the merge sorting object. We'll see. But based from discussions with @samuelgarcia , my understanding was that it was much faster to work at the spike vector level, and this is why I skipped it. We should discuss. Currently, everything is now working except the metrics.

yger:
At the very least, maybe I should harmonize the way we are handling the notations of merges. In MergeUnitSorting the signature is different, since it is given as units_to_merge, new_unit_ids. 
Here, I made the choice of {unit_id : list_of_unit_ids}. We should discuss which signature do we want to chose

zm711:
yeah that's a good discussion to have. @JoeZiminski likes being part of API discussions so I'm tagging him as well!

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -737,13 +766,53 @@ def select_units(self, unit_ids, format="memory", folder=None) -> "SortingAnalyz
...
+            A new unit_ids for merged units. If given, it needs to have the same length as `units_to_merge`. If None,
+            merged units will have the first unit_id of every lists of merges
+
+        folder : Path or None
+            The new folder where selected waveforms are copied
+        format:
+        a
+        Returns
+        -------
+        we :  SortingAnalyzer
```
alejoe91, line 793:
```suggestion
        analyzer :  SortingAnalyzer
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -737,13 +766,53 @@ def select_units(self, unit_ids, format="memory", folder=None) -> "SortingAnalyz
...
+            A list of lists for every merge group. Each element needs to have at least two elements (two units to merge),
+            but it can also have more (merge multiple units at once).
+        new_unit_ids : None or list
+            A new unit_ids for merged units. If given, it needs to have the same length as `units_to_merge`. If None,
+            merged units will have the first unit_id of every lists of merges
+
+        folder : Path or None
+            The new folder where selected waveforms are copied
+        format:
+        a
```
alejoe91, line 790: something wrong here :)

yger: Thanks for spotting !

yger:
The branch is now almost working, with all the needed features. Everything can be recomputed, even the metrics. I just need to modify a bit the quality_metric_calculator to only recompute metrics for some ids and not all. By passing a sorting_analyzer to the merge function, and by assuming the extensions are properly sorted (given the _sort_extensions functions), we have all the requiered extension to compute everything on the fly

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -86,7 +86,12 @@ def _select_extension_data(self, unit_ids):
         new_data = dict(metrics=new_metrics)
         return new_data
 
-    def _run(self, verbose=False, **job_kwargs):
+    def _merge_extension_data(self, units_to_merge, new_unit_ids, new_sorting_analyzer):
+        metrics = self._compute_metrics(new_sorting_analyzer)
+        new_data = dict(metrics=metrics)
+        return new_data
```
alejoe91, line 92: So now this recomputes everything regardless of the `units_to_merge`?

yger: Now this has been improved, and metrics are recomputed only for merged indices. Other values are taken from already computed metrics

alejoe91: Parfait!

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -86,7 +86,23 @@ def _select_extension_data(self, unit_ids):
         new_data = dict(metrics=new_metrics)
         return new_data
 
-    def _run(self, verbose=False, **job_kwargs):
+    def _merge_extension_data(self, units_to_merge, new_unit_ids, new_sorting_analyzer):
+        old_metrics = self.data["metrics"]
+        import pandas as pd
+
+        metrics = pd.DataFrame(index=new_sorting_analyzer.unit_ids)
```
alejoe91, line 93: If you instantiate this with column names as well, I think you can avoid the loop and set everything with a couple of `loc` calls. I can take a stab at this on Monday

yger:
Actually, we should have a meeting, because there are some corner issues that should be discussed. In fact, if when we do a merge if we change the sparsity mask of the merged units (taking the union as I did), then we might need to reextract the waveforms, because otherwise they do not have the same layouts as before. 
More generally, handling everything in the generic case, where users could merge everything, is more complicated than assuming only similar templates are merged (this could save quite a lot of computations). So this has to be discussed

yger:
This is good to go, everything is working, with a soft and a hard mode @alejoe91 @samuelgarcia . Several observations:
- could be good to have a recompute mode (instead of the hard), the hard beeing an in between with union for the sparsity masks. I don't know
- when slicing the masks during an intersection, everything is fine at the template levels (they are stored in a dense manner), but for the waveforms, I guess we might need to reorder/slice the channels for the merged units, even if no recompute is triggered
- the censor_ms is now properly handled as a boolean mask, and this PR should be compatible with #2941

alejoe91:
> * could be good to have a recompute mode (instead of the hard), the hard beeing an in between with union for the sparsity masks. I don't know

What do you mean?

> * when slicing the masks during an intersection, everything is fine at the template levels (they are stored in a dense manner), but for the waveforms, I guess we might need to reorder/slice the channels for the merged units, even if no recompute is triggered

Yes, we probably need to slice the waveforms according to the new sparsity


@yger is it ok i

src/spikeinterface/core/sorting_tools.py
```diff
@@ -220,3 +220,102 @@ def random_spikes_selection(
...
+        assert len(units_to_merge[count]) > 1, "A merge should have at least two units"
+        for unit_id in units_to_merge[count]:
+            assert unit_id in sorting.unit_ids, "Merged ids should be in the sorting"
+        if new_unit_ids is None:
+            for unit_id in units_to_merge[count][1:]:
+                merged_unit_ids.discard(unit_id)
+        else:
+            for unit_id in units_to_merge[count]:
+                merged_unit_ids.discard(unit_id)
+            merged_unit_ids = merged_unit_ids.union([new_unit_ids[count]])
```
alejoe91, line 237: I think I prefer to give a new unit id to the merge, rather than keeping the first entry:

You can port [this function](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/curation/mergeunitssorting.py#L52-L74) to `sorting_tools`. In case unit ids are strings, the new unit will be `unit1-unit2-...`

This is another advocate IMO to force unit ids to be strings....

src/spikeinterface/core/sorting_tools.py
```diff
@@ -220,3 +220,102 @@ def random_spikes_selection(
...
+        if censor_ms is not None:
+            times_list += [spikes["sample_index"][s0:s1][to_keep[s0:s1]]]
+            labels = spikes["unit_index"][s0:s1][to_keep[s0:s1]]
+            labels_list += [labels]
+        else:
+            times_list += [spikes["sample_index"][s0:s1]]
+            labels = spikes["unit_index"][s0:s1]
+            labels_list += [labels]
+
+    sorting = NumpySorting.from_times_labels(times_list, labels_list, sorting.sampling_frequency)
```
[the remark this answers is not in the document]
yger: Cool, will do that, indeed I forgot this

yger:
I mean that currently, we have a soft mode which does not recompute anything and a hard mode that recompute everything. But there could be an in-between, where merges that could be done with none or little recomputing couple be made instead of recomputing everything. Maybe not worth it, just a suggestion.

src/spikeinterface/core/sorting_tools.py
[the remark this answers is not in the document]
yger: Actually, this is not that straightforward isn't it? Because NumpySorting assumes index are sorted with no gaps. While currently, when I'm editing in place the unit_index, some are removed (because of merges), some are added. i'll think about it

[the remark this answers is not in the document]
alejoe91: I think it should be. If there are missing unit_indices, that should still work as long as the `all_unit_ids` has the "missing" ids

yger:
The waveforms was quite a headache, but ready to read on my side, and all extensions are working! 
In the soft mode, no recomputation based on the traces is triggered whatsoever, everyhing is performed on the fly.

src/spikeinterface/core/sorting_tools.py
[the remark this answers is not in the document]
yger: Yes, but I'll need to remove these extra ids afterward. If we assume that users can not assign a merge to an already existing unit_id, then this is simpler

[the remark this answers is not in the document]
alejoe91: couldn't we just add the `new_unit_ids` and remove the `units_to_merge` ids?

src/spikeinterface/core/sorting_tools.py
```diff
@@ -220,3 +220,100 @@ def random_spikes_selection(
...
+        has been kept, given the refractory period violations (None if censor_ms is None)
+    """
+    spikes = sorting.to_spike_vector().copy()
+
+    if censor_ms is None:
+        to_keep = None
+    else:
+        to_keep = np.ones(len(spikes), dtype=bool)
+
+    from spikeinterface.curation.curation_tools import get_new_unit_ids_for_merges
```
alejoe91, line 271: can you move this at the top of the function?

src/spikeinterface/curation/curation_tools.py
```diff
@@ -133,3 +133,39 @@ def find_duplicated_spikes(
         return _find_duplicated_spikes_keep_last_iterative(spike_train.astype(np.int64), censored_period)
     else:
         raise ValueError(f"Method '{method}' isn't a valid method for find_duplicated_spikes. Use one of {_methods}")
+
+
+def get_new_unit_ids_for_merges(sorting, units_to_merge, new_unit_ids):
```
alejoe91, line 138: thinking more about this: if left here, importing this will import numba and jit compile the numba functions. 

Maybe worth moving this function too to sorting_tools? It's general enough

yger: Done

yger:
Tests are not passing because, for some unknown reasons, the select_units() in the sorting_tools functions is triggering issues. My guess is that this is not directly related to this PR, but would need some help there

yger:
Actually, While everything is working, there is a remaining question and issue for PC. I will reslice them as I did for the waveforms, but should I do with the model? I guess they will be obsolete, since the merged units will have different sparsity masks, and thus not sure that pretrained models can be kept

alejoe91:
> Actually, While everything is working, there is a remaining question and issue for PC. I will reslice them as I did for the waveforms, but should I do with the model? I guess they will be obsolete, since the merged units will have different sparsity masks, and thus not sure that pretrained models can be kept

@yger I think that PC scores are ok. Here's why:

The PC models are trained by channel and using all available waveforms. The only obsoleter aspect would be if you have a censored period, because some of the training waveforms will be dropped. Honestly, that's not a big deal.
After the merge, the projections will therefore not change. 
As for sparsity, we can use the union and:
- either leave the undefined projections as 0s (they will be peripheral anyways)
- or project the "tails" on the existing models

alejoe91:
@yger merge with main to fix CI errors ;)

yger:
PCA is also handled, so now everything is good to go. I've simplified a bit the waveforms, assuming that in soft merges, we can only have reduction of the sparsity masks and no union

alejoe91:
Missing:

- [ ] template metrics

yger:
What are these metrics ? The similarity is handled

alejoe91:
Added! These are like peak to peak, halfwidth, etc.

I now added a merge units step in every extension tests, and spotted a couple of errors:
- extended CCG to multi-segment in merge_extension_data
- fixed dense PCA merge in merge_extension_data

yger:
Cool, thanks: However, the merge for CCGs is not working anymore because of the update of the compute methods for CCG I guess. I'll try to have a look later

alejoe91:
@yger in the last couple of commits a fixed some bugs in the waveforms and PCa metrics:

- in both cases, the mask implementation was extracting multiple times large arrays, now it's better in terms of memory
- the `get_some_*` were returning an array of unit indices that you were using as indices of the spike vector! Instead, the function now correctly return the selected indices for the spike vector

In addition I added some fixes in terms of extension params back-compatibility with the `load_waveforms` function

src/spikeinterface/postprocessing/principal_component.py
```diff
@@ -269,18 +269,13 @@ def get_some_projections(self, channel_ids=None, unit_ids=None, kept_indices=Non
...
-                sparse_projection = sparse_projection[:, :, channel_mask]
-                local_chan_inds = local_chan_inds[channel_mask]
+                if sum(channel_mask) != len(channel_mask):
+                    sparse_projection = sparse_projection[:, :, channel_mask]
 
                 spike_mask = np.flatnonzero(spike_unit_indices == unit_index)
-                proj = np.zeros((spike_mask.size, num_components, channel_indices.size), dtype=dtype)
-                # inject in requested channels
-                channel_mask = np.isin(channel_indices, local_chan_inds)
-                proj[:, :, channel_mask] = sparse_projection
```
alejoe91, line 233: this array was usless and just filling up memory

yger: Tests are not passing... We've also worked this morning with @samuelgarcia to refactor some key portions of this PR. We should merge them and then rebase the two PR (merging_units and meta_merging). Let me know how you want to proceed

alejoe91: Ok,let me do that because I have some local changes with improved testing for the get_some_waveforms/projections

yger: As you want, then go ahaed. Also, I think that in get_some_projections and get_some_waveforms, we should gain some memory by looking at the number of channels in the union of all the sparsity mask, for the unit_ids that we want to select, and only create a buffer with such a size. But I'll double check

yger:
Looks like something is strange with the diff here. Too many files are marked as modified...

src/spikeinterface/core/tests/test_sortinganalyzer.py
```diff
@@ -142,6 +142,27 @@ def test_SortingAnalyzer_tmp_recording(dataset):
         sorting_analyzer.set_temporary_recording(recording_sliced)
 
 
+def test_SortingAnalyzer_merge_all_extensions(dataset):
```
alejoe91, line 145: I added these tests in the common extension tests. Core will not have all dependencies required by all extensions.

alejoe91:
I think I messed it up with the `meta_merging` branch here: https://github.com/SpikeInterface/spikeinterface/pull/3043/commits/9e9c3fd7b3a812b3b895e735877e2a3df7c78e91

alejoe91:
Removed all changes in the last commit. Let's remember to squash and merge!!!

alejoe91:
and merged with #3154

There are a couple of points left:

1. @yger I commented out your test all extensions in core. There are added tests to the postprocessing module and core doesn't have all required dependencies. If you're ok I'll remove it
2. I'd like to have a `sorting_analyzer.curation` (which could be saved to json) with the merges applied to it, so we can keep track internally. The `create` function could have an optional `curation_dict` argument which if passed, is stored as `self.curation` and saved to json/zarr

Apart from these, this is ready fo final review!

src/spikeinterface/core/sorting_tools.py
```diff
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import numpy as np
+from spikeinterface.core import NumpySorting
+from spikeinterface.core import NumpySorting
```
alejoe91, line 5:
```suggestion
```

src/spikeinterface/core/tests/test_sortinganalyzer.py
```diff
@@ -331,4 +420,5 @@ def test_extensions_sorting():
     test_SortingAnalyzer_zarr(tmp_path, dataset)
     test_SortingAnalyzer_tmp_recording(dataset)
     test_extension()
+    test_SortingAnalyzer_merge_all_extensions()
```
alejoe91, line 423:
```suggestion
```

src/spikeinterface/core/waveforms_extractor_backwards_compatibility.py
```diff
@@ -543,7 +544,11 @@ def _read_old_waveforms_extractor_binary(folder, sorting):
         ext = new_class(sorting_analyzer)
         with open(ext_folder / "params.json", "r") as f:
             params = json.load(f)
-        ext.params = params
+        # update params
+        new_params = ext._set_params()
+        updated_params = make_ext_params_up_to_date(ext, params, new_params)
```
alejoe91, line 549: added this here because if we don't do this old waveform extractors loaded as analyzers won't be able to undergo merging because of conflicts with parameters.

yger:
This is now ready to read, and the run times are now mucchh better (they were really slow with the waveforms/pca) @samuelgarcia @alejoe91  the only question that remains is if we want to make a json for the merged units, but otherwise this is good for me now

samuelgarcia:
Salut Pierre.
I will have a look this afternoon.
Do not push anymore stuff to avoid collision.

yger:
ok camarade

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -76,6 +76,19 @@ def _select_extension_data(self, unit_ids):
...
 
+    def _merge_extension_data(
+        self, units_to_merge, new_unit_ids, new_sorting_analyzer, keep_mask=None, verbose=False, **job_kwargs
+    ):
+        new_data = dict()
+        if keep_mask is not None:
+            valid = keep_mask[self.sorting_analyzer.get_extension("random_spikes")._get_data()]
+            nb_skipped = np.cumsum(~keep_mask)
+            new_data["random_spikes_indices"] = np.flatnonzero(valid)
+            new_data["random_spikes_indices"] -= nb_skipped[new_data["random_spikes_indices"]]
```
samuelgarcia, line 87: This hard to read and not so fast.
Why not simple boolean mask on the new space ? and then nonzeor on it ?

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -76,6 +76,19 @@ def _select_extension_data(self, unit_ids):
...
+        else:
+            new_data["random_spikes_indices"] = self.data["random_spikes_indices"]
```
samuelgarcia, line 89: we shold make copy

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -224,18 +237,75 @@ def _select_extension_data(self, unit_ids):
...
+
+        waveforms = self.data["waveforms"]
+        some_spikes = self.sorting_analyzer.get_extension("random_spikes").get_random_spikes()
+
+        if keep_mask is not None:
+            spike_indices = self.sorting_analyzer.get_extension("random_spikes").get_data()
+            valid = keep_mask[spike_indices]
+            some_spikes = some_spikes[valid]
+            waveforms = waveforms[valid]
+
```
samuelgarcia, line 253: we should make copy of waveforms in a else

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -248,6 +318,92 @@ def get_waveforms_one_unit(
...
+                # inject in requested channels
+                spike_mask = np.flatnonzero(spike_unit_indices == unit_index)
+                channel_mask_channel_inds = np.isin(channel_indices, local_chan_inds)
+                if sum(channel_mask_channel_inds) < len(channel_mask_channel_inds):
+                    # inject in requested channels
+                    wfs = np.zeros((spike_mask.size, num_samples, channel_indices.size), dtype=dtype)
+                    wfs[:, :, channel_mask_channel_inds] = sparse_waveforms
+                else:
+                    wfs = sparse_waveforms
+                some_waveforms[spike_mask] = wfs
```
samuelgarcia, line 403: This is should be outside in helper function.
Herte we have a weird design.

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -410,11 +558,21 @@ def _compute_and_append_from_waveforms(self, operators):
 
     @property
     def nbefore(self):
-        return self.params["nbefore"]
+        waveforms_extension = self.sorting_analyzer.get_extension("waveforms")
+        if waveforms_extension is not None:
+            nbefore = waveforms_extension.nbefore
+        else:
+            nbefore = int(self.params["ms_before"] * self.sorting_analyzer.sampling_frequency / 1000.0)
+        return nbefore
```
samuelgarcia, line 566: bad idea.
The wavefortms extesion could be deleted in bteween!
The params must captured on the run.

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -410,11 +558,21 @@ def _compute_and_append_from_waveforms(self, operators):
...
 
     @property
     def nafter(self):
-        return self.params["nafter"]
+        waveforms_extension = self.sorting_analyzer.get_extension("waveforms")
+        if waveforms_extension is not None:
+            nafter = waveforms_extension.nafter
+        else:
+            nafter = int(self.params["ms_after"] * self.sorting_analyzer.sampling_frequency / 1000.0)
+        return nafter
```
samuelgarcia, line 575: Same here.

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -582,6 +770,12 @@ def _select_extension_data(self, unit_ids):
         # this do not depend on units
         return self.data
 
+    def _merge_extension_data(
+        self, units_to_merge, new_unit_ids, new_sorting_analyzer, keep_mask=None, verbose=False, **job_kwargs
+    ):
+        # this do not depend on units
+        return self.data
```
samuelgarcia, line 777: we should make a copy.

src/spikeinterface/core/analyzer_extension_core.py
yger: We need to keep only the indices of the valid spikes. Then substract to each value the number of spikes that has been removed before this particular index. But open to suggestion

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -629,36 +630,89 @@ def set_temporary_recording(self, recording: BaseRecording):
...
+                    id = np.flatnonzero(unit_ids == unit_id)[0]
+                    to_be_merged = units_to_merge[id]
+                    indices = self.sorting.ids_to_indices(to_be_merged)
+                    union_mask = np.sum(self.sparsity.mask[indices], axis=0) > 0
+                    intersection_mask = np.prod(self.sparsity.mask[indices], axis=0) > 0
+                    if merging_mode == "soft":
+                        thr = np.sum(intersection_mask) / np.sum(union_mask)
+                        assert (
+                            thr > sparsity_overlap
+                        ), f"A sparsity threshold of {thr} has been found, can not use soft mode"
```
samuelgarcia, line 681: SHould be a bit more explicit.

src/spikeinterface/core/analyzer_extension_core.py
yger: ok

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -224,18 +237,75 @@ def _select_extension_data(self, unit_ids):
...
+                )
+                num_chans = new_waveforms.shape[2]
+                new_data["waveforms"][spike_indices, :, :num_chans] = new_waveforms
+            new_data["waveforms"] = new_data["waveforms"][:, :, :max_num_chans]
+        else:
+            new_data["waveforms"] = waveforms
+
+        return new_data
+
+    def get_waveforms_one_unit(self, unit_id, force_dense: bool = False, kept_waveforms=None, kept_spikes=None):
```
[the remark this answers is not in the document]
yger: I know, but what is your suggestion ? We need to provide to these functions the subarrays of valid waveforms and spikes, in order to avoid constently recomputing them by slicing. This is much faster.

src/spikeinterface/core/analyzer_extension_core.py
yger: done

src/spikeinterface/core/analyzer_extension_core.py
yger: This is exactly the same design that used to be there for the PC.

src/spikeinterface/core/analyzer_extension_core.py
yger: The set_params() can only save params that are in the init, nothing more. Otherwise, we have issues when reloading extensions. Let's discuss that

src/spikeinterface/core/analyzer_extension_core.py
yger: same here

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -691,10 +745,29 @@ def _save_or_select(self, format="binary_folder", folder=None, unit_ids=None) ->
...
+                        keep_mask=keep_mask,
+                        verbose=verbose,
+                        **job_kwargs,
+                    )
+                elif merging_mode == "hard":
+                    recompute_dict[extension_name] = extension.params
+            else:
+                new_sorting_analyzer.extensions[extension_name] = extension.copy(
+                    new_sorting_analyzer, unit_ids=unit_ids
+                )
```
samuelgarcia, line 767: We need to capture the params fore the recompute. no ?
recompute_dict is not update  no ?

src/spikeinterface/core/analyzer_extension_core.py
yger: done

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -729,21 +802,102 @@ def select_units(self, unit_ids, format="memory", folder=None) -> "SortingAnalyz
...
+    ) -> "SortingAnalyzer":
+        """
+        This method is equivalent to `save_as()`but with a list of merges that have to be achieved.
+        Merges units by creating a new sorting analyzer object in a new folder with appropriate merges
+
+        Extensions are also updated to display the merged unit ids.
+
+        Parameters
+        ----------
+        units_to_merge : list/tuple of lists/tuples
```
samuelgarcia, line 837: I am not super happy with this naming.
I know I did the same in the other PR but it is time to find a cristal clear naming

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -629,36 +630,89 @@ def set_temporary_recording(self, recording: BaseRecording):
...
             sorting_provenance = sorting_provenance.select_units(unit_ids)
+        else:
+            from spikeinterface.core.sorting_tools import apply_merges_to_sorting
+
+            sorting_provenance, keep_mask = apply_merges_to_sorting(
+                sorting=sorting_provenance,
+                units_to_merge=units_to_merge,
+                new_unit_ids=unit_ids,
+                censor_ms=censor_ms,
+                return_kept=True,
```
samuelgarcia, line 712: we need the option of new_id_strategy here.

src/spikeinterface/core/sortinganalyzer.py
yger: I've tried something

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -729,21 +802,102 @@ def select_units(self, unit_ids, format="memory", folder=None) -> "SortingAnalyz
...
+        return self._save_or_select_or_merge(
+            format=format,
+            folder=folder,
+            units_to_merge=units_to_merge,
+            unit_ids=new_unit_ids,
+            censor_ms=censor_ms,
+            merging_mode=merging_mode,
+            sparsity_overlap=sparsity_overlap,
+            verbose=verbose,
+            **job_kwargs,
```
samuelgarcia, line 893: new_id_strategy need to be propagate no ?

src/spikeinterface/core/sortinganalyzer.py
yger: No, because this is handled before, and new_unit_ids here is never None

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1664,6 +1824,26 @@ def copy(self, new_sorting_analyzer, unit_ids=None):
...
+        units_to_merge=None,
+        new_unit_ids=None,
+        keep_mask=None,
+        verbose=False,
+        **job_kwargs,
+    ):
+        new_extension = self.__class__(new_sorting_analyzer)
+        new_extension.params = self.params.copy()
+        if units_to_merge is None:
+            new_extension.data = self.data
```
samuelgarcia, line 1839: copy ?

src/spikeinterface/core/sortinganalyzer.py
yger: lists_of_ids_to_merge?

src/spikeinterface/core/tests/test_sortinganalyzer.py
```diff
@@ -141,6 +142,44 @@ def test_SortingAnalyzer_tmp_recording(dataset):
         sorting_analyzer.set_temporary_recording(recording_sliced)
 
 
+# TODO: Alessio's opinion: this doesn't belong here
```
samuelgarcia, line 145: I agree

src/spikeinterface/core/sortinganalyzer.py
yger: recompute_dict is filled with the params of the extension at the time of the merge

src/spikeinterface/core/tests/test_sortinganalyzer.py
yger: ok, then let's remove it

src/spikeinterface/core/sortinganalyzer.py
yger: new_id_strategy is handled at the merge_units() level, and then, after, new_units_id is never None

src/spikeinterface/postprocessing/amplitude_scalings.py
```diff
@@ -114,6 +114,21 @@ def _select_extension_data(self, unit_ids):
...
+    ):
+        new_data = dict()
+        new_data["amplitude_scalings"] = self.data["amplitude_scalings"]
+        if self.params["handle_collisions"]:
+            new_data["collision_mask"] = self.data["collision_mask"]
+
+        if keep_mask is not None:
+            new_data["amplitude_scalings"] = new_data["amplitude_scalings"][keep_mask]
+            if self.params["handle_collisions"]:
+                new_data["collision_mask"] = new_data["collision_mask"][keep_mask]
```
samuelgarcia, line 128: lets do if .. else .. and use copy when not None no ?

src/spikeinterface/core/sortinganalyzer.py
yger: done

src/spikeinterface/postprocessing/isi.py
```diff
@@ -56,6 +56,26 @@ def _select_extension_data(self, unit_ids):
...
+        all_new_units = new_sorting_analyzer.unit_ids
+        new_isi_hists = np.zeros((len(all_new_units), num_dims), dtype=arr.dtype)
+        for unit_ind, unit_id in enumerate(all_new_units):
+            if unit_id not in new_unit_ids:
+                keep_unit_index = self.sorting_analyzer.sorting.id_to_index(unit_id)
+                new_isi_hists[unit_ind] = arr[keep_unit_index]
+            else:
+                new_sorting = new_sorting_analyzer.sorting.select_units([unit_id])
+                new_hist, _ = _compute_isi_histograms(new_sorting, **self.params)
+                new_isi_hists[unit_ind] = new_hist
```
samuelgarcia, line 74: I have the intuition that this is sub optimal to make many select_units instead of one.

src/spikeinterface/postprocessing/principal_component.py
```diff
@@ -101,6 +101,42 @@ def _select_extension_data(self, unit_ids):
...
+            max_num_chans = np.sum(new_sorting_analyzer.sparsity.mask, 1).max()
+            new_data["pca_projection"] = pca_projections.copy()
+            for to_be_merge, unit_id in zip(units_to_merge, new_unit_ids):
+                new_channel_ids = sparsity.unit_id_to_channel_ids[unit_id]
+                new_projections, spike_indices = self.get_some_projections(
+                    new_channel_ids, to_be_merge, kept_projections=new_data["pca_projection"], kept_spikes=some_spikes
+                )
+                num_chans = new_projections.shape[2]
+                new_data["pca_projection"][spike_indices, :, :num_chans] = new_projections
+            new_data["pca_projection"] = new_data["pca_projection"][:, :, :max_num_chans]
```
samuelgarcia, line 130: We need tools shared between waveforms and PCa no ?

src/spikeinterface/postprocessing/principal_component.py
```diff
@@ -120,25 +156,26 @@ def get_pca_model(self):
             pca_models = self.data[f"pca_model_{mode}"]
         return pca_models
 
-    def get_projections_one_unit(self, unit_id, sparse=False):
+    def get_projections_one_unit(self, unit_id, sparse=False, kept_projections=None, kept_spikes=None):
```
samuelgarcia, line 159: same comment that waveforms.
This new signature is a bit allien and use for the merge which should be internal

src/spikeinterface/postprocessing/spike_locations.py
```diff
@@ -92,6 +92,17 @@ def _select_extension_data(self, unit_ids):
...
 
+    def _merge_extension_data(
+        self, units_to_merge, new_unit_ids, new_sorting_analyzer, keep_mask=None, verbose=False, **job_kwargs
+    ):
+        new_spike_locations = self.data["spike_locations"]
+        if keep_mask is not None:
+            new_spike_locations = new_spike_locations[keep_mask]
+
+        ### In theory here, we should recompute the locations since the peak positions
+        ### in a merged could be different. Should be discussed
```
samuelgarcia, line 103: I agree but we were OK that mode=soft merge is not 100% extact.
But this should detail in the doc somewhere. explicitly.

src/spikeinterface/core/waveforms_extractor_backwards_compatibility.py
```diff
@@ -604,6 +609,21 @@ def _read_old_waveforms_extractor_binary(folder, sorting):
     return sorting_analyzer
 
 
+def make_ext_params_up_to_date(ext, old_params, new_params):
```
samuelgarcia, line 612: this will not handle magically the nboefore to ms_before.
no ?

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -150,24 +155,45 @@ def _select_extension_data(self, unit_ids):
...
+        metrics = pd.DataFrame(index=new_sorting_analyzer.unit_ids)
+        new_metrics = self._compute_metrics(new_sorting_analyzer, new_unit_ids, verbose, **job_kwargs)
+        for unit_id in new_sorting_analyzer.unit_ids:
+            if unit_id in new_unit_ids:
+                source = new_metrics
+            else:
+                source = old_metrics
+            for k in source.loc[unit_id].keys():
+                metrics.loc[unit_id, k] = source.loc[unit_id, k]
+        new_data = dict(metrics=metrics)
```
samuelgarcia, line 174: I have the intuition that pandas could help us to do that without looping and with one line.

yger: maybe :-)

src/spikeinterface/core/waveforms_extractor_backwards_compatibility.py
yger: This is an addition of @alejoe91 , i'll let him comment

src/spikeinterface/postprocessing/spike_amplitudes.py
```diff
@@ -81,6 +81,17 @@ def _select_extension_data(self, unit_ids):
...
         return new_data
 
+    def _merge_extension_data(
+        self, units_to_merge, new_unit_ids, new_sorting_analyzer, keep_mask=None, verbose=False, **job_kwargs
+    ):
+        new_data = dict()
+        new_data["amplitudes"] = self.data["amplitudes"]
+
+        if keep_mask is not None:
+            new_data["amplitudes"] = new_data["amplitudes"][keep_mask]
```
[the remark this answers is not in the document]
yger: I've made copy everywhere, since this is what has to be done

src/spikeinterface/postprocessing/principal_component.py
yger: Ok, but let's discuss what do do, because as said, we would nee to provide sliced waveforms/spikes to be efficient

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -63,6 +64,63 @@ def _select_extension_data(self, unit_ids):
...
+                        template1,
+                        template2,
+                        method=self.params["method"],
+                        num_shifts=num_shifts,
+                        support=self.params["support"],
+                        sparsity=sparsity1,
+                        other_sparsity=sparsity2,
+                    )
+                else:
+                    new_similarity[unit_ind1, unit_ind2] = arr[i, j]
```
samuelgarcia, line 118: we should have the same machinery for run and merge oerthwise bug and improvement will not be sync.

src/spikeinterface/postprocessing/tests/common_extension_tests.py
```diff
@@ -133,6 +133,11 @@ def _check_one(self, sorting_analyzer, extension_class, params):
         sliced = sorting_analyzer.select_units(some_unit_ids, format="memory")
         assert np.array_equal(sliced.unit_ids, sorting_analyzer.unit_ids[::2])
 
+        some_merges = [sorting_analyzer.unit_ids[:2].tolist()]
+        num_units_after_merge = len(sorting_analyzer.unit_ids) - 1
+        merged = sorting_analyzer.merge_units(some_merges, format="memory", mode="soft", sparsity_overlap=0.0)
+        assert len(merged.unit_ids) == num_units_after_merge
```
samuelgarcia, line 139: This is really a small test no ?

src/spikeinterface/postprocessing/isi.py
yger: This is really fast, not sure this is worth optimizing

src/spikeinterface/postprocessing/unit_locations.py
```diff
@@ -49,17 +49,43 @@ def __init__(self, sorting_analyzer):
...
+                new_unit_location[unit_ind] = arr[keep_unit_index]
+            else:
+                id = np.flatnonzero(new_unit_ids == unit_id)[0]
+                unit_ids = units_to_merge[id]
+                keep_unit_indices = self.sorting_analyzer.sorting.ids_to_indices(unit_ids)
+                weights = np.zeros(len(unit_ids), dtype=np.float32)
+                for count, id in enumerate(unit_ids):
+                    weights[count] = counts[id]
+                weights /= weights.sum()
+                new_unit_location[unit_ind] = (arr[keep_unit_indices] * weights[:, np.newaxis]).sum(0)
```
samuelgarcia, line 81: why not a clear recompute ? This is not always linear.

src/spikeinterface/postprocessing/template_similarity.py
yger: This is the same function called internally. I'm ok to recompute everything (as done with CCGs), but this is slower when you have only few merges

src/spikeinterface/postprocessing/unit_locations.py
yger: What do you mean? We make a weighted sum of the former templates, based on the spike counts. You want to recompute everthing instead?

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -86,7 +86,26 @@ def _select_extension_data(self, unit_ids):
...
+        new_metrics = self._compute_metrics(new_sorting_analyzer, new_unit_ids, verbose, **job_kwargs)
+        for unit_id in new_sorting_analyzer.unit_ids:
+            if unit_id in new_unit_ids:
+                source = new_metrics
+            else:
+                source = old_metrics
+            for k in source.loc[unit_id].keys():
+                metrics.loc[unit_id, k] = source.loc[unit_id, k]
```
samuelgarcia, line 104: same comments that template metrics

src/spikeinterface/postprocessing/tests/common_extension_tests.py
yger: Yes, we could/should extend that

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
yger: Same response :-). At least I've made a copy() also here of former metrics

yger:
Indeed, we should add more tests, I'll try to make some more. This would be great to test all extensions. And about documentation, where should it go?

src/spikeinterface/core/analyzer_extension_core.py
samuelgarcia: This should be set in the run

yger: Ok, then I'll let the two of you do that. But this means that if the extension is not runed, this params can not be obtained

src/spikeinterface/core/analyzer_extension_core.py
samuelgarcia: done

src/spikeinterface/core/analyzer_extension_core.py
alejoe91: done

src/spikeinterface/core/analyzer_extension_core.py
alejoe91: done

src/spikeinterface/core/sortinganalyzer.py
samuelgarcia: done

src/spikeinterface/core/sortinganalyzer.py
samuelgarcia: We finally `merge_unit_groups`

zm711:
I'm also willing to do some real world tests with one of my labmates as well. Just ping me when this PR is basically ready (based on comments it looks like there are a few more details being ironed out) and she and I can install from this PR to test.

src/spikeinterface/postprocessing/template_metrics.py
samuelgarcia: done

src/spikeinterface/postprocessing/unit_locations.py
samuelgarcia: Yes I think we should recompte the locations only for merges instead of make weithed average

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
samuelgarcia: done

samuelgarcia:
@yger @alejoe91 : I proposed a simpler sparsity re alegnement for swaveforms and pca.
Can you check it ?

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -91,7 +91,7 @@ def _merge_extension_data(
     ):
         import pandas as pd
 
-        old_metrics = self.data["metrics"]
+        old_metrics = self.data["metrics"].copy()
```
samuelgarcia, line 94: no this is useless!

src/spikeinterface/postprocessing/spike_locations.py
```diff
@@ -95,8 +95,10 @@ def _select_extension_data(self, unit_ids):
     def _merge_extension_data(
         self, merge_unit_groups, new_unit_ids, new_sorting_analyzer, keep_mask=None, verbose=False, **job_kwargs
     ):
-        new_spike_locations = self.data["spike_locations"]
+        
         if keep_mask is not None:
+            new_spike_locations = self.data["spike_locations"].copy()
+        else:
             new_spike_locations = new_spike_locations[keep_mask]
```
samuelgarcia: you iverted the is None and new_spike_locartion must be self.data

yger:
We should have a quick call @samuelgarcia , because you've messed up some internal structures for the merges. I can patch it, but we should discuss before

samuelgarcia:
> We should have a quick call @samuelgarcia , because you've messed up some internal structures for the merges. I can patch it, but we should discuss before

OK

yger:
All tests are fixed, I uncommented the get_some_waveforms() and get_some_projections() function just in case. They are not used anymore in merges (there is the inplace modification made by @samuelgarcia ), but they were properly tested, and they could still be useful for users I guess. I've also fixed the issue with indices, we need the cumsum and the substraction when doing the merges

yger:
No idea why the qm tests on d_prime is not passing.... The functions returns NaN without obvisous reasons

zm711:
Quick user question before I test this in general.

 Is the idea that if I have a binary folder or zarr analyzer that merging units will create a new analyzer where I need to specify a new folder or will it kill the old analyzer or will it return an analyzer in memory that I need to save to a new folder.

zm711:
Perfect thanks @yger.

src/spikeinterface/postprocessing/tests/common_extension_tests.py
samuelgarcia: big test will be implemented in postprocessing/test_multiextension

yger:
Thanks for the tests! Actually I made something very similar, but @samuelgarcia told me not to push anything so I was waiting for his green light.... But it will barely be ready for the release

alejoe91:
@samuelgarcia curation docs updated! Ready to merge on my end

alejoe91 (approved):

Issue #3031, opened 2024-06-14
`SortingAnalyzer` issues


khl02007:
@alejoe91 @samuelgarcia 
I just started trying the new sorting analyzer and wanted to share some issues that I have encountered:

- It seems that every time you run `sorting_analyzer.compute`, you recompute all the extensions. I think that it would be better if it first checks if the extension exists with the same parameters and not re-do it if possible. Maybe there could be a `recompute=True` parameter that you can pass.
- Some of the extensions depend on other extensions. For example, computing PC metrics requires the principal components extension. But the parameters passed may differ. For example, `principal_components` extension could be computed with `n_components=5`, but a PC metric that depends on it may ask for `n_components=10`. In that case it seems like the user should get at least a warning, or a error message saying that the parameters are not compatible. 
- Another idea is for the user to just be able to compute the quality metrics from the sorting analyzer, and the sorting analyzer to silently compute the extensions if they don't exist. I prefer this, as I don't really want to think about what kind of extensions are required for me to get my metrics.
- In general from the user's perspective it is very opaque as to how the extensions depend on one another. Perhaps this can be better documented.
- I feel that `si.load_sorting_analyzer` should have `load_extensions=False` by default (rather than load all the extensions by default, which could take a long time)
- I'm not sure how I can add / attach a recording back to a sorting analyzer. I created a sorting analyzer with a sorting and a recording, and computed some extensions. Then later I tried to load the sorting analyzer and compute more extensions. But when I loaded the sorting analyzer with `load_sorting_analyzer` and tried to compute the extensions, I got an error saying that it requires a recording and the sorting analyzer doesn't have one. I can load the recording independently, but I don't know how it can be associated with the loaded sorting analyzer again. Would I 'recreate' the sorting analyzer again with `create_sorting_analyzer` again in that case?

h-mayorquin:
Just a comment, there is documentation for the extension dependency graph but it might not be easily discoverable:

https://spikeinterface.readthedocs.io/en/latest/modules/postprocessing.html#extensions-as-analyzerextensions

khl02007:
@h-mayorquin I see, thanks, should have read that

zm711:
> Another idea is for the user to just be able to compute the quality metrics from the sorting analyzer, and the sorting analyzer to silently compute the extensions if they don't exist. I prefer this, as I don't really want to think about what kind of extensions are required for me to get my metrics.

[...]

PR #3045, opened 2024-06-19
Add force_recompute to SortingAnalyzer.compute

alejoe91:
This is for avoiding recomputing extensions when they are:
- already computed
- have the same parameters
This also avoids deleting dependent extensions by mistake!

A `force_recompute` option is added to  override this behavior

Partially addresses #3031

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -957,8 +957,8 @@ class SortingAnalyzer:
             If not then the extension will only live in memory as long as the object is deleted.
             save=False is convenient to try some parameters without changing an already saved extension.
 
-        **kwargs:
-            All other kwargs are transmitted to extension.set_params() or job_kwargs
+        **kwargs: keyword arguments
+            All other keyword arguments are transmitted to extension.set_params() or job_kwargs
 
         Returns
         -------
@@ -1033,7 +1033,7 @@ class SortingAnalyzer:
             If True, the extension is recomputed even if it has already been computed with the provided parameters.
             If False, the extension is not recomputed if it has already been computed with the provided parameters.
         **job_kwargs : keyword arguments
-            All other kwargs are transmitted to job_kwargs
+            All other keyword arguments are transmitted to job_kwargs
 
         Returns
         -------
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -887,7 +887,7 @@ class SortingAnalyzer:
             If True, the extension is recomputed even if it has already been computed with the provided parameters.
             If False, the extension is not recomputed if it has already been computed with the provided parameters.
         **kwargs:
-            All other kwargs are transmitted to extension.set_params() (if input is a string) or job_kwargs
+            All other keyword arguments are transmitted to extension.set_params() (if input is a string) or job_kwargs
 
         Returns
         -------
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -864,7 +864,7 @@ class SortingAnalyzer:
         return self.sorting.get_num_units()
 
     ## extensions zone
-    def compute(self, input, save=True, extension_params=None, verbose=False, **kwargs):
+    def compute(self, input, save=True, extension_params=None, verbose=False, force_recompute=False, **kwargs):
         """
         Compute one extension or several extensiosn.
         Internally calls compute_one_extension() or compute_several_extensions() depending on the input type.
@@ -883,6 +883,9 @@ class SortingAnalyzer:
         extension_params : dict or None, default: None
             If input is a list, this parameter can be used to specify parameters for each extension.
             The extension_params keys must be included in the input list.
+        force_recompute : bool, default: False
+            If True, the extension is recomputed even if it has already been computed with the provided parameters.
+            If False, the extension is not recomputed if it has already been computed with the provided parameters.
         **kwargs:
             All other kwargs are transmitted to extension.set_params() (if input is a string) or job_kwargs
 
@@ -912,11 +915,15 @@ class SortingAnalyzer:
         )
         """
         if isinstance(input, str):
-            return self.compute_one_extension(extension_name=input, save=save, verbose=verbose, **kwargs)
+            return self.compute_one_extension(
+                extension_name=input, save=save, verbose=verbose, force_recompute=force_recompute, **kwargs
+            )
         elif isinstance(input, dict):
             params_, job_kwargs = split_job_kwargs(kwargs)
             assert len(params_) == 0, "Too many arguments for SortingAnalyzer.compute_several_extensions()"
-            self.compute_several_extensions(extensions=input, save=save, verbose=verbose, **job_kwargs)
+            self.compute_several_extensions(
+                extensions=input, save=save, verbose=verbose, force_recompute=force_recompute, **job_kwargs
+            )
         elif isinstance(input, list):
             params_, job_kwargs = split_job_kwargs(kwargs)
             assert len(params_) == 0, "Too many arguments for SortingAnalyzer.compute_several_extensions()"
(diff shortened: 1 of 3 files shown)


src/spikeinterface/core/tests/test_analyzer_extension_core.py
```diff
@@ -218,7 +218,7 @@ def test_delete_on_recompute(create_cache_folder):
     sorting_analyzer.compute("templates")
 
     # re compute random_spikes should delete waveforms and templates
-    sorting_analyzer.compute("random_spikes")
+    sorting_analyzer.compute("random_spikes", force_recompute=True)
```
alejoe91, line 221: this triggers recompute and deletes children extensions

zm711:
Seems tests are failing at the moment :)

alejoe91:
Indeed, the test was wrong :)

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1015,6 +1027,13 @@ def compute_several_extensions(self, extensions, save=True, verbose=False, **job
             It the extension can be saved then it is saved.
             If not then the extension will only live in memory as long as the object is deleted.
             save=False is convenient to try some parameters without changing an already saved extension.
+        verbose : bool, default: False
+            If True, print more information about the computation.
+        force_recompute : bool, default: False
+            If True, the extension is recomputed even if it has already been computed with the provided parameters.
+            If False, the extension is not recomputed if it has already been computed with the provided parameters.
+        **job_kwargs : keyword arguments
+            All other kwargs are transmitted to job_kwargs
```
zm711, line 1036:
```suggestion
            All other keyword arguments are transmitted to job_kwargs
```

I still have a slight preference to say keyword arguments here because even though kwarg is well known in the python world, it is not apparent to new python users (ie a big chunk of users). So giving people a chance to kwarg and keyword argument together will help them make the connection in the future.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1108,6 +1142,25 @@ def compute_several_extensions(self, extensions, save=True, verbose=False, **job
...
+            if extension_class.need_job_kwargs:
+                params, job_kwargs = split_job_kwargs(kwargs)
+            else:
+                params = kwargs
+
+            extension_instance = extension_class(self)
+            extension_instance.set_params(**params)
+
+            loaded_extension = self.get_extension(extension_name)
+            if loaded_extension.params == extension_instance.params:
```
zm711, line 1159: Does this check always work? I didn't realize you could do == for a dict? Or am I misunderstanding this?

zm711 (commented):
One plea for a docstring although I could be overridden on pythonic-ness.

And then a question on the check itself.

src/spikeinterface/core/sortinganalyzer.py
alejoe91: it seems so :)

samuelgarcia:
I am not super happy with this!
For me this is dangerous.
At the moment when you compute something it force the recompute always and this is a good behavior I think.

Are we sure that when do this in a chain where steps has dependencies that every child nodes is recompute if a parent has change parameters ? This should be tested. I am not seeing in the code a parent-child checking for this.
In short if you recompute random_spike with new params and waveforms with same params both should be computed.

And also I am strongly/totaly disagree with the default being False.

To be honest I do not see when we need this.

alejoe91:
Why is this dangerous? If the params are the same in my opinion we shouldn't recompute. 

The dependency deletion is kept. Each compute function will force the deletion of its children if recompute is triggered. Chris added a test for it.

src/spikeinterface/core/sortinganalyzer.py
chrishalcrow: Yes. One tiny interesting thing. Dictionaries in python are ordered (since 3.6), but their comparison is order-independent. So if
```
a={1:2, 2:4}
b={2:4, 1:2}
```
then
```
a == b
a.keys() == b.keys
```
both return `True`, but 
```
list(a.keys()) == list(b.keys)
```
returns `False`.

src/spikeinterface/core/tests/test_analyzer_extension_core.py
```diff
@@ -218,7 +218,7 @@ def test_delete_on_recompute(create_cache_folder):
...
     assert sorting_analyzer.get_extension("templates") is None
     assert sorting_analyzer.get_extension("waveforms") is None
```
chrishalcrow, line 224:
```suggestion
def test_delete_on_parm_change(create_cache_folder):
    cache_folder = create_cache_folder
    sorting_analyzer = get_sorting_analyzer(cache_folder, format="memory", sparse=False)

    sorting_analyzer.compute("random_spikes", seed=1, force_recompute=False)
    sorting_analyzer.compute("waveforms")
    sorting_analyzer.compute("templates")

    sorting_analyzer.compute("random_spikes", seed=1, force_recompute=False)
    assert sorting_analyzer.get_extension("waveforms") is not None

    sorting_analyzer.compute("random_spikes", seed=2, force_recompute=False)
    assert sorting_analyzer.get_extension("waveforms") is None
    assert sorting_analyzer.get_extension("templates") is None
```

I'd be up for another test here, just to make sure that the `random_spikes` parameter change doesn't delete waveforms when a parameter isn't changed, but does delete it when it is.

samuelgarcia: typically for me random_spike recompute with the same number of spikes and seed=None (so same params) should delete everything but I guess it will not be the case with this PR

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1027,9 +1046,15 @@ def compute_several_extensions(self, extensions, save=True, verbose=False, **job
         >>> sorting_analyzer.compute_several_extensions({"waveforms": {"ms_before": 1.2}, "templates" : {"operators": ["average", "std"]}})
 
         """
+        extensions_to_recompute = {}
+        if not force_recompute:
+            for extension_name, extension_params in extensions.items():
+                if self.does_extension_need_recompute(extension_name, **extension_params):
+                    extensions_to_recompute[extension_name] = extension_params
```
chrishalcrow, line 1053: Hello, there's a problem here.
If I do something like:
```
sorting_analyzer.compute(["random_spikes", "waveforms", "templates", "spike_locations"])
sorting_analyzer.compute(["random_spikes", "waveforms", "templates", "spike_locations"], 
    extension_params={
        'waveforms': {'ms_before': 1.5},
        'spike_locations' : {'ms_after': 0.4}
    }
)
```
Then it only includes `waveforms` and `spike_locations` in the recomputation. But we should also recompute templates. So you need to include this too. Something like:
```
if self.does_extension_need_recompute(extension_name, **extension_params):
    extensions_to_recompute[extension_name] = extension_params
        for ext_dependency in set(_get_children_dependencies(extension_name)) & set(extensions.keys()):
            extensions_to_recompute[ext_dependency] = extensions[ext_dependency]
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1108,6 +1142,25 @@ def compute_several_extensions(self, extensions, save=True, verbose=False, **job
             else:
                 self.compute_one_extension(extension_name, save=save, verbose=verbose, **extension_params)
 
+    def does_extension_need_recompute(self, extension_name, **kwargs):
+        if not self.has_extension(extension_name):
+            return True
+        else:
+            extension_class = get_extension_class(extension_name)
+            if extension_class.need_job_kwargs:
+                params, job_kwargs = split_job_kwargs(kwargs)
```
chrishalcrow, line 1151: should `job_kwargs` is unused, should it be used somewhere?

chrishalcrow (changes requested):
Need to fix the `compute_several_extensions()` issue. Otherwise looks good!

src/spikeinterface/core/tests/test_analyzer_extension_core.py
chrishalcrow: Yup, for this PR it wouldn't recompute or delete with seed=None, you'd have to turn on `force_recompute=True`. I do think `random_spikes` feels special for this kinda thing, compared to the other extensions. But I'm not sure what the solution should be.

samuelgarcia:
I think we need to discuss the semantic globally here.
I am not against this behavior at al but this should be done a a separated method could be named `load_or_compute()` for me `compute()` should not be a hidden load a cache.

PR #3051, opened 2024-06-19
Propagate FrameSlice default behavior to frame_slice and time_slice

h-mayorquin:
`FrameSlice` sets defaults of 0 and num_samples when start_frame and end_frame are not passed respectively:

https://github.com/h-mayorquin/spikeinterface/blob/bd626b0b4fbf4bfa5cfa68e857fb8ea997784c56/src/spikeinterface/core/frameslicerecording.py#L33-L45

But currently, this behavior is not available in frame_slice. This PR propagates this behavior to both frame_slice and time_slice to make it consistent.


samuelgarcia (approved):

samuelgarcia:
merci.

Issue #2985, opened 2024-06-06
feature request: bad channel exclusion from common_reference function


WeissShahaf:
Dear devs,
I appreciate your hard work. spikeinterface is becoming a really great system for nerual data processing. 

Please consider implementing a bad channel exclusion option for common_reference function.
similiar to catGT's -chnexcl={prb;chans}     ;this probe, exclude listed chans from ap loccar, gblcar, gfix

maybe something like this:
si.common_reference(rec, operator="median",chnexcl=bad_channel_ids)

Thank you,
Shahaf

alejoe91:
Hi @WeissShahaf 

Thanks for the support :)

You can already do that! If you have a list of channels to exclude, you can use the `remove_channels()` function:
```
bad_channel_ids = [...+

recording_clean = recording.remove_channels(bad_channel_ids)
recording_clean_cmr = spre.common_reference(recording_clean)
```
Is this what you had in mind?

WeissShahaf:
@alejoe91 
no, the functionality i'm describing is to exclude channels from the CMR function, but keep them in the recording.

h-mayorquin:
Does this accomplish what you want: @WeissShahaf 

```python
from spikeinterface.core import generate_recording, aggregate_channels
from spikeinterface.preprocessing import common_reference


recording = generate_recording(num_channels=4, durations=[10], seed=0)
recording = recording.rename_channels(new_channel_ids=["a", "b", "c", "d"])
bad_channels = ["b"]
recording_for_crm = recording.remove_channels(remove_channel_ids=bad_channels)
recordin_bad_chanenls = recording.select_channels(channel_ids=bad_channels)


common_reference = common_reference(recording_for_crm, reference='global')

recording_where_some_channels_are_excluded_from_crm = aggregate_channels([common_reference, recordin_bad_chanenls])
recording_where_some_channels_are_excluded_from_crm.get_traces()

```

alejoe91:
@WeissShahaf fixed here #3139, but with a slight different design:

You can optionally pass a list of channels to compute the median on with the `ref_channel_ids` existing argument

h-mayorquin:
To be more precise you would do

```python
from spikeinterface.core import generate_recording
from spikeinterface.preprocessing import common_reference


recording = generate_recording(num_channels=4, durations=[10], seed=0)
recording = recording.rename_channels(new_channel_ids=["a", "b", "c", "d"])
good_channels = ["a", "b", "c"]


[...]

Issue #2958, opened 2024-06-02
[Feature request] Add return_scaled option for spikeinterface.core.write_binary_recording


jnjnnjzch:
Hello, I'm using spikeinterface to convert my data from openEphys to a binary file for further analysis.
I found that the `spikeinterface.core.write_binary_recording()` function just directly write the original data into binary without any scale factors such as `gain_to_uV` or `offset_to_uV` which can be read from `recording`. 

From the view of the project, it is necessary not to change the raw recording data, and I compeletely understand why spikeinterface keeps the raw data as it was.
However, from the persepective of a user, when I call the `write_binary_recording()` function, not seeing a `return_scaled` option makes me feel like the function has done it automatically, especially I can use the sorters naturally without considering if the data scale is returned or not.

It did confuse me for couple of days figuring out why the converted binary data is "wrong". Now, I'm using a code snippet to deal with it, which is not so elegent.
```python
info = recording.to_dict()['properties']
gain_to_uV = info['gain_to_uV'][0]
offset_to_uV = info['offset_to_uV'][0]
recording = spre.scale(recording, gain=gain_to_uV, offset=offset_to_uV)
si.write_binary_recording(recording, file_paths=file_path, dtype=dtype, **job_kwargs)
```

For the sake of user experience, would you consider adding a `return_scaled` option for  the `write_binary_recording()` function? If it's inconvenient for you, I can learn to submit a PR since I'm new to GitHub.

h-mayorquin:
I want to keep the function `write_binary_recording` with as few arguments as we can because it is easier to test and optimize. The reason for this is that it is a very central function in the library.


```python

gain_to_uV = recording.get_channel_gains()
offset_to_uV = info['offset_to_uV'][0]
recording = spre.scale(recording, gain=gain_to_uV, offset=offset_to_uV)
si.write_binary_recording(recording, file_paths=file_path, dtype=dtype, **job_kwargs)
```

jnjnnjzch:
Hi, @h-mayorquin, sorry for my late reply. Here are my thoughts.

It is quite hard to "define" an elegent code style because such styles can be defined from many perspectives. And I agree with your thought that one should minimize the number of arguments for those core functions and to build an API-provided function via simple function combinations.

[...]

PR #3053, opened 2024-06-20
Add `scale_to_uV` preprocessing

h-mayorquin:
The justification from:

https://github.com/SpikeInterface/spikeinterface/issues/2958#issuecomment-2154947452

> Now that I think about it myself what I would like is to have a preprocessing called scale_to_uV that which allows me to write:
> 
> ```python
> recording_in_uV = spre.scale_to_uV(recording=raw_recording)
> si.write_binary_recording(recording_in_uV, file_paths=file_path, dtype=dtype, **job_kwargs)
> ```
> 
> I think this is both concise while at the same time generalizes to situations beyond `write_binary_recording`. What do you think @jnjnnjzch, do you feel this would make the API more in line with your concept of elegance?
> 
> I actually would like your take on this @JoeZiminski as you have been thinking about preprocessing. My thinking here is that I think is better to combine simple functions and apply them flexibly depending on the situation at hand rather than adding a new parameter to each of the core functions for every situation. The latter also has the downside of making the core functions more difficult to optimize and document. See here for another instance where I think this question arises: #2985. I recognize there might be performance limitations that might limit this approach but I would rather take it as a baseline.


src/spikeinterface/preprocessing/preprocessinglist.py
```diff
@@ -54,6 +56,7 @@
     ScaleRecording,
     CenterRecording,
     ZScoreRecording,
+    ScaleTouV,
```
zm711, line 59: Everything else has Recording in the name. Are you just against that?

h-mayorquin: Not really but I feel the patttern does not fit. 
ScaleRecoding : ScaleTheRecording 
ZScoreRecording: ZScoreTheRecording
CenterRecording: CenterTheRecording

Pattern: Verb(To)Recording

I think the pattern should be (Adective)Recording  like ScaledRecording, CenteredRecording but I guess that shiped has sailed. 

Would you like me to add Recording? I have the vague feeling that is marginally better to be consistent.

zm711: It is just hard for me to visually parse `ScaleTouV` because micro is the first word so it should be `ScaleToUV`, but then you have the double capital which is also tricky. I'm just hoping that `ScaleTouVRecording` feels better, but it is still a little tricky for me to view. I don't really have a solution, but consistency like you said at least gives us a marginal benefit.

h-mayorquin: Done.

src/spikeinterface/preprocessing/scale.py
```diff
@@ -0,0 +1,46 @@
...
+from spikeinterface.core import BaseRecording
+from spikeinterface.preprocessing.basepreprocessor import BasePreprocessor
+
+
+class ScaleTouVRecording(BasePreprocessor):
+    """
+    Scale raw traces to microvolts (µV).
+
+    This preprocessor uses the channel-specific gain and offset information
+    stored in the recording extractor to convert the raw traces to µV units.
```
JoeZiminski, line 12: What happens if the raw data dtype is `int`, is it changed to float? Is it worth mentioning this here?

src/spikeinterface/preprocessing/scale.py
```diff
@@ -0,0 +1,46 @@
...
+
+    Parameters
+    ----------
+    recording : BaseRecording
+        The recording extractor to be scaled. The recording extractor must
+        have gains and offsets otherwise an error will be raised.
+
+    Raises
```
JoeZiminski, line 20: cool! I did not know about this section

src/spikeinterface/preprocessing/tests/test_scaling.py
```diff
@@ -0,0 +1,36 @@
...
+import numpy as np
+from spikeinterface.core.testing_tools import generate_recording
+from spikeinterface.preprocessing import ScaleTouVRecording  # Replace 'your_module' with your actual module name
+
+
+def test_scale_to_uv():
+    # Create a sample recording extractor with fake gains and offsets
+    num_channels = 4
+    sampling_frequency = 30_000.0
+    durations = [1]  # seconds
```
JoeZiminski, line 11: could the multi-segment case be tested?

src/spikeinterface/preprocessing/tests/test_scaling.py
```diff
@@ -0,0 +1,36 @@
...
+    recording = generate_recording(
+        num_channels=num_channels,
+        durations=durations,
+        sampling_frequency=sampling_frequency,
+    )
+
+    rng = np.random.default_rng(0)
+    gains = rng.random(size=(num_channels)).astype(np.float32)
```
JoeZiminski, line 19: is it worth parameterising over some extreme cases (e.g. very small values, typical values, extremely large values) for gains and offsets?

src/spikeinterface/preprocessing/tests/test_scaling.py
```diff
@@ -0,0 +1,36 @@
...
+    offsets = rng.random(size=(num_channels)).astype(np.float32)
+    recording.set_channel_gains(gains)
+    recording.set_channel_offsets(offsets)
+
+    # Apply the preprocessor
+    scaled_recording = ScaleTouVRecording(recording=recording)
+
+    # Check if the traces are indeed scaled
+    expected_traces = recording.get_traces(return_scaled=True)
```
JoeZiminski, line 28: what happens if `return_scaled` is `False` on the `ScaleTouVRecording?` intuitively I might expect it always returns scaled. Though of course it makes more sense for it only to return scaled to uV when `return_scaled=True`. Maybe this could be added to docstring? and a quick additional test that the `return_scaled=False` case does not change?

src/spikeinterface/preprocessing/tests/test_scaling.py
```diff
@@ -0,0 +1,36 @@
...
+    scaled_traces = scaled_recording.get_traces()
+
+    np.testing.assert_allclose(scaled_traces, expected_traces)
+
+    # Test for the error when recording doesn't have scaleable traces
+    recording.set_channel_gains(None)  # Remove gains to make traces unscaleable
+    with pytest.raises(AssertionError):
```
JoeZiminski, line 35: Maybe it is worth raising a `ValueError` or `RuntimeError`?

src/spikeinterface/preprocessing/scale.py
```diff
@@ -0,0 +1,46 @@
...
+    ------
+    AssertionError
+        If the recording extractor does not have scaleable traces.
+    """
+
+    name = "scale_to_uV"
+
+    def __init__(self, recording: BaseRecording):
+        assert recording.has_scaleable_traces(), "Recording must have scaleable traces"
+        from spikeinterface.preprocessing.normalize_scale import ScaleRecordingSegment
```
JoeZiminski, line 30: A more general question out of scope for this PR. Often I get confused about imports not at the top of the file and have an impulse to move them to the top, even if this is very wrong. However I am still always interested in why the import is not at the top, I guess the two reasons can be 1) speed up imports 2) avoid circular imports. Do you think it is in general worth adding comments with the reason for placing such imports or is this too much information / clutter?

src/spikeinterface/preprocessing/tests/test_scaling.py
```diff
@@ -0,0 +1,36 @@
+import pytest
+import numpy as np
+from spikeinterface.core.testing_tools import generate_recording
+from spikeinterface.preprocessing import ScaleTouVRecording  # Replace 'your_module' with your actual module name
```
JoeZiminski, line 4: stray comment

JoeZiminski (commented):
This is great! I'm constantly getting confused about how scaling and dtype is handled and this is very straightforward. A short how-to or tutorial would be useful in future (I could add suggestion to documentation thread) which could exemplify use of this preprocessor.

src/spikeinterface/preprocessing/scale.py
h-mayorquin: Yes, it does, I think is worth it.

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: Yep. I will add that.

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: Mmm do you have some suggestions of what would be the purpose more specifically? I agree with this philosophy (e.g. "think about what will break the test") but nothing concrete comes to mind other than testing overflow problems. Let me think more but if you come with anything it would be good.

I would not like to stop the PR if we don't come with good extreme case criteria though : P

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: This is a good idea and something that I did not think before.

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: Mmm ValueError is when you pass values that are incorrect? What is the advantage of RuntimeError over the generic one? 

I agree that it might be a more specific exception but I can't think on the one that would be more appropriate, how are you thinking about this?

src/spikeinterface/preprocessing/scale.py
h-mayorquin: Yes, in order of relevance:
1) Modularization: that is, people do not have that module (think that pynwb is only required for a small section of the library so most people who use the library don't have it) and we want to avoid an improt error.
2) Improve speed ups like with numba, scipy, etc.
3) Avoid circular imports. This case. 

I can add a comment if you think this would be helpful here.

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: Indeed.

src/spikeinterface/preprocessing/scale.py
zm711: If you scale they always have to be float, no matter what right? For neo our scale analog chunk function allows for any dtype with a default of float.

src/spikeinterface/preprocessing/tests/test_scaling.py
zm711: Shouldn't we raise an error instead. scaling to uv necessarily requires us to return_scaled. If the user is not okay with that, then they shouldn't use this class? Why should we let them @JoeZiminski ? What use-case?

zm711:
@alejoe91 ,

yeah I mean that also makes sense to me. And way less surface area to have to test and maintain, which I like @h-mayorquin likes too!

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: Mmm the thing is that return_scaled is False by default. Having to set `return_scaled=True` even when passing this class defeats the purpose of using it I feel.

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: I can change to run time error sure but I think  the message is the important thing not the error in this case.

src/spikeinterface/preprocessing/scale.py
JoeZiminski: Thanks Herberto I think in this case it is probably out of scope for the PR, I was just kind of raising it as a general thought but with no real purpose 😅 I think it's okay to leave for now and can raise it more generally another time.

src/spikeinterface/preprocessing/scale.py
JoeZiminski: Yes I think I am generally misunderstanding how the scaling works, I will ask more clarifying questions 😅 (thanks for your patience, I promise this will be useful as will write a docs article :D) :

Are the gains and offsets always to uV? (rather than say V). Like I thought it could be possible that you have an integer recording and the associated gains and offsets are to V, so when you do `return_scaled=True` you get the recording in V. Then one thing this might do is change the recording such that it is still integer but when `return_scaled=True` it returns as `uV` rather than `V`.

Instead, (I believe) the purpose of this to change the raw datatype such that the integer data is set to float, and scaled by the scaling factors, and then the gain and offset are changed to 1 and 0 forever more? and `return_scaled=True` and `return_scaled=False` have the same behaviour?

If the original recording is a float, but also has some gains and offsets such that `return_scaled=False` and `return_scaled=True` give different output (the latter also float but scaled to `uV`), this will simply scale the raw data to `uV` also? (and set gain 1 offset 0).

src/spikeinterface/preprocessing/tests/test_scaling.py
JoeZiminski: My understanding is that the convention is for anything user-facing to be raised as a full Error, any internal checks (that could be removed anytime or replaced by a unit test) are `assert`. But agree the message is the main thing. 

Agree `ValueError` is not good, `RuntimeError` is okay, maybe `AttributeError`?

src/spikeinterface/preprocessing/tests/test_scaling.py
JoeZiminski: Oh yes, if `return_scaled=False` had to be set explicitly I would agree Zach in this case though I guess we let it slide as it is the default. 

If I understand everything correctly, maybe a test can just check the output of `return_scaled=True` and `return_scaled=False` are identical?

src/spikeinterface/preprocessing/scale.py
h-mayorquin: I like discussing. I do this random question in PRs myself. Feel free to ask whatever!

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: I would like to expand on that not because I feel we disagree but because I have seen this on many places.

Yes, you find that advice in many places.

But I really don't think it matters.

Reason is that:
*  asserts are said to be bad because they will be ommited if people run python with the -o flag.
* BUT, removing assertions IS THE ONLY THING that the -o flag does.

That behavior is locked and the developers don't really intend to change it, check Guido's Van Rossum take here:
https://discuss.python.org/t/stop-ignoring-asserts-when-running-in-optimized-mode/13132/8

So, in practice the behavior of asserts as other exceptions is kind of frozen and you end up with a matter of synatx preference. People love one liners and that's why you end up with a bunch of assertions all the over place.

I think that unless used in a try-exept block (you are trying to catch up something), focusing on the type of error or assertion is a red hering. What matters the most for the final user is a clear actionable error message and there is where the focus should be.

src/spikeinterface/preprocessing/scale.py
h-mayorquin: Yes, by design we write gains and scales so when you multiply by the scale and add the offset you end up with units in uV. Always uV.

No, this preprocessor should not alter the gains and offsets of the preprocessed recording. It just does the scaling at some point of the chain so you can be sure that anything you do either will be in "natural" units in opposition to "storage" units.

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: What I meant above when I have to think more is that the real problem is if someone calls `return_scaled=True` then the scaling would be done twice. I need to think about this.

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: My idea was that the random gains kind of cover for all the common use cases save the stress (outlier / weird) part. For the overflow problem I think we could think on someting but I am leaning on testing overflow in a more general check. I actually do feel that that one needs generalization to be useful.

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: This is critical in cases of things like the sorting_analyzer (which I think is unique?) that have `return_scaled=True`.

src/spikeinterface/preprocessing/tests/test_scaling.py
zm711: I find this comment interesting (though I don't know the internals). 

> Part of the problem with assert is the fact that a failed assert does not abort the process. It’s too easy and too common to catch and ignore the result of a failed assert. Python’s assert should really work more like C assert. In C a failed [assert(3)](https://man7.org/linux/man-pages/man3/assert.3.html) results in an abnormal termination of the program. In Python assert False raises an AssertionError, which is a subclass of Exception. We should have changed the base class to BaseException or treat a failed assert as unrecoverable error (e.g. abort with Py_FatalError).

I mean one argument I would have for a better error type would be your plexon PR right now. For example if you saw that as a IOError you could make the choice to reattempt because it is an IO problem (maybe bad connection, etc). But an AssertionError doesn't tell you as the developer the domain of the problem.

I think we all agree the messaging is the most important part, especially for the end-user.

That being said I also think it is lower priority to switch these in this library and much much higher priority to add assert messaging for the end-user.

src/spikeinterface/preprocessing/scale.py
zm711: I think a comment like this:

> Modularization: that is, people do not have that module (think that pynwb is only required for a small section of the library so most people who use the library don't have it) and we want to avoid an improt error.
Improve speed ups like with numba, scipy, etc.
Avoid circular imports. This case.

is super useful and I know in a month or 6 months something like this will come up and I won't be able to find this very useful comment to link to :) and I'll have to re-explain that exact same thing in a (no doubt) less succinct way.

thanks Heberto!

src/spikeinterface/preprocessing/scale.py
zm711: I think this is in our documents somewhere, but this would be another great point to add to a glossary. Anything scaled will be in uV. It's a choice the library has made so any other units have to be accomplished by the end-user.

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: Yes, that's wnat I meant above that the type matters if you are a in try-except block trying to catch the error.  But you really need to have a good type for your error for this to work!

src/spikeinterface/preprocessing/scale.py
h-mayorquin: : D
I am happy to hear it was useful for you!

src/spikeinterface/preprocessing/tests/test_scaling.py
zm711: I see. Then I agree with you :)

src/spikeinterface/preprocessing/scale.py
h-mayorquin: Yes!

samuelgarcia:
I am not against this but I think that write_binary_recording() with a option scale to micro volt would be easier for end user no ? In short I am not sure to share the starting point of this.

h-mayorquin:
1) This solution can be used anywhere where a user needs it. If a need to scale to uV a function other than `write_binary` appears, we can point them out to this instead of adding another parameter to every function. Teach a man how to fish instead of giving him fish kind of thing.
2) The return_scale works post segment traces scaling, this one, like scaling, works before. In other words, you can do your operation in natural units if that is whath you need.

Easier to mantain instead of adding parameters everywhere and more general.

JoeZiminski:
After having got stuck / confused trying to write a recording recently, I'm now +1 on adding `return_scaled` as an argument to `write_binary_recording()` (and any other relevant writer functions). 

The reasons are:
1) keeps with the `get_traces()` convention that whenever you are 'accessing' the data, you have the option to return scaled or not. At the moment the behaviour is ambiguous for writer functions because this pattern is broken. 
2) it is maximally convenient for the user (i) the option is very easy to see as a function arg rather than behaviour hidden further down in docstring ii) have to read about and understand behaviour of a new function (which is not necessarily clear, considering the complexities of scaling and dtype).

But I think that and this PR are mutually exclusive I still think `scale_to_uV` is a great addition that should be included.

src/spikeinterface/preprocessing/tests/test_scaling.py
JoeZiminski: Definitely agree in general, I do think there is intrinsic value in the user-facing vs. developer distinction. If all errors are `asserts` it is not clear reading the code if an error is being raised because the user passed a wrong value or because we are checking some internal log that, if broken, might have severe consequences / indicates a bug. This can be inferred from a good comment but is easier to see with this convention. Anyways here I don't mind, it's a broader discussion for another day :D!

src/spikeinterface/preprocessing/tests/test_scaling.py
JoeZiminski: Interesting yes, in this cause should the gain and offset be set to 1 and 0? Is there any other workaround except `return_scaled=True` raising an error? Would there be any issue in changing these apart from lost provenance about the original recording? I think it is 'accurate' in some sense (the gain and offset to uV is really now 1 and 0 after `scale_to_uV`).

src/spikeinterface/preprocessing/tests/test_scaling.py
JoeZiminski: Yes agree it's not super important to check the overflow errors in sporadic tests and would be better centralised somewhere. I like the rand but isnt rand bounded [0, 1] and gains can in theory be (0, (some large number?]? For me it seems most efficient rather than guessing bounds on realistic values just to test very small, commonplace, very large number and for sure cover all cases.

zm711:
I'm currently slight on Heberto's side. In that this makes the step explicit which puts the onus on the user to know they are doing this and I also hate when functions have 1000 arguments especially when they start getting super mutually exclusive. So I prefer the unix philosophy of small tools do one things and string the tools together. (I also never write scaled binary so I'm not the target audience).

h-mayorquin:
@JoeZiminski 
> After having got stuck / confused trying to write a recording recently, I'm now +1 on adding return_scaled as an argument to write_binary_recording()

Can you describe more about your confusion? I think people should not be using `write_binary_recording` to save their data. It loses a lot of information.  I am fine with adding `return_scaled` it in a bunch of places (like it is on the widgets for example) but there are core functions like `write_binary_recording`, `to_dict`,  etc that I want to keep as simple as possible because they are used in way too many places. I want to keep complexity out of the core and close to the application interface for two reasons: 1) the core is simpler to optimize, document and mantain, 2) the errors make sense and are easier to debug when they surface close to the application layer.

But maybe I am wrong about the purpose of  `write_binary_recording` and it should be user facing. The user in #2958 proposed that we can have a public `write_binary_recording` that is user facing with a lot of bell of whistles and I think that is fine idea. If the powers that be (@alejoe91 and @samuelgarcia ) decide that it is the way to go I would like to have a private `_write_binary_function` that does not have those bells and whistles and is used on core. The function in this PR enables that workflow.

> 2\. ) applying scaling requires having to read about and understand behaviour of a new function (which is not necessarily clear, considering the complexities of scaling and dtype).

Don't the same complexities appear when using `return_scaled`? That is, this, these two should do the same:
```python
recording_in_uV = scale_to_uv(recording)
some_other_function(scaled_recording, **kwargs)
````

```python
some_other_function(scaled_recodring, return_scaled=True, **kwargs)
```

What is a complexity that appears on the first that does not appear on the second?

> it is maximally convenient for the user 

We can argue about which one a better API (I prefer fluents API) and that's fine but the argument for this is the following 


* With the kwargs approach you need to document that everywhere. This is simpler than job_kwargs but look at how that looks to see the difficulties. Plus, as per the discussion in #2958  `return_scaled` is not really an intuitive parameter either.
* If we forgot it to put it somewhere then the users have no recourse until we decide to add it and enable it there.
* Finally, the function proposed here allows you to operate in the preprocessing phase in physical units. This workflow can't be enabled with `return_scaled` which only does the scaling after all the preprocessing pipe is ran. 

The later point is crucial for the following reason:
We have taken advantage of the fact that most formats use linear scaling and thanks to magic of linear algebra a lot of preprocessing can be done in the raw format and then scale in the end.  This workflow might have  some rough corners like the numeric acrobatics with z-score recording and all the rounding that we do so histograms are calculated like we expect but it works and is a bit cheaper memory wise in some places. But this is an accident, some formats do not have linear conversion between raw and units representation and you can't not operate on the raw space and hope your results to hold. One example is the EDF format that uses a polynomial interpolation. Finally, I personally think that is easier to reason about preprocessing when is units of uV, I don't have to be concerned about rounding or whether the result will mean what I think (does a PCA the waveforms in raw has the same statistical properties than a PCA of the waveforms in natural units for example). The cost is more memory, but I rather understand something first, know what to expect, and then optimize memory and run the pipeline in raw data once I am done and I want to fine tune performance. 

Fuck I write a lot. Sorry.

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: OK, but I don't think there is a good error here. I will change to `RunTimeError` but maybe what we need here is an error of our own if that's the goal. Will change this and mark this as resolved.

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: OK, to move this forward I propose the following:
1) You give me the numbers you want to test.
2) I increase the bounds of float to be very large.

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: Thanks for pushing on this. I think that yes, the gains and offsets are propagated in preprocessing chains here:

https://github.com/h-mayorquin/spikeinterface/blob/a98e81af11f06c0ee47338b42e3132964f558d76/src/spikeinterface/preprocessing/basepreprocessor.py#L25-L26

I think we should modify the gains and offsets so any step down the chain should get 1 and 0s as gains and offsets respectively. The provenance is not lost because the dict still keeps the properties at the parent levels. Let me think abit more about it.

JoeZiminski:
Thanks @h-mayorquin I'm glad you write a lot that was super interesting! I think we are largely in agreement and problems that with the clarity of `write_binary_recording` can be solved in other ways instead of adding `return_scaled`. To summarise:

1) I definitely think this is a useful PR and also prefer the way of working you describe. I think whatever happens with `write_binary_recording` this should be merged. 
2) The problem I was trying to solve with the `return_scaled` to `write_binary_recording` is that at first, the data-scaling issue is very confusing and to understand properly requires quite a deep knowledge of how SI represents data. Often it is not feasible to properly understand this in one sitting and requires some familiarity with the codebase to use the `scale` functions. As such having an easy-to-interpret function argument is nice so it is explicit to newer users how their data is handled.
3) However, I think the above can be solved by as you say finessing the API to make it super-clear to users what to use. A big problem is that `recording.save()` is not properly listed in the API docs, for me still I am not sure the recommended way to save data. So I would agree make `write_binary_recording` private and have a super-clear, well documented workflow for saving data (however that may look).

I will stop bogging down your PR with this discussion and we can continue elsewhere, maybe #2958 can prompt a more general documentation / refactoring of the data-saving pipeline. LGTM!

h-mayorquin:
I have been convinced since https://github.com/SpikeInterface/spikeinterface/issues/2958 that we need an IO tutorial.

jnjnnjzch:
How about creating a new function in `spre` or `sw` etc? I mean, **let the core be the core**. Calling a `spikeinterface.core` function explicitly in the pipeline **should not be encouraged** anyway. 

So here is my suggestion: create another function like `save_to_binary` (which hasa consistent naming format with the existing 'save_to_folder') somewhere else. This function contains the `return_scale` or `return_to_uV` argument. Additionally, the documentation might need to be changed to discourage direct calls to `write_binary_recording`.

The final code snippet could be:
```python
recording.save_to_binary(return_scale=True, **kwargs)
```
Maybe this code is convenient for the user and keeps with the `get_traces()` convention? @JoeZiminski 
Also in this way there is no need to change the `write_binary_recording` function and keeps its purity... @h-mayorquin 

P.S. I'm not sure if I should write in PR comments since it seems like a developer's discussion area. If it bothers, sorry.

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: Added. Marking this as resolved.

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: Removed. Thanks.

h-mayorquin:
@JoeZiminski feel free to keep making comments, man.  I discovered an error by changing the assertion to an error and your comments about how to handle double scaling have been very useful. I don't think I would have noticed that problem as quick as I did thanks to you. And I think your solution is the one that will work. I am thankful for your feedback.

@jnjnnjzch 
No, you are welcome, thanks for your feedback. You made us think about it. Basically now I think we:
1) **might need to rethink** the naming of `return_scaled` which makes sense when is close to `get_traces` but makes less sense far away from it (like in the widgets). Maybe something like `data_in_uV` but we need to think about it and definitley document the behavior better.
2) Add an IO tutorial. 
3) The `save_to_folder` uses `write_to_binary` and has had this TODO since forever to change the name to `save_to_binary_folder`:

https://github.com/h-mayorquin/spikeinterface/blob/62485d5ee5c9dbb30085b3eaad9ea07a448d2ace/src/spikeinterface/core/base.py#L848-L856

zm711:
Thanks @jnjnnjzch I also appreciate people commenting. Sometimes this is a space for *future developers*  ;)

I just want to clarify one thing you said:

> I mean, let the core be the core. Calling a spikeinterface.core function explicitly in the pipeline should not be encouraged anyway.

We actually very much encourage using the spikeinterface.core module as it has the core components. You're right that we need to move more functions to private status to make it clear the external API from the internal, but the module itself is completely public!

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: Done in the last commit.

src/spikeinterface/preprocessing/tests/test_scaling.py
h-mayorquin: OK, I added this, the provenance is mantained because the gains and offsets are copied from parent on further forward on the chain which is exactly what we want. I added a test that guarantess that return_scaled=True has no effect once gain_to_uV is part of the preprocessing chain.

h-mayorquin:
Ok, two changes:
* I have a test to handle the case of users using `return_scaled=True` . This preprocessor is not affected by it and the result is as expected.
* I changed the definition from a class to a function as @alejoe91 suggested.

alejoe91 (approved):
Awesome! LGTM!

src/spikeinterface/preprocessing/tests/test_scaling.py
JoeZiminski: This sounds good, this was a suggestion that I'm not super-strong on and I wouldn't want it to be a blocker! Please feel free to proceed however you think best, personally I would do something like parameterize over small, medium and large number for gain and offset:

```
@pytest.mark.parametrize("gains", [1e-6,, 50, 1e6])
@pytest.mark.parametrize("offsets", [1e-6,, 50, 1e6])
```
I'm not advocating this as necessarily the best approach but it is the kind of habit I have fallen into to test bounds.

JoeZiminski: (to avoid repeatedly generating a recording in such an case, the recording can be generated in a session-scoped fixture)

JoeZiminski (approved):

samuelgarcia (approved):

samuelgarcia:
merci.

PR #3054, opened 2024-06-20
Fix intan kwargs

h-mayorquin:
The `ignore_integrity_checks` keyword that was added recently was not propagated to the kwargs. So running thing in parallel with a context other than fork was not working. This PR fixes it.

This also indicates that there are some missing tests there if this behavior was missing. I will think on how to add a general test to see that al kwargs are catched in another PR.

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/intan.py
+++ b/src/spikeinterface/extractors/neoextractors/intan.py
@@ -53,12 +53,8 @@ class IntanRecordingExtractor(NeoBaseRecordingExtractor):
         )
 
         self._kwargs.update(dict(file_path=str(Path(file_path).absolute())))
-        import packaging
-        import neo
-
-        neo_version = packaging.version.parse(neo.__version__)
-        if neo_version > packaging.version.parse("0.13.1"):
-            self._kwargs.update(dict(ignore_integrity_checks=ignore_integrity_checks))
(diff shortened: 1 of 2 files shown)


samuelgarcia (approved):

src/spikeinterface/extractors/neoextractors/intan.py
```diff
@@ -53,6 +53,12 @@ def __init__(
         )
 
         self._kwargs.update(dict(file_path=str(Path(file_path).absolute())))
+        import packaging
+        import neo
+
+        neo_version = packaging.version.parse(neo.__version__)
+        if neo_version > packaging.version.parse("0.13.1"):
+            self._kwargs.update(dict(ignore_integrity_checks=ignore_integrity_checks))
```
samuelgarcia, line 61: This should be done in line 45 also no ?

h-mayorquin: It is already handled inside the mapping function:

https://github.com/SpikeInterface/spikeinterface/blob/5ae3a24a467c821d7eb7a5be61944874d27740ca/src/spikeinterface/extractors/neoextractors/intan.py#L70-L75

h-mayorquin:
Tested for my own data with forkserve.

PR #3058, opened 2024-06-20
Remove duplicate function from common test suite.

JoeZiminski:
This PR removes a duplicate-named function from the `postprocessing` common test class. All tests pass, I think this function was just overridden by the below version and was silent.


zm711:
It's saying it can get your branch for some of the tests. I have no clue what that means.

JoeZiminski:
Thanks Zach that was really wild! I've never seen that before either. Luckily the classic 'do nothing and hope the problem is gone tomorrow' approach worked well this time.

alejoe91 (approved):

Issue #2839, opened 2024-05-13
error with running KS4 from spikeinterface


paolahydra:
Hi, can you please tell me what is going wrong here?
Trying to call KS4 from spikeinterface, I get the error "slice indices must be integers or None or have an __index__ method".
Please see code and full error report below.

Using:
NPX2.0 (two shanks)
spikeinterface 0.100.6
kilosort 4.0.7

Thanks a lot,
Paola

```
recording_raw = se.read_openephys(folder_path=path, stream_id="{}".format(index), load_sync_channel=False) 
fs = recording_raw.get_sampling_frequency()

# preprocessing steps
recording = st.correct_lsb(recording_raw, verbose=1);
recording = st.bandpass_filter(recording, freq_min=300, freq_max=6000)
recording = st.phase_shift(recording) 
[bad_channel_ids, channel_labels] = st.detect_bad_channels(recording=recording)  
recording = recording.remove_channels(remove_channel_ids=bad_channel_ids)  

grouped_recordings = recording.split_by(property='group')
recgrouplist_hpsf = [st.highpass_spatial_filter(recording=grouped_recordings[k]) for k in grouped_recordings.keys()]  
recording_hpsf = si.aggregate_channels(recgrouplist_hpsf)
print(recording_hpsf)

print("Starting KS4")
t_start = time.perf_counter()
sorting_ks4_prop = ss.run_sorter_by_property("kilosort4", recording=recording_hpsf, grouping_property="group", working_folder=ks_folder, verbose=True)
t_stop = time.perf_counter()
elapsed_prop = np.round(t_stop - t_start, 2)
print(f"Elapsed time by property: {elapsed_prop} s")
```

Error running kilosort4
---------------------------------------------------------------------------
SpikeSortingError                         Traceback (most recent call last)
Cell In[37], line 117
    115 print("Starting KS4")
    116 t_start = time.perf_counter()
--> 117 sorting_ks4_prop = ss.run_sorter_by_property("kilosort4", recording=recording_hpsf, grouping_property="group", working_folder=ks_folder, verbose=True)
    118 t_stop = time.perf_counter()
    119 elapsed_prop = np.round(t_stop - t_start, 2)

[...]

Issue #2871, opened 2024-05-17
Should `get_traces` with `return_scaled=True` when the base dtype is float.


h-mayorquin:
From the discussion inhttps://github.com/SpikeInterface/spikeinterface/pull/2864

We have the following lines:

https://github.com/h-mayorquin/spikeinterface/blob/bc0cbd9f74f30cfed9376f984091cd57f3634169/src/spikeinterface/core/baserecording.py#L369-L373

But if the original dtype is float64 this will downcast. Maybe we should change the code a little bit.

JoeZiminski:
Agree it makes sense if the dtype of the recording is already `float64` to have `get_traces` return `float64` when `get_traces` is used. So overall if I understand SI behaviour correctly the (new) system would be:

Original          Unscaled.         Scaled
`int`                  `int`              `float32`
`float32`       `float32 `        `float32`
`float64 `      `float64 `        `float64 `

h-mayorquin:
Yes, int should be cast to float.

 I think that float should not be cast at all. Maybe float16 if that appears for some reason? Does it appear?

JoeZiminski:
Good point, yes `float16` seems to be [supported](https://github.com/search?q=repo%3ASpikeInterface%2Fspikeinterface%20float16&type=code) and would agree it makes sense that the `dtype` for any float should be preserved between `return_scaled` and `return_unsclaed`.

PR #3059, opened 2024-06-21
Check start_frame/end_frame in `BaseRecording.get_traces()`

alejoe91:
Following up on #3055, this PR moves the logic of checking and setting `start/end_frame` in the `BaseRecording.get_traces` instead of leaving it to the segments. 

This simplifies a lot the `segment.get_traces()`

@JoeZiminski

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -335,8 +335,6 @@ class BaseRecording(BaseRecordingSnippets):
         end_frame = int(min(end_frame, rs.get_num_samples())) if end_frame is not None else rs.get_num_samples()
         if start_frame < 0:
             raise ValueError("start_frame cannot be negative")
-        if start_frame > end_frame:
-            raise ValueError("start_frame cannot be greater than end_frame")
         traces = rs.get_traces(start_frame=start_frame, end_frame=end_frame, channel_indices=channel_indices)
         if order is not None:
             assert order in ["C", "F"]
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -331,7 +331,7 @@ class BaseRecording(BaseRecordingSnippets):
         segment_index = self._check_segment_index(segment_index)
         channel_indices = self.ids_to_indices(channel_ids, prefer_slice=True)
         rs = self._recording_segments[segment_index]
-        start_frame = int(start_frame) if start_frame is not None else 0
+        start_frame = int(max(0, start_frame)) if start_frame is not None else 0
         end_frame = int(min(end_frame, rs.get_num_samples())) if end_frame is not None else rs.get_num_samples()
         traces = rs.get_traces(start_frame=start_frame, end_frame=end_frame, channel_indices=channel_indices)
         if order is not None:
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -332,7 +332,8 @@ class BaseRecording(BaseRecordingSnippets):
         channel_indices = self.ids_to_indices(channel_ids, prefer_slice=True)
         rs = self._recording_segments[segment_index]
         start_frame = int(start_frame) if start_frame is not None else 0
-        end_frame = int(min(end_frame, rs.get_num_samples())) if end_frame is not None else rs.get_num_samples()
+        num_samples = rs.get_num_samples()
+        end_frame = int(min(end_frame, num_samples)) if end_frame is not None else num_samples
         if start_frame < 0:
             raise ValueError("start_frame cannot be negative")
         traces = rs.get_traces(start_frame=start_frame, end_frame=end_frame, channel_indices=channel_indices)
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -331,6 +331,8 @@ class BaseRecording(BaseRecordingSnippets):
         segment_index = self._check_segment_index(segment_index)
         channel_indices = self.ids_to_indices(channel_ids, prefer_slice=True)
         rs = self._recording_segments[segment_index]
+        start_frame = int(start_frame) if start_frame is not None else 0
+        end_frame = int(min(end_frame, rs.get_num_samples())) if end_frame is not None else rs.get_num_samples()
         traces = rs.get_traces(start_frame=start_frame, end_frame=end_frame, channel_indices=channel_indices)
         if order is not None:
             assert order in ["C", "F"]
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -331,8 +331,12 @@ class BaseRecording(BaseRecordingSnippets):
         segment_index = self._check_segment_index(segment_index)
         channel_indices = self.ids_to_indices(channel_ids, prefer_slice=True)
         rs = self._recording_segments[segment_index]
-        start_frame = int(max(0, start_frame)) if start_frame is not None else 0
+        start_frame = int(start_frame) if start_frame is not None else 0
         end_frame = int(min(end_frame, rs.get_num_samples())) if end_frame is not None else rs.get_num_samples()
+        if start_frame < 0:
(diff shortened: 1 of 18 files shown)


JoeZiminski:
This looks good @alejoe91, this should also fix #2839. I don't want to slow this PR down but it has always been my dream to add some `preprocessing` tests that iterate through all available preprocessors and perform some standard checks (e.g. try indexing with `uint`, check `dtype` on output is as expected (e.g. #2871). However it might end up proving very difficult if they require different input arguments. Nonetheless would now be a good time to try and add such a set of tests?

alejoe91:
Yeah we could have a common preprocessing test suite!

@samuelgarcia suggested we go one step further here and rename the `BaseRecordingSegment.get_traces()` to `_get_traces()`.

I agree, but would vote to keep this PR as is (so it can be included easily in #3061) and make a follow up PR.

@JoeZiminski @zm711 @chrishalcrow @h-mayorquin all in favor?

src/spikeinterface/core/baserecording.py
```diff
@@ -331,6 +331,8 @@ def get_traces(
         segment_index = self._check_segment_index(segment_index)
         channel_indices = self.ids_to_indices(channel_ids, prefer_slice=True)
         rs = self._recording_segments[segment_index]
+        start_frame = int(max(0, start_frame)) if start_frame is not None else 0
```
h-mayorquin, line 334: I think you guys wanted to raise an error if start_frame was lower than 0? 

https://github.com/SpikeInterface/spikeinterface/issues/1989

h-mayorquin (approved):
I think this is a a good idea, good that you are casting to ints. 

Let's discuss the use of the hook pattern (_get_traces()) separtely as it is easy to get it wrong I feel.

h-mayorquin:
> This looks good @alejoe91, this should also fix #2839. I don't want to slow this PR down but it has always been my dream to add some `preprocessing` tests that iterate through all available preprocessors and perform some standard checks (e.g. try indexing with `uint`, check `dtype` on output is as expected (e.g. #2871). However it might end up proving very difficult if they require different input arguments. Nonetheless would now be a good time to try and add such a set of tests?


@JoeZiminski check out https://github.com/SpikeInterface/spikeinterface/issues/1989 for things that should be there.

src/spikeinterface/core/baserecording.py
alejoe91: done. Also a check that start frame cannot be greater than end_frame

h-mayorquin: What about the end_frame > num_samples?

alejoe91: Yep, I reread the thread and I'll add that on Monday! Might need to tweak some failing tests then

h-mayorquin: Maybe it would be beter to leave this like this and then add that condition and fix the tests on another PR? 

I also think that you will get failures that will bow you down and this is and improvement already as it is.

alejoe91: Great idea!

samuelgarcia:
I globally OK for this.
Checking in every RecordingSegment was crazy.
But here, we are cliping only at `recording.get_trace` level.
There is no more protection at `recording_segment.get_trace` level.

The best would to have a clip and force to to int also in recording_segment.
The very best solution would to have a get_trace in base recording segment class and then move 
`recording_segment.get_trace` to `recording_segment._get_trace` for all classes.


I will do this after the merge.

src/spikeinterface/core/baserecording.py
```diff
@@ -331,6 +331,10 @@ def get_traces(
         segment_index = self._check_segment_index(segment_index)
         channel_indices = self.ids_to_indices(channel_ids, prefer_slice=True)
         rs = self._recording_segments[segment_index]
+        start_frame = int(start_frame) if start_frame is not None else 0
+        end_frame = int(min(end_frame, rs.get_num_samples())) if end_frame is not None else rs.get_num_samples()
```
alejoe91, line 335:
```suggestion
        num_samples = rs.get_num_samples()
        end_frame = int(min(end_frame, num_samples)) if end_frame is not None else num_samples
```

src/spikeinterface/core/baserecording.py
```diff
@@ -331,6 +331,11 @@ def get_traces(
         segment_index = self._check_segment_index(segment_index)
         channel_indices = self.ids_to_indices(channel_ids, prefer_slice=True)
         rs = self._recording_segments[segment_index]
+        start_frame = int(start_frame) if start_frame is not None else 0
+        num_samples = rs.get_num_samples()
+        end_frame = int(min(end_frame, num_samples)) if end_frame is not None else num_samples
+        if start_frame < 0:
+            raise ValueError("start_frame cannot be negative")
```
samuelgarcia, line 338: Why having a clip for upper bounds and making a raise for lower bounds ?
we could also have a max(0, start_frame) similar to min(end_frame, num_samples)
no ?

src/spikeinterface/core/tests/test_binaryrecordingextractor.py
```diff
@@ -33,7 +33,7 @@ def test_BinaryRecordingExtractor(create_cache_folder):
 
 def test_round_trip(tmp_path):
     num_channels = 10
-    num_samples = 50
+    num_samples = 500
```
samuelgarcia, line 36: why ?
did the new change trigered a new errors with veru small signals ?

src/spikeinterface/preprocessing/decimate.py
```diff
@@ -123,13 +123,6 @@ def get_num_samples(self):
         return int(np.ceil((parent_n_samp - self._decimation_offset) / self._decimation_factor))
 
     def get_traces(self, start_frame, end_frame, channel_indices):
-        if start_frame is None:
-            start_frame = 0
-        if end_frame is None:
-            end_frame = self.get_num_samples()
-        end_frame = min(end_frame, self.get_num_samples())
-        start_frame = min(start_frame, self.get_num_samples())
```
samuelgarcia, line 131: This specific last lines is strange but this was put here for a reason a guess.
Here this (useless) protection has gone.

src/spikeinterface/core/baserecording.py
alejoe91: we decided to raise errors in #1989

src/spikeinterface/core/tests/test_binaryrecordingextractor.py
alejoe91: because the values in the tests ar start_frame 200/end_frame 400, so I think this was wrong

src/spikeinterface/preprocessing/decimate.py
alejoe91: I agree it was strange, but I don't think there was a reason. We can improve this global behavior later to follow #1989

src/spikeinterface/core/baserecording.py
h-mayorquin: And you voted for exaclty that against me 💔 : (

alejoe91: me???

h-mayorquin: I meant Sam : )

samuelgarcia: So we need also to raise error here also no ?

samuelgarcia: for upper bound I mean

alejoe91, line 338:
```suggestion
```
@h-mayorquin Sam made a good point about making another PR where we add all the checks discussed in #1989 

So let's remove it for now!

samuelgarcia (approved):

src/spikeinterface/core/baserecording.py
h-mayorquin: Yeah, the point was that the other bound will introduce errors I think so no point on getting into that quagmire here. But this one was for free already. 

Anyway, both  work, maybe it is better to do everything in one go as you suggested.

PR #3062, opened 2024-06-21
Dredge lfp

samuelgarcia:
Here the initial port of dredgelib into spikeinterface.
This come with some small refactoring in spikeinterface and folder reorganization.

@cwindolf tell me if this is OK for you. It is only an intial port. I also want to make a class to wrap into estimate_motion(method="drege_lfp")


src/spikeinterface/sortingcomponents/motion/tests/test_motion_estimation.py
```diff
@@ -49,6 +47,12 @@ def setup_module(tmp_path_factory):
     return recording, sorting, cache_folder
 
 
+@pytest.fixture(scope="module", name="setup_module")
+def setup_module_fixture(tmp_path_factory):
```
alejoe91, line 51: You can use the `create_cache_folder` global fixture ;)

samuelgarcia:
@cwindolf the last commit is a quite big refactoring in estimate motion to integrate the 'dredge_lfp' method.
Maybe you would like have a look.

samuelgarcia:
@cwindolf I could resists in porting also the dredge_ap and make a class.
This is not working yet because we need to fix the histogram2d that mimic all imrovements.

src/spikeinterface/sortingcomponents/motion/dredge.py
```diff
@@ -0,0 +1,1469 @@
...
+    mincorr=0.1,
+    # weights arguments
+    do_window_weights=True,
+    weights_threshold_low=0.2,
+    weights_threshold_high=0.2,
+    mincorr_percentile=None,
+    mincorr_percentile_nneighbs=None,
+    # raster arguments
+    amp_scale_fn=None,  ## @Charlie this one is not used anymore
+    post_transform=np.log1p,  ###@this one is directly transimited to weight_correlation_matrix() and so get_wieiith()
```
cwindolf, line 156: actually we can probably not even transmit it. we can apply the function to the threshold in advance before passing it to weight_correlation_matrix to reduce the number of params. I can do this in my PR.

src/spikeinterface/sortingcomponents/motion/dredge.py
```diff
@@ -0,0 +1,1469 @@
...
+    histogram_depth_smooth_um=1,
+    histogram_time_smooth_s=1,
+    avg_in_bin=False,
+    count_masked_correlation=False,
```
cwindolf, line 160: let's remove these count-related things

src/spikeinterface/sortingcomponents/motion/dredge.py
```diff
@@ -0,0 +1,1469 @@
...
+        What library to use for computing cross-correlations.
+        If numpy, falls back to the scipy correlate function.
+
+    Returns
+    -------
+    corr : tensor
+    """
+
+
+    if conv_engine == "torch":
```
cwindolf, line 1236: add `assert HAVE_TORCH` or similar

src/spikeinterface/sortingcomponents/motion/dredge.py
```diff
@@ -0,0 +1,1469 @@
...
+            )
+            max_corr, best_disp_inds = torch.max(corr, dim=2)
+            best_disp = possible_displacement[best_disp_inds.cpu()]
+            D[i : i + batch_size, j : j + batch_size] = best_disp.T
+            C[i : i + batch_size, j : j + batch_size] = max_corr.cpu().T
+
+    return D, C
+
+
+# TODO charlie sam : at the moment this is a duplicate with small differences see motion_estimate.py same function
```
cwindolf, line 1183: I checked through these two functions. There is no functional difference between them, and it should be fine to switch to the new one everywhere.

cwindolf: The main difference is this `xmasks` thing which is not used, but I think it's fine to keep it if you don't mind. Or else in my PR I can remove it, but I think let's stick to the new function if that's fine just for other tiny changes.

src/spikeinterface/sortingcomponents/motion/dredge.py
```diff
@@ -0,0 +1,1469 @@
...
+# @TODO : Charlie I started very small refactoring, I let you continue
+def dredge_ap(
+    recording,
+    peaks,
+    peak_locations,
+    direction="y",
+    rigid=False,
+    # nonrigid window construction arguments
+    win_shape="gaussian",
+    win_step_um=400,
```
cwindolf, line 140: I still think that these defaults are the best in typical NP1 data :)

src/spikeinterface/sortingcomponents/motion/dredge.py
```diff
@@ -0,0 +1,1469 @@
...
+        will not be cross-correlated. So, if your data has nonstationarities or changes which
+        could lead to bad cross-correlations at some timescale, it can help to input that
+        value here. If this is too small, it can make the motion estimation unstable.
+    mincorr : float, between 0 and 1
+        Correlation threshold. Pairs of frames whose maximal cross correlation value is smaller
+        than this threshold will be ignored when solving for the global displacement estimate.
+    thomas_kw, xcorr_kw, raster_kw, weights_kw
+        These dictionaries allow setting parameters for fine control over the registration
+    device : str or torch.device
+        What torch device to run on? E.g., "cpu" or "cuda" or "cuda:1".
```
cwindolf, line 87: I'll add more documentation of the other parameters in my PR.

cwindolf (commented):
I'm happy with stuff pretty much! Most of my "review" here is a todo for me to come back in the next PR for dredge AP.

src/spikeinterface/sortingcomponents/motion/dredge.py
```diff
@@ -0,0 +1,1469 @@
...
+import numpy as np
+
+import gc
+
+from .motion_utils import Motion, get_spatial_windows, get_window_domains, scipy_conv1d, make_2d_motion_histogram, get_spatial_bin_edges
+
+
+# todo use gaussian_filter1d in historgam 2d
+# put smotthing inside the histogram function
+# put the log for weight inhitstogram
```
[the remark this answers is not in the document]
samuelgarcia: Yes. I left a todo for you inside the code.

src/spikeinterface/sortingcomponents/motion/dredge.py
samuelgarcia: ok. I left the code  like this for the moment.

src/spikeinterface/sortingcomponents/motion/dredge.py
samuelgarcia: We are not anymore making import at module level but import locally and so this HAVE_TORCH is dessapearing slowly.

src/spikeinterface/sortingcomponents/motion/dredge.py
samuelgarcia: ok done

src/spikeinterface/sortingcomponents/motion/dredge.py
samuelgarcia: OK.
I remove the one in decentralized

src/spikeinterface/sortingcomponents/motion/iterative_template.py
```diff
@@ -0,0 +1,292 @@
...
+            spatial_bin_centers=spatial_bin_centers,
+            rigid=rigid,
+            win_margin_um=win_margin_um,
+            win_step_um=win_step_um,
+            win_scale_um=win_scale_um,
+            win_shape=win_shape,
+            zero_threshold=None,
+        )
+
+        # make a 3D histogram
```
yger, line 99: maybe do a if verbose: print('Computing histograms') as in dredge because otherwise verbose is never used

src/spikeinterface/sortingcomponents/motion/motion_cleaner.py
```diff
@@ -0,0 +1,72 @@
+import numpy as np
+
+# TODO this need a full rewrite with motion object
+
+
+def clean_motion_vector(motion, temporal_bins, bin_duration_s, speed_threshold=30, sigma_smooth_s=None):
+    """
+    Simple machinery to remove spurious fast bump in the motion vector.
+    Also can applyt a smoothing.
```
yger, line 9: typo "apply"

src/spikeinterface/sortingcomponents/motion/motion_cleaner.py
```diff
@@ -0,0 +1,72 @@
...
+            # take care of the shift
+            bins = (np.arange(motion_clean.shape[0]) - half_size + 1) * bin_duration_s
+        else:
+            bins = (np.arange(motion_clean.shape[0]) - half_size) * bin_duration_s
+        smooth_kernel = np.exp(-(bins**2) / (2 * sigma_smooth_s**2))
+        smooth_kernel /= np.sum(smooth_kernel)
+        smooth_kernel = smooth_kernel[:, None]
+        motion_clean = scipy.signal.fftconvolve(motion_clean, smooth_kernel, mode="same", axes=0)
+
+    return motion_clean
```
yger, line 72: Where is this function used? Should it be an option at the end of estimate_motion, extending the signature with motion_cleaner_kwargs as a dict, and if not None, it is launched?

src/spikeinterface/sortingcomponents/motion/motion_utils.py
```diff
@@ -0,0 +1,566 @@
...
+    if depth_smooth_um is not None:
+        motion_histogram = gaussian_filter1d(motion_histogram, depth_smooth_um / bin_um, axis=1, mode="constant")
+
+    if time_smooth_s is not None:
+        motion_histogram = gaussian_filter1d(motion_histogram, time_smooth_s / bin_s, axis=0, mode="constant")
+
+    return motion_histogram, temporal_bin_edges, spatial_bin_edges
+
+
+def make_3d_motion_histograms(
```
yger, line 484: This is a bit of a pity that make_2d_motion_histogram and makd_3d_motion_histogram are not the same function, one being a particular case of the other. Because conceptually, they should. We could also propose some smoothing for 3d histogram over time, amplitudes and depth. What do you think? One single function could ease the code maintenance

yger (dismissed):
Very nice! I'll test it today, and you need to fix tests also with some non appropriate imports that are still there

src/spikeinterface/sortingcomponents/motion/motion_utils.py
alejoe91: let's make it in a follow up PR

src/spikeinterface/sortingcomponents/motion/motion_cleaner.py
alejoe91: I think it's not used currently, but it's an option. Let's discuss this later

alejoe91:
@samuelgarcia @yger 

- I cleaned up the docs, looking good now :)

- There was a failing test due to this: the test recording probe is too short for the default win size and margin, resulting in `num_windows` = -1. I extended the `get_spatial_windows` to deal with this and switching to rigid motion in case window sizes/margins are too large for the probe depth. This produces a warning.

- I changed `contact_depth` to `contact_depths`, since it's more accurate!

src/spikeinterface/preprocessing/tests/test_motion.py
```diff
@@ -13,7 +13,7 @@ def test_estimate_and_correct_motion(create_cache_folder):
     if folder.exists():
         shutil.rmtree(folder)
 
-    rec_corrected = correct_motion(rec, folder=folder)
+    rec_corrected = correct_motion(rec, folder=folder, estimate_motion_kwargs={"win_step_um": 20})
```
alejoe91, line 16: the test recording depth is not that high!

alejoe91 (approved):

alejoe91:
One last fix needed to use `numpy.broadcast_to` instead of `torch.broadcast_to` if `conv_engine` is `numpy`

alejoe91:
In last commit: propagate API changes and docs

src/spikeinterface/sortingcomponents/motion/motion_cleaner.py
samuelgarcia: It is now separated from the main estimate. I think this is better. and honestly nobody used it.

PR #3063, opened 2024-06-21
Fix the probe handling tutorial

zm711:
If you do 

```python
>>> a_dict = {0: 'a', 1: 'b'}

>>> a, b = a_dict

>>>print(a)
0
>>>print(b)
1

```

which means people will do this wrong. I think maybe we switched the default from list to dict as some point. So I added a note in the explanation that you can choose the output and then fixed the example for the dict output instead.

### DIFF
--- a/examples/tutorials/core/plot_3_handle_probe_info.py
+++ b/examples/tutorials/core/plot_3_handle_probe_info.py
@@ -54,16 +54,15 @@ plot_probe(recording_2_shanks.get_probe())
 # or as a list of recordings with :code:`outputs='list'`.
 
 print(recording_2_shanks)
-print(f'/nGroup Property: {recording_2_shanks.get_property("group")}/n')
+print(f'\nGroup Property: {recording_2_shanks.get_property("group")}\n')
 
 # Here we split as a dict
 sub_recording_dict = recording_2_shanks.split_by(property="group", outputs='dict')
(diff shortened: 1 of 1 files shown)


examples/tutorials/core/plot_3_handle_probe_info.py
```diff
@@ -47,16 +47,23 @@
 plot_probe(recording_2_shanks.get_probe())
 
 ###############################################################################
-# Now let's check what we have loaded. The `group_mode='by_shank'` automatically
+# Now let's check what we have loaded. The :code:`group_mode='by_shank'` automatically
 # sets the 'group' property depending on the shank id.
-# We can use this information to split the recording into two sub-recordings:
+# We can use this information to split the recording into two sub-recordings.
+# We can acccess this information either as a dict with :code:`outputs='dict'` (default)
```
alejoe91, line 53:
```suggestion
# We can access this information either as a dict with :code:`outputs='dict'` (default)
```

zm711: Thanks that's what I get for typing fast!

h-mayorquin (approved):
LGTM.

alejoe91 (approved):

samuelgarcia:
merci. This was very bad.

Issue #2309, opened 2023-12-07
Concerns Regarding Unit Measurement in compute_template_metrics Function for Waveform Extractor Objects


Anyesz:
My concern arises when applying the function compute_template_metrics from the postprocessing module to waveform objects. According to the documentation, the metrics repolarization_slope and recovery_slope are provided in V/m, but in my case, these values are 6-digit numbers, which seems extremely large. I am questioning whether the unit of measurement is incorrect or if there might be a missing conversion somewhere in the process.

For instance, if we consider the approximate peak-to-trough duration of a realistic cell as 0.5 ms, and the membrane potential changes by 100 mV, the expected repolarization speed should be 100 mV/0.5 ms = 200 V/s. However, the values I am obtaining are around 200,000, which, in my opinion, appears to be in mV/s instead of V/s.

JoeZiminski:
@alejoe91 I am confused about this, [e.g. for `repolarization_slope`](https://github.com/SpikeInterface/spikeinterface/blob/328bba007391c47a5f08a1f0897119b5ea09bc05/src/spikeinterface/postprocessing/template_metrics.py#L407) the slope is calculated directly from the template, is the template in V? or this depends on the recording scaling?

chrishalcrow:
Hello, I had a go at updated the docs in response to this issue.

@Anyesz I think the correct unit are (unit of your template)/seconds. The unit of your template will depend on your recording: our lab is using Neuropixels 2.0 and the open ephys format. This uses the `bit_volts` parameter to convert from the integer raw data to "unitful data". For us this number is 0.195 and I'm pretty sure it converts integers to mV. So our templates are in mV, and we get similar results to you. Maybe you could check to see if your templates are in mV too, which would explain why your values are around 200,000.

alejoe91:
@chrishalcrow 0.195 is the conversion factor to uV ;)

chrishalcrow:
Hello, the docs have been updated (https://github.com/SpikeInterface/spikeinterface/pull/3066) so I'm closing this. @Anyesz if you think it's not clear feel free to re-open.

PR #3066, opened 2024-06-24
Correct docs and docstrings for compute_template_metric units

chrishalcrow:
In response to issue #2309.

Correct documentation on `compute_template_metrics` which had incorrect units for `repolarization_slope` and `recovery_slope`. The correct units are "units of the templates" per second, rather than `V/s`.

Also added a few more details to the docs, and updated docstrings of several of the template metrics to be more precise about units.

### DIFF
--- a/doc/modules/core.rst
+++ b/doc/modules/core.rst
@@ -180,7 +180,7 @@ Importantly, the :py:class:`~spikeinterface.core.SortingAnalyzer` handles the *s
 Sparsity defines the channels on which waveforms and templates are calculated using, for example,  a
 physical distance from the channel with the largest peak amplitude (see the :ref:`Sparsity` section). Scaling, set by
 the :code:`return_scaled` argument, determines whether the data is converted from integer values to :math:`\mu V` or not.
-By default, it is converted and all traces have units of :math:`\mu V`.
+By default, :code:`return_scaled` is true and all processed data voltage values are in :math:`\mu V` (e.g., waveforms, templates, spike amplitudes, etc.).
 
 Now we will create a :code:`SortingAnalyzer` called :code:`sorting_analyzer`.
 
--- a/doc/modules/core.rst
+++ b/doc/modules/core.rst
@@ -21,7 +21,7 @@ All classes support:
   * data on-demand (lazy loading)
   * multiple segments, where each segment is a contiguous piece of data (recording, sorting, events).
 
-
+.. _core-recording:
 Recording
 ---------
 
--- a/doc/modules/core.rst
+++ b/doc/modules/core.rst
@@ -162,7 +162,7 @@ Internally, any sorting object can construct 2 internal caches:
   2. a unique numpy.array with structured dtype aka "spikes vector". This is useful for processing by small chunks of
      time, like for extracting amplitudes from a recording.
 
-
+.. _core-sorting-analyzer:
 SortingAnalyzer
 ---------------
 
@@ -179,9 +179,8 @@ to perform further analysis, such as calculating :code:`waveforms` and :code:`te
 Importantly, the :py:class:`~spikeinterface.core.SortingAnalyzer` handles the *sparsity* and the physical *scaling*.
 Sparsity defines the channels on which waveforms and templates are calculated using, for example,  a
 physical distance from the channel with the largest peak amplitude (see the :ref:`Sparsity` section). Scaling, set by
-the :code:`return_scaled` argument, says whether the data has been converted from integer values to physical units such as
-Voltage (see the end of the :ref:`Recording` section).
-
+the :code:`return_scaled` argument, determines whether the data is converted from integer values to :math:`\mu V` or not.
+By default, it is converted and all traces have units of :math:`\mu V`.
 
 Now we will create a :code:`SortingAnalyzer` called :code:`sorting_analyzer`.
 
--- a/doc/modules/postprocessing.rst
+++ b/doc/modules/postprocessing.rst
@@ -301,29 +301,30 @@ template_metrics
 This extension computes commonly used waveform/template metrics.
 By default, the following metrics are computed:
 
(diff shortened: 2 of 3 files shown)


src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -454,12 +455,10 @@ def get_recovery_slope(template_single, sampling_frequency, peak_idx=None, **kwa
     Return the recovery slope of input waveforms. After repolarization,
     the neuron hyperpolarizes until it peaks. The recovery slope is the
     slope of the action potential after the peak, returning to the baseline
-    in dV/dT. The slope is computed within a user-defined window after
+    in dV/dT. The returned slope is in units of (unit of template)
+    per second. The slope is computed within a user-defined window after
```
zm711, line 459: I actually like below that you put: usually um/s for the distance below. Maybe we put something like if `return_scaled=True` then in uV/s or whatever since that is what we force all voltages to be when scaled? So we give the units for the most common but also make it clear that user's need to pay attention.

doc/modules/postprocessing.rst
```diff
@@ -304,11 +304,19 @@ By default, the following metrics are computed:
...
+* "repolarization_slope": speed to repolarize from the positive peak to 0
 * "num_positive_peaks": the number of positive peaks
 * "num_negative_peaks": the number of negative peaks
 
+The units of the results depend on the input. Voltages are based on the units of the
+template, usually :math:`\mu V` (this depends on the :code:`return_scaled`
+parameter, read more here: :ref:`core-recording`). Distances are based on the unit of the
+underlying recording's probe's :code:`channel_locations`, usually :math:`\mu m`.
+Times are always in seconds. E.g. if the templates are in units of :math:`mV` and channel
+locations in :math:`\mu m` then: :code:`repolarization_slope` is in :math:`mV / s`;
```
alejoe91, line 317: Channel locations are expected to be um.

For templates, we either have unscaled (`analyzer.return_scaled` is False) or `uV`

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -619,7 +618,7 @@ def fit_velocity(peak_times, channel_dist):
 
 def get_velocity_above(template, channel_locations, sampling_frequency, **kwargs):
     """
-    Compute the velocity above the max channel of the template.
+    Compute the velocity above the max channel of the template in units (unit of channel locations) per second, usually um/s.
```
alejoe91, line 621:
```suggestion
    Compute the velocity above the max channel of the template in um/s.
```
This is always the case. The convention is that channel_locations are in um. Maybe we should make this clearer in the docs

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -697,7 +696,7 @@ def get_velocity_above(template, channel_locations, sampling_frequency, **kwargs
 
 def get_velocity_below(template, channel_locations, sampling_frequency, **kwargs):
     """
-    Compute the velocity below the max channel of the template.
+    Compute the velocity below the max channel of the template in units (unit of channel locations) per second, usually um/s.
```
alejoe91, line 699: same here

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -775,7 +774,8 @@ def get_velocity_below(template, channel_locations, sampling_frequency, **kwargs
 
 def get_exp_decay(template, channel_locations, sampling_frequency=None, **kwargs):
     """
-    Compute the exponential decay of the template amplitude over distance.
+    Compute the exponential decay of the template amplitude over distance. The returned value
+    is in the same units as `channel_locations`, usually um.
```
alejoe91, line 778: same here

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -853,7 +858,8 @@ def exp_decay(x, decay, amp0, offset):
 
 def get_spread(template, channel_locations, sampling_frequency, **kwargs):
     """
-    Compute the spread of the template amplitude over distance.
+    Compute the spread of the template amplitude over distance. The returned value
+    is in the same units as `channel_locations`, usually um.
```
alejoe91, line 862: same here

src/spikeinterface/postprocessing/template_metrics.py
chrishalcrow: I've had another go.

doc/modules/core.rst
```diff
@@ -179,9 +179,8 @@ to perform further analysis, such as calculating :code:`waveforms` and :code:`te
 Importantly, the :py:class:`~spikeinterface.core.SortingAnalyzer` handles the *sparsity* and the physical *scaling*.
 Sparsity defines the channels on which waveforms and templates are calculated using, for example,  a
 physical distance from the channel with the largest peak amplitude (see the :ref:`Sparsity` section). Scaling, set by
-the :code:`return_scaled` argument, says whether the data has been converted from integer values to physical units such as
-Voltage (see the end of the :ref:`Recording` section).
-
+the :code:`return_scaled` argument, determines whether the data is converted from integer values to :math:`\mu V` or not.
+By default, it is converted and all traces have units of :math:`\mu V`.
```
alejoe91, line 183:
```suggestion
By default, :code:`return_scaled` is true and all processed data voltage values are in :math:`\mu V` (e.g., waveforms, templates, spike amplitudes, etc.).
```

alejoe91: @chrishalcrow what do you think about this? The other changes look great!

chrishalcrow: Beautiful!

alejoe91 (approved):

zm711 (approved):
This looks good to me as well.

PR #3068, opened 2024-06-24
Add plot_drift_map widget

alejoe91:
Separate drift maps into a separate widget:

image


src/spikeinterface/widgets/driftmap.py
```diff
@@ -0,0 +1,143 @@
...
+        cmap="inferno",
+        clim=None,
+        alpha=1,
+        backend=None,
+        **backend_kwargs,
+    ):
+        if segment_index is None:
+            assert (
+                len(np.unique(peaks["segment_index"])) == 1
+            ), "segment_index must be specified if there is only one segment in the peaks array"
```
zm711, line 62: Isn't this if it is multi-segment you need to specify....

zm711 (commented):
Couple initial doc things.

src/spikeinterface/widgets/driftmap.py
```diff
@@ -0,0 +1,143 @@
...
+class DriftMapWidget(BaseWidget):
+    """
+    Plot the a drift map from a motion info dictionary.
+
+    Parameters
+    ----------
+    peaks : np.array
+        The peaks array, with dtype ("sample_index", "channel_index", "amplitude", "segment_index")
+    peak_locations : np.array
+        The peak locations, with dtype ("x", "y") or ("x", "y", "z")
```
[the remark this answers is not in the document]
alejoe91: maybe just, as computed from the `localize_peaks` or `spike_locations` extension?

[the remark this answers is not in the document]
zm711: That's fair so that the user knows that the input needs to be exactly from that function/extension.

src/spikeinterface/widgets/driftmap.py
alejoe91, line 62:
```suggestion
            ), "segment_index must be specified if there are multiple segments in the peaks array"
```

samuelgarcia:
Could we move this into the motion.py to centralize everything related in the palce ?

src/spikeinterface/widgets/driftmap.py
```diff
@@ -0,0 +1,143 @@
...
+    clim : tuple or None, default: None
+        The min and max amplitude to display, if None (min and max of the amplitudes)
+    alpha : float, default: 1
+        The alpha of the scatter points
+    """
+
+    def __init__(
+        self,
+        peaks,
+        peak_locations,
```
alejoe91, line 45: add `analyzer` option

src/spikeinterface/widgets/driftmap.py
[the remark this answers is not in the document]
JoeZiminski: Another use for the glossary @zm711 !

src/spikeinterface/widgets/driftmap.py
```diff
@@ -0,0 +1,143 @@
+from __future__ import annotations
+
+import numpy as np
+
+from .base import BaseWidget, to_attr
+
+
+class DriftMapWidget(BaseWidget):
+    """
+    Plot the a drift map from a motion info dictionary.
```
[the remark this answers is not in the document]
JoeZiminski: Maybe a little more detail on the drift-map concept? (or, maybe this can go in a 'Notes'). "The drift map displays shows the vertical position of detected peaks as a function of recording time. The y-axis is probe depth and the x-axis is time."

src/spikeinterface/widgets/driftmap.py
```diff
@@ -0,0 +1,143 @@
...
+        recording=None,
+        sampling_frequency=None,
+        segment_index=None,
+        depth_lim=None,
+        color_amplitude=True,
+        scatter_decimate=None,
+        cmap="inferno",
+        clim=None,
+        alpha=1,
+        backend=None,
```
JoeZiminski, line 56: this `backend` argument must be `None` or `matplotlib` so it is undocumented. I think it is nicer to either document and state it's value is fixed, or (probably better?) remove as an option.

src/spikeinterface/widgets/driftmap.py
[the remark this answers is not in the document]
alejoe91: good point! Adding it!

JoeZiminski (approved):
Nice! Approved with some suggested changes.

src/spikeinterface/widgets/driftmap.py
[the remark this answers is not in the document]
alejoe91: it gets documented later, see [widget_list](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/widgets/widget_list.py#L83-L110)

src/spikeinterface/widgets/driftmap.py
```diff
@@ -0,0 +1,143 @@
...
+        dp = to_attr(data_plot)
+
+        assert backend_kwargs["axes"] is None, "axes argument is not allowed in MotionWidget"
+
+        self.figure, self.axes, self.ax = make_mpl_figure(**backend_kwargs)
+        fig = self.figure
+
+        if dp.times is None:
+            # temporal_bins_plot = dp.temporal_bins
+            x = dp.peaks["sample_index"] / dp.sampling_frequency
```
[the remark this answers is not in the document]
alejoe91: done!

src/spikeinterface/widgets/driftmap.py
```diff
@@ -0,0 +1,143 @@
...
+    def plot_matplotlib(self, data_plot, **backend_kwargs):
+        import matplotlib.pyplot as plt
+        from .utils_matplotlib import make_mpl_figure
+        from matplotlib.colors import Normalize
+
+        from spikeinterface.sortingcomponents.motion_interpolation import correct_motion_on_peaks
+
+        dp = to_attr(data_plot)
+
+        assert backend_kwargs["axes"] is None, "axes argument is not allowed in MotionWidget"
```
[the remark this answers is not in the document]
alejoe91: done

src/spikeinterface/widgets/tests/test_widgets.py
```diff
@@ -581,32 +584,39 @@ def test_plot_multicomparison(self):
...
+    def test_drift_raster_map(self):
+        peaks = self.motion_info["peaks"]
+        recording = self.recording
+        peak_locations = self.motion_info["peak_locations"]
+        analyzer = self.sorting_analyzer_sparse
 
+        possible_backends = list(sw.MotionWidget.get_possible_backends())
+        for backend in possible_backends:
+            if backend not in self.skip_backends:
+                # with recoridng
```
JoeZiminski, line 604:
```suggestion
                # with recording
```

src/spikeinterface/widgets/motion.py
```diff
@@ -96,55 +94,241 @@ def plot_matplotlib(self, data_plot, **backend_kwargs):
...
+        from spikeinterface.sortingcomponents.motion_interpolation import correct_motion_on_peaks
+
+        dp = to_attr(data_plot)
+
+        assert backend_kwargs["axes"] is None, "axes argument is not allowed in DriftRasterMapWidget. Use ax instead."
+
+        self.figure, self.axes, self.ax = make_mpl_figure(**backend_kwargs)
+
+        if dp.times is None:
+            peak_times = dp.peaks["sample_index"] / dp.sampling_frequency
```
JoeZiminski, line 240: Sorry completely forgot, is this a place to use `time_from_sample_index` etc. to handle `t_start` and `time vector`?

JoeZiminski (approved):
Changes look good, sorry forgot two things!

src/spikeinterface/widgets/motion.py
```diff
@@ -4,29 +4,31 @@
...
-    segment_index : None | int
-        If Motion is multi segment, the must be not None
-    mode : "auto" | "line" | "map"
-        How to plot map or lines.
-        "auto" make it automatic if the number of depth is too high.
+        The motion object.
+    segment_index : int | None, default: None
+        If Motion is multi segment, the must be not None.
+    mode : "auto" | "line" | "map", default: "line"
+        How to plot map or lines. "auto" makes it automatic if the number of motion depths is too high.
```
zm711, line 22: Just thinking about this (but fine to say out of scope for this PR, but "map" vs "line" isn't the clearest right? I mean "line" makes more sense, but thinking from a new user's perspective would I know what "map" meant. What do you think @JoeZiminski?

src/spikeinterface/widgets/motion.py
```diff
@@ -96,55 +94,241 @@ def plot_matplotlib(self, data_plot, **backend_kwargs):
...
+    depth_lim : tuple or None, default: None
+        The min and max depth to display, if None (min and max of the recording).
+    scatter_decimate : int, default: None
+        If > 1, the scatter points are decimated.
+    color_amplitude : bool, default: True
+        If True, the color of the scatter points is the amplitude of the peaks.
+    cmap : str, default: "inferno"
+        The colormap to use for the amplitude.
+    color : str, default: "Gray"
+        The color of the scatter points if color_amplitude is False.
```
zm711, line 133: something for us to discuss. We are super inconsistent with

`color_amplitude` vs `'color_amplitude'` vs ``color_amplitude`` (with ticks, but they are being rendered so you can't see them in this third option....)

src/spikeinterface/widgets/motion.py
```diff
@@ -96,55 +94,241 @@ def plot_matplotlib(self, data_plot, **backend_kwargs):
...
+
 class MotionInfoWidget(BaseWidget):
     """
-        Plot motion information from the motion_info dict returned by correct_motion().
-        This plot:
-          * the motion iself
-          * the peak depth vs time before correction
-          * the peak depth vs time after correction
+    Plot motion information from the motion_info dictionary returned by the `correct_motion()` funciton.
+    This widget plots::
```
zm711, line 283:
```suggestion
    This widget plots:
```

did we need the two : ?

zm711 (commented):
I think just one more correction from me and then two discussion points more for the future than to have to do right now :)

src/spikeinterface/widgets/motion.py
alejoe91: The problem here is that recording can be None (in that case sampling_frequency is required)

src/spikeinterface/widgets/motion.py
alejoe91: I kind of like map, but we can be clearer about it's definition. Agree that it probably deserves a different issue

src/spikeinterface/widgets/motion.py
alejoe91: Right. I like ticks because it's clear they refer to arguments and " for strings. We need another unification effort here

src/spikeinterface/widgets/motion.py
zm711: I meant should we define "map" in all our docstrings, not change the argument :)

src/spikeinterface/widgets/motion.py
zm711: I agree with you. I like ticks to emphasize arguments. I'll open an issue for the doc team :)

alejoe91:
All done! thanks for the feedback :) Will merge after tests so I can update the hybrid tools `How to`: https://spikeinterface--2436.org.readthedocs.build/en/2436/

Issue #1973, opened 2023-09-07
Contamination ratio


llobetv:
Just a comment about  compute_isi_violations from Hill (qualitymetrics/misc_metrics.py l. 240)

It is write that :
 "You can interpret an ISI violations ratio value of 0.5 as meaning that contaminating spikes are
   occurring at roughly half the rate of "true" spikes for that unit."
However Hill et al. converge to isi_contamination close to 0.5 when true contamination converge to 1:
![image](https://github.com/SpikeInterface/spikeinterface/assets/22896067/60bf5303-0616-42c1-85ca-2eebb6b5be12)

This equation works well for contamination estimation when contamination is low.

code to check:

import numpy as np
import spikeinterface.full as si

duration = 2200   # in s
neuron = np.arange(0, duration, 100e-3) # neuron that spike once every 100ms during 2200 s => 22 000 spikes
noise = np.random.uniform(0, duration, 220000) # create 220 000 spike with random timing 

final =  np.sort(np.concatenate([noise, neuron])) # create noisy neuron
calc_isi_violation = si.misc_metrics.isi_violations([final], duration)[0]
true_contamination = len(noise)/len(final)

print("Calculate isi_violation is : {}".format(calc_isi_violation))
print("True contamination is : {}".format(true_contamination))

@DradeAW

zm711:
Nick Steinmetz actually put in the correction based on @llobetv/@DradeAW paper 3 days ago into sortingQuality which is the basis for the SI implementation of the function `si.isi_violations()`. So now sortingQuality runs the equivalent of `si.compute_refrac_period_violations()` under the hood rather than a modification from the Hill et al paper. Not sure what the best solution is (ie merge the functions vs change documentation to be more explicit).

https://github.com/cortex-lab/sortingQuality/blob/70c8659adc60484434be828d617e16eb83e94cca/core/ISIViolations.m#L21-L46

zm711:
Good point :)

h-mayorquin:
Is there an actionable to-do so we could close this issue? Should the documentation be modified somewhere to make this point explicit?

zm711:
@h-mayorquin 

I think the actionable item would be 1 of 2 things:

1) change documentation (rtd and docstring) to state the function is still based on the original `sortingQuality` modification of the Hill et al conception of violations 

OR

[...]

PR #3070, opened 2024-06-24
Unify compute_isi_violation docs and add UltraMegaSort2000 citation

### LINKED ISSUE #2050: Unify notation in ISI violations docs
While the page on ISI violations is super useful, we should unify the notations between differen methods to improve readibility:

https://spikeinterface.readthedocs.io/en/latest/modules/qualitymetrics/isi_violations.html

chrishalcrow:
An attempt to close #2050

Two things:
1) Unify the notation between `isi_violation` and `rp_violation` in the docs so that the functions are easier to compare.
2) Get a correct citation for `isi_violation`: the current implementation is not actually from the Hill paper, it's a simpler function originally from UltraMegaSort2000 (https://github.com/danamics/UMS2K, https://github.com/danamics/UMS2K/blob/master/quality_measures/rpv_contamination.m) which is also written by Hill. I think citing this directly is much less confusing (see more confusion here: https://github.com/SpikeInterface/spikeinterface/issues/1973, and see https://github.com/cortex-lab/sortingQuality/blob/70c8659adc60484434be828d617e16eb83e94cca/core/ISIViolations.m#L17 for another discussion. Here our `isi_violation` formula is "the approximation previously used" which "wasn't accurate to the Hill paper" )


src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -261,22 +261,23 @@ def compute_isi_violations(sorting_analyzer, isi_threshold_ms=1.5, min_isi_ms=0,
...
-    In cases of highly contaminated units, the ISI violations ratio can sometimes be greater than 1.
+    The returned ISI violations ratio measures the approximate fraction of spikes in each
+    unit which are contaminted. This interpretation is good when the ratio is small, and
+    becomes worse as it grows. In cases of highly contaminated units, the ISI  violations
+    ratio can sometimes be greater than 1.
 
     References
     ----------
-    Based on metrics described in [Hill]_
+    Based on metrics originally implemented in [UMS]_
```
JoeZiminski, line 277: What is `[UMS]_`?

JoeZiminski: On nice it is a link for the docs very cool, how about `implemented in Ultra Mega Sort [UMS]_`?

doc/modules/qualitymetrics/isi_violations.rst
```diff
@@ -10,40 +10,37 @@ Calculation
...
-The following quantities are required:
-
-- :math:`T` the duration of the recording.
-- :math:`N` the number of spikes in the unit's spike train.
-- :math:`t_r` the duration of the unit's refractory period.
-- :math:`n_v` the number of violations of the refractory period.
-
-The estimated contamination :math:`C` can be calculated with 2 extreme scenarios. In the first one, the contaminant spikes are completely random (or come from an infinite number of other neurons). In the second one, the contaminant spikes come from a single other neuron:
+The estimated contamination :math:`C` is calculated in 2 extreme scenarios. In the first, the contaminant spikes
+ are completely random (or come from an infinite number of other neurons). In the second, the contaminant spikes
```
JoeZiminski, line 42: I am getting a slight formatting error here:

Screenshot 2024-06-27 at 10 17 34

doc/modules/qualitymetrics/isi_violations.rst
```diff
@@ -10,40 +10,37 @@ Calculation
...
+We aim to calculate the contamination rate :math:`C`, measuring the ratio of isi violations in the spike-train of a unit.
 The calculation works under the assumption that the contaminant events happen randomly or come from another neuron that is not correlated with our unit.
 A correlation will lead to an overestimation of the contamination, whereas an anti-correlation will lead to an underestimation.
 
-Different formulas have been developed over the years.
+Different formulas have been developed, but all require:
 
-Calculation from the [Hill]_ paper
-----------------------------------
+- :math:`T` the duration of the recording.
```
JoeZiminski, line 19: (s) I guess?

doc/modules/qualitymetrics/isi_violations.rst
```diff
@@ -10,40 +10,37 @@ Calculation
...
+- :math:`N` the number of spikes in the unit's spike train.
+- :math:`t_r` the duration of the unit's refractory period.
```
JoeZiminski, line 21: same (s)?

doc/modules/qualitymetrics/isi_violations.rst
```diff
@@ -10,40 +10,37 @@ Calculation
...
+- :math:`n_v` the number of violations of the refractory period.
 
-The following quantities are required:
+Calculation from the [UMS]_ package
```
JoeZiminski, line 24: Is it worth somewhere directly linking to the relevant [line on their github](https://github.com/danamics/UMS2K/blob/07e99b11e12a874d044e1367e548baeeb63e507b/quality_measures/rpv_contamination.m#L75)?

doc/modules/qualitymetrics/isi_violations.rst
JoeZiminski, line 42: I wonder in general if it is better to replace instances of the word 'neuron' here with the word 'unit'. I actually hate the word 'unit' 😆  But I feel it better captures the context 'a putative neuron, the sorted output from extracellular ephys data' whereas neuron I get confused - do you mean a true, real neuron? are you recording intracellularly? But it's just a feeling.

zm711: I struggle with that too a bit, but I think in this case the model itself assumes neurons right? As in assuming a real biological system with either 1) infinite neurons or 2) a single neuron then what would that look like for spikes. Then we take those models and apply them to our units to decide if they are putative neurons. So in this case I think neuron is the correct word until we get to the actual metric in which we work on units. But that's my two cents :)

doc/modules/qualitymetrics/isi_violations.rst
chrishalcrow: I think a link to the matlab file is good, since the file only contains this calculation (and explains the calculation!)

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -324,7 +325,7 @@ def compute_refrac_period_violations(
     Calculate the number of refractory period violations.
 
     This is similar (but slightly different) to the ISI violations.
-    The key difference being that the violations are not only computed on consecutive spikes.
+    The key differences being that the violations are not only computed on consecutive spikes.
```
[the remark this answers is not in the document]
chrishalcrow: Hello, I've put something in the Notes for both methods. Does that make it clear?

doc/modules/qualitymetrics/isi_violations.rst
chrishalcrow: Unless anyone has strong feelings I'll err on the side of caution (laziness) and not change anything.

doc/modules/qualitymetrics/isi_violations.rst
JoeZiminski: Oh I didn't see that, yes sounds good!

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -351,6 +356,12 @@ def compute_refrac_period_violations(
     -----
     Requires "numba" package
 
+    This method counts the number of violations which occur during the refactory period.
+    If there are three spikes within `refractory_period_ms`, the second and third spikes
```
JoeZiminski, line 360:
```suggestion
    For example, if there are three spikes within `refractory_period_ms`, the second and third spikes
```

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -351,6 +356,12 @@ def compute_refrac_period_violations(
...
+    violate the first spikes and the third spike violates the second spike. Hence there
```
JoeZiminski, line 361:
```suggestion
    violate the first spike and the third spike violates the second spike. Hence there
```

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -351,6 +356,12 @@ def compute_refrac_period_violations(
...
+    are three violations. This is in contrast to `compute_isi_violations`, which
+    computes the number of spikes which have been violated.
```
JoeZiminski, line 363: cool!

doc/modules/qualitymetrics/isi_violations.rst
JoeZiminski: Yes I think that makes sense, this is discussion of the theory, then in application 'neurons' -> 'units'. Happy to keep as is!

JoeZiminski (approved):
great stuff!

doc/references.rst
```diff
@@ -50,9 +50,10 @@ If you use the :code:`qualitymetrics` module, i.e. you use the :code:`analyzer.c
 or :code:`compute_quality_metrics()` methods, please include the citations for the :code:`metric_names` that were particularly
 important for your research:
 
-- :code:`amplitude_cutoff` or  :code:`isi_violation` [Hill]_
+- :code:`amplitude_cutoff` [Hill]_
 - :code:`amplitude_median` or :code:`sliding_rp_violation` [IBL]_
```
zm711, line 54: Shouldn't we just move this one too while we at it?

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -261,22 +261,28 @@ def compute_isi_violations(sorting_analyzer, isi_threshold_ms=1.5, min_isi_ms=0,
...
         Number of violations.
 
     Notes
     -----
-    You can interpret an ISI violations ratio value of 0.5 as meaning that contaminating spikes are
-    occurring at roughly half the rate of "true" spikes for that unit.
-    In cases of highly contaminated units, the ISI violations ratio can sometimes be greater than 1.
+    The returned ISI violations ratio measures the approximate fraction of spikes in each
+    unit which are contaminted. This interpretation is good when the ratio is small, and
+    becomes worse as it grows. In cases of highly contaminated units, the ISI  violations
```
zm711, line 272: This is 100% clear. Worse being what exactly... "As it grows past 0.xx it begins to diverge from xx" something like that based on llobet's comments in the issue.

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -261,22 +261,28 @@ def compute_isi_violations(sorting_analyzer, isi_threshold_ms=1.5, min_isi_ms=0,
...
+    which counts the number of violations.
 
     References
     ----------
-    Based on metrics described in [Hill]_
+    Based on metrics originally implemented in Ultra Mega Sort [UMS]_.
 
-    Originally written in Matlab by Nick Steinmetz (https://github.com/cortex-lab/sortingQuality)
-    and converted to Python by Daniel Denman.
+    This implementation is based on one written in Matlab by Nick Steinmetz
```
zm711, line 284:
```suggestion
    This implementation is based on one of the original implementations written in Matlab by Nick Steinmetz
```

maybe something like this. Because if people follow this link the implementation has been changed so we probably want to make it clear that it was originally based on the matlab but has since diverged? What do you think?

zm711 (commented):
Nothing to stop merging, but a couple potential tweaks around the edges from my perspective.

src/spikeinterface/qualitymetrics/misc_metrics.py
chrishalcrow: Mmm. I was deliberately vague because the success of the approximation depends on lots of factors: when the contamination is large is it because there are LOTS spikes near the same time a few times, or many violations of where one spike violates only one other spike? The Llobet comment is on one specific example (uniformly generated random spikes) so you shouldn't take precise numbers from it. 

All the models agree when $n_vT/N^2t_$ is small (much less than one), so we could say that the interpretation only works in that situation.

src/spikeinterface/qualitymetrics/misc_metrics.py
chrishalcrow: Yeah, sounds good. It's mainly there to give credit to Nick, and the new link to UMS is nearby.

doc/references.rst
chrishalcrow: Move what where?

src/spikeinterface/qualitymetrics/misc_metrics.py
zm711: That's fair. Okay so maybe something more like:

This intrepretation is good when the ratio is small, but as the ratio approaches 1 it is a less reliable indicator (or something like that? ). I know the ratio can go above one so maybe we don't even want that. But "becomes worse" is he vague part for me. ie becomes worse at what?

doc/references.rst
zm711: Sorry the highlighting wasn't clear.... My bad. You moved isi_violation to its own line so I thought just for aesthetics it might be nice to also move the `sliding_rp_violation` to it's own line as well?

chrishalcrow: Ahh, not for aesthetics. Each line was a list of metrics with the appropriate citation. But now there's only one citation with more than 1 metric. Agree that it looks better in two lines if there's one outlier - thanks!

src/spikeinterface/qualitymetrics/misc_metrics.py
chrishalcrow: I've realised that's also not true: you can cook up examples (where each spike is contaminated by exactly one other spike) where the ratio is a reliable indicator. So we can only really say: sometimes this is a good indicator, sometimes it isn't (like most of the quality metrics!). I suggest we refer to the Llobet paper, which discusses some of the different models? I've had a go.

samuelgarcia (approved):

PR #3073, opened 2024-06-24
Use keyword arguments in all `kilosort4` function calls.

JoeZiminski:
This PR updates the `kilosort4.py` wrapper to call kilosort functions using kwargs rather than position arguments, following the discussion [here](https://github.com/SpikeInterface/spikeinterface/pull/3055#pullrequestreview-2132263072).

This is looking okay as-is, will carefully double check tomorrow but might as well wait for an upcoming kilosort testing PR, to make sure it didn't break anything.

Some notes:

in KS v4.0.11 the `shift` and `scale` arguments were added to the `BinaryFiltered` class. I'm not sure exactly what these are doing, but when skipping kilosort preprocessing, should these be passed to from `params` to `BinaryFiltered` for versions >= 4.0.11?

At present, if I understand correctly all of the default params stored in kilosorts [`DEFAULT_SETTINGS`](https://github.com/MouseLand/Kilosort/blob/59c03b060cc8e8ac75a7f1a972a8b5c5af3f41a6/kilosort/parameters.py#L164) are overwritten by the `_default_params` set in SpikeInterface. I'm unsure of the best way to handle this. Would it not be best to always use the KS-set `DEFAULT_SETTINGS` to ensure the wrapper is faithful to the underlying kilosort version? KS may conceivably make important changes to their defaults but as I understand it these would not be propagated to the spikeinterface wrapper.

[EDIT] closing as superceeded by #3085, will transfer comments, thanks!

### DIFF
--- a/src/spikeinterface/sorters/external/kilosort4.py
+++ b/src/spikeinterface/sorters/external/kilosort4.py
@@ -243,7 +243,7 @@ class Kilosort4Sorter(BaseSorter):
 
         # Set preprocessing and drift correction parameters
         if not params["skip_kilosort_preprocessing"]:
-            ops = compute_preprocessing(ops, device, tic0=tic0, file_object=file_object)
+            ops = compute_preprocessing(ops=ops, device=device, tic0=tic0, file_object=file_object)
         else:
             print("Skipping kilosort preprocessing.")
             bfile = BinaryFiltered(
--- a/src/spikeinterface/sorters/external/kilosort4.py
+++ b/src/spikeinterface/sorters/external/kilosort4.py
@@ -17,9 +17,6 @@ class Kilosort4Sorter(BaseSorter):
     requires_locations = True
     gpu_capability = "nvidia-optional"
 
-    # Q: Should we take these directly from the KS defaults? https://github.com/MouseLand/Kilosort/blob/59c03b060cc8e8ac75a7f1a972a8b5c5af3f41a6/kilosort/parameters.py#L164
-    # I see these overwrite the `DEFAULT_SETTINGS`. Do we want to do this? There is benefit to fixing on the SI side, but users switching KS version would expect
-    # the defaults to represent the KS version. This could lead to divergence in result between users running KS directly vs. the SI wrapper.
     _default_params = {
         "batch_size": 60000,
         "nblocks": 1,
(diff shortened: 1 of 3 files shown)


src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -205,34 +205,58 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
...
+                n_chan_bin=n_chan_bin,
+                fs=fs,
+                nT=NT,
+                nt=nt,
+                nt0min=twav_min,
+                chan_map=chan_map,
                 hp_filter=None,
                 device=device,
-                do_CAR=do_CAR,
+                do_CAR=do_CAR,  # TODO: should this always be False if we are in skipping KS preprocessing land?
```
chrishalcrow, line 259: So, at the moment, if a user specifies `skip_kilosort_preprocessing = True`, then `do_CAR` is still `True` by default and is still applied. I guess: is this the expected behaviour? 

And then what do we want to do if someone explicitly specifies `skip_kilosort_preprocessing = True` and `do_CAR=True` ?

alejoe91: Good point! Maybe we should set it to False id `skip_kilosort_preprocessing` is set

alejoe91:
Thanks @JoeZiminski 

> in KS v4.0.11 the `shift` and `scale` arguments were added to the `BinaryFiltered` class. I'm not sure exactly what these are doing, but when skipping kilosort preprocessing, should these be passed to from `params` to `BinaryFiltered` for versions >= 4.0.11?

Honestly, I was a bit torn about including these two, simply because in SI we have better scaling management, so why would someone that launch KS4 from SI want to use scale and shift there? Maybe, let's just remove them!

> At present, if I understand correctly all of the default params stored in kilosorts [`DEFAULT_SETTINGS`](https://github.com/MouseLand/Kilosort/blob/59c03b060cc8e8ac75a7f1a972a8b5c5af3f41a6/kilosort/parameters.py#L164) are overwritten by the `_default_params` set in SpikeInterface. I'm unsure of the best way to handle this. Would it not be best to always use the KS-set `DEFAULT_SETTINGS` to ensure the wrapper is faithful to the underlying kilosort version? KS may conceivably make important changes to their defaults but as I understand it these would not be propagated to the spikeinterface wrapper.

That is potentially a good idea, but I think that having the dictionaries "spelled" out is better for readibility and design and it complies with the pattern with other sorters. Since we discussed about having automated CI for KS4 versions, we could throw in parameter checks there ;)

src/spikeinterface/sorters/external/kilosort4.py
chrishalcrow: Yeah, I'd vote to set 'do_CAR=False' if 'skip_kilosort_preprocessing` is True and do_CAR is unspecified.

But should we let the user skip preprocessing but just do CAR? i.e. should 'skip_kilosort_preprocessing=True` and 'do_CAR=True' apply _just_ CAR and nothing else? Maybe someone would want to do this?

PR #3076, opened 2024-06-25
Docstrings extractors update, fix PR01 and PR02 #3016

chrishalcrow:
Update docstrings in public Extractors module functions to be compliant with PR01 and PR02 from numpydoc.
(also fix a whitespace error in Preprocessing)

Two controversial changes:
- This update BREAKS `BiocamRecordingExtractor`, `SpikeGadgetsRecordingExtractor` and `BlackrockRecordingExtractor`, which all had an unused `block_index` parameter in their `__init__`. I think it's better to delete these than to document something which is unused => can create unexpected behaviour. Is this a problem @alejoe91 @samuelgarcia ?
- `read_openephys` still fails PR01 and PR02 because the Parameters are bundled in a `kwargs` dict. But the Parameters are very helpful and, from the user point of view, can be passed as "normal" parameters. Not sure what to do. Code here: https://github.com/SpikeInterface/spikeinterface/blob/d963aa26e31a9b6bc0bea2a4973f76d4dbafb062/src/spikeinterface/extractors/neoextractors/openephys.py#L304

The rules are...

https://numpydoc.readthedocs.io/en/latest/validation.html#
"PR01": "Parameters {missing_params} not documented",
"PR02": "Unknown parameters {unknown_params}",

...and they ensure that the function argument agrees with the Parameters listed in docstrings.

### DIFF
--- a/src/spikeinterface/extractors/cbin_ibl.py
+++ b/src/spikeinterface/extractors/cbin_ibl.py
@@ -27,9 +27,11 @@ class CompressedBinaryIblExtractor(BaseRecording):
     load_sync_channel : bool, default: False
         Load or not the last channel (sync).
         If not then the probe is loaded.
-    stream_name : str, default: "ap".
+    stream_name : {"ap", "lp"}, default: "ap".
         Whether to load AP or LFP band, one
         of "ap" or "lp".
+    cbin_file : str or None, default None
+        The cbin file of the recording. If None, searches in `folder_path` for file.
 
     Returns
     -------
--- a/src/spikeinterface/extractors/neoextractors/intan.py
+++ b/src/spikeinterface/extractors/neoextractors/intan.py
@@ -28,7 +28,9 @@ class IntanRecordingExtractor(NeoBaseRecordingExtractor):
         check we perform is that timestamps are continuous. Setting this to True will ignore this check and set
         the attribute `discontinuous_timestamps` to True in the underlying neo object.
     use_names_as_ids : bool, default: False
-        If True, use channel names as IDs. If False, use default IDs inherited from neo.
+        If False, use default IDs inherited from Neo. If True, use channel names as IDs.
+
+
     """
 
     mode = "file"
--- a/src/spikeinterface/extractors/neoextractors/intan.py
+++ b/src/spikeinterface/extractors/neoextractors/intan.py
@@ -27,6 +27,8 @@ class IntanRecordingExtractor(NeoBaseRecordingExtractor):
         If True, data that violates integrity assumptions will be loaded. At the moment the only integrity
         check we perform is that timestamps are continuous. Setting this to True will ignore this check and set
         the attribute `discontinuous_timestamps` to True in the underlying neo object.
+    use_names_as_ids : bool or None, default: None
+        If True, use channel names as IDs. If None, use default IDs.
     """
 
     mode = "file"
--- a/src/spikeinterface/extractors/neoextractors/intan.py
+++ b/src/spikeinterface/extractors/neoextractors/intan.py
@@ -27,8 +27,8 @@ class IntanRecordingExtractor(NeoBaseRecordingExtractor):
         If True, data that violates integrity assumptions will be loaded. At the moment the only integrity
         check we perform is that timestamps are continuous. Setting this to True will ignore this check and set
         the attribute `discontinuous_timestamps` to True in the underlying neo object.
-    use_names_as_ids : bool or None, default: None
-        If True, use channel names as IDs. If None, use default IDs.
+    use_names_as_ids : bool, default: False
+        If True, use channel names as IDs. If False, use default IDs inherited from neo.
     """
 
     mode = "file"
--- a/src/spikeinterface/extractors/toy_example.py
+++ b/src/spikeinterface/extractors/toy_example.py
@@ -62,13 +62,13 @@ def toy_example(
     seed : int or None, default: None
         Seed for random initialization.
     upsample_factor : None or int, default: None
-        A upsampling factor used only when templates are not provided.
+        An upsampling factor, used only when templates are not provided.
     num_columns : int, default:  1
         Number of columns in probe.
     average_peak_amplitude : float, default: -100
-        Average peak amplitude of generated templates
+        Average peak amplitude of generated templates.
     contact_spacing_um : float, default: 40.0
-        Spacing between probe contacts.
+        Spacing between probe contacts in micrometers.
 
     Returns
     -------
--- a/src/spikeinterface/extractors/toy_example.py
+++ b/src/spikeinterface/extractors/toy_example.py
@@ -57,12 +57,18 @@ def toy_example(
         Spike time in the recording
     spike_labels : np.array or list[nparray] or None, default: None
         Cluster label for each spike time (needs to specified both together).
-    # score_detection : int (between 0 and 1)
-    #    Generate the sorting based on a subset of spikes compare with the trace generation
     firing_rate : float, default: 3.0
         The firing rate for the units (in Hz)
     seed : int or None, default: None
         Seed for random initialization.
+    upsample_factor : None or int, default: None
+        A upsampling factor used only when templates are not provided.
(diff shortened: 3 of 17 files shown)


alejoe91 (approved):

alejoe91:
Thanks Chris! Ok to remove unused `block_index` in my side. @samuelgarcia ok?

JoeZiminski:
(Not knowing anything about the actual code functionality) it would be very cool to remove `**kwargs` from `read_openephys()` and replace with the full accepted params list. My preference is to avoid `**kwargs` wherever possible for public functions,  even if there is duplication, as more robust. With `**kwargs` it is possible for the user to pass a mis-spelled argument with no ill affect, it will not crash just silently not perform the desired action.

src/spikeinterface/extractors/cbin_ibl.py
```diff
@@ -27,9 +27,11 @@ class CompressedBinaryIblExtractor(BaseRecording):
     load_sync_channel : bool, default: False
         Load or not the last channel (sync).
         If not then the probe is loaded.
-    stream_name : str, default: "ap".
+    stream_name : {"ap", "lp"}, default: "ap".
```
JoeZiminski, line 30: This is cool I don't know this syntax, it means `"ap" | "lp"`?

src/spikeinterface/extractors/neoextractors/intan.py
```diff
@@ -27,6 +27,8 @@ class IntanRecordingExtractor(NeoBaseRecordingExtractor):
         If True, data that violates integrity assumptions will be loaded. At the moment the only integrity
         check we perform is that timestamps are continuous. Setting this to True will ignore this check and set
         the attribute `discontinuous_timestamps` to True in the underlying neo object.
+    use_names_as_ids : bool or None, default: None
+        If True, use channel names as IDs. If None, use default IDs.
```
JoeZiminski, line 31: What are the default IDs? `np.arange(num_chans)`?

src/spikeinterface/extractors/toy_example.py
```diff
@@ -57,12 +57,18 @@ def toy_example(
...
     spike_labels : np.array or list[nparray] or None, default: None
         Cluster label for each spike time (needs to specified both together).
-    # score_detection : int (between 0 and 1)
-    #    Generate the sorting based on a subset of spikes compare with the trace generation
     firing_rate : float, default: 3.0
         The firing rate for the units (in Hz)
     seed : int or None, default: None
         Seed for random initialization.
+    upsample_factor : None or int, default: None
+        A upsampling factor used only when templates are not provided.
```
JoeZiminski, line 65:
```suggestion
        A upsampling factor, used only when templates are not provided.
```

src/spikeinterface/extractors/toy_example.py
```diff
@@ -57,12 +57,18 @@ def toy_example(
...
+    num_columns : int, default:  1
+        Number of columns in probe.
+    average_peak_amplitude : float, default: -100
+        Average peak amplitude of generated templates
```
JoeZiminski, line 69:
```suggestion
        Average peak amplitude of generated templates.
```

src/spikeinterface/extractors/toy_example.py
```diff
@@ -57,12 +57,18 @@ def toy_example(
...
+    contact_spacing_um : float, default: 40.0
+        Spacing between probe contacts.
```
JoeZiminski, line 71:
```suggestion
        Spacing between probe contacts in micrometers.
```

src/spikeinterface/preprocessing/filter.py
```diff
@@ -10,15 +10,14 @@
...
-            The order of the filter
-        filter_mode :  "sos" | "ba", default: "sos"
-            Filter form of the filter coefficients:
-            - second-order sections ("sos")
-            - numerator/denominator : ("ba")
-        ftype : str, default: "butter"
-            Filter type for `scipy.signal.iirfilter` e.g. "butter", "cheby1".
-    """
+            filter_order : order
+                The order of the filter
```
JoeZiminski, line 14:
```suggestion
                The order of the filter. Note as filtering is applied with scipy's `filtfilt` functions (i.e. acausal, zero-phase) the effective order will be double the `filter_order`.
```

JoeZiminski: This however might be more relevant for #2942, at present all filters are applied with `filtfilt` but that will introduce causal `filt` and the docstring should be expanded to highlight the difference in cases.

alejoe91:
> (Not knowing anything about the actual code functionality) it would be very cool to remove `**kwargs` from `read_openephys()` and replace with the full accepted params list. My preference is to avoid `**kwargs` wherever possible for public functions, even if there is duplication, as more robust. With `**kwargs` it is possible for the user to pass a mis-spelled argument with no ill affect, it will not crash just silently not perform the desired action.

The "problem" is that the function dispatches wither the legacy or the binary format. We could extend the signature, but then we'll need to route the params accordingly!

JoeZiminski (approved):
Nice! I think it is okay to remove these unused kwargs, they should either be removed or raise an error now and removed later. One option is to raise an error to explain, and deprecating in the version-after-next? More overhead thought to remember to remove, would have to add as an issue + milestone. If I understand correctly these are not critical arguments (e.g. it will be obvious what data blocks are returned, not like it can cause silent or unknown changes to results). In which case just remove I think!

src/spikeinterface/extractors/cbin_ibl.py
chrishalcrow: It signifies that your have a finite set of options, https://numpydoc.readthedocs.io/en/latest/format.html#parameters
Numpy just puts the default first, but I still like having the default explicitly stated

src/spikeinterface/extractors/neoextractors/intan.py
chrishalcrow: They are inherited from neo's signal_channels["id"], and seems to depend on the Extractor. See here: https://github.com/NeuralEnsemble/python-neo/blob/1a7f60e2d75176b6df00fd7a2d1344772d223174/neo/rawio/intanrawio.py#L228 for the intan one. Maybe it's enough to just say that it's inherited from neo? And probably put the default first. Something like:

`If False, use default IDs inherited from neo. If True, use channel names as IDs.`

But thanks for pointing this out: just realised the default is incorrect for lots of these. Once we decide on the wording for this, I'll update all of them.

src/spikeinterface/preprocessing/filter.py
chrishalcrow: Just to check. You're saying we should implement the change you suggest here, but we will need to add more details once #2942 is merged?

src/spikeinterface/extractors/neoextractors/intan.py
zm711: Hey guys, neo has 3 things
channel_names: non-immutable (could be set by user) and so we can guarantee uniqueness
channel_ids: the hardware "names" of the recording apparatus sometimes just stringified indexes of the channels if the recording equipment doesn't provide unique names
channel_indices: the indices of the names and ids such that channel_names[1] is the same as channel_ids[1]

Basically setting that to True, the channel_names are used which lose the safety guarantees of neo but are clearer if the user actually "renamed" their channels. For example Intan has a native name which is assigned by the hardware and a custom name which is either the native name or a user specified name.

JoeZiminski: Thanks @zm711! thats super useful / informative. In that case I think @chrishalcrow it would be enough to say default channel IDs from NEO. BTW (out of scope comment for this PR): should `use_names_as_ids` just accept bool only? (i.e. `None` and `False` are presumably redundant).

src/spikeinterface/preprocessing/filter.py
JoeZiminski: I left it ambiguous as I couldn't decide 😅 . If your happy with the suggestion I'd say let's go for it here and we can update on #2942 when required.

PR #3079, opened 2024-06-25
better get_traces() check : toward strict mode

samuelgarcia:
I think this is a better place to check start/stop than: #3059
but this needed a big but fast refactoring.

@alejoe91 @JoeZiminski @h-mayorquin @pauladkisson @zm711

I am not sure if the strict mode should be before the relase. Maybe not. We can postpone this behavior after the release.

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -887,7 +887,7 @@ class BaseRecordingSegment(BaseSegment):
         """
         # @alessio 
         start_frame = int(start_frame) if start_frame is not None else 0
-        num_samples = self.get_num_samples()
+        num_samples = int(self.get_num_samples())
         end_frame = int(min(end_frame, num_samples)) if end_frame is not None else num_samples
 
         # @ramon @alessio @zach @joe @paul


h-mayorquin:
Can you say why this is better place to check?

samuelgarcia:
> Can you say why this is better place to check?

Sure.

In the actual design (since 0.90):
 * Recording has the responsabibility to handle metadata and also the ability to dump/load
 * RecordingSegment has the responsability to handle : get trace + segment size + time info

recording has a convinient get trace with a segment index but nothing prevent to use RecordingSegment get trace because it is the orginal design.

Very often RecordingSegment is retrieve trace from the "parent" with RecordingSegment and not from Recording.
So the protection of int + cliping (or raise) has to be here rather than the get trace in Recording.
This is why we need this change.

alejoe91:
@samuelgarcia so you suggest protection at both levels? Recording + segment?

samuelgarcia:
> @samuelgarcia so you suggest protection at both levels? Recording + segment?

no, I think we could remove the protect you did in the other PR. I wanted to discuss this with you first.

alejoe91:
In my opinion we need at the traces level. If you have a chain, you will redo the same check many times. If it's at the recording level, it will be done once. Since you are moving the segment get traces to private, it's not facing the end user. 99.9% of the calls are through the recording, not the segment

samuelgarcia:
> In my opinion we need at the traces level. If you have a chain, you will redo the same check many times. If it's at the recording level, it will be done once. Since you are moving the segment get traces to private, it's not facing the end user. 99.9% of the calls are through the recording, not the segment

You mean at recording level get trace level ? Good point.
RecordingSegment could only handle the int/None story and the Recording the raise.

alejoe91:
Yes I would be ok checking for None and casting to int in the segment as well, but the very important one is at the `BaseRecording.get_traces`

h-mayorquin:
@samuelgarcia Thanks for the explanation.

I would rather have checks only `BaseRecording.get_traces()` to avoid checking multiple times in preprocessing chains as @alejoe91 mention.


> when recording_2.get_traces() is called at the moment, it first enters the actual subclassed segment.get_traces() **and then works its way back up the chain**, to the BaseRecording.get_traces(), which gets the actual data. Then the data is preprocessed back down the chain and returned from segment.get_traces().

@JoeZiminski  Maybe I am misreading but this sounds wrong to me, usually segments call `segment.get_traces` and not  `BaseRecording.get_traces()`.

JoeZiminski:
Thanks @h-mayorquin! I indeed was completely wrong, I hope my second attempt makes more sense. 

Assuming it is correct, then I like the idea of making segment `get_traces()` private but, yes why not funnel all calls through `recording.get_traces()`? This would help decrease the surface API / help maintainability. It would also be clearer to users too, I distinctly remember being very confused as to why the `get_traces()` I was looking at in the source code (segment level) had a completely different signature to the `recording.get_traces()` I was trying to call in python.

h-mayorquin:
As Alessio mentions in practice most calls happen through `BaseRecording.get_traces()`, `BaseRecordingSegment.get_traces()` is mostly called internally. 

But Sam points out that this is not how the design intent was created and I understand he wants to preserve that intent.

zm711:
I'm a bit late to this, but I've also never worried about segments as an end-user. I understand the concept and original design choice, but do we know how many people actually use the segment machinery vs just concatenating into a mono-segment when necessary?

PR #3084, opened 2024-06-25
Add machinery to run test only on changed files

h-mayorquin:
Companion of https://github.com/SpikeInterface/spikeinterface/pull/2937 where tests to control which test modules are run as a function of changes is imported (and simplfied) from the ubuntu test


src/spikeinterface/widgets/widget_list.py
```diff
@@ -1,3 +1,5 @@
+# Another inocent comment
```
samuelgarcia, line 1:
```suggestion
# Yep
```

samuelgarcia:
I think I will like this a lot.

.github/detect_module_changes.py
```diff
@@ -0,0 +1,30 @@
...
+
+changed_files = os.getenv("CHANGED_FILES", "").split()
+
+for change, patterns in changes.items():
+    for file in changed_files:
+        if any(pathlib.PurePath(file).match(pattern) for pattern in patterns):
+            print(f"{change}=true")
+            with open(os.getenv("GITHUB_OUTPUT"), "a") as env_file:
+                env_file.write(f"{change}=true\n")
+            break
```
alejoe91, line 30: brillant :)

.github/workflows/all-tests.yml
```diff
@@ -32,13 +32,77 @@ jobs:
...
+            ["/sorters/internal"]="SORTERS_INTERNAL_CHANGED"
+            ["/comparison/"]="COMPARISON_CHANGED"
+            ["/curation/"]="CURATION_CHANGED"
+            ["/widgets/"]="WIDGETS_CHANGED"
+            ["/exporters/"]="EXPORTERS_CHANGED"
+            ["/sortingcomponents/"]="SORTINGCOMPONENTS_CHANGED"
+            ["/generation/"]="GENERATION_CHANGED"
+          )
+
+          changed_files="${{ steps.changed-files.outputs.all_changed_files }}"
```
alejoe91, line 64: why not using the python script here?

JoeZiminski:
This is really cool, it must save a lot of time. Just to check my understanding, is there some stage in the CI where everything is tested together, changed or not? For example, to check if a change in `core` breaks some other part of the code that was not explicitly changed but depends on `core`.

src/spikeinterface/widgets/widget_list.py
h-mayorquin: : D

.github/workflows/all-tests.yml
h-mayorquin: That was the coal it is just that I had to test something. 

Now all the logic to decide which tests to run lives on this script.

h-mayorquin:
> This is really cool, it must save a lot of time. Just to check my understanding, is there some stage in the CI where everything is tested together, changed or not? For example, to check if a change in `core` breaks some other part of the code that was not explicitly changed but depends on `core`.

Yes, two things:
1) Here, if the core, the pyproject.toml or the neo base class are changed, we run all the tests.
2) We run a cron job with the full test daily. That's the one that is used to measure code coverage.

h-mayorquin:
@alejoe91 @samuelgarcia this is ready and will solve some of the errors that you are seeing on the CI. Apparently, datalad sanctioned method stopped working yesterday for some reason but the ongoing action for installing git-annex has been working for a while on linux so I changed the script for that.

alejoe91 (approved):

alejoe91:
Heroic!!! Thanks @h-mayorquin

PR #3085, opened 2024-06-26
Add kilosort4 wrapper tests

JoeZiminski:
This PR makes some refactorings and small changes to `kilosort4.py` (Kilosort4 Wrapper) and adds tests for the wrapper across all released versions.

### Changes to `kilosort4.py`

- Changes the `_default_params` based on version to always match kilosort's `DEFAULT_SETTINGS`.
- All KS4 functions are called via kwargs rather than position arguments.
- Removed `keep_good_only` and `scaleproc` which I think were left over from earlier KS versions but are not implemented in KS4.
- KS4 `.__version__` attribute is `4` for all versions < `4.0.10` so I used `importlib.version` for versioning here.

### CI tests
- The CI tests use a python script to grab the versions of all kilosort released on pypi and set them in a matrix to run tests over, inspired by @tabedzki [related PR](https://github.com/SpikeInterface/spikeinterface-dockerfiles/pull/76) on auto-building the docker images.
- Tests for running KS4 natively vs. through SpikeInterface wrapper are added (details on these in the module header). These are stored in `.github/scripts` so are not part of the test suite. The CI script can be adjusted to run on cron job only. I ran into problems with the module tagging in `conftest.py` because the script is not a submodule of the package src, so added a try/except. Maybe there is a better solution (e.g. running the script from a different location) happy to adjust this.
- 

### Some points for discussion
- Possible kwargs to fully expose: `shift`, `scale`, `tmin`, `tmax`, `save_preprocesed_copy`.
- docker is not tested, but I think as the docker code is not related to how failthfully the wrapper is, I think thats okay not to test it?
- `save_extra_vars` (KS4) is `save_extra_kwargs` in SpikeInterface, should it be renamed for consistency?

### KIlosort versions <= `4.0.4`

For some reason on the tests in CI, there are completely wrong results between SI wrapper vs. KS4 in `4.0.4`. Also, versions `<4.0.4` seem to be susceptible to a bug ` ValueError: negative axis 0 index: -2147483648` which looks like a KS4-side overflow error that's since fixed. Also, as all KS4 `DEFAULT_SETTINGS` changes that are a little tricky to handle (see below comment to Alessio) are `<=4.0.4` I suggest we just don't support versions earlier than 4.0.5? 

#### Some TODOs:
- KS4 next version will deprecate `duplicate_spike_bins` for `duplicate_spike_ms`, also mention `save_preprocesed_copy` typo.
- could test a multi-segment recording to check concatenation is OK.


src/spikeinterface/temp_test_file_dir/test_kilosort4_new.py
```diff
@@ -13,6 +13,7 @@
 from packaging.version import parse
 from importlib.metadata import version
 
+# TODO: save_preprocesed_copy is misspelled in KS4.
```
alejoe91, line 16: nice catch!

src/spikeinterface/core/datasets.py
```diff
@@ -14,56 +14,80 @@ def download_dataset(
...
+    a list of all the files in the folder and then use pooch to download them one by one.
 
     Parameters
     ----------
     repo : str, default: "https://gin.g-node.org/NeuralEnsemble/ephy_testing_data"
         The repository to download the dataset from
     remote_path : str, default: "mearec/mearec_test_10s.h5"
         A specific subdirectory in the repository to download (e.g. Mearec, SpikeGLX, etc)
-    local_folder : str, default: None
+    local_folder : str, optional
```
alejoe91, line 31: we use `default: None` elsewhere ;)

JoeZiminski: Thanks @alejoe91!! Though please ignore  most of this PR ATM, it contains a lot of stray commits from random other PRs that are screwing everything up and I that need to drop! 😅

alejoe91: Ok! I'll refrain myself ;)

src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -205,40 +205,69 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
...
+        )
+
         if version.parse(cls.get_sorter_version()) >= version.parse("4.0.12"):
-            ops = initialize_ops(settings, probe, recording.get_dtype(), do_CAR, invert_sign, device, False)
+            # TODO: save_preprocesed_copy added
+            ops = initialize_ops(
+                settings=settings,
+                probe=probe,
+                data_dtype=recording.get_dtype(),
+                do_CAR=do_CAR,
```
JoeZiminski, line 224: Ported from #3073 

> Chris
> So, at the moment, if a user specifies `skip_kilosort_preprocessing = True`, then `do_CAR` is still `True` by default and is still applied. I guess: is this the expected behaviour?
> 
> And then what do we want to do if someone explicitly specifies `skip_kilosort_preprocessing = True` and `do_CAR=True` ?

> So, at the moment, if a user specifies `skip_kilosort_preprocessing = True`, then `do_CAR` is still `True` by default and is still applied. I guess: is this the expected behaviour?
> 
> And then what do we want to do if someone explicitly specifies `skip_kilosort_preprocessing = True` and `do_CAR=True` ?

>Alessio
> So, at the moment, if a user specifies `skip_kilosort_preprocessing = True`, then `do_CAR` is still `True` by default and is still applied. I guess: is this the expected behaviour?

>Chris
> And then what do we want to do if someone explicitly specifies `skip_kilosort_preprocessing = True` and `do_CAR=True` ?

JoeZiminski: I think because `do_CAR` is the default `True` it is not possible to determine if the user wanted to apply `CAR` or just left the default on. In this case I think most users would expect `skip_kilosort_preprocessing` to skip all preprocessing and would be surprised if CAR was still performed by default. I would suggest to raise an error if `skip_kilosort_preprocessing=False` and `do_CAR=True`. It will be annoying for users first time because by default it will raise an error, but at least it is an easy fix and super explicit. If we get issues raised from people who want to skip KS preprocessing but keep CAR (I would imagine this is very small % of users) we could reconsider.

Further on this, I believe that the KS drift correction is performed in the [code block](https://github.com/MouseLand/Kilosort/blob/c4ed5dd093e6b979756f2e5c85a781ff100079b3/kilosort/io.py#L726) that also performs whitening. So when whitening is turned off as part of `skip_kilosort_preprocessing` then drift correct is also turned off. This makes it even weirder for `skip_kilosort_preprocessing` to allow only CAR. Also (if this is correct) this behaviour can be added to the dosctring and I'd suggest raise an error if `do_correction=True` while `skip_kilosort_preprocessing=True`. Even though it will raise a lot of errors by default when trying to use `skip_kilosort_preprocessing` I feel it is quite a complex / confusing case and so best to be super explicit.

JoeZiminski:
(In quotes, Alessio original message ported from #3073)

> [On `shift` and `scale` adding to `BinaryFiltered`] Honestly, I was a bit torn about including these two, simply because in > SI we have better scaling management, so why would someone that launch KS4 from SI want to use scale and shift there? >Maybe, let's just remove them!

Yes I can definitely see the advantage in not including everything, on balance though as a philosophy I think SpikeInterface, when it comes to sorters, is most effective as as a wrapper that exposes all available sorter options. In some ways it's also easier to maintain because avoids having to decide what is / isn't worthwhile to expose, and avoids potential future issues users might raise wanting things included. As such I'd probably advocate towards exposing `shift`, `scale`, `tmin`, `tmax`, `save_preprocessed_copy` as part of this PR.

>[On setting default params in SpikeInterface side `_default_params` vs. KS4 `DEFAULT_SETTINGS`. That is potentially >a good idea, but I think that having the dictionaries "spelled" out is better for readibility and design and it complies with the >pattern with other sorters. Since we discussed about having automated CI for KS4 versions, we could throw in parameter >checks there ;)

Agree, for some reason I thought this was reading to some large differences in results but its not. They are the same for all versions except two cases where the defaults changed in early KS versions. I made the `_param_defaults` condition on the version in these cases, but this duplicates the `cls.get_sorter_version()` but I can see no way around this as the class is not instantiated yet. Anyways, glad its easy to handle!

.github/workflows/test_kilosort4.yml
```diff
@@ -0,0 +1,74 @@
...
+          echo "matrix=$(jq -c . < .github/scripts/kilosort4-latest-version.json)" >> $GITHUB_OUTPUT
+
+  test:
+    needs: versions
+    name: ${{ matrix.ks_version }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.12"]  # TODO: just checking python version is not cause of failing test.
```
JoeZiminski, line 49:
```suggestion
        python-version: ["3.12"] 
```

src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -129,9 +128,8 @@ def is_installed(cls):
 
     @classmethod
     def get_sorter_version(cls):
-        import kilosort as ks
-
-        return ks.__version__
+        """kilosort version <4.0.10 is always '4'"""
```
JoeZiminski, line 131:
```suggestion
        """kilosort.__version__ <4.0.10 is always '4'"""
```

.github/scripts/test_kilosort4_ci.py
```diff
@@ -0,0 +1,634 @@
...
+            expected_arguments.pop(-1)
+            expected_arguments.extend(["shift", "scale", "file_object"])
+
+        self._check_arguments(
+            BinaryFiltered,
+            expected_arguments
+        )
+
+    def _check_arguments(self, object_, expected_arguments):
+        sig = signature(object_)
```
JoeZiminski, line 339:
```suggestion
        """
        Check that the argument signature of  `object_` is as expected 
        (i..e has not changed across kilosort versions).
        """
        sig = signature(object_)
```

carlacodes:
is there any easy way to change save_preprocessed_copy to true within the existing spikeinterface framework?

JoeZiminski:
Hi @carlacodes unfortunately not, although I think it would be quick to add. @alejoe91 would you be happy to expose `save_preprocessed_copy` in the `kilosort4` wrapper?

alejoe91:
@JoeZiminski sure!

.github/scripts/test_kilosort4_ci.py
```diff
@@ -0,0 +1,634 @@
...
+                assert params[key] == DEFAULT_SETTINGS[key], f"{key} is not the same across versions."
+
+    # Testing Arguments ###
+    def test_set_files_arguments(self):
+        self._check_arguments(
+            set_files,
+            ["settings", "filename", "probe", "probe_name", "data_dir", "results_dir"]
+        )
+
+    def test_initialize_ops_arguments(self):
```
JoeZiminski, line 257: e.g. here

.github/workflows/test_kilosort4.yml
```diff
@@ -0,0 +1,74 @@
+name: Testing Kilosort4
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 12 * * 0"  # Weekly on Sunday at noon UTC
+  pull_request:
+    types: [synchronize, opened, reopened]
+    branches:
+      - main
```
alejoe91, line 10: @JoeZiminski I would remove this, otherwise it will run on any open PR...no?

JoeZiminski: @alejoe91 oh yes good point, should this be a weekly cron job like [here](https://github.com/SpikeInterface/spikeinterface/blob/694f8620e7e79cc914678bf2c8a3a7d70df3af29/.github/workflows/all-tests.yml#L6)?

so just

```
on:
  workflow_dispatch:
  schedule:
    - cron: "0 12 * * 0"  # Weekly on Sunday at noon UTC
```

JoeZiminski: Actually, there is a conflict with exposing `save_preprocesed_copy`, will fix now

h-mayorquin (commented):
Thanks for working on this!

PR #3088, opened 2024-06-26
Adding option to overwrite while doing correct_motion and saving to a folder

yger:
Currently, when doing motion correction, if saving to a folder we can not overwrite and an error is launched only at then end, after motion correction has been made. This PR adds the option to overwrite the motion folder


src/spikeinterface/preprocessing/motion.py
```diff
@@ -316,6 +319,14 @@ def correct_motion(
 
     if folder is not None:
         folder = Path(folder)
+        if overwrite:
+            if folder.exists():
```
alejoe91, line 323:
```suggestion
            if folder.is_dir():
```

src/spikeinterface/preprocessing/motion.py
```diff
@@ -316,6 +319,14 @@ def correct_motion(
...
+                import shutil
+
+                shutil.rmtree(folder)
+        else:
+            assert not folder.exists(), f"Folder {folder} already exists"
```
alejoe91, line 328:
```suggestion
            assert not folder.is_dir(), f"Folder {folder} already exists"
```

src/spikeinterface/preprocessing/motion.py
alejoe91: otherwise `shutil.rmtree` will fail

alejoe91 (approved):

src/spikeinterface/preprocessing/motion.py
```diff
@@ -204,6 +204,7 @@ def correct_motion(
     recording,
     preset="nonrigid_accurate",
     folder=None,
+    overwrite=False,
```
samuelgarcia, line 207: This is not consistent with some otherplace in spikeinterface.

alejoe91:
@yger why did you remove the code that does the overwriting?

yger:
aie, mistake then, i'll check

yger:
I've mixed up some git history I guess, so mayeb the best is to delete this PR and start a new one? I don't know how to properly revert, there are way too many changes in my commit that are not mine here

alejoe91 (dismissed):

alejoe91:
the diff in files look really weird!

@yger I would suggest making a clean PR without messing up with git history ;)

yger:
Now this has been fixed and the diff is legit. Just squash everything

alejoe91 (approved):

PR #3089, opened 2024-06-26
extractor_dict_iterator for solving path detection in object `kwargs`

h-mayorquin:
Extractor dicts (outputs of `BaseExtractor.to_dict()`) are nested structures where some arguments are paths.

For solving the problem in https://github.com/SpikeInterface/spikeinterface/issues/3013 we need a way of filtering by the arguments that correspond to real paths (e.g. Path(argument).exists() is True) and only modify those.

The current approach is to use the `recursive_path_modifier`:
https://github.com/alejoe91/spikeinterface/blob/fa2ca6d7c804fc8ab3df93dbd463d9576bc7723b/src/spikeinterface/core/core_tools.py#L186

I guess at the beginning the purpose of the function was both iterating overt the dict and also modifying the values. 
But then, some other functions appeared like `get_path_list` that use `recursive_path_modifier` to iterate over the extractor and accumulate path-like values.

This PR creates two functions that can accomplish the same as `recursive_path_modifier` but are decoupled:

1. First, it introduces `extractor_dict_iterator` that generates an iterator over all elements with three attributes: value, name and access path.
2. Second, a new `set_value_in_extractor_dict` function that uses the access path of the elements from the iterator to modify the value is also introduced.

With this logic, functions like `find_recording_folders` that only use the first part can avoid the part of `recursive_path_modifier` that modifies the values. Avoiding computational waste and simplifying intent. Moreover,, the problem in #3014 of having to check for path existence before or after a modification can be easily stated as a single line where elements are filtered instead of modifying the logic of `recursive_path_modifier`. In general it enables working with extractor dicts in the folowing way:

1) Collect items from the extractor dict
2) Filter to select some of them
3) Modify them 

Note that this does not even need to be about paths. Any type of filter, reduce or concistency check and modification can be implemented with this.

I also modified the tests of `make_relative_paths` and `make_absolute_paths` to work on real paths so we can directly test the type of problem causes #3013


src/spikeinterface/core/core_tools.py
```diff
@@ -250,15 +320,17 @@ def recursive_path_modifier(d, func, target="path", copy=True) -> dict:
                     raise ValueError(f"{k} key for path  must be str or list[str]")
 
 
-def _get_paths_list(d):
-    # this explore a dict and get all paths flatten in a list
-    # the trick is to use a closure func called by recursive_path_modifier()
-    path_list = []
+# This is the current definition that an element in a recording_dict is a path
```
alejoe91, line 323:
```suggestion
# This is the current definition that an element in an extractor_dict is a path
```

alejoe91 (changes requested):
@h-mayorquin I love this! just some renaming suggestions

alejoe91:
@h-mayorquin this should cloe #3014 right?

src/spikeinterface/core/core_tools.py
```diff
@@ -183,6 +184,75 @@ def is_dict_extractor(d: dict) -> bool:
     return is_extractor
 
 
+recording_dict_element = namedtuple(typename="recording_dict_element", field_names=["value", "name", "access_path"])
```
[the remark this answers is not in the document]
h-mayorquin: Yes, I made that mistake at the beginning and this renaming escape me. Will change!

h-mayorquin:
@alejoe91 did the correction and yes, this should close #3041

alejoe91:
@h-mayorquin can you make the previous test function in `test_core_tools` (`test_path_utils_functions`) also to use `pytest.skip`?

h-mayorquin:
Yes, done.

alejoe91:
This looks good to me. Much much cleaner than the old presentation!

@samuelgarcia what do you think?

alejoe91 (approved):

samuelgarcia:
> I guess at the beginning the purpose of the function was both iterating overt the dict and also modifying the values. But then, some other functions appeared like `get_path_list` that use `recursive_path_modifier` to iterate over the extractor and accumulate path-like values.

Excellent guess and analysis of the histical reason!
I wanted to do something similar to this PR this when impleneted _get_paths_list but was to lazy to refacor with the yield like you did.

Thanks a lot for this effort.

src/spikeinterface/core/core_tools.py
```diff
@@ -183,6 +184,75 @@ def is_dict_extractor(d: dict) -> bool:
...
+    """
+
+    def _extractor_dict_iterator(dict_list_or_value, access_path=(), name=""):
+        if isinstance(dict_list_or_value, dict):
+            for k, v in dict_list_or_value.items():
+                yield from _extractor_dict_iterator(v, access_path + (k,), name=k)
+        elif isinstance(dict_list_or_value, list):
+            for i, v in enumerate(dict_list_or_value):
+                yield from _extractor_dict_iterator(
+                    v, access_path + (i,), name=name
```
samuelgarcia, line 219: This is really smart to have the path both with key and index for list!!

src/spikeinterface/core/core_tools.py
```diff
@@ -183,6 +184,75 @@ def is_dict_extractor(d: dict) -> bool:
     return is_extractor
 
 
+extractor_dict_element = namedtuple(typename="extractor_dict_element", field_names=["value", "name", "access_path"])
```
samuelgarcia, line 187: very good idea!

PR #3091, opened 2024-06-27
Mark failing sorter test on Windows*Python3.12 as xfail

alejoe91:
See https://github.com/SpikeInterface/spikeinterface/actions/runs/9693230325/job/26748253219?pr=3090

Let's not waste too much time here for the time being since it seems a 3rd party problem

### DIFF
--- a/src/spikeinterface/sorters/tests/test_runsorter.py
+++ b/src/spikeinterface/sorters/tests/test_runsorter.py
@@ -1,7 +1,9 @@
 import os
+import platform
 import pytest
 from pathlib import Path
 import shutil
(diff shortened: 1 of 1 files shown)


src/spikeinterface/sorters/tests/test_runsorter.py
```diff
@@ -19,6 +21,10 @@ def generate_recording():
     return _generate_recording()
 
 
+@pytest.mark.xfail(
```
h-mayorquin, line 24: @alejoe91, was the equality sign in python meant to be >= instead of just larger?

PR #3092, opened 2024-06-27
estimate_sparsity arg ordering

### LINKED ISSUE #3086: Standardise order of recording, sorting args
Hi all! I've noticed that for different functions which require both a recording and a sorting, the order in which these should be passed is not consistent. E.g. `si.create_sorting_analyzer(sorting, recording)` but `si.estimate_sparsity(recording, sorting)`

Not exactly high priority, but it might be nice to standardise this order in the API at some point? Although I can also see how it might be better to have the first argument be the most relevant object for whatever that function is doing

jakeswann1:
Fixes #3086 , if wanted


alejoe91:
@jakeswann1 I guess you based this on a branch that was not in sync, so you're getting a bunch of conflicts! Maybe easier to close, sync, and push again!

PR #3094, opened 2024-06-27
Reordering recording, sorting args

### LINKED ISSUE #3086: Standardise order of recording, sorting args

jakeswann1:
Would fix #3086, if we want to break `estimate_sparsity`


zm711 (commented):
this looks good to me. I think @alejoe91 and @samuelgarcia will decide on whether they are okay with the break. But I think it is the right idea.

alejoe91:
This is ok with me! Thanks @jakeswann1 

@samuelgarcia ?

samuelgarcia:
This is lotally OK for me.
I hope to not regret that we merge 2 weeks before the relase.
Lets see if any bombs blow up after push the merge button.

samuelgarcia (approved):

PR #3097, opened 2024-06-27
Add `*sg_execution_times.rst` to gitignore.

JoeZiminski:
Further to the comment [here](https://github.com/SpikeInterface/spikeinterface/pull/2879#discussion_r1656732438), this PR adds `*sg_execution_times.rst` to the gitignore so it is ignored wherever it may be found. This file is generated automatically when building docs.


h-mayorquin (approved):
Thanks, there is also some zarr files that are created? I remember thinking something else from the docs build process was missing.

Issue #3100, opened 2024-06-27
One CI script to rule them all. Remaining improvements to testing


h-mayorquin:
- [x] Add windows and macos to the CI. https://github.com/SpikeInterface/spikeinterface/pull/2937
- [x] Add caching jobs for each os  https://github.com/SpikeInterface/spikeinterface/pull/3075
- [x] Drop caching of dependencies. https://github.com/SpikeInterface/spikeinterface/pull/3080
- [x] Add machinery for only testing the changed files in python https://github.com/SpikeInterface/spikeinterface/pull/3084
- [ ] Add plexon action to tests.
- [ ] Drop linux exclusive full tests script as this is already done in the single script.
- [ ] Add streaming tests and drop them. Now that we don't have ross they don't really need that isolation and they consume time and our priority on the queue. 
- [ ] Check that core behaves as in the core tests and drop core tests [mainly dependencies]
- [ ] Order test for less costly to more costly so they fail faster when they have to.
- [ ] Add dependencies for each of the modules to delay installation and make testing more efficiency.
- [ ] Use uv installer on the CI to decrease dependency installation time.
- [ ] Solve the need for editable installs in the CI. This probably indicates there is something wrong with our import structure.

Anyone interested in taking some of those points to learn CI? If you fancy learning about the import machinery and the mess that [PEP 660](https://peps.python.org/pep-0660/) introduced the editable install problem is a good intro. Ordering tests by time should rather easy as well as well as creating module based dependencies (e.g. the dependencies that are needed for running that specific module dependencies).

PR #3101, opened 2024-06-27
Add plexon2 tests action for install wine and drop full (linux only) tests

h-mayorquin:
Related to an effort to simplify tests in #3100. 

This PR modifes the wine action so it works with the three operating systems. Other modifications to the testing suite:
* Modifying plexon triggers the extractors test
* Full testing deependencies delayed farther away after pre-processing
* The data from gin is not downloaded unless the extractor suite is tested.


h-mayorquin:
This will need this merge on neo:

https://github.com/NeuralEnsemble/python-neo/pull/1494

h-mayorquin:
Ready to go @alejoe91

samuelgarcia (approved):

pyproject.toml
```diff
@@ -85,8 +85,10 @@ streaming_extractors = [
     "s3fs"
 ]
 
+
 preprocessing = [
     "scipy",
+    "ibllib>=2.36.0", # for IBL
```
alejoe91, line 91: @h-mayorquin 

After discussing with Sam, we think that this is not a `preprocessing` dependency per se, since it's only used for testing.

Maybe this extra should be `test_preprocessing` instead? (as we have for `test_extractors`)

pyproject.toml
```diff
@@ -85,8 +85,10 @@ streaming_extractors = [
     "s3fs"
 ]
 
+
 preprocessing = [
```
alejoe91, line 89:
```suggestion
test_preprocessing = [
```

pyproject.toml
```diff
@@ -124,10 +126,19 @@ test_core = [
...
+
+test_extractors = [
+    # Functions to download data in neo test suite
+    "pooch>=1.8.2",
+    "datalad>=1.0.2",
+    "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",
+    "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",
 ]
 
+
```
alejoe91, line 141: these cen also be factored out at module level:

- `test_sortingcomponents`
```
    "xarray",
    "huggingface_hub",
```

- `test_sorters`
```
    # tridesclous
    "numba",
    "hdbscan>=0.8.33",
``` 
- `test_widgets`
```
    # for sortingview backend
    "sortingview",
```


What do you guys think?

alejoe91: Maybe in a different PR

zm711: I think this is a great idea, but is a different PR from this current one :)  Since we have the dev install this will lighten that install and then the developer can layer on specific module dependencies if they want to work on it or just have the core developer dependencies if they don't.

h-mayorquin: Yes, @alejoe91, moving slowly in that direction. Will link this comment in the roadmap for testing improvement.

pyproject.toml
h-mayorquin: All right.

pyproject.toml
h-mayorquin: https://github.com/SpikeInterface/spikeinterface/issues/3100

alejoe91 (approved):

pyproject.toml
h-mayorquin: @alejoe91 , what aobut postprocessing and comparisons?

PR #3104, opened 2024-06-28
Since the gui now has curation we should allow users to set curation when making sorting summary

zm711:
Do we need to add a version check? Or can we always just put in the curation?


alejoe91 (approved):

samuelgarcia:
Thanks.
We also need to add options to transport the curation setting dict.
So lets merge and do it later.

PR #3105, opened 2024-06-29
update to run_sorter_jobs() and slurm

MarinManuel:
I am trying to run spikeinterface sorters on a HPC cluster and needed to pass specific parameters to `slurm`/`sbatch`.

I added the option to pass arbitrary arguments to `sbatch` in the `engine_kwargs`.

Let me know if that's causing any issues

### DIFF
--- a/src/spikeinterface/sorters/launcher.py
+++ b/src/spikeinterface/sorters/launcher.py
@@ -26,14 +26,14 @@ _default_engine_kwargs = dict(
     joblib=dict(n_jobs=-1, backend="loky"),
     processpoolexecutor=dict(max_workers=2, mp_context=None),
     dask=dict(client=None),
-    slurm=dict(tmp_script_folder=None, cpus_per_task=1, mem="1G"),
+    slurm=dict(tmp_script_folder=None, sbatch_args=dict(cpus_per_task=1, mem="1G")),
 )
 
 
 _implemented_engine = list(_default_engine_kwargs.keys())
 
 
-def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=False):
+def run_sorter_jobs(job_list, engine="loop", engine_kwargs=None, return_output=False):
     """
     Run several :py:func:`run_sorter()` sequentially or in parallel given a list of jobs.
 
@@ -55,10 +55,9 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=Fal
 
     Where *blocking* means that this function is blocking until the results are returned.
     This is in opposition to *asynchronous*, where the function returns `None` almost immediately (aka non-blocking),
-    but the results must be retrieved by hand when jobs are finished. No mechanisim is provided here to be know
-    when jobs are finish.
+    but the results must be retrieved by hand when jobs are finished. No mechanism is provided here to know
+    when jobs are finished.
     In this *asynchronous* case, the :py:func:`~spikeinterface.sorters.read_sorter_folder()` helps to retrieve individual results.
-
 
     Parameters
     ----------
@@ -68,10 +67,12 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=Fal
         The engine to run the list.
         * "loop" : a simple loop. This engine is
     engine_kwargs : dict
+        In the case of engine="slum", arguments to sbatch can be passed via sbatch_args, which is a dictionary whose
+        keys correspond to the --args to be passed to sbatch.
 
     return_output : bool, dfault False
         Return a sortings or None.
-        This also overwrite kwargs in  in run_sorter(with_sorting=True/False)
+        This also overwrite kwargs in run_sorter(with_sorting=True/False)
 
     Returns
     -------
@@ -81,6 +82,7 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=Fal
 
     assert engine in _implemented_engine, f"engine must be in {_implemented_engine}"
 
+    engine_kwargs = {} if None else engine_kwargs
     engine_kwargs_ = dict()
     engine_kwargs_.update(_default_engine_kwargs[engine])
     engine_kwargs_.update(engine_kwargs)
@@ -150,8 +152,6 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=Fal
         if tmp_script_folder is None:
             tmp_script_folder = tempfile.mkdtemp(prefix="spikeinterface_slurm_")
         tmp_script_folder = Path(tmp_script_folder)
-        cpus_per_task = engine_kwargs["cpus_per_task"]
-        mem = engine_kwargs["mem"]
 
         tmp_script_folder.mkdir(exist_ok=True, parents=True)
 
@@ -180,8 +180,8 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=Fal
                 )
                 f.write(slurm_script)
                 os.fchmod(f.fileno(), mode=stat.S_IRWXU)
-
-            subprocess.Popen(["sbatch", str(script_name.absolute()), f"-cpus-per-task={cpus_per_task}", f"-mem={mem}"])
+            sbatch_args = ' '.join(['--{k}={v}' for k, v in engine_kwargs['sbatch_args'].items()])
+            subprocess.Popen("sbatch", str(script_name.absolute()), sbatch_args)
 
     return out
 
--- a/src/spikeinterface/sorters/launcher.py
+++ b/src/spikeinterface/sorters/launcher.py
@@ -26,7 +26,7 @@ _default_engine_kwargs = dict(
     joblib=dict(n_jobs=-1, backend="loky"),
     processpoolexecutor=dict(max_workers=2, mp_context=None),
     dask=dict(client=None),
-    slurm=dict(tmp_script_folder=None, sbatch_args=dict(cpus_per_task=1, mem="1G")),
+    slurm={"tmp_script_folder": None, "sbatch_executable_path": "sbatch", "cpus-per-task": 1, "mem": "1G"},
 )
 
 
@@ -67,10 +67,16 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs=None, return_output=F
         The engine to run the list.
         * "loop" : a simple loop. This engine is
     engine_kwargs : dict
-        In the case of engine="slum", arguments to sbatch can be passed via sbatch_args, which is a dictionary whose
-        keys correspond to the --args to be passed to sbatch.
+        In the case of engine="slum", possible kwargs are:
+        - tmp_script_folder: str, default None
+            the folder in which the job scripts are created. Default: directory created by
+            the `tempfile` library
+        - sbatch_executable_path: str, default 'sbatch'
+            the path to the `sbatch` executable
+        - other kwargs are interpreted as arguments to sbatch, and are translated to the --args to be passed to sbatch.
+          see the [documentation for `sbatch`](https://slurm.schedmd.com/sbatch.html) for a list of possible arguments
 
-    return_output : bool, dfault False
+    return_output : bool, default False
         Return a sortings or None.
         This also overwrite kwargs in run_sorter(with_sorting=True/False)
 
@@ -82,7 +88,8 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs=None, return_output=F
 
     assert engine in _implemented_engine, f"engine must be in {_implemented_engine}"
 
-    engine_kwargs = {} if None else engine_kwargs
+    if engine_kwargs is None:
+        engine_kwargs = dict()
     engine_kwargs_ = dict()
     engine_kwargs_.update(_default_engine_kwargs[engine])
     engine_kwargs_.update(engine_kwargs)
@@ -148,10 +155,18 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs=None, return_output=F
 
     elif engine == "slurm":
         # generate python script for slurm
-        tmp_script_folder = engine_kwargs["tmp_script_folder"]
+        tmp_script_folder = engine_kwargs.pop("tmp_script_folder")
         if tmp_script_folder is None:
             tmp_script_folder = tempfile.mkdtemp(prefix="spikeinterface_slurm_")
         tmp_script_folder = Path(tmp_script_folder)
+        sbatch_executable = engine_kwargs.pop("sbatch_executable_path")
+
+        # for backward compatibility with previous version
+        if "cpus_per_task" in engine_kwargs:
+            warnings.warn("cpus_per_task is deprecated, use cpus-per-task instead", DeprecationWarning)
+            cpus_per_task = engine_kwargs.pop("cpus_per_task")
+            if "cpus-per-task" not in engine_kwargs:
+                engine_kwargs["cpus-per-task"] = cpus_per_task
 
         tmp_script_folder.mkdir(exist_ok=True, parents=True)
 
@@ -180,8 +195,14 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs=None, return_output=F
                 )
                 f.write(slurm_script)
                 os.fchmod(f.fileno(), mode=stat.S_IRWXU)
-            sbatch_args = ' '.join(['--{k}={v}' for k, v in engine_kwargs['sbatch_args'].items()])
-            subprocess.Popen("sbatch", str(script_name.absolute()), sbatch_args)
+
+            progr = [sbatch_executable]
+            for k, v in engine_kwargs.items():
+                progr.append(f"--{k}")
+                progr.append(f"{v}")
+            progr.append(str(script_name.absolute()))
+            p = subprocess.run(progr, capture_output=True, text=True)
+            print(p.stdout)
 
     return out
 


alejoe91:
@MarinManuel thank you! This is great!

The SLURM deployment has been unexplored.on our side, so we are very glad that it works for you. 

Would you mind sharing the script that you are using? I'd love to add an How To page on how to deploy multiple spike sorting jobs on SLURM

src/spikeinterface/sorters/launcher.py
```diff
@@ -68,10 +67,18 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=Fal
         The engine to run the list.
         * "loop" : a simple loop. This engine is
     engine_kwargs : dict
-
-    return_output : bool, dfault False
+        In the case of engine="slum", possible kwargs are:
+        - tmp_script_folder: str, default None
```
zm711, line 71:
```suggestion
        - tmp_script_folder : str, default: None
```

src/spikeinterface/sorters/launcher.py
```diff
@@ -68,10 +67,18 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=Fal
...
+            the folder in which the job scripts are created. Default: directory created by
```
zm711, line 72:
```suggestion
            the folder in which the job scripts are created. If None, it will be the directory created by
```

src/spikeinterface/sorters/launcher.py
```diff
@@ -68,10 +67,18 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=Fal
...
+            the `tempfile` library
+        - sbatch_executable_path: str, default 'sbatch'
```
zm711, line 74:
```suggestion
        - sbatch_executable_path : str, default: 'sbatch'
```

src/spikeinterface/sorters/launcher.py
```diff
@@ -68,10 +67,18 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=Fal
...
+            the path to the `sbatch` executable
+        - other kwargs are interpreted as arguments to sbatch, and are translated to the --args to be passed to sbatch.
+          see the [documentation for `sbatch`](https://slurm.schedmd.com/sbatch.html) for a list of possible arguments
+
+    return_output : bool, default False
```
zm711, line 79:
```suggestion
    return_output : bool, default: False
```

zm711 (commented):
@alejoe91 / @samuelgarcia know the launcher stuff better, is popping the best strategy or could you just pull out the kwargs without changing the dict as you go? 

Otherwise, I just added some fixes for our docstring style :)

src/spikeinterface/sorters/launcher.py
MarinManuel: Happy to follow whatever guidelines you recommend. The advantage of pop-ing out the few arguments that we want to use for something else is to not have to specify which aruments to `sbatch` we accept. That way, if `sbatch` were to add an additional argument, for instance, then we do not need to update spikeinterface to match.

zm711: Since I don't use slurm I'm not sure. Some times there are json dump of some params for provenance, so I would wait to see whatever Sam and Alessio say regarding that.

src/spikeinterface/sorters/launcher.py
```diff
@@ -26,14 +26,14 @@
...
-    slurm=dict(tmp_script_folder=None, cpus_per_task=1, mem="1G"),
+    slurm={"tmp_script_folder": None, "sbatch_executable_path": "sbatch", "cpus-per-task": 1, "mem": "1G"},
 )
 
 
 _implemented_engine = list(_default_engine_kwargs.keys())
 
 
-def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=False):
+def run_sorter_jobs(job_list, engine="loop", engine_kwargs=None, return_output=False):
```
JoeZiminski, line 36: 🙌

src/spikeinterface/sorters/launcher.py
```diff
@@ -26,14 +26,14 @@
     joblib=dict(n_jobs=-1, backend="loky"),
     processpoolexecutor=dict(max_workers=2, mp_context=None),
     dask=dict(client=None),
-    slurm=dict(tmp_script_folder=None, cpus_per_task=1, mem="1G"),
+    slurm={"tmp_script_folder": None, "sbatch_executable_path": "sbatch", "cpus-per-task": 1, "mem": "1G"},
```
JoeZiminski, line 29: I would lean towards having something like 'sbatch_kwargs' as a sub-dict, with all kwargs that are passed directly to sbatch held there. The reason is that later `engine_kwargs` is iterated over an all parameters passed to `sbatch`. In future when editing people may miss this and not `pop` new arguments before `engine_kwargs` is iterated over, so accidentally pass unrelated arguments to `sbatch`. So could make sense to have `"sbatch_kwargs"` as a sub-dict to make the separation very clear then iterate only over `engine_kwargs["sbatch_kwargs"]`.

src/spikeinterface/sorters/launcher.py
JoeZiminski, line 74: This is cool, out of interest what is the use case? will `sbatch` alone not always work?

src/spikeinterface/sorters/launcher.py
```diff
@@ -68,10 +67,18 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=Fal
...
+            the `tempfile` library
+        - sbatch_executable_path: str, default 'sbatch'
+            the path to the `sbatch` executable
+        - other kwargs are interpreted as arguments to sbatch, and are translated to the --args to be passed to sbatch.
+          see the [documentation for `sbatch`](https://slurm.schedmd.com/sbatch.html) for a list of possible arguments
+
+    return_output : bool, default False
         Return a sortings or None.
-        This also overwrite kwargs in  in run_sorter(with_sorting=True/False)
+        This also overwrite kwargs in run_sorter(with_sorting=True/False)
```
JoeZiminski, line 81:
```suggestion
        This also overwrites kwargs in run_sorter(with_sorting=True/False)
```

src/spikeinterface/sorters/launcher.py
```diff
@@ -146,12 +155,18 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=Fal
 
     elif engine == "slurm":
         # generate python script for slurm
-        tmp_script_folder = engine_kwargs["tmp_script_folder"]
+        tmp_script_folder = engine_kwargs.pop("tmp_script_folder")
```
JoeZiminski, line 158: Here the popping is done so these arguments are not passed to sbatch. If using a subdict (comment above) I would revert these just to passing the variable so the code-reader does not have to track the state of `engine_kwargs` when reading.

src/spikeinterface/sorters/launcher.py
```diff
@@ -146,12 +155,18 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=Fal
...
         if tmp_script_folder is None:
             tmp_script_folder = tempfile.mkdtemp(prefix="spikeinterface_slurm_")
         tmp_script_folder = Path(tmp_script_folder)
-        cpus_per_task = engine_kwargs["cpus_per_task"]
-        mem = engine_kwargs["mem"]
+        sbatch_executable = engine_kwargs.pop("sbatch_executable_path")
+
+        # for backward compatibility with previous version
+        if "cpus_per_task" in engine_kwargs:
+            warnings.warn("cpus_per_task is deprecated, use cpus-per-task instead", DeprecationWarning)
```
JoeZiminski, line 166: Nice, I guess this is to match the syntax of SLURM? 👍

src/spikeinterface/sorters/launcher.py
```diff
@@ -146,12 +155,18 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=Fal
...
+            cpus_per_task = engine_kwargs.pop("cpus_per_task")
+            if "cpus-per-task" not in engine_kwargs:
```
JoeZiminski, line 168: I wonder if better (though a bit more verbose) to raise an error here, if the user has accidentally passed two versions of the same kwarg it would be better to let them know I think.

src/spikeinterface/sorters/launcher.py
```diff
@@ -181,7 +196,13 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=Fal
                 f.write(slurm_script)
                 os.fchmod(f.fileno(), mode=stat.S_IRWXU)
 
-            subprocess.Popen(["sbatch", str(script_name.absolute()), f"-cpus-per-task={cpus_per_task}", f"-mem={mem}"])
+            progr = [sbatch_executable]
+            for k, v in engine_kwargs.items():
+                progr.append(f"--{k}")
+                progr.append(f"{v}")
+            progr.append(str(script_name.absolute()))
+            p = subprocess.run(progr, capture_output=True, text=True)
```
JoeZiminski, line 204: will `subprocess.run()` mean this is no longer run asynchronously vs. `subprocess.Popen()`?

JoeZiminski (commented):
This is very cool! really opens up the SLURM options here. I have a couple of comments now, and will try it out on our HPC ASAP!

src/spikeinterface/sorters/launcher.py
MarinManuel: This was what I had originally done, but then I thought it made the kwargs pretty cluttered and changed my mind. But I can see how that would help avoid mistakes in the future, I will change the code.

src/spikeinterface/sorters/launcher.py
MarinManuel: Just in case sbatch is not in the PATH. Unlikely but why not ?

src/spikeinterface/sorters/launcher.py
MarinManuel: Sbatch returns immediately after submitting the job to the queue, so it's still going to be asynchronous. The run() syntax is just a simpler way to interact with the subprocess module and the one I have more experience with, that's all

JoeZiminski: Okay nice ofc sbatch will just return, agree `run` is the preferred way to use `subprocess`. Nice!

src/spikeinterface/sorters/launcher.py
JoeZiminski: I can see why this would be useful but I think in terms of codebase maintainability it would be better that the user handles this on their end by adding `sbatch` to PATH, given that it is (I think?) a real edge case. Otherwise, if this is a general pattern that should be adopted it will require adding such arguments everywhere (e.g. if a sorter uses `matlab`, it will have to expose a path to the `matlab` executable, similar for anything that calls a system executable).

MarinManuel: That's fair

src/spikeinterface/sorters/launcher.py
```diff
@@ -68,10 +67,17 @@ def run_sorter_jobs(job_list, engine="loop", engine_kwargs={}, return_output=Fal
...
         * "loop" : a simple loop. This engine is
     engine_kwargs : dict
-
-    return_output : bool, dfault False
+        In the case of engine="slum", possible kwargs are:
+        - tmp_script_folder: str, default None
+            the folder in which the job scripts are created. Default: directory created by
+            the `tempfile` library
+        - sbatch_args: dict
+          arguments to be passed to sbatch. They are translated to the --args form.
```
JoeZiminski, line 75: Maybe slightly reword to make a little clearer, that the only modification is the addition of `--`. There might be some confusion as previously it could take underscores for `cps_per_task` but I think this now must be `cpus-per-task`. So many something like 'arguments to be passed directly to sbatch. They will be automatically prefixed with `--`.  Arguments must be in the format slurm specify, see [documentation for sbatch]().

JoeZiminski:
Hey @MarinManuel sorry for the delay in testing this locally, I tried it on our cluster and seems to be working well! I had a couple of other comments below, it might also be worth adding a test for this. I have added pasted one in the dropdown below, I don't think I can push to your fork, but feel free to add it to `src/spikeinterface/sorters/tests/test_launcher.py` underneath `test_run_sorter_jobs_slurm`. 

If you think of any functionality this is missing, please let me know / feel free to expand the test. This will also require adding `pytest-mock` to the `dev` dependencies in `pyproject.toml` in the main repo folder. It would be cool to also test the generated `.py` files sometime, but that's outside the scope of this PR. 

<details><summary>The test</summary>

```
def test_run_sorter_jobs_slurm_kwargs(mocker, tmp_path, job_list):
    """
    Mock `subprocess.run()` to check that engine_kwargs are
    propagated to the call as expected.
    """
    # First, mock `subprocess.run()`, set up a call to `run_sorter_jobs`
    # then check the mocked `subprocess.run()` was called with the
    # expected signature. Two jobs are passed in `jobs_list`, first
    # check the most recent call.
    mock_subprocess_run = mocker.patch(
        "spikeinterface.sorters.launcher.subprocess.run"
    )

    tmp_script_folder = tmp_path / "slurm_scripts"

    engine_kwargs = engine_kwargs=dict(
        tmp_script_folder=tmp_script_folder,
        sbatch_args={
            "cpus-per-task": 32,
            "mem": "32G",
            "gres": "gpu:1",
            "any_random_kwarg": 12322,
        }
    )
    run_sorter_jobs(
        job_list,
        engine="slurm",
        engine_kwargs=engine_kwargs,
    )

    script_0_path = f"{tmp_script_folder}/si_script_0.py"
    script_1_path = f"{tmp_script_folder}/si_script_1.py"

    expected_command = [
        "sbatch", "--cpus-per-task", "32", "--mem", "32G", "--gres", "gpu:1", "--any_random_kwarg", "12322", script_1_path
    ]
    mock_subprocess_run.assert_called_with(expected_command, capture_output=True, text=True)


    # Next, check the fisrt call (which sets up `si_script_0.py`)
    # also has the expected arguments.
    expected_command[9] = script_0_path
    assert mock_subprocess_run.call_args_list[0].args[0] == expected_command

    # Next, check that defaults are used properly when no kwargs are
    # passed. This will default to `_default_engine_kwargs` as
    # set in `launcher.py`
    run_sorter_jobs(
        job_list,
        engine="slurm",
        engine_kwargs={"tmp_script_folder": tmp_script_folder},
    )
    expected_command = [
        "sbatch", "--cpus-per-task", "1", "--mem", "1G", script_1_path
    ]
    mock_subprocess_run.assert_called_with(expected_command, capture_output=True, text=True)

    # Finally, check that the `tmp_script_folder` is generated on the
    # fly as expected. A random foldername is generated, just check that
    # the folder to which the scripts are saved is in the `tempfile` format.
    run_sorter_jobs(
        job_list,
        engine="slurm",
        engine_kwargs=None,  # TODO: test defaults
    )
    tmp_script_folder = "_".join(tempfile.mkdtemp(prefix="spikeinterface_slurm_").split("_")[:-1])
    assert tmp_script_folder in mock_subprocess_run.call_args_list[-1].args[0][5]
```

</details>

Some other small things:
- I think that the other, currently existing test (`test_run_sorter_jobs_slurm`) might need to be updated to the new way of passing the `sbatch_args`. 
- I think it is worth adding an error if any other key-value pair exists in `engine_kwargs` except for `tmp_script_folder` or `sbatch_args`. The reason is this PR is backward-incompatible with previous versions but will not error, only silently fail to pass the kwargs onwards to slurm (e.g. if `cpus_per_task` is passed to `engine_kwargs` directly as it was previously, it will just be ignored). So good to raise an error here to let the user know about the new API.
- Also, might as well also assert / raise an error if that `engine_kwargs` is only not `None` when `engine="slurm"`/

MarinManuel:
Hi @JoeZiminski 
I tried to integrate your comments to my PR. Hopefully I did it right.
The only thing I did not do was

> Also, might as well also assert / raise an error if that engine_kwargs is only not None when engine="slurm"/

I wasn't entirely sure what you meant, but I think `sbatch` would launch with default parameters if engine_kwargs=None, so that shouldn't be a problem?
Also, I didn't know how to write a test to check that the exception was raised correctly.

src/spikeinterface/sorters/launcher.py
samuelgarcia: honestly we really should discuss this more!
I think I really do not like it.

JoeZiminski: I'll make a thread on this / cc on the existing discussion to continue the chat from last week, I think this is also a good example to consider. If your happy to leave as-is for now I'll link it on that issue and we can change to later with the rest of such cases.

JoeZiminski: #3226

src/spikeinterface/sorters/tests/test_launcher.py
```diff
@@ -126,6 +126,154 @@ def test_run_sorter_jobs_slurm(job_list, create_cache_folder):
...
+        job_list,
+        engine="slurm",
+        engine_kwargs=None,  # TODO: test defaults
+    )
+    tmp_script_folder = "_".join(tempfile.mkdtemp(prefix="spikeinterface_slurm_").split("_")[:-1])
+    assert tmp_script_folder in mock_subprocess_run.call_args_list[-1].args[0][5]
+
+
+@pytest.mark.skip("Slurm launcher need a machine with slurm")
+def test_run_sorter_jobs_slurm_kwargs(mocker, tmp_path, job_list):
```
JoeZiminski, line 204: I think this test got duplicated in the git history (there are now two copies). I think one of the two copies can be deleted.

src/spikeinterface/sorters/tests/test_launcher.py
```diff
@@ -126,6 +126,154 @@ def test_run_sorter_jobs_slurm(job_list, create_cache_folder):
...
+    run_sorter_jobs(
+        job_list,
+        engine="slurm",
+        engine_kwargs=None,  # TODO: test defaults
+    )
+    tmp_script_folder = "_".join(tempfile.mkdtemp(prefix="spikeinterface_slurm_").split("_")[:-1])
+    assert tmp_script_folder in mock_subprocess_run.call_args_list[-1].args[0][5]
+
+
+@pytest.mark.skip("Slurm launcher need a machine with slurm")
```
JoeZiminski, line 203: I think you can remove this, the test uses monkeypatching so `subprocess.run` call to SLURM is never actually run

JoeZiminski: BTW I think this will result in the test always being skipped, [`skipif`](https://docs.pytest.org/en/stable/how-to/skipping.html#id1) can be used to skip under the condition that SLURM is not found on the system (but that's not required here)

JoeZiminski:
Hey @MarinManuel, so sorry for the delay with this! I'm around again now, please feel free to ping me and we can merge this ASAP.

Just a couple of small things I think:
- please see a couple of small comments on the tests
- there are a few github suggestions (e.g. from Zach) about. These can be commited by selected 'commit suggestion', changing the commit message (e.g. just 'docstring fix' or something is okay) and then 'commit changes'. It's a cool github feature!
- Any other comments that are outdated or have been addressed can be resolved.

> Also, might as well also assert / raise an error if that engine_kwargs is only not None when engine="slurm"

Sorry this was the other way round, in case a user who is unfamiliar with SLURM e.g. using 'dask' say but passes the `engine_kwargs` under some false impression these will do something for all engines. Alternatively we could rename `engine_kwargs` to `slurm_engine_kwargs` as they only relate to slurm. Actually this is better, maybe they can be called `slurm_engine_kwargs` or just `slurm_kwargs`?

After this I can do one last pass and test locally again then I think it's good to go! Apologies the review took so long, this is a super-useful addition.

MarinManuel:
Hi @JoeZiminski 
Please let me know how that looks

JoeZiminski:
Hi @MarinManuel, I'm incredibly sorry I misunderstood from the docstring how engine kwargs was intended, I thought it was just for SLURM from the below:

<details> <summary> old docstring </summary>

```
engine_kwargs : dict
        In the case of engine="slum", possible kwargs are:
        - tmp_script_folder: str, default None
            the folder in which the job scripts are created. Default: directory created by
            the `tempfile` library
        - sbatch_args: dict
          arguments to be passed to sbatch. They will be automatically prefixed with --.
          Arguments must be in the format slurm specify, see the
          [documentation for `sbatch`](https://slurm.schedmd.com/sbatch.html) for a list of possible arguments
```

</details>

It might be worth reverting that change such that the all kwargs come under 'engine kwargs' as before. The new docstring can look as below. I'm happy to do this if you give me access to your fork as I appreciate it will be annoying to undo something you just did due to my mistake!

<details> <summary> New docstring </summary>

```
    engine_kwargs : dict
        Parameters to be passed to the underlying engine.
        Defaults are:
        * loop : None
        * joblib : n_jobs=1, backend="loky"
        * multiprocessing : max_workers=2, mp_context=None
        * dask : client=None
        * slurm : tmp_script_folder=None
        for the `slurm` entry:
            This dictionary contains arguments to be passed to sbatch.
            They will be automatically prefixed with --.
            Arguments must be in the format slurm specify, see the
            [documentation for `sbatch`](https://slurm.schedmd.com/sbatch.html) for a list of possible arguments
```
</details>

Some other very minor things that came up when manually testing, then I think good to go!

- Need to add `pytest-mock` [here](https://github.com/SpikeInterface/spikeinterface/blob/73f4d58d2d542e607155d49d30716cc28215e12f/pyproject.toml#L123) and [here](https://github.com/SpikeInterface/spikeinterface/blob/73f4d58d2d542e607155d49d30716cc28215e12f/pyproject.toml#L148) in the `pyproject.toml` as it was a new dependency added in the test.
- the test is failing on windows as some `os` commands used in the `run_sorter_job` assume we are on Linux. In the end we should skip only when not running on Linux. We need to add `from platform import system` at the top of `test_launcher.py` and `@pytest.mark.skipif("system() != 'Linux'", "Assumes we are on Linux to run SLURM")`.
- I was manually testing and it was hard to figure out what was being passed to slurm by the CLI. I think we can add a quick print to before the subprocess call to make this clear, e.g. adding `print(f"subprocess called with command {progr}\n")` [here](https://github.com/SpikeInterface/spikeinterface/blob/e7071704b01a6db7580255993a9c58cfb18d39fe/src/spikeinterface/sorters/launcher.py#L206).
-
</details>

Then all good from my end, sorry about this last-minute back and forth and thanks again for this super useful addition!

JoeZiminski:
Hi @MarinManuel thanks for this, apologies please resend the invite I will fix the tests and conflicts, thanks!

samuelgarcia:
Hi @MarinManuel. Thanks a lot for this improvement. 
We ultra sorry for the long delay.
This is OK for us.

Thank you @JoeZiminski for taking care of this.

samuelgarcia (approved):

PR #3115, opened 2024-07-01
Out-of-place multiply needed when using amplitude scaling in the drifty hybrid recording

cwindolf:


cwindolf:
(Oops Pierre, I didn't mean to do that, but if you have the time haha...)

alejoe91 (approved):

Issue #2515, opened 2024-02-26
Times not propagated by `.save_to_memory()`


cwindolf:
I was a bit surprised to see that the following is possible:

```
rec = si.read_binary_folder(...)
rec1 = rec.save_to_memory()
print(rec.sample_index_to_time(0), rec1.sample_index_to_time(0))
# => 2451.763033333333, 0.0
```

I'm working around this by `rec1.set_times(rec.get_times())`, but it seems to me that since `save()` maintains the times when saving to other formats (like the binary folder), it should do that for memory mode as well.

samuelgarcia:
Is it based on time vector or t_start ?

cwindolf:
This particular recording has t_start and no time_vector

cwindolf:
To get it working, I ended up doing
```
rec1._recording_segments[0].t_start = rec0._recording_segments[0].t_start
rec1._kwargs["t_starts"] = [rec0._recording_segments[0].t_start]
```

h-mayorquin:
Related:
https://github.com/SpikeInterface/spikeinterface/issues/2627

PR #3072, opened 2024-06-24
Add 'How SpikeInterface handles time' documentation

JoeZiminski:
(The documentation as rewritten since first posted. See the drop down for original PR message).

This PR updates the existing tutorial for times [here](https://spikeinterface.readthedocs.io/en/stable/tutorials/core/plot_6_handle_times.html#sphx-glr-tutorials-core-plot-6-handle-times-py) to highlight some of the new time handling methods.

Currently the section on 'shift_start_times' doesn't exist but I wrote this doc to see how this API would feel, so it is a kind of prototype docs for the new method on #3509. This PR will await #3509. The build docs are [here](https://spikeinterface--3072.org.readthedocs.build/en/3072/tutorials/core/plot_6_handle_times_ver2.html).

TODO:
- sample_index_to_to_time` and `time_to_sample_index` are currently not in the API docs
<details> <summary> Old PR message </summary>
- check all sections that are not running, because they are below the line that crashes because #3509 function doesn't exist.

<details> <summary> Old PR message </summary>

In-progress draft of a page detailing how time is represented in SpikeInterface. closes #2627. Once a first draft is finished will require a lot of revision I'm sure as I'm not certain on a lot of this.

Worth discussion is where this goes, at present is a How-to but it is a little tenuous. Maybe it can be reworded to 'How to handle time on SpikeInterface recordings' with a bit more code added.

(An aside, please excuse if there already solutions to this I have missed). As the documentation grows API changes will frequently break existing documentation. One solution is to add a template to all PR that has a tick-box stating that all relevant documentation must be reviewed and updated before merge. This is good practice anyway but is error-prone. There seem to be some existing solutions to testing code snippets within documentation: pylit[https://pypi.org/project/pylit/] [sphinx](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html). Actually I see from searching @h-mayorquin might be able to advise on this!

EDIT: some things to do before continuing
- currently I don't think it is possible to manually set `t_start` on a recording object. Instead it is only loaded from certain extractors. As far as I can tell. It would be cool to make this accessible, this would then make the tutorial make more sense.
- currently generate_ground_truth_recording has no `t_start` properly and it is not possible to set times that then relate to the sorting.  Would be useful to add this before continuing here. discussed in #3071.

<details>

Issue #2627, opened 2024-03-26
Understanding how Spikeinterface handles recording times.


JoeZiminski:
I always thought that SI was essentially time-agnostic and computed times on the fly based on sample number and sampling rate. However I had the same experience as @yger when reading [this thread](https://github.com/SpikeInterface/spikeinterface/discussions/2612#discussioncomment-8877517) and was surprised time information is tracked.

I just seen [this docs page](https://spikeinterface.readthedocs.io/en/latest/modules_gallery/core/plot_6_handle_times.html) which introduces the concept which is great, I had not seen it before. However it might still be useful to flesh it out a bit and add what happens in certain scenarios e.g. 1) slicing recording 2) concatenate recording 3) appending / concatenating segments.

I'd be happy to do this ASAP (just (hopefully 🤞 ) finishing off an unrelated project and will have more time for ephys). But I thought I'd get the ball rolling here in case there are other cases I had not considered.

h-mayorquin:
This is great.
This would be useful for us in neuroconv.

If you have time to improve the docs on that I will be happy to review and discuss.

I remember that @samuelgarcia changed his views about time recently after a discussion with @cwindolf .

We should sort this out as well:
https://github.com/SpikeInterface/spikeinterface/issues/2515

samuelgarcia:
I changed my view about time handling in the motion.
In short we are about to create a motion object that will handle multi segment and also absolut time.

For recording, we are time agnostic by default but you can have a t_start and also a time vector.

We wanted to move this into base so that Sorting also also inherits this 3 times mode (t_start=None, t_start=something or times vector) But this is not started yet.

Time handliong is a bit hidden here : https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/baserecording.py#L689

alejoe91:
Thanks @JoeZiminski ! this sounds great :)

Note that for some extractors, e.g. `read_openephys` you can load the syncrhonized timestamps directly ([`load_sync_timestamps` param](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/extractors/neoextractors/openephys.py#L122)). It would be important to mention!

PR #3117, opened 2024-07-01
Expose `t_start` in `BaseRecording`

JoeZiminski:
This PR exposes `t_start` by allowing it to be set through `set_times()`. Now `set_times` `times` parameter can take a vector (`np.ndarray`) or `int | float`. If the format it is treated as a time vector and if the latter a `t_start`. Previously `t_start` were only available if set when loading from file.

The PR extends the tests to cover the `t_start` cases and another couple of cases such as `save_to_memory()`, which currently has a bug (#2515). In creating these tests another couple of small issues were found, I cherry-picked the fixes to different PRs to keep the diffs easier to review. For ease though if everyone is happy all can stay here. They will fail until this branch is rebased back on master once the other PRs are reviewed.

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -458,12 +458,26 @@ class BaseRecording(BaseRecordingSnippets):
         return d["time_vector"] is not None
 
     def set_times(self, times, segment_index=None, with_warning=True):
-        """Set times for a recording segment.
+        """Set times for a recording segment. Any existing times
+        will be overwritten.
+
+        Times can be manually set on the recording segment. If times are
+        not set, the sample index and sampling frequency are used to
+        calculate time. Otherwise, `t_start` or `time_vector` can be
+        provided:
+
+        `t_start` - the start time for the segment. The times for
+            this recording segment will be calculated as
+            t_start + sample_index * (1 / sampling_frequency)
+
+        `time_vector` - A vector of length segment.get_num_samples()
+            that holds the exact time for each sample in the recording.
 
         Parameters
         ----------
-        times : 1d np.array
-            The time vector
+        times : float | 1d np.array
+            If `int`, this is the `t_start` for the segment,
+            otherwise, it is the time vector.
         segment_index : int or None, default: None
             The segment index (required for multi-segment)
         with_warning : bool, default: True
@@ -472,11 +486,18 @@ class BaseRecording(BaseRecordingSnippets):
         segment_index = self._check_segment_index(segment_index)
         rs = self._recording_segments[segment_index]
 
-        assert times.ndim == 1, "Time must have ndim=1"
-        assert rs.get_num_samples() == times.shape[0], "times have wrong shape"
+        if isinstance(times, float) or isinstance(times, int):
+            rs.t_start = times
+            rs.time_vector = None
+        elif isinstance(times, np.ndarray):
 
-        rs.t_start = None
-        rs.time_vector = times.astype("float64", copy=False)
+            assert times.ndim == 1, "Time must have ndim=1"
+            assert rs.get_num_samples() == times.shape[0], "times have wrong shape"
+
+            rs.t_start = None
+            rs.time_vector = times.astype("float64", copy=False)
+        else:
+            raise TypeError("`times` must be an integer / float (`t_start`) or " "numpy array (`time_vector`).")
 
         if with_warning:
             warnings.warn(
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -475,8 +475,8 @@ class BaseRecording(BaseRecordingSnippets):
 
         Parameters
         ----------
-        times : float | 1d np.array
-            If `int`, this is the `t_start` for the segment,
+        times : int | float | 1d np.array
+            If `int` or `float`, this is the `t_start` for the segment,
             otherwise, it is the time vector.
         segment_index : int or None, default: None
             The segment index (required for multi-segment)
--- a/src/spikeinterface/core/tests/test_time_handling.py
+++ b/src/spikeinterface/core/tests/test_time_handling.py
@@ -1,9 +1,328 @@
+import copy
+
 import pytest
 import numpy as np
 
 from spikeinterface.core import generate_recording, generate_sorting
+import spikeinterface.full as si
 
 
+class TestTimeHandling:
+
+    # Fixtures #####
+    @pytest.fixture(scope="session")
+    def raw_recording(self):
+        """
+        A three-segment raw recording without times added.
+        """
+        durations = [10, 15, 20]
+        recording = generate_recording(num_channels=4, durations=durations)
+        return recording
+
+    @pytest.fixture(scope="session")
+    def time_vector_recording(self, raw_recording):
+        """
+        Add time vectors to the recording, returning the
+        raw recording, recording with time vectors added to
+        segments, and list a the time vectors added to the recording.
+        """
+        return self._get_time_vector_recording(raw_recording)
+
+    @pytest.fixture(scope="session")
+    def t_start_recording(self, raw_recording):
+        """
+        Add a t_starts to the recording, returning the
+        raw recording, recording with t_starts added to segments,
+        and a list of the time vectors generated from adding the
+        t_start to the recording times.
+        """
+        return self._get_t_start_recording(raw_recording)
+
+    def _get_time_vector_recording(self, raw_recording):
+        """
+        Loop through all recording segments, adding a different time
+        vector to each segment. The time vector is the original times with
+        a t_start and irregularly spaced offsets to mimic irregularly
+        spaced timeseries data. Return the original recording,
+        recoridng with time vectors added and list including the added time vectors.
+        """
+        times_recording = copy.deepcopy(raw_recording)
+        all_time_vectors = []
+        for segment_index in range(raw_recording.get_num_segments()):
+
+            t_start = segment_index + 1 * 100
+            offsets = np.arange(times_recording.get_num_samples(segment_index)) * (
+                1 / times_recording.get_sampling_frequency()
+            )
+            time_vector = t_start + times_recording.get_times(segment_index) + offsets
+
+            all_time_vectors.append(time_vector)
+            times_recording.set_times(times=time_vector, segment_index=segment_index)
+
+            assert np.array_equal(
+                times_recording._recording_segments[segment_index].time_vector,
+                time_vector,
+            ), "time_vector was not properly set during test setup"
+
+        return (raw_recording, times_recording, all_time_vectors)
+
+    def _get_t_start_recording(self, raw_recording):
+        """
+        For each segment in the recording, add a different `t_start`.
+        Return a list of time vectors generating from the recording times
+        + the t_starts.
+        """
+        t_start_recording = copy.deepcopy(raw_recording)
+
+        all_t_starts = []
+        for segment_index in range(raw_recording.get_num_segments()):
+
+            t_start = (segment_index + 1) * 100
+
+            all_t_starts.append(t_start + t_start_recording.get_times(segment_index))
+            t_start_recording.set_times(times=t_start, segment_index=segment_index)
+
+            assert np.array_equal(
+                t_start_recording._recording_segments[segment_index].t_start,
+                t_start,
+            ), "t_start was not properly set during test setup"
+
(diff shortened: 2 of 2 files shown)


src/spikeinterface/core/tests/test_time_handling.py
```diff
@@ -1,9 +1,317 @@
...
+
+        sorting = generate_sorting(num_units=10, durations=durations)
+
+        sorting.register_recording(recording_to_attach)
+        assert sorting.has_recording()
+
+        return sorting
 
 
+# TODO: deprecate original implementations ###
```
JoeZiminski, line 314: TODO: deleting was messing up the diff so will do at end.

src/spikeinterface/core/tests/test_time_handling.py
```diff
@@ -1,9 +1,328 @@
...
```
JoeZiminski, line 325: TODO: this was messing up the diff so left for the end.

h-mayorquin:
We do use the time machiner of spikeinterface in neuroconv. I would like to participate on the review. Where should I start?

JoeZiminski:
Thanks @h-mayorquin! That would be great, I think:

1) https://github.com/SpikeInterface/spikeinterface/pull/3119 is a very quick one mostly unrelated to others
2) Is also mostly standalone https://github.com/SpikeInterface/spikeinterface/pull/3118 only related to the `time_vector` concept.
3) I think then this PR (I will undraft it now) for the changes to `t_start`
4) then https://github.com/SpikeInterface/spikeinterface/pull/3120 which fixes a failing test case here.

Cheers!

src/spikeinterface/core/baserecording.py
```diff
@@ -458,12 +458,26 @@ def has_time_vector(self, segment_index=None):
...
+            t_start + sample_index * (1 / sampling_frequency)
+
+        `time_vector` - A vector of length segment.get_num_samples()
+            that holds the exact time for each sample in the recording.
 
         Parameters
         ----------
-        times : 1d np.array
-            The time vector
+        times : int | float | 1d np.array
```
h-mayorquin, line 478: I personally would prefer to not overload this function and create a new one instead. `set_t_start`.

 What are the advantages of overloading this? How are you thinking about it?

src/spikeinterface/core/tests/test_time_handling.py
```diff
@@ -1,9 +1,328 @@
...
+
+    def _get_time_vector_recording(self, raw_recording):
+        """
+        Loop through all recording segments, adding a different time
+        vector to each segment. The time vector is the original times with
+        a t_start and irregularly spaced offsets to mimic irregularly
+        spaced timeseries data. Return the original recording,
+        recoridng with time vectors added and list including the added time vectors.
+        """
+        times_recording = copy.deepcopy(raw_recording)
```
h-mayorquin, line 49: we have clone for this as an extractor method but if you really require this, why make the raw recording fixture per session?

h-mayorquin (changes requested):
What is the purpose of this? I think knowing about how you intend this to be used would help clarify the debate and inform the API.

I find the testing a bit complicated to read.  I think there is too much indirection of the fixtures. I will have to do a second reading but I think that's a code smell.

Plus, it seems that many of the tests don't really correspond hee. Why are the tests of memory no in the memory PR. If you are iterating over the segments already on some of the fixtures you can just set their t_start attribute directly, you don't need a special method at the `BaseRecording` level,  it is a public private at the `BaseRecordingSegment`

I am requesting changes here because I think some of the testing does not correspond to this PR.

src/spikeinterface/core/baserecording.py
JoeZiminski: Yes I couldn't decide on this vs. separate functions and @samuelgarcia suggested this approach. On balance I think I prefer it as `t_start` and `times_vector` are mutually exclusive ways of setting custom times, so it makes sense to change in one place. It would be slightly strange to call `set_t_start()` and this removes the `time_vector` attribute. But I'm not sure either way.

src/spikeinterface/core/tests/test_time_handling.py
JoeZiminski: The benefit of the `raw_recording` fixture is that `durations` only needs to be defined once, then copied as `set_times()` is in place. But I agree it is a lot of indirection and it is probably more readable to incorporate into the individual fixtures, possibly with `DURATIONS=[...]` set at the top of the script?

src/spikeinterface/core/baserecording.py
h-mayorquin: Interesting. Thanks for sharing, I will add this as something to discuss at the meeting.

h-mayorquin: I guess that my take is the following:

We use `set_times` because the sampling recording is _slightly_ irregular when we want more precision. Why we would need to set `t_start`. 

The use case that I can think off is that we would like to shift all the recording to the right or the left on time. But if that is the use case I would be better to have a method that shift the recording times and works independently of whether times are handled internally with t_start and sampling frequency or a time vector.

If it is not for shifting I can't thik on other use of setting `t_start`.

JoeZiminski:
Thanks @h-mayorquin! The problme is that `t_start` and `time_vecetor` are mutually exclusive and so if you set `time_vector` the `t_start` is removed (and, it requires the save vice-versa). Therefore some kind of setter is required to remove `time_vector` if it is set when setting `t_start`. Also, `t_start` and `time_vector` are core ways of handling custom times in SpikeInterface so for me it seems unusual to have a setter for one but the other is set through a public attribute on a segment.

Agree the tests are kind of messed up, in adding these for the `t_start` I found some other bugs and split them into separate PRs to ease review, but was too lazy to do the tests 😅 . The tests in the other PRs require some of the machinery in this PR, so happy to merge this PR first removing some of those tests, then add them in other PRs once this is merged. 

Tomorrow I can try and remove some of the indrection in the tests and remove tests that do not correspond directly to this PR, or at least, are failing if included in this PR. WDYT?

src/spikeinterface/core/baserecording.py
JoeZiminski: I think the use case would be if you had separate sessions in a single day, for example a session, 10 minute break to change some equiptment, and another recording session. If these sessions are held as different segments on a recording (or, as separate recordings) the researcher may want to hold the true recording times (e.g. session 1 started at 1 pm, session two at 1.30 pm).

h-mayorquin:
> Thanks @h-mayorquin! The problme is that t_start and time_vecetor are mutually exclusive and so if you set time_vector the t_start is removed (and, it requires the save vice-versa). Therefore some kind of setter is required to remove time_vector if it is set when setting t_start. Also, t_start and time_vector are core ways of handling custom times in SpikeInterface so for me it seems unusual to have a setter for one but the other is set through a public attribute on a segment.

Yes, I think this is at the core of why I don't like this approach. It requires the final user to be aware of internal details of spikeinterface. I feel it mixes how spikeinterface handles time internally with what the user wants with its recording object. Let me elaborate:

I think that a good interface will abstract those implementations details away from the user and will allow them to express things that they want. Let me illustrate a different approach with an example:

* My times are iregular so I will set the correct times that I got from a TTL process with `set_times`.
* But also, this is the second segment of an experiment and it should start an hour after the first.

What I calim is that this should be as easy as 
```python
set_times(times_from_ttl)
set_t_start(t=time_when_the_second_segment_started)
```

The interface proposed in this PR will say: **sorry, those things are exclusive because of this internal spikeinterface reason**, what you need to do is to compose on your own the time_vector, shift it, and then use set_times differently for each segment.

What I am proposing is that we have a `set_segment_start_time` that is agnostic to the internal implementation details of spikeinterface: It should do what the user wants

* If the recording has a time_vector because the format is like that or the user set those times before,  then when the user calls `set_segment_start_time` it should modify the time vector such that the time_vector[0] is t_start.
* If the recording handles time with sampling_frequency and t_start then it should modify `t_start` to what the user wants.

Advance users or testing casses like the one in this PR can modify the attributes directly for stuff that is not covered.

src/spikeinterface/core/baserecording.py
h-mayorquin: OK, change my answer below what I think this type of case is not well served with the current implementation.

JoeZiminski:
Thanks Herberto, I like the semantics of the `set_segment_start_time()` it is very clear. I also don't like the mutual excusivility of the `t_start` and `time_vector` mechanisms and its requirement for the user to understand the SpikeInterface internals. I have put some feedback on the proposed approach in the dropdown below (as not to bloat the response).



I think this is a very important discussion to have but it would require a major reworking of how SpikeInterface handles time which I think is outside the scope of the PR. I think we should make an issue to discuss this, I'd be very keen as it is important for #3072.

For now, let me better motivate this PR with respect to the existing implementation of times:

1) A present, custom times can be represented as `t_start` or `time_vector`. `time_vector` has it's own setter `set_times()`. However `t_start` is not possible to set except for directly on the segment, which is an unused pattern in the SpikeInterface API. The only proper way to set `t_start` through the API is through some extractors when loading certain file types. So `t_start` is in this weird state of sometimes turning up in the codebase but isn't actually settable generally. 
2) Therefore I think there are two possible approaches. Either `t_start` is not a useful concept and it should be removed from SpikeInterface (maybe if `t_start` is required in extractors it can be set to a full `time_vector` during load) (I am not averse to this). Alternatively, it should be fully exposed and documented in the SpikeInterface API (this PR takes the second approach).
3) Tests are added to cover all uses of `time_vector` and `t_start` in the core machinery.

Let me know what you think of the above, I think there is definately room for further optimisation for the API generally, but I think this PR represents a digestible improvement on the current situation (but alternatively would agree with removing `t_start` from the SpikeInterface side and relying entirely on `time_vector`).

h-mayorquin:
> Or, maybe set_times can take a time_vector that do not start at zero. But now there are two way to set the segment start times for time vector (directly through the set time vector or through this second approach).

Yes, that's how I am thinking about it. This is the way that is now. I don't understand the first two examples but maybe they rely on this behavior not being available? Maybe not? If not, can you illustrate the examples?

> It could lead to the case where somone sets set_segment_start_time then later adds the time_vector, but now their time vector is wrong because the previously set t_starts are interacting. This would be quite stange behaviour that you set_times() then get_times() and the get_times() is different.

This one I agree is confusing. The current behavior is that `set_times` is the source of true so I will keep that overwrite behavior. `set_segment_start_time` is a utility that works either if you have a time vector or not but if you set the time vector then that overides what `set_segment_start_time` did before.

> A present, custom times can be represented as t_start or time_vector. time_vector has it's own setter set_times(). However t_start is not possible to set except for directly on the segment, which is an unused pattern in the SpikeInterface API. The only proper way to set t_start through the API is through some extractors when loading certain file types. So t_start is in this weird state of sometimes turning up in the codebase but isn't actually settable generally.

Yeah, mabye is an improvement, but I don't think is a big one, setting the t_start at the segment level is fine for the purposes of this PR for example which is testing but if we are gona discuss how to expose a simpler API to the end user then I want to have this more general discussion that we are having.

We should not try to remove the concept of handling the internal timings with t_start because that is there for memory efficiency reasons but I maybe the user does not need to interact with it. Another possibility to the one I am proposing is only setings times with a time vector but transforming them to t_start, sampling_frequency internally if they are regular enough. I general I think we should separate internal handing from how the user interacts with him and avoid a power user bias.

JoeZiminski:
Thanks! Agree these can be split into two separate discussions. Would you agree with:

1) Move the from tests here to #3120 (with some refactoring of the tests) and close this PR
2) Open an issue to discuss the time API and how this should be presented in #2627

h-mayorquin:
Mmm this is just my opinion, I don't know if I have convinced you. I think that Sam will side with you if you want to make your life more convenient here

 But I think not using this machinery for the tests is a good idea regardless of whether we make this a user interface or not.

JoeZiminski:
I think it's worthwhile discussing further, if we pick this lane (i.e. what is introduced in this PR) and write the documentation it will be hard to undo and there may be better options. I definately agree there is room to make this cleaner and more intutiive, for me though it is safer to keep `t_start` and `time_vector` distinct concepts rather than allow them to interact. I'll move the tests to the other PR ASAP, and then we can keep this PR with bare minimum changes and discuss the API further here.

JoeZiminski:
I will close this PR, the tests can be included in #3120 #3118 and we are converging on a much better time API in #3157

PR #3118, opened 2024-07-01
Add time vector case to `get_durations`.

JoeZiminski:
This PR extends `BaseRecording.get_duration()` (and the associated `get_total_duration()`) but allowing it to handle the `time_vector` case. It is tested in #3117 (see that PR for why).

In reality, this should only have an extremely minimal effect, only in cases with some irregularities in the time-interval spacing. Nonetheless I guess it makes sense to use the true `time_vector` for the duration if it exists as it will be different form estimating based on number of sample and sampling frequency.

The calculation is `t_stop - t_start + sampling_step`. The sampling step is introduced due to fencepost error. For example, if you have `ts=0.1`, `num_samples=10`, `duration=1`, you will have `10*0.1=1`, a 1s recording as expected. These will be represented by times `[0, 0.1, ..., 0.9]`. Therefore we need to add 1x `ts` to `t_stop - t_start` to get the true duration. As the `ts` is usually a very small number the duration calculated in this way is not nice and round as it usually is with `num_samples / sampling_frequency`.


src/spikeinterface/core/baserecording.py
```diff
@@ -233,8 +233,14 @@ def get_duration(self, segment_index=None) -> float:
             The duration in seconds
         """
         segment_index = self._check_segment_index(segment_index)
-        segment_num_samples = self.get_num_samples(segment_index=segment_index)
-        segment_duration = segment_num_samples / self.get_sampling_frequency()
+
+        if self.has_time_vector(segment_index):
+            times = self.get_times(segment_index)
+            segment_duration = times[-1] - times[0] + (1 / self.get_sampling_frequency())
```
h-mayorquin, line 239: Got you, makes sense. Also, if you were to have only one sample this will give you a segment_duration of 0 which does not make sense while your formula gives you 1/ sampling_rate which is correct.

That said, it might be that something set times because the sampling rate is not quite right or something else and while that might be a good approximation it would be better if you did:  `duration_of_samples = np.diff(times).mean()` 

It does not look very costly:

image

samuelgarcia:
We agree with this but we should check carfully that the internal use of this get_total_duration/get_dutaion is correcyly use anywhere in the code.
And also the `sortinganalyzer.get_ total_duration` should be change also.

src/spikeinterface/core/baserecording.py
JoeZiminski: Nice! I wasn't sure how to handle this, thanks for the profiling also! Will add

h-mayorquin: Wait, I think that's wrong. Maybe not:

image

JoeZiminski: I see yes, on my crappy laptop I can't allocate the space to create the 24h recording 😆 In this case, because the `sampling_rate` should be a good approximation to what the true sampling rate will be (I think?), and it is only adding one sample step value at the end, maybe we stick with the sampling rate? WDYT?

h-mayorquin: Yes, my hunch is to go for the performant code (adding the 1/sr) giving that is a small adjustment.

JoeZiminski:
[Currently editing / testing]
Good point I will list their uses in the codebase below. I don't *think* any of these should be affected:

#### `recording.get_duration()`
- In [baserecording](https://github.com/SpikeInterface/spikeinterface/blob/52372a60b8bbdd1838caa792ed630b13cf488b9d/src/spikeinterface/core/baserecording.py#L69) it is only used for representation methods.
- In `widgets/traces.py`, it is used to set the end time of the recording to display. TODO: TEST
- In `NoiseGeneratorRecording` for [silence_periods](https://github.com/SpikeInterface/spikeinterface/blob/52372a60b8bbdd1838caa792ed630b13cf488b9d/src/spikeinterface/preprocessing/silence_periods.py#L82) TODO: TEST
- In [generate_byhrid_recordings](https://github.com/SpikeInterface/spikeinterface/blob/52372a60b8bbdd1838caa792ed630b13cf488b9d/src/spikeinterface/generation/hybrid_tools.py#L402) where they are passed to `generate_sorting_kwargs` for `generate_sorting`. TODO: TEST 
- In [`test_benchmark_motion_interpolation()`][https://github.com/SpikeInterface/spikeinterface/blob/52372a60b8bbdd1838caa792ed630b13cf488b9d/src/spikeinterface/sortingcomponents/benchmark/tests/test_benchmark_motion_interpolation.py#L31] I think this is OK, as is currently working and would require the test to be changed to break in some way.

#### `get_total_duration()` (`recording` and `sorting_analyzer`)

The two places the function is used that is not tests / docs is:
- `widgets/peak_activivity.py`](https://github.com/SpikeInterface/spikeinterface/blob/52372a60b8bbdd1838caa792ed630b13cf488b9d/src/spikeinterface/widgets/peak_activity.py#L75)
- [`qualitymetrics/misc_metrics.py`](https://github.com/SpikeInterface/spikeinterface/blob/52372a60b8bbdd1838caa792ed630b13cf488b9d/src/spikeinterface/qualitymetrics/misc_metrics.py#L92)

JoeZiminski:
Oh yes @samuelgarcia the sorting analyzer needs changing too! How to do you think it is best to structure this, to avoid duplciating the generation code across sorter and recording? Can the `sorting_analyzer` function call the `get_total_duration()` on the `self.recording`?

alejoe91:
@JoeZiminski since the analyzer can be recordingless, I would do the following:
```
def get_total_duration(self) -> float:
    if self.has_recording() or self.has_temporary_recording():
        duration = self.recording.get_total_duration()
    else:
        duration = self.get_total_samples() / self.sampling_frequency
    return duration
```

JoeZiminski:
Cool thanks @alejoe91 I will add this mode. For testing, I can't figure out how to make a recordingless analyser 😄 I guess it is not with `create_sorting_analzyer`?

alejoe91:
> Cool thanks @alejoe91 I will add this mode. For testing, I can't figure out how to make a recordingless analyser 😄 I guess it is not with `create_sorting_analzyer`?

Not the most elegant of solutions, but this works:
```
sorting_analyzer._recording = None
```
:)

alejoe91:
@JoeZiminski I pushed a small fix in hybrid tools! Thanks for adding tests. Let us know when it's done!

JoeZiminski:
Thanks @alejoe91 will push final commit shortly!

samuelgarcia (approved):

Issue #3029, opened 2024-06-14
Proposal for internal handling of `time_vector`


cwindolf:
Following some discussions with @samuelgarcia and @JoeZiminski , I wanted to open an issue to work out the details of handling time_vectors in a way that won't crash everyone's computers. I'm opening a separate issue from https://github.com/SpikeInterface/spikeinterface/issues/2627, since this will focus on implementation details and will not change the logic.

Summary of what exists:
 - User-facing API: users interact with times through `extractor.sample_index_to_time` and `extractor.time_to_sample_index`
   - This means that we have freedom to change the implementation details without users knowing
 - There are two logical cases:
   - `time_seconds = t_start + sample_index / sampling_frequency`, where `t_start=0` by default
   - `time_seconds = time_vector[sample_index]`

Summary of problems:
 - The first case causes no issues.
 - The second causes two issues:
   - For an hour long recording, a time vector takes ~100M samples * (8 bytes / sample) ~= 1GB of memory. This becomes a problem for parallel processing.
   - Going from times in seconds to sample indices with a time_vector currently [means a `np.searchsorted` for each sample](https://github.com/SpikeInterface/spikeinterface/blob/bd89c998afb29c204167a4f7aff360a1cd631da5/src/spikeinterface/core/baserecording.py#L826), which could get expensive for two reasons: time_vectors are long so individual searches can be expensive, and doing lots of them is even worse.

I propose that we can still efficiently implement time vectors with two changes:
 - Store them in a SharedMemory when doing multiprocessing
 - Don't do a searchsorted for every sample input into `extractor.sample_index_to_time`
   - For instance, to speed up individual searches, we could initialize the search at (time_seconds - t_start) * sampling_frequency and look left or right depending on what value we find there.
   - To speed up multiple searches, if the array of times we're searching for is also sorted than we can use a merge or something if it seems helpful.

I think this could be addressed by a PR that doesn't affect any other code, and would allow time_vectors in the motion correction interpolation.

cwindolf:
A PR here could try to fix https://github.com/SpikeInterface/spikeinterface/issues/2515 as well

h-mayorquin:
Related:
https://github.com/SpikeInterface/spikeinterface/issues/2627

h-mayorquin:
Question:
In what case do we need to call `time_to_sample_index` repeteadly?

My prior is that np.searchsorted is not gona be the bottleneck for anything reasonable if we accept vectors in the corresponding functions:
[...]

PR #3119, opened 2024-07-01
Round instead of int for `time_to_sample_index`.

JoeZiminski:
In `BaseRecordingSegment.time_to_sample_index()` the calculation from time was cast from `float` to `int`. By default this will use floor method of rounding, which can lead to wrong results e.g. doing:

```
sample = 500
time_ = recording.sample_index_to_time(sample)
sample_again = recording.time_to_sample_index(time_)
```
Could lead to `sample` not equalling `sampling_again`.

This PR is tested in #3118.

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -848,8 +848,10 @@ class BaseRecordingSegment(BaseSegment):
                 sample_index = time_s * self.sampling_frequency
             else:
                 sample_index = (time_s - self.t_start) * self.sampling_frequency
+            sample_index = round(sample_index)
         else:
             sample_index = np.searchsorted(self.time_vector, time_s, side="right") - 1
+
         return int(sample_index)
 
     def get_num_samples(self) -> int:


alejoe91 (approved):

h-mayorquin (approved):
LGTM. I also think that at some point we should vectorize these functions as discussed in  https://github.com/SpikeInterface/spikeinterface/issues/3029

samuelgarcia (approved):

zm711:
I just learned a fun fact about rounding in python

```python
>>> round(4.5)
4
```

Python tries to reduce bias in rounding and using the .5 rule actual introduces an upward bias. (I think @h-mayorquin and @JoeZiminski would be most interested in this).

JoeZiminski:
Cool @zm711 I did not know about that!! They taught us wrong in school!

h-mayorquin:
I don't remember that they taught us the rounding up rule in school so maybe our system was different. 

I actually never read about how rounding works in python and the different rules, thanks Zach for this rabbit hole. Check this out:

```python
In [12]: [i + .5 for i in range(-5, 5)]
Out[12]: [-4.5, -3.5, -2.5, -1.5, -0.5, 0.5, 1.5, 2.5, 3.5, 4.5]

In [13]: [round(i + .5) for i in range(-5, 5)]
Out[13]: [-4, -4, -2, -2, 0, 0, 2, 2, 4, 4]
```

I wonder if @DradeAW is aware of this as I remember we discussed some of this behavior when changing int to round at some point.

zm711:
It does raise the question of expectation vs the choice to reduce bias. I would have predicted that round followed the rule that Joe and I learned in school as a default rather than trying to eliminate biases. So round is less consistent than float unless we change the rules ourselves. But It is super cool. Definitely would love to hear what @DradeAW thinks. I also can't remember if we do any `np.round` here or in Neo and whether the behaviors are the same.

DradeAW:
I was not aware of that, this is super weird!
I was always taught to round away from 0 when it was perfectly x.5

Do you think having exactly .5 happens in real cases?

zm711:
My prediction is that all of our schools just taught a simple binary decision because that is easy for humans. Knowing that .5 rounds up for some numbers and down for others so it hard for a person to keep track of. For a computer we can actually implement sophisticated rules to try to do this statistically. But that's just my guess. I haven't checked but I heard there are multiple implementations possible for rounding in general all with pros and cons (typically speed vs bias).

DradeAW:
I think for float, 0.5 is almost never going to happen, unless you have a specific conversion to µV (in which case it could be frequent).

But I think this problem is less problematic than `int()` which doubles the amount of zeros ...

PR #3120, opened 2024-07-01
Fix `t_starts` not propagated to `save_to_memory`.

### LINKED ISSUE #2515: Times not propagated by `.save_to_memory()`

JoeZiminski:
This PR closes #2515 by propagated `t_starts` to memory recording objects during `save_to_memory`. This is tested in #3117. 

This works to propagate the times to save to memory, and tests are passing, but I havn't considered other uses for these recording classes. So it would be great if others could let me know if this might break anything elsewhere! I can't see why it would as it uses default kwargs that maintain the old behaviour, but still I'm paranoid 😅 

Tests are added here that cover propagation of time information when stored as `t_start` and `time_vector` to various methods of saving the data. Also ,tests for other time-related functions (e.g. `sample_time_to_index()` are added).


alejoe91 (approved):

h-mayorquin (changes requested):
Let's move the corresponding tests from https://github.com/SpikeInterface/spikeinterface/pull/3117.

I suspect that you tested this in #3117 because you wanted to have a canonical way for seting the t_start in your tests but you don't really need it, t_start is a public attribute of the class. 

Plus, you are adding interface in #3117 and the tests should not depend on the newly added interface.

src/spikeinterface/core/baserecording.py
```diff
@@ -545,11 +545,11 @@ def _save(self, format="binary", verbose: bool = False, **save_kwargs):
             if kwargs.get("sharedmem", True):
                 from .numpyextractors import SharedMemoryRecording
 
-                cached = SharedMemoryRecording.from_recording(self, **job_kwargs)
+                cached = SharedMemoryRecording.from_recording(self, t_starts=t_starts, **job_kwargs)
```
samuelgarcia, line 548: Cool. thanks a lot I forgot it.

src/spikeinterface/core/numpyextractors.py
```diff
@@ -85,7 +85,7 @@ def __init__(self, traces_list, sampling_frequency, t_starts=None, channel_ids=N
         }
 
     @staticmethod
-    def from_recording(source_recording, **job_kwargs):
+    def from_recording(source_recording, t_starts=None, **job_kwargs):
```
samuelgarcia, line 88: Not sure to agree with this. The t_tsarts should be taken from the source recording. No ?

src/spikeinterface/core/numpyextractors.py
```diff
@@ -211,18 +212,16 @@ def __del__(self):
                 shm.unlink()
 
     @staticmethod
-    def from_recording(source_recording, **job_kwargs):
+    def from_recording(source_recording, t_starts=None, **job_kwargs):
```
samuelgarcia, line 215: same here.

src/spikeinterface/core/numpyextractors.py
samuelgarcia: Mayybe we should make a private method for this
https://github.com/SpikeInterface/spikeinterface/pull/3120/files#diff-6b93c507ba0be7d51533ba5d6d6bb00e9b4481bd887646d4e436496164ba0ec8R503-R508

and use it at several places. No ?

alejoe91: done in last commit

src/spikeinterface/core/numpyextractors.py
alejoe91: done

JoeZiminski:
Thanks both, I will move the tests over now

src/spikeinterface/core/tests/test_time_handling.py
```diff
@@ -1,9 +1,292 @@
...
+
+        sorting = generate_sorting(num_units=10, durations=durations)
+
+        sorting.register_recording(recording_to_attach)
+        assert sorting.has_recording()
+
+        return sorting
 
 
+# TODO: deprecate original implementations ###
```
JoeZiminski, line 289: Deleting these led to the diff becoming very messed up and hard to read. If everyone is happy, I'll remove these original tests (the new tests include cover these cases).

JoeZiminski:
@h-mayorquin @samuelgarcia @alejoe91 I re-requested review as added tests from #3117 to here, these cover all saving behaviour for both time representations + some other time-related functions. Herberto I tried to reduce the indirection on the fixtures by a level, let me know if it is still not clear.

alejoe91 (approved):
LGTM!!

Removed the deprecated tests ;)

PR #3121, opened 2024-07-02
Remove unit_params_range from generate.py

### LINKED ISSUE #3077: Expose `unit_params_range` in test file generation
There is a cool argument to `generate_ground_truth_recording()` called `unit_params_range` that should sent the bounds of the template shapes. However as far as I can tell it is not actually linked up to anything. A solution would be to take `unit_params_range` as an argument [here](https://github.com/SpikeInterface/spikeinterface/blob/d963aa26e31a9b6bc0bea2a4973f76d4dbafb062/src/spikeinterface/core/generate.py#L1408) and use it as default / update the defaults if it is not empty.

JoeZiminski:
closes #3077 by removing the unused `unit_params_range` in `generate.py`. This previously allowed the user to set the parameters for templates injected into synthetic recordings. However I think it was superceeded by `unit_params`. @samuelgarcia you were correct on #3077 `unit_params` can take a tuple in which case it will use these as bounds to generate template parameters randomly. So `unit_params` performs the job of `unit_params_range`.

### DIFF
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -1440,7 +1440,6 @@ def generate_templates(
     dtype="float32",
     upsample_factor=None,
     unit_params=None,
-    unit_params_range=None,
     mode="ellipsoid",
 ):
     """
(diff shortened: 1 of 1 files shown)


zm711:
Any chance this needs a deprecation? Or is this new enough that it can just be deleted without annoying users?

JoeZiminski:
Good question, my impression is it is quite new and it is not widely used. Also, it was not connected to anything so if was used it wasn't doing anything. As such, I think either remove or raise an error if trying to use saying it doesn't do anything, use `unit_params` instead, and remove version after next. Happy with both approaches 🤔

samuelgarcia:
Merci Joe
I am agree that we could remove this without warning it was somhow a hidden option that nobody used.

Issue #2626, opened 2024-03-26
A thread to discuss inter-session alignment.


JoeZiminski:
Following a discussion on the Slack channel about inter-session alignment I thought I'd start an issue to summarise / discuss these things and come up with a plan on how this could be handled in Spikeinterface.

There are a couple of (not necessarily mutually exclusive) approaches:

**1) post-hoc unit-matching**
This approach matches units from independently sorted sessions. Current approaches includes Unitmatch(https://www.biorxiv.org/content/10.1101/2023.10.12.562040v1.full.pdf). SI and Unitmatch teams are working together to ensure compatibility such that SI-sorted data can be easily read in Unitmatch. There is also the approach outlined in [this paper](https://elifesciences.org/reviewed-preprints/92495) which looks interesting. Typically the motion correction in these methods relies on putative matches between units across sessions, which may fail in the presence of large (e.g. >50 um?) drift. Therefore alignment of either the traces (pre-sorting) or templates (post-sorting) across sessions, would compliment these methods.

**2) Concatenating sessions and motion correcting**
Another approach is to concatenate two sessions together and perform motion correction as if they were a single session, then sort. A concern is how current intra-session methods will perform in the presence of significant drift between sessions (manifesting as a sudden, large 'jump' in the middle of the concatenated recording). In the NP2 paper the Kilosort team suggest this approach does not work so well with the iterative template method*. In the [Dredge paper](https://www.biorxiv.org/content/10.1101/2023.10.24.563768v1) this approach appears to be used successfully. It would be interesting to see at what level of inter-session drift (linear and nonlinear) this approach fails for the available methods. I guess the machinery for such tests has already been largely developed [here](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC10897502/). 

**3) Alignment of raw data pre-sorting**
This approach is described in the NP2 paper, roughly it goes like 1) within-session motion correction for two sessions 2) compute drift between session and add compensation to within-session offsets 3) interpolate. In theory it should also be quite possible to do with the existing `correct_motion` machinery, just requiring some additional logical to incorporate inter-session registration offsets prior to interpolation. 

**4) Alignment of template waveforms post-sorting**
[...]

PR #3126, opened 2024-07-02
Add `correct_inter_session_displacement` function

JoeZiminski:
This PR requires #3231 , on which it is rebased. Currently this PR contains commits from that PR so it makes sense for that to be merged before reviewing this in depth as the diff will be confusing.

The problem to be solved is inter-session movement ('displacement') of the probe. This is very similar problem to the inter-session motion correction, and can use many of the same tools. The idea is accept as input a list of recording objects (and, optionally, a list of intra-session motion correction outputs). For each session, an activity histogram will be generated, and then the inter-session drift and necessary correction estimated. The correction is then applied to the recording object, but a `displacement_info` object (similar to `motion_info` object) will be output. This could be used to correct peaks directly and optionally later, correct templates (see #2626).

Under the hood it essentially using the kilosort drift correction approach (cross correlation for linear alignment, and rigid alignment within multiple segments along the probe y-axis for non-rigid alignment). To generate the histogram for each session, the session is first split into segments and a histogram calculated for each segment, and then either the mean or median taken. The idea is to downweight periods of a session that may have noise. I tried a few different approaches (e.g. maximum value, first eigenvalue) but decided it was overkill and just kept the mean / median for now).

This PR makes a few changes to existing code to refactor anything I needed so it can be shared across modules, for the most part though the changes are contained within new modules.

Tests are added and a new tutorial is added in the documentation.

Question
- One issue I had is that if I add the module to preprocessing `__init__` I get a circular import issue because I am calling some `sortingcomponents.motion` modules. Please see the comment mode in the `preprocessing/__init__.py` and uncomment to reproduce. I think this can be solved by moving some of the motion code into core? For now, the functions are not in the API docs for this reason.

<details><summary> Example Script </summary>

```
from spikeinterface.generation.session_displacement_generator import generate_session_displacement_recordings

from spikeinterface.preprocessing.inter_session_alignment import (
    session_alignment,
)
from spikeinterface.widgets import plot_session_alignment, plot_activity_histogram_2d

si.set_global_job_kwargs(n_jobs=10)


if __name__ == '__main__':

    # --------------------------------------------------------------------------------------
    # Load / generate some recordings
    # --------------------------------------------------------------------------------------

    recordings_list, _ = generate_session_displacement_recordings(
        num_units=5,
        recording_durations=[25, 25, 25],
        recording_shifts=((0, 0), (0, -200), (0, 150)),
        non_rigid_gradient=0.1,
        seed=55,
    )

    peaks_list, peak_locations_list = session_alignment.compute_peaks_locations_for_session_alignment(
        recordings_list,
        detect_kwargs={"method": "locally_exclusive"},
        localize_peaks_kwargs={"method": "grid_convolution"},
    )

    # --------------------------------------------------------------------------------------
    # Do the estimation
    # --------------------------------------------------------------------------------------
    # For each session, an 'activity histogram' is generated. This can be `entire_session`
    # or the session can be chunked into segments and some summary generated taken over then.
    # This might be useful if periods of the recording have weird kinetics or noise.
    estimate_histogram_kwargs = session_alignment.get_estimate_histogram_kwargs()
    estimate_histogram_kwargs["method"] = "chunked_median"
    estimate_histogram_kwargs["log_transform"] = False
    
    corrected_recordings_list, extra_info = session_alignment.align_sessions(
        recordings_list,
        peaks_list,
        peak_locations_list,
        alignment_order="to_session_1",  # "to_session_X" or "to_middle"
        estimate_histogram_kwargs=estimate_histogram_kwargs,
    )

    plot_session_alignment(
        recordings_list,
        peaks_list,
        peak_locations_list,
        extra_info["session_histogram_list"],
        **extra_info["corrected"],
        spatial_bin_centers=extra_info["spatial_bin_centers"],
        drift_raster_map_kwargs={"clim":(-250, 0), "scatter_decimate": 10}
    )
    plt.show()

    if estimate_histogram_kwargs["histogram_type"]  == "2d":
        plot_activity_histogram_2d(
            extra_info["session_histogram_list"],
            extra_info["spatial_bin_centers"],
            extra_info["corrected"]["corrected_session_histogram_list"]
        )
        plt.show()
```

</details>

image


debugging/session_alignment.py
```diff
@@ -0,0 +1,977 @@
...
+    the chunked histograms.
+
+    Returns
+    -------
+    A dictionary with entries:
+
+    "bin_um" : number of spatial histogram bins. As the estimated peak
+        locations are continuous (i.e. real numbers) this is not constrained
+        by the number of channels.
+    "method" : may be "chunked_mean", "chunked_median", "chunked_supremum",
```
JoeZiminski, line 41: missing "entire_session"

debugging/session_alignment.py
```diff
@@ -0,0 +1,977 @@
...
+    The estimated displacement will be added to the existing recording.
+
+    Parameters
+    ----------
+    recordings_list : list[BaseRecording]
+        A list of motion-corrected (`InterpolateMotionRecording`) recordings.
+    motion_info_list : list[dict]
+        A list of `motion_info` objects, as output from `correct_motion`.
+        Each entry should correspond to a recording in `recording_list`.
+    rigid: bool
```
JoeZiminski, line 266: delete this

debugging/session_alignment.py
```diff
@@ -0,0 +1,977 @@
...
+
+    Returns
+    -------
+    `corrected_recordings_list : list[BaseRecording]
+        List of displacement-corrected recordings (corresponding
+        in order to `recordings_list`). If an input recordings is
+        an  InterpolateMotionRecording` recording, the corrected
+        output recording will be a copy of the input recording with
+        the additional displacement correction added.
+    `motion_objects_list : list[Motion]
```
JoeZiminski, line 179: this is wrong now

src/spikeinterface/sortingcomponents/motion/motion_utils.py
```diff
@@ -168,6 +168,18 @@ def get_spatial_bin_edges(recording, direction, hist_margin_um, bin_um):
     return spatial_bins
 
 
+def get_spatial_bins(recording, direction, hist_margin_um, bin_um):
+    # TODO: could this be merged with the above function?
```
JoeZiminski, line 172: I guess where `get_spatial_bin_edges` is used this could be used instead

src/spikeinterface/widgets/motion.py
```diff
@@ -203,6 +203,7 @@ def __init__(
                 peak_amplitudes = peak_amplitudes[peak_mask]
 
         from matplotlib.pyplot import colormaps
+        from matplotlib.colors import Normalize
```
JoeZiminski, line 206: small fixes in this module

Antorhythms:
Hi @JoeZiminski (and everyone following along) — first, thank you for this PR. I finally got a chance to use it on a multi-session Neuropixels dataset, and inter-session alignment via align_sessions_after_motion_correction works well (qualitatively better than any standard motion correction at session boundaries). Would you (and the dev team) be open to rebasing this PR onto the current main and merging it? I believe it would be a great addition to spikeinterface and the SOTA for multisession alignment. I would be very interested in this being merged to be able to scale up its use on a computing cluster in an environment that uses the latest spikeinterface. 

In addition, while using it, I built two extensions on top of your branch that improved my results, which I'd be happy to add in a follow-up once this branch is merged.
1. Sequential alignment mode — In addition to "to_middle"/"to_session_N", I added a mode that aligns each session to its immediate predecessor (k → k-1) and accumulates the pairwise shifts (cumsum) into a single coordinate frame, reusing _compute_session_alignment pairwise under the hood. For datasets spanning many sessions or days, where non-adjacent sessions don't always correlate well enough for a global "to_middle"/reference-session estimate, chaining adjacent-only comparisons gave me more reliable shifts.
2. Non-rigid spatial bin inheritance — When non-rigid inter-session alignment is used after non-rigid within-session motion correction, I pass the same win_step_um/win_scale_um/win_margin_um used at motion-correction time so the inter-session window centers match motion.spatial_bins_um by construction. I also included an np.interp-based fallback that remaps shifts onto the target bins on the occasion they don't line up exactly. This avoids the assertion in _add_displacement_to_interpolate_recording (assert np.array_equal(motion_ref.spatial_bins_um, new_non_rigid_window_centers)) failing when bins drift apart.

JoeZiminski:
Hi @Antorhythms thanks for this! Apologies I was at FENS when this was posted, coincidentally I happened to see your poster on motor cortex projections! I think it was presented by your colleague so didn't see you in person, cool work!

Thanks for reviving this and testing it out, I'm glad to hear it's been useful. I don't think there is any reason it can't be merged it just got a little lost over time. It will require #3231 to be merged first but that should be relatively quick. I'll make sure to align these back with main and discuss with @samuelgarcia and @alejoe91 

Both those additions sound excellent, once this is merged please do go ahead and open the PRs they would be great additions. Will get back to you ASAP on this!

Antorhythms:
Hey Joe! Thanks a lot for getting back to this and also checking out the poster at FENS! Unfortunately, I could not be there this time.

Just let me know when this PR is merged so I can make a new one with updates and thanks a lot for the hard work on this!

JoeZiminski:
Thanks @Antorhythms! Sounds good will keep you updated, I forgot people are mostly away in August but will push this in September, cheers!

PR #3129, opened 2024-07-02
Add streaming extractor tests to main  test script

h-mayorquin:
Part of the effort in https://github.com/SpikeInterface/spikeinterface/issues/3100

Two improvements for the streaming extractors:
1. The streaming extractors are tested on all the OSs.
2. We have one workflow less, less cue, more priority, etc.

I am thinking that maybe we can only one test for all extractors and not have a separate sets of tests but I am less sure about IBL which was problematic at some point.

Maybe let's keep it like this for a while and then we can change in the future if does not give problems for a while.


src/spikeinterface/extractors/iblextractors.py
```diff
@@ -309,7 +309,7 @@ class IblSortingExtractor(BaseSorting):
     name = "ibl"
     installation_mesg = "IBL extractors require ibllib as a dependency." " To install, run: \n\n pip install ibllib\n\n"
 
-    def __init__(self, pid, good_clusters_only=False, load_unit_properties=True, one=None):
+    def __init__(self, pid: str, good_clusters_only: bool = False, load_unit_properties: bool = True, one=None):
```
samuelgarcia, line 312: So good.

samuelgarcia (approved):

samuelgarcia:
merci a lot.

src/spikeinterface/extractors/iblextractors.py
h-mayorquin: Sorry Sam, needed a cange to trigger the CI : P

PR #3130, opened 2024-07-03
Fix serializability of `InjectDriftingTemplatesRecording`

### LINKED ISSUE #3125: `.save_to_folder()` fails for the `InjectDriftingTemplatesRecording`
When trying to save a hybrid recording, I get the error `TypeError: Object of type DriftingTemplates is not JSON serializable`. This makes sense, because DriftingTemplates inherits from Templates which is a dataclass and not a BaseExtractor, so it will not be `.to_dict()`ed by https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/base.py#L423 even though it has a `.to_dict()` method.

I'm not sure how to resolve this? One could maybe make Templates subclass BaseExtractor while still being a dataclass? Although it does feel a bit strange to json-serialize a large array of templates. Anyway, wanted to see what you all think, thanks!

alejoe91:
Fixes #3125

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -7,7 +7,6 @@ import warnings
 import weakref
 import json
 import pickle
-import os
 import random
 import string
 from packaging.version import parse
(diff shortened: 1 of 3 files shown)


src/spikeinterface/sortingcomponents/peak_selection.py
```diff
@@ -76,19 +76,18 @@ def select_peaks(
...
     if margin is not None:
         to_keep = np.zeros(len(selected_peaks), dtype=bool)
-        offset = 0
-        for segment_index in range(recording.get_num_segments()):
-            duration = recording.get_num_frames(segment_index)
+        for segment_index in range(num_segments):
+            num_samples_in_segment = recording.get_num_samples(segment_index)
             i0, i1 = np.searchsorted(selected_peaks["segment_index"], [segment_index, segment_index + 1])
-            while selected_peaks["sample_index"][i0] <= margin[0] + offset:
+            while selected_peaks["sample_index"][i0] <= margin[0]:
```
alejoe91, line 86: @samuelgarcia @yger 

This is a fix for #2980 :

1. The spikes need to be lexsorted based on segment and sample indices
2. sample indices start from 0 for each segment, so you don't need an offset!

alejoe91: I can make a separate PR for it if you prefer

alejoe91: made a separate PR #3132

Issue #1963, opened 2023-09-06
Seed behavior for generate


h-mayorquin:
I did not have time to participate in the review of https://github.com/SpikeInterface/spikeinterface/pull/1948. 

We added the following auxiliary function to generate:

https://github.com/SpikeInterface/spikeinterface/blob/3359647eb7a75a4f258590a46b3f665b7aac9d1d/src/spikeinterface/core/generate.py#L16-L24

Which generates a different seed at runtime when no default seed is provided.

I think I understand the rationality, it tries to achieve two things.
1) The results of the generator should be the same after saving it and loading (@samuelgarcia, thanks for adding tests for this!).
2) Every call to a generator should generate new traces if the seed is not specified. Let's call this variability by default.

I don't agree completly with 2. When less experienced users utilize let's stay `generate_ground_truth_recording` with default parameters I want the function to have deterministic behavior. This is better for debugging and error finding as when I run the function on my system it wiill be as similar as possible as their run. With the current behavior given by `ensure_seed` I will not be able to reproduce their exact results unless I have their specific seed that was generated at runtime. This can be extracted but it will be difficult for new users and it will create one more round of questions and requests when discussing an issue. A more severe version of this problem is presented if users find an error when they run `generate_ground_truth_recording` (because the specific seed was problematic), but then they run the function again and the error dissappear. This undermines their trust in their algorithm (they just know something is wrong but can't reproduce) and makes it hard for them to report the problem to us.

I think that @samuelgarcia made this decision from a power user point of views as this is more convenient and it makes sense semantically: `NoiseGeneratorRecording` should generate different noise everytime you call it (just like `np.random.rand`). But I think error reporting and debugging is an imporant use case of these functions which I think is better served by having a specific seed by default. This is better for less experienced users. Power users that want to actually generate variability are in a better position to know what a seed is, change it themselves and roll a function like `ensure_seed` by themselves if needed. 

Two other minor advantages of having a seed by default instead of generating at runtime:
1) It is better for provenance, `kwargs` saves exactly what the user inputed instead of a value generated on the fly.
2) Less code in our library and easier to document the behavior.

h-mayorquin:
I haved changed my view somehow and I think that most of the functions that interact with the user should actually have the behavior that @samuelgarcia implemented.

zm711:
@h-mayorquin is this good to close then?

h-mayorquin:
Yes. Thanks for checking it out.

PR #3131, opened 2024-07-03
Seeding sorters

florian6973:
Hi!

To better control reproducibility and study sorters' variability, I was envisioning to add some seeding parameters to the existing sorter configurations. 

I've done ms5 and ks4 which are quite popular, I do not know if you would like something more standardized before merging.

Thanks!

### DIFF
--- a/src/spikeinterface/sorters/external/kilosort4.py
+++ b/src/spikeinterface/sorters/external/kilosort4.py
@@ -57,6 +57,7 @@ class Kilosort4Sorter(BaseSorter):
         "skip_kilosort_preprocessing": False,
         "scaleproc": None,
         "torch_device": "auto",
+        "seed": 1
     }
 
     _params_description = {
@@ -99,6 +100,7 @@ class Kilosort4Sorter(BaseSorter):
         "skip_kilosort_preprocessing": "Can optionally skip the internal kilosort preprocessing",
         "scaleproc": "int16 scaling of whitened data, if None set to 200.",
         "torch_device": "Select the torch device auto/cuda/cpu",
+        "seed": "Kilosort random seed"
     }
 
     sorter_description = """Kilosort4 is a Python package for spike sorting on GPUs with template matching.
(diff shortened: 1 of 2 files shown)


h-mayorquin (commented):
Hi, thanks for your contribution.

This seems like a good idea but I don't know enough about how sorters work to see if this is done correctly or whether we can test it (probably not?).

Two comments though:
* You should not tamper with the global seed the way you are doing, see: https://blog.scientific-python.org/numpy/numpy-rng/
* We have a way of setting the seed when is not passed described here:

https://github.com/h-mayorquin/spikeinterface/blob/6645dbba97f638703b7d544154af7915af1651e3/src/spikeinterface/core/generate.py#L19-L26

And debated to some deegree here:
https://github.com/SpikeInterface/spikeinterface/issues/1963

florian6973:
Thanks for your feedback!

So I will change the default seed param to None and call ensure_seed first.

Regarding setting the global seed, I agree it is a bad behavior but otherwise we would need to go into the spike sorter code to change how numbers are generated. And for KS4, the spikeinterface code already uses this global seeding, so I was just replicating this technique for MS5. Do you feel it is worth to work on MS5 and KS4 to change their seed strategy please?

Regarding testing, we could do an empirical test to check the consistency, but this is a bit tricky because, for instance, MS5 can be deterministic depending on the settings (no spike sampling). Would you like me to add some basic tests like this?

h-mayorquin:
> Regarding setting the global seed, I agree it is a bad behavior but otherwise we would need to go into the spike sorter code to change how numbers are generated. And for KS4, the spikeinterface code already uses this global seeding, so I was just replicating this technique for MS5. Do you feel it is worth to work on MS5 and KS4 to change their seed strategy please?

I see. As I mentioned I am out of my depth here, I will let people who are more familiar with those sorters comment. Thanks for the clarification.

zm711:
Actually one tiny point about seeding KS4. The idea there was that the current wrapper is meant to mimic KS4 behavior (at the time of writing), which didn't allow seeding it just set its own global seed. @samuelgarcia thought this was a bad idea, but since it was mimicking KS4 we left it. So for KS4 I think it is actually important for the default to mimic the native KS4 behavior, but I think it is fine to give the user the option to change that seeding. But this is just my 2 cents. For MS5 I would suggest @magland weigh in how he thinks seeding should be handled.

alejoe91:
Hi @florian6973 

Sorry about the massive delay here!!!
Kilosort recently added its own seed: https://github.com/MouseLand/Kilosort/blob/main/kilosort/run_kilosort.py#L273-L275

For mountainsort, the best solution would be to have the MS package accept a seed and use it internally. The proposed change is not recommended by numpy, which suggests using a `RandomGenerator` object instead. We'll forward this to Jeremy, but let's close this for now.

PR #3132, opened 2024-07-03
Fix select peaks

alejoe91:
@samuelgarcia @yger

This is a fix for https://github.com/SpikeInterface/spikeinterface/pull/2980 :

The spikes need to be lexsorted based on segment and sample indices
sample indices start from 0 for each segment, so you don't need an offset!

### DIFF
--- a/src/spikeinterface/sortingcomponents/peak_selection.py
+++ b/src/spikeinterface/sortingcomponents/peak_selection.py
@@ -76,19 +76,18 @@ def select_peaks(
 
     selected_indices = select_peak_indices(peaks, method=method, seed=seed, **method_kwargs)
     selected_peaks = peaks[selected_indices]
+    num_segments = len(np.unique(selected_peaks["segment_index"]))
 
     if margin is not None:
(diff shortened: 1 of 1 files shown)


yger:
Ok cool. Yes, I tried to please @samuelgarcia but it was merged too quickly ! This seems good to me, I'll test that. Thanks a lot !!

src/spikeinterface/sortingcomponents/peak_selection.py
```diff
@@ -284,7 +283,9 @@ def select_peak_indices(peaks, method, seed, **method_kwargs):
         )
 
     selected_indices = np.concatenate(selected_indices)
-    selected_indices = selected_indices[np.argsort(peaks[selected_indices]["sample_index"])]
+    selected_indices = selected_indices[
+        np.lexsort((peaks[selected_indices]["sample_index"], peaks[selected_indices]["segment_index"]))
+    ]
```
samuelgarcia, line 288: the best would be stop sort selected_indices.
But this is ok.

samuelgarcia (approved):

PR #3138, opened 2024-07-04
Fix nn pca_metric computation and update tests

chrishalcrow:
Fixes a bug where `nn_isolation` and `nn_noise_overlap` were not being calculated.

In the `compute_pc_metrics` function, which calls the `pca_metrics_one_unit` function, the pca computations were surrounded by try/except statements. These meant that the computations were failing but the tests were passing.

The `nn_isolation` and `nn_noise_overlap` computations don’t fit into the quality metrics parallelisation easily, because they require waveforms. Hence they had to be moved so that they are calculated separately, and without parallelisation. It would be good idea to speed them up soon!

The tests involving the nn metrics were taking a long time (5mins+ on my machine) so I swapped in the `small_sorting_analyzer`, which sped things up. I also played a little with the `small_sorting_analyzer` to make sure all quality_metrics had non-NaN or non-trivial results for both the pca and non-pca quality metrics.


src/spikeinterface/qualitymetrics/pca_metrics.py
```diff
@@ -156,16 +160,16 @@ def compute_pc_metrics(
         )
         items.append(func_args)
 
-    if not run_in_parallel:
+    if not run_in_parallel and metric_names:
         units_loop = enumerate(unit_ids)
         if progress_bar:
-            units_loop = tqdm(units_loop, desc="calculate_pc_metrics", total=len(unit_ids))
+            units_loop = tqdm(units_loop, desc="calculate non nn pc_metrics", total=len(unit_ids))
```
alejoe91, line 166:
```suggestion
            units_loop = tqdm(units_loop, desc="calculate pc_metrics", total=len(unit_ids))
```
I would keep this simple ;)

src/spikeinterface/qualitymetrics/pca_metrics.py
```diff
@@ -176,6 +180,44 @@ def compute_pc_metrics(
...
+                total=len(unit_ids),
+            )
+
+        for unit_ind, unit_id in units_loop:
+            nn_noise_overlap = nearest_neighbors_noise_overlap(
+                sorting_analyzer,
+                unit_id,
+                n_spikes_all_units=n_spikes_all_units,
+                fr_all_units=fr_all_units,
+                seed=2205,
```
alejoe91, line 216:
```suggestion
                seed=seed,
```

src/spikeinterface/qualitymetrics/pca_metrics.py
```diff
@@ -176,6 +180,44 @@ def compute_pc_metrics(
...
+                **qm_params["nn_noise_overlap"],
+            )
+            pc_metrics["nn_noise_overlap"][unit_id] = nn_noise_overlap
```
alejoe91, line 219: if both metrics are in the list, we can do a single loop right?

src/spikeinterface/qualitymetrics/pca_metrics.py
```diff
@@ -955,11 +1005,9 @@ def pca_metrics_one_unit(args):
     pc_metrics = {}
     # metrics
     if "isolation_distance" in metric_names or "l_ratio" in metric_names:
-        try:
-            isolation_distance, l_ratio = mahalanobis_metrics(pcs_flat, labels, unit_id)
-        except:
-            isolation_distance = np.nan
-            l_ratio = np.nan
```
alejoe91, line 962: what if they fail? IMO, we should keep the try except and maybe add a warning.

If I'm computing 10 quality metrics for 500 units and one pc metric fail, this will stop the computation, or am I reading it wrong?

alejoe91: same for all other sklearn based metrics, sometimes the fitting fails and that's ok :)

chrishalcrow: Ok. So, keep these but make sure the tests, which we know should give non-Nan results, give non-Nan results? That sounds good!

src/spikeinterface/qualitymetrics/tests/test_pca_metrics.py
```diff
@@ -1,82 +1,24 @@
...
-
-
-def test_nearest_neighbors_noise_overlap(sorting_analyzer_simple):
-    sorting_analyzer = sorting_analyzer_simple
-    this_unit_id = sorting_analyzer.unit_ids[0]
-    nearest_neighbors_noise_overlap(sorting_analyzer, this_unit_id)
-
+    for metric_name in res1.columns:
+        if metric_name != "nn_unit_id":
+            assert not np.all(np.isnan(res1[metric_name].values))
```
chrishalcrow, line 21: I added a non-Nan test here

src/spikeinterface/qualitymetrics/tests/test_pca_metrics.py
```diff
@@ -1,82 +1,24 @@
...
     res1 = pd.DataFrame(res1)
 
-    res2 = compute_pc_metrics(sorting_analyzer, n_jobs=2, progress_bar=True)
+    res2 = compute_pc_metrics(sorting_analyzer, n_jobs=2, progress_bar=True, seed=1205)
     res2 = pd.DataFrame(res2)
 
-    for k in res1.columns:
-        mask = ~np.isnan(res1[k].values)
-        if np.any(mask):
-            assert np.array_equal(res1[k].values[mask], res2[k].values[mask])
```
chrishalcrow, line 63: Deleted these tests, as the test above does a more rigorous test of these functions, within `compute_pc_metrics`

src/spikeinterface/qualitymetrics/tests/conftest.py
```diff
@@ -0,0 +1,37 @@
+import pytest
```
chrishalcrow, line 1: Added this file, to share fixtures across test files. See: https://docs.pytest.org/en/6.2.x/fixture.html#conftest-py-sharing-fixtures-across-multiple-files

src/spikeinterface/qualitymetrics/pca_metrics.py
```diff
@@ -176,6 +179,32 @@ def compute_pc_metrics(
...
+        units_loop = enumerate(unit_ids)
+        if progress_bar:
+            units_loop = tqdm(units_loop, desc=f"calculate {metric_name} metric", total=len(unit_ids))
+
+        func = _nn_metric_name_to_func[metric_name]
+        metric_params = qm_params[metric_name] if metric_name in qm_params else {}
+
+        for unit_ind, unit_id in units_loop:
+
+            res = func(
```
alejoe91, line 192: I would still protect this with a try need try except clause

src/spikeinterface/qualitymetrics/pca_metrics.py
```diff
@@ -156,16 +159,16 @@ def compute_pc_metrics(
         )
         items.append(func_args)
 
-    if not run_in_parallel:
+    if not run_in_parallel and metric_names:
```
zm711, line 162: what does this do? I guess a user could supply metrics_names = [] as an empty list, but if it is None it is supplied. Just a little confused about this.

chrishalcrow: The nearest neighbour metrics are stripped out of metric_names earlier (line 114). [Ultimately this is because the non-nn metrics are parallelised using`pca_metrics_one_unit` while the nn ones must be calculated in a different way.]

So this is here to avoid running the code underneath in the case of a user just running nn metrics. Without it, if you just run nn metrics the user still sees a progress bar saying "calculating pca metrics...". 

Might be clearer to make a new list called `non_nn_metric_names`?

zm711: Yeah I think that would be clearer from a developer perspective. But I guess if it works it works.

zm711:
Also in general on Windows we are having super variable test times and quality metrics is one of the modules that can be super variable. In #3173, one of the actions for windows took 64min. I'm wondering if there is some sort of Windows specific thing in the tests that is causing this. Since you've been digging around in these metrics has anything jumped out as a slowdown for testing?

src/spikeinterface/qualitymetrics/pca_metrics.py
chrishalcrow: I think it is clearer: have done it

chrishalcrow:
> Also in general on Windows we are having super variable test times and quality metrics is one of the modules that can be super variable. In #3173, one of the actions for windows took 64min. I'm wondering if there is some sort of Windows specific thing in the tests that is causing this. Since you've been digging around in these metrics has anything jumped out as a slowdown for testing?

Goodness, those times shouldn't be that long. Especially because, before this gets merged, it isn't even calculating the slow metrics... Could it be something to do with `n_jobs`? Lots of the quality_metrics tests are done with `n_jobs=2`? Although the gh ci machines should have at least 2 cores? We should investigate....

zm711:
We changed the way n_jobs works to limit to the number of cores of the machine. I think some of the runners have 4 cores so we could think about bumping some up to 3. But yeah, I'm not sure. I need to dig into more. We are also seeing crazy slowdowns in sorting components sometimes, which is what you saw locally for sc2 right?

src/spikeinterface/qualitymetrics/pca_metrics.py
```diff
@@ -176,6 +177,37 @@ def compute_pc_metrics(
...
 
+    for metric_name in nn_metrics:
+        units_loop = enumerate(unit_ids)
+        if progress_bar:
+            units_loop = tqdm(units_loop, desc=f"calculate {metric_name} metric", total=len(unit_ids))
+
+        func = _nn_metric_name_to_func[metric_name]
+        metric_params = qm_params[metric_name] if metric_name in qm_params else {}
+
+        for unit_ind, unit_id in units_loop:
```
samuelgarcia, line 188: since we do not need unit_ind I would remove it because this is very error prone and ambiguous when unit_ids is given and is not teh full list.

chrishalcrow: done!

alejoe91 (approved):

PR #3139, opened 2024-07-04
Add option to use ref_channel_ids in global common reference

### LINKED ISSUE #2985: feature request: bad channel exclusion from common_reference function

alejoe91:
Fixes #2985

Add the option to select a subset of channels when reference is `global` via the `ref_channel_ids` argument.

### DIFF
--- a/src/spikeinterface/preprocessing/tests/test_common_reference.py
+++ b/src/spikeinterface/preprocessing/tests/test_common_reference.py
@@ -11,7 +11,7 @@ from tqdm import trange
 
 def _generate_test_recording():
     recording = generate_recording(durations=[1.0], num_channels=4)
-    recording = recording.channel_slice(recording.channel_ids, np.array(["a", "b", "c", "d"]))
+    recording = recording.rename_channels(np.array(["a", "b", "c", "d"]))
     return recording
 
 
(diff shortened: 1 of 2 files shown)


src/spikeinterface/preprocessing/tests/test_common_reference.py
```diff
@@ -23,12 +23,14 @@ def recording():
...
     # Test simple case
     rec_cmr = common_reference(recording, reference="global", operator="median")
+    rec_cmr_ref = common_reference(recording, reference="global", operator="median", ref_channel_ids=["a", "b", "c"])
     rec_car = common_reference(recording, reference="global", operator="average")
     rec_sin = common_reference(recording, reference="single", ref_channel_ids=["a"])
     rec_local_car = common_reference(recording, reference="local", local_radius=(20, 65), operator="median")
 
     traces = recording.get_traces()
     assert np.allclose(traces, rec_cmr.get_traces() + np.median(traces, axis=1, keepdims=True), atol=0.01)
+    assert np.allclose(traces, rec_cmr_ref.get_traces() + np.median(traces[:, :3], axis=1, keepdims=True), atol=0.01)
```
h-mayorquin, line 33: I had to run this to understand it but makes sense.

src/spikeinterface/preprocessing/tests/test_common_reference.py
```diff
@@ -11,7 +11,7 @@
 
 def _generate_test_recording():
     recording = generate_recording(durations=[1.0], num_channels=4)
-    recording = recording.channel_slice(recording.channel_ids, np.array(["a", "b", "c", "d"]))
+    recording = recording.rename_channels(np.array(["a", "b", "c", "d"]))
```
h-mayorquin, line 14: Yei!

h-mayorquin (approved):
OK, initially I felt that we were caving to adding yet another keyword for yet another specific case but after reading this I think you hit the jackpot.

This falls naturally in the semantics of  `ref_channel_ids`, it involves minimal code changes and within the complexities of the interaction between groups and ref_channel_ids it is relatively easy to document and the explanation does not feel convoluted.

As an aside you can support exactly the use case in #2985 as a side effect of this.

Great!

PR #3141, opened 2024-07-04
Add test_extractors to full test with codecov

alejoe91:


h-mayorquin (approved):
Right. 
At some point I will move the coverage to the other script but this makes sense!

PR #3142, opened 2024-07-04
Add a new widget to show the units topological distribution on the probe

FrancescoNegri:
This contribution aims at introducing a novel widget displaying the spatial distribution of sorted units with respect to the probe layout.
The supported visualizations are:
1.  Unidimensional histogram (along the depth axis)
2. Bidimensional histogram
3.  Bidimensional kernel density estimates (KDE)

image


src/spikeinterface/widgets/widget_list.py
```diff
@@ -10,10 +10,9 @@
 from .autocorrelograms import AutoCorrelogramsWidget
 from .crosscorrelograms import CrossCorrelogramsWidget
 from .isi_distribution import ISIDistributionWidget
-from .motion import DriftRasterMapWidget, MotionWidget, MotionInfoWidget
+from .motion import MotionWidget
```
alejoe91, line 13: I think you based your PR on a previous version of `main` where these widgets wee not implemented yet. Can you revert?

src/spikeinterface/widgets/widget_list.py
```diff
@@ -10,10 +10,9 @@
...
 from .multicomparison import MultiCompGraphWidget, MultiCompGlobalAgreementWidget, MultiCompAgreementBySorterWidget
 from .peak_activity import PeakActivityMapWidget
-from .peaks_on_probe import PeaksOnProbeWidget
```
alejoe91, line 16: Revert

src/spikeinterface/widgets/widget_list.py
```diff
@@ -44,15 +44,12 @@
     ConfusionMatrixWidget,
     ComparisonCollisionBySimilarityWidget,
     CrossCorrelogramsWidget,
-    DriftRasterMapWidget,
```
alejoe91, line 47: Same here

src/spikeinterface/widgets/widget_list.py
```diff
@@ -119,15 +117,12 @@
 plot_confusion_matrix = ConfusionMatrixWidget
 plot_comparison_collision_by_similarity = ComparisonCollisionBySimilarityWidget
 plot_crosscorrelograms = CrossCorrelogramsWidget
-plot_drift_raster_map = DriftRasterMapWidget
```
alejoe91, line 122: Same here

alejoe91 (changes requested):
Thanks @FrancescoNegri 

See my comments. Can you also enable maintainers to push to your PR so it gets automatically formatted?

yger:
Cool, looks very nice? Is there a particular reason why the histogram would be restricted to 1D only? Could we have it also 2D, for planar MEA?

src/spikeinterface/widgets/widget_list.py
FrancescoNegri: Sure, I will fix those lines!

FrancescoNegri:
> Thanks @FrancescoNegri
> 
> See my comments. Can you also enable maintainers to push to your PR so it gets automatically formatted?

Yeah, I can see the comments and I will fix the imports.
I am trying to enable you to push to this PR, but I cannot find the check box 😅

FrancescoNegri:
> Cool, looks very nice! Is there a particular reason why the histogram would be restricted to 1D only? Could we have it also 2D, for planar MEA?
> 
> @edit: it seems that this is already handling 2D, I only judged based on the plots, sorry for that

I considered the `UnitDepthsWidget` class as reference, thus with the idea of a `depth_axis`. However, we can make that parameter optional and if not specified the function can assume planar 2D MEAs.

alejoe91:
Can you then run `pre-commit` on your end?

alejoe91:
Thanks @FrancescoNegri, can you add a simple test in the `tests/test_widgets.py`? Even just running the function with a few params changes is fine

src/spikeinterface/widgets/unit_spatial.py
```diff
@@ -0,0 +1,152 @@
+from __future__ import annotations
+
+import numpy as np
+from probeinterface import Probe
+from probeinterface.plotting import get_auto_lims
+from seaborn import color_palette
```
alejoe91, line 6: you should move these to the `plot_matplotlib` function. Also `seaborn` is not installed by default. Could you use a matplotlib palette?

FrancescoNegri: I can definitely use a `matplotlib` palette, but I was thinking to add an optional `seaborn` import if available to enable also `seaborn` palettes. I used `seaborn` also in other parts of the function, I will try to fix it, though I am not sure I will be able to easily plot kernel desnity estimates with `matplotlib` only.

Is there a specific reason to move probeinterface imports to the `plot_matplotlib` function? I use its `get_auto_lims` function to compute `xrange` and `yrange`, that are independent on the visualization backend.

alejoe91: yes that souns good! The reason are the failing tests ;)

Core tests only install minimal dependencies. Upon collecting tests across modules, if something is not installed it'll throw an error.

FrancescoNegri: I checked the core tests. It seems that `probeinterface` is installed, but `matplotlib` is not, thus `probeinterface.plotting` cannot be imported. Is that right?

FrancescoNegri:
I spent some time trying to replace all `seaborn` dependencies with `matplotlib` ones. The issue is that it is not as much straightforward as I thought and some functionalities of the `UnitSpatialDistributionsWidget` class will be probably dropped.

I have seen that there are other classes relying on `seaborn` such as the `StudyPerformances` widget. Wouldn't be worth it to add `seaborn` to the dependencies of `SpikeInterface`? Since `matplotlib` is already there, it would come with minimal overhead.

alejoe91:
Thanks for trying! I'm ok in adding it to the "widgets" installation :)

FrancescoNegri:
Just a quick check. I was going through the `probeinterface` code. The `get_auto_lims` function defined in the `plotting` namespace does not use `matplotlib` and it could be useful in other contexts where `matplotlib` might not be available (for instance when estimating the probe limits for different plotting backends).
Wouldn't be better to move it to the `utils` namespace then? The `get_auto_lims` function relies exclusively on vanilla Python and `numpy`.
In case you agree, I can open a pull request with those changes in the probeinterface repo :)

alejoe91:
That sounds good! 😊

FrancescoNegri:
I opened [this](https://github.com/SpikeInterface/probeinterface/pull/281) PR in the `probeinterface` repo!

alejoe91:
@FrancescoNegri is this ready for review?

FrancescoNegri:
@alejoe91 not really, I've been off the whole month of August. I will be back to work in a week. Hopefully, I should push a ready-for-review version by mid September.

alejoe91:
> @alejoe91 not really, I've been off the whole month of August. I will be back to work in a week. Hopefully, I should push a ready-for-review version by mid September.

No rush! just checking in!

FrancescoNegri:
I updated all of the code and is now ready to be pushed, though I am having an issue and I would like to discuss it with you.
The widget is configured such that it can show:

1.  Only the _"joint"_ plot, thus the histogram/kde on the probe shape.
2. The _"joint"_ plot and 1 _"marginal"_ plot, If the probe has a preferential direction (`depth_axis` parameter) I am likely only interested in a 1D marginal histogram along that direction.
3. The _"joint"_ plot and 2 _"marginal"_ plots. This is useful when we deal with 2D MEAs and we might want to visualize the marginal histograms on both x and y axes.

Given this starting point, I am trying to automatically determine the proportions of the _"joint"_ and _"marginal"_ plot(s) given the default or user-specified `figsize` and adding the _"marginal"_ axes to the _"joint"_ one by using an `AxisDivider`. However, given the huge diversity in probe shapes, I am not finding an optimal solution.

Do you by chance have suggestions, ideas or comments about this? @alejoe91 @yger 
Thank you in advance! :)

alejoe91:
@FrancescoNegri sorry for the long delay on this from our side.

I was testing the PR and it still looks unfinished, right? 
- if `bins` are not provided externally, it fails.
- `depth_axis` is not used
- can you add docstrings for all arguments

Next time it woudl be better to make the PR from you personal GH account, so we can contribute :)

FrancescoNegri:
@alejoe91 yes, this was paused, but I can try to wrap it up by the end of summer. Next time I will surely create the PR from my personal account! :)

PR #3150, opened 2024-07-04
Improved better conditions for closing nwb when using hdf5 backend

h-mayorquin:
So we have been battling this for a while. Our access pattern of nwb when using the hdf5 backend directly (instead of the pynwb API) leaves files references dangling. 

When looking a this, I see that we are keeping references to some groups and datasets even after closing the main reference. I am not sure if it is the weak reference or the way we access some properties were we don't copy at read.

At some point I have to look deeply into this but meanwhile this should alleviate some of the problems. This uses the low level API of h5py to access all the references to  a particular id and tries to close them when calling del. It did solve my issue that I was facing but I am too skeptical to think this is the final nail on the coffin.

### DIFF
--- a/src/spikeinterface/extractors/nwbextractors.py
+++ b/src/spikeinterface/extractors/nwbextractors.py
@@ -626,13 +626,29 @@ class NwbRecordingExtractor(BaseRecording):
             "file": file,
         }
 
+    def _close_hdf5_file(self):
+        has_hdf5_backend = hasattr(self, "_file")
+        if has_hdf5_backend:
+            import h5py
+
+            main_file_id = self._file.id
+            open_object_ids_main = h5py.h5f.get_obj_ids(main_file_id, types=h5py.h5f.OBJ_ALL)
+            for object_id in open_object_ids_main:
+                object_name = h5py.h5i.get_name(object_id).decode("utf-8")
+                try:
+                    object_id.close()
+                except:
+                    import warnings
+
+                    warnings.warn(f"Error closing object {object_name}")
+
     def __del__(self):
         # backend mode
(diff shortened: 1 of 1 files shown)


src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -626,13 +626,29 @@ def __init__(
...
+            main_file_id = self._file.id
+            open_object_ids_main = h5py.h5f.get_obj_ids(main_file_id, types=h5py.h5f.OBJ_ALL)
+            for object_id in open_object_ids_main:
+                object_name = h5py.h5i.get_name(object_id).decode("utf-8")
+                try:
+                    object_id.close()
+                except:
+                    import warnings
+
+                    warnings.warn(f"Error closing object {object_name}")
```
alejoe91, line 643: since this is the same function as the `Sorting` object, why not making a separate unique function?
```
def _close_hdf5_file(extraxctor):
    has_hdf5_backend = hasattr(extraxctor, "_file")
    ....
```

then you would just call it in the `__del__` with `_close_hdf5_file(self)`

alejoe91 (changes requested):
Thanks Ramon, see comment

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: You are right that this repetition is unecessary but I wanted to use self, check out the latest commit. I implemented a mixing class that should hold the methods that need the state of the nwb extractor and are common to both recording and sorting.

alejoe91 (approved):

PR #3153, opened 2024-07-05
Remove unused legacy class name/mode from extractors

alejoe91:
Needs to be merged to unlock #2784 and following up on #2874

All extractors had a legacy "name" and "mode" class attributes that we never use. This PR removes them and with it also removes the unused: `recording_extractor_full_dict` / `sorting_extractor_full_dict` / `get_recording_extractor_from_name` / `get_sorting_extractor_from_name`


h-mayorquin (approved):
Thanks for doing this and also moving and fixing some imports

samuelgarcia:
I am a bit sad with the remove of mode="file"/"folder" because I really wanted to make a GUI that would have take this attr.
Lets be realistic I will never have time to do it.

And putting this back to classe would cost nothing :)

This is OK for me

samuelgarcia (approved):

PR #3160, opened 2024-07-07
Create SECURITY.md

Arilsrinivas:


zm711:
@Arilsrinivas,

hey I'm not quite sure how this fits into the project. We use `.rst` files for most things not `.md` and although I think security is important this is a scientific analysis library not a web interface. Also your PR has nothing to do with the current state of the library. (ie we are version 0.100.8-> 0.101.0, not 4 and 5 like you have in this PR).

My advice for the future would be open an issue and see if your PR is something that fits with the library first. This gives you a chance to ask questions like rst vs md. I would also provide a description of what your PR hopes to do and why. Right now this is just empty with no description.

I don't think we need this but I guess @alejoe91 and @samuelgarcia can decide.

zm711:
I'm just going to close this one at this point, but @Arilsrinivas feel free to open an issue where moving forward we could discuss how to fit something like this into the project. I think @JoeZiminski has a great idea that you could instead make a PR explaining safe download practices within the developer documentation.

PR #3169, opened 2024-07-09
Build extractor dicts automatically

### LINKED ISSUE #3168: `get_neo_stream` broken after removing extractor class `name` attributes

alejoe91:
Fixes #3168

### DIFF
--- a/src/spikeinterface/extractors/extractorlist.py
+++ b/src/spikeinterface/extractors/extractorlist.py
@@ -116,3 +116,21 @@ event_extractor_full_list = [MaxwellEventExtractor]
 event_extractor_full_list += neo_event_extractors_list
 
 snippets_extractor_full_list = [NpySnippetsExtractor, WaveClusSnippetsExtractor]
+
+recording_extractor_full_dict = {}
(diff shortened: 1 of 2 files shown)


zm711 (commented):
Should there be a test for this then just to make sure the machinery of `get_neo_streams` doesn't break again?

h-mayorquin (approved):
LGTM

What do you think of deprecating this function in another PR?

The same behavior can be achieved by using the class method `get_streams` and that seems better because the user does not need to know in advance the string that corresponds to the extractor (e.g. does not need to know to know that `spikeglx` corresponds to `read_spikeglx`). Plus, because we have the class we can have better error messages and import the io reader directly without having to build an ad-hoc dictionary as we do here.

h-mayorquin:
We would just need to change the docs and throw a deprecation here.

alejoe91:
I think that the `get_neo_streams` is super useful and I use it all the time :)

alejoe91:
> Should there be a test for this then just to make sure the machinery of `get_neo_streams` doesn't break again?

Let's merge for now and open an issue about adding this test

Issue #3148, opened 2024-07-04
relative paths


rat-h:
We currently use spikeinterface in a pipe, when for each new recording the script
- creates a new directory,
- does preprocessing saved into `(recording directory)/preproc`, 
- runs spikesorter of choice, 
- saves sorting into directory  `(recording directory)/sorting-saved` and 
- runs wave extractor(s) which saves results into `(recording directory)/waves`.
I can then use the `waves` folder to see the results.

However, we save all data and processing in a central storage. After copying the results there and downloading them onto another computer, nothing works. The culprits are absolute paths in JSON files. 

Would it be possible to use relative paths that will be agnostic of the absolute location of the result folders?

h-mayorquin:
Yes, we have that functionality. But, how are you saving your data? `recording.save()`?

zm711:
I think the waveform extractor has a parameter that must be explicit set see [here](https://github.com/SpikeInterface/spikeinterface/blob/1197aad12b7170432eec8b887e24b9bc29921eab/src/spikeinterface/core/waveform_extractor.py#L264). If you haven't been setting this to true (it's default is false) then it has been using absolute paths.

rat-h:
> Yes, we have that functionality. But, how are you saving your data? `recording.save()`?

Yes, but only preprocessed data - after filtering and referencing

 

> I think the waveform extractor has a parameter that must be explicit set see [here](https://github.com/SpikeInterface/spikeinterface/blob/1197aad12b7170432eec8b887e24b9bc29921eab/src/spikeinterface/core/waveform_extractor.py#L264).

That is awesome! Do we have something like that for `sorting.save()` ? It seems there are a few files which have absolute paths.
```bash
$ find sorting-saved  -name '*.json' -exec grep $HOME {} + 
sorting-saved/provenance.json:        "file_path": "/home/spikeinterface/20230508T195033Z-continuous-hdsort-local-T0.45-20240602-155052-20230508T195033Z-HDsortLx36x21x4-3031626-final-final-84/sorting-workingdir/sorter_output/hdsort_output/hdsort_output_results.mat",
sorting-saved/provenance.json:                    "folder_path": "/home/spikeinterface/20230508T195033Z-continuous-hdsort-local-T0.45-20240602-155052-20230508T195033Z-HDsortLx36x21x4-3031626-final-final-84/preprocessed"
[...]

PR #3170, opened 2024-07-09
Better error message for `BaseExtractor.load()` 

h-mayorquin:
As in the title, related to issue https://github.com/SpikeInterface/spikeinterface/issues/3148

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -766,6 +766,7 @@ class BaseExtractor:
             if "warning" in d:
                 print("The extractor was not serializable to file")
                 return None
+
             extractor = BaseExtractor.from_dict(d, base_folder=base_folder)
             return extractor
 
@@ -798,7 +799,10 @@ class BaseExtractor:
             return extractor
 
         else:
-            raise ValueError("spikeinterface.Base.load() file_path must be an existing folder or file")
+            error_msg = f"Failed to load the file_path {file_path} must be an existing folder or file"
+            if base_folder:
+                error_msg = f"\n base folder provided is {base_folder}"
+            raise ValueError(error_msg)
 
     def __reduce__(self):
         """
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -799,7 +799,7 @@ class BaseExtractor:
             return extractor
 
         else:
-            error_msg = f"Failed to load the file_path {file_path} must be an existing folder or file"
+            error_msg = f"{file_path} is not a file or a folder"
             if base_folder:
                 error_msg = f"\n The provided base_folder is {base_folder}"
             raise ValueError(error_msg)
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -799,9 +799,8 @@ class BaseExtractor:
             return extractor
 
         else:
-            error_msg = f"{file_path} is not a file or a folder"
-            if base_folder:
-                error_msg = f"\n The provided base_folder is {base_folder}"
+            error_msg = f"{file_path} is not a file or a folder. It should point out to either a json, pickle file or a
+            folder that is the result of extractor.save(...)"
             raise ValueError(error_msg)
 
     def __reduce__(self):
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -801,7 +801,7 @@ class BaseExtractor:
         else:
             error_msg = f"Failed to load the file_path {file_path} must be an existing folder or file"
             if base_folder:
-                error_msg = f"\n base folder provided is {base_folder}"
+                error_msg = f"\n The provided base_folder is {base_folder}"
             raise ValueError(error_msg)
 
     def __reduce__(self):
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -799,8 +799,8 @@ class BaseExtractor:
             return extractor
 
         else:
-            error_msg = f"{file_path} is not a file or a folder. It should point out to either a json, pickle file or a
-            folder that is the result of extractor.save(...)"
+            error_msg = f"{file_path} is not a file or a folder. It should point out to either a json, pickle file or a "
+            "folder that is the result of extractor.save(...)"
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/base.py
```diff
@@ -798,7 +799,10 @@ def load(file_path: Union[str, Path], base_folder: Optional[Union[Path, str, boo
             return extractor
 
         else:
-            raise ValueError("spikeinterface.Base.load() file_path must be an existing folder or file")
+            error_msg = f"Failed to load the file_path {file_path} must be an existing folder or file"
+            if base_folder:
+                error_msg = f"\n base folder provided is {base_folder}"
```
alejoe91, line 804:
```suggestion
            error_msg = f"Failed to load the file_path {file_path}: it must be an existing folder or file"
            if base_folder:
                error_msg = f"\nThe provided base_folder provided is: {base_folder}"
```

h-mayorquin: Your suggestion repeats provided but yes, my grammar is weird on that one. Let me fix it and push.

src/spikeinterface/core/base.py
```diff
@@ -798,7 +799,10 @@ def load(file_path: Union[str, Path], base_folder: Optional[Union[Path, str, boo
             return extractor
 
         else:
-            raise ValueError("spikeinterface.Base.load() file_path must be an existing folder or file")
+            error_msg = f"Failed to load the file_path {file_path} must be an existing folder or file"
```
zm711, line 802:
```suggestion
            error_msg = f"Failed to load the file_path {file_path}. The file_path must be an existing folder or file"
```

Or something like this. You failed to load the file_path `.` (why?) because it must be an existing folder or file.

zm711 (commented):
Even with the grammar improvement I have no clue what the problem with the `base_folder` is. At least for the `file_path` you're saying that based on your tests it doesn't seem to exist. But just telling me that I gave a base_folder doesn't tell me how to fix my problem. 

And the error rule for Heberto is it must be actionable.

src/spikeinterface/core/base.py
h-mayorquin: Yes, this should be corrected.

Concerning of how to make the information of `base_folder` better I agree and thought about it before. But I honestly don't know how to improve this. The specification of how to use `base_folder` is sort of implicit and left to the user. People used `relative_to` to save with respect to this folder and then need to correct their paths with respect to this and pass it like this.

Any ideas on how to make this better are welcomed but I think perfect should not be the enemy of good here.

Btw, I believe that warnings should be actionable. I am not sure I believe that about errors but it makes sense as well as an ideal. I will think about it.

zm711: In that case it seems to me like the message should be something like

```
The provided base_folder is {base_folder}. Please ensure this is the correct root for the relative paths  to your extractor
```

or something along that vein. I don't know this part of the code well enough to know for sure this is the correct statement. But this reminds the user that the base_folder is given for the relative_to machinery. 

> Btw, I believe that warnings should be actionable. I am not sure I believe that about errors but it makes sense as well as an ideal. I will think about it.

I stand corrected. But in my schema a ValueError, TypeError, FileNotFoundError are things that ought to be fixable right. So if we can nudge the person to fix it that should be the goal, I think. So warning + those would be my metric. Something like a RunTimeError for me says nope this is fundamentally wrong.

h-mayorquin:
Github seems to have some connection errors today.

src/spikeinterface/core/base.py
h-mayorquin: > or something along that vein. I don't know this part of the code well enough to know for sure this is the correct statement. 

We have the same problem.

Ok. I think I am removing this part. Maybe that error should be deeper and I need to think better about relative paths. About every 3 months I try to think on the relative_path machinery and I never feel that I completly grok it.

Your heuristics about those type of errors makes sense. Maybe less clear about file not found. You can say what is the problem but you don't know if the file does no exists, is in the wrong place, is corrupted, wrong extension, wrong mount, etc.

zm711: Yeah, then I would remove it. Because technically the user can just press the up-arrow and see what they put in the `base_folder` argument.

For me filenotfound implies all those things and the user needs to figure out which, but maybe that is expecting too much.

But yeah one of these days you and I can have a zoom and just stare at the code together until we both understand it :)

h-mayorquin: Ok, removed this.

src/spikeinterface/core/base.py
```diff
@@ -798,7 +799,9 @@ def load(file_path: Union[str, Path], base_folder: Optional[Union[Path, str, boo
             return extractor
 
         else:
-            raise ValueError("spikeinterface.Base.load() file_path must be an existing folder or file")
+            error_msg = f"{file_path} is not a file or a folder. It should point out to either a json, pickle file or a
+            folder that is the result of extractor.save(...)"
```
zm711, line 803:
```suggestion
            error_msg = f"{file_path} is not a file or a folder. It should point out to either a json, pickle file or a "
            "folder that is the result of extractor.save(...)"
```
the lines were missing quotes :)

zm711:
I think pre-commit broke this :/

h-mayorquin:
Should be solved int the last commit.

src/spikeinterface/core/base.py
```diff
@@ -797,7 +798,11 @@ def load(file_path: Union[str, Path], base_folder: Optional[Union[Path, str, boo
             return extractor
 
         else:
-            raise ValueError("spikeinterface.Base.load() file_path must be an existing folder or file")
+            error_msg = (
+                f"{file_path} is not a file or a folder. It should point out to either a json, pickle file or a "
```
chrishalcrow, line 802:
```suggestion
                f"{file_path} is not a file or a folder. It should point to either a json, pickle file or a "
```

Tiny grammar improvement, then all good

h-mayorquin: Thanks, phrasal verbs are difficult for me in English.

chrishalcrow (approved):

alejoe91 (approved):

PR #3173, opened 2024-07-09
Add explainer of compute always computing in the analyzer (vs WaveformExtractor behavior)

### LINKED ISSUE #3031: `SortingAnalyzer` issues

zm711:
Fixes #3031.

I would say that we've addressed all the issues in #3031 except making it clear in the docs that compute will compute no matter what. So I've added two lines:

One in the wvf_ext -> sorting_analyer docs explain load_if_exists no longer exists in the sorting analyzer (for people use to the old way)
One in the core docs just emphasizing that compute will recompute any extension it is called on.

EDIT: 3 addition in the analyzer core tutorial @chrishalcrow .


chrishalcrow:
Hello, looks good. We could also add to the [plot_4_sorting_analyzer.py](https://github.com/SpikeInterface/spikeinterface/blob/main/examples/tutorials/core/plot_4_sorting_analyzer.py).

Maybe we could change:
`Note that there are a few different ways to calculate extensions. You might find it easier to keep the calculation parameters in a dictionary. Let’s recalculate everything using this notation:`

to 

`Note that there are a few different ways to calculate extensions. You might find it easier to keep the calculation parameters in a dictionary. Let’s recalculate everything using this notation (note that extensions will be _recomputed_ when we call compute, even if the parameters are the same):`

just to point it out.

PS: during reviews can you add suggestions to files that the author hasn't touched. ie in this situation, can I add a review suggestion to `plot_4_sorting_analyzer.py`?

zm711:
> PS: during reviews can you add suggestions to files that the author hasn't touched. ie in this situation, can I add a review suggestion to plot_4_sorting_analyzer.py?

Not that I know of, but I will add your change!

chrishalcrow (approved):
Looks good!

doc/modules/core.rst
```diff
@@ -360,6 +360,9 @@ examples is seen below:
         ["random_spikes", "waveforms", "templates", "noise_levels"]
     )
 
+Note that any extension entered into the :code:`compute` function will be computed. Thus, even if an extension has already been computed
+it will be recomputed when :code:`compute` is called even if the parameters are the same.
```
alejoe91, line 364: @zm711 I think it would be good here to mention how to check if an extension is computed and how to load it:

```suggestion
it will be recomputed when :code:`compute` is called even if the parameters are the same. 
To check if an extansion has already been computed and reload it, you can simply do:

.. code-block:: python

    if sorting_analyzer.has_extension("templates"):
        templates_extension = sorting_analyzer.get_extension("templates")
```

What do you think?

zm711: Done :)

zm711: I thought about that and then promptly forgot to mention it.

alejoe91 (approved):

doc/modules/core.rst
```diff
@@ -360,6 +360,15 @@ examples is seen below:
...
+To check if an extansion has already been computed and reload it, you can simply do:
```
zm711, line 365:
```suggestion
To check if an extension has already been computed and reload it, you can simply do:
```

h-mayorquin (approved):
Makes sense to me.

h-mayorquin:
> PS: during reviews can you add suggestions to files that the author hasn't touched. ie in this situation, can I add a review suggestion to `plot_4_sorting_analyzer.py`?

@chrishalcrow you mean within github web interface? not that I am aware

alejoe91 (approved):

PR #3174, opened 2024-07-10
Parallel computation for similarity matrices

yger:
I've exactly copied the parallelism used in sortingcomponents.clustering.split in order to allow several cores to be used while computing the similartiy matrix. Otherwise, for large number of units and non zero lags, the computation is really slow...


yger:
I was curious to give it a go @samuelgarcia , no hurry to review it, but I've tried my best to do exactly as you did in split(). This would tremendously boost the computation of similarity matrices for large number of units and/or large lags

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -5,6 +5,11 @@
 
 from spikeinterface.core.sortinganalyzer import register_result_extension, AnalyzerExtension
 from ..core.template_tools import get_dense_templates_array
+from spikeinterface.core.job_tools import fix_job_kwargs
```
alejoe91, line 8:
```suggestion
```

PR #3178, opened 2024-07-10
Fix pca transform error

### LINKED ISSUE #3166: Bug in `spykingcircus2` sorting processing
Sorry for the mistake. I'll update the question here. 
When I run sorter "spykingcircus2" with the method "wobble", there is an error.
image

The data is my data. There are 2 channels for sorting. Should I do something else to make it sure?

alejoe91:
I have actually encountered #3166 a few times, and it should be fixed with this simple solution. We were a bit too strict with the exception type.

Fixes #3166


h-mayorquin:
Dude, how did you realize from the traces in https://github.com/SpikeInterface/spikeinterface/issues/3166 that this is the solution? : O

I envy your knowledge!

samuelgarcia (approved):

PR #3179, opened 2024-07-10
Update nwbextractors.py

bendichter:
use latest tooling for streaming data

### DIFF
--- a/src/spikeinterface/extractors/nwbextractors.py
+++ b/src/spikeinterface/extractors/nwbextractors.py
@@ -497,9 +497,10 @@ class NwbRecordingExtractor(BaseRecording, _BaseNWBExtractor):
     >>> from dandi.dandiapi import DandiAPIClient
     >>>
     >>> # get s3 path
-    >>> dandiset_id, filepath = "101116", "sub-001/sub-001_ecephys.nwb"
+    >>> dandiset_id = "001054"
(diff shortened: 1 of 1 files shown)


h-mayorquin (approved):
Sounds good for me.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -498,11 +498,11 @@ class NwbRecordingExtractor(BaseRecording, _BaseNWBExtractor):
     >>>
     >>> # get s3 path
     >>> dandiset_id, filepath = "101116", "sub-001/sub-001_ecephys.nwb"
-    >>> with DandiAPIClient("https://api-staging.dandiarchive.org/api") as client:
```
bendichter, line 501: oh, actually we may need to keep this, since I guess the sample data is in staging

h-mayorquin: I guess we coudl jsut select other dataset that is not in staging area which I think would be better. What do you think?

alejoe91: @h-mayorquin so should we pick another dataset?

h-mayorquin: Yes.I think it will be better if the docstring does not have that complication.

alejoe91 (approved):

PR #3180, opened 2024-07-10
Units aggregation preserve unit ids of aggregated sorters

### LINKED ISSUE #3152: Unit aggregation should preserve unit ids.
Like https://github.com/SpikeInterface/spikeinterface/issues/2789 but for UnitAggregation. Currently it does not:

```python
from spikeinterface.core import generate_sorting


sorting1 = generate_sorting(num_units=3)
sorting1 = sorting1.rename_units(new_unit_ids=["unit1", "unit2", "unit3"])

sorting2 = generate_sorting(num_units=3)
sorting2 = sorting2.rename_units(new_unit_ids=["unit4", "unit5", "unit6"])

from spikeinterface.core import aggregate_units

sorting = aggregate_units([sorting1, sorting2])
sorting.get_unit_ids()
array([0, 1, 2, 3, 4, 5])
```

h-mayorquin:
This should close #3152

### DIFF
--- a/src/spikeinterface/core/unitsaggregationsorting.py
+++ b/src/spikeinterface/core/unitsaggregationsorting.py
@@ -34,7 +34,8 @@ class UnitsAggregationSorting(BaseSorting):
             )
             unit_ids = list(renamed_unit_ids)
         else:
-            all_ids_are_same_type = np.unique([sort.get_unit_ids().dtype for sort in sorting_list]).size == 1
+            unit_ids_dtypes = [sort.get_unit_ids().dtype for sort in sorting_list]
+            all_ids_are_same_type = np.unique(unit_ids_dtypes).size == 1
             all_units_ids_are_unique = False
             if all_ids_are_same_type:
                 combined_ids = np.concatenate([sort.get_unit_ids() for sort in sorting_list])
@@ -44,7 +45,10 @@ class UnitsAggregationSorting(BaseSorting):
                 unit_ids = combined_ids
             else:
                 default_unit_ids = [str(i) for i in range(num_all_units)]
-                unit_ids = default_unit_ids
+                if all_ids_are_same_type and np.issubdtype(unit_ids_dtypes[0], np.integer):
+                    unit_ids = np.arange(num_all_units, dtype=dtype)
+                else:
+                    unit_ids = default_unit_ids
 
         # unit map maps unit ids that are used to get spike trains
         u_id = 0
--- a/src/spikeinterface/core/unitsaggregationsorting.py
+++ b/src/spikeinterface/core/unitsaggregationsorting.py
@@ -34,7 +34,17 @@ class UnitsAggregationSorting(BaseSorting):
             )
             unit_ids = list(renamed_unit_ids)
         else:
-            unit_ids = list(np.arange(num_all_units))
+            all_ids_are_same_type = np.unique([sort.get_unit_ids().dtype for sort in sorting_list]).size == 1
+            all_units_ids_are_unique = False
+            if all_ids_are_same_type:
+                combined_ids = np.concatenate([sort.get_unit_ids() for sort in sorting_list])
+                all_units_ids_are_unique = np.unique(combined_ids).size == num_all_units
+
+            if all_ids_are_same_type and all_units_ids_are_unique:
+                unit_ids = combined_ids
+            else:
+                default_unit_ids = [str(i) for i in range(num_all_units)]
+                unit_ids = default_unit_ids
 
         # unit map maps unit ids that are used to get spike trains
         u_id = 0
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/unitsaggregationsorting.py
```diff
@@ -34,7 +34,17 @@ def __init__(self, sorting_list, renamed_unit_ids=None):
...
+            all_ids_are_same_type = np.unique([sort.get_unit_ids().dtype for sort in sorting_list]).size == 1
+            all_units_ids_are_unique = False
+            if all_ids_are_same_type:
+                combined_ids = np.concatenate([sort.get_unit_ids() for sort in sorting_list])
+                all_units_ids_are_unique = np.unique(combined_ids).size == num_all_units
+
+            if all_ids_are_same_type and all_units_ids_are_unique:
+                unit_ids = combined_ids
+            else:
+                default_unit_ids = [str(i) for i in range(num_all_units)]
```
zm711, line 46: Only thing here is you're changing the default behavior from int to str. I guess strings are better, but this is at the public level so people may have relied on the type right? I guess I'm not strongly against the typing as long as the error that occurs (which you put into base right?) will make it clear to the end user that the type has switched to strings.

h-mayorquin: Reading this again I am not sure what do you mean by your last sentence about the error. You mean the other PR about improving the checking?

Sadly, actually there is no such check for units ids. I think it should be though : )

zm711: I honestly forgot if the new error message was super base or just for recordings. But yeah I think we need some sort of error that let's people know the type of ids (or just lists the possibilities). Especially if we are changing the default type from what the person is putting in to what they are getting out.

h-mayorquin: I agree that we should have a check for this. Will do in another PR. But just to clarify, we are not changing the type of the ids that put people put in, in fact, we are preserving it.

Previously, unit aggregation would by default just build new ids of type int. The new behavior is:
1) If you ids are compatible (same dtype and not repeated) then the resulting ids are the concatenate ids of the sorting components. So, same type in same type out.
2) If you ids are not compatible we build new ids of type string (previously they were of type int).

zm711: Sorry I wasn't being clear. I agree with point 1. That is the nice new feature. 

My point was exclusively about point 2. For a real world use case if I sort a probegroup by probe with MS5 it will return two sortings with non-compatible IDS (because they both start at 0) but of dtype int. Then when I aggregate them I will now have new ids with strings instead of the ints I was expecting because I put ints into. As long as the error message is good I think people will understand that their ids were switched to strings, but I personally wouldn't expect this if I put two sortings with a dtype of int for their ids to have a new sorting with a dtype of string.

h-mayorquin: Yeah, but that's your current situation if you have two sortings of type string that are incompatible. You get a new sorting with type int in the current version.

zm711: I guess my question is what is the dtype for a sorting unit ids in general. Using KS2, 2.5, 3, and MS5 all return sortings with int ids. So I'm thinking from the perspective that most people likely don't rename their ids and so most people will have int ids. But maybe I'm wrong. so int -> str is more disruptive than str -> int in the wild. But I could be wrong.

h-mayorquin: Yeah, as mentioned in the linked thread, it is a small movement in the direction of:
https://github.com/SpikeInterface/spikeinterface/issues/1432

WHen you say that the sorters return ids as strings, isnt' that a choice that we made over there? Or do you mean in their native APIs they return spikeinterface sorting objects and they have ids that are of type int?

zm711: I'm saying that if I do

```python
ms_sorting = ss.run_sorter('mountainsort5', rec)
```

The ids will be ints in the `ms_sorting`. And I know from past experience that KS returns its values and saves them in an npy file with a dtype of int. So if I do

```python

ks_sorting = se.read_kilosort(folder_path)
```

`ks_sorting` will have int ids as well. Because the previous discussion said that no magical casting under the hood yet.

h-mayorquin: Thanks. Yes,  I think the first is our choice the second I am less sure.

zm711: I think this issue would all go away if we just made a choice to represent everything as strings  then people would know that even in a format like KS (which stores their ids an ints) that they just use the stringified ints. 

But my (and again I stress this is just a personal preference) preference would be to keep the dtype of the inputs. So if both or three recordings to be joined are ints then keep ints and if the dtypes are not the same then maybe default to strings? That way if someone puts all strings in they get strings. If they put all ints in they get ints and for mix it's safer to just default to strings. Unless we agree to make everything strings.

h-mayorquin: OK, fair enough, You have convinced me that that's the most user friendly way to move this forward.

h-mayorquin: Ok @zm711, check the last commit.

zm711 (commented):
This looks great to me now!

alejoe91 (approved):

PR #3184, opened 2024-07-11
Change installation instructions in readme to be compatible with all OS

h-mayorquin:
So double quotes works on all operating systems, see:

https://github.com/catalystneuro/neuroconv/issues/947

https://discuss.python.org/t/pep-621-how-to-specify-dependencies/4599/61

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ You can also have a look at the [spikeinterface-gui](https://github.com/SpikeInt
 
 ## How to install spikeinterface
 
-You can install the latest version of `spikeinterface` version with pip (using quotes ensures `pip install` works in all terminals):
+You can install the latest version of `spikeinterface` version with pip (using quotes ensures `pip install` works in all terminals/shells):
 
 ```bash
 pip install "spikeinterface[full]"
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ You can also have a look at the [spikeinterface-gui](https://github.com/SpikeInt
 
 ## How to install spikeinterface
 
-You can install the latest version of `spikeinterface` version with pip:
+You can install the latest version of `spikeinterface` version with pip (using quotes ensures `pip install` works in all terminals):
 
 ```bash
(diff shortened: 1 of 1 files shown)


zm711:
This is only necessary for additional dependencies. So although this works for everything I'm always left with the question of will this make people confused that normally they can do `pip install numpy` but for our package they have to do `pip install "spikeinterface[full]"`. Would it make sense to add a note in text that explains that adding the quotes ensures this works in all types of terminals?

h-mayorquin:
Na, I don't think so:

```bash
(test_env) @h-laptop$ pip install "numpy"
```

This works and it is probably to part of the way that works everywhere.

But on the other hand I don't think it hurts. Feel free to add a clarification note if you feel it might help someone.

zm711:
I know pip install "numpy" works. I meant no one ever does that. But yeah I'll add one tiny note.

zm711:
Added a tiny note. Let me if you hate it :)

README.md
```diff
@@ -88,18 +88,18 @@ You can also have a look at the [spikeinterface-gui](https://github.com/SpikeInt
 
 ## How to install spikeinterface
 
-You can install the latest version of `spikeinterface` version with pip:
+You can install the latest version of `spikeinterface` version with pip (using quotes ensures `pip install` works in all terminals):
```
h-mayorquin, line 91: No hay que dejar decaer la pendatería en estos temas, hermano.

```suggestion
You can install the latest version of `spikeinterface` version with pip (using quotes ensures `pip install` works in all terminals/shells):
```

h-mayorquin:
> Added a tiny note. Let me if you hate it :)

No, thanks, I think is an improvement : )

README.md
zm711: aprendo muchas palabras contigo! La pendateria.

h-mayorquin: Vas a ocupar mucho esa palabra si pasas más tiempo conmigo x D

alejoe91 (approved):

PR #3189, opened 2024-07-12
Add a couple of seeds in tests

### LINKED ISSUE #3186: Potential Stochastic Test Failure in `test_generate_sorting_to_inject`
This only failed on one OS/one python version. Not sure if this is a problem in generate_sorting_to_inject which can cause some failures. We can leave this issue open to monitor if this problem happens multiple times.

From action: https://github.com/SpikeInterface/spikeinterface/actions/runs/9898102287/job/27344129816?pr=3185

```
_______________________ test_generate_sorting_to_inject ________________________

    def test_generate_sorting_to_inject():
        durations = [10.0, 20.0]
        sorting = generate_sorting(num_units=10, durations=durations, sampling_frequency=30000, firing_rates=1.0)
>       injected_sorting = generate_sorting_to_inject(
            sorting, [int(duration * sorting.sampling_frequency) for duration in durations]
        )

src/spikeinterface/core/tests/test_generate.py:549: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/spikeinterface/core/generate.py:293: in generate_sorting_to_inject
    min_diff = np.min(np.abs(injected_spike_train[:, None] - spike_train[None, :]), axis=1)
/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/numpy/core/fromnumeric.py:2953: in min
    return _wrapreduction(a, np.minimum, 'min', axis, None, out,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

obj = array([], shape=(0, 0), dtype=int64), ufunc = <ufunc 'minimum'>
method = 'min', axis = 1, dtype = None, out = None
kwargs = {'initial': <no value>, 'keepdims': <no value>, 'where': <no value>}
passkwargs = {}

    def _wrapreduction(obj, ufunc, method, axis, dtype, out, **kwargs):
        passkwargs = {k: v for k, v in kwargs.items()
                      if v is not np._NoValue}
    
        if type(obj) is not mu.ndarray:
            try:
                reduction = getattr(obj, method)
            except AttributeError:
                pass
            else:
                # This branch is needed for reductions like any which don't
                # support a dtype.
                if dtype is not None:
                    return reduction(axis=axis, dtype=dtype, out=out, **passkwargs)
                else:
                    return reduction(axis=axis, out=out, **passkwargs)
    
>       return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
E       ValueError: zero-size array to reduction operation minimum which has no identity
```

alejoe91:
Related to #3186, but not closing because we should investigate the nature of the failures

### DIFF
--- a/src/spikeinterface/core/tests/test_generate.py
+++ b/src/spikeinterface/core/tests/test_generate.py
@@ -547,7 +547,7 @@ def test_generate_sorting_to_inject():
     durations = [10.0, 20.0]
     sorting = generate_sorting(num_units=10, durations=durations, sampling_frequency=30000, firing_rates=1.0)
     injected_sorting = generate_sorting_to_inject(
-        sorting, [int(duration * sorting.sampling_frequency) for duration in durations]
+        sorting, [int(duration * sorting.sampling_frequency) for duration in durations], seed=2308
     )
     num_spikes = sorting.count_num_spikes_per_unit()
     num_injected_spikes = injected_sorting.count_num_spikes_per_unit()
--- a/src/spikeinterface/preprocessing/tests/test_silence.py
+++ b/src/spikeinterface/preprocessing/tests/test_silence.py
@@ -16,7 +16,7 @@ def test_silence(create_cache_folder):
 
     rec = generate_recording()
 
-    rec0 = silence_periods(rec, list_periods=[[[0, 1000], [5000, 6000]], []], mode="zeros")
+    rec0 = silence_periods(rec, list_periods=[[[0, 1000], [5000, 6000]], []], mode="zeros", seed=2308)
     rec0.save(verbose=False)
     traces_in0 = rec0.get_traces(segment_index=0, start_frame=0, end_frame=1000)
     traces_in1 = rec0.get_traces(segment_index=0, start_frame=5000, end_frame=6000)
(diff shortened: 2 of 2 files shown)


zm711 (approved):
This looks good to me.

JoeZiminski:
Looks good but do we know why the original tests were failing? Maybe this is revealing some unexpected behaviour of the underlying machinery that would be good to fix. I say this having no idea of the underlying code or test 😅 but checking just in case.

alejoe91:
> Looks good but do we know why the original tests were failing? Maybe this is revealing some unexpected behaviour of the underlying machinery that would be good to fix. I say this having no idea of the underlying code or test 😅 but checking just in case.

Good point! I suggest we don't close @zm711 issue so it stays on our radar

zm711:
We can just add failures and see if we find a pattern with more data. I think some of the generator code failures sometimes because with certain seeds it doesn't generate good data. But whether that is fixable or not I'm not completely sure. usually if I generate a sorting and it fails I rerun the function and then it works fine. So for me it make sense just to seed in the tests and hope that people in the wild just rerun the function.

Issue #3194, opened 2024-07-12
Enable the option to use names as ids for all neo extractors


h-mayorquin:
This comes from a discussion in neo here:

https://github.com/NeuralEnsemble/python-neo/issues/1507

There, it surfaces that it would be better for the user to work with neo names instead of ids as the channel ids. However, this option is not exposed to the user for most extractors. I think we should enable the user choice at all levels without changing the default  behavior. 

I am also adding comments here on how to document this change from the #3193 here

> So the template of the docstring is:

> The abstract "I don't make any claims" think here.
Some clarification about what each of them are (for example in Intan we can mentioned what is what).
An example of how they look like.

https://github.com/SpikeInterface/spikeinterface/pull/3193#discussion_r1676125976

PR #3195, opened 2024-07-12
Propagate option for using neo channel names as ids in all neo extractors without changing the deafult behavior

h-mayorquin:
Should fix https://github.com/SpikeInterface/spikeinterface/issues/3194

This just make the option available but does not switch default behavior.


samuelgarcia (approved):

h-mayorquin:
> I guess this will help you for neuroconv. I think if we prefer name instead of IDs it is because maybe we made the wrong choie of IDs in neo side. No ?

Yes, probably. I guess we can long term make better choices but that implies a degree of knowledge of the formats and is a lot of work. Maybe better to leave the options open, then change defaults and maybe at some point propagate back. We also have checks for uniquenes  of names on spikeinterface side so there is that.

PR #3198, opened 2024-07-12
Fix naming when `all_annotations=True` in neo base extractors

h-mayorquin:
So this is related to #2784.

There, we made `name` an attribute for representation purposes which is great.
However, this clashes with the use of `all_annotations=True` for neo based extractors that load the stream_name as an annotation wose key is "name". This means that you get these monsters in the repr:

image

(see, the strem name is displayed because the annotation was loaded by neo as name).

This PR corrects that by setting the neo annotation called name as stream_name which is what it actually is.

Another minor fix is to avoid loding ids again as a property and channel_names when the latter are used as ids.


samuelgarcia (approved):

samuelgarcia:
OK

PR #3200, opened 2024-07-12
Remove name class attribute in preprocessing module

### LINKED ISSUE #3199: Name property is not propagated to preprocessing
When we did #2784 one of the goals is that we could set a name and propagate it across preprocessing:

https://github.com/SpikeInterface/spikeinterface/pull/2784#issuecomment-2090177104
@zm711 

However the dream is still not realized:
image

The reason is because most pre-processors still have a name attribute. Examples:

https://github.com/SpikeInterface/spikeinterface/blob/b3383575feade4bf70d623937d6ea2db9cd05b37/src/spikeinterface/preprocessing/normalize_scale.py#L187-L207

And
https://github.com/SpikeInterface/spikeinterface/blob/b3383575feade4bf70d623937d6ea2db9cd05b37/src/spikeinterface/preprocessing/filter.py#L172-L196

The solution seems easy, just remove the names there and this should work.

h-mayorquin:
Related and progresses on  #2784

Should close #3199

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -46,7 +46,6 @@ class BaseExtractor:
     # these properties are skipped by default in copy_metadata
     _skip_properties = []
 
-    installed = True
     installation_mesg = ""
 
     def __init__(self, main_ids: Sequence) -> None:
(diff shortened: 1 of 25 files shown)


alejoe91:
Thanks @h-mayorquin 

Can you remove the installed=True as well so we make a proper cleanup? :)

h-mayorquin:
Ok, done.

samuelgarcia (approved):

PR #2934, opened 2024-05-30
Improve and benchmark auto-merging functions

yger:
This PR adds a meta merging benchmarks with various methods that can be tested/debugged. Plots are in progress, but the plan is to be able to properly benchmark the final step of the sorters

PR #3203, opened 2024-07-15
Port auto-merge changes and refactor

alejoe91:
Port the `curation` part of #2934 and refactor


yger:
I thought drift_tools was excluded from this PR, am I right?

alejoe91:
@yger @samuelgarcia ready to review

I did some renaming to unify the argument names. Let me know what you think!

src/spikeinterface/curation/auto_merge.py
```diff
@@ -520,14 +781,379 @@ def check_improve_contaminations_score(
...
+
+    res = [scipy.stats.binom.sf(x, n_, p) for n_ in n_array]
+    f = scipy.interpolate.interp1d(n_array, res, kind="quadratic")
+
+    return f(n)
+
+
+if HAVE_NUMBA:
+
+    @numba.jit((numba.float32,), nopython=True, nogil=True, cache=True)
```
alejoe91, line 941:
```suggestion
    @numba.jit(nopython=True, nogil=True, cache=False)
```

src/spikeinterface/curation/auto_merge.py
```diff
@@ -520,14 +781,379 @@ def check_improve_contaminations_score(
...
+        border_low = math.floor(max_time)
+        p_high = 0.5 * (max_time - border_high + 1) ** 2
+        p_low = 0.5 * (1 - (max_time - border_low) ** 2) + (max_time - border_low)
+
+        if border_low == 0:
+            p_low -= 0.5 * (-max_time + 1) ** 2
+
+        return border_low, border_high, p_low, p_high
+
+    @numba.jit((numba.int64[:], numba.float32), nopython=True, nogil=True, cache=True)
```
alejoe91, line 973:
```suggestion
    @numba.jit(nopython=True, nogil=True, cache=False)
```

src/spikeinterface/curation/auto_merge.py
```diff
@@ -520,14 +781,379 @@ def check_improve_contaminations_score(
...
+                if diff == border_high:
+                    n_violations_high += 1
+                elif diff == border_low:
+                    n_violations_low += 1
+                else:
+                    n_violations += 1
+
+        return n_violations + p_high * n_violations_high + p_low * n_violations_low
+
+    @numba.jit((numba.int64[:], numba.int64[:], numba.float32), nopython=True, nogil=True, cache=True)
```
alejoe91, line 1014:
```suggestion
    @numba.jit(nopython=True, nogil=True, cache=False)
```

src/spikeinterface/curation/auto_merge.py
```diff
@@ -1,53 +1,74 @@
...
+from ..core import SortingAnalyzer, Templates
 from ..core.template_tools import get_template_extremum_channel
 from ..postprocessing import compute_correlograms
 from ..qualitymetrics import compute_refrac_period_violations, compute_firing_rates
 
 from .mergeunitssorting import MergeUnitsSorting
+from .curation_tools import resolve_merging_graph
+
+
+_possible_presets = ["default", "xcontamination", "temporal_splits", "feature_neighbors"]
```
yger, line 23: I would put xcontaminations with an s, but maybe this is too much. And then maybe default is also not really informative, when all the others start to be. What about xcorrelograms?

src/spikeinterface/curation/auto_merge.py
```diff
@@ -61,193 +82,373 @@ def get_potential_auto_merge(
...
-            censored_period_ms,
-        )
+            else:
+                templates_ext = sorting_analyzer.get_extension("templates")
+                assert (
+                    templates_ext is not None
+                ), "auto_merge with template_similarity requires a SortingAnalyzer with extension templates"
+                templates_array = templates_ext.get_data(outputs="numpy")
+
+                templates_diff = compute_templates_diff(
```
yger, line 347: I know we want to keep the default behavior, but we want to quickly enforce the computation of template_similarity widget, since it will be more accurate than the one performed with lussac (because of the adaptative masks). Maybe for the refactoring of the fonction, we should list, for every preset, the requiered extensions and compute them if not present in the analyzer

yger (approved):

src/spikeinterface/curation/auto_merge.py
samuelgarcia: I am agree. We could put "legacy" or better name.

src/spikeinterface/curation/curation_tools.py
```diff
@@ -133,3 +134,44 @@ def find_duplicated_spikes(
         return _find_duplicated_spikes_keep_last_iterative(spike_train.astype(np.int64), censored_period)
     else:
         raise ValueError(f"Method '{method}' isn't a valid method for find_duplicated_spikes. Use one of {_methods}")
+
+
+def remove_empty_units(sorting_or_sorting_analyzer, minimum_spikes=10):
```
samuelgarcia, line 139: We already have this as a method Sorting no ?
I would put 0 by default.
We should make a doc

src/spikeinterface/curation/auto_merge.py
alejoe91: done

src/spikeinterface/curation/curation_tools.py
alejoe91: removed

alejoe91:
@yger we decided to rely on precomputed extensions, and this breaks SC2. Is it ok to change it there?

yger:
Can I push in the branch to fix that?

PR #3215, opened 2024-07-16
Implement a simple system to have backward compatibility for Analyzer extension

samuelgarcia:
This simple system enabling to hack on load `self.params` and `self.data` when a new features are implemented in an analyzer extension that imply new parameters and new variable in data.

It was for instace the case for template similarity


zm711 (commented):
Do we not need to set need_backward_compatibility_on_load for the similarity?

samuelgarcia:
oups

zm711:
Here's what I don't understand. Now that we've added the `need_backward_compatibility_on_load` now it is going back to failing with the first error where it can't find the `max_lag_ms`. I need to read this a bit better why it worked when that line was missing, but isn't working now.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1814,6 +1819,9 @@ def load(cls, sorting_analyzer):
         ext = cls(sorting_analyzer)
         ext.load_params()
         ext.load_data()
+        if cls.need_backward_compatibility_on_load:
+            self._handle_backward_compatibility_on_load()
```
zm711, line 1823: Should this be `self or cls`

zm711: No you're right it is should be self.

zm711:
Since the most recent commit, nothing I do works. I tried going back in and just deleting `need_backward` so it was like your original commit. But now regardless of what I do it gives me the missing `max_lag_ms` error.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -777,6 +777,10 @@ def _save_or_select_or_merge(
         # make a copy of extensions
         # note that the copy of extension handle itself the slicing of units when necessary and also the saveing
         sorted_extensions = _sort_extensions_by_dependency(self.extensions)
+        # hack: quality metrics are computed at last
+        qm_extension_params = sorted_extensions.pop("quality_metrics")
+        if qm_extension_params is not None:
+            sorted_extensions["quality_metrics"] = qm_extension_params
```
alejoe91, line 783: this is a trick needed for recompute, since quality metrics don't explicitly depend on principal_components, but if PCA metrics have been computed we need to enforce quality metrics to be computed at last

src/spikeinterface/postprocessing/unit_locations.py
```diff
@@ -42,10 +42,17 @@ class ComputeUnitLocations(AnalyzerExtension):
...
+    need_backward_compatibility_on_load = True
 
     def __init__(self, sorting_analyzer):
         AnalyzerExtension.__init__(self, sorting_analyzer)
 
+    def _handle_backward_compatibility_on_load(self):
+        if "method_kwargs" in self.params:
+            # make compatible analyzer created between february 24 and july 24
+            method_kwargs = self.params.pop("method_kwargs")
+            self.params.updated(**method_kwargs)
```
zm711, line 54:
```suggestion
            self.params.update(**method_kwargs)
```

typo causing load to fail. Let me push this then I'll try again.

src/spikeinterface/postprocessing/unit_locations.py
```diff
@@ -51,7 +51,7 @@ def _handle_backward_compatibility_on_load(self):
         if "method_kwargs" in self.params:
             # make compatible analyzer created between february 24 and july 24
             method_kwargs = self.params.pop("method_kwargs")
-            self.params.updated(**method_kwargs)
+            self.params.update(**method_kwargs)
```
alejoe91, line 54: oups!

zm711:
Here's the current bug in merge units
```

  File ~\Documents\GitHub\spikeinterface\src\spikeinterface\core\sortinganalyzer.py:939 in merge_units
    if len(units) < 2:

TypeError: object of type 'int' has no len()
```

zm711:
I have to finish up a presentation at the moment, so I can test another round in the background, but I can't track down this bug. Just let me know when ready for another test :)

zm711:
Never mind I put the list in as a list instead of list of list. Let me try again :)

zm711:
Okay the merge code works from loading an analyzer from 08MAY24 :) !!

src/spikeinterface/core/sortinganalyzer.py
samuelgarcia: you need a pop ("quality_metrics", None) no ?

alejoe91:
> Okay the merge code works from loading an analyzer from 08MAY24 :) !!

For me it works for waveform extractor folders loaded as analyzers from last year :P

samuelgarcia:
I hope you both have tried sigui for this dataset.

zm711:
It has a probegroup. Has that been patched yet?

src/spikeinterface/core/sortinganalyzer.py
alejoe91: it's done already! Fixed it

yger:
Not sure the problem is fully solved. While I can load the analyzer you gave me @alejoe91 for testing the merging, when applying new functions recursively, I still have errors such as 

nbefore = int(self.params["ms_before"] * self.sorting_analyzer.sampling_frequency
return nbefore

KeyError: 'ms_before'

alejoe91:
@yger I know, will push a fix tomorrow!

alejoe91:
@yger fixed! The templates were not an extensions so they required a fix before in the code

alejoe91 (approved):

src/spikeinterface/core/waveforms_extractor_backwards_compatibility.py
```diff
@@ -614,13 +616,12 @@ def make_ext_params_up_to_date(ext, old_params, new_params):
     old_name = ext.extension_name
     updated_params = old_params.copy()
     for p, values in old_params.items():
-        if isinstance(values, dict):
+        if p not in new_params:
+            warnings.warn(f"Removing legacy param {p} from {old_name} extension")
```
zm711, line 620:
```suggestion
            warnings.warn(f"Removing legacy parameter {p} from {old_name} extension")
```

should we be more explicit for the end user. We call it param in the code base, but the end user would understand parameter better, no?

zm711 (commented):
Just one tiny error comment.

src/spikeinterface/core/waveforms_extractor_backwards_compatibility.py
```diff
@@ -548,6 +552,8 @@ def _read_old_waveforms_extractor_binary(folder, sorting):
         new_params = ext._set_params()
         updated_params = make_ext_params_up_to_date(ext, params, new_params)
         ext.set_params(**updated_params, save=False)
+        if ext.need_backward_compatibility_on_load:
+            ext._handle_backward_compatibility_on_load()
```
samuelgarcia, line 552: This must be moved after the `ext.data = ...` because this function can hack the two sub dict (params + data).

So move line 612

alejoe91: You mean to move this at the end of the function?

alejoe91: I don't think we need to modify the data, since we build it on the fly based on the current API. Or am I getting this wrong?

PR #3217, opened 2024-07-17
Fix bug in extractor_dict_iterator if element is not str/Path

alejoe91:
@samuelgarcia


samuelgarcia (approved):

h-mayorquin:
How was this failing?

h-mayorquin:
Ah, short circuit to only do the second container test on str and Path types. Make sense. My bad.

alejoe91:
Yep indeed! :)

PR #3218, opened 2024-07-17
Fix estimate_motion when time_vector is set

alejoe91:
@samuelgarcia 

Also added tests


alejoe91:
> you shoudll check in the tests that the motion bins are iside the time vector no ?

done in last commit!

samuelgarcia (approved):

PR #3227, opened 2024-07-18
Remove unused rng from `drifting_generator.py`

JoeZiminski:
This PR removes an unused `rng = np.random.default_rng(seed=seed)` from `generate_drifting_recording()`. I think this is a better pattern vs. setting global seed (detailed in [this](https://blog.scientific-python.org/numpy/numpy-rng/) nice article @h-mayorquin originally posted) but is currently unused so removed. An alternative is to make refactorings to pass this around?


h-mayorquin (approved):
Yes. Left around for some reason.

JoeZiminski:
@alejoe91 is this okay to merge?

PR #3229, opened 2024-07-18
Add `int` type to `num_samples` on `InjectTemplatesRecording`.

JoeZiminski:
Adds a missing case for the `num_samples` parameter type hint on `InjectTemplatesRecording`. @zm711 @h-mayorquin was the consensus to change `Union` and `Optional` to `|` in type hints? If so should these be changed in this module?


h-mayorquin:
Yes, we have been using `|` instead of Union or Optional. I think is less verbose.

zm711:
Also we have switched to all native typing rather than the typing module types. So all the uses of `List` can be changed to `list`.

JoeZiminski:
Great thanks both I'll make some changes to this module then request review

zm711:
It'll be the sixth republic before Sam let's us do static type checking. So we've worked under the `from __future__ import annotations` in so much of the code base that changing this would be a drop in the bucket. I wouldn't worry about it.

src/spikeinterface/core/generate.py
```diff
@@ -1690,13 +1688,13 @@ class InjectTemplatesRecording(BaseRecording):
         Can be None (no scaling).
         Can be scalar all spikes have the same factor (certainly useless).
         Can be a vector with same shape of spike_vector of the sorting.
-    parent_recording: BaseRecording | None
+    parent_recording: BaseRecording | None, default: None
         The recording over which to add the templates.
         If None, will default to traces containing all 0.
-    num_samples: list[int] | int | None
+    num_samples: list[int] | int | None, default: None
```
zm711, line 1694:
```suggestion
    num_samples : list[int] | int | None, default: None
```

src/spikeinterface/core/generate.py
```diff
@@ -1690,13 +1688,13 @@ class InjectTemplatesRecording(BaseRecording):
         Can be None (no scaling).
         Can be scalar all spikes have the same factor (certainly useless).
         Can be a vector with same shape of spike_vector of the sorting.
-    parent_recording: BaseRecording | None
+    parent_recording: BaseRecording | None, default: None
```
zm711, line 1691:
```suggestion
    parent_recording : BaseRecording | None, default: None
```

src/spikeinterface/core/generate.py
```diff
@@ -1690,13 +1688,13 @@ class InjectTemplatesRecording(BaseRecording):
...
         The recording over which to add the templates.
         If None, will default to traces containing all 0.
-    num_samples: list[int] | int | None
+    num_samples: list[int] | int | None, default: None
         The number of samples in the recording per segment.
         You can use int for mono-segment objects.
-    upsample_vector: np.array or None, default: None.
+    upsample_vector: np.array | None, default: None.
```
zm711, line 1697:
```suggestion
    upsample_vector : np.array | None, default: None.
```

src/spikeinterface/core/generate.py
```diff
@@ -1710,11 +1708,11 @@ def __init__(
...
-        nbefore: Union[List[int], int, None] = None,
-        amplitude_factor: Union[List[List[float]], List[float], float, None] = None,
-        parent_recording: Union[BaseRecording, None] = None,
-        num_samples: Optional[List[int]] = None,
-        upsample_vector: Union[List[int], None] = None,
+        nbefore: list[int] | int | None = None,
+        amplitude_factor: list[float] | float | None = None,
+        parent_recording: BaseRecording | None = None,
+        num_samples: list[int] | int | None = None,
+        upsample_vector: np.array | None = None,
```
zm711, line 1715:
```suggestion
        nbefore : list[int] | int | None = None,
        amplitude_factor : list[float] | float | None = None,
        parent_recording : BaseRecording | None = None,
        num_samples : list[int] | int | None = None,
        upsample_vector : np.array | None = None,
```

src/spikeinterface/core/generate.py
```diff
@@ -1846,10 +1844,10 @@ def __init__(
...
         templates: np.ndarray,
         nbefore: int,
-        amplitude_vector: Union[List[float], None],
-        upsample_vector: Union[List[float], None],
-        parent_recording_segment: Union[BaseRecordingSegment, None] = None,
-        num_samples: Union[int, None] = None,
+        amplitude_vector: list[float] | None,
+        upsample_vector: list[float] | None,
+        parent_recording_segment: BaseRecordingSegment | None = None,
+        num_samples: int | None = None,
```
zm711, line 1850:
```suggestion
        amplitude_vector : list[float] | None,
        upsample_vector : list[float] | None,
        parent_recording_segment : BaseRecordingSegment | None = None,
        num_samples : int | None = None,
```

src/spikeinterface/core/generate.py
```diff
@@ -1869,9 +1867,9 @@ def __init__(
 
     def get_traces(
         self,
-        start_frame: Union[int, None] = None,
-        end_frame: Union[int, None] = None,
-        channel_indices: Union[List, None] = None,
+        start_frame: int | None = None,
+        end_frame: int | None = None,
+        channel_indices: list | None = None,
```
zm711, line 1872:
```suggestion
        start_frame : int | None = None,
        end_frame : int | None = None,
        channel_indices : list | None = None,
```

src/spikeinterface/core/generate.py
```diff
@@ -1089,11 +1089,11 @@ def __init__(
...
-        durations: List[float],
-        noise_levels: float = 1.0,
-        cov_matrix: Optional[np.array] = None,
-        dtype: Optional[Union[np.dtype, str]] = "float32",
-        seed: Optional[int] = None,
+        durations: list[float],
+        noise_levels: float | np.array = 1.0,
+        cov_matrix: np.array | None = None,
+        dtype: np.dtype | str | None = "float32",
+        seed: int | None = None,
```
zm711, line 1096:
```suggestion
        durations : list[float],
        noise_levels : float | np.array = 1.0,
        cov_matrix : np.array | None = None,
        dtype : np.dtype | str | None = "float32",
        seed : int | None = None,
```

src/spikeinterface/core/generate.py
```diff
@@ -1060,15 +1060,15 @@ class NoiseGeneratorRecording(BaseRecording):
         The number of channels.
     sampling_frequency : float
         The sampling frequency of the recorder.
-    durations : List[float]
+    durations : list[float]
         The durations of each segment in seconds. Note that the length of this list is the number of segments.
     noise_levels: float or array, default: 1
         Std of the white noise (if an array, defined by per channels)
-    cov_matrix: np.array, default None
+    cov_matrix: np.array | None, default: None
```
zm711, line 1067:
```suggestion
    cov_matrix : np.array | None, default: None
```

src/spikeinterface/core/generate.py
```diff
@@ -333,10 +333,10 @@ class TransformSorting(BaseSorting):
...
         self,
         sorting: BaseSorting,
-        added_spikes_existing_units=None,
-        added_spikes_new_units=None,
-        new_unit_ids: Optional[List[Union[str, int]]] = None,
-        refractory_period_ms: Optional[float] = None,
+        added_spikes_existing_units: np.array | None = None,
+        added_spikes_new_units: np.array | None = None,
+        new_unit_ids: list[str | int] | None = None,
+        refractory_period_ms: float | None = None,
```
zm711, line 339:
```suggestion
        added_spikes_existing_units : np.array | None = None,
        added_spikes_new_units : np.array | None = None,
        new_unit_ids : list[str | int] | None = None,
        refractory_period_ms : float | None = None,
```

src/spikeinterface/core/generate.py
```diff
@@ -313,13 +313,13 @@ class TransformSorting(BaseSorting):
     ----------
     sorting : BaseSorting
         The sorting object.
-    added_spikes_existing_units : np.array (spike_vector)
+    added_spikes_existing_units : np.array (spike_vector) | None, default: None
         The spikes that should be added to the sorting object, for existing units.
-    added_spikes_new_units: np.array (spike_vector)
+    added_spikes_new_units: np.array (spike_vector) | None, default: None
```
zm711, line 318:
```suggestion
    added_spikes_new_units : np.array (spike_vector) | None, default: None
```

src/spikeinterface/core/generate.py
```diff
@@ -313,13 +313,13 @@ class TransformSorting(BaseSorting):
...
         The spikes that should be added to the sorting object, for new units.
-    new_units_ids: list
+    new_units_ids: list[str, int] | None, default: None
```
zm711, line 320:
```suggestion
    new_units_ids : list[str, int] | None, default: None
```

src/spikeinterface/core/generate.py
```diff
@@ -253,13 +253,13 @@ def generate_sorting_to_inject(
     num_samples: list of size num_segments.
         The number of samples in all the segments of the sorting, to generate spike times
         covering entire the entire duration of the segments.
-    max_injected_per_unit: int, default 1000
+    max_injected_per_unit: int, default: 1000
```
zm711, line 256:
```suggestion
    max_injected_per_unit : int, default: 1000
```

src/spikeinterface/core/generate.py
```diff
@@ -253,13 +253,13 @@ def generate_sorting_to_inject(
...
         The maximal number of spikes injected per units.
-    injected_rate: float, default 0.05
+    injected_rate: float, default: 0.05
```
zm711, line 258:
```suggestion
    injected_rate : float, default: 0.05
```

src/spikeinterface/core/generate.py
```diff
@@ -253,13 +253,13 @@ def generate_sorting_to_inject(
...
         The rate at which spikes are injected.
-    refractory_period_ms: float, default 1.5
+    refractory_period_ms: float, default: 1.5
```
zm711, line 260:
```suggestion
    refractory_period_ms : float, default: 1.5
```

src/spikeinterface/core/generate.py
```diff
@@ -253,13 +253,13 @@ def generate_sorting_to_inject(
...
         The refractory period that should not be violated while injecting new spikes.
-    seed: int, default None
+    seed: int, default: None
```
zm711, line 262:
```suggestion
    seed : int, default: None
```

src/spikeinterface/core/generate.py
```diff
@@ -44,13 +44,13 @@ def generate_recording(
         The number of channels in the recording.
     sampling_frequency : float, default: 30000. (in Hz)
         The sampling frequency of the recording, default: 30000.
-    durations: List[float], default: [5.0, 2.5]
+    durations: list[float], default: [5.0, 2.5]
```
zm711, line 47:
```suggestion
    durations : list[float], default: [5.0, 2.5]
```

zm711 (commented):
numpydoc standard requires a space before the colon after an argument. So the final format is:

```
arg : float, default: 5.0
```

we put a space before the first colon since it is required for the linter, but not for the second one. This is Chris's project, but I figure we can help him out a little ;)

src/spikeinterface/core/generate.py
h-mayorquin: Should it not be optional in the case of None?

JoeZiminski: As in the `default: None` part? I'm not sure on this @zm711 may know, happy to make any changes as required

JoeZiminski:
Thanks both! I've applied changes and made some further changes to docstrings, re-requested review

src/spikeinterface/core/generate.py
zm711: Yeah, this is an issue where we had previously agreed that None would be `default: None` so much of the code base has been transitioning over to that style. But more recently we had a discussion about using `, num_samples : list[int] | int | None, optional` as an alternative style. Maybe we should add this to a maintenance meeting that way we can all discuss this in person and make sure we are all on the same page with the style. We could do a quick vote to make sure everyone agrees on which of the two styles.

src/spikeinterface/core/generate.py
```diff
@@ -793,20 +795,20 @@ def synthesize_random_firings(
 
     Parameters
     ----------
-    num_units : int
+    num_units : int, default: 20
         Number of units.
-    sampling_frequency : float
+    sampling_frequency : float, default: 30000.0 (in Hz)
```
zm711, line 800: For these sampling frequency default values I don't know if they should be (in Hz) vs having the (in Hz) down in the actual description. So for me I would prefer:

```
sampling_frequency : float, default: 30000.0
    The sampling rate entered in Hz. (or samples/second)
```

opinions @chrishalcrow @h-mayorquin ?

src/spikeinterface/core/generate.py
```diff
@@ -1279,11 +1281,14 @@ def generate_recording_by_size(
...
-    seed : int, default: None
+    seed : int | None, default: None
         The seed for np.random.default_rng.
-
+    strategy : "tile_pregenerated" | "on_the_fly", default: "tile_pregenerated"
+        The strategy of generating noise chunk:
+          * "tile_pregenerated": pregenerate a noise chunk of noise_block_size sample and repeat it
+                                 very fast and cusume only one noise block.
+          * "on_the_fly": generate on the fly a new noise block by combining seed + noise block index
+                          no memory preallocation but a bit more computaion (random)
```
zm711, line 1291:
```suggestion
                          no memory preallocation but a bit more computation (random)
```

src/spikeinterface/core/generate.py
```diff
@@ -1554,6 +1559,10 @@ def generate_templates(
             * array of the same length of units
             * scalar, then an array is created
             * tuple, then this difine a range for random values.
+    mode : "ellipsoid" | "sphere", default: "ellipsoid"
+        Method used to calculate the distance between unit and channel location.
+        Ellipoid injects some anisotropy dependent on unit shape, sphere is equivalent
```
zm711, line 1564:
```suggestion
        Ellipsoid injects some anisotropy dependent on unit shape, sphere is equivalent
```

zm711:
@JoeZiminski one more discussion point and then a couple of typos :)

src/spikeinterface/core/generate.py
chrishalcrow: I'd personally prefer the Hz to be in the description, so that a user can copy and paste the default value and it would work.

JoeZiminski: Agree I prefer in the description I will make the change

src/spikeinterface/core/generate.py
JoeZiminski: Nice this sounds good, I think I prefer `default: None` for consistency and it is clearer than `optional` if you don't know much about typing. If most of the code is `default: None` shall I leave this for now and we can discuss at the meeting?

JoeZiminski:
Thanks @zm711! After all that talk of codespell I should actually remember to run it locally 😆

src/spikeinterface/core/generate.py
zm711: I'm fine with that for now :)

h-mayorquin (approved):
LGTM

src/spikeinterface/core/generate.py
h-mayorquin: All right if you find those semantics clearer.

PR #3230, opened 2024-07-18
Add check for None in 'NoiseGeneratorRecordingSegment' get_traces().

JoeZiminski:
At the moment the default value of `NoiseGeneratorRecordingSegment.get_traces()` is `None` but it is not handled in the code and would error if called with default values. This PR sets to `start_frame=0` and `end_frame=self.get_num_samples()`. A bug caught thanks to use of type hints and mypy!


samuelgarcia:
Hi this is related also to this #3079

JoeZiminski:
@samuelgarcia @h-mayorquin should this be closed in favour of the upcoming changes? I guess these `None` defaults will be removed in the future PR? If so, is it worth adding this protection in the meantime? sorry 3x questions 😄

h-mayorquin:
I think it is fine to have in the meantime as well.

alejoe91 (approved):

PR #3237, opened 2024-07-19
Handle case where channel count changes from probeA to probeB

cwindolf:
I think that this small change is necessary to handle interpolating templates from one probe to another when they have different channel counts

### DIFF
--- a/src/spikeinterface/generation/drift_tools.py
+++ b/src/spikeinterface/generation/drift_tools.py
@@ -40,9 +40,13 @@ def interpolate_templates(templates_array, source_locations, dest_locations, int
     source_locations = np.asarray(source_locations)
     dest_locations = np.asarray(dest_locations)
     if dest_locations.ndim == 2:
-        new_shape = templates_array.shape
+        new_shape = (*templates_array.shape[:2], len(dest_locations))
     elif dest_locations.ndim == 3:
-        new_shape = (dest_locations.shape[0],) + templates_array.shape
+        new_shape = (
+            dest_locations.shape[0],
+            *templates_array.shape[:2],
+            dest_locations.shape[1],
+        )
     else:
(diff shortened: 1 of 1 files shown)


JoeZiminski:
Hey @cwindolf this looks good to me (just reading the code, not testing in practice). Maybe these new cases can be added to the [interpolation tests](https://github.com/SpikeInterface/spikeinterface/blob/8b84711c0272ce9d35f248a3ad20c22fa3f51730/src/spikeinterface/generation/tests/test_drift_tools.py#L74). i.e. pass cases where the number of source and destination channel locations does / does not match, under the `n_dim=2` or `n_dim=3` conditions, and check the template output shape is as expected?

As an aside, at present the current test is a 'smoke test' that check `interpolate_motion` runs without error but does not check that output is as expcted. It would be cool to also extend this to check the maximum loading channel of the template changes as expected after interpolation, but I think this is outside the scope of this PR!

alejoe91 (approved):

src/spikeinterface/generation/drift_tools.py
```diff
@@ -40,9 +40,13 @@ def interpolate_templates(templates_array, source_locations, dest_locations, int
     source_locations = np.asarray(source_locations)
     dest_locations = np.asarray(dest_locations)
     if dest_locations.ndim == 2:
-        new_shape = templates_array.shape
+        new_shape = (*templates_array.shape[:2], len(dest_locations))
```
samuelgarcia, line 43: using len() is always ambiguous I think for arrays.
len = dest_locations.shape[0] but here I think you wanted dest_locations.shape[1]
@cwindolf  : can you check this ? @alejoe91 merged this too quickly and I think we need a new PR now.
no sure.

PR #3238, opened 2024-07-19
Fix #3236

### LINKED ISSUE #3236: Assertion error when using backend="ipywidgets" indicating to set `%matplotlib widget`
If I run the following code in SI version 0.100.7, it works fine, but in versions 0.100.8 or 0.101.0rc1, it doesn't.

Here is the code running in the VS Code Jupyter Notebook on a Windows 11 computer:

`%matplotlib widget`
`si.plot_traces(raw_rec, backend="ipywidgets", mode='line', channel_ids=raw_rec.channel_ids[:8])`

This gives the following error:
---> [13](file:///C:/Users/pdrfl/Anaconda3/envs/si_101rc1/Lib/site-packages/spikeinterface/widgets/utils_ipywidgets.py:13) assert "ipympl" in mpl_backend, "To use the 'ipywidgets' backend, you have to set %matplotlib widget"

AssertionError: To use the 'ipywidgets' backend, you have to set %matplotlib widget

The problem is only with `backend="ipywidgets"`, while using other options like `backend="ephyviewer"` works fine.

h-mayorquin:
It would be nice if someone knew the exact expected behavior of this. The behavior is not documented in matplotilb.get_backend() as far as I can tell but this is how it works in my system:

https://matplotlib.org/stable/api/matplotlib_configuration_api.html#matplotlib.get_backend

Maybe something that changed recently:

image


chrishalcrow:
Can confirm this fixes the bug on my Linux machine, and doesn't break anything on my Mac.

zm711:
That would explain why I thought it worked fine. I had been doing it all on my mac for this stuff....

JoeZiminski:
Can confirm this doesn't break anything on Windows either (could not reproduce #3236 on Windows), interestingly the output is completely different. I'm not sure why it's the same for both backends, I can confirm I have 

```
ipympl 0.9.4
ipywidgets 8.1.3
matplotlib 
```
and restarting the kernel between the two:

image


image

zm711:
This could be a very specific version issue, I don't which package. Not sure if we want to search for the root since it is so inconsistent or just use this PR to not worry about the assert.

zm711:
Same as @JoeZiminski , I just tested this on Windows and couldn't replicate the issue. I get the same backend as Joe. But if I do `%matplotlib ipympl` it gives me a warning and says it will automatically switch to widget mode for me. But works fine.

JoeZiminski:
Okay that's great @zm711 that it repliaces for you too, weird how different it is across platforms! I guess if its working cross-platform at the moment it makes sense to keep.

alejoe91 (approved):

h-mayorquin:
I will merge this as someone in #3236 is asking for this.

Issue #3222, opened 2024-07-17
Set subplot parameters in UnitSummaryWidget


florian6973:
Hi!

I am currently using the UnitSummaryWidget that I find very practical, however I noticed that we cannot change the parameters of the subplots like the AmplitudesWidget: in the parameters of the AmplitudeWidget we can change the number of bins but here we do not have access. 
![image](https://github.com/user-attachments/assets/b5c3f195-e0cb-42ed-9213-948b3000ad37)

Could we potentially add to the unit summary widget additional fields to pass parameters to subwidgets please? I can make a small PR if needed. Thanks!

Best,

zm711:
I think this is a great idea. I've been meaning to suggest something like this for a while.  As long as @samuelgarcia or @alejoe91 don't have a reason not to I would say go for the PR. I would definitely use this!

alejoe91:
Yep! Please go ahead with the PR!

zm711:
This is all done so I'll close now :)

PR #3242, opened 2024-07-22
Add subwidget parameters for UnitSummaryWidget

florian6973:
Hi,

As discussed in #3222, here is a very small PR to add this feature. Please let me know if you would like any changes before merging. 

To test:
```
import spikeinterface.widgets as sw
rec, sort = si.generate_ground_truth_recording(num_channels=4)
sa.compute(["random_spikes", "templates", "spike_amplitudes", "waveforms"])
sw.plot_unit_summary(sa, 0, unitwaveformswidget_params={"plot_channels": True}, amplitudeswidget_params={"bins": 20})
```

Best,


zm711:
Maybe @samuelgarcia would feel differently, but I would actually like the idea of a nested dict instead of multiple dicts. Maybe @h-mayorquin would also like to weigh in on the design choice. So it would be something like

```python
widget_param_dict = dict(amplitude_params=xx, etc)
```

And then if None you make an empty dict and just iterate through the keys to create your sub dictionaries. That way if we changes the widgets it would be easy to change that internally without breaking the external API.

florian6973:
Thanks for your feedback! I just did the change to a nested dict :)

src/spikeinterface/widgets/unit_summary.py
```diff
@@ -30,6 +31,13 @@ class UnitSummaryWidget(BaseWidget):
     sparsity : ChannelSparsity or None, default: None
         Optional ChannelSparsity to apply.
         If SortingAnalyzer is already sparse, the argument is ignored
+    widget_params : dict or None, default: None
+        Parameters for the subwidgets in a nested dictionary
+            unitlocations_params: UnitLocationsWidget (see UnitLocationsWidget for details)
+            unitwaveforms_params: UnitWaveformsWidget (see UnitWaveformsWidget for details)
+            unitwaveformdensitymap_params : UnitWaveformDensityMapWidget (see UnitWaveformDensityMapWidget for details)
+            autocorrelograms_params : AutoCorrelogramsWidget (see AutoCorrelogramsWidget for details)
+            amplitudes_params : AmplitudesWidget (see AmplitudesWidget for details)
```
alejoe91, line 40: @florian6973 I would actually drop the `_params` and add underscores (it's clear from the `widgets_params` already):
```suggestion
            unit_locations : UnitLocationsWidget (see UnitLocationsWidget for details)
            unit_waveforms : UnitWaveformsWidget (see UnitWaveformsWidget for details)
            unit_waveform_density_map : UnitWaveformDensityMapWidget (see UnitWaveformDensityMapWidget for details)
            autocorrelograms : AutoCorrelogramsWidget (see AutoCorrelogramsWidget for details)
            amplitudes : AmplitudesWidget (see AmplitudesWidget for details)
```

alejoe91 (changes requested):
Thanks @florian6973 

I have a suggestion on the naming of the additional params

florian6973:
Thanks for your suggestion! Just implemented the changes.

Example:
```python
import spikeinterface.widgets as sw
rec, sort = si.generate_ground_truth_recording(num_channels=4)
sa.compute(["random_spikes", "templates", "spike_amplitudes", "waveforms"])
sw.plot_unit_summary(sa, 0, widget_params=dict(unit_waveforms={"plot_channels": True}, amplitudes={"bins": 5}))
```

src/spikeinterface/widgets/unit_summary.py
```diff
@@ -70,6 +81,13 @@ def plot_matplotlib(self, data_plot, **backend_kwargs):
         unit_colors = dp.unit_colors
         sparsity = dp.sparsity
 
+        widget_params = defaultdict(lambda: dict(), dp.widget_params)
```
JoeZiminski, line 84: cool! I did not know about that `defaultdict`. Might be adding a quick comment like `# Returns an empty dict if the key is not in passed in dp.widget_params`, but maybe this is quite widely known and doesn't need the comment

JoeZiminski:
Hey @florian6973 this is really cool! I played around with it and it was working well. I had one small point above and two below, please feel free to ignore them they are minor.

- the `widget_params` could be renamed to `widget_kwargs` which I think (?) is usually the convention in SI for kwargs that are passed down the call chain to subfunctions / classes. Maybe this could also be `subwidget_kwargs`  to distinguish it from the other arguments of the function (technically all of the params passed are 'widget params', for the `UnitSummaryWidget`).
- I naively tried to pass `unit_colors` to one of the subwidget and it raised a slightly confusing error (the subwidget received two duplicate `unit_colors` arguments), as that is already handled by the `UnitSummaryWidget`. So it's not possible to override this in the `widget_params`.  Some options would be to raise an error saying this is not supported, or just add to the docstring to that effect, or to override the `unit_colors` for the subwidget if it is passed in the dict. This is an edge-case so docstring is probably easiest!

florian6973:
Thanks for your feedback! Just pushed some changes to rename the variables and add a clearer error message for this edge case.

```python
import spikeinterface.widgets as sw
rec, sort = si.generate_ground_truth_recording(num_channels=4)
sa.compute(["random_spikes", "templates", "spike_amplitudes", "waveforms"])
sw.plot_unit_summary(sa, 0, unit_colors=['r'], subwidget_kwargs=dict(unit_waveforms={"plot_channels": True}, amplitudes={"bins": 5}))
                     
from matplotlib import pyplot as plt
plt.show()
```

src/spikeinterface/widgets/unit_summary.py
florian6973: Just added a comment :)

src/spikeinterface/widgets/unit_summary.py
```diff
@@ -17,19 +18,27 @@ class UnitSummaryWidget(BaseWidget):
...
+    unit_id: int or str
         The unit id to plot the summary of
-    unit_colors : dict or None, default: None
+    unit_colors: dict or None, default: None
         If given, a dictionary with unit ids as keys and colors as values,
-    sparsity : ChannelSparsity or None, default: None
+    sparsity:  ChannelSparsity or None, default: None
         Optional ChannelSparsity to apply.
         If SortingAnalyzer is already sparse, the argument is ignored
+    subwidget_kwargs: dict or None, default: None
```
alejoe91, line 34:
```suggestion
    subwidget_kwargs : dict or None, default: None
```

src/spikeinterface/widgets/unit_summary.py
```diff
@@ -17,19 +18,27 @@ class UnitSummaryWidget(BaseWidget):
     """
     Plot a unit summary.
 
-    If amplitudes are alreday computed they are displayed.
+    If amplitudes are alreday computed, they are displayed.
 
     Parameters
     ----------
-    sorting_analyzer : SortingAnalyzer
+    sorting_analyzer: SortingAnalyzer
```
alejoe91, line 25: @florian6973 the space is needed for the docs and to comply with numpydocs. Can you re-add the space before the : to all arguments?

alejoe91 (changes requested):
Add spae before colon

src/spikeinterface/widgets/unit_summary.py
alejoe91: done in last commit

alejoe91 (approved):

florian6973:
Oh so sorry! Thank you for added them again

zm711 (approved):
Thanks Florent!

PR #3250, opened 2024-07-24
Sc2 fixes

yger:
Some patches post release for SC2


src/spikeinterface/sortingcomponents/clustering/circus.py
```diff
@@ -152,9 +153,9 @@ def main_function(cls, recording, peaks, params):
                 sub_data = sub_data.reshape(len(sub_data), -1)
 
                 if all_pc_data.shape[1] > params["n_svd"][1]:
-                    tsvd = TruncatedSVD(params["n_svd"][1])
+                    tsvd = PCA(params["n_svd"][1], whiten=True)
```
zm711, line 156: For my own learning why should you whiten here if you've already whitened the recording above?

yger: The recording has been whiten, but it can not hurt to rewhiten the individual waveforms snippets that have been collected before clustering.

zm711: I wasn't sure if this was something similar to if you double filter a recording you actually change the order of the filter, which may not necessarily be a bad thing. thanks Pierre.

alejoe91:
@yger I have a request for SC2: could you add a `gather_mode` param to the `find_spikes_from_templates` so that we don't run out of me for very long recordings?

yger:
Arg yes, I never had the issues, but indeed this could be important. I'll do that

samuelgarcia (approved):

Issue #3257, opened 2024-07-25
SortingAnalyzer extension names "pca" vs "principal_components"


jonahpearl:
The docs appear to have a small bug, in the bottom snippet on [this page](https://spikeinterface.readthedocs.io/en/stable/modules/qualitymetrics.html#quality-metrics-module), there is `sorting_analyzer.compute(input={"pca": dict(...)...` but using "pca" raises a [ValueError](https://github.com/SpikeInterface/spikeinterface/blob/0a60f7c37b3a0be7cafff46c32d30e828dc5f3a9/src/spikeinterface/core/sortinganalyzer.py#L1633), and it looks like the correct name is "principal_components", based on [here](https://github.com/SpikeInterface/spikeinterface/blob/0a60f7c37b3a0be7cafff46c32d30e828dc5f3a9/src/spikeinterface/core/sortinganalyzer.py#L2124). 

Thanks!

zm711:
Eagle eye :) 

Any interest in submitting the patch :) you're right it is `"principal_components"`. If not we will open a PR shortly.

jonahpearl:
Sure, which branch should I submit the PR to?

zm711:
Thanks for asking. Our preference is for you to always make a branch on your fork to allow us to push changes if necessary and then submit the PR to main on the repo!

PR #3258, opened 2024-07-25
fix name of principal_components ext in docs

jonahpearl:
Addresses #3257


zm711:
Same thing here as the other. You have the stray commit from the other PR on this. If you remove that stray commit then it is easier to just merge a doc update. Want to fix that then we can trigger the workflows :)

PR #3270, opened 2024-07-29
Protect median against nans in get_prototype_spike

alejoe91:
as in title! nan median is needed in case some waveforms are flat!

### DIFF
--- a/src/spikeinterface/sortingcomponents/tools.py
+++ b/src/spikeinterface/sortingcomponents/tools.py
@@ -70,10 +70,10 @@ def extract_waveform_at_max_channel(rec, peaks, ms_before=0.5, ms_after=1.5, **j
 
 
 def get_prototype_spike(recording, peaks, ms_before=0.5, ms_after=0.5, nb_peaks=1000, **job_kwargs):
+    from spikeinterface.sortingcomponents.peak_selection import select_peaks
+
     nbefore = int(ms_before * recording.sampling_frequency / 1000.0)
     nafter = int(ms_after * recording.sampling_frequency / 1000.0)
-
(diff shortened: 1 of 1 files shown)


h-mayorquin (commented):
I have always wondering if there is a performance penalty for using these safe functions.

alejoe91:
@h-mayorquin seems very minor, and it solves the issue!

alejoe91:
Nothing seems wrong with the data honestly. Maybe the fix is not working as it should? Do we have a proper test? Note that I was not getting any failure, just all estimated spike locations were 0

yger (approved):

yger:
No proper tests of course, but I'll have a look. Meanwhile, the fix can not hurt and indeed. However before the fix, at least there was messages with numpywarning during the estimation of the prototype if nan/0 were found. In your case, before the fix, you saw such messages before getting your 0 positions?

alejoe91:
Nope, because it's protected by the `with np.errstate(divide="ignore", invalid="ignore"):` ...

h-mayorquin:
> @h-mayorquin seems very minor, and it solves the issue!

Yeah, sorry, I was not trying to argue against just wondering for myself. I agree that performant code that is buggy is worse than ultra-slow-code that is sound. I was just taking the opportunity to see if you guys knew. I feel that is good to be aware of these things.

A reading of the source indicates that it doubles (or triples?) the memory cost:
https://github.com/numpy/numpy/blob/v2.0.0/numpy/lib/_nanfunctions_impl.py#L968-L1071

The strategy for both nanmean and mean seem to sum and count so I expect the most costly loop to be the same.

PR #3276, opened 2024-07-31
Expose save preprocessing in ks4

JoeZiminski:
This PR edits `kilosort4.py` to expose the `save_preprocessed_copy` parameter. Now, `temp_wh.dat` is saveed to the sorting output and `params.py` is edited to point `phy` to the `temp_wh.dat.` @carlacodes if you have some time maybe you could check off this fork/branch this is behaving as expected? I tested with the below code:

<details><summary> Test code </summary>

```
import numpy as np

recording, _, sorting = si.generate_drifting_recording(
    duration=10.0,
    generate_displacement_vector_kwargs=dict(
        displacement_sampling_frequency=5.0,
        drift_start_um=[0, 20],
        drift_stop_um=[0, -20],
        drift_step_um=1,
        motion_list=[
            dict(
                drift_mode="zigzag",
                non_rigid_gradient=None,
                t_start_drift=None,
                t_end_drift=None,
                period_s=200,
            ),
        ],
    ),
)

si.run_sorter("kilosort4", recording, save_preprocessed_copy=True)
```

</details>

At present this is not tested in #3085 but that can be extended after this PR is merged. 

More generally @alejoe91, at present `kilosort4.py` replicates the body of `run_sorter.py` but as versions progress, this might prove difficult to maintain and requires a lot of testing e.g. #3085 to check nothing was missed. I wonder if long-term (assuming I am correct in thinking this is done to skip: preprocessing and drift correction), we can ask / work with KS4 team to expose options to skip preprocessing if they are interested. Then, we could just wrap the `run_kilosort` function directly and testing would just be checking the kwargs are passed properly (rather than having to check the outputs). Maybe there are other considerations I have overlooked though, WDYT?

### DIFF
--- a/src/spikeinterface/sorters/external/kilosort4.py
+++ b/src/spikeinterface/sorters/external/kilosort4.py
@@ -217,7 +217,7 @@ class Kilosort4Sorter(BaseSorter):
                 do_CAR,
                 invert_sign,
                 device,
-                save_preprocesed_copy=save_preprocessed_copy,
+                save_preprocesed_copy=save_preprocessed_copy,  # this kwarg is correct (typo)
             )
             n_chan_bin, fs, NT, nt, twav_min, chan_map, dtype, do_CAR, invert, _, _, tmin, tmax, artifact, _, _ = (
                 get_run_parameters(ops)
--- a/src/spikeinterface/sorters/external/kilosort4.py
+++ b/src/spikeinterface/sorters/external/kilosort4.py
@@ -99,7 +99,7 @@ class Kilosort4Sorter(BaseSorter):
         "save_extra_kwargs": "If True, additional kwargs are saved to the output",
         "skip_kilosort_preprocessing": "Can optionally skip the internal kilosort preprocessing",
         "scaleproc": "int16 scaling of whitened data, if None set to 200.",
-        "save_preprocessed_copy": "save a pre-processed copy of the data (including drift correction) to temp_wh.dat in the results directory",
+        "save_preprocessed_copy": "save a pre-processed copy of the data (including drift correction) to temp_wh.dat in the results directory and format Phy output to use that copy of the data",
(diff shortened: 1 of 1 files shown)


alejoe91:
@JoeZiminski this LGTM. @carlacodes were you able to test this?

carlacodes (approved):
LGTM; tested on my own data with ks 4.0.12

src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -207,7 +210,15 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
...
         if version.parse(cls.get_sorter_version()) >= version.parse("4.0.12"):
-            ops = initialize_ops(settings, probe, recording.get_dtype(), do_CAR, invert_sign, device, False)
+            ops = initialize_ops(
+                settings,
+                probe,
+                recording.get_dtype(),
+                do_CAR,
+                invert_sign,
+                device,
+                save_preprocesed_copy=save_preprocessed_copy,  # this kwarg is correct (typo)
```
zm711, line 220: No way..... I have a feeling is is going to be fixed and then we will need another version just to account for this....

alejoe91: @zm711 you need to reviow KS PRs ahahahah

alejoe91: @JoeZiminski maybe we have to go back to positional arguments here? :)

JoeZiminski: Yes I think so @alejoe91 or the number of conditionals will explode, shall we merge this and then continue the discussion on how best to proceed [here](https://github.com/SpikeInterface/spikeinterface/pull/3316#issuecomment-2298683709)?

alejoe91: Let's do that!

PR #3279, opened 2024-08-01
Proposal for adding Examples to docstrings

zm711:
To increase documentation for users that don't read the docs but instead use `help` or their IDE docstring reader we could begin to add examples to the code base. `read_intan` is the clearest for what I would like examples to be, but I think even beginning to add examples that we flesh out later would be an incremental improvement (for example intan stream_id ='0' is for amplifier data. I could also add an example with how to use name.

What do people think about this in general?


alejoe91:
Thanks @zm711 

I'm totally in favor of this! Anything that can make things easier for users is highly welcome :)

samuelgarcia (approved):

samuelgarcia:
Merci l'ami.

PR #3287, opened 2024-08-06
Refactor `set_property` in base

h-mayorquin:
I added two things around `set_property` and `set_annotations` in `base.py`:
1) Typing to inputs
2) More meaningful errors that describe the nature of the problem to the user.

I started adding tests but then I found a bug that I will report in an issue. I will fix the bug after we merge this.

[EDIT] This is the error that I stumbuled upon https://github.com/SpikeInterface/spikeinterface/issues/3288

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -208,7 +208,7 @@ class BaseExtractor:
     def set_property(
         self,
         key,
-        values: list | np.ndarray | tuple | None,
+        values: list | np.ndarray | tuple,
         ids: list | np.ndarray | tuple | None = None,
         missing_value: Any = None,
     ) -> None:
@@ -236,12 +236,6 @@ class BaseExtractor:
             The missing_value has to be specified for types int and unsigned int.
         """
 
-        # This deletes the values but we have `delete_property` maybe we should eliminate this?
-        if values is None:
-            if key in self._properties:
-                self._properties.pop(key)
-            return
-
         size = self._main_ids.size
         values = np.asarray(values)
         dtype = values.dtype
@@ -267,7 +261,7 @@ class BaseExtractor:
                         if dtype_kind not in self.default_missing_property_values.keys():
                             raise ValueError(
                                 f"Can't infer a natural missing value for dtype {dtype_kind}. "
-                                "Please provide it with the missing_value argument"
+                                "Please provide it with the `missing_value` argument"
                             )
                         else:
                             missing_value = self.default_missing_property_values[dtype_kind]
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -229,12 +229,14 @@ class BaseExtractor:
             Array of values for the property
         ids : list/np.array, default: None
             List of subset of ids to set the values, default: None
+            if None which is the default all the ids are set or changed
         missing_value : object, default: None
             In case the property is set on a subset of values ("ids" not None),
             it specifies the how the missing values should be filled.
             The missing_value has to be specified for types int and unsigned int.
         """
 
+        # This deletes the values but we have `delete_property` maybe we should eliminate this?
         if values is None:
             if key in self._properties:
                 self._properties.pop(key)
@@ -257,16 +259,15 @@ class BaseExtractor:
                 non_unique_ids = [id for id in ids if np.count_nonzero(ids == id) > 1]
                 raise ValueError(f"IDs are not unique: {non_unique_ids}")
 
-            # This branch is used for unit and channel aggregation
+            # Not clear where this branch is used, perhaps on aggregation of extractors?
             if ids.size < size:
                 if key not in self._properties:
 
                     if missing_value is None:
                         if dtype_kind not in self.default_missing_property_values.keys():
-                            raise Exception(
-                                "For values dtypes other than float, string, object or unicode, the missing value "
-                                "cannot be automatically inferred. Please specify it with the 'missing_value' "
-                                "argument."
+                            raise ValueError(
+                                f"Can't infer a natural missing value for dtype {dtype_kind}. "
+                                "Please provide it with the missing_value argument"
                             )
                         else:
                             missing_value = self.default_missing_property_values[dtype_kind]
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/base.py
```diff
@@ -164,7 +164,7 @@ def id_to_index(self, id) -> int:
     def annotate(self, **new_annotations) -> None:
         self._annotations.update(new_annotations)
 
-    def set_annotation(self, annotation_key, value: Any, overwrite=False) -> None:
+    def set_annotation(self, annotation_key: str, value: Any, overwrite=False) -> None:
```
zm711, line 167: So value could be like a lamda function? like

```python
set_annotation(annotation_key = 'my_lambda', value = lamda x)
```

that might be a little pendantic, but it really just comes back to our debate about explicit vs too much info.

src/spikeinterface/core/base.py
```diff
@@ -205,7 +205,13 @@ def get_annotation(self, key, copy: bool = True) -> Any:
     def get_annotation_keys(self) -> List:
         return list(self._annotations.keys())
 
-    def set_property(self, key, values: Sequence, ids: Optional[Sequence] = None, missing_value: Any = None) -> None:
+    def set_property(
+        self,
+        key,
+        values: list | np.ndarray | tuple | None,
```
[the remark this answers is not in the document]
h-mayorquin: We have a branch for that, check the comment I think we should eliminate it!

src/spikeinterface/core/base.py
```diff
@@ -223,12 +229,14 @@ def set_property(self, key, values: Sequence, ids: Optional[Sequence] = None, mi
...
         ids : list/np.array, default: None
             List of subset of ids to set the values, default: None
+            if None which is the default all the ids are set or changed
         missing_value : object, default: None
             In case the property is set on a subset of values ("ids" not None),
             it specifies the how the missing values should be filled.
             The missing_value has to be specified for types int and unsigned int.
         """
 
+        # This deletes the values but we have `delete_property` maybe we should eliminate this?
```
h-mayorquin, line 239: @alejoe91

src/spikeinterface/core/base.py
h-mayorquin: Ah, but I just remember you don't think ArrayLike is very useful, right? Probably less so in this case where you can pass ints, floats and strings that are array like but not tuple, liste or np.ndarray.

I am fine with the way it is now. I think that Aurelien added this so his linter would not look ugly : P

What would you have here? Any suggestions?

zm711: I'm fine with how it is now. I guess for this it was more supporting Sam in that sometimes typing in python is a mess :) 

But for now I think `Any` is fine.

h-mayorquin: I disagree with this assesment. What's the mess here? It is just hard to type but it is also hard to code. We really have not set those restrictions in place any way.

If anything it reveals that we have not constrain our schema of annotations enough.

zm711: To be fair I never use annotations, so my knowledge base here isn't clear enough to be super useful. But since this can't be `Any` I think it is better to explicit:

`str | float | int` or whatever types even if it gets long.

h-mayorquin: You can really put almost anything on the dictionary value. First we need to make a distinction of what we want on it and then find the typing.

I think that you are assuming that we already know what could be there but that's a decision to make. Or maybe I am reading you wrong.

zm711: Nope, I read your comment wrong. I see now what you meant. I could imagine annotations of things like int, float, string, datetime that people would want to annotate. I guess the question is how much of this annotation machinery would duplicate the goal of what NWB does. I mean neo also has an annotation mechanism, but I've never used it.

h-mayorquin: What I have seen until now is datetime, int, str, float, and nested dictionaries and list of those (for probe properties) . I would wait till something breaks so we can dicuss what should not be there : P

zm711: Now that seems like a challenge :P

src/spikeinterface/core/base.py
alejoe91: Thanks @h-mayorquin 

Yeah I think we should remove this weird behavior and suggest to use `delete_property`

src/spikeinterface/core/base.py
```diff
@@ -240,23 +248,26 @@ def set_property(self, key, values: Sequence, ids: Optional[Sequence] = None, mi
...
 
                     if missing_value is None:
                         if dtype_kind not in self.default_missing_property_values.keys():
-                            raise Exception(
-                                "For values dtypes other than float, string, object or unicode, the missing value "
-                                "cannot be automatically inferred. Please specify it with the 'missing_value' "
-                                "argument."
+                            raise ValueError(
+                                f"Can't infer a natural missing value for dtype {dtype_kind}. "
+                                "Please provide it with the missing_value argument"
```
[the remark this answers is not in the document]
h-mayorquin: That make sense.

src/spikeinterface/core/base.py
h-mayorquin: Great! Doing it now.

src/spikeinterface/core/base.py
h-mayorquin: Yeah, I mean, if you put a file stream in annotations and then you try to pickle it should break. Or you can put a large numpy arrray  or something of the sort

src/spikeinterface/core/base.py

src/spikeinterface/core/base.py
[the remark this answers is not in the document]
h-mayorquin: I think I did it. I will do another PR after this to improve some of the error strings anyway after this if I missed any.

src/spikeinterface/core/base.py
h-mayorquin: This is actually used (I don't think that intentionally)

https://github.com/h-mayorquin/spikeinterface/blob/7b37ee3f45abe56debc6f05d2131f3114e94c10f/src/spikeinterface/core/base.py#L1138-L1140

But I would rather address that into another PR.

src/spikeinterface/core/base.py
```diff
@@ -240,23 +247,26 @@ def set_property(self, key, values: Sequence, ids: Optional[Sequence] = None, mi
...
             self._properties[key] = values
         else:
             ids = np.array(ids)
-            assert np.unique(ids).size == ids.size, "'ids' are not unique!"
+            unique_ids = np.unique(ids)
+            if unique_ids.size != ids.size:
+                non_unique_ids = [id for id in ids if np.count_nonzero(ids == id) > 1]
+                raise ValueError(f"IDs are not unique: {non_unique_ids}")
 
+            # Not clear where this branch is used, perhaps on aggregation of extractors?
```
samuelgarcia, line 261: because you could set all ids (like the None) but with the vector.

samuelgarcia (approved):

PR #3289, opened 2024-08-06
Test IBL skip when the setting up the one client fails

h-mayorquin:
Seeing if this works.

### DIFF
--- a/src/spikeinterface/extractors/tests/test_iblextractors.py
+++ b/src/spikeinterface/extractors/tests/test_iblextractors.py
@@ -19,12 +19,15 @@ class TestDefaultIblRecordingExtractorApBand(TestCase):
         from one.api import ONE
 
         cls.eid = EID
-        cls.one = ONE(
-            base_url="https://openalyx.internationalbrainlab.org",
-            password="international",
-            silent=True,
-            cache_dir=None,
-        )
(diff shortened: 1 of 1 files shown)


h-mayorquin:
This seems to be working but I am concerned because the tests are passing for me locally without this. That is, I run:

```python
(work) @h-laptop$ pytest /.../src/spikeinterface/extractors/tests/test_iblextractors.py
```

I wonder why the requests from github are being blocked. I wonder if it is a new policy of ibl?

alejoe91:
@h-mayorquin I think that webclient has some internal caching mechanism. I'm getting this warning on my side: 
```
/home/alessio/anaconda3/envs/si/lib/python3.10/site-packages/one/webclient.py:126: RuntimeWarning: Failed to connect, returning cached response
  warnings.warn('Failed to connect, returning cached response', RuntimeWarning)
```

h-mayorquin:
So you think is passing for me locally because I have something cached?

h-mayorquin:
Then I guess this should work. We should just be weary of the test not every working again : P

alejoe91 (approved):

alejoe91:
Thanks @h-mayorquin 

I really like the smart solution! :)

Issue #3256, opened 2024-07-25
New quality metrics overwrite old ones


jonahpearl:
Hi all — I recently decided to add the `nearest_neighbors` quality metric into my pipeline, but when I tried to compute just that metric, I was annoyed to find that it overwrote all the other previously calculated metrics. This seems like non-optimal behavior — imagine the user computes slow quality metric X, then wants to also add slow quality metric Y the next day, they will also have to re-compute X, or otherwise manually futz with the saved CSVs. 

This behavior seems to be implemented [here](https://github.com/SpikeInterface/spikeinterface/blob/be7ce8d21354863dadb6fd6f06d4010f5da28c1e/src/spikeinterface/qualitymetrics/quality_metric_calculator.py#L137). Instead of creating a new df each time, why not check for an existing one, and merge it with any new metrics created? I understand that overwriting is perhaps a better default, in case the user has curated the units or otherwise changed pre-processing, and is trying to compute metrics de novo, but there could be a "keep_existing" kwarg or something that allows the user to specify not to overwrite what's already there.

Thanks!

zm711:
@chrishalcrow suggested we should fix this at our last meeting. Still in the planning phase of figuring this out. Thanks for the report!

zm711:
I also want to link this discussion which makes a similar request just so that if there is any debate we know that this is a very requested feature

#3251

jakeswann1:
Just adding my wish for this to come true, too!

chrishalcrow:
Yes, I'm keen for this too!

zm711:
I will close this now that #3292 is merged :)

PR #3292, opened 2024-08-07
Do not delete quality and template metrics on recompute

chrishalcrow:
Possible solution to #3256 and some problems in #3251

This PR alters the way quality_metrics are computed, so that ones which are computed are appended to the extension data, rather than replacing it. This is done by using a new initialisation of the quality_metrics extension.

This might be controversial. Some points in its favour:
- If you forget to calculate your favourite quality_metric, you can just calculate as you'd expect.
- If you compute a quality_metric with new parameters, it replaces the old one. So your quality_metrics are always up to date.
- If parents of the quality_metrics extension are recomputed, the quality_metrics are deleted. Hence it is impossible to have some quality_metrics calculated using waveforms with one parameter set and some waveforms with other parameters. So provenance seems intact to me.

Happy to see other solutions to this problem: @samuelgarcia suggested a new method called `append` (or something similar) but I couldn't figure out nice syntax for this.


src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -71,8 +83,14 @@ def _set_params(self, metric_names=None, qm_params=None, peak_sign=None, seed=No
             if "peak_sign" in qm_params_[k] and peak_sign is not None:
                 qm_params_[k]["peak_sign"] = peak_sign
 
+        try:
+            existing_metric_names = self.sorting_analyzer.get_extension("quality_metrics").params.get("metric_names")
+            metric_names_for_params = np.concatenate((existing_metric_names, metric_names))
+        except:
```
zm711, line 89: I don't know if a naked except is the best move. This can hid bugs that become very difficult to track down. Is there a better exception to catch?

chrishalcrow: Yeah, good point. There are possible exceptions: there can be no extension or the list of metric_names can be empty. So I've replaced this with two `if`s which should be more explicit. Whad'ya think?

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -71,8 +84,15 @@ def _set_params(self, metric_names=None, qm_params=None, peak_sign=None, seed=No
             if "peak_sign" in qm_params_[k] and peak_sign is not None:
                 qm_params_[k]["peak_sign"] = peak_sign
 
+        metric_names_for_params = metric_names
+        qm_extension = self.sorting_analyzer.get_extension("quality_metrics")
+        if qm_extension:
```
zm711, line 89:
```suggestion
        if qm_extension is not None:
```

this is more precise right? If qm extension doesn't exist none is returned so checking for None is more precise and will help us keep track of what the code is doing. Right? Otherwise I think the ifs are better :)

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
zm711: See my new comment below!

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
chrishalcrow: Yeah, sure. Thanks!

zm711:
On a slightly more global discussion point this/or a similar strategy would be needed for template metrics as well right? So that if a new metric is implemented someone can just add that to their old template metrics rather than have to re-calculate.

chrishalcrow:
@samuelgarcia Does this look ok? I'll do a similar thing to `template_metrics` shortly.

zm711:
much like adding `fast` to the end of a method a commit of `tests now pass` is always a dangerous gamble :P

chrishalcrow:
> much like adding `fast` to the end of a method a commit of `tests now pass` is always a dangerous gamble :P

I like to live dangerously

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -64,6 +64,8 @@ class ComputeTemplateMetrics(AnalyzerExtension):
         For more on generating a ChannelSparsity, see the `~spikeinterface.compute_sparsity()` function.
     include_multi_channel_metrics : bool, default: False
         Whether to compute multi-channel metrics
+    delete_existing_metrics : bool, default: False
+        If True, deletes any quality_metrics attached to the `sorting_analyzer`
```
JoeZiminski, line 68: Could this state what happens when `False`? (i.e. previously computed are propagated, but will be deleted if kwargs don't match).

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -141,8 +145,29 @@ def _set_params(
             metrics_kwargs_ = _default_function_kwargs.copy()
             metrics_kwargs_.update(metrics_kwargs)
 
+        all_metric_names = metric_names
```
JoeZiminski, line 148: I got a bit confused with `all_metric_names`, `metric_names` and `existing_metric_names`. Could these be renamed to be slightly more informative? I am quite naive to the `template_metrics` / `sorting_analyzer` stuff but am a little confused by what is going on here, could a short comment be added at the top of this section to explain?

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -141,8 +145,29 @@ def _set_params(
...
+                warnings.warn(
+                    "The parameters used to calculate the previous template metrics are different than those used now. Deleting previous template metrics..."
+                )
+                self.sorting_analyzer.get_extension("template_metrics").data["metrics"] = pd.DataFrame(
+                    index=self.sorting_analyzer.unit_ids
+                )
+                self.sorting_analyzer.get_extension("template_metrics").params["metric_names"] = []
+                existing_metric_names = []
+
+            for metric_name in existing_metric_names:
```
JoeZiminski, line 165: Is my understanding correct that the existing metrics are re-computed? they are added to the params dict here, and then the call `_run` -> `_compute_metrics` will re-compute everything? (or, maybe there is some skipping of existing templates in the `compute_metrics` logic I have missed)

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -141,8 +145,29 @@ def _set_params(
...
+        if delete_existing_metrics is False and tm_extension is not None:
+            existing_metric_names = tm_extension.params["metric_names"]
+            existing_params = tm_extension.params["metrics_kwargs"]
+
+            # checks that existing metrics were calculated using the same params
+            if existing_params != metrics_kwargs_:
+                warnings.warn(
+                    "The parameters used to calculate the previous template metrics are different than those used now. Deleting previous template metrics..."
+                )
+                self.sorting_analyzer.get_extension("template_metrics").data["metrics"] = pd.DataFrame(
```
JoeZiminski, line 159: Are this and the next three lines all functioning to completely delete existing metrics? If so this could be a candidate for a new function `self.delete_existing_metrics()`, just so if people want to do this in future there is no risk of forgetting one of these steps and half-deleting the metrics.

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -141,8 +145,29 @@ def _set_params(
...
+                self.sorting_analyzer.get_extension("template_metrics").params["metric_names"] = []
+                existing_metric_names = []
+
+            for metric_name in existing_metric_names:
+                if metric_name not in metric_names:
+                    all_metric_names.append(metric_name)
+
         params = dict(
-            metric_names=[str(name) for name in np.unique(metric_names)],
+            metric_names=[str(name) for name in np.unique(all_metric_names)],
```
JoeZiminski, line 170: Under what circumstances could a metric name not be a string here? Would `list(set(all_metric_names))` suffice?

src/spikeinterface/postprocessing/template_metrics.py
JoeZiminski: I think (if my below understanding is correct) then maybe just `all_metric_names` -> `metric_names_to_compute` could work? (+ the short comment explainer)

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -283,11 +308,20 @@ def _compute_metrics(self, sorting_analyzer, unit_ids=None, verbose=False, **job
                 template_metrics.at[index, metric_name] = value
         return template_metrics
 
-    def _run(self, verbose=False):
+    def _run(self, delete_existing_metrics=False, verbose=False):
         self.data["metrics"] = self._compute_metrics(
             sorting_analyzer=self.sorting_analyzer, unit_ids=None, verbose=verbose
         )
 
+        tm_extension = self.sorting_analyzer.get_extension("template_metrics")
```
JoeZiminski, line 316: This makes me think actually that the previously computed metrics are not recomputed. But if so I have two questions:

1) where is that logic to avoid in `_compute_metrics`?
2) at this stage, is `self.sorting_analzyer.get_extension("template_metrics")` an old version of the template metrics object that is due to be overwritten by the current one, but has not been yet? so we can pass the previously computed metrics from this instantiation to the current one?

If so that's very nice, a couple of lines to this function docstring to explain I think would make this clear.

src/spikeinterface/postprocessing/tests/test_template_metrics.py
```diff
@@ -3,6 +3,36 @@
 import pytest
 
 
+def test_compute_new_template_metrics(small_sorting_analyzer):
```
JoeZiminski, line 6: nice

src/spikeinterface/postprocessing/tests/test_template_metrics.py
```diff
@@ -3,6 +3,36 @@
...
+    """
+    Computes template metrics then computes a subset of quality metrics, and checks
+    that the old quality metrics are not deleted.
+
+    Then computes template metrics with new parameters and checks that old metrics
+    are deleted.
+    """
+
+    small_sorting_analyzer.compute("template_metrics")
```
JoeZiminski, line 15: apologies for my ignorance, what does this call do vs. the one underneath? by default will this compute every template metric? and then `assert "half_width" in list(template_metric_extension.get_data().keys())` tests that they are not deleted, using `half_width` as an example? This all makes sense, however assuming I am correct above it the logic flow would be clearer by either a) only computing `half_width` in the first call or b) testing against all template metrics from the default template metrics name list in the assert.

src/spikeinterface/postprocessing/tests/test_template_metrics.py
```diff
@@ -3,6 +3,36 @@
...
+    assert "half_width" in list(template_metric_extension.get_data().keys())
+
+    # check that, when parameters are changed, the old metrics are deleted
+    small_sorting_analyzer.compute(
+        {"template_metrics": {"metric_names": ["exp_decay"], "metrics_kwargs": {"recovery_window_ms": 0.6}}}
+    )
+
+    template_metric_extension = small_sorting_analyzer.get_extension("template_metrics")
+
+    assert "half_width" not in list(template_metric_extension.get_data().keys())
```
JoeZiminski, line 31: same here might as well check against full default list if it was indeed all computed in the first call

chrishalcrow:
> Hey @chrishalcrow this is looking nice! I am not very familiar with the `sorting_analyzer` and it's extensions so am using this PR as a learning opportunity 😅 apologies for all the questions. It's nice that the quality-metrics / template-metrics solution takes the same approach here, after I understand the template_metrics a bit more I will take a quick look at the quality metrics, play around with it in code and give any last feedback.

Thanks - I had totally messed up the template_metrics. The quality_metrics are a bit simpler, I hope!

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -69,6 +69,9 @@ def compute_num_spikes(sorting_analyzer, unit_ids=None, **kwargs):
     return num_spikes
 
 
+_default_params["num_spikes"] = {}
```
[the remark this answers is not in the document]
chrishalcrow: They are called, fairly implicitly, here

https://github.com/SpikeInterface/spikeinterface/blob/c24015504166b47368875e80390a81a6224f69e8/src/spikeinterface/qualitymetrics/quality_metric_calculator.py#L67

Before, they were passed into a `dict` and it didn't matter if they were `None` or an empty dict. Now, I add any existing params to them, so I'd like an empty dict.

src/spikeinterface/postprocessing/tests/test_template_metrics.py
chrishalcrow: I've added some comments. Without any metric names, it will compute everything. With some names it will recompute those metrics. I was testing to make sure thing weren't being deleted. But yeah, it'd be better to test:
 1) _only_ half width is computed 
2) everything is computed 
3) nothing is deleted with half width is computed
4) everything is deleted when we turn `delete_existing_metric` on
Gimmie a sec...

chrishalcrow: Done!

src/spikeinterface/postprocessing/template_metrics.py
chrishalcrow: I've simplified this, so that we don't need to reset everything here. Before, we needed non-empty results here, which were then appended to when we computed the new metrics. Now, the results dict is initialised at `_compute_metrics`, then the existing results are appended. So we don't need to initialise an empty DataFrame here.

src/spikeinterface/postprocessing/template_metrics.py
chrishalcrow: They were but that was a mistake. Hopefully clearer now.

src/spikeinterface/postprocessing/template_metrics.py
chrishalcrow: Oooh yes yes yes!

chrishalcrow:
> Hey @chrishalcrow this is looking nice! I am not very familiar with the `sorting_analyzer` and it's extensions so am using this PR as a learning opportunity 😅 apologies for all the questions. It's nice that the quality-metrics / template-metrics solution takes the same approach here, after I understand the template_metrics a bit more I will take a quick look at the quality metrics, play around with it in code and give any last feedback.

Hello, hopefully `template_metrics` is simpler and correct now! 

One big bug I missed: you can't pass any directly parameters to `run` (which is what `compute` calls to actually calculate things). See here-ish:
https://github.com/SpikeInterface/spikeinterface/blob/c24015504166b47368875e80390a81a6224f69e8/src/spikeinterface/core/sortinganalyzer.py#L1250
So you'd like `set_params` to just deal with parameters and `run` to just deal with computing, but you can't tell `run` which metrics to calculate. So you've got to sneak them in using a property of the extension class.

I also wasn't meaning to recompute everything but it was tricky to see because the end result was the same and these are pretty fast to compute. Not sure how you could test this.

Don't look at `quality_metrics` until I've made a few more commits!

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -131,6 +136,10 @@ def _set_params(
         if include_multi_channel_metrics:
             metric_names += get_multi_channel_template_metric_names()
 
+        # `run` cannot take parameters, so need to find another way to pass this
+        self.delete_existing_metrics = delete_existing_metrics
+        self.metric_names = metric_names
```
samuelgarcia, line 141: We should not have self attribute because of reload it do not works.
So everything should be either in self.parms or self.data.
In that case lets put this to self.params.

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -141,8 +150,24 @@ def _set_params(
...
+            if existing_params != metrics_kwargs_:
+                warnings.warn(
+                    "The parameters used to calculate the previous template metrics are different than those used now. Deleting previous template metrics..."
+                )
+                tm_extension.params["metric_names"] = []
+                existing_metric_names = []
+            else:
+                existing_metric_names = tm_extension.params["metric_names"]
+
+            metric_names = list(set(existing_metric_names + metric_names))
```
samuelgarcia, line 167: The set is making a reordering.
Maybe this could be usefull for a user to keep the same order of metric_names inside the params no ?

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -284,10 +311,30 @@ def _compute_metrics(self, sorting_analyzer, unit_ids=None, verbose=False, **job
         return template_metrics
 
     def _run(self, verbose=False):
-        self.data["metrics"] = self._compute_metrics(
-            sorting_analyzer=self.sorting_analyzer, unit_ids=None, verbose=verbose
+
+        delete_existing_metrics = self.delete_existing_metrics
```
samuelgarcia, line 315: Lets get it from self.params.

src/spikeinterface/postprocessing/template_metrics.py
chrishalcrow: Yes - the automated curation stuff uses metrics and ordering is important. So let's keep any user-specified order. Done!

alejoe91:
> Hello, response to @samuelgarcia done and `quality_metrics` now updated to look a lot more like what `template_metrics` ended up as.
> 
> One bit of a pain: some metrics are named one thing, like `drift`, but compute several things, like `drift_ptp`, `drift_std` and `drift_mad. It's helpful to know these so I put them in a dict and used it at some point. I've wanted this dict before, so I think it's good to have.
>

@chrishalcrow good point! I think it would be great to keep an internal dict of quality metric name to metric table column names!

src/spikeinterface/postprocessing/template_metrics.py
chrishalcrow: Ok done - we also need to propogate `metrics_to_compute`, which looks a bit more out of place. Alternately we could repeat the code near line 90 of `quality_metric_calculator.py` .

src/spikeinterface/qualitymetrics/quality_metric_list.py
```diff
@@ -53,3 +53,29 @@
     "drift": compute_drift_metrics,
     "sd_ratio": compute_sd_ratio,
 }
+
+# a dict converting the name of the metric for computation to the output of that computation
```
chrishalcrow, line 57: @alejoe91 You are in luck my friend (will figure out how to test this soon...)

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -158,6 +185,7 @@ def _merge_extension_data(
     ):
         import pandas as pd
 
+        metric_names = self.params["metric_names"]
```
alejoe91, line 188: Not `metrics_to_compute`?

alejoe91: ah no this is merge! Ok :)

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -134,7 +165,7 @@ def _compute_metrics(self, sorting_analyzer, unit_ids=None, verbose=False, **job
 
         import pandas as pd
 
-        metrics = pd.DataFrame(index=unit_ids)
+        metrics = pd.DataFrame(index=sorting_analyzer.sorting.unit_ids)
```
alejoe91, line 168:
```suggestion
        metrics = pd.DataFrame(index=sorting_analyzer.unit_ids)
```
A small suggestion, but let's use the strengths of the `SortingAnalyzer` :)

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -139,12 +144,34 @@ def _set_params(
...
             metrics_kwargs_.update(metrics_kwargs)
 
+        metrics_to_compute = metric_names
+        tm_extension = self.sorting_analyzer.get_extension("template_metrics")
+        if delete_existing_metrics is False and tm_extension is not None:
+
+            existing_params = tm_extension.params["metrics_kwargs"]
+            # checks that existing metrics were calculated using the same params
+            if existing_params != metrics_kwargs_:
+                warnings.warn(
```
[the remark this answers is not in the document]
chrishalcrow: In defence of the warning:
Suppose a user is going through a large set of `sorting_analyzer`s that they computed over a long time span. They want to compute a new `template_metric` that they’ve heard about. They know it takes a long time to recompute everything, so they just specify the new metric to compute. Sometimes all the other metrics are deleted! Sometimes they are not!! What is going on?? This warning would have told them. Plus it’s quite rare to get the warning. The default behaviour (compute everything) will never trigger it.

Another option: if the kwargs are different, there is an error telling the user to either use the same kwargs are before or to include all the previous metrics in their compute (quite a complicated error)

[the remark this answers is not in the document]
h-mayorquin: I am not familiar enough with the sorting analyzer workflow to have a useful opinion on this one.  

I think it would be better if you had the parameters (the new and the old) in the argument to give the user a better workflow.

Something to think about as well is whether this will be called in a context where the warning will make no sense at all for the user. Say, somebody wraps a sorting analyzer and keeps recomputing then the user gets a warning that they don't know anything about. For those cases, it is good to have an option to turn off a sure warning. 

But I don't really know how this is used to judge if this last concern is valid or not.

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -290,10 +319,30 @@ def _compute_metrics(self, sorting_analyzer, unit_ids=None, verbose=False, **job
         return template_metrics
 
     def _run(self, verbose=False):
-        self.data["metrics"] = self._compute_metrics(
-            sorting_analyzer=self.sorting_analyzer, unit_ids=None, verbose=verbose
+
+        delete_existing_metrics = self.params["delete_existing_metrics"]
+        metrics_to_compute = self.params["metrics_to_compute"]
```
alejoe91, line 324: @chrishalcrow one last comment. IMO the `metrics_to_compute` is not really a `param`, but rather a run-time thing. Would it make sense to move the logic here?  As a user, seeing that `metrics_to_compute` parameter in the json file would make me feel weird!

src/spikeinterface/postprocessing/template_metrics.py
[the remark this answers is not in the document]
chrishalcrow: Ooh yes, I like the idea of printing the old and new parameters!

src/spikeinterface/postprocessing/template_metrics.py
alejoe91: Yes, what I'm saying is since the metrics to compute matter when it's time to run, why not just move the code here to avoid having to save the `metrics_to_compute` parameter? We have access here to metric names and delete existing, so we have everything we need!

chrishalcrow: I had a go but it’s not so simple. We need to set the `metric_names` in `set_params`, since this is used for the metadata and in other places (like in `get_data`). So the `params` dict needs to know all the metric which are gonna be contained in the extension: everything that has been and will be computed.

So we have access to what’s already been computed, in the previous extensions `params`, and the complete list of metrics in self.params["metric_names"]. Using this, we can’t distinguish between:
1. a metric which the user included in `metric_names` and was already computed. We want to recompute this because the user has explicitly specified it.
2. a metric which the user has _not_ specified in `metric_names` and was already computed. We don’t want to recompute this because it was already computed.

So we can either
- Change the behavior so that we don’t re-compute metrics that the user specifies and exists already (I think this is bad: the user might be re-comuting because a metric was updated/fixed, and this wouldn’t overwrite the old result, even when explicitly included by the user)
- Give `metrics_to_compute` a better name in params. Something like a catchier version of “metrics_computed_on_most_recent_run`?
- Keep it as is.
- Be cleverer than I've been so far???

alejoe91: I see! Makes total sense. Let's leave it as is then and thanks for explaining.

From my comment I didn't consider the option that the user is passing a metric name that is already computed in order to recompute...

alejoe91 (approved):

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -134,7 +165,7 @@ def _compute_metrics(self, sorting_analyzer, unit_ids=None, verbose=False, **job
 
         import pandas as pd
 
-        metrics = pd.DataFrame(index=unit_ids)
+        metrics = pd.DataFrame(index=sorting_analyzer.unit_ids)
```
samuelgarcia, line 168: The unit_ds is super super important, because it allow to compute only a subsets in case of the merging process.
```suggestion
        metrics = pd.DataFrame(index=unit_ids)
```

src/spikeinterface/qualitymetrics/quality_metric_list.py
```diff
@@ -53,3 +53,29 @@
...
+    "sd_ratio": ["sd_ratio"],
+    "isolation_distance": ["isolation_distance"],
+    "l_ratio": ["l_ratio"],
+    "d_prime": ["d_prime"],
+    "nearest_neighbor": ["nn_hit_rate", "nn_miss_rate"],
+    "nn_isolation": ["nn_isolation", "nn_unit_id"],
+    "nn_noise_overlap": ["nn_noise_overlap"],
+    "silhouette": ["silhouette"],
+    "silhouette_full": ["silhouette_full"],
+}
```
samuelgarcia, line 81: We should maybe have a class approch one day...

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
alejoe91: I agree this should be controlled by `unit_ids`, otherwise [this merging snippet](https://github.com/SpikeInterface/spikeinterface/blob/406f99bf7d2f68b68db3e4535ba6e004331de419/src/spikeinterface/qualitymetrics/quality_metric_calculator.py#L131-L133) will not behave as it should

samuelgarcia (approved):

samuelgarcia:
Merci Chris!

PR #3294, opened 2024-08-08
Fix sampling frequency repr

h-mayorquin:
The current repr is rounded to int even it is not which is deceiving. I discussed this with @alejoe91 in person. 

Screenshoot of the new behavior:

 
image

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -96,14 +96,18 @@ class BaseRecording(BaseRecordingSnippets):
     def _repr_header(self):
         num_segments = self.get_num_segments()
         num_channels = self.get_num_channels()
-        sf_hz = self.get_sampling_frequency()
-        sf_khz = sf_hz / 1000
         dtype = self.get_dtype()
 
         total_samples = self.get_total_samples()
         total_duration = self.get_total_duration()
         total_memory_size = self.get_total_memory_size()
-        sampling_frequency_repr = f"{sf_khz:0.1f}kHz" if sf_hz > 10_000.0 else f"{sf_hz:0.1f}Hz"
+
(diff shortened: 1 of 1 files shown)


zm711 (commented):
I'm sorry I'm being dense. So you check if it is not an integer and then display the whole float. But you're going to have some floating point error so I would expect you to still limit the decimal point (even if it was like 4 decimal places). Then in the case the values are integer (for Hz) you still convert to a float instead of having it display as an integer. (for kHz it makes sense to switch to float because of the division issue).

h-mayorquin:
Just to clarify, because I am not sure I am understanding full. The function `value.is_integer()` returns True if you value is float but has no decimal values: `3.000000.is_integer()` returns True.

In other words, the logic there is that if the representation has decimals, then write is as it is. 

Do your concern about floating point error still apply? I just wanted to be sure that we are on the same page.

zm711:
But your checking that it is not an integer (`not value.is_integer()`)? so it would have to have decimal values. Right?

h-mayorquin:
> I read yours too quick. Yes I'm saying if it has decimal values are we sure that all of the decimals are good and not floating point error.

I don't know how we would figure that one out.

 I kind of think it is better to let the user know anyway. In the current representation the decimals are hidden no matter if they are good or bad.

zm711:
Okay. I was just making sure I understand the logic and I think sharing is fine unless we end up with something where I have something like
30,000.0000000015 which for me should be an integer but I'm sure if `is_integer` works for that. But even then I think most people would ignore the extra stuff.

I've said my piece. I support this improvement. :)

samuelgarcia (approved):

samuelgarcia:
merci.

Issue #3295, opened 2024-08-09
BinaryRecordingExtractor opens the binary file, but doesn't close it, preventing later deleting or modifying binary file


cheydrick:
I attempted to programmatically delete a .bin file I had used after performing operations using a BinaryRecordingExtractor object, but was getting an exception saying that the file was in use by another process.

Brief pseudo-code example:

```
# create BinaryRecordingExtractor instance
recording = si.read_binary("file.bin", num_channels=256, sampling_frequency=40000, dtype='int16')
# Perform an operation
recording.save_to_folder(folder='saved', overwrite=True)
# attempting to delete the .bin file fails
os.remove(binfile) 
```

I found that the BinaryRecordingSegment class (used within BinaryRecordingExtractor) opens the binary file when it's instanced, but never closes it.

I moved the file open operation to the get_traces() method, wrapping the relevant code in a 'with open()' statement, which fixed the problem. I will submit a PR shortly.

alejoe91:
@cheydrick have you tried `del recording` before removing the file?

cheydrick:
Ah, that did release it and allowed the file to be deleted.

I had first tried setting `recording = None` figuring that would cause the garbage collector to destruct the instance, but it's not likely it would happen immediately.

I'll close this issue and pick up the discussion in the PR.

cheydrick:
I'm going to re-open this issue to note that deleting the recording object does not release the file if Kilosort 4 is first run.

For example:

```
# create BinaryRecordingExtractor instance
recording = si.read_binary("file.bin", num_channels=256, sampling_frequency=40000, dtype='int16')
# Perform an operation
sorting = ss.run_sorter(sorter_name='kilosort4', recording=recording, output_folder="ks4_output")
# delete the objects, which should release the file
del recording
del sorting
# deleting the .bin file fails
os.remove(binfile) 
```

I can't tell if it's something within the run_sorter() code or within Kilosort itself that's holding on to the file. I haven't yet tried it with other sorters.

zm711:
KS4 makes its own file handle which is likely a different reference than the `recording` reference, so that could be the issue. del would delete the reference from `recording`, but would not necessarily delete the KS4 reference to the same file.

cheydrick:
Perhaps, though when I modified the BinaryRecordingSegment class to open the .bin file only when it needs it, and close right afterwards, the problem went away (#3296). 

[...]

PR #3296, opened 2024-08-09
Moved file-open operation in BinaryRecordingSegment out of constructor

cheydrick:
I recently report issue #3295 (BinaryRecordingExtractor opens the binary file, but doesn't close it, preventing later deleting or modifying binary file).

This PR fixes the problem that I described in that issue. I don't think it has any unwanted side-effects, though hopefully someone more familiar with the bigger picture will point out if it does!


h-mayorquin:
Sup, @cheydrick, hope you are doing well in Texas.

The behavior in your PR is what I wanted to implement but @samuelgarcia pointed out some possible complication that we have not had time to test/experiment.

The problem is that if the file is remote it might take a while to open it (this is what we want to do experiments with). If I remember well @samuelgarcia did some small experiments with its clusters and find some small effect in that direction. I still believe that this effect is a small price to pay for IO security but leaving the file handles open was the compromise we reached.

Can you test the solution that @alejoe91 proposed on the linked issue? (to `del` the recording).
If that does not work, we can provide a method that iterates over all the open files and closes them.

We should do more conclusive experiments to assess the cost of re-opening the file on every call when the files are remote but now that I think about it maybe we could solve this with a flag. We will have to wait till Sam returns from his vacations to discuss it.

cheydrick:
The suggestion that @alejoe91 made worked.

I had first tried setting `recording = None` thinking that would force the cleanup, but using `del recording` worked to release the files.

I closed the issue, but will wait for someone else to close this PR in case further discussion as you described should happen.

Thanks for the replying @alejoe91 and @h-mayorquin!

paolahydra:
> If that does not work, we can provide a method that iterates over all the open files and closes them.

Hi @h-mayorquin, I would be interested in this patch since I am having the same problem!
Thanks,
Paola

samuelgarcia:
Hi.
@h-mayorquin exactly. I am pretty sure that open/close file have a cost on network drive but I do not have number.
So I would not be in favor of this.

For me, when we do `rec = read_something()` then it is natural that the file is opened (ressource locked) and cannot be delete until `del rec`. Otherwise what would be the behavior of `rec.get_trace()` after deletion ?

JoeZiminski:
It is the case that every time a data chunk is accessed the file would need to be reopened and memory mapped? e.g. preprocessing a 1h recording in 2 second chunks, this would result in 1800 reopen-memory map-close cycles? I am not so familiar with the performance costs of this operation but it seems suboptimal.

However I've also got confused trying to figure out how to close the file handle to the recording object in a similar situation. Can this be solved by documentation and API e.g. can we add a `recording.close_file()` method? in this way the file handle can be closed without deleting the entire recording object. Also, we can mention in the docs that extractors will keep the file handle open unless explicitly closed with `close_file()`.

zm711:
I think there is some work on the Neo side for changing how memmapping will work for recording objects, but it has been on the backburner for a while.

h-mayorquin:
> It is the case that every time a data chunk is accessed the file would need to be reopened and memory mapped? e.g. preprocessing a 1h recording in 2 second chunks, this would result in 1800 reopen-memory map-close cycles? I am not so familiar with the performance costs of this operation but it seems suboptimal if it is the case.

Locally they don't matter:

image

samuelgarcia:
I am afraid that on a nextwork drive this will have a high cost but we should make deeper benchmark.

h-mayorquin:
In a dire attempt against everything that is beautiful we will move forward on this one with a flag that switches the behavior for closing on every call to the current one so the user can choose to play in safe or performance mode if they wish

alejoe91:
@cheydrick @h-mayorquin 

we are picking up old PRs :)

Would the explicit `__del__` implemented here https://github.com/SpikeInterface/spikeinterface/commit/857fbbe47f166f1ce52a78fa128d92f8b657688d address the original issue? In that case, should we close this?

cheydrick:
No, it still exhibits the same behavior I described in #3295.

However, if I had come across this today I wouldn't have reported it. I'm not sure why I expected the recording object to not have ownership of the file while the object was still in scope.

I changed my general pattern to something like this:

```
def do_work():
    recording = si.read_binary(binfile, num_channels=256, sampling_frequency=40000, dtype='int16')
    recording.save_to_folder(folder='saved', overwrite=True)

do_work()
os.remove(binfile) 
```

The recording goes out of scope and as expected releases the file. I think the issue/PR can be closed, though I appreciate the efforts!

PR #3299, opened 2024-08-13
Allow quality and template metrics in sortingview's unit table

jonahpearl:
Hi all — I'm wondering about the status of sortingview (see https://github.com/magland/sortingview/issues/233), and how close it is to being a phy replacement. In the meantime, I implemented a small change whereby you can add in any quality and template metrics that have been computed to the sortingview window — this gets it one step closer to being like Phy :) happy to accommodate any suggestions. Thanks!

### DIFF
--- a/src/spikeinterface/widgets/utils_sortingview.py
+++ b/src/spikeinterface/widgets/utils_sortingview.py
@@ -86,8 +86,6 @@ def generate_unit_table_view(analyzer, unit_properties=None, similarity_scores=N
         for prop_name in unit_properties:
 
             # Get property values from correct location
-            # import pdb
-            # pdb.set_trace()
             if prop_name in sorting_props:
                 property_values = sorting.get_property(prop_name)
             elif prop_name in qm_props:
--- a/src/spikeinterface/widgets/utils_sortingview.py
+++ b/src/spikeinterface/widgets/utils_sortingview.py
@@ -52,14 +52,14 @@ def generate_unit_table_view(analyzer, unit_properties=None, similarity_scores=N
     sorting = analyzer.sorting
 
     # Find available unit properties from all sources
-    sorting_props = sorting.get_property_keys()
+    sorting_props = list(sorting.get_property_keys())
     if analyzer.get_extension("quality_metrics") is not None:
-        qm_props = analyzer.get_extension("quality_metrics").get_data().columns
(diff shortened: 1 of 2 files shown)


zm711:
I think you hit the nail on the head with the lack of splitting. Splitting along the amplitude view isn't too bad in my opinion, but the big issue is that people often try to split along the PC views which are more problematic. They typically don't show all the spikes and may in fact be a poor representation of some data (especially as recordings get more and more channels). So deciding how to split in a reproducible more pipeline like structure isn't super clear. We have people that individually split out opto artifact waveforms using phy which I guess is one use, but again if they resorted their data and still had artifacts they would have to manually do everything again. 

I'm wondering how do you specifically use the splitting in Phy? Do you try to do it on the PC? Do you try to confirm with the waveform view after? Do you only do it on the amplitudes?

jonahpearl:
Maybe we can move this to a discussion since it's not really related to the PR — but I normally only split if I see clear evidence in the waveform view of two units (e.g. clearly two sets of waveforms overlapping each other, like in the screenshot I posted in the sortingview issue) or in the amplitude view (e.g. two full bands of spikes with distinct amplitudes, not just a bit of drift). I have certainly used the PC view to then do those splits, but in those cases it's usually pretty clear how to split it. I've been a bit confused trying to track what happens to the split units in Phy so I haven't been able to verify after each one, but when I'm finished, overall the output looks ok, so I think it's working. I haven't ever gone through thinking, oh, those PCs look a little bimodal, maybe I'll split this one.

src/spikeinterface/widgets/utils_sortingview.py
```diff
@@ -56,8 +81,22 @@ def generate_unit_table_view(sorting, unit_properties=None, similarity_scores=No
...
         values = {}
         valid_unit_properties = []
+
+        # Create columns for each property
         for prop_name in unit_properties:
-            property_values = sorting.get_property(prop_name)
+
+            # Get property values from correct location
+            # import pdb
+            # pdb.set_trace()
```
samuelgarcia, line 90:
```suggestion
```

samuelgarcia (approved):

samuelgarcia:
merci beaucoup.
This is cool.

alejoe91:
Tested and working great! Thanks again @jonahpearl

PR #3304, opened 2024-08-14
Adding return type annotations

tabedzki:
Type annotations have been added to a minor subset of functions. The main one I need annotated is `get_event_times`.

Annotated added were typically `np.ndarray`. If `NDArray` should have been used, I am open to changes in the type annotatonos.

Code review questions: 
- [x]  In `auto_merge.py`, `compute_presence_distance` says it returns a `list` but returns `NDArray`
- [x] In `remove_redundant.py`, ` remove_redundant_units` can return BaseSorting, but I would have to import the class for it be recognized. Please advise whether it should
- [ ] In `remove_redundant.py` `find_redundant_units` says it returns two lists but appears to return one list
- [ ] In `sortingview_curation.py`, `# TODO @alessio you remove this after testing`
- [x] In `amplitude_scalings.py`, `find_collisions` docstring says it returns a `np.array` but returns a `dict`
- [x] In `misc_metrics.py`, `get_synchrony_counts` says it returns `dict` but returns `np.ndarray`
- [x] In `comparisontools.py`, `make_match_count_matrix` indicates that it would return an `ndarray` but instead returns a `pd.DataFrame`. Should the documentation be updated?
- [x]  In `comparisontools.py`, `make_agreement_scores` indicates that it would return an `ndarray` but instead returns a `pd.DataFrame`. Should the documentation be updated?
- [x] Should we add `import pandas as pd` to the top instead of including it within each function?
- [x] In `comparisontools.py`, `make_possible_match` indicates it would return `tuple[pd.Series, pd.Series]` but returns `tuple[dict,dict]`
- [x] In `core_tools.py`, `set_value_in_extractor_dict` documentation says that the function is both an inplace modification but also returns a dict


zm711:
We use the `from __future__ import annotations` which actually allows us to use typing from any version. You'll notice in many places we do the `|` which only became available in 3.10. So we actually prefer `list` over `List` and we are working on cleaning up the old style typing in the places it is still present. 

I think most people are fine with np.ndarray. Sometimes people put np.array so we are little lax about that, but I think ndarry is accurate enough. Most people on the team are cool with typing, except Sam. I actually think the return typing is important for some of the functions that return multiple arrays since most of our functions return only one thing! So thanks on my end. 

As far as coverage, it is not strictly enforced here. If you're adding a new feature there have to be tests, but no one is micromanaging coverage changing by 0.5% or something like that!

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -8,7 +8,7 @@
 import numpy as np
 
 
-def count_matching_events(times1, times2, delta=10):
+def count_matching_events(times1: list, times2: list, delta=10):
```
zm711, line 11: Is list enforced here. We've been having this debate recently to find a better name, but most of the times these functions can take in:

```
list | np.ndarray | tuple
```
So unless this can only take a list I would probably do all three.

zm711, line 11: Separate comment if you are adding typing you want to add one for delta as well. Is that only int or can it take float? If it is a float I would change default to 10.0 for us.

zm711 (commented):
Couple clarifications.

tabedzki:
I ignored any "Returns" within the class docstrings. If VSCode was able to detect any types, I generally left the function unannotated. If there was something that seemed like it benefited from not being annotated, I didn't annotate (example: `get_pca_models` in `principal_components.py`).

If it was unable to detect (`Any`), then I added what the docstring said it returns. There were some conflicts between the docstring and the actual value. Here is a compounded list

- In `auto_merge.py`, `compute_presence_distance` says it returns a `list` but returns `NDArray`
- In `remove_redundant.py`, ` remove_redundant_units` can return BaseSorting, but I would have to import the class for it be recognized. Please advise whether it should
- In `remove_redundant.py` `find_redundant_units` says it returns two lists but appears to return one list
- In `sortingview_curation.py`, `# TODO @alessio you remove this after testing`
- In `amplitude_scalings.py`, `find_collisions` docstring says it returns a `np.array` but returns a `dict`
- In `misc_metrics.py`, `get_synchrony_counts` says it returns `dict` but returns `np.ndarray`

zm711:
If you can't use the right type hint. The next best is to use the stringified type hint.

so if something returned a `BaseSorting` but you don't have access to the hint (especially if we need to avoid circular imports) you would do:

`"BaseSorting"`.

The type linters aren't perfectly happy about this but it is an accepted python practice.

I think most of your fixes are fine! I forget if numpydoc only works with parameters or if will also check for the return value of the docstring. But we definitely need those fixed.

tabedzki:
Thanks Zach for the suggestion. If the editor detected something like `np.array | Any` for the autocomplete, I just ended up treating it as "detected".  Additionally, when it was multiple optional returns or multiple possible returns (`np.array | np.NDArray | tuple[np.array  | np.NDArray]` etc) I left it alone since I believed the autocomplete would still detect the proper type in the end.

tabedzki:
@zm711  thanks for mentioning it. Resolved. 
@alejoe91  Is this amount of notations good or are you seeking more return type annotations?

alejoe91:
> @zm711 thanks for mentioning it. Resolved. @alejoe91 Is this amount of notations good or are you seeking more return type annotations?

Looks great to me!

tabedzki:
@alejoe91  thanks for catching that. I didn't realize that was what was causing it to fail.

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -514,7 +514,7 @@ def compute_sliding_rp_violations(
 )
 
 
-def get_synchrony_counts(spikes, synchrony_sizes, all_unit_ids):
+def get_synchrony_count(spikes, synchrony_sizes, all_unit_ids):
```
alejoe91, line 517: can you revert this? I think the naming is correct because it returns multiple counts for different synchrony sizes. (plus all tests are failing!!!)
```suggestion
def get_synchrony_counts(spikes, synchrony_sizes, all_unit_ids):
```

zm711:
>  Should we add import pandas as pd to the top instead of including it within each function?

This is because pandas is an optional dependency. Having it at the top would cause the code to crash if the person doesn't have it.

>  In remove_redundant.py,  remove_redundant_units can return BaseSorting, but I would have to import the class for it be recognized. Please advise whether it should

Use "BaseSorting"

I think for the others where the type is wrong it would be right to switch although in those case you might want to talk to @chrishalcrow since as part of his work for numpydoc compliance he is trying to fix a lot of those. I think it is fine to wrap this up (the bigger we make this PR the more likely you will keep getting hit with merge conflicts!).  Maybe @alejoe91 has a different opinion but I think a few more smaller PRs with changes makes everything easier to merge for this type of change since so many of these modules are being actively worked on.

chrishalcrow:
> I think for the others where the type is wrong it would be right to switch although in those case you might want to talk to @chrishalcrow since as part of his work for numpydoc compliance he is trying to fix a lot of those.

In places it's wrong, please do feel free to update! I've been slowly going through docstrings but not made it to checking types systematically yet.

tabedzki (commented):
Okay, I've updated to cross a few things off of the code review. There are still a few things:

> - In `remove_redundant.py`, `find_redundant_units` says it returns two lists but appears to return one list

I looked at it again, and it returns a list of a 2 member list. Is there a reason that this is a list of list, or can it be a list of a tuple?

> In `core_tools.py`, `set_value_in_extractor_dict` documentation says that the function is both an inplace modification but also returns a dict

I've modified the code to match the docstring and so now it both modifies in place and also returns the modified dict. I think this was a wrong decision and instead the docstring should've been modified to state it does not return anything. Should I make this change?

src/spikeinterface/comparison/comparisontools.py
```diff
@@ -3,12 +3,13 @@
...
 
 from __future__ import annotations
+from typing import Iterable
 
 
 import numpy as np
 
 
-def count_matching_events(times1, times2, delta=10):
+def count_matching_events(times1: Iterable, times2: Iterable, delta: int = 10):
```
zm711, line 12: As @h-mayorquin has pointed out to many many many times (because I make the same mistake), is that `str` is also an `Iterable` so in this case this type hint would be too permissive. We've discussed this before that it `ArrayLike` is probably the best type, but I personally don't like it because I don't think it is beginner friendly. I think this is an edge case, could you remind me based on your reading of this code: is this function public or private? If private I would probably just give it a type of np.ndarray or list since that is likely what it users internally. If public then this is a debate between doing 
```
list | np.ndarray | tuple
```
vs
```
ArrayLike
```
which I think actually comes from numpy.typing.

tabedzki: Thanks for pointing that out! I didn't even consider the overly permissiveness of `Iterable`. Since this function is private (can't find a reference to it on readthedocs), I will go ahead with just giving it a np.ndarray. 

This function annotation is something that I put since i was starting to annotate everything before realizing that annotating the return types is more useful than the inputs. I can also just remove the hint since it is not necessary.

src/spikeinterface/core/core_tools.py
```diff
@@ -258,6 +258,7 @@ def set_value_in_extractor_dict(extractor_dict: dict, access_path: tuple, new_va
     for key in access_path[:-1]:
         current = current[key]
     current[access_path[-1]] = new_value
+    return current
```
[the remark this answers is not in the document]
alejoe91: this is a question for @h-mayorquin :)

src/spikeinterface/comparison/comparisontools.py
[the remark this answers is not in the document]
alejoe91: yes, I think it should :)

src/spikeinterface/comparison/comparisontools.py
alejoe91: @tabedzki has this been updated?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1122,7 +1122,7 @@ def get_num_units(self) -> int:
         return self.sorting.get_num_units()
 
     ## extensions zone
-    def compute(self, input, save=True, extension_params=None, verbose=False, **kwargs):
+    def compute(self, input, save=True, extension_params=None, verbose=False, **kwargs) -> "SortingAnalyzer | None":
```
alejoe91, line 1125:
```suggestion
    def compute(self, input, save=True, extension_params=None, verbose=False, **kwargs) -> "AnalyzerExtension | None":
```

src/spikeinterface/curation/remove_excess_spikes.py
```diff
@@ -85,7 +85,7 @@ def get_unit_spike_train(
         return spike_train[min_spike:max_spike]
 
 
-def remove_excess_spikes(sorting, recording):
+def remove_excess_spikes(sorting: BaseSorting, recording):
```
alejoe91, line 88:
```suggestion
def remove_excess_spikes(sorting: BaseSorting, recording: BaseRecording):
```

alejoe91 (changes requested):
@tabedzki thank you so much for this!

A couple of suggestions

src/spikeinterface/comparison/comparisontools.py
tabedzki: Has been now to reflect zm711's suggestion for simple internal type hint

src/spikeinterface/comparison/comparisontools.py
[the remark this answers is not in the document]
tabedzki: Great, the documentation now reflects `pd.DataFrame`

src/spikeinterface/core/core_tools.py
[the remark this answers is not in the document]
h-mayorquin: My bad, the docstring is wrong there. The return value should not be added and if you can please correct the docstring.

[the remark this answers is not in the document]
tabedzki: Okay sounds good. I'll remove `return current` and keep the fact that this is inplace.

tabedzki:
The last two things that I request feedback on before I think this is good to merge:
- [ ]  the return value of `find_redundant_units` in `remove_redundant.py`. It currently returns a list of a two member list. 
  - Should we update that to be a list of 2 member tuples? Is there a reason why it is current a list of lists?
- [ ] In `sortingview_curation.py`, there is `# TODO @alessio you remove this after testing`. I have not annotated this function since it would require importing `BaseSorting` since even in quotes, Pylance does not recognize it. 

---

A closing thought: while I have gone through and done the return values, I think there are instances where defining input type hints can be useful as the input type definitions are sometimes implicitly used in the return type annotation. For example, if we annotate `CurationSorting` to have `"BaseSorting"` as part of the input in `def __init__(self, sorting: "BaseSorting", make_graph=False, properties_policy="keep"):`, then the return values of `.sorting()` and `.current_sorting()` are automatically provided instead of `Any`. The importance of this may not be high as these are internal functions rather than external functions.

h-mayorquin:
> The last two things that I request feedback on before I think this is good to merge:
> 
> * [ ]   the return value of `find_redundant_units` in `remove_redundant.py`. It currently returns a list of a two member list.
>   
>   * Should we update that to be a list of 2 member tuples? Is there a reason why it is current a list of lists?
> * [ ]  In `sortingview_curation.py`, there is `# TODO @alessio you remove this after testing`. I have not annotated this function since it would require importing `BaseSorting` since even in quotes, Pylance does not recognize it.
> 
> A closing thought: while I have gone through and done the return values, I think there are instances where defining input type hints can be useful as the input type definitions are sometimes implicitly used in the return type annotation. For example, if we annotate `CurationSorting` to have `"BaseSorting"` as part of the input in `def __init__(self, sorting: "BaseSorting", make_graph=False, properties_policy="keep"):`, then the return values of `.sorting()` and `.current_sorting()` are automatically provided instead of `Any`. The importance of this may not be high as these are internal functions rather than external functions.

Maybe we should merge this PR here. Otherwise, conflicts might accrue as this affects quite a lot of files already.

zm711:
> Maybe we should merge this PR here. Otherwise, conflicts might accrue as this affects quite a lot of files already.

+1. I think iterative changes will reduce risk of conflicts continually popping up.

tabedzki:
Agreed about merging this as soon as possible. We should make a new issue about type hints for the input values. 

Should the questions about the tuples and the removal of the legacy command be in a new issue as well or can we address them here?

zm711:
I would probably do the list/tuple later. It seems pretty minor and I'm not sure why it was done that way. But we can solve that later. And its hard to know why pylance isn't working. Could be a pylance thing that it is not recognizing the "BaseSorting". I don't think it is worth importing the class just to get a type hint so I would be against that so I think it would be fine to discuss that one later too.

Alessio will make the final call so if he really wants those two things addressed he will let us know soon.

alejoe91 (approved):

alejoe91:
Agreed! @tabedzki do you mind opening an issue with the missing items so we keep in mind?

PR #3312, opened 2024-08-18
Add dtype in load_waveforms and analyzer.is_filtered()

alejoe91:
A couple of small fixes to ensure back-compatibility with old waveform folders.

Also added th `sorting_analyzer.is_filtered()` to make the export to phy fulli recordingless.


src/spikeinterface/core/waveforms_extractor_backwards_compatibility.py
```diff
@@ -446,6 +446,10 @@ def _read_old_waveforms_extractor_binary(folder, sorting):
     else:
         rec_attributes["probegroup"] = None
 
+    if "dtype" not in rec_attributes:
+        warnings.warn("dtype not found in rec_attributes. Setting to float32")
+        rec_attributes["dtype"] = "float32"
+
```
zm711, line 452: My only question is how often would things be float32? Is the assumption people would have run preprocessing so the majority would have already been converted at some step? I'm just wondering if there is a way to know? 

Or maybe this isn't really used and so the value doesn't really matter?

I'm not sure. Just curious how you're thinking about this.

alejoe91: Thanks @zm711 

I think this is only used when setting a temporary recording. Maybe it would be better to not set it and skip the check if non-existing? Then it can be set from the temporary recording object in memory!

alejoe91: @zm711 changed strategy to improve and add options to the `do_recording_attributes_match`. The only real need in terms of back-compatibility is to only test for `dtype` if the `rec_attributes` have it, which was not the case for old waveform extractors loaded as analyzers

src/spikeinterface/core/recording_tools.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import annotations
 from copy import deepcopy
-from typing import Literal
+from typing import Literal, Tuple
```
zm711, line 3:
```suggestion
from typing import Literal
```

src/spikeinterface/core/recording_tools.py
```diff
@@ -929,7 +929,9 @@ def get_rec_attributes(recording):
     return rec_attributes
 
 
-def do_recording_attributes_match(recording1, recording2_attributes) -> bool:
+def do_recording_attributes_match(
+    recording1: "BaseRecording", recording2_attributes: bool, check_is_filtered: bool = True, check_dtype: bool = True
+) -> Tuple[bool, str]:
```
zm711, line 934:
```suggestion
) -> tuple[bool, str]:
```

I think we want to keep to the builtins going forward no?

src/spikeinterface/core/recording_tools.py
```diff
@@ -939,22 +941,52 @@ def do_recording_attributes_match(recording1, recording2_attributes) -> bool:
...
+        If True, check if the recordings are filtered
+    check_dtype : bool, default: True
+        If True, check if the recordings have the same dtype
 
     Returns
     -------
     bool
         True if the recordings have the same attributes
+    str
+        A string with the an exception message with attributes that do not match
```
zm711, line 954:
```suggestion
        A string with the exception message with the attributes that do not match
```

src/spikeinterface/exporters/to_phy.py
```diff
@@ -167,7 +167,7 @@ def export_to_phy(
         f.write(f"dtype = '{dtype_str}'\n")
         f.write(f"offset = 0\n")
         f.write(f"sample_rate = {fs}\n")
-        f.write(f"hp_filtered = {sorting_analyzer.recording.is_filtered()}")
+        f.write(f"hp_filtered = {sorting_analyzer.is_filtered()}")
```
zm711, line 170: I know we need this for compatibility, but I'm pretty sure Phy has stopped checking the filter status ages ago. KS just always wrote this so we have to keep it around. The cost of tech debt...

src/spikeinterface/core/recording_tools.py
```diff
@@ -939,22 +941,52 @@ def do_recording_attributes_match(recording1, recording2_attributes) -> bool:
...
+        if not recording1_attributes["is_filtered"] == recording2_attributes["is_filtered"]:
+            attributes_match = False
+            non_matching_attrs.append("is_filtered")
+    # dtype is optional
+    if "dtype" in recording1_attributes and "dtype" in recording2_attributes:
+        if check_dtype:
+            if not recording1_attributes["dtype"] == recording2_attributes["dtype"]:
+                attributes_match = False
+                non_matching_attrs.append("dtype")
+
```
zm711, line 986: This works, but if you want to reduce the boiler plate you could rewrite this so that you just check
```
if len(non_matching_attrs) > 0:
    attributes_match = False
```

Then you can delete all that repeated attributes_match = False.

zm711 (commented):
Just one suggestion (for debate), one suggestion (for style) and then a typo fix :)

src/spikeinterface/core/recording_tools.py
alejoe91: done

src/spikeinterface/core/recording_tools.py
alejoe91: done

src/spikeinterface/core/recording_tools.py
alejoe91: done

src/spikeinterface/core/recording_tools.py
```diff
@@ -939,22 +941,52 @@ def do_recording_attributes_match(recording1, recording2_attributes) -> bool:
...
-        and recording1_attributes["num_channels"] == recording2_attributes["num_channels"]
-        and recording1_attributes["num_samples"] == recording2_attributes["num_samples"]
-        and recording1_attributes["is_filtered"] == recording2_attributes["is_filtered"]
-        and recording1_attributes["dtype"] == recording2_attributes["dtype"]
-    )
+    attributes_match = True
+    non_matching_attrs = []
+
+    if not np.array_equal(recording1_attributes["channel_ids"], recording2_attributes["channel_ids"]):
+        attributes_match = False
```
zm711, line 965:
```suggestion
```

src/spikeinterface/core/recording_tools.py
```diff
@@ -939,22 +941,52 @@ def do_recording_attributes_match(recording1, recording2_attributes) -> bool:
...
+        non_matching_attrs.append("channel_ids")
+    if not recording1_attributes["sampling_frequency"] == recording2_attributes["sampling_frequency"]:
+        attributes_match = False
```
zm711, line 968:
```suggestion
```

src/spikeinterface/core/recording_tools.py
```diff
@@ -939,22 +941,52 @@ def do_recording_attributes_match(recording1, recording2_attributes) -> bool:
...
+        non_matching_attrs.append("sampling_frequency")
+    if not recording1_attributes["num_channels"] == recording2_attributes["num_channels"]:
+        attributes_match = False
```
zm711, line 971:
```suggestion
```

src/spikeinterface/core/recording_tools.py
```diff
@@ -939,22 +941,52 @@ def do_recording_attributes_match(recording1, recording2_attributes) -> bool:
...
+        non_matching_attrs.append("num_channels")
+    if not recording1_attributes["num_samples"] == recording2_attributes["num_samples"]:
+        attributes_match = False
```
zm711, line 974:
```suggestion
```

src/spikeinterface/core/recording_tools.py
```diff
@@ -939,22 +941,52 @@ def do_recording_attributes_match(recording1, recording2_attributes) -> bool:
...
+        non_matching_attrs.append("num_samples")
+    if check_is_filtered:
+        if not recording1_attributes["is_filtered"] == recording2_attributes["is_filtered"]:
+            attributes_match = False
```
zm711, line 978:
```suggestion
```

src/spikeinterface/core/recording_tools.py
```diff
@@ -939,22 +941,52 @@ def do_recording_attributes_match(recording1, recording2_attributes) -> bool:
...
+            non_matching_attrs.append("is_filtered")
+    # dtype is optional
+    if "dtype" in recording1_attributes and "dtype" in recording2_attributes:
+        if check_dtype:
+            if not recording1_attributes["dtype"] == recording2_attributes["dtype"]:
+                attributes_match = False
```
zm711, line 984:
```suggestion
```

zm711 (commented):
These can all be removed now that you have the one global check at the bottom.

src/spikeinterface/core/recording_tools.py
alejoe91: done in last commit :)

samuelgarcia (approved):

PR #3314, opened 2024-08-19
Enable cloud-loading for analyzer Zarr

alejoe91:
A few modifications needed to load a `SortingAnalzyer` directly from the cloud

### DIFF
--- a/src/spikeinterface/core/core_tools.py
+++ b/src/spikeinterface/core/core_tools.py
@@ -684,3 +684,20 @@ def measure_memory_allocation(measure_in_process: bool = True) -> float:
         memory = mem_info.total - mem_info.available
 
     return memory
+
+
+def is_path_remote(path: str | Path) -> bool:
+    """
+    Returns True if the path is a remote path (e.g., s3:// or gcs://).
+
+    Parameters
+    ----------
+    path : str or Path
+        The path to check.
+
+    Returns
+    -------
+    bool
+        Whether the path is a remote path.
+    """
+    return "s3://" in str(path) or "gcs://" in str(path)
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -23,7 +23,7 @@ from .basesorting import BaseSorting
 
 from .base import load_extractor
 from .recording_tools import check_probe_do_not_overlap, get_rec_attributes, do_recording_attributes_match
-from .core_tools import check_json, retrieve_importing_provenance
+from .core_tools import check_json, retrieve_importing_provenance, is_path_remote
 from .sorting_tools import generate_unit_ids_for_merge_group, _get_ids_after_merging
 from .job_tools import split_job_kwargs
 from .numpyextractors import NumpySorting
@@ -195,6 +195,7 @@ class SortingAnalyzer:
         format=None,
         sparsity=None,
         return_scaled=True,
+        storage_options=None,
     ):
         # very fast init because checks are done in load and create
         self.sorting = sorting
@@ -204,6 +205,7 @@ class SortingAnalyzer:
         self.format = format
         self.sparsity = sparsity
         self.return_scaled = return_scaled
+        self.storage_options = storage_options
         # this is used to store temporary recording
         self._temporary_recording = None
 
@@ -276,17 +278,15 @@ class SortingAnalyzer:
         return sorting_analyzer
 
     @classmethod
-    def load(cls, folder, recording=None, load_extensions=True, format="auto"):
+    def load(cls, folder, recording=None, load_extensions=True, format="auto", storage_options=None):
         """
         Load folder or zarr.
         The recording can be given if the recording location has changed.
         Otherwise the recording is loaded when possible.
         """
-        folder = Path(folder)
-        assert folder.is_dir(), "Waveform folder does not exists"
         if format == "auto":
             # make better assumption and check for auto guess format
-            if folder.suffix == ".zarr":
+            if Path(folder).suffix == ".zarr":
                 format = "zarr"
             else:
                 format = "binary_folder"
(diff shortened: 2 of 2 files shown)


src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -276,30 +278,34 @@ def create(
...
+        if is_path_remote(str(folder)):
+            sorting_analyzer.folder = folder
+            # in this case we only load extensions when needed
+        else:
+            sorting_analyzer.folder = Path(folder)
 
-        if load_extensions:
-            sorting_analyzer.load_all_saved_extension()
+            if load_extensions:
+                sorting_analyzer.load_all_saved_extension()
```
alejoe91, line 308: This can be very slow from a remote storage, so let's load as needed

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -276,30 +278,34 @@ def create(
...
         if format == "binary_folder":
             sorting_analyzer = SortingAnalyzer.load_from_binary_folder(folder, recording=recording)
         elif format == "zarr":
-            sorting_analyzer = SortingAnalyzer.load_from_zarr(folder, recording=recording)
+            sorting_analyzer = SortingAnalyzer.load_from_zarr(
+                folder, recording=recording, storage_options=storage_options
+            )
 
-        sorting_analyzer.folder = folder
+        if is_path_remote(str(folder)):
```
h-mayorquin, line 301: What was this library that you mention that extends pathlib to work with remote assets? Not suggestion to use it but I just remembed think and it would simplify all of this logic.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -552,25 +560,22 @@ def create_zarr(cls, folder, sorting, recording, sparsity, return_scaled, rec_at
...
-        assert folder.is_dir(), f"This folder does not exist {folder}"
-
-        zarr_root = zarr.open(folder, mode="r")
+        zarr_root = zarr.open_consolidated(str(folder), mode="r", storage_options=storage_options)
 
         # load internal sorting in memory
-        # TODO propagate storage_options
         sorting = NumpySorting.from_sorting(
-            ZarrSortingExtractor(folder, zarr_group="sorting"), with_metadata=True, copy_spike_vector=True
+            ZarrSortingExtractor(folder, zarr_group="sorting", storage_options=storage_options),
```
h-mayorquin, line 570: Side question: why not use the ZarrSortingExtractor directly? : O

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -470,7 +476,9 @@ def load_from_binary_folder(cls, folder, recording=None):
     def _get_zarr_root(self, mode="r+"):
         import zarr
 
-        zarr_root = zarr.open(self.folder, mode=mode)
+        if is_path_remote(str(self.folder)):
+            mode = "r"
```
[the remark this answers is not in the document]
h-mayorquin: Doesn't try-except work here? Or is too costly to send two GETs.

src/spikeinterface/core/core_tools.py
```diff
@@ -684,3 +684,20 @@ def measure_memory_allocation(measure_in_process: bool = True) -> float:
...
+    ----------
+    path : str or Path
+        The path to check.
+
+    Returns
+    -------
+    bool
+        Whether the path is a remote path.
+    """
+    return "s3://" in str(path) or "gcs://" in str(path)
```
h-mayorquin, line 703: Another alternative is to check for the store attribute of the object (for example it is not `DirectoryStore`) to determine this:

https://zarr.readthedocs.io/en/stable/api/storage.html#zarr.storage.DirectoryStore

h-mayorquin (approved):
Seems like a working solution. 

Two comments:
1) Maybe this should be handled at the extension level. That is, the extension should be loaded but the data of the extension should not be loaded until you call it. This will make the object available but not the `__init__` costly.
2) I am thinking on whether it makes sense to have the mode (e.g. "r", "r+", etc) as an attribute of the sorting analyzer in general. I think this a well know metaphor and will avoid accidentally ovewriting data that is loaded..

h-mayorquin (changes requested):
Ah, did not see that the tests are failing. Consolidated maybe?

src/spikeinterface/core/core_tools.py
alejoe91: @h-mayorquin this is before the zarr object is instantiated. In particular, we need to avoid casting to `Path` a remote string because it messes up with the formatting!

src/spikeinterface/core/sortinganalyzer.py
alejoe91: `universal-pathlib`[https://pypi.org/project/universal-pathlib/]

src/spikeinterface/core/sortinganalyzer.py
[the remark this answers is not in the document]
alejoe91: Not sure though if an `r+` would fail only if you try to write to it

src/spikeinterface/core/sortinganalyzer.py
alejoe91: Because this loads it in memory as a spike vector and spike-train-based operations will be fast

src/spikeinterface/core/sortinganalyzer.py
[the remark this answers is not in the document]
h-mayorquin: Make and erase an atribute? : P

I think this is fine! Just wondering about your question.

src/spikeinterface/core/sortinganalyzer.py
h-mayorquin: Thanks.

src/spikeinterface/core/sortinganalyzer.py
h-mayorquin: Oh, I thought that zarr sorting already had a spike_vector so it would not be different.

src/spikeinterface/core/core_tools.py
h-mayorquin: In the code that you have the only one that is done before is the one for mode but that can be easily fixed.

The casting to path could be conditione on the file existing in general I think.

```python
folder_path = Path(folder_path) if Path(folder_path).exists() else folder_path
```

But I think this work as well. I don't see how one is better than the other.

h-mayorquin (approved):
LGTM, tests are passing.

src/spikeinterface/core/core_tools.py
samuelgarcia, line 703: maybe ":" should be enought for a path to be remote. No ?

alejoe91: I thought about it, but it could be `C:\\...` :)

PR #3316, opened 2024-08-20
Added bad channels for kilosort >= 4.0.14

### LINKED ISSUE #3310: set_files requires bad_channels positional argument for kilosort v4.0.16
The 'set_files' method in kilosort4 version 4.0.16 now requires a 'bad_channels' positional argument which seems to call an error on line 208 in src/spikeinterface/sorters/external/kilosort4.py (..the issue can be easily mitigated by downgrading kilosort to 4.0.13)

e.g.:
```
Error running kilosort4                                                                                                                
  File "/home/jack/code/ephys_tools/run_si_pipeline.py", line 303, in <module>                                                         
    main(sys.argv[1:])                                                                                                                 
  File "/home/jack/code/ephys_tools/run_si_pipeline.py", line 295, in main                                                             
    sorting = run_spike_sorting(recording_corrected, ks_path, sorter)                                                                  
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                  
  File "/home/jack/code/ephys_tools/run_si_pipeline.py", line 175, in run_spike_sorting                                                
    sorting = si.run_sorter(                                                                                                           
              ^^^^^^^^^^^^^^                                                                                                           
  File "/home/jack/.miniconda3/envs/si/lib/python3.11/site-packages/spikeinterface/sorters/runsorter.py", line 216, in run_sorter      
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                           
  File "/home/jack/.miniconda3/envs/si/lib/python3.11/site-packages/spikeinterface/sorters/runsorter.py", line 276, in run_sorter_local
    SorterClass.run_from_folder(folder, raise_error, verbose)
  File "/home/jack/.miniconda3/envs/si/lib/python3.11/site-packages/spikeinterface/sorters/basesorter.py", line 301, in run_from_folder
    raise SpikeSortingError(                                       
spikeinterface.sorters.utils.misc.SpikeSortingError: Spike sorting error trace:                                                        
  File "/home/jack/.miniconda3/envs/si/lib/python3.11/site-packages/spikeinterface/sorters/basesorter.py", line 261, in run_from_folder
    SorterClass._run_from_folder(sorter_output_folder, sorter_params, verbose)                                                          
  File "/home/jack/.miniconda3/envs/si/lib/python3.11/site-packages/spikeinterface/sorters/external/kilosort4.py", line 208, in _run_from_folder
    filename, data_dir, results_dir, probe = set_files(settings, filename, probe, probe_name, data_dir, results_dir)
                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                   
TypeError: set_files() missing 1 required positional argument: 'bad_channels'                                                           
                                                                                                                                       
Spike sorting failed
```

chrishalcrow:
Hopefully fixes #3310

Kilosort introduced `bad_channels` in 4.0.14; this PR includes it too.

### DIFF
--- a/src/spikeinterface/sorters/external/kilosort4.py
+++ b/src/spikeinterface/sorters/external/kilosort4.py
@@ -57,6 +57,7 @@ class Kilosort4Sorter(BaseSorter):
         "skip_kilosort_preprocessing": False,
         "scaleproc": None,
         "torch_device": "auto",
+        "bad_channels": None,
     }
 
     _params_description = {
@@ -99,6 +100,7 @@ class Kilosort4Sorter(BaseSorter):
         "skip_kilosort_preprocessing": "Can optionally skip the internal kilosort preprocessing",
         "scaleproc": "int16 scaling of whitened data, if None set to 200.",
         "torch_device": "Select the torch device auto/cuda/cpu",
+        "bad_channels": "A list of channel indices (rows in the binary file) that should not be included in sorting. Listing channels here is equivalent to excluding them from the probe dictionary.",
     }
 
     sorter_description = """Kilosort4 is a Python package for spike sorting on GPUs with template matching.
@@ -205,7 +207,17 @@ class Kilosort4Sorter(BaseSorter):
         # NOTE: Also modifies settings in-place
         data_dir = ""
         results_dir = sorter_output_folder
-        filename, data_dir, results_dir, probe = set_files(settings, filename, probe, probe_name, data_dir, results_dir)
+        bad_channels = params["bad_channels"]
+
+        if version.parse(cls.get_sorter_version()) >= version.parse("4.0.14"):
+            filename, data_dir, results_dir, probe = set_files(
+                settings, filename, probe, probe_name, data_dir, results_dir, bad_channels
+            )
+        else:
+            filename, data_dir, results_dir, probe = set_files(
+                settings, filename, probe, probe_name, data_dir, results_dir
+            )
+
         if version.parse(cls.get_sorter_version()) >= version.parse("4.0.12"):
             ops = initialize_ops(settings, probe, recording.get_dtype(), do_CAR, invert_sign, device, False)
             n_chan_bin, fs, NT, nt, twav_min, chan_map, dtype, do_CAR, invert, _, _, tmin, tmax, artifact, _, _ = (


JoeZiminski:
Thanks @chrishalcrow! this looks good, but could we wait for #3276 to be merged? I think it's good to go, there are some other changes required for the newer version e.g. `save_preprocesed_recording` kwarg was renamed, maybe we can take care of all these here? @alejoe91 

Also @alejoe91 I wonder if it's possible to merge #3085 then we can add a few new tests to cover the cases that have arisen since the new version. Maybe we can split these tests such that the lightweight ones that check the kilosort function kwarg orders (e.h. [here](https://github.com/SpikeInterface/spikeinterface/blob/23d2c77533a2bc65791bd6d07eda9b8723133c33/.github/scripts/test_kilosort4_ci.py#L257))  are included in the spikeinterface CI tests, and the heavier tests that check actual outputs are run elsewhere.

Further to the discussion [here](https://github.com/SpikeInterface/spikeinterface/pull/3276#discussion_r1722813491), I'm in two minds about reverting to positional kwargs, we can compensate elsewhere (e.g. testing kwarg order, testing outputs), but there is always the risk of tests diverging over the years. However, maintaining the KS4 tests and adding conditionals for each new version will a) be super messy b) may itself lead to bugs when it becomes very long and hard to maintain. So i'm not sure there is a good solution. For now I'm happy to either use positionals arguments and then test the argument order of the KS4 functions is as expected, but in general I think this wrapper will take a lot of care as the risk of bugs is very high considering pace of KS4 development and how much of the KS4 internals are exposed in the wrapper. In the short-medium term I think we should try and work with the KS4 team to introduce functions that skip preprocessing etc. into the KS4 codebase, if they are happy to. That way we only need to worry about wrapping the top-level function. I wonder what people think about this.

alejoe91:
> Thanks @chrishalcrow! this looks good, but could we wait for #3276 to be merged? I think it's good to go, there are some other changes required for the newer version e.g. `save_preprocesed_recording` kwarg was renamed, maybe we can take care of all these here? @alejoe91
> 
> Also @alejoe91 I wonder if it's possible to merge #3085 then we can add a few new tests to cover the cases that have arisen since the new version. Maybe we can split these tests such that the lightweight ones that check the kilosort function kwarg orders (e.h. [here](https://github.com/SpikeInterface/spikeinterface/blob/23d2c77533a2bc65791bd6d07eda9b8723133c33/.github/scripts/test_kilosort4_ci.py#L257)) are included in the spikeinterface CI tests, and the heavier tests that check actual outputs are run elsewhere.
> 
> Further to the discussion [here](https://github.com/SpikeInterface/spikeinterface/pull/3276#discussion_r1722813491), I'm in two minds about reverting to positional kwargs, we can compensate elsewhere (e.g. testing kwarg order, testing outputs), but there is always the risk of tests diverging over the years. However, maintaining the KS4 tests and adding conditionals for each new version will a) be super messy b) may itself lead to bugs when it becomes very long and hard to maintain. So i'm not sure there is a good solution. For now I'm happy to either use positionals arguments and then test the argument order of the KS4 functions is as expected, but in general I think this wrapper will take a lot of care as the risk of bugs is very high considering pace of KS4 development and how much of the KS4 internals are exposed in the wrapper. In the short-medium term I think we should try and work with the KS4 team to introduce functions that skip preprocessing etc. into the KS4 codebase, if they are happy to. That way we only need to worry about wrapping the top-level function. I wonder what people think about this.

What about dropping support for versions<4.0.15?

alejoe91:
@JoeZiminski goot to merge #3085 first, but see my comment there

JoeZiminski:
Oh yes of course, that's such a good solution, I wish I'd thought of that 😆 

I'm not familiar with the current release cycle and not sure if it's feasible, but would it be possible to:

- merge #3085 (pending my response)
- make a patch release

Then we can say everything SI < 0.101.1 supports kilosort <= 4.0.12 and everything 0.101.1 (and onwards, for the time being) supports KS >= 4.0.14? (or maybe just 4.0.16, depending on what changes were made, if it is requested we can add in some backward compatibility stuff to extend down to 4.0.13 if needed, but I'd imagine most people would be happy to start with 4.0.16).

chrishalcrow:
> (...but I'd imagine most people would be happy to start with 4.0.16).

I agree with this; and support us not-supporting old versions of kilosort 4.0.x.

zm711:
I will also throw my weight behind not supporting old versions of KS4. It seems like they are still actively working on bugs, which means that supporting old versions are just us supporting buggy versions. This CUDA error seems important to users so I would says something like only supporting the minor releases maybe (let's see how they do with minor releases) instead of supporting each point release (maintain only the current point release for any particular minor).

alejoe91:
Ok, to sum up we should:

- [ ] add a check for KS>=4.0.16 and raise an error for older versions. @chrishalcrow can you take care of that?

- [ ]  modify the CI action for testing against old KS versions and only test against newer versions. @JoeZiminski can you do this?

- [ ]  update KS4 docker version to 4.0.16 (I can take care of this!)

Sounds good?

chrishalcrow:
This branch has conflicts after #3276 was merged, and our new solution goes beyond what this PR is about, so I'll close it and start a new PR.

JoeZiminski:
Thanks this sounds good @alejoe91, can we also make a patch release for SI 0.101.1 so it is easy to differentiate when KS < 4.0.16 was supported?

PR #3318, opened 2024-08-20
Lazy loading of zarr timestamps

alejoe91:
This was previously discussed here: https://github.com/SpikeInterface/spikeinterface/pull/2828

To sum up, loading and decompressing zarr timestamps for very long recordings can be quite time consuming, so we want to avoid doing that at init. When fetching the timestamps though, if they ar not a numpy array they are cast and cached as np.arryas, to avoid re-reading and re-decompressing at every call

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -811,10 +811,9 @@ class BaseRecordingSegment(BaseSegment):
 
     def get_times(self):
         if self.time_vector is not None:
-            if isinstance(self.time_vector, np.ndarray):
-                return self.time_vector
-            else:
-                return np.array(self.time_vector)
+            if not isinstance(self.time_vector, np.ndarray):
+                self.time_vector = np.array(self.time_vector)
+            return self.time_vector
         else:
             time_vector = np.arange(self.get_num_samples(), dtype="float64")
             time_vector /= self.sampling_frequency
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/baserecording.py
```diff
@@ -811,10 +811,9 @@ def __init__(self, sampling_frequency=None, t_start=None, time_vector=None):
 
     def get_times(self):
         if self.time_vector is not None:
-            if isinstance(self.time_vector, np.ndarray):
-                return self.time_vector
-            else:
-                return np.array(self.time_vector)
+            if not isinstance(self.time_vector, np.ndarray):
```
h-mayorquin, line 814: I think you can just always call np.asaray() which by default will just pass the data along if it is already and np.ndarray but will create a copy if it is hdf5, zarr or a memmap.

h-mayorquin (commented):
Yes, I think it is better. Zarr is not the format were we want the data to be read when called. 

Can you add the typing to `get_times()` -> np.ndaarray, this will make it clear that is returning and in memory object and also we should add a docstring describing this behavior.

src/spikeinterface/core/baserecording.py
h-mayorquin: Yes!

alejoe91: Then done in last commit :)

h-mayorquin (approved):

PR #3319, opened 2024-08-20
Fix DeprecationWarnings from distance calculations

rkim48:
This PR fixes a minor warning that results from postprocessing/tests/test_template_similarity.py:

_DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)_

NumPy version: 1.26.4
scikit-learn version: 1.5.1
SciPy version: 1.12.0


alejoe91:
LGTM!

Issue #3221, opened 2024-07-17
docker error when using singularity image to run ks4


a-zmz:
Hello,

i have been running ks4 with singularity successfully in the past two weeks, but today i got an error saying docker module not found:

```
ModuleNotFoundError                       Traceback (most recent call last)                                                                                                                                                                   
File ~/Documents/git/analysis/analysis/preprocessing/test_ks4.py:122                                                                                                                                                                          
    119 os.chdir(preprocess_folder)                                                                                                                                                                                                           
    121 # sort spikes                                                                                                                                                                                                                         
--> 122 sorting = ss.run_sorter(                                                                                                                                                                                                              
    123     sorter_name='kilosort4',                                                                                                                                                                                                          
    124     recording=ap_mcd,                                                                                                                                                                                                                 
    125     #folder=str(base_folder/"ks4_full"),                                                                                                                                                                                              
    126     singularity_image=True,                                                                                                                                                                                                           
    127     remove_existing_folder=True,                                                                                                                                                                                                      
    128     #**ks4_params,                                                                                                                                                                                                                    
    129 )                                                                                                                                                                                                                                     
    131 # curate                                                                                                                                                                                                                              
    132 sorting = scur.remove_duplicated_spikes(                                                                                                                                                                                              
    133     sorting,                                                                                                                                                                                                                          
    134     censored_period_ms=0.3,                                                                                                                                                                                                           
    135     method="keep_first_iterative",                                                                                                                                                                                                    
    136 )                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                              
File ~/Documents/git/spikeinterface/src/spikeinterface/sorters/runsorter.py:210, in run_sorter(sorter_name, recording, folder, remove_existing_folder, delete_output_folder, verbose, raise_error, docker_image, singularity_image, delete_con
tainer_files, with_output, output_folder, **sorter_params)                                                                                                                                                                                    
    204         if not has_spython():                      
    205             raise RuntimeError(                    
    206                 "The python `spython` package must be installed to "                                           
    207                 "run singularity. Install with `pip install spython`"                                          
    208             )                                      
--> 210     return run_sorter_container(                   
    211         container_image=container_image,           
    212         mode=mode,                                 
    213         **common_kwargs,                           
    214     )                                              
    216 return run_sorter_local(**common_kwargs)

[...]

PR #3335, opened 2024-08-25
fix: download apptainer images without docker client

mhhennig:
Previous code failed when docker could not be run on the host system. Now getting the docker image via spython.

### DIFF
--- a/src/spikeinterface/sorters/container_tools.py
+++ b/src/spikeinterface/sorters/container_tools.py
@@ -99,7 +99,7 @@ class ContainerClient:
                 singularity_image = sif_file
             else:
 
-                docker_image = Client.load('docker://'+container_image)
+                docker_image = Client.load("docker://" + container_image)
                 if docker_image and len(docker_image.tags) > 0:
                     tag = docker_image.tags[0]
                     print(f"Building singularity image from local docker image: {tag}")
--- a/src/spikeinterface/sorters/container_tools.py
+++ b/src/spikeinterface/sorters/container_tools.py
@@ -99,7 +99,7 @@ class ContainerClient:
                 singularity_image = sif_file
             else:
 
-                docker_image = self._get_docker_image(container_image)
+                docker_image = Client.load('docker://'+container_image)
                 if docker_image and len(docker_image.tags) > 0:
                     tag = docker_image.tags[0]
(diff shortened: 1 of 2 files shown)


mhhennig:
Btw, this should solve #3221

src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -232,7 +233,7 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
                 do_CAR=do_CAR,
                 invert_sign=invert_sign,
                 device=device,
-                save_preprocesed_copy=save_preprocessed_copy,  # this kwarg is correct (typo)
+                save_preprocessed_copy=save_preprocessed_copy,  # this kwarg is correct (typo)
```
alejoe91, line 236: @mhhennig this was a typo in a previous KS version (see https://github.com/SpikeInterface/spikeinterface/pull/3316). Can you revert the KS changes? It's better to have a dedicated PR to handle them

alejoe91 (changes requested):
See comment

mhhennig:
Sorry, this ended up in the same PR. So here we now have:
- fix for failing to pull containers when only singularity is available
- fix for recent changed in Kilosort4 API (latest pypi version (4.0.16 from 14 August) which prevents it from running; note this change will break older versions, but the code is not compatible

src/spikeinterface/sorters/external/kilosort4.py
mhhennig: Ok

mhhennig:
Removed the KS commit.

alejoe91 (approved):
Thansk @mhhennig 

It's working on my end too after uninstallinf docker!

src/spikeinterface/sorters/container_tools.py
```diff
@@ -99,7 +99,7 @@ def __init__(self, mode, container_image, volumes, py_user_base, extra_kwargs):
                 singularity_image = sif_file
             else:
 
-                docker_image = self._get_docker_image(container_image)
+                docker_image = Client.load("docker://" + container_image)
```
samuelgarcia, line 102: The patch should have been done in `_get_docker_image`

PR #3265, opened 2024-07-29
Fix KS2/2.5/3 skip_kilosort_preprocessing

alejoe91:
Fixes #3263

PR #3339, opened 2024-08-26
Updates to kilosort 4: version >= 4.0.16, `bad_channels`, `clear_cache`, `use_binary_file`

chrishalcrow:
As discussed in #3316, we decided to only support the latest version of kilosort. This PR checks the version and makes the arguments consistent with the kilosort arguments for version 4.0.16.

I added the version check by extending `initialize_folder`, which is where the installation check is. It's a little awkward because we should do the installation check before the version check, but this doesn't quite fit in the structure of the `BaseSorter` `initialize_folder` method. So we actually check that kilosort is installed twice.


src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -222,36 +244,22 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
...
-                device=device,
-            )
+        ops = initialize_ops(
+            settings=settings,
+            probe=probe,
+            data_dtype=recording.get_dtype(),
+            do_CAR=do_CAR,
+            invert_sign=invert_sign,
+            device=device,
+            save_preprocessed_copy=save_preprocessed_copy,  # this kwarg is correct (typo)
```
zm711, line 257: Just to doublecheck. As of the most recent version this kwarg is corrected or not?

chrishalcrow: Yes - corrected since 4.0.13

alejoe91:
@chrishalcrow @JoeZiminski 

I would suggest using this PR also to:

- [ ] Update the Kilosort4 action
- [ ] Make sure docker image is up-to-date: PR here https://github.com/SpikeInterface/spikeinterface-dockerfiles/pull/81

alejoe91:
@chrishalcrow I'll push a change for the `use_binary_file` directly to this PR (updated title)

src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -186,10 +224,22 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
         recording = cls.load_recording_from_folder(sorter_output_folder.parent, with_warnings=False)
         probe = load_probe(probe_path=probe_filename)
         probe_name = ""
-        filename = ""
 
-        # this internally concatenates the recording
-        file_object = RecordingExtractorAsArray(recording_extractor=recording)
+        if params["use_binary_file"]:
```
alejoe91, line 228: Actually, if the recording is binary I would always do this path. It's faster.

@zm711 @samuelgarcia do you agree?

zm711: That makes sense to me. KS4 likes the binary so if that's an option I think feeding it that is better.

alejoe91: done! the only *caveat* is that if one has a binary recording, but wants to force the `RecordingExtractorAsArray` mode, the only way is to modify the recording, e.g.: `recording_non_bin = spre.scale(recording_bin)`

alejoe91:
@chrishalcrow @zm711 @JoeZiminski also updated the KS4 actions. Now only versions >= 4.0.16 are tested.

A test run on a tmp branch is running here: https://github.com/SpikeInterface/spikeinterface/actions/runs/10670770848/job/29575421194

Let me know what you think!

src/spikeinterface/sorters/external/kilosort4.py
alejoe91: done and added a test that checks that results are the same: https://github.com/chrishalcrow/spikeinterface/blob/only-allow-above-ks-4-16/.github/scripts/test_kilosort4_ci.py#L401

alejoe91:
@JoeZiminski I added a trigger for changes to the `kilosort4` file and a general cleanup of the ks4 tests.

Let me know what you think!

.github/scripts/test_kilosort4_ci.py
```diff
@@ -554,16 +562,18 @@ def _get_kilosort_native_settings(self, recording, paths, param_key, param_value
...
-        if param_key in RUN_KILOSORT_ARGS:
-            run_kilosort_kwargs = {param_key: param_value}
-        else:
-            if param_key != "change_nothing":
-                settings.update({param_key: param_value})
-            run_kilosort_kwargs = {}
+            if param_key in RUN_KILOSORT_ARGS:
+                run_kilosort_kwargs = {param_key: param_value}
+            else:
+                if param_key != "change_nothing":
```
JoeZiminski, line 574: was this removed? so should be `if param_key`?

JoeZiminski:
Hey @alejoe91 sorry I never got around this in good time, thanks for taking care of it and the clean up! Looks great, also learnt some cool SI functions I did not know about it. Just one comment above. Cheers!

.github/scripts/test_kilosort4_ci.py
alejoe91: good catch!

alejoe91: done :)

alejoe91 (approved):

JoeZiminski:
Great! @alejoe91 before merging this does it make sense to release SI patch 0.101.1?

alejoe91:
> Great! @alejoe91 before merging this does it make sense to release SI patch 0.101.1?

I would actually include it in the release:) It's a great upgrade!

src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -51,14 +56,16 @@ class Kilosort4Sorter(BaseSorter):
...
+        "duplicate_spike_ms": 0.25,
         "do_correction": True,
         "keep_good_only": False,
         "save_extra_kwargs": False,
         "skip_kilosort_preprocessing": False,
         "scaleproc": None,
         "save_preprocessed_copy": False,
         "torch_device": "auto",
+        "bad_channels": None,
+        "use_binary_file": False,
```
alejoe91, line 68: Make default None to as "auto"

alejoe91: Final change after discussing with Sam: by default `use_binary_file` is `None`, which means that KS will run on the binary file is the recording is binary compatible, otherwise on the wrapper. Setting it to `True` or `False` will force either options, which will require creating a `recording.dat` file in case it's true and the rec is not binary compatible.

For that reason, I added the `delete_recording_dat`, which only deletes a newly created `recording.dat` file (true by default)

JoeZiminski:
Sorry for the back and forth @alejoe91, unless I'm mistaken a lot of the commits for finalising KS4 versions <4.0.16 were added after 0.101.0 was released (e.g. #3805, #3276, I think #3265). So it would be cleaner to make a patch version before merging this PR for a clean break. However, it's not the end of the world if not, but the 'transition point' would be the commit before this is merged rather than a full version which might be confusing.

alejoe91:
Hey @JoeZiminski 

#3265 is for KS<4. For the other changes, they are all to update the most recent KS4 version, including the CI changes. So I think it's ok to have the new release as a breaking point.

In addition, since we rely on many external tools, I think that in this case a patch release is also ok. We should encourage people to update KS4 anyways

JoeZiminski:
I think #3276 adds `save_preprocesed_copy` (old spelling) which is for KS <4.0.13 (the typo is fixed in 4.0.13). That CI PR is to test <4.0.13 also I think and then is overwritten here, but the CI stuff doesn't matter so much. I think if you want the 'final final version' of spikeinterface that supports KS4 <4.0.13 including tests you would need to go to the commit prior to this one as 0.101.0 will be missing `save_preprocessed_copy` (and tests, that matters less). I think this is an extreme edge case though and so happy to squash it into 0.101.1, in practice I think it's unlikely anyone will need such a specific version!

.github/scripts/test_kilosort4_ci.py
```diff
@@ -10,111 +10,101 @@
...
+
+PARAMS_TO_TEST = list(PARAMS_TO_TEST_DICT.keys())
+
+PARAMETERS_NOT_AFFECTING_RESULTS = [
+    "artifact_threshold",
+    "ccg_threshold",
+    "acg_threshold",
+    "cluster_downsampling",
+    "cluster_pcs",
+    "duplicate_spike_ms",  # this is because gorund-truth spikes don't have violations
```
chrishalcrow, line 95:
```suggestion
    "duplicate_spike_ms",  # this is because ground-truth spikes don't have violations
```

.github/scripts/test_kilosort4_ci.py
```diff
@@ -516,6 +539,7 @@ def monkeypatch_filter_function(self, X, ops=None, ibatch=None):
         )
 
         monkeypatch.undo()
+        si.read_kilosort(kilosort_output_dir)
```
chrishalcrow, line 542: What does this do?

.github/scripts/test_kilosort4_ci.py
```diff
@@ -537,33 +561,16 @@ def monkeypatch_filter_function(self, X, ops=None, ibatch=None):
...
         # These are all extremely close (usually just 1 spike, 1 idx different).
-        results = self._get_sorting_output(kilosort_output_dir, spikeinterface_output_dir)
+        results = {}
+        results["ks"] = {}
+        results["ks"]["st"] = np.load(kilosort_output_dir / "spike_times.npy")
+        results["ks"]["clus"] = np.load(kilosort_output_dir / "spike_clusters.npy")
+        results["si"] = {}
+        results["si"]["st"] = np.load(spikeinterface_output_dir / "sorter_output" / "spike_times.npy")
+        results["si"]["clus"] = np.load(spikeinterface_output_dir / "sorter_output" / "spike_clusters.npy")
         assert np.allclose(results["ks"]["st"], results["si"]["st"], rtol=0, atol=1)
```
chrishalcrow, line 571: Should we also check `results["ks"]["clus"] == results["si"]["clus"]` here?

.github/scripts/test_kilosort4_ci.py
alejoe91: Nothing! It's a leftover :P

alejoe91, line 542:
```suggestion
```

chrishalcrow:
Looks good, and is working for me locally.

I don't think we need a release that supports in-between kilosort versions, since it was rapidly developing in its first months. I'd vote to have that _if_ there had been a long-stable release, followed by a significant update.

src/spikeinterface/sorters/external/kilosort4.py
zm711: I think that is a great solution overall!

src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -101,6 +110,11 @@ class Kilosort4Sorter(BaseSorter):
         "scaleproc": "int16 scaling of whitened data, if None set to 200.",
         "save_preprocessed_copy": "save a pre-processed copy of the data (including drift correction) to temp_wh.dat in the results directory and format Phy output to use that copy of the data",
         "torch_device": "Select the torch device auto/cuda/cpu",
+        "bad_channels": "A list of channel indices (rows in the binary file) that should not be included in sorting. Listing channels here is equivalent to excluding them from the probe dictionary.",
+        "use_binary_file": "If True then Kilosort is run using a binary file. In this case, if the input recording is not binaru compatible, it is written to a binary file in the output folder. "
```
zm711, line 114:
```suggestion
        "use_binary_file": "If True then Kilosort is run using a binary file. In this case, if the input recording is not binary compatible, it is written to a binary file in the output folder. "
```

.github/scripts/test_kilosort4_ci.py
alejoe91: done

src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -101,6 +111,12 @@ class Kilosort4Sorter(BaseSorter):
         "scaleproc": "int16 scaling of whitened data, if None set to 200.",
         "save_preprocessed_copy": "save a pre-processed copy of the data (including drift correction) to temp_wh.dat in the results directory and format Phy output to use that copy of the data",
         "torch_device": "Select the torch device auto/cuda/cpu",
+        "bad_channels": "A list of channel indices (rows in the binary file) that should not be included in sorting. Listing channels here is equivalent to excluding them from the probe dictionary.",
+        "clear_cache": "If True, force pytorch to free up memory reserved for its cache in between memory-intensive operations. Note that setting `clear_cache=True` is NOT recommended unless you encounter GPU out-of-memory errors, since this can result in slower sorting.",
```
zm711, line 115: Tiny question here. I really like this explanation. Do we want to make it clear for the user that this is being handled by KS4 and not by spikeinterface or do you think it doesn't matter?

alejoe91: I think that then we should do it with every KS argument, no? :)

zm711: Yeah I think so. The issue is as soon as we start mixing in spikeinterface commands and KS4 commands I can imagine users getting confused. So letting them know when it is a KS4 argument causing problems vs us will make the issue tracker easier to manage. But we could also say it is the user's responsibility to know what is KS4/spikeinterface. So although I lean toward being explicit one could argue that it is not necessary :)

guptadivyansh:
I would just like to report here that `use_binary_file` makes a big difference to KS4 sorting performance. Writing the binary does take a while but overall I managed to sort a ~2 hour recording in ~4 hours with `use_binary_file = True` compared to ~22 hours with the default `use_binary_file = False` for the same recording. So, thank you for adding this param!

I would even suggest making it the default behavior unless there is a good reason/policy against using disk space.

Workstation: Intel i7 8700K, Nvidia 3060, 64GB RAM, SSD

zm711:
I argued for making that the default so I'm in favor! It might actually be worth writing this out in a new separate issue since this is a merged PR. That will gives us the opportunity to discuss it again and make it discoverable to other users so we can see if there is community desire for this. Finally it will allow other users to discover this feature so they might test it out :)

samuelgarcia:
I was the actual default but given this enormous diff, I guess we should move.
It is hard to understand why this is so slow using the spikeinterface API.

PR #3346, opened 2024-08-28
Clean-up identity merges in `get_potential_auto_merges`

### LINKED ISSUE #3345: `get_potential_auto_merges` returns units themselves as potential merges when using `feature_neighbors`
We were testing the new presets and `feature_neighbors` looked pretty good with the defaults, but it returned a bunch of units as potential merges with themselves.

```python
[(1,1), (2,2), (1,2)]
```

things like that. Is there a reason for that or is it okay to add a clean-up step before we return the merges?

@yger @DradeAW (I forget which of you implemented this one? )

zm711:
Fixes #3345

### DIFF
--- a/src/spikeinterface/curation/auto_merge.py
+++ b/src/spikeinterface/curation/auto_merge.py
@@ -361,6 +361,9 @@ def get_potential_auto_merge(
     ind1, ind2 = np.nonzero(pair_mask)
     potential_merges = list(zip(unit_ids[ind1], unit_ids[ind2]))
 
+    # some methods return identities ie (1,1) which we can cleanup first.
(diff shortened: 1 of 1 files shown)


alejoe91 (approved):

zm711:
Just to add some more info. This did not happen with every single unit. Only a subset of the units were suggested as self-merges. So maybe it is providing some data about cluster quality, but I haven't looked at the implementation, yet @yger.

yger:
Sorry for the late reply, these functions and auto_merges are still WIP see #2934. In theory this should not happen, since the KNN tree should only tests dissimilar units. I'll check the implementation, but thanks for the patch

PR #3347, opened 2024-08-28
Analyzer extension exit status

### LINKED ISSUE #3329: Sorting analyzer extensions don't know if they've been interrupted
Hi all — sometimes a sorting analyzer's computation of an extension is interrupted, either accidentally (i.e. a SLURM job times out), or manually (I want to change a parameter). Normally, the way one would deal with this, in a script, is to check if the extension has been completed with `get_extension`/`has_extension`, but these only check for the presence of the corresponding dir and the file `params.json` (see [here](https://github.com/SpikeInterface/spikeinterface/blob/39218069ca61529575abf70095d1ce7912a410d3/src/spikeinterface/core/sortinganalyzer.py#L1364)), and so if the computation has been interrupted partway through, the code would treat it as completed, but downstream analyses would fail / be wrong. Currently, for me, I'm manually deleting extension folders if this occurs and I need to fix it. Is there an existing better solution? Otherwise, I'd propose simply having a `is_finished` or `exit_status` flag in `info.json` that could be used for `has_extension`, rather than just checking that the folder / params are there. Thanks!

jonahpearl:
Fixes #3329

Here's a first pass at adding information about an extension's completion (or not) to its metadata.

This PR adds a `run_info.json` file to each extension's folder. This contains keys:
* `run_completed` (whether or not the call to `AnalyzerExtension._run()` finished),
* `data_loadable` (whether or not `AnalyzerExtension.load_data()` can be called without an error) 
* `runtime_s` (the runtime of `AnalyzerExtension._run()` in seconds). 

The core logic is implemented in `AnalyzerExtension.run()` and wraps the call to the extension-specific `_run()`. Then when `AnalyzerExtension.get_extension()` is called (and redirects to `AnalyzerExtension.load()` internally), if the run wasn't completed, it simply returns None as if the extension doesn't exist, and the user would be able to catch that and re-run the extension.
 
There is a mild complication with extensions that use the `run_node_pipeline` if called through `compute_several_extensions`, because then `AnalyzerExtension.run()` is never actually called. Instead, it looks like `AnalyzerExtension.save()` is used, so I added the relevant lines there to catch that, and assume that if the code makes it there, the run is completed.

TODO:
* I'm not sure if the implementations in `merge()` and `copy()` are correct, I don't totally understand when those methods are expected to be called / what's wrapping them.
* Write some tests?
 
Notes:
* It is a bit of a slow-down to check if the data are loadable after each run. I'm ambivalent as to whether that check is actually useful, especially since data corruption is more likely to happen sometimes significantly after the run, not immediately after.
* It seemed more reasonable to add a separate `run_info.json` file than trying to cram this into `info.json`, since that is more about the code itself.

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -2186,7 +2186,9 @@ class AnalyzerExtension:
         return self._get_pipeline_nodes()
 
     def get_data(self, *args, **kwargs):
-        assert self.run_info["run_completed"], f"You must run the extension {self.extension_name} before retrieving data"
+        assert self.run_info[
+            "run_completed"
+        ], f"You must run the extension {self.extension_name} before retrieving data"
         assert len(self.data) > 0, "Extension has been run but no data found."
         return self._get_data(*args, **kwargs)
 
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -1947,10 +1947,10 @@ class AnalyzerExtension:
                     # this load in memmory
                     ext_data = np.array(ext_data_)
 
-        if ext_data is None:
+        if ext_data is not None:
+            self.data[ext_data_name] = ext_data    
+        else:
             warnings.warn(f"Found no data for {self.extension_name}, extension should be re-computed.")
-
-        self.data[ext_data_name] = ext_data
 
     def copy(self, new_sorting_analyzer, unit_ids=None):
         # alessio : please note that this also replace the old select_units!!!
@@ -2183,7 +2183,8 @@ class AnalyzerExtension:
         return self._get_pipeline_nodes()
 
     def get_data(self, *args, **kwargs):
-        assert len(self.data) > 0, f"You must run the extension {self.extension_name} before retrieving data"
+        assert self.run_info["run_completed"], f"You must run the extension {self.extension_name} before retrieving data"
+        assert len(self.data) > 0, "Extension has been run but no data found — data file might be missing. Re-compute the extension."
         return self._get_data(*args, **kwargs)
 
 
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -2002,6 +2002,8 @@ class AnalyzerExtension:
         self._save_importing_provenance()
         self._save_data(**kwargs)
         self.run_info["data_loadable"] = self._check_data_loadable()
+        if self.run_info["data_loadable"]:
+            self.run_info["run_completed"] = True  # extensions that go through compute_several_extensions() and then run_node_pipeline() never have ext_instance.run() called, so need to check here (or at least somewhere) instead
         self._save_run_info()
 
     def _save_data(self, **kwargs):
@@ -2122,6 +2124,7 @@ class AnalyzerExtension:
         if save:
             self._save_params()
             self._save_importing_provenance()
+            self._save_run_info()
 
     def _save_params(self):
         params_to_save = self.params.copy()
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -19,8 +19,6 @@ import probeinterface
 
 import spikeinterface
 
-from zarr.errors import ArrayNotFoundError
-
 from .baserecording import BaseRecording
 from .basesorting import BaseSorting
 
@@ -1339,6 +1337,7 @@ extension_params={"waveforms":{"ms_before":1.5, "ms_after": "2.5"}}\
 
             job_name = "Compute : " + " + ".join(extensions_with_pipeline.keys())
 
+            t_start = perf_counter()
             results = run_node_pipeline(
                 self.recording,
                 all_nodes,
@@ -1348,10 +1347,14 @@ extension_params={"waveforms":{"ms_before":1.5, "ms_after": "2.5"}}\
                 squeeze_output=False,
                 verbose=verbose,
             )
+            t_end = perf_counter()
+            # for pipeline node extensions we can only track the runtime of the run_node_pipeline
+            runtime_s = np.round(t_end - t_start, 1)
 
             for r, result in enumerate(results):
                 extension_name, variable_name = result_routage[r]
                 extension_instances[extension_name].data[variable_name] = result
+                extension_instances[extension_name].run_info["runtime_s"] = runtime_s
 
             for extension_name, extension_instance in extension_instances.items():
                 self.extensions[extension_name] = extension_instance
@@ -1859,13 +1862,20 @@ class AnalyzerExtension:
         ext = cls(sorting_analyzer)
         ext.load_params()
         ext.load_run_info()
-        if ext.run_info["run_completed"]:
+        if ext.run_info is not None:
+            if ext.run_info["run_completed"]:
+                ext.load_data()
+                if cls.need_backward_compatibility_on_load:
+                    ext._handle_backward_compatibility_on_load()
+                if len(ext.data) > 0:
+                    return ext
+        else:
+            # this is for back-compatibility of old analyzers
             ext.load_data()
             if cls.need_backward_compatibility_on_load:
                 ext._handle_backward_compatibility_on_load()
             if len(ext.data) > 0:
                 return ext
-
         # If extension run not completed, or data has gone missing,
         # return None to indicate that the extension should be (re)computed.
         return None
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1930,19 +1990,33 @@ def merge(
 
     def run(self, save=True, **kwargs):
         if save and not self.sorting_analyzer.is_read_only():
-            # this also reset the folder or zarr group
+            # NB: this call to _save_params() also resets the folder or zarr group
             self._save_params()
             self._save_importing_provenance()
+            self._save_run_info()
 
+        start = time()
```
samuelgarcia, line 1998: Can you use `time.perf_counter()` instead ?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1898,12 +1934,35 @@ def load_data(self):
...
                     ext_data = np.array(ext_data_)
-                self.data[ext_data_name] = ext_data
+
+        if ext_data is None:
+            warnings.warn(f"Found no data for {self.extension_name}, extension should be re-computed.")
+
+        if keep:
+            self.data[ext_data_name] = ext_data
+
+    def _check_data_loadable(self):
```
alejoe91, line 1944: After discussing with @samuelgarcia , we agree this is an overkill. It's enough for the purpose of this PR to add `run_info` with run time and run_completed. Can you remove this @jonahpearl ?

src/spikeinterface/core/sortinganalyzer.py
jonahpearl: All set

src/spikeinterface/core/sortinganalyzer.py
jonahpearl: All set

jonahpearl:
I also fixed the edge case where (for some reason) the run completes, but then the data file gets deleted; it will now raise a warning upon loading the sorting analyzer, and return None on calls to get_extension(), since the extension is no longer usable and should be re-computed as if it had never been.

jonahpearl:
Also — I still think my naive / default usage of this code would be to check `has_extension()`, and if that's true, assume everything is ok. Sure, the "check for None" pattern can be shown in the docs, but it's still odd that `has_extension()` now essentially means, is there a folder, regardless of if there's data. It looks like internally, sometimes it's used to mean, "is there a folder", like [here](https://github.com/jonahpearl/spikeinterface/blob/8e9acfc66bd22b2232914274c12319f43e197559/src/spikeinterface/core/sortinganalyzer.py#L1234), but then sometimes it means, "is there usable data", like [here](https://github.com/jonahpearl/spikeinterface/blob/8e9acfc66bd22b2232914274c12319f43e197559/src/spikeinterface/widgets/motion.py#L177) or [here](https://github.com/jonahpearl/spikeinterface/blob/8e9acfc66bd22b2232914274c12319f43e197559/src/spikeinterface/qualitymetrics/misc_metrics.py#L801) or [here](https://github.com/jonahpearl/spikeinterface/blob/8e9acfc66bd22b2232914274c12319f43e197559/src/spikeinterface/widgets/motion.py#L186). 

Just taking the third example:
```python
if sorting_analyzer.has_extension("spike_amplitudes"):
    peak_amplitudes = sorting_analyzer.get_extension("spike_amplitudes").get_data()
else:
    peak_amplitudes = None
```
right now, that will fail if the extension doesn't have data, since `has_extension()` will come back `True` but then `get_extension()` will return `None` and the call to `get_data()` will raise an `AttributeError`. I guess currently the preferred implementation would be
```python
spike_amps = sorting_analyzer.get_extension("spike_amplitudes")
if spike_amps:
    peak_amplitudes = spike_amps.get_data()
else:
    peak_amplitudes = None
```
(or with a tertiary operator plus assignment expression it can fit in one line)
```python
peak_amplitudes = _ext.get_data() if (_ext := sorting_analyzer.get_extension("spike_amplitudes")) else None
```
but the pattern still just feels confusing 🤷

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1897,7 +1931,6 @@ def load_data(self):
                     ext_data = pickle.load(ext_data_file.open("rb"))
                 else:
                     continue
-                self.data[ext_data_name] = ext_data
```
alejoe91, line 1900: this is wrong!!! Many extensions have multiple files they need to load (and that's why tests are fortunately failing!) :)

I'll push a fix with some other changes if you don't mind @jonahpearl

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1944,24 +1995,36 @@ def merge(
...
+        self.run_info["run_completed"] = True
+        self._save_run_info()
+
     def save(self, **kwargs):
         self._save_params()
         self._save_importing_provenance()
         self._save_data(**kwargs)
+        self.run_info["run_completed"] = (
+            True  # extensions that go through compute_several_extensions() and then run_node_pipeline() never have ext_instance.run() called, so need to change run_completed here (or somewhere, at least)
+        )
```
alejoe91, line 2026:
```suggestion
        self.run_info["run_completed"] = True
```
This is done in the last commit :) I just set the `run_pipeline_node` time for each pipelinable extension

src/spikeinterface/core/sortinganalyzer.py
jonahpearl: ah, good catch!

alejoe91:
@samuelgarcia this is good to merge on my side. I added a back-compatibility logic for loading folders/zarr produced prior to this change

alejoe91 (approved):

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1888,6 +1934,7 @@ def load_data(self):
                     # and have a link to the old buffer on windows then it fails
                     # ext_data = np.load(ext_data_file, mmap_mode="r")
                     # so we go back to full loading
+                    print(f"{ext_data_file} is numpy!")
```
alejoe91, line 1937:
```suggestion
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1944,24 +1994,34 @@ def merge(
...
-            # this also reset the folder or zarr group
+            # NB: this call to _save_params() also resets the folder or zarr group
             self._save_params()
             self._save_importing_provenance()
+            self._save_run_info()
 
+        t_start = perf_counter()
         self._run(**kwargs)
+        t_end = perf_counter()
+        self.run_info["runtime_s"] = np.round(t_end - t_start, 1)
```
alejoe91, line 2011:
```suggestion
        self.run_info["runtime_s"] = t_end - t_start
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1345,10 +1347,14 @@ def compute_several_extensions(self, extensions, save=True, verbose=False, **job
                 squeeze_output=False,
                 verbose=verbose,
             )
+            t_end = perf_counter()
+            # for pipeline node extensions we can only track the runtime of the run_node_pipeline
+            runtime_s = np.round(t_end - t_start, 1)
```
alejoe91, line 1352:
```suggestion
            runtime_s = t_end - t_start
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1345,10 +1347,14 @@ def compute_several_extensions(self, extensions, save=True, verbose=False, **job
...
 
             for r, result in enumerate(results):
                 extension_name, variable_name = result_routage[r]
                 extension_instances[extension_name].data[variable_name] = result
+                extension_instances[extension_name].run_info["runtime_s"] = runtime_s
```
alejoe91, line 1357:
```suggestion
                extension_instances[extension_name].run_info["runtime_s"] = runtime_s
                extension_instances[extension_name].run_info["run_completed"] = True
```

alejoe91 (approved):

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1345,10 +1347,15 @@ def compute_several_extensions(self, extensions, save=True, verbose=False, **job
...
                 verbose=verbose,
             )
+            t_end = perf_counter()
+            # for pipeline node extensions we can only track the runtime of the run_node_pipeline
+            runtime_s = t_end - t_start
 
             for r, result in enumerate(results):
                 extension_name, variable_name = result_routage[r]
                 extension_instances[extension_name].data[variable_name] = result
+                extension_instances[extension_name].run_info["runtime_s"] = runtime_s
```
samuelgarcia, line 1357: OK for me but if we want the total run ttime then suming of run_time will be sring because this run time is shared.

samuelgarcia (approved):

samuelgarcia:
Merci beaucoup Jonah et Alessio.

src/spikeinterface/core/sortinganalyzer.py
alejoe91: This is the best estimate we can get :)

Issue #3348, opened 2024-08-28
Stray .nfs file preventing deletion of waveforms extension folder


jonahpearl:
Hi all — I'm on a Linux / NFS system (CentOS 7 specifically, on my school's HPCC), and I encounter an OSError every time I try to re-compute the waveforms extension of a sorting analyzer. The core problem is that there's a stray .nfs temporary file hanging out in the waveforms folder. If I do `ls -a [that folder]` I can see it clearly, it's always something like `.nfs9eb83351caf602ec00005734`. 

I can see that it's python using the file with `lsof`:
```
COMMAND   PID    USER   FD   TYPE DEVICE  SIZE/OFF                 NODE NAME
python  30135 jop9552  mem    REG   0,45 128896688 11436947680097862380 [...]/analyzer_folder/extensions/waveforms/.nfs9eb83351caf602ec00005734
python  30135 jop9552   93u   REG   0,45 128896688 11436947680097862380 [...]/analyzer_folder/extensions/waveforms/.nfs9eb83351caf602ec00005734
```
and if I call `kill` on that process, my notebook's kernel crashes, so I'm pretty sure it's coming from spikeinterface.

The offending function is at `/spikeinterface/core/sortinganalyzer.py:2090`: `shutil.rmtree(extension_folder)`.

### MRE:
```python
recording = si.load_extractor([any recording])
sorting = si.read_sorter_folder([any sorting])
analyzer = si.create_sorting_analyzer(
    sorting=sorting, 
    recording=recording, 
    format='binary_folder', 
    folder="./tmp_analyzer",
)
to_compute = ["random_spikes", "waveforms"]
analyzer.compute(to_compute)
analyzer.compute("waveforms")  # tries to reset folder before recomputing but fails at deletion
```

Not sure how to go about debugging this but any advice is welcome. This is v0.101.0. Thanks!

zm711:
What is the `OSError` that is actually given :) I don't see it in the issue.

h-mayorquin:
It is probably in the recording and sorting that keeps a reference file handle. Probably related to:

https://github.com/SpikeInterface/spikeinterface/issues/3295

jonahpearl:
> What is the `OSError` that is actually given :) I don't see it in the issue.

Added the traceback :)

jonahpearl:
Also the same issue happens if I load sorting output from mountainsort5, so I don't think it's related to KS4 as being discussed in #3295.

[...]

PR #3349, opened 2024-08-28
Fix zarr folder suffix handling

jonahpearl:
While troubleshooting for #3348, I noticed some odd behavior with zarr-based sorting analyzers. Specifically, in some parts of the code, it seemed like the user would be permitted to pass a folder name with the `.zarr` suffix included (ie `folder=/my/favorite/analyzer/` instead of `folder=/my/favorite/analyzer.zarr/`). This led to two error cases when using `create_sorting_analyzer` if the user tried to make a new folder without the suffix: 
1) it would be created by [create_zarr](https://github.com/SpikeInterface/spikeinterface/blob/007b6efd0e675e60658199112d0541fba2b432cd/src/spikeinterface/core/sortinganalyzer.py#L272) [with the suffix](https://github.com/SpikeInterface/spikeinterface/blob/007b6efd0e675e60658199112d0541fba2b432cd/src/spikeinterface/core/sortinganalyzer.py#L492), but then would fail to load directly after because the `load_from_zarr` function didn't add the `.zarr` suffix.
2) when checking for the folder's existence when `overwrite=True` in the [beginning](https://github.com/SpikeInterface/spikeinterface/blob/007b6efd0e675e60658199112d0541fba2b432cd/src/spikeinterface/core/sortinganalyzer.py#L114) of the function, it would not delete the folder (since the real folder has the `.zarr` extension, but the user's `folder` arg doesn't). But then later in `SortingAnalyzer.create_zarr` there is another check for whether the folder exists, and when the suffix'd folder does indeed exist, it raises an error, defeating the point of the `overwrite` kwarg.

This PR fixes that by:
1) creating a single `clean_zarr_folder_name` in `core_tools.py` that will check for a `.zarr` suffix and add one if there isn't already
2) cleaning the user-provided folder to ensure that if `mode="zarr"`, the folder passed to internal funcs always has `.zarr` appended. Cleaning happens in:
* create_sorting_analyzer
* load_sorting_analyzer
* merge_units
* save_as
* select_units
* remove_units
* [not copy, since that goes just into memory]

I also added cleaning in `SortingAnalyzer.create_zarr()` and `SortingAnalyzer._save_or_select_or_merge()` themselves which might be redundant. Hopefully I didn't miss any other entry points but I haven't extensively tested.

Obviously another solution here would be to insist that the user just put the `.zarr` suffix on themselves. But I like this way, as it allows for minimal change to a user's code (mode from binary_folder to zarr) and they will get the desired zarr behavior.

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -17,6 +17,7 @@ import numpy as np
 from .globals import get_global_tmp_folder, is_set_global_tmp_folder
 from .core_tools import (
     check_json,
+    clean_zarr_folder_name,
     is_dict_extractor,
     SIJsonEncoder,
     make_paths_relative,
@@ -1061,9 +1062,7 @@ class BaseExtractor:
                         print(f"Use zarr_path={zarr_path}")
         else:
             if storage_options is None:
-                folder = Path(folder)
-                if folder.suffix != ".zarr":
-                    folder = folder.parent / f"{folder.stem}.zarr"
+                folder = clean_zarr_folder_name(folder)
                 if folder.is_dir() and overwrite:
                     shutil.rmtree(folder)
                 zarr_path = folder
--- a/src/spikeinterface/core/core_tools.py
+++ b/src/spikeinterface/core/core_tools.py
@@ -153,6 +153,13 @@ def check_json(dictionary: dict) -> dict:
     return json.loads(json_string)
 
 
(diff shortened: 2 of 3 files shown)


samuelgarcia:
merci.
This look ok.

alejoe91 (approved):

samuelgarcia (approved):

PR #3351, opened 2024-08-29
Propagate storage_options to load_sorting_analyzer

alejoe91:


src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -157,16 +157,17 @@ def load_sorting_analyzer(folder, load_extensions=True, format="auto"):
...
+        For open buckets, it doesn't need to be specified.
 
     Returns
     -------
     sorting_analyzer : SortingAnalyzer
         The loaded SortingAnalyzer
 
     """
-    if format == "zarr":
-        folder = clean_zarr_folder_name(folder)
```
alejoe91, line 168: the path could be remote, so we have to avid this here

samuelgarcia (approved):

PR #3354, opened 2024-08-29
Fix widgets tests and add test on unit_table_properties

alejoe91:
Not sure how but https://github.com/SpikeInterface/spikeinterface/pull/3299 broke a bunch of widgets tests that were uncaught and are showing up in the daily full tests.

The fix here is that the `generate_unit_table_view` needs to support both a `SortingAnalyzer` and a `BaseSorting`, because it's used by other widgets as well

### DIFF
--- a/.github/workflows/all-tests.yml
+++ b/.github/workflows/all-tests.yml
@@ -202,7 +202,7 @@ jobs:
         shell: bash
         if: env.RUN_WIDGETS_TESTS == 'true'
         run: |
-          pip install -e .[full]
+          pip install -e .[full,widgets]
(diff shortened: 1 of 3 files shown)


zm711:
That must mean that we are reusing something in the full test process that is not being passed around in the widget-only tests right? That happened with the template shading and the analyzer before right?

alejoe91:
@zm711 mystery solved: https://github.com/SpikeInterface/spikeinterface/pull/3354/commits/3b9922e648a0dfc8132ab41b817b36f1bce03f4f

zm711:
That's a simple fix! Nice.

PR #3358, opened 2024-09-01
Fix an issue arising because merge_units does not return new unit ids

grero:
I ran into this issue while trying to manually curate a spike sorting using spikeinterface_gui.


alejoe91:
Hi @grero 

I don't think your PR is correct! You see that the function asks for `new_unit_ids` [here](https://github.com/SpikeInterface/spikeinterface/pull/3358/files#diff-4a0ff9d8cfb969b795daca2d042c3f89dde3ac38a7738d4c910749d3ddf423d3R349) and in fact the change make the tests fail.

Maybe the problem is on the SpikeInterface GUI side? Can you paste the entire traceback?

@samuelgarcia what do you think?

grero:
Hm that's strange. I re-installed spikeinterface and now I don't see the error anymore. Perhaps I was using an older version? I think you can just close this PR then. Sorry for the noise!

PR #3359, opened 2024-09-03
Improve noise level machinery

samuelgarcia:
* Add machinery to compute noise level in paralell
* Add `get_random_recording_slices()` to implement more futur) strategy for random chunk (aka non overlaping,
   regular, uniform...)

A very important change is that now the seed=None (instead of seed=0) in the function which I think is the good way. seed must be explicit and no inplicit. So the consequence is:
  * all test that are running twice the `get_random_data_chunk()` (sometimes this isn hidden) are not garanted 
     anymore  to have the  same results. The solution is to explicitly seed everywhere which is a good practice.

@yger 

@cwindolf  : have a look to this, this is a first step to have a better noise levels estimate in SI.


src/spikeinterface/core/recording_tools.py
```diff
@@ -509,6 +511,87 @@ def determine_cast_unsigned(recording, dtype):
...
 
+
+
+def get_random_recording_slices(recording,
+                                method="legacy",
+                                num_chunks_per_segment=20,
+                                chunk_duration="500ms",
+                                chunk_size=None,
+                                margin_frames=0,
+                                seed=None):
```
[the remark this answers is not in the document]
samuelgarcia: No I prefer seed=None so if you really want to set it is on purpose.
I will change the docstring.

src/spikeinterface/core/recording_tools.py
```diff
@@ -634,12 +731,42 @@ def get_closest_channels(recording, channel_ids=None, num_channels=None):
...
+    
+    one_chunk = recording.get_traces(
+        start_frame=start_frame,
+        end_frame=end_frame,
+        segment_index=segment_index,
+        return_scaled=worker_ctx["return_scaled"],
+    )
+
+
+    if worker_ctx["method"] == "mad":
```
h-mayorquin, line 745: Should pre-allocate to reduce memory footprint.

src/spikeinterface/core/recording_tools.py
```diff
@@ -634,12 +731,42 @@ def get_closest_channels(recording, channel_ids=None, num_channels=None):
...
+    return worker_ctx
+
 def get_noise_levels(
     recording: "BaseRecording",
     return_scaled: bool = True,
     method: Literal["mad", "std"] = "mad",
     force_recompute: bool = False,
-    **random_chunk_kwargs,
+    **kwargs,
+    # **random_chunk_kwargs,
```
h-mayorquin, line 768: Why just not pass different dics? This is more complicated to read, document and has caused bugs in the past before with the verbose and job_kwargs.

h-mayorquin (commented):
Two general questions:
1) Will this not fail with formats that lock IO access to the same region if the chunks overlap?
2) I am still confused on why computing noise requires so many samples. The methods we use assume normality (we use MAD to estimate the std) but then we go and sample far way more than the converge criteria of normal distributions would naively suggests. What gives? Is there some empirical work on this? Now that we have a lot of open data available estimating sampling requirements for a variety of neural data (species, areas, etc) coul be done. It appears to me that this could be a quick and informative paper that we could put out for the community if there is no previous work.

src/spikeinterface/core/recording_tools.py
samuelgarcia: ok

samuelgarcia: this was to keep backward compatibility I guess.

samuelgarcia:
> 1. Will this not fail with formats that lock IO access to the same region if the chunks overlap?

Really good point! The access is read only.
We will be able to add a none overlapiing option in random slices

> 2. I am still confused on why computing noise requires so many samples. The methods we use assume normality (we use MAD to estimate the std) but then we go and sample far way more than the converge criteria of normal distributions would naively suggests. What gives? Is there some empirical work on this? Now that we have a lot of open data available estimating sampling requirements for a variety of neural data (species, areas, etc) coul be done. It appears to me that this could be a quick and informative paper that we could put out for the community if there is no previous work.

Honestly I was pretty sure that the number of sample used to be enough.
After discussion with @cwindolf  I get the impression that we should have more ...
Charlie any coment ?

src/spikeinterface/core/recording_tools.py
```diff
@@ -509,6 +511,87 @@ def determine_cast_unsigned(recording, dtype):
...
+                warnings.warn(
+                    f"chunk_size is greater than the number "
+                    f"of samples for segment index {segment_index}. "
+                    f"Using {chunk_size}."
+                )
+        rng = np.random.default_rng(seed)
+        recording_slices = []
+        low = margin_frames
+        size = num_chunks_per_segment
+        for segment_index in range(num_segments):
```
yger, line 581: There could be an option to avoid overlapping chunks. This was not really necessary as long as we had not too many chunks given the size of the recording, but if we are taking more, maybe this is worth considering it.

cwindolf:
Yeah... in my experience, more blocks helps to stabilize the estimate (let's say we want numbers within x% of each other across runs with different seeds). The data certainly is not Gaussian, it has spikes, and spike activity can vary wildly across a recording. So using very few blocks, they will by chance disproportionately land in higher or lower activity regions (maybe in different ways across channels). You need a good number of blocks to reduce that effect -- for short or super consistent recordings maybe fewer blocks is fine.

Also, it would be cool if `si.zscore()` and the other normalize_scale stuff can use these tools :)

h-mayorquin:
But if the data is not Gaussian would it mean that using MAD to estimate std is wrong? This assumes normality:

https://en.wikipedia.org/wiki/Median_absolute_deviation
(see relationship to Relation to standard deviation here)

Anyway, if your experience is that more samples stabilize the estimator that I think trumps these considerations.

cwindolf:
Yeah, it's wrong! But I don't have any better ideas. Ideally one would be able to subtract away all of the spikes and then MAD the residuals (which would ideally be only Gaussian noise, but even that is not 100% true...), but that requires sorting, which requires some kind of standardization...

h-mayorquin:
Agree on the limitation. Thanks for answering my questions.

src/spikeinterface/core/job_tools.py
```diff
@@ -187,6 +187,21 @@ def ensure_n_jobs(recording, n_jobs=1):
 
     return n_jobs
 
+def chunk_duration_to_chunk_size(chunk_duration, recording):
```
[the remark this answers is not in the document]
samuelgarcia: could be an idea could but I have the feeling that we will overbalance into a over ingeneering zone no ?
Lets discuss later.

src/spikeinterface/core/recording_tools.py
```diff
@@ -509,6 +511,87 @@ def determine_cast_unsigned(recording, dtype):
...
+    """
+    Get random slice of a recording across segments.
+
+    This is used for instance in get_noise_levels() and get_random_data_chunks() to estimate noise on traces.
+
+    Parameters
+    ----------
+    recording : BaseRecording
+        The recording to get random chunks from
+    methid : "legacy"
```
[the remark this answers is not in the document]
samuelgarcia: Done

src/spikeinterface/core/recording_tools.py
samuelgarcia: for a futur PR

src/spikeinterface/core/recording_tools.py
samuelgarcia: we do not have garanty that chaunk have the same size.
I think this will handle more globally when adding `out=...` in `get_traces()`

src/spikeinterface/core/recording_tools.py
samuelgarcia: This is somethign we should discuss. I was for what you propose but this will break backward compatibility.

src/spikeinterface/core/recording_tools.py
```diff
@@ -637,19 +703,52 @@ def get_closest_channels(recording, channel_ids=None, num_channels=None):
...
     Estimate noise for each channel using MAD methods.
     You can use standard deviation with `method="std"`
 
     Internally it samples some chunk across segment.
-    And then, it use MAD estimator (more robust than STD)
+    And then, it use MAD estimator (more robust than STD) ot the STD on each chunk.
+    Finally the average on all MAD is performed.
+
+    The result is cached in a property of the recording.
+    Next call on the same recording will use the cache unless force_recompute=True.
```
alejoe91, line 751:
```suggestion
    And then, it uses the MAD estimator (more robust than STD) or the STD on each chunk.
    Finally the average of all MAD/STD values is performed.

    The result is cached in a property of the recording, so that the next call on the same
    recording will use the cached result unless `force_recompute=True`.
```

src/spikeinterface/core/recording_tools.py
```diff
@@ -662,9 +761,11 @@ def get_noise_levels(
         The method to use to estimate noise levels
     force_recompute : bool
         If True, noise levels are recomputed even if they are already stored in the recording extractor
-    random_chunk_kwargs : dict
-        Kwargs for get_random_data_chunks
-
+    random_slices_kwargs : dict
+        Options transmited to  get_random_recording_slices(), please read documentation from this
+        function for more details.
+    **job_kwargs:
```
alejoe91, line 767: I would actually copy/paste the kwargs in the docstring here, since it's much higher level API

alejoe91 (changes requested):
Super @samuelgarcia!!! 

Just a few suggestions and failing tests to fix

alejoe91 (approved):

PR #3362, opened 2024-09-04
Fix streaming extractor condition in the CI

### LINKED ISSUE #3361: Streaming extractors test running when it should not
Here an example:

https://github.com/SpikeInterface/spikeinterface/actions/runs/10693710896/job/29644312631?pr=3360

There is something wrong with the CI.

h-mayorquin:
Fix #3361


alejoe91 (approved):
Nice catch!

PR #3364, opened 2024-09-04
Add extra protection for template metrics

alejoe91:
Being extra careful is never enough! We simply don't want a full pipeline to fail because computing `spread` for a single unit fails...

### DIFF
--- a/src/spikeinterface/postprocessing/template_metrics.py
+++ b/src/spikeinterface/postprocessing/template_metrics.py
@@ -8,11 +8,9 @@ from __future__ import annotations
 
 import numpy as np
 import warnings
-from typing import Optional
 from copy import deepcopy
 
 from ..core.sortinganalyzer import register_result_extension, AnalyzerExtension
-from ..core import ChannelSparsity
 from ..core.template_tools import get_template_extremum_channel
 from ..core.template_tools import get_dense_templates_array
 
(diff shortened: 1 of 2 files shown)


zm711:
This was the strategy we have for the quality metrics. [See here](https://github.com/SpikeInterface/spikeinterface/blob/740e7799c554387888afad655a10f4f3ace7f9dd/src/spikeinterface/qualitymetrics/pca_metrics.py#L188-L211) Because you are doing the try on each individual function and we assume that any problem means just use a nan. I agree it could mask bugs. 

>  why not catching the underlying bug in one of many functions and handling the cases with nan.
no ?

What do you mean exactly by this? Like if it is a `ValueError` you're okay with nan, but you're not okay to let each function that fails for any reason just return a nan?

alejoe91:
> not super happy with this. This try expcet is too high level for my taste. why not catching the underlying bug in one of many functions and handling the cases with nan. no ?
> 
> For me doing this will hide bugs for long time instead of fixing then. lets discuss this.

That would be the ideal solution, but currently I don't have time for it :) Let's open an issue to keep it as a reminder and merge this one! Ok?

zm711:
If it helps I often have spike cutoff fail for quality metrics because I don't have enough spikes for some units to calculate. It would be so annoying for the quality metrics to break just because I don't have enough spikes sometimes. So I'm fine with Alessio's strategy. We let metrics pass for now. Open an issue and work on it later to be cleaner. If there is a bug and it is always skipped and returns a nan then I would assume someone will open an issue. Like how Chris is finally working on updating nn metrics for quality.

alejoe91:
> Also in favour of lower down checks in the future.
> 
> For now, we should also wrap the `metrics_multi_channel` calculations at line ~275 in a try/except.

Done! Ready on my end

samuelgarcia (approved):

PR #3365, opened 2024-09-05
"quality" property to be read as string instead of object in `BasePhyKilosortSortingExtractor`

h-mayorquin:
At the moment, this specific column which we know is a string is unnecessarily read as an object.

This is a simple patch so this property is converted as a string.

Another possible more general solution is to use this function from pandas after reading the cluster info data frames and merging

https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.convert_dtypes.html 

I think this could simplify the logic here where we do the casting:

https://github.com/h-mayorquin/spikeinterface/blob/5953d6cdaac993928988a715725b824edf11d00a/src/spikeinterface/extractors/phykilosortextractors.py#L167-L183

@pauladkisson


alejoe91:
Thanks @h-mayorquin 

Let's merge this for now, but I like the `convert_dtypes` approach as a more general solution :)

alejoe91 (approved):

PR #3369, opened 2024-09-05
Extend estimate_sparsity, add from_amplitude and deprecate from_ptp

### LINKED ISSUE #3320: Non default Sparsity kwargs when creating SortingAnalyzer
I'd like to create a `SortingAnalyzer` object with sparsity computed from channel groups rather than the default setting. However, it seems sparsity arguments passed to `spikeinterface.core.create_sorting_analyzer()` are passed to `spikeinterface.core.estimate_sparsity()`, which does not handle such method, instead of to `spikeinterface.core.compute_sparsity()`.

At any rate, even with the relatively in-depth documentation specific to explaining sparsity, the many ways to generate it seem redundant/confusing. It seems `compute_sparsity()` is the most full-fledged function for generating it, but the fact it needs `SortingAnalyzer` to do so means one would have create `SortingAnalyzer` (possibly writing to disk etc) with `sparse=False`, generate sparsity and then create it (and write to disk) again. 

Furthermore, it seems that default behavior is for the group property to be referred as `'group'` for recordings, but when a grouped sorting is created from a grouped recording, this key changes automatically to `'channel_group'`, so `compute_sparsity()` will return an error when trying the routine I described above.

alejoe91:
Fixes #3320 

This PR extends the methods supported by the `estimate_sparsity` function. In particular, the `by_property` method is also added and handled differently.

Note that the implementation of the `from_ptp` is deprecated, in favor of `from_snr(..., amplitude_mode="peak_to_peak")`.

A new `from_amplitude` sparsity option is added, to select channels based on uV amplitudes.


yger:
Why was it wrong? Don't you want the ptp to be normalized to noise levels ? I mean, an absolute value do not really make sense, no?

alejoe91:
> Why was it wrong? Don't you want the ptp to be normalized to noise levels ? I mean, an absolute value do not really make sense, no?

Oups! You're definitely right then. Then what's the difference with SNR?

alejoe91:
Actually, I think it makes sense also to give an absolute ptp in uV. Say you want to keep all channels with a template ptp > 20uV. It could be a valid option no?

yger:
Yes, but the problem is that sometimes, the value in the SortingAnalyzer are not even scaled, so absolute values then might be confusing. But I agree it could be an option, if values are scaled. As said, for now, the ptp was mostly an extension of the SNR, taking also into account the pos/neg peaks instead of just the neg, and this is why the noise_levels were requiered to normalize.

alejoe91:
Ok after discussing with @yger we agreed on the following:

- Adding another param called `snr_amplitude_mode`, which is propagated to the `get_template_amplitudes`. This substitutes the `ptp` method which is equivalent to the `snr_amplitude_mode="peak_to_peak"
- Keep the `ptp` method with an absolute threshold

src/spikeinterface/core/sparsity.py
```diff
@@ -25,14 +25,16 @@
...
 
-    peak_sign : str
-        Sign of the template to compute best channels ("neg", "pos", "both")
+    peak_sign : "neg" | "pos" | "both"
+        Sign of the template to compute best channels
     num_channels : int
         Number of channels for "best_channels" method
     radius_um : float
         Radius in um for "radius" method
     threshold : float
```
JoeZiminski, line 34: does this also set the threshold for ptp?

src/spikeinterface/core/sparsity.py
```diff
@@ -25,14 +25,16 @@
...
         Threshold in SNR "threshold" method
```
JoeZiminski, line 35: does this also set the threshold for ptp?

src/spikeinterface/core/sparsity.py
```diff
@@ -25,14 +25,16 @@
         * "by_property" : sparsity is given by a property of the recording and sorting(e.g. "group").
```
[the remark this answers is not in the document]
JoeZiminski: similar for 'energy', although I cannot hazard a guess on how this is computed ⚡

src/spikeinterface/core/sparsity.py
```diff
@@ -353,41 +357,35 @@ def from_snr(cls, templates_or_sorting_analyzer, threshold, noise_levels=None, p
         return cls(mask, unit_ids, channel_ids)
 
     @classmethod
-    def from_ptp(cls, templates_or_sorting_analyzer, threshold, noise_levels=None):
+    def from_ptp(cls, templates_or_sorting_analyzer, threshold):
```
JoeZiminski, line 360: Is it worth adding a warning here that this method of computation was changed in this version, and to use `snr` with `peak_to_peak` to replicate the old method?

src/spikeinterface/core/sparsity.py
```diff
@@ -605,43 +609,55 @@ def estimate_sparsity(
...
-    templates = Templates(
-        templates_array=templates_array,
-        sampling_frequency=recording.sampling_frequency,
-        nbefore=nbefore,
-        sparsity_mask=None,
-        channel_ids=recording.channel_ids,
-        unit_ids=sorting.unit_ids,
-        probe=probe,
-    )
+    if method != "by_property":
```
JoeZiminski, line 612: Would it be possible to add a little more to the docstring here, I don't quite understand the use for `by_property`. I am used to a property per channel e.g. "group" returns two recordings split by channel group. I can't align that in my head with e.g.  `sorting_analyzer.sorting.get_property(by_property)[unit_ind]` the property is index by unit not channel. So I am confused 😅

src/spikeinterface/core/sparsity.py
```diff
@@ -25,14 +25,16 @@
...
-        Sign of the template to compute best channels ("neg", "pos", "both")
+    peak_sign : "neg" | "pos" | "both"
+        Sign of the template to compute best channels
     num_channels : int
         Number of channels for "best_channels" method
     radius_um : float
         Radius in um for "radius" method
     threshold : float
         Threshold in SNR "threshold" method
+    snr_amplitude_mode : "extremum" | "at_index" | "peak_to_peak"
```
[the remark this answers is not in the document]
alejoe91: yep I'll do that!

src/spikeinterface/core/tests/test_sparsity.py
```diff
@@ -212,6 +240,9 @@ def test_compute_sparsity():
     sparsity = compute_sparsity(sorting_analyzer, method="best_channels", num_channels=2, peak_sign="neg")
     sparsity = compute_sparsity(sorting_analyzer, method="radius", radius_um=50.0, peak_sign="neg")
     sparsity = compute_sparsity(sorting_analyzer, method="snr", threshold=5, peak_sign="neg")
+    sparsity = compute_sparsity(
```
JoeZiminski, line 243: Usually during a review I would run the code under the different arguments to check for any regressions introduced by the refactorings (it looks good to me reading the code, but I guess you never know!). However there are a lot of different options so it might be easier just to extend these tests. It might be worth now or in a follow up PR extending these tests to check that, given a sorting analyzer with some set of fixed templates, the results of all these methods are as expected (i.e. the machinery around the computation is properly propagating the kwargs etc).

src/spikeinterface/core/sparsity.py
alejoe91: Actually, I doupt anyone except @yger used it. I wouldn't overload with an additional warning :)

src/spikeinterface/core/sparsity.py
alejoe91: so we have a `run_sorter_by_property` function. In that case, a unit is assigned a property value corresponding to the property group it was found on (e.g. `tetrode0`)

src/spikeinterface/core/sparsity.py
JoeZiminski: Okay makes sense, anyone who is particularly interested can find it on the version release notes anyways

src/spikeinterface/core/sparsity.py
alejoe91: I think it might make sense to add the run sorter by group example for sure!

src/spikeinterface/core/sparsity.py
```diff
@@ -13,28 +13,33 @@
...
-        * "radius" : radius around the best channel. Use the "radius_um" argument to specify the radius in um
+                           number of channels.
+        * "radius" : radius around the best channel. Use the "radius_um" argument to specify the radius in um.
         * "snr" : threshold based on template signal-to-noise ratio. Use the "threshold" argument
-                 to specify the SNR threshold (in units of noise levels)
+                 to specify the SNR threshold (in units of noise levels) and the "amplitude_mode" argument
+                 to specify the mode to compute the amplitude of the templates.
         * "ptp" : threshold based on the peak-to-peak values on every channels. Use the "threshold" argument
-                to specify the ptp threshold (in units of noise levels)
+                 to specify the ptp threshold (in units of amplitude).
```
samuelgarcia, line 22: Does this means that for ptp, the previous behavior was relative and now it is absolut ?
I think this is a string change for a patch release no ?

src/spikeinterface/core/sparsity.py
samuelgarcia: Yes but here it is a strong change for a patch release no ?

src/spikeinterface/core/sparsity.py
```diff
@@ -353,41 +417,47 @@ def from_snr(cls, templates_or_sorting_analyzer, threshold, noise_levels=None, p
...
         """
         Construct sparsity from a thresholds based on template peak-to-peak values.
-        Use the "threshold" argument to specify the SNR threshold.
+        Use the "threshold" argument to specify the peak-to-peak threshold.
+
+        Parameters
+        ----------
+        templates_or_sorting_analyzer : Templates | SortingAnalyzer
+            A Templates or a SortingAnalyzer object.
+        threshold : float
```
samuelgarcia, line 429: I would use threshold_uv if now it is always absolut to sure we avoid mistake

src/spikeinterface/core/sparsity.py
alejoe91: then let's release 0.102.0 ;)

I think Pierre was the sole user of this :P

src/spikeinterface/core/sparsity.py
alejoe91: well that depends on whether return scaled is true. Shoudl we throw an error if it's not so that this only works with scaled templates?

src/spikeinterface/core/sparsity.py
```diff
@@ -396,14 +466,27 @@ def from_energy(cls, sorting_analyzer, threshold):
...
         Construct sparsity from a threshold based on per channel energy ratio.
         Use the "threshold" argument to specify the SNR threshold.
+        This method requires the "waveforms" and "noise_levels" extensions to be computed.
+
+        Parameters
+        ----------
+        sorting_analyzer : SortingAnalyzer
+            A SortingAnalyzer object.
+        threshold : float
+            Threshold for "energy" method (in units of noise levels).
```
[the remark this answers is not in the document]
alejoe91: yes it's still True, but thie method requires the noise levels extension (see line 469)

samuelgarcia (approved):

samuelgarcia:
Quel talent!
merci beaucoup

PR #3372, opened 2024-09-05
Remove `from __future__ import annotations` after dropping 3.8

alejoe91:


zm711:
We have some 3.10 + only... the | is 3.10 :(

alejoe91:
My bad!

zm711:
No worries. Valiant effort. I think after 3.10 all the typing stuff has calmed down so it should be easy :)

h-mayorquin:
Numpy is already at minimal 3.10 https://numpy.org/neps/nep-0029-deprecation_policy.html x D

PR #3378, opened 2024-09-06
Ensure sorting analyzer in zarr are consolidated

alejoe91:
As in title

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -622,7 +622,7 @@ class SortingAnalyzer:
             format="zarr",
             sparsity=sparsity,
             return_scaled=return_scaled,
-            storage_options=storage_options
+            storage_options=storage_options,
         )
 
         return sorting_analyzer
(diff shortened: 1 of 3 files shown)


src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -563,11 +566,13 @@ def create_zarr(cls, folder, sorting, recording, sparsity, return_scaled, rec_at
...
         recording_info = zarr_root.create_group("extensions")
 
+        zarr.consolidate_metadata(zarr_root.store)
+
     @classmethod
     def load_from_zarr(cls, folder, recording=None, storage_options=None):
         import zarr
 
-        zarr_root = zarr.open(str(folder), mode="r", storage_options=storage_options)
+        zarr_root = zarr.open_consolidated(str(folder), mode="r", storage_options=storage_options)
```
samuelgarcia, line 575: we have several zarr.open in this module can you check then all ?

alejoe91:
@samuelgarcia this is fixed. The failing tests wa sa tricky one to find: basically we have to consolidate metadata AFTER run info is updated (and it was not the case)

alejoe91:
As discussed with @samuelgarcia, I also added the `ext.delete()` function :)

samuelgarcia (approved):

samuelgarcia:
This sounds a bit annoying to do ` zarr.consolidate_metadata` after eevry changes. No ?
mets got with this for now.

PR #3379, opened 2024-09-06
Fix time handling test memory

alejoe91:
@JoeZiminski I realized that the `test_sorting_analyzer_get_durations` test was using up 10s of GBs of RAM!

This is because the cumulative offsets ended up creating durations of over 30'000 seconds, used to generate sorting objects. I refactored the test to "manually" extend the duration with a simple linspace

### DIFF
--- a/src/spikeinterface/core/tests/test_time_handling.py
+++ b/src/spikeinterface/core/tests/test_time_handling.py
@@ -280,7 +280,6 @@ class TestTimeHandling:
         """
         _, times_recording, _ = time_vector_recording
 
-        durations = [times_recording.get_duration(s) for s in range(times_recording.get_num_segments())]
         sorting = si.generate_sorting(
             durations=[times_recording.get_duration(s) for s in range(times_recording.get_num_segments())]
         )
--- a/src/spikeinterface/core/tests/test_time_handling.py
+++ b/src/spikeinterface/core/tests/test_time_handling.py
@@ -49,19 +49,14 @@ class TestTimeHandling:
         spaced timeseries data. Return the original recording,
         recoridng with time vectors added and list including the added time vectors.
         """
-        times_recording = copy.deepcopy(raw_recording)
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/tests/test_time_handling.py
```diff
@@ -285,6 +280,7 @@ def test_sorting_analyzer_get_durations_from_recording(self, time_vector_recordi
         """
         _, times_recording, _ = time_vector_recording
 
+        durations = [times_recording.get_duration(s) for s in range(times_recording.get_num_segments())]
```
alejoe91, line 283:
```suggestion
```

samuelgarcia (approved):

JoeZiminski:
****! good spot! sorry about that

The purpose of that bit of code was to generate an unevenly spaced time series, just to check this case is tested as I think this might be possible in a real recording , i.e. `np.diff(time_vector)` should not all be identical. One much neater way to generate this would be:

```
exp_increasing = np.linspace(0, 1, raw_recording.get_num_samples(segment_index)) ** 2
time_vector = exp_increasing * t_stop + t_start
```
but this only creates spacing offsets which differ by around 9 decimal places. I think the time of the recording was increased so much to get these spacings a little larger, but didn't realise of course it would create such huge sortings! Also it's pointless to do so because any small offset in a true time vector is going to be to high decimal place anyway.

I think its best to leave it as is now and have an explicit test for unevenly spaced time series, as it's quite confusing testing this implicitly by setting it up like this in the test data generation. I'll make an issue to add such a test this test if that makes sense.

alejoe91:
Thanks @JoeZiminski 

Yeah I agree it's better to add a specific test. I was about confused by the implementation so it needed a bit of doggimg to figure out why the test was crashing my laptop :)

JoeZiminski:
Great that sounds good, apologies for the oversight / laptop crashing 😅

Issue #3324, opened 2024-08-21
Inconsistent handling of time offsets in plot_traces


jonahpearl:
Hi all — I was doing some manual digging around to deal with artifacts caused by some custom lighting in our rigs, and I wanted to find the moment in the recording where the artifacts began. I noticed that depending on whether I used `recording.time_slice()` or `recording.plot_traces`, there are two different behaviors for dealing with the case where t0 != 0 (this is coming from open ephys, so it's often the case that the first timestamp isn't 0 in the recording).

Case 1: the "give me X many seconds into the recording" strategy. This appears to be used by `recording.plot_traces`, and I think is a bug.

Case 2: the "give me the data where the timestamps equal this value" strategy. This appears to be used by `recording.time_slice`, and I think is the correct way (?).

For example, here is the plot from `plot_traces` for the time range (25.5, 26.5): you can see the artifacts start ~halfway through the window.
<img width="725" alt="image" src="https://github.com/user-attachments/assets/aa7c9c75-e109-4d6d-8fa9-38414856f29c">

Now if I run this code to try to show the same thing but with `recording.time_slice()`, we instead get a different moment: 
```python
tmp_rec = recording_car.time_slice(25.5, 26.5)
t = tmp_rec.get_traces(channel_ids=['CH40'])
plt.figure(figsize=(10,3))
plt.plot(t, lw=0.5)
```
![image](https://github.com/user-attachments/assets/3f6bf379-6309-4068-b478-9ee096f92016)

If we adjust the time slicing to account for the inconsistent behavior, we recover the same moment in the data:
```python
t0 = recording_car.get_times()[0]
tmp_rec = recording_car.time_slice(25.5 + t0, 26.5 + t0)  # this will get data from where the timestamp equals 25.5 + t0
t = tmp_rec.get_traces(channel_ids=['CH40'])
plt.figure(figsize=(10,3))
plt.plot(t, lw=0.5)
```
![image](https://github.com/user-attachments/assets/c386624e-9863-482d-9801-3b25758a647e)

[...]

PR #3380, opened 2024-09-06
Reset-times: segment must have either time vector or sampling frequency

alejoe91:
Fixes a bug in reset times that causes issues downstream, since rec segments must have either a `sampling_frequency` or a `time_vector`

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -506,6 +506,7 @@ class BaseRecording(BaseRecordingSnippets):
                 rs = self._recording_segments[segment_index]
                 rs.t_start = None
                 rs.time_vector = None
+                rs.sampling_frequency = self.sampling_frequency
 
     def sample_index_to_time(self, sample_ind, segment_index=None):
         """
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -497,16 +497,17 @@ class BaseRecording(BaseRecordingSnippets):
 
     def reset_times(self):
         """
-        Reset times in-memory for all segments that have a time vector.
+        Reset time information in-memory for all segments that have a time vector.
         If the timestamps come from a file, the files won't be modified. but only the in-memory
-        attributes of the recording objects are deleted.
+        attributes of the recording objects are deleted. Also `t_start` is set to None and the
+        segment's sampling frequency is set to the recording's sampling frequency.
         """
         for segment_index in range(self.get_num_segments()):
             if self.has_time_vector(segment_index):
                 rs = self._recording_segments[segment_index]
-                rs.t_start = None
                 rs.time_vector = None
-                rs.sampling_frequency = self.sampling_frequency
+            rs.t_start = None
+            rs.sampling_frequency = self.sampling_frequency
 
     def sample_index_to_time(self, sample_ind, segment_index=None):
         """
--- a/src/spikeinterface/core/tests/test_baserecording.py
+++ b/src/spikeinterface/core/tests/test_baserecording.py
@@ -292,7 +292,11 @@ def test_BaseRecording(create_cache_folder):
     # reset times
     rec.reset_times()
     for segm in range(num_seg):
+        time_info = rec.get_time_info(segment_index=segm)
         assert not rec.has_time_vector(segment_index=segm)
+        assert time_info["t_start"] is None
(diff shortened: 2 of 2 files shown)


zm711:
I don't really touch the time vector stuff (I work in sample for everything). Should there be some additional testing with the time vector stuff to prevent this? Or are these some one off issues with the latest changes?

alejoe91:
The problem arose in a preprocessing step. Basically a segment must always have either a time vector or a sampling frequency, otherwise instantiation fail

JoeZiminski:
I think similar to #3324, if a `self.t_start` is set (but no time vector) this will not be reset? Should it be `if self.has_time_vector(segment_index) or self.t_start is not None`?

Out of interest, in what cases can a segment have a time vector but no sampling frequency set? is the sampling frequency not always propagated from parent recording to segment?

alejoe91:
> I think similar to #3324, if a `self.t_start` is set (but no time vector) this will not be reset? Should it be `if self.has_time_vector(segment_index) or self.t_start is not None`?

I think that if we set the time vector, then the `segment.t_start` should be forced to `None`, since they are mutually exclusive ways to set times. 

> Out of interest, in what cases can a segment have a time vector but no sampling frequency set? is the sampling frequency not always propagated from parent recording to segment?

The segment works in samples, so time information can be regarded as metadata. The `sampling_frequency` instead needs to be specified at the recording level.

JoeZiminski:
Cool thanks I did not know that about sampling frequency!

I think if a time vector has been explicitly set by the user then that is the case. But it is possible that `t_start` is set in some extractors, and then [`get_times()`](https://github.com/SpikeInterface/spikeinterface/blob/0503867fa6fd264e718a9eeccbfc0a3ab6511ef6/src/spikeinterface/core/baserecording.py#L828) will generate the `time_vector` ad hoc but not save it to the recording. In this case you could have something like:

```
recording.get_times()  # 10, 10.001, ...., 11 (because recording.t_start = 10)
recording.reset_times()
recording.get_times() # # 10, 10.001, ...., 11
```

alejoe91:
I see, got it! So then the `reset_times` should also have a `segment.t_start = None`?

JoeZiminski:
I think the `t_start` on the segment is set to `None` in `reset_times()`, but the only problem is that this conditional will not be triggered in the case that `t_start` is set but not `time_vector`. And so even if the `t_start` is set, it will not be reset on `reset_times()`. I think it would have to be like:

```
for segment_index in range(self.get_num_segments()):

    rs = self._recording_segments[segment_index]
    if self.has_time_vector(segment_index) or rs.t_start is not None:
        rs.t_start = None
        rs.time_vector = None
        rs.sampling_frequency = self.sampling_frequency
```

Maybe there is a more elegant way to do this, I'm also not sure how the `t_start` case interacts with the `rs.sampling_frequency` case. But this will definitely make sure that times are reset even when only a `t_start` is on the segment but no `time_vector`.

I am wondering a bit more about the `sampling_frequency` on the segment itself, to me it's a bit confusing. My understanding is that it is enforced that all segments should have the same sampling frequency as the parent recording. A  segment in will always have a baseline sampling frequency inherited from the recording regardless of whether a specific time vector has been set. And this sampling frequency will be the same as all the other segments (?). So for me it would be more intuitive for segments to be given a generic `get_sampling_frequency()` that just returns the parent recording sampling frequency, rather than each segment having its own optional attribute which implies the sampling frequencies can be different across segments, or somehow that a segment does not even have a sampling frequency. (sorry, this is somewhat outside the scope of this PR, if this makes any sense I can make an issue).

alejoe91:
Got it! Then the `reset_times()` should always also set `t_start=None`, regardless of the segment having a time vector.

As for the `sampling_frequency`, note that the segment doesn't have any reference to the parent recording, so `sampling_frequency` must be passed externally. 
Currently, we check that either a `sampling_frequency` or `time_vector` is passed, since these are the two alternative ways to convert samples/times and viceversa. The segment itself only works in the samples space to retrieve traces

samuelgarcia (approved):

JoeZiminski:
@alejoe91 @samuelgarcia  I think this [current implementation](https://github.com/alejoe91/spikeinterface/blob/a3fef7055d3702481cee99f17821f910b84921c3/src/spikeinterface/core/baserecording.py#L506)  will error out if `self.has_time_vector` is `False` right? because `rs` will not be defined

PR #3388, opened 2024-09-10
Fix integer overflow of the total sample number when concatenating long recordings

jiumao2:
Fix integer overflow of the total sample number when concatenating long recordings.

image
The bug occurs when concatenating recordings exceeding 24 hours in total length at a 30000 Hz sampling rate, leading to integer overflow in `np.sum` function.

Simply adding `dtype=np.int64` in `np.sum` resolves this bug.
image

### DIFF
--- a/src/spikeinterface/core/segmentutils.py
+++ b/src/spikeinterface/core/segmentutils.py
@@ -157,7 +157,7 @@ class ProxyConcatenateRecordingSegment(BaseRecordingSegment):
         self.parent_segments = parent_segments
         self.all_length = [rec_seg.get_num_samples() for rec_seg in self.parent_segments]
         self.cumsum_length = [0] + [sum(self.all_length[: i + 1]) for i in range(len(self.all_length))]
-        self.total_length = int(np.sum(self.all_length, dtype=np.int64))
+        self.total_length = int(sum(self.all_length))
 
     def get_num_samples(self):
         return self.total_length
@@ -450,7 +450,7 @@ class ProxyConcatenateSortingSegment(BaseSortingSegment):
         self.parent_segments = parent_segments
         self.parent_num_samples = parent_num_samples
         self.cumsum_length = np.cumsum([0] + self.parent_num_samples)
-        self.total_num_samples = np.sum(self.parent_num_samples, dtype=np.int64)
+        self.total_num_samples = int(sum(self.parent_num_samples))
 
     def get_num_samples(self):
         return self.total_num_samples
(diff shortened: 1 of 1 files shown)


alejoe91:
Thanks! @h-mayorquin has been spotting a bunch of these recently, so let's wait for him to take a look!

src/spikeinterface/core/segmentutils.py
```diff
@@ -157,7 +157,7 @@ def __init__(self, parent_segments, sampling_frequency, ignore_times=True):
         self.parent_segments = parent_segments
         self.all_length = [rec_seg.get_num_samples() for rec_seg in self.parent_segments]
         self.cumsum_length = [0] + [sum(self.all_length[: i + 1]) for i in range(len(self.all_length))]
-        self.total_length = int(np.sum(self.all_length))
+        self.total_length = int(np.sum(self.all_length, dtype=np.int64))
```
zm711, line 160: The issue with this as a fix would be that if we are doing 10 day recordings in the future then this still might fail. Happy to hear what @h-mayorquin things, but something like:

```
int(sum(list(all_length)))
```
provides us the protection of python integers which shouldn't overflow. We get a performance hit since we won't be working at the numpy vectorized speed, but I think overflows are worse. And overflow behavior is different between OSs < 2.0 (I forget the issue I found that Heberto and I discussed). Hopefully Heberto has a more performant solution than mine though.

h-mayorquin:
Yep, using int64 or uint64 is good but it would be better to use python native int type. I suggest doing the latter as it would be more robust for the reasons that @zm711 mentions.

src/spikeinterface/core/segmentutils.py
jiumao2: That's definitely a better solution for robustness!

jiumao2:
New commit made as @zm711 suggested!

h-mayorquin (approved):
Thanks for your contributions. 

We should open an issue to write tests for this. It seems easy to forget why it was done this way, change it and then introduce a regression.

@JoeZiminski

h-mayorquin:
Also, sorry for citing you without adding context. What I forgot to say is that we have had some disagreement in the past about whether to test the gain multiplication for really high values and low values but in the subject of regression testing I am totally in your side and I think this is a good case.

JoeZiminski:
Hey @h-mayorquin thanks for this! I agree this is a good case. I thought more about the subject of testing extreme values for a while and still can't decide if its a good or bad idea 😆 but I'm inclined to agree with you on that too.  BTW happy to chat at a maintainers meeting about deciding on the general testing approach like you mentioned that time, would be very valuable. I think like you said the key concern of testing is to ensure no regressions slip into the code base!

Issue #3383, opened 2024-09-07
Sorting analyzer behavior with multi probe recording


RobertoDF:
Hi,
I have a problem with the following pipeline:
I have a 2 NPX1 recording.
I sort each probe individually.
I curate each one in phy.
I load each sorting with read_phy and aggregate them with aggregate_units.
I compute extensions on this aggregated sorting.

The waveforms from this aggregated sorting analyzer are, however, off. If I plot them using plot_units_waveforms, I see four columns of waveforms, two columns are fine, two columns are flat lines (I'll post a pic later). Notably, if I plot using a sorting_analyzer created from an individual sorting (not the aggregated one) waveforms are fine. I guess the waveform extraction is not noticing the probe group of each channel and picking the channels corresponding to the peak template in both probes? Not sure though.

Should I use the sorting analyzer on not aggregated sorting objects only?if yes shouldn't there be some correspondent method to aggregate_units for the sorting analyzers too?

Thanks!

zm711:
I think the picture would help. I also sort with probe groups and I haven't noticed this problem before. The sparsity should take care of this I believe. Are you creating your analyzer with sparsity = False to get dense info?

RobertoDF:
They look all like this: 
<img width="417" alt="Screenshot 2024-09-07 at 23 50 51" src="https://github.com/user-attachments/assets/cd709184-057f-4d6d-8878-b88ecdeae9f8">
<img width="329" alt="Screenshot 2024-09-07 at 23 51 37" src="https://github.com/user-attachments/assets/8ed0da99-1e54-4eea-8973-16d943506a38">

RobertoDF:
this is the code I use, I don´t explicitly set sparsity: 
```
sortings = []
for probe_n in tqdm(range(len(raw_rec.get_probes()))):
            sortings.append(read_phy(Path(f"{path_recording_folder}/spike_interface_output/probe{probe_n}/sorter_output")))

 sorting = aggregate_units(sortings)

analyzer = create_sorting_analyzer(sorting, raw_rec, sparse=True, format="binary_folder",
                                           folder=f"{path_recording_folder}/spike_interface_output/sorting_analyzer_post_curation",
                                           overwrite=True, **job_kwargs)
```

zm711:
How are you sorting to start with? What is the script for sorting to get to the phy data step? Are you doing run_sorter_by_property or something else?

alejoe91:
[...]

PR #3389, opened 2024-09-10
solving  #3383

RobertoDF:
This solves my problem #3383 in which the channel indices are scrambled. It shouldn't impact other cases. but more testing is needed I guess.

### DIFF
--- a/src/spikeinterface/core/baserecordingsnippets.py
+++ b/src/spikeinterface/core/baserecordingsnippets.py
@@ -357,7 +357,12 @@ class BaseRecordingSnippets(BaseExtractor):
                 # check that multiple probes are non-overlapping
                 check_probe_do_not_overlap(all_probes)
                 all_positions = np.vstack([probe.contact_positions for probe in all_probes])
-                positions = all_positions[channel_indices]
+                probes_channel_indices = np.concatenate([probe.device_channel_indices for probe in all_probes])
+
+                sorted_probe_idx = np.argsort(probes_channel_indices)
+                sorted_positions_idx = np.searchsorted(probes_channel_indices[sorted_probe_idx], channel_indices)
+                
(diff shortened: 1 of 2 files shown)


h-mayorquin:
Can we have a test for this? I would illustrate the problem and serve to avoid regressions. We can use generate_probe I guess.

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -357,7 +357,12 @@ def get_channel_locations(self, channel_ids=None, axes: str = "xy"):
                 # check that multiple probes are non-overlapping
                 check_probe_do_not_overlap(all_probes)
                 all_positions = np.vstack([probe.contact_positions for probe in all_probes])
-                positions = all_positions[channel_indices]
+                probes_channel_indices = np.concatenate([probe.device_channel_indices for probe in all_probes])
+
+                sorted_probe_idx = np.argsort(probes_channel_indices)
+                sorted_positions_idx = np.searchsorted(probes_channel_indices[sorted_probe_idx], channel_indices)
+
+                positions = all_positions[sorted_probe_idx[sorted_positions_idx]]
```
samuelgarcia, line 365: OK for me but this is really strange because when attached to a recording the device_channel_indices should always be sorted and should be =[0...n-1].
Maybe there is something deeper in the bug.

samuelgarcia:
Hi @RobertoDF 
After discussing with Alessio.
I think we found the deep bug.
I will propose another PR ti fix channel locations when we have a probegroup.

samuelgarcia:
I did another PR here #3392

the problem was deeper I think and I prefer the other fix because normally the contact_vector is always ordered.

RobertoDF:
solved by #3392

PR #3392, opened 2024-09-11
Fix proposal for channel location when probegroup

samuelgarcia:
This should solve #3389 and replace #3389

@RobertoDF : could you check on your side that it solve the original problem.

This is untested at the moment, if this works I will push some tests and continue to improve this way.

EDIT:

@JoeZiminski 
Here the problem:
 1. Having a `Probe` is similar to having the `contact_vector` respresentation
 2. in a recording the `contact_vector`  is always sorted so that `device_channel_indices` = np.arange(num_channels)
 3. For the ProbeGroup this is the same only and only the channel of Probes are NOT interlaved otherwise 
     having the `ProbeGroup` (which is a list Probe) is different from having the full `contact_vector` because the 
     ProbeGroup do not ensue th step 2. when channels of probes of interleaved

### DIFF
--- a/src/spikeinterface/core/baserecordingsnippets.py
+++ b/src/spikeinterface/core/baserecordingsnippets.py
@@ -348,17 +348,22 @@ class BaseRecordingSnippets(BaseExtractor):
         if channel_ids is None:
             channel_ids = self.get_channel_ids()
         channel_indices = self.ids_to_indices(channel_ids)
-        if self.get_property("contact_vector") is not None:
-            if len(self.get_probes()) == 1:
-                probe = self.get_probe()
-                positions = probe.contact_positions[channel_indices]
-            else:
-                all_probes = self.get_probes()
-                # check that multiple probes are non-overlapping
-                check_probe_do_not_overlap(all_probes)
-                all_positions = np.vstack([probe.contact_positions for probe in all_probes])
-                positions = all_positions[channel_indices]
-            return select_axes(positions, axes)
+        contact_vector = self.get_property("contact_vector")
+        if contact_vector is not None:
+            # to avoid the get_probes() when only one probe do check unique probe_id
+            num_probes = np.unique(contact_vector["probe_index"]).size
+            if num_probes > 1:
+                # get_probes() is called only when several probes check_overlaps
+                # TODO make this check_probe_do_not_overlap() use only the contact_vector instead of constructing the probe
+                check_probe_do_not_overlap(self.get_probes())
+
+            # here we bypass the probe reconstruction so this works both for probe and probegroup
+            ndim = len(axes)
+            all_positions = np.zeros((contact_vector.size, ndim), dtype="float64")
+            for i, dim in enumerate(axes):
+                all_positions[:, i] = contact_vector[dim]
+            positions = all_positions[channel_indices]
+            return positions
         else:
             locations = self.get_property("location")
             if locations is None:
--- a/src/spikeinterface/core/baserecordingsnippets.py
+++ b/src/spikeinterface/core/baserecordingsnippets.py
@@ -239,7 +239,7 @@ class BaseRecordingSnippets(BaseExtractor):
 
         warning_msg = (
             "`set_probes` is now a private function and the public function will be "
-            "removed in 0.103.0. Please use `set_probe` or `set_probegroups` instead"
+            "removed in 0.103.0. Please use `set_probe` or `set_probegroup` instead"
         )
 
         warn(warning_msg, category=DeprecationWarning, stacklevel=2)
--- a/src/spikeinterface/core/baserecordingsnippets.py
+++ b/src/spikeinterface/core/baserecordingsnippets.py
@@ -145,6 +145,11 @@ class BaseRecordingSnippets(BaseExtractor):
         else:
             raise ValueError("must give Probe or ProbeGroup or list of Probe")
 
+        # check that the probe do not overlap
+        num_probes = len(probegroup.probes)
+        if num_probes > 1:
+            check_probe_do_not_overlap(probegroup.probes)
+
         # handle not connected channels
         assert all(
             probe.device_channel_indices is not None for probe in probegroup.probes
@@ -350,13 +355,6 @@ class BaseRecordingSnippets(BaseExtractor):
         channel_indices = self.ids_to_indices(channel_ids)
         contact_vector = self.get_property("contact_vector")
         if contact_vector is not None:
-            # to avoid the get_probes() when only one probe do check unique probe_id
-            num_probes = np.unique(contact_vector["probe_index"]).size
-            if num_probes > 1:
-                # get_probes() is called only when several probes check_overlaps
-                # TODO make this check_probe_do_not_overlap() use only the contact_vector instead of constructing the probe
-                check_probe_do_not_overlap(self.get_probes())
-
             # here we bypass the probe reconstruction so this works both for probe and probegroup
             ndim = len(axes)
             all_positions = np.zeros((contact_vector.size, ndim), dtype="float64")
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -1102,7 +1102,7 @@ class SortingAnalyzer:
         # important note : contrary to recording
         # this give all channel locations, so no kwargs like channel_ids and axes
         probegroup = self.get_probegroup()
-        probe_as_numpy_array = probegroup.to_numpy()
+        probe_as_numpy_array = probegroup.to_numpy(complete=True)
         # we need to sort by device_channel_indices to ensure the order of locations is correct
         probe_as_numpy_array = probe_as_numpy_array[np.argsort(probe_as_numpy_array["device_channel_indices"])]
         ndim = probegroup.ndim
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -1101,9 +1101,14 @@ class SortingAnalyzer:
     def get_channel_locations(self) -> np.ndarray:
         # important note : contrary to recording
         # this give all channel locations, so no kwargs like channel_ids and axes
-        all_probes = self.get_probegroup().probes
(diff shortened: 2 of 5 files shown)


samuelgarcia:
@zm711 @h-mayorquin @alejoe91 : 
* we should make the `check_probe_do_not_overlap` when `set_probe_group()` and not at every
   `get_channel_locations()` I do not remember I/we did this.
* this PR try to skip `get_probe()` because it is slower we should do it at several places
* the bug was due to the fact that you can have several channels that are interleaved in the contact vector.
 I think that when I wrote the original implementation in my mind several probe was consecutive blocks on channel and not interleaved

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -348,17 +348,22 @@ def get_channel_locations(self, channel_ids=None, axes: str = "xy"):
...
-                positions = all_positions[channel_indices]
-            return select_axes(positions, axes)
+        contact_vector = self.get_property("contact_vector")
+        if contact_vector is not None:
+            # to avoid the get_probes() when only one probe do check unique probe_id
+            num_probes = np.unique(contact_vector["probe_index"]).size
+            if num_probes > 1:
+                # get_probes() is called only when several probes check_overlaps
+                # TODO make this check_probe_do_not_overlap() use only the contact_vector instead of constructing the probe
+                check_probe_do_not_overlap(self.get_probes())
```
alejoe91, line 358: Then let's move this to `_set_probes`? :)

RobertoDF:
I ll test it now

RobertoDF:
dont we also need to modify the get_channel_locations method in sorting_analyzer .py?

RobertoDF:
Everything works if I modify the sortinganalyzer method as in my pull request
Capture

alejoe91:
Good catch @RobertoDF !

Yes Sam can you modify the logic [here](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/sortinganalyzer.py#L1101-L1106) as well?

RobertoDF:
Do we have access to contact vector fromn the sorting_analyzer though?

alejoe91:
> Do we have access to contact vector fromn the sorting_analyzer though?

No, we need a different strategy. I think that something like `probegroup.to_numpy()["x"]/["y"]` should work, since it's sorted

alejoe91:
@RobertoDF I think this is fixed with the latest commit. Can you test again?

(Thanks for the help BTW!)

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1101,9 +1101,14 @@ def get_probe(self):
     def get_channel_locations(self) -> np.ndarray:
         # important note : contrary to recording
         # this give all channel locations, so no kwargs like channel_ids and axes
-        all_probes = self.get_probegroup().probes
-        all_positions = np.vstack([probe.contact_positions for probe in all_probes])
-        return all_positions
+        probegroup = self.get_probegroup()
+        probe_as_numpy_array = probegroup.to_numpy()
+        # duplicate positions to "locations" property
```
alejoe91, line 1106:
```suggestion
```

RobertoDF:
Somehow the order of contact vector and probe_as_numpy_array is different so the waveforms are still messed up
Capture

alejoe91:
right, because thr probe group is constructed probe by probe...then we need something similar to your initial fix there :) give me 2 minutes

alejoe91:
let's see now :)

RobertoDF:
```
File ~\miniforge3\envs\spikesorting\lib\site-packages\spikeinterface\core\sortinganalyzer.py:1107, in SortingAnalyzer.get_channel_locations(self)
   1105 probe_as_numpy_array = probegroup.to_numpy()
   1106 # we need to sort by device_channel_indices to ensure the order of locations is correct
-> 1107 probe_as_numpy_array = probe_as_numpy_array[np.argsort(probe_as_numpy_array["device_channel_indices"])]
   1108 ndim = probegroup.ndim
   1109 locations = np.zeros((probe_as_numpy_array.size, ndim), dtype="float64")

ValueError: no field of name device_channel_indices
```

I think we should access that by probe? `probes_channel_indices = np.concatenate([probe.device_channel_indices for probe in all_probes])`

alejoe91 (approved):

JoeZiminski:
Cool this looks very interesting! Is it possible to update the PR message which a couple of lines to explain: i) what the problem is and ii) what the implemented solution is? I'm looking through the various issues and PRs but its a bit hard to follow and this looks super-important to be aware of. Thanks!

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1101,9 +1101,15 @@ def get_probe(self):
     def get_channel_locations(self) -> np.ndarray:
         # important note : contrary to recording
         # this give all channel locations, so no kwargs like channel_ids and axes
-        all_probes = self.get_probegroup().probes
-        all_positions = np.vstack([probe.contact_positions for probe in all_probes])
-        return all_positions
+        probegroup = self.get_probegroup()
+        probe_as_numpy_array = probegroup.to_numpy(complete=True)
+        # we need to sort by device_channel_indices to ensure the order of locations is correct
+        probe_as_numpy_array = probe_as_numpy_array[np.argsort(probe_as_numpy_array["device_channel_indices"])]
```
samuelgarcia, line 1107: This correct but we have a internal problem that should be solved about this.
Becasue the contact_vector should be always ordered I think.
In recording this is the case but no in analyzer because we have ProbeGroup instead of the contact_vector representation

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1101,9 +1101,15 @@ def get_probe(self):
...
+        ndim = probegroup.ndim
+        locations = np.zeros((probe_as_numpy_array.size, ndim), dtype="float64")
+        for i, dim in enumerate(["x", "y", "z"][:ndim]):
+            locations[:, i] = probe_as_numpy_array[dim]
+        return locations
```
samuelgarcia, line 1112: we should clip to 2 dimensions with a `axes="xy"` like we do in the recording

zm711:
> Is it possible to update the PR message which a couple of lines to explain: i) what the problem is and ii) what the implemented solution is? I'm looking through the various issues and PRs but it's a bit hard to follow and this looks super-important to be aware of. Thanks!

Just historically, when the analyzer was first created it was designed only to work with probes and support for probe groups was my request. So I think the original design was hacking on the original probe only design. So this PR is important for improving and refining probe group support. This doesn't fully address your specific request, but it is the original context of what may have contributed to this issue.

h-mayorquin:
Regression test please. Otherwise we are stuck with explanations for code that say "we vaguely remember this fixed something and I don't want to change it" when someone asks.

samuelgarcia:
> Regression test please. Otherwise we are stuck with explanations for code that say "we vaguely remember this fixed something and I don't want to change it" when someone asks.

I will do it very very soon. Sure. We are not anarchist. (apart for some of us)

h-mayorquin:
Your anarchism is the the capitalist one, man, mine is the Bakunian where we die discussing in long unending assemblies where nothing is decided.

src/spikeinterface/core/sortinganalyzer.py
alejoe91: Fixed

alejoe91:
> > Regression test please. Otherwise we are stuck with explanations for code that say "we vaguely remember this fixed something and I don't want to change it" when someone asks.
> 
> I will do it very very soon. Sure. We are not anarchist. (apart for some of us)



> > Regression test please. Otherwise we are stuck with explanations for code that say "we vaguely remember this fixed something and I don't want to change it" when someone asks.
> 
> I will do it very very soon. Sure. We are not anarchist. (apart for some of us)

@h-mayorquin I added tests :)

src/spikeinterface/core/tests/test_sortinganalyzer.py
```diff
@@ -178,6 +178,27 @@ def test_SortingAnalyzer_tmp_recording(dataset):
         sorting_analyzer.set_temporary_recording(recording_sliced)
 
 
+def test_SortingAnalyzer_interleaved_probegroup(dataset):
```
h-mayorquin, line 181: Thanks.

h-mayorquin (commented):
lgtm

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1101,9 +1101,16 @@ def get_probe(self):
     def get_channel_locations(self) -> np.ndarray:
         # important note : contrary to recording
         # this give all channel locations, so no kwargs like channel_ids and axes
-        all_probes = self.get_probegroup().probes
-        all_positions = np.vstack([probe.contact_positions for probe in all_probes])
-        return all_positions
+        probegroup = self.get_probegroup()
+        probe_as_numpy_array = probegroup.to_numpy(complete=True)
+        # we need to sort by device_channel_indices to ensure the order of locations is correct
```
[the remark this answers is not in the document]
alejoe91: same order as channel ids is correct :)

src/spikeinterface/core/tests/test_sortinganalyzer.py
```diff
@@ -178,6 +178,27 @@ def test_SortingAnalyzer_tmp_recording(dataset):
...
+    probe1 = generate_linear_probe(num_elec=num_channels // 2, ypitch=20.0)
+    probe2 = probe1.copy()
+    probe2.move([100.0, 100.0])
+
+    probegroup = ProbeGroup()
+    probegroup.add_probe(probe1)
+    probegroup.add_probe(probe2)
+    probegroup.set_global_device_channel_indices(np.random.permutation(num_channels))
+
+    recording.set_probegroup(probegroup)
```
samuelgarcia, line 195:
```suggestion
    recording = recording.set_probegroup(probegroup)
```

samuelgarcia:
> Your anarchism is the the capitalist one, man, mine is the Bakunian where we die discussing in long unending assemblies where nothing is ever decided.

I am under attack.

PR #3393, opened 2024-09-11
Correct handling of time in plot_traces

### LINKED ISSUE #3324: Inconsistent handling of time offsets in plot_traces

alejoe91:
Fixes #3324

### DIFF
--- a/src/spikeinterface/widgets/traces.py
+++ b/src/spikeinterface/widgets/traces.py
@@ -125,8 +125,10 @@ class TracesWidget(BaseWidget):
 
         if not rec0.has_time_vector(segment_index=segment_index):
             times = None
-            t_start = 0
-            t_end = rec0.get_duration(segment_index=segment_index)
+            t_start = rec0.sample_index_to_time(0, segment_index=segment_index)
+            t_end = rec0.sample_index_to_time(
+                rec0.get_num_samples(segment_index=segment_index), segment_index=segment_index
+            )
         else:
             times = rec0.get_times(segment_index=segment_index)
             t_start = times[0]
@@ -673,13 +675,17 @@ def _get_trace_list(recordings, channel_ids, time_range, segment_index, return_s
         assert all(
             rec.has_scaleable_traces() for rec in recordings.values()
         ), "Some recording layers do not have scaled traces. Use `return_scaled=False`"
+    frame_range = np.array(
+        [
+            rec0.time_to_sample_index(time_range[0], segment_index=segment_index),
+            rec0.time_to_sample_index(time_range[1], segment_index=segment_index),
+        ]
+    )
     if times is not None:
-        frame_range = np.searchsorted(times, time_range)
         times = times[frame_range[0] : frame_range[1]]
     else:
-        frame_range = (time_range * fs).astype("int64", copy=False)
-        a_max = rec0.get_num_frames(segment_index=segment_index)
-        frame_range = np.clip(frame_range, 0, a_max)
+        num_samples = rec0.get_num_samples(segment_index=segment_index)
+        frame_range = np.clip(frame_range, 0, num_samples)
         time_range = frame_range / fs
         times = np.arange(frame_range[0], frame_range[1]) / fs
 


src/spikeinterface/core/frameslicerecording.py
```diff
@@ -34,7 +34,9 @@ def __init__(self, parent_recording, start_frame=None, end_frame=None):
         if start_frame is None:
             start_frame = 0
         else:
-            assert 0 <= start_frame < parent_size
+            assert (
+                0 <= start_frame < parent_size
+            ), f"'start_frame' must be fewer than number of samples in parent: {parent_size}"
```
zm711, line 39:
```suggestion
            ), f"`start_frame` must be fewer than number of samples in parent: {parent_size}"
```

src/spikeinterface/widgets/traces.py
```diff
@@ -149,6 +151,11 @@ def __init__(
             )
             time_range[1] = t_end
 
+        if time_range[0] < t_start or time_range[1] < t_start:
+            raise ValueError(f"All time_range values must be greater than {t_start}")
```
zm711, line 155:
```suggestion
            raise ValueError(f"All `time_range` values must be greater than {t_start}")
```

src/spikeinterface/widgets/traces.py
```diff
@@ -149,6 +151,11 @@ def __init__(
...
+        if time_range[1] <= time_range[0]:
+            raise ValueError("time_range[1] must be greater than time_range[0]")
```
zm711, line 157:
```suggestion
            raise ValueError("`time_range[1]` must be greater than `time_range[0]`")
```
maybe not ticks? I guess this is an edge case where it is indexing of the variable and not the variable itself?

src/spikeinterface/widgets/traces.py
```diff
@@ -673,13 +680,17 @@ def _get_trace_list(recordings, channel_ids, time_range, segment_index, return_s
         assert all(
             rec.has_scaleable_traces() for rec in recordings.values()
         ), "Some recording layers do not have scaled traces. Use `return_scaled=False`"
+    frame_range = np.array(
```
zm711, line 683: Optionally, although I'm super curious about the history of this. We could start moving some of these internal variables over to sample instead of frame. Just for consistency. The function says `time_to_sample_index`. I think most people get the frame-sample equivalence, but it might be good to try to converge a bit more on one term for consistency. Trickiest thing is that the public function is `frame_slice` instead of `sample_slice`

zm711 (commented):
Optional tick-fying and one discussion point that is not completely in scope of PR, so let me know if you want to move it to an issue.

src/spikeinterface/core/frameslicerecording.py
jonahpearl, line 39: This is a bit confusing in the case where start_frame is less than 0. It should just be two different checks:
```
assert 0 <= start_frame, "Requested `start_frame` is negative (requested timestamp may be earlier than the first timestamp in the recording)"
assert start_frame < parent_size, f"`start_frame` must be fewer than number of samples in parent: {parent_size}"
```

zm711: I thought about saying this too, but I vaguely remember we discussed (and documented) not allowing negative slicing so I wasn't sure if this point was moot or not. It can't hurt to be more explicit though.

jonahpearl: I get this error trying to run `recording_car.time_slice(0,1)` when the recording starts around t0=15 seconds, I'm just assuming it's giving a negative start_frame. If it's somehow failing with start_frame > parent _size, that seems like a different issue

zm711: No I agree that they are different problems. Although your error should likely be handled in the time_slice function since saying you have negative frames/samples isn't super helpful. @h-mayorquin wrote the time_slice right? so maybe that should be dealt with at the time level instead? What do you think Alessio and Heberto? If you give a bad time you should just know it right away in my opinion rather than getting a nonsensical frame?

h-mayorquin: Dividing the asssertion in two each one with a message revealing what is the problem is good. Having specific assertions with proper messages in the time_slice is better. I also think  all of that can be done in another PR as it is not really central to this.

JoeZiminski:
It would be super nice to have tests for this as it's a little confusing as there are a lot of possible use cases to keep in mind. I think there was some discussion to avoid adding tests for widgets until a big widget refactor that was coming. Maybe we can make an issue to track these tests-to-add, otherwise adding tests now would be useful as they will catch regressions introduced in the impending refactor!

src/spikeinterface/core/frameslicerecording.py
zm711: Agreed. Thanks Heberto.

JoeZiminski (changes requested):
Apologies, I wrote this comment ages ago and then did not publish the review!

src/spikeinterface/widgets/traces.py
```diff
@@ -125,8 +125,10 @@ def __init__(
 
         if not rec0.has_time_vector(segment_index=segment_index):
             times = None
-            t_start = 0
-            t_end = rec0.get_duration(segment_index=segment_index)
+            t_start = rec0.sample_index_to_time(0, segment_index=segment_index)
```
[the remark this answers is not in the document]
alejoe91: This was done to avoid loading the time vector if not needed. In the `_get_traces_list`, only a local time vector is generated on the fly

alejoe91:
this was fixed elsewhere

PR #3403, opened 2024-09-12
Add `get_channel_locations` to the base recording api

h-mayorquin:
This will make the method appear on the BaseRecording API on the docs.

Plus, I added a docstring to the method. 

No changes on behavior with this diff.

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -746,6 +746,30 @@ class BaseRecording(BaseRecordingSnippets):
 
         return SelectSegmentRecording(self, segment_indices=segment_indices)
 
+    def get_channel_locations(
+        self,
+        channel_ids: list | np.ndarray | tuple | None = None,
+        axes: "xy" | "yz" | "xz" = "xy",
+    ) -> np.ndarray:
+        """
+        Get the physical locations of specified channels.
+
+        Parameters
+        ----------
+        channel_ids : array-like, optional
+            The IDs of the channels for which to retrieve locations. If None, retrieves locations
+            for all available channels. Default is None.
+        axes : str, optional
+            The spatial axes to return, specified as a string (e.g., "xy", "xyz"). Default is "xy".
+
+        Returns
+        -------
+        np.ndarray
+            A 2D or 3D array of shape (n_channels, n_dimensions) containing the locations of the channels.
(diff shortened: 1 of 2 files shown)


alejoe91 (approved):

src/spikeinterface/core/baserecording.py
```diff
@@ -746,6 +746,30 @@ def _select_segments(self, segment_indices):
...
+    ) -> np.ndarray:
+        """
+        Get the physical locations of specified channels.
+
+        Parameters
+        ----------
+        channel_ids : array-like, optional
+            The IDs of the channels for which to retrieve locations. If None, retrieves locations
+            for all available channels. Default is None.
+        axes : str, optional
```
zm711, line 762:
```suggestion
        axes : "xy" | "yz" | "xz" | "xyz", default: "xy"
```

I know we need to rehash the `None`/ `optional` argument with the fuller team, but I think it was not being debated to do this part this way

src/spikeinterface/core/baserecording.py
```diff
@@ -746,6 +746,30 @@ def _select_segments(self, segment_indices):
 
         return SelectSegmentRecording(self, segment_indices=segment_indices)
 
+    def get_channel_locations(
+        self,
+        channel_ids: list | np.ndarray | tuple | None = None,
+        axes: "xy" | "yz" | "xz" = "xy",
```
zm711, line 752:
```suggestion
        axes: "xy" | "yz" | "xz" | "xyz" = "xy",
```
based on the possible return we are missing the 3d option. Or the return needs to be fixed!

src/spikeinterface/core/baserecording.py
```diff
@@ -746,6 +746,30 @@ def _select_segments(self, segment_indices):
...
+    ) -> np.ndarray:
+        """
+        Get the physical locations of specified channels.
```
zm711, line 755: I think this is likely improvement enough for merging, but since contact vs channel is still a point of confusion for many users we may want to workshop better ways to talk about channel_locations for public functions.

zm711 (commented):
Small tweaks :)

zm711 (approved):
Like I said my other point was more for us to think about long-term. This PR is a great doc improvement.

alejoe91 (approved):

PR #3405, opened 2024-09-13
Optimizations for template_similarity (numba and dependencies)

yger:
- Add the possibility to compute the template similarities (rather slow for the moment) using a numba kernel. 
- Remove the dependencies to sklearn in template_similarity without numba, since we do not plan anymore to add metric
- Also adds prange in some other numba functions, with the keyword parallel=True (otherwise, prange is not activated, as far as I understood).

### DIFF
--- a/src/spikeinterface/postprocessing/template_similarity.py
+++ b/src/spikeinterface/postprocessing/template_similarity.py
@@ -263,6 +263,7 @@ if HAVE_NUMBA:
 
                     norm_i = 0
                     norm_j = 0
+                    distances[count, i, j] = 0
 
                     for k in range(len(src)):
                         if metric == 0:
--- a/src/spikeinterface/postprocessing/template_similarity.py
+++ b/src/spikeinterface/postprocessing/template_similarity.py
@@ -155,9 +155,6 @@ compute_template_similarity = ComputeTemplateSimilarity.function_factory()
 
 
 if HAVE_NUMBA:
-    
-    from numba import prange
-
 
     @numba.jit(nopython=True, parallel=True, fastmath=True, cache=True, nogil=True)
     def _compute_similarity_matrix(
@@ -183,7 +180,7 @@ if HAVE_NUMBA:
         for count, shift in enumerate(shift_loop):
             src_sliced_templates = templates_array[:, num_shifts : num_samples - num_shifts]
             tgt_sliced_templates = other_templates_array[:, num_shifts + shift : num_samples - num_shifts + shift]
-            for i in prange(num_templates):
+            for i in numba.prange(num_templates):
                 src_template = src_sliced_templates[i]
                 overlapping_templates = np.flatnonzero(np.sum(mask[i], 1))
                 tgt_templates = tgt_sliced_templates[overlapping_templates]
--- a/src/spikeinterface/postprocessing/template_similarity.py
+++ b/src/spikeinterface/postprocessing/template_similarity.py
@@ -7,6 +7,13 @@ from spikeinterface.core.sortinganalyzer import register_result_extension, Analy
 from ..core.template_tools import get_dense_templates_array
 from ..core.sparsity import ChannelSparsity
 
+try:
+    import numba
+
+    HAVE_NUMBA = True
+except ImportError:
+    HAVE_NUMBA = False
+
 
 class ComputeTemplateSimilarity(AnalyzerExtension):
     """Compute similarity between templates with several methods.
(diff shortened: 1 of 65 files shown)


zm711:
One thing to keep in mind is that the process of creating the threads is pretty expensive, so depending on the number of computations it can actually be beneficial to do the calculation serially rather than do the parallel operation. Not that we shouldn't do this, but we should just thing about common use case vs worst case scenario.

yger:
Yes, but here on my machine the speed up is rather drastic, but I can do more benchmarks if needed. In its current form, the template_similarity is too slow (if max_lag_ms is not 0 this is really bad) for large number of templates. So one must optimize it. I've discussed with @samuelgarcia  and made two proposals. One in PR #3174 involving PoolExecutor and this one, relying on numba. This is simpler I guess, but this is worth letting @samuelgarcia comment :-)

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -147,10 +154,144 @@ def _get_data(self):
 compute_template_similarity = ComputeTemplateSimilarity.function_factory()
 
 
+if HAVE_NUMBA:
```
h-mayorquin, line 157: When we were making the import faster I remember that we discussed whether this compiles at init time or not. Does anyone have a better memory than me @zm711 @JoeZiminski ?

zm711: I think the compiling at init is mainly due to type hints. I was actually listening to someone talk about numba the other day and they recommend not to do the type hints for numba and let it infer them. So I think we should be fine. We should see the slow down in the import action right? or we could use tuna (I forget the name @JoeZiminski )

h-mayorquin: Ah yes, it was because Aurelien specified the actual types on the numba decorator. Now I remember.

zm711: Exactly.

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -147,10 +154,144 @@ def _get_data(self):
...
+            shift_loop = range(-num_shifts, num_shifts + 1)
+
+        for count, shift in enumerate(shift_loop):
+            src_sliced_templates = templates_array[:, num_shifts : num_samples - num_shifts]
+            tgt_sliced_templates = other_templates_array[:, num_shifts + shift : num_samples - num_shifts + shift]
+            for i in numba.prange(num_templates):
+                src_template = src_sliced_templates[i]
+                overlapping_templates = np.flatnonzero(np.sum(mask[i], 1))
+                tgt_templates = tgt_sliced_templates[overlapping_templates]
+                for gcount, j in enumerate(overlapping_templates):
```
samuelgarcia, line 185: not sure if enumerate is good for numba or not.

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -147,10 +154,144 @@ def _get_data(self):
...
+                        # no need exhaustive looping when same template
+                        continue
+                    src = src_template[:, mask[i, j]].flatten()
+                    tgt = (tgt_templates[gcount][:, mask[i, j]]).flatten()
+
+                    norm_i = 0
+                    norm_j = 0
+
+                    for k in range(len(src)):
+                        if method == "l1":
```
samuelgarcia, line 198: Maybe we should use constant with number to avoid testing against a str object in any inner loops.
 Something in C style like
```python
L1 = 0
L2 = 1
COSINE = 2

method_int = {"l1": L1, "l2": L2, "cosine": COSINE}[method]

for ... in ... :
    for ... in ... :
       for ... in ... :
            if method_int == L1:




```

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -147,10 +154,144 @@ def _get_data(self):
...
+                overlapping_templates = np.flatnonzero(np.sum(mask[i], 1))
+                tgt_templates = tgt_sliced_templates[overlapping_templates]
+                for gcount, j in enumerate(overlapping_templates):
+
+                    # symmetric values are handled later
+                    if same_array and j < i:
+                        # no need exhaustive looping when same template
+                        continue
+                    src = src_template[:, mask[i, j]].flatten()
+                    tgt = (tgt_templates[gcount][:, mask[i, j]]).flatten()
```
samuelgarcia, line 192: Cool but how does numba deal with method ? Is it fast ?

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -147,10 +154,144 @@ def _get_data(self):
...
+                            distances[count, i, j] += (src[k] - tgt[k]) ** 2
+                        elif method == "cosine":
+                            distances[count, i, j] += src[k] * tgt[k]
+                            norm_i += src[k] ** 2
+                            norm_j += tgt[k] ** 2
+
+                    if method == "l1":
+                        distances[count, i, j] /= norm_i + norm_j
+                    elif method == "l2":
+                        norm_i = np.sqrt(norm_i)
```
samuelgarcia, line 214: Do we need np.sqrt or sqrt in numba ?

src/spikeinterface/postprocessing/template_similarity.py
zm711: https://numba.pydata.org/numba-doc/0.17.0/reference/pysupported.html

it is supported. Not sure if it is a good idea.

src/spikeinterface/postprocessing/template_similarity.py
zm711: https://numba.pydata.org/numba-doc/dev/reference/numpysupported.html

it is supported, for "c" order only. Which I think we always follow? Again not sure on speed.

yger: This is supported

src/spikeinterface/postprocessing/template_similarity.py
yger: No speed difference

src/spikeinterface/postprocessing/template_similarity.py
yger: I've made something like that, but no speed difference

src/spikeinterface/postprocessing/template_similarity.py
yger: I've remove the enumerate, but speed difference is very small

yger:
For 500 templates with 384 channels (Neuropixels like) we have on a laptop (for 0.1 ms of max_lag_ms) 

* default implementation: 5.1s
* numba implementation: compilation time 10.6s, run time 1.1s
* poolexecutor: 9s (but speed up on a desktop with more cores, i guess we see mostly overhead here)

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -147,10 +154,144 @@ def _get_data(self):
...
+        if same_array:
+            # optimisation when array are the same because of symetry in shift
+            shift_loop = range(-num_shifts, 1)
+        else:
+            shift_loop = range(-num_shifts, num_shifts + 1)
+
+        for count, shift in enumerate(shift_loop):
+            src_sliced_templates = templates_array[:, num_shifts : num_samples - num_shifts]
+            tgt_sliced_templates = other_templates_array[:, num_shifts + shift : num_samples - num_shifts + shift]
+            for i in numba.prange(num_templates):
```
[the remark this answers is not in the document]
yger: I've tried to invert the loops, and it does not make any difference

alejoe91:
@yger could you add a test that checks that numba and numpy implementations give the same results?

yger:
Done, tests are passing

samuelgarcia (approved):

samuelgarcia:
Merci Pierre. This is really cool.

PR #3407, opened 2024-09-13
Check run info completed only if it exists (back-compatibility)

alejoe91:
@samuelgarcia


zm711 (commented):
Just to confirm since I didn't fully follow this story, even if the run_info.json doesn't exist it won't error. It will just load as a None and that's why checking for None works?

alejoe91:
> Just to confirm since I didn't fully follow this story, even if the run_info.json doesn't exist it won't error. It will just load as a None and that's why checking for None works?

Exactly :)

zm711 (approved):
This works for me then :)

PR #3412, opened 2024-09-15
Refactor pandas save load and convert dtypes

alejoe91:
We found out that zarr consolidation doens't seem to play well with our way of saving/loading dataframes to zarr, using xarray.

`xarray` was only used to save/load pandas dataframes to zarr, and this PR modifies that by saving each column and the index directly. This is similar to how xarray saves to zarr, so it should be backcompatible (testing now).

To make sure we don't run in such problems in the future, I added a roundtrip test in the common extension tests that asserts that data reloaded is the same as the original ones.

As sugegsted by @h-mayorquin here #3365, the generated and reloaded dataframes are also converted to numpy dtypes with the `convert_dtypes` function. We just have to make sure to call a `Series.to_numpy` to cast pandas dtypes to numpy ones.

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -591,7 +591,7 @@ class SortingAnalyzer:
                 warnings.warn(
                     "The zarr store was not properly consolidated prior to v0.101.1. "
                     "This may lead to unexpected behavior in loading extensions. "
-                    "Please consider re-saving the SortingAnalyzer object."
+                    "Please consider re-generating the SortingAnalyzer object."
                 )
 
         # load internal sorting in memory
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -1970,12 +1970,14 @@ class AnalyzerExtension:
                 if "dict" in ext_data_.attrs:
                     ext_data = ext_data_[0]
                 elif "dataframe" in ext_data_.attrs:
-                    import xarray
+                    import pandas as pd
 
-                    ext_data = xarray.open_zarr(
-                        ext_data_.store, group=f"{extension_group.name}/{ext_data_name}"
-                    ).to_pandas()
-                    ext_data.index.rename("", inplace=True)
+                    index = ext_data_["index"]
+                    ext_data = pd.DataFrame(index=index)
+                    for col in ext_data_.keys():
+                        if col != "index":
+                            ext_data.loc[:, col] = ext_data_[col][:]
+                    ext_data = ext_data.convert_dtypes()
                 elif "object" in ext_data_.attrs:
                     ext_data = ext_data_[0]
                 else:
@@ -2031,12 +2033,21 @@ class AnalyzerExtension:
         if save and not self.sorting_analyzer.is_read_only():
             self._save_run_info()
             self._save_data(**kwargs)
+            if self.format == "zarr":
+                import zarr
+
+                zarr.consolidate_metadata(self.sorting_analyzer._get_zarr_root().store)
 
     def save(self, **kwargs):
         self._save_params()
         self._save_importing_provenance()
-        self._save_data(**kwargs)
         self._save_run_info()
+        self._save_data(**kwargs)
+
+        if self.format == "zarr":
+            import zarr
+
+            zarr.consolidate_metadata(self.sorting_analyzer._get_zarr_root().store)
 
     def _save_data(self, **kwargs):
         if self.format == "memory":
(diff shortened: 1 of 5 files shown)


src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -287,7 +287,7 @@ def _compute_metrics(self, sorting_analyzer, unit_ids=None, verbose=False, **job
                     warnings.warn(f"Error computing metric {metric_name} for unit {unit_id}: {e}")
                     value = np.nan
                 template_metrics.at[index, metric_name] = value
-        return template_metrics
+        return template_metrics.convert_dtypes()
```
zm711, line 290: This is a weak recommendation ,but maybe we put this on its own line with a comment. Just from reading this I have no clue why we need to do this and doing this in the return line is even more confusing. So something like
```python
# see xx
template_metrics.convert_dtypes()
return template_metrics
```

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -185,7 +185,7 @@ def _compute_metrics(self, sorting_analyzer, unit_ids=None, verbose=False, **job
         if len(empty_unit_ids) > 0:
             metrics.loc[empty_unit_ids] = np.nan
 
-        return metrics
+        return metrics.convert_dtypes()
```
zm711, line 188: same here. From the code it is not clear why we need to convert dtypes so I would refer to divide this into a convert step and then only return the converted. That way we can have a comment explaining why we need to convert.

zm711 (commented):
Since I haven't fully tested zarr yet I want to make sure. We have an appropriate pandas warning somewhere for users so they know they need pandas for these features. I know we have a warning for qualitymetrics do we have one for templatemetrics?

alejoe91:
> Since I haven't fully tested zarr yet I want to make sure. We have an appropriate pandas warning somewhere for users so they know they need pandas for these features. I know we have a warning for qualitymetrics do we have one for templatemetrics?

We don't have warnings anywhere. If a user tries to compute template or quality metrics without pandas, it will throw an interpetable `ModuleNotFoundError` :)

src/spikeinterface/postprocessing/template_metrics.py
alejoe91: better?

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
alejoe91: added comment and convert step

alejoe91:
One last comment: for analyzers saved to `zarr` in version 0.101.0, the consolidation step was missing after the computation of each extension. I added a check and a consolitation step, that raises a warning if it fails

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -139,7 +139,9 @@ def _compute_metrics(self, sorting_analyzer, unit_ids=None, verbose=False, metri
         """
         Compute quality metrics.
         """
+        import pandas as pd
 
+        metric_names = self.params["metric_names"]
```
alejoe91, line 144:
```suggestion
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -579,6 +580,20 @@ def load_from_zarr(cls, folder, recording=None, storage_options=None):
...
+            # v0.101.0 did not have a consolidate metadata step after computing extensions.
+            # Here we try to consolidate the metadata and throw a warning if it fails.
+            try:
+                zarr_root_a = zarr.open(str(folder), mode="a", storage_options=storage_options)
+                zarr.consolidate_metadata(zarr_root_a.store)
+            except Exception as e:
+                warnings.warn(
+                    "The zarr store was not properly consolidated prior to v0.101.1. "
+                    "This may lead to unexpected behavior in loading extensions. "
+                    "Please consider re-saving the SortingAnalyzer object."
```
zm711, line 594: with a `save_as(format='zarr')`?  I just want to make sure the error is as clear as possible.

src/spikeinterface/postprocessing/template_metrics.py
```diff
@@ -318,6 +318,10 @@ def _compute_metrics(self, sorting_analyzer, unit_ids=None, verbose=False, metri
                     warnings.warn(f"Error computing metric {metric_name} for unit {unit_id}: {e}")
                     value = np.nan
                 template_metrics.at[index, metric_name] = value
+
+        # we use the convert_dtypes to convert the columns to the most appropriate dtype and avoid object columns
+        # (in case of NaN values)
+        template_metrics = template_metrics.convert_dtypes()
```
zm711, line 324: Thanks I think that's great!

zm711 (commented):
This looks pretty good to me!

src/spikeinterface/core/sortinganalyzer.py
alejoe91: Not really..since the problem is consolidation, the save as may fail to discover all the pieces of the folder. I changed it to `re-generating`.

Honestly, I don't think this will be an issue since it will only happen if:
- you saved to zarr between 0.101.0 and 0.101.1
- you don't have write access to the data

samuelgarcia (approved):

zm711:
Thanks Alessio!

PR #3415, opened 2024-09-16
Auto-cast recording to float prior to interpolation

### LINKED ISSUE #3404: Can't interpolate traces with dtype('int16')
Hello!
I've been having the same error as was seen in [https://github.com/SpikeInterface/spikeinterface/issues/3146#issue-2391280072](this thread). I was happy to switch to float32, as they did, but it seems like this is causing problems further down in our processing chain with Phy. We are running Kilosort4 from SI after performing motion correction—what's the best course of action for us? Should we somehow convert our data back to int16 prior to running kilosort?

  File "----------", line 122, in <module>
    rec_corrected, motion_info = si.correct_motion(rec, preset='nonrigid_accurate', interpolate_motion_kwargs={'border_mode':'force_extrapolate'},folder=result_folder, output_motion_info=True)
  File "---------/miniconda3/envs/kilosort/lib/python3.9/site-packages/spikeinterface/preprocessing/motion.py", line 433, in correct_motion
    recording_corrected = InterpolateMotionRecording(recording, motion, **interpolate_motion_kwargs)
  File "-----------/miniconda3/envs/kilosort/lib/python3.9/site-packages/spikeinterface/sortingcomponents/motion/motion_interpolation.py", line 346, in __init__
    raise ValueError(f"Can't interpolate traces of recording with non-floating dtype={recording.dtype=}.")
ValueError: Can't interpolate traces of recording with non-floating dtype=recording.dtype=dtype('int16').
srun: error: gpu-n40: task 0: Exited with exit code 1

alejoe91:
Fixes #3404

Thanks @jiumao2 for the suggestion

### DIFF
--- a/src/spikeinterface/preprocessing/motion.py
+++ b/src/spikeinterface/preprocessing/motion.py
@@ -13,6 +13,7 @@ from spikeinterface.core.job_tools import _shared_job_kwargs_doc
 from spikeinterface.core.core_tools import SIJsonEncoder
 from spikeinterface.core.job_tools import _shared_job_kwargs_doc
 
+
 motion_options_preset = {
     # dredge
     "dredge": {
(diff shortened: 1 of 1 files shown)


samuelgarcia (approved):

JoeZiminski:
Is it worth printing / warning a statement when the conversion is done? Just to indicate the conversion is done and file will be larger if saved now.

alejoe91:
Actually I'll just go ahead and merge..I think that an extra warnign will be confusing for most usere :)

jiumao2:
> Since the `correct_motion` is intended to be as high-level as possible, we thought we would just keep it silent and do whatever's needed to avoid failures. @samuelgarcia @jiumao2 what do you guys think?

That's great! I’ve actually already added that code to my spikeinterface!

PR #3370, opened 2024-09-05
Prepare release 0.101.1

alejoe91:
- [x] Update pyproject toml (remove git installs)
- [x] Update `DEV_MODE` in init
- [x] Add release notes
- [x] Update `whatisnew`
- [x] Tag KS4 4.0.16 Docker image as `latest`

PR #3417, opened 2024-09-16
Fix metrics widgets for convert_dtypes

alejoe91:
Moved some bug fixes out of #3370 for better logging :)


alejoe91:
@zm711 if you could approve this I'll go ahead wiht the review!

zm711 (commented):
One tiny question. Should we unify that someway? Why return None sometimes and np.nan other times. It might work better long-term if we can have consistent behavior?

zm711 (approved):
Works for me. Thanks.

PR #3422, opened 2024-09-18
Enable testing arm64 Mac architecture in the CI.

h-mayorquin:
Moving forward with #3100.

### DIFF
--- a/src/spikeinterface/extractors/tests/test_neoextractors.py
+++ b/src/spikeinterface/extractors/tests/test_neoextractors.py
@@ -234,7 +234,7 @@ class BlackrockSortingTest(SortingCommonTestSuite, unittest.TestCase):
     ExtractorClass = BlackrockSortingExtractor
     downloads = ["blackrock"]
     entities = [
-        "blackrock/FileSpec2.3001.nev",
+        dict(file_path=local_folder / "blackrock/FileSpec2.3001.nev", sampling_frequency=30_000.0),
         dict(file_path=local_folder / "blackrock/blackrock_2_1/l101210-001.nev", sampling_frequency=30_000.0),
     ]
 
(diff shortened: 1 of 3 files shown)


h-mayorquin:
Uh, that's great, it just... worked.

h-mayorquin:
Ah, of course it was so easy, we did not have logic for running everything if the workflow files from github actions are changed. I added that.

h-mayorquin:
Ok, this is ready. I also eliminated a warning from the CI to keep things tidy  🧹 🧹 🧹 (   x D )

zm711:
I think the only potentially slow step would be if we do a bunch of torch things, since pytorch defaults to the Mac CPU instead of GPU, but I don't even know what the runner GPU story is.

src/spikeinterface/extractors/tests/test_neoextractors.py
```diff
@@ -278,8 +278,8 @@ class Spike2RecordingTest(RecordingCommonTestSuite, unittest.TestCase):
 
 
 @pytest.mark.skipif(
-    version.parse(platform.python_version()) >= version.parse("3.10"),
-    reason="Sonpy only testing with Python < 3.10!",
+    version.parse(platform.python_version()) >= version.parse("3.10") or platform.system() == "Darwin",
+    reason="Sonpy only testing with Python < 3.10 and not supported on macOS!",
```
zm711, line 282: Is this an Arm issue?

h-mayorquin: Yes, want to give it a shot to have something that excludes intel cpus from mac? I think they stopped selling two years ago only so many people might still have them.

That said, I did not want to allocate extra effort to something that won't work in a year anyway (see the python restriction)

zm711 (approved):
For me this is good to go. Based on a google intel Macs officially went end of life in JUNE2024. So only testing on officially supported hardware seems fair to me.  And it looks like it successfully tested everything for complete tests so I think this should be good to go.

alejoe91 (approved):

PR #3426, opened 2024-09-18
Fix integer overflow in parallel computing

jiumao2:
`frame_starts = np.arange(n) * chunk_size` will overflow when the recording is ~24 hours long and `chunk_size = "1s"`

Another integer overflow issue on long recordings following #3388

### DIFF
--- a/src/spikeinterface/core/job_tools.py
+++ b/src/spikeinterface/core/job_tools.py
@@ -136,11 +136,8 @@ def divide_segment_into_chunks(num_frames, chunk_size):
     else:
         n = num_frames // chunk_size
 
-        frame_starts = np.arange(n) * chunk_size
-        frame_stops = frame_starts + chunk_size
-
-        frame_starts = frame_starts.tolist()
-        frame_stops = frame_stops.tolist()
+        frame_starts = [i * chunk_size for i in range(n)]
+        frame_stops = [(i+1) * chunk_size for i in range(n)]
 
         if (num_frames % chunk_size) > 0:
             frame_starts.append(n * chunk_size)
--- a/src/spikeinterface/core/job_tools.py
+++ b/src/spikeinterface/core/job_tools.py
@@ -137,7 +137,7 @@ def divide_segment_into_chunks(num_frames, chunk_size):
         n = num_frames // chunk_size
 
         frame_starts = [i * chunk_size for i in range(n)]
-        frame_stops = [(i + 1) * chunk_size for i in range(n)]
+        frame_stops = [frame_start + chunk_size for frame_start in frame_starts]
 
         if (num_frames % chunk_size) > 0:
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/job_tools.py
```diff
@@ -136,11 +136,8 @@ def divide_segment_into_chunks(num_frames, chunk_size):
     else:
         n = num_frames // chunk_size
 
-        frame_starts = np.arange(n) * chunk_size
-        frame_stops = frame_starts + chunk_size
-
-        frame_starts = frame_starts.tolist()
-        frame_stops = frame_stops.tolist()
+        frame_starts = [i * chunk_size for i in range(n)]
+        frame_stops = [(i + 1) * chunk_size for i in range(n)]
```
zm711, line 140: This works. I tend to lean toward explicit so something like

```

frame_stops = [frame_start + chunk_size for frame_start in frame_starts]
```

Just because it is clearer rather than have to multiple things out. But I think this is a good protection against overflow. Your long recordings are really finding all the overflows :)

zm711:
@h-mayorquin

h-mayorquin (approved):
Yeah, this is a good idea.

I also prefer @zm711 writing but this is your PR and that would firmly in style territory : P

src/spikeinterface/core/job_tools.py
jiumao2: Yeah... I am surprised no one has tried to deal with long recordings before. Are there any other long-recording users?

zm711:
>  firmly in style territory : P

pero somos elegantes, carnal.

zm711 (approved):

src/spikeinterface/core/job_tools.py
zm711: The most I've heard of recently is closer to the 10-15 hour mark. But haven't really followed up on those. These things work until they don't. So we are trying to track them down, but they are easier to find when they break.

jiumao2:
We all love @zm711 writing!

src/spikeinterface/core/job_tools.py
jiumao2: Hope that long recording is not a bad idea :)

alejoe91: @jiumao2 we are planning to run week long recordings. You are in good company :)

jiumao2: Good to hear that! Not alone anymore :)

jiumao2: Hi @alejoe91 ,

Are you beginning to work with the week-long recordings? I'm not satisfied with the results from Kilosort and have run into a number of issues. Could you share any strategies or tools you're using, such as different sorters, that might be helpful?

Thank you!

Issue #3367, opened 2024-09-05
cannot detect the installed mountainsort4


betty-hbr:
Hi team,

I'm trying to use mountainsort4 in Spikeinterface 0.101.0. Although I have correctly installed the mountainsort4 to the same environment as Spikeinterface, it cannot detect or to run it. 

The output of ss.installed_sorters() are ['mountainsort5', 'simple', 'spykingcircus2', 'tridesclous2'].

I also tried with Spikeinterface 0.97.0 and 0.98.0. Same issues for them.

Can you please advice on this?

Many thanks

alejoe91:
Can you try to import it?
```
import mountainsort4
```

The `installed_sorters` will list sorters that are importable

betty-hbr:
Thank you so much for the prompt response! I tried with your suggestion. It wasn't importable, because I didn't have `requests` installed. The problem was fixed after  `pip install requests`

PR #3431, opened 2024-09-21
Add pip install requests to mountainsort4 install instructions

rly:
I just ran into the same issue as in #3367 and was quite puzzled. I think `pip install requests` should be included in the mountainsort4 installation instructions here in the spikeinterface docs.

### DIFF
--- a/doc/get_started/install_sorters.rst
+++ b/doc/get_started/install_sorters.rst
@@ -189,7 +189,7 @@ Mountainsort4
 * Authors: 	Jeremy Magland, Alex Barnett, Jason Chung, Loren Frank, Leslie Greengard
 * Installation::
 
-      pip install mountainsort4
+      pip install mountainsort4 requests
 
 Mountainsort5
 ^^^^^^^^^^^^^


rly:
I also found that I needed to pip install pybind11 before mountainsort4 (which lists pybind11 as a requirement) on my Mac M1. I don't know how universal that is though.

zm711:
If I have time later today I can try on a Windows computer too and if it is necessary on Mac and Windows then I would say we should ping a comment. Otherwise this looks good to me!

alejoe91:
> I also found that I needed to pip install pybind11 before mountainsort4 (which lists pybind11 as a requirement) on my Mac M1. I don't know how universal that is though.

Yeah that is needed also on Linux. @rly could you add that too? Or @magland could add both to `mountainsort4` :)

magland:
@alejoe91 

I have updated mountainsort4 to no longer depend on requests (and released to pypi)
https://github.com/magland/mountainsort4/pull/9

As stated, it already has a requirement of pybind11... but I think that sometimes needs to be installed prior to ms4 installation.

rly:
@magland released mountainsort4 1.0.6 which should resolve the need to install pybind11 as well. At least it does on my mac. Thanks!

PR #3432, opened 2024-09-23
Fix bug in saving zarr recordings

alejoe91:
The times array loaded from zarr are converted to numpy arrays only when `get_times` is called. This resulted in an error when saving a zarr-based recording, since we were accessing times vector directly.


zm711 (approved):
This makes sense to me.

PR #3433, opened 2024-09-23
Fix compute analyzer pipeline with tmp recording

alejoe91:
Fixes a bug that prevented runnning pipeline-based extensions with a temprary recording alone.

Also some other small fixes (e.g., removing some `we` variable names in pca postprocessing)

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -230,8 +230,10 @@ class SortingAnalyzer:
         txt = f"{clsname}: {nchan} channels - {nunits} units - {nseg} segments - {self.format}"
         if self.is_sparse():
             txt += " - sparse"
-        if self.has_recording() or self.has_temporary_recording():
+        if self.has_recording():
             txt += " - has recording"
+        if self.has_temporary_recording():
+            txt += " - has temporary recording"
         ext_txt = f"Loaded {len(self.extensions)} extensions: " + ", ".join(self.extensions.keys())
         txt += "\n" + ext_txt
         return txt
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1355,7 +1355,9 @@ def compute_several_extensions(self, extensions, save=True, verbose=False, **job
 
             for extension_name, extension_params in extensions_with_pipeline.items():
                 extension_class = get_extension_class(extension_name)
-                assert self.has_recording(), f"Extension {extension_name} need the recording"
+                assert (
+                    self.has_recording() or self.has_temporary_recording()
```
alejoe91, line 1359: this is the main fix!

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -230,7 +230,7 @@ def __repr__(self) -> str:
         txt = f"{clsname}: {nchan} channels - {nunits} units - {nseg} segments - {self.format}"
         if self.is_sparse():
             txt += " - sparse"
-        if self.has_recording():
+        if self.has_recording() or self.has_temporary_recording():
```
zm711, line 233: One tiny idea (but more for discussion not a hard suggestion). Would it be beneficial to update the repr such that it says whether it is `has recording` or `has temp recording` so the user has a quick display to remind them if this was the original that made the object or a temp they just attached? Might be overkill, but I could imagine wanting to double check if I were trying to reanalyze a bunch of datasets some with original and some where I attached a temp.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1355,7 +1355,9 @@ def compute_several_extensions(self, extensions, save=True, verbose=False, **job
 
             for extension_name, extension_params in extensions_with_pipeline.items():
                 extension_class = get_extension_class(extension_name)
-                assert self.has_recording(), f"Extension {extension_name} need the recording"
+                assert (
+                    self.has_recording() or self.has_temporary_recording()
+                ), f"Extension {extension_name} need the recording"
```
zm711, line 1360:
```suggestion
                ), f"Extension {extension_name} requires the recording"
```
while we are fixing this. at minimum it needs an "s" added on. I prefer the word requires, but you could also change to needs :)

alejoe91: done, plus some other docs fixes

zm711 (approved):
Looks good.

PR #3435, opened 2024-09-24
Refactoring auto_merge

yger:
This PR is made because of @samuelgarcia request. It will ease the design of recursive auto_merging functions, requiered by SC2 and better methods of automated merging. The idea is that one can provide preset (as before), and params are exposed as dictionary in auto_merges. The old behavior get_potential_auto_merges() is kept, but under the hood, we are now calling the new fonction. The fonction is based as much as possible on extensions.


src/spikeinterface/curation/auto_merge.py
```diff
@@ -25,35 +23,43 @@
...
+    template_similarity_kwargs={"template_diff_thresh": 0.25},
+    presence_distance_kwargs={"presence_distance_thresh": 100},
+    knn_kwargs={"k_nn": 10},
+    cross_contamination_kwargs={
+        "cc_thresh": 0.1,
+        "p_value": 0.2,
+        "refractory_period_ms": 1.0,
+        "censored_period_ms": 0.3,
+    },
+    quality_score_kwargs={"firing_contamination_balance": 2.5, "refractory_period_ms": 1.0, "censored_period_ms": 0.3},
```
samuelgarcia, line 57: I think we should have a single dict with key for the step and value is a sub dict for step params.
This will avoid the horrible eval() in the middle and will be more easy to handle externally.

src/spikeinterface/curation/auto_merge.py
```diff
@@ -230,12 +201,24 @@ def get_potential_auto_merge(
...
             for ext in _required_extensions[step]:
-                if not sorting_analyzer.has_extension(ext):
-                    raise ValueError(f"{step} requires {ext} extension")
+                if compute_needed_extensions:
+                    if step in _templates_needed:
+                        template_ext = sorting_analyzer.get_extension("templates")
+                        if template_ext is None:
+                            sorting_analyzer.compute(["random_spikes", "templates"], **job_kwargs)
+                    params = eval(f"{step}_kwargs")
+                    params = params.get(ext, dict())
```
samuelgarcia, line 217: What is the expented signature to have a nested dict for the extension params ?
This is a bit strange.
For me we should do it externally and when this is done automatically then we use the default

src/spikeinterface/curation/auto_merge.py
```diff
@@ -248,55 +231,62 @@ def get_potential_auto_merge(
...
             pair_mask[:, to_remove] = False
+            outs["num_spikes"] = to_remove
 
         # STEP : remove units with too small SNR
         elif step == "snr":
             qm_ext = sorting_analyzer.get_extension("quality_metrics")
             if qm_ext is None:
-                sorting_analyzer.compute("noise_levels")
-                sorting_analyzer.compute("quality_metrics", metric_names=["snr"])
+                sorting_analyzer.compute(["noise_levels"], **job_kwargs)
```
samuelgarcia, line 243: we should check that this exists

src/spikeinterface/curation/auto_merge.py
yger: Done

src/spikeinterface/curation/auto_merge.py
yger: Done

src/spikeinterface/curation/auto_merge.py
yger: Ok done, now if not present default is used (and a message is displayed to warn the user)

src/spikeinterface/curation/auto_merge.py
```diff
@@ -354,19 +352,16 @@ def get_potential_auto_merge(
...
+                params["censored_period_ms"],
             )
             outs["pairs_decreased_score"] = pairs_decreased_score
 
     # FINAL STEP : create the final list from pair_mask boolean matrix
     ind1, ind2 = np.nonzero(pair_mask)
     potential_merges = list(zip(unit_ids[ind1], unit_ids[ind2]))
 
-    # some methods return identities ie (1,1) which we can cleanup first.
-    potential_merges = [(ids[0], ids[1]) for ids in potential_merges if ids[0] != ids[1]]
```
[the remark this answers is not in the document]
zm711: Maybe we add a test then instead. I'm fine to have this removed if the identity merges are fixed in the refactor.

src/spikeinterface/curation/auto_merge.py
```diff
@@ -25,35 +23,48 @@
...
+        "p_value": 0.2,
+        "refractory_period_ms": 1.0,
+        "censored_period_ms": 0.3,
+    },
+    "quality_score": {"firing_contamination_balance": 2.5, "refractory_period_ms": 1.0, "censored_period_ms": 0.3},
+}
 
-def get_potential_auto_merge(
+
+def auto_merges(
```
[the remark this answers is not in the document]
yger: done

src/spikeinterface/curation/auto_merge.py
```diff
@@ -230,12 +202,9 @@ def get_potential_auto_merge(
...
             ]
-
-    for step in steps:
-        if step in _required_extensions:
-            for ext in _required_extensions[step]:
-                if not sorting_analyzer.has_extension(ext):
-                    raise ValueError(f"{step} requires {ext} extension")
+    if force_copy and compute_needed_extensions:
+        # To avoid erasing the extensions of the user
+        sorting_analyzer = sorting_analyzer.copy()
```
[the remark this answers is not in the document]
yger: Now this might not be needed anymore. This was because at the beginning of this function, one could use a sorting analyzer with already precomputed extension and recompute some with new params for the merges. But this is not possible anymore. Either extensions have been already computed with custom params, either we use default. So this copy is not needed anymore I guess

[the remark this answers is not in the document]
yger: We can leave that here in case user would not want to save some extra extensions that might have been needed during the merges

src/spikeinterface/curation/auto_merge.py
```diff
@@ -376,6 +371,182 @@ def get_potential_auto_merge(
...
+        Returned only when extra_outputs=True
+        A dictionary that contains data for debugging and plotting.
+
+    References
+    ----------
+    This function is inspired and built upon similar functions from Lussac [Llobet]_,
+    done by Aurelien Wyngaard and Victor Llobet.
+    https://github.com/BarbourLab/lussac/blob/v1.0.0/postprocessing/merge_units.py
+    """
+    presence_distance_kwargs = presence_distance_kwargs or dict()
```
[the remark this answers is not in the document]
yger: I've added a sentence to say it has been inspired, but we should keep the reference I think

src/spikeinterface/curation/auto_merge.py
```diff
@@ -98,205 +141,145 @@ def get_potential_auto_merge(
...
     unit_ids = sorting.unit_ids
 
-    # to get fast computation we will not analyse pairs when:
-    #    * not enough spikes for one of theses
-    #    * auto correlogram is contaminated
-    #    * to far away one from each other
+    if preset is None and steps is None:
+        raise ValueError("You need to specify a preset or steps for the auto-merge function")
+    elif steps is not None:
+        # steps has presendance on presets
```
alejoe91, line 188:
```suggestion
        # steps has precedence on presets
```

yger:
Forget what I said @samuelgarcia , I found the bug in the other PR, this one is working fine and good to merge on my side

samuelgarcia (approved):

samuelgarcia:
@alejoe91 this is OK for me.

alejoe91 (approved):

PR #3438, opened 2024-09-25
Add PreprocessingPipeline

chrishalcrow:
Add a `PreprocessingPipeline` class, which contains ordered preprocessing steps and their kwargs in a dictionary. 

Docs: 
https://spikeinterface--3438.org.readthedocs.build/en/3438/modules/preprocessing.html#the-preprocessing-pipeline
https://spikeinterface--3438.org.readthedocs.build/en/3438/how_to/build_pipeline_with_dicts.html

You can `apply_preprocessing_pipeline` to a recording to make a preprocessed recording:

``` python
preprocessor_dict = {'bandpass_filter': {'freq_max': 3000}, 'common_reference': {}}

from spikeinterface.preprocessing import apply_preprocessing_pipeline
preprocessed_recording = apply_preprocessing_pipeline(recording, preprocessor_dict)
```

Under the hood, this uses the `_apply` method of the new `PreprocessingPipeline`. 

Also adds a function which takes in a `provenance.json`, `provenance.pkl`, 'recording.json` or `recording.pkl` provenance file and makes a `preprocessor_dict`. So it's easy to extract preprocessing steps from a saved recording.

``` python
from spikeinterface.preprocessing import get_preprocessing_dict_from_json
my_dict = get_preprocessing_dict_from_provenance('/path/to/provenance.json')
```

Can also get it from an analyzer, using `get_preprocessing_dict_from_analyzer` function.

After you load this, you can either apply the `precomputable_kwargs` or ignore them and compute on application:

``` python
# this will apply the precomputed stuff, like the `M` and `W` matrices from whitening:
pp_rec = si.apply_preprocessing_pipeline(rec, my_dict, apply_precomputed_kwargs=True)
# this will ignore this stuff, and recompute the kwargs on application:
pp_rec = si.apply_preprocessing_pipeline(rec, my_dict, apply_precomputed_kwargs=False)
```

PR allow for some cool things:
1. Users can pass a single dictionary to construct a preprocessed recording (as above). Hence it completes the “dictionary workflow”; since you can use dicts in sorting, `run_sorter`,  and postprocessing in `compute`.
2. Users can easily visualise their preprocessing pipeline using the repr, including an HTML repr in Jupyter notebook
3. Increases portability between labs, and should make giving advice to users easier (from us, and from spike sorting developers), since we can just say "Oh, for KS4 NP2.0 we use this dict for preprocessing".
4. Increases the usefulness of our provenance system, since you can reconstruct human-readable preprocessing steps from the `provenance.json` file without the original recording (and worrying about paths).

The repr currently looks like this:
Screenshot 2025-05-20 at 15 33 45


src/spikeinterface/preprocessing/pipeline.py
```diff
@@ -0,0 +1,274 @@
...
+    """
+
+    def __init__(self, preprocessor_dict):
+        for preprocessor in preprocessor_dict:
+            if preprocessor not in pp_names_to_functions.keys():
+                raise TypeError(
+                    f"'{preprocessor}' is not supported by the `PreprocessingPipeline`. \
+                    To see the list of supported steps, run:\n\t>>> from spikeinterface.preprocessing \
+                    import preprocessor_dict\n\t>>> print(preprocessor_dict.keys())"
+                )
```
zm711: Wouldn't it make sense to check all preprocessors add them to a list if not possible and raise a type error with all problematic preprocessors at once rather than make the person run it and then try again only to realize they made a typo in their third key?

src/spikeinterface/preprocessing/pipeline.py
```diff
@@ -0,0 +1,274 @@
...
+
+            non_rec_kwargs = {key: value for key, value in kwargs.items() if key not in dont_apply_kwargs}
+            pp_output = pp_names_to_functions[preprocessor_name](recording, **non_rec_kwargs)
+            recording = pp_output
+
+        return recording
+
+
+def apply_pipeline(
+    recording: BaseRecording, pipeline_or_dict: dict | PreprocessingPipeline = {}, apply_precomputed_kwargs=True
```
zm711: Isn't an empty dict as the default dangerous for inplace modification? Would it make more sense to be required and not add the default?

src/spikeinterface/preprocessing/pipeline.py
```diff
@@ -0,0 +1,274 @@
...
+        with open(recording_provenance_path, "r") as f:
+            provenance_dict = json.load(f)
+    elif str(recording_provenance_path).endswith(".pkl") or str(recording_provenance_path).endswith(".pickle"):
+        import pickle
+
+        with open(recording_provenance_path, "rb") as f:
+            provenance_dict = pickle.load(f)
+
+    pipeline_dict_from_provenance = {}
+    _load_pp_from_dict(provenance_dict, pipeline_dict_from_provenance)
```
zm711: shouldn't this return something?

zm711 (commented):
I guess I'm out of the loop on this. So I'm not sure how helpful my review of code would be. I would prefer to hear what Alessio and Sam think to feel safe. I think my only concern is that dictionaries since 3.7 maintain order of insertion, but since preprocessing steps are ordered (doing filtering then whitening is not the same as whitening then filtering) I would feel more comfortable if we had some way to guarantee the order rather than relying on dictionary implementation details since they've changed once before. Right?

src/spikeinterface/preprocessing/pipeline.py
chrishalcrow: oooh, nice idea. Done!

src/spikeinterface/preprocessing/pipeline.py
chrishalcrow: yes yes yes, 100 times yes. Thanks!

src/spikeinterface/preprocessing/pipeline.py
chrishalcrow: This `_load_pp_from_dict` is a gross recursive function since our provenance json/pkl are gross recursive dicts. Due to this structure, it returns the kwargs of the recursive step it's at. So it does return something, but the important bit of the function is that it modifies `pipeline_dict_from_provenance`. It does return something, so maybe it's best practice to put that returned value somewhere (which I've done). I've also significantly updated the `_load_pp_from_dict` docstring and variable names, which should hopefully make the purpose of the function clearer.

chrishalcrow:
> I guess I'm out of the loop on this. So I'm not sure how helpful my review of code would be. I would prefer to hear what Alessio and Sam think to feel safe. I think my only concern is that dictionaries since 3.7 maintain order of insertion, but since preprocessing steps are ordered (doing filtering then whitening is not the same as whitening then filtering) I would feel more comfortable if we had some way to guarantee the order rather than relying on dictionary implementation details since they've changed once before. Right?

It was incredibly helpful - thanks!!

They made the change in 3.6, and from 3.7 it was added to the language spec ("Changed in version 3.7: Dictionary order is guaranteed to be insertion order." from the bottom of https://docs.python.org/3/library/stdtypes.html#dict), as ruled by Guido (https://mail.python.org/pipermail/python-dev/2017-December/151283.html). So I think it's pretty safe.

Not against implementing `OrderedDict`s under the hood, but I'd prefer to allow the user to pass a plain dict, to avoid them having to `from collections import OrderedDict`. If we used `OrderdDict`s in the codebase, it would be pretty easy to update if python broke dictionary ordering. But I'm personally in favour of keeping it the way it is. The tests will (sometimes) fail if ordered is removed.

doc/modules/preprocessing.rst
```diff
@@ -53,6 +53,53 @@ CMR, and save it to a binary file in the "/path/to/preprocessed" folder. The :co
 
 **NOTE:** all sorters will automatically perform the saving operation internally.
 
+The Preprocessing Pipeline
+--------------------------
+
+The module also contains the :code:`PreprocessingPipeline` object which aims to allow users to easily share pipelines across
+labs. The input to create the pipeline is a dictionary of preprocessing steps whose keys are the names of the steps
+and values are dictionaries of parameters. For example, to construct a pipeline consisting of highpass filtering
+with a minimum frequency of 250 Hz followed by whitening with default parameters, we first make the appropriate dictionary
```
alejoe91: missing detect and remove step

doc/modules/preprocessing.rst
```diff
@@ -53,6 +53,53 @@ CMR, and save it to a binary file in the "/path/to/preprocessed" folder. The :co
...
+
+Alternatively, we can construct a :code:`PreprocessingPipeline`, allowing us to investigate the pipeline before
+using it.
+
+.. code-block:: python
+
+    preprocessing_pipeline = PreprocessingPipeline(recording, preprocessing_dict)
+    # to view the pipeline:
+    preprocessing_pipeline
+
```
alejoe91, line 89: You could paste html here, right?

src/spikeinterface/preprocessing/pipeline.py
```diff
@@ -0,0 +1,293 @@
...
+    elif isinstance(pipeline_or_dict, dict):
+        pipeline = PreprocessingPipeline(pipeline_or_dict)
+    else:
+        raise TypeError("`pipeline_or_dict` must be a `PreprocessingPipeline` or a dict")
+
+    preprocessed_recording = pipeline._apply(recording, apply_precomputed_kwargs)
+    return preprocessed_recording
+
+
+def get_preprocessing_dict_from_provenance(recording_provenance_path):
```
alejoe91: I don't think this needs to be `provenance`, since you could pass it any json/pkl recording file. What about `get_preprocessing_pipeline_from_file`?

src/spikeinterface/preprocessing/pipeline.py
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        pipeline_dict[preprocessor_function.__name__] = pp_kwargs
+
+    return pipeline_dict
+
+
+def _load_pp_from_dict(prov_dict, kwargs_dict):
+    """
+    Recursive function used to iterate through a recording provenance dictionary,
```
alejoe91:
```suggestion
    Recursive function used to iterate through a recording dictionary,
```
ditto

src/spikeinterface/preprocessing/pipeline.py
```diff
@@ -0,0 +1,293 @@
...
+            for key, value in pipeline_dict_from_provenance[preprocessor].items()
+            if key not in ["recording", "parent_recording"]
+        }
+
+        pipeline_dict[preprocessor_function.__name__] = pp_kwargs
+
+    return pipeline_dict
+
+
+def _load_pp_from_dict(prov_dict, kwargs_dict):
```
alejoe91, line 283: I would make this public. It could be helpful

src/spikeinterface/preprocessing/pipeline.py
```diff
@@ -0,0 +1,293 @@
...
+    def _apply(self, recording, apply_precomputed_kwargs=False):
+        """
+        Creates a preprocessed recording by applying the `PreprocessingPipeline` to
+        `recording`.
+
+        Parameters
+        ----------
+        recording : RecordingExtractor
+            The initial recording
+        apply_precomputed_kwargs : Bool, default: False
```
alejoe91:
```suggestion
        apply_precomputed_kwargs : bool, default: False
```

src/spikeinterface/preprocessing/pipeline.py
```diff
@@ -0,0 +1,293 @@
...
+        return html_text
+
+    def _apply(self, recording, apply_precomputed_kwargs=False):
+        """
+        Creates a preprocessed recording by applying the `PreprocessingPipeline` to
+        `recording`.
+
+        Parameters
+        ----------
+        recording : RecordingExtractor
```
alejoe91:
```suggestion
        recording : BaseRecording
```

src/spikeinterface/preprocessing/pipeline.py
alejoe91, line 283: @chrishalcrow could you add a `load_preprocessing_pipeline_from_analyzer`? It could be very helpful (both from binary_folder and zarr)

alejoe91 (changes requested):
Thanks for this amazing work Chris! My main point is to make the functions more generally available to recordings, not just provenance files.

src/spikeinterface/preprocessing/pipeline.py
chrishalcrow: Yup, sounds good. 

Note: I prefer to return the `dict` than a `PreprocessingPipeline` itself since this allows the user to then use `apply_pipeline(recording, pipeline_dict)` and to share it easily. But I could be convinced otherwise!

src/spikeinterface/preprocessing/pipeline.py
chrishalcrow: I've added a `get_preprocessing_dict_from_analyzer` function plus tests.

chrishalcrow: I'm against this. Mostly because this funcitons main purpose is to modify `kwargs_dict`. So it doesn't return what you would naively expect. It's like this to make the recursive loading work. In this last round of refactoring, I've added a `_make_pipeline_dict_from_recording_dict` - maybe that could be made public?? Do you really think it's helpful?

chrishalcrow: And yes - a great idea!!

chrishalcrow:
Tests failing due to #3990

alejoe91:
LGTM! Thanks Chris this is going to be supeeeer helpful!

examples/how_to/build_pipeline_with_dicts.py
```diff
@@ -0,0 +1,194 @@
...
+#     display_name: .venv
+#     language: python
+#     name: python3
+# ---
+
+# # Build a full Sorting pipeline with dicts
+#
+# When using `SpikeInterface` there are two phases. First, you should play: try to figure out
+# any special steps or parameters you need to play with to get everything working with your
+# data. Once your happy, you then need to build a sturdy, consistent pipeline to process all
```
zm711:
```suggestion
# data. Once you're happy, you then need to build a sturdy, consistent pipeline to process all
```
:P

doc/how_to/build_pipeline_with_dicts.rst
```diff
@@ -0,0 +1,196 @@
+Build a full Sorting pipeline with dicts
+========================================
+
+When using ``SpikeInterface`` there are two phases. First, you should
+play: try to figure out any special steps or parameters you need to play
+with to get everything working with your data. Once your happy, you then
```
zm711:
```suggestion
with to get everything working with your data. Once you're happy, you then
```
```suggestion
with to get everything working with your data. Once your happy, you then
```

zm711:
Not required I guess, but we could fix both of them once tests pass so that we don't forget to fix in the future.

chrishalcrow:
Just fixed a couple of lil doc things - happy once tests pass :)

src/spikeinterface/preprocessing/pipeline.py
```diff
@@ -0,0 +1,356 @@
...
+from pathlib import Path
+import inspect
+from spikeinterface.core import BaseRecording
+from spikeinterface.core.core_tools import is_dict_extractor
+from spikeinterface.core.zarrextractors import super_zarr_open
+from spikeinterface.preprocessing.preprocessing_classes import preprocessor_dict, _all_preprocesser_dict
+
+
+pp_names_to_functions = {preprocessor.__name__: preprocessor for preprocessor in preprocessor_dict.values()}
+pp_names_to_classes = {pp_function.__name__: pp_class for pp_class, pp_function in _all_preprocesser_dict.items()}
```
zm711, line 11: As an aside if we are actually going to use this private dict then I probably shouldn't have left the typo as an homage to history :). We should consider fixing that in another PR....

zm711 (approved):
I didn't reread the code, but at least from the documentation side this looks good to me too

alejoe91:
Ok to merge for me. @samuelgarcia ok?

PR #3443, opened 2024-09-25
Allow to save recordingless analyzer as

alejoe91:
This PR allows to save an analyzer as binary or zarr even if recordingless

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -2262,15 +2262,16 @@ class AnalyzerExtension:
             extension_group.attrs["info"] = info
 
     def _save_run_info(self):
-        run_info = self.run_info.copy()
+        if self.run_info is not None:
+            run_info = self.run_info.copy()
 
-        if self.format == "binary_folder":
-            extension_folder = self._get_binary_extension_folder()
-            run_info_file = extension_folder / "run_info.json"
-            run_info_file.write_text(json.dumps(run_info, indent=4), encoding="utf8")
-        elif self.format == "zarr":
-            extension_group = self._get_zarr_extension_group(mode="r+")
-            extension_group.attrs["run_info"] = run_info
+            if self.format == "binary_folder":
+                extension_folder = self._get_binary_extension_folder()
+                run_info_file = extension_folder / "run_info.json"
+                run_info_file.write_text(json.dumps(run_info, indent=4), encoding="utf8")
+            elif self.format == "zarr":
+                extension_group = self._get_zarr_extension_group(mode="r+")
+                extension_group.attrs["run_info"] = run_info
 
     def get_pipeline_nodes(self):
         assert (
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -2023,7 +2023,10 @@ class AnalyzerExtension:
             new_extension.data = self.data
         else:
             new_extension.data = self._select_extension_data(unit_ids)
-        new_extension.run_info = self.run_info.copy()
+        if self.run_info is not None:
+            new_extension.run_info = self.run_info.copy()
+        else:
+            new_extension.run_info = None
         new_extension.save()
         return new_extension
 
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -2044,7 +2044,10 @@ class AnalyzerExtension:
         new_extension.data = self._merge_extension_data(
             merge_unit_groups, new_unit_ids, new_sorting_analyzer, keep_mask, verbose=verbose, **job_kwargs
         )
-        new_extension.run_info = self.run_info.copy()
+        if self.run_info is not None:
+            new_extension.run_info = self.run_info.copy()
+        else:
+            new_extension.run_info = None
         new_extension.save()
         return new_extension
 
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -397,7 +397,6 @@ class SortingAnalyzer:
         rec_attributes_file = folder / "recording_info" / "recording_attributes.json"
         rec_attributes_file.parent.mkdir()
         if rec_attributes is None:
-            assert recording is not None
             rec_attributes = get_rec_attributes(recording)
             rec_attributes_file.write_text(json.dumps(check_json(rec_attributes), indent=4), encoding="utf8")
             probegroup = recording.get_probegroup()
@@ -560,7 +559,6 @@ class SortingAnalyzer:
         recording_info = zarr_root.create_group("recording_info")
 
         if rec_attributes is None:
-            assert recording is not None
             rec_attributes = get_rec_attributes(recording)
             probegroup = recording.get_probegroup()
         else:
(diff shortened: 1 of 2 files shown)


chrishalcrow:
Hello, if I've already saved a `sorting_analyzer` in `my_sa` and then run the following code:

```
sa = si.load_sorting_analyzer("my_sa")
sa._recording = None
sa_zarr = sa.save_as(format="zarr", folder="my_zarr_sa")
```

Then the saved zarr folder has the correct folder structure but all the files are things like called things like `0` and `0.0`. If you save `sa_zarr` as a binary folder, it works. So something is wrong with the zarr saving.

alejoe91:
Those are the name of zarr chunks :) The `recording` will be saved as an empty dict, so it makes sense that it's few bytes. everything else should be ok (but I'll double check)

chrishalcrow:
> Those are the name of zarr chunks :) The `recording` will be saved as an empty dict, so it makes sense that it's few bytes. everything else should be ok (but I'll double check)

Oh right! Everything is named like this, even e.g. sorting_provenance?

Screenshot 2024-09-26 at 08 18 41

alejoe91:
https://probeinterface.readthedocs.io/en/main/releases/0.2.18.html

> > Those are the name of zarr chunks :) The `recording` will be saved as an empty dict, so it makes sense that it's few bytes. everything else should be ok (but I'll double check)
> 
> Oh right! ~Everything is named like this, even e.g. sorting_provenance?~ EDIT: ok I've read about zarr, cool!
> 
> Screenshot 2024-09-26 at 08 18 41

If you're interested, you can check the hidden files. You should have a `.zarray` describing the properties of the datasets and a `.zattrs` which is a dictionary with additional metadata ;)

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -2015,7 +2023,10 @@ def copy(self, new_sorting_analyzer, unit_ids=None):
             new_extension.data = self.data
         else:
             new_extension.data = self._select_extension_data(unit_ids)
-        new_extension.run_info = self.run_info.copy()
+        if self.run_info is not None:
```
chrishalcrow, line 2026: If you `import copy from copy` you can use `copy(self.run_info)` which returns `None` if `self.run_info` is `None`, which would save some lines of code and indentations here, at line 2047 and line 2268.

samuelgarcia:
I am not sure to like this.
Lets discuss before any merger please.
How do we get rec_attributes if not recording ?

alejoe91:
From the analyzer rec_attributes, which is always set! This is essential especially when loading an old waveform_extractor

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -2251,15 +2260,16 @@ def _save_importing_provenance(self):
...
-        elif self.format == "zarr":
-            extension_group = self._get_zarr_extension_group(mode="r+")
-            extension_group.attrs["run_info"] = run_info
+        if self.run_info is not None:
+            run_info = self.run_info.copy()
+
+            if self.format == "binary_folder":
+                extension_folder = self._get_binary_extension_folder()
+                run_info_file = extension_folder / "run_info.json"
+                run_info_file.write_text(json.dumps(run_info, indent=4), encoding="utf8")
```
zm711, line 2269: One question I have here in general not related to this PR is if we encode with 'utf-8' will that then mess with other encoding systems? Or do we think that the encoding was really just limited to shell script issues? Is there any place where bouncing between encodings may bite us?

alejoe91: Since we are building the dictionaries we are sure that they can be UTF-8 encoded and decoded, so I think it's safe!

zm711: What about the case for paths where the paths include Chinese or Japanese characters, which can't be encoded with `uft-8`? For run_info that doesn't matter, but I'm wondering about other json files we make. It might be that we just let people know that this works for utf-8 and may work for other encodings. But since most people on the team all use utf-8 I think it will be hard for us to know for sure for this. Doesn't really hold anything up I guess.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -352,8 +353,6 @@ def create_memory(cls, sorting, recording, sparsity, return_scaled, rec_attribut
     def create_binary_folder(cls, folder, sorting, recording, sparsity, return_scaled, rec_attributes):
         # used by create and save_as
 
-        assert recording is not None, "To create a SortingAnalyzer you need to specify the recording"
```
samuelgarcia, line 355: Could we move this to the function def create_sorting_analyzer ?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -372,11 +371,17 @@ def create_binary_folder(cls, folder, sorting, recording, sparsity, return_scale
...
+            # save recording and sorting provenance
+            if recording.check_serializability("json"):
+                recording.dump(folder / "recording.json", relative_to=folder)
+            elif recording.check_serializability("pickle"):
+                recording.dump(folder / "recording.pickle", relative_to=folder)
+        else:
+            assert rec_attributes is not None, "recording or rec_attributes must be provided"
+            # write an empty recording.json
+            with open(folder / "recording.json", mode="w") as f:
+                json.dump({}, f, indent=4)
```
samuelgarcia, line 384: I would write nothing I think.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -372,11 +371,17 @@ def create_binary_folder(cls, folder, sorting, recording, sparsity, return_scale
         # NumpyFolderSorting.write_sorting(sorting, folder / "sorting")
         sorting.save(folder=folder / "sorting")
 
-        # save recording and sorting provenance
-        if recording.check_serializability("json"):
-            recording.dump(folder / "recording.json", relative_to=folder)
-        elif recording.check_serializability("pickle"):
-            recording.dump(folder / "recording.pickle", relative_to=folder)
+        if recording is not None:
```
samuelgarcia, line 374: we should have a mutualy exclusive between recording and rec_attributes in this function at some point

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -519,20 +524,24 @@ def create_zarr(cls, folder, sorting, recording, sparsity, return_scaled, rec_at
...
+                zarr_rec = np.array([check_json(rec_dict)], dtype=object)
+                zarr_root.create_dataset("recording", data=zarr_rec, object_codec=numcodecs.JSON())
+            elif recording.check_serializability("pickle"):
+                # zarr_root.create_dataset("recording", data=rec_dict, object_codec=numcodecs.Pickle())
+                zarr_rec = np.array([rec_dict], dtype=object)
+                zarr_root.create_dataset("recording", data=zarr_rec, object_codec=numcodecs.Pickle())
+            else:
+                warnings.warn(
+                    "SortingAnalyzer with zarr : the Recording is not json serializable, the recording link will be lost for future load"
+                )
```
samuelgarcia, line 540: This warning should be also in the other format no ?

alejoe91: Is it not there? I'll check

samuelgarcia (approved):

PR #3446, opened 2024-09-27
Expose `backend_options` at the analyzer level to set `storage_options` and `saving_options`

alejoe91:
Generally, default compression works fine, but I've been hitting some errors with chunks being too large when analyzing very long NP recordings (>8h). 

This PR gives the possibility to modify the zarr arguments used to create a dataset. These can be passed at the `create_sorting_analyzer` or `save_as` step. The update test shows it in action

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -828,7 +828,7 @@ class SortingAnalyzer:
             )
             new_sorting_analyzer = SortingAnalyzer.load_from_zarr(folder, recording=recording)
             new_sorting_analyzer.folder = folder
-            new_sorting_analyzer._zarr_kwargs = zarr_kwargs
+            new_sorting_analyzer.set_zarr_kwargs(zarr_kwargs)
         else:
             raise ValueError(f"SortingAnalyzer.save: unsupported format: {format}")
 
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -219,6 +219,9 @@ class SortingAnalyzer:
         # this is used to store temporary recording
         self._temporary_recording = None
 
+        # for zarr format, we store the kwargs to create zarr datasets (e.g., compression)
+        self._zarr_kwargs = {}
+
         # extensions are not loaded at init
         self.extensions = dict()
 
@@ -500,7 +503,7 @@ class SortingAnalyzer:
         return zarr_root
 
     @classmethod
-    def create_zarr(cls, folder, sorting, recording, sparsity, return_scaled, rec_attributes):
+    def create_zarr(cls, folder, sorting, recording, sparsity, return_scaled, rec_attributes, **zarr_kwargs):
         # used by create and save_as
         import zarr
         import numcodecs
@@ -531,7 +534,8 @@ class SortingAnalyzer:
             zarr_root.create_dataset("recording", data=zarr_rec, object_codec=numcodecs.Pickle())
         else:
             warnings.warn(
-                "SortingAnalyzer with zarr : the Recording is not json serializable, the recording link will be lost for future load"
+                "SortingAnalyzer with zarr : the Recording is not json serializable, "
+                "the recording link will be lost for future load"
             )
 
         # sorting provenance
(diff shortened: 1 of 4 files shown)


src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -807,10 +824,11 @@ def _save_or_select_or_merge(
             assert folder is not None, "For format='zarr' folder must be provided"
             folder = clean_zarr_folder_name(folder)
             SortingAnalyzer.create_zarr(
-                folder, sorting_provenance, recording, sparsity, self.return_scaled, self.rec_attributes
+                folder, sorting_provenance, recording, sparsity, self.return_scaled, self.rec_attributes, **zarr_kwargs
             )
             new_sorting_analyzer = SortingAnalyzer.load_from_zarr(folder, recording=recording)
             new_sorting_analyzer.folder = folder
+            new_sorting_analyzer._zarr_kwargs = zarr_kwargs
```
alejoe91, line 831:
```suggestion
            new_sorting_analyzer.set_zarr_kwargs(zarr_kwargs)
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -531,7 +534,8 @@ def create_zarr(cls, folder, sorting, recording, sparsity, return_scaled, rec_at
             zarr_root.create_dataset("recording", data=zarr_rec, object_codec=numcodecs.Pickle())
         else:
             warnings.warn(
-                "SortingAnalyzer with zarr : the Recording is not json serializable, the recording link will be lost for future load"
+                "SortingAnalyzer with zarr : the Recording is not json serializable, "
+                "the recording link will be lost for future load"
```
zm711, line 538: This is a bit meta, but this warning isn't super helpful. Heberto has converted me to the camp that a warning should help. If I want to do zarr and I want the recording link how do I "fix" this warning. If i'm a new user I likely won't know what the pros vs cons of losing my recording link even mean. Maybe not in this PR, but I'll try to do another assert/warning PR for terminology soon.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -219,6 +219,9 @@ def __init__(
         # this is used to store temporary recording
         self._temporary_recording = None
 
+        # for zarr format, we store the kwargs to create zarr datasets (e.g., compression)
+        self._zarr_kwargs = {}
```
samuelgarcia, line 223: why not using the storage option ?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -683,7 +698,7 @@ def _save_or_select_or_merge(
         sparsity_overlap=0.75,
         verbose=False,
         new_unit_ids=None,
-        **job_kwargs,
+        **kwargs,
```
samuelgarcia, line 701: I would create a clear spearate dict here like `storage_option={}`.
we do this to create an analyzer we should do the same for the save_as() no ?

samuelgarcia:
Lets have a call for this.
I think we should have something more geenric.
Could we handle zarr specifics in on dict like the "storage_option" and propagate it at the good places ?
I known that the storage option are more the crdential and username but we could use a  universal dict for all backend specific stuff like compression also no ?
The code here is a bit too zarr specific no ?

src/spikeinterface/core/sortinganalyzer.py
alejoe91: what about `_backend_kwargs`? Storage option could be one of those, but it's not the same. It's a keyword argument for the `zarr.open`

src/spikeinterface/extractors/iblextractors.py
```diff
@@ -105,6 +105,8 @@ def get_stream_names(eid: str, cache_folder: Optional[Union[Path, str]] = None,
             An instance of the ONE API to use for data loading.
             If not provided, a default instance is created using the default parameters.
             If you need to use a specific instance, you can create it using the ONE API and pass it here.
+        stream_type : "ap" | "lf" | None, default: None
```
alejoe91, line 108: Somehow changes to this file got here from #3449. We should merge that first!

alejoe91:
@samuelgarcia I refactored it to have a more general `backend_kwargs` that could support multiple backends (also in the future). Let me know what you think, but I'm sure you'll like it :)

PR #3452, opened 2024-09-30
Fix `reset_global_job_kwargs`

alejoe91:
Fixes missing key in `reset_gloabl_job_kwrags`, that was causing full tests to fail!
https://github.com/SpikeInterface/spikeinterface/actions/runs/11092646529/job/30817949101


zm711 (approved):
Looks right to me. There should be jobs, chunk, mp, progress bar, and threads. Verbose was completely removed. So

PR #3455, opened 2024-10-01
Avoid warnings in sorting analyzer

samuelgarcia:


alejoe91:
For context, this is a weird behavior that happened due to the `run_info` addition and the QM additional computation which triggered a warning about run info being not present (because it's actually deleted priot to computing additional quality metrics).

The fix is to use the extension which is still loaded in memory

alejoe91 (approved):

PR #3456, opened 2024-10-01
fix random_spikes_selection()

### LINKED ISSUE #3450: Error at "random_spikes_selection" ?
Hello, I think I found an error at "core.sorting_tools.random_spikes_selection". 

A note about how I came to this error:
I was trying to create a sorting_analyzer from a single electrode recording with only one cluster and very few spikes (n=10). This was because of a high 'detect_threshold' value. When I lowered the value, more spikes were detected, and everything worked fine. However, I found the following mismatch when selecting random spikes:

At the step when the spike vector is converted to indices (core.sorting_tools ~ line: 200):

`spike_indices = spike_vector_to_indices(spikes, sorting.unit_ids) `

I got the following:
`
spike_indices =
{0: {1: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int64)}}`

Which makes sense given that I only have 10 spikes. 

However, when removing the spikes from the margin, the code uses sample_index, not relative indices. In other words, the variable "margin_size" which removes spikes from each border of segments, is defined in units of the number of samples to avoid the edges of the segment, but the code uses relative indexes to remove it (line 206):

```
inds_in_seg = spike_indices[segment_index][unit_id] + cum_sizes[segment_index]
if margin_size is not None:
    inds_in_seg = inds_in_seg[inds_in_seg >= margin_size]
    inds_in_seg = inds_in_seg[inds_in_seg < (num_samples[segment_index] - margin_size)] 
```
 
In my example data of 10 spikes, I got an empty array from "random_spikes_selection" because my "margin_size" is 18 (ms_after = 1.6 milliseconds at 30KHz) I don't have any index higher than 18 (my indexes are 0 to 9). But it shouldn't be based on the relative index, it should be on the sample index of the spikes.

 I think this part of the code is a legacy of using "spike_vector_to_spike_trains" instead of "spike_vector_to_indices". Because, when when I use:

`
spike_samples = spike_vector_to_spike_train(spikes, sorting.unit_ids) `

I got:

```
spike_samples = 
{0: {1: array([   251331,   3037977,   3237922,  13033056,  21827320,  24649558,
        35249946,  53788634,  70662738, 103737779], dtype=int64)}}
```

Which makes sense for the computations related to margin_size. 

The computation of removing spike indexes from the margin needs to be updated. Otherwise, it will always remove the first "margin_size" spikes and never remove spikes close to the end of the segment. 

I hope these comments make sense. 

Thanks. 
Pepe

samuelgarcia:
fix random_spikes_selection()
fix #3450

@borrepp could you check this please ?

### DIFF
--- a/src/spikeinterface/core/sorting_tools.py
+++ b/src/spikeinterface/core/sorting_tools.py
@@ -197,17 +197,21 @@ def random_spikes_selection(
         cum_sizes = np.cumsum([0] + [s.size for s in spikes])
 
         # this fast when numba
-        spike_indices = spike_vector_to_indices(spikes, sorting.unit_ids)
+        spike_indices = spike_vector_to_indices(spikes, sorting.unit_ids, absolute_index=False)
 
         random_spikes_indices = []
         for unit_index, unit_id in enumerate(sorting.unit_ids):
             all_unit_indices = []
             for segment_index in range(sorting.get_num_segments()):
-                inds_in_seg = spike_indices[segment_index][unit_id] + cum_sizes[segment_index]
+                # this is local index
+                inds_in_seg = spike_indices[segment_index][unit_id]
                 if margin_size is not None:
-                    inds_in_seg = inds_in_seg[inds_in_seg >= margin_size]
-                    inds_in_seg = inds_in_seg[inds_in_seg < (num_samples[segment_index] - margin_size)]
-                all_unit_indices.append(inds_in_seg)
+                    local_spikes = spikes[segment_index][inds_in_seg]
+                    mask = (local_spikes["sample_index"] >= margin_size) & (local_spikes["sample_index"] < (num_samples[segment_index] - margin_size))
+                    inds_in_seg = inds_in_seg[mask]
+                # go back to absolut index
+                inds_in_seg_abs = inds_in_seg + cum_sizes[segment_index]
+                all_unit_indices.append(inds_in_seg_abs)
             all_unit_indices = np.concatenate(all_unit_indices)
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/sorting_tools.py
```diff
@@ -197,17 +197,23 @@ def random_spikes_selection(
         cum_sizes = np.cumsum([0] + [s.size for s in spikes])
```
zm711, line 197: Should we think about the possibility of overflow using the `np.cumsum` here? Max number of spikes for int64 is
9,223,372,036,854,775,807, which should give us a lot of leeway. Even just doing a back of the envelope isn't does seem like we would hit the limit yet, but it is worth keeping in mind that if we have  enough spikes this could overflow.

samuelgarcia: Here it is the spikes indices not the traces shape.
It would be a good thing but I not sure we will handle 10e19 spikes very soon.

alejoe91:
@borrepp could you test that this fixes your issue?

borrepp:
> @borrepp could you test that this fixes your issue?
Hi @alejoe91 and @samuelgarcia 

I tested it and it worked fine. 

Thank you so much

src/spikeinterface/core/sorting_tools.py
zm711: Heberto and I will demand one saucisson/per overflow :P

alejoe91 (approved):

PR #3461, opened 2024-10-03
Added vspacing_factor as a param for TracesWidget

jakeswann1:
It would be useful for me (and maybe others?) to be able to adjust the vertical spacing of traces in the TracesWidget, so I've added an adjustable parameter which defaults (as before) to 1.5x max trace amplitude


zm711 (commented):
Any opinions from people on tacking this to the end in case of positional arguments? I know we are trying to force people do use keyword, but food for thought?

alejoe91:
I think it's on in this case since they are all plotting kwargs. I don't think anyone would use positional to that death! :)

alejoe91 (approved):

PR #3474, opened 2024-10-14
Sparsify the weights

yger:
We can speed up the matched filtering with sparse weights


src/spikeinterface/sortingcomponents/peak_detection.py
```diff
@@ -631,6 +631,7 @@ def __init__(
         weight_method={},
     ):
         PeakDetector.__init__(self, recording, return_output=True)
+        import scipy
```
zm711, line 634: This is a super meaty import. What if you just import scr_matrix instead?

yger: yes, of course, could be clever, I agree... I'll do that asap

samuelgarcia (approved):

samuelgarcia:
cool.

Issue #3451, opened 2024-09-29
Need to save analyzer extension's run_info before run starts (bug from #3347?)


jonahpearl:
> @ alejoe91, can you explain why this call to `self._save_run_info()` line got deleted? I think it's important to have it saved before the run starts, otherwise if the run crashes there is no run_info present to be loaded. 

 _Originally posted by @jonahpearl in [ff07ac6](https://github.com/SpikeInterface/spikeinterface/commit/ff07ac603bb8552ef40bd1f43962f201a72df17c#r147358628)_

jonahpearl:
Sorry idk how well the commenting works, but here's a screenshot of the line i'm referencing, specifically the deletion of line 2017:
<img width="779" alt="image" src="https://github.com/user-attachments/assets/e46728f2-75f1-41ac-bbd7-77993dc30bf1">

alejoe91:
The run info dict is instantiated with the object, so there is no need to save it before it has run.

Is it triggering a specific error for you? All tests are running fine, so I don't think it's crashing

jonahpearl:
It doesn't accomplish the intended use case, which was to allow the user (or the code) to check the run info dict for run_completed as being true or false in case the analyzer code crashes. As it is, one could check for the existence of the run info json file as a proxy for whether the run finished, I suppose, but that's a bit counterintuitive.

PR #3478, opened 2024-10-15
Only load `template_metrics` extension on compute if keeping some metrics

### LINKED ISSUE #3471: Error when loading template_metrics extension
I've been getting some errors when computing/loading extensions from a `SortingAnalyzer` containig a `template_metrics` extension. I calculate extensions from a dict as follows:

```py
phy_exts = {"random_spikes": {},
            "waveforms": {},
            "templates":{},
            "noise_levels": {},
            "principal_components": {"n_components":5, "mode":"by_channel_local"},
            "template_similarity": {"method": "cosine_similarity"},
            "spike_amplitudes": {},
            "quality_metrics": {},
            "template_metrics": {}}

analyzer_obj.compute(phy_exts)
```

After successfully computing waveforms, fitting PCAs and projecting waveforms, I get these errors:

```py
c:\Users\systemses\Anaconda3\envs\si_env\Lib\site-packages\spikeinterface\core\sortinganalyzer.py:1924: UserWarning: Found no run_info file for template_metrics, extension should be re-computed.
  warnings.warn(f"Found no run_info file for {self.extension_name}, extension should be re-computed.")
c:\Users\systemses\Anaconda3\envs\si_env\Lib\site-packages\spikeinterface\core\sortinganalyzer.py:1931: UserWarning: Found no run_info file for template_metrics, extension should be re-computed.
  warnings.warn(f"Found no run_info file for {self.extension_name}, extension should be re-computed.")
c:\Users\systemses\Anaconda3\envs\si_env\Lib\site-packages\spikeinterface\core\sortinganalyzer.py:2004: UserWarning: Found no data for template_metrics, extension should be re-computed.
  warnings.warn(f"Found no data for {self.extension_name}, extension should be re-computed.")
```

After which it will proceed to calculate `spike_amplitudes` successfully but then hang indefinitely.
I couldn't manually delete the folder with the `waveforms` extension, so code seemed to have hang there.

chrishalcrow:
Might fix #3471
(Could you try this out please @jonpedros)

When `template_metrics` is computed and `delete_existing_metrics = False`, any old metrics which aren't being recomputed are kept. To do this, the `_run` method loads the old template metric extension (if it exists). However, if this is the first time it has run, it has already created the extension folder. In this case `_run` sees the newly-created folder and tries to load it. Bu there's not much in it, so a no `run_info` warning appears.

To avoid this, this PR only loads the extension folder if there are metrics to be kept.

I think the previous implementation of `load_run_info` was sending two warnings if a `run_info` file didn't exist. I think I've made the logic a bit simpler. Could one of the `run_info` experts take a look please (@jonahpearl @alejoe91 ) - thanks!

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -2033,19 +2033,18 @@ class AnalyzerExtension:
         return None
 
     def load_run_info(self):
+        run_info = None
         if self.format == "binary_folder":
             extension_folder = self._get_binary_extension_folder()
             run_info_file = extension_folder / "run_info.json"
             if run_info_file.is_file():
                 with open(str(run_info_file), "r") as f:
                     run_info = json.load(f)
-            else:
-                warnings.warn(f"Found no run_info file for {self.extension_name}, extension should be re-computed.")
-                run_info = None
 
         elif self.format == "zarr":
             extension_group = self._get_zarr_extension_group(mode="r")
             run_info = extension_group.attrs.get("run_info", None)
+
         if run_info is None:
             warnings.warn(f"Found no run_info file for {self.extension_name}, extension should be re-computed.")
         self.run_info = run_info
--- a/src/spikeinterface/postprocessing/template_metrics.py
+++ b/src/spikeinterface/postprocessing/template_metrics.py
@@ -335,13 +335,18 @@ class ComputeTemplateMetrics(AnalyzerExtension):
         )
 
         existing_metrics = []
-        tm_extension = self.sorting_analyzer.get_extension("template_metrics")
-        if (
-            delete_existing_metrics is False
-            and tm_extension is not None
-            and tm_extension.data.get("metrics") is not None
-        ):
-            existing_metrics = tm_extension.params["metric_names"]
+
+        # Check if we need to propogate any old metrics. If so, we'll do that.
+        # Otherwise, we'll avoid attempting to load an empty template_metrics.
+        if set(self.params["metrics_to_compute"]) != set(self.params["metric_names"]):
+
+            tm_extension = self.sorting_analyzer.get_extension("template_metrics")
+            if (
+                delete_existing_metrics is False
+                and tm_extension is not None
+                and tm_extension.data.get("metrics") is not None
+            ):
+                existing_metrics = tm_extension.params["metric_names"]
 
         # append the metrics which were previously computed
         for metric_name in set(existing_metrics).difference(metrics_to_compute):


jonahpearl:
The `run_info` change looks fine. Incidentally, this is exactly the reason that I think it would be useful to save `run_info` before the run has occurred, and perhaps as soon as the extension is created (@alejoe91 and see #3451) — because then it's very simple to check if the extension has already been run. If not, you can skip all the `delete_existing_metrics` logic, avoid loading an empty folder, etc.

As written here, I think it's this condition ` if set(self.params["metrics_to_compute"]) != set(self.params["metric_names"])` that's being used to check if the newly created folder is empty? I suspect it would be clearer to just check `if not run_info["run_completed"]` (if we can make run_info get saved before the run happens) but lmk if I'm misunderstanding.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -2033,19 +2033,18 @@ def load(cls, sorting_analyzer):
...
     def load_run_info(self):
+        run_info = None
         if self.format == "binary_folder":
             extension_folder = self._get_binary_extension_folder()
             run_info_file = extension_folder / "run_info.json"
             if run_info_file.is_file():
                 with open(str(run_info_file), "r") as f:
                     run_info = json.load(f)
-            else:
-                warnings.warn(f"Found no run_info file for {self.extension_name}, extension should be re-computed.")
```
samuelgarcia, line 2043: The warning was here to be aware that the analyzer was computed before this run info.

chrishalcrow: Take a look at the full current implementation here: https://github.com/SpikeInterface/spikeinterface/blob/65d4b1ed28e2972e03edc3796942d7ba4445ab10/src/spikeinterface/core/sortinganalyzer.py#L2035 
If the first warning is triggered, the second is always triggered (I think!). So the first warning isn’t needed, and this is deleted here.

jonahpearl: I think the second if block was meant to be indented under the `zarr` case, but I agree it works just as well to assign it to None in the first place and then do the warning at the end if it's still None.

chrishalcrow:
Making the `run_info.json` file asap sounds reasonable to me.

> As written here, I think it's this condition ` if set(self.params["metrics_to_compute"]) != set(self.params["metric_names"])` that's being used to check if the newly created folder is empty? I suspect it would be clearer to just check `if not run_info["run_completed"]` (if we can make run_info get saved before the run happens) but lmk if I'm misunderstanding.

To be honest, it's quite a good check to do anyway: it will now only try to load the template_metrics extension it needs propagate some already-existing metrics from an old run. So in this case I'm happy with the solution.

src/spikeinterface/core/sortinganalyzer.py
chrishalcrow: Yes, I did wonder if it was a missing indentation. This solution's a tiny bit neater as we only need one copy of the warning message. I like the idea of the `run_info` in general: great suggestion @jonahpearl  :)

alejoe91:
> Making the `run_info.json` file asap sounds reasonable to me.
> 
> > As written here, I think it's this condition ` if set(self.params["metrics_to_compute"]) != set(self.params["metric_names"])` that's being used to check if the newly created folder is empty? I suspect it would be clearer to just check `if not run_info["run_completed"]` (if we can make run_info get saved before the run happens) but lmk if I'm misunderstanding.
> 
> To be honest, it's quite a good check to do anyway: it will now only try to load the template_metrics extension it needs propagate some already-existing metrics from an old run. So in this case I'm happy with the solution.

I think I removed it because it didn't play nice with loading old waveform extractor folders as an analyzer. This was part as a larger fix to improve back-compatibility, so I'm happy to test if this PR works on old data on my end!

jonpedros:
Tested it and now all the extensions I need to compute are outputted as expected. I do still get the warnings, though, but I guess that was not changed?

For `template_metrics`:
```py
[c:\Users\systemses\Anaconda3\envs\si_env\Lib\site-packages\spikeinterface\core\sortinganalyzer.py:2043](file:///C:/Users/systemses/Anaconda3/envs/si_env/Lib/site-packages/spikeinterface/core/sortinganalyzer.py:2043): UserWarning: Found no run_info file for template_metrics, extension should be re-computed.
  warnings.warn(f"Found no run_info file for {self.extension_name}, extension should be re-computed.")
[c:\Users\systemses\Anaconda3\envs\si_env\Lib\site-packages\spikeinterface\core\sortinganalyzer.py:2050](file:///C:/Users/systemses/Anaconda3/envs/si_env/Lib/site-packages/spikeinterface/core/sortinganalyzer.py:2050): UserWarning: Found no run_info file for template_metrics, extension should be re-computed.
  warnings.warn(f"Found no run_info file for {self.extension_name}, extension should be re-computed.")
[c:\Users\systemses\Anaconda3\envs\si_env\Lib\site-packages\spikeinterface\core\sortinganalyzer.py:2125](file:///C:/Users/systemses/Anaconda3/envs/si_env/Lib/site-packages/spikeinterface/core/sortinganalyzer.py:2125): UserWarning: Found no data for template_metrics, extension should be re-computed.
  warnings.warn(f"Found no data for {self.extension_name}, extension should be re-computed.")
```

For `spike_amplitudes`:
```py
[c:\Users\systemses\Anaconda3\envs\si_env\Lib\site-packages\numpy\core\_methods.py:206](file:///C:/Users/systemses/Anaconda3/envs/si_env/Lib/site-packages/numpy/core/_methods.py:206): RuntimeWarning: Degrees of freedom <= 0 for slice
  ret = _var(a, axis=axis, dtype=dtype, out=out, ddof=ddof,
[c:\Users\systemses\Anaconda3\envs\si_env\Lib\site-packages\numpy\core\_methods.py:163](file:///C:/Users/systemses/Anaconda3/envs/si_env/Lib/site-packages/numpy/core/_methods.py:163): RuntimeWarning: invalid value encountered in divide
  arrmean = um.true_divide(arrmean, div, out=arrmean,
[c:\Users\systemses\Anaconda3\envs\si_env\Lib\site-packages\numpy\core\_methods.py:198](file:///C:/Users/systemses/Anaconda3/envs/si_env/Lib/site-packages/numpy/core/_methods.py:198): RuntimeWarning: invalid value encountered in divide
  ret = ret.dtype.type(ret / rcount)
```

jonahpearl:
> Hey @jonahpearl 

@jonpedros I think you meant :)

jonpedros:
> If you had time to do that, re-run and confirm that you don't get two identical "Found no run_info file for template_metrics, extension should be re-computed" warning; that'd be great.

Tested and its gone! Only the one from spike_amplitudes remains. Thanks!

alejoe91 (approved):

Issue #3477, opened 2024-10-15
Question about changing device_channel_indices from str into int


Lingsheng-Meng:
Hi, I used OpenEphys to do recording from a single 32-channel probe. When I used probeinterface to write my probe information, the channel indices must be int. But the channel indices from OpenEphys recording is a string like 'CH1'. 
So when I add probe information to my BaseRecording, I meet the error below. Is there a method to change the device_channel_indices into int to match the probe file generated from probeinterface?
![image](https://github.com/user-attachments/assets/f17f63c1-83b5-416d-88fd-8e449cf22bc7)
![image](https://github.com/user-attachments/assets/2ee1b96b-d118-43e0-87a2-5ec59914fc37)
Thanks!

alejoe91:
Hi! `device_channel_indices` should be zero-based :) so it should be from 0 to 31

Lingsheng-Meng:
> Hi! `device_channel_indices` should be zero-based :) so it should be from 0 to 31

That works! Are the device_channel_indices automatically matching '0' to 'CH1' and '31' to 'CH32'?
Thanks for your help!

zm711:
It's not automatically matching in that sense. But because the channels are "CH1" -> "CH31" they would match in this case. As a more concrete example if I made my device channel indices `[31, 30,29...]` instead then it would do the channels in the opposite order ie the first channel in that case would be "CH31". Does that make sense how the process is working? These indices are just like doing indexing in a numpy array.

zm711:
This is specifically what I was referring to to get you or Chris to help make an illustration (like the one you started for Probe interface ages ago), since our docs just aren't doing this justice. I think we need to visually explain this both with Pandas DFs and an idealized cartoon probe->adaptor->headstage/device.

h-mayorquin:
Sadly, I won't be able to jump back into graphics until December because of all the work I have : /

If that is still needed it, happy to help at that point.

Lingsheng-Meng:
That makes sense. The device channel indices are just indices, not the channel names. Thanks for your help!

zm711:
We can close this since my PR is only tangentially related to this issue. Feel free to open more issues if things come up!

> Sadly, I won't be able to jump back into graphics until December because of all the work I have : /
If that is still needed it, happy to help at that point.

No worries, I know you all stay busy. I'm just terrible at art. So we can revisit later if needed :)

PR #3479, opened 2024-10-15
Improve error message for `set_probegroup`/`set_probe`

zm711:
Hey @alejoe91 the confusion in #3477 has come up multiple times. How about this as a change:

1) we check if the list is the wrong length (this is a rare mistake)
2) we see if they forgot to 0-index the list (this is the common mistake)

Happy to workshop this a bit if needed :) 

Just for clarity the current error saying you made a mistake because you have 32 values and you should have 32 values is super confusing, but the problem is that is actually the max value.

### DIFF
--- a/src/spikeinterface/core/baserecordingsnippets.py
+++ b/src/spikeinterface/core/baserecordingsnippets.py
@@ -172,8 +172,10 @@ class BaseRecordingSnippets(BaseExtractor):
         number_of_device_channel_indices = np.max(list(device_channel_indices) + [0])
         if number_of_device_channel_indices >= self.get_num_channels():
             error_msg = (
-                f"The given Probe have 'device_channel_indices' that do not match channel count \n"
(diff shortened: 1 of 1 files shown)


alejoe91 (approved):

h-mayorquin:
Related to the discussion today. This would be clearer with a map of ids to contact ids where we could specify which contacts are not mapped I think.

Because we are doing indices the error is harder to pint-point.

zm711:
Yeah we should think about that in general more. From my perspective documentation is the most important thing. This is error-prone because it is tedious, not because it is hard. So if people better understand the steps then we are in better shape. One issue that came up in the meeting today is that the reversed but similar concept in Kilosort in Matlab is one-indexed so people thinking this is a similar idea to KS will likely do 1-32 instead of 0-31 values. I think we also allow for disconnects by including -1 values which is also not super well documented everywhere.

h-mayorquin:
Uh, I forgot about the disconnects.

samuelgarcia (approved):

PR #3480, opened 2024-10-15
Cap python 3.13 until we can remove the cap on numpy <2.0

h-mayorquin:
From the meeting.  @zm711 please correct if I got anything wrong
python 3.13 only supports numpy 2.0 on windows.
We can't support numpy 2.0 because of neo issues.

We are adding this temporary cap to avoid users getting a complicated error until we fix neo numpy 2.0 support.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ authors = [
 ]
 description = "Python toolkit for analysis, visualization, and comparison of spike sorting output"
 readme = "README.md"
-requires-python = ">=3.9,<4.0"
+requires-python = ">=3.9,<3.14"  # Only numpy 2.0 supported on python 3.12 for windows. We need to wait for fix on neo
 classifiers = [
     "Programming Language :: Python :: 3 :: Only",
     "License :: OSI Approved :: MIT License",


pyproject.toml
```diff
@@ -7,7 +7,7 @@ authors = [
 ]
 description = "Python toolkit for analysis, visualization, and comparison of spike sorting output"
 readme = "README.md"
-requires-python = ">=3.9,<4.0"
+requires-python = ">=3.9,<3.14"  # Only numpy 2.0 supported on python 3.12 for windows. We need to wait for fix on neo
```
zm711, line 10:
```suggestion
requires-python = ">=3.9,<3.13"  # Only numpy 2.0 supported on python 3.12 for windows. We need to wait for fix on neo
```

This still allows 3.13 :)

h-mayorquin: Thanks, silly mistake of mine!

h-mayorquin:
We should set the token! Let's do this on the next maintenance meeting.

pyproject.toml
```diff
@@ -7,7 +7,7 @@ authors = [
 ]
 description = "Python toolkit for analysis, visualization, and comparison of spike sorting output"
 readme = "README.md"
-requires-python = ">=3.9,<4.0"
+requires-python = ">=3.9,<3.13"  # Only numpy 2.0 supported on python 3.13 for windows. We need to wait for fix on neo
```
zm711, line 10:
```suggestion
requires-python = ">=3.9,<3.13"  # Only numpy 2.1 supported on python 3.13 for windows. We need to wait for fix on neo
```

Actually if we want to be super accurate (when I explained this I did a lazy job), it is actually 2.0 is good for python 3.9-3.12 and then 2.1 is when 3.13 support was added. Really based on my neo testing there was no issue adding 2.1 at the exact same time as 2.0, so this doesn't practically matter, but if we want the comment to be accurate then technically we should say 2.1. Although based on the Neo CI 3.13 did work with 2.0 on Linux it just isn't officially supported.

h-mayorquin: Thanks. I read the issue on numpy that you linked on neo. It left me a bit confused about the whole thing. Might have to check it twice.

alejoe91 (approved):

Issue #3448, opened 2024-09-27
How to use NEO format in SpikeInterface?


DerfelLee:
Hi! I'm trying to build a pipeline combining different python packages including SpikeInterface. I want to read the data using 'neo.rawio' since it's supported by SpikeInterface. I wonder how I can put a 'neo.rawio' data into SpikeInterface? Also, after using the SpikeInterface, how can I export the data back to NEO format?

I saw someone discussing creating an 'export_to_neo' function in 2019, but it seems like this is not in the API or code.

Thanks!

DerfelLee:
I'm only using the recording right now, but I only saw NumpySorting has a 'from_neo_spiketrain' function

zm711:
Basically all of neo.rawio is incorporated under the hood inside the `extractors` module. So you would do

```python
import spikeinterface.extractors as se
recording = se.read_xx()
```
where the xx is the format you want. For example blackrock, intan, neuralynx, spikeglx, etc. See all formats [here](https://spikeinterface.readthedocs.io/en/stable/modules/extractors.html). 

When you use the `se.read_xx` function (for example `se.read_intan()` you are actually using the neo.rawio api under the hood. If this doesn't answer your question could you provide more context about which format you want to use?

DerfelLee:
I've read these functions in the documents, but they seem to read the file into SpikeInterface formats. I want to use SpikeInterface with other Python packages like Elephant without converting and writing the file on disk, which may save me a lot of time and make things easier. Therefore, I just want to ask if I can just simply use functions to convert NEO-loaded files with SpikeInterface Extractors.

zm711:
I think you'll need to explain the actual work flow you want. SpikeInterface Recording objects are also lazy when possible. If you want to sort data then it will need to be in a recording objects and some sorters require files to be written to disk to be used (for example KS1-3, MS5) so you'll have to write files anyway. If you want to do spike train analysis, we haven't gotten around to writing a SpikeInterface->Neo spiketrain function yet, but you could make it yourself by extracting the spike train information from a `Sorting` and then initializing your own Neo spiketrain. But without understanding what you specifically are hoping to do we can't really give you more specific tips.

DerfelLee:
[...]

PR #3482, opened 2024-10-15
Remove Non-Serializable writing of json

zm711:
The note says remove after 0.100 and this caused an issue recently #3448 with the error not being super clear (because recording was None) the  error the user saw was that None didn't have get_num_channels). Since this isn't supported we shouldn't write this here, right? I have run local tests because I need to run, so will update after the CI if needed.

### DIFF
--- a/src/spikeinterface/sorters/basesorter.py
+++ b/src/spikeinterface/sorters/basesorter.py
@@ -145,9 +145,8 @@ class BaseSorter:
         elif recording.check_serializability("pickle"):
             recording.dump(output_folder / "spikeinterface_recording.pickle", relative_to=output_folder)
         else:
-            # TODO: deprecate and finally remove this after 0.100
-            d = {"warning": "The recording is not serializable to json"}
-            rec_file.write_text(json.dumps(d, indent=4), encoding="utf8")
+            raise RuntimeError("This recording is not serializable and so can not be sorted. Consider `recording.save()` to save a "
+                               "compatible binary file.")
 
         return output_folder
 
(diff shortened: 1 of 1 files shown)


alejoe91 (approved):

src/spikeinterface/sorters/basesorter.py
```diff
@@ -145,9 +145,10 @@ def initialize_folder(cls, recording, output_folder, verbose, remove_existing_fo
         elif recording.check_serializability("pickle"):
             recording.dump(output_folder / "spikeinterface_recording.pickle", relative_to=output_folder)
         else:
-            # TODO: deprecate and finally remove this after 0.100
-            d = {"warning": "The recording is not serializable to json"}
-            rec_file.write_text(json.dumps(d, indent=4), encoding="utf8")
+            raise RuntimeError(
+                "This recording is not serializable and so can not be sorted. Consider `recording.save()` to save a "
+                "compatible binary file."
+            )
```
samuelgarcia, line 151: I think we should keep the possibility of running a sorting on a recording living only in memory.
And so no dump into a file.
This could be usefull for benchmark of simulated in memeory.
So I would `pass` silently no ?

zm711: This was inspired by the issue I linked where we asked you to comment. fine to let it pass if you're okay with it, but do you want to read that issue and comment there or here on your opinions about it.

zm711: Based on your comment in the issue what I would argue is we add this runtime error but with a message like:

```
This recording is not serializable and so can not be sorted. Consider recording.save() to generate a compatible binary file. We plan to support in memory sorting for some sorters in the future
```

zm711: Sam and I discussed this and he is okay with the error for now. Any opinions on adding either 

1) a comment saying we should support in-memory in the future
2) saying in the error message that we plan to support in the future

@alejoe91

alejoe91: No I think it's fine to as is :)

PR #3487, opened 2024-10-18
Improve merging and iterative merging

yger:
This branch depends on #3435 , and adds the possibility to benchmark only meta merging, i.e. the final steps of a sorting procedure. Such step is crucial in the context of drift-corrected recordings. Note that it will replace #2934 that will be closed, since now the code is easier to read, cleaner, and more in line with @samuelgarcia point of view. It can still be discussed however, to know if we should do a proper merging step for the sortingcomponents. I think we should, because several options could be tried and benchmark. 

Functions to split by times and/or amplitudes have been rewritten without any dependencies, in a separate file as requested in #2934 (splitting_tools.py in generation)


src/spikeinterface/generation/splitting_tools.py
```diff
@@ -0,0 +1,139 @@
...
+    else:
+        to_split_ids = unit_ids
+
+    spikes = sa.sorting.to_spike_vector()
+    new_spikes = spikes.copy()
+    max_index = np.max(spikes["unit_index"])
+    new_unit_ids = list(sa.sorting.unit_ids.copy())
+    splitted_pairs = []
+    for unit_id in to_split_ids:
+        ind_mask = spikes["unit_index"] == sa.sorting.id_to_index(unit_id)
```
samuelgarcia, line 58: This is very slow you should use spike_vector_to_indices() from sorting tools.py

src/spikeinterface/generation/splitting_tools.py
```diff
@@ -0,0 +1,139 @@
...
+        If specified, only cells with a snr higher than min_snr might be splitted
+    seed : int | None, default: None
+        The seed for random generator.
+
+    Returns
+    -------
+    new_sorting, splitted_pairs : The new splitted sorting, and the pairs that have been splitted
+    """
+
+    sa = sorting_analyzer
```
samuelgarcia, line 32: you should use analyzer instead of sa for readability no ?

src/spikeinterface/generation/splitting_tools.py
```diff
@@ -0,0 +1,139 @@
+import numpy as np
+from spikeinterface.core.numpyextractors import NumpySorting
+
+
+def split_sorting_by_times(
```
samuelgarcia, line 5: this is really a good idea.

src/spikeinterface/sortingcomponents/merging/main.py
```diff
@@ -0,0 +1,109 @@
...
+import numpy as np
+from spikeinterface.core.sortinganalyzer import create_sorting_analyzer
+from spikeinterface.core.sparsity import ChannelSparsity
+from spikeinterface.core.analyzer_extension_core import ComputeTemplates
+
+
+merging_methods = ["circus", "auto_merges"]
+
+
+def create_sorting_analyzer_with_templates(sorting, recording, templates, remove_empty=True):
```
samuelgarcia, line 12: If this is a need we should have this function somewhere else

src/spikeinterface/sortingcomponents/merging/main.py
```diff
@@ -0,0 +1,109 @@
...
+        non_empty_sorting = sorting
+
+    sa = create_sorting_analyzer(non_empty_sorting, recording, format="memory", sparsity=sparsity)
+    sa.extensions["templates"] = ComputeTemplates(sa)
+    sa.extensions["templates"].params = {"ms_before": templates.ms_before, "ms_after": templates.ms_after}
+    sa.extensions["templates"].data["average"] = templates_array
+    return sa
+
+
+def merging_circus(
```
samuelgarcia, line 33: move this to sc2

src/spikeinterface/sortingcomponents/merging/main.py
```diff
@@ -0,0 +1,109 @@
...
+
+    template_diff_thresh = np.arange(0.05, 0.25, 0.05)
+    presets_params = [{"template_similarity": {"template_diff_thresh": i}} for i in template_diff_thresh]
+    presets = ["x_contaminations"] * len(template_diff_thresh)
+    return iterative_merges(
+        sorting_analyzer, presets=presets, presets_params=presets_params, extra_outputs=extra_outputs, **job_kwargs
+    )
+
+
+def merge_spikes(
```
samuelgarcia, line 56: I think we should use only the 
`merged_analyzer = auto_merge(analyzer, ...)`
which I have push in #3435 because it do the same.
The final sorting is accessed by `merged_analyzer.sorting`.

src/spikeinterface/sortingcomponents/merging/__init__.py
```diff
@@ -0,0 +1 @@
+from .main import merge_spikes
```
samuelgarcia, line 1: I would remove this and use only the curation module.
I think we all the machinery there.

src/spikeinterface/benchmark/benchmark_merging.py
```diff
@@ -0,0 +1,158 @@
...
+    plot_unit_templates,
+    plot_amplitudes,
+    plot_crosscorrelograms,
+)
+
+import numpy as np
+from .benchmark_base import Benchmark, BenchmarkStudy
+
+
+class MergingBenchmark(Benchmark):
```
samuelgarcia, line 16: super chouette

src/spikeinterface/generation/splitting_tools.py
yger: This has beeen fixes, and tests have been added

src/spikeinterface/generation/splitting_tools.py
yger: Done

src/spikeinterface/benchmark/benchmark_merging.py
```diff
@@ -0,0 +1,158 @@
...
+        self.recording = recording
+        self.splitted_sorting = splitted_sorting
+        self.method = params["method"]
+        self.gt_sorting = gt_sorting
+        self.splitted_cells = splitted_cells
+        self.method_kwargs = params["method_kwargs"]
+        self.result = {}
+
+    def run(self, **job_kwargs):
+        self.result["sorting"], self.result["merges"], self.result["outs"] = merge_spikes(
```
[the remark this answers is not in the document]
yger: I can not, because I want to use a higher level function here that can launch iteratively compute_merge_unit_groups. So only think I can do is to can modify auto_merge_units into auto_merge_units_iterative (and the two will be unified), and use this function in the benchmark

yger:
This one is also almost good after the discussion we had yesterday @samuelgarcia @alejoe91. Now the auto_merge_units() function can handle lists of presets, and the benchmark is also there and working. Some questions:
- are all default params optimal for extension (similarity uses cosine, no lag for example)
- should we add to auto_merge_units the notion of "strategies", i.e. for example if someone has a good list of presets and want to use it?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -812,6 +812,57 @@ def get_sorting_property(self, key: str, ids: Optional[Iterable] = None) -> np.n
         """
         return self.sorting.get_property(key, ids=ids)
 
+    def can_perform_merges(
```
zm711, line 815: what about `is_mergeable`? That fits with things like `is_scaled` `has_extension`?  Others have opinions?

src/spikeinterface/curation/auto_merge.py
```diff
@@ -369,19 +371,123 @@ def compute_merge_unit_groups(
         return merge_unit_groups
 
 
-def auto_merge_units(
-    sorting_analyzer: SortingAnalyzer, compute_merge_kwargs: dict = {}, apply_merge_kwargs: dict = {}, **job_kwargs
+def resolve_pairs(existing_merges, new_merges):
```
zm711, line 374: did this also take care of the identity pairs that were generated at some point?

zm711 (commented):
couple initial comments.

src/spikeinterface/core/sortinganalyzer.py
yger: let's go for is_mergeable, this is indeed better.

src/spikeinterface/curation/auto_merge.py
yger: This has been taken care of, but this is not in this function. I've extented the docstrings, but this function is mostly only for internal usage and debug purposes

src/spikeinterface/curation/auto_merge.py
```diff
@@ -356,6 +355,9 @@ def compute_merge_unit_groups(
             )
             outs["pairs_decreased_score"] = pairs_decreased_score
 
+        # ind1, ind2 = np.nonzero(pair_mask)
```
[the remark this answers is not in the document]
yger: yes

doc/modules/curation.rst
```diff
@@ -153,6 +153,42 @@ merges. It offers multiple "presets" and the flexibility to apply individual ste
     # here we apply the merges
     analyzer_merged = analyzer.merge_units(merge_unit_groups=merge_unit_groups)
 
+There is also the convenient :py:func:`~spikeinterface.curation.auto_merge_units` function that combines the
+:py:func:`~spikeinterface.curation.compute_merge_unit_groups` and :py:func:`~spikeinterface.core.SortingAnalyzer.merge_units` functions.
+This is a high level function that allows you to apply either one or several presets/lists of steps in one go. For example, let's
```
zm711, line 158: We probably need to add a note that this is a "permanent" merge so that users know they will return a new analyzer that cannot be used to recreate the old one right? I think that would be a very annoying accident if I did:
```python
analyzer = analyzer.merge_units()
```
and then found out that I had lost my old analyzer. People are still used to curation with phy where some number of steps can be reversed.

yger: I've added a bold comment at the end, but you are right. However, by default in this function, a copy will be done in memory, so even if you might loose the reference to your object, the folders themselves will not be modified because it will ony be in memory. If you want to explicitly overwrite everything, then there is a flag in the arguments

doc/modules/curation.rst
```diff
@@ -153,6 +153,43 @@ merges. It offers multiple "presets" and the flexibility to apply individual ste
...
+    from spikeinterface import create_sorting_analyzer
+    from spikeinterface.curation import auto_merge_units
+
+    analyzer = create_sorting_analyzer(sorting=sorting, recording=recording)
+
+    # some extensions are required
+    analyzer.compute(["random_spikes", "templates", "template_similarity", "correlograms"])
+    analyzer.compute("unit_locations", method="monopolar_triangulation")
+
+    presets = ["x_contaminations"] * len(template_diff_thresh)
```
alejoe91, line 175: `template_diff_thresh` is not defined right?

yger: Indeed, good catch

src/spikeinterface/curation/tests/test_auto_merge.py
```diff
@@ -81,60 +82,82 @@ def test_compute_merge_unit_groups(sorting_analyzer_for_curation, preset):
...
+
+
+def test_auto_merge_units(sorting_analyzer_for_curation):
+    recording = sorting_analyzer_for_curation.recording
+    job_kwargs = dict(n_jobs=-1)
+    new_sorting, _ = split_sorting_by_times(sorting_analyzer_for_curation)
+    new_sorting_analyzer = create_sorting_analyzer(new_sorting, recording, format="memory")
+    merged_analyzer = auto_merge_units(new_sorting_analyzer, presets="x_contaminations", **job_kwargs)
+    assert len(merged_analyzer.unit_ids) < len(new_sorting_analyzer.unit_ids)
+
```
zm711, line 145:
```suggestion
    merged_analyzer = auto_merge_units(new_sorting_analyzer, presets="x_contaminations", **job_kwargs)
    assert len(merged_analyzer.unit_ids) < len(new_sorting_analyzer.unit_ids)
    
    # test for steps integration
    step_merged_analyzer = auto_merge_units(new_sorting_analyzer, steps = ["num_spikes", "removed_contaminated", "unit_locations", "template_similarity", "quality_scores"], step_params = dict(num_spikes=150), **job_kwargs)
    assert len(step_merged_analyzer.unit_ids) < len(new_sorting_analyzer.unit_ids)

```

what about something like this? Might have a typo so we may need to fix since it is hard to code in this code review window? My goal here is to :
1) test for steps instead of presets to make sure the steps flow is sound
2) test changing a preset in a way I imagine a user might want to do

yger: This has been done, I've added the test and I consolidated the functions such that it could be used the way you thought

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -812,6 +812,57 @@ def get_sorting_property(self, key: str, ids: Optional[Iterable] = None) -> np.n
         """
         return self.sorting.get_property(key, ids=ids)
 
+    def is_mergeable(
```
alejoe91, line 815:
```suggestion
    def are_units_mergeable(
```

src/spikeinterface/curation/auto_merge.py
```diff
@@ -369,19 +368,126 @@ def compute_merge_unit_groups(
...
     )
 
-    merged_analyzer = sorting_analyzer.merge_units(merge_unit_groups, **apply_merge_kwargs, **job_kwargs)
-    return merged_analyzer
+    if extra_outputs:
+        merge_unit_groups, outs = merge_unit_groups
+
+    merged_units = len(merge_unit_groups) > 0
+    if merged_units:
+        is_mergeable = sorting_analyzer.is_mergeable(merge_unit_groups, **apply_merge_kwargs)
```
alejoe91, line 467: See comment above. If the `sorting_analyzer.is_mergeable` returned a dictionary with mergeability for each merge_unit_group, then here we can be much more "surgical" and just skip groups that cannot be merged. What do you think @yger ?

src/spikeinterface/curation/auto_merge.py
```diff
@@ -572,6 +678,138 @@ def get_potential_auto_merge(
...
+    steps: list[str] | None = None,
+    apply_merge_kwargs: dict = {},
+    recursive: bool = False,
+    merging_strategy: str = "warning",
+    extra_outputs: bool = False,
+    force_copy: bool = True,
+    **job_kwargs,
+) -> SortingAnalyzer:
+    """
+    Wrapper to conveniently be able to launch several presets for auto_merge_units in a row, as a list.
```
alejoe91, line 694:
```suggestion
    Automatically finds and apply merges.
    This function enables one to launch several merging presets in sequence and also to apply each
    step recursively.
```
I think we shouls make this docstring more high level since this will be the main merging entry point :)

src/spikeinterface/curation/auto_merge.py
yger: This has been implemented. Now the functions are_units_mergeables return a dict whose keys are the new unit ids, and values boolean telling if the merge is possible or not

src/spikeinterface/curation/auto_merge.py
```diff
@@ -369,19 +368,129 @@ def compute_merge_unit_groups(
...
+        for new_unit_id, merge_group in zip(mergeable_units.keys(), merge_unit_groups):
+            if not mergeable_units[new_unit_id]:
+                if raise_error:
+                    raise ValueError(
+                        f"Units {merge_group} can not be merged with the current sparsity_threshold. Merging is stopped"
+                    )
+                else:
+                    warnings.warn(
+                        "Units {merge_group} can not be merged with the current sparisty_threshold. Merging is skipped",
+                        UserWarning,
```
alejoe91, line 477:
```suggestion
                        f"Units {merge_group} can not be merged with the current sparisty_threshold. Merging is skipped"
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -812,6 +812,70 @@ def get_sorting_property(self, key: str, ids: Optional[Iterable] = None) -> np.n
...
+        -------
+        can_perform_merges : bool
+            If soft merging can be performed.
+        """
+        assert merging_mode in ["hard", "soft"], "merging_mode must be 'hard' or 'soft'"
+
+        if new_unit_ids is None:
+            new_unit_ids = generate_unit_ids_for_merge_group(
+                self.unit_ids, merge_unit_groups, new_unit_ids, new_id_strategy
+            )
```
alejoe91, line 847: Why do we need `new_unit_ids` here? We can just retrun the dict whose keys are the merge unit group!

yger: Not sure, because this is a bit nested. new_unit_ids are needed to check what are the new unit ids, in all_unit_ids, depending on the new_id_strategy

alejoe91 (approved):

yger:
I've changed sparsity_threshold to sparsity_overlap for consistency, and remove some changes related to clustering in SC2 and that will go in one other PR. Looks good to me and seems to work

PR #3493, opened 2024-10-21
Allow `export_report` to run without waveforms

chrishalcrow:
Small edit to `export_report` so that it will generate a report without waveforms.

When we're being speedy, we often don't compute or don't copy over waveforms. This PR just changes `export_report` so that it does/doesn't include the `UnitWaveformDensityMapWidget` if the sorting_analyzer does/doesn't have waveforms computed.

A unit report without waveforms:

computed using

```
rec, sort = si.generate_ground_truth_recording()
sa.compute(["random_spikes",  "templates", "spike_amplitudes", "correlograms", "template_similarity"])
si.export_report(sorting_analyzer=sa, output_folder="my_report")
```

### DIFF
--- a/src/spikeinterface/widgets/unit_summary.py
+++ b/src/spikeinterface/widgets/unit_summary.py
@@ -108,7 +108,7 @@ class UnitSummaryWidget(BaseWidget):
         fig = self.figure
         nrows = 2
         ncols = 2
-        if sorting_analyzer.has_extension("correlograms") and sorting_analyzer.has_extension("template_similarity"):
+        if sorting_analyzer.has_extension("correlograms"):
             ncols += 1
         if sorting_analyzer.has_extension("waveforms"):
             ncols += 1
@@ -172,7 +172,7 @@ class UnitSummaryWidget(BaseWidget):
             col_counter += 1
             ax_waveform_density.set_ylabel(None)
 
-        if sorting_analyzer.has_extension("correlograms") and sorting_analyzer.has_extension("template_similarity"):
+        if sorting_analyzer.has_extension("correlograms"):
             ax_correlograms = fig.add_subplot(gs[:2, col_counter])
             AutoCorrelogramsWidget(
                 sorting_analyzer,
--- a/src/spikeinterface/widgets/unit_summary.py
+++ b/src/spikeinterface/widgets/unit_summary.py
@@ -108,7 +108,7 @@ class UnitSummaryWidget(BaseWidget):
         fig = self.figure
         nrows = 2
         ncols = 2
-        if sorting_analyzer.has_extension("correlograms") or sorting_analyzer.has_extension("spike_amplitudes"):
+        if sorting_analyzer.has_extension("correlograms") and sorting_analyzer.has_extension("template_similarity"):
             ncols += 1
         if sorting_analyzer.has_extension("waveforms"):
             ncols += 1
@@ -117,31 +117,30 @@ class UnitSummaryWidget(BaseWidget):
         gs = fig.add_gridspec(nrows, ncols)
         col_counter = 0
 
-        if sorting_analyzer.has_extension("unit_locations"):
-            ax1 = fig.add_subplot(gs[:2, col_counter])
-            # UnitLocationsPlotter().do_plot(dp.plot_data_unit_locations, ax=ax1)
-            w = UnitLocationsWidget(
-                sorting_analyzer,
-                unit_ids=[unit_id],
-                unit_colors=unit_colors,
-                plot_legend=False,
-                backend="matplotlib",
-                ax=ax1,
-                **unitlocationswidget_kwargs,
-            )
-            col_counter = col_counter + 1
+        # Unit locations and unit waveform plots are always generated
+        ax_unit_locations = fig.add_subplot(gs[:2, col_counter])
+        _ = UnitLocationsWidget(
+            sorting_analyzer,
+            unit_ids=[unit_id],
+            unit_colors=unit_colors,
+            plot_legend=False,
+            backend="matplotlib",
+            ax=ax_unit_locations,
+            **unitlocationswidget_kwargs,
+        )
+        col_counter += 1
 
-            unit_locations = sorting_analyzer.get_extension("unit_locations").get_data(outputs="by_unit")
-            unit_location = unit_locations[unit_id]
-            x, y = unit_location[0], unit_location[1]
-            ax1.set_xlim(x - 80, x + 80)
-            ax1.set_ylim(y - 250, y + 250)
-            ax1.set_xticks([])
-            ax1.set_xlabel(None)
-            ax1.set_ylabel(None)
+        unit_locations = sorting_analyzer.get_extension("unit_locations").get_data(outputs="by_unit")
+        unit_location = unit_locations[unit_id]
+        x, y = unit_location[0], unit_location[1]
+        ax_unit_locations.set_xlim(x - 80, x + 80)
+        ax_unit_locations.set_ylim(y - 250, y + 250)
+        ax_unit_locations.set_xticks([])
+        ax_unit_locations.set_xlabel(None)
+        ax_unit_locations.set_ylabel(None)
 
-        ax2 = fig.add_subplot(gs[:2, col_counter])
-        w = UnitWaveformsWidget(
+        ax_unit_waveforms = fig.add_subplot(gs[:2, col_counter])
+        _ = UnitWaveformsWidget(
             sorting_analyzer,
             unit_ids=[unit_id],
             unit_colors=unit_colors,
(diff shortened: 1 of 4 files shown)


zm711:
I think this is an improvement. Another request we should think about is to have more flexibility in selecting what to plot in general. There are complaints about the slowness of making the summary when you only want some features to be saved. So something like this could be extended to help speed things up for people.

samuelgarcia (approved):

samuelgarcia:
Good idea. thanks Chris.

chrishalcrow:
> I think this is an improvement. Another request we should think about is to have more flexibility in selecting what to plot in general. There are complaints about the slowness of making the summary when you only want some features to be saved. So something like this could be extended to help speed things up for people.

Mmm, good idea. So at the moment we include 5 widgets. Your suggestion would be to allow the user to select which ones they'd like, and we format the report appropriately? Sounds doable (and extendable for other widgets!) in another PR.

zm711:
> in another PR.

agreed. :)

zm711 (approved):

src/spikeinterface/widgets/unit_summary.py
```diff
@@ -136,37 +140,41 @@ def plot_matplotlib(self, data_plot, **backend_kwargs):
...
+                unit_ids=[unit_id],
+                unit_colors=unit_colors,
+                use_max_channel=True,
+                same_axis=False,
+                backend="matplotlib",
+                ax=ax3,
+                **unitwaveformdensitymapwidget_kwargs,
+            )
+            ax3.set_ylabel(None)
+            col_counter = col_counter + 1
```
JoeZiminski, line 174: these could be `col_counter += 1`

src/spikeinterface/widgets/unit_summary.py
```diff
@@ -107,15 +107,18 @@ def plot_matplotlib(self, data_plot, **backend_kwargs):
...
+        if sorting_analyzer.has_extension("waveforms"):
+            ncols += 1
         if sorting_analyzer.has_extension("spike_amplitudes"):
             nrows += 1
         gs = fig.add_gridspec(nrows, ncols)
+        col_counter = 0
 
         if sorting_analyzer.has_extension("unit_locations"):
-            ax1 = fig.add_subplot(gs[:2, 0])
+            ax1 = fig.add_subplot(gs[:2, col_counter])
```
JoeZiminski, line 121: ATM this is `ax1` and the next axis is `ax2` but if `unit_locations` is not found then `ax2` will actually be the first axis. Is it worth naming these by their plot type? e.g. `unit_locations_axis` (but maybe this is confusing in other ways as you loose any sense of order)

src/spikeinterface/widgets/unit_summary.py
```diff
@@ -107,15 +107,18 @@ def plot_matplotlib(self, data_plot, **backend_kwargs):
         # and use custum grid spec
         fig = self.figure
         nrows = 2
-        ncols = 3
+        ncols = 2
```
JoeZiminski, line 110: It might be worth expanding the docs just to briefly introduce what things are plot and under what conditions. For example are some plots conditional on the steps run while others will always be plot? I got a little confused here because `unit_locations` is under a conditional but it's inclusion doesn't add to the number `ncols`. Maybe the initial `ncols=2` is reserved for `unit_locations`? but then I don't think `unit_locations` need to be under the conditional if it will always be plot. I tried skipping unit locations and there was some unused space in the plot so maybe this is the case 🤔

JoeZiminski: (I just read the diff more carefully and see this is not actually related to your changes in this PR, sorry! but maybe you know the answer

src/spikeinterface/widgets/unit_summary.py
```diff
@@ -136,37 +140,41 @@ def plot_matplotlib(self, data_plot, **backend_kwargs):
...
             ax1.set_ylabel(None)
 
-        ax2 = fig.add_subplot(gs[:2, 1])
+        ax2 = fig.add_subplot(gs[:2, col_counter])
         w = UnitWaveformsWidget(
             sorting_analyzer,
             unit_ids=[unit_id],
             unit_colors=unit_colors,
             plot_templates=True,
+            plot_waveforms=sorting_analyzer.has_extension("waveforms"),
```
JoeZiminski, line 149: If `UnitWaveformsWidget` is the only plot that is always plot, could it be moved to the top of this section? This would help make it clearer what plots are conditional and what plots are always plot (i.e. always plot at the top, then conditional).

src/spikeinterface/widgets/unit_summary.py
```diff
@@ -107,15 +107,18 @@ def plot_matplotlib(self, data_plot, **backend_kwargs):
         # and use custum grid spec
         fig = self.figure
         nrows = 2
-        ncols = 3
+        ncols = 2
         if sorting_analyzer.has_extension("correlograms") or sorting_analyzer.has_extension("spike_amplitudes"):
             ncols += 1
+        if sorting_analyzer.has_extension("waveforms"):
```
JoeZiminski, line 113: Sorry not related to this line but just above, two small things not from this diff but now seems a good time to fix:

- `            # UnitLocationsPlotter().do_plot(dp.plot_data_unit_locations, ax=ax1)` this commented line could be removed
- the variables `w` as output from both calls to `UnitLocationsWidget` is unused and could be removed.

src/spikeinterface/widgets/unit_summary.py
JoeZiminski: Actually I guess this doesn't work if this plot should be after the first `UnitWaveformsWidget` (if it exists). Maybe a comment above like `#This plot is always displayed, if 'unit_locations' exists it will be plot in the column before it`.

JoeZiminski (approved):
Nice @chrishalcrow this is really cool, the export report is a great feature and making it more flexible is very useful. Also agree that custom selecting of subplots would be very nice (in another PR). 

I ran it and it all worked nicely. Some small changes suggested, apologies some are outside this diff as I was reading the code in another editor, but I guess now would be a good time to change them.

chrishalcrow:
Thanks @JoeZiminski - will take a look on Monday!

src/spikeinterface/widgets/unit_summary.py
chrishalcrow: Hello - just noticed that at the moment the `export_report` function computes `unit_locations` if they're not already computed (line 90 in `report.py`). So the first plot also always appears.

So either:
- remove the conditional on the unit locations plot
- remove the automatic computation

Opinions @samuelgarcia @zm711 ?
I'd vote to remove the automatic computation, since I don't think we should be computing anything in a plotting function.

chrishalcrow: Sounds ok to me, Zach! I've updated to remove the conditional on the unit locations plot.

src/spikeinterface/widgets/unit_summary.py
chrishalcrow: I know that `c++` isn't in python, and this somehow merged in my mind with the `+=` operator. Omg I've been coding in python for 9 months now... thanks Joe!

chrishalcrow:
All updated - also fixed a column counting issue that occurred when correlograms and spike_amplitudes were on, but waveforms were off. Also added in a check for `template_similarity`, which is needed to compute correlograms it seems.

zm711:
> template_similarity, which is needed to compute correlograms it seems.

yeah the correlograms are only calculated for templates deemed similar enough. That's annoyed me in the past, but I understand why it is there.

chrishalcrow:
Hello, could this be merged please(@samuelgarcia )? Just to sum up, here is every possible report output, made from

```
rec, sort = si.generate_ground_truth_recording(durations=[60])
```
then...

`sa.compute(["random_spikes", "unit_locations","templates","template_similarity"])`


`sa.compute(["random_spikes", "unit_locations", "templates","template_similarity", "correlograms"])`


`sa.compute(["random_spikes", "unit_locations", "templates","template_similarity", "waveforms"])`


`sa.compute(["random_spikes", "unit_locations", "templates","template_similarity", "spike_amplitudes"])`


`sa.compute(["random_spikes", "unit_locations", "templates", "spike_amplitudes", "correlograms"])`
(note: no template similarity => no correlograms!)


`sa.compute(["random_spikes", "unit_locations","templates","template_similarity", "waveforms", "correlograms"])`


`sa.compute(["random_spikes", "unit_locations","templates","template_similarity", "waveforms", "spike_amplitudes"])`


`sa.compute(["random_spikes", "unit_locations","templates","template_similarity", "waveforms", "correlograms", "spike_amplitudes"])`

alejoe91:
@chrishalcrow @zm711 I pushed an update that removes the need to compute `template_similarity`, which is counter-intuituve. In fact, this is only needed for cross-correlograms, but not for auto-correlogram plots.

Tested on my side and works like a charm!

alejoe91 (approved):

chrishalcrow:
> @chrishalcrow @zm711 I pushed an update that removes the need to compute `template_similarity`, which is counter-intuituve. In fact, this is only needed for cross-correlograms, but not for auto-correlogram plots.
> 
> Tested on my side and works like a charm!

Beautiful - works on my side too.

Issue #3494, opened 2024-10-21
Template loading issue and extensions documentation.


OlivierPeron:
Hi everyone,

I was recently experimenting with the analyzer extensions and found it challenging to locate the parameters for each extension. Perhaps I missed it, but I couldn't find any dedicated page or documentation listing this information. While the `get_default_extension_params` function from the `SortingAnalyzer` does provide default parameters, it doesn't offer the full range of available options.

I'm primarily raising this issue not because of that, but rather due to some problems I encountered while using templates. Similar issues might occur with other extensions, but I haven't tested those yet.

Specifically, I used 'median' as the operator to compute my templates from the waveforms, and I saved the template extension in my extension folder. The problem arises when the function `get_dense_templates_array` is called, as it assumes that templates were saved using the 'average' operator. Here's the relevant code:

```python
if isinstance(one_object, Templates):
    templates_array = one_object.get_dense_templates()
elif isinstance(one_object, SortingAnalyzer):
    if return_scaled != one_object.return_scaled:
        raise ValueError(
            f"get_dense_templates_array: return_scaled={return_scaled} is not compatible; SortingAnalyzer has the opposite setting."
        )
    ext = one_object.get_extension("templates")
    if ext is not None:
        templates_array = ext.data["average"]
    else:
        raise ValueError("SortingAnalyzer requires the 'templates' extension to retrieve templates")
else:
    raise ValueError("Input must be either Templates or SortingAnalyzer")
```

A possible solution could be to check for the computed operator such as :

```
if ext is not None:
      operator = ext.params['operators'][0]
      templates_array = ext.data[operator]
```

Thanks again for all the work !

PERON Olivier

alejoe91:
Hi @OlivierPeron 

I think it woould be safest to do something like this:
```
    templates_array = ext.data.get("average") or ext.data.get("median")
    assert templates_array is not None, "Average or median templates have not been computed."
```

Do you feel like trying to make a pull request?

zm711:
since a PR related to this has been merged and this appears to be solved I'll close this one :) Let us know if something else comes up!

PR #3495, opened 2024-10-21
Template loading average or median

OlivierPeron:
Loading templates whatever the operator in the `get_dense_templates_array` function. 
Following #3494.

### DIFF
--- a/src/spikeinterface/core/template_tools.py
+++ b/src/spikeinterface/core/template_tools.py
@@ -31,8 +31,12 @@ def get_dense_templates_array(one_object: Templates | SortingAnalyzer, return_sc
             )
         ext = one_object.get_extension("templates")
         if ext is not None:
-            templates_array = ext.data.get("average") or ext.data.get("median")
-            assert templates_array is not None, "Average or median templates have not been computed."
+            if "average" in ext.data:
+                templates_array = ext.data.get("average")
+            elif "median" in ext.data:
+                templates_array = ext.data.get("median")
+            else:
+                raise ValueError("Average or median templates have not been computed.")
         else:
             raise ValueError("SortingAnalyzer need extension 'templates' to be computed to retrieve templates")
     else:
--- a/src/spikeinterface/core/template_tools.py
+++ b/src/spikeinterface/core/template_tools.py
@@ -31,7 +31,8 @@ def get_dense_templates_array(one_object: Templates | SortingAnalyzer, return_sc
             )
         ext = one_object.get_extension("templates")
         if ext is not None:
-            templates_array = ext.data["average"]
+            templates_array = ext.data.get("average") or ext.data.get("median")
+            assert templates_array is not None, "Average or median templates have not been computed."
         else:
             raise ValueError("SortingAnalyzer need extension 'templates' to be computed to retrieve templates")
     else:


alejoe91 (approved):

src/spikeinterface/core/template_tools.py
```diff
@@ -31,7 +31,8 @@ def get_dense_templates_array(one_object: Templates | SortingAnalyzer, return_sc
             )
         ext = one_object.get_extension("templates")
         if ext is not None:
-            templates_array = ext.data["average"]
+            templates_array = ext.data.get("average") or ext.data.get("median")
+            assert templates_array is not None, "Average or median templates have not been computed."
```
alejoe91, line 35:
```suggestion
            if "average" in ext.data:
                templates_array = ext.data.get("average")
            elif "median" in ext.data:
                templates_array = ext.data.get("median")
            else:
                raise ValueError("Average or median templates have not been computed.")
```

@OlivierPeron let's refactor this way since the current implementation is throwing an error

alejoe91:
LGTM, @samuelgarcia @zm711 wanna have a look?

src/spikeinterface/core/template_tools.py
```diff
@@ -31,7 +31,12 @@ def get_dense_templates_array(one_object: Templates | SortingAnalyzer, return_sc
             )
         ext = one_object.get_extension("templates")
         if ext is not None:
-            templates_array = ext.data["average"]
+            if "average" in ext.data:
+                templates_array = ext.data.get("average")
+            elif "median" in ext.data:
+                templates_array = ext.data.get("median")
+            else:
```
zm711, line 38: One question:
Are percentiles where this else is triggered? 

One small suggestion:
my only suggestion which is very light would be call it `template_ext` instead of `ext` to be more explicit, but this is less crucial.

OlivierPeron:
I found something similar in the `ComputeTemplate` class of the analyzer_extension_core module. I share it here as a comment because I don't know how to add it to this PR.
```python
    def _get_data(self, operator="average", percentile=None, outputs="numpy"):
        if operator != "percentile":
            key = operator
        else:
            assert percentile is not None, "You must provide percentile=..."
            key = f"percentile_{percentile}"

        templates_array = self.data[key]

        if outputs == "numpy":
            return templates_array
        elif outputs == "Templates":
            return Templates(
                templates_array=templates_array,
                sampling_frequency=self.sorting_analyzer.sampling_frequency,
                nbefore=self.nbefore,
                channel_ids=self.sorting_analyzer.channel_ids,
                unit_ids=self.sorting_analyzer.unit_ids,
                probe=self.sorting_analyzer.get_probe(),
            )
        else:
            raise ValueError("outputs must be numpy or Templates")
```

Here again, operator is initialized as "average". Although it is possible to initialized differently, It could also automatically check for the operator of the saved extension. It returns a error when having used 'median' operator and only calling such as `templates = tmp.get_data()`, instead of `templates = tmp.get_data('median')` which require more knowledge of the process and I think the former is more intuitive and broadly used.

zm711:
> templates = tmp.get_data('median') which require more knowledge of the process and I think the former is more intuitive and broadly used.

I think that is an interesting point. Our `get_data` function likely needs more documentation in general. I agree with that. What is the error you actually get there? I think the concern is what if you do `average` and `median` how do we decide which to do? Whereas forcing you to request the "non-standard" makes sure you understand what you are grabbing. I could image our error being better like: 

```python
ValueError: You have selected the averages, but the options are : ['median', 'percentile']
```

OlivierPeron:
If I run `templates.get_data()` and that I computed the templates with the median operator, I get a `KeyError: 'average'`, from this line `templates_array = self.data[key]`

zm711:
Yeah, for me I think a better error is easier to fit into this scheme. How would you envision autodetect in the case you have multiple keys?

OlivierPeron:
I don't know about this. Either, I would returns a dict with all possible key. For templates is either 'average' or 'median' and will therefore be such as :
```python
templates = {}
operators = self.params['operators']
for operator in operators:
# here check for normal operator or percentile, I didn't dig into this option so I won't try handling it
  templates[operator] = self.data[operator]
```
Either improve documentation, which is a need anyway as mentioned in #3494, regarding the extensions inputs parameters.

zm711:
I would argue we merge your current PR and then you either open a separate PR with your proposal for a change or I can open a PR with an improved error message. If you don't want to work on the PR without more debate then I would open a separate issue about this other issue so the whole team can comment on this :)

OlivierPeron:
Yes let's merge this PR. Unfortunately I won't have time to work extensively on this so I think is better if you handle it on your side !
Thanks !

src/spikeinterface/core/template_tools.py
samuelgarcia: I think I would have done the reverse.
In case we have both median and average which one do we want ? median maybe no ?

zm711: I think the better case is to force the user to make the choice, but it comes back to ease of use vs maximal control. And I think this project has become big enough that more of the users want ease of use (ie intelligent defaults) vs needing to control everything. 

I'm curious what's your view on median vs average for this? I think average is used more commonly no? Although median is cool especially if you have some outlier artifact "spikes" I guess median would be better.

PR #3497, opened 2024-10-22
Fix dtype of quality metrics before and after merging

zm711:
### MRE

```python

import pandas as pd
df = pd.DataFrame({'test': [1,2,3]})
new_df = pd.DataFrame(index=df.index, columns=df.columns)

df.test.dtype
Out[7]: dtype('int64')

new_df.test.dtype
Out[8]: dtype('O')
```

Basically when you copy a dataframe from a previous dataframe columns it forces the dtype to be object instead of numeric. 

### Easy Solution

Using the `to_numeric` will bring us back to numeric values.

### Caveats

This switches the dtype from the Pandas `Float64` to the numpy `float64`. I don't think this is too bad, since doing the queries should still be fine no? 

### Testing
i added a small test to test merging, but let me know if we'd prefer not to have it.


src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -125,7 +125,13 @@ def _merge_extension_data(
         all_unit_ids = new_sorting_analyzer.unit_ids
         not_new_ids = all_unit_ids[~np.isin(all_unit_ids, new_unit_ids)]
 
+        # this creates a new metrics dictionary, but the dtype for everything will be
+        # object
         metrics = pd.DataFrame(index=all_unit_ids, columns=old_metrics.columns)
+        # we can iterate through the columns and convert them back to numbers with
+        # pandas.to_numeric. coerce allows us to keep the nan values.
+        for column in metrics.columns:
+            metrics[column] = pd.to_numeric(metrics[column], errors="coerce")
```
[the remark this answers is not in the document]
zm711: Maybe. I agree Pandas is making their own dtypes like NADType which doesn't play nicely with numpy in my scipts I tend to just query based on numpy stuff). So I don't know for sure.  I could test that later. Although for me I would prefer to coerce everything to numpy types since that's what I'm used to. None of my tables are big enough that I worry about dtype inefficiency stuff that Pandas has been working on with the new backend.

JoeZiminski (approved):
Hey @zm711 this looks great! super useful test too. Weird behaviour from pandas! Some minor comments:

- I think `new_df = pd.DataFrame(index=df.index, columns=df.columns, dtype=np.float64)` will have the same effect. You loose the coerce on error behaviour but assuming the data is always going to be filled with `NaN` this shouldn't be a problem. However it is more implicit and provides less information on the weird pandas behaviour than the loop approach.

- The results of this operation mean all columns are `np.float64` but in the original `metrics` as returned from `compute_quality_metrics` some columns are `Int64Dtype`. This seems to be dynamic based on contents (e.g. in the test run presence ratio were all `1` and it's dtype is `Int64Dtype`but presumably it would be a float under most circumstances. `num_spikes` I guess will always be `int`. The only time I can imagine this being a problem is if some equality check is performed e.g.`num_spikes == 1` which might work for the original `compute_quality_metrics` output but fail after merging as data will be float. So maybe it is simplest just to cast `num_spikes` -> `Int64Dtype` and leave the rest as float?

zm711:
Thanks so much @JoeZiminski!

>  think new_df = pd.DataFrame(index=df.index, columns=df.columns, dtype=np.float64) 

I'm no Pandas expert so I'm happy to have changes here if they are better! I just don't know have an intuition for what is the smartest strategy so if you know Pandas really well then I'll make the change :)

> e.g. in the test run presence ratio were all 1 and it's dtype is Int64Dtypebut presumably it would be a float under most circumstances.

True. This is our mistake for letting Pandas infer. presence ratio is a float between 0 and 1. But if they are all 0 or all 1 it casts to int for memory purposes. Users should never assume it is an int although it could be in extreme cases. It would be better for us to explicitly make it a float and take the memory hit in my opinion.

> num_spikes I guess will always be int.

This is true and when I scanned the table I forgot about this one. It would be better to make that one a float. I don't think a user should ever do `num_spikes == x`. I could imagine a query of `num_spikes>=x`. I think (BUT correct me if I'm wrong) that for floats only the decimal is unstable such that a `float(1) >= int(1)`. In this case testing against a min number of spikes should not be a problem. If I'm wrong then I think we have to cast that one back to int in a separate step.

> super useful test too
<3 Thanks. I figure we really need to protect ourselves from some of these small regressions. So I'm trying :)

zm711:
@alejoe91, do you have any opinions of implementing this? Happy to change to a different method if you prefer something. I think the only thing we are failing to maintain is `num_spikes`.

alejoe91 (approved):

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
alejoe91, line 134:
```suggestion
            metrics[column] = pd.to_numeric(metrics[column], errors="coerce")
            if np.all(np.mod(metrics[column], 1) == 0):
                metrics[column] = metrics[column].astype(int)
```
@zm711 @JoeZiminski what about this to restore integers?

src/spikeinterface/qualitymetrics/tests/test_quality_metric_calculator.py
```diff
@@ -48,6 +48,34 @@ def test_compute_quality_metrics(sorting_analyzer_simple):
...
+
+    # we should copy over the metrics after merge
+    for column in metrics.columns:
+        assert column in new_metrics.columns
+
+    # 10 units vs 9 units
+    assert len(metrics.index) > len(new_metrics.index)
+
+    # dtype should be fine after merge but is cast from Float64->float64
+    assert np.float64 == new_metrics["snr"].dtype
```
alejoe91, line 76: we can add a test on int coercion if we end up using the suggestion here: https://github.com/SpikeInterface/spikeinterface/pull/3497/files#r1827487180

zm711:
So the problem is that Pandas will infer the dtype and sometimes this is actually wrong. Like the presence ratio above which should technically always be a float between 0.0 and 1.0, but if it is all 1s and 0s will be stored as an int. Then if we merge and get a fraction then the dtype is wrong. I think it might be better to hard code the int64 for `num_spikes` since everything else is a float.

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
zm711: This actually doesn't work because at this point everything is a nan until data is loaded.

zm711:
Okay so changes in this PR

1) now we ensure all columns have the dtype specified in our name_to_dtype dict (pandas is not allowed to infer)
2) test for empty units was changed because we no longer put nans in num_spikes (because that should be 0 for empty units)--let me know if we should discuss
3) merging now recasts the columns to the original dtypes
4) test added for merging

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -229,10 +238,25 @@ def _compute_metrics(self, sorting_analyzer, unit_ids=None, verbose=False, metri
...
         metrics = metrics.convert_dtypes()
+
+        # we do this because the convert_dtypes infers the wrong types sometimes.
+        # the actual types for columns can be found in column_name_to_column_dtype dictionary.
+        for column in metrics.columns:
+            # we have one issue where the name of the columns for synchrony are named based on
+            # what the user has input as arguments so we need a way to handle this separately
+            # everything else should be handled with the column name.
+            if "sync" in column:
+                metrics[column] = metrics[column].astype(column_name_to_column_dtype["sync"])
```
zm711, line 255: I would argue we keep this for backward compatibility no?  I could add a comment saying we can simplify this in a couple versions.

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -16,7 +16,8 @@
     compute_pc_metrics,
     _misc_metric_name_to_func,
     _possible_pc_metric_names,
-    qm_compute_name_to_column_names,
+    compute_name_to_column_names,
```
zm711, line 19:
```suggestion
```

zm711:
Never mind. After Chris's updates I thought this would go in cleanly other than one conflict. His changes actually require me to make additional updates to this PR...

zm711:
Now we are finally ready :)

src/spikeinterface/qualitymetrics/quality_metric_list.py
```diff
@@ -66,7 +66,11 @@
     "amplitude_cutoff": ["amplitude_cutoff"],
     "amplitude_median": ["amplitude_median"],
     "amplitude_cv": ["amplitude_cv_median", "amplitude_cv_range"],
-    "synchrony": ["sync_spike_2", "sync_spike_4", "sync_spike_8"],
+    "synchrony": [
+        "sync_spike_2",
+        "sync_spike_4",
+        "sync_spike_8",
+    ],  # we probably shouldn't hard code this. This is determined by the arguments in the function...
```
alejoe91, line 73: but I think we agreed that we can just hard-code this at the QM level, so it should be ok.

Let's keep the comment until this is actually hard-coded!

src/spikeinterface/qualitymetrics/quality_metric_list.py
```diff
@@ -79,3 +83,36 @@
...
+    "isi_violations_ratio": float,
+    "isi_violations_count": float,
+    "rp_violations": float,
+    "rp_contamination": float,
+    "sliding_rp_violation": float,
+    "amplitude_cutoff": float,
+    "amplitude_median": float,
+    "amplitude_cv_median": float,
+    "amplitude_cv_range": float,
+    "sync": float,
```
alejoe91, line 102: then this could become `sync_spike_2`, `sync_spike_4`, `sync_spike_8` as well

src/spikeinterface/qualitymetrics/quality_metric_list.py
alejoe91: Actually, this was done already: https://github.com/SpikeInterface/spikeinterface/pull/3559

src/spikeinterface/qualitymetrics/quality_metric_list.py
alejoe91, line 102:
```suggestion
    "sync_spike_2": float,
    "sync_spike_4": float,
    "sync_spike_8": float,
```

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -229,10 +237,25 @@ def _compute_metrics(self, sorting_analyzer, unit_ids=None, verbose=False, metri
...
+        # we do this because the convert_dtypes infers the wrong types sometimes.
+        # the actual types for columns can be found in column_name_to_column_dtype dictionary.
+        for column in metrics.columns:
+            # we have one issue where the name of the columns for synchrony are named based on
+            # what the user has input as arguments so we need a way to handle this separately
+            # everything else should be handled with the column name.
+            if "sync" in column:
+                metrics[column] = metrics[column].astype(column_name_to_column_dtype["sync"])
+            else:
+                metrics[column] = metrics[column].astype(column_name_to_column_dtype[column])
```
alejoe91, line 257:
```suggestion
            metrics[column] = metrics[column].astype(column_name_to_column_dtype[column])
```
https://github.com/SpikeInterface/spikeinterface/pull/3559

src/spikeinterface/qualitymetrics/quality_metric_list.py
zm711: We can't do this because in the old days people could make this name anything they wanted. So technically some could've used si 101 and had a sync_spike_10 and we would break here. Right?

src/spikeinterface/qualitymetrics/quality_metric_calculator.py
```diff
@@ -229,10 +237,22 @@ def _compute_metrics(self, sorting_analyzer, unit_ids=None, verbose=False, metri
...
         # we use the convert_dtypes to convert the columns to the most appropriate dtype and avoid object columns
         # (in case of NaN values)
         metrics = metrics.convert_dtypes()
+
+        # we do this because the convert_dtypes infers the wrong types sometimes.
+        # the actual types for columns can be found in column_name_to_column_dtype dictionary.
+        for column in metrics.columns:
+            # we have one issue where the name of the columns for synchrony are named based on
+            # what the user has input as arguments so we need a way to handle this separately
+            # everything else should be handled with the column name.
```
alejoe91, line 253:
```suggestion
```
we don't need this comment anymore

alejoe91:
@zm711 since we hard-coded the sync sizes in #3559 I simplified the PR, so we don't have to deal with them differently.

For back compatibility, I just added a check that casts the column values only if the column name is in `column_name_to_column_dtype`. So for example, if one computed a `sync_spike_32`, it won't be cast but would still work.

zm711:
Okay cool. Works for me :)

PR #3501, opened 2024-10-23
Add error messaging around use of get data in templates

zm711:
This is a light changing of the error messaging in the core extension file along with some docstring touch ups for numpydoc.

Related to #3495

### DIFF
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -23,7 +23,7 @@ from .sorting_tools import random_spikes_selection
 class ComputeRandomSpikes(AnalyzerExtension):
     """
     AnalyzerExtension that select somes random spikes.
-    This is allows for a subsampling of spikes for further calculations and is important
+    This allows for a subsampling of spikes for further calculations and is important
     for managing that amount of memory and speed of computation in the analyzer.
 
     This will be used by the `waveforms`/`templates` extensions.
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -28,7 +28,7 @@ class ComputeRandomSpikes(AnalyzerExtension):
 
     This will be used by the `waveforms`/`templates` extensions.
 
-    This internally use `random_spikes_selection()` parameters.
+    This internally uses `random_spikes_selection()` parameters.
 
     Parameters
     ----------
@@ -106,7 +106,7 @@ class ComputeRandomSpikes(AnalyzerExtension):
         return self._some_spikes
 
     def get_selected_indices_in_spike_train(self, unit_id, segment_index):
-        # useful for Waveforms extractor backwars compatibility
+        # useful for WaveformExtractor backwards compatibility
         # In Waveforms extractor "selected_spikes" was a dict (key: unit_id) of list (segment_index) of indices of spikes in spiketrain
         sorting = self.sorting_analyzer.sorting
         random_spikes_indices = self.data["random_spikes_indices"]
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -380,7 +380,12 @@ class ComputeTemplates(AnalyzerExtension):
         assert isinstance(operators, list)
         for operator in operators:
             if isinstance(operator, str):
-                assert operator in ("average", "std", "median", "mad")
+                if operator not in ("average", "std", "median", "mad"):
+                    error_msg = (
+                        f"You have entered an operator {operator} in your `operators` argument which is "
+                        f"not supported. Please use any of ['average', 'std', 'median', 'mad'] instead."
+                    )
+                    raise ValueError(error_msg)
             else:
                 assert isinstance(operator, (list, tuple))
                 assert len(operator) == 2
@@ -549,8 +554,16 @@ class ComputeTemplates(AnalyzerExtension):
         if operator != "percentile":
             key = operator
         else:
-            assert percentile is not None, "You must provide percentile=..."
+            assert percentile is not None, "You must provide percentile=... if `operator=percentile`"
             key = f"percentile_{percentile}"
+
+        if key not in self.data.keys():
+            error_msg = (
+                f"You have entered `operator={key}`, but the only operators calculated are "
+                f"{list(self.data.keys())}. Please use one of these as your `operator` in the "
+                f"`get_data` function."
+            )
+            raise ValueError(error_msg)
 
         templates_array = self.data[key]
 
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -22,21 +22,23 @@ from .sorting_tools import random_spikes_selection
 
 class ComputeRandomSpikes(AnalyzerExtension):
     """
-    AnalyzerExtension that select some random spikes.
+    AnalyzerExtension that select somes random spikes.
+    This is allows for a subsampling of spikes for further calculations and is important
+    for managing that amount of memory and speed of computation in the analyzer.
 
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -22,21 +22,23 @@
 
 class ComputeRandomSpikes(AnalyzerExtension):
     """
-    AnalyzerExtension that select some random spikes.
+    AnalyzerExtension that select somes random spikes.
+    This is allows for a subsampling of spikes for further calculations and is important
```
zm711, line 26:
```suggestion
    This allows for a subsampling of spikes for further calculations and is important
```

zm711 (commented):
Fix my own typo haha.

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -22,21 +22,23 @@
 
 class ComputeRandomSpikes(AnalyzerExtension):
     """
-    AnalyzerExtension that select some random spikes.
+    AnalyzerExtension that select somes random spikes.
+    This allows for a subsampling of spikes for further calculations and is important
```
JoeZiminski, line 26: This is very nice addition, could 'for further calculations' and the below ' This will be used by the `waveforms`/`templates` extensions.'  be combined and emphasise that this choice could potentially have important consequences for results. e.g. 'The samples spikes will be used for calculating waveforms and templates and as such determine many downstream parameters (e.g. quality metrics). Therefore it is important that spikes a sufficient number of spikes are sampled and that these are distributed evenly through the dataset'.

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -22,21 +22,23 @@
...
+    for managing that amount of memory and speed of computation in the analyzer.
 
     This will be used by the `waveforms`/`templates` extensions.
 
-    This internally use `random_spikes_selection()` parameters are the same.
+    This internally use `random_spikes_selection()` parameters.
```
JoeZiminski, line 31:
```suggestion
    This internally uses `random_spikes_selection()` parameters.
```

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -380,7 +382,12 @@ def _set_params(self, ms_before: float = 1.0, ms_after: float = 2.0, operators=N
         assert isinstance(operators, list)
         for operator in operators:
             if isinstance(operator, str):
-                assert operator in ("average", "std", "median", "mad")
+                if operator not in ("average", "std", "median", "mad"):
```
JoeZiminski, line 385: I was thinking about this earlier, do you think it is worth adding the type  : `"average" | "std" | "median" | "mad" | None` to the `_set_params ` signature here? I do not think this is documented anywhere except for inside the function body

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -380,7 +382,12 @@ def _set_params(self, ms_before: float = 1.0, ms_after: float = 2.0, operators=N
...
+                    error_msg = (
+                        f"You have entered an operator {operator} in your `operators` argument which is "
+                        f"not supported. Please use any of ['average', 'std', 'median', 'mad'] instead."
+                    )
```
JoeZiminski, line 389: Oh nice, in that case discard above comment 😆

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -405,9 +412,13 @@ def _run(self, verbose=False, **job_kwargs):
...
         else:
-            for operator in self.params["operators"]:
-                if operator not in ("average", "std"):
-                    raise ValueError(f"Computing templates with operators {operator} needs the 'waveforms' extension")
+            bad_operator_list = [
+                operator for operator in self.params["operators"] if operator not in ("average", "std")
+            ]
+            if len(bad_operator_list) > 0:
+                raise ValueError(
+                    f"Computing templates with operators {bad_operator_list} requires the 'waveforms' extension"
```
JoeZiminski, line 420: Interesting I assumed waveforms always needed to be computed for templates, what other ways ways of doing it are there?

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -405,9 +412,13 @@ def _run(self, verbose=False, **job_kwargs):
             self._compute_and_append_from_waveforms(self.params["operators"])
 
         else:
-            for operator in self.params["operators"]:
-                if operator not in ("average", "std"):
-                    raise ValueError(f"Computing templates with operators {operator} needs the 'waveforms' extension")
+            bad_operator_list = [
+                operator for operator in self.params["operators"] if operator not in ("average", "std")
+            ]
+            if len(bad_operator_list) > 0:
```
JoeZiminski, line 418: maybe if `any(bad_operator_list)` is slightly more readable (?)

JoeZiminski (dismissed):
Nice! am currently updating some old teaching code to sorting analyzer so appreciating the utility of such additions :pray:

src/spikeinterface/core/analyzer_extension_core.py
zm711: I think that is stylistic. I don't use `any` that often in code I think checking the length of the list is just my default. I don't know what others think?

src/spikeinterface/core/analyzer_extension_core.py
zm711: You should dig into the templates code. Alessio + Sam ( I think it was just them but maybe Heberto too) developed an accumulator method that uses the waveforms to make an average without saving them so you would

analyzer.compute(['random_spikes', 'templates']) and it will read the waveforms from random spikes while making the templates then discard them to save on memory. I think the only think it breaks would be doing PCA later, but it is way less storage intensive since you don't save all the extra waveforms. It is limited in the types of operators it can do though as you can see from the error.

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -22,21 +22,23 @@
 
 class ComputeRandomSpikes(AnalyzerExtension):
     """
-    AnalyzerExtension that select some random spikes.
+    AnalyzerExtension that select somes random spikes.
+    This allows for a subsampling of spikes for further calculations and is important
+    for managing that amount of memory and speed of computation in the analyzer.
```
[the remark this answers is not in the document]
zm711: I'm torn. I was just trying to clarify. I prefer 

one line short summary
params
return
examples
notes

And do that big explanation in the notes. The statement you're making I would put into notes, but I didn't want to change stuff too much although since I made the diff it's my fault you looked. 

The real point of this PR is just to improve error messaging. I think we could improve further the docstrings etc in a separate PR (and this is a reminder for me to focus on my task so we don't get bogged down on other stuff for a PR with a specific purpose :) ).

src/spikeinterface/core/analyzer_extension_core.py
samuelgarcia: yep this accumulator was a great stuf during the analyzer dev.
the only drawback is that MAD cannot be computed that way. but saving ram and disk space is cool.
you can see the idea in waveform_tools.py
every worker accumulate snippet in parralel and the sum + divide is done at teh end.
this is quite fast

samuelgarcia (approved):

samuelgarcia:
ok for me.
@alejoe91 you want to have a look ?

src/spikeinterface/core/analyzer_extension_core.py
alejoe91: I think it's clear enough :)

src/spikeinterface/core/analyzer_extension_core.py
JoeZiminski: cool! will check that out

alejoe91 (approved):

PR #3503, opened 2024-10-24
Installation tips using uv

samuelgarcia:


installation_tips/README.md
```diff
@@ -1,36 +1,53 @@
 ## Installation tips
 
-If you are not (yet) an expert in Python installations (conda vs pip, mananging environements, etc.),
-here we propose a simple recipe to install `spikeinterface` and several sorters inside an anaconda
-environment for windows/mac users.
+If you are not (yet) an expert in Python installations, a main difficulty is choosing the installation procedure.
+The main ideas you need to know before starting:
+ * python itself can be distributed and installed many many ways.
+ * python itself do not contain so many features for scientifique computing you need to install "packages".
```
zm711, line 6:
```suggestion
 * python itself does not contain so many features for scientific computing you need to install "packages".
```

installation_tips/README.md
```diff
@@ -1,36 +1,53 @@
...
+   Numpy, scipy, matplotlib, spikeinterface, ... are python packages that have a complicated dependency graph between then. "uv"
```
zm711, line 7:
```suggestion
   numpy, scipy, matplotlib, spikeinterface, ... are python packages that have a complicated dependency graph between then. "uv"
```

installation_tips/README.md
```diff
@@ -1,36 +1,53 @@
...
+ * installing package  can be distributed and installed several ways (pip, conda, uv, mamba, ...)
+ * installing many packages at once is challenging (because of the depenency graph) so you need to do it in an "isolated environement"
```
zm711, line 9:
```suggestion
 * installing many packages at once is challenging (because of their dependency graphs) so you need to do it in an "isolated environement"
```

installation_tips/README.md
```diff
@@ -1,36 +1,53 @@
...
+   to not destroy any previous installation. You need to see an "environement" as a sub installtion in dedicated folder.
```
zm711, line 10:
```suggestion
   to not destroy any previous installation. You need to see an "environment" as a sub installation in a dedicated folder.
```

installation_tips/README.md
```diff
@@ -1,36 +1,53 @@
...
+
+Choosing the installator + a environement manager + a package installer is a nightmare for beginners.
```
zm711, line 12:
```suggestion
Choosing the installer + an environment manager + a package installer is a nightmare for beginners.
```

installation_tips/README.md
```diff
@@ -1,36 +1,53 @@
...
+The main options are:
+  * use "anaconda" that do everything. The most popular but bad idea because : ultra slow and agressive licensing (not free anymore)
```
zm711, line 14:
```suggestion
  * use "anaconda", which does everything. The most popular but bad idea because : ultra slow and aggressive licensing (not always free anymore)
```

installation_tips/README.md
```diff
@@ -1,36 +1,53 @@
...
+  * use python from the system or python.org + venv + pip : good idea for linux users.
+  * use "uv" : a new, fast and simple. We recommand this for beginners on evry os.
```
zm711, line 16:
```suggestion
  * use "uv" : a new, fast and simple package manager. We recommend this for beginners on every os.
```

installation_tips/README.md
```diff
@@ -1,36 +1,53 @@
...
+
+Here we propose a steps by step recipe for beginers based on "uv".
```
zm711, line 18:
```suggestion
Here we propose a step by step recipe for beginers based on "uv".
```

installation_tips/README.md
```diff
@@ -1,36 +1,53 @@
...
+We used to propose here a solution based on anaconda. It is kept here for a while but we do not recommand it anymore.
```
zm711, line 19:
```suggestion
We used to recommend installing with anaconda. It will be kept here for a while but we do not recommend it anymore.
```

installation_tips/README.md
```diff
@@ -1,36 +1,53 @@
...
+ * kilosort4
 
 Kilosort, Ironclust and HDSort are MATLAB based and need to be installed from source.
 
 ### Quick installation
 
-Steps:
+1. On macOS and Linux. Open a terminal and do
+   `$ curl -LsSf https://astral.sh/uv/install.sh | sh`
+1. On windows. Open a powershell and do
```
zm711, line 33: Hey Sam we recommend users use Command Prompt since our shell scripts for sorters only work in command prompt. It might be better if they have a command prompt version of instructions.

installation_tips/README.md
```diff
@@ -1,36 +1,53 @@
...
 
+More detail on [uv here](https://github.com/astral-sh/uv).
 
-Done! Before running a spikeinterface script you will need to "select" this "environment" with `conda activate si_env`.
+## Installing before release
 
-Note for **linux** users : this conda recipe should work but we recommend strongly to use **pip + virtualenv**.
+Some tools in the spikeinteface ecosystem are getting regular bug fixes (spikeinterface, spikeinterface-gui, probeinterface, python-neo, sortingview).
+We are making releases 2 to 4 times a year. In between releases if you want to install from source you can use the `requirements_rolling.txt` file to create the environment. This will install the packages of the ecosystem from source.
+This is a good way to test if patch fix your issue.
```
zm711, line 50:
```suggestion
This is a good way to test if a patch fixes your issue.
```

installation_tips/README.md
```diff
@@ -62,6 +79,28 @@ This script tests the following:
   * opening the spikeinterface-gui
   * exporting to Phy
 
+### Legacy installation using anaconda (not recomanded)
```
zm711, line 82:
```suggestion
### Legacy installation using anaconda (not recommended)
```

zm711 (commented):
just edits and one comment before we go live with this.

installation_tips/README.md
JoeZiminski, line 14: Maybe it is better to lead with the recommended one (i.e. uv first bullet point)?

On conda, is true in some cases but newer versions use `mamba` backend which is quite fast, and `conda-forge` does not have licensing issues. A problem is that HPC often come with the older slow versions and the default channel is anaconda's. 

No point going into too much detail but a suggestion to amend to say that that the default anaconda channel has agressive licensing and that it can be slow / is slower than uv. It is probably worth mentioning it can handle non-python dependencies while the other two can't, but this is not required for spikeinterface.

JoeZiminski (changes requested):
Hey @samuelgarcia this is very nice, I actually havn't tried `uv` yet so will run the commands to test on windows later. I request changes for some minor typos but feel free to ignore the other suggestions. BTW the creator of `uv` also makes another tool called `Ruff` 😏 maybe we should try it!

installation_tips/README.md
zm711: Agreed. Wasn't sure what he meant there. haha.

installation_tips/README.md
samuelgarcia: I will try to modify a bit

installation_tips/README.md
samuelgarcia: could you test if the windows command also works in CMD ?
uv docs recommand powershell (I tested and it works)

samuelgarcia:
thanks a lot to @zm711  and @JoeZiminski.
At the moment the idea is to test th recipe.
We can reformulate later.

installation_tips/README.md
```diff
@@ -1,36 +1,53 @@
...
- * phy
- * tridesclous
+ * kilosort4
 
 Kilosort, Ironclust and HDSort are MATLAB based and need to be installed from source.
 
 ### Quick installation
 
-Steps:
+1. On macOS and Linux. Open a terminal and do
```
h-mayorquin, line 31: Two comments:
1) I think it would be better to refer them to the uv installation guide. If it changes, the our docs will be outdated for no reason and they have os dependent paths:

https://docs.astral.sh/uv/getting-started/installation/

2) Why are we creating new requirements? for the stabel if uv installs from pip it should suffice, for the rolling, uv can install directly from github and that would cover that. 

In both cases it would use the pyproject.toml.

zm711: I think this is a good point for these external tools!

samuelgarcia: I think I do not want to overload the pyproject
The idea to have a working env for a beginner for a tutorial that contains spikeinterface but also external tools.
sigui sortingview ks.
The idea of having a entire and unqiue recipe and not refering to uv is to avoid a a benginer to understand evreything and compiling informations from several places.

samuelgarcia:
>  BTW the creator of `uv` also makes another tool called `Ruff` 😏 maybe we should try it!

I know but I do not plan to use a linter very soon... I figth very hard to disable then in editors.

installation_tips/README.md
h-mayorquin: I don't understand, you don't need to add anything to the pyproject. The point of the proposal is to have less files on the repo.

samuelgarcia: Here the requirements is bug : PySide6, jupyter.
I do not see this requirements as optional dependencies of spikeinterface but more a minimal stack for a beginner.
what would be the labels in pyproject.toml for theses 2 requirements files ?

h-mayorquin: Fair enough, the right way would be to use the newly accepted [PEP 735](https://peps.python.org/pep-0735/) that covers just this but that is not available yet. 

I would suggest that it would be better for you (so you don't have to remember about dependencies in two places) if the requirements files only had the dependencies that are not in the pyprojec.toml there. Then you could do:

```bash
uv pip install spikeinterface
uv pip install `the_beginnger_required_stack.txt` 
```

But maybe you don't mind having these not working some time.

installation_tips/check_your_install.py
```diff
@@ -13,15 +16,20 @@ def check_import_si_full():
...
+    )
+    rec.save(folder='./toy_example_recording', **job_kwargs)
 
 
 def _run_one_sorter_and_analyzer(sorter_name):
-    job_kwargs = dict(n_jobs=-1, progress_bar=True, chunk_duration="1s")
     import spikeinterface.full as si
     recording = si.load_extractor('./toy_example_recording')
-    sorting = si.run_sorter(sorter_name, recording, output_folder=f'./sorter_with_{sorter_name}', verbose=False)
+    sorting = si.run_sorter(sorter_name, recording, folder=f'./sorter_with_{sorter_name}', verbose=False)
```
zm711, line 32: Should we set n_jobs here? It might be a bit annoying during the install check that users see a million warnings about "n_jobs" not set.

installation_tips/README.md
```diff
@@ -40,29 +60,48 @@ If you want to test the spikeinterface install you can:
...
-2. Open the Anaconda Command Prompt (Windows) or Terminal (Mac)
-3. If not in your "Documents" folder type `cd Documents`
-4. Run this:
-    ```
-    conda activate si_env
-    python check_your_install.py
-    ```
-5. If a windows user to clean-up you will also need to right click + save [`cleanup_for_windows.py`](https://raw.githubusercontent.com/SpikeInterface/spikeinterface/main/installation_tips/cleanup_for_windows.py)
+2. Open the CMD (Windows) or Terminal (Mac/Linux)
+3. Acticate your is_env : `source si_env/bin/activate`
```
zm711, line 64:
```suggestion
3. Activate your si_env : `source si_env/bin/activate` (Max/Linux), `si_env\Scripts\activate` (CMD prompt)
```

installation_tips/README.md
```diff
@@ -1,36 +1,56 @@
...
+1. On macOS and Linux. Open a terminal and do
+   `$ curl -LsSf https://astral.sh/uv/install.sh | sh`
+1. On windows. Open a powershell and do
+   `powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"`
+2. exit session and log again.
+3. Download with right click and save this file corresponding in "Documents" folder:
+    * [`requirements_stable.txt`](https://raw.githubusercontent.com/SpikeInterface/spikeinterface/main/installation_tips/requirements_stable.txt)
+4. open terminal or powershell
+5. `uv venv si_env --python 3.11`
+6. `source si_env/bin/activate` (you should have `(si_env)` in your terminal)
```
zm711, line 43:
```suggestion
6. For Mac/Linux `source si_env/bin/activate` (you should have `(si_env)` in your terminal) or for Powershell `si_env\Scripts\activate` 
```

zm711 (commented):
@samuelgarcia to get the env to work with the standard powershell and command prompt you can't use source + bin. So you need to fix these two commands.

installation_tips/README.md
cwindolf: conda+mamba via https://github.com/conda-forge/miniforge is my favorite, doesn't have licensing problems, and almost required when setting up gpu dependencies

installation_tips/full_spikeinterface_environment_linux_dandi.yml
```diff
@@ -5,7 +5,7 @@ channels:
 dependencies:
   - python=3.11
   - pip
-  - numpy
+  - numpy<2
```
alejoe91, line 8:
```suggestion
  - numpy
```

zm711:
@samuelgarcia do you want me to push some idiomatic fixes directly to this or would you prefer to merge and then I do a separate PR to do some final writing cleanup?

alejoe91:
@zm711 I'm doing a pass over this! and fix outdated stuff. Let me check with @samuelgarcia an push here

alejoe91:
@zm711 done on my end!

zm711:
I should have time to read later today (but can't promise). If you'd prefer to get it merged that's fine by me. I can update anything I see in a future PR.

installation_tips/README.md
```diff
@@ -1,70 +1,104 @@
...
+Done! Before running a spikeinterface script you will need to "select" this "environment" with `conda activate si_env`.
+
+Note for **Linux** users: this conda recipe should work but we recommend strongly to use **pip + virtualenv**.
+
+
 
-Some tools in the spikeinteface ecosystem are getting regular bug fixes (spikeinterface, spikeinterface-gui, probeinterface, python-neo, sortingview).
-We are making releases 2 to 4 times a year. In between releases if you want to install from source you can use the `full_spikeinterface_environment_rolling_updates.yml` file to create the environment. This will install the packages of the ecosystem from source.
-This is a good way to test if patch fix your issue.
+[^1]: Although uv installation instructions are for the Powershell. Our sorter scripts are for the CMD Prompt. Using both terminals is possible.
```
zm711: This is good by me now, unless people think this statement is now confusing ("using both terminals is possible"-- I meant it in the sense of you have to do Powershell first then use CMD Prompt, but if people think that users will interpret this as going back and forth then maybe I should fix it)?

zm711: I made this clearer. Since it is a footnote I think having the full explanation is fine.

PR #3505, opened 2024-10-25
Whitening fix - compute covariance matrix in float

JoeZiminski:
In `whiten.py` there is a datatype argument to set the datatype of the output recording. The casting is performed after whitening has been performed and does not affect the datatype used for computing the whitening matrix. Up until `0.100.8`, the covariance matrix was always computed in `float32`, but in `0.101.0` the line `random_data = np.astype(random_data, np.float32)` that was [under this line](https://github.com/SpikeInterface/spikeinterface/blob/4a1a45a67db4534150d591bd5324e271ae8af740/src/spikeinterface/preprocessing/whiten.py#L187) was removed. Now if the recording is `int16` the covariance matrix estimation is performed in `int16` and will overflow in nearly all cases.

This PR reinstates that line, for now casting to `float64`. The recording `dtype` will be cast back to the user-specified `dtype` or the input recording `dtype`, so I don't think there is any downside in using maximum precision. The only thing is, maybe we will want to move to GPU support in future, and using float32 will keep it consistent. 

Unless I'm missing something, I think it would be worth rolling this out in an update ASAP and making a note of this on the release note header. My understanding is the majority of recording datatypes are stored `int16` and so this regression may catch out quite a lot of users since `0.101.0` and is quite hard to track down. 

TODOs
- [ ] extend tests

### DIFF
--- a/src/spikeinterface/preprocessing/whiten.py
+++ b/src/spikeinterface/preprocessing/whiten.py
@@ -20,7 +20,7 @@ class WhitenRecording(BasePreprocessor):
         The recording extractor to be whitened.
     dtype : None or dtype, default: None
         Datatype of the output recording (covariance matrix estimation
-        and whitening are performed in float64.
+        and whitening are performed in float64).
         If None the the parent dtype is kept.
         For integer dtype a int_scale must be also given.
     mode : "global" | "local", default: "global"
--- a/src/spikeinterface/preprocessing/whiten.py
+++ b/src/spikeinterface/preprocessing/whiten.py
@@ -76,8 +76,9 @@ class WhitenRecording(BasePreprocessor):
         dtype_ = fix_dtype(recording, dtype)
 
         if dtype_.kind == "i":
-            assert int_scale is not None, ("For recording with dtype=int you must set the output dtype to float "
-                                           " OR set a int_scale")
+            assert int_scale is not None, (
+                "For recording with dtype=int you must set the output dtype to float " " OR set a int_scale"
+            )
 
         if W is not None:
             W = np.asarray(W)
--- a/src/spikeinterface/preprocessing/whiten.py
+++ b/src/spikeinterface/preprocessing/whiten.py
@@ -124,7 +124,7 @@ class WhitenRecordingSegment(BasePreprocessorSegment):
     def get_traces(self, start_frame, end_frame, channel_indices):
         traces = self.parent_recording_segment.get_traces(start_frame, end_frame, slice(None))
         traces_dtype = traces.dtype
-        # if uint --> force int
+        # if uint --> force float
         if traces_dtype.kind == "u":
             traces = traces.astype("float32")
 
@@ -185,6 +185,7 @@ def compute_whitening_matrix(
 
     """
     random_data = get_random_data_chunks(recording, concatenated=True, return_scaled=False, **random_chunk_kwargs)
+    random_data = random_data.astype(np.float64)
 
     regularize_kwargs = regularize_kwargs if regularize_kwargs is not None else {"method": "GraphicalLassoCV"}
 
--- a/src/spikeinterface/preprocessing/whiten.py
+++ b/src/spikeinterface/preprocessing/whiten.py
@@ -19,6 +19,8 @@ class WhitenRecording(BasePreprocessor):
     recording : RecordingExtractor
         The recording extractor to be whitened.
     dtype : None or dtype, default: None
+        Datatype of the output recording (covariance matrix estimation
+        and whitening are performed in float64.
         If None the the parent dtype is kept.
         For integer dtype a int_scale must be also given.
     mode : "global" | "local", default: "global"
@@ -74,7 +76,8 @@ class WhitenRecording(BasePreprocessor):
         dtype_ = fix_dtype(recording, dtype)
 
         if dtype_.kind == "i":
-            assert int_scale is not None, "For recording with dtype=int you must set dtype=float32 OR set a int_scale"
+            assert int_scale is not None, ("For recording with dtype=int you must set the output dtype to float "
+                                           " OR set a int_scale")
 
         if W is not None:
             W = np.asarray(W)
--- a/src/spikeinterface/preprocessing/whiten.py
+++ b/src/spikeinterface/preprocessing/whiten.py
@@ -77,7 +77,7 @@ class WhitenRecording(BasePreprocessor):
 
         if dtype_.kind == "i":
             assert int_scale is not None, (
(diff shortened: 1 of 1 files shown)


zm711:
Out of curiosity since kilosort 1-3 (and even 4 prefers) int16 how do they deal with their whitening? I have looked.

JoeZiminski:
I was curious about this also, they:
- have `int16` raw data
- scale to `float32` for CAR, filtering, whitening
- scale up by 200 and convert back to `int16`.

**kilosort4**

In kilosort4, the processing is done in `float32`, as mentioned [here](https://github.com/MouseLand/Kilosort/blob/9bdd8f2814eb433fa4f881c4f735aabce605f871/kilosort/io.py#L1040). The covariance [matrix is computed](https://github.com/MouseLand/Kilosort/blob/9bdd8f2814eb433fa4f881c4f735aabce605f871/kilosort/preprocessing.py#L96) in `float32` (`X` is in `float32` and `torch.zeros` initialises in `float32` by default). You can see the resacling by 200 and `int16` conversion for example [here](https://github.com/MouseLand/Kilosort/blob/9bdd8f2814eb433fa4f881c4f735aabce605f871/kilosort/io.py#L1079).

Curiously, I can see in the data loading class that [here](https://github.com/MouseLand/Kilosort/blob/9bdd8f2814eb433fa4f881c4f735aabce605f871/kilosort/io.py#L565) that if the data is `uint16` it is [cast to float](https://github.com/MouseLand/Kilosort/blob/9bdd8f2814eb433fa4f881c4f735aabce605f871/kilosort/io.py#L727). But in the class `__init__` it states the default datatype expected is `int16`. I can't find ATM where this cast from `int16` to `float32` is performed (it is definitely performed for `uint16`). I need to look into this further, from the other comments and code sections the data is definitely expected to be in `float32` for processing.

**kilosort<4** (based on 2.5)

First, the whitening matrix is [computed in `float32`](https://github.com/MouseLand/Kilosort/blob/f2e570c61b6e920d361bcae6e2350bea03a16506/preProcess/get_whitening_matrix.m#L42) (from data processed by [`gpufilter.py`](https://github.com/MouseLand/Kilosort/blob/f2e570c61b6e920d361bcae6e2350bea03a16506/preProcess/gpufilter.m#L18) which performs CAR and filtering in `float32`). Then float32 conversion + filtering and CAR is performed in the main [preprocessing loop](https://github.com/MouseLand/Kilosort/blob/f2e570c61b6e920d361bcae6e2350bea03a16506/preProcess/preprocessDataSub.m#L97), as well as whitening. Then cast back to int16 is performed [here](https://github.com/MouseLand/Kilosort/blob/f2e570c61b6e920d361bcae6e2350bea03a16506/preProcess/preprocessDataSub.m#L108), with the 200 scaling [incorporated into the whitening matrix](https://github.com/MouseLand/Kilosort/blob/f2e570c61b6e920d361bcae6e2350bea03a16506/preProcess/get_whitening_matrix.m#L69).

zm711:
Cool I guess they cast too. did you find the commit+PR where it was changed was there a comment about the dropping? Some of the preprocessing dtype stuff is a mystery to me so maybe someone had a reason?

JoeZiminski:
Thanks @zm711 good point I was just looking at the version diff, I think it was [here](https://github.com/SpikeInterface/spikeinterface/pull/2744/files/a0b96e86579401236f8d4aa56d63e1bcbec14da3#diff-0f43097f0e039881c32b35975143e51b78fb23b7b861dba0a59bbb42cd2cf35fL165).

yger:
Then indeed, this is a big mistake and random_data should always be broadcasted to float32 before estimating the whitening matrix

src/spikeinterface/preprocessing/whiten.py
```diff
@@ -19,6 +19,8 @@ class WhitenRecording(BasePreprocessor):
     recording : RecordingExtractor
         The recording extractor to be whitened.
     dtype : None or dtype, default: None
+        Datatype of the output recording (covariance matrix estimation
+        and whitening are performed in float64.
```
alejoe91, line 23:
```suggestion
        and whitening are performed in float64).
```

alejoe91 (approved):

src/spikeinterface/preprocessing/whiten.py
```diff
@@ -185,6 +189,7 @@ def compute_whitening_matrix(
 
     """
     random_data = get_random_data_chunks(recording, concatenated=True, return_scaled=False, **random_chunk_kwargs)
```
zm711, line 191: Could someone remind me why we want to `return_scaled=False` here?

alejoe91: because scaling only happens at the `BaseRecording` step. The whitening is happening at the `BaseSegment` step, which returns unscaled traces

zm711: Cool. Thanks.

alejoe91:
@JoeZiminski are you planning to extend tests here or is it good to merge?

zm711:
Could we add the regression test in an issue to add? We use MS5 which relies on the whitening machinery. Would love to have this on main for everyone.

JoeZiminski:
Hey @alejoe91 @zm711 hey both, sorry got waylaid with something but will add some tests today! agree this will be good to merge ASAP

src/spikeinterface/preprocessing/whiten.py
```diff
@@ -74,7 +76,9 @@ def __init__(
         dtype_ = fix_dtype(recording, dtype)
 
         if dtype_.kind == "i":
-            assert int_scale is not None, "For recording with dtype=int you must set dtype=float32 OR set a int_scale"
+            assert int_scale is not None, (
+                "For recording with dtype=int you must set the output dtype to float " " OR set a int_scale"
```
JoeZiminski, line 80:
```suggestion
                "For recording with dtype=int you must set the output dtype to float OR set a int_scale"
```

JoeZiminski:
Hey @alejoe91 @zm711 in the end adding test tests was taking longer than expected, I'll make a new PR for this. Feel free to merge! At the next meeting we can discuss where to include a warning (e.g. maybe 0.101.0 release notes, or the release notes for a new version).

In the end I reverted to `np.float32` to a) match what was there before b) match other implementation (kilosort) c) match GPU if this is implemented. The overflowing computation is on int16 ephys data and is the dot product between two channel timeseries of length determined by the random chunk, by default  it is 10k samples. In some insane extreme case, if somehow the raw data is all around +1000 (I don't think possible), demean is not selected, and the user specifies 1M samples instead of 10k, the value will be 1e+12, float32 max value is 3.4 * 1e+39, so I think we'll be fine!

Issue #3157, opened 2024-07-05
Proposal for handling the user interaction with time


h-mayorquin:
From today's meeting.

## Context:
Spikeinterface has two internal representation time:
cheap: sampling_frequency and t_start
expensive: time_vector

## Fine control of the time
Right now `t_start` is a public attribute at the RecordingSegment level and it is harder to access than maybe it needs be. We need a more comfortable setter. Plus, the setter should handle the fact that the cheap and expensive representations are mutually exclusive.

*  A proposal implementation that overloads the current `set_times` is in #3117. 
* @alejoe91 proposal, `set_times_info` as a setter that handles the above. I think this is a good idea because it reflects the other method we have `get_times_kwargs` and it is a nice symmetry from getters to setters. I am +1 on that

## End user API
My proposal for how end users should interact with time:
1) set_times always sets the the time vector `set_times` and this is the final source of truth after setting. 
2) I propose adding a second function `set_starting_time` or `shift_starting_time` (do not know what's better) that is agnostic to the internal implementation. If there is a time vector it shifts so `time_vector[0] = start_time` and if there is a `t_start` then it just shifts or changes that values.

Setting the correct times and shifting the current times are natural concepts that don't rely on internal implementation details and the functions above don't leak internal details to users.

Some extra notes:
1) This enables the workflow of shifting or doing whatever you want with the `time_vector` before setting it. So it does not get in the way of  @JoeZiminski workflow  (`set_times` is still `set_times`).
2) @alejoe91 mention that you could use the workflow of `get_times()` , modify the vector and finally use `set_times` again. I think this is good and very close to what I want but has the drawback that you always end with the expensive representation internally which I think we can avoid. Plus, right now this workflow throws a warning
3) @samuelgarcia mentioned that the same mechanism should be implemented for the sorting. Maybe using the registered recording method is an overkill.

In brief, I think we should create an API that does not leak the internal implementation details of how spikeinterface represents time. We might decide down the line to change the internal representation of time and this should not affect the API. Thus, I think those functions and their docstrings should not make references or rely on understanding of `t_start` and `time_vector`.

samuelgarcia:
Shitting and also compressing times!
For instance if you want to synchronize recording with distinct source clock (like 2 neuropixel probes) on long recording we need also to modify the sampling_freq to compensate for for the relative clock drift.

alejoe91:
> Shitting

Not sure what you mean with this! Ahahahha

[...]

PR #3509, opened 2024-10-28
Add `shift start time` function.

JoeZiminski:
This PR adds a 'shift start time' function that a user can use to set a new start time for their recording.

This follows from discussions on the time API #3117 and #3157. Briefly, there are two ways to handle time in SI, a `t_start` recording attribute that is used to set the start time. Under this regime, the (regularly) spaced timeseries is regenerated on the fly from `t_start`, num samples and sampling frequency. Alternatively, a (more memory heavy) `time_vector` can be set to explicitly set the time array for a recording. As I understand it this is useful in the case there are gaps in the recording or there is some clock drift. From the user perspective, they always get a time vector back from `get_times()`.

This PR implements a `shift_start_time` as suggested by @h-mayorquin as a convenient way for users to shift the start time of their recording, abstracting away the technical details of the implementation under the hood (`t_start` vs `time_vector`).

I have updated the PR #3072  containing documentation on times, the relevant proposed section can be found [here](https://spikeinterface--3072.org.readthedocs.build/en/3072/tutorials/core/plot_6_handle_times_ver2.html#shifting-the-start-time).

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -509,6 +509,26 @@ class BaseRecording(BaseRecordingSnippets):
             rs.t_start = None
             rs.sampling_frequency = self.sampling_frequency
 
+    def shift_start_time(self, shift, segment_index=None):
+        """
+        Shift the starting time of the times.
+
+        shift : int | float
+            The shift to apply to the first time point. If positive,
+            the current start time will be increased by `shift`. If
+            negative, the start time will be decreased.
+
+        segment_index : int | None
+            The segment on which to shift the times.
+        """
+        segment_index = self._check_segment_index(segment_index)
+        rs = self._recording_segments[segment_index]
+
+        if self.has_time_vector():
+            rs.time_vector += shift
+        else:
+            rs.t_start += shift
+
     def sample_index_to_time(self, sample_ind, segment_index=None):
         """
         Transform sample index into time in seconds
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -509,7 +509,7 @@ class BaseRecording(BaseRecordingSnippets):
             rs.t_start = None
             rs.sampling_frequency = self.sampling_frequency
 
-    def shift_start_time(self, shift, segment_index=None):
+    def shift_start_time(self, shift: int | float, segment_index: int | None = None) -> None:
         """
         Shift the starting time of the times.
 
@@ -536,15 +536,14 @@ class BaseRecording(BaseRecordingSnippets):
             else:
                 rs.t_start += shift
 
-    def sample_index_to_time(self, sample_ind, segment_index=None):
-        """
-        Transform sample index into time in seconds
-        """
+    def sample_index_to_time(self, sample_ind: int, segment_index: int | None = None):
+        """ """
         segment_index = self._check_segment_index(segment_index)
         rs = self._recording_segments[segment_index]
         return rs.sample_index_to_time(sample_ind)
 
-    def time_to_sample_index(self, time_s, segment_index=None):
+    def time_to_sample_index(self, time_s: float, segment_index: int | None = None):
+        """ """
         segment_index = self._check_segment_index(segment_index)
         rs = self._recording_segments[segment_index]
         return rs.time_to_sample_index(time_s)
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -526,8 +526,6 @@ class BaseRecording(BaseRecordingSnippets):
             The segment on which to shift the times. if `None`, all
             segments will be shifted.
         """
-        self._check_segment_index(segment_index)  # Check the segment index is valid only
-
         if segment_index is None:
             segments_to_shift = range(self.get_num_segments())
         else:
@@ -536,7 +534,7 @@ class BaseRecording(BaseRecordingSnippets):
         for idx in segments_to_shift:
             rs = self._recording_segments[idx]
 
-            if self.has_time_vector():
+            if self.has_time_vector(segment_index=idx):
                 rs.time_vector += shift
             else:
                 rs.t_start += shift
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -539,14 +539,15 @@ class BaseRecording(BaseRecordingSnippets):
             else:
                 rs.t_start += shift
 
-    def sample_index_to_time(self, sample_ind: int, segment_index: int | None = None):
-        """ """
+    def sample_index_to_time(self, sample_ind, segment_index=None):
+        """
+        Transform sample index into time in seconds
+        """
         segment_index = self._check_segment_index(segment_index)
         rs = self._recording_segments[segment_index]
         return rs.sample_index_to_time(sample_ind)
 
-    def time_to_sample_index(self, time_s: float, segment_index: int | None = None):
-        """ """
+    def time_to_sample_index(self, time_s, segment_index=None):
         segment_index = self._check_segment_index(segment_index)
         rs = self._recording_segments[segment_index]
         return rs.time_to_sample_index(time_s)
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -511,8 +511,7 @@ class BaseRecording(BaseRecordingSnippets):
 
     def shift_times(self, shift: int | float, segment_index: int | None = None) -> None:
         """
-        Shift all times by a scalar value. The default behaviour is to
-        shift all segments uniformly.
+        Shift all times by a scalar value.
 
         Parameters
         ----------
@@ -523,8 +522,8 @@ class BaseRecording(BaseRecordingSnippets):
             started earlier.
 
         segment_index : int | None
-            The segment on which to shift the times. if `None`, all
-            segments will be shifted.
+            The segment on which to shift the times.
+            If `None`, all segments will be shifted.
         """
         if segment_index is None:
             segments_to_shift = range(self.get_num_segments())
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -509,19 +509,24 @@ class BaseRecording(BaseRecordingSnippets):
             rs.t_start = None
             rs.sampling_frequency = self.sampling_frequency
 
-    def shift_start_time(self, shift: int | float, segment_index: int | None = None) -> None:
+    def shift_times(self, shift: int | float, segment_index: int | None = None) -> None:
         """
-        Shift the starting time of the times.
+        Shift all times by a scalar value. The default behaviour is to
+        shift all segments uniformly.
 
+        Parameters
+        ----------
         shift : int | float
-            The shift to apply to the first time point. If positive,
-            the current start time will be increased by `shift`. If
-            negative, the start time will be decreased.
+            The shift to apply. If positive, times will be increased by `shift`.
+            e.g. shifting by 1 will be like the recording started 1 second later.
+            If negative, the start time will be decreased i.e. as if the recording
+            started earlier.
 
         segment_index : int | None
-            The segment on which to shift the times.
+            The segment on which to shift the times. if `None`, all
+            segments will be shifted.
         """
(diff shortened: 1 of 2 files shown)


zm711 (commented):
I'm wondering if somewhere we should indicate that if `t_start` it shifts that, but if `time_vector` then it shifts all times within the vector.

samuelgarcia:
HI Joe.
I am not sure this approach is persistent to json dump/load cycle. If this is a need then we should have a dedicated class for that handle explicitly the shbift in the kwargs

JoeZiminski:
Hey @zm711 this is a good point, we (@h-mayorquin) decided that from a user perspective the implementation under the hood should not leak through the API. So as far as the user is concerned the user just as 'times' and there is no distinction between the `t_start` implementation and `time_vector`. But maybe this is useful to have somewhere in the docstring for the more interested user, maybe notes?

Thanks @samuelgarcia I'm not 100% sure I understand. This PR is editing the existing segment attributes, will these not be propagated to a newly saved version of the recording? This would be an important test.

e.g. (pseudocode, I can't remember the exact commands)

```
recording.save("some_path")  # save with original times

recording.shift_start_time(10)  # shift times

recording.save("some_path")  # the saved data will not be updated?
load_recording - si.load("some_path")  # this will not have the updated times?
```

src/spikeinterface/core/baserecording.py
```diff
@@ -509,6 +509,26 @@ def reset_times(self):
             rs.t_start = None
             rs.sampling_frequency = self.sampling_frequency
 
+    def shift_start_time(self, shift, segment_index=None):
```
h-mayorquin, line 512: Typing x D ?

The function could also be called "shift_times" to avoid making a reference to the internal detail of "t_start", etc but I think this works as well.

src/spikeinterface/core/baserecording.py
```diff
@@ -509,6 +509,26 @@ def reset_times(self):
...
+        """
+        Shift the starting time of the times.
+
+        shift : int | float
+            The shift to apply to the first time point. If positive,
+            the current start time will be increased by `shift`. If
+            negative, the start time will be decreased.
+
+        segment_index : int | None
```
h-mayorquin, line 521: Interesting, I was thinking that all segments should be shift at the same times but  adding more fine grain control might make sense. 

How are you thinking about this? Do you think having the default as shifting all the segments makes sense?

h-mayorquin (commented):
Thanks for pushing this forward despite all the noise that we have created. 

Things to test for:
1) recording with times vs recording with t_start implementation, see that the shift works.
2) Json and pickling of base recording object.
3) I think this is not properly propagated to preprocessing (vague memory), in my view, it is fine if we deal with that into another PR but is a good thing to keep on mind.

h-mayorquin:
Also, @JoeZiminski @zm711 I am against adding notes about internal implementation details on the docstring (for the reason that Joel gave) but I think a top level comment on the function can work.

My reasoning is that if you change the internal implementation then you need to update the docstring which should be user facing and unaffected by this.

src/spikeinterface/core/baserecording.py
zm711: I think I agree. I would name the function whatever is user facing more. So if the "times" are then times, but if the start is more important than start.

zm711:
I'm fine seeing myself off of this thread. I honestly never use this machinery for my purposes so it is really hard for me to imagine the point of this and how I as a user would interact with these functions. So what is implementation vs what is user facing is hard for me to parse. Maybe a call at some point (dedicated just to time machinery) between those who care would be helpful (at least for me). I agree with Heberto not to leak internal stuff, but so I would like to understand internal better.

src/spikeinterface/core/baserecording.py
JoeZiminski: Typing!! Need to add! 🎉 

Yes I think I agree, my only thought is, is there a world where 'shift_times' is not clear because it might imply the possibility to make completely non-uniform shifts e.g. supply a vector of shifts? The only benefit of something like `shift_start_times` is it provides a reference point from where the shift will occur. Maybe it is also more intuitive as if you have a recording it is natural to imagine starting it at a different time. But maybe it is confusing with the `t_start` implementation. I'm basically 50/50 so happy to take the way of the majority!

src/spikeinterface/core/baserecording.py
JoeZiminski: Good point I wasn't really thinking of this in reference to a real world use-case. My only thinking was if you want to shift all segments, this approach allows you to (a bit verbose though) but vice-versa, you are stuck if the implementation automatically shifts all. I can imagine if you are starting/stopping a recording and storing e.g. for long trial-like periods with the parts in different segments, you may want to adjust each segment separately to align with some trigger?

JoeZiminski:
Thanks @h-mayorquin I agree a top level comment is nice for this. Cheers for the feedback @zm711! Yes I think a call would be nice, @samuelgarcia also mentioned at some point he will do some time-alignment stuff so that might be a good opportunity.

src/spikeinterface/core/baserecording.py
h-mayorquin: I am in the same train. 50/50, feel free to pick what your hearth thinks is better!

src/spikeinterface/core/baserecording.py
h-mayorquin: That's a good point. Having control per segment sounds like a good idea. What do you think of having `segment_index=None` shift all the segments by default?

src/spikeinterface/core/baserecording.py
JoeZiminski: Haha okay ❤️ I will decide at the last moment! Thanks for the test suggestions, also good point on the preprocessing steps I am not sure either, I opened #3534

src/spikeinterface/core/baserecording.py
JoeZiminski: Yes that's much nicer, I think that makes more sense than `segmnet_index=None` only shifting the first segment or something. Should `segment_index` be default `None` or a required argument to make the behaviour explicit?

h-mayorquin: My preference is by default you shift everything and then shift a segment if you specify. But I am highly biased because that makes my workflow convenient and I have not think this through. A somewhat weak argument for it is that if I say "shift_times" I want to shift all the times of the object.

zm711: I think I may lean toward Heberto on this. Why the first segment if `None`. There is no obvious logic for that that I can see? I think the options are None does all or you require explicit. I don't think `None` should implicitly choose a segment for the user. But that's just how I would think to use this as a user.

For a use case maybe I do an `AppendRecording` (I forget the actual class name haha) where I fuse three monosegments together but realized I need to shift them. I like None in this case if I need to move all of them as once otherwise if we don't have a None option then I would have to do a for loop going through each segment. If I only need to do one of the segments I always have the option to specify it rather than have `None` or 0 being the first segment.

zm711:
```

File: src/spikeinterface/core/baserecording.py (hunk 2 above)
Comment (JoeZiminski) [reply src/spikeinterface/core/baserecording.py:525]: Great thanks both I agree this feels natural, let's have `None` shift all segments and the default `segment_index=None`.

BTW, a complete aside. I just noticed `times = recording.get_times()` returns a veiw to the time vector (if it exists) otherwise an auto-generated vector of times. Therefore if you edit the `times` (and the recording has a time vector) you'll actually edit the times on the recording. I think this is probably a bad idea, and also means there is different behaviour using `time_vector` vs. `t_start` (in which editing the returned `times` does not change anything on the recording itself). 

What do people think of introducing a `copy=True` (i.e. default on) option for `get_times()` that will make a copy of  `time_vector` (if it exists). If `copy=False`, the time vector will not be copied. Unfortunately this will leak the implementation to the docstring, but I guess this option is necessary in the case of very large time arrays.

Comment (h-mayorquin) [reply src/spikeinterface/core/baserecording.py:524]: Mmm interesting point. I need to think on this.  My first instinct is that I would rather not have an argument and always return a copy.

Comment (JoeZiminski) [reply src/spikeinterface/core/baserecording.py:524]: I agree actually I think that is much better, much cleaner. There's no point trying to preemptively support a super-edge case (and in such a case the time vector can be accessed directly anyways).

Comment (h-mayorquin) [reply src/spikeinterface/core/baserecording.py:524]: Yes, internally for memory efficient reasons we can access the attribute or create a private method. Great point!

File: src/spikeinterface/core/tests/test_time_handling.py [hunk 3]
```diff
@@ -305,7 +311,87 @@ def test_sorting_analyzer_get_durations_no_recording(self, time_vector_recording
...
+    @pytest.mark.parametrize("fixture_name", ["time_vector_recording", "t_start_recording"])
+    @pytest.mark.parametrize("shift", [-123.456, 123.456])
+    def test_shift_times_different_segments(self, request, fixture_name, shift):
+        """
+        Shift each segment separately, and check the shifted segment only
+        is shifted as expected.
+        """
+        _, times_recording, all_times = self._get_fixture_data(request, fixture_name)
+
+        num_segments, orig_seg_data = self._store_all_times(times_recording)
```
h-mayorquin, line 341: Maybe you could just do recording.copy?

h-mayorquin (approved):
LGTM.

src/spikeinterface/core/tests/test_time_handling.py
JoeZiminski: Nice this will allow to remove a lot of boiler plate and make the assert more explicit. However check this out!:

```
        import copy

        _, times_recording, all_times = self._get_fixture_data(request, fixture_name)

        orig_times_recording = copy.deepcopy(times_recording)

        # num_segments, orig_seg_data = self._store_all_times(times_recording)

        times_recording.shift_times(shift)  # use default `segment_index=None`

        for idx in range(num_segments):
            assert np.allclose(
                orig_times_recording.get_times(segment_index=idx),
                times_recording.get_times(segment_index=idx) - shift,
                rtol=0, atol=1e-8
            )
```
This fails,  `orig_times_recording.get_times(segment_index=idx)` gives 
```
# array([0.00000000e+00, 3.33333333e-05, 6.66666667e-05, ...,
 #      9.99990000e+00, 9.99993333e+00, 9.99996667e+00])
```

It seems the time vector info is not being propagated in the copy, though it seemed to work okay for the save/ load 🤔 will have to come back to this, do you have ideas why this might be happening?

h-mayorquin: Mmm yeah, I think that that's probably by design. I think cloning uses to do_dict mechanism and that coupled with saving to disk and serializing. Because timestamps can be large we probably don't want that there.

The solution is to change copy (which should be a in memory thing) to copy the time vector but I think that should be done in another PR.

zm711: A little out of scope, but do you remember when we were looking up copy vs clone at Edinburgh. I forget what it said (but basically one is deep and one is suppose to be shallow). It would probably be worth us making clear in our docstrings if our copy's are shallow or not (also for another PR).

alejoe91: > I'm a little confused sorry, what is `clone` vs `copy`? BTW I forgot to say based on the suggested command, I got an error:
> 
> ```
> (Pdb) times_recording.copy()
> *** AttributeError: 'NoiseGeneratorRecording' object has no attribute 'copy'
> ```
> 
> should this have worked?

No I don't think this would work. I guess @h-mayorquin was suggesting to do a:
```
from copy import copy

copy(times_recording)
```

zm711: > I'm a little confused sorry, what is clone vs copy?

Some projects use `copy(deep=True/False)` to indicate whether to just do metadata vs actual array data whereas other projects use `clone` vs `copy` to make the deep vs shallow distinction (again deep being everything and shallow just being metadata). Because there is no requirement for clone and copy being shallow or deep it would probably be better for us as a project to either make it clear in the docstring or for us to use the `clone/copy` names to make clear which functions are deep and which are shallow. Is that clearer? Again not crucial to this PR, but related to the fact that the array is not being copied (ie our copy is shallow).

JoeZiminski:
@alejoe91 this is good to go from my end

Issue #3328, opened 2024-08-23
Question about interpolate with motion of a different sampling rate than recording


DaohanZhang:
Hey, as the document mentioned, some drift-correcting methods like dredge_lfp require preprocessing steps, including resampling. I'm curious about whether I need to resample the motion data back to match the original recordings. If I need to resample back, we use low-frequency (LF) recordings from Neuropixels to obtain motion data sometimes and apply it to action-potential (AP) recordings, but the sampling ratio isn't an integer (not exactly 12) due to timing differences. If I don't resample the motion data back, I encounter an error when saving recordings to Zarr files. The index 208455 corresponds to the timestep of the motion data, not the recording itself.
```python
import numpy as np
import matplotlib.pyplot as plt

import spikeinterface.full as si  # import core only
import spikeinterface.extractors as se
import spikeinterface.preprocessing as spre
import spikeinterface.sorters as ss
import spikeinterface.postprocessing as spost
import spikeinterface.qualitymetrics as sqm
import spikeinterface.comparison as sc
import spikeinterface.exporters as sexp
import spikeinterface.curation as scur
import spikeinterface.widgets as sw

lfpraw = se.read_spikeglx("/home/zhangdaohan20h/public_data/NPX_examples/Pt01/", load_sync_channel=False, 
                          stream_id="imec0.lf")
## the file named Pt02.imec0.ap.bin & Pt02.imec0.ap.meta
lfpraw
# convert to floating point
lfprec = si.astype(lfpraw, np.float32)
# ensure depth order
lfprec = si.depth_order(lfprec)

cutoff_um = 8000
if cutoff_um is not None:
    geom = lfprec.get_channel_locations()
    lfprec = lfprec.remove_channels(lfprec.channel_ids[geom[:, 1] > cutoff_um])

# bandpass filter
# we do an aggressive one since we plan to downsample
lfprec = si.bandpass_filter(
    lfprec,
    freq_min=0.5,
    freq_max=250,
    margin_ms=1000,
    filter_order=3,
    dtype="float32",
    add_reflect_padding=True,
)
# fancy bad channels detection and removal from the International Brain Lab
bad_chans, labels = si.detect_bad_channels(lfprec, psd_hf_threshold=1.4, num_random_chunks=100, seed=0)
print("Found bad channels", bad_chans)
lfprec = lfprec.remove_channels(bad_chans)
# correct for ADC sample shifts
lfprec = si.phase_shift(lfprec)
# common median reference
[...]

PR #3517, opened 2024-11-04
Fix a cross-band interpolation bug, and allow time_vector in interpolate_motion

cwindolf:
Hi all, this PR addresses two issues related to motion interpolation, specifically estimating motion in the LFP band and applying the correction to the AP band.
 - First, @DaohanZhang's IndexErrors, discussed at:
   - https://github.com/evarol/dredge/issues/1
   - fixes https://github.com/SpikeInterface/spikeinterface/issues/3328
   - I was able to reproduce this issue in real data and the changes here and at https://github.com/evarol/dredge/commit/5cfe179f5896de859fc4b658bd5330f499dd57c8 resolve this issue. A test case has been added.
 - Second, in fixing that issue I found it easiest just to handle time vectors with a searchsorted. Hopefully this is not controversial anymore?
   - As discussed in https://github.com/SpikeInterface/spikeinterface/issues/3029, this is nowhere near a performance issue. Your computer will spend 1e6 amount of time in the matrix multiply.
   - The implementation here is commented and tested and allows for unevenly spaced motion and recording time bins
   - I would consider that this closes https://github.com/SpikeInterface/spikeinterface/issues/3029 -- @JoeZiminski and others, would you agree?

I also added a small test for cross-band registration. The idea was to generate an "LFP" and an "AP" recording, which are zeros except for a channel of 1s which jumps after 3s. Motion is estimated from the LFP recording and then used to interpolate the AP recording, and we assert that the result has no drift at all.

Interestingly, since the LFP recording's time bins are treated as bin centers by `interpolate_motion` but as left bin edges by `time_to_sample_index()`, there is a half-bin offset between the motion and the recording which I had to account for in this test. But, I actually think this is the right way to go? 2 reasons:
 - `interpolate_motion`: It makes sense to treat motion time bins as centers even when they correspond to LFP recording samples; this artificial case is not realistic and in real life we believe that behavior leads to less error.
 - `time_to_sample_index()`: obviously, would not want to change the behavior here, everybody expects the time in seconds to be "floored" to the previous sample.
   - If I'm wrong about that, then the `searchsorted()` line needs to be changed in `time_to_sample_index()` to map to the nearest bin, and logic similar to what I've written in `interpolate_motion_on_traces()` can be used. But, that's a separate issue from this PR -- just wanted to mention that in case my understanding is off.

(And, to @DaohanZhang: this PR together with a recent dredge repo commit (https://github.com/evarol/dredge/commit/5cfe179f5896de859fc4b658bd5330f499dd57c8) should fix your issue, if you'd like to test them. I've updated the [demo notebook there](https://github.com/evarol/dredge/blob/main/notebook/lfp_registration_and_interpolation_demo.ipynb) to reflect the latest updates -- dredge's understanding of spikeinterface's time handling needed to be updated.)

### DIFF
--- a/src/spikeinterface/sortingcomponents/motion/motion_interpolation.py
+++ b/src/spikeinterface/sortingcomponents/motion/motion_interpolation.py
@@ -130,7 +130,7 @@ def interpolate_motion_on_traces(
 
     # -- determine the blocks of frames that will land in the same interpolation time bin
     if interpolation_time_bin_centers_s is None and interpolation_time_bin_edges_s is None:
-        interpolation_time_bin_centers_s = motion.temporal_bin_centers_s[segment_index]
+        interpolation_time_bin_centers_s = motion.temporal_bins_s[segment_index]
         interpolation_time_bin_edges_s = motion.temporal_bin_edges_s[segment_index]
     else:
         interpolation_time_bin_centers_s, interpolation_time_bin_edges_s = ensure_time_bins(


src/spikeinterface/sortingcomponents/motion/tests/test_motion_interpolation.py
```diff
@@ -148,5 +206,6 @@ def test_InterpolateMotionRecording():
 if __name__ == "__main__":
     # test_correct_motion_on_peaks()
     # test_interpolate_motion_on_traces()
-    test_interpolation_simple()
-    test_InterpolateMotionRecording()
+    # test_interpolation_simple()
+    # test_InterpolateMotionRecording()
+    test_cross_band_interpolation()
```
zm711, line 211: Are you like Sam and use python testing rather than pytest : 0 ?

src/spikeinterface/sortingcomponents/motion/motion_interpolation.py
```diff
@@ -122,14 +122,23 @@ def interpolate_motion_on_traces(
...
+    bin_left = np.zeros_like(time_bins)
+    # it's fine to use the first bin center for the first left edge
+    bin_left[0] = time_bins[0]
+    bin_left[1:] = 0.5 * (time_bins[1:] + time_bins[:-1])
+    bin_inds = np.searchsorted(bin_left, times, side="right") - 1
+
     # the time bins may not cover the whole set of times in the recording,
     # so we need to clip these indices to the valid range
-    np.clip(bin_inds, 0, time_bins.size, out=bin_inds)
+    np.clip(bin_inds, 0, time_bins.size - 1, out=bin_inds)
```
zm711, line 141: And why the -1 here? To get to the left side again? Just so we have a record of the though process.

zm711: Okay cool. I'm always super paranoid about indexing issues since I was taught Matlab first :)

src/spikeinterface/core/baserecording.py
```diff
@@ -921,11 +917,11 @@ def time_to_sample_index(self, time_s):
                 sample_index = time_s * self.sampling_frequency
             else:
                 sample_index = (time_s - self.t_start) * self.sampling_frequency
-            sample_index = round(sample_index)
+            sample_index = np.round(sample_index).astype(int)
```
cwindolf, line 920: Flagging this change! Sorry if this is not relevant enough to this PR, but I thought that while I was working on time logic it would be good to fix this last small quality of life thing (vectorizing time_to_sample_index -- note that the scalar case still behaves the same).

cwindolf:
Hi @samuelgarcia -- updated this PR with the stuff we talked about. To recap: we need the interpolation time bin edges to interpolate traces, which we were recomputing over and over for each interpolation chunk. Now they are cached. They are cached in the motion object, but also in the interpolated recording object if for example a user specifies their own interpolation time bin size.

I also added a small thing to vectorize time_to_sample_index, flagged in a comment above. If that feels outside the scope here I'm happy to put it in another PR!

src/spikeinterface/core/baserecording.py
cwindolf: I guess this follows @h-mayorquin 's [comment](https://github.com/SpikeInterface/spikeinterface/pull/3119#pullrequestreview-2154668157)

src/spikeinterface/sortingcomponents/motion/tests/test_motion_interpolation.py
```diff
@@ -67,18 +65,20 @@ def test_interpolate_motion_on_traces():
...
+                times,
+                channel_locations,
+                motion,
+                channel_inds=None,
+                spatial_interpolation_method=method,
+                interpolation_time_bin_centers_s=interpolation_time_bin_centers_s,
+                # spatial_interpolation_kwargs={},
+                spatial_interpolation_kwargs={"force_extrapolate": True},
+            )
+            assert traces.shape == traces_corrected.shape
```
JoeZiminski, line 80: Outside the scope of this PR (as tests were like this anyways), but it would be nice to extend these tests to explicitly check the values are correct. Testing the shape will not pick up any regressions that mess up the actual computation but leave the shape in tact. Unfortunately these are the least likely to be picked up as any erroneous shapes will probably crash at runtime anyway.

I am wondering if traces is a simple 3x3 array (3 channels, 3 timepoints) it would be relatively easy to compute manually the expected results of kriging, idw and NN interpolation and check against the output of this function? Of course, this is outside the scope of the PR so feel free to ignore and I can write an issue!

JoeZiminski (commented):
Hey Charlie this is very nice, I agree `searchsorted` is succinct and should pose no issues with performance. The new test is great will help the code be robust going forward and also helps explain what the functions are doing. Please see some general comments / suggestions none of these are major so feel free to ignore.

src/spikeinterface/core/baserecording.py
JoeZiminski: Interesting! This is nice, I think the only consideration is possible overflow for longer recordings, as `int64` is capped but python `int()` is not capped. @h-mayorquin has been focussing on this more, but looking at a quick example below it should be fine.

`int64` max value is `9,223,372,036,854,775,807`. If we take a neuropixels recording, continuous for 2 months (not unfeasible these days) we have `(30,000 * 60 * 60 * 24 * 60) = 165888000000`  (samples per s x seconds per minute x minutes per hour x hours per day x ~days in 2 month) (please check). But, maybe in 5 years people are sampling at 100 kHz and doing year long recordings 😆 we would have max index of `(3.1536e+12)`. So I think should be sufficient under all feasible uses, but something to consider.

cwindolf: Interesting... wait, maybe I'm doing the math wrong, but don't we have:

```python
# int64 max val     | samples/sec | sec/min | min/hr | hr/day
9223372036854775807 /     100_000 /      60 /     60 /     24
# => 1_067_519_911.6730065 days
```
which is quite a long time? ([wolfram double check](https://www.wolframalpha.com/input?i=9223372036854775807+*+10+microseconds+%3D+%3F+days))

cwindolf:
Thanks a ton for the review @JoeZiminski ! I think I've made changes corresponding to all of your comments. No clue why some of my comments got duplicated... that's what I get for trying out this github code spaces thing!

samuelgarcia (approved):

samuelgarcia:
This is OK for me.

PR #3518, opened 2024-11-05
Sc2 making use of new recording slices

yger:
This PR adds the possibility for SC2 to make use of the newly introduced recording slices. I.e. while detecting peaks, chunks are shuffled given a seed, and when enough peaks have been collected for clustering, we stop looking for peaks instead of collecting everything, then slicing


yger:
Depends on #3533, but need to understand why this one is failing

src/spikeinterface/sorters/internal/spyking_circus2.py
```diff
@@ -51,7 +51,7 @@ class Spykingcircus2Sorter(ComponentsBasedSorter):
             },
         },
         "clustering": {"legacy": True},
-        "matching": {"method": "circus-omp-svd"},
+        "matching": {"method": "wobble"},
```
samuelgarcia, line 54: what ?

yger: Just playing, will be reverted asap

src/spikeinterface/sortingcomponents/tools.py
```diff
@@ -151,3 +300,21 @@ def fit_sigmoid(xdata, ydata, p0=None):
...
+    chunk_size = ensure_chunk_size(recording, **job_kwargs)
+    recording_slices = []
+    for segment_index in range(recording.get_num_segments()):
+        num_frames = recording.get_num_samples(segment_index)
+        chunks = divide_segment_into_chunks(num_frames, chunk_size)
+        recording_slices.extend([(segment_index, frame_start, frame_stop) for frame_start, frame_stop in chunks])
+
+    if seed is not None:
+        rng = np.random.RandomState(seed)
+        recording_slices = rng.permutation(recording_slices)
```
samuelgarcia, line 318:
```suggestion

    recording_slices = rng.permutation(recording_slices)
```

yger: this will not work if rng is not created, without seed, will it?

samuelgarcia, line 318:
```suggestion

    rng = np.random.default_rng(seed)
    recording_slices = rng.permutation(recording_slices)
```

samuelgarcia (approved):

PR #3519, opened 2024-11-05
Don't let decimate mess with times and skim tests

alejoe91:
Deep bug in decimate, that deletes existing time vectors for segments!

This PR modifies the behavior so that `time_vectors` are also decimated, which is very intuitive.

Also skimmed a but the tests (previously 2100 tests with parameterized, now "only" 25)

### DIFF
--- a/src/spikeinterface/preprocessing/decimate.py
+++ b/src/spikeinterface/preprocessing/decimate.py
@@ -105,7 +105,7 @@ class DecimateRecordingSegment(BaseRecordingSegment):
             if parent_recording_segment.t_start is None:
                 t_start = None
             else:
-                t_start = parent_recording_segment.t_start + decimation_offset / parent_rate
+                t_start = parent_recording_segment.t_start + (decimation_offset / parent_rate)
 
         # Do not use BasePreprocessorSegment bcause we have to reset the sampling rate!
         BaseRecordingSegment.__init__(
--- a/src/spikeinterface/preprocessing/decimate.py
+++ b/src/spikeinterface/preprocessing/decimate.py
@@ -63,18 +63,15 @@ class DecimateRecording(BasePreprocessor):
                 f"Consider combining DecimateRecording with FrameSliceRecording for fine control on the recording start/end frames."
             )
         self._decimation_offset = decimation_offset
-        resample_rate = self._orig_samp_freq / self._decimation_factor
+        decimated_sampling_frequency = self._orig_samp_freq / self._decimation_factor
 
-        BasePreprocessor.__init__(self, recording, sampling_frequency=resample_rate)
+        BasePreprocessor.__init__(self, recording, sampling_frequency=decimated_sampling_frequency)
 
-        # in case there was a time_vector, it will be dropped for sanity.
-        # This is not necessary but consistent with ResampleRecording
         for parent_segment in recording._recording_segments:
-            parent_segment.time_vector = None
             self.add_recording_segment(
                 DecimateRecordingSegment(
                     parent_segment,
-                    resample_rate,
+                    decimated_sampling_frequency,
                     self._orig_samp_freq,
                     decimation_factor,
                     decimation_offset,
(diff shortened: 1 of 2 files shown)


src/spikeinterface/preprocessing/decimate.py
```diff
@@ -93,22 +90,26 @@ class DecimateRecordingSegment(BaseRecordingSegment):
...
+            time_vector = parent_recording_segment.time_vector[decimation_offset::decimation_factor]
+            decimated_sampling_frequency = None
+            t_start = None
         else:
-            new_t_start = parent_recording_segment.t_start + decimation_offset / parent_rate
+            time_vector = None
+            if parent_recording_segment.t_start is None:
+                t_start = None
+            else:
+                t_start = parent_recording_segment.t_start + decimation_offset / parent_rate
```
zm711, line 108: This is just a stylistic comment, but I prefer parentheses to help me with order of operations. It makes sense we need the offset/rate, but for something like this I'm always worried where we wanted to add first then divide or divide then add. The parentheses make it explicit, but I'm fine being overruled there.

zm711 (commented):
Couple comments. I think Joe will definitely be able to comment better on the time component.

src/spikeinterface/preprocessing/decimate.py
```diff
@@ -93,22 +90,26 @@ class DecimateRecordingSegment(BaseRecordingSegment):
     def __init__(
         self,
         parent_recording_segment,
-        resample_rate,
+        decimated_sampling_frequency,
```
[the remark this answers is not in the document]
alejoe91: yes, but it's soo deep in the API that I'm 100% sure it wouldn't affect anyone's workflow.

If it were at the `DecimateRecording` level, than we should have worried about back-compatibility because of saved objects/JSON files. But since it's the segment which is instantiated on the fly we don't have to worry about it (and I think the naming is much more in line with the overll API)

src/spikeinterface/preprocessing/decimate.py
```diff
@@ -93,22 +90,26 @@ class DecimateRecordingSegment(BaseRecordingSegment):
...
         parent_rate,
         decimation_factor,
         decimation_offset,
         dtype,
     ):
-        if parent_recording_segment.t_start is None:
-            new_t_start = None
+        if parent_recording_segment.time_vector is not None:
+            time_vector = parent_recording_segment.time_vector[decimation_offset::decimation_factor]
+            decimated_sampling_frequency = None
```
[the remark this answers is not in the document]
alejoe91: currently `time_vector` and `sampling_frequency`/`t_start` representation of time in the segment are still mutually exclusive, so we need to set the sampling_freq to None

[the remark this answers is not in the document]
zm711: Okay cool. Maybe we can have a chat about this at some point. I don't want to take up developer meeting time for this necessarily, but since I don't use the time api I don't know it well enough. :)

src/spikeinterface/preprocessing/tests/test_decimate.py
```diff
@@ -8,39 +8,74 @@
...
-@pytest.mark.parametrize("decimation_factor", [1, 9, 10, 11, 100, 101])
-@pytest.mark.parametrize("start_frame", [0, 1, 5, None, 1000])
-@pytest.mark.parametrize("end_frame", [0, 1, 5, None, 1000])
-def test_decimate(N_segments, decimation_offset, decimation_factor, start_frame, end_frame):
-    rec = generate_recording()
-
-    segment_num_samps = [101 + i for i in range(N_segments)]
-
+@pytest.mark.parametrize("num_segments", [1, 2])
+@pytest.mark.parametrize("decimation_offset", [0, 5, 21, 101])
```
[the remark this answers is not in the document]
alejoe91: added

zm711:
RTD has been so flaky. It has consistently failed on one of my Neo PRs for no reason....

alejoe91:
> RTD has been so flaky. It has consistently failed on one of my Neo PRs for no reason....

Yeah we can look at that later..let me know if it looks good!

src/spikeinterface/preprocessing/tests/test_decimate.py
```diff
@@ -8,39 +8,74 @@
...
+    times = [rec.get_times(0) + 10, rec.get_times(1) + 20]
+    for i, t in enumerate(times):
+        rec.set_times(t, i)
+
+    decimation_factor = 2
+    decimation_offset = 1
+    decimated_rec = DecimateRecording(rec, decimation_factor, decimation_offset=decimation_offset)
+
+    for segment_index in range(rec.get_num_segments()):
+        assert np.allclose(
```
zm711, line 62: should we explicitly set the tolerance that we tolerate? We often get flakiness due to floating point rounding. does `get_times` return floats?

alejoe91: good point. I think here all equal will work too. I'll push an update

alejoe91: @zm711 use `np.testing/assert_array_almost_equal(..., decimal=10)`

zm711: You think 10 decimal places? It looks like assert_array_almost_equal also deals with nan's which is nice. I vaguely remembering Heberto having to slowly relax one of these style of tests because it keep failing. 10 decimals seems super exact. Why not the default? (which is 7 based on most recent docs).

alejoe91: In my tests the absolute error was e-15 :)

zm711 (approved):
This works for me :)

Issue #2843, opened 2024-05-14
[feature request] Export to IBL Atlas tool?


jonahpearl:
Hi Si folks — I'm wondering if there's a feature (or any plan to develop one) to allow exporting a recording to the IBL's neat [Atlas Electrophysiology tool](https://www.youtube.com/watch?v=HJC08wW0H4w&list=PLfhWmWntvjl7kljKozClpjS29DoY8V5pB&index=16), which allows post-hoc realignment of probe tracks based on characteristics of the ephys data.

You can read their description of how to export data for their tool [here](https://github.com/int-brain-lab/iblapps/wiki/4.-Preparing-data-for-ephys-GUI#preparing-ephys-data). Unfortunately, it seems to only support data acquired with spikeGLX. I've been using open-ephys, so I'm out of luck, and it seems at least a few others have [run into this problem](https://github.com/int-brain-lab/iblapps/issues/47) before. As I wrote in that issue, it seems like the exported data is quite similar to what phy needs, and in the same format, so I imagine that si's `export_to_phy` function is a good starting point. (You can download example data [here](https://ibl.flatironinstitute.org/public/), "ephys_alignment_sample_data.zip").

I think there's a good chance that a few folks in my lab will develop an exporter for this if one doesn't already exist, but just checking in case someone has already done it...

Thank you!

samuelgarcia:
Hi,
feel free to propose a PR for implementing `export_to_ibl()`.
The other way which is more work would be to patch this ibl data tools to support spikeinterface object directly as input!

alejoe91:
@jonahpearl We actually have some code lying around for this: 
https://github.com/AllenNeuralDynamics/aind-ephys-ibl-alignment-gui-converter/blob/main/code/run_capsule.py

jonahpearl:
Aha! Thanks for the tip — I will take a look and see if I can modify it for my use case.

jonahpearl:
[...]

PR #3522, opened 2024-11-06
Ibl exporter

jonahpearl:
Function for #2843
Would be great if a few others could test on their data. Thanks!


zm711:
This looks like it is off of the bug fixes branch. @jonahpearl you want to switch this to main instead?

.github/workflows/core-test.yml
```diff
@@ -5,6 +5,7 @@ on:
     types: [synchronize, opened, reopened]
     branches:
       - main
+      - 0.100-bug-fixes
```
alejoe91, line 8:
```suggestion
```

alejoe91:
@jonahpearl thanks for this! Is it ok if I push to this PR?

I've been working on refining the export function so I'd be good to be in sync :)

alejoe91:
> This looks like it is off of the bug fixes branch. @jonahpearl you want to switch this to main instead?

I think it's correctly using the `SortingAnalyzer`. Maybe the added branch in the actions yaml was confusing, so let's just remove it

alejoe91 (changes requested):

src/spikeinterface/postprocessing/tests/test_template_similarity.py
```diff
@@ -25,6 +25,12 @@
 
 SKIP_NUMBA = pytest.mark.skipif(not HAVE_NUMBA, reason="Numba not available")
 
+from spikeinterface.core import extract_waveforms
+from spikeinterface.extractors import toy_example
+from spikeinterface.comparison import compare_templates
+
+import numpy as np
```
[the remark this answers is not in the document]
jonahpearl: Yes, not sure how this slipped in.

src/spikeinterface/exporters/to_phy.py
```diff
@@ -194,7 +194,7 @@ def export_to_phy(
         templates[unit_ind, :, :][:, : len(chan_inds)] = template
         templates_ind[unit_ind, : len(chan_inds)] = chan_inds
 
-    if not sorting_analyzer.has_extension("template_similarity"):
+    if sorting_analyzer.get_extension("template_similarity") is None:
```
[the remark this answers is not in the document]
jonahpearl: You're suggesting we put this back to "has_extension"? Any reason this one and not the other ones? The ibl export relies on doing a Phy export, and as discussed elsewhere this "has_extension" pattern is no longer safe if the folder exists but without any data inside.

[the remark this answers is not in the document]
alejoe91: yes, all of them. I think it's much easier to read :)

src/spikeinterface/exporters/to_ibl.py
```diff
@@ -0,0 +1,305 @@
+from __future__ import annotations
+
+import os
+import shutil
+from pathlib import Path
+from typing import Optional
+
+import numpy as np
+import numpy.typing as npt
+from scipy.signal import welch
```
alejoe91, line 10: `import scipy` should be done within the functions using it because it's not a mandatory requirement (see failing tests)

src/spikeinterface/exporters/to_ibl.py
```diff
@@ -0,0 +1,309 @@
...
+        If True and "output_folder" exists, it is removed and overwritten
+    template_mode: str, default: "median"
+        Parameter "mode" to be given to WaveformExtractor.get_template()
+    dtype: dtype or None, default: None
+        Dtype to save binary data
+    verbose: bool, default: True
+        If True, output is verbose
+    use_relative_path : bool, default: False
+        If True and `copy_binary=True` saves the binary file `dat_path` in the `params.py` relative to `output_folder` (ie `dat_path=r"recording.dat"`). If `copy_binary=False`, then uses a path relative to the `output_folder`
+        If False, uses an absolute path in the `params.py` (ie `dat_path=r"path/to/the/recording.dat"`)
```
zm711, line 79:
```suggestion
    ----------
    analyzer : SortingAnalyzer
        The sorting analyzer object.
    output_folder : str | Path
        The output folder where the phy template-gui files are saved
    rms_win_length_sec : float, default: 3
        The window length in seconds for the RMS calculation.
    welch_win_length_samples : int, default: 1024
        The window length in samples for the Welch method.
    total_secs : int, default: 100
        The total number of seconds to use for the spectral density calculation.
    only_ibl_specific_steps : bool, default: False
        If True, only the IBL specific steps are run (i.e. skips calling `export_to_phy`)
    compute_pc_features : bool, default: False
        If True, pc features are computed
    compute_amplitudes : bool, default: True
        If True, waveforms amplitudes are computed
    sparsity : ChannelSparsity or None, default: None
        The sparsity object (currently only respected for phy part of the export)
    copy_binary : bool, default: True
        If True, the recording is copied and saved in the phy "output_folder"
    remove_if_exists : bool, default: False
        If True and "output_folder" exists, it is removed and overwritten
    template_mode : str, default: "median"
        Parameter "mode" to be given to WaveformExtractor.get_template()
    dtype : dtype or None, default: None
        Dtype to save binary data
    verbose : bool, default: True
        If True, output is verbose
    use_relative_path : bool, default: False
        If True and `copy_binary=True` saves the binary file `dat_path` in the `params.py` relative to `output_folder` (ie `dat_path=r"recording.dat"`). If `copy_binary=False`, then uses a path relative to the `output_folder`
        If False, uses an absolute path in the `params.py` (ie `dat_path=r"path/to/the/recording.dat"`)
```

We are working on implementing numpydoc linting in the future and it requires there to be a space after the parameter.

src/spikeinterface/exporters/to_ibl.py
```diff
@@ -0,0 +1,309 @@
...
+    {}
+
+    """
+
+    try:
+        from scipy.signal import welch
+    except ImportError as e:
+        raise ImportError("Please install scipy to use the export_to_ibl function.") from e
```
zm711, line 87: I've never seen this before. How does this raise error from e work?

zm711 (commented):
Just a docstring fix while you're at it. I believe we add this to the dev docs after this PR was started which is why this was missed.

src/spikeinterface/exporters/to_ibl.py
alejoe91: I would just import it in the function. The problem is that core tests will look in all files for tests, and script is not installed in core tests

src/spikeinterface/exporters/to_phy.py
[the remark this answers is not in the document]
alejoe91: That's a separate issue :) we should delay folder creation until dependencies are checked. Can you open another issue about it?

src/spikeinterface/exporters/to_ibl.py
```diff
@@ -0,0 +1,309 @@
...
+
+    # the window generator will generates window indices
+    wingen = WindowGenerator(ns=total_samples_ap, nswin=rms_win_length_samples_ap, overlap=0)
+    win = {
+        "TRMS": np.zeros((wingen.nwin, analyzer.recording.get_num_channels())),
+        "nsamples": np.zeros((wingen.nwin,)),
+        "fscale": fscale(welch_win_length_samples, 1 / fs_ap, one_sided=True),
+        "tscale": wingen.tscale(fs=fs_ap),
+    }
+    win["spectral_density"] = np.zeros((len(win["fscale"]), analyzer.recording.get_num_channels()))
```
samuelgarcia, line 170: I think we should have a proper function to compute the spectrum if you need it.
So we could have proper tests for this.

src/spikeinterface/exporters/to_ibl.py
```diff
@@ -0,0 +1,309 @@
...
+    # Now we need to add the extra IBL specific files
+    (channel_inds,) = np.isin(analyzer.recording.channel_ids, analyzer.channel_ids).nonzero()
+
+    ### Run spectral density and rms ###
+    fs_ap = analyzer.recording.sampling_frequency
+    rms_win_length_samples_ap = 2 ** np.ceil(np.log2(fs_ap * rms_win_length_sec))
+    total_samples_ap = int(np.min([fs_ap * total_secs, analyzer.recording.get_num_samples()]))
+
+    # the window generator will generates window indices
+    wingen = WindowGenerator(ns=total_samples_ap, nswin=rms_win_length_samples_ap, overlap=0)
```
samuelgarcia, line 163: we have already some function to generate window in spikeinterface.

samuelgarcia:
Hi @jonahpearl
thanks for this contribution to spikeinetrface and making a bridge to ibl tools.

About   `to_ibl_utils.py` I would identify function by function what the need and if there is an equivalent function in spikeinterface for it. Then I would implement then at the good place (core, preprocessing, postprocessing, ...) and also would make some tests function per function.

About the spectrum, I would implement it in a speparate file using the spikeinterface machinery on chunk parralelisation.

jonahpearl:
Fair enough -- I don't have time to refactor this right now but I can work on it in the coming months.

PR #3527, opened 2024-11-08
Removing useless dependencies

yger:
Some extensions computed by the sorting analyzer should relax their dependencies, and do not need recording, as far as I am aware. Looks like template_similarity and template_metrics, once you have the templates, could be recomputed without the recording


alejoe91 (approved):
LGTM, not sure why we had that flag

PR #3533, opened 2024-11-13
Clustering components api

yger:
In order to harmonize the API for the components, job_kwargs is now handled properly at the clustering level


yger:
@samuelgarcia Any ideas why the tests are failing by timing out? Can I know more about it? The changes made in this PR are not drastic, and tests are locally passed on my machine (for internals sorters, the ones that might be affected by changes)

PR #3536, opened 2024-11-13
Load peaks from kilosort output

JoeZiminski:
[In progress]


chrishalcrow:
Hey Joe, I think the existence #4202 means we can close this. Let me know if we should reopen :)

PR #3548, opened 2024-11-21
Add `stream_mode` as extra_requirements for NWB when streaming

### LINKED ISSUE #3540: Missing requirements in docker container for spike sorting with kilosort3
Two requirements are missing to run kilosort3 container "remfile" and "pynwb"-

e.g., The code below produces ModuleNotFoundError for these two dependencies: 

```
# spike sort with kilosort 3 from container (uses apptainer)
sorting_KS3 = ss.run_sorter(sorter_name="kilosort3", singularity_image=True, recording=recording)
```

Currently the container runs successfully by adding the extra_requirements argument:

```
# spike sort with kilosort 3 from container (uses apptainer)
sorting_KS3 = ss.run_sorter(sorter_name="kilosort3", singularity_image=True, recording=recording, extra_requirements=["remfile","pynwb"])
```

alejoe91:
Fixes #3540

### DIFF
--- a/src/spikeinterface/extractors/nwbextractors.py
+++ b/src/spikeinterface/extractors/nwbextractors.py
@@ -599,6 +599,8 @@ class NwbRecordingExtractor(BaseRecording, _BaseNWBExtractor):
         else:
             gains, offsets, locations, groups = self._fetch_main_properties_backend()
             self.extra_requirements.append("h5py")
+        if stream_mode is not None:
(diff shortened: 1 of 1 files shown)


h-mayorquin (approved):
LGTM. 

I am curios on what the performance will be with this. 

This will throw an error if someone tries to use ros3 I think but we are getting rid of that.

PR #3559, opened 2024-12-02
Hard code `synchony_sizes`

### LINKED ISSUE #3549: fix handling of synchrony columns in qualitymetrics
@chrishalcrow

I'm sorry I missed this during review, but we need to find a better way of dealing with synchrony quality metrics. I just discovered that we have the names `sync_spike_2` etc hardcoded, but those are named based on user input with 2,4,8 or 2,6,8 being the defaults. So we should think about a more robust way to deal with that. Not sure how, but I discovered this in #3497 and added a note in comment for us to think on it.

chrishalcrow:
Fix #3549

Currently there is one metric, `synchony`, whose column names depends on a kwarg: `synchony_sizes`. If the user inputs `synchony_sizes = [3,5,7]` the columns in the `quality_metrics` dataframe are `sync_spikes_3`, `sync_spikes_5` and `sync_spikes_7`. This is a bit awkward to deal with in a few places. Computing these custom `synchrony_sizes` are a bit of an edge case, so me @alejoe91 and @zm711 decided to just hard code the default values of `synchony_sizes = [2,4,8]`.

To help users who _really_ want to compute custom synchony sizes, I've left the `_compute_synchonry_counts` function (which is called by the user-facing `compute_synchrony_metrics` function) with the ability to do this, but made it private. 

No backwards compatibility issues spring to mind: using this PR you can still load an old `sorting_analyzer` with `sync_spikes_3` columns. Also, it shouldn't be breaking: if you pass the `synchony_sizes` kwarg to `compute_synchrony_metrics` it still computes, just with a warning.

### DIFF
--- a/src/spikeinterface/qualitymetrics/misc_metrics.py
+++ b/src/spikeinterface/qualitymetrics/misc_metrics.py
@@ -520,7 +520,7 @@ _default_params["sliding_rp_violation"] = dict(
 )
 
 
-def get_synchrony_counts(spikes, synchrony_sizes, all_unit_ids):
+def _get_synchrony_counts(spikes, all_unit_ids, synchrony_sizes=np.array([2, 4, 8])):
     """
     Compute synchrony counts, the number of simultaneous spikes with sizes `synchrony_sizes`.
 
@@ -528,10 +528,10 @@ def get_synchrony_counts(spikes, synchrony_sizes, all_unit_ids):
     ----------
     spikes : np.array
         Structured numpy array with fields ("sample_index", "unit_index", "segment_index").
-    synchrony_sizes : numpy array
-        The synchrony sizes to compute. Should be pre-sorted.
     all_unit_ids : list or None, default: None
         List of unit ids to compute the synchrony metrics. Expecting all units.
+    synchrony_sizes : numpy array
+        The synchrony sizes to compute. Should be pre-sorted.
 
     Returns
     -------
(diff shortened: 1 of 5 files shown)


src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -565,37 +565,36 @@ def get_synchrony_counts(spikes, synchrony_sizes, all_unit_ids):
...
     """
-    assert min(synchrony_sizes) > 1, "Synchrony sizes must be greater than 1"
-    # Sort the synchrony times so we can slice numpy arrays, instead of using dicts
-    synchrony_sizes_np = np.array(synchrony_sizes, dtype=np.int16)
-    synchrony_sizes_np.sort()
 
-    res = namedtuple("synchrony_metrics", [f"sync_spike_{size}" for size in synchrony_sizes_np])
+    if synchrony_sizes is not None:
+        warning_message = "Custom `synchrony_sizes` is deprecated; the `synchrony_metrics` will be computed using `synchrony_sizes = [2,4,8]`"
+        warnings.warn(warning_message)
```
alejoe91, line 593: we should use `DeprecationWarning` and stacklevel=2 here (see other deprecation warnings in code base :) )

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -520,18 +520,18 @@ def compute_sliding_rp_violations(
 )
 
 
-def get_synchrony_counts(spikes, synchrony_sizes, all_unit_ids):
```
zm711, line 523: This does change the order in order to have the default. I guess I don't see a way around this, so it's okay.

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -520,18 +520,18 @@ def compute_sliding_rp_violations(
...
 
     Parameters
     ----------
     spikes : np.array
         Structured numpy array with fields ("sample_index", "unit_index", "segment_index").
-    synchrony_sizes : numpy array
-        The synchrony sizes to compute. Should be pre-sorted.
     all_unit_ids : list or None, default: None
         List of unit ids to compute the synchrony metrics. Expecting all units.
+    synchrony_sizes : numpy array
```
zm711, line 533:
```suggestion
    synchrony_sizes : numpy array, default: np.array([2,4,8])
```
something like this right?

zm711 (commented):
Couple requests and one discussion point along with Alessio's comment. :)

src/spikeinterface/qualitymetrics/misc_metrics.py
chrishalcrow: Yeah, I think you're right. The function is only used in `compute_synchrony_metrics`, and this allows people to use `get_synchrony_counts` with the `synchrony_sizes` in another context and keeps the order in tact. Nice.

zm711 (approved):
This is good by me now!

PR #3570, opened 2024-12-05
Correct method default in docstring

### LINKED ISSUE #3566: Default `plot_unit_locations` method is not what is documented
Hello,

Our docstring for `compute_unit_locations` says:
 ```
method : "center_of_mass" | "monopolar_triangulation" | "grid_convolution", default: "center_of_mass"
```

But I think the default is actually `monopolar_triangulation`, as set here https://github.com/SpikeInterface/spikeinterface/blob/853d8a48abcf40f5a47401ed2f7baa83e1c3295f/src/spikeinterface/postprocessing/unit_locations.py#L56

I think we want to use `center_of_mass` by default, as is done for `compute_spike_locations`. @samuelgarcia @alejoe91 can you confirm?

chrishalcrow:
Fixes #3566

Correct the default `method` docstring in `compute_unit_locations`. Was `center_of_mass` but `monopolar_triangulation` is actually the default.

### DIFF
--- a/src/spikeinterface/postprocessing/unit_locations.py
+++ b/src/spikeinterface/postprocessing/unit_locations.py
@@ -26,7 +26,7 @@ class ComputeUnitLocations(AnalyzerExtension):
     ----------
     sorting_analyzer : SortingAnalyzer
         A SortingAnalyzer object
-    method : "center_of_mass" | "monopolar_triangulation" | "grid_convolution", default: "center_of_mass"
+    method : "monopolar_triangulation" or  "center_of_mass" or "grid_convolution", default: "monopolar_triangulation"
         The method to use for localization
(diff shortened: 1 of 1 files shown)


zm711 (commented):
why `or` instead of `|`? We had decided on doing `|` in general, but have never enforced it. Do you want to make an argument for using or instead. The `|` in type hints is 3.10+, but in docstrings there are no rules.

chrishalcrow:
> why `or` instead of `|`? We had decided on doing `|` in general, but have never enforced it. Do you want to make an argument for using or instead. The `|` in type hints is 3.10+, but in docstrings there are no rules.

Oh, crud. Yes indeed, that's in our docstring conventions - my mistake!

zm711 (approved):
good by me.

PR #3580, opened 2024-12-13
Adding SharedMemoryTemplates

yger:
For template matching engines, it might be good to have some datastructures that would be shared in memory. For example the Templates (with the current PR), and maybe also the spatio-temporal overlaps (might be sorter specific, not sure we need a generic object for that)


src/spikeinterface/core/template.py
```diff
@@ -6,6 +6,8 @@
 from probeinterface import Probe
 from pathlib import Path
 from .sparsity import ChannelSparsity
+from multiprocessing.shared_memory import SharedMemory
+from .core_tools import make_shared_array
```
alejoe91, line 10:
```suggestion
from multiprocessing.shared_memory import SharedMemory

from .sparsity import ChannelSparsity
from .core_tools import make_shared_array
```

alejoe91 (approved):

PR #3587, opened 2024-12-17
Add `NwbTimeSeriesExtractor` to extract TimeSeries data from NWB

h-mayorquin:
Some users would like to access data that is not an ElectricalSeries to analyze it with SpikeInterface. TimeSeries data can be loaded lazily to Spikeinterface but it does not have all the ElectricalSeries infrastructure like electrodes so it can not be loaded by the current extractor. This PR adds another extractor that can be used to load TimeSeries data.

@steevelaquitaine  @borrepp

### DIFF
--- a/src/spikeinterface/extractors/nwbextractors.py
+++ b/src/spikeinterface/extractors/nwbextractors.py
@@ -585,6 +585,7 @@ class NwbRecordingExtractor(BaseRecording, _BaseNWBExtractor):
                 segment_data,
                 times_kwargs,
             ) = self._fetch_recording_segment_info_backend(file, cache, load_time_vector, samples_for_rate_estimation)
+
         BaseRecording.__init__(self, channel_ids=channel_ids, sampling_frequency=sampling_frequency, dtype=dtype)
         recording_segment = NwbRecordingSegment(
             electrical_series_data=segment_data,
@@ -595,12 +596,9 @@ class NwbRecordingExtractor(BaseRecording, _BaseNWBExtractor):
         # fetch and add main recording properties
         if use_pynwb:
             gains, offsets, locations, groups = self._fetch_main_properties_pynwb()
-            self.extra_requirements.append("pynwb")
         else:
             gains, offsets, locations, groups = self._fetch_main_properties_backend()
-            self.extra_requirements.append("h5py")
-        if stream_mode is not None:
-            self.extra_requirements.append(stream_mode)
+
         self.set_channel_gains(gains)
         self.set_channel_offsets(offsets)
         if locations is not None:
@@ -658,6 +656,19 @@ class NwbRecordingExtractor(BaseRecording, _BaseNWBExtractor):
             "stream_cache_path": stream_cache_path,
             "file": file,
         }
+
+        # Set extra requirements for the extractor, so they can be installed from docker
+        if use_pynwb:
+            self.extra_requirements.append("pynwb")
+        else:
+            if self.backend == "hdf5":
+                self.extra_requirements.append("h5py")
+            if self.backend == "zarr":
+                self.extra_requirements.append("zarr")
+        if self.stream_mode == "fsspec":
+            self.extra_requirements.append("fsspec")
+        if self.stream_mode == "remfile":
+            self.extra_requirements.append("remfile")
 
     def _fetch_recording_segment_info_pynwb(self, file, cache, load_time_vector, samples_for_rate_estimation):
         self._nwbfile = read_nwbfile(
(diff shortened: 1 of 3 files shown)


src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -659,6 +657,19 @@ def __init__(
             "file": file,
         }
 
+        # Set extra requirements for the extractor, so they can be installed from docker
```
zm711, line 660:
```suggestion
        # Set extra requirements for the extractor, so they can be installed when using docker
```
or on. from would be you get the package from docker rather than install into. So you say into docker. but I think when using is probably best :)

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1364,8 +1375,371 @@ def get_unit_spike_train(
         return frames[start_index:end_index].astype("int64", copy=False)
 
 
+def _find_timeseries_from_backend(group, path="", result=None, backend="hdf5"):
+    """
+    Recursively searches for groups with TimeSeries neurodata_type in hdf5 or zarr object,
+    and returns a list with their paths.
+    """
+    if backend == "hdf5":
+        import h5py
```
zm711, line 1384: I don't know how big h5py is or zarr for that matter. Is there any benefit to just import the exact thing we need (ie import h5py.Group rather than the full package?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1364,8 +1375,371 @@ def get_unit_spike_train(
...
+        if not using the `file` parameter.
+    timeseries_path : str | None
+        The path to the TimeSeries object within the NWB file. This parameter is required
+        when the NWB file contains multiple TimeSeries objects. The path corresponds to
+        the location within the NWB file hierarchy, e.g. 'acquisition/MyTimeSeries'.
+    load_time_vector : bool, default: False
+        If True, the time vector is loaded into the recording object. Useful when
+        precise timing information is needed.
+    samples_for_rate_estimation : int, default: 1000
+        The number of timestamp samples used for estimating the sampling rate when
```
zm711, line 1420: I'm trying to understand the writing. This would just be the number of timestamps right? I'm not quite sure what it would mean to say timestamp samples? But I don't know this format so I could be totally off on this :)

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1364,8 +1375,371 @@ def get_unit_spike_train(
...
+        timestamps are used instead of a fixed rate.
+    stream_mode : Literal["fsspec", "remfile", "zarr"] | None, default: None
+        Determines the streaming mode for reading the file.
+    file : BinaryIO | None, default: None
+        A file-like object representing the NWB file. Use this parameter if you have
+        an in-memory representation of the NWB file instead of a file path.
```
zm711, line 1426:
```suggestion
        an in-memory representation of the NWB file instead of a file path given by `file_path`.
```
Maybe do a call back to the actual argument?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1364,8 +1375,371 @@ def get_unit_spike_train(
...
+    cache : bool, default: False
+        If True, the file is cached locally when using streaming.
+    stream_cache_path : str | Path | None, default: None
+        Local path for caching. Only used if cache is True.
```
zm711, line 1430:
```suggestion
        Local path for caching. Only used if `cache` is True.
```
we are trying to have the arguments marked out moving forward.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1364,8 +1375,371 @@ def get_unit_spike_train(
...
+        recording_segment = NwbTimeSeriesSegment(
+            timeseries_data=segment_data,
+            times_kwargs=times_kwargs,
+        )
+        self.add_recording_segment(recording_segment)
+
+        if storage_options is not None and stream_mode == "zarr":
+            warnings.warn(
+                "The `storage_options` parameter will not be propagated to JSON or pickle files "
+                "for security reasons, so the extractor will not be JSON/pickle serializable."
```
zm711, line 1500:
```suggestion
                "for security reasons, so this extractor will not be JSON/pickle serializable."
```
the is not wrong, but this emphasizes that it the current version of the extractor. Up to you if you want that emphasis.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1364,8 +1375,371 @@ def get_unit_spike_train(
...
+            self.extra_requirements.append("pynwb")
+        else:
+            if self.backend == "hdf5":
+                self.extra_requirements.append("h5py")
+            if self.backend == "zarr":
+                self.extra_requirements.append("zarr")
+        if self.stream_mode == "fsspec":
+            self.extra_requirements.append("fsspec")
+        if self.stream_mode == "remfile":
+            self.extra_requirements.append("remfile")
```
zm711, line 1527: Would this be worth changing to a if-elif-else where you add an error if they use the wrong streammode? Or is the concern that the stream_mode can be None if not streaming? I'm just trying to understand why there are two if statements which seem mutually exclusive?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1364,8 +1375,371 @@ def get_unit_spike_train(
...
+        if self.timeseries_path is not None:
+            if self.timeseries_path not in time_series_dict:
+                raise ValueError(f"TimeSeries {self.timeseries_path} not found in file")
+
+        else:
+            if len(time_series_dict) == 1:
+                self.timeseries_path = list(time_series_dict.keys())[0]
+            else:
+                raise ValueError(
+                    f"Multiple TimeSeries found! Specify 'timeseries_path'. Options: {list(time_series_dict.keys())}"
```
zm711, line 1557: :)

zm711 (commented):
Just a couple cosmetic things.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Yeah, confusing, the thing is that the object/data container is called timestamps. That time series (timestamps) has samples.

Let me think this through.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Good idea.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: OK. Sorry, I keep forgeting this.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Makes sense. Thanks.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: It could be better, yes. At least for the streams.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: I welcome any suggestion though.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: I changed it.

src/spikeinterface/extractors/nwbextractors.py
zm711: I think I would need to see an example. Maybe I'll look at the test and see if I can better understand how this works. If it is confusing naming then we just have to live with it.

src/spikeinterface/extractors/nwbextractors.py
zm711: I honestly just didn't know the size of these packages. Something like scipy would be so slow and heavy for this kind of check, but if these are rather small/fast (which has been my experience playing with h5py at least) then it doesn't matter.

alejoe91 (approved):

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1364,8 +1378,372 @@ def get_unit_spike_train(
...
+        if not using the `file` parameter.
+    timeseries_path : str | None
+        The path to the TimeSeries object within the NWB file. This parameter is required
+        when the NWB file contains multiple TimeSeries objects. The path corresponds to
+        the location within the NWB file hierarchy, e.g. 'acquisition/MyTimeSeries'.
+    load_time_vector : bool, default: False
+        If True, the time vector is loaded into the recording object. Useful when
+        precise timing information is needed.
+    samples_for_rate_estimation : int, default: 1000
+        The number of timestamp samples used for estimating the sampling rate when
```
alejoe91, line 1423:
```suggestion
        The number of timestamps used for estimating the sampling rate when
```
I think this is clear :)

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Yeah, what I mean is that usually both of them are equivalent because importing one attribute imports the full module in most python packages. Anyway, here are the measurements:

```python
@h-laptop$ conda activate work
(work) @h-laptop$ python -m timeit -s "import h5py" 
50000000 loops, best of 5: 5.43 nsec per loop
(work) @h-laptop$ python -m timeit -s "from h5py import Group" 
50000000 loops, best of 5: 6.33 nsec per loop
(work) @h-laptop$ python -m timeit -s "from zarr import Group" 
50000000 loops, best of 5: 5.39 nsec per loop
(work) @h-laptop$ python -m timeit -s "import zarr" 
50000000 loops, best of 5: 5.31 nsec per loop
```

h-mayorquin:
Can I merge this? anyone else comments?

alejoe91:
LGTM

PR #3588, opened 2024-12-17
Use strings as ids in generators

h-mayorquin:
As discussed in the maintenance meeting of 2024/12/17.

Minor change but let's see what the testing says.

### DIFF
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -134,7 +134,7 @@ def generate_sorting(
     seed = _ensure_seed(seed)
     rng = np.random.default_rng(seed)
     num_segments = len(durations)
-    unit_ids = np.arange(num_units)
+    unit_ids = [str(id) for id in np.arange(num_units)]
 
     spikes = []
     for segment_index in range(num_segments):
@@ -1111,7 +1111,7 @@ class SortingGenerator(BaseSorting):
 
         """
 
-        unit_ids = np.arange(num_units)
+        unit_ids = [str(id) for id in np.arange(num_units)]
         super().__init__(sampling_frequency, unit_ids)
 
         self.num_units = num_units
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/generate.py
```diff
@@ -1111,7 +1111,7 @@ def __init__(
 
         """
 
-        unit_ids = np.arange(num_units)
+        unit_ids = [str(id) for id in np.arange(num_units)]
```
alejoe91, line 1114:
```suggestion
        unit_ids = [str(idx) for idx in np.arange(num_units)]
```

src/spikeinterface/core/generate.py
```diff
@@ -134,7 +134,7 @@ def generate_sorting(
     seed = _ensure_seed(seed)
     rng = np.random.default_rng(seed)
     num_segments = len(durations)
-    unit_ids = np.arange(num_units)
+    unit_ids = [str(id) for id in np.arange(num_units)]
```
alejoe91, line 137:
```suggestion
    unit_ids = [str(idx) for idx in np.arange(num_units)]
```

alejoe91 (changes requested):
@h-mayorquin small comment to avoid using `id` as a var name

src/spikeinterface/core/generate.py
h-mayorquin: Thanks.  This makes sense.

src/spikeinterface/core/generate.py
```diff
@@ -1177,7 +1179,8 @@ def __init__(
             self.refractory_period_seconds = np.full(num_units, self.refractory_period_seconds, dtype="float64")
 
         self.segment_seed = seed
-        self.units_seed = {unit_id: self.segment_seed + hash(unit_id) for unit_id in range(num_units)}
+        self.units_seed = {unit_id: abs(self.segment_seed + hash(unit_id)) for unit_id in unit_ids}
```
samuelgarcia, line 1182:
```suggestion
        self.units_seed = {unit_id: self.segment_seed + abs(hash(unit_id)) for unit_id in unit_ids}
```

src/spikeinterface/core/generate.py
```diff
@@ -1280,7 +1283,7 @@ def __init__(
         noise_block_size: int = 30000,
     ):
 
-        channel_ids = np.arange(num_channels)
+        channel_ids = [str(id) for id in np.arange(num_channels)]
```
alejoe91, line 1286:
```suggestion
        channel_ids = [str(idx) for idx in np.arange(num_channels)]
```

alejoe91 (approved):

PR #3598, opened 2025-01-07
Update zarr version

alejoe91:
Zarr 2.18.0 used to trigger some errors. Let's see if latest releases fixed the issues.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = [
     "numpy>=1.20, <2.0",  # 1.20 np.ptp, 1.26 might be necessary for avoiding pickling errors when numpy >2.0
     "threadpoolctl>=3.0.0",
     "tqdm",
-    "zarr>=2.16,<2.18",
+    "zarr>=2.18.4,<3",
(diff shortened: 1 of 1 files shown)


alejoe91:
@h-mayorquin 

seems like the most recent hdmf-zarr/pynwb versions makes the `use_pynwb` option to fail. Not sure why honestly. The zarr files seem to be ok, and in fact when using zarr directly (`use_pynwb=False`) all tests pass.

Can you help me look into this? One option could be to simply enforce  `use_pynwb=False` when a `zarr` backend is detected.

zm711:
Maybe related to the issue at neo. pynwb calling hdmf caused our tests to fail due to changes at the hdmf level in allowable symbols in strings. They started forbidding `:` and `\` maybe due to path parsing?

h-mayorquin:
Yeah, I can take a look.

alejoe91:
hdmf-issue has been fixed in 0.11.0. I think we can merge this! :)

samuelgarcia (approved):

PR #3607, opened 2025-01-10
Add a fast correlogram merge

chrishalcrow:
Have added a fast method to re-compute correlograms when units are merged.

When two units are merged, their correlograms with other units become the sum of the previous correlograms. Let $C_{i,j}$ be the cross-correlogram of units $i$ and $j$. If we merge units 2, 4 and 6 into a new unit 10, the cross-correlograms of unit 10 with unit 1 would be:

$$C_{10,1} = C_{2,1} + C_{4,1} + C_{6,1}$$
$$C_{1,10} = C_{1,2} + C_{1,4} + C_{1,6}$$

Similarly, the auto-correlogram of the new unit 10 is

$$C_{10,10} = (C_{2,2} + C_{2,4} + C_{2,6}) + (C_{4,2} + C_{4,4} + C_{4,6}) + (C_{6,2} + C_{6,4} + C_{6,6})$$

You can implement this with two matrix operations (sum all merged-unit columns together, then  sum all merged-unit rows together), which I've implemented.

The most annoying thing is tracking the `new_unit_index`, and there's more code which deals with this than the actual new algorithm. It's much easier to deal with the `take_first` new_id_strategy than the `append` one. Any advice here is welcome.

If you don't believe my maths, I've added tests to show that the results are the same if you use the fast merging method or if we merge, then re-compute the correlograms. If you can think of more painful tests than I've written, let me know!

This is a big speedup for my kilosort'd NP2.0 sortings (on my laptop, 25s -> 0.2s for a single pairwise merge). And yup, I do have numba installed.

It only becomes a lot faster for generated recordings when the correlograms become a bit more full. This requires a high firing rate and small refractory period. Here's some benchmarking code:

``` python
from time import perf_counter

nums_units = [40,80,160,320,640]
times = []
for num_units in nums_units:
    
    rec, sort = si.generate_ground_truth_recording(seed=1205,num_units=num_units, generate_sorting_kwargs = {'firing_rates': 50, 'refractory_period_ms': 1.0})
    sa.compute('correlograms')
    
    t1 = perf_counter()
    sa.merge_units(merge_unit_groups=[['0','4']], sparsity_overlap=0.5)
    t2 = perf_counter()
    times.append(t2-t1)

old_times = dict(zip(nums_units, times))

plt.plot(nums_units, new_times.values())
plt.plot(nums_units, old_times.values())
plt.legend(['new method', 'old method'])
plt.xlabel('num units')
plt.ylabel('time (seconds)')
```

and I get:

``` python
print(old_times)
{40: 0.007854124996811152,
 80: 0.02739199995994568,
 160: 0.11776154092513025,
 320: 1.210061291931197,
 640: 6.01522733294405}

print(new_times)
{40: 0.0016532500740140676,
 80: 0.003977375105023384,
 160: 0.013842040905728936,
 320: 0.05750433378852904,
 640: 0.20545295905321836}
```

giving the following plot

Screenshot 2025-01-10 at 15 13 28


JoeZiminski:
cool!

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -93,9 +94,46 @@ def _select_extension_data(self, unit_ids):
     def _merge_extension_data(
         self, merge_unit_groups, new_unit_ids, new_sorting_analyzer, censor_ms=None, verbose=False, **job_kwargs
     ):
-        # recomputing correlogram is fast enough and much easier in this case
-        new_ccgs, new_bins = _compute_correlograms_on_sorting(new_sorting_analyzer.sorting, **self.params)
-        new_data = dict(ccgs=new_ccgs, bins=new_bins)
+
+        method = "keep_first"
```
zm711, line 98: I guess I get the point of these strings, but it is a bit confusing to me to have this hard-code. For me at least it would make more sense to have a bool something like

```python
need_to_append = False
```

Then you have the if statement check the boolean. The strings make me think that the user should be able to control this.

zm711 (commented):
This is cool. Someone else can check the math :P. Just one comment :) . And then I would request a little explanation of the strategy either as docstring or comment so we can look back and know why this works.

src/spikeinterface/postprocessing/correlograms.py
chrishalcrow: Sure!

chrishalcrow:
> This is cool. Someone else can check the math :P. Just one comment :) . And then I would request a little explanation of the strategy either as docstring or comment so we can look back and know why this works.

My hope is that the tests are robust enough that no one has to check the maths!

Docstring done

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -93,9 +94,54 @@ def _select_extension_data(self, unit_ids):
...
+
+            new_row = np.sum(correlograms[:, merge_unit_group_indices, :], axis=1)
+            correlograms[:, merge_unit_group_indices[0], :] = new_row
+            correlograms[:, merge_unit_group_indices[1:], :] = 0
+
+            if new_unit_id not in merge_unit_group:
+                need_to_append = True
+                delete_from = 0
+
+        if need_to_append is True:
```
zm711, line 127: Not that it really matters, but since `need_to_append` is truthy as bool you don't need to check with is. Just the fact that it is `True/False` will trigger or not the branch. :)

chrishalcrow: Hello, I prefer what I've done, just to be a bit more explicit. Just find truthiness a bit opaque - more so with lists and dicts than bools.

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -93,9 +94,54 @@ def _select_extension_data(self, unit_ids):
     def _merge_extension_data(
         self, merge_unit_groups, new_unit_ids, new_sorting_analyzer, censor_ms=None, verbose=False, **job_kwargs
     ):
-        # recomputing correlogram is fast enough and much easier in this case
-        new_ccgs, new_bins = _compute_correlograms_on_sorting(new_sorting_analyzer.sorting, **self.params)
-        new_data = dict(ccgs=new_ccgs, bins=new_bins)
```
alejoe91, line 98: @chrishalcrow I thinnk this doesn't handle the case when `censor_ms` is not None. In that case, individual spike trains from the merged units might be trimmed to remove the duplicates, so we cannot apply this new strategy (or am I missing something?)

What we could do is the following:
- if `censor_period` is None, use new method
- if not, check the len of the new merged spiketrains are the same as the sum of the pre-merged list. If so, use new method, else use old one

chrishalcrow: Hello, thanks, good catch! Because `censor_ms` removes exactly the spikes in the censor period, we can just set the correlogram to be equal to zero in the censor period. There's an edge case where `censor_ms` isn't on the boundary of a bin, but we don't guarantee `soft` mode to be exact, so I think it's fine to `ceil` censor_ms to the closest bin edge?

chrishalcrow: Implement @alejoe91 's suggesion!

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -93,9 +94,77 @@ def _select_extension_data(self, unit_ids):
...
+
+                old_spike_train_lengths = len(
+                    np.concatenate(
+                        [self.sorting_analyzer.sorting.get_unit_spike_train(unit_id) for unit_id in merge_unit_group]
+                    )
+                )
+
+                if merged_spike_train_length != old_spike_train_lengths:
+                    can_apply_soft_method = False
+                    break
```
alejoe91, line 121: fantastic!

alejoe91 (approved):

chrishalcrow:
I reckon this is good to go!

alejoe91:
@samuelgarcia wanted to check this

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -93,9 +94,77 @@ def _select_extension_data(self, unit_ids):
...
+            C_{l,k} = C_{l,k} + C_{l,j}
+        Here, we apply this formula to quickly compute correlograms for merged units.
+        """
+
+        can_apply_soft_method = True
+        if censor_ms is not None:
+            # if censor_ms has no effect, can apply "soft" method. Check if any spikes have been removed
+            for new_unit_id, merge_unit_group in zip(new_unit_ids, merge_unit_groups):
+
+                merged_spike_train_length = len(new_sorting_analyzer.sorting.get_unit_spike_train(new_unit_id))
```
samuelgarcia, line 111: Did you test this when you have several segments ?
I think this will not work.
The multi segment way of spikeinterface make the correlogram more tedious because need an extra loop.

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -93,9 +94,77 @@ def _select_extension_data(self, unit_ids):
...
+
+                old_spike_train_lengths = len(
+                    np.concatenate(
+                        [self.sorting_analyzer.sorting.get_unit_spike_train(unit_id) for unit_id in merge_unit_group]
```
samuelgarcia, line 115: same here

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -93,9 +94,77 @@ def _select_extension_data(self, unit_ids):
...
+            if need_to_append is True:
+                old_num_units = np.shape(correlograms)[0]
+                correlograms = np.pad(correlograms, ((0, len(new_unit_ids)), (0, len(new_unit_ids)), (0, 0)))
+                for a, (new_unit_id, merge_unit_group) in enumerate(zip(new_unit_ids, merge_unit_groups)):
+
+                    old_loc = self.sorting_analyzer.sorting.ids_to_indices([merge_unit_group[0]])[0]
+                    new_loc = old_num_units + a
+
+                    correlograms[:, [old_loc, new_loc], :] = correlograms[:, [new_loc, old_loc], :]
+                    correlograms[[old_loc, new_loc], :, :] = correlograms[[new_loc, old_loc], :, :]
```
samuelgarcia, line 159: Is it working if some units have a id and some other have the same id ?

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -93,9 +94,77 @@ def _select_extension_data(self, unit_ids):
...
+
+        if can_apply_soft_method is False:
+            new_ccgs, new_bins = _compute_correlograms_on_sorting(new_sorting_analyzer.sorting, **self.params)
+            new_data = dict(ccgs=new_ccgs, bins=new_bins)
+        else:
+
+            need_to_append = False
+            delete_from = 1
+
+            correlograms, new_bins = deepcopy(self.get_data())
```
samuelgarcia, line 131: The "inplace" inplementation make the read a bit difficult for my slow brain.
Maybe having 2 matrics for old and new correlogram would have make the code more easy.
But this is OK for me. Justa comment.

samuelgarcia:
Hi Chris. Thanks a lot for this, you are a true heroe.
I did a comment for the multi segment handling.
Maybe we could have a tests for it.

The idea is really great, the implementation is a bit hard to follow with all the numpy tricks.

src/spikeinterface/postprocessing/correlograms.py
chrishalcrow: Thanks, added the segment loop and a test to make sure it work

src/spikeinterface/postprocessing/correlograms.py
chrishalcrow: The algorithm is quite "inplace", because you do the sum on the rows, _then_ the sum on the columns (or vice-versa).

I've added some comments which should help anyone reading this.

src/spikeinterface/postprocessing/correlograms.py
chrishalcrow: I don't understand. Your `sorting_analyzer` can't have units with the same ids, right?

chrishalcrow:
> Hi Chris. Thanks a lot for this, you are a true heroe. 
I'm doing it for my own evil intentions, don't worry

> I did a comment for the multi segment handling. Maybe we could have a tests for it.
Fixed!

> The idea is really great, the implementation is a bit hard to follow with all the numpy tricks.
Comments added!

src/spikeinterface/postprocessing/correlograms.py
alejoe91: I think Sam means if the merged_unit_id is chosen to be one of the merged unit ids (and I think we have that option)

chrishalcrow: We are too kind

chrishalcrow: That was harder than I thought! I've refactored things a bit. The function now constructs a change-of-unit_index matrix which maps the old unit_indices to the new unit_indices. You compute all the correlograms using the old matrices (in the "old unit index basis") then apply a change of basis:

$$ C \to TCT^T $$

to get the data in the "new unit index basis" (who remembers their undergrad linear algebra??). The nice thing about this: the same idea works for all methods, so the awkward `if` chains are confined to one bit of code. In the future, we should probably take this logic out of this function: making the transformation matrix is probably super helpful elsewhere!

Added a new test to make sure all new_id_strategies work.

chrishalcrow:
Wait! Don't merge! `einsum` is super slow. Will replace...

chrishalcrow:
Ok, done! New method is 3x slower than the original implementation - might try an optimise another time. Still 10x faster if you have lots of units!

```
old: {40: 0.007347875274717808, 80: 0.02676966693252325, 160: 0.11491125030443072, 320: 1.1197055419906974, 640: 5.708826333284378}
new: {40: 0.002904416993260383, 80: 0.01057270821183919, 160: 0.03658070880919695, 320: 0.145080625079572, 640: 0.634408749639988}
```

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -93,9 +94,98 @@ def _select_extension_data(self, unit_ids):
...
+                for merge_unit_group_index in merge_unit_group_indices:
+                    correlograms[:, merge_unit_group_index, :] = new_row
+
+            new_correlograms = np.zeros(
+                (len(new_sorting_analyzer.unit_ids), len(new_sorting_analyzer.unit_ids), correlograms.shape[2])
+            )
+            for old_index_1, new_index_1 in old_to_new_unit_index_map.items():
+                for old_index_2, new_index_2 in old_to_new_unit_index_map.items():
+                    new_correlograms[new_index_1, new_index_2, :] = correlograms[old_index_1, old_index_2, :]
+                    new_correlograms[new_index_2, new_index_1, :] = correlograms[old_index_2, old_index_1, :]
```
samuelgarcia, line 186: Excelent! This is super clear for me!!!!!

samuelgarcia: But is slower ?

alejoe91 (approved):

src/spikeinterface/postprocessing/tests/test_extension_merges.py
```diff
@@ -0,0 +1,61 @@
...
+    sorting_analyzer.compute("correlograms")
+
+    trial_merges = [
+        [["1", "2"]],
+        [["2", "4", "6", "8"]],
+        [["1", "4", "7"], ["2", "8"]],
+        [["4", "1", "8"], ["2", "7", "0"], ["3", "9"], ["5", "6"]],
+    ]
+
+    new_unit_ids = [["2"], ["4"], ["4", "2"], ["1", "2", "9", "100"]]
```
samuelgarcia, line 25: Crazy case but good to test! :)

samuelgarcia (approved):

src/spikeinterface/postprocessing/tests/test_extension_merges.py
chrishalcrow: Haha

src/spikeinterface/postprocessing/correlograms.py
chrishalcrow: I think the whole new method is easier to understand - using the `old_to_new_unit_index_map` is nice but it doesn't lend itself to numpy. This bit is slow, but I couldn't figure out how to `numpy-ifty` it. Probably easiest to `numba-fy` the whole thing later.

chrishalcrow:
Very happy to merge this - I might have snuck in a wee bit of Scottish ;)

PR #3619, opened 2025-01-14
Switch RTD build to fail on gallery example failure

### LINKED ISSUE #3603: Should we fail doc build on broken example moving forward?
Just a discussion based on the fact @chrishalcrow recently had to fix the examples because of a change in generator. I think the project is big enough that we need to make sure our changes don't break the docs to keep things look as professional as possible. Should we make it so rtd does the red check if the examples fail to build rather than allow them to pass with failed examples to alert us to these changes at the time of the PR instead of later when one of us checks? ie whoever broke it should fix it at the time?

zm711:
I included our options in a comment to make it easier to edit in the future.

Fixes #3603

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -119,7 +119,9 @@ from sphinx_gallery.sorting import FileNameSortKey
 
 # for sphinx gallery plugin
 sphinx_gallery_conf = {
-    'only_warn_on_example_error': True,
+    # This is the default but including here explicitly. Should build all docs and fail on gallery failures only.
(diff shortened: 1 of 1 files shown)


h-mayorquin:
This will require a change on the settings to add the passing of this check as a requirement for the pull request to be merged, right?

zm711:
I don't think we wanted it required. More just the shame of the red x rather than the green check. So that we are aware the PR caused problems.

alejoe91 (approved):

h-mayorquin:
Ah, got it wrong. Thanks, guys.

PR #3620, opened 2025-01-14
Improve docstring of `get_neuropixels_sample_shifts` 

h-mayorquin:
I am [working on improving the NWB annotation of this property](https://github.com/catalystneuro/neuroconv/pull/1177) and I took the time to improve the docstring of this function. I also added type hints.


src/spikeinterface/extractors/neuropixels_utils.py
```diff
@@ -1,39 +1,47 @@
...
-    num_cycles: int or None, default: None
-        The number of cycles in the ADC on the probe.
-        Neuropixels 1.0 probes have 13 cycles for AP and 12 for LFP.
+    num_cycles : int or None, default: None
+        Number of cycles in the ADC sampling sequence.
+        Neuropixels 1.0 probes have 13 cycles for AP (action potential) signals
+        and 12 for LFP (local field potential) signals.
         Neuropixels 2.0 probes have 16 cycles.
-        If None, the num_channels_per_adc is used.
+        If None, defaults to the value of num_channels_per_adc.
```
zm711, line 38:
```suggestion
        If None, defaults to the value of `num_channels_per_adc`.
```
:)

src/spikeinterface/extractors/neuropixels_utils.py
```diff
@@ -42,10 +50,11 @@ def get_neuropixels_sample_shifts(num_channels=384, num_channels_per_adc=12, num
         np.arange(num_channels), 2
     )
 
-    sample_shifts = np.zeros_like(adc_indices)
+    sample_shifts = np.zeros_like(adc_indices, dtype=float)
```
zm711, line 53: remind me the benefit of using python float in the case of numpy functions? This just auto scales the dtype between float32 and float64? Could this create float16 that could overflow? I honestly and super rusty on this stuff :)

zm711 (commented):
One question and adding the ticks.

src/spikeinterface/extractors/neuropixels_utils.py
h-mayorquin: Thanks, will change in the next commit.

src/spikeinterface/extractors/neuropixels_utils.py
h-mayorquin: It is an alias that will auto scale depending on the OS if I remember well. But now that I think about it, I don't want to introduce a change on behavior on this PR so I am removing this. Thanks for the catch.

zm711 (approved):
This looks good to me now. Technically one small code change but really just docs.

alejoe91 (approved):

PR #3622, opened 2025-01-15
Fix dataframe index dtype due to csv in  quality metrics 

samuelgarcia:


alejoe91:
I think it's a perfectly valid hack and wouldn't be so dramatic on the CSV choice :)

alejoe91 (approved):

PR #3623, opened 2025-01-15
Add `get_start_time`/`get_end_time` functions and use them in `get_duration`

alejoe91:
Since the `__repr__` uses the `get_duration`s function, I added an option to avoid using timestamps to estimate duration. 
In fact, this can be quite slow for long zarr datasets with timestamps, since the whole array is loaded in mem when calling `get_times()`.

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -233,7 +233,7 @@ class BaseRecording(BaseRecordingSnippets):
             The duration in seconds
         """
         segment_duration = (
-            self.get_stop_time(segment_index) - self.get_start_time(segment_index) + (1 / self.get_sampling_frequency())
+            self.get_end_time(segment_index) - self.get_start_time(segment_index) + (1 / self.get_sampling_frequency())
         )
         return segment_duration
 
@@ -456,7 +456,7 @@ class BaseRecording(BaseRecordingSnippets):
         rs = self._recording_segments[segment_index]
         return rs.get_start_time()
 
-    def get_stop_time(self, segment_index=None) -> float:
+    def get_end_time(self, segment_index=None) -> float:
         """Get the stop time of the recording segment.
 
         Parameters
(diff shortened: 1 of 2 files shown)


h-mayorquin:
Where do you think the loading into memory is happening?
Maybe here in the asarray:

https://github.com/h-mayorquin/spikeinterface/blob/80820853354fbc2c6a56bb5dade4c94c8dacb321/src/spikeinterface/core/baserecording.py#L895-L898

Because the duration only requires the last and the first timestamps:

https://github.com/h-mayorquin/spikeinterface/blob/80820853354fbc2c6a56bb5dade4c94c8dacb321/src/spikeinterface/core/baserecording.py#L237-L240

Maybe we can just not transform to array if the timestamps are memmap and or zarr provided they are read only?

alejoe91:
I think that transforming to array is very convenient to cache the timestamps in memory and overall.it improves performance. Another option could be to have a get_start_time and get_end_time, which will only access the required samples.

h-mayorquin:
Fair enough, changing the way that `get_times` work would probably require more discussion.

Adding private methods like this:
>  get_start_time and get_end_time, which will only access the required samples

Seems like a great idea to me. That way, we can use that in get duration without touching anything else.

alejoe91:
@h-mayorquin done :) much faster now

h-mayorquin:
This looks good but there are some tests that are failing due to tolerance I guess?

Two points for discussion:
1) We have `start_frame` and `end_frame` in the signature of get_traces. Shouldn't we have `get_start_time` and `get_end_time` to be consistent in the nomenclature? what do you think? To be honest, a mistake that I commonly make when calling get_traces is using stop_frame instead of end_frame...
2) Are you OK making these methods public? I think that's a good idea but the cautious thing would be to make them private and release them later in case we need to make some fixes that we find through usage.

alejoe91:
Changed to `get_end_time` and fixed its behavior (we have good tests for time handling! thanks @JoeZiminski!)

I'm ok in exposing these functions as public since their behavior is quite straightforward.

h-mayorquin (approved):
LGTM

Issue #3626, opened 2025-01-17
LFP channels vs AP channels


taningh86:
Hi,
This may be more of a general question and possibly more suited for Neuropixels slack channel. But, I figured I would still ask here because I am sure this question must have been raised in the past in the forum and also I am using spikeinterface to process LFP data.
Looks like LFP channels are different than AP channels. I haven't looked in depth on 'Why' but i could make sume asumptions why that would be. But If anyone knows about it in certain terms then please feel free to share. I use the code below to process LFP data and get  the LFP channel ID list. The ID's are long and seems to span multiple channels. Am I getting something wrong here or this is actually the case? if so, then why only these channels have LFP signals and not others.
`lfprec2 = si.bandpass_filter(
    raw_rec,
    freq_min=50,
    freq_max=150,

    margin_ms=1850.,
    filter_order=3,
    dtype="float32",
    add_reflect_padding=True,
)
lfprec2 = si.phase_shift(lfprec2)
lfprec2 = si.resample(lfprec2, resample_rate=150, margin_ms=100)

lfprec2 = si.directional_derivative(lfprec2, order=2, edge_order=1)
lfprec2 = si.average_across_direction(lfprec2)

print(lfprec2)`

`print(lfprec2.get_channel_ids())`

And below is the list of LFP channel it outputs:

`['imec0.lf#LF48-imec0.lf#LF49' 'imec0.lf#LF50-imec0.lf#LF51'
 'imec0.lf#LF52-imec0.lf#LF53' 'imec0.lf#LF54-imec0.lf#LF55'
 'imec0.lf#LF56-imec0.lf#LF57' 'imec0.lf#LF58-imec0.lf#LF59'
 'imec0.lf#LF60-imec0.lf#LF61' 'imec0.lf#LF62-imec0.lf#LF63'
 'imec0.lf#LF64-imec0.lf#LF65' 'imec0.lf#LF66-imec0.lf#LF67'
 'imec0.lf#LF68-imec0.lf#LF69' 'imec0.lf#LF70-imec0.lf#LF71'
 'imec0.lf#LF72-imec0.lf#LF73' 'imec0.lf#LF74-imec0.lf#LF75'
 'imec0.lf#LF76-imec0.lf#LF77' 'imec0.lf#LF78-imec0.lf#LF79'
 'imec0.lf#LF80-imec0.lf#LF81' 'imec0.lf#LF82-imec0.lf#LF83'
[...]

PR #3631, opened 2025-01-20
Fix channels bug in average_across_direction, and add a test

cwindolf:
Fixes https://github.com/SpikeInterface/spikeinterface/issues/3626

channel_indices was unused in the segment's get_traces()! My apologies. This fixes and adds a test.

### DIFF
--- a/src/spikeinterface/preprocessing/average_across_direction.py
+++ b/src/spikeinterface/preprocessing/average_across_direction.py
@@ -132,7 +132,7 @@ class AverageAcrossDirectionRecordingSegment(BasePreprocessorSegment):
         # now, divide by the number of channels at that position
         traces /= self.n_chans_each_pos
 
-        return traces
+        return traces[:, channel_indices]
 
 
 # function for API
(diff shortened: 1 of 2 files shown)


alejoe91 (approved):

src/spikeinterface/preprocessing/average_across_direction.py
```diff
@@ -132,7 +132,7 @@ def get_traces(self, start_frame, end_frame, channel_indices):
         # now, divide by the number of channels at that position
         traces /= self.n_chans_each_pos
 
-        return traces
+        return traces[:, channel_indices]
```
samuelgarcia, line 135: if channel_indices is None then it is adding one dimmension.
could tou test for this ?

cwindolf: Oh oops, didn't realize that can be None. Will do. I wasn't sure if I can reopen this PR (for the future -- is that possible?) so I did a new one here: https://github.com/SpikeInterface/spikeinterface/pull/3636

PR #3635, opened 2025-01-21
Fix precision/recall in hybrid example

### LINKED ISSUE #3633: Precision and recall interpretation in documentation
I'm a bit confused by my mental picture of how precision and recall are interpreted, compared to what I read in the documentation [here](https://spikeinterface.readthedocs.io/en/latest/how_to/benchmark_with_hybrid_recordings.html#plot-performances).

It mentions `Kilosort2.5 performs well when looking at precision (finding all spikes in a hybrid unit), at the cost of lower recall (finding spikes when it shouldn’t).` but in my mind it's the opposite.
As seen in the picture below, I would expect the "relevant elements" to be all spikes of the (hybrid) ground truth. And so the recall would then be akin to "finding all spikes" and the precision would related to not finding irrelevant spikes (false positives).
<p><a href="https://commons.wikimedia.org/wiki/File:Precisionrecall.svg#/media/File:Precisionrecall.svg"></a><br>By <a href="//commons.wikimedia.org/wiki/User:Walber" title="User:Walber">Walber</a> - <span class="int-own-work" lang="en">Own work</span>, <a href="https://creativecommons.org/licenses/by-sa/4.0" title="Creative Commons Attribution-Share Alike 4.0">CC BY-SA 4.0</a>, <a href="https://commons.wikimedia.org/w/index.php?curid=36926283">Link</a></p>

Is this a switchup in the documentation or am I missing something?

alejoe91:
Fixes #3633

### DIFF
--- a/doc/how_to/benchmark_with_hybrid_recordings.rst
+++ b/doc/how_to/benchmark_with_hybrid_recordings.rst
@@ -2531,9 +2531,8 @@ Although non of the sorters find all units perfectly, ``Kilosort2.5``,
 ``Kilosort4``, and ``SpyKING CIRCUS 2`` all find around 10-12 hybrid
 units with accuracy greater than 80%. ``Kilosort4`` has a better overall
 curve, being able to find almost all units with an accuracy above 50%.
-``Kilosort2.5`` performs well when looking at precision (finding all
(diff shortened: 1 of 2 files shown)


zm711 (approved):
Looks good to me.

PR #3638, opened 2025-01-23
Make `time_range` conversion to array dtype `float64` 

JoeZiminski:
One of the most edge-case, inconsequential bugs ever: If your recording is < 1 second long and if you specify `time_range=(0, 1)` instead of `(0, 1.0)` then the cast from `tuple` to `array` will generate an `int` array. When`t_end` is assigned to it it will round to `0` leading to `time_range=(0, 0)` and failure to retrieve traces.

This is for a super-short example file I am using😅


JoeZiminski:
Also, apologies I was using Githubs web interface to raise the commit and it did it from this repo not my fork, please let me know if its a problem and I will redo

alejoe91 (approved):

PR #3639, opened 2025-01-23
Force latest hdmf-zarr version

alejoe91:


zm711:
It's weird that the Win32 file permission error would happen stochastically rather than deterministically with our testing.... But I've definitely had users report SC2 and TDC2 issues with files not being released when using caching options on Windows. So maybe we should doublecheck if there is a global solution for this issue. Not the point of this PR, but maybe worth connecting to one of my old Windows issues floating around? What do you think?

h-mayorquin (approved):
lgtm

zm711 (approved):
I agree, the error is unrelated, so this should be merged, but I think it's important to log these Windows errors :)

PR #3481, opened 2024-10-15
Add python 313 to CI test

h-mayorquin:
To track if everything works.

PR #3640, opened 2025-01-24
Support numpy 2.0

alejoe91:
Starting from @h-mayorquin PR https://github.com/SpikeInterface/spikeinterface/pull/3481, but one thing at a time.

This is only for numpy 2.0 (with NEO 0.14.0)


pyproject.toml
```diff
@@ -20,11 +20,11 @@ classifiers = [
 
 
 dependencies = [
-    "numpy>=1.20, <2.0",  # 1.20 np.ptp, 1.26 might be necessary for avoiding pickling errors when numpy >2.0
+    "numpy>=1.20",
```
zm711, line 23: We are sure we have enough of these pickling tests to feel okay deleted Heberto's note?

zm711 (approved):
Looks good to me.

h-mayorquin (approved):
LGTM

pyproject.toml
alejoe91: I think so! :) let's see

PR #3646, opened 2025-01-27
Adding colorbar support to peak activity widget

abhinavsns:
I am following the tutorial located at: https://spikeinterface.readthedocs.io/en/stable/tutorials/widgets/plot_4_peaks_gallery.html
The feature for plotting and animating was already present. This PR adds support for the color bar which is not implemented.

Attached are the screenshot and gifs for the same tutorial when using `with_color_bar` and `fixed_color_range` arguments.
no_color_bar
with_color_bar
colorbarmoving
colorbarfixed


alejoe91:
Hi @abhinavsns 

Thank you for the PR! Can you add more information in the PR description and a screenshot/screencast of the widget? This would be the first animated widget, so it might require some discussion.

src/spikeinterface/widgets/peak_activity.py
```diff
@@ -100,21 +151,51 @@ def animate_func(i):
...
+        with_channel_ids,
+        with_contact_color,
+        with_interpolated_map,
+        with_color_bar,
+        vmin=None,
+        vmax=None,
+    ):
+        rates = self._compute_rates(rec, peaks, duration)
+
+        self.artists = ()
```
zm711, line 198: why did we need to put the artists into the class? It seems like they only exist inside the function or am I misreading?

src/spikeinterface/widgets/peak_activity.py
```diff
@@ -82,14 +88,59 @@ def plot_matplotlib(self, data_plot, **backend_kwargs):
         probe = probes[0]
 
         if dp.bin_duration_s is None:
-            self._plot_one_bin(
-                rec, probe, peaks, duration, dp.with_channel_ids, dp.with_contact_color, dp.with_interpolated_map
-            )
+            if type(dp.fixed_color_range) in [tuple, list]:
```
zm711, line 91: docstring only says tuple. This is a tiny point though so not sure we need to change.

src/spikeinterface/widgets/peak_activity.py
```diff
@@ -82,14 +88,59 @@ def plot_matplotlib(self, data_plot, **backend_kwargs):
...
+                self.vmin, self.vmax = dp.fixed_color_range
```
zm711, line 92: if you're just putting them into the function why do you need to pull them out and put them in the class?

zm711 (commented):
Couple initial questions from me. But I personally think this cool!

src/spikeinterface/widgets/peak_activity.py
abhinavsns: I was using it to access the artists after creating the `plot_peak_activity` object. But I think it is not necessary so will remove.

src/spikeinterface/widgets/peak_activity.py
abhinavsns: fixed

src/spikeinterface/widgets/peak_activity.py
```diff
@@ -30,8 +30,15 @@ class PeakActivityMapWidget(BaseWidget):
         Plot rates with interpolated map
     with_channel_ids : bool, default: False
         Add channel ids text on the probe
-
-
+    fixed_color_range : bool or tuple/list of length=2, default: False
+        Fixes the color bar range when animating or plotting
+        in case of a length 2 tuple [vmin,vmax], uses that to fixes the colorbar range
```
zm711, line 35:
```suggestion
        in case of a length 2 tuple [vmin,vmax], uses that to fix the colorbar range
```

zm711 (commented):
One docstring and this if fine by me. I leave @alejoe91 and @samuelgarcia to do the final review.

alejoe91:
@abhinavsns after we discussed internally, we agreed that it doesn't make sense to have `fixed_color_range=False`. 
I updated and simplified the code a bit, so a user can either pass a custom `color_range`, or the `fixed_color_range` will be default.

alejoe91 (approved):

PR #3649, opened 2025-01-28
Add LocationsWidget and plot_locations

chrishalcrow:
Add a `LocationsWidget` and `plot_locations`, to complement the `AmplitudesWidget`  and `plot_amplitudes`.

This plots the spike locations as a function of time, and is 99% the same as the `plot_amplitudes` function. You can control which direction you want to plot using e.g. `directions_axis = 'x'` and plot a histogram if you'd like.

An example:
``` python
_, drift_rec, sort = si.generate_drifting_recording(
    num_units=2, 
    duration=30,
    generate_displacement_vector_kwargs=dict(
        displacement_sampling_frequency=20.0,
        drift_start_um=[0, 50],
        drift_stop_um=[0, -50],
        drift_step_um=5,
        motion_list=[
            dict(
                drift_mode="zigzag",
                non_rigid_gradient=None,
                t_start_drift=0.0,
                t_end_drift=None,
                period_s=300,
            ),
        ],
    )
)

sorting_analyzer = si.create_sorting_analyzer(sorting=sort, recording=drift_rec)
sorting_analyzer.compute(['random_spikes', 'templates', 'spike_locations'])
si.plot_locations(sorting_analyzer, unit_ids=['0'])
```
Screenshot 2025-01-28 at 19 06 54

Or

```
si.plot_locations(sorting_analyzer, unit_ids=['0'], plot_histograms=True)
```

Screenshot 2025-01-28 at 19 06 30

Can plot multiple units, e.g.

```
si.plot_locations(sorting_analyzer, unit_ids=['0','1'])
```
Screenshot 2025-01-28 at 19 11 37

Up for debate:
- is this a useful widget? I've been using it, so I think it is!
- is a histogram useful?
- Naming. We already have `plot_spike_locations` which plots the spike locations on the probe. This naming is designed to pair with `plot_amplitudes`.

### DIFF
--- a/src/spikeinterface/widgets/spike_locations_by_time.py
+++ b/src/spikeinterface/widgets/spike_locations_by_time.py
@@ -1,0 +1,263 @@
+from __future__ import annotations
+
+import numpy as np
+from warnings import warn
+
+from .base import BaseWidget, to_attr
+from .utils import get_some_colors
+
+from ..core.sortinganalyzer import SortingAnalyzer
+
+
+class LocationsWidget(BaseWidget):
+    """
+    Plots spike locations as a function of time
+
+    Parameters
+    ----------
+    sorting_analyzer : SortingAnalyzer
+        The input sorting analyzer
+    unit_ids : list or None, default: None
+        List of unit ids
+    segment_index : int or None, default: None
+        The segment index (or None if mono-segment)
+    max_spikes_per_unit : int or None, default: None
+        Number of max spikes per unit to display. Use None for all spikes
+    hide_unit_selector : bool, default: False
+        If True the unit selector is not displayed
+        (sortingview backend)
+    plot_histogram : bool, default: False
+        If True, an histogram of the locations is plotted on the right axis
+        (matplotlib backend)
+    bins : int or None, default: None
+        If plot_histogram is True, the number of bins for the location histogram.
+        If None this is automatically adjusted
+    plot_legend : bool, default: True
+        True includes legend in plot
+    locations_axis : str, default: 'y'
+        Which location axis to use when plotting locations.
+    """
+
+    def __init__(
+        self,
+        sorting_analyzer: SortingAnalyzer,
+        unit_ids=None,
+        unit_colors=None,
+        segment_index=None,
+        max_spikes_per_unit=None,
+        hide_unit_selector=False,
+        plot_histograms=False,
+        bins=None,
+        plot_legend=True,
+        locations_axis="y",
+        backend=None,
+        **backend_kwargs,
+    ):
+
+        sorting_analyzer = self.ensure_sorting_analyzer(sorting_analyzer)
+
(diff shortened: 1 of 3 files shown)


samuelgarcia:
Good idea. I think it can be usefull.
Maybe we could move it to the same file than spike_location.

Another idea (not sure it is a good one) could be to move this logic in the spike_location widget using having options like: 
  * `along_times=True/False`
  * `with_histogram=True/False`

JoeZiminski:
Looks really nice! Under the hood this looks similar to the [`DriftRasterMapWidget`](https://github.com/SpikeInterface/spikeinterface/pulls) as shown [here](https://spikeinterface.readthedocs.io/en/stable/how_to/handle_drift.html#plot-the-results) but with the spikes coloured by unit membership rather than amplitude? In this case maybe there is scope to leverage the existing raster-plot backend and use the array of unit ids for colouring?

Incorporating this directly with `DraftRasterMapWidget` would be confusing but there might be scope for a common raster-plot backend (I don't think there is anything particularly drift-specific about the `DriftRasterMapWidget`, it is just spike depth x time) that these various functions could call, with different spike-colouring options. There is also a [`RasterWidget`](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/widgets/rasters.py#L9), I'm not sure the scope / intersection across these different widgets.

The histogram look really nice, just in case useful here is another version on [this PR](https://github.com/SpikeInterface/spikeinterface/pull/3499) that I never got over the line.

chrishalcrow:
> Looks really nice! Under the hood this looks similar to the [`DriftRasterMapWidget`](https://github.com/SpikeInterface/spikeinterface/pulls) as shown [here](https://spikeinterface.readthedocs.io/en/stable/how_to/handle_drift.html#plot-the-results) but with the spikes coloured by unit membership rather than amplitude? In this case maybe there is scope to leverage the existing raster-plot backend and use the array of unit ids for colouring?
> 
> Incorporating this directly with `DraftRasterMapWidget` would be confusing but there might be scope for a common raster-plot backend (I don't think there is anything particularly drift-specific about the `DriftRasterMapWidget`, it is just spike depth x time) that these various functions could call, with different spike-colouring options. There is also a [`RasterWidget`](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/widgets/rasters.py#L9), I'm not sure the scope / intersection across these different widgets.
> 
> The histogram look really nice, just in case useful here is another version on [this PR](https://github.com/SpikeInterface/spikeinterface/pull/3499) that I never got over the line.

That's a great idea, Joe! I just hacked together a 90%-the-same plot using `DriftRasterMapWidget` in a few lines. I think a common raster-plot backend is a good idea - I'll give it a go!

Thanks for the other feedback, all!

alejoe91 (approved):

PR #3651, opened 2025-01-29
Extend `si.load` capabilities

alejoe91:
This PR extends the `si.load` magic function to load also `SortingAnalyzers`.

In addition, a better logic is implemented to load zarr files, both locally and remotely, including trying to open the folder with different options: `zarr.open`/`zarr.open_consolidated` and `anon False/True` for remote assets. This simplifies handling of zarr opening everywhere


pyproject.toml
```diff
@@ -123,6 +123,7 @@ test_core = [
     "pytest",
     "pytest-dependency",
     "psutil",
+    "s3fs", # for streaming tests
```
samuelgarcia, line 126: This is a big one for core tests no ?

samuelgarcia:
Good idea but I am not sure to like the design with try except.
We need to do clear check on json files or zarr attribute (also) json for information on chich object it is.

samuelgarcia:
And also Template

samuelgarcia:
@alejoe91  : I refactor everything to have more general case.
This looks big but this is an easier design which avoid nested if else evrywhere.
You will see that there are 2 steps guess the object and then loading.
This is more easy to maintains.

src/spikeinterface/core/loading.py
```diff
@@ -18,110 +29,274 @@ def load(file_or_folder_or_dict, base_folder=None) -> BaseExtractor:
...
+    elif object_type == "Sorting":
+        from .zarrextractors import read_zarr_sorting
+
+        storage_options = kwargs.get("storage_options", None)
+        sorting = read_zarr_sorting(folder_or_url, storage_options=storage_options)
+        return sorting
+    elif object_type == "Recording|Sorting":
+        # This case shoudl deprecated soon because the read_zarr is ultra ambiguous
+        # just testing if the zarr contains unit_ids or channel_ids but many object also contains it (see template)!!!!
+        from .zarrextractors import read_zarr
```
samuelgarcia, line 298: @alejoe91 : any idea how to avoid using this read_zarr stuff ?
If think that read_zarr should deprecated because this is too ambiguous.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -659,11 +642,12 @@ def create_zarr(cls, folder, sorting, recording, sparsity, return_scaled, rec_at
     @classmethod
     def load_from_zarr(cls, folder, recording=None, backend_options=None):
         import zarr
+        from .loading import load
 
         backend_options = {} if backend_options is None else backend_options
         storage_options = backend_options.get("storage_options", {})
 
-        zarr_root = zarr.open_consolidated(str(folder), mode="r", storage_options=storage_options)
+        zarr_root = super_zarr_open(str(folder), mode="r", storage_options=storage_options)
```
samuelgarcia, line 650: @alejoe91  : Here, is it a problem that the zarr_root could be eventually not consolidated ?

samuelgarcia:
@alejoe91  : really cool.
The only stuff we should change is the s3fs for core testing. I would move it to a less frequent testing like extrator even if it is in the core because the core tests is havilly launched  and any eternal downloading would make it more slow I think.

src/spikeinterface/core/sortinganalyzer.py
alejoe91: It could for some older assets. Plus in append mode we can't use consolidated

samuelgarcia (approved):

PR #3652, opened 2025-01-29
Use SpikeInterface-ported pre-trained models in doc

alejoe91:
We ported the UnitRefine models in the SpikeInterface org in HF. This PR updates the models examples to reflect it


chrishalcrow (approved):
Checked links - all works for me.

PR #3657, opened 2025-01-30
Handling segments while interpolating Motion

yger:
Actually, I've found that multi-segment is introducing a bug when saving/reloading corrected recordings. This should patch the issue

The bug can be reproduce in main by simply doing
>> static, drifting, sorting = si.generate_drifting_recording(probe_name='Neuropixel-128')
>> corrected = si.correct_motion(drifting)
>> corrected.dump("rec.pickle")
>> corrected2 = si.load('rec.pickle')

The patch should fix the issue


samuelgarcia:
I think your patch is uncorrect I will try something else.

samuelgarcia:
@yger : can you test #3659 instead ?

PR #3660, opened 2025-01-30
keep same dtype after scale_to_uV

RobertoDF:
Right now rec type gets converted silently when applying scale_to_uV. This commit preserves dtype.

### DIFF
--- a/src/spikeinterface/preprocessing/scale.py
+++ b/src/spikeinterface/preprocessing/scale.py
@@ -34,7 +34,7 @@ def scale_to_uV(recording: BasePreprocessor) -> BasePreprocessor:
     gain = recording.get_channel_gains()
     offset = recording.get_channel_offsets()
 
-    scaled_to_uV_recording = ScaleRecording(recording, gain=gain, offset=offset, dtype="float32")
+    scaled_to_uV_recording = ScaleRecording(recording, gain=gain, offset=offset, dtype=recording.get_dtype())
 
     # We do this so when get_traces(return_scaled=True) is called, the return is the same.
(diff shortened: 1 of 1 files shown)


zm711:
I think the issue with this is if your recording has a dtype of int then you could either overflow or truncate your actual result. There have been debates about--I think @h-mayorquin would be good to comment on this. But I think in general if you're scaling the dtype should be a float. But maybe instead of `float32` we should just make the default be `float` to allow for `float64` if necessary. 

What is the use case you're thinking of?

alejoe91:
not sure about this. IMO microvolts are float by definition, since it's a physical unit. So it makes sense that you would need another astype if you want to coerce to ints

RobertoDF:
I see, I have a recording with [ONIX](https://open-ephys.github.io/onix-docs/index.html) and these binaries are np.uint16 so this hidden conversion was causing me problems. So I should first convert to float and afterward scale?

RobertoDF:
Right now in their [official docs](https://open-ephys.github.io/bonsai-onix1-docs/articles/hardware/np1e/load-data.html) they propose: 
```
ap_scalar = 1.2e6 / (2 ** bit_depth) / ap_gain
ap_offset = (2 ** (bit_depth - 1)) * ap_scalar
ap_recording = se.read_binary(os.path.join(data_directory,  f"np1-spike_{suffix}.raw"), 
                           3e5, 
                           np.uint16, 
                           num_channels, 
                           gain_to_uV=ap_scalar, 
                           offset_to_uV=-ap_offset) 
```

PR #3661, opened 2025-01-31
Add `BaseRasterWidget`

chrishalcrow:
Added `BaseRasterWidget` widget, and re-factored `AmplitudeWidget`, `RasterWidget` and `DriftRasterMapWidget` to use it.

Following on from @JoeZiminski's comment in #3649, I’ve made a `BaseRasterWidget`, which can be used to make spike times vs anything plots. Some advantages:

- Unifies the plotting logic for `AmplitudeWidget`, `RasterWidget` and `DriftRasterMapWidget` (and later LocationsWidget)
- Plotting kwargs from some widgets can now be used by others. E.g. `y_lim` and `scatter_decimate` from `DriftRasterMapWidget` is now available for `AmplitudeWidget`.

Now users (and devs!) can easily make custom raster plots, e.g. if you have a sorting analyzer called `sa` with spike_amplitudes computed, you can do:

``` python
all_spiketrains = {
    unit_id: sa.sorting.get_unit_spike_train(unit_id, segment_index=0, return_times=True) 
for unit_id in sa.sorting.unit_ids}

locations = sa.get_extension("spike_amplitudes").get_data(outputs="by_unit")[0]
x_locs_squares = {
    unit_id: pow(locations[unit_id],2)
for unit_id in sa.sorting.unit_ids}

BaseRasterWidget(spike_train_data=all_spiketrains, y_axis_data=x_locs_squares, unit_ids=[2,6], backend="matplotlib", y_lim=[0,50_000], title="Amplitudes Squared!!!", y_label="I love axis labels")

```
which makes this:
amp_squared

The goal was for this to be totally backend and not change any plots, but when doing it, I think it was a chance to unify some widget style choices. E.g. drawing just the left and bottom axes, and making the data fill the plot. It's a little difficult to test all this stuff. Here's a bunch of plots using `main` and this PR. I've also checking that SortingView can still do an amplitudes plot and played with `ipywidgets` a lot.

OLD vs NEW

lotsa_plots

### DIFF
--- a/src/spikeinterface/widgets/amplitudes.py
+++ b/src/spikeinterface/widgets/amplitudes.py
@@ -3,13 +3,16 @@ from __future__ import annotations
 import numpy as np
 from warnings import warn
 
+from .rasters import BaseRasterWidget
 from .base import BaseWidget, to_attr
 from .utils import get_some_colors
 
 from ..core.sortinganalyzer import SortingAnalyzer
 
+from spikeinterface.core import SortingAnalyzer
 
-class AmplitudesWidget(BaseWidget):
+
+class AmplitudesWidget(BaseRasterWidget):
     """
     Plots spike amplitudes
 
@@ -19,6 +22,9 @@ class AmplitudesWidget(BaseWidget):
         The input waveform extractor
     unit_ids : list or None, default: None
         List of unit ids
+    unit_colors : dict | None, default: None
+        Dict of colors with key : unit, value : color. If None, uses `get_some_colors` to
+        generate colors.
     segment_index : int or None, default: None
         The segment index (or None if mono-segment)
     max_spikes_per_unit : int or None, default: None
@@ -61,9 +67,6 @@ class AmplitudesWidget(BaseWidget):
         if unit_ids is None:
             unit_ids = sorting.unit_ids
 
-        if unit_colors is None:
-            unit_colors = get_some_colors(sorting.unit_ids)
-
         if sorting.get_num_segments() > 1:
             if segment_index is None:
                 warn("More than one segment available! Using segment_index 0")
@@ -73,13 +76,11 @@ class AmplitudesWidget(BaseWidget):
         amplitudes_segment = amplitudes[segment_index]
         total_duration = sorting_analyzer.get_num_samples(segment_index) / sorting_analyzer.sampling_frequency
 
-        spiketrains_segment = {}
-        for i, unit_id in enumerate(sorting.unit_ids):
-            times = sorting.get_unit_spike_train(unit_id, segment_index=segment_index)
-            times = times / sorting.get_sampling_frequency()
-            spiketrains_segment[unit_id] = times
+        all_spiketrains = {
+            unit_id: sorting.get_unit_spike_train(unit_id, segment_index=0, return_times=True)
+            for unit_id in sorting.unit_ids
+        }
 
-        all_spiketrains = spiketrains_segment
         all_amplitudes = amplitudes_segment
         if max_spikes_per_unit is not None:
             spiketrains_to_plot = dict()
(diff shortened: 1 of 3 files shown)


zm711:
I have to admit I like your style better with the edges removed. Looking classier to me. Something I've been meaning to suggest for a long time!

src/spikeinterface/widgets/amplitudes.py
```diff
@@ -19,10 +22,17 @@ class AmplitudesWidget(BaseWidget):
         The input waveform extractor
     unit_ids : list or None, default: None
         List of unit ids
+    unit_colors : dict | None, default: None
+        Dict of colors with key : unit, value : color. If None, uses `get_some_colors` to
```
zm711, line 26: Maybe we need to explain what `get_some_colors` is since that is an internal utility function and not a well known python function. Also should we be better explaining what "color" is? Is it any representation of color (hex, string, one letter abbreviation) or does it require something specific.

src/spikeinterface/widgets/amplitudes.py
```diff
@@ -19,10 +22,17 @@ class AmplitudesWidget(BaseWidget):
...
+        generate colors.
     segment_index : int or None, default: None
         The segment index (or None if mono-segment)
     max_spikes_per_unit : int or None, default: None
         Number of max spikes per unit to display. Use None for all spikes
+    y_lim : tuple or None, default: None
+        The min and max depth to display, if None (min and max of the amplitudes).
+    scatter_decimate : int, default: 1
+        If > 1, the scatter points are decimated.
```
zm711, line 35: What does 2 mean in this case? I think for users we should be more explicit that scatter_decimate cuts data in half or say it takes very other or that it takes every nth datapoint. Although the word decimate means take every 1/10 I think people that aren't used to this might not fully understand how this works.

src/spikeinterface/widgets/amplitudes.py
```diff
@@ -73,13 +82,11 @@ def __init__(
         amplitudes_segment = amplitudes[segment_index]
         total_duration = sorting_analyzer.get_num_samples(segment_index) / sorting_analyzer.sampling_frequency
 
-        spiketrains_segment = {}
-        for i, unit_id in enumerate(sorting.unit_ids):
-            times = sorting.get_unit_spike_train(unit_id, segment_index=segment_index)
-            times = times / sorting.get_sampling_frequency()
-            spiketrains_segment[unit_id] = times
+        all_spiketrains = {
+            unit_id: sorting.get_unit_spike_train(unit_id, segment_index=0, return_times=True)
```
zm711, line 86:
```suggestion
            unit_id: sorting.get_unit_spike_train(unit_id, segment_index=segment_index, return_times=True)
```
this only defaults to 0 if the `sorting_index=None` so in the case the user gives an index we should use it.

src/spikeinterface/widgets/rasters.py
```diff
@@ -4,16 +4,284 @@
...
 
 from .base import BaseWidget, to_attr, default_backend_kwargs
+from .utils import get_some_colors
 
 
-class RasterWidget(BaseWidget):
+class BaseRasterWidget(BaseWidget):
+    """
+    Make a raster plot with spike times on the x axis and arbritary data on the y axis.
+    Can customise plot with histograms, title, labels, ticks etc.
```
zm711, line 13:
```suggestion
    Can customize plot with histograms, title, labels, ticks etc.
```
Nice try ;)

src/spikeinterface/widgets/rasters.py
```diff
@@ -4,16 +4,284 @@
...
+
+        import ipywidgets.widgets as W
+        from IPython.display import display
+        from .utils_ipywidgets import check_ipywidget_backend, UnitSelector
+
+        check_ipywidget_backend()
+
+        self.next_data_plot = data_plot.copy()
+
+        cm = 1 / 2.54
```
zm711, line 197: If you used inches you wouldn't have to worry about pesky conversions.

src/spikeinterface/widgets/rasters.py
```diff
@@ -4,16 +4,284 @@
...
+
+        width_cm = backend_kwargs["width_cm"]
+        height_cm = backend_kwargs["height_cm"]
+
+        ratios = [0.15, 0.85]
```
zm711, line 202: Where does this come from?

src/spikeinterface/widgets/rasters.py
```diff
@@ -25,64 +293,64 @@ class RasterWidget(BaseWidget):
...
+            raise Exception("Should supply either a sorting or a sorting_analyzer, not both")
+        if sorting_analyzer is not None:
+            sorting = sorting_analyzer.sorting
+
         sorting = self.ensure_sorting(sorting)
 
         if segment_index is None:
             if sorting.get_num_segments() != 1:
                 raise ValueError("You must provide segment_index=...")
             segment_index = 0
```
zm711, line 318: This seems like a bug. You must provide segment_index but then it ignores it? I think we should fix his no?

zm711: IE this should be part of the else statement.

src/spikeinterface/widgets/rasters.py
```diff
@@ -25,64 +293,64 @@ class RasterWidget(BaseWidget):
...
 
-        if time_range is None:
-            frame_range = [0, sorting.to_spike_vector()[-1]["sample_index"]]
-            time_range = [f / sorting.sampling_frequency for f in frame_range]
-        else:
+        if unit_ids is None:
+            unit_ids = sorting.unit_ids
+
+        all_spiketrains = {
+            unit_id: sorting.get_unit_spike_train(unit_id, segment_index=0, return_times=True) for unit_id in unit_ids
```
zm711, line 324:
```suggestion
            unit_id: sorting.get_unit_spike_train(unit_id, segment_index=segment_index, return_times=True) for unit_id in unit_ids
```
should still be user-specifiable. But I think we need to fix the logic above to make sure nothing bad sneaks through.

zm711 (commented):
first round of comments. I think the main thing is actually making sure the segment handling is working as expected. I think a bug from before this refactor was being propagated.

src/spikeinterface/widgets/amplitudes.py
chrishalcrow: Great, this was propogated forward, so I think this counts as you adding a new feature @zm711 !

src/spikeinterface/widgets/rasters.py
chrishalcrow: I've added it to the if/else for logical clarity, but it doesn't change behaviour. I think that's ok: if you've not specificed segment index, we'll check if it's multi-segment. If it is multi-segment, we'll use the only segment: segment 0. I believe a one-segment sorting always has `segment_index = 0`.

e.g.
```
sub_sort = si.select_segment_sorting(sorting, segment_indices=1)
sub_sort.get_unit_spike_train(segment_index=0, unit_id='4')
```
...runs fine.

src/spikeinterface/widgets/rasters.py
chrishalcrow: It's from `SpikeAmplitudesWidget`, not sure how these were chosen. But everything looks nice! @samuelgarcia ?

src/spikeinterface/widgets/rasters.py
chrishalcrow: Don't tempt me...

Screenshot 2025-02-18 at 16 31 19

src/spikeinterface/widgets/rasters.py
zm711: Now that I'm re-reading this isn't the problem.

chrishalcrow:
There's currently different behaviour for plot_rasters and plot_amplitudes for a multi-segment analyzers.
- For `plot_rasters`, an error is raised if you forget to supply a segment_index (for a multi-segment recording). 
- For `plot_amplitudes`, the `0`th segment is chosen by default if no segment index is chosen.

Should we keep the different behaviours? Or unify the behaviour?

@samuelgarcia @zm711

src/spikeinterface/widgets/rasters.py
alejoe91: Probably me :)

src/spikeinterface/widgets/amplitudes.py
chrishalcrow: `unit_colors` is documented 14 times in the docstrings... so let's make sure we get a good description!

`Dict of colors with unit ids as keys and colors as values. Colors can be any type accepted by matplotlib. If None, a default set of colours is chosen using the `get_some_colors` function.`
??

I don't think we need to describe the `get_some_colors` function beyond this?

zm711:
I would say unify the behaviors in general. I only deal in mono segments so if Alessio is good with warn and do the first then that works for me for everything. I think I would lean toward error to force people to specify a segment so they realize this is only plotting one segment, but I'm also okay with a warning.

src/spikeinterface/widgets/rasters.py
chrishalcrow: Hopefully resolved with new default behaviour?

src/spikeinterface/widgets/all_amplitudes_distributions.py
```diff
@@ -19,8 +19,9 @@ class AllAmplitudesDistributionsWidget(BaseWidget):
         The SortingAnalyzer
     unit_ids : list
         List of unit ids, default None
-    unit_colors : None or dict
-        Dict of colors with key : unit, value : color, default None
+    unit_colors : dict | None, default: None
+        Dict of colors with unit ids as keys and colors as values. Colors can be any type accepted
+        by matplotlib. If None, default colours are chosen using the `get_some_colors` function.
```
zm711, line 24:
```suggestion
        by matplotlib. If None, default colors are chosen using the `get_some_colors` function.
```
I know it's hard ....

src/spikeinterface/widgets/amplitudes.py
```diff
@@ -19,10 +22,17 @@ class AmplitudesWidget(BaseWidget):
         The input waveform extractor
     unit_ids : list or None, default: None
         List of unit ids
+    unit_colors : dict | None, default: None
+        Dict of colors with unit ids as keys and colors as values. Colors can be any type accepted
+        by matplotlib. If None, default colours are chosen using the `get_some_colors` function.
```
zm711, line 27:
```suggestion
        by matplotlib. If None, default colors are chosen using the `get_some_colors` function.
```

src/spikeinterface/widgets/amplitudes.py
```diff
@@ -31,7 +41,7 @@ class AmplitudesWidget(BaseWidget):
         (matplotlib backend)
     bins : int or None, default: None
         If plot_histogram is True, the number of bins for the amplitude histogram.
-        If None this is automatically adjusted
+        If None, use 100 bins.
```
zm711, line 44:
```suggestion
        If None, uses 100 bins.
```
right? the function is the singular subject?  or you're commanding it I guess.

src/spikeinterface/widgets/crosscorrelograms.py
```diff
@@ -31,8 +31,9 @@ class CrossCorrelogramsWidget(BaseWidget):
         this argument is ignored
     hide_unit_selector : bool, default: False
         For sortingview backend, if True the unit selector is not displayed
-    unit_colors : dict or None, default: None
-        If given, a dictionary with unit ids as keys and colors as values
+    unit_colors : dict | None, default: None
+        Dict of colors with unit ids as keys and colors as values. Colors can be any type accepted
+        by matplotlib. If None, default colours are chosen using the `get_some_colors` function.
```
zm711, line 36:
```suggestion
        by matplotlib. If None, default colors are chosen using the `get_some_colors` function.
```

src/spikeinterface/widgets/metrics.py
```diff
@@ -24,8 +24,9 @@ class MetricsBaseWidget(BaseWidget):
         If given, a list of quality metrics to skip, default: None
     include_metrics: list or None, default: None
         If given, a list of quality metrics to include, default: None
-    unit_colors :  dict or None, default: None
-        If given, a dictionary with unit ids as keys and colors as values
+    unit_colors : dict | None, default: None
+        Dict of colors with unit ids as keys and colors as values. Colors can be any type accepted
+        by matplotlib. If None, default colours are chosen using the `get_some_colors` function.
```
zm711, line 29:
```suggestion
        by matplotlib. If None, default colors are chosen using the `get_some_colors` function.
```

src/spikeinterface/widgets/rasters.py
```diff
@@ -4,16 +4,284 @@
...
+        Duration of spike_train_data in seconds.
+    plot_histograms : bool, default: False
+        Plot histogram of y-axis data in another subplot
+    bins : int | None, default: None
+        Number of bins to use in histogram. If None, use 1/30 of spike train sample length.
+    scatter_decimate : int | None, default: None
+        If equal to n, each nth spike is kept for plotting.
+    unit_colors : dict | None, default: None
+        Dict of colors with unit ids as keys and colors as values. Colors can be any type accepted
+        by matplotlib. If None, default colours are chosen using the `get_some_colors` function.
```
zm711, line 34:
```suggestion
        by matplotlib. If None, default colors are chosen using the `get_some_colors` function.
```

src/spikeinterface/widgets/spike_locations.py
```diff
@@ -23,8 +23,9 @@ class SpikeLocationsWidget(BaseWidget):
         Number of max spikes per unit to display. Use None for all spikes.
     with_channel_ids : bool, default: False
         Add channel ids text on the probe
-    unit_colors :  dict or None, default: None
-        If given, a dictionary with unit ids as keys and colors as values
+    unit_colors : dict | None, default: None
+        Dict of colors with unit ids as keys and colors as values. Colors can be any type accepted
+        by matplotlib. If None, default colours are chosen using the `get_some_colors` function.
```
zm711, line 28:
```suggestion
        by matplotlib. If None, default colors are chosen using the `get_some_colors` function.
```

src/spikeinterface/widgets/spikes_on_traces.py
```diff
@@ -32,9 +32,9 @@ class SpikesOnTracesWidget(BaseWidget):
     sparsity : ChannelSparsity or None, default: None
         Optional ChannelSparsity to apply
         If SortingAnalyzer is already sparse, the argument is ignored
-    unit_colors : dict or None, default: None
-        If given, a dictionary with unit ids as keys and colors as values
-        If None, then the get_unit_colors() is internally used. (matplotlib backend)
+    unit_colors : dict | None, default: None
+        Dict of colors with unit ids as keys and colors as values. Colors can be any type accepted
+        by matplotlib. If None, default colours are chosen using the `get_some_colors` function.
```
zm711, line 37:
```suggestion
        by matplotlib. If None, default colors are chosen using the `get_some_colors` function.
```

zm711 (commented):
I'm sorry for this.

alejoe91 (approved):

samuelgarcia (approved):

PR #3664, opened 2025-02-03
Trying to delete sorting for windows

yger:


zm711:
It seems like you only write the `spikes.npy` in debugging mode. Can we shut off debugging mode just for the Windows tests and see if that fixes it? L332

```python

                if params["debug"]:
                    fitting_folder = sorter_output_folder / "fitting"
                    fitting_folder.mkdir(parents=True, exist_ok=True)
                    np.save(fitting_folder / "spikes", spikes)

                if verbose:
                    print("Found %d spikes" % len(spikes))
```

zm711:
Okay so after some more research I think I has a lead. So the `NumpyFolderSorting` relies on generating a memmap of the spikes and the numpy docs make it clear that the closing of the memmaps are never guaranteed even if you call `memmap.close()` you would have to wait for a garbage college for this to work. Now the difference between 3.9 and 3.12 is that 3.9 has numpy 2.0 and 3.12 downloads 2.2 so maybe there is a tiny change in memmapping between the versions. 

so I think the problem is:

https://github.com/SpikeInterface/spikeinterface/blob/5b2f2f30ad8437dfbee27e8cdbdd7e2367ab199a/src/spikeinterface/sorters/internal/spyking_circus2.py#L394

you write out the folder. And then when we call `load` in the `run_sorter` function we do the memmap here:

https://github.com/SpikeInterface/spikeinterface/blob/5b2f2f30ad8437dfbee27e8cdbdd7e2367ab199a/src/spikeinterface/core/sortingfolder.py#L40

This memmap is not always closed and so often leads to the test on Windows to fail because of this line

https://github.com/SpikeInterface/spikeinterface/blob/5b2f2f30ad8437dfbee27e8cdbdd7e2367ab199a/src/spikeinterface/sorters/runsorter.py#L277

So the solution would be either we change the memmapping to make this work with Windows or we could make the sorting writing out npz file instead (I think) which wouldn't make the memmap. Maybe Sam should weigh in, but do you think there is a huge benefit to using the `NumpyFolderSorting` in this case? Especially if the memmaps aren't being closed when we need them to?

yger:
Ok, the suggestion of @zm711  is working, if we are saving to npz_folder, both for SC2 (and I think it has also to be done for TDC), then tests are passing

src/spikeinterface/sorters/internal/spyking_circus2.py
```diff
@@ -391,7 +391,7 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
         if folder_to_delete is not None:
             shutil.rmtree(folder_to_delete)
 
-        sorting = sorting.save(folder=sorting_folder)
+        sorting = sorting.save(folder=sorting_folder, format="npz_folder")
```
zm711, line 394:
```suggestion
        # `numpy_folder doesn't release the memmap on Windows see https://github.com/SpikeInterface/spikeinterface/pull/3664
        sorting = sorting.save(folder=sorting_folder, format="npz_folder")
```

should we include a note if this gets approved so we know why? Or you think you guys will change a bunch and so the note isn't necessary?

src/spikeinterface/sorters/internal/tridesclous2.py
```diff
@@ -247,6 +247,6 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
         final_spikes["segment_index"] = spikes["segment_index"]
 
         sorting = NumpySorting(final_spikes, sampling_frequency, labels_set)
-        sorting = sorting.save(folder=sorter_output_folder / "sorting")
+        sorting = sorting.save(folder=sorter_output_folder / "sorting", format="npz_folder")
```
zm711, line 250:
```suggestion
        # `numpy_folder doesn't release the memmap on Windows see https://github.com/SpikeInterface/spikeinterface/pull/3664
        sorting = sorting.save(folder=sorter_output_folder / "sorting", format="npz_folder")
```

zm711 (commented):
Maybe we add a comment with a link to this PR for provenance in case we forget the Windows issue in the future ?

yger:
Good idea, thanks for the suggestion

alejoe91:
LGTM, @samuelgarcia do you approve?

src/spikeinterface/sorters/internal/si_based.py
```diff
@@ -21,7 +21,11 @@ def _setup_recording(cls, recording, output_folder, params, verbose):
     @classmethod
     def _get_result_from_folder(cls, output_folder):
         sorting = load(output_folder / "sorting")
-        return sorting
+        from spikeinterface.core.numpyextractors import SharedMemorySorting
+
+        shm_sorting = SharedMemorySorting.from_sorting(sorting)
+        del sorting
+        return shm_sorting
```
samuelgarcia, line 28: I do not think we should use the sharedmem always.
Whey do we need this ?

yger: I was trying to get rid of the memmap, but @zm711  found the patch, and thought it could not hurt anything to leave that here, can it?

yger: This has been removed

samuelgarcia:
I do not validate this.
The deep problem is the memmap when reading sorting.
Lets make another PR.

PR #3665, opened 2025-02-03
Adding support for lowpass filter

jesusdpa1:
Adding lowpass filter class, useful for processing emg recordings.

Copying the class from highpass filter and applying the require changes

### DIFF
--- a/src/spikeinterface/preprocessing/filter.py
+++ b/src/spikeinterface/preprocessing/filter.py
@@ -301,6 +301,7 @@ class LowpassFilterRecording(FilterRecording):
         self._kwargs = dict(recording=recording, freq_max=freq_max, margin_ms=margin_ms, dtype=dtype.str)
         self._kwargs.update(filter_kwargs)
 
+
 class NotchFilterRecording(BasePreprocessor):
     """
     Parameters
@@ -357,6 +358,7 @@ bandpass_filter = define_function_from_class(source_class=BandpassFilterRecordin
 notch_filter = define_function_from_class(source_class=NotchFilterRecording, name="notch_filter")
 highpass_filter = define_function_from_class(source_class=HighpassFilterRecording, name="highpass_filter")
 lowpass_filter = define_function_from_class(source_class=LowpassFilterRecording, name="lowpass_filter")
+
 
 def causal_filter(
     recording,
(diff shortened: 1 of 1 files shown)


yger:
This is an interesting feature, but be careful about the "magic" parameter margin_ms, that should be increased for small low pass frequencies. Indeed, this is the margin that will be consider to avoid border effect while filtering a chunk. While this parameter of 5ms is rather safe for high frequencies typically used in spike sorting, it might be too small for larger frequencies if you can for tens of Herz. (and then, the advanges of chunk based filtering might be lost also, because margin will be as long if not longer than chunk themselves). In fact, I think it would be good if margin_ms could be set automatically, to avoid any misusage by users. Why not setting it automatically to the lowest bounds of the frequencies considered ?

jesusdpa1:
Thanks @yger! I think I understand but just to make sure you mean setting margin_ms to 

```py
fs = recording.get_sampling_frequency()
margin_ms = (1 / (2.5*freq_max)) * fs * 1000
```

yger:
Not sure your formula is correct. To me, the limiting factor is freq_min (for bandpass, so maybe freq_max for your lowpass), because for lower frequencies, if you want the filtfilt procedure to be accurate without any border effects, you need longer margin. So something like int(fs/(2.5*freq_min)) ? This should be discussed, I'm not saying we should do that, but maybe otherwise throw a warning if there seems to be a big mismatch, to warn the user that filters, at the borders (especially if chunk_size is small) will be inaccurate. What do you think @samuelgarcia @alejoe91 ?

samuelgarcia:
Unfortunatly because we are doing chunk processing every where in spikeinterface we cannot support lowpass filter like this.

In short, for a real DC recorded signals, how we could make a lowpass with cutoff  0.5Hz when the chunk is 1s.
The result would be totally wrong unless very very long margin. The best is to not support it to avoid mistake, I think.

alejoe91:
Yeah I agree with Sam. The closest to a lowpass filter will be a bandpass with a very very low freq min (e.g. 0.001)!

jesusdpa1:
Sounds good! I'll apply low pass outside and bring the recording back into spike interface if it makes sense. I do see that this was mentioned "when the chunk is 1s." keeping this function outside of spikeinterface, is it possible to adjust this value `[chunk size]` so I can apply lowpass filter with the correct chunk size?

zm711:
I'll close this since it won't work with the chunking, but appreciate the the PR!

PR #3671, opened 2025-02-05
Remove mmap mode in numpy folder

alejoe91:
Could replace #3664

### DIFF
--- a/src/spikeinterface/core/sortingfolder.py
+++ b/src/spikeinterface/core/sortingfolder.py
@@ -37,7 +37,7 @@ class NumpyFolderSorting(BaseSorting):
 
         BaseSorting.__init__(self, sampling_frequency, unit_ids)
 
-        self.spikes = np.load(folder_path / "spikes.npy", mmap_mode="r")
+        self.spikes = np.load(folder_path / "spikes.npy")
 
         for segment_index in range(num_segments):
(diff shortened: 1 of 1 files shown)


zm711:
This is also a valid solution. But I think Sam needs to comment on if he thinks the memmap is needed in which case we need to use a memmap we make ourselves rather than a numpy version (I think). That's what he and Heberto worked on somewhere. Or we just load the file itself and not work about memmapping. But then what is the difference between using the numpy and the npz? file format only?

alejoe91:
I made this with @samuelgarcia :)

samuelgarcia:
Hi Zach. This PR was my suggestion.
We removed th memmap loading because when the vector is small then the memmaping make too much annoying effect for very few benefit. For loadin waveforms this would be different.

The old NPZ format is very no efficient because it is a per segment a dictionary of spiketrain.
When implemented the spike vector  we switched to the new format to be more efficient on loading. We should not go back top npz. @yger  is more lazy than me sometimes.

zm711:
Okay as long as you are cool with not-memmapping in general. I thought maybe you wanted to keep the memmap vs dictionary options, but if it is just small array vs dict as the two options then that makes sense.

alejoe91:
Ok tests pass! We can merge this

zm711 (approved):

Issue #3675, opened 2025-02-06
How to easily extract spike times related to a recording chunk.


Djoels:
Just a question: how can one easily create selection of a time fragment along a recording.
For the recording you can make a call to select_channels and time_slice, to get a chunk of the recording.
However when trying to extract the affiliated spike times to this time_slice, I find I'm having to play with the sampling frequency (given that I don't have a time vector).

Am I missing some elegant way of doing this?

zm711:
I always just do this myself by converting to samples from time (as you said using the sampling_frequency), but maybe @alejoe91 or @samuelgarcia know of a machinery in spikeinterface to easily do this (in which case I might switch myself :) )

chrishalcrow:
Hello! There's a few ways to look at spike trains: `to_spike_vector`, `get_unit_spike_train` and I recently found `spike_vector_to_spike_train`(!). Of these, the `get_unit_spike_train` method has a `return_times` parameter. So if you have a sorting object called `sort` you can do something like

``` python
unit_id = 5
unit_spike_train = sort.get_unit_spike_train(unit_id = unit_id, return_times=True)

t_start = 1
t_end = 4
interval_train = unit_spike_train[(1 < unit_spike_train) & (unit_spike_train < 4)]
```

But if you're working with the `to_spike_vector` output, you'll need to deal with the sampling_frequency yourself.

alejoe91:
Another option is to use the `sorting.frame_slice()` to get a new sorting object. I think we could add a `time_slice`, which automatically does the conversion/checks the registered recording for timestamps. 

@chrishalcrow can you give it a try?

chrishalcrow:
Hey @Djoels , thanks to your comment we've just added a new `time_slice` method for the sorting object, so you can do things like

```
reduced_spike_vector = sorting.time_slice(start_time=60, end_time=120).to_spike_vector()
```

The `reduced_spike_vector`  will still give you the spike times using sample indices, so you'll need to convert at the end, but hopefully this will make life easier!

Djoels:
Wow, thank you very much for this!

PR #3677, opened 2025-02-07
Add `time_slice` method to `BaseSorting`

chrishalcrow:
In response to #3675

This PR allows the user to slice a sorting, so that you can do something like

``` python
reduced_sorting = sorting.time_slice(start_time = 60, end_time=120)
reduced_spike_train = reduced_sorting.to_spike_vector()
```

Code and test were essentially copied and pasted from the `BaseRecording` object's `time_slice` method. Then had to make a new `time_to_sample_index` function for the `BaseSorting`, whose logic was taken from `get_unit_spike_trains` here: 
https://github.com/SpikeInterface/spikeinterface/blob/4e34fc747115e91dde06bf2b42ea6bfe4d42616c/src/spikeinterface/core/basesorting.py#L169
This uses the attached recording if there is one, checks if there's a `_t_start`, and sets `_t_start` to zero if there isn't.

Only works for single-segment sorting objects, which matches the capabilities of `time_slice` for the `BaseRecording` object.

### DIFF
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -472,6 +472,30 @@ class BaseSorting(BaseExtractor):
         )
         return sub_sorting
 
+    def time_slice(self, start_time: float | None, end_time: float) -> BaseSorting:
+        """
+        Returns a new sorting with sliced time. Note that this operation is not in place.
+
+        Parameters
+        ----------
+        start_time : float, optional
+            The start time in seconds. If not provided it is set to 0.
+        end_time : float, optional
+            The end time in seconds. If not provided it is set to the total duration.
+
+        Returns
+        -------
+        BaseSorting
+            A new sorting object with only samples between start_time and end_time
+        """
+
+        assert self.get_num_segments() == 1, "Time slicing is only supported for single segment sortings."
+
+        start_frame = self.time_to_sample_index(start_time) if start_time else None
+        end_frame = self.time_to_sample_index(end_time) if end_time else None
+
+        return self.frame_slice(start_frame=start_frame, end_frame=end_frame)
+
     def get_all_spike_trains(self, outputs="unit_id"):
         """
         Return all spike trains concatenated.
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -501,7 +501,7 @@ class BaseSorting(BaseExtractor):
         Transform time in seconds into sample index
         """
         if self.has_recording():
-            sample_index = self._recording.time_to_sample_index(time)
+            sample_index = self._recording.time_to_sample_index(time, segment_index=segment_index)
         else:
             segment = self._sorting_segments[segment_index]
             t_start = segment._t_start if segment._t_start is not None else 0
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -472,15 +472,15 @@ class BaseSorting(BaseExtractor):
         )
         return sub_sorting
 
-    def time_slice(self, start_time: float | None, end_time: float) -> BaseSorting:
+    def time_slice(self, start_time: float | None, end_time: float | None) -> BaseSorting:
         """
         Returns a new sorting with sliced time. Note that this operation is not in place.
 
         Parameters
         ----------
-        start_time : float, optional
+        start_time : float | None, default: None
             The start time in seconds. If not provided it is set to 0.
-        end_time : float, optional
+        end_time : float | None, default: None
             The end time in seconds. If not provided it is set to the total duration.
 
         Returns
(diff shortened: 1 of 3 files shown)


src/spikeinterface/core/basesorting.py
```diff
@@ -472,6 +472,43 @@ def frame_slice(self, start_frame, end_frame, check_spike_frames=True):
...
+        end_frame = self.time_to_sample_index(end_time, segment_index=0) if end_time else None
+
+        return self.frame_slice(start_frame=start_frame, end_frame=end_frame)
+
+    def time_to_sample_index(self, time, segment_index=0):
+        """
+        Transform time in seconds into sample index
+        """
+        if self.has_recording():
+            sample_index = self._recording.time_to_sample_index(time)
```
zm711, line 504: Just to be clear should we put the segment_index into this function? to be explicit. Or why are you okay with the implicit segment_index for the if statement?

alejoe91: We should!

alejoe91 (approved):

src/spikeinterface/core/basesorting.py
```diff
@@ -472,6 +472,43 @@ def frame_slice(self, start_frame, end_frame, check_spike_frames=True):
         )
         return sub_sorting
 
+    def time_slice(self, start_time: float | None, end_time: float | None) -> BaseSorting:
+        """
+        Returns a new sorting with sliced time. Note that this operation is not in place.
```
zm711, line 477: Sorry this can definitely be merged but one other minor point of discussion. 

returns a new sorting object sliced by time? I've never heard sliced time... Is this usual. 

Also do we want to say not in place. Basically all of our functions are not in place.. So I don't want to confuse people by calling this out in one function. Why do this?

JoeZiminski:
Hey @chrishalcrow this is really nice, sorry don't have time for a deep look at the moment. Its great that the underlying recording to managee the times (if it exists). If somone has a recordingless sorting, sets some times, and then adds a recording with `register_recording`, what should happen to the time representations? e.g. if they are different should they be forced to be reconciled?

alejoe91:
> Hey @chrishalcrow this is really nice, sorry don't have time for a deep look at the moment. Its great that the underlying recording to managee the times (if it exists). If somone has a recordingless sorting, sets some times, and then adds a recording with `register_recording`, what should happen to the time representations? e.g. if they are different should they be forced to be reconciled?

@JoeZiminski you can't set times for a recordingless sorting (just `t_start`). If a recording is registered, it should take precedence for time management, since recordings have a better representation of time than sortings :)

src/spikeinterface/core/basesorting.py
chrishalcrow: Thanks for the comments - I copied the docstring from `time_slice` in `BaseRecording`, so was just prioritising consistency with that. I guess the slicing is python terminology. Happy to change both. It's quite hard to write about - Pynapple uses _epochs_ which I think sound too serious. How about:

"""
Returns a new sorting object, restricted to the time interval `[start_time, end_time]`.
"""

Suggestions?

zm711: I like that much better! 

I think the ultimate consistency for me is to remove set_probe in place (maybe requiring a name change) and then nothing in the library is in place. :) But I think as far the one-liner that is much better!

src/spikeinterface/core/basesorting.py
```diff
@@ -472,6 +472,43 @@ def frame_slice(self, start_frame, end_frame, check_spike_frames=True):
...
+    def time_to_sample_index(self, time, segment_index=0):
+        """
+        Transform time in seconds into sample index
+        """
+        if self.has_recording():
+            sample_index = self._recording.time_to_sample_index(time, segment_index=segment_index)
+        else:
+            segment = self._sorting_segments[segment_index]
+            t_start = segment._t_start if segment._t_start is not None else 0
+            sample_index = round((time - t_start) * self.get_sampling_frequency())
```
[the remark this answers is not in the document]
chrishalcrow: Thanks - the big idea is duplicated but the code is inverted. In `get_unit_spike_trains` you get the spike_times_in_seconds from the spike_times_in_samples, whereas here we get the sample_index from the time_in_seconds_index. So, afraid it's not possible.

alejoe91:
> Makes sense! Is it worth adding a note on the docstring of `register_recording` that the recording will be used for all times once registered? I was thinking a warning but that might be overkill.

Yep that's fair! @chrishalcrow can you add a note? I think it fits well into this PR since it's all.time handling :)

zm711 (approved):
Looks good to me now.

lcui809:
@chrishalcrow This is super helpful—thank you!
I was wondering if it's possible to merge or concatenate time slices from a sorting object. Let's assume the time slices of interest are periodic, and I've already extracted them. Would it be possible to merge those extracted slices into a single object so I can create a sorting analyzer for downstream processing?

chrishalcrow:
Hey @lcui809 if you started with a sorting from one long recording then split it up, you can put it back together again using: `append_sortings` which takes in a list of sortings and makes a multi-segment sorting, then `concatenate_sortings` which takes in a multi-segment sorting and makes a single segment sorting. Both get imported from full, so e.g.

``` python
big_sorting = si.concatenate_sortings(si.append_sortings( [ sorting_0, sorting_1] ))
```

should work. I don't really know why you'd want to split up a sorting, then stitch it back together again. What's your use case??

PR #3683, opened 2025-02-11
Test Python 3.13

alejoe91:


alejoe91:
All good but Windows 3.13 @chrishalcrow

JoeZiminski:
Looks very similar (identical?) to this: https://github.com/numpy/numpy/issues/27894

but the numpy version in use is `numpy-1.26.4`

Oh maybe `1.26.4` does not work on python 3.13 https://numpy.org/devdocs/release/1.26.4-notes.html

I'm not sure why it is resolving to numpy `1.26.4`, it might be worth forcing to `numpy>=2` if python 3.13, this should be possible in the pyproject.toml with [dependency specifiers](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#dependency-specifiers). This stuff always gives me a headache, maybe something like (untested):
```
    "numpy>=1.20" ; python_version<'3.13'",
    "numpy>=2.0.0" ; python_version>='3.13'",
```
(assuming the numpy version is the issue)

chrishalcrow:
What fun! New error is that elephant 1.1.0 imports a depreciated scipy function; if we could get to elephant 1.1.1,  that'd be fixed.

alejoe91:
This is mainly due to MEArec, which uses elephant. Just made a release to have local elephant imports, since it's only used at generation. Let's retrigger and see how it goes :)

alejoe91:
depends on #3689

chrishalcrow (approved):
Well done @alejoe91 , looks good to me!

PR #3685, opened 2025-02-12
Add `RemoveBadChannelsRecording` class and `remove_bad_channels`

chrishalcrow:
Part of plan to `class`-ifty all preprocessing steps. Here's an implementation of a `RemoveBadChannelsRecording` class, inheriting from `ChannelSliceRecording`.

Users can now use
``` python
recording_with_bad_channels_removed = remove_bad_channels(recording)
```
and this will detect and remove the bad channels.

Most difficult thing was the kwargs. `RemoveBadChannelsRecording` shares much of the same signature as `detect_bad_channels`, and I didn't want to duplicate the default parameters, but did want to save them in the `RemoveBadChannelsRecording._kwargs`. Ended up using `inspect` from `signature`, as is also done in the motion module. Happy to discuss different implementations.

### DIFF
--- a/src/spikeinterface/preprocessing/detect_bad_channels.py
+++ b/src/spikeinterface/preprocessing/detect_bad_channels.py
@@ -4,8 +4,122 @@ import warnings
 import numpy as np
 from typing import Literal
 
-from .filter import highpass_filter
-from ..core import get_random_data_chunks, order_channels_by_depth, BaseRecording
+from spikeinterface.core.core_tools import define_function_from_class
+from spikeinterface.preprocessing.filter import highpass_filter
+from spikeinterface.core import get_random_data_chunks, order_channels_by_depth, BaseRecording
+from spikeinterface.core.channelslice import ChannelSliceRecording
+
+from inspect import signature
+
+
+class RemoveBadChannelsRecording(ChannelSliceRecording):
+    """
+    Removes bad channels.
+
+    {}
+
+    Returns
+    -------
+    removed_bad_channels_recording : RemoveBadChannelsRecording
+        The recording with bad channels removed
+    """
+
+    params_doc = """Different methods are implemented:
+
+* std : threhshold on channel standard deviations
+    If the standard deviation of a channel is greater than `std_mad_threshold` times the median of all
+    channels standard deviations, the channel is flagged as noisy
+* mad : same as std, but using median absolute deviations instead
+* coeherence+psd : method developed by the International Brain Laboratory that detects bad channels of three types:
+    * Dead channels are those with low similarity to the surrounding channels (n=`n_neighbors` median)
+    * Noise channels are those with power at >80% Nyquist above the psd_hf_threshold (default 0.02 uV^2 / Hz)
+        and a high coherence with "far away" channels"
+    * Out of brain channels are contigious regions of channels dissimilar to the median of all channels
+        at the top end of the probe (i.e. large channel number)
+* neighborhood_r2
+    A method tuned for LFP use-cases, where channels should be highly correlated with their spatial
+    neighbors. This method estimates the correlation of each channel with the median of its spatial
+    neighbors, and considers channels bad when this correlation is too small.
+
+Parameters
+----------
+recording : BaseRecording
+    The recording for which bad channels are detected
+method : "coeherence+psd" | "std" | "mad" | "neighborhood_r2", default: "coeherence+psd"
+        The method to be used for bad channel detection
+std_mad_threshold : float, default: 5
+    The standard deviation/mad multiplier threshold
+psd_hf_threshold : float, default: 0.02
+    For coherence+psd - an absolute threshold (uV^2/Hz) used as a cutoff for noise channels.
+    Channels with average power at >80% Nyquist larger than this threshold
+    will be labeled as noise
+dead_channel_threshold : float, default: -0.5
+    For coherence+psd - threshold for channel coherence below which channels are labeled as dead
+noisy_channel_threshold : float, default: 1
+    Threshold for channel coherence above which channels are labeled as noisy (together with psd condition)
+outside_channel_threshold : float, default: -0.75
+    For coherence+psd - threshold for channel coherence above which channels at the edge of the recording are marked as outside
+    of the brain
+outside_channels_location : "top" | "bottom" | "both", default: "top"
+    For coherence+psd - location of the outside channels. If "top", only the channels at the top of the probe can be
+    marked as outside channels. If "bottom", only the channels at the bottom of the probe can be
+    marked as outside channels. If "both", both the channels at the top and bottom of the probe can be
+    marked as outside channels
+n_neighbors : int, default: 11
+    For coeherence+psd - number of channel neighbors to compute median filter (needs to be odd)
+nyquist_threshold : float, default: 0.8
+    For coherence+psd - frequency with respect to Nyquist (Fn=1) above which the mean of the PSD is calculated and compared
+    with psd_hf_threshold
+direction : "x" | "y" | "z", default: "y"
+    For coherence+psd - the depth dimension
+highpass_filter_cutoff : float, default: 300
+    If the recording is not filtered, the cutoff frequency of the highpass filter
+chunk_duration_s : float, default: 0.5
+    Duration of each chunk
+num_random_chunks : int, default: 100
+    Number of random chunks
+    Having many chunks is important for reproducibility.
+welch_window_ms : float, default: 10
+    Window size for the scipy.signal.welch that will be converted to nperseg
+neighborhood_r2_threshold : float, default: 0.95
+    R^2 threshold for the neighborhood_r2 method.
+neighborhood_r2_radius_um : float, default: 30
+    Spatial radius below which two channels are considered neighbors in the neighborhood_r2 method.
+seed : int or None, default: None
+    The random seed to extract chunks
+
+    """
+
+    def __init__(
+        self,
+        recording: BaseRecording,
+        **detect_bad_channels_kwargs,
+    ):
+
+        # get the default parameters from `detect_bad_channels`, and update with any user-specified parameters.
+        sig = signature(detect_bad_channels)
+        updated_detect_bad_channels_kwargs = {k: v.default for k, v in sig.parameters.items() if k != "recording"}
+        updated_detect_bad_channels_kwargs.update(detect_bad_channels_kwargs)
+
+        bad_channel_ids, channel_labels = detect_bad_channels(recording=recording, **updated_detect_bad_channels_kwargs)
+
+        self._main_ids = recording.get_channel_ids()
+        new_channel_ids = self.channel_ids[~np.isin(self.channel_ids, bad_channel_ids)]
+
+        ChannelSliceRecording.__init__(
+            self,
+            recording,
+            channel_ids=new_channel_ids,
+        )
+
+        self._kwargs.update({"channel_labels": channel_labels})
+        self._kwargs.update(updated_detect_bad_channels_kwargs)
+
+
+remove_bad_channels = define_function_from_class(source_class=RemoveBadChannelsRecording, name="remove_bad_channels")
+
+
+RemoveBadChannelsRecording.__doc__ = RemoveBadChannelsRecording.__doc__.format(RemoveBadChannelsRecording.params_doc)
 
 
 def detect_bad_channels(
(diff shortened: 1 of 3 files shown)


JoeZiminski (approved):
This looks good to me, few suggestions / questions. Really love this as it solves a big problem for me!

src/spikeinterface/preprocessing/detect_bad_channels.py
```diff
@@ -4,8 +4,122 @@
...
+        recording: BaseRecording,
+        **detect_bad_channels_kwargs,
+    ):
+
+        # get the default parameters from `detect_bad_channels`, and update with any user-specified parameters.
+        sig = signature(detect_bad_channels)
+        updated_detect_bad_channels_kwargs = {k: v.default for k, v in sig.parameters.items() if k != "recording"}
+        updated_detect_bad_channels_kwargs.update(detect_bad_channels_kwargs)
+
+        bad_channel_ids, channel_labels = detect_bad_channels(recording=recording, **updated_detect_bad_channels_kwargs)
```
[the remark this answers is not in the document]
chrishalcrow: Yes, your right, and seems safer, thanks!

src/spikeinterface/preprocessing/tests/test_detect_bad_channels.py
```diff
@@ -18,6 +20,26 @@
...
+    recording = generate_recording(durations=[5, 6], seed=1205, num_channels=8)
+    recording.set_channel_offsets(0)
+    recording.set_channel_gains(1)
+    # set noisy_channel_threshold so that we do detect some bad channels
+    new_rec = remove_bad_channels(recording, noisy_channel_threshold=0)
+
+    # make sure they are removed
+    assert len(set(new_rec._kwargs["bad_channel_ids"]).intersection(new_rec.channel_ids)) == 0
+    # and that the kwarg is propogatged to the kwargs of new_rec.
+    assert new_rec._kwargs["noisy_channel_threshold"] == 0
```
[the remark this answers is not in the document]
chrishalcrow: The ordering of the channel ids?

src/spikeinterface/preprocessing/detect_bad_channels.py
```diff
@@ -32,69 +146,7 @@ def detect_bad_channels(
...
-        Having many chunks is important for reproducibility.
-    welch_window_ms : float, default: 10
-        Window size for the scipy.signal.welch that will be converted to nperseg
-    neighborhood_r2_threshold : float, default: 0.95
-        R^2 threshold for the neighborhood_r2 method.
-    neighborhood_r2_radius_um : float, default: 30
-        Spatial radius below which two channels are considered neighbors in the neighborhood_r2 method.
-    seed : int or None, default: None
-        The random seed to extract chunks
+    {}
```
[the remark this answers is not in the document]
chrishalcrow: I think it's quite hard to do this, since we can't add comments to docstrings. The pattern is used a fair bit in the package, and I do remember getting confused. Hmmm.

src/spikeinterface/preprocessing/detect_bad_channels.py
```diff
@@ -4,8 +4,122 @@
...
+
+    {}
+
+    Returns
+    -------
+    removed_bad_channels_recording : RemoveBadChannelsRecording
+        The recording with bad channels removed
+    """
+
+    params_doc = """Different methods are implemented:
```
alejoe91, line 27: I would move this outside of the class and just call it `_bad_channel_detection_kwargs_doc`

src/spikeinterface/preprocessing/detect_bad_channels.py
```diff
@@ -4,8 +4,122 @@
...
+        recording: BaseRecording,
+        **detect_bad_channels_kwargs,
+    ):
+
+        # get the default parameters from `detect_bad_channels`, and update with any user-specified parameters.
+        sig = signature(detect_bad_channels)
+        updated_detect_bad_channels_kwargs = {k: v.default for k, v in sig.parameters.items() if k != "recording"}
+        updated_detect_bad_channels_kwargs.update(detect_bad_channels_kwargs)
+
+        bad_channel_ids, channel_labels = detect_bad_channels(recording=recording, **detect_bad_channels_kwargs)
```
alejoe91, line 104:
```suggestion
       bad_channel_ids = detect_bad_channels_kwargs.get("bad_channel_ids")
        if bad_channel_ids is None:
            bad_channel_ids, channel_labels = detect_bad_channels(recording=recording, **detect_bad_channels_kwargs)
```
We don't have to rerun bad channel detecion if bad channels have already been computed. Otherwise, if we parallelize this then each process/thread will rerun detection.

We should also keep the `ProcessingPipeline` in mind. In case we have a processing pipeline, we should differentiate somehow the lazy/non-lazy computation steps.
An option could be to extend the `self._kwargs` mechanism, and add a `self._precomputed_kwargs` (in this case this would have the `bad_channel_ids`).
When loading an object, we could then specify whether we want to use or ignore the precomputed kwargs. When applying the pipeline to another object, you would ignore the precomputed, so that channel detection will be repreformed on the new recording.

Does it make sense?

chrishalcrow: Oh you're very right, great point. For the `ProcessingPipeline` I was thinking that the main distinction is if the parameter is recording-specific or not. If the kwarg (e.g. `bad_channel_ids` or `whitening_matrix` - these might be the only examples??) depends on which recording it's applied to, it shouldn't be used by the pipeline. I reckon this is the exact same set of kwargs that could be precomputed, since you'll always need a recording to do a precomputation?

chrishalcrow: `_kwargs` are handled by the subclasses, so would we need to add `_precomputed_kwargs` to all the preprocessing subclasses?

alejoe91: This could be done in the base object. It will need to be there because we will have to handle it upon loading and give the option to add them to the main _kwargs (default), or skip them. @samuelgarcia what do you think?

chrishalcrow: Oh no, we could use `hasattr` to check if a preprocessing class has possible `_precomputed_kwargs`, nice.

alejoe91: yes, that is actually probably the best option. But: if we save to json, we need to somehow merge them with the _kwargs, otherwise they get lost!

JoeZiminski: In this case, what happens if someone runs the function, then changes the kwargs (e.g. maybe set them stricter to remove more channels)? would nothing happen on the second call?

src/spikeinterface/preprocessing/detect_bad_channels.py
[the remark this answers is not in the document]
JoeZiminski: Although, I kind of withdraw comment 😅 as I didn't realise you need the full dict to save the kwargs later. Maybe in the end better to keep the same dict passed as `kwargs` and stored as `kwargs` just in case, maybe add a quick comment here to explain why this approach is taken.

src/spikeinterface/preprocessing/detect_bad_channels.py
JoeZiminski: sorry that wasn't clear I meant the kwargs on the public function like:

```
chan_removed_rec = remove_bad_channels(recording, noisy_channel_threshold=5)

# do some plotting or something to check the remove channels, and decide want to remove a few more

chan_removed_rec = remove_bad_channels(chan_removed_rec, noisy_channel_threshold=1)

```
Ideally people would pass in the original recording but I don't think the above scenario would be so unlikely. Can I warning be thrown?

chrishalcrow: Yeah, having `bad_channel_ids` as a separate argument sounds ok to me.

I think your example wouldn't cause a problem because they are two independent `RemoveBadChannelRecording` classes each with their own kwargs.  Will check...

The difficult bit is that there are two things we want. First, if you load a recording from the `recording.json` file, it should quickly be able to reconstruct the preprocessed recording. For this to work and be fast, we need to save the `bad_channel_ids` in the kwargs and apply them when we load the recording, rather than re-computing.

On the other hand, thinking about the PreprocessingPipeline, we want people to be able to share their pipelines and apply a pipeline from another lab to their own data. To do this, we need to be able to share which kwargs were used to detect the bad channels, but *not* the bad channels themselves; since different recordings have different bad channels.

So we need to mark out that as a special kwarg: apply it when you load the recording, but don't add it to a PreprocessingPipeline. Does that make sense?

chrishalcrow: To summarise the problem (and clarify in my head...)

The difficult bit is that there are two things we want. First, if you load a recording from the `recording.json` file, it should quickly be able to reconstruct the preprocessed recording. For this to work and be fast, we need to save the `bad_channel_ids` in the kwargs and apply them when we load the recording, rather than re-computing.

On the other hand, thinking about the PreprocessingPipeline, we want people to be able to share their pipelines and apply a pipeline from another lab to their own data. To do this, we need to be able to share which kwargs were used to detect the bad channels, but *not* the bad channels themselves; since different recordings have different bad channels.

So we need to mark out that as a special kwarg: apply it when you load the recording, but don't add it to a PreprocessingPipeline. Does that make sense?

alejoe91: Yes that makes total sense

src/spikeinterface/preprocessing/detect_bad_channels.py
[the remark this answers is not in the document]
h-mayorquin: https://github.com/matplotlib/matplotlib/issues/9414

Relevant

h-mayorquin:
> We need a strong refactoring to separate kwargs in 2 parts, the parametrsation and caching some heavy computation. This is on my todo list since 3 years now... Lets do it before.

+1 on discussing this. 

Sam mentioned it on the last meeting. Would be good to bring it back.

alejoe91:
> Just a quick thought on this, there is the [interpolate_bad_channels](https://github.com/SpikeInterface/spikeinterface/blob/17c3d328f1c039e9cb70b576086eb19beb2c4ffb/src/spikeinterface/preprocessing/interpolate_bad_channels.py#L115) class, but that takes a list of channel ids. Given the function in this PR (and in general) you might expect it to detect-and-interpolate bad channels. I wonder if it makes sense to change that to `interpolate_channels()` (similar to `recording.remove_channels()` and then have `remove_bad_channels` and `interpolate_bad_channels` as convenience functions to detect-and-remove. However, this might be a confusing change re: backwards compatibility.

Yes it makes total sense @JoeZiminski 

What about having two classes: 
- `DetectAndRemoveBadChannels` (function `detect_and_remove_bad_channels`)
- `DetectAndInterpolateBadChannel` (function: `detect_and_interpolate_bad_channels`)

They are a bit verbose, but it binds the `bad_channel_ids` of the remove/interpolate part to the detection.

@chrishalcrow what do yo think?

chrishalcrow:
> > Just a quick thought on this, there is the [interpolate_bad_channels](https://github.com/SpikeInterface/spikeinterface/blob/17c3d328f1c039e9cb70b576086eb19beb2c4ffb/src/spikeinterface/preprocessing/interpolate_bad_channels.py#L115) class, but that takes a list of channel ids. Given the function in this PR (and in general) you might expect it to detect-and-interpolate bad channels. I wonder if it makes sense to change that to `interpolate_channels()` (similar to `recording.remove_channels()` and then have `remove_bad_channels` and `interpolate_bad_channels` as convenience functions to detect-and-remove. However, this might be a confusing change re: backwards compatibility.
> 
> Yes it makes total sense @JoeZiminski
> 
> What about having two classes:
> 
> * `DetectAndRemoveBadChannels` (function `detect_and_remove_bad_channels`)
> * `DetectAndInterpolateBadChannel` (function: `detect_and_interpolate_bad_channels`)
> 
> They are a bit verbose, but it binds the `bad_channel_ids` of the remove/interpolate part to the detection.
> 
> @chrishalcrow what do yo think?

I think the verboseness makes it very clear that the function is dong two things, which is great! Thanks for the suggestion @JoeZiminski 

So you'd use `detect_bad_channel` when you're playing with parameters and getting things going, then use `detect_and_remove_bad_channels` once your pipeline is a bit more settled. I think it's a nice workflow.

chrishalcrow:
Hello @JoeZiminski and @alejoe91 - I've added a `detect_and_interpolate_bad_channels` function. At the moment, this does a very similar thing to `detect_and_remove_bad_channels` but calls the `InterpolateBadChannels` class rather than `ChannelSlice`.

chrishalcrow:
> Nice! That looks great @chrishalcrow cheers this will be super useful. I had a quick look at the code and looks good, I've got a slight preference to refactor the shared code:
> 
> ```
> 
>         # get the default parameters from `detect_bad_channels`, and update with any user-specified parameters.
>         sig = signature(detect_bad_channels)
>         updated_detect_bad_channels_kwargs = {k: v.default for k, v in sig.parameters.items() if k != "recording"}
>         updated_detect_bad_channels_kwargs.update(detect_bad_channels_kwargs)
> 
>         if bad_channel_ids is None:
>             bad_channel_ids, channel_labels = detect_bad_channels(recording=recording, **detect_bad_channels_kwargs)
>         else:
>             channel_labels = None
> 
>         self._main_ids = recording.get_channel_ids()
> ```
> 
> into a new function on `detect_bad_channels.py` which is shared between both, but not very strong on that.

I've refactored (and moved) the gross `updated_detect_bad_channels_kwargs` code. I think the other bits are helpful to keep for understandability.

src/spikeinterface/preprocessing/detect_bad_channels.py
chrishalcrow: Hello, I think this is sorted now. We've added `_precomputable_kwarg_names = ["bad_channel_ids"]` and as a class property, and we'll use this on `load` or when making a `pipeline` to include/not include the bad channel ids.

Note: when you load a saved file from dict, it makes an e.g. `DetectAndRemoveBadChannelsRecording` but skips the detection stage, since it knows the `bad_channel_ids`. Hence you can load this recording very fast!

chrishalcrow:
Hello, I think this is ready @alejoe91

src/spikeinterface/preprocessing/interpolate_bad_channels.py
```diff
@@ -82,6 +88,58 @@ def check_inputs(self, recording, bad_channel_ids):
...
+        recording: BaseRecording,
+        bad_channel_ids=None,
+        **detect_bad_channels_kwargs,
+    ):
+        if bad_channel_ids is None:
+            bad_channel_ids, channel_labels = detect_bad_channels(recording=recording, **detect_bad_channels_kwargs)
+        else:
+            channel_labels = None
+
+        self._main_ids = recording.get_channel_ids()
```
alejoe91, line 119: Why is this needed? The `InterpolateBadChannelsRecording` should call the `BasePreprocessor` init which does that automatically

chrishalcrow: Thanks - you're right! It _is_ needed when you `remove_bad_channels`, because `ChannelSliceRecording` needs the list of "good" channels. To make these we get the difference between all units (generated by `self.channel_ids` which is a method which basically just returns `self._main_ids`) and the bad units.

But yeah, not needed for interpolate.

chrishalcrow:
Hello, I now _really_ think this is ready. @alejoe91 
Docs failing because of gin. None of my doc changes actually get run.

doc/modules/preprocessing.rst
```diff
@@ -247,6 +247,18 @@ interpolated with the :code:`interpolate_bad_channels()` function (channels labe
...
+functions to do everything at once. These return a Preprocessor class, so are consistent with
+the "chain" concept for this module. For example:
+
+.. code-block:: python
+
+    # detect and remove bad channels
+    rec_only_good_channels = detect_and_remove_bad_channels(recording=rec)
+
+    # detect and interpolate the bad channels
+    rec_interpolated_channels = detect_and_interpolate_bad_channels(remove_channel_ids=bad_channel_ids)
```
alejoe91:
```suggestion
    rec_interpolated_channels = detect_and_interpolate_bad_channels(recording=rec)
```

alejoe91:
@chrishalcrow looks perfect! Just something of fin the docs and the merge conflict and ready to merge on my end

src/spikeinterface/preprocessing/detect_bad_channels.py
```diff
@@ -4,8 +4,148 @@
...
+        If given, these are labels given to the channels by the
+        detection process. Only intended for use when loading.
+
+    Returns
+    -------
+    removed_bad_channels_recording : DetectAndRemoveBadChannelsRecording
+        The recording with bad channels removed
+    """
+
+    _precomputable_kwarg_names = ["bad_channel_ids"]
```
samuelgarcia:
```suggestion
    _precomputable_kwarg_names = ["bad_channel_ids", "channel_labels"]
```

alejoe91: @chrishalcrow should channel_labels be here too?

chrishalcrow: Yess, we'd need them to reproduce the exact preprocessed recording. Will need to test that this works ok when they are `None` or `not None`  in e.g. #3438

PR #3693, opened 2025-02-14
Peak SVD motion extraction

samuelgarcia:
This implements sparse svd decompostion which has the option of "motion aware".
The decompostion are interpolated back to stable position using cubic interpolation.
This should help the clustering for SC2 or TDC2 in case of motion.

@yger

Note:
 * To achieve this a small hack to node_pipeline has been done optionaly for node computation to be aware of the true time.
 * a small bugfix also in `correct_motion_on_peaks()` when using multi segments.

### DIFF
--- a/src/spikeinterface/core/node_pipeline.py
+++ b/src/spikeinterface/core/node_pipeline.py
@@ -29,6 +29,11 @@ spike_peak_dtype = base_peak_dtype + [
 
 
 class PipelineNode:
+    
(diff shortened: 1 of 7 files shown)


samuelgarcia:
@yger ready to review

yger:
I'll review it, but tests are not passing

alejoe91:
> I'll review it, but tests are not passing

it's pre-commit :) Sam needs to allow edits eheheh

yger (approved):

PR #3694, opened 2025-02-14
Add `SortingAnalyzer.set_sorting_property()/get_sorting_property()` functions

alejoe91:
Helper function to add properties to `SortingAnalyzer.sorting` and make them persistent to disk

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -231,10 +231,10 @@ class BaseExtractor:
         ids : list/np.array, default: None
             List of subset of ids to set the values, default: None
             if None which is the default all the ids are set or changed
-        missing_value : object, default: None
+        missing_value : Any, default: None
             In case the property is set on a subset of values ("ids" not None),
-            it specifies the how the missing values should be filled.
-            The missing_value has to be specified for types int and unsigned int.
+            This argument specifies how to fill missing values.
+            The `missing_value` is required for types int and unsigned int.
         """
 
         if values is None:
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -145,7 +145,7 @@ class BaseExtractor:
             non_existent_ids = [id for id in ids if id not in self._main_ids]
             if non_existent_ids:
                 error_msg = (
-                    f"IDs {non_existent_ids} are not channel ids of the extractor. \n"
+                    f"IDs {non_existent_ids} are not ids of the extractor. \n"
                     f"Available ids are {self._main_ids} with dtype {self._main_ids.dtype}"
                 )
                 raise ValueError(error_msg)
@@ -293,6 +293,11 @@ class BaseExtractor:
                     ), f"Mismatch between existing property dtype {existing_property.kind} and provided values dtype {dtype_kind}."
 
                 indices = self.ids_to_indices(ids)
+                if dtype_kind == "U":
+                    # re-adjust the size of the property
+                    existing_unicode_size = max(len(s) for s in self._properties[key])
+                    new_unicode_size = max(max(len(s) for s in values), existing_unicode_size)
+                    self._properties[key] = self._properties[key].astype(f"<U{new_unicode_size}")
                 self._properties[key][indices] = values
             else:
                 indices = self.ids_to_indices(ids)
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from typing import Literal, Optional
+from typing import Literal, Optional, Any
 
 from pathlib import Path
 from itertools import chain
@@ -741,6 +741,53 @@ class SortingAnalyzer:
         if self._recording is not None:
             warnings.warn("SortingAnalyzer recording is already set. The current recording is temporarily replaced.")
         self._temporary_recording = recording
+
+    def set_unit_property(
+        self,
+        key,
+        values: list | np.ndarray | tuple,
+        ids: list | np.ndarray | tuple | None = None,
+        missing_value: Any = None,
+    ) -> None:
+        """
+        Set property vector for unit ids.
+
+        If the SortingAnalyzer backend is in memory, the property will be only set in memory.
+        If the SortingAnalyzer backend is in binary_folder or zarr, the property will also
+        be saved to to the backend.
+
+        Parameters
+        ----------
+        key : str
+            The property name
+        values : np.array
+            Array of values for the property
+        ids : list/np.array, default: None
+            List of subset of ids to set the values, default: None
+            if None which is the default all the ids are set or changed
+        missing_value : object, default: None
+            In case the property is set on a subset of values ("ids" not None),
+            it specifies the how the missing values should be filled.
+            The missing_value has to be specified for types int and unsigned int.
+        """
+        self.sorting.set_property(key, values, ids=ids, missing_value=missing_value)
+        if not self.is_read_only():
+            if self.format == "binary_folder":
+                np.save(self.folder / "sorting" / "properties" / f"{key}.npy", self.sorting.get_property(key))
+            elif self.format == "zarr":
+                import zarr
+
+                zarr_root = self._get_zarr_root(mode="r+")
+                prop_values = self.sorting.get_property(key)
+                if prop_values.dtype.kind == "O":
+                    warnings.warn(f"Property {key} not saved because it is a python Object type")
+                else:
+                    if key in zarr_root["sorting"]["properties"]:
+                        zarr_root["sorting"]["properties"][key][:] = prop_values
+                    else:
+                        zarr_root["sorting"]["properties"].create_dataset(name=key, data=prop_values, compressor=None)
+                    # IMPORTANT: we need to re-consolidate the zarr store!
+                    zarr.consolidate_metadata(zarr_root.store)
 
     def _save_or_select_or_merge(
         self,
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -223,13 +223,13 @@ class SortingAnalyzer:
 
     def __init__(
         self,
-        sorting=None,
-        recording=None,
-        rec_attributes=None,
-        format=None,
-        sparsity=None,
-        return_scaled=True,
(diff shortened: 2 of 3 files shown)


src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -742,6 +742,53 @@ def set_temporary_recording(self, recording: BaseRecording, check_dtype: bool =
...
+        key,
+        values: list | np.ndarray | tuple,
+        ids: list | np.ndarray | tuple | None = None,
+        missing_value: Any = None,
+    ) -> None:
+        """
+        Set property vector for unit ids.
+
+        If the SortingAnalyzer backend is in memory, the property will be only set in memory.
+        If the SortingAnalyzer backend is in binary_folder or zarr, the property will also
```
zm711, line 756:
```suggestion
        If the SortingAnalyzer backend is `binary_folder` or `zarr`, the property will also
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -742,6 +742,53 @@ def set_temporary_recording(self, recording: BaseRecording, check_dtype: bool =
...
+        be saved to to the backend.
+
+        Parameters
+        ----------
+        key : str
+            The property name
+        values : np.array
+            Array of values for the property
+        ids : list/np.array, default: None
+            List of subset of ids to set the values, default: None
```
zm711, line 766:
```suggestion
            List of subset of ids to set the values.
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -742,6 +742,53 @@ def set_temporary_recording(self, recording: BaseRecording, check_dtype: bool =
...
+            if None which is the default all the ids are set or changed
```
zm711, line 767:
```suggestion
            if None all the ids are set or changed
```
that's definitely an an assisted docstring ;)

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -742,6 +742,53 @@ def set_temporary_recording(self, recording: BaseRecording, check_dtype: bool =
...
+        missing_value : object, default: None
+            In case the property is set on a subset of values ("ids" not None),
+            it specifies the how the missing values should be filled.
+            The missing_value has to be specified for types int and unsigned int.
```
zm711, line 771:
```suggestion
           This argument specifies how to fill missing values
            The `missing_value` has to be specified for types int and unsigned int.
```
Is this true? It has to be int?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -742,6 +742,53 @@ def set_temporary_recording(self, recording: BaseRecording, check_dtype: bool =
...
+        if not self.is_read_only():
+            if self.format == "binary_folder":
+                np.save(self.folder / "sorting" / "properties" / f"{key}.npy", self.sorting.get_property(key))
+            elif self.format == "zarr":
+                import zarr
+
+                zarr_root = self._get_zarr_root(mode="r+")
+                prop_values = self.sorting.get_property(key)
+                if prop_values.dtype.kind == "O":
+                    warnings.warn(f"Property {key} not saved because it is a python Object type")
```
zm711, line 783: Can we tell the user how to fix it or is this just a you're out of luck?

src/spikeinterface/core/tests/test_sortinganalyzer.py
```diff
@@ -66,6 +66,12 @@ def test_SortingAnalyzer_memory(tmp_path, dataset):
     )
     assert not sorting_analyzer.return_scaled
 
+    # test set_unit_property
+    sorting_analyzer.set_unit_property(key="quality", values=["good"] * len(sorting_analyzer.unit_ids))
```
zm711, line 70: because values is a string here so why does missing value have to be an int?

zm711 (commented):
can we get a test for the missing_value argument?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -742,6 +742,53 @@ def set_temporary_recording(self, recording: BaseRecording, check_dtype: bool =
             warnings.warn("SortingAnalyzer recording is already set. The current recording is temporarily replaced.")
         self._temporary_recording = recording
 
+    def set_unit_property(
+        self,
+        key,
+        values: list | np.ndarray | tuple,
+        ids: list | np.ndarray | tuple | None = None,
+        missing_value: Any = None,
```
samuelgarcia, line 750: I would explicitly make a `save=True/False`, option here.

src/spikeinterface/core/sortinganalyzer.py
alejoe91: I think this is by construction, so we can re-open the zarr folder also outside of Python

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -742,6 +742,53 @@ def set_temporary_recording(self, recording: BaseRecording, check_dtype: bool =
...
+        Parameters
+        ----------
+        key : str
+            The property name
+        values : np.array
+            Array of values for the property
+        ids : list/np.array, default: None
+            List of subset of ids to set the values, default: None
+            if None which is the default all the ids are set or changed
+        missing_value : object, default: None
```
alejoe91, line 768:
```suggestion
        missing_value : Any, default: None
```

src/spikeinterface/core/tests/test_sortinganalyzer.py
alejoe91: it doesn't have to be an int. It is required if the values are integers. I changed the docstrings to make it more clear

chrishalcrow:
Hello, found a bug:

``` python
rec, sort = si.generate_ground_truth_recording()
sa.set_unit_property(key="quality", values=["bad"]*10)
sa.set_unit_property(key="quality", values=["good","good","good"], ids=['1','7','8'])
print(sa.get_sorting_property("quality"))
>>> array(['bad', 'goo', 'bad', 'bad', 'bad', 'bad', 'bad', 'goo', 'goo',
       'bad'], dtype='<U3')
```

It's saving the labels as `U3` then using this later, so that only the first three chracters of `'good'` are saved.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -742,6 +743,56 @@ def set_temporary_recording(self, recording: BaseRecording, check_dtype: bool =
             warnings.warn("SortingAnalyzer recording is already set. The current recording is temporarily replaced.")
         self._temporary_recording = recording
 
+    def set_unit_property(
```
chrishalcrow, line 746: Should we use `set_sorting_property` to match `get_sorting_property`? I prefer get/set_unit_... but we've already used sorting.

chrishalcrow:
The error message when you put in a non_existent id mentions `channel_ids`, which might be confusing:

e.g. forgetting that generated ids are strings...
``` python
sa.set_unit_property(key="beep", values=[5,4,3,2,1], ids=[1,2,3,4,5])
>>> ValueError: IDs [np.int64(1), np.int64(2), np.int64(3), np.int64(4), np.int64(5)] are not channel ids of the extractor.
Available ids are ['0' '1' '2' '3' '4' '5' '6' '7' '8' '9'] with dtype <U1
```

(The fact this dtype is <U1 looks a big dangerous!!)

Error message here:
https://github.com/SpikeInterface/spikeinterface/blob/91c2a0d8f57bd5b8ef999e1dc9452643acac4f83/src/spikeinterface/core/base.py#L148

src/spikeinterface/core/sortinganalyzer.py
alejoe91: makes sense :)

alejoe91:
@chrishalcrow I fixed the unicode size issue in the core `get_property` by resizing the dtype according to existing and new values.

Also renamed to `set_sorting_property`, added `get_sorting_property`, and removed reference to "channel IDs" in the error message.

Good to go on my end!

chrishalcrow (approved):
Thanks Alessio - really helpful feature! This
``` python
sa.select_units(sa.unit_ids[sa.get_sorting_property('brain_area')=='MEC'])
```
feels good!!

All working for me :)

zm711 (approved):
Thanks the docs are clearer now!

alejoe91:
Also fixes #3699 @chrishalcrow

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1186,11 +1259,16 @@ def get_sorting_provenance(self):
...
-            else:
-                sorting_provenance = None
+                # try-except here is because it's not required to be able
+                # to load the sorting provenance, as the user might have deleted
+                # the original sorting folder
+                try:
+                    sort_dict = zarr_root["sorting_provenance"][0]
+                    sorting_provenance = load(sort_dict, base_folder=self.folder)
+                except:
+                    pass
```
samuelgarcia, line 1271: I think we should raise an error when the user really wants to load this no ?
I am not sure to like theses try except at many place when dealing with zarr

samuelgarcia:
Nobody is giving me a chance to read teh PR before the merge now...

zm711:
Is it finally time for us to have a ready-to-merge tag?

alejoe91:
> Is it finally time for us to have a ready-to-merge tag?

maybe! but this is just @samuelgarcia with his daily dose of complaints :)

src/spikeinterface/core/base.py
```diff
@@ -293,6 +293,11 @@ def set_property(
                     ), f"Mismatch between existing property dtype {existing_property.kind} and provided values dtype {dtype_kind}."
 
                 indices = self.ids_to_indices(ids)
+                if dtype_kind == "U":
```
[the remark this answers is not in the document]
zm711: So your idea here would be no matter what annotation as long as it is a string it can be combined and then you'll resize? I guess that makes sense right. All strings ought to be combinable? but something like a str and an int shouldn't be ?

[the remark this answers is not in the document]
chrishalcrow: Another thing: numpy allows for concatenation of different unicode types:
```np.concat([np.array(['hello', 'there']),  np.array(['you'])])```
is fine. So we don't necessarily need to update the data ourselves - we just need to check that it's all unicode.

PR #3701, opened 2025-02-17
Set max_channel property of MdaSortingExtractor

### LINKED ISSUE #3695: Storing channel index from MDA sorting in MdaSortingExtractor for downstream use
I am using spikeinterface and NeuroConv to convert a firings.mda sorting file from an older version of MountainSort to an NWB file. Currently this creates an NWB Units table with spike times and unit labels from MountainSort, but the table lacks information about which channel a unit was maximally detected on, which is available in the MDA file. NWB has an "electrodes" column in the Units table to store that.

`MdaSortingSegment._max_channels` (https://github.com/SpikeInterface/spikeinterface/blob/dd0c6d33f98f071362fa06abb631a7629b97de58/src/spikeinterface/extractors/mdaextractors.py) stores channel/electrode IDs for each spike event from the MDA file but those are not used anywhere.

In the MDA files that I have, for each unit, every spike assigned to that unit has the same max channel index. Is that always the case, @magland?

If so, I propose making `MdaSortingSegment._max_channels` public, as in drop the leading underscore in the variable name, so that downstream tools, e.g., NeuroConv, can use that information, e.g., when writing the sorting to an NWB file.

Does that sound reasonable? I am not an expert in spikeinterface and neuroconv. If there is a better approach, please let me know. Thanks!

Reference for firings.mda file format: https://mountainsort.readthedocs.io/en/latest/first_sort.html#format-of-the-firings-mda

rly:
Fix #3695

### DIFF
--- a/src/spikeinterface/extractors/mdaextractors.py
+++ b/src/spikeinterface/extractors/mdaextractors.py
@@ -202,6 +202,25 @@ class MdaSortingExtractor(BaseSorting):
         sorting_segment = MdaSortingSegment(firings)
         self.add_sorting_segment(sorting_segment)
 
+        # Store the max channel for each unit
+        # Every spike assigned to a unit (label) has the same max channel
+        # ref: https://github.com/SpikeInterface/spikeinterface/issues/3695#issuecomment-2663329006
+        label_to_channel = dict()
+        segment = self._sorting_segments[0]
+        for spike_event_index in range(len(segment._labels)):
+            label = segment._labels[spike_event_index]
+            max_channel = segment._max_channels[spike_event_index]
+            label_to_channel[label] = max_channel
+        
+        # Put channels into a list in the same order as labels, and set as property
+        # NOTE: The max_channel values (1-indexed) index the channels given as input to the sorter
+        # If sorting was run on a subset of channels of the recording, then the max_channel values are
+        # based on that subset, so care must be taken when associating these values with a recording.
+        # NOTE: if additional sorting segments are added to this sorting extractor, then max_channel
+        # will not be updated
+        max_channels = [value for key, value in sorted(d.items())]
+        self.set_property(key="max_channel", values=max_channels)
+
         self._kwargs = {
             "file_path": str(Path(file_path).absolute()),
             "sampling_frequency": sampling_frequency,
@@ -209,7 +228,7 @@ class MdaSortingExtractor(BaseSorting):
 
     @staticmethod
     def write_sorting(sorting, save_path, write_primary_channels=False):
-        assert sorting.get_num_segments() == 1, "MdaSorting.write_sorting() can only write a single segment " "sorting"
+        assert sorting.get_num_segments() == 1, "MdaSorting.write_sorting() can only write a single segment sorting"
         unit_ids = sorting.get_unit_ids()
         times_list = []
         labels_list = []
--- a/src/spikeinterface/extractors/mdaextractors.py
+++ b/src/spikeinterface/extractors/mdaextractors.py
@@ -218,7 +218,7 @@ class MdaSortingExtractor(BaseSorting):
         # based on that subset, so care must be taken when associating these values with a recording.
         # NOTE: if additional sorting segments are added to this sorting extractor, then max_channel
         # will not be updated
-        max_channels = [value for key, value in sorted(d.items())]
+        max_channels = [value for key, value in sorted(label_to_channel.items())]
         self.set_property(key="max_channel", values=max_channels)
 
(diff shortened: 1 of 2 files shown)


src/spikeinterface/extractors/mdaextractors.py
```diff
@@ -202,14 +202,33 @@ def __init__(self, file_path, sampling_frequency):
...
+            max_channel = segment._max_channels[spike_event_index]
+            label_to_channel[label] = max_channel
+
+        # Put channels into a list in the same order as labels, and set as property
+        # NOTE: The max_channel values (1-indexed) index the channels given as input to the sorter
+        # If sorting was run on a subset of channels of the recording, then the max_channel values are
+        # based on that subset, so care must be taken when associating these values with a recording.
+        # NOTE: if additional sorting segments are added to this sorting extractor, then max_channel
+        # will not be updated
+        max_channels = [value for key, value in sorted(d.items())]
```
alejoe91, line 221:
```suggestion
        max_channels = [value for key, value in sorted(label_to_channel.items())]
```
?

rly: oops

src/spikeinterface/extractors/mdaextractors.py
```diff
@@ -202,14 +202,33 @@ def __init__(self, file_path, sampling_frequency):
...
 
+        # Store the max channel for each unit
+        # Every spike assigned to a unit (label) has the same max channel
+        # ref: https://github.com/SpikeInterface/spikeinterface/issues/3695#issuecomment-2663329006
+        label_to_channel = dict()
+        segment = self._sorting_segments[0]
+        for spike_event_index in range(len(segment._labels)):
+            label = segment._labels[spike_event_index]
+            max_channel = segment._max_channels[spike_event_index]
+            label_to_channel[label] = max_channel
```
alejoe91, line 213:
```suggestion
        max_channels = []
        segment = self._sorting_segments[0]
        for unit_id in self.unit_ids:
            label_mask = segment._labels == unit_id
            # since all max channels are the same, we can just grab the first occurrence for the unit
            max_channel = segment._max_channels[label_mask][0]
            max_channels.append(max_channel)
```
You can just loop over unit ids, to make sure it's already sorted accordingly to the `Sorting.unit_id`.
Also, since max channels are all the same, we can just grab the first value

rly: Sounds great

rly:
I think the comment about max_channel values being 1-indexed and being based on the input data to the sorter is useful to document somewhere. Perhaps in the docstring of `MdaSortingExtractor`?

rly:
I haven't run tests locally. I was hoping to see if the workflows pass.

src/spikeinterface/extractors/mdaextractors.py
```diff
@@ -180,6 +180,13 @@ def get_traces(
 class MdaSortingExtractor(BaseSorting):
     """Load MDA format data as a sorting extractor.
 
+    NOTE: As in the MDA format, the max_channel property indexes the channels that are given as input
+    to the sorter.
+    If sorting was run on a subset of channels of the recording, then the max_channel values are
+    based on that subset, so care must be taken when associating these values with a recording.
+    If additional sorting segments are added to this sorting extractor after initialization,
+    then max_channel will not be updated. The max_channel indices begin at 1.
```
alejoe91, line 188: Should we then set the property with a -1?

rly: Probably. Though if you are familiar with the MDA format, you might know and expect that the channel indices begin at 1. I leave that for you to decide

alejoe91: No strong preference here! Just thinking that if downstream or during NWB export we assume these are channel indices we would assume they start from 0.

@magland any thought here?

alejoe91 (approved):

src/spikeinterface/extractors/mdaextractors.py
alejoe91: Letàs leave 1-based, since this is a reader :)

rly:
Is there a way to add a description to the `max_channel` property so that when it is converted to NWB, the description of the `Units.max_channel` property is set to that description? It would be nice to document what "max_channel" is in the file itself.

alejoe91:
@rly no unfortunately we don't have property descriptions. I think that this could be done with a custom snippet on the NeuroConv side

rly:
Sounds good. This is ready to merge IMO.

PR #3711, opened 2025-02-19
Allow preprocessing on dicts of recordings

chrishalcrow:
This PR would allow user to pass dicts of recordings to all preprocessing classes (for now: just common reference!)

This would allow for the following syntax:

``` python
rec, _ = si.generate_ground_truth_recording(num_channels=16)
rec.set_property(key="tetrode_number", values=[0,0,0,0,1,1,1,1,2,2,2,2,3,3,3,3])
pp_rec = si.common_reference(rec.split_by("tetrode_number"))
pp_rec
>>> {0: GroundTruthRecording (CommonReferenceRecording): 4 channels - 25.0kHz - 1 segments 
                       250,000 samples - 10.00s - float32 dtype - 3.81 MiB,
 1: GroundTruthRecording (CommonReferenceRecording): 4 channels - 25.0kHz - 1 segments 
                       250,000 samples - 10.00s - float32 dtype - 3.81 MiB,
 2: GroundTruthRecording (CommonReferenceRecording): 4 channels - 25.0kHz - 1 segments 
                       250,000 samples - 10.00s - float32 dtype - 3.81 MiB,
 3: GroundTruthRecording (CommonReferenceRecording): 4 channels - 25.0kHz - 1 segments 
                       250,000 samples - 10.00s - float32 dtype - 3.81 MiB}
```

which can be chained without problem. Also allows for users to take in two sessions from e.g. different days and ensure the same preprocessing steps are applied to them. Or helpful if you want to split your recording in half and sort seperately (UnitMatch almost do this).

This would certainly remove a few annoying for loops from our labs' pipeline.

We agreed to not touch the `BaseRecording` class. I made the `define_function_or_dict_from_class `, which looks through the `args` and `kwargs` for a recording or dict of recordings, then applies either the normal function (e.g. `CommonReferenceRecording`) or produces a dict, where each recording has had the normal function applied to it, keeping the original keys. I've copied the signature, doc and set a new name in the function, rather than calling another decorator (ala https://github.com/SpikeInterface/spikeinterface/blob/c837b3bbbdea799ac03c4a41dd8e488a963f6bd0/spikeinterface/core/core_tools.py#L25)  . I think this is simpler??

Also, note that this only works for the functions. So `common_reference(dict_of_recs)` is good but `CommonReferenceRecording(dict_of_recs)` does not work. This seems ok from the user point of view, but maybe there's some deeper issue with loading that I'm not thinking of??


src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -259,4 +259,6 @@ def slice_groups(self, channel_indices):
         return zip(group_indices, selected_channels, group_channels)
 
 
-common_reference = define_function_from_class(source_class=CommonReferenceRecording, name="common_reference")
+@_is_documented_by(CommonReferenceRecording)
```
alejoe91, line 262: why not modifying the `define_function_from_class` to do both the `support_extractor_or_dict` and deal with docs?

zm711: the library doesn't do much with decorators (outside of the class property stuff). Is this a pattern that we want to introduce in general? Not for against, but something I think worth discussing at the meeting.

chrishalcrow: Yeah, I'd be interested to see a different implementation. I really couldn't figure out any other way to do it though! You need to set the docs when this is defined. E.g. putting below the def `common_reference.__doc__ = CommonReferenceRecording.__doc__` doesn't work, since this isn't run on import.

chrishalcrow:
Had another go and it feels more beautiful! Added the `silence_period` example because it has another arg (rather than kwarg). @alejoe91 @samuelgarcia, would appreciate feedback!

src/spikeinterface/core/core_tools.py
```diff
@@ -9,10 +9,49 @@
 import importlib
 from math import prod
 from collections import namedtuple
+import inspect
 
 import numpy as np
 
 
+def dict_of_source_classes(rec_or_dict_of_recs, source_class, *args, **kwargs):
```
zm711, line 17: I'm wondering if we should name this differently. Most of our functions are verbs no? so something like `create_dict_of_source_classes` or `generate_dict_of_source_classes`. As is this seems like an object/attribute rather than a function, no?

zm711: And the signature of this function only accepts `dict_of_recs` so maybe re-named to make it clearer what goes in (or add a really nice docstring).

src/spikeinterface/core/core_tools.py
```diff
@@ -9,10 +9,49 @@
...
+    preprocessed_recordings_dict = {
+        property_id: source_class(recording, *args, **kwargs) for property_id, recording in rec_or_dict_of_recs.items()
+    }
+    return preprocessed_recordings_dict
+
+
+def define_rec_or_dict_function(source_class, name):
```
zm711, line 24: same for symmetry maybe `define_rec_or_dict_function_from_class` so that we see that this function is the same as below function but to have a `dict_function`.

zm711 (commented):
A couple naming things to think about while you wait for Alessio and Sam.

src/spikeinterface/core/core_tools.py
chrishalcrow: How about: `define_function_or_dict_from_class`, which is the same as below, but I think is more accurate?

src/spikeinterface/core/core_tools.py
chrishalcrow: Cool, I like `create_dict_of_source_classes`

src/spikeinterface/core/core_tools.py
zm711: Works for me. I guess the real full name would be `define_function_or_dict_of_function_from_class` but at some point the name is just too long :P.

alejoe91: I think we can keep tho original name qnd explain what it does on the docstting for simplicity :)

chrishalcrow: I think Zach is right: it's nice to have the same grammar as `define_function_from_class`. Also added docstrings to both new functions

src/spikeinterface/core/core_tools.py
```diff
@@ -9,10 +9,55 @@
...
+            return source_class(*args, **kwargs)
+        else:
+            # Edit args & kwargs to pass the dict of recordings but _not_ the original recording
+            new_kwargs = {key: kwarg for key, kwarg in kwargs.items() if key != "recording"}
+            if recording_in_kwargs:
+                new_args = args
+            else:
+                new_args = args[1:]
+
+            return create_dict_of_source_classes(rec_or_dict_of_recs, source_class, *new_args, **new_kwargs)
```
samuelgarcia, line 52: I would do the logic here instead of external function for more readability.

samuelgarcia:
This looks good.
We should check the behavior when the input is a list of recordings for instance when concatenating.
What do we want here ?

src/spikeinterface/core/core_tools.py
```diff
@@ -9,10 +9,55 @@
...
+
+        recording_in_kwargs = False
+        if rec_or_dict_of_recs := kwargs.get("recording"):
+            recording_in_kwargs = True
+        else:
+            rec_or_dict_of_recs = args[0]
+
+        if isinstance(rec_or_dict_of_recs, BaseRecording):
+            return source_class(*args, **kwargs)
+        else:
```
samuelgarcia, line 44: I would do 
```suggestion
        elif isinstance(, dict):
```

and then raise an error when it is something else

chrishalcrow: Yess, great idea

src/spikeinterface/core/core_tools.py
chrishalcrow: Yup, sounds good - it's a one liner.

src/spikeinterface/core/core_tools.py
chrishalcrow: This now doesn't exist - the simplest solution when having trouble naming a function.

samuelgarcia:
This is OK for me.
Do you want to make doc inside this PR ? 
You should for instance change the `process_by_channel_group.rst` (or maybe make a gallery example) ?

src/spikeinterface/preprocessing/clip.py
```diff
@@ -169,5 +169,7 @@ def get_traces(self, start_frame, end_frame, channel_indices):
         return traces
 
 
-clip = define_function_from_class(source_class=ClipRecording, name="clip")
-blank_staturation = define_function_from_class(source_class=BlankSaturationRecording, name="blank_staturation")
+clip = define_function_handling_dict_from_class(source_class=ClipRecording, name="clip")
+blank_staturation = define_function_handling_dict_from_class(
+    source_class=BlankSaturationRecording, name="blank_staturation"
```
chrishalcrow, line 174:
```suggestion
    source_class=BlankSaturationRecording, name="blank_saturation"
```

chrishalcrow:
All looks good: I corrected the spelling on `blank_saturation`.

 Thanks for adding it to all the preprocessors @alejoe91  !

alejoe91:
Perfect! As part of this PR, can you modify [this doc page](https://github.com/chrishalcrow/spikeinterface/blob/group-splitting-in-pp/doc/how_to/process_by_channel_group.rst) to show both strategies?

chrishalcrow:
> Perfect! As part of this PR, can you modify [this doc page](https://github.com/chrishalcrow/spikeinterface/blob/group-splitting-in-pp/doc/how_to/process_by_channel_group.rst) to show both strategies?

Hello, that page also shows how to aggregate the recordings back into one, and how to sort grouped recordings. I think we should add 1) allow aggregate channels to combine dicts 2) allow run_sorter to take dicts of recordings and then update the docs. Without these, the workflow doesn't really make sense.

chrishalcrow:
Oops!

alejoe91:
Oups! I merged the absolute imports and a bunch of small conflicts showed up...

chrishalcrow:
Ughhh, got a new machine and have made a mess... gimme a minute...

chrishalcrow:
I made too much of a git mess. See new PR here: #3773

zm711:
[I did something bad](https://github.com/SpikeInterface/spikeinterface/pull/3711/commits/070896b82fde5914e5da03ba7fc4de32ec2f0566)

Favorite commit message maybe ever.

PR #3715, opened 2025-02-21
Remove artefacts based of an enveloppe

yger:
Because this code has been asked by several people, I have the feeling that there might be the need for such a preprocessing Node, to remove artefacts. The idea is the following: the node creates an low-pass filtered version of the rectified signal, compute the noise levels of this new signal, and detect thresholds crossings (onsets/offsets). We are then using these detected time periods to feed a silence_periods nodes, such that periods spotted by the enveloppe will be blanked.

Do you think such a node can be useful? For people using tetrode/few channels it seems to matter quite a lot. For denser probes, this could be debatted, and especially one could think about extending the silence_periods node to work on a per_channel basis instead of across all channels. But this could be debated

### DIFF
--- a/src/spikeinterface/preprocessing/preprocessinglist.py
+++ b/src/spikeinterface/preprocessing/preprocessinglist.py
@@ -43,6 +43,7 @@ from .directional_derivative import DirectionalDerivativeRecording, directional_
 from .depth_order import DepthOrderRecording, depth_order
 from .astype import AstypeRecording, astype
 from .unsigned_to_signed import UnsignedToSignedRecording, unsigned_to_signed
+from .silence_artefacts import SilenceArtefactsRecording, silence_artefacts
 
 
 preprocessers_full_list = [
@@ -79,6 +80,7 @@ preprocessers_full_list = [
     DirectionalDerivativeRecording,
     AstypeRecording,
     UnsignedToSignedRecording,
+    SilenceArtefactsRecording,
 ]
 
 preprocesser_dict = {pp_class.name: pp_class for pp_class in preprocessers_full_list}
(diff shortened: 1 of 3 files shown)


zm711:
We already have [remove_artifacts.py](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/preprocessing/remove_artifacts.py) with the US spelling of artifacts. Should probably keep the spelling unified for all artifact based stuff. (Chris is really going to hate me).

src/spikeinterface/preprocessing/silence_artifacts.py
```diff
@@ -0,0 +1,201 @@
...
+        mode="zeros",
+        noise_levels=None,
+        seed=None,
+        list_periods=None,
+        **random_slices_kwargs,
+    ):
+
+        self.enveloppe = RectifyRecording(recording)
+        self.enveloppe = GaussianFilterRecording(self.enveloppe, freq_min=None, freq_max=freq_max)
+        self.enveloppe = CommonReferenceRecording(self.enveloppe)
```
alejoe91:
```suggestion
        self.envelope = RectifyRecording(recording)
        self.envelope = GaussianFilterRecording(self.enveloppe, freq_min=None, freq_max=freq_max)
        self.envelope = CommonReferenceRecording(self.enveloppe)
```

alejoe91 (dismissed):
@yger can you add some docs on the detect_onsets?

Also `enveloppe` -> `envelope` in english :)

src/spikeinterface/preprocessing/silence_artifacts.py
```diff
@@ -0,0 +1,202 @@
...
+def detect_period_artifacts(
+    recording,
+    detect_threshold=5,
+    min_duration_ms=50,
+    freq_max=20.0,
+    seed=None,
+    noise_levels=None,
+    **noise_levels_kwargs,
+):
+
```
samuelgarcia, line 85: add doc string

samuelgarcia:
@alejoe91 I am OK with this. Do you want to read again ? and validate ?

samuelgarcia (approved):

PR #3716, opened 2025-02-21
Fix `extra_properties` propagation in sorting summary and SV string properties

alejoe91:


src/spikeinterface/widgets/utils.py
```diff
@@ -283,6 +284,10 @@ def make_units_table_from_sorting(sorting, units_table=None):
         values = sorting.get_property(col)
         if values.dtype.kind in "iuUSfb" and values.ndim == 1:
             units_table.loc[:, col] = values
+        else:
+            warn(
+                f"Property {col} not added to the units table because it has ndim > 1 or dtype not in 'iuUSfb'",
```
zm711, line 289:
```suggestion
                f"Property {col} not added to the units table because it has ndim > 1 or dtype not in 'iuUSfb'",
```

That error is fine for people that know numpy, but seems a bit too intense for new folks who aren't going to know what iuU etc means. Do we want succinct or would it be better to bring it down to a beginner level and list them out? I think this is an open question. I always prefer making it as simple as possible since the target audience is the neuroscientist and not the developer.

zm711 (commented):
One (maybe annoying) suggestion for the warning.

src/spikeinterface/widgets/utils.py
alejoe91, line 289:
```suggestion
                f"Property {col} not added to the units table because it has ndim > 1 or dtype not supported",
```

src/spikeinterface/widgets/utils.py
```diff
@@ -342,5 +347,9 @@ def make_units_table_from_analyzer(
             # the ndim = 1 is important because we need  column only for the display in gui.
             if values.dtype.kind in "iuUSfb" and values.ndim == 1:
                 units_table.loc[:, col] = values
+            else:
+                warn(
+                    f"Extra property {col} not added to the units table because it has ndim > 1 or dtype not in 'iuUSfb'",
```
alejoe91, line 352:
```suggestion
                    f"Extra property {col} not added to the units table because it has ndim > 1 or dtype not supported",
```

alejoe91 (commented):
suggestions

src/spikeinterface/widgets/utils.py
alejoe91: Done :)

zm711: Thanks :) since I don't use sorting view much is there someone better to doublecheck for correctness? It looks fine to me, but we can always be patient for a review by Sam.

alejoe91: I'll ask him to check :)

src/spikeinterface/widgets/utils.py
```diff
@@ -283,6 +284,10 @@ def make_units_table_from_sorting(sorting, units_table=None):
         values = sorting.get_property(col)
         if values.dtype.kind in "iuUSfb" and values.ndim == 1:
             units_table.loc[:, col] = values
+        else:
+            warn(
+                f"Property {col} not added to the units table because it has ndim > 1 or dtype not supported",
+            )
```
samuelgarcia, line 290: Honestly, this is very annoying.
For instance you often have location.

samuelgarcia, line 290:
```suggestion
```

src/spikeinterface/widgets/utils.py
```diff
@@ -342,5 +347,9 @@ def make_units_table_from_analyzer(
             # the ndim = 1 is important because we need  column only for the display in gui.
             if values.dtype.kind in "iuUSfb" and values.ndim == 1:
                 units_table.loc[:, col] = values
+            else:
+                warn(
+                    f"Extra property {col} not added to the units table because it has ndim > 1 or dtype not supported",
+                )
```
samuelgarcia, line 353:
```suggestion
```

samuelgarcia (approved):

src/spikeinterface/widgets/utils.py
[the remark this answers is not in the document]
alejoe91: here is ok because they are extra properties!

PR #3718, opened 2025-02-24
Fix IBL streaming tests

alejoe91:
Somehow the same PID resulted in a different number of units.

@oliche anything changed on the ONE-api side? Is it possible that a PID is now pointing to a different insertion?

### DIFF
--- a/src/spikeinterface/extractors/tests/test_iblextractors.py
+++ b/src/spikeinterface/extractors/tests/test_iblextractors.py
@@ -12,6 +12,7 @@ PID = "80f6ffdd-f692-450f-ab19-cd6d45bfd73e"
 
 
 @pytest.mark.streaming_extractors
+@pytest.mark.xfail(reason="We need to fix ibllib/one-api dependency")
 class TestDefaultIblRecordingExtractorApBand(TestCase):
     @classmethod
     def setUpClass(cls):
@@ -106,6 +107,7 @@ class TestDefaultIblRecordingExtractorApBand(TestCase):
 
 
 @pytest.mark.streaming_extractors
+@pytest.mark.xfail(reason="We need to fix ibllib/one-api dependency")
 class TestIblStreamingRecordingExtractorApBandWithLoadSyncChannel(TestCase):
     @classmethod
     def setUpClass(cls):
@@ -180,6 +182,7 @@ class TestIblStreamingRecordingExtractorApBandWithLoadSyncChannel(TestCase):
 
 
 @pytest.mark.streaming_extractors
+@pytest.mark.xfail(reason="We need to fix ibllib/one-api dependency")
 class TestIblSortingExtractor(TestCase):
     def test_ibl_sorting_extractor(self):
         """
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,15 +67,15 @@ extractors = [
     "sonpy;python_version<'3.10'",
     "lxml", # lxml for neuroscope
     "scipy",
-    "ONE-api>=3.0.0", # alf sorter and streaming IBL
-    "ibllib>=3.3.0", # streaming IBL
+    "ONE-api>=2.7.0,<3.0.0", # alf sorter and streaming IBL
+    "ibllib>=2.36.0,<3.0.0", # streaming IBL
     "pymatreader>=0.0.32", # For cell explorer matlab files
     "zugbruecke>=0.2; sys_platform!='win32'", # For plexon2
 ]
 
 streaming_extractors = [
-    "ONE-api>=3.0.0", # alf sorter and streaming IBL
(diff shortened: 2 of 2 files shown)


src/spikeinterface/extractors/tests/test_iblextractors.py
```diff
@@ -198,9 +198,9 @@ def test_ibl_sorting_extractor(self):
         except:
             pytest.skip("Skipping test due to server being down.")
         sorting = read_ibl_sorting(pid=PID, one=one)
-        assert len(sorting.unit_ids) == 733
+        assert len(sorting.unit_ids) == 1091
```
zm711, line 201: This seems really risky to have a gt data set and then just test that we load that dataset. Especially if they can change. When I was looking at the failed tests it also said a failure between subtracting datetime.datetime and NoneType. Is that error related to this too? Seems like changes in IBL overall no?

zm711: That seems safer for a test from my perspective.

src/spikeinterface/extractors/tests/test_iblextractors.py
```diff
@@ -198,9 +198,9 @@ def test_ibl_sorting_extractor(self):
...
         sorting_good = read_ibl_sorting(pid=PID, good_clusters_only=True)
-        assert len(sorting_good.unit_ids) == 108
+        assert len(sorting_good.unit_ids) == 155
```
alejoe91, line 203:
```suggestion
        assert len(sorting_good.unit_ids) < len(sorting.unit_ids)
```

src/spikeinterface/extractors/tests/test_iblextractors.py
oliche: Yes, I can show you how to use the `revision` argument that pins the query to a version of the dataset.

alejoe91: Thanks that would be great!

oliche:
Hi Alessio,
This is what I would propose to pin the IBL dataset and also provide the users with a way to do it themselves if they wish so. https://github.com/alejoe91/spikeinterface/pull/25

zm711:
And this should fix the datetime issue too?

alejoe91:
> And this should fix the datetime issue too?

No :( @oliche can you help us debug this error as well?

```
self.ssl = SpikeSortingLoader(one=one, pid=pid)
<string>:18: in __init__
    ???
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/brainbox/io/one.py:815: in __post_init__
    self.eid, self.pname = self.one.pid2eid(self.pid)
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/one/util.py:162: in wrapper
    self.refresh_cache(mode=mode)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = One (online, https://openalyx.internationalbrainlab.org),/ mode = 'auto'

    def refresh_cache(self, mode='auto'):
        """Check and reload cache tables.
    
        Parameters
        ----------
        mode : {'local', 'refresh', 'auto', 'remote'}
            Options are 'local' (don't reload); 'refresh' (reload); 'auto' (reload if expired);
            'remote' (don't reload).
    
        Returns
        -------
        datetime.datetime
            Loaded timestamp.
        """
        # NB: Currently modified table will be lost if called with 'refresh';
        # May be instances where modified cache is saved then immediately replaced with a new
        # remote cache. Also it's too slow :(
        # self.save_cache()  # Save cache if modified
        if mode in {'local', 'remote'}:
            pass
        elif mode == 'auto':
>           if datetime.now() - self._cache['_meta']['loaded_time'] >= self.cache_expiry:
E           TypeError: unsupported operand type(s) for -: 'datetime.datetime' and 'NoneType'
```

alejoe91:
@oliche any lead on this?

alejoe91:
I tried upgrading to the latest IBLlib and one-api versions, but it seems to be causing issues with the envs. 

@zm711 I suggest to skip the IBL streaming tests for now and work on a separate branch to fix them. Maybe @oliche could help us out there :)

zm711:
Would it make sense instead of skipping to pin to older version and make sure they pass. Then in the PR we unpin and can work on a fix?

alejoe91:
I tried! But I wasn't able to find a good combination of versions that works unfortunately :(

zm711:
Okay then we live with what we have!

alejoe91:
I'll open an issue :)

zm711 (approved):

PR #3721, opened 2025-02-25
Handle automatic RAM allocation for chunks

yger:
When using large number of cores, we might want to automatically set the chunk_size as function of the available RAM and/or number of cores. While I know that currently, in SI, the preprocessing nodes can consumes more RAM than chunk_size, at least this will avoid swapping and automatically adjust the sizes

### DIFF
--- a/src/spikeinterface/sorters/internal/spyking_circus2.py
+++ b/src/spikeinterface/sorters/internal/spyking_circus2.py
@@ -15,6 +15,7 @@ from spikeinterface.sortingcomponents.tools import (
     cache_preprocessing,
     get_prototype_and_waveforms_from_recording,
     get_shuffled_recording_slices,
+    set_optimal_chunk_size
 )
 from spikeinterface.core.basesorting import minimum_spike_dtype
 from spikeinterface.core.sparsity import compute_sparsity
@@ -55,6 +56,7 @@ class Spykingcircus2Sorter(ComponentsBasedSorter):
         "apply_preprocessing": True,
         "matched_filtering": True,
         "cache_preprocessing": {"mode": "memory", "memory_limit": 0.5, "delete_cache": True},
+        "chunk_preprocessing": {"memory_limit": 0.5},
         "multi_units_only": False,
         "job_kwargs": {"n_jobs": 0.5},
         "seed": 42,
@@ -84,6 +86,8 @@ class Spykingcircus2Sorter(ComponentsBasedSorter):
         "matched_filtering": "Boolean to specify whether circus 2 should detect peaks via matched filtering (slightly slower)",
         "cache_preprocessing": "How to cache the preprocessed recording. Mode can be memory, file, zarr, with extra arguments. In case of memory (default), \
                          memory_limit will control how much RAM can be used. In case of folder or zarr, delete_cache controls if cache is cleaned after sorting",
+        "chunk_preprocessing": "How much RAM (approximately) should be devoted to load data chunks. memory_limit will control how much RAM can be used\
+                         as a fraction of available memory. Otherwise, use total_memory to fix a hard limit", 
         "multi_units_only": "Boolean to get only multi units activity (i.e. one template per electrode)",
         "job_kwargs": "A dictionary to specify how many jobs and which parameters they should used",
         "seed": "An int to control how chunks are shuffled while detecting peaks",
@@ -126,8 +130,8 @@ class Spykingcircus2Sorter(ComponentsBasedSorter):
 
         job_kwargs = fix_job_kwargs(params["job_kwargs"])
         job_kwargs.update({"progress_bar": verbose})
-
         recording = cls.load_recording_from_folder(sorter_output_folder.parent, with_warnings=False)
+        job_kwargs = set_optimal_chunk_size(recording, job_kwargs, **params["chunk_preprocessing"])
 
         sampling_frequency = recording.get_sampling_frequency()
         num_channels = recording.get_num_channels()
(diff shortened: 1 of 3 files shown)


src/spikeinterface/sortingcomponents/tools.py
```diff
@@ -249,19 +249,132 @@ def check_probe_for_drift_correction(recording, dist_x_max=60):
...
+    """
+    job_kwargs = fix_job_kwargs(job_kwargs)
+    n_jobs = job_kwargs["n_jobs"]
+    if HAVE_PSUTIL:
+        assert 0 < memory_limit < 1, "memory_limit should be in ]0, 1["
+        memory_usage = memory_limit * psutil.virtual_memory().available
+        n_jobs = int(min(n_jobs, memory_usage // ram_requested))
+        job_kwargs.update(dict(n_jobs=n_jobs))
+    else:
+        print("psutil is required to use only a fraction of available memory")
```
zm711, line 327: are these prints or warnings? in general.

src/spikeinterface/sortingcomponents/tools.py
```diff
@@ -249,19 +249,132 @@ def check_probe_for_drift_correction(recording, dist_x_max=60):
...
+
+    if mode == "memory":
+        if total_memory is None:
+            if HAVE_PSUTIL:
+                assert 0 < memory_limit < 1, "memory_limit should be in ]0, 1["
+                memory_usage = memory_limit * psutil.virtual_memory().available
+                if recording.get_total_memory_size() < memory_usage:
+                    recording = recording.save_to_memory(format="memory", shared=True, **job_kwargs)
+                else:
+                    print("Recording too large to be preloaded in RAM...")
```
zm711, line 370: same for these prints.

src/spikeinterface/sortingcomponents/tools.py
```diff
@@ -249,19 +249,132 @@ def check_probe_for_drift_correction(recording, dist_x_max=60):
...
+
+    job_kwargs: dict
+        The updated job kwargs
+    """
+    job_kwargs = fix_job_kwargs(job_kwargs)
+    n_jobs = job_kwargs["n_jobs"]
+    if HAVE_PSUTIL:
+        assert 0 < memory_limit < 1, "memory_limit should be in ]0, 1["
+        memory_usage = memory_limit * psutil.virtual_memory().available
+        n_jobs = int(min(n_jobs, memory_usage // ram_requested))
```
zm711, line 324: if ram_requested is a dict how are you dividing here?

src/spikeinterface/sorters/internal/spyking_circus2.py
```diff
@@ -73,6 +75,8 @@ class Spykingcircus2Sorter(ComponentsBasedSorter):
         "matched_filtering": "Boolean to specify whether circus 2 should detect peaks via matched filtering (slightly slower)",
         "cache_preprocessing": "How to cache the preprocessed recording. Mode can be memory, file, zarr, with extra arguments. In case of memory (default), \
                          memory_limit will control how much RAM can be used. In case of folder or zarr, delete_cache controls if cache is cleaned after sorting",
+        "chunk_preprocessing": "How much RAM (approximately) should be devoted to load data chunks. memory_limit will control how much RAM can be used\
```
zm711, line 78: I think you need to say what units of RAM this will work with. I'm confused by this with the way it is currently written. I think users will need a bit more info to really understand how to use this.

zm711 (commented):
Couple more cosmetic questions.

src/spikeinterface/sortingcomponents/tools.py
yger: Sorry, wrong docstrings, the is an int of course

src/spikeinterface/sortingcomponents/tools.py
yger: I'll use the warning package instead

src/spikeinterface/sortingcomponents/tools.py
yger: same here

b-grimaud:
With a few modifications, it can process a minute long recording in 8h45.

- In the main sorting function,  it would often crash at `estimate_templates`. I added a call to `get_optimal_n_jobs` with the same `ram_requested` as in `clustering/circus.py`.
- In cases where the requested RAM is high and/or the memory limit is low, `int` will round `n_jobs` down to 0, which will return an error. I think `get_optimal_n_jobs` should ensure that `n_jobs` is at least 1, or at least return a specific warning.
- In `clustering/circus.py`, I also had crashes at `remove_duplicates_via_matching`. Passing the same `job_kwargs` as those used in `estimate_templates` does the trick, but another call to `get_optimal_n_jobs` with a proper estimate of the requested amount of RAM could be more appropriate.
- In `clustering_tools.py`, `detect_mixtures` sets `n_jobs` to 1 in its main loop. I'm guessing this part can't be parallelized ?

yger:
Ok, good to know, I'll update the PR. Meanwhile, I'll also make 
- a PR to bypass the estimate_templates() call needed at the end of the clustering. In fact, templates could be inferred (at the cost of slight smoothing) from the SVD components that have been precomputed and are saved in RAM. 
- a PR to use the new graph-based clustering that we developed with sam. Such a clustering, taking the whole probe into account as a whole, would not need to deal with duplicates and thus detect_mixtures() should not be needed anymore. That's being said, I think that this function could also be skipped in your case (and in recent version of circus2). Because the final merging has been improved, keeping some duplicates should not hurt too much, and you are welcome to give it a try. How many are removed by such a function in your case? 

Thanks again for the feedback

PR #3724, opened 2025-02-27
Renaming: `from_time_labels` -> `from_samples_labels`

### LINKED ISSUE #3723: mis-informing naming in se.NumpySorting
in se.NumpySorting.from_times_labels, a user would expect times_list to be time (in seconds). but it is actually in frames
maybe rename int to from_frames_labels, or times_list to frame_list?

or even better if times_list  is given in integers assume its frames and if float it's time?

alejoe91:
and adding `from_times_labels`

fixes #3723

### DIFF
--- a/doc/modules/core.rst
+++ b/doc/modules/core.rst
@@ -711,7 +711,7 @@ In this example, we create a recording and a sorting object from numpy objects:
         spike_trains += spike_trains_i
         labels += labels_i
 
-    sorting_memory = NumpySorting.from_times_labels(times=spike_trains, labels=labels,
+    sorting_memory = NumpySorting.from_samples_labels(times=spike_trains, labels=labels,
                                                     sampling_frequency=sampling_frequency)
 
 
--- a/src/spikeinterface/core/numpyextractors.py
+++ b/src/spikeinterface/core/numpyextractors.py
@@ -236,7 +236,8 @@ class NumpySorting(BaseSorting):
 
     But we have convenient class methods to instantiate from:
       * other sorting object: `NumpySorting.from_sorting()`
-      * from time+labels: `NumpySorting.from_times_labels()`
+      * from samples+labels: `NumpySorting.from_samples_labels()`
+      * from times+labels: `NumpySorting.from_times_labels()`
       * from dict of list: `NumpySorting.from_unit_dict()`
       * from neo: `NumpySorting.from_neo_spiketrain_list()`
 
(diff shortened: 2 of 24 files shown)


yger (approved):
Looks good to me, and I agree it would be easier to understand

zm711 (dismissed):
I think this is clutch for understandability. I think the other longer term thing we should think about (pre-1.0 in my opinion) is moving from the frame -> sample terminology in all places.

doc/modules/core.rst
```diff
@@ -711,7 +711,7 @@ In this example, we create a recording and a sorting object from numpy objects:
         spike_trains += spike_trains_i
         labels += labels_i
 
-    sorting_memory = NumpySorting.from_times_labels(times=spike_trains, labels=labels,
+    sorting_memory = NumpySorting.from_samples_labels(times=spike_trains, labels=labels,
```
h-mayorquin, line 714: How comes the argument is still called times here, is this something we should change?

h-mayorquin:
Two request:
1) Can we add a deprecation so we don't break code.
2) Can the method be `from_samples_and_labels` and `from_times_and_labels`

I find the lack of **and** very confusing in how I read the naming. As in, if the labels were the labels of the times. I am writing, aren't the labels the unit_ids? Shouldn't we use that terminology? `from_samples_and_units_ids` or something like that? I feel less strong about this last point but I feel strongly about the and.

zm711:
I think those are great points Heberto! I was thinking something similar to point 2, but didn't want the double break. But I think you're right that if this breaking change is being done  might as well as make it as clear as possible moving forward.  I like the `from_samples_and_labels` but I take your point that in spikeinterface terminology we use `unit_ids` as our neuron "labels".

samuelgarcia:
I like the idea but we need to be backward compatible here (with a warning)
So maybe the idea of @h-mayorquin is the good way and so we keep the old one.

samuelgarcia (approved):

src/spikeinterface/core/numpyextractors.py
```diff
@@ -337,6 +339,41 @@ def from_times_labels(times_list, labels_list, sampling_frequency, unit_ids=None
...
+            times_list = [times_list]
+            labels_list = [labels_list]
+
+        sample_list = [np.round(t * sampling_frequency).astype("int64") for t in times_list]
+        return NumpySorting.from_samples_and_labels(sample_list, labels_list, sampling_frequency, unit_ids)
+
+    @staticmethod
+    def from_times_labels(times_list, labels_list, sampling_frequency, unit_ids=None) -> "NumpySorting":
+        warnings.warn(
+            "from_times_labels is deprecated, use from_sample_and_labels instead",
```
zm711, line 371: Could we get a removal version so we don't forget to take this out in a couple version?
```suggestion
            "from_times_labels is deprecated and will be removed in 0.104.0, use from_sample_and_labels instead",
```

zm711 (commented):
One tweak to deprecation just so we don't forget to eventually remove.

Issue #3727, opened 2025-03-03
Wrong gain inference for Plexon2 files


FrancescoNegri:
When the `NeoBaseRecordingExtractor` class set the `gain_to_uV` property, it starts from the gains retrieved by `Neo` and applies an `additional_gain` according to the signal unit such that the signal is automatically scaled to microvolts (uV).
This is done in a hardcoded way by checking the `units` array values against a predefined set of alternatives`["V", "Volt", "mV", "uV"]`.

Unluckily, Plexon2 files encode the unit as `"Volts"` and this is not recognized by `NeoBaseRecordingExtractor`, hence a standard `additional_gain` of `1` is applied, leading to a wrong `gain_to_uV` value.

The situation is illustrated in the attached image, that is the `NeoBaseRecordingExtractor` class during debug.

A quick fix would be to add a line checking for `"Volts"`:
```python
additional_gain[units == "Volts"] = 1e6
```

However, this workaround might still fail in future, whenever a unit is not recognized.
Additionally, the `has_non_standard_units` property of `NeoBaseRecordingExtractor` is correctly set to `True`, but no warning or error is raised.

![Image](https://github.com/user-attachments/assets/e69e6b7c-80e6-4169-8848-e1577a4822d5)

alejoe91:
Thanks @FrancescoNegri 

Would you minf making a PR? Also adding a warning for non standard units sounds fair!

zm711:
Is this worth also mentioning on Neo since Plexon has been working with us over there to ensure their DLL plays well with Neo. We could ask for clarity from them on their policies for units moving forward or whether they would refuse to offer any such guarantees.

h-mayorquin:
OK, this should be fixed by #3728. Thanks a lot @FrancescoNegri!

PR #3728, opened 2025-03-03
Fix gain inference for "Volts" unit in Plexon2 files

FrancescoNegri:
This pull request fixes issue #3727.

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/neobaseextractor.py
+++ b/src/spikeinterface/extractors/neoextractors/neobaseextractor.py
@@ -260,16 +260,14 @@ class NeoBaseRecordingExtractor(_NeoBaseExtractor, BaseRecording):
         units = signal_channels["units"]
 
         # mark that units are V, mV or uV
+        standard_units_and_additional_gains = {"V": 1e6, "Volt": 1e6, "mV": 1e3, "uV": 1.0}
         self.has_non_standard_units = False
-        if not np.all(np.isin(units, ["V", "Volt", "mV", "uV"])):
+        if not np.all(np.isin(units, list(standard_units_and_additional_gains.keys()))):
             self.has_non_standard_units = True
 
         additional_gain = np.ones(units.size, dtype="float")
-        additional_gain[units == "V"] = 1e6
-        additional_gain[units == "Volt"] = 1e6
-        additional_gain[units == "mV"] = 1e3
-        additional_gain[units == "uV"] = 1.0
-        additional_gain = additional_gain
+        for key, value in standard_units_and_additional_gains.items():
+            additional_gain[units == key] = value
 
         final_gains = gains * additional_gain
         final_offsets = offsets * additional_gain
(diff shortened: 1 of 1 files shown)


src/spikeinterface/extractors/neoextractors/neobaseextractor.py
```diff
@@ -260,16 +260,17 @@ def __init__(
         units = signal_channels["units"]
 
         # mark that units are V, mV or uV
+        standard_units_and_additional_gains = {"V": 1e6, "Volt": 1e6, "Volts": 1e6, "mV": 1e3, "uV": 1.0}
```
h-mayorquin, line 263:
```suggestion
        standard_units_to_additional_gains = {"V": 1e6, "Volt": 1e6, "Volts": 1e6, "mV": 1e3, "uV": 1.0}
```

src/spikeinterface/extractors/neoextractors/neobaseextractor.py
```diff
@@ -260,16 +260,17 @@ def __init__(
...
         self.has_non_standard_units = False
-        if not np.all(np.isin(units, ["V", "Volt", "mV", "uV"])):
+        if not np.all(np.isin(units, list(standard_units_and_additional_gains.keys()))):
             self.has_non_standard_units = True
+            warnings.warn(
```
h-mayorquin, line 267: I don't think this warning is very informative as it is. It would not be clear for the user what standard means and which ones are the non-standard. 

I think the warning (if you want to add it, otherwise I can add it into another PR) should indicate which channels have units that are not recognized by spikeinterface, the units, and the recognized units.

src/spikeinterface/extractors/neoextractors/neobaseextractor.py
FrancescoNegri: Alright, I updated the name.

src/spikeinterface/extractors/neoextractors/neobaseextractor.py
```diff
@@ -260,16 +260,17 @@ def __init__(
         units = signal_channels["units"]
 
         # mark that units are V, mV or uV
+        standard_units_and_additional_gains = {"V": 1e6, "Volt": 1e6, "Volts": 1e6, "mV": 1e3, "uV": 1.0}
         self.has_non_standard_units = False
-        if not np.all(np.isin(units, ["V", "Volt", "mV", "uV"])):
+        if not np.all(np.isin(units, list(standard_units_and_additional_gains.keys()))):
```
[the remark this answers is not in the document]
FrancescoNegri: Yes, casting from dict_keys to list is necessary for np.isin to work.

src/spikeinterface/extractors/neoextractors/neobaseextractor.py
h-mayorquin: Yeah, my idea, is that the users will know which channels might be fishy and which units we automatically assign.

alejoe91 (approved):

PR #3736, opened 2025-03-04
Warn instead of print in solve_monopolar

alejoe91:
Found it! This should be a warning, so it can be silenced :)


h-mayorquin (approved):
LGTM

PR #3739, opened 2025-03-04
Add warning message for stream with mixed units

h-mayorquin:
After reviewing [PR #3728](https://github.com/SpikeInterface/spikeinterface/pull/3728), I don't think we should be raising an error when non-voltage units are found. Loading data with NIDQ should be fine, and in that case, the warning isn’t particularly useful—it just adds noise.

What we should warn the user about is when the extractor contains mixed units (voltage and non-voltage), as this likely indicates that the stream should be subsetted. This PR improves this and gives the user the tools to figure out which channels might have to be removed.

As a side benefit, this also addresses a request that @samuelgarcia made a while ago to move this warning logic from the base extractor to Neo, which is an added benefit.

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/neobaseextractor.py
+++ b/src/spikeinterface/extractors/neoextractors/neobaseextractor.py
@@ -257,26 +257,37 @@ class NeoBaseRecordingExtractor(_NeoBaseExtractor, BaseRecording):
         else:
             self.inverted_gain = False
 
-        units = signal_channels["units"]
+        channel_units = signal_channels["units"]
 
-        # mark that units are V, mV or uV
-        standard_units_to_additional_gains = {"V": 1e6, "Volt": 1e6, "Volts": 1e6, "mV": 1e3, "uV": 1.0}
-        self.has_non_standard_units = False
-        if not np.all(np.isin(units, list(standard_units_to_additional_gains.keys()))):
-            self.has_non_standard_units = True
-            warnings.warn(
-                f"The raw file has non standard units: {np.unique(units)}. Standard units are: {list(standard_units_to_additional_gains.keys())}. 'uV' are assumed."
+        # Define standard voltage units and their conversion factors to uV
+        voltage_units_to_gains = {"V": 1e6, "Volt": 1e6, "Volts": 1e6, "mV": 1e3, "uV": 1.0}
+        supported_voltage_units = list(voltage_units_to_gains.keys())
+
+        is_channel_in_voltage = np.isin(channel_units, supported_voltage_units)
+        has_voltage_channels = np.any(is_channel_in_voltage)
+        has_non_voltage_channels = not np.all(is_channel_in_voltage)
+
+        # Streams with mixed units are to be used with caution
+        has_mixed_units = has_non_voltage_channels and has_voltage_channels
+        if has_mixed_units:
+            self.set_property("neo_units", channel_units)
+            warning_msg = (
+                "Found a mix of voltage and non-voltage units. "
+                'Proceed with caution. Check channel units with `recording.get_property("neo_units")`.'
             )
+            warnings.warn(warning_msg)
 
-        additional_gain = np.ones(units.size, dtype="float")
-        for key, value in standard_units_to_additional_gains.items():
-            additional_gain[units == key] = value
+        # We assume gains of 1.0 by default and modify that for voltage channels
+        additional_gain = np.ones(channel_units.size, dtype="float")
+        for unit, gain in voltage_units_to_gains.items():
+            additional_gain[channel_units == unit] = gain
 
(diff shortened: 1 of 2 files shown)


src/spikeinterface/extractors/neoextractors/neobaseextractor.py
```diff
@@ -257,26 +257,37 @@ def __init__(
...
+            warning_msg = (
+                "Found a mix of voltage and non-voltage units. "
+                'Proceed with caution. Check channel units with `recording.get_property("neo_units")`.'
             )
+            warnings.warn(warning_msg)
 
-        additional_gain = np.ones(units.size, dtype="float")
-        for key, value in standard_units_to_additional_gains.items():
-            additional_gain[units == key] = value
+        # We assume gains of 1.0 by default and modify that for voltage channels
```
h-mayorquin, line 280: Non voltage units should be have gain_to_UV equal NAN.

src/spikeinterface/extractors/neoextractors/neobaseextractor.py
```diff
@@ -246,37 +246,53 @@ def __init__(
...
-        if not np.all(np.isin(units, list(standard_units_to_additional_gains.keys()))):
-            self.has_non_standard_units = True
-            warnings.warn(
-                f"The raw file has non standard units: {np.unique(units)}. Standard units are: {list(standard_units_to_additional_gains.keys())}. 'uV' are assumed."
+        # Define standard voltage units and their conversion factors to uV
+        voltage_units_to_gains = {"V": 1e6, "Volt": 1e6, "Volts": 1e6, "mV": 1e3, "uV": 1.0}
+        supported_voltage_units = list(voltage_units_to_gains.keys())
+
+        channel_units = signal_channels["units"]
+        fill_value = 1.0  # This should be np.nan but will break a lot of tests
```
h-mayorquin, line 265: @alejoe91 @samuelgarcia 

This is the line that you guys want to change. At the moment this breaks the tests and it also changes the behavior in the following way:

Someone who used to call `get_traces(return_scaled=True)` was getting the get wrong values (because the gains were one) now they will be getting np.nans. So both are wrong but in a different way.

I would like to get this PR through to get current warning out of my way. We can decide how to move forward with the proposed change in the future. What do you think?

zm711: I think fill values are so difficult. Everyone has their own solution (nan vs noise vs arbitrary value). This might be good for a bigger discussion maybe even at the monthly meeting to see if the field as a whole could get closer to agreement or at least within our realm of the field.

h-mayorquin: Nan was the agreement on the monthly meeting

zm711: I thought you were referring to the maintenance meeting. I don't remember discussing this at the Spikes meeting where we could have more community engagement.

zm711: Yeah I think that makes sense from a spikeinterface perspective. 

Many systems have digital/adc data that can also be scaled (ie they are packed as ints and so need a gain/offset to be converted to V for example). In this case the user would need to know that they raw data instead of actual data. (I think our "events" machinery is pretty limited to a few extractors no?). I'm wondering if there is a way we can ensure users know this. Sorry for missing the meeting because it sounds like this PR was part of the discussion so maybe you've already thought of this.

alejoe91: If it's any Volt unit, then it can be safely scaled to uV so it will work. 

The problem is really when there are non-Volt streams. We need a mechanism to support these but throw right errors when users try to do something which is not possible, like converting to uV.
Note that the problem only arises when calling `return_scaled=True`!

zm711: So the only major failure would be any stream that supports digital signals where the user should be okay just setting `return_scaled=False`? I'm sure there are other minor cases but I think that would be the biggest.

This makes sense to me!

alejoe91: yes exactly!

h-mayorquin: This PR also keeps the units of neo so somebody can do the scaling by themselves (which we will do in neuroconv) but at the moment they original units are only available through the header (the gains of nonv-voltage channels are set to 1).

So if anything, this PR makes it better for non-voltage channels. You won't be able to use `return_scaled` but the quantities that are necessary to scale will be available at the Spikeinterface level.

samuelgarcia (approved):

samuelgarcia:
OK for me.
thanks

PR #3744, opened 2025-03-05
Fix dict representation in aggregation (numpy 2.0)

h-mayorquin:
Because numpy 2.0 changes the string representation of string scalars the aggregated dict is displaying strangely. This fixes it:

image


alejoe91 (approved):

zm711:
Good to know. Thanks Heberto!

PR #3746, opened 2025-03-06
Clean up all docs warnings

alejoe91:
@zm711 this is for you :)


zm711:
Thanks Alessio :). I will double check tomorrow when I'm back from the seminar series :)

zm711 (commented):
Looks good on my review. Let me flip through the doc build tomorrow and then we will call it great.

PR #3747, opened 2025-03-06
Re-enable n_jobs in multi-comparison

alejoe91:
@paul-aparicio I realized that the `n_jobs` parameter was actually not used.

This PR adds it back and updates the parallelization using the `ProcessPoolExecutor`. This will run each comparison in parallel.
A simple test shows that the results are >10x faster (on my machine - 8 CPUs).

Can you test this PR? you just need to add the `n_jobs=-1` to the `compare_multiple_sorters`

### DIFF
--- a/src/spikeinterface/comparison/basecomparison.py
+++ b/src/spikeinterface/comparison/basecomparison.py
@@ -2,8 +2,10 @@ from __future__ import annotations
 
 from copy import deepcopy
 from typing import OrderedDict
+from concurrent.futures import ProcessPoolExecutor
 import numpy as np
 
+from spikeinterface.core.job_tools import fix_job_kwargs
 from .comparisontools import make_possible_match, make_best_match, make_hungarian_match
 
 
@@ -27,7 +29,7 @@ class BaseMultiComparison(BaseComparison):
     It handles graph operations, comparisons, and agreements.
     """
 
-    def __init__(self, object_list, name_list, match_score=0.5, chance_score=0.1, verbose=False):
+    def __init__(self, object_list, name_list, match_score=0.5, chance_score=0.1, verbose=False, n_jobs=1):
         import networkx as nx
 
         BaseComparison.__init__(
@@ -41,6 +43,7 @@ class BaseMultiComparison(BaseComparison):
         self.graph = None
         self.subgraphs = None
         self.clean_graph = None
+        self.n_jobs = n_jobs
 
     def _compute_all(self):
         self._do_comparison()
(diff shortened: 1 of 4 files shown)


paul-aparicio:
Absolutely! Thank you so much

src/spikeinterface/comparison/basecomparison.py
```diff
@@ -96,18 +99,38 @@ def _do_comparison(
...
+            for i in range(len(self.object_list)):
+                for j in range(i + 1, len(self.object_list)):
+                    if self.name_list is not None:
+                        name_i = self.name_list[i]
+                        name_j = self.name_list[j]
+                    else:
+                        name_i = "object i"
+                        name_j = "object j"
+                    job_list.append((i, j))
+                    job_names.append((name_i, name_j))
```
alejoe91, line 129: Refactor this

alejoe91:
As long as you have spikeinterface installed from sources, yes! You can also try the other agreement method (`counts`), since this will parallelize over comparisons across sorters

paul-aparicio:
Hi, I completed my tests. I did not find much difference, maybe I downloaded the incorrect patch? I tested agreement_method='counts' (took ~7.5 hours to complete both before and after) and agreement_method='count' (took ~7.75 hours to compete before and after the change). I am supposed to patch with #3747? Thanks!

alejoe91:
How did you install from this PR?

alejoe91:
When runnint the new `test_parallel` test locally I'm getting a large speedup:
```
Elapsed 1 job: 1.7032024950021878
Elapsed 8 jobs: 0.21884372199565405
```

h-mayorquin (approved):
This looks OK to me. It is adding an option that is not enabled by default. @alejoe91 profiled for correctness and is faster on his computer.

src/spikeinterface/comparison/basecomparison.py
```diff
@@ -96,6 +99,10 @@ def _do_comparison(
             print("Multicomparison step 1: pairwise comparison")
 
         self.comparisons = {}
+
+        n_jobs = fix_job_kwargs({"n_jobs": self.n_jobs})["n_jobs"]
```
samuelgarcia, line 103: Here this will not have the general behavior.
In short if a user do
```python
set_global_job_kwargs(n_jobs=-1)
comp  = BaseMultiComparison()
```
then the n_jobs will be 1 and so not parralel.

src/spikeinterface/comparison/basecomparison.py
```diff
@@ -104,10 +111,17 @@ def _do_comparison(
...
+                job_names.append((name_i, name_j))
+
+        if self.n_jobs == 1:
+            for job_name, job in zip(job_names, job_list):
+                self.comparisons[job_name] = self._compare_ij(*job)
+        else:
+            with ProcessPoolExecutor(max_workers=n_jobs) as executor:
+                results = list(executor.map(self._compare_ij, *zip(*job_list)))
+            for job_name, result in zip(job_names, results):
+                self.comparisons[job_name] = result
```
samuelgarcia, line 124: We could also implement a progress_bar=True/False no ?

samuelgarcia:
This is ok for me BUT this do not follow the main job_kwargs behavior anywhere else. because it is one by default and this has precedence over the global ones.

Issue #3745, opened 2025-03-05
Scaling bug in plotting waveforms/templates for checkerboard-like probes


zzhmark:
Plotting waveforms for my probe yielding very small waveforms, unless a scale around 1000 is specified. I found the y_scale factor in the plotting functions are very small, due to the layout of my channel positions, which is a non-uniform checkerboard. Some of the y coords may have miniscule differences, and it causes the program to perceive it as a very small inter channel distance. You might need a more adaptive strategy with such scenario.

zm711:
Absolutely agree. I was having this issue too because of a the nature of the x changes between my contacts on the probe leading to x squashing. I think we need to work on a better solution overall! I think a lot of us are busy. You want to give this a try? Otherwise we will put this on the todo list!

zm711:
@alejoe91 and @samuelgarcia review the widgets stuff so I'll tag them to take a look at your proposal.

zm711:
Closing as we have merged the PR :) Thanks again for the contribution!

PR #3748, opened 2025-03-06
Fitting scaling factors in waveform/template plots for irregular probe layouts

zzhmark:
delta_y will be inversely weighted by the x distances to get a proper interval over y dimension. In this case contacts that are close vertically but far laterally will be less represented.

The issue:
https://github.com/SpikeInterface/spikeinterface/issues/3745#issuecomment-2702562018


alejoe91:
@zzhmark can you provide some screenshots before/after? :)

alejoe91:
@zm711 @samuelgarcia what do you think? 

I believe the implementation is too complex for a plotting function at this point, so I would rather go for a middle-way solution that gives OK results, but is not too complex

zm711:
That implementation looks much better to me. I'm in the middle of some analyses, but I'll have to test this myself with a couple probe layouts and make sure it works.

zm711:
@chrishalcrow, don't you think the x on the tetrodes is just a hair to close? Again I don't think we will ever be perfect and I agree NP is more common so we want to make sure that always looks nice. I want to test one other probe (It has a v-shape so it's X is a bit weird so I'd like to test to see how this handles some of setups. 

As I see it the normal geometries are 
diamonds (eg. tetrodes)
squares/rectangle (eg. NP)
checkerboard (repeating diamonds really) (eg other NP)

then you could have some "weirder"

v -shaped
homemade

So I would love to just have an idea of how as many of these would look. I assume when Alessio/Sam made this originally it was with a specific subset of probes in mind (maybe some NPs)

chrishalcrow:
Looks good for NP2.0 and the ProbeInterface tetrode!

np_after_32
tet_after_32

alejoe91 (approved):

PR #3753, opened 2025-03-06
Add Intan port groups by default

h-mayorquin:
@zm711 

We discussed this a while ago. 

This will be useful for stream users so they can sort or split by group which are (in most cases not all) mapped to different probes or devices:

image

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/intan.py
+++ b/src/spikeinterface/extractors/neoextractors/intan.py
@@ -1,10 +1,10 @@
 from __future__ import annotations
-
 from pathlib import Path
 
-from spikeinterface.core.core_tools import define_function_from_class
+import numpy as np
 
-from .neobaseextractor import NeoBaseRecordingExtractor, NeoBaseSortingExtractor
+from spikeinterface.core.core_tools import define_function_from_class
+from .neobaseextractor import NeoBaseRecordingExtractor
 
 
 class IntanRecordingExtractor(NeoBaseRecordingExtractor):
@@ -64,7 +64,7 @@ class IntanRecordingExtractor(NeoBaseRecordingExtractor):
             **neo_kwargs,
         )
 
-        # self.set_channel_groups()
+        self._add_channel_groups()
 
         self._kwargs.update(
             dict(file_path=str(Path(file_path).resolve()), ignore_integrity_checks=ignore_integrity_checks),
@@ -75,6 +75,26 @@ class IntanRecordingExtractor(NeoBaseRecordingExtractor):
 
         neo_kwargs = {"filename": str(file_path), "ignore_integrity_checks": ignore_integrity_checks}
         return neo_kwargs
+
+    def _add_channel_groups(self):
+
+        analog_ports = ["A", "B", "C", "D", "E", "F"]
+        num_channels = self.get_num_channels()
+        original_ids = self.neo_reader.header["signal_channels"]["id"][:num_channels]
+        groups = np.zeros(shape=num_channels, dtype="int")
+        group_names = np.zeros(shape=num_channels, dtype="str")
+
+        # The hard-coded IDS of intan ids is "Port-Number" for amplifier channels
+        channel_ports = [id[:1] for id in original_ids if id[1] == "-"]
+
+        for port in analog_ports:
+            channel_index = np.where(np.array(channel_ports) == port)
+            if channel_index[0].size > 0:
+                group_names[channel_index] = port
+                groups[channel_index] = analog_ports.index(port) + 1
+
+        self.set_channel_groups(groups)
+        self.set_property(key="group_names", values=group_names)
 
 
 read_intan = define_function_from_class(source_class=IntanRecordingExtractor, name="read_intan")
--- a/src/spikeinterface/extractors/neoextractors/intan.py
+++ b/src/spikeinterface/extractors/neoextractors/intan.py
@@ -64,7 +64,9 @@ class IntanRecordingExtractor(NeoBaseRecordingExtractor):
             **neo_kwargs,
         )
 
-        self._add_channel_groups()
+        amplifier_streams = ["RHS2000 amplifier channel", "RHD2000 amplifier channel"]
+        if self.stream_name in amplifier_streams:
+            self._add_channel_groups()
 
         self._kwargs.update(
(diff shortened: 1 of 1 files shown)


src/spikeinterface/extractors/neoextractors/intan.py
```diff
@@ -64,23 +64,39 @@ def __init__(
...
+            self._add_channel_groups()
+
+        self._kwargs.update(
+            dict(file_path=str(Path(file_path).resolve()), ignore_integrity_checks=ignore_integrity_checks),
+        )
 
     @classmethod
     def map_to_neo_kwargs(cls, file_path, ignore_integrity_checks: bool = False):
 
-        # Only propagate the argument if the version is greater than 0.13.1
```
zm711, line 74: why do you want to remove backward compatibility?

src/spikeinterface/extractors/neoextractors/intan.py
```diff
@@ -64,23 +64,39 @@ def __init__(
...
-        if neo_version > packaging.version.parse("0.13.1"):
-            neo_kwargs = {"filename": str(file_path), "ignore_integrity_checks": ignore_integrity_checks}
-        else:
-            neo_kwargs = {"filename": str(file_path)}
+        neo_kwargs = {"filename": str(file_path), "ignore_integrity_checks": ignore_integrity_checks}
         return neo_kwargs
 
+    def _add_channel_groups(self):
+
+        analog_ports = ["A", "B", "C", "D", "E", "F", "G", "H"]
```
zm711, line 83: Is this the hard limit (I forget). Our machine is a 512 and it goes to D so it makes sense their bigger 1024 would go to H but did you confirm?

src/spikeinterface/extractors/neoextractors/intan.py
```diff
@@ -64,23 +64,39 @@ def __init__(
...
+        num_channels = self.get_num_channels()
+        original_ids = self.neo_reader.header["signal_channels"]["id"][:num_channels]
```
zm711, line 85: why are we doing 84 and 85. why get num channels instead of all data? Are we saying that we will always guarantee amplifier channels are stored first?

src/spikeinterface/extractors/neoextractors/intan.py
```diff
@@ -64,23 +64,39 @@ def __init__(
...
+        group_names = np.zeros(shape=num_channels, dtype="str")
+
+        # The hard-coded IDS of intan ids is "Port-Number" for amplifier channels
+        channel_ports = [id[:1] for id in original_ids if id[1] == "-"]
+
+        for port in analog_ports:
+            channel_index = np.where(np.array(channel_ports) == port)
+            if channel_index[0].size > 0:
+                group_names[channel_index] = port
+                groups[channel_index] = analog_ports.index(port) + 1
```
zm711, line 96: why do we need + 1

zm711 (commented):
first round of questions!

src/spikeinterface/extractors/neoextractors/intan.py
h-mayorquin: The floor for neo version is higher than this now on the pyproject.toml.

src/spikeinterface/extractors/neoextractors/intan.py
h-mayorquin: That's what the photo shows

src/spikeinterface/extractors/neoextractors/intan.py
zm711: Sounds good!

src/spikeinterface/extractors/neoextractors/intan.py
zm711: But I haven't checked their website lately. Do they make a 2048? My question was more have you checked their website to make sure you found the biggest recorder possible?

h-mayorquin: I did not find anything but we can just make it larger as a precaution?

zm711: Probably can't hurt since you're search on the letters and you won't letters that don't exist. Like I said our machine is A-D so the extra letters should be fine for the smaller recorder. But up to you. I guess the other option would be to find some way to see if you missed channels and then report to the user which channels weren't grouped? But that gets messy with non-amplifier channels, right?

src/spikeinterface/extractors/neoextractors/intan.py
h-mayorquin: I need to filter for amplifier channels as they are the ones that follow this pattern. But I think you make a good point about the guarantees which do not exists. I will do the neo pattern of filtering the header by signal stream and take it form there. Thanks.

src/spikeinterface/extractors/neoextractors/intan.py
h-mayorquin: We don't. Removed it. Let's go with zero indexing.

src/spikeinterface/extractors/neoextractors/intan.py
h-mayorquin: Check the new implementations that sidesteps this by just getting the port from the available ids.

src/spikeinterface/extractors/neoextractors/intan.py
```diff
@@ -64,23 +64,42 @@ def __init__(
...
+        groups = np.zeros(shape=num_channels, dtype="uint16")
+        group_names = np.zeros(shape=num_channels, dtype="str")
+
+        signal_header = self.neo_reader.header["signal_channels"]
+        amplifier_signal_header = signal_header[signal_header["stream_id"] == self.stream_id]
+        original_ids = amplifier_signal_header["id"]
+
+        # The hard-coded IDS of intan ids is "Port-Number" (e.g. A-001, C-017, B-020, etc) for amplifier channels
+        channel_ports = [id[:1] for id in original_ids if id[1] == "-"]
+        analog_ports = np.unique(channel_ports).tolist()
```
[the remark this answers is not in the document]
h-mayorquin: Good point.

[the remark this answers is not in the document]
h-mayorquin: changed.

zm711 (approved):
This works for me. @alejoe91 did you want to take a look too?

PR #3756, opened 2025-03-09
[Draft] Ruff as linter

zm711:
I just wanted to draft out so we could see. Right now if I did it right--though I doubt it haha. This should just check for unused imports... If we like this we could remove black auto commit and switch with a commit action in the future. But just want to see how this flows.


zm711:
That took ~3 seconds to run. That's insane! Just for unused imports right now. I think we would exclude init files so we could get fancier. But might be worth thinking about.

samuelgarcia:
Mayeb it is time to switch no ?

zm711:
Well if we do switch we need to decide what we want. Currently this is set to lint imports only. But we can fine tune this. We could start by switching this to actually do the black formatting and then add rules if we need to get what we want. We can put this on a meeting agenda!

zm711:
@zm711 note to self. Turn this into a pre-commit action. Think about which rules we want to start with--probably just black-like formatting.

PR #3757, opened 2025-03-10
Pin sortingview version

alejoe91:
Sortingview has been updated, so we need some modifications to the CI to make it work with the latest versiobn

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -164,7 +164,7 @@ test = [
     "hdbscan>=0.8.33",  # Previous version had a broken wheel
 
     # for sortingview backend
-    "sortingview<0.14.0",
+    "sortingview>=0.12.0,<0.14.0",
 
     # for motion and sortingcomponents
     "torch",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -164,7 +164,7 @@ test = [
     "hdbscan>=0.8.33",  # Previous version had a broken wheel
 
     # for sortingview backend
-    "sortingview",
```
zm711, line 167: should this one also have the lower limit like above?

pyproject.toml
```diff
@@ -164,7 +164,7 @@ test = [
...
+    "sortingview<0.14.0",
```
alejoe91, line 167: That's done @zm711

zm711:
```suggestion
    "sortingview>=0.12.0,<0.14.0",
```
I was saying this? Do we need the lower limit like we have in above so that we don't allow people to build the broken one.

alejoe91: Ehehehe you're right! Let me add that!

h-mayorquin (approved):
Until fixes I guess.

PR #3760, opened 2025-03-11
Add pydantic curation model, improve merging rules, and add splitting model

alejoe91:
This PR goes in the direction of adding more structure to the curation format.

By defining a Pydantic model, we can add proper description, types, and validation strategies for the curation.
This will make it easier to validate and adopt by third party software

### DIFF
--- a/src/spikeinterface/generation/hybrid_tools.py
+++ b/src/spikeinterface/generation/hybrid_tools.py
@@ -399,8 +399,10 @@ def generate_hybrid_recording(
     probe = recording.get_probe()
     num_segments = recording.get_num_segments()
     dtype = recording.dtype
-    durations = np.array([recording.get_duration(segment_index) for segment_index in range(num_segments)])
     num_samples = np.array([recording.get_num_samples(segment_index) for segment_index in range(num_segments)])
+    # since the recording can have timestamps with some small gaps, we use the number of samples to compute
+    # the duration used for the sorting generation
+    durations = num_samples / sampling_frequency
     channel_locations = probe.contact_positions
 
     assert (
--- a/src/spikeinterface/preprocessing/tests/test_average_across_direction.py
+++ b/src/spikeinterface/preprocessing/tests/test_average_across_direction.py
@@ -9,14 +9,8 @@ def test_average_across_direction():
     # gradient recording with 100 samples and 10 channels
     rec_arr = np.arange(6, dtype="float32")[None, :] * np.ones((100, 1))
 
-    # test geometry with chans at the same y position
-    geom = np.zeros((6, 2), dtype="float32")
-    geom[[0, 1], 1] = 1
-    geom[[2, 3], 1] = 2
-    geom[[4, 5], 1] = 3
-
-    # have the x position change as well to test the geometry
-    geom[[4, 5], 0] = [1, 2]
+    # construct a 2 col geometry
+    geom = np.array([[0, 1], [1, 1], [0, 2], [1, 2], [0, 3], [1, 3]])
 
     rec = NumpyRecording(rec_arr, 10)
     rec.set_dummy_probe_from_locations(geom)
@@ -30,8 +24,10 @@ def test_average_across_direction():
     assert np.all(traces[:, 1] == 2.5)
     assert np.all(traces[:, 2] == 4.5)
     geom_avgy = rec_avgy.get_channel_locations()
-    assert np.all(geom_avgy[:2, 0] == 0)
-    assert np.all(geom_avgy[2, 0] == 1.5)
+    assert np.all(geom_avgy[:, 0] == 0.5)
+    assert geom_avgy[0, 1] == 1.0
+    assert geom_avgy[1, 1] == 2.0
+    assert geom_avgy[2, 1] == 3.0
 
     # test with channel ids
     # use chans at y in (1, 2)
@@ -43,13 +39,14 @@ def test_average_across_direction():
     # test averaging across x
     rec_avgx = average_across_direction(rec, direction="x")
     traces = rec_avgx.get_traces()
-    assert traces.shape == (100, 3)
+    assert traces.shape == (100, 2)
     # correct averages
     # the chans at x=0 are [0, 1, 2, 3]
-    assert rec_avgx.get_channel_ids()[0] == "0-1-2-3"
-    assert np.all(traces[:, 0] == 1.5)
-    assert np.all(traces[:, 1] == 4)
-    assert np.all(traces[:, 2] == 5)
+    assert rec_avgx.get_channel_ids()[0] == "0-2-4"
+    assert rec_avgx.get_channel_ids()[1] == "1-3-5"
+
+    assert np.all(traces[:, 0] == 2)
+    assert np.all(traces[:, 1] == 3)
 
     # int16 test
     rec = NumpyRecording(rec_arr.astype("int16"), 10)
--- a/src/spikeinterface/preprocessing/tests/test_depth_order.py
+++ b/src/spikeinterface/preprocessing/tests/test_depth_order.py
@@ -20,20 +20,23 @@ def test_depth_order():
 
     # geometry flat -- needs to be stable!
     geom = np.zeros((5, 2), dtype="float32")
+    geom[:, 0] = np.arange(5)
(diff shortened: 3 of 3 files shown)


src/spikeinterface/curation/curation_model.py
```diff
@@ -0,0 +1,144 @@
...
+        return v
+
+    @field_validator("label_definitions", mode="before")
+    def add_label_definition_name(cls, v):
+        if v is None:
+            v = {}
+        else:
+            v_copy = v.copy()
+            for key in v_copy.keys():
+                v[key]["name"] = key
```
alejoe91, line 43:
```suggestion
            for key in list(v.keys()):
                v[key]["name"] = key
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1138,18 +1138,18 @@ def remove_units(self, remove_unit_ids, format="memory", folder=None) -> "Sortin
...
-        folder=None,
-        verbose=False,
+        merge_unit_groups: list[list[str | int]] | list[tuple[str | int]],
+        new_unit_ids: list[int | str] | None = None,
+        censor_ms: float | None = None,
+        merging_mode: str = "soft",
+        sparsity_overlap: float = 0.75,
+        new_id_strategy: str = "append",
+        return_new_unit_ids: bool = False,
+        format: str = "memory",
```
zm711, line 1148: One general question that will matter for typing going forward. We have been moving toward doing:

`"append" | "new"` for typing but type analysis programs don't like this so I assume pydantic won't either. `str` however is not accurate either because it doesn't expect any string, but specific strings. So in this case should we move the library over to 
`Literal['append' | 'new']`

zm711: I don't know. On vscode I only get a warning saying that the type "append" | "new" are not defined. And others (I think Heberto) have commented about why not use Literal['append' | 'new'] so maybe he is seeing the typing warning too. I just want to make sure we fit in the pydantic model but also be useful to the end user. Saying `str` is not useful to the end-user that uses type hints because it is actually a Literal. I think adding Literal clutters stuff, but if we are now relying on a tool that expects Literal then we have to use it and we should move the whole code base in that direction for consistency. 

I think the static type analysis programs think that "append" should be a type because we are not specifying it is a literal. So although python allows it, I think static type checkers don't know what to do with it. It is a little similar to the `Optional`, `optional` debate in type hinting in python.

alejoe91: Actually, I think using `"append" | "new"` is not supported...

zm711: That is exactly what I'm saying!

I prefer it, but it is not supported. So we need to switch! I don't want us to switch to `str` I want us to switch to `Literal`.

alejoe91: Got it! That makes sense to me :)

alejoe91:
@samuelgarcia this is now only the new pydantic model, including splitting, and a general validation clean up in the curation format

alejoe91:
ready to review

samuelgarcia:
Hi.
I read it very quickly.
I have the feeling taht you also also changes the format itself.
Now the merges are list of dict which certainly better.
Can we have a call for this ? 

Then we should have a clear  v2 in the format.
We need too handle backward compatibility.

samuelgarcia:
Could you also change the curation.rst format ?

src/spikeinterface/curation/curation_model.py
```diff
@@ -0,0 +1,360 @@
+from pydantic import BaseModel, Field, model_validator, field_validator
+from typing import List, Dict, Union, Optional, Literal, Tuple
+from itertools import chain, combinations
+import numpy as np
+
+
+class LabelDefinition(BaseModel):
+    name: str = Field(..., description="Name of the label")
+    label_options: List[str] = Field(..., description="List of possible label options", min_length=2)
```
samuelgarcia: We we really need this `...` everywhere ?
I think this is clear that they are all mandatory expecet whe there is a default no ?
Is this common pydantic ?

src/spikeinterface/curation/curation_model.py
```diff
@@ -0,0 +1,360 @@
...
+    split_mode: Literal["indices", "labels"] = Field(
+        default="indices",
+        description=(
+            "Mode of the split. The split can be defined by indices or labels. "
+            "If indices, the split is defined by the a list of lists of indices of spikes within spikes "
+            "belonging to the unit (`split_indices`). "
+            "If labels, the split is defined by a list of labels for each spike (`split_labels`). "
+        ),
+    )
+    split_indices: Optional[Union[List[List[int]]]] = Field(default=None, description="List of indices for the split")
```
samuelgarcia: why Union ?

samuelgarcia: how does this play with numpy array ?

src/spikeinterface/curation/curation_model.py
```diff
@@ -0,0 +1,360 @@
...
+                        else:
+                            raise ValueError(f"Manual label {unit_id} has no value for label {label}")
+                if unit_id not in unit_ids:
+                    raise ValueError(f"Manual label unit_id {unit_id} is not in the unit list")
+                manual_labels[i] = manual_label
+            values["manual_labels"] = manual_labels
+        return values
+
+    @classmethod
+    def check_merges(cls, values):
```
samuelgarcia, line 101: The structure of split_data should be desribe by mode here.
The split_data type is unclear.

src/spikeinterface/curation/curation_model.py
```diff
@@ -0,0 +1,360 @@
...
+            # Check new unit id not already used
+            if merge.merge_new_unit_id is not None:
+                if merge.merge_new_unit_id in unit_ids:
+                    raise ValueError(f"New unit ID {merge.merge_new_unit_id} is already in the unit list")
+
+        values["merges"] = merges
+        return values
+
+    @classmethod
+    def check_splits(cls, values):
```
samuelgarcia, line 156: The structure of split_data should be desribe by mode here.
The split_data type is unclear.

src/spikeinterface/curation/curation_model.py
```diff
@@ -0,0 +1,360 @@
...
+
+
+class ManualLabel(BaseModel):
+    unit_id: Union[int, str] = Field(..., description="ID of the unit")
+    labels: Dict[str, List[str]] = Field(..., description="Dictionary of labels for the unit")
+
+
+class Merge(BaseModel):
+    merge_unit_group: List[Union[int, str]] = Field(..., description="List of groups of units to be merged")
+    merge_new_unit_id: Optional[Union[int, str]] = Field(default=None, description="New unit IDs for the merge group")
```
samuelgarcia: If we now have a nested dict why not 
```suggestion
    group: List[Union[int, str]] = Field(..., description="List of groups of units to be merged")
    new_unit_id: Optional[Union[int, str]] = Field(default=None, description="New unit IDs for the merge group")
```
because it is obvisouly a merge. no ?

src/spikeinterface/curation/curation_model.py
alejoe91: The `...` is needed for required fields

src/spikeinterface/curation/curation_model.py
alejoe91: good catch, union is useless here

src/spikeinterface/curation/curation_model.py
alejoe91: yep!

src/spikeinterface/curation/curation_model.py
alejoe91: mmm, no It should be `Union[List[int], List[List[int]]]`: The `List[int]` is for `labels` mode, the `List[List[int]]` for `indices` mode

src/spikeinterface/curation/curation_model.py
alejoe91: done, also for splits

src/spikeinterface/curation/curation_model.py
alejoe91: done

samuelgarcia (approved):

PR #3767, opened 2025-03-13
Allow `aggregate_channels` to accept a dict of recordings

chrishalcrow:
Related to #3711

Allows `aggregate_channels` to accept a dict of recordings. This allows for the workflow (once #3711 is merged) of the kind:

``` python
split_recordings = recording.split_by('group')
preprocessed_recordings = si.common_reference(si.bandpass_filter(split_recordings))
aggregated_preprocessing_recording = si.aggregate_channels(preprocessed_recordings)
```

or the truly horrifying
``` python
preprocessed_recording =  si.aggregate_channels(si.common_reference(si.bandpass_filter(recording.split_by('group'))))
```

edit: PR is mostly new tests

### DIFF
--- a/src/spikeinterface/core/channelsaggregationrecording.py
+++ b/src/spikeinterface/core/channelsaggregationrecording.py
@@ -25,7 +25,10 @@ class ChannelsAggregationRecording(BaseRecording):
                 # If default 'group'ing (all equal 0), we label the recordings using the 'group' property
                 recording_groups = []
                 for recording in recording_list:
-                    recording_groups.extend(recording.get_property("group"))
+                    if (group_property := recording.get_property("group")) is not None:
+                        recording_groups.extend(group_property)
+                    else:
+                        recording_groups.extend([0])
                 if np.all(np.unique(recording_groups) == np.array([0])):
                     for group_id, recording in enumerate(recording_list):
                         recording.set_property("group", group_id * np.ones(recording.get_num_channels()))
--- a/src/spikeinterface/core/channelsaggregationrecording.py
+++ b/src/spikeinterface/core/channelsaggregationrecording.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+import warnings
 
 import numpy as np
 
@@ -236,13 +237,17 @@ class ChannelsAggregationRecordingSegment(BaseRecordingSegment):
         return np.concatenate(traces, axis=1)
 
 
-def aggregate_channels(recording_list, renamed_channel_ids=None):
+def aggregate_channels(
+    recording_list_or_dict=None,
+    renamed_channel_ids=None,
+    recording_list=None,
+):
     """
     Aggregates channels of multiple recording into a single recording object
 
     Parameters
     ----------
-    recording_list: list | dict
+    recording_list_or_dict: list | dict
         List or dict of BaseRecording objects to aggregate.
     renamed_channel_ids: array-like
         If given, channel ids are renamed as provided.
@@ -252,4 +257,13 @@ def aggregate_channels(recording_list, renamed_channel_ids=None):
     aggregate_recording: ChannelsAggregationRecording
         The aggregated recording object
     """
-    return ChannelsAggregationRecording(recording_list, renamed_channel_ids)
+
+    if recording_list is not None:
+        warnings.warn(
+            "`recording_list` is deprecated and will be removed in 0.105.0. Please use `recording_list_or_dict` instead.",
+            category=DeprecationWarning,
+            stacklevel=2,
+        )
+        recording_list_or_dict = recording_list
+
+    return ChannelsAggregationRecording(recording_list_or_dict, renamed_channel_ids)
--- a/src/spikeinterface/core/channelsaggregationrecording.py
+++ b/src/spikeinterface/core/channelsaggregationrecording.py
@@ -17,27 +17,21 @@ class ChannelsAggregationRecording(BaseRecording):
 
         if isinstance(recording_list_or_dict, dict):
             recording_list = list(recording_list_or_dict.values())
+            recording_ids = list(recording_list_or_dict.keys())
         elif isinstance(recording_list_or_dict, list):
             recording_list = recording_list_or_dict
-
-            # Check if the recordings were previously split using `split_by`
-            if recording_list[0].get_annotation("split_by_property") is None:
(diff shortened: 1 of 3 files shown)


src/spikeinterface/core/channelsaggregationrecording.py
```diff
@@ -13,7 +13,16 @@ class ChannelsAggregationRecording(BaseRecording):
 
     """
 
-    def __init__(self, recording_list, renamed_channel_ids=None):
+    def __init__(self, recording_list_or_dict, renamed_channel_ids=None):
+
+        if isinstance(recording_list_or_dict, dict):
+            recording_list = list(recording_list_or_dict.values())
```
alejoe91, line 19:
```suggestion
            recording_list = list(recording_list_or_dict.values())
```
In this case, I think it would make sense to check if each recording's group corresponds to the key.
If all groups are 0 (which is default), we could set channel groups according to keys

chrishalcrow: At the moment there is nothing special about "group". As in, a user can do
```
rec.split_by("user_defined_label")
```
And we don't keep track of the label since `split_by` just returns a dictionary. So this would make "group" a special label. I think this is ok: we can recommend users use it when we document `split_by`?

alejoe91: You're abslutely right. We could actually track this as an annotation though :)

chrishalcrow: Oh nice! Lemme have a go...

src/spikeinterface/core/channelsaggregationrecording.py
```diff
@@ -13,7 +13,29 @@ class ChannelsAggregationRecording(BaseRecording):
...
+                # If default 'group'ing (all equal 0), we label the recordings using the 'group' property
+                recording_groups = []
+                for recording in recording_list:
+                    if (group_property := recording.get_property("group")) is not None:
+                        recording_groups.extend(group_property)
+                    else:
+                        recording_groups.extend([0])
+                if np.all(np.unique(recording_groups) == np.array([0])):
+                    for group_id, recording in enumerate(recording_list):
+                        recording.set_property("group", group_id * np.ones(recording.get_num_channels()))
```
alejoe91, line 34: This should be done after the `else`, because it applies to both dicts and lists inputs. Right?

src/spikeinterface/core/tests/test_channelsaggregationrecording.py
```diff
@@ -94,6 +126,37 @@ def test_channel_aggregation_preserve_ids():
     assert list(aggregated_recording.get_channel_ids()) == ["a", "b", "c", "d", "e"]
 
 
+def test_aggretion_labelling_for_lists():
```
alejoe91, line 129:
```suggestion
def test_aggregation_labeling_for_lists():
```

src/spikeinterface/core/channelsaggregationrecording.py
```diff
@@ -13,7 +13,29 @@ class ChannelsAggregationRecording(BaseRecording):
...
+            recording_list = list(recording_list_or_dict.values())
+        elif isinstance(recording_list_or_dict, list):
+            recording_list = recording_list_or_dict
+
+            # Check if the recordings were previously split using `split_by`
+            if recording_list[0].get_annotation("split_by_property") is None:
+                # If default 'group'ing (all equal 0), we label the recordings using the 'group' property
+                recording_groups = []
+                for recording in recording_list:
+                    if (group_property := recording.get_property("group")) is not None:
```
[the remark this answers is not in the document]
alejoe91: Actually, maybe this is the case only when setting a probe, so ok to leave it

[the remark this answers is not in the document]
chrishalcrow: Yeah - the generated recordings have `None` for instance.

src/spikeinterface/core/channelsaggregationrecording.py
chrishalcrow: I think we're both right: we should use the dictionary keys for `group` if we don't know how the recording was split up. I've added this feature.

alejoe91 (approved):

src/spikeinterface/core/channelsaggregationrecording.py
```diff
@@ -13,10 +13,26 @@ class ChannelsAggregationRecording(BaseRecording):
 
     """
 
-    def __init__(self, recording_list, renamed_channel_ids=None):
+    def __init__(self, recording_list_or_dict, renamed_channel_ids=None):
```
zm711, line 16: Do we want a deprecation for the argument for people that use keyword. So in that case it would be 

```
__init__(self, recording_list_or_dict, renamed_channels_ids=None, recording_list=None):
```

That way we can deprecated over a couple version. Especially for something that is in core. No?

src/spikeinterface/core/channelsaggregationrecording.py
```diff
@@ -207,8 +242,8 @@ def aggregate_channels(recording_list, renamed_channel_ids=None):
 
     Parameters
     ----------
-    recording_list: list
-        List of BaseRecording objects to aggregate
+    recording_list: list | dict
```
zm711, line 245: why not changing name here if changed above? I understand the annoyance, but if we are planning on changing the parameter of above doesn't it make sense to be consistent? (again with the deprecation in my opinion).

zm711 (commented):
Two small comments (I think)

src/spikeinterface/core/channelsaggregationrecording.py
chrishalcrow: Oh nice, good catch!

src/spikeinterface/core/channelsaggregationrecording.py
chrishalcrow: Oh, just ignorance and sloppiness from me! Will fix!

src/spikeinterface/core/channelsaggregationrecording.py
zm711: If we've made it clear that this part is private them I'm okay with not changing it! I only looked at the diff not the surrounding :)

zm711 (approved):
Me too

samuelgarcia:
Parfait.

PR #3768, opened 2025-03-13
Fix KS4 tests

alejoe91:
Versions 4.0.26/27 are buggy. v4.0.28 added a new `verbose` param ot some functions!

### DIFF
--- a/.github/scripts/check_kilosort4_releases.py
+++ b/.github/scripts/check_kilosort4_releases.py
@@ -1,10 +1,9 @@
 import os
-import re
 from pathlib import Path
 import requests
 import json
 from packaging.version import parse
-import spikeinterface
+
 
 def get_pypi_versions(package_name):
     """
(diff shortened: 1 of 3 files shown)


zm711 (approved):
This also looks fine to me, but didn't we have a discussion about ratcheting up the version we support so that we only support the last few versions. Or did we change that policy? I guess if the burden isn't too high then supporting as many as possible is nice.

chrishalcrow (approved):
Looks good, and works for me.

JoeZiminski:
Looks good to me! @zm711 I think that would make sense, if people are desperate to use an older version for some reason then they can use an older version of SI that supports it. 

As you say I guess if old versions are passing the CI tests then might as well allow their use, but if for some reason it's a pain for support a old version I don't think we should put much effort into it.

alejoe91:
Yeah. One thing we have to consider for dropping versions is to make sure the containers are updated..but this is currently manual. We might need some similar CI to auto-build/auto-deploy

zm711:
You're doing all of that right Alessio? I definitely think at this point we should find a way to automate that to lessen your burden, especially for something as active as KS4 is.

alejoe91:
> You're doing all of that right Alessio? I definitely think at this point we should find a way to automate that to lessen your burden, especially for something as active as KS4 is.

Yes. If we can automate some parts of it it would be great! We also have some outdated Py3.8 images there...

PR #3772, opened 2025-03-14
Update `load_motion_info` to load legacy motion folders

alejoe91:
When we added the `Motion.load` to the `load_motion_info`, we forgot to handle back-compatibility.

This PR just tries to load a legacy motion in case the `Motion.load` fails.

### DIFF
--- a/src/spikeinterface/preprocessing/motion.py
+++ b/src/spikeinterface/preprocessing/motion.py
@@ -513,14 +513,14 @@ def load_motion_info(folder):
         else:
             motion_info[name] = None
 
-    try:
+    if (folder / "motion").is_dir():
         motion = Motion.load(folder / "motion")
-    except IOError as e:
-        warnings.warn("The Motion object could not be loaded. Trying to load the legacy format")
+    else:
+        warnings.warn("Trying to load Motion from the legacy format")
         required_files = ["spatial_bins.npy", "temporal_bins.npy", "motion.npy"]
         for required_file in required_files:
             if not (folder / required_file).is_file():
-                raise IOError("The provided folder is not a valid legacy motion folder")
+                raise IOError("The provided folder is not a valid motion folder")
         spatial_bins_um = np.load(folder / "spatial_bins.npy")
         temporal_bins_s = [np.load(folder / "temporal_bins.npy")]
         displacement = [np.load(folder / "motion.npy")]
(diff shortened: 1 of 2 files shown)


zm711 (commented):
I know Sam prefers bool check over doing try/except (at least he seems to). What if we flipped the logic and checked for the legacy files and if all legacy files are present we load legacy. If not we load the current format and let the current format raise the IOError if its neither type. try/except are expensive if the except is used. This is probably the simplest version, but just wanted to raise the point.

Otherwise this looks fine to me.

EDIT: error appears to be related to building scipy. Not this PR.

alejoe91:
Good point @zm711 

Got rid of try-except :)

zm711:
Looks good to me. You want to do another PR to figure out the scipy build issue?

And do we have any way to test? Or do we not have a way to generate (simulated) legacy motion?

src/spikeinterface/preprocessing/motion.py
```diff
@@ -511,6 +513,19 @@ def load_motion_info(folder):
...
+        warnings.warn("Trying to load Motion from the legacy format")
+        required_files = ["spatial_bins.npy", "temporal_bins.npy", "motion.npy"]
+        for required_file in required_files:
+            if not (folder / required_file).is_file():
+                raise IOError("The provided folder is not a valid motion folder")
+        spatial_bins_um = np.load(folder / "spatial_bins.npy")
+        temporal_bins_s = [np.load(folder / "temporal_bins.npy")]
+        displacement = [np.load(folder / "motion.npy")]
+
+        motion = Motion(displacement=displacement, temporal_bins_s=temporal_bins_s, spatial_bins_um=spatial_bins_um)
```
samuelgarcia, line 528:
```suggestion
        motion = Motion(displacement=displacement, temporal_bins_s=[temporal_bins_s], spatial_bins_um=[spatial_bins_um])
```

alejoe91: No :) I'm ok making the list later if it's better readable

alejoe91: Actually, let's leave it as is!

zm711:
The failure was ibl? Struggles.

samuelgarcia (approved):

PR #3776, opened 2025-03-14
Playing with `pytest-xdist`

JoeZiminski:
This may have been tried before, there is a package `pytest-xdist` which can run tests in parallel. In the past I've tried this and had concurrency issues parallelising tests within the same module, but realized they also have a parallelise-by-module option. Just testing it out to see here to see whether it speeds up tests on the CI.


JoeZiminski:
The full tests took like 30 minutes longer 😆 I guess each individual test group (e.g. preprocessing, sorting) takes 1-2 minutes, so paralleling within this wont help much, I guess the setup time for the processes is longer than the gain across the 2-4 nodes on CI!

zm711:
Haha. Nice try @JoeZiminski. Would've been cool if it worked. Especially since Windows is still pretty slow.

PR #3779, opened 2025-03-17
add `load_extension()` in `MockWaveformExtractor` for backwards compatibility

guptadivyansh:
Looks like this one function got missed while adding backward compatibility for the old WaveformExtractor extensions.

Without this fix, I get the error:

```
waves = si.load_waveforms(paths['si_waveforms'], with_recording=False, sorting = sorting)
locs = waves.load_extension('unit_locations').get_data()


AttributeError: 'MockWaveformExtractor' object has no attribute 'load_extension'
```

### DIFF
--- a/src/spikeinterface/core/waveforms_extractor_backwards_compatibility.py
+++ b/src/spikeinterface/core/waveforms_extractor_backwards_compatibility.py
@@ -248,6 +248,9 @@ class MockWaveformExtractor:
     def has_extension(self, extension_name: str) -> bool:
         return self.sorting_analyzer.has_extension(extension_name)
 
+    def load_extension(self, extension_name: str):
+        return self.sorting_analyzer.load_extension(extension_name)
+
     def select_units(self, unit_ids):
         return self.sorting_analyzer.select_units(unit_ids)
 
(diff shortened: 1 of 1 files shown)


alejoe91:
This is because all extensions are loaded at init because some need a bit of restructuring. I recommend using the `get_extension` function downstream, which should work for both the MockWaveformExtractor and SortingAnalyzer

guptadivyansh:
I was avoiding rewriting the code. So I guess backwards compatibility
refers to being able to use data that was analyzed with the old versions,
not that old scripts will continue to work as-is with the new versions?

I can close the PR if you think adding this can do more harm than good
going forward.

src/spikeinterface/core/waveforms_extractor_backwards_compatibility.py
```diff
@@ -248,6 +248,9 @@ def format(self):
     def has_extension(self, extension_name: str) -> bool:
         return self.sorting_analyzer.has_extension(extension_name)
 
+    def load_extension(self, extension_name: str):
+        return self.sorting_analyzer.load_extension(extension_name)
```
alejoe91, line 252:
```suggestion
        return self.sorting_analyzer.get_extension(extension_name)
```
All extensions are pre-loaded

alejoe91 (approved):

guptadivyansh:
Awesome, thank you @alejoe91 !

PR #3780, opened 2025-03-18
exclude unicode subtypes from type checking in unitsaggregationsorting.py

RobertoDF:
Hi,

I was aggreagating 2 sortings, one had quality types "mua", "sua" dtype('<U3'), the other one included a "good" so dtype('<U4'). For this reason "quality" was skipped (  print(f"Skipping property '{prop_name}: difference in dtype between sortings'")). With this modification the code will avoid distinguishing by number of charachters.

### DIFF
--- a/src/spikeinterface/core/unitsaggregationsorting.py
+++ b/src/spikeinterface/core/unitsaggregationsorting.py
@@ -88,7 +88,7 @@ class UnitsAggregationSorting(BaseSorting):
                     existing_dtype = property_keys[prop_name]
                     new_dtype = sort.get_property(prop_name).dtype
                     if existing_dtype != new_dtype:
-                        if not (existing_dtype.kind == 'U' and new_dtype.kind == 'U'):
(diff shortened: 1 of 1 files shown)


chrishalcrow:
Hey @RobertoDF - I found the same bug on Friday and have a PR open #3778 to fix it too! I don't know which
 implementation is better, but I'd recommend looking at the changes I made to the tests.

RobertoDF:
Ah! what a coincindence, I also just saw ur comment here: _"Another idea: we check if it's a string, and deal with that in a special way. Then for the non-strings, we do a similar type check that already existed."_

chrishalcrow:
Hello, these problems have been cropping up a lot so we discussed making a general strategy at the last maintainers meeting. We've agreed to use `dtype.kind` to check for type equivalence, then let numpy try to concatenate. I had a go at implementing that in the other PR (#3778); and I ended up refactoring the tests and code a bit. So we'll merge that PR and not this one. Thanks for highlighting the problem though @RobertoDF !

PR #3781, opened 2025-03-18
Add `_precomputable_kwarg_names` to `BaseExtractor`

chrishalcrow:
As discussed in https://github.com/SpikeInterface/spikeinterface/pull/3685#discussion_r1954036875 and in maintainer meetings, it would be handy to have a `_precomputable_kwargs` in the `BaseExtractor`, especially for Preprocessors.

These are kwargs which can be precomputed for use in a preprocessor. For instance, if you know the whitening matrices, you can pass these directly instead of computing them. If you know the bad channel ids, you can pass these to `RemoveBadChannels` (once it exists #3685) and if you know your motion object, you could pass this to a `CorrectMotion` class (once this exists too!).

This can also be used when saving preprocessed recordings. If we save the provenance and the precomputed data, we can load a preprocessed recording much faster. This will be implemented in a later PR.

This PR has no functionality so there are no new tests.

**Questions**:
- Naming. @alejoe91  had suggested on `_precomputed_kwargs` but I think this is confusing since these kwargs aren't always precomputed, it's just that they _can_ be precomputed. So I've gone for `_precomputable_kwargs`. Opinions?
- Should this be in `BaseExtractor`? Could be in `BasePreprocessor`? Any use cases for it in scenarios other than preprocessing?

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -54,6 +54,9 @@ class BaseExtractor:
         # store init kwargs for nested serialisation
         self._kwargs = {}
 
+        # kwargs which can be precomputed before being used by the extractor
+        self._precomputable_kwargs = []
+
         # "main_ids" will either be channel_ids or units_ids
         # They are used for properties
         self._main_ids = np.array(main_ids)
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -48,7 +48,7 @@ class BaseExtractor:
     _skip_properties = []
 
     # kwargs which can be precomputed before being used by the extractor
-    _precomputable_kwargs = []
+    _precomputable_kwarg_names = []
 
     installation_mesg = ""
     installed = True
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -47,15 +47,15 @@ class BaseExtractor:
     # these properties are skipped by default in copy_metadata
     _skip_properties = []
 
+    # kwargs which can be precomputed before being used by the extractor
+    _precomputable_kwargs = []
+
     installation_mesg = ""
     installed = True
 
     def __init__(self, main_ids: Sequence) -> None:
         # store init kwargs for nested serialisation
         self._kwargs = {}
-
-        # kwargs which can be precomputed before being used by the extractor
-        self._precomputable_kwargs = []
 
         # "main_ids" will either be channel_ids or units_ids
         # They are used for properties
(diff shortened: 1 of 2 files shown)


alejoe91:
@samuelgarcia ok to you?

alejoe91 (approved):

h-mayorquin:
My two cents: 
1) Kwargs is usually used (in this codebase as well) for dict objects. I think `._precomputable_kwargs` is a good name but it should be a dict and follow kwarg conventions.
2) I say isolate to `BasePocessor` and then generalize when and if we needed. It is good to keep complexity away from the core as long as we can. 


I can't really say a lot because I have not seen it in use. Perhaps a use case other than #3781 is the estimated noise and std of recordings. We already have some mechanism to treat that as special quantities and maybe they should be unified with this.Also, how will this work with the json serialization is not clear to me. Maybe you discussed this on the meeting that I missed.

All of that said, these is a private attribute so I am not too concerned if we need to change it later.

alejoe91:
> My two cents:
> 
> 1. Kwargs is usually used (in this codebase as well) for dict objects. I think `._precomputable_kwargs` is a good name but it should be a dict and follow kwarg conventions.
> 2. I say isolate to `BasePocessor` and then generalize when and if we needed. It is good to keep complexity away from the core as long as we can.
> 
> I can't really say a lot because I have not seen it in use. Perhaps a use case other than #3781 is the estimated noise and std of recordings. We already have some mechanism to treat that as special quantities and maybe they should be unified with this.Also, how will this work with the json serialization is not clear to me. Maybe you discussed this on the meeting that I missed.
> 
> All of that said, these is a private attribute so I am not too concerned if we need to change it later.

What about `_precomputable_kwarg_names`?

chrishalcrow:
When we discussed, we agreed _not_ to use this to store any kwargs themselves, just the information about which kwargs could be precomputed. So I think a list is the correct structure; very happy to call `_precomputable_kwarg_names`.

I'm indifferent to putting it in `BaseExtractor` or `BaseProcessor`.

alejoe91:
> When we discussed, we agreed _not_ to use this to store any kwargs themselves, just the information about which kwargs could be precomputed. So I think a list is the correct structure; very happy to call `_precomputable_kwarg_names`.
> 
> I'm indifferent to putting it in `BaseExtractor` or `BaseProcessor`.

Yes, but what @h-mayorquin is saying is that the name `_precomputable_kwargs` suggests that it contains kwargs (a dict).

h-mayorquin (approved):
OK, we will see how it is used.

It could also be kwargs_keys to keep with the dict semantics but names is just fine if not more expressive.

samuelgarcia (approved):

samuelgarcia:
super

Issue #3783, opened 2025-03-18
[Question] Annotating sorted spikes with channel positions


lcui809:
Hi there,

I have a list of sorted spikes and would like to annotate each spike with the channel position where it was originally generated. I have already computed the spike locations using sorting_analyzer, which has been very useful. However, after considerable effort, I still haven't figured out how to link each sorted spike to its source channel position.

I am using an HD-MEA from Maxwell. Any guidance would be greatly appreciated!

chrishalcrow:
Hi @lcui809, I thought we didn't have a method for this, but digging into the code, we do! It's not documented, so I'll add some documentation soon.

Turns out you can pass the `peak_channel` method when you compute spike locations. This will give you the location of the channel with the peak waveform on it. If you have a `sorting_analyzer` called `sa`, you can do this like so:

``` python
sa.compute({'spike_locations': {'method': 'peak_channel'}})
print(sa.get_extension('spike_locations').get_data())
>>> array([( 0., 20.), (20., 20.), ( 0.,  0.), ..., ( 0.,  0.), ( 0., 20.),
       ( 0., 20.)], dtype=[('x', '<f8'), ('y', '<f8')])
```

Hope that works!

zm711:
My deeper question is why do you care about the channel itself? One of the benefits of the fancier methods is to estimate the "true" position rather than just find the nearest channel which could be 10s of microns away. Is there a need for channel vs "estimated" true location which is likely more accurate unless a neuron is right on top of a channel?

yger:
Note sure to follow here. Internally, in SI, peaks are attached to a channel (because we need this information to extract sparse mask and do efficient computations within components), and spikes are not. However, given spikes, one can estimate their putative positions depending on various methods (grid/monopolar/barycenter), and I think this is more relevant at the end of the pipelines than working at the channel level. Otherwise, as @chrishalcrow found, we can get the spike locations at the channel levels, and not with any fancy inference scheme

lcui809:
Thanks, everyone, for answering my question — your insights are extremely helpful!

@chrishalcrow  Hi Chris, thanks again for resolving my question. I have a follow-up if I may.
Would it be possible to extract waveform features (e.g., peak-to-valley) or waveform metrics for all spikes? We're considering some downstream analyses, such as grouping spikes by frequency and extracting waveform features from different subgroups within the same unit. If this is feasible, could you please point me to the relevant NPY file where these metrics can be found?

[...]

PR #3784, opened 2025-03-19
Add `peak_channel` to `compute_spike_locations` docstring and docs

chrishalcrow:
Hello, we have a nice undocumented feature: compute spike locations by peak channel! This PR adds the documentation. Thanks to #3783!

**Question** @samuelgarcia @yger : for spike locations the peak_channel method is called `peak_channel`. For unit locations, the peak_channel method is called `max_channel`. I think the `peak_channel` is only just in the SortingComponents module. Maybe we can unify these names now and just use `max_channel` (this one has been documented for a while, so some users might be using it already, so I'd vote to keep this one).

### DIFF
--- a/doc/modules/postprocessing.rst
+++ b/doc/modules/postprocessing.rst
@@ -256,8 +256,9 @@ spike_locations
 
 This extension estimates the location of each spike in the sorting output. Spike location estimates can be done
 with center of mass (:code:`method="center_of_mass"` - fast, but less accurate), a monopolar triangulation
-(:code:`method="monopolar_triangulation"` - slow, but more accurate), or with the method of grid convolution
-(:code:`method="grid_convolution"`)
+(:code:`method="monopolar_triangulation"` - slow, but more accurate), the method of grid convolution
+(:code:`method="grid_convolution"`), or by finding the location of the channel containing the extremum amplitude
+(:code:`method="peak_channel"`).
 
 **NOTE:** computing spike locations is required to compute :ref:`drift_metrics`.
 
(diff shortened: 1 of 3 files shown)


zm711:
I guess the only concern with max is that it implies positive values vs min. Whereas peak could be pos peak or neg peak if its not specified.

chrishalcrow:
> I guess the only concern with max is that it implies positive values vs min. Whereas peak could be pos peak or neg peak if its not specified.

Yeah, prefer `peak_channel` too.

samuelgarcia:
and what about `extremum_channel`

zm711:
I'm fine with `extremum` too although it sounds more "mathy" to the less comp/more basic neuro people. :)

doc/modules/postprocessing.rst
```diff
@@ -285,7 +286,7 @@ unit_locations
 
 This extension is similar to the :code:`spike_locations`, but instead of estimating a location for each spike
 based on individual waveforms, it calculates at the unit level using templates. The same localization methods
-(:code:`method="center_of_mass" | "monopolar_triangulation" | "grid_convolution"`) are available.
+(:code:`method="center_of_mass" | "monopolar_triangulation" | "grid_convolution" | "max_channel"`) are available.
```
alejoe91, line 289:
```suggestion
(:code:`method="center_of_mass" | "monopolar_triangulation" | "grid_convolution" | "peak_channel"`) are available.
```
I would go with `peak_channel`

h-mayorquin:
peak channel makes sense to me. Both extremum and peak are good but we already use peak here and there on the library.

chrishalcrow:
Ok, consensus seems to be `peak`, so now unified localisation method names to `peak_channel`.

h-mayorquin:
I am fine with delaying, thanks for checking @samuelgarcia!

chrishalcrow:
Ok, thanks for the info Sam

PR #3786, opened 2025-03-20
Add job kwargs to KS4

alejoe91:
Fixes #3785


zm711:
Was this a discussion issue ? I can't find the link to an issue that you're fixing. Let the record show I argued for this from the beginning so I'm happy to approve this, but I'm unsure of the context?

alejoe91:
It was linked to a Discussion, whcih I closed :)

The think is that if `use_binary_file` is used, but the recording is not binary, we should propagate the job kwargs.

zm711 (approved):
Works for me.

PR #3787, opened 2025-03-20
Update tj-actions changes-files for security

h-mayorquin:
Here. After we do this we should change the keys.

### DIFF
--- a/.github/workflows/issue-on-change-matlab.yml
+++ b/.github/workflows/issue-on-change-matlab.yml
@@ -18,7 +18,7 @@ jobs:
         uses: actions/checkout@v4
       - name: Get changed files
         id: changed-files
-        uses: tj-actions/changed-files@v41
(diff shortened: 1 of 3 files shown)


.github/workflows/issue-on-change-matlab.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         uses: actions/checkout@v4
       - name: Get changed files
         id: changed-files
-        uses: tj-actions/changed-files@v41
+        uses: tj-actions/changed-files@v46
```
alejoe91, line 21:
```suggestion
        uses: tj-actions/changed-files@v46.0.1
```
?

h-mayorquin: Yes

alejoe91 (approved):

alejoe91:
Changed the API keys ;) Thanks Ramon!

PR #3788, opened 2025-03-20
Better naming for SI units 

h-mayorquin:
@chrishalcrow 

Discussed on the meeting of 20 of March 2025

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/neobaseextractor.py
+++ b/src/spikeinterface/extractors/neoextractors/neobaseextractor.py
@@ -288,7 +288,7 @@ class NeoBaseRecordingExtractor(_NeoBaseExtractor, BaseRecording):
         if has_mixed_units:
             warning_msg = (
                 "Found a mix of voltage and non-voltage units. "
-                'Proceed with caution. Check channel units with `recording.get_property("original_unit")`.'
+                'Proceed with caution. Check channel units with `recording.get_property("physical_unit")`.'
             )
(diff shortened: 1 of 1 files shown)


alejoe91 (approved):

chrishalcrow:
I promise I'm not trolling: I'd say `gains_to_physical_unit` and `offsets_to_physical_unit` since the gains and offsets are plural.

@zm711 @JoeZiminski

zm711:
I don't know if there is a right answer. For intan there is one gain per stream so it is `gain` to physical unit, but as you said there is other equipment that has `gains` /`offsets` on a channel basis. Right? So the property in this case needs to be for the channels I would name this variable based on the most common thing users will see. I would see based on the issue tracker the most common equipment right now are neuropixels by far followed by intan followed by everything else. (Although if someone wants to argue this let me know what you think). So I would do whatever neuropixels do (if they have an offset per channel then do `gains` and `offsets`, if they have one gain/offset per stream then I would do `gain` and `offset`).

alejoe91:
> I promise I'm not trolling: I'd say `gains_to_physical_unit` and `offsets_to_physical_unit` since the gains and offsets are plural.
> 
> @zm711 @JoeZiminski

@chrishalcrow we have `gain_to_uV` and `offset_to_uV`, so I would be consistent in naming the property with singulars. Then each unit will have a value so in the end you'll have unit gains to uV

zm711 (approved):
I agree with Alessio!

PR #3794, opened 2025-03-21
`return_scaled` to `return_in_uV` 

h-mayorquin:
OK, so I open this PR for two reasons:

1) As you can see this will be a massive change and I think that someone (cc @samuelgarcia) really dislikes dealing with merge conflicts so maybe we could just change this in `BaseRecording` and co `get_traces` and deal with annoying deprecation messages for a while.

2) The proposed name `scale_to_uV` will clash with the preprocessing of the same name. I have chosen `return_in_UV` as the interpolation of `return_scaled` and the discussed `scale_to_uV` but I am happy to change. `return_in_UV` is regex-friendly so it should be easy.


h-mayorquin:
`scaled_to_uV` avoids the clashing.

yger:
Templates has also a field is_scaled. Should it be changed also ? If yes, sortingcomponents needs quite a lot of changes, because we are heavily using return_scaled and is_scaled

zm711:
> Templates has also a field is_scaled. Should it be changed also ? If yes, sortingcomponents needs quite a lot of changes, because we are heavily using return_scaled and is_scaled

I guess the better named attribute is `is_in_uV`. But is this suppose to be a private attribute or a public one (in the dev sense). We all know what is_scaled means so do you and Sam hope that some day people will use `Template` objects to make their own sorters? If so we should be consistent. If this is private and never should be used by others than it is less important to change.

h-mayorquin:
I saw the template thing and I wanted to discussed on the previous maintenance meeting but I forgot. In any case, I think that changes should be its own PR.

src/spikeinterface/core/testing.py
```diff
@@ -15,11 +16,41 @@ def check_sorted_arrays_equal(a1, a2):
...
+        DEPRECATED. Use return_in_uV instead.
+        If True, compare scaled traces
+    return_in_uV : bool, default: True
+        If True, compare scaled traces.
+    force_dtype : dtype, default: None
+        If not None, force the dtype of the traces before comparison
+    check_annotations : bool, default: False
+        If True, check annotations
+    check_properties : bool, default: False
+        If True, check properties
```
zm711, line 44: Tests are passing and I know a lot of like to use keyword, but others on the team tend toward using positional arguments internally. So we risk:

```python
check_recordings_equal(rec1, rec2, True, 'float32', True, False)
```
which would raise errors rather then putting the deprecation at the end which would allow things to work with positional as well and raise the error if using the wrong keyword. How are you thinking about this?

zm711: But my point is deprecating across multiple functions and I just took this one as an example.

h-mayorquin: We can move the old parameter to the end if you want.

samuelgarcia:
Genial. Merci infiniment.
Lets merge this no ?

@yger Lets make another PR after the merge for `Templates.is_scaled`  > `Templates.is_in_uV`

PR #3795, opened 2025-03-22
Cleanup imports and use `importlib.util.find_spec` rather than try-except

zm711:
try-except is expensive especially to import something like torch just to check if the user has it. What if we move more toward `importlib.util.find_spec` instead? Also cleaned up a few imports.

in draft for feedback--I could do more. For example we could change the numba check to this instead which I think would be useful. Doing a single test on my Mac showed that trying to import numba took 0.1997 seconds whereas doing the find_spec took 0.00016 seconds. They are marginal gains for sure, but having a really smooth import experience for users keeps them with a good feeling for the package. Thoughts all?

### DIFF
--- a/src/spikeinterface/sorters/external/herdingspikes.py
+++ b/src/spikeinterface/sorters/external/herdingspikes.py
@@ -95,12 +95,13 @@ class HerdingspikesSorter(BaseSorter):
 
     @classmethod
     def is_installed(cls):
-        try:
-            import herdingspikes as hs
+        import importlib.util
 
-            HAVE_HS = True
-        except ImportError:
+        spec = importlib.util.find_spec("herdingspikes")
+        if spec is None:
             HAVE_HS = False
+        else:
+            HAVE_HS = True
         return HAVE_HS
 
     @classmethod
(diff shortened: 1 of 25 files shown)


src/spikeinterface/sorters/external/herdingspikes.py
```diff
@@ -95,12 +95,13 @@ class HerdingspikesSorter(BaseSorter):
 
     @classmethod
     def is_installed(cls):
-        try:
-            import herdingspikes as hs
+        import importlib.util
```
h-mayorquin, line 98: This can be top level.

h-mayorquin (commented):
+1 on this.

I see ~10x improvements as well on my tests. 

find_spec on init and normal import at run time seems like a good policy to me.

src/spikeinterface/sorters/external/herdingspikes.py
zm711: So you think we should move all checks to top level?

zm711: sounds good. I'll move that to top level import.

zm711: Done :)

zm711:
importlib is the preferred way to do a lot of this as of 3.8 so since our minimum is 3.9 we should be good for moving everything over that direction :)

samuelgarcia:
Formidable!

alejoe91 (approved):

PR #3797, opened 2025-03-23
Switch `numba` check to `importlib.util.find_spec`

zm711:
Same idea as #3795, but moving our numba check to find_spec so that we don't waste time importing. The import is moved to run time so that only users using a numba based function pay the cost. 

I was hoping the `torch.nn.functional` check would work so I was just winging that one. I'll fix it soon :)


samuelgarcia:
genial.

alejoe91 (approved):

PR #3798, opened 2025-03-24
fix #3288

### LINKED ISSUE #3288: Erorr in `set_property` with missing values for string
The following is a bug when setting missing properties with unicode strings in `base.set_property`:

```python

num_channels = 5
recording = generate_recording(num_channels=5, durations=[1.0])
channel_ids = ["a", "b", "c", "d", "e"]
recording = recording.rename_channels(new_channel_ids=channel_ids)



incomplete_values = ["value"] * (num_channels - 1)
# Passs a missing value
recording.set_property(
    key="missing_property",
    ids=channel_ids[:-1],
    values=incomplete_values,
    missing_value="missing",
)

property_in_recording = recording.get_property("missing_property")
expected_array = np.array(incomplete_values + ["missing"])
assert np.array_equal(property_in_recording, expected_array)

```

The output in property_in_recording is:
```python
array(['value', 'value', 'value', 'value', 'missi'], dtype='<U5')
```

I found the error while working on https://github.com/SpikeInterface/spikeinterface/pull/3287. I can fix it after we are done with that one.

h-mayorquin:
Fix #3288

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -284,7 +284,7 @@ class BaseExtractor:
                 non_unique_ids = [id for id in ids if np.count_nonzero(ids == id) > 1]
                 raise ValueError(f"IDs are not unique: {non_unique_ids}")
 
-            # Not clear where this branch is used, perhaps on aggregation of extractors?
+            # Sam: this is used because you could set all ids (like the None) but with the vector.
             if ids.size < size:
                 if key not in self._properties:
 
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/base.py
```diff
@@ -306,7 +306,16 @@ def set_property(
...
+
+                    # For string (unicode) types, make sure dtype can fit the missing value
+                    if dtype_kind == "U" and missing_value is not None:
+                        max_val_len = np.max(np.char.str_len(values)) if values.size > 0 else 0
+                        missing_val_len = len(missing_value)
+                        max_len = max(max_val_len, missing_val_len)
+                        empty_values = np.full(shape, missing_value, dtype=f"<U{max_len}")
+                    else:
+                        empty_values = np.zeros(shape, dtype=dtype)
+
```
chrishalcrow, line 318: Hello, I'm maybe overlooking something, but could we just do
```
empty_values = np.full(shape, missing_value)
```
replacing all this, the the line below.

samuelgarcia (approved):

src/spikeinterface/core/base.py
chrishalcrow: There's a bit of a stringy hack just below this bit too, so I'll make a PR sometime to make both of these nicer. But good to have a bug fix in anyhow!

h-mayorquin: Thanks @chrishalcrow

PR #3811, opened 2025-03-27
Switch `print` to `warnings.warn`

zm711:
Just a quick search of the codebase fond these prints that should probably be warnings. 

I left alone a couple prints. The main type being during `verbose=True` I think it is fine to keep a print because the user is asking for additional info. If we were to have a verbose=1 vs a verbose=2 then I think we could think about a warning vs print divide, but with the current state of the code most things are warnings unless they are specifically under a `if verbose:`


src/spikeinterface/sorters/container_tools.py
```diff
@@ -102,10 +103,10 @@ def __init__(self, mode, container_image, volumes, py_user_base, extra_kwargs):
                 docker_image = Client.load("docker://" + container_image)
                 if docker_image and len(docker_image.tags) > 0:
                     tag = docker_image.tags[0]
-                    print(f"Building singularity image from local docker image: {tag}")
+                    warnings.warn(f"Building singularity image from local docker image: {tag}")
```
samuelgarcia, line 106: Is it really a warning ?

zm711: No. But warnings can be turned off for people who don't want to see them. Whereas prints don't give that option. So for me I'm using warning for that rather than a "true" warning. I'm okay with changing back.

h-mayorquin: Yes, it is good to avoid sending things to standard output unless requested.

zm711: The way I'm thinking about it Sam is if we include a `if verbose` check then that can be print because the user is requesting extra info. But if we are not giving the user the option then by housing in warnings we give them the option to turn off warnings. So Heberto and I agree here, but if you have a strong reason to need this to be print just let me know!

samuelgarcia: ok. keep the warnings. it is not important

alejoe91 (approved):

PR #3815, opened 2025-03-28
Fix error message and pass through function for generating `NumpySorting`

### LINKED ISSUE #3814: `mountainsort5` failing due to change in `from_times_labels`
### MRE

```python
import probeinterface as pi
import numpy as np

recording, _ = si.generate_ground_truth_recording(durations=[600], sampling_frequency=30_000, num_channels=128)
# this code is for the Cambridge neurotech option to glue to probes together which naturally fits well with a probe group :)

probe = pi.get_probe("cambridgeneurotech", "ASSY-77-H5")
probe1 = probe.copy()
probe1.move([650,0])
wiring = np.arange(0,128)
multishank = pi.ProbeGroup()
multishank.add_probe(probe)
multishank.add_probe(probe1)
multishank.set_global_device_channel_indices(wiring)

recording = recording.set_probegroup(multishank)

# adding this because my normal work flow removes some channels based on bad impedance values.
# I don't know if this matters yet.
recording = recording.remove_channels(['10', '35', '67'])

ms_sorting = si.run_sorter_by_property(sorter_name="mountainsort5", recording=recording, grouping_property="group", folder='test')
```

### Error
This fails with the following error

```
extract_snippets
    snippets[j] = traces[t1:t2]
    ~~~~~~~~^^^
ValueError: could not broadcast input array from shape (0,64) into shape (40,64)
```

But this happens no matter what data I'm working with (both simulated or real data). On version 0.102.1 this error does not happen.

### Actual Problem
Not a clue yet. Unfortunately the error happens in sorting so I'm assuming something goes wrong either in channel slicing or in preprocessing. Mountainsort5 internally runs our `bandpass_filter` and our `whiten` so either of those functions since the update to preprocessing could be the problem. I need to run more tests. But I'm pretty busy today, so if someone else wants to pick this up too that would be great.

### Environment
```
spikeinterface: main branch 
neo: main branch
numpy 1.26.0
scipy 1.11.3
python 3.11.6
``` 

EDIT typed out my script too fast... edited to fix :)

zm711:
Fixes #3814. 

We can see that a package that relies on si was using our wrongly named function based on the what we told them to do (function requested times but need samples). We fixed the name by making new functions but connected the old function to the wrong new function which means the time <-> sample conversion was backwards. We should keep it connected to the wrong function for at least a version so that the warning can be read by people expecting the behavior and then fixed before we change the behavior in what could be a silent fashion.

https://github.com/flatironinstitute/mountainsort5/blob/a2f4195927b041210bed8c7113eda5f7ffd367e7/mountainsort5/schemes/sorting_scheme2.py#L328

@magland this affects mountainsort. (I'm happy to push a PR for you over there to deal with this. But I would need to know if you want to put a new floor of spikeinterface to 0.102.2 or whether you want me to account for old behavior and new behavior )

### DIFF
--- a/src/spikeinterface/core/numpyextractors.py
+++ b/src/spikeinterface/core/numpyextractors.py
@@ -367,12 +367,20 @@ class NumpySorting(BaseSorting):
 
     @staticmethod
     def from_times_labels(times_list, labels_list, sampling_frequency, unit_ids=None) -> "NumpySorting":
+        warning_msg = (
+            "`from_times_labels` is deprecated and will be removed in 0.104.0. Note this function requires"
+            "samples rather than times so should not be used for clarity purposes. For those working in samples please"
+            "use `from_samples_and_labels` instead. For those working in time units (seconds) please use "
+            "`from_times_and_labels` instead."
+        )
+
         warnings.warn(
-            "from_times_labels is deprecated and will be removed in 0.104.0, use from_sample_and_labels instead",
+            warning_msg,
             DeprecationWarning,
             stacklevel=2,
         )
-        return NumpySorting.from_times_and_labels(times_list, labels_list, sampling_frequency, unit_ids)
+        # despite the naming of the original function this required samples
+        return NumpySorting.from_samples_and_labels(times_list, labels_list, sampling_frequency, unit_ids)
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


alejoe91 (approved):

h-mayorquin (approved):
Thanks for the improved warnings message.

PR #3819, opened 2025-03-29
allow UnitWaveformsWidget to use probeinterface plot_probe

jakeswann1:
Allows the UnitWaveformsWidget to use probeinterface `plot_probe` for channel locations, and implements axes_equal arg which was not previously implemented. This would depend on ProbeInterface [PR 334](https://github.com/SpikeInterface/probeinterface/pull/334) to allow the plot_probe function not to plot directly to an set of axes.

Would this be safe enough (do we expect the sorting_analyzer_or_templates object to always have a probe attached, or should I include a fallback to the original implementation too?

### DIFF
--- a/src/spikeinterface/widgets/unit_waveforms.py
+++ b/src/spikeinterface/widgets/unit_waveforms.py
@@ -383,11 +383,11 @@ class UnitWaveformsWidget(BaseWidget):
 
             # plot channels
             if dp.plot_channels:
-                from probeinterface.plotting import plot_probe
+                from probeinterface.plotting import create_probe_polygons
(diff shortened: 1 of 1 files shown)


chrishalcrow:
We demand there's a probe attached already in `get_channel_locations` here: https://github.com/SpikeInterface/spikeinterface/blob/aa4c79200b835de2792b48095b89a3ff1ab0440f/src/spikeinterface/widgets/unit_waveforms.py#L125
so we can safely assume there's a probe around :+1:

alejoe91 (approved):

PR #3822, opened 2025-03-31
add test

### LINKED ISSUE #3816: `remove_channels` should either warn or raise an error when giving non-present channels
Carry over from #3814.

### MRE
```python
>>> import spikeinterface.full as si
>>> recording , _ si.generate_ground_truth_recording([100], num_channels=64)
>>> rec = recording.remove_channels(1)
>>> print(rec.get_num_channels())
64

```
This will sometimes silently sometimes with a numpy warning do nothing. This is because the dtype of the channels is string, but I'm trying to remove an int. 

### Solution
I would argue that we should flat out error since it is too easy to ignore a warning if someone gives channels that can't be removed (either due to dtype or not existing). But at minimum we should warn the user that they are doing it wrong since nothing is happening.

@chrishalcrow.

chrishalcrow:
Fixes #3816. In `remove_channels` this PR adds an `Assertion` that all given channels-to-remove are in the recording ids. And adds a test.

Example message:

``` python
rec = generate_recording(num_channels=4, durations=durations, set_probe=False, seed=seed)
rec_sliced = rec.remove_channels(remove_channel_ids=[0, "1"])
```
```
>>> AssertionError: `remove_channel_ids` [0] are not in recording ids ['0' '1' '2' '3'].
```

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -735,7 +735,7 @@ class BaseRecording(BaseRecordingSnippets):
         non_present_channel_ids = list(set(remove_channel_ids).difference(recording_channel_ids))
         assert (
             len(non_present_channel_ids) == 0
-        ), f"channel ids {non_present_channel_ids} are not in recording ids {recording_channel_ids}."
+        ), f"`remove_channel_ids` {non_present_channel_ids} are not in recording ids {recording_channel_ids}."
 
         new_channel_ids = self.channel_ids[~np.isin(self.channel_ids, remove_channel_ids)]
         sub_recording = ChannelSliceRecording(self, new_channel_ids)
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -731,6 +731,12 @@ class BaseRecording(BaseRecordingSnippets):
     def _remove_channels(self, remove_channel_ids):
         from .channelslice import ChannelSliceRecording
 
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/baserecording.py
```diff
@@ -731,6 +731,12 @@ def _channel_slice(self, channel_ids, renamed_channel_ids=None):
     def _remove_channels(self, remove_channel_ids):
         from .channelslice import ChannelSliceRecording
 
+        recording_channel_ids = self.get_channel_ids()
+        non_present_channel_ids = list(set(remove_channel_ids).difference(recording_channel_ids))
+        assert (
+            len(non_present_channel_ids) == 0
+        ), f"`remove_channel_ids` {non_present_channel_ids} are not in recording ids {recording_channel_ids}."
+
```
zm711, line 739: This is a huge fight in the python world in general. But what about instead of assertion we do a value error.

so if len(non_present_channel_ids)>0:
raise ValueError.

The only issue (though Heberto and I have discussed this often) is that you can run python while shutting off all asserts with a flag. So internal functions can have assert, but if it is an api the user faces then it shouldn't be an assert. We don't enforce that but I would request it if you're amenable.

chrishalcrow: Yeah, sure, was just copying `select_channels`.

zm711 (approved):
Good by me as long as @alejoe91 is good with it :)

PR #3830, opened 2025-04-02
benchmark plot : go back to previous with no level in plot_run_times and plot_unit_ counts

samuelgarcia:
benchmark plot : go back to previous with no level in plot_run_times and plot_unit_ counts


zm711:
@samuelgarcia just for context for everyone, why do you want to revert the changes? What was the problem with the newer version?

samuelgarcia:
Yes, the multi level introduce by Alessio is really great but was breaking some basic plotting fonctions for non multi level cases.
And we need then with Pierre to continue making benchmark figures.

alejoe91 (approved):

Issue #3823, opened 2025-03-31
Unclosed file after opening the recording


OlivierPeron:
Hi ! 

I recently have had an error when opening the binary file :

```bash
sys:1: ResourceWarning: unclosed file <_io.BufferedReader name='my\\dat\\file.dat'> 
```

I used `tramalloc` to identify where the file was being opened and it is here:

```python
class BinaryRecordingSegment(BaseRecordingSegment):
    def __init__(self, file_path, sampling_frequency, t_start, num_channels, dtype, time_axis, file_offset):
        BaseRecordingSegment.__init__(self, sampling_frequency=sampling_frequency, t_start=t_start)
        self.num_channels = num_channels
        self.dtype = np.dtype(dtype)
        self.file_offset = file_offset
        self.time_axis = time_axis
        self.file_path = file_path
        self.file = open(self.file_path, "rb")
        self.bytes_per_sample = self.num_channels * self.dtype.itemsize
        self.data_size_in_bytes = Path(file_path).stat().st_size - file_offset
        self.num_samples = self.data_size_in_bytes // self.bytes_per_sample
```

It is opening the file `self.file = open(self.file_path, "rb")`, but never close it, could it be an issue, I never had this error before so I don't know what led to it. Am I missing something?

zm711:
I've been trying to track these instances. It would be helpful if you could also indicate your OS/if you're working from a server. 

There are a couple places where we don't close files. Unfortunately sometimes they are memmaps and numpy.memmaps offer no guarantees that they can be closed. So if we know as much info as possible about what type of file you were trying to write that would also be useful!

OlivierPeron:
I have Windows 11 Pro for Workstation, and I first opened the recording saved as a .dat file (from intan), did the lazy recording, and then when launching the sorting, I am getting this error that the file is unclosed.

zm711:
Okay that's good to know. Windows file management is a bit harder to deal with than Mac/Linux. (I know our workstations are also Windows). Does it give any windows errors or just the sys error?

OlivierPeron:
Only the sys error

samuelgarcia:
> It is opening the file `self.file = open(self.file_path, "rb")`, but never close it, could it be an issue? 

All reader are opening a file.. The file is close when you detsroy the object (the recording) itself.
We have this by design everywhere.

This should not be a problem as long as you do no try to write the same file again while the provious recording is not yet destroy.

[...]

PR #3833, opened 2025-04-03
Add del dunder to binary recording

h-mayorquin:
Trying to ameliorate the situation on #3823

Adding del dunders to the binary recording extractor to make the desired behavior explicit.


h-mayorquin:
I would not know how to test it. If the user wants to that would be great.

zm711:
My question was does a dunder del work automatically or would we need to call a del in the code to test this? The user already tried running delete so we could see if this works better, but I was curious if we would need to add something in our code first before they test or whether testing as is with the method was enough.

h-mayorquin:
> My question was does a dunder del work automatically or would we need to call a del in the code to test this? 

Neither exactly:

https://docs.python.org/3/reference/datamodel.html#object.__del__

But calling del would be better indeed. 

I am just making the action more explicit.

samuelgarcia:
I am OK with this but this is strange that it is not done already.
More explicit do not hurt.
We must have a cyclic ref somewhere no ?

samuelgarcia (approved):

src/spikeinterface/core/binaryrecordingextractor.py
```diff
@@ -148,6 +148,18 @@ def get_binary_description(self):
...
+        """
+        Ensures that all segment resources are properly cleaned up when this recording extractor is deleted.
+        Closes any open file handles in the recording segments.
+        """
+        # Close all recording segments
+        if hasattr(self, "_recording_segments"):
+            for segment in self._recording_segments:
+                # This will trigger the __del__ method of the BinaryRecordingSegment
+                # which will close the file handle
+                del segment
```
alejoe91, line 161: Why not moving this to the `BaseRecording`? I think deleting segments is a safe thing to do for any object! We could do it for sortings as well!

h-mayorquin:
> We must have a cyclic ref somewhere no ?

Yes. I think so, this will just make me more confident when we try to debug that. As you say, be explicit to be 100 % sure.

src/spikeinterface/core/binaryrecordingextractor.py
h-mayorquin: I can do this but I would like to get this through and do the general change in another PR. In case the general change fails or there is something else we are missing at least we have this feature already published.

would you be OK with that?

alejoe91: yep. Merging as is then :)

alejoe91 (approved):

alejoe91:
Test failures are unrelated and were passing before.

PR #3834, opened 2025-04-04
Function to estimate templates from svd

yger:
We can avoid one pass on the raw data (for IO bottleneck) if we estimate templates from the inverse of the svd model, for clustering solutions that rely on svd. This can be used by KS clustering, graph clustering, and all methods relying on split too @samuelgarcia

### DIFF
--- a/src/spikeinterface/sortingcomponents/clustering/tools.py
+++ b/src/spikeinterface/sortingcomponents/clustering/tools.py
@@ -317,7 +317,7 @@ def get_templates_from_peaks_and_svd(
                 data = np.mean(local_svd[sub_mask, :, count], 0)
             elif operator == "median":
                 data = np.median(local_svd[sub_mask, :, count], 0)
-            templates_array[unit_ind, :, i] = svd_model.inverse_transform(data.reshape(1,-1))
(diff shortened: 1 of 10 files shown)


src/spikeinterface/sortingcomponents/clustering/tools.py
```diff
@@ -192,3 +192,58 @@ def apply_waveforms_shift(waveforms, peak_shifts, inplace=False):
...
+    fs = recording.get_sampling_frequency()
+    nbefore = int(ms_before * fs / 1000.0)
+    nafter = int(ms_after * fs / 1000.0)
+    num_channels = recording.get_num_channels()
+
+    if svd_model is None:
+        templates_array = estimate_templates(
+            recording, peaks, labels, nbefore, nafter, return_scaled=False, job_name=None, **job_kwargs
+        )
+    else:
```
samuelgarcia, line 224: I would not this magic if else.

I think I prefer a clear function that do only the second part.
For instance `compute_template_from_svd_and_label()`
Lets do the if else outside this fonction.

samuelgarcia (approved):

PR #3835, opened 2025-04-04
Enhance split with various options

yger:
As we already discussed @samuelgarcia , some extensions for the internal split, backward compatible


samuelgarcia:
Yep.

samuelgarcia (approved):

PR #3836, opened 2025-04-04
Minor improvements to extract_peak_svd

yger:
Here we can handle the mask (needed if we want this function to be used by KS-like clustering and compare), and clean the outliers while collecting the peaks


samuelgarcia (approved):

samuelgarcia:
Cool merci.

samuelgarcia (approved):

PR #3838, opened 2025-04-04
neoextractor biocam: add paramaters for sparse event based recordings

mahlzahn:
see https://github.com/NeuralEnsemble/python-neo/pull/1446

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/biocam.py
+++ b/src/spikeinterface/extractors/neoextractors/biocam.py
@@ -23,6 +23,14 @@ class BiocamRecordingExtractor(NeoBaseRecordingExtractor):
         The inter-electrode distance (pitch) between electrodes.
     electrode_width : float, default: None
         Width of the electrodes in um.
+    fill_gaps_strategy: "zeros" | "synthetic_noise" | None, default: None
+        The strategy to fill the gaps in the data when using event-based
+        compression. If None and the file is event-based compressed,
+        you need to specify a fill gaps strategy:
+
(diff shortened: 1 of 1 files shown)


src/spikeinterface/extractors/neoextractors/biocam.py
```diff
@@ -86,9 +86,11 @@ def __init__(
 
     @classmethod
     def map_to_neo_kwargs(cls, file_path, true_zeroes, use_synthetic_noise):
-        neo_kwargs = {"filename": str(file_path),
-                      "true_zeroes": true_zeroes,
-                      "use_synthetic_noise": use_synthetic_noise}
+        neo_kwargs = {
+            "filename": str(file_path),
+            "true_zeroes": true_zeroes,
+            "use_synthetic_noise": use_synthetic_noise,
```
alejoe91, line 92: Hi Robert, we just discussed about the PR on NEO and are about to merge it, but we decided to change the args a bit. I'll push a fix and ping you there

alejoe91: The idea is the following: using a `fill_gaps_strategy` argument instead of the `true_zeroes`/`use_synthetic_noise` which can be `zeros` or `synthetic_noise`. `zeros` will be the same as `true_zeros`, since I don't think it makes sense to just set 0s givent the 12bits...

By defautl though, this will be None and an error is raised if there ise event-based compression but the strategy is not set. How does it sound?

PR #3839, opened 2025-04-04
Backwards compatibility for sync stream in OpenEphysBinary

### LINKED ISSUE #3837: Upcoming change in OpenEphysBinary extractor
The PR in NEO https://github.com/NeuralEnsemble/python-neo/pull/1668 will change the way the binary extractor works when a single stream is available and when there are multiple streams, but one only wants to get the neural streams listed with the `get_neo_streams` function.

@h-mayorquin we should work on a patch here as soon as we can :)

h-mayorquin:
Should fix #3837

As requested by @chrishalcrow and @alejoe91

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/openephys.py
+++ b/src/spikeinterface/extractors/neoextractors/openephys.py
@@ -139,21 +139,48 @@ class OpenEphysBinaryRecordingExtractor(NeoBaseRecordingExtractor):
     all_annotations : bool, default: False
         Load exhaustively all annotation from neo
 
+    Notes
+    -----
+    If no stream is explicitly specified and there are exactly two streams (neural data and
+    synchronization data), the neural data stream will be automatically selected.
     """
 
     NeoRawIOClass = "OpenEphysBinaryRawIO"
 
     def __init__(
         self,
-        folder_path,
-        load_sync_channel=False,
-        load_sync_timestamps=False,
-        experiment_names=None,
-        stream_id=None,
-        stream_name=None,
-        block_index=None,
-        all_annotations=False,
+        folder_path: str | Path,
+        load_sync_channel: bool = False,
+        load_sync_timestamps: bool = False,
+        experiment_names: str | list | None = None,
+        stream_id: str = None,
+        stream_name: str = None,
+        block_index: int = None,
+        all_annotations: bool = False,
     ):
+
+        # Not sure if we should deprecate this, discuss with Alessio
+        # if load_sync_channel:
+        #     import warnings
+
+        #     warning_message = (
+        #         "OpenEphysBinaryRecordingExtractor: load_sync_channel is deprecated and will"
+        #         "be removed in version 0.104, use the stream_name or stream_id to load the sync stream if needed"
+        #     )
+        #     warnings.warn(warning_message, DeprecationWarning, stacklevel=2)
+
+        stream_is_not_specified = stream_name is None and stream_id is None
+        if stream_is_not_specified:
+            available_stream_names, _ = self.get_streams(folder_path, load_sync_channel, experiment_names)
+
+            # Auto-select neural data stream when there are exactly two streams (neural + sync)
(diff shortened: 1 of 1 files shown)


chrishalcrow:
Hello,

I'm having fun learning about Neo this morning! Main question: if users want to load the sync stream by itself, I think the new recommended way to do this is

``` python
sync_stream = si.read_openephys("path/to/rec", stream_name = "CH_SYNC-0")
# well, that's the name in my case
```

Is that right? (And then we're keeping the "load_sync_channel=True` to add a sync channel, just for backwards compatibility; and we hope to retire this one day.)

At the moment, the above code breaks for me because it can't figure out the `record_node` here:
https://github.com/SpikeInterface/spikeinterface/blob/d30fc3269789e71cbe11f66303359bc5f3c3fdc9/src/spikeinterface/extractors/neoextractors/openephys.py#L176

``` python
  File "<stdin>", line 1, in <module>
  File "/home/nolanlab/Work/Developing/fromgit/spikeinterface/src/spikeinterface/extractors/neoextractors/openephys.py", line 384, in read_openephys
    recording = OpenEphysBinaryRecordingExtractor(folder_path, **kwargs)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nolanlab/Work/Developing/fromgit/spikeinterface/src/spikeinterface/extractors/neoextractors/openephys.py", line 213, in __init__
    exp_ids = sorted(list(self.neo_reader.folder_structure[record_node]["experiments"].keys()))
                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
KeyError: ''
```

If I set `record_node = "Record node 109"` (which is the correct name in my case) everything works perfectly. Don't know how to get this info though!

chrishalcrow:
> > At the moment, the above code breaks for me because it can't figure out the record_node here:
> 
> That does not happen on the main branch?

Nope. Note it's only when I try and get the sync stream using `stream_name = "CH_SYNC-0"` that this error occurs. I can't even try that on the main branch, since there's no stream for the sync channel.

chrishalcrow:
Some more debugging: in the past we get the `record_node` name by splitting the name of the main neural stream (which is `Record Node 109#Neuropix-PXI-103.ProbeA` in my case). Without this name, it doesn't know the keys of `self.neo_reader.folder_structure`, which in my case are just `["Record Node 109"]`

alejoe91:
@h-mayorquin I thinkw e should update the stream_name (not the id) in NEO to have record-node#oe_stream-SYNC. It is important to know where the sync came from. What do you think?

h-mayorquin:
> @h-mayorquin I thinkw e should update the stream_name (not the id) in NEO to have record-node#oe_stream-SYNC. It is important to know where the sync came from. What do you think?

Yes, just did this.

chrishalcrow:
Hello - this is now working for all my "real life" tests! Thanks @h-mayorquin!!

Just wanted to point out when this new code breaks. It will break if someone has the NEW PR in neo but is using OLD spikeinterface. I think this would be a fairly rare case? In that rare case, if you do `openephys(path_to_rec)` you get

``` python
ValueError: This reader have several streams: 
`stream_names`: ['Record Node 200#Neuropix-PXI-103.ProbeA', 'Record Node 200#Neuropix-PXI-103.ProbeASYNC']
`stream_ids`: ['0', 'Record Node 200#Neuropix-PXI-103.ProbeASYNC']. 
Specify it from the options above with the 'stream_name' or 'stream_id' arguments
```

which seems like a pretty descriptive error, which few people will see.

alejoe91 (approved):

alejoe91:
@chrishalcrow @h-mayorquin 

LGTM!

Ok to merge now that the NEO PR is merged?

chrishalcrow:
Yup, good for me :)

PR #3843, opened 2025-04-07
BlackrockSortingExtractor - load only .nev files, ignore nsX

luiztauffer:
There are cases when we'd like to load only the `.nev` file containing the events data, and ignore the local `.nsX` files. One example for this is when an experimenter pauses and then resumes the recording. This creates a different number of segments between nev and nsX, which breaks neo reader.
Thankfully, neo already allows for it, all we need to do is to explicitly pass the `nsx_to_load` argument. So, in this PR I'm proposing adding it as an optional argument to `BlackrockSortingExtractor`.

Ps.: It's probably be a good idea tho eventually allow passing forward all kwargs available at the neo extractors directly from the SI wrappers `__init__` methods, a `neo_kwargs: dict` or something


luiztauffer:
@alejoe91 @h-mayorquin could you take a look at this whenever you get a chance? thanks!

alejoe91:
Thanks @luiztauffer 

I wonder if the sorting extractor should automatically just load the nev file only. The other files shouldn't be required for loading spike trains only. What do you think?

luiztauffer:
@alejoe91 check it out now, let me know what you think

h-mayorquin:
> important: this change depends on the current master branch of neo, so maybe we should push for a release over there before merging this?

which feature that we merge recently? There will be release soon.

> I wonder if the sorting extractor should automatically just load the nev file only. The other files shouldn't be required for loading spike trains only. What do you think?

Agree with this.


@alejoe91 the tests are failing because of pinnin numcodes. Oh, I see you merged already.

h-mayorquin:
> @h-mayorquin I believe it's this one: [NeuralEnsemble/python-neo@b87b8f2](https://github.com/NeuralEnsemble/python-neo/commit/b87b8f277d6a53925ee30c1a09f900993ca39cf9)

Ah, the overflow of Thinh. Yes, there will be a release soon according to the neo team.

src/spikeinterface/extractors/neoextractors/neobaseextractor.py
```diff
@@ -95,18 +95,22 @@ def build_stream_id_to_sampling_frequency_dict(self) -> Dict[str, float]:
...
-        # Get unique pairs of channel_stream_id and channel_sampling_frequencies
-        channel_sampling_frequencies = signal_channels["sampling_rate"]
-        channel_stream_id = signal_channels["stream_id"]
-        unique_pairs = np.unique(np.vstack((channel_stream_id, channel_sampling_frequencies)).T, axis=0)
-
-        # Form a dictionary of stream_id to sampling_frequency
-        stream_to_sampling_frequencies = {}
-        for stream_id, sampling_frequency in unique_pairs:
-            stream_to_sampling_frequencies[stream_id] = float(sampling_frequency)
+        if "signal_channels" in neo_header and neo_header["signal_channels"].size > 0:
```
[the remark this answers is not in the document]
h-mayorquin: > I'm not sure, actually. But this won't be a problem if we overwrite the method

OK, so that is the critical thing here. How could we confirm this? Does blackrock has a spec sheet of the format? Can you ask the people from the conversion project? If we are not sure we should add an optional keyword as an escape valve so we don't leave incorrect behavior as the default.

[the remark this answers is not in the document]
luiztauffer: `wf_sampling_rate` seems to be the common `neo` variable name for writing the header. So I think we're safe here

h-mayorquin:
Ok, this is hardcoded here on the python neo side anyway.

https://github.com/NeuralEnsemble/python-neo/blob/be8e663e5912dcab70d5c87a9ade80f576681702/neo/rawio/blackrockrawio.py#L301

https://github.com/NeuralEnsemble/python-neo/blob/be8e663e5912dcab70d5c87a9ade80f576681702/neo/rawio/blackrockrawio.py#L253

So no need to override the function at all. If we are going to use the wf we might as well just hardcoded in the init of blackrock and avoid all the coding complications.

luiztauffer:
@h-mayorquin right, nice catch! Yeah, from the SI wrapper side, I believe it would be better to wait for [your PR](https://github.com/NeuralEnsemble/python-neo/pull/1685) to get merged and then we can retrieve the sampling rate from the neo reader, instead of hard coding it here again. This way we keep true to whatever neo says, including if there's any future changes to that parameter

h-mayorquin:
@alejoe91

If we consider `wf_sampling_rate` from the Blackrock headers a reliable heuristic for determining the sorter's sampling frequency (@samuelgarcia maybe knows better), I suggest the following path forward:

1. Expose this value as a class attribute on the Neo side, so users don’t need to parse the headers to access the main sampling frequency. See: https://github.com/NeuralEnsemble/python-neo/pull/1685

2. Add a keyword argument to `BlackrockSortingExtractor` like `use_main_blackrock_frequency: bool = True`,  and pass the main sampling frequency from the header here:  
   https://github.com/SpikeInterface/spikeinterface/blob/4072c3c8a17074498ef90f64192e648de8a86820/src/spikeinterface/extractors/neoextractors/blackrock.py#L100-L105

This approach has a few advantages:
* No need to override any functionality.
* Clearly communicates the heuristic being used.
* Couples the logic to Neo, so improvements in Neo automatically propagate here.
* Provides users with an escape hatch in case the heuristic doesn’t apply in their case.

luiztauffer:
It looks like Blackrock won’t need `_infer_sampling_frequency_from_analog_signal` at all, or am I missing something?

If we want to keep having `sampling_frequency` as an optional argument for SI extractor to override whatever neo has in place, ok. 

But I don’t think it’s reasonable for a Neuroconv DataInterface to have a **required** argument for something which is a fixed attribute down the line, in the `neo` reader (and it seems to be correct for that system). And I’m not even sure if leaving this as an optional override for NWB conversions would be recommended.

h-mayorquin:
> But I don’t think it’s reasonable for a Neuroconv DataInterface to have a required argument for something which is a fixed attribute down the line, in the neo reader (**and it seems to be correct for that system**). And I’m not even sure if leaving this as an optional override for NWB conversions would be recommended. 

[bold mine]

Yes, I think that’s the core of the issue. The Neo sampling frequency is undocumented, and the last time I discussed the extractor with Sam, he mentioned he hasn’t touched it in a while. I also asked him about using the frequency from the spike header, and he said it might not be accurate. In other words, I don’t fully trust that value, and I’d prefer to leave an escape valve in case we're wrong—so we don’t have to tell users to wait for another PR to correct the mistake.

> It looks like Blackrock won’t need _infer_sampling_frequency_from_analog_signal at all, or am I missing something?

If there’s only one signal channel, the heuristic gets the sampling frequency from that channel, which should be the signal that was sorted. Otherwise, if there’s more than one signal, the user gets an error telling them to set the sampling frequency manually. 

Regardless of all that, it's important to point out that this is a problem for us (people who work with other people's data) because we often don’t know the correct sampling frequency. Hopefully, most users will be the ones who generated the data (fingers crossed), in which case it's better if they always set the frequency themselves. Maybe in this case we should have an especial error for when the segments in the sorting are not the same as the signal which was the initial problem, right?

I’ve come around to agree with Sam that users should be responsible for setting the sampling frequency as neo is not the right tool to fetch this from and he, reasonably, does not want to maintain that. All that said, I’m okay with adding the heuristic you’re suggesting here for the sake of expediency and give the user more options. Let’s see what others think.

src/spikeinterface/extractors/neoextractors/blackrock.py
```diff
@@ -117,6 +118,33 @@ def map_to_neo_kwargs(cls, file_path):
...
+
+        Raises
+        ------
+        AssertionError
+            If there are no signal streams available from which to extract the sampling frequencies.
+        """
+        neo_header = self.neo_reader.header
+        if "spike_channels" in neo_header and neo_header["spike_channels"].size > 0:
+            channels = neo_header["spike_channels"]
+            channel_sampling_frequencies = channels["wf_sampling_rate"]
```
samuelgarcia, line 141: In neo this is not necessary the sampling frequency of the spikes (even it is very often the same).
This is the frequency of the waveforms snippet. And this field has no garnty to be here if the snipet are not recorded

luiztauffer: yeah, I guess for now we're going with the fixed sampling rate

luiztauffer:
just adding now the option to to pass two relevant kwargs to the neo reader: `nsx_to_load` and `sampling_frequency`

alejoe91 (approved):

h-mayorquin (approved):
+1 on this. 
Will allow users to specify the sampling frequency and to load data when there are more than one segment on the recording but not in the sorting.

src/spikeinterface/extractors/neoextractors/blackrock.py
```diff
@@ -2,8 +2,8 @@
 
 from pathlib import Path
 from packaging import version
-from typing import Optional
-
+from typing import Optional, Dict
+import numpy as np
```
h-mayorquin, line 6:
```suggestion
```

src/spikeinterface/extractors/neoextractors/blackrock.py
```diff
@@ -2,8 +2,8 @@
 
 from pathlib import Path
 from packaging import version
-from typing import Optional
-
+from typing import Optional, Dict
```
h-mayorquin, line 5:
```suggestion
from typing import Optional
```

src/spikeinterface/extractors/neoextractors/blackrock.py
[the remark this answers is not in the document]
h-mayorquin: they were probably left from a past iteration.

src/spikeinterface/extractors/neoextractors/blackrock.py
```diff
@@ -92,16 +94,18 @@ class BlackrockSortingExtractor(NeoBaseSortingExtractor):
     def __init__(
         self,
         file_path,
-        sampling_frequency: Optional[float] = None,
         stream_id: Optional[str] = None,
         stream_name: Optional[str] = None,
+        sampling_frequency: Optional[float] = None,
+        nsx_to_load: Optional[int | list | str] = None,
```
h-mayorquin, line 100: the typing of this seems strange, can you double check @luiztauffer?

Sam is going to vacations so if he does not have time to check it before that I will merge this next Monday but just double check this.

luiztauffer: @h-mayorquin this comes from here: https://github.com/NeuralEnsemble/python-neo/blob/c425a529958d582919607a0eb940154a4b5c4d7e/neo/rawio/blackrockrawio.py#L97

PR #3773, opened 2025-03-14
Allow preprocessing on dicts of recordings (again)

chrishalcrow:
Copy of #3711, before I somehow made a mess of it.

See #3711 for information and discussion.

Edit: read the new docs here: https://spikeinterface--3773.org.readthedocs.build/en/3773/how_to/process_by_channel_group.html#preprocessing-a-recording-by-channel-group

PR #3844, opened 2025-04-07
Add documentation to handle (physical) units of recording channels

h-mayorquin:
Ok guys, I will need your feedback. I am also adding another preprocessor wrapper that handles this for the user.

### DIFF
--- a/doc/how_to/combine_recordings.rst
+++ b/doc/how_to/combine_recordings.rst
@@ -4,7 +4,7 @@ Combine recordings in SpikeInterface
 In this tutorial we will walk through combining multiple recording objects. Sometimes this occurs due to hardware
 settings (e.g. Intan software has a default setting of new files every 1 minute) or the experimenter decides to
 split their recording into multiple files for different experimental conditions. If the probe has not been moved,
-however, then during sorting it would likely make sense to combine these individual reocrding objects into one
+however, then during sorting it would likely make sense to combine these individual recording objects into one
 recording object.
 
 **Why Combine?**
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "spikeinterface"
-version = "0.102.3"
+version = "0.102.4"
 authors = [
   { name="Alessio Buccino", email="alessiop.buccino@gmail.com" },
   { name="Samuel Garcia", email="sam.garcia.die@gmail.com" },
@@ -126,8 +126,8 @@ test_core = [
 
     # for github test : probeinterface and neo from master
     # for release we need pypi, so this need to be commented
-    # "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",
-    # "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",
+    "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",
+    "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",
 ]
 
 test_extractors = [
@@ -135,8 +135,8 @@ test_extractors = [
     "pooch>=1.8.2",
     "datalad>=1.0.2",
     # Commenting out for release
-    # "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",
-    # "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",
+    "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",
+    "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",
 ]
 
 test_preprocessing = [
@@ -174,8 +174,8 @@ test = [
 
     # for github test : probeinterface and neo from master
     # for release we need pypi, so this need to be commented
-    # "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",
-    # "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",
+    "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",
+    "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",
 ]
 
 docs = [
@@ -200,8 +200,8 @@ docs = [
     "datalad>=1.0.2",
 
     # for release we need pypi, so this needs to be commented
-    # "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",  # We always build from the latest version
-    # "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",  # We always build from the latest version
+    "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",  # We always build from the latest version
+    "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",  # We always build from the latest version
 ]
 
 dev = [
--- a/src/spikeinterface/__init__.py
+++ b/src/spikeinterface/__init__.py
@@ -27,5 +27,5 @@ import spikeinterface.full as si
 # This flag must be set to False for release
 # This avoids using versioning that contains ".dev0" (and this is a better choice)
 # This is mainly useful when using run_sorter in a container and spikeinterface install
-# DEV_MODE = True
-DEV_MODE = False
+DEV_MODE = True
+# DEV_MODE = False
--- a/src/spikeinterface/benchmark/benchmark_base.py
+++ b/src/spikeinterface/benchmark/benchmark_base.py
@@ -594,15 +594,22 @@ class Benchmark:
             elif format == "sorting":
                 from spikeinterface.core import load_extractor
 
-                result[k] = load(folder / k)
+                sorting_folder = folder / k
+                if sorting_folder.exists():
+                    result[k] = load(sorting_folder)
             elif format == "Motion":
                 from spikeinterface.core.motion import Motion
 
-                result[k] = Motion.load(folder / k)
+                motion_folder = folder / k
+                if motion_folder.exists():
+                    result[k] = Motion.load(motion_folder)
             elif format == "zarr_templates":
                 from spikeinterface.core.template import Templates
 
-                result[k] = Templates.from_zarr(folder / k)
+                zarr_folder = folder / k
+                if zarr_folder.exists():
+
+                    result[k] = Templates.from_zarr(zarr_folder)
 
         return result
 
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -757,9 +757,9 @@ class BaseRecording(BaseRecordingSnippets):
         Parameters
         ----------
         start_frame : int, optional
-            The start frame, if not provided it is set to 0
+            Start frame index. If None, defaults to the beginning of the recording (frame 0).
         end_frame : int, optional
-            The end frame, it not provided it is set to the total number of samples
+            End frame index. If None, defaults to the last frame of the recording.
 
         Returns
         -------
@@ -779,9 +779,9 @@ class BaseRecording(BaseRecordingSnippets):
         Parameters
         ----------
         start_time : float, optional
-            The start time in seconds. If not provided it is set to 0.
+            Start time in seconds. If None, defaults to the beginning of the recording.
         end_time : float, optional
-            The end time in seconds. If not provided it is set to the total duration.
+            End time in seconds. If None, defaults to the end of the recording.
 
         Returns
         -------
--- a/src/spikeinterface/extractors/nwbextractors.py
+++ b/src/spikeinterface/extractors/nwbextractors.py
@@ -759,10 +759,13 @@ class NwbRecordingExtractor(BaseRecording, _BaseNWBExtractor):
         if "starting_time" in electrical_series.keys():
             t_start = electrical_series["starting_time"][()]
             sampling_frequency = electrical_series["starting_time"].attrs["rate"]
+            timestamps = None
         elif "timestamps" in electrical_series.keys():
             timestamps = electrical_series["timestamps"][:]
             t_start = timestamps[0]
             sampling_frequency = 1.0 / np.median(np.diff(timestamps[:samples_for_rate_estimation]))
+        else:
+            raise ValueError("TimeSeries must have either starting_time or timestamps")
 
         if load_time_vector and timestamps is not None:
             times_kwargs = dict(time_vector=electrical_series["timestamps"])
@@ -1572,6 +1575,8 @@ class NwbTimeSeriesExtractor(BaseRecording, _BaseNWBExtractor):
             timestamps = timeseries.timestamps
             sampling_frequency = 1.0 / np.median(np.diff(timestamps[:samples_for_rate_estimation]))
             t_start = timestamps[0]
+        else:
+            raise ValueError("TimeSeries must have either starting_time or timestamps")
 
         if load_time_vector and timestamps is not None:
             times_kwargs = dict(time_vector=timestamps)
--- a/src/spikeinterface/postprocessing/spike_amplitudes.py
+++ b/src/spikeinterface/postprocessing/spike_amplitudes.py
@@ -17,38 +17,19 @@ class ComputeSpikeAmplitudes(AnalyzerExtension):
     Computes the spike amplitudes.
 
     Needs "templates" to be computed first.
-    Localize spikes in 2D or 3D with several methods given the template.
+    Computes spike amplitudes from the template's peak channel for every spike.
 
     Parameters
     ----------
     sorting_analyzer : SortingAnalyzer
         A SortingAnalyzer object
-    ms_before : float, default: 0.5
-        The left window, before a peak, in milliseconds
-    ms_after : float, default: 0.5
-        The right window, after a peak, in milliseconds
-    spike_retriver_kwargs : dict
-        A dictionary to control the behavior for getting the maximum channel for each spike
-        This dictionary contains:
-          * channel_from_template: bool, default: True
-              For each spike is the maximum channel computed from template or re estimated at every spikes
-              channel_from_template = True is old behavior but less acurate
-              channel_from_template = False is slower but more accurate
-          * radius_um: float, default: 50
-              In case channel_from_template=False, this is the radius to get the true peak
-          * peak_sign, default: "neg"
-              In case channel_from_template=False, this is the peak sign.
-    method : "center_of_mass" | "monopolar_triangulation" | "grid_convolution", default: "center_of_mass"
-        The localization method to use
-    **method_kwargs : dict, default: {}
-        Kwargs which are passed to the method function. These can be found in the docstrings of `compute_center_of_mass`, `compute_grid_convolution` and `compute_monopolar_triangulation`.
-    outputs : "numpy" | "by_unit", default: "numpy"
-        The output format, either concatenated as numpy array or separated on a per unit basis
+    peak_sign : "neg" | "pos" | "both", default: "neg"
+        Sign of the template to compute extremum channel used to retrieve spike amplitudes.
 
     Returns
     -------
-    spike_locations: np.array
-        All locations for all spikes and all units are concatenated
+    spike_amplitudes: np.array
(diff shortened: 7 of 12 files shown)


doc/how_to/physical_units.rst
```diff
@@ -0,0 +1,96 @@
+Working with Units in SpikeInterface Recordings
```
chrishalcrow, line 1: "Units" could refer to sorted units. Maybe
```suggestion
Working with physical units in SpikeInterface recordings
```

doc/how_to/physical_units.rst
```diff
@@ -0,0 +1,96 @@
...
+.. code-block:: text
+
+    physical_value = raw_value * gain + offset
+
+
+Converting to Physical Units
+-------------------------
+
+SpikeInterface provides two preprocessing classes for converting recordings to physical units. Both wrap the
+``RecordingExtractor`` class and ensures that the data is returned in physical units when calling get_traces()
```
chrishalcrow, line 29: I was gonna suggest we link to some docs on `get_traces`, but we don't really have any... maybe we could link to the API
```suggestion
``RecordingExtractor`` class and ensures that the data is returned in physical units when calling `get_traces <https://spikeinterface.readthedocs.io/en/stable/api.html#spikeinterface.core.BaseRecording.get_traces>`_
```

doc/how_to/physical_units.rst
```diff
@@ -0,0 +1,96 @@
...
+
+1. ``scale_to_uV``: The primary function for extracellular recordings. SpikeInterface is centered around
+   extracellular recordings, and this function is designed to convert the data to microvolts (µV). Many plottinf
```
chrishalcrow, line 32: spelling
```suggestion
   extracellular recordings, and this function is designed to convert the data to microvolts (µV). Many plotting
```

doc/how_to/physical_units.rst
```diff
@@ -0,0 +1,96 @@
...
+
+While most extractors automatically set the appropriate ``gain_to_uV`` and ``offset_to_uV`` values,
+there might be cases where you want to set custom physical units. In these cases, you can set
+the following properties:
+
+* ``physical_unit``: The target physical unit (e.g., 'uV', 'mV', 'N')
+* ``gain_to_unit``: The gain to convert from raw values to the target unit
+* ``offset_to_unit``: The offset to convert from raw values to the target unit
+
+Here's an example of how to set custom units:
```
chrishalcrow, line 77: I think the code below makes more sense when you know that you have to set the units for every channel. Could we add something like:

```
You need to set these properties for every channel, which allows for the case when there are different gains and offsets on different channels. Here's an example:
```

chrishalcrow:
Looks great - doc was super clear and easy to read!

src/spikeinterface/preprocessing/scale.py
```diff
@@ -6,6 +6,55 @@
 from spikeinterface.preprocessing.basepreprocessor import BasePreprocessor
 
 
+def scale_to_physical_units(recording: BasePreprocessor) -> BasePreprocessor:
```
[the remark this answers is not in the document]
h-mayorquin: Sure, I was not aware this could not be chained. What fails?

doc/how_to/physical_units.rst
h-mayorquin: Thanks for the suggestions!

doc/how_to/physical_units.rst
h-mayorquin: Yes, this makes sense.

src/spikeinterface/preprocessing/scale.py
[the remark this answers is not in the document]
chrishalcrow: Oh, sorry, it does chain nicely when you're using it. It just fails on dicts (#3773), and it is written slightly differently than the other preprocessing steps. But maybe it doesn't matter - might be overkill to introduce two new classes for this. I'm not sure.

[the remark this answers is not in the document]
h-mayorquin: I am OK with making it a proper class so up to you : )

doc/how_to/physical_units.rst
```diff
@@ -0,0 +1,96 @@
+Working with physical units in SpikeInterface recordings
+===============================================
```
zm711, line 2: rendering is impossible for the suggestion but this needs to go to the end of the heading.

zm711 (commented):
Before being really careful my one concern with this doc as written is that for the average user we don't need to `scale_to_uV`. It happens with argument flags in various functions. The way this documentation works it sounds like I always have to create a scaled recording rather than keep my unscaled and scale at the function level.  I think the doc about other physical units is great, but I'm worried that this might actually confuse users on how to use the library since the scaling for uV is baked in to other functions the class is for people that just need their recording scaled for other purposes right?

h-mayorquin:
>  but I'm worried that this might actually confuse users on how to use the library since the scaling for uV is baked in to other functions

what would the confusion exactly?

zm711:
The confusion is do I need to do a `scale_to_uV` step in all of my pipelines? ie

```python
recording = read_intan()
scaled_rec = scale_to_uV(recording)
sorting = run_sorter(xx)
# or spikeglx
recording = read_spikeglx()
scaled_rec= scale_to_uV(recording)
sorting=run_sorter()
```

The scaling is not required in a pipeline. But when I read these docs I feel like I have to explicitly run this step no matter what. I think it is helpful for people to know scaling is important but the main way the library has people interact with ephys scaling is with the `return_scaled` soon to be `return_in_uV` (or whatever we decided) as a function argument. Not as a preprocessing step. Does that make sense? I think the docs are good, but I'm worried people will not realize when they should scale outside of function calls vs inside. For another example. If I take a `scale_to_uV recording` what happens if I try to return_scaled? I shouldn't rescale again (I think you may have added an error, but then that mechanism seems weird. Because data should be scaled. Maybe this is worth an actual discussion because I'm worried I'm not being clear.

doc/how_to/physical_units.rst
h-mayorquin: Generated by copilot before.
Fixed.

doc/how_to/physical_units.rst
```diff
@@ -0,0 +1,96 @@
...
+
+The conversion formula is:
+
+.. code-block:: text
+
+    physical_value = raw_value * gain + offset
+
+
+Converting to Physical Units
+-------------------------
```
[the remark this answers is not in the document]
h-mayorquin: Thanks, fixed.

doc/how_to/physical_units.rst
h-mayorquin: Changed.

src/spikeinterface/preprocessing/scale.py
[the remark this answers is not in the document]
h-mayorquin: Ok, I changed this to a class.

doc/how_to/physical_units.rst
```diff
@@ -0,0 +1,103 @@
...
+SpikeInterface provides tools to handle both situations.
+
+It's important to note that **most spike sorters work fine on raw digital (ADC) units** and not scaling is needed.
+Many preprocessing tools are also linear transformations, and if the ADC is implemented as a linear transformation, the overall effect can be preserved.
+That is, **preprocessing steps can often be applied either before or after unit conversion without affecting the outcome.**
+
+Therefore, **it is usually safe to work in raw ADC units unless a specific tool or analysis requires physical units**.
+If you are interested in visualizations, comparability across devices, or outputs with interpretable physical scales (e.g., microvolts), converting to physical units is recommended.
+Otherwise, remaining in raw units can simplify processing and preserve performance.
+
```
zm711, line 15: We might want Alessio or Sam to comment on some of the internal tooling. I think the scaling is automatic for some of our stuff so we should make it clear if/when we do that. I just don't remember off the top of my head. If Alessio doesn't have the time to look this over I can doublecheck in the code.

doc/how_to/physical_units.rst
```diff
@@ -0,0 +1,103 @@
...
+You need to set these properties for every channel, which allows for the case when there are different gains and offsets on different channels. Here's an example:
+
+.. code-block:: python
+
+    # Set custom physical units
+    num_channels = recording.get_num_channels()
+    values = ["volts"] * num_channels
+    recording.set_property(key='physical_unit', value=values)
+
+    values = [0.001] * num_channels  # Convert from ADC to volts
```
zm711, line 93: Maybe we say
```
gain_values = [0.001] * num_channels
```
to be even more clear and then below we would say
`offset_values`

just so we aren't using the same variable being overwritten for both? This is definitely optional.

src/spikeinterface/preprocessing/scale.py
```diff
@@ -2,8 +2,60 @@
...
+    ------
+    ValueError
+        If the recording extractor does not have gain_to_physical_unit and offset_to_physical_unit properties.
+    """
+
+    name = "recording_in_physical_units"
+
+    def __init__(self, recording):
+        if "gain_to_physical_unit" not in recording.get_property_keys():
+            raise ValueError("Recording must have 'gain_to_physical_unit' property to convert to physical units")
```
zm711, line 39: Any interest in adding the way to do this in the error? for example. adding a line like 

please use the `set_property` function in order to set "gain_to_physical_unit"

zm711 (commented):
Few more comments.

doc/how_to/physical_units.rst
```diff
@@ -0,0 +1,103 @@
...
+* **gain**: A multiplicative factor to scale the raw values
+* **offset**: An additive factor to shift the values
+
+The conversion formula is:
+
+.. code-block:: text
+
+    physical_value = raw_value * gain + offset
+
+
```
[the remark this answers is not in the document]
h-mayorquin: Can you add a suggestion?

src/spikeinterface/preprocessing/scale.py
h-mayorquin: Good idea.

h-mayorquin: Done.

doc/how_to/physical_units.rst
h-mayorquin: Done.

doc/how_to/physical_units.rst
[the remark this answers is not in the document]
zm711: Yep will do. Have some experiments all day today, but my hope is tomorrow should be a little freer.

doc/how_to/physical_units.rst
h-mayorquin: I think this is only missing part to move this forward?

zm711: Yeah I agree @alejoe91 could you comment here when you have a moment :)

src/spikeinterface/preprocessing/scale.py
```diff
@@ -1,6 +1,69 @@
...
+        # Reset gain/offset since data is now in physical units
+        self.set_property(key="gain_to_physical_unit", values=np.ones(recording.get_num_channels(), dtype="float32"))
+        self.set_property(key="offset_to_physical_unit", values=np.zeros(recording.get_num_channels(), dtype="float32"))
+
+        # Also reset channel gains and offsets
+        self.set_channel_gains(gains=1.0)
+        self.set_channel_offsets(offsets=0.0)
+
+
+scale_to_physical_units = ScaleToPhysicalUnits
```
chrishalcrow, line 66: Hello, could you change this to 
``` python
from spikeinterface.core.core_tools import define_function_handling_dict_from_class
scale_to_physical_units = define_function_handling_dict_from_class(ScaleToPhysicalUnits, name="scale_to_physical_units")
```
then will works with dicts of recs

src/spikeinterface/preprocessing/scale.py
```diff
@@ -1,6 +1,69 @@
...
+        If the recording extractor does not have gain_to_physical_unit and offset_to_physical_unit properties.
+    """
+
+    name = "recording_in_physical_units"
+
+    def __init__(self, recording):
+        if "gain_to_physical_unit" not in recording.get_property_keys():
+            error_msg = (
+                "Recording must have 'gain_to_physical_unit' property to convert to physical units. \n"
+                "Set the gain using `recording.set_property(key='gain_to_physical_unit', value=values)`."
```
chrishalcrow, line 41:
```suggestion
                "Set the gain using `recording.set_property(key='gain_to_physical_unit', values=values)`."
```

src/spikeinterface/preprocessing/scale.py
```diff
@@ -1,6 +1,69 @@
...
+            )
+            raise ValueError(error_msg)
+        if "offset_to_physical_unit" not in recording.get_property_keys():
+            error_msg = (
+                "Recording must have 'offset_to_physical_unit' property to convert to physical units. \n"
+                "Set the offset using `recording.set_property(key='offset_to_physical_unit', value=values)`."
```
chrishalcrow, line 47:
```suggestion
                "Set the offset using `recording.set_property(key='offset_to_physical_unit', values=values)`."
```

src/spikeinterface/preprocessing/scale.py
h-mayorquin: Done, thanks!

src/spikeinterface/preprocessing/scale.py
h-mayorquin: Done

This function `define_function_handling_dict_from_class` needs way more documentation of what it has now.

src/spikeinterface/preprocessing/scale.py
```diff
@@ -1,6 +1,70 @@
...
+        # Reset gain/offset since data is now in physical units
+        self.set_property(key="gain_to_physical_unit", values=np.ones(recording.get_num_channels(), dtype="float32"))
+        self.set_property(key="offset_to_physical_unit", values=np.zeros(recording.get_num_channels(), dtype="float32"))
+
+        # Also reset channel gains and offsets
+        self.set_channel_gains(gains=1.0)
+        self.set_channel_offsets(offsets=0.0)
+
+
+scale_to_physical_units = define_function_handling_dict_from_class(ScaleToPhysicalUnits)
```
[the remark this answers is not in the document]
chrishalcrow: Yes - probably missed due to the minimal documentation @h-mayorquin mentioned. I'm on it!

doc/how_to/physical_units.rst
```diff
@@ -0,0 +1,103 @@
+Working with physical units in SpikeInterface recordings
+========================================================
+
+In neurophysiology recordings, data is often stored in raw ADC (Analog-to-Digital Converter) units but needs to be analyzed in physical units.
```
zm711, line 4:
```suggestion
In neurophysiology recordings, data is often stored in raw ADC (Analog-to-Digital Converter) units but at various points in an analysis pipeline the end-user may want to observe their data in physical units.
```

I think this line saying "needs to be analyzed" sets us up for my concern for beginners. Because we are basically saying you have to do this. Then the next lines says it's optional. And that is confusing. So maybe we write saying that the user can switch to physical units at anytime they want with this tooling. Then we have the rest of your notes. 

What do you think?

doc/how_to/physical_units.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Converting to Physical Units
+----------------------------
+
+SpikeInterface provides two preprocessing classes for converting recordings to physical units. Both wrap the
+``RecordingExtractor`` class and ensures that the data is returned in physical units when calling `get_traces <https://spikeinterface.readthedocs.io/en/stable/api.html#spikeinterface.core.BaseRecording.get_traces>`_
+
+1. ``scale_to_uV``: The primary function for extracellular recordings. SpikeInterface is centered around
+   extracellular recordings, and this function is designed to convert the data to microvolts (µV). Many plotting
+   and analyzing functions in SpikeInterface expect data in microvolts, so this is the recommended approach for most users.
```
zm711, line 40: The issue is most plotting functions don't need this because we offer a `return_scaled` so maybe we change this statement. 

If you want to implement your own plotting functions you'll need to make sure to display units in microvolts so you can use this function. You can also use this for SpikeInterace plotting functions although plotting functions in spikeinterface also offer automatically scaling using the `return_scaled` argument

doc/how_to/physical_units.rst
```diff
@@ -0,0 +1,103 @@
...
+    recording_physical = scale_to_physical_units(recording)
+
+Both preprocessors automatically:
+
+1. Detect the appropriate gain and offset from the recording properties
+2. Apply the conversion to all channels
+3. Update the recording properties to reflect that data is now in physical units
+
+Setting Custom Physical Units
+---------------------------
```
zm711, line 74: ----- need to be added :)

zm711 (commented):
I have reread this I request two minor changes and then I feel good about this.

doc/how_to/physical_units.rst
h-mayorquin: Your proposal is better than what we have. I would still use analyze instead of observe in your assertion but let's move forward with this as it is.

doc/how_to/physical_units.rst
h-mayorquin: Yes, this statement is misleading, I agree. Let me take a look.

doc/how_to/physical_units.rst
h-mayorquin: Thanks!

doc/how_to/physical_units.rst
h-mayorquin: I just removed it.

doc/how_to/physical_units.rst
h-mayorquin: Changed them.

h-mayorquin:
OK, I implemented changed based on the last round of review @zm711 .

doc/how_to/physical_units.rst
```diff
@@ -0,0 +1,107 @@
+Working with physical units in SpikeInterface recordings
+========================================================
+
+In neurophysiology recordings, data is often stored in raw ADC (Analog-to-Digital Converter) integer values but needs to be analyzed in physical units.
+For extracellular recordings, this is typically microvolts (µV), but some recording devices may use different physical units.
+SpikeInterface provides tools to handle both situations.
+
+It's important to note that **most spike sorters work fine on raw digital (ADC) units** and scaling is not needed. Going a step further, some sorters, such as Kilosort 3, require their input to be in raw ADC units.
+The specific behavior however depends on the spike sorter, so it's important to understand the specific input requirements in a case per case basis.
```
zm711, line 9:
```suggestion
The specific behavior however depends on the spike sorter, so it's important to understand the specific input requirements on a case per case basis.
```
:)

doc/how_to/physical_units.rst
```diff
@@ -0,0 +1,107 @@
...
+
+Many preprocessing tools are also linear transformations, and if the ADC is implemented as a linear transformation which is fairly common, then the overall effect can be preserved.
+That is, **preprocessing steps can often be applied either before or after unit conversion without affecting the outcome.**. That said, there are rough edges to this approach.
```
zm711, line 12:
```suggestion
That is, **preprocessing steps can often be applied either before or after unit conversion without affecting the outcome.**. That being said, there are rough edges to this approach.
```
this is the way to say that

zm711 (commented):
Two grammar fixes :) 

still slight preference for a mentioned of `return_scaled` but after merging the grammar fixes I'm fine with approving and then if we have user issues/confusion we can adjust to what they need.

h-mayorquin:
> Two grammar fixes :)
> 
> still slight preference for a mentioned of `return_scaled` but after merging the grammar fixes I'm fine with approving and then if we have user issues/confusion we can adjust to what they need.

I don't want to mention this until https://github.com/SpikeInterface/spikeinterface/pull/3794 is done.

zm711:
> I don't want to mention this until https://github.com/SpikeInterface/spikeinterface/pull/3794 is done.

Yeah I forgot about that until you pushed another commit :) 

That works for me!

zm711 (approved):

Issue #3722, opened 2025-02-25
SpykingCircus2 clustering crash at template estimation


b-grimaud:
I'm trying to run SC2 on 3Brain HD-MEA data (4096 channels, 20 kHz) with mostly default parameters : 

```
p = {
 'apply_motion_correction': False,
 'apply_preprocessing': True,
 'cache_preprocessing': {'delete_cache': True,
                         'memory_limit': 0.5,
                         'mode': 'zarr'},
 'clustering': {'legacy': True},
 'debug': False,
 'detection': {'detect_threshold': 5, 'peak_sign': 'neg'},
 'filtering': {'filter_order': 2,
               'freq_max': 7000,
               'freq_min': 150,
               'ftype': 'bessel',
               'margin_ms': 10},
 'general': {'ms_after': 2, 'ms_before': 2, 'radius_um': 75},
 'job_kwargs': {'n_jobs': 40,},# 'total_memory': '50G'},
 'matched_filtering': True,
 'matching': {'method': 'circus-omp-svd'},
 'merging': {'auto_merge': {'corr_diff_thresh': 0.25, 'min_spikes': 10},
             'correlograms_kwargs': {},
             'similarity_kwargs': {'max_lag_ms': 0.2,
                                   'method': 'cosine',
                                   'support': 'union'}},
 'motion_correction': {'preset': 'dredge_fast'},
 'multi_units_only': False,
 'seed': 42,
 'selection': {'method': 'uniform',
               'min_n_peaks': 100000,
               'n_peaks_per_channel': 5000,
               'seed': 42,
               'select_per_channel': False},
 'sparsity': {'amplitude_mode': 'peak_to_peak',
              'method': 'snr',
              'threshold': 0.25},
 'whitening': {'mode': 'local', 'regularize': False}}
```

The only trace I get is : 

```
spykingcircus2 could benefit from using torch. Consider installing it
Preprocessing the recording (bandpass filtering + CMR + whitening)
noise_level (workers: 20 processes): 100%|███████████████████████████████████████████████████| 20/20 [00:24<00:00,  1.23s/it]
Use zarr_path=/tmp/spikeinterface_cache/tmpdx30z0db/CCLII4NL.zarr
write_zarr_recording 
engine=process - n_jobs=40 - samples_per_chunk=19,753 - chunk_memory=308.64 MiB - total_memory=12.06 GiB - chunk_duration=1.00s (999.96 ms)
[...]

PR #3847, opened 2025-04-08
Returned svd

yger:
Make an example, with SC2, of how one can avoid template estimation from raw data with SVD components with graph-based clustering (or circus clustering). Note that there is also a slight refactoring of SC2 to make the code more component-friendly while writing the paper. With such a version, there should be no memory problem with very large number of templates and/or electrodes. This is solving #3722


src/spikeinterface/sortingcomponents/clustering/circus.py
```diff
@@ -314,4 +228,4 @@ def main_function(cls, recording, peaks, params, job_kwargs=dict()):
         if verbose:
             print("Kept %d non-duplicated clusters" % len(labels))
 
-        return labels, peak_labels
+        return labels, peak_labels, svd_model, peaks_svd, sparse_mask
```
samuelgarcia, line 231: Non, il faut une option pour avoir plus de sorti sinon ça va casser les benchmark

src/spikeinterface/sortingcomponents/clustering/graph_clustering.py
```diff
@@ -191,7 +195,7 @@ def main_function(cls, recording, peaks, params, job_kwargs=dict()):
         labels_set = np.unique(peak_labels)
         labels_set = labels_set[labels_set >= 0]
 
-        return labels_set, peak_labels
+        return labels_set, peak_labels, svd_model, peaks_svd, sparse_mask
```
samuelgarcia, line 198: idem

src/spikeinterface/sortingcomponents/clustering/tools.py
```diff
@@ -228,17 +231,34 @@ def get_templates_from_peaks_and_recording(
...
-    labels = np.unique(peak_labels[mask])
+    valid_peaks = peaks[mask]
+    valid_labels = peak_labels[mask]
+    labels, indices = np.unique(valid_labels, return_inverse=True)
+
     fs = recording.get_sampling_frequency()
     nbefore = int(ms_before * fs / 1000.0)
     nafter = int(ms_after * fs / 1000.0)
 
+    sorting = np.zeros(valid_peaks.size, dtype=minimum_spike_dtype)
```
samuelgarcia, line 245:
```suggestion
    spikes = np.zeros(valid_peaks.size, dtype=minimum_spike_dtype)
```

yger: done

yger:
This is good to me @samuelgarcia am, and now, with such a PR, we can even use KS clustering internally in SC2. Keeping everything else constant, the perfs are really good! Clustering is still currently the remaining component that should be optimized (graph_clustering is an attempt)

b-grimaud:
I'm getting very different results than with the older version with default params on the same data.

Previous one : 
```
Found 24348 spikes
Kept 1220 units after final merging
spykingcircus2 run time 31510.87s
```

New one : 
```
Found 8092 spikes                         
Kept 26 units after final merging 
spykingcircus2 run time 648.18s
```

I also get some warnings : 

```
Build distance graph over channels:  99%|██████████████████████████████████████████████████████▌| 4062/4095 [00:13<00:00, 305.77it/s]
/home/user/.conda/envs/spike/lib/python3.12/site-packages/sklearn/decomposition/_truncated_svd.py:275: RuntimeWarning: invalid value encountered in divide
  self.explained_variance_ratio_ = exp_var / full_var
```

However, if I lower the detection threshold this warning disappears so I'm guessing it's due to too few spikes ?

I also get this one at the end, which I didn't before, presumably when saving the analyzer ?

```
/home/user/Documents/GitHub/spikeinterface/src/spikeinterface/core/sortinganalyzer.py:431: UserWarning: The Recording is not serializable! The recording link will be lost for future load
  warnings.warn("The Recording is not serializable! The recording link will be lost for future load")
```

yger:
Can you pull again and retry? I put back the circus clustering as a default (so you should not see graph clustering and face these warnings), but with svd estimation of the templates. This is faster, and in my case finding only hundreds of templates, that do make sense without memory saturation hopefully. You should not get only 26, or are we taking about the same data ? Otherwise I'll check tomorrow, but this is clearly not the numer I am getting, so something must be strange

b-grimaud:
Yes, that was my bad, I got the most recent commits and it is indeed working better.

With this PR, I get : 
```
Kept 138396 peaks for clustering
Found 3508 raw clusters, starting to clean with matching
```
Then OOM error as it hits `remove_duplicates_via_matching`.

I then merged with #3856 and get : 
```
Kept 304231 peaks for clustering
Found 4079 raw clusters, starting to clean with matching
```
Then OOM again.

Looking at the commits, it doesn't look like `remove_duplicates_via_matching` is under any condition so I guess it's still running ?

yger:
Arg yes, sorry, I'll make an option to make this optional and you should be good to go. Comment the line for now, but this will be exposed at a higher level. At least, templates are computed from svd in memory and thus not more errors because of estimate_templates()

yger:
#3856 is a different issue, not for your own purpose. This is a way, for low density probes, to avoid template matching when templates are found, but instead assign labels to all peaks given the SVD projections. Because the clustering is performed only on a subset of the peaks. But this is only interesting for low density probes, where template matching can sometimes do more harm than good. In your case (dense, lots of channels), you should not activate such a mode

yger:
Now this is good to go for you, remove_mixtures is disabled by default, this should work on your side, let me know

samuelgarcia (approved):

samuelgarcia:
bien joué!

h-mayorquin:
I think this broke some tests.

b-grimaud:
Apologies for the delay, I was away and had not been able to try it out more thoroughly.

I can process a 10 minutes long recording in about 18 hours with `torch` and ~30 jobs. The slowest step seems to be the final merging process, which is still pretty memory intensive.

I had to set up a less-than-elegant solution of having a 500 GB swap file on an SSD for some other computing tasks, which might explain why things are slow. I'm currently trying again with lower `n_jobs` to see if I can avoid swapping and actually run things faster, but at some point there might be diminishing returns. 

Results look good with default parameters, but the quality metrics suggest some unit contamination. I'll try to mess around with clustering params to see if it can be improved.

yger:
This has been merged to main, but strange that you need such a large amount of swap. Nothing is written to disk, so I'm rather curious... Note also that torch is not really faster in case of wobble and circus-omp-svd, maybe this could be the reason also of latencies. I'll investigate why the final merging is so long, because this should not be the case. How many units are found?

b-grimaud:
The size of the swap space is because of something unrelated, I think at most SC2 used 25-30 GB of it.
I'll try without torch and see if it makes a difference. I also just noticed a warning at the beginning : the full recording is around 90 GB so it cannot be fully preloaded in RAM, I'll try adjusting the preprocessing memory limit just enough.
Final unit count is a little under 5000.

yger:
Ok, then I'll make some tests with such high number of units. If you update from main your branch, the memory consumption should have been reduced during matching (if using circus-omp-svd) because of #3889 that I've just merged

PR #3849, opened 2025-04-08
WhiteMatterRecordingExtractor

### LINKED ISSUE #3848: Feature Request: WhiteMatter
I am currently working on a conversion project for the Schneider Lab at NYU and they record ephys data using WhiteMatter: https://white-matter.com/

This data gets saved in a minimal binary .bin file with just the raw data.

pauladkisson:
Fixes #3848 

Pretty minimal wrapper around BinaryRecordingExtractor with a specified file_offset, dtype, and gain. As far as I know these should be constant for WhiteMatter files, but I only have one example, and no formal spec from WhiteMatter (so far). So, I think we should keep an eye out for WhiteMatter files in future conversions and be prepared to update this extractor as necessary.


src/spikeinterface/extractors/whitematterrecordingextractor.py
```diff
@@ -0,0 +1,78 @@
...
+
+        Parameters
+        ----------
+        recording : RecordingExtractor
+            The recording extractor object to be saved in .dat format
+        file_paths : str
+            The path to the file.
+        """
+        job_kwargs["byte_offset"] = 8
+        BinaryRecordingExtractor.write_recording(recording, file_paths=file_paths, dtype="int16", **job_kwargs)
```
alejoe91, line 74: I would remove this and just write it in the test

src/spikeinterface/extractors/whitematterrecordingextractor.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+# Define function equivalent for convenience
+read_whitematter = define_function_from_class(source_class=WhiteMatterRecordingExtractor, name="read_whitematter")
```
alejoe91, line 78: can you add this to the `api.rst` too?

src/spikeinterface/core/tests/test_whitematterrecordingextractor.py
```diff
@@ -0,0 +1,107 @@
...
+
+    file_paths = [cache_folder / f"test_WhiteMatterRecordingExtractor_{i}.raw" for i in range(num_seg)]
+    for i in range(num_seg):
+        np.memmap(file_paths[i], dtype=dtype, mode="w+", shape=(num_samples, num_channels))
+
+    rec = WhiteMatterRecordingExtractor(
+        file_paths=file_paths,
+        sampling_frequency=sampling_frequency,
+        num_channels=num_channels,
+    )
```
alejoe91, line 24: How is this working? You are writing the file without an offset, and the `WhiteMatterRecording` excpects an 8-bit offset..

I would remove this test and just keep the other two

src/spikeinterface/extractors/whitematterrecordingextractor.py
```diff
@@ -0,0 +1,78 @@
...
+    ----------
+    file_paths : list or Path
+        List of paths to the binary files.
+    sampling_frequency : float
+        The sampling frequency.
+    num_channels : int
+        Number of channels in the recording.
+    channel_ids : list or None, default: None
+        A list of channel ids. If None, channel_ids = list(range(num_channels)).
+    time_axis : int, default: 0
```
h-mayorquin, line 25: I don't think`time_axis` should be an argument here as this is fxied right?
Also, why provide more than one file_path, I think the data that we had is mono-segment, right?

h-mayorquin: Yeah, I guess that the point of having this extractor is to simplify the interface and fix the things that we can fix (like you are doing with offset).

Why we would not fix these two if we know that the time axis and the number of segments?

src/spikeinterface/core/tests/test_whitematterrecordingextractor.py
pauladkisson: Done!

src/spikeinterface/extractors/whitematterrecordingextractor.py
pauladkisson: Done!

src/spikeinterface/extractors/whitematterrecordingextractor.py
pauladkisson: I think we should fix them (see latest), just didn't consider it that deeply.

src/spikeinterface/core/tests/test_whitematterrecordingextractor.py
```diff
@@ -0,0 +1,78 @@
...
+    # Ttest for a sub-set of the traces
+    start_frame = 20
+    end_frame = 40
+    smaller_traces = recording.get_traces(start_frame=start_frame, end_frame=end_frame)
+    binary_smaller_traces = binary_recorder.get_traces(start_frame=start_frame, end_frame=end_frame)
+
+    np.allclose(smaller_traces, binary_smaller_traces)
+
+
+def test_sequential_reading_of_small_traces(tmp_path):
```
h-mayorquin, line 38: This is not really necessary here because we are testing that in the binary extractor

src/spikeinterface/core/tests/test_whitematterrecordingextractor.py
```diff
@@ -0,0 +1,78 @@
+import pytest
+import numpy as np
+
+from spikeinterface.extractors import WhiteMatterRecordingExtractor, BinaryRecordingExtractor
+from spikeinterface.core.numpyextractors import NumpyRecording
+
+
+def test_round_trip(tmp_path):
```
h-mayorquin, line 8: This is OK but I would prefer testing data. Can we add something to gin? This should be easy to stub.

h-mayorquin (commented):
This LGTM.
I think the read small traces test should be removed. No need to test here was is tested already.

I would like to have testing data on gin for this but I think @alejoe91 should make the call if this is necessary or not for this for merging.

Otherwise this has my +1.

zm711:
Two minor comments unrelated to the PR content.
1) We typically do a forking workflow. Our docs do say that for dev, but lately we've been sneaking in more PRs from project branches. Do we want to loosen that requirement for people affiliated with the project or do we want to be strict about this (other than CI/release branches)
2) Is there a reason why we wouldn't want to implement this at the Neo level just to keep things separate? is it a concern about Neo releases? If so we could do the extractor like this for now, but work on a neo level and then convert? I think there is a benefit to keep as much non-sorting extractor stuff hived off. But if there is a good reason for us to not do that I would love to know people's opininons.

alejoe91:
I agree it's not worth including in NEO at this point!

src/spikeinterface/extractors/whitematterrecordingextractor.py
```diff
@@ -0,0 +1,62 @@
...
+    file_path : Path
+        Path to the binary file.
+    sampling_frequency : float
+        The sampling frequency.
+    num_channels : int
+        Number of channels in the recording.
+    channel_ids : list or None, default: None
+        A list of channel ids. If None, channel_ids = list(range(num_channels)).
+    is_filtered : bool or None, default: None
+        If True, the recording is assumed to be filtered. If None, is_filtered is not set.
```
zm711, line 26:
```suggestion
        If True, the recording is assumed to be filtered. If None, `is_filtered` is not set.
```

src/spikeinterface/extractors/whitematterrecordingextractor.py
```diff
@@ -0,0 +1,62 @@
+from pathlib import Path
+from typing import List, Union, Optional
```
zm711, line 2: We are far enough along in python that we can use the built-in list, but since you're pulling in Union and Optional if you want to use List it doesn't cost extra I suppose.

src/spikeinterface/extractors/whitematterrecordingextractor.py
```diff
@@ -0,0 +1,62 @@
...
+        num_channels: int,
+        channel_ids: Optional[List] = None,
+        is_filtered: Optional[bool] = None,
+    ):
+        # Specific parameters for WhiteMatter format
+        dtype = "int16"
+        gain_to_uV = 6.25e3 / 32768
+        offset_to_uV = 0.0
+        file_offset = 8
+        time_axis = 0
```
zm711, line 44: Again more cosmetic I suppose but we could put these at the top of the file with a better note about why we think these are the parameters.  And then they could be CAPS since they are constant for this system. Open to everyone hating this idea. But I'm just worried that this format is based on one file example and is not accurate to the format so I'm just trying to think of the best way for us to represent that in the file so that if we get a bug and someone else tries to look at it they can instantly see that we must've had faulty assumptions.

src/spikeinterface/extractors/whitematterrecordingextractor.py
zm711: This is a policy we enacted, but might not be in the dev docs yet. We put ticks around arguments/attributes to highlight that fact.

zm711 (commented):
Couple comments :)

src/spikeinterface/extractors/whitematterrecordingextractor.py
pauladkisson: Seems like a reasonable concern to me. What do you think of this?

https://github.com/SpikeInterface/spikeinterface/blob/3b3bde15816ca99cd35a3f8f9a8594a5d489d0a0/src/spikeinterface/extractors/whitematterrecordingextractor.py#L7-L15

zm711: Yeah that's exactly what I was thinking. As long as others are cool with this.

pauladkisson:
Any other issues holding up this PR?

h-mayorquin:
Data and removing of unnecessary test for me. Either Alessio and Sam should take a look as well.

src/spikeinterface/extractors/whitematterrecordingextractor.py
h-mayorquin: I am neutral about the whole Fortran caps thing but I think it would be better if they were class attributes to avoid floating variables as we don't guard imports.

src/spikeinterface/core/tests/test_whitematterrecordingextractor.py
pauladkisson: Ok, this is awaiting gin access. Should be up in a day or two.

pauladkisson: See https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/155

h-mayorquin: Ok, waiting for the test and should be good to go after that.

pauladkisson: > Ok, waiting for the test and should be good to go after that.

Done!

pauladkisson:
Not sure why the file is not being found...

zm711:
> Not sure why the file is not being found...

I think our tests do not update the ephy_cache all the time. So I'm wondering if your specific file changes aren't triggering a cache check and so you're not getting the updated version. GIN is down this weekend for maintenance, but if this still isn't working Monday after a cache refresh from the cron job then I think we should explore this more.

src/spikeinterface/core/tests/test_whitematterrecordingextractor.py
h-mayorquin, line 1: this should not be in core tests but in extractors test.

src/spikeinterface/core/tests/test_whitematterrecordingextractor.py
```diff
@@ -0,0 +1,57 @@
...
+    start_frame = 20
+    end_frame = 40
+    smaller_traces = recording.get_traces(start_frame=start_frame, end_frame=end_frame)
+    binary_smaller_traces = binary_recorder.get_traces(start_frame=start_frame, end_frame=end_frame)
+
+    np.allclose(smaller_traces, binary_smaller_traces)
+
+
+file_path = (
+    get_global_dataset_folder()
```
h-mayorquin, line 40: This is not meant to work this way.

You need to use `download_dataset`:

https://github.com/h-mayorquin/spikeinterface/blob/f73c3e6ccb48dd113b6a9348b59245adc26ec106/src/spikeinterface/core/datasets.py#L12-L17

See an use here in the common test suite:

https://github.com/h-mayorquin/spikeinterface/blob/f73c3e6ccb48dd113b6a9348b59245adc26ec106/src/spikeinterface/extractors/tests/common_tests.py#L16-L25

And we should check that the data gets in the cache.

h-mayorquin (changes requested):
Let's make the test work and this should be ready to go.

pauladkisson:
Tests are passing now! (minus linux which should be fixed by #3877)

h-mayorquin (approved):
I am OK with this.

alejoe91 (approved):

PR #3854, opened 2025-04-10
Sc2 deterministic mode

yger:
Add the option (as requested by users) to get a deterministic mode to SC2. It will also ease the tracking of errors. False by default, this makes sure seeds are propagated everwhere, and the newly introduced features skip_n_peaks is dealt with. This depends on #3847


samuelgarcia (approved):

alejoe91:
Can we squash and merge this?

Issue #3853, opened 2025-04-10
plot_unit_locations error with scaling


jamcdan:
I seem to run into a problem when trying to plot unit locations from a Neuropixel UHD probe. The scaling seems to be very off for the electrode positions which squishes the actual data to about 1/6th the size of the probe. I say 1/6th because it might be related to the 6um spacing of the probes and some bug related to that? any help sorting this out is appreciated!
![Image](https://github.com/user-attachments/assets/bf8bf671-b1e5-4f06-a6c8-aefa5755a400)

chrishalcrow:
Hi @jamcdan , I suspect the plot is using the "full" probe with all of the electrodes from the NPU device, and you've recorded from the electrodes at the very bottom. Does that sound plausible? Depending on the situation, you might want to plot this, or not. So I think the solution here is to give the user more flexibility in the x/y-limits of the plot (as we do for raster plots and some others). I'll make a PR.

jmcdaniel2:
Yeah that sounds likely, I used the 2x192 configuration just at the bottom of the probe

chrishalcrow:
Great, PR underway. In the meantime @alejoe91 also points out the following:

"we can also control the limits later:
```
w = sw.plot_unit_locations(...)
w.ax.set_xlim(...)
w.ax.set_ylim(...)
```
"
https://github.com/SpikeInterface/spikeinterface/pull/3855#issuecomment-2792166362

zm711:
@jmcdaniel2 did you want to test out the code on main to see if this fixes your issue? Or use the matplotlib strategy? Just trying to figure out if we are good to close :)

jmcdaniel2:
I used the matplotlib way and that worked for me. I can check the new code to see if that method works as well

zm711:
I close this as the issue itself has been solved. But if you are able to test the PR feel free to comment in the closed issue so future users can see :)

PR #3855, opened 2025-04-10
Add docs for custom plotting and change default `limits` in `UnitLocationsWidget`

chrishalcrow:
Fix for #3853

Add to docs (https://spikeinterface--3855.org.readthedocs.build/en/3855/how_to/customize_a_plot.html) and docstrings to show how to customize plots that come from widgets.
Changes the default limits in `UnitLocationsWidget` to use unit info, rather than probe info.

Plots using new defaults:

generated tetrode data:
unit_locs_50_50

real NP2.0 data:
unit_locs_np_50_50

(sorry, should probably be two PRs)


chrishalcrow:
I've never even _considered_ doing that - will add it to the docs to-do list!

Playing with the unit locations, it's not obvious to me what to do. You don't wanna use the actual max/min positions, because then it cuts off the extremum points:

unit_locs_0_0

So you could do something like scale the interval between min and max points by 1.2 or something. Here's those plots for fake tetrode data and real 4-shank NP data:

unit_locs_scale
unit_locs_np_scale

Or you could just add 50 microns to the unit min/max: 

unit_locs_50_50
unit_locs_np_50_50

Thoughts?

alejoe91:
I think that adding a margin of 50 by default and exposing it as an argument would be best

chrishalcrow:
Done!

src/spikeinterface/widgets/unit_locations.py
```diff
@@ -52,6 +53,19 @@ def __init__(
...
+
+        x_locations = all_unit_locations[:, 0]
+        x_min = np.min(x_locations)
+        x_max = np.max(x_locations)
+        x_lim = (x_min - 50, x_max + 50)
+
+        y_locations = all_unit_locations[:, 1]
+        y_min = np.min(y_locations)
+        y_max = np.max(y_locations)
+        y_lim = (y_min - 50, y_max + 50)
```
alejoe91, line 67: Can you explose the marging argument?

doc/how_to/customize_a_plot.rst
```diff
@@ -0,0 +1,139 @@
+Customize a plot
```
alejoe91, line 1: Did you make a .py in the examples/how_to? You should push that as well!

alejoe91 (changes requested):
Thansk @chrishalcrow 

See comments

alejoe91 (approved):

doc/how_to/customize_a_plot.rst
```diff
@@ -0,0 +1,140 @@
...
+================
+
+The ``SpikeInterface`` widgets are designed to have reasonable default
+plotting options, but sometimes you’ll want to make adjustments to the
+plots. For doing this, we expose the underlying ``matplotlib`` objects
```
zm711, line 6: We might want to make it clearer that we expose a widget that gives easy access to the matplotlib objects? I'm used to other libraries returning an `ax` object but if we are returning something more complicated/not an `ax`/`fig` I think we should make that clear.

doc/how_to/customize_a_plot.rst
```diff
@@ -0,0 +1,140 @@
...
+
+    <class 'matplotlib.figure.Figure'>
+    <class 'matplotlib.axes._axes.Axes'>
+
+
+
+.. image:: customize_a_plot_files/customize_a_plot_4_1.png
+
+
+By getting with the matplotlib objects, we gain access to the full
```
zm711, line 60: ? This sounds vaguely inappropriate, no? By gaining access to the matplotlib objects maybe?

So 

```
By gaining access to the matplotlib objects, we are able to use the .... machinery
```

doc/how_to/customize_a_plot.rst
```diff
@@ -0,0 +1,140 @@
...
+``matplotlib`` machinery: adding custom titles, axis labels, ticks, more
+plots etc. Let’s cusomtise our unit locations plot. (Note: the
```
zm711, line 62:
```suggestion
plots etc. Let’s customize our unit locations plot. (Note: the
```
both a typo and a Brittishism.

zm711 (commented):
I put notes on the RST (oops) please just transfer my thoughts to the .py for conversion.

doc/how_to/customize_a_plot.rst
```diff
@@ -0,0 +1,140 @@
...
+
+.. code::
+
+    # Get the widget
+    fig_units = si.plot_unit_locations(sorting_analyzer)
+
+    # Modify the widget's `axis`` to set the title and axes labels
+    fig_units.ax.set_title("My favorite units", fontname = "Comic Sans MS")
+    fig_units.ax.set_xlabel("x probe location (um)")
+    fig_units.ax.set_xlabel("y probe location (um)")
```
zm711, line 74:
```suggestion
    fig_units.ax.set_yabel("y probe location (um)")
```
found another tiny guy :)

zm711 (approved):
This looks good to me too.

PR #3857, opened 2025-04-10
ake 2 on export_to_ibl

### LINKED ISSUE #2843: [feature request] Export to IBL Atlas tool?

jonahpearl:
Attempts to resolve #2843, again :) This version doesn't require any utility functions outside of what spikeinterface already has, plus some functionality from numpy and scipy.

I expect that someone who knows SI a bit better than me will be able to take the `_get_rms` function and convert it into a class that can use the ChunkExtractor to speed things up with multi-processing. If no one gets around to it by the next month I can have a go, I just have some deadlines in the meantime. The LFP spectrum is currently just calculated on a single snippet of LFP data, so I don't think it needs to be parallelized.

### DIFF
--- a/src/spikeinterface/exporters/to_ibl.py
+++ b/src/spikeinterface/exporters/to_ibl.py
@@ -1,0 +1,294 @@
+from __future__ import annotations
+
+import os
+import shutil
+import warnings
+from pathlib import Path
+from typing import Optional
+
+import numpy as np
+import numpy.typing as npt
+from tqdm.auto import tqdm
+
+from spikeinterface.core import ChannelSparsity, SortingAnalyzer
+from spikeinterface.core.job_tools import divide_segment_into_chunks
+from spikeinterface.core.template_tools import get_template_extremum_channel
+from spikeinterface.exporters import (
+    export_to_phy,
+)
+
+
+def export_to_ibl(
+    analyzer: SortingAnalyzer,
+    output_folder: str | Path,
+    lfp_recording=None,
+    rms_win_length_sec=3,
+    welch_win_length_samples=2**14,
+    total_secs_spec_dens=100,
+    only_ibl_specific_steps=False,
+    sparsity: Optional[ChannelSparsity] = None,
+    remove_if_exists: bool = False,
+    verbose: bool = True,
+    **job_kwargs,
+):
+    """
+    Exports a sorting analyzer to the IBL gui format (similar to the Phy format with some extras).
+
+    Parameters
+    ----------
+    analyzer: SortingAnalyzer
+        The sorting analyzer object to use for spike information. 
+        Should also contain the pre-processed recording to use for AP-band data.
+    output_folder: str | Path
+        The output folder for the exports.
+    lfp_recording: Any SI Recording object, default None
+        The pre-processed recording to use for LFP data. If None, the LFP data is not exported.
+    rms_win_length_sec: float, default: 3
+        The window length in seconds for the RMS calculation (on the LFP data).
+    welch_win_length_samples: int, default: 2^14
+        The window length in samples for the Welch spectral density computation (on the LFP data).
+    total_secs_spec_dens: int, default: 100
+        The total number of seconds to use for the spectral density calculation.
+    only_ibl_specific_steps: bool, default: False
+        If True, only the IBL specific steps are run (i.e. skips calling `export_to_phy`)
+    sparsity: ChannelSparsity or None, default: None
+        The sparsity object (currently only respected for phy part of the export)
+    remove_if_exists: bool, default: False
+        If True and "output_folder" exists, it is removed and overwritten
+    verbose: bool, default: True
+        If True, output is verbose
+
+    """
+
+    try:
+        from scipy.signal import welch
+    except ImportError as e:
+        raise ImportError(
+            "Please install scipy to use the export_to_ibl function."
+        ) from e
+
+    # Output folder checks
+    if isinstance(output_folder, str):
+        output_folder = Path(output_folder)
+    output_folder = Path(output_folder).absolute()
+    if output_folder.is_dir():
+        if remove_if_exists:
+            shutil.rmtree(output_folder)
+        else:
+            raise FileExistsError(f"{output_folder} already exists")
+    else:
+        pass
+        # don't make the output dir yet, b/c export_to_phy will do that for us.
+
+    if verbose:
+        print("Exporting recording to IBL format...")
+
+    # Compute any missing extensions
+    available_extension_names = analyzer.get_saved_extension_names()
+    required_exts = [
+        "templates",
+        "template_similarity",
+        "spike_locations",
+        "noise_levels",
+        "quality_metrics",
+    ]
+    required_qms = ["amplitude_median", "isi_violations_ratio", "amplitude_cutoff"]
+    for ext in required_exts:
+        if ext not in available_extension_names:
+            if ext == "quality_metrics":
+                kwargs = {"skip_pc_metrics": False}
+            else:
+                kwargs = {}
+            analyzer.compute(ext, verbose=verbose, **kwargs)
+        elif ext == "quality_metrics":
+            qm = analyzer.get_extension("quality_metrics").get_data()
+            for rqm in required_qms:
+                if rqm not in qm:
+                    analyzer.compute(
+                        "quality_metrics",
+                        metric_names=[rqm],
+                        verbose=verbose,
+                    )
+
+    # # Start by just exporting to phy
+    if not only_ibl_specific_steps:
+        if verbose:
+            print("Doing phy-like export...")
+        export_to_phy(
+            analyzer,
+            output_folder,
+            compute_amplitudes=True,
+            compute_pc_features=False,
+            sparsity=sparsity,
+            copy_binary=False,
+            template_mode="median",
+            verbose=verbose,
+            use_relative_path=False,
+            **job_kwargs,
+        )
+
+    # Make sure output dir exists, in case user skips export_to_phy
+    if not output_folder.is_dir():
+        os.makedirs(output_folder)
+
+    if verbose:
+        print("Running IBL-specific steps...")
+
+    # Now we need to add the extra IBL specific files
+    # See here for docs on the format: https://github.com/int-brain-lab/iblapps/wiki/3.-Overview-of-datasets#input-histology-data
+
+    # Subset channels in case some were excluded from spike sorting
+    (channel_inds,) = np.isin(
+        analyzer.recording.channel_ids, analyzer.channel_ids
+    ).nonzero()
+
+    # TODO: put this into a chunk extractor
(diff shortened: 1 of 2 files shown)


src/spikeinterface/exporters/to_ibl.py
```diff
@@ -0,0 +1,274 @@
...
+    verbose: bool, default: True
+        If True, output is verbose
+
+    """
+
+    try:
+        from scipy.signal import welch
+    except ImportError as e:
+        raise ImportError("Please install scipy to use the export_to_ibl function.") from e
+
```
zm711, line 67: Use `importlib.util.find_spec` it is much less costly for users that don't have the function. https://docs.python.org/3/library/importlib.html

Then you raise the error without the expense of the try. Once you know you have the package then you just import.

src/spikeinterface/exporters/to_ibl.py
```diff
@@ -0,0 +1,274 @@
...
+    spike_locations = analyzer.load_extension("spike_locations").get_data()
+    spike_depths = spike_locations["y"]
+
+    # convert clusters and squeeze
+    clusters = np.load(output_folder / "spike_clusters.npy")
+    np.save(output_folder / "spike_clusters.npy", np.squeeze(clusters.astype("uint32")))
+
+    # convert times and squeeze
+    times = np.load(output_folder / "spike_times.npy")
+    np.save(output_folder / "spike_times.npy", np.squeeze(times / 30000.0).astype("float64"))
```
zm711, line 212: Where did you get the 30000 from? Some machines can be run at different rates?

src/spikeinterface/exporters/to_ibl.py
```diff
@@ -0,0 +1,274 @@
...
+    np.save(output_folder / "channel_inds.npy", np.arange(len(channel_inds), dtype="int"))
+
+    # # save templates
+    cluster_channels = []
+    cluster_peakToTrough = []
+    cluster_waveforms = []
+    templates = analyzer.get_extension("templates").get_data()
+    extremum_channel_indices = get_template_extremum_channel(analyzer, outputs="index")
+
+    for unit_idx, unit_id in enumerate(analyzer.unit_ids):
```
zm711, line 229: light preference for writing out index going forward so unit_index just to be extra clear to everyone.

src/spikeinterface/exporters/to_ibl.py
```diff
@@ -0,0 +1,274 @@
...
+        waveform = templates[unit_idx, :, :]
+        extremum_channel_index = extremum_channel_indices[unit_id]
+        peak_waveform = waveform[:, extremum_channel_index]
+        peakToTrough = (np.argmax(peak_waveform) - np.argmin(peak_waveform)) / analyzer.sampling_frequency
```
zm711, line 233: I think we prefer keeping things pythonic except when saving dict values for other formats. so this would stay `peak_to_trough`.

zm711: same for other variable names.

zm711 (commented):
I think this is cool. I think Alessio would have to take the `rms` function to deal with the chunking. But keeping things within the si framework is super nice. a couple initial comments that should be super easy to do.

src/spikeinterface/exporters/to_ibl.py
alejoe91: yeah this should be `analyzer.sampling_frequency`

alejoe91:
Will do! Thanks for the re-writing @jonahpearl !!

jonahpearl:
-- Added a test (passes locally for me), made a few fixes based on that
-- Loosened the IBL unit standards a bit, since it's presumably ok to have some mua here. Could allow users to pass their own criteria?
-- Fixed skip_pc_metrics to be True (which was my original intention, just got the sign of the parameter flipped in my head). This should speed things up for large recordings, and if anyone really wants the PC metrics they can use the export to phy feature with those on.
-- Slightly simplified the bit that checks the pre-computed qm's.

src/spikeinterface/exporters/to_ibl.py
```diff
@@ -0,0 +1,280 @@
...
+        The sparsity object (currently only respected for phy part of the export)
+    remove_if_exists: bool, default: False
+        If True and "output_folder" exists, it is removed and overwritten
+    verbose: bool, default: True
+        If True, output is verbose
+
+    """
+
+    if find_spec("scipy") is None:
+        raise ImportError("Please install scipy to use the export_to_ibl function.")
```
zm711, line 64:
```suggestion
        raise ImportError("Please install scipy to use the `export_to_ibl` function.")
```

just our convention to put ticks around library things (so functions or args). Since we just started this convention not everything is done this way, but with stuff moving forward we want to enforce it :)

src/spikeinterface/exporters/to_ibl.py
```diff
@@ -0,0 +1,280 @@
...
+            else:
+                kwargs = {}
+            analyzer.compute(ext, verbose=verbose, **kwargs)
+
+    # Check in case user pre-calculated a small set of qm's that aren't enough for IBL
+    required_qms = ["amplitude_median", "isi_violation", "amplitude_cutoff"]
+    qm = analyzer.get_extension("quality_metrics").get_data()
+    for rqm in required_qms:
+        if rqm not in qm:
+            analyzer.compute(
```
zm711, line 109: Just to make sure we aren't missing out on any parallelization now or in the future wouldn't it make more sense to do something like

```python

required_qms = xx
qm = analyzer.get_extension('quality_metrics').get_data()
rqm = [metric for metric in required_qms if metric not in qm]
analyzer.compute(...)
```
that way we get the small speed up of a list comprehension and we only run the analyzer compute once at this step.

src/spikeinterface/exporters/to_ibl.py
```diff
@@ -0,0 +1,280 @@
...
+            traces = rec.get_traces(start_frame=start_frame, end_frame=stop_frame)
+            chunk_rms[iChunk, :] = np.sqrt(np.mean(traces**2, axis=0))
+            chunk_start_times[iChunk] = start_frame / rec.sampling_frequency
+        chunk_rms = chunk_rms[:, channel_inds]
+        chunk_rms = chunk_rms.astype(np.float32)
+        chunk_start_times = chunk_start_times.astype(np.float32)
+        return chunk_rms, chunk_start_times
+
+    # Get RMS for the AP data. We will use a window of length rms_win_length_sec seconds slid over the entire recording.
+    ap_rec = analyzer.recording
```
zm711, line 161: Since we don't always have a recording for an analyzer we should probably do some sort of error checker above to ensure the current analyzer has access to a recording first. If this isn't required then I would make it optional so that users can use this if they don't have a recording with their analyzer. (For me I tend to store my recs on a server after I make my analyzers).

src/spikeinterface/exporters/to_ibl.py
```diff
@@ -0,0 +1,280 @@
...
+            os.path.join(output_folder, "_iblqc_ephysTimeRmsLF.timestamps.npy"),
+            chunk_start_times,
+        )
+
+        # Get spectral density on a snippet of LFP data
+        end_frame = int(total_secs_spec_dens * lfp_recording.sampling_frequency)
+        traces = lfp_recording.get_traces(start_frame=0, end_frame=end_frame)  # time x channels
+        spec_density = np.zeros((welch_win_length_samples // 2 + 1, traces.shape[1]))
+        for iCh in range(traces.shape[1]):
+            f, Pxx = welch(
```
zm711, line 189: Heberto (and I) really prefer not using single letter variable names. even if it is the standard coming out of the function. Not required to change but something that we do recommend in our dev docs.

src/spikeinterface/exporters/to_ibl.py
```diff
@@ -0,0 +1,280 @@
...
+                traces[:, iCh],
+                fs=lfp_recording.sampling_frequency,
+                nperseg=welch_win_length_samples,
+            )
+            spec_density[:, iCh] = Pxx
+        spec_density = spec_density[:, channel_inds]  # only keep channels that were used for spike sorting
+        spec_density = spec_density.astype(np.float32)
+        f = f.astype(np.float32)
+        assert spec_density.shape[0] == len(f)
```
zm711, line 198: I love assert messages (although I personally prefer errors--but that is a battle I lost here). So if you're willing to write a message for users that would be lovely.

zm711 (commented):
Hey Jonah few more things on my end. I don't want to generate too much work before Alessio hops onto this PR.  But I think we definitely need to protect for the case where the user no longer has a useable recording.

src/spikeinterface/exporters/to_ibl.py
jonahpearl: it was just for my testing, removed it

zm711:
Thanks Jonah. I'll leave this alone now until Alessio deals with it just to ensure we don't end up with conflicting stuff! :)

alejoe91:
Hi @jonahpearl 

I finally found some time to review and work on this.

Here are some updates:

- I added parallelization to the compute rms function
- improved chunk retrieval for psd estimation
- remove export to phy and file mapping (I realized we have everything we need in the analyzer!)
- removed auto-computation of extensions and QC
- made the definition of good units more flexible, with a `good_unit_query` that acts on the qm dataframe
- extended tests

Can you test it on your side and make sure that the IBL GUI works well with the output?

jonahpearl:
Just tried it out — works well and fast. Seems like there's one test failing just because it still has the old function name.

alejoe91:
> Just tried it out — works well and fast. Seems like there's one test failing just because it still has the old function name.

Fixed now!

alejoe91 (approved):

alejoe91:
Awesome! I'll merge then :)

jonahpearl:
Thanks Alessio!

PR #3862, opened 2025-04-14
Better Neuroscope support.

theodchrn:
Hi!
This is the first time i'm doing a pull request on an "official" project, and don't really know how to do it. If you have any remark/advice on how to do it properly i'd be glad to have them!

Basically i would just like to share my small improvements on reading neuroscope-type data (amplifier.dat coupled with amplifier.xml). I don't think this format is widely used to date, but for the few of us that still do I think this might be a step forward to using SpikeInterface more easily.

What I've added for my personal use and think might benefit everyone:
- fetches the xml with channel groups
- automatically adds it as a group in the recording
- parses the real colors and positions (offsets) directly from Neuroscope and adds them as properties of the object (handy for ephyviewer)

Will allow for parallel spike sorting, processing... as if it were separate recordings/structures (for now it's only anatomy-based, but we could do the same with the "spike groups" identified after spike sorting by Neuroscope).

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/neuroscope.py
+++ b/src/spikeinterface/extractors/neoextractors/neuroscope.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import warnings
 from pathlib import Path
 from typing import Union, Optional
+from xml.etree import ElementTree as Etree
 
 import numpy as np
 
@@ -64,6 +65,7 @@ class NeuroScopeRecordingExtractor(NeoBaseRecordingExtractor):
         if xml_file_path is not None:
             xml_file_path = str(Path(xml_file_path).absolute())
         self._kwargs.update(dict(file_path=str(Path(file_path).absolute()), xml_file_path=xml_file_path))
+        self.split_recording_by_channel_groups(xml_file_path=xml_file_path if xml_file_path is not None else Path(file_path).with_suffix(".xml"))
 
     @classmethod
     def map_to_neo_kwargs(cls, file_path, xml_file_path=None):
@@ -77,6 +79,65 @@ class NeuroScopeRecordingExtractor(NeoBaseRecordingExtractor):
             neo_kwargs = {"filename": Path(file_path)}
 
         return neo_kwargs
+
+    def _parse_xml_file(self, xml_file_path):
+        """
+        Comes from NeuroPhy package by Diba Lab
+        """
+        tree = Etree.parse(xml_file_path)
+        myroot = tree.getroot()
+
+        for sf in myroot.findall("acquisitionSystem"):
+            n_channels = int(sf.find("nChannels").text)
+
+        channel_groups, skipped_channels, anatomycolors = [], [], {}
+        for x in myroot.findall("anatomicalDescription"):
+            for y in x.findall("channelGroups"):
+                for z in y.findall("group"):
+                    chan_group = []
+                    for chan in z.findall("channel"):
+                        if int(chan.attrib["skip"]) == 1:
+                            skipped_channels.append(int(chan.text))
+
+                        chan_group.append(int(chan.text))
+                    if chan_group:
+                        channel_groups.append(np.array(chan_group))
+
+        for x in myroot.findall("neuroscope"):
+            for y in x.findall("channels"):
+                for i, z in enumerate(y.findall("channelColors")):
+                    try:
+                        channel_id = str(z.find("channel").text)
+                        color = z.find("color").text
+
+                    except AttributeError:
+                        channel_id = i 
+                        color = "#0080ff"
+                    anatomycolors[channel_id] = color
+
+        discarded_channels = np.setdiff1d(np.arange(n_channels), np.concatenate(channel_groups))
+        kept_channels = np.setdiff1d(np.arange(n_channels),np.concatenate([skipped_channels, discarded_channels]))
+
+        return channel_groups, kept_channels, discarded_channels, anatomycolors
+
+    def split_recording_by_channel_groups(self):
+        n = self.get_num_channels()
+        group_ids = np.full(n, -1, dtype=int)  # Initialize all positions to -1
+
+        channel_groups, kept_channels, discarded_channels, colors = (
+            self._parse_xml_file(self.xml_file_path)
+        )
+        for group_id, numbers in enumerate(channel_groups):
+            group_ids[numbers] = (
+                group_id  # Assign group_id to the positions in `numbers`
+            )
+        self.set_property("group", group_ids)
+        discarded_ppty = np.full(n, False, dtype=bool)
+        discarded_ppty[discarded_channels] = True
+        self.set_property("discarded_channels", discarded_ppty)
+        self.set_property(
+            "colors", values=list(colors.values()), ids=list(colors.keys())
+        )
 
 
 class NeuroScopeSortingExtractor(BaseSorting):


zm711:
Thanks for the contribution @theodchrn. Super cool. I think @h-mayorquin recently did something similar for our Intan reading capabilities. Maybe he would be good to give this a glance over?

src/spikeinterface/extractors/neoextractors/neuroscope.py
```diff
@@ -78,6 +81,59 @@ def map_to_neo_kwargs(cls, file_path, xml_file_path=None):
...
+        group_ids = np.full(n, -1, dtype=int)  # Initialize all positions to -1
+
+        channel_groups, kept_channels, discarded_channels, colors = self._parse_xml_file(self.xml_file_path)
+        for group_id, numbers in enumerate(channel_groups):
+            group_ids[numbers] = group_id  # Assign group_id to the positions in `numbers`
+        self.set_property("group", group_ids)
+        discarded_ppty = np.full(n, False, dtype=bool)
+        discarded_ppty[discarded_channels] = True
+        self.set_property("discarded_channels", discarded_ppty)
+        self.set_property("colors", values=list(colors.values()), ids=list(colors.keys()))
```
h-mayorquin, line 135: Unsure about adding colors here but I think that @samuelgarcia and @alejoe91 are more familiar with viewers so I defer to them.

theodchrn:
Hey @h-mayorquin ! Thanks for the review :) I'll try my best to answer those.

1. Channel groups are basically user-derived channels segmentation, e.g. channel ids for the hippocampus, PFC, ACC... They each have their own "group" and group-color in neuroscope. 
2. Discarded just means that the user chose not to visualize the channels in the final recording - eg ref, ground, noisy signal...
3. about the maintainers: the update i'm suggesting only comes because I am using "old" data that has already been processed and visualized through neuroscope - otherwise I would simply start from scratch. So this all_annotations = True would be super handy for people like me!

src/spikeinterface/extractors/neoextractors/neuroscope.py
```diff
@@ -78,6 +81,59 @@ def map_to_neo_kwargs(cls, file_path, xml_file_path=None):
...
+                        channel_id = i
+                        color = "#0080ff"
+                    anatomycolors[channel_id] = color
+
+        discarded_channels = np.setdiff1d(np.arange(n_channels), np.concatenate(channel_groups))
+        kept_channels = np.setdiff1d(np.arange(n_channels), np.concatenate([skipped_channels, discarded_channels]))
+
+        return channel_groups, kept_channels, discarded_channels, anatomycolors
+
+    def split_recording_by_channel_groups(self):
```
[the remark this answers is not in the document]
theodchrn: what do you mean by private ? i'm not super familiar with advanced objects. Is it the __split_recording caracter ?

src/spikeinterface/extractors/neoextractors/neuroscope.py
```diff
@@ -78,6 +81,59 @@ def map_to_neo_kwargs(cls, file_path, xml_file_path=None):
...
+        n = self.get_num_channels()
+        group_ids = np.full(n, -1, dtype=int)  # Initialize all positions to -1
+
+        channel_groups, kept_channels, discarded_channels, colors = self._parse_xml_file(self.xml_file_path)
+        for group_id, numbers in enumerate(channel_groups):
+            group_ids[numbers] = group_id  # Assign group_id to the positions in `numbers`
+        self.set_property("group", group_ids)
+        discarded_ppty = np.full(n, False, dtype=bool)
+        discarded_ppty[discarded_channels] = True
```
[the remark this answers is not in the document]
theodchrn: it just means it wont be shown in the final viewer and was probably considered as noise/unplugged channel on the eib

src/spikeinterface/extractors/neoextractors/neuroscope.py
```diff
@@ -78,6 +81,59 @@ def map_to_neo_kwargs(cls, file_path, xml_file_path=None):
...
+                    try:
+                        channel_id = str(z.find("channel").text)
+                        color = z.find("color").text
+
+                    except AttributeError:
+                        channel_id = i
+                        color = "#0080ff"
+                    anatomycolors[channel_id] = color
+
+        discarded_channels = np.setdiff1d(np.arange(n_channels), np.concatenate(channel_groups))
```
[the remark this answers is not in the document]
theodchrn: i just pushed [this commit](https://github.com/MobsLab/spikeinterface/commit/9b0595f13fa26a44913d6324818bbca82731db7e) to account for that

src/spikeinterface/extractors/neoextractors/neuroscope.py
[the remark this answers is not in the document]
zm711: private in general would be
`_set_groups` so one underscore. The `__` leads to name mangling, which I personally hate. But `_` underscore is a flag to the user that this function is for internal use and that the implementation and arguments could change at any time.

src/spikeinterface/extractors/neoextractors/neuroscope.py
zm711: I think we just need to understand is this something all users should think about or is this super specific. If colors are always generated in a consistent (at least to the user) way then it might be worth adding (again with Heberto's final suggestion to fit this in with `all_annotations` for a future PR). 

Basically we want properties set here to be truly meaningful to all users. Other properties can be set on a per-user basis with our property setting machinery.

theodchrn:
> Thanks for your answers.
> 
> > 1. Channel groups are basically user-derived channels segmentation, e.g. channel ids for the hippocampus, PFC, ACC... They each have their own "group" and group-color in neuroscope.
> 
> Can they be anything else than brain areas? if not, then the property should be named "brain_area". "groups" in SpikeInterface are usually vaguely things that you would sort separately (e.g. a shank in a multi shank neuropixel probe).
> 
Yes it could be something else than brain area - for example in my lab we have EKG and EMG recordings as well, you could probably add breathing data... I called them groups because (at least in my use case) i run the sorting in SpikeInterface based on those groups (they could represent a shank!) - I just skip the groups that are not brain areas. 

> > 2. Discarded just means that the user chose not to visualize the channels in the final recording - eg ref, ground, noisy signal...
> 
> Mmm I am unsure about this. Is this general neuroscope or specific to your lab workflow?
Tbh i dont know. I assumed it was general neuroscope, but maybe it's only us! In my opinion it's also one of the main advantages of Neuroscope, as you directly visualize relevant data, already sorted for the different brain/body areas you're recording.

Let me know if this is not clear enough! English is not my 1st language and I am not very familiar with neuroscope usage outside of of my lab scope.

src/spikeinterface/extractors/neoextractors/neuroscope.py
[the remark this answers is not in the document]
h-mayorquin: ok

h-mayorquin:
> Yes it could be something else than brain area - for example in my lab we have EKG and EMG recordings as well, you could probably add breathing data... I called them groups because (at least in my use case) i run the sorting in SpikeInterface based on those groups (they could represent a shank!) - I just skip the groups that are not brain areas.

Then, maybe we should call them `neuroscope_groups` (or something like that) just to distinguish it from groups. 

It would be nice if we could have more meaningful names but it seems that those properties mean different things for different users.

I need to think more about this but i would like someone who works with visualization or knows neuroscope to review this as I think we need more knowledge for this system  so we don't overfit to a visualization workflow or add things that are not meaningful in other contexts. 

Maybe a compromise so the user can have facilities but at the same time we don't make strong commitments is to add a new method "prepare_neuroscope_for_ephyviewer" which loads these properties but the user needs to call explicitly (I am not sure this work with ephyviewer). I am slightly against adding yet another keyword because it will crash with all_annotations

src/spikeinterface/extractors/neoextractors/neuroscope.py
samuelgarcia: OK for color but maybe we could rename this to *neuroscope_color* no ?

src/spikeinterface/extractors/neoextractors/neuroscope.py
```diff
@@ -78,6 +81,64 @@ def map_to_neo_kwargs(cls, file_path, xml_file_path=None):
...
+        These group ids are usually different brain/body anatomical areas, or shanks from multi-shank probes.
+        The group ids are set as a property of the recording extractor.
+        """
+        n = self.get_num_channels()
+        group_ids = np.full(n, -1, dtype=int)  # Initialize all positions to -1
+
+        channel_groups, kept_channels, discarded_channels, colors = self._parse_xml_file(self.xml_file_path)
+        for group_id, numbers in enumerate(channel_groups):
+            group_ids[numbers] = group_id  # Assign group_id to the positions in `numbers`
+        self.set_property("group", group_ids)
```
samuelgarcia: I agree with @h-mayorquin  for *neuroscope_group*

samuelgarcia:
Hi @theodchrn 
Thanks a lot for this.
I am OK with this.

theodchrn:
Hi
Apologies for the very late replies, I completely forgot to check my github.
You should find the appropriate changes in the last commit.

src/spikeinterface/extractors/neoextractors/neuroscope.py
```diff
@@ -78,6 +80,71 @@ def map_to_neo_kwargs(cls, file_path, xml_file_path=None):
...
```
h-mayorquin:
```suggestion
        self.set_property("neuroscope_group", group_ids)
```

h-mayorquin (approved):
Now that this is divided into a function that can needs to be called intentionally I am OK with it. 

I still think that it would be better to be specific about the group being neuroscope group but I am not sure if this will still work with the viewer.

src/spikeinterface/extractors/neoextractors/neuroscope.py
theodchrn: Done ! I forgot to add it in the previous commit, thanks for the reminder.

theodchrn:
Hi, just so you know the checks are failing for windows, when building wheel for pyedflib. I don't fully understand if it's linked to my changes or not - I don't see why it would be.

zm711:
I don't think it is. We will have to dig into what changed about the package on windows. We recently had to fiddle with versioning for that library so maybe they changed something again and we will need to adjust again. But need to figure out if this is our level or a neo level issue.

theodchrn:
Ok thanks for the confirmation:)

zm711:
Hopefully we have the fix in main now. Just updated your branch so we will see :)

PR #3866, opened 2025-04-15
Incorporated kilosort4's setup_logger() function to ensure generation of kilosort4.log file in sorter_output folder

### LINKED ISSUE #3782: SpikeInterface wrapper for Kilosort4 does not invoke default Kilosort4 logging function
The Kilosort4 API's entry point for the algorithm is a function called "run_kilosort()" located in kilosort/run_kilosort.py. 

The SpikeInterface wrapper for Kilosort4--located at src/spikeinterface/sorters/external/kilosort4.py--does not call "run_kilosort()" directly. Instead, it reproduces the individual function calls contained within run_kilosort(). 

One of the functions calls in the original run_kilosort() is setup_logger(), which is responsible for the generation of "kilosort4.log" in the ks4 output folder (the one that contains all the .npy files). This function call is missing from the SpikeInterface wrapper. 

Including this function call would make debugging various issues more convenient. Would it be possible to do so?

@h-mayorquin @alejoe91 @zm711 @cheydrick

nikhilchandra:
Fixes #3782

### DIFF
--- a/.github/scripts/test_kilosort4_ci.py
+++ b/.github/scripts/test_kilosort4_ci.py
@@ -70,7 +70,6 @@ PARAMS_TO_TEST_DICT = {
     "nearest_chans": 8,
     "nearest_templates": 35,
     "max_channel_distance": 5,
-    "templates_from_data": False,
     "n_templates": 10,
     "n_pcs": 3,
     "Th_single_ch": 4,
@@ -108,6 +107,10 @@ if parse(kilosort.__version__) >= parse("4.0.24"):
     )
     # max_peels is not affecting the results in this short dataset
     PARAMETERS_NOT_AFFECTING_RESULTS.append("max_peels")
+
+if parse(kilosort.__version__) >= parse("4.0.33"):
+    PARAMS_TO_TEST_DICT.update({"cluster_neighbors": 11})
+    PARAMETERS_NOT_AFFECTING_RESULTS.append("cluster_neighbors")
 
 
 PARAMS_TO_TEST = list(PARAMS_TO_TEST_DICT.keys())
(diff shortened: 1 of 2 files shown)


zm711:
At least based on my reading of the errors it seems like they may have played with the arguments for this logging. Hopefully you don't end up needing to do too much logic. Also note we haven't updated for 4.0.31 yet so that test will fail even if you get the logging working on all other versions.

nikhilchandra:
@zm711 I think it works now up to v4.0.30, please confirm!

zm711:
yeah that seems to be working for all of them before 0.31. I think we review this now :) I'm a bit busy today and tomorrow, but I can look over this this weekend unless someone else has time before that.

zm711 (commented):
I think the only thing I'm not sure about is should we only do this logging if the user enters that they want spikeinterface in verbose mode? It seems like you removed the check for verbose @nikhilchandra but we try to require the user request verbose rather than have verbosity be the default. So I wonder if we should only have the logger if the user specifies verbose as a kwarg?

nikhilchandra:
In the original code, when verbose=True then the logging level for what gets printed to the console is INFO. Note that the default logging level is always WARNING.

In the new code, the newly inserted setup_logger() function from kilosort creates a StreamHandler that manages console output. In newer versions of Kilosort, setup_logger() has a function called verbose_console. When True, the StreamHandler's logging level is set to DEBUG. When False, it is set to INFO.

Thus, to replicate the original behavior of spike interface

1. I always set the verbose_console argument to False-->this means that the StreamHandler's logging level is INFO
2. If the spike interface user sets the verbose flag to True, then we can leave the StreamHandler's logging level as is
3. If the spike interface user sets the verbose flag to False, then we can set the StreamHandler's logging level to WARNING

Note: older versions of Kilosort4 (4.0.16-4.0.18) don't name the logger "kilosort", instead opting to use a blank string "". In other versions, setup_logger() does not accept a verbose_console argument. The rest of my changes are just geared towards ensuring that the above logic works across all supported versions of Kilosort.

zm711:
Sorry I actually missed line 197 where you do the switch to warning if not verbose. This looks fine to me then. :)

alejoe91:
Actually, I tested this again and it works. The problem is only that with `verbose=False` we still get progress bars...But I think this is a KS problem

alejoe91:
See https://github.com/MouseLand/Kilosort/pull/935

PR #3872, opened 2025-04-18
Update KS4 wrapper for versions 4.0.31+

zm711:
I think this is the fix we need, but I'll let our CI see whether this is enough for KS4. Seems like the new versions modify the st and tF that need to be saved.


zm711:
@alejoe91 I think this fixes our KS4 wrapper so that we can propagate this to a couple other PRs :)

alejoe91 (approved):

alejoe91:
CI approved!

PR #3877, opened 2025-04-21
Change datalad installation in the CI for linux

### LINKED ISSUE #3876: ubuntu tests failing due to inability to connect to git-annex download
Hey guys

I thought maybe this was a one day issue, but since this started Thursday or Friday and it's now lasted multiple days we may need to troubleshoot this more deeply

```
Connecting to downloads.kitenet.net (downloads.kitenet.net)|66.228.36.95|:443... connected.
HTTP request sent, awaiting response... 403 Forbidden
2025-04-21 12:11:40 ERROR 403: Forbidden.
```
seems like the site where we were downloading git-annex is no longer allowing us access. @h-mayorquin you worked on this so maybe you have knowledge? If not when I have some time I can also try to read up from datalad to see if there are other ways on ubuntu to gain access to git-annex.

h-mayorquin:
Let's test if this works. Will fix #3876 if it does.

### DIFF
--- a/.github/workflows/all-tests.yml
+++ b/.github/workflows/all-tests.yml
@@ -107,14 +107,7 @@ jobs:
         run: |
           pip install datalad-installer
           if [ ${{ runner.os }} = 'Linux' ]; then
-            wget https://downloads.kitenet.net/git-annex/linux/current/git-annex-standalone-amd64.tar.gz
-            mkdir /home/runner/work/installation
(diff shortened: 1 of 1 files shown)


h-mayorquin:
Ok, so this method is included in the datalad documentation which is better. I think we should change to this. I have no memory on why we were using this one, probably my fault as I wanted to avoid neurodebian repositories.

zm711:
I will likely need to propagate this over to neo since I think in the end we took this si ci to get datalad working on python 3.12.

PR #3892, opened 2025-04-29
Skip loading extensions when instantiating benchmarks

alejoe91:
With many cases and large recordings, loading everything in RAM can cause out of mem issues.

We don't really need loading extensions at this point

### DIFF
--- a/src/spikeinterface/benchmark/benchmark_base.py
+++ b/src/spikeinterface/benchmark/benchmark_base.py
@@ -176,17 +176,11 @@ class BenchmarkStudy:
         self.levels = self.info["levels"]
 
         for key, folder in self.analyzers_path.items():
-            analyzer = load_sorting_analyzer(folder)
+            analyzer = load_sorting_analyzer(folder, load_extensions=False)
             self.analyzers[key] = analyzer
             # the sorting is in memory here we take the saved one because comparisons need to pickle it later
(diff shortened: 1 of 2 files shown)


zm711:
Related to this @alejoe91 / @samuelgarcia I think the real heavy extension is the waveforms right? I like to keep my waveforms around (really for provenance), but once I have my templates/PCA done, I don't really want to have to reload them. I can't think of a good mechanism to only autoload the "final" extensions (since that could probably be different for different people), so it would be nice if we could do that, but I guess the only way currently is to not load the extensions and then load them one-by-one?

chrishalcrow:
> Related to this @alejoe91 / @samuelgarcia I think the real heavy extension is the waveforms right? I like to keep my waveforms around (really for provenance), but once I have my templates/PCA done, I don't really want to have to reload them. I can't think of a good mechanism to only autoload the "final" extensions (since that could probably be different for different people), so it would be nice if we could do that, but I guess the only way currently is to not load the extensions and then load them one-by-one?

Agree with @zm711  - we do this in our lab all the time. My proposal: expand the `load_extensions` to accept lists of strings. E.g.
``` python
sa = si.load_sorting_analyzer(
    folder="my_sa", 
    load_extensions=["unit_locations", "templates", "spike_amplitudes"],
)
```
For another PR though!

samuelgarcia (approved):

PR #3893, opened 2025-04-29
Fix plot traces when t_start is defined

alejoe91:
We were not handling correctly the case when time_vector is None, but t_starts are present

### DIFF
--- a/src/spikeinterface/widgets/motion.py
+++ b/src/spikeinterface/widgets/motion.py
@@ -214,7 +214,9 @@ class DriftRasterMapWidget(BaseRasterWidget):
                 amps /= q_95
                 c = cmap(amps)
             else:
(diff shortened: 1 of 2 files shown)


zm711 (commented):
In line 138 we verify that we have an appropriate segment_index for multisegment? Wouldn't it make sense to move those lines before we do the time vector stuff?

alejoe91:
> In line 138 we verify that we have an appropriate segment_index for multisegment? Wouldn't it make sense to move those lines before we do the time vector stuff?
> 
> Otherwise this PR looks good to me. So if you'd prefer to discuss in a different PR just let me know.

Totally! Moved in last commit

chrishalcrow:
Works for me, and looks good

chrishalcrow (approved):

zm711 (approved):
Good by me too :)

PR #3895, opened 2025-04-30
Clarify docstrings in neuropixels_utils

### LINKED ISSUE #3894: Incorrect assumptions in `neuropixels_utils`
Various comments in neuropixels_utils.py regarding the number of ADCs on probes are wrong. Have not checked if code is also wrong.

alejoe91:
Fixes #3894

@billkarsh Do you mind taking a look? I believe that the main error and confusion point was in the use of `num_channels_per_adc` and its description

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/spikeglx.py
+++ b/src/spikeinterface/extractors/neoextractors/spikeglx.py
@@ -87,11 +87,13 @@ class SpikeGLXRecordingExtractor(NeoBaseRecordingExtractor):
                 self.set_probe(probe, in_place=True)
 
             # load num_channels_per_adc depending on probe type
-            ptype = probe.annotations["probe_type"]
+            model_name = probe.annotations.get("model_name", None)
+            if model_name is None:
+                model_name = probe.annotations["probe_name"]
 
-            if ptype in [21, 24]:  # NP2.0
+            if "2.0" in model_name:  # Neuropixels 2.0
                 num_channels_per_adc = 16
-                num_cycles_in_adc = 16  # TODO: Check this.
+                num_cycles_in_adc = 16
                 total_channels = 384
             else:  # NP1.0
                 num_channels_per_adc = 12
--- a/src/spikeinterface/extractors/neuropixels_utils.py
+++ b/src/spikeinterface/extractors/neuropixels_utils.py
@@ -5,7 +5,7 @@ from typing import Optional
 
 
 def get_neuropixels_sample_shifts(
-    num_channels: int = 384, num_channels_per_adc: int = 12, num_cycles: Optional[int] = None
+    num_channels: int = 384, num_adcs: int = 12, num_cycles: Optional[int] = None
 ) -> np.ndarray:
     """
     Calculate the relative sampling phase (inter-sample shifts) for each channel
@@ -26,16 +26,16 @@ def get_neuropixels_sample_shifts(
     num_channels : int, default: 384
         Total number of channels in the recording.
         Neuropixels probes typically have 384 channels.
-    num_channels_per_adc : int, default: 12
+    num_adcs : int, default: 12
         Number of channels assigned to each ADC on the probe.
         Neuropixels 1.0 probes have 12 ADCs, each handling 32 channels.
-        Neuropixels 2.0 probes have 16 ADCs.
+        Neuropixels 2.0 probes have 16 ADCs, each handling 24 channels.
     num_cycles : int or None, default: None
         Number of cycles in the ADC sampling sequence.
         Neuropixels 1.0 probes have 13 cycles for AP (action potential) signals
         and 12 for LFP (local field potential) signals.
         Neuropixels 2.0 probes have 16 cycles.
-        If None, defaults to the value of `num_channels_per_adc`.
+        If None, defaults to the value of `num_adcs`.
 
     Returns
     -------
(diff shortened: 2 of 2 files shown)


chrishalcrow:
Hello, is there some NeuroPixels documentation/manual which describes this stuff in more detail? Could we link to it?

alejoe91:
> Hello, is there some NeuroPixels documentation/manual which describes this stuff in more detail? Could we link to it?

I couldn't find any good reference. Maybe @jsiegle and @billkarsh could provide some resources we could point at for the estimation of the channel inter sample shifts?

billkarsh:
The operations are documented here: https://github.com/billkarsh/CatGT/blob/main/Build/Tool.cpp::FFT::init(), FFT::timeShiftChannel().

The mux tables are in the public probeTable repo: https://github.com/billkarsh/ProbeTable.

The description of the mux table is here: https://billkarsh.github.io/SpikeGLX/Sgl_help/Metadata_30.html#imec.

The mux table is organized as (nADC,nGrp)(grp)(grp)… Each (grp) are the (acquired) channel indices that are digitized together.

The shift amount delT(igrp) is the time difference between group-igrp and group-0. Group-0 does not need to be shifted.

delT(igrp) = igrp/(nGrp*sample_rate).

Note that for 2.0-like probes (that are full band only), nGrp should be used as is from the table. In these cases the multiplexer clock runs at rate nGrp*sample_rate. That is, nGrp sets of channels are digitized during each effective sample period (1/sample_rate).

However, for 1.0-like probes that have an AP and an LF band, the multiplexer runs at (nGrp+1)*ap_sample_rate. In all cases so far, nGrp=12. During each sample period, the multiplexer digitizes {nGrp=12 AP channel groups + 1 LF group}. The LF groups are the same groups as the AP groups in the table, however, it takes 12 AP sample periods to digitize all of the LF groups. That’s how the LF sample rate becomes exactly 1/12 of the AP sample rate.

Bill

billkarsh:
I also put these comments on github issue, and edited as follows:
However, for 1.0-like probes that have an AP and an LF band, the multiplexer runs at (nGrp+1) * ap_sample_rate. In all cases so far, nGrp=12. During each sample period, the multiplexer digitizes {nGrp=12 AP channel groups + 1 LF group}. The LF groups are the same groups as the AP groups in the table, however, it takes 12 AP sample periods to digitize all of the LF groups. That’s how the LF sample rate becomes exactly 1/12 of the AP sample rate.
So when doing a 1.0-like AP file, delT(igrp) = igrp/(13 * ap_sample_rate).
When doing an LF file, delT(igrp) = igrp/(12 * lf_sample_rate), as if this was single-band data.

jsiegle:
1.0 probes have 32 ADCs (each of which receives multiplexed input from 12 channels), while 2.0 probes have 24 ADCs (each of which receives multiplexed input from 16 channels). This is not explained anywhere in IMEC's public-facing documentation, as far as I'm aware.

The key piece of info for determining the sample shifts is knowing the order in which channels are sampled by a given ADC, rather than the total number of ADCs. And also keeping in mind that there are 13 cycles of digitization for the 1.0 probes (12 AP + 1 LFP), as Bill pointed out. 

We created an illustration for the Open Ephys GUI documentation that may be helpful. Here, simultaneously sampled channels have the same color:

Neuropixels Sampling Pattern

alejoe91:
Thanks @jsiegle and @billkarsh!

I reverted the naming to `num_channels_per_adc`, which is correctly set to 12/16 for NP1/2 probes. The number of cycles (13 for NP1 and 16 for NP2) is also correct (both for [Open Ephys](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/extractors/neoextractors/openephys.py#L242-L255) and [SpikeGLX](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/extractors/neoextractors/spikeglx.py#L92-L99)).

@billkarsh I think that the mux table is super clear, but it is not available for older SpikeGLX versions and for Open Ephys. I checked a couple of muxTables entries in our test files and they correspond to the schema that Josh provided, which also corresponds to the "logic" implemented in the function:
```
# these correspond to the ADC#
adc_indices = np.floor(np.arange(num_channels) / (num_channels_per_adc * 2)) * 2 + np.mod(np.arange(num_channels), 2)

for adc_index in adc_indices:
    sample_shifts[adc_indices == adc_index] = np.arange(num_channels_per_adc) / num_cycles
```
I think this is correct for standard configurations, since `np.nonzero(adc_indices == adc_index)` results in the same mux table entries as some test files (e.g, [this one](https://github.com/SpikeInterface/probeinterface/blob/main/tests/data/spikeglx/non_human_primate_long_staggered.imec0.ap.meta#L55)). I'm not sure though if this logic is robust against custom configurations by the user or when only a subset of electrodes is selected.

@billkarsh how is the mux table generated by SpikeGLX? Is it retrieved by the probe registers or is there some logic in SpikeGLX to compute them? For the latter case, we can port the logic here to make sure it works for different SpikeGLX versions and for Open Ephys.

On the probeinterface side, we should are planning to sync all the NP info with ProbeTable repo. @billkarsh would it be possible to add the `nGrp` (i.e., `num_channels_per_adc`) to each entry? So we can propagate it to the metadata and read it from spikeinterface for each probe without hardcoding it.

alejoe91:
Thanks @billkarsh 

If I understand correctly, then the mux tables are fixed and don't change if the user changes the selected electrodes.
Therefore:
- whatever the user chooses as configuration the at-most 384 channels will be sampled by the ADCs as in the `z_mux_tables`
- if a subset of channels are used, we can just slice the mux table values accordingly (e.g., if 100 channels are used, we'll take values until 100)

I think that then what I'll do is to refactor probeinterface to use this information and add the mux table index as an array annotation to the contact, so we can read it and compute the correct sample shift on the spikeinterface side.

Here is the issue on probeinterface: https://github.com/SpikeInterface/probeinterface/issues/341

samuelgarcia (approved):

PR #3897, opened 2025-05-02
Prepare 0.102.3

alejoe91:


zm711:
Or we need a PR that allows for parsing pre-post 0.14.1 neo args. To be honest I think we should probably raise our minimum version. Each release of neo is getting more and more fixes for numpy 2.0 value overflows in various readers. So I think I agree with @mahlzahn that it is probably safer right now for us to just ratchet up our min neo each release to account for these previously unknown overflows.

alejoe91:
> Or we need a PR that allows for parsing pre-post 0.14.1 neo args. To be honest I think we should probably raise our minimum version. Each release of neo is getting more and more fixes for numpy 2.0 value overflows in various readers. So I think I agree with @mahlzahn that it is probably safer right now for us to just ratchet up our min neo each release to account for these previously unknown overflows.

done!

samuelgarcia (approved):

samuelgarcia:
Perfect. Tests are passing.
I merge and release.

PR #3898, opened 2025-05-02
Remove classes from extractor and preprocessing `__init__`

zm711:
I went with the hard ones to start with hahah. I figured a lot of the other ones would be easier to do. Let me know what you think about this strategy.

For preprocessors this changes nothing-- it just ensures that the class and function are lined up and then only imports the functions. Example here:
image

For extractors this is much harder and makes one potentially controversial change. When we convert from class to function we were keeping the name of the class (in the python sense not in a class attribute sense, but to make this work we need to change the class name to be the `read_function` name. Since we want our users to only see the function I think this should be fine, but maybe someone is super against this. Or maybe we make the wrapper function a bit more complex like for the dict example that @chrishalcrow did? I think keeping the name is probably better for provenance actually.

 Small example here:
image


src/spikeinterface/extractors/tests/test_neoextractors.py
```diff
@@ -8,7 +8,41 @@
 import pytest
 
 from spikeinterface import get_global_dataset_folder
-from spikeinterface.extractors import *
+from spikeinterface.extractors.neoextractors import (
```
h-mayorquin, line 11: why don't we import this from extractor_list?

src/spikeinterface/extractors/extractorlist.py
```diff
@@ -17,19 +14,21 @@
...
-from .neoextractors import NeuroScopeSortingExtractor, MaxwellEventExtractor
+# keep for reference Currently pulling from neoextractor __init__
+# from .neoextractors import NeuroScopeSortingExtractor, MaxwellEventExtractor
 
 # NWB sorting/recording/event
 from .nwbextractors import (
     NwbRecordingExtractor,
     NwbSortingExtractor,
-    NwbTimeSeriesExtractor,
+    NwbTimeSeriesExtractor,  # @Heberto/Alessio what do we do with this
```
h-mayorquin, line 31: make a class `read_nwb_timeseries` or something like that? 

Or what is the question?

src/spikeinterface/extractors/extractorlist.py
```diff
@@ -77,70 +76,93 @@
...
+
+_snippets_extractor_full_dict = {
+    NpySnippetsExtractor: "read_npy_snippets",
+    WaveClusSnippetsExtractor: "read_waveclus_snippets",
+}
+
+#############################################################################################
+# Organize the possible extractors into an easy to use format
+
+recording_extractor_full_dict = {
```
h-mayorquin, line 137: Do we need those dicts?

Where are they used?

h-mayorquin (commented):
There are many things that I don't remember about the importing system so my comments are questions more than requests.

Suggestions:
I think if we want to make change easier for people that relied on the class names we can create something like the `extractor_list` but with a more explicit names

`extractor_classees`?

So instead of:

```python
from spikeinterface.extractors import AnExtractror`
```

they do:
```python
from spikeinterface.extractors.extractor_classes import AnExtractor
```

Seems like an easy change to make that would also allow us to keep the privilege naming that you guys want (the `read_extractor` functions) as the unique access at `spikeinterface.extractors`.

src/spikeinterface/extractors/extractorlist.py
zm711: I never use these functions so I just wasn't sure if I should make a `read_nwb_timeseries` so that answers that question. The other question was is that an event extractor or something else? and should I just list `read_nwb` for all of them or should I give each one separately?

src/spikeinterface/extractors/extractorlist.py
zm711: Alessio likes them to be accessible to users so they can see what the options are for file types. So if you do 
```python
>>> import spikeinterface.extractor as se
>>> se.recording_extractor_full_dict['intan']
'read_intan'
```
it will tell you the function to use.

Otherwise the dicts are necessary because the `__all__` feature of an init requires strings rather than function names and because our `read_xx` function inherits the `.__name__` of the extractor class we can't just pull the name like in preprocessors.

Did I answer your question?

zm711:
> There are many things that I don't remember about the importing system so my comments are questions more than requests.
> 
> Suggestions: I think if we want to make change easier for people that relied on the class names we can create something like the `extractor_list` but with a more explicit names
> 
> `extractor_classees`?
> 
> So instead of:
> 
> ```python
> from spikeinterface.extractors import AnExtractror`
> ```
> 
> they do:
> 
> ```python
> from spikeinterface.extractors.extractor_classes import AnExtractor
> ```
> 
> Seems like an easy change to make that would also allow us to keep the privilege naming that you guys want (the `read_extractor` functions) as the unique access at `spikeinterface.extractors`.

I like this. I would be happy to change the file name and switch to extractor classes to make this clear as long as @alejoe91 and @samuelgarcia are okay with that :)

src/spikeinterface/extractors/tests/test_neoextractors.py
h-mayorquin: I would say if that makes your life more complicated no.

I just think that there is a lot of lists/dicts to maintain on synch. 

But you probably have thought better about this.

src/spikeinterface/extractors/extractorlist.py
chrishalcrow: Should I not be using `read_nwb`?? It's what I use to test that our labs nwb files are working as expected.

zm711: sounds good from my end. I'll expose `read_nwb` but won't preference. :) Best of luck Chris :)

src/spikeinterface/extractors/extractorlist.py
zm711: Sorry I don't understand this comment. Current it is a dict:

```
{ 
'intan': 'read_intan',
 'blackrock': 'read_blackrock'
}
Where do you want to expose the classes instead? We prep the dicts so we have the class and the read_function sync'ed in the private dicts and then we expose a nice clean dict for users? Could you give a pseudo example of what you want?

File: src/spikeinterface/extractors/tests/test_neoextractors.py (hunk 1 above)
Comment (zm711) [reply src/spikeinterface/extractors/tests/test_neoextractors.py:11]: I could give it a quick try. It shouldn't be too difficult. The issue is it is very easy for everything to get exposed. Since we aren't adding `__all__` at the init level anything that gets to the init is exposed, but I think I can safely allow the neo classes into the extractorlist because we funnel out only the functions with the `__all__` there so that the init of extractors will only have the functions and classes we want to expose.

File: src/spikeinterface/extractors/extractorlist.py (hunk 3 above)
Comment (alejoe91) [reply src/spikeinterface/extractors/extractorlist.py:137]: Instead of the str as value, can we have the function itself?

I'd like to do something like this:
```
recording = recording_extractor_full_dict["intan"](**reader_keargs)
```

Comment (zm711) [reply src/spikeinterface/extractors/extractorlist.py:137]: Yes of course. but it will display like this:

```
>>> recording_extractor_full_dict['intan']
<function at x00047>
```

so it will be easy to use, but will not tell the user anything. We can either make it good for displaying info or have it return the function for use. Is the function what we were hoping for in this case?

Comment (alejoe91) [reply src/spikeinterface/extractors/extractorlist.py:137]: Sure, but I think that we have all readers in the documentation for the end user. This dictionaries were useful as a map between reader name and actual class/function :)

Comment (zm711) [reply src/spikeinterface/extractors/extractorlist.py:137]: It will be done in the next commit :)

File: src/spikeinterface/extractors/tests/test_neoextractors.py (hunk 1 above)
Comment (zm711) [reply src/spikeinterface/extractors/tests/test_neoextractors.py:11]: It's super easy to delete the `__all__` from neoextractors which should bring the classes and functions into the `extractorlist` so I just made this change in the last commit :)

File: src/spikeinterface/extractors/extractorlist.py (hunk 3 above)
Comment (zm711) [reply src/spikeinterface/extractors/extractorlist.py:141]: Done :) 
Tested locally and worked for me to input the extractor_kwargs in the dict with the key (it gave me an error since I didn't have an actual file, but the error looked appropriate for the specific extractor )
```
>>> recording = se.recording_extractor_full_dict['intan'](file_path='')
```
and when we just type it in without kwargs it gives
```
>>> se.recording_extractor_full_dict['intan']
<class 'spikeinterface.extractors.neoextractors.intan.IntanRecordingExtractor'>
```

File: src/spikeinterface/preprocessing/preprocessinglist.py [hunk 4]
```diff
@@ -45,41 +45,50 @@
...
+
+# pp_function.__name__ gives the name of the function that users should use
+__all__ = [pp_function.__name__ for pp_function in _all_preprocesser_dict.values()]
+__all__.extend(
+    [scale_to_uV.__name__, compute_whitening_matrix.__name__, train_deepinterpolation.__name__, causal_filter.__name__]
+)
 
-preprocesser_dict = {pp_class.name: pp_class for pp_class in preprocessers_full_list}
+preprocessor_dict = {
+    pp_class.__name__: pp_function.__name__ for pp_class, pp_function in _all_preprocesser_dict.items()
```
zm711, line 92:
```suggestion
    pp_class.__name__: pp_function for pp_class, pp_function in _all_preprocesser_dict.items()
```
return function rather than name of function.

src/spikeinterface/preprocessing/preprocessinglist.py
```diff
@@ -45,41 +45,48 @@
...
+
+# pp_function.__name__ gives the name of the function that users should use
+__all__ = [pp_function.__name__ for pp_function in _all_preprocesser_dict.values()]
+__all__.extend(
+    [scale_to_uV.__name__, compute_whitening_matrix.__name__, train_deepinterpolation.__name__, causal_filter.__name__]
+)
 
-preprocesser_dict = {pp_class.name: pp_class for pp_class in preprocessers_full_list}
+preprocessor_dict = {pp_class.__name__: pp_function for pp_class, pp_function in _all_preprocesser_dict.items()}
+__all__.append("preprocessor_dict")
```
chrishalcrow, line 92: Hello, I'd quite like the user to be able to get a list of available preprocessing steps, like we have for sorters: https://github.com/SpikeInterface/spikeinterface/blob/d142fe41d7b6177aea2591d68196b7cf4b6354a3/src/spikeinterface/sorters/sorterlist.py#L56 

So I'd like to expose a list, which is something like:
```
available_preprocessors = [foo.__name__ for foo in pp.preprocessor_dict.values()]
```

chrishalcrow: Yeah, let's think. Would be cool to have consistent logic and API for `available_preprocessors`, `available_sorters`, `available_extractors` etc! `available_sorters` is a function, but seems like it'd be easier for new users if it were just a list? @alejoe91 @samuelgarcia do you agree it'd be nice to have this? And in what form?

src/spikeinterface/extractors/tests/test_cbin_ibl_extractors.py
```diff
@@ -1,8 +1,7 @@
 import pytest
-import numpy as np
 import unittest
 
-from spikeinterface.extractors import CompressedBinaryIblExtractor, read_cbin_ibl
+from spikeinterface.extractors.extractorlist import CompressedBinaryIblExtractor, read_cbin_ibl
```
zm711, line 4:
```suggestion
from spikeinterface.extractors.extractor_classes import CompressedBinaryIblExtractor, read_cbin_ibl
```

src/spikeinterface/extractors/tests/test_iblextractors.py
```diff
@@ -6,7 +6,7 @@
 from numpy.testing import assert_array_equal
 import pytest
 
-from spikeinterface.extractors import read_ibl_recording, read_ibl_sorting, IblRecordingExtractor
+from spikeinterface.extractors.extractorlist import read_ibl_recording, read_ibl_sorting, IblRecordingExtractor
```
zm711, line 9:
```suggestion
from spikeinterface.extractors.extractor_classes import read_ibl_recording, read_ibl_sorting, IblRecordingExtractor
```

src/spikeinterface/extractors/tests/test_mdaextractors.py
```diff
@@ -2,7 +2,7 @@
 from pathlib import Path
 from spikeinterface.core.testing import check_recordings_equal, check_sortings_equal
 from spikeinterface.core import generate_ground_truth_recording
-from spikeinterface.extractors import MdaRecordingExtractor, MdaSortingExtractor
+from spikeinterface.extractors.extractorlist import MdaRecordingExtractor, MdaSortingExtractor
```
zm711, line 5:
```suggestion
from spikeinterface.extractors.extractor_classes import MdaRecordingExtractor, MdaSortingExtractor
```

src/spikeinterface/extractors/tests/test_nwbextractors.py
```diff
@@ -5,7 +5,7 @@
 import pytest
 import numpy as np
 
-from spikeinterface.extractors import NwbRecordingExtractor, NwbSortingExtractor, NwbTimeSeriesExtractor
+from spikeinterface.extractors.extractorlist import NwbRecordingExtractor, NwbSortingExtractor, NwbTimeSeriesExtractor
```
zm711, line 8:
```suggestion
from spikeinterface.extractors.extractor_classes import NwbRecordingExtractor, NwbSortingExtractor, NwbTimeSeriesExtractor
```

src/spikeinterface/extractors/tests/test_nwbextractors_streaming.py
```diff
@@ -7,7 +7,7 @@
 from spikeinterface import load
 from spikeinterface.core.testing import check_recordings_equal
 from spikeinterface.core.testing import check_recordings_equal, check_sortings_equal
-from spikeinterface.extractors import NwbRecordingExtractor, NwbSortingExtractor
+from spikeinterface.extractors.extractorlist import NwbRecordingExtractor, NwbSortingExtractor
```
zm711, line 10:
```suggestion
from spikeinterface.extractors.extractor_classes import NwbRecordingExtractor, NwbSortingExtractor
```

src/spikeinterface/extractors/tests/test_shybridextractors.py
```diff
@@ -2,7 +2,7 @@
 
 from spikeinterface.core import generate_ground_truth_recording
 from spikeinterface.core.testing import check_recordings_equal, check_sortings_equal
-from spikeinterface.extractors import SHYBRIDRecordingExtractor, SHYBRIDSortingExtractor
+from spikeinterface.extractors.extractorlist import SHYBRIDRecordingExtractor, SHYBRIDSortingExtractor
```
zm711, line 5:
```suggestion
from spikeinterface.extractors.extractor_classes import SHYBRIDRecordingExtractor, SHYBRIDSortingExtractor
```

src/spikeinterface/extractors/tests/test_whitematterrecordingextractor.py
```diff
@@ -3,7 +3,7 @@
 from pathlib import Path
 import pickle
 
-from spikeinterface.extractors import WhiteMatterRecordingExtractor, BinaryRecordingExtractor
+from spikeinterface.extractors.extractorlist import WhiteMatterRecordingExtractor, BinaryRecordingExtractor
```
zm711, line 6:
```suggestion
from spikeinterface.extractors.extractor_classes import WhiteMatterRecordingExtractor, BinaryRecordingExtractor
```

zm711 (commented):
missing extractor class change

src/spikeinterface/preprocessing/preprocessinglist.py
zm711: I'll resolve this for cleanup but I copied over to #3928 for a future PR.

zm711:
Everything is fixed except:

RTD is not building because how do we want to deal with `NpzSortingExtractor` and `BinaryRecordingExtractor`. I initially had just the `read_xx` function, but in our tutorials we use them to write so it feels weird to say

read_binary.write_recording(xx). So maybe we leave those as classes instead?

And second maybe we do a dep of the paths like this https://github.com/catalystneuro/neuroconv/issues/1346#issuecomment-2885087428

src/spikeinterface/extractors/extractor_classes.py
```diff
@@ -0,0 +1,172 @@
...
+# most important extractor are in spikeinterface.core
+from spikeinterface.core import (
+    BinaryFolderRecording,
+    BinaryRecordingExtractor,
+    NumpyRecording,
+    NpzSortingExtractor,
+    NumpySorting,
+    NpySnippetsExtractor,
+    ZarrRecordingExtractor,
+    ZarrSortingExtractor,
```
zm711, line 13: To be more explicit I'm wondering if we want to keep these as classes? since they are from `core`. We can still allow people to use the convenience functions, but having these be classes can signal they are for devs and for people that want just the read functionality they can use the `read_xx`. But not sure?

alejoe91: I would keep them since they are core extractors :)

zm711: Cool. I'll do that in the next commit. Heberto and I are meeting later to try to deprecate the class import for neuroconv. It seems like I have a pretty easy way to do it for the extractors so once he and I talk I'll do both those commits for review. Then I want everyone to test importing on their setups to make sure they like how docs and warnings appear :)

zm711:
So to summarize this PR (docs failing our now unrelated to the PR)

1) extractorlist/preprocessorlist -> extractor_classes/preprocessor_classes
2) only functions are imported rather than classes, except core classes which remain as classes
3) deprecation of classes such that a warning is given and where to find the classes is given then the function version is returned
    * We should remove this as it will be better when we can remove the dep in a version or two as this will speed up imports, allow better doc flow-through and allow us not to worry about an __all__ mechanism at the init level
4) docs and testing fixed for api change
5) sorting extractors fixed for api change

src/spikeinterface/extractors/neoextractors/__init__.py
```diff
@@ -42,42 +42,45 @@
...
-    OpenEphysLegacyRecordingExtractor,
-    PlexonRecordingExtractor,
-    Plexon2RecordingExtractor,
-    Spike2RecordingExtractor,
-    SpikeGadgetsRecordingExtractor,
-    SpikeGLXRecordingExtractor,
-    TdtRecordingExtractor,
-    NeuroExplorerRecordingExtractor,
-]
+neo_recording_extractors_dict = {
```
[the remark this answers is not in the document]
chrishalcrow: Oh wait, they are here, but don't show the deprecation message...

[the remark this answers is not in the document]
chrishalcrow: Ahhh, it's because e.g. `read_neuronexus` (which works) is defined from the class:

``` python
read_neuronexus = define_function_from_class(source_class=NeuroNexusRecordingExtractor, name="read_neuronexus")
```

but `read_openephys` and `read_mearec` are functions.

Messy solution for deprecation message, to be added when constructing `all_extractors`:

``` python
from .neoextractors import OpenEphysBinaryEventExtractor, OpenEphysBinaryRecordingExtractor, OpenEphysLegacyRecordingExtractor, MEArecSortingExtractor, MEArecRecordingExtractor, SpikeGLXEventExtractor
all_extractors += [OpenEphysBinaryEventExtractor, OpenEphysBinaryRecordingExtractor, OpenEphysLegacyRecordingExtractor, MEArecSortingExtractor, MEArecRecordingExtractor, SpikeGLXEventExtractor]
```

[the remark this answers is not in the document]
zm711: Great catch. Will fix :)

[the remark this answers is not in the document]
zm711: Done.

src/spikeinterface/extractors/__init__.py
```diff
@@ -30,6 +30,16 @@ def __getattr__(extractor_name):
     all_extractors += list(sorting_extractor_full_dict.values())
     all_extractors += list(event_extractor_full_dict.values())
     all_extractors += list(snippets_extractor_full_dict.values())
+    # special cases because they don't have simple wrappers
```
chrishalcrow, line 33: magnifico

src/spikeinterface/extractors/extractor_classes.py
```diff
@@ -0,0 +1,175 @@
...
+# events only from neo
+_event_extractor_full_dict = neo_event_extractors_dict
+
+_snippets_extractor_full_dict = {
+    NpySnippetsExtractor: {"read_npy_snippets": read_npy_snippets},
+    WaveClusSnippetsExtractor: {"read_waveclus_snippets": read_waveclus_snippets},
+}
+
+#############################################################################################
+# Organize the possible extractors into an easy to use format
```
h-mayorquin, line 140: easy-to-use is not very specific. Better to write what you are accomplish.

src/spikeinterface/extractors/extractor_classes.py
```diff
@@ -0,0 +1,175 @@
...
+# snippers
+from .waveclussnippetstextractors import WaveClusSnippetsExtractor, read_waveclus_snippets
+
+
+# misc
+from .alfsortingextractor import ALFSortingExtractor, read_alf_sorting
+
+
+###############################################################################################
+# first we line up each class with its function
```
h-mayorquin, line 80: Can we not call this function, I keep getting confused about it and I think other users that are not us will get confused. 

They are two uses for this:
1) You guys like snake_case naming scheme for this.
2) Wrap in magic like the Chris' trick for its dict preprocessing scheme

Can we call them wrappers? This way we don't say something that is false (calling this things functions) and we also leave more space to other magic that we want to add.

src/spikeinterface/extractors/extractor_classes.py
```diff
@@ -0,0 +1,175 @@
...
+# for this to work we need the actual function along with a string version of the function name
+# so we make a private nested dict that we can use to make the exposed dicts as well as load the
+# correct function into the extractors __init__
+# note that some formats (binary and numpy) still use the class format as they aren't read-only
+
```
h-mayorquin, line 85: I think that you should mention here the purpose of this dict. I think you have the name in strings name so you can add it in `__all__` dunder and this way you reduce the number of lines you write? And then we also have the wrapper object because that's what we want?

I also think that the nested dict is not a good data structure for this.

A dict with named tuples when you add specific name to each component would be a better fit here than a nested dict I feel. A dict that maps class to string name and another one from class to wrapper would be better.

h-mayorquin (commented):
I ran the tests suite on neuroconv and it is working. Let me know when it is ready so I can run them again.

src/spikeinterface/extractors/extractor_classes.py
zm711: Okay let's fix this once we address your other comment about structure.

src/spikeinterface/extractors/extractor_classes.py
zm711: I'm fine calling them wrappers.

src/spikeinterface/extractors/extractor_classes.py
zm711: Could you give an example with one extractor. Are you thinking like a structure array? I guess I don't know what the value of the named tuple would be? 

(I agree nested dict is not ideal, though), but what is the advantage you're hoping for? Speed, memory management or readability?

h-mayorquin: Readability. The purpose of this dicts and what each of their members is accomplishing is not clear. This can be solved with documentation above or with better naming in the data structure:

```python

WrapperInfo = namedtuple("WrapperInfo", ["wrapper_string", "wrapper_class"])

_recording_extractor_full_dict = {
    BinaryFolderRecording: WrapperInfo("BinaryFolderRecording", BinaryFolderRecording),
    SHYBRIDRecordingExtractor: WrapperInfo("read_shybrid_recording", read_shybrid_recording),
}
```

Or a data class which allows named initilization:

```
from dataclasses import dataclass

@dataclass(frozen=True)
class WrapperInfo:
    wrapper_string: str
    wrapper_class: object

_recording_extractor_full_dict = {
    BinaryFolderRecording: WrapperInfo(
        wrapper_string="BinaryFolderRecording",
        wrapper_class=BinaryFolderRecording
    ),
    SHYBRIDRecordingExtractor: WrapperInfo(
        wrapper_string="read_shybrid_recording",
        wrapper_class=read_shybrid_recording
    ),
}
```

Things that would help me tell why do we need this structure and what are its parts.

chrishalcrow: I love the NamedTuple option - gonna steal that for use elsewhere :)

zm711: Only issue is it can be used as a tuple. So if I do WrapperInfo[0] is will also return "BinaryFolderRecording". So sometimes people forget that. Whereas using the dict structure would prevent that since I can't do my_dict[0]. I have to use the key which in this case is "wrapper_string".

zm711: I've thought about it and I think I prefer the DataClass approach too. I don't use dataclasses that often. Is there any major speed hiccup from this? You'd know better @h-mayorquin in general. I don't want our imports to get ruined by doing it this way rather than dict methods.

src/spikeinterface/extractors/__init__.py
```diff
@@ -7,3 +7,58 @@
...
+from warnings import warn
+
+
+# deprecation of class import idea from neuroconv
+# this __getattr__ is only triggered if the normal lookup fails so import
+# any of our functions is fine but if someone tries to import a class this raises
+# the warning and then returns the "function" version which will look the same
+# to the end-user
+# to be removed after version 0.105.0
+def __getattr__(extractor_name):
```
samuelgarcia, line 20: Amazing, you are doing very sophisticated stuff. Maybe too much for me but I really admire.

src/spikeinterface/extractors/extractor_classes.py
samuelgarcia: Not sure to follow the debate here.
Is there cases where the str do not match the function name. I would have use only the function. I must miss something.

samuelgarcia:
Vous êtes tous incroyables!
Merci beaucoup pour ce grand nettoyage.
I totaly trust all of you on this.
You can merge when you want.

src/spikeinterface/extractors/extractor_classes.py
zm711: Hey @samuelgarcia, Heberto is saying that this nested dict is hard to read because it is not clear. I was being lazy so that I could just use dict methods, so the options are either use a better structure to be more readability or make my comment better to understand how the poorly designed structure works. I've just been to busy to do either at the moment. But will soon. 

We do have some naming issues. Our preprocessors change their names from the base classes because they now are able to do rec or dict so we don't want them to have the same name. Extractors are just wrappers over the classes so they have the same name. So we need slightly different logic for different modules.

src/spikeinterface/extractors/__init__.py
zm711: It was a pain to get working. But Heberto makes a great point that we don't want to pull the rug out from under our devs who build off of spikeinterface. So we warn them for a couple versions then wee can clean this all up.

src/spikeinterface/extractors/extractor_classes.py
zm711: @h-mayorquin want to take a quick look at the nested dict I made?

src/spikeinterface/extractors/extractor_classes.py
zm711: I changed this.

src/spikeinterface/extractors/extractor_classes.py
zm711: New comment. Let me know if you like it better!

src/spikeinterface/extractors/extractor_classes.py
```diff
@@ -0,0 +1,185 @@
...
+_snippets_extractor_full_dict = {
+    NpySnippetsExtractor: dict(wrapper_string="read_npy_snippets", wrapper_class=read_npy_snippets),
+    WaveClusSnippetsExtractor: dict(wrapper_string="read_waveclus_snippets", wrapper_class=read_waveclus_snippets),
+}
+
+############################################################################################################
+# Organize the possible extractors into a user facing format with keys being extractor names
+# (e.g. 'intan' , 'kilosort') and values being the appropriate Extractor class (e.g. IntanRecordingExtractor,
+# KiloSortingSortingExtractor)
```
zm711:
```suggestion
# KiloSortSortingExtractor)
```
this is what I get for typing too fast!

src/spikeinterface/extractors/neoextractors/__init__.py
```diff
@@ -42,42 +42,47 @@
...
+    AlphaOmegaRecordingExtractor: dict(wrapper_string="read_alphaomega", wrapper_class=read_alphaomega),
+    AxonaRecordingExtractor: dict(wrapper_string="read_axona", wrapper_class=read_axona),
+    BiocamRecordingExtractor: dict(wrapper_string="read_biocam", wrapper_class=read_biocam),
+    BlackrockRecordingExtractor: dict(wrapper_string="read_blackrock", wrapper_class=read_blackrock),
+    CedRecordingExtractor: dict(wrapper_string="read_ced", wrapper_class=read_ced),
+    EDFRecordingExtractor: dict(wrapper_string="read_edf", wrapper_class=read_edf),
+    IntanRecordingExtractor: dict(wrapper_string="read_intan", wrapper_class=read_intan),
+    MaxwellRecordingExtractor: dict(wrapper_string="read_maxwell", wrapper_class=read_maxwell),
+    MEArecRecordingExtractor: dict(wrapper_string="read_mearec", wrapper_class=read_mearec),
```
zm711, line 54: You good with MEArec being dealt with this way? @alejoe91

src/spikeinterface/extractors/extractor_classes.py
chrishalcrow: Yo, this looks good and works for me.

chrishalcrow:
Been using the preprocessing imports from this PR when developing #3438 and they work a treat. So all good from me :)

src/spikeinterface/extractors/neoextractors/__init__.py
alejoe91: So here the function actually returns both the recording and sorting, but we don't have separate functions for it so super fine with it!

zm711: Perfect.

zm711:
Last thing we need is for @h-mayorquin to run neuroconv testing suite one more time to make sure I didn't make a mistake when converting the nested dict. Then once testing is fixed I think I'm done with this!

alejoe91 (approved):

alejoe91:
LGTM! You can merge when you want @h-mayorquin @zm711

h-mayorquin:
OK, for me.

src/spikeinterface/extractors/extractor_classes.py
```diff
@@ -0,0 +1,185 @@
...
+from .waveclussnippetstextractors import WaveClusSnippetsExtractor, read_waveclus_snippets
+
+
+# misc
+from .alfsortingextractor import ALFSortingExtractor, read_alf_sorting
+
+
+###############################################################################################
+# first we line up each class with its wrapper that returns a snakecase version of the class
+# that Alessio and Zach like to call the function-version.
```
h-mayorquin: it looks like a duck but does not quack guys!

src/spikeinterface/extractors/extractor_classes.py
```diff
@@ -0,0 +1,185 @@
...
+    event_class.__name__.replace("Event", "").replace("Extractor", "").lower(): event_func["wrapper_class"]
+    for event_class, event_func in _event_extractor_full_dict.items()
+}
+snippets_extractor_full_dict = {
+    snippets_class.__name__.replace("Snippets", "").replace("Extractor", "").lower(): snippets_func["wrapper_class"]
+    for snippets_class, snippets_func in _snippets_extractor_full_dict.items()
+}
+
+
+# we only do the functions in the init rather than pull in the classes
```
h-mayorquin, line 187:
```suggestion
# we only do the functions in the init rather than pull in the classes. Importantly, `__all__` only works with strings.
```

src/spikeinterface/extractors/extractor_classes.py
```diff
@@ -0,0 +1,185 @@
...
+}
+
+############################################################################################################
+# Organize the possible extractors into a user facing format with keys being extractor names
+# (e.g. 'intan' , 'kilosort') and values being the appropriate Extractor class (e.g. IntanRecordingExtractor,
+# KiloSortSortingExtractor)
+# for example if we wanted to create a recording from an intan file we could do the following:
+# >>> recording = se.recording_extractor_full_dict['intan'](file_path='path/to/data.rhd')
+
+
```
h-mayorquin, line 168:
```suggestion
# Dictionary of format to wrapper/function class (e.g. "SpikeGLX": read_spikeglx)
```
is this correct?

src/spikeinterface/preprocessing/preprocessing_classes.py
```diff
@@ -45,41 +45,49 @@
 from .astype import AstypeRecording, astype
 from .unsigned_to_signed import UnsignedToSignedRecording, unsigned_to_signed
 
-
-preprocessers_full_list = [
+_all_preprocesser_dict = {
```
h-mayorquin, line 48:
```suggestion
_all_preprocessors_dict = {
```
?
actual uestion

src/spikeinterface/extractors/extractor_classes.py
zm711: almost: we use the lower string method so it would be
"spikeglx": read_spikeglx

src/spikeinterface/preprocessing/preprocessing_classes.py
zm711: It was a typo before but since I was switching it to a private structure I left the typo for the memories. If you look at the non-private version of the dict I correct the typo there.

zm711: If it bothers you I can fix it.

src/spikeinterface/extractors/extractor_classes.py
```diff
@@ -0,0 +1,185 @@
...
+###############################################################################################
+# first we line up each class with its wrapper that returns a snakecase version of the class
+# that Alessio and Zach like to call the function-version.
+# for this to work we need the actual wrapper class (ie the function) along with a string version of wrapper name
+# so we make a private nested dict that we can load the wrapper string into the __all__ attribute which means only the
+# correct "function"/wrapper class are loaded into the extractors __init__
+# note that some formats (binary and numpy) still use the class format as they aren't read-only (ie they have no wrapper)
+
```
h-mayorquin, line 100: The only thing missing for someone to understand the purpose is why we need the string wrapper? that is because `__all__` only takes a string.

There are three things that we need:
* A mapping from the original class to its wrapper (because that's what we want to expose)
* A mapping from the original class to its wrapper string (because of all)
* A mapping from format to the class wrapper for convenience.

You also wanted to structure all of this information so it can be modified in a single line. 

So we have to dictionaries, one that fulfills the first two purposes and then you do preprocessing to get another dict with the third. 

I am writing this for me, if you think some of these ideas can improve the explanation above feel free to take them.

src/spikeinterface/extractors/extractor_classes.py
zm711: happy to add once we settle the other questions.

src/spikeinterface/extractors/extractor_classes.py
zm711: If it hurts your soul then I can remove that :P

src/spikeinterface/preprocessing/preprocessing_classes.py
h-mayorquin: Na, no worries.

src/spikeinterface/extractors/extractor_classes.py
zm711: I could add/rewrite to include the bullet list:

>There are three things that we need:

>A mapping from the original class to its wrapper (because that's what we want to expose)
>A mapping from the original class to its wrapper string (because of all)
>A mapping from format to the class wrapper for convenience.

src/spikeinterface/extractors/extractor_classes.py
h-mayorquin: I think the comment accurately describes the situation and is good to have. 

Just note that wrapper is both for extractor as a processors as that is what we are actually doing. As in, you guys decide in the future that you might want to do more than just renaming to snake case like we do for preprocessors and it will still be a wrapper.

src/spikeinterface/extractors/extractor_classes.py
h-mayorquin: sounds good.

src/spikeinterface/extractors/__init__.py
```diff
@@ -1,7 +1,7 @@
 from .extractor_classes import *
 
-from .toy_example import toy_example
-from .bids import read_bids
+from .toy_example import toy_example as toy_example
```
h-mayorquin, line 3: eh?

zm711: You saw it. I was trying to sneak that in since it fixes linters....

src/spikeinterface/extractors/extractor_classes.py
h-mayorquin: Looks good to me. Thanks.

src/spikeinterface/extractors/__init__.py
h-mayorquin: Got you, that makes sense.

PR #3900, opened 2025-05-03
Fix ks4 tests and support ks4>=4.0.34

alejoe91:
- skip `"templates_from_data"` from testing (since it results in failures: https://github.com/MouseLand/Kilosort/issues/932)
- support KS API changes from 4.0.34

### DIFF
--- a/.github/scripts/test_kilosort4_ci.py
+++ b/.github/scripts/test_kilosort4_ci.py
@@ -110,7 +110,7 @@ if parse(kilosort.__version__) >= parse("4.0.24"):
 
 if parse(kilosort.__version__) >= parse("4.0.33"):
     PARAMS_TO_TEST_DICT.update({"cluster_neighbors": 11})
-    PARAMETERS_NOT_AFFECTING_RESULTS.append("position_limit")
+    PARAMETERS_NOT_AFFECTING_RESULTS.append("cluster_neighbors")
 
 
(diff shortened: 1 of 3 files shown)


.github/scripts/test_kilosort4_ci.py
```diff
@@ -533,33 +539,60 @@ def test_kilosort4_skip_preprocessing_correction(self, tmp_path, monkeypatch, pa
...
+                if self.hp_filter is not None:
+                    pass
+
+                if self.artifact_threshold < np.inf:
+                    if torch.any(torch.abs(X) >= self.artifact_threshold):
+                        return torch.zeros_like(X)
+
+                if self.whiten_mat is not None:
+                    pass
+                return X
```
zm711, line 595: This is brutal..... Sorry.

zm711 (approved):
Looks fine to me and tests are passing :)

I only wish we could find a way to help deal with the changing number of arguments. Your idea of setting a kwargs_dict seems cool, I wonder if we can apply that in more places such that the dict changes with the versions so when we unpack it, it returns the correct things. Maybe for a future PR we think about that :)

.github/scripts/test_kilosort4_ci.py
alejoe91: Hopefully this doesn't happen too soon! Anyways, kudos to @JoeZiminski since the tests to check the arguments of different KS functions are very useful to understand what's changed!

PR #3904, opened 2025-05-06
Add Literals to lots of function arguments

chrishalcrow:
This PR improves the user experience by adding Literals to LOTS of function arguments (eventually...). They are in a few function arguments at the moment.

Allows for auto-complete for string arguments, like so:

Screenshot 2025-05-06 at 12 38 11

Plan is to go through each module and add `Literal`s everywhere. Will be _especially_ helpful for `compute`! Before I do so, thought I'd use one concrete example to generate some discussion. I've added Literals for the `method` argument of `detect_bad_channels`. To do so, I've done three things:

1. Added the Literal to the typing in the argument list
2. Moved the list of allowed arguments outside the function, just above it. So that we can...
3. Add a test to compare the Literal to the list of allowed arguments

As far I can figure out, Literals have to be hardcoded for static type checkers to pick up on them. To check this hardcoding, I think it's worth adding tests like this for any arguments that might be extended in the future. Any other opinions?

EDIT: The steps 1. 2. 3. might be overkill?? Should I just do 1.?

### DIFF
--- a/src/spikeinterface/preprocessing/tests/test_detect_bad_channels.py
+++ b/src/spikeinterface/preprocessing/tests/test_detect_bad_channels.py
@@ -8,7 +8,7 @@ from spikeinterface.core import generate_recording
 from spikeinterface.preprocessing import detect_bad_channels, highpass_filter
 from spikeinterface.preprocessing.detect_bad_channels import detect_bad_channels_method_names
 
-from typing import get_type_hints, get_args
+from typing import get_args, Literal
 
 try:
     # WARNING : this is not this package https://pypi.org/project/neurodsp/
@@ -26,7 +26,7 @@ def test_literal_hardcoded_values():
     The possible strings allowed by the `method` argument are hardcoded. Here we check they are consistent
     with the methods list.
     """
-    detect_bad_channels_method_literals = get_args(get_type_hints(detect_bad_channels)["method"])
+    detect_bad_channels_method_literals = get_args(eval(detect_bad_channels.__annotations__["method"]))
     assert set(detect_bad_channels_method_literals) == set(detect_bad_channels_method_names)
 
 
--- a/src/spikeinterface/preprocessing/tests/test_detect_bad_channels.py
+++ b/src/spikeinterface/preprocessing/tests/test_detect_bad_channels.py
@@ -6,6 +6,9 @@ from probeinterface import generate_linear_probe
 
 from spikeinterface.core import generate_recording
 from spikeinterface.preprocessing import detect_bad_channels, highpass_filter
+from spikeinterface.preprocessing.detect_bad_channels import detect_bad_channels_method_names
+
+from typing import get_type_hints, get_args
 
 try:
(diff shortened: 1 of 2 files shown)


zm711:
I think you need a from future import annotations for 3.9 :)

chrishalcrow:
The "|" notation for types was only added in 3.10 (https://peps.python.org/pep-0604/)!!

So the following code errors on 3.9:
``` python
from typing import get_type_hints, get_args
get_type_hints(si.detect_bad_channels)
```

Does annotations help somehow here, @zm711 ??

This notation is used _everywhere_ in the codebase and 3.14 comes out in 5 months, so I'd vote to skip these tests on python < 3.10?

zm711:
Yes that's the reason we use that import statement at the top. It allows us to bring in `|` into 3.9 :)

zm711:
Just to be clear it is
```python
```

chrishalcrow:
Hello, thanks! Turns out `get_type_hints` still doesn't work with `annotations from __future__`. Instead you need to use `__annotations__` to get the types, and un-stringify the output (more: https://blog.derlin.ch/python-type-hints-and-future-annotations). This requires going from

``` python
detect_bad_channels_method_literals = get_args(get_type_hints(detect_bad_channels)["method"])
```

to

``` python
from typing import Literal
detect_bad_channels_method_literals = get_args(eval(detect_bad_channels.__annotations__['method']))
```
(the Literal import is needed for the `eval`)

src/spikeinterface/preprocessing/tests/test_detect_bad_channels.py
```diff
@@ -18,6 +21,15 @@
     HAVE_NPIX = False
 
 
+def test_literal_hardcoded_values():
```
zm711, line 24: I guess if we think about it what is the value of this test? This is basically saying that you don't trust the type hints of python and want to confirm they exist? We wouldn't add this test to every file right? 

What is it that you want to test?

chrishalcrow: I want to protect against a developer adding a new method and forgetting to add it to the typing.

zm711: I see so this is just a change detector to annoy Sam :P

zm711: So after carefully reading your PR discussion + your edits I think this:

> EDIT: The steps 1. 2. 3. might be overkill?? Should I just do 1.?

I think we start with just 1 and then if we decide to add a test later we do that. I'm actually fine with 2 as well because it allows us to easily add in a useful error message (ie you have entered x but only [y,z,a,b] are allowed). But 2 just adds so much work and you're right that if we do 2 then if we want to ensure it is enforced we have to do 3, but then that will cause test failures in refactor potentially. So for me 2 and 3 should really be a 1.0 type move when we are saying the API is 100% stable.

PR #3905, opened 2025-05-06
Refactor `correct_motion` into compute motion, then make interpolation

chrishalcrow:
PR to re-factor the `correct_motion` function into two parts: `compute_motion` (which returns _just_ a `motion_info` object) and `interpolate_motion`.

We expose the `compute_motion` function and edit the docs to use it. This encourages users to do the motion correction computation and check their results before interpolation (as Sam always tells us to do!). At the moment, to do this, the user has to do
``` python
_, motion, motion_info = si.correct_motion(recording=rec, preset="dredge_fast", output_motion=True, output_motion_info=True)
```
Then analyse the results. Now they can do
``` python
motion_info = si.compute_motion(recording=rec, preset="dredge_fast")
```
Which hopefully feels nicer.

Alessio also requested a `raise_error` argument, so that if the `estimate_motion` fails, the peaks and peak locations are still returned. Motion estimation often fails on e.g. a single shank-bank of a NP probe. But it can still be helpful to do a raster plot of the peaks. Hence this seems helpful. Note: it returns `motion_info['motion'] = None` at the moment.

I've done some little refactors, and updated some things to improve provenance. These are also in preparation for adding a new class `ComputeMotionAndInterpolateRecording` soon.

The `correct_motion` function should work exactly as before.


src/spikeinterface/preprocessing/motion.py
```diff
@@ -384,20 +372,30 @@ def correct_motion(
...
+
+    if no_selection_kwargs:
         # maybe do this directly in the folder when not None, but might be slow on external storage
         gather_mode = "memory"
         # node detect
-        method = detect_kwargs.pop("method", "locally_exclusive")
-        method_class = detect_peak_methods[method]
-        node0 = method_class(recording, **detect_kwargs)
+        detect_peaks_method = detect_kwargs["method"]
+        method_class = detect_peak_methods[detect_peaks_method]
```
chrishalcrow, line 382: The previous code was `pop`ing some kwargs from the dict, which `pop`'d them out of the parameters dict, losing provenance.

src/spikeinterface/preprocessing/motion.py
```diff
@@ -384,20 +372,30 @@ def correct_motion(
...
+        detect_kwargs_without_method = {
+            key: detect_kwarg for key, detect_kwarg in detect_kwargs.items() if key != "method"
```
chrishalcrow, line 384: The previous code was `pop`ing some kwargs from the dict, which `pop`'d them out of the parameters dict, losing provenance. We now make a `detect_kwargs_without_method` so that we don't edit the `detect_kwargs` dict,

src/spikeinterface/preprocessing/motion.py
```diff
@@ -435,13 +431,20 @@ def correct_motion(
...
-    motion = estimate_motion(recording, peaks, peak_locations, progress_bar=progress_bar, **estimate_motion_kwargs)
+    try:
+        motion = estimate_motion(recording, peaks, peak_locations, progress_bar=progress_bar, **estimate_motion_kwargs)
+    except Exception as err:
+        error_message = f"Motion estimation failed. Error given: {err}."
+        if raise_error:
+            raise RuntimeError(error_message)
+        else:
+            warnings.warn(error_message)
+        motion = None
```
chrishalcrow, line 442: If `raise_error` is set to `False` and the `estimate_motion` function fails, the function still returns `motion_info`, just with `motion = None`.

src/spikeinterface/preprocessing/tests/test_motion.py
```diff
@@ -45,7 +46,21 @@ def test_get_motion_parameters_preset():
     pprint(params)
 
 
+def test_estimate_motion_fails():
```
alejoe91, line 49:
```suggestion
def test_compute_motion_fails():
```

examples/how_to/handle_drift.py
```diff
@@ -176,6 +176,21 @@ def preprocess_chain(rec):
...
+# Suppose we decide to use the `nonrigid_accurate` preset to make the interpolated recording. We do this as follows
+
+from spikeinterface.sortingcomponents.motion import interpolate_motion
+preset = "nonrigid_accurate"
+folder = base_folder / "motion_folder_dataset1" / preset
+motion_info = si.load_motion_info(folder)
+motion = motion_info['motion']
+interpolated_recording = interpolate_motion(recording=rec, motion=motion)
+
+# You can then use the interpolated recording for e.g. spike sorting.
```
alejoe91, line 192: We should also add that one can use `correct_motion` to combine compute and interpolate

src/spikeinterface/preprocessing/motion.py
```diff
@@ -293,86 +319,48 @@ def correct_motion(
...
     """
+
     # local import are important because "sortingcomponents" is not important by default
     from spikeinterface.sortingcomponents.peak_detection import detect_peaks, detect_peak_methods
     from spikeinterface.sortingcomponents.peak_selection import select_peaks
     from spikeinterface.sortingcomponents.peak_localization import localize_peaks, localize_peak_methods
-    from spikeinterface.sortingcomponents.motion import estimate_motion, InterpolateMotionRecording
     from spikeinterface.core.node_pipeline import ExtractDenseWaveforms, run_node_pipeline
+    from spikeinterface.sortingcomponents.motion import estimate_motion
+    from spikeinterface.sortingcomponents.motion.motion_estimation import estimate_motion_methods
```
alejoe91, line 341:
```suggestion
    from spikeinterface.sortingcomponents.motion.motion_estimation import estimate_motion, estimate_motion_methods
```
Since you are already importing from the submodule, I think this is cleaner

src/spikeinterface/preprocessing/motion.py
alejoe91: This was my wish! Thank you! :)

src/spikeinterface/preprocessing/motion.py
```diff
@@ -453,6 +456,108 @@ def correct_motion(
...
+        detect_kwargs=detect_kwargs,
+        select_kwargs=select_kwargs,
+        localize_peaks_kwargs=localize_peaks_kwargs,
+        estimate_motion_kwargs=estimate_motion_kwargs,
+        **job_kwargs,
+    )
+
+    motion = motion_info["motion"]
+
+    recording_corrected = interpolate_motion(recording, motion, **interpolate_motion_kwargs)
```
alejoe91, line 559: So clean!

alejoe91 (changes requested):
Just some tiny comments! Looks super :)

Have you tried medicine in the handle drift tutorial?

examples/how_to/handle_drift.py
chrishalcrow: Done

chrishalcrow:
Hello, I'm trying `medicine` now - might be easier to leave it for another PR (I'll working more on this very soon!)

src/spikeinterface/preprocessing/motion.py
alejoe91: I think that the save_motion_info / and load_motion_info should also be updated to handle motion=None. Maybe worth adding a test with a specified folder?

chrishalcrow: Done - now if `motion = None`, `save_motion_info` makes a folder with no `motion` subfolder. When `load_motion_info` finds this folder structure, it sets `motion` to `None`.

src/spikeinterface/preprocessing/motion.py
```diff
@@ -250,41 +261,56 @@ def get_motion_parameters_preset(preset):
...
+
+def _update_interpolation_kwargs(preset, interpolation_kwargs):
+
+    params = motion_options_preset[preset]
+    interpolation_kwargs = dict(params["interpolate_motion_kwargs"], **interpolation_kwargs)
+
+    return interpolation_kwargs
+
+
+def compute_motion(
```
alejoe91, line 283: suggestion from Sam:
this function should only return `Motion` and have an optional `output_motion_info` argument that optionally returns `motion, motion_info`. The save mechanism etc will not need any change.

chrishalcrow: done

alejoe91 (approved):

src/spikeinterface/preprocessing/motion.py
```diff
@@ -453,6 +456,106 @@ def correct_motion(
...
+        overwrite=overwrite,
+        detect_kwargs=detect_kwargs,
+        select_kwargs=select_kwargs,
+        localize_peaks_kwargs=localize_peaks_kwargs,
+        estimate_motion_kwargs=estimate_motion_kwargs,
+        output_motion_info=True,
+        **job_kwargs,
+    )
+
+    recording_corrected = interpolate_motion(recording, motion, **interpolate_motion_kwargs)
```
alejoe91, line 557: One very last comment @chrishalcrow 

In case folder is given, we should add the `interpolate_motion_kwargs` to the `parameters.json`, right?

chrishalcrow: I don't think so. The compute motion is independent of the interpolation, and the interpolation kwargs are already propagated to the `recording_corrected._kwargs`.

alejoe91: You're absolutely right!

PR #3915, opened 2025-05-12
Expose `match_score` to benchmark compute_results

alejoe91:
By default, match_score is 0.5, which means that all performance metrics for units with agreement below that are 0. This PR at least exposes the option to pass match_score

### DIFF
--- a/src/spikeinterface/benchmark/benchmark_matching.py
+++ b/src/spikeinterface/benchmark/benchmark_matching.py
@@ -40,7 +40,11 @@ class MatchingBenchmark(Benchmark):
 
     def compute_result(self, with_collision=False, **result_params):
         sorting = self.result["sorting"]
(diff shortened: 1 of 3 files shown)


src/spikeinterface/benchmark/benchmark_matching.py
```diff
@@ -40,7 +40,11 @@ def run(self, **job_kwargs):
 
     def compute_result(self, with_collision=False, **result_params):
```
samuelgarcia:
```suggestion
    def compute_result(self, with_collision=False, match_score=0.5,**result_params):
```

samuelgarcia:
Ciao @alejoe91. I would be more explicit and put the variable in the signature.

src/spikeinterface/benchmark/benchmark_matching.py
alejoe91: @samuelgarcia should we expose the exhaustive_gt parameter then as well? I think they are at the same level

Issue #3920, opened 2025-05-14
documentation minor issues


pas-calc:
On the API documentation

- `compute_spike_amplitudes` at 
https://spikeinterface.readthedocs.io/en/stable/api.html#spikeinterface.postprocessing.compute_spike_amplitudes
has "Localize spikes in 2D or 3D with several methods given the template." but this is for `compute_spike_locations`.

- `read_kilosort` at
https://spikeinterface.readthedocs.io/en/stable/api.html#spikeinterface.extractors.read_kilosort
has "Path to the output Phy folder (containing the params.py)." but this is for `read_phy`  (however they share the same parent class)

zm711:
> compute_spike_amplitudes at
https://spikeinterface.readthedocs.io/en/stable/api.html#spikeinterface.postprocessing.compute_spike_amplitudes

Thanks yeah that was probably someone doing a bad copy :) 

> read_kilosort at
https://spikeinterface.readthedocs.io/en/stable/api.html#spikeinterface.extractors.read_kilosort
has "Path to the output Phy folder (containing the params.py)." but this is for read_phy (however they share the same parent class)

we tend to call a phy folder that set of files that ks makes that phy needs to read (it is also the set of files we make with our export to phy function). So this was intentional, but if you think it is unclear we could definitely do the Phy/Kilosort folder). Since the two projects were originally written in the same lab (but have diverged a bit) we just call the global format phy.

Any interest in submitting a PR for those fixes? Contributions are definitely welcome!

zm711:
I'll close as you've fixed both of these. Thanks again for the contributions!

PR #3924, opened 2025-05-15
phykilosortextractors.py Phy/Kilosort

pas-calc:
few corrections/suggestions
https://github.com/SpikeInterface/spikeinterface/issues/3920


zm711 (approved):
This is fine by me unless @alejoe91 would prefer to de-emphasize the KS nature of this and still focus on the Phy side only.

alejoe91 (approved):

PR #3925, opened 2025-05-15
spike_amplitudes description

pas-calc:
documentation
https://github.com/SpikeInterface/spikeinterface/issues/3920

### DIFF
--- a/src/spikeinterface/postprocessing/spike_amplitudes.py
+++ b/src/spikeinterface/postprocessing/spike_amplitudes.py
@@ -17,38 +17,19 @@ class ComputeSpikeAmplitudes(AnalyzerExtension):
     Computes the spike amplitudes.
 
     Needs "templates" to be computed first.
-    Localize spikes in 2D or 3D with several methods given the template.
+    Spike amplitudes from templates on peak channel for every spike in spike train.
 
     Parameters
     ----------
     sorting_analyzer : SortingAnalyzer
         A SortingAnalyzer object
-    ms_before : float, default: 0.5
-        The left window, before a peak, in milliseconds
-    ms_after : float, default: 0.5
-        The right window, after a peak, in milliseconds
-    spike_retriver_kwargs : dict
-        A dictionary to control the behavior for getting the maximum channel for each spike
-        This dictionary contains:
-          * channel_from_template: bool, default: True
-              For each spike is the maximum channel computed from template or re estimated at every spikes
-              channel_from_template = True is old behavior but less acurate
-              channel_from_template = False is slower but more accurate
-          * radius_um: float, default: 50
-              In case channel_from_template=False, this is the radius to get the true peak
-          * peak_sign, default: "neg"
-              In case channel_from_template=False, this is the peak sign.
-    method : "center_of_mass" | "monopolar_triangulation" | "grid_convolution", default: "center_of_mass"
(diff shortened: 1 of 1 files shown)


src/spikeinterface/postprocessing/spike_amplitudes.py
```diff
@@ -17,38 +17,19 @@ class ComputeSpikeAmplitudes(AnalyzerExtension):
...
-          * peak_sign, default: "neg"
-              In case channel_from_template=False, this is the peak sign.
-    method : "center_of_mass" | "monopolar_triangulation" | "grid_convolution", default: "center_of_mass"
-        The localization method to use
-    **method_kwargs : dict, default: {}
-        Kwargs which are passed to the method function. These can be found in the docstrings of `compute_center_of_mass`, `compute_grid_convolution` and `compute_monopolar_triangulation`.
-    outputs : "numpy" | "by_unit", default: "numpy"
-        The output format, either concatenated as numpy array or separated on a per unit basis
+    peak_sign : "pos" or "neg"
+        to get the extremum_channels_indices
```
zm711, line 27: Did you copy this from somewhere? :) We should rewrite that. The peak is whether you are looking for negative spikes (standard for extracellular potentials) or positive spikes (true for intracellular but a little contentious for extracellular experiments--I fall in that camp that things these are axonal).

zm711 (commented):
Thanks for the PR. Just one tweak on my end.

src/spikeinterface/postprocessing/spike_amplitudes.py
zm711: yeah it's just the `_set_params`.

I was saying you did it right, but we should rewrite the argument to explain it better :) The end-user might be confused how a peak_sign gets the `extremum_channels_indices` so instead we should explain what "peak" means. The internal details are less important vs the overall point of the argument. 

Our documentation in general is written with the devs in mind, but we are slowly converting it to user facing docs as the project is now being used by a continuum of users from beginner to advanced. Does that make sense?

pas-calc: Yes I see, and I appreciate this to make the docs more understandable for a broad audience. The technical details however are also important. For example it was not clear for me whether spike amplitudes can be also computed by all extremum modes/peak_sign as it is the case for waveform/templates? That's why I looked up compute_spike_amplitudes in the docs.

alejoe91, line 27:
```suggestion
    peak_sign : "neg" | "pos" | "both", default: "neg"
        Sign of the template to compute extremum channel used to retrieve spike amplitudes.
```

src/spikeinterface/postprocessing/spike_amplitudes.py
```diff
@@ -17,38 +17,19 @@ class ComputeSpikeAmplitudes(AnalyzerExtension):
     Computes the spike amplitudes.
 
     Needs "templates" to be computed first.
-    Localize spikes in 2D or 3D with several methods given the template.
+    Spike amplitudes from templates on peak channel for every spike in spike train.
```
alejoe91, line 20:
```suggestion
    Computes ppike amplitudes from the template's peak channel for every spike.
```

alejoe91 (approved):

PR #3931, opened 2025-05-16
Read spikeglx neuropixels data without ap meta

### LINKED ISSUE #3921: Load SpikeGLX lf data when meta for ap is not available
This comes from a [discussion](https://github.com/catalystneuro/neuroconv/issues/1340) with @rly.

Currently, when loading LF, we load the probe from the AP meta:

https://github.com/SpikeInterface/spikeinterface/blob/e45a8be7fd2f65546d8dabd89e5a42fec35e3147/src/spikeinterface/extractors/neoextractors/spikeglx.py#L76-L83

The reason is that the shank map is only available in the AP meta, and that is needed to load the probe:

https://github.com/SpikeInterface/spikeinterface/pull/596#issuecomment-1130194250

I propose that if the AP meta is not present for any reason, we load the data without the probe so users can at least access it. I think this might prevent loading the inter_sample_shift property?

But maybe with all the new metadata changes in Neuropixels, we can retrieve the probe information and the inter-sample shift in a simpler way.

Thoughts?

h-mayorquin:
Should fix #3921

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/spikeglx.py
+++ b/src/spikeinterface/extractors/neoextractors/spikeglx.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from pathlib import Path
+import warnings
 
 import probeinterface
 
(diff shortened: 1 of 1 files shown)


alejoe91:
@h-mayorquin so in case the ap meta is missing the is no probe info, right?

h-mayorquin:
Yes, so you only read the data but you won't have groups and locations.

PR #3933, opened 2025-05-16
More informative errors in frame slicing

h-mayorquin:
This comes from here:

https://github.com/catalystneuro/neuroconv/issues/1355

where a users has a slicing error and I don't have enough info to debug it.

### DIFF
--- a/src/spikeinterface/core/frameslicerecording.py
+++ b/src/spikeinterface/core/frameslicerecording.py
@@ -40,9 +40,9 @@ class FrameSliceRecording(BaseRecording):
         else:
             assert (
                 0 < end_frame <= parent_size
-            ), f"'end_frame' must be fewer than number of samples in parent: {parent_size}"
+            ), f"'{end_frame=}' must be fewer than number of samples in parent: {parent_size=}"
 
-        assert end_frame > start_frame, "'start_frame' must be smaller than 'end_frame'!"
+        assert end_frame > start_frame, f"{start_frame=} must be smaller than 'end_frame' {end_frame=}!"
 
         BaseRecording.__init__(
             self,
--- a/src/spikeinterface/core/frameslicerecording.py
+++ b/src/spikeinterface/core/frameslicerecording.py
@@ -30,20 +30,15 @@ class FrameSliceRecording(BaseRecording):
         num_segments = parent_recording.get_num_segments()
         assert num_segments == 1, f"FrameSliceRecording only works with one segment but found {num_segments}"
 
-        parent_size = parent_recording.get_num_samples(segment_index=0)
-        if start_frame is None:
-            start_frame = 0
-        else:
-            assert 0 <= start_frame < parent_size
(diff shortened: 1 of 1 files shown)


h-mayorquin:
Oh, the rate limit.

src/spikeinterface/core/frameslicerecording.py
```diff
@@ -40,9 +40,9 @@ def __init__(self, parent_recording, start_frame=None, end_frame=None):
         else:
             assert (
                 0 < end_frame <= parent_size
-            ), f"'end_frame' must be fewer than number of samples in parent: {parent_size}"
+            ), f"'{end_frame=}' must be fewer than number of samples in parent: {parent_size=}"
```
zm711, line 43: Does the user know what `parent_size` is? or is that an internal variable to represent the size of the parent. I'm wondering if this is better as {parent_size} instead. The rest looks good to me.

h-mayorquin: If you think that's better I am OK with it. I think that parent size is a bad variable name overall.

zm711: It's a super minor point. I don't see the value of saying 

```
>>> 
... samples in parent: parent_size=xx
```
instead of
```
>>> samples in parent: xx
```

But I agree parent size is not super clear to the end-user as a variable name, which is why I don't think we even show that variable name in the error.

h-mayorquin: Changed. Let me know what you think.

src/spikeinterface/core/frameslicerecording.py
```diff
@@ -27,22 +27,18 @@ class FrameSliceRecording(BaseRecording):
...
-        else:
-            assert 0 <= start_frame < parent_size
-
-        if end_frame is None:
-            end_frame = parent_size
-        else:
-            assert (
-                0 < end_frame <= parent_size
-            ), f"'end_frame' must be fewer than number of samples in parent: {parent_size}"
+        samples_in_sliced_recording = parent_recording.get_num_samples(segment_index=0)
```
zm711, line 33: I'm not sure? When I first read this I thought it was referring to the new child sliced recording rather than the parent. "sliced" can semantically be either child or parent. To be clear you would have to say `samples_in_the_recording_to_be_sliced`, but that is even worse. what about `number_samples_in_recording` or `samples_in_recording`? Because the user will assume the current recording? or you could say `samples_in_current_recording` to more emphasize it is the current recording? 

How were you understanding `samples_in_sliced_recording`?

h-mayorquin: samples in recording is fine for me if you think that's clear.

zm711 (commented):
I'm good with this. Did you want @alejoe91 to take a look too?

samuelgarcia:
Merci beaucoup mon ami.
Parafit pour moi.
@alejoe91 do you want to check ?

samuelgarcia (approved):

alejoe91 (approved):

PR #3939, opened 2025-05-20
Fix recording slicing docstrings and documentation typo

### LINKED ISSUE #3938: Confusing default behavior for `time_slice` docstring
The way that this docstring is phrased makes it appear that the argument can be omitted but instead it should say that if the start time is `None`
https://github.com/SpikeInterface/spikeinterface/blob/03c818c62548a3b489e6a668ab5bee4260e2b1c1/src/spikeinterface/core/baserecording.py#L774

kushaangupta:
Closes #3938

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -749,9 +749,9 @@ class BaseRecording(BaseRecordingSnippets):
         Parameters
         ----------
         start_frame : int, optional
-            The start frame, if not provided it is set to 0
+            Start frame index. If None, defaults to the beginning of the recording (frame 0).
         end_frame : int, optional
-            The end frame, it not provided it is set to the total number of samples
+            End frame index. If None, defaults to the last frame of the recording.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/baserecording.py
```diff
@@ -771,9 +771,9 @@ def time_slice(self, start_time: float | None, end_time: float | None) -> BaseRe
         Parameters
         ----------
         start_time : float, optional
-            The start time in seconds. If not provided it is set to 0.
+            Start time in seconds. If None, defaults to the beginning of the recording.
```
h-mayorquin, line 774: This an important change. 0 is in fact wrong.

Thanks for the fix.

h-mayorquin (approved):
Thanks for your contribution. This is more explicit and fixes a place where the documentation was wrong. 

As mentioned in the issue I think we should set default values None but this should be done in a different PR.

zm711 (approved):
Thanks! And thanks for checking on the actual behavior!

I thought the idea with no default was that we force the user to make a choice rather than silently do anything. So I agree better for a different PR where we can discuss what we think default behavior should be.

PR #3852, opened 2025-04-09
Updated KilosortBase's channel map generation to support json probe maps with kcoord information.

nikhilchandra:
Fixes #3704

PR #3944, opened 2025-05-21
Fix kcoords for non-int groups

alejoe91:
Introduced by #3852, if the channel groups are not strings they cannot be cast to float. This makes a map to integers

### DIFF
--- a/src/spikeinterface/sorters/external/kilosort4.py
+++ b/src/spikeinterface/sorters/external/kilosort4.py
@@ -440,7 +440,7 @@ class Kilosort4Sorter(BaseSorter):
         yc = positions[:, 1]
         unique_groups = set(groups)
         group_map = {group: idx for idx, group in enumerate(unique_groups)}
-        kcoords = np.array(group_map.values(), dtype=int)
+        kcoords = np.array(list(group_map.values()), dtype=int)
 
         probe = {
             "chanMap": chanMap,
--- a/src/spikeinterface/sorters/external/kilosort4.py
+++ b/src/spikeinterface/sorters/external/kilosort4.py
@@ -438,9 +438,9 @@ class Kilosort4Sorter(BaseSorter):
         chanMap = np.arange(n_chan)
         xc = positions[:, 0]
         yc = positions[:, 1]
-        unique_groups = np.unique(groups)
+        unique_groups = set(groups)
         group_map = {group: idx for idx, group in enumerate(unique_groups)}
-        kcoords = np.array([group_map[group] for group in groups], dtype=int)
+        kcoords = np.array(group_map.values(), dtype=int)
 
         probe = {
             "chanMap": chanMap,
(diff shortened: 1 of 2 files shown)


src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -438,7 +438,9 @@ def _setup_json_probe_map(cls, recording, sorter_output_folder):
         chanMap = np.arange(n_chan)
         xc = positions[:, 0]
         yc = positions[:, 1]
-        kcoords = groups.astype(float)
+        unique_groups = np.unique(groups)
```
h-mayorquin, line 441: Remember the bug with np.unique, we should use set.

alejoe91: Right! Let me change that!

src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -438,7 +438,9 @@ def _setup_json_probe_map(cls, recording, sorter_output_folder):
...
+        group_map = {group: idx for idx, group in enumerate(unique_groups)}
+        kcoords = np.array([group_map[group] for group in groups], dtype=int)
```
zm711, line 443: I guess the other option here is 

```
kcoords = np.array(group_map.values(), dtype=int)
```

This just makes it more readable for me, but doesn't help with the code so feel free to ignore.

src/spikeinterface/sorters/external/kilosort4.py
alejoe91, line 441:
```suggestion
        unique_groups = set(groups)
```

src/spikeinterface/sorters/external/kilosort4.py
alejoe91: @zm711 just realized that doesn't work. We need kcoords to be the same length as x and y. Reverted

src/spikeinterface/sorters/external/kilosort4.py
chrishalcrow: What's the `np.unique` bug? We use it a lot!

alejoe91: I vaguely remember a potential overflow issue since numpy is returning np.int* instead of Python int (of course in case the elements are ints). @h-mayorquin how wrong am I?

src/spikeinterface/sorters/external/kilosort4.py
alejoe91: No because it returns a flat list of inta from 0 to Ngroups-1

zm711: my bad. I actually had the same thought as I was falling asleep last night and was going to post when I woke up!

h-mayorquin (approved):
LGTM

PR #3946, opened 2025-05-22
basesorting.py `get_spike_trains` method

pas-calc:
convienient function to get spike trains of all units as dict

an alternative method could be
```
spike_vector = sorting.to_spike_vector()
    spikes = spike_vector[spike_vector["unit_index"]==sorting.id_to_index(unit_id)]
    # np.unique(spike_vector["unit_index"],return_counts=True) # number of spikes per unit
    ...
```

### DIFF
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -653,6 +653,18 @@ class BaseSorting(BaseExtractor):
 
         return spikes
 
+    def get_spike_trains(self):
+        """
+        spike trains of sorter in recording time (seconds) of all units
+
+        Returns: dictionary with unit_id as keys and spike train in seconds as values
+        """
+        spike_trains = {}
+        for unit_id in self.get_unit_ids():
+            spike_trains[unit_id] = self.get_unit_spike_train(unit_id) / self.sampling_frequency
+        # self.spike_trains = spike_trains
+        return spike_trains
+    
     def to_numpy_sorting(self, propagate_cache=True):
         """
         Turn any sorting in a NumpySorting.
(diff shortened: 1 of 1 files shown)


chrishalcrow:
Hey @pas-calc , I think this is a great idea! I suspect there will be some debate about how to implement this. Just to note: there's another way to do this as follows:

``` python
spike_trains_samples = si.spike_vector_to_spike_trains(sorting.to_spike_vector(concatenated=False), unit_ids=sort.unit_ids)

# go from sample index to seconds
spike_trains = {segment_id : {unit_id: spike_train/sorting.sampling_frequency for unit_id, spike_train in  spike_trains_s\
egment.items()} for segment_id, spike_trains_segment in spike_trains_samples.items()}
```

`spike_vector_to_spike_trains` uses a numba implementation so should be a lot faster. It might worth benchmarking on some real data. And you also need to worry about segments, I'm afraid.

I would also vote to call it `to_spike_trains` to match the `to_spike_vector` method.

h-mayorquin:
The `get_unit_spike_train` is basically that dict with some extra functionality:

https://github.com/SpikeInterface/spikeinterface/blob/2914abdb42c4c8359493dc3401156e2cf3ea6be0/src/spikeinterface/core/basesorting.py#L131-L174

samuelgarcia:
Hi.
THanks for this proposal.
I think we need to debate on the naming. `to_spike_trains_dict` could be a candidate.
The missing piece here is the segment. What this function shoudl return when multi segment ? a list of dict ?
And then when unique segment ? also a list of dict with one element ?

h-mayorquin:
Thinking more about this I am against adding a method to the sorting extractor class that can be accomplished in one line:

```python
spike_trains_dict = {unit_id: sorting.get_unit_spike_train(unit_id=unit_id, segment_index=0) for unit_id in sorting.unit_ids}
```
Let's not increase the API exposure area at the core level, this leaks complexity to all the other places in the library.

alejoe91:
Hi @pas-calc 

Thanks for the input but we discussed internally and we agree with @h-mayorquin. Since it's a one liner to get the dict of spiketrains let's not overload the core API!

PR #3956, opened 2025-05-27
Expose hidden Kilosort parameters and descriptions

### LINKED ISSUE #3943: Kilosort 1 customize parameters
I'm trying to run kilosort 1 with custom parameters but the default parameters that are exposed for manipulation seem only a subset of the original parameters. I do see these parameters being hard-coded here, so I was wondering if it's even possible to change these parameters

https://github.com/SpikeInterface/spikeinterface/blob/2914abdb42c4c8359493dc3401156e2cf3ea6be0/src/spikeinterface/sorters/external/kilosort.py#L134-L217

kushaangupta:
Closes #3943

### DIFF
--- a/src/spikeinterface/sorters/external/kilosort.py
+++ b/src/spikeinterface/sorters/external/kilosort.py
@@ -46,6 +46,32 @@ class KilosortSorter(KilosortBase, BaseSorter):
         "wave_length": 61,
         "delete_tmp_files": ("matlab_files",),
         "delete_recording_dat": False,
+        "parfor": 0.0,
+        "nNeighPC": None,
+        "nNeigh": 16.0,
+        "whitening": "full",
+        "nSkipCov": 1.0,
+        "whiteningRange": 32.0,
+        "Nrank": 3.0,
+        "nfullpasses": 6.0,
+        "maxFR": 20000,
+        "scaleproc": 200.0,
+        "Th": [4.0, 10.0, 10.0],
+        "lam": [5.0, 5.0, 5.0],
+        "nannealpasses": 4.0,
+        "momentum": [1 / 20, 1 / 400],
+        "shuffle_clusters": 1.0,
+        "mergeT": 0.1,
+        "splitT": 0.1,
+        "initialize": "fromData",
+        "loc_range": [3.0, 1.0],
+        "long_range": [30.0, 6.0],
+        "maskMaxChannels": 5.0,
+        "crit": 0.65,
+        "nFiltMax": 10000.0,
+        "fracse": 0.1,
+        "epu": np.inf,
+        "ForceMaxRAMforDat": 20e9,
     }
 
     _params_description = {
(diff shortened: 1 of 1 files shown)


src/spikeinterface/sorters/external/kilosort.py
```diff
@@ -150,65 +202,82 @@ def _get_specific_options(cls, ops, params) -> dict:
...
-        ops["nfullpasses"] = 6.0  # number of complete passes through data during optimization (6)
-        ops["maxFR"] = 20000  # maximum number of spikes to extract per batch (20000)
+        ops["Nrank"] = params["Nrank"]  # matrix rank of spike template model (3)
+        ops["nfullpasses"] = params["nfullpasses"]  # number of complete passes through data during optimization (6)
+        ops["maxFR"] = params["maxFR"]  # maximum number of spikes to extract per batch (20000)
         ops["fshigh"] = params["freq_min"]  # frequency for high pass filtering
         ops["fslow"] = params["freq_max"]  # frequency for low pass filtering (optional)
         ops["ntbuff"] = params["ntbuff"]  # samples of symmetrical buffer for whitening and spike detection
-        ops["scaleproc"] = 200.0  # int16 scaling of whitened data
+        ops["scaleproc"] = params["scaleproc"]  # int16 scaling of whitened data
```
zm711, line 244: My only concern is that there are a few parameters that are always fixed (I guess someone could fork KS1 and change it themselves), but I think they always do the scaling at 200 for their own thing. So I will wait for @alejoe91 to comment on this since I started using KS at KS2 so maybe he knows better if we shouldn't expose certain parameters. Unless you really want to expose everything?

kushaangupta: Yeah maybe not everything, some parameters apparently aren't meant to be changed. I can remove those. It's hard for me to distinguish the internal parameters from the internally configurable ones. But some parameters that were earlier hardcoded like `max/minFR`, `Th`, `lam` etc were useful to be configured definitely

alejoe91: Hi guys! Yeah this looks good to me! :)

src/spikeinterface/sorters/external/kilosort.py
```diff
@@ -150,33 +200,42 @@ def _get_specific_options(cls, ops, params) -> dict:
...
-        )  # visualization only (Phy): number of channnels to mask the PCs, leave empty to skip (12)
-        ops["nNeigh"] = 16.0  # visualization only (Phy): number of neighboring templates to retain projections of (16)
+
+        # ops["nNeighPC"] = min(12.0, ops["Nchan"]) # Original Kilosort default logic
+        if params["nNeighPC"] is not None:
+            ops["nNeighPC"] = params["nNeighPC"]
+        else:
+            # Kilosort's default behavior if nNeighPC is None in params (from _default_params)
+            ops["nNeighPC"] = min(
+                12.0, ops.get("Nchan", 12.0)
```
zm711: Previous logic assumes that ops has an "Nchan" get. Is there a reason why you want to change the logic such that you default to 12? Is there ever a time when ops doesn't have "Nchan"?

src/spikeinterface/sorters/external/kilosort.py
```diff
@@ -150,33 +200,42 @@ def _get_specific_options(cls, ops, params) -> dict:
...
-        ops["whiteningRange"] = (
-            32.0  # how many channels to whiten together (Inf for whole probe whitening, should be fine if Nchan<=32)
-        )
+        ops["whitening"] = params[
+            "whitening"
+        ]  # type of whitening (default 'full', for 'noSpikes' set options for spike detection below)
+        ops["nSkipCov"] = params["nSkipCov"]  # compute whitening matrix from every N-th batch (1)
+        ops["whiteningRange"] = params[
+            "whiteningRange"
+        ]  # how many channels to whiten together (Inf for whole probe whitening, should be fine if Nchan<=32)
```
zm711, line 240: Based on this comment it seems like if we are allowing the user to set their own number then we should either prevent them from using > 32 or warn them. But we do say inf is possible. 

Do you know if we should be doing some sort of check here?

src/spikeinterface/sorters/external/kilosort.py
```diff
@@ -188,27 +247,30 @@ def _get_specific_options(cls, ops, params) -> dict:
...
-        ops["nannealpasses"] = 4.0  # should be less than nfullpasses (4)
-        ops["momentum"] = [1 / 20, 1 / 400]  # start with high momentum and anneal (1./[20 1000])
-        ops["shuffle_clusters"] = 1.0  # allow merges and splits during optimization (1)
-        ops["mergeT"] = 0.1  # upper threshold for merging (.1)
-        ops["splitT"] = 0.1  # lower threshold for splitting (.1)
-
-        ops["initialize"] = "fromData"  # 'fromData' or 'no'
+        ops["Th"] = params["Th"]  # threshold for detecting spikes on template-filtered data ([6 12 12])
+        ops["lam"] = params["lam"]  # large means amplitudes are forced around the mean ([10 30 30])
+        ops["nannealpasses"] = params["nannealpasses"]  # should be less than nfullpasses (4)
```
zm711, line 261: Same here. Our comment says should be less than "nfullpasses". If this is actually required then we should enforce it.

zm711 (commented):
Just a few comments @kushaangupta

src/spikeinterface/sorters/external/kilosort.py
kushaangupta: Oh I think I didn't realize that the Nchan was being created while generating ops file in kilosortbase.py

src/spikeinterface/sorters/external/kilosort.py
kushaangupta: I'm not sure if I fully understand. Isn't it saying that it can be infinity if Nchan <=32?

zm711: The issue is that users should use Nchan <=32, but alternatively they could whitened based on a whole probe by doing infinite, but this is not recommended. So the question would be do we 

1) assert <=32 to ensure that people use the more appropriate local whitening rather than allowing global whitening
2) keep what you've done and let people do whatever they want, but default to 32
3) let people do what they want, but warn if > 32

I'm note completely sure what the right answer is. I'm leaning toward 3, but if you have strong opinions I would love to hear arguments for 1 or 2.

kushaangupta: I updated it for 3

zm711: Close. So what I would do instead is:

```python
if params["whiteningRange"] > 32:
    n_channels_whitening = params["whiteningRange"] if np.isfinite(params["whiteningRange"]) else "all"
    warn(f"Kilosort recommends whitening with 32 or fewer channels you are whitening with {n_channels_whitening} channels"
ops["whiteningRange"] = params['whiteningRange"]
```

But in this case at the top of the file you would need to add 

```python
from warnings import warn
```

Want to give that a go instead? If not I can push this change myself.

src/spikeinterface/sorters/external/kilosort.py
```diff
@@ -188,27 +252,31 @@ def _get_specific_options(cls, ops, params) -> dict:
...
+        assert ops["nannealpasses"] < ops["nfullpasses"]
```
zm711:
```suggestion
        assert ops["nannealpasses"] < ops["nfullpasses"], f"{nannealpasses=} should be less than {nfullpasses=}"
```
or you could make it smoother if you wanted to write out a proper error message!  But we prefer when possible to show the end-user what they are doing wrong :)

kushaangupta: Having a description is certainly helpful

src/spikeinterface/sorters/external/kilosort.py
```diff
@@ -188,27 +252,31 @@ def _get_specific_options(cls, ops, params) -> dict:
...
-        ops["momentum"] = [1 / 20, 1 / 400]  # start with high momentum and anneal (1./[20 1000])
-        ops["shuffle_clusters"] = 1.0  # allow merges and splits during optimization (1)
-        ops["mergeT"] = 0.1  # upper threshold for merging (.1)
-        ops["splitT"] = 0.1  # lower threshold for splitting (.1)
-
-        ops["initialize"] = "fromData"  # 'fromData' or 'no'
+        ops["Th"] = params["Th"]  # threshold for detecting spikes on template-filtered data ([6 12 12])
+        ops["lam"] = params["lam"]  # large means amplitudes are forced around the mean ([10 30 30])
+        ops["nannealpasses"] = params["nannealpasses"]  # should be less than nfullpasses (4)
+        assert ops["nannealpasses"] < ops["nfullpasses"], f"{nannealpasses=} should be less than {nfullpasses=}"
```
zm711:
```suggestion
        assert ops["nannealpasses"] < ops["nfullpasses"], f"{ops['nannealpasses']=} should be less than {ops['nfullpasses']=}"
```
I meant write something like that :) If you wanted to commit my suggestion we should have done the above.

zm711:
Tests aren't working right now. Seems like pytest released a new version today. We are in the process of fixing the testing!

src/spikeinterface/sorters/external/kilosort.py
kushaangupta: oops sorry about that

src/spikeinterface/sorters/external/kilosort.py
zm711: Now that testing is working again did you want to give a go editing to something like what I did above? Or would you like me to just push it myself?

alejoe91:
LGTM! Thanks @kushaangupta

PR #3958, opened 2025-05-27
keep probe_planar_contour when aggregating channels 

### LINKED ISSUE #3947: probe_planar_contour lost when splitting & aggregating channels
Hi,
I noticed that probe inside raw_rec after the splitting and aggregating lose .probe_planar_contour.

```
        probes = read_probeinterface(probe_json)
        raw_rec = raw_rec.set_probegroup(probes, group_mode="by_shank")
        split_recording_dict = raw_rec.split_by("group")
        proc_rec = spre.highpass_filter(recording=split_recording_dict)
        proc_rec = spre.common_reference(recording=proc_rec, operator="median", reference="global")
        raw_rec = aggregate_channels(proc_rec)
        raw_rec.clear_channel_groups()

```

so I am forced to call again       `raw_rec = raw_rec.set_probegroup(probes, group_mode="by_shank")` afterward, is this fixable?

RobertoDF:
Fix #3947

### DIFF
--- a/src/spikeinterface/core/channelsaggregationrecording.py
+++ b/src/spikeinterface/core/channelsaggregationrecording.py
@@ -105,9 +105,13 @@ class ChannelsAggregationRecording(BaseRecording):
                 "Locations are not unique! " "Cannot aggregate recordings!"
             )
 
-        planar_contour_keys = [key for recording in recording_list for key in recording.get_annotation_keys() if 'planar_contour' in key]
-        if len(planar_contour_keys)>0:
-            if all(k == planar_contour_keys[0] for k in planar_contour_keys):  # we add the 'planar_contour' annotations only if there is a unique one in the recording_list
+        planar_contour_keys = [
+            key for recording in recording_list for key in recording.get_annotation_keys() if "planar_contour" in key
+        ]
+        if len(planar_contour_keys) > 0:
+            if all(
+                k == planar_contour_keys[0] for k in planar_contour_keys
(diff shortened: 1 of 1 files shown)


chrishalcrow:
Hello, looking great - good bug catch! There are other annotations we're not forwarding during this function (like `probes_info`). It'd be amazing if your code generalised to all annotations. I think your basic structure will abstract pretty nicely to all annotations. You'd do something like

```
for annotation_key in recording.get_annotation_keys():
    check annotation_key is a key in all sub-recordings
    check annotation_value is the same for all sub-recordings
    If it is:
        set the annotation (provided it's not already set: e.g is_filtered gets set somehow)
```

I think the hard thing is to make the `all` function work more generally. In my testing, something like
``` python
all( np.all(k == first_annotation_value) for k in annotation_values))
```
seems ok, but maybe someone has a better idea.

samuelgarcia (approved):

samuelgarcia:
Thanks for this.
I am ok with it.
ANd also ok with @chrishalcrow  comments.

RobertoDF:
@chrishalcrow I think your idea will work well with small modifications from what I proposed. I won´t have time to test this this week though!

h-mayorquin:
I would like to point out that this is a data point of just passing the whole probe as an attribute and fetch the information from the probe when it is available. Spreading the probe information across annotations and properties makes this type of error harder to trace.

samuelgarcia:
thanks you. Lets make an issue to make it for all annotations

PR #3970, opened 2025-06-03
Fix for handle_drift bullet points

zm711:
I don't have access to the handle_drift data to rerun myself. I think this is a quick fix for the rst file in the meantime until someone decides to fix the root .py file. Not sure that this will fix it so I'll wait for ci to render. I tried to potentially fix the underlying py file, but again without the data I can't know for sure by running locally.

### DIFF
--- a/doc/how_to/handle_drift.rst
+++ b/doc/how_to/handle_drift.rst
@@ -1455,38 +1455,32 @@ Plot the results
 We load back the results and use the widgets module to explore the
 estimated drift motion.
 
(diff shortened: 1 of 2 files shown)


zm711:
Screenshot 2025-06-03 at 17 42 58

This PR fixes the RST issue. But like I said not sure if it fixes the .py issue. I would say we merge for now since we don't know the next time someone will run the drift docs.

alejoe91 (approved):

PR #3971, opened 2025-06-03
cleanup __init__ of exporters + cleanup imports

zm711:
importing with
```python
from x import y as y
```

allows for full linter support in the future. If this is too contentious feel free to close this one :)


zm711:
I missed some sort of monkeypatching going on. Closing.

PR #3974, opened 2025-06-04
Add SorterStudyWithoutGroundTruth

samuelgarcia:
This is the very beginnings of a benchmark class without groundtruth using internally the multi comparison strategy and then doing some quality metrics.

This also include an analysis of the residual.

This is here for discussion at the moment.

@yger


samuelgarcia:
big mess in my branches.

Issue #3969, opened 2025-06-03
resample() converts dtype to int16 even when dtype=‘uint16’ is specified


carmengasco:
Hello!

When I load a binary file using read_binary, I specify dtype='uint16'. As expected, the trace I get is indeed of type uint16.

However, the issue arises when I use .resample(recording, 2500, dtype='uint16'). Even though I explicitly set the output dtype, the resulting data is automatically converted to int16.

Is there a way to preserve the original uint16 dtype during resampling?

Thanks a lot!

h-mayorquin:
Which method is resample? Can you share a script that reproduces your error?

carmengasco:
This is the resample method: 
https://spikeinterface.readthedocs.io/en/stable/api.html#spikeinterface.preprocessing.resample

<img width="503" alt="Image" src="https://github.com/user-attachments/assets/1ebbeca2-c35a-4e95-b86d-366647f645e6" />

zm711:
It seems like we have a note in the code that we need to force uint -> int
https://github.com/SpikeInterface/spikeinterface/blob/604c049bd078f08fd00c284419aba2d09eafa7c1/src/spikeinterface/preprocessing/filter.py#L406-L410

Not sure why though.

h-mayorquin:
I think this is for @alejoe91 and @samuelgarcia

h-mayorquin:
The reason for having this behavior is that some scipy functions fail silently when taking uint as an input. So we have this all over preprocessing. I think that @alejoe91 recommended decimate as an alternative preprocessing method here but can't remember well.

zm711:
Since we moved forward with #3982 this is true for all preprocessing and a warning will be raised explaining issues with with using unsigned int's within the code base so I'll close this. But if you have more questions just let us know!

PR #3975, opened 2025-06-04
Add documentation to `fix_dtype`

h-mayorquin:
From the meeting today. Just adding a docstring so we remember why this is here.

Maybe we should change the name to `force_signed_int` to make it more clear? we can do this in another PR.

### DIFF
--- a/src/spikeinterface/preprocessing/filter.py
+++ b/src/spikeinterface/preprocessing/filter.py
@@ -400,8 +400,27 @@ highpass_filter.__doc__ = highpass_filter.__doc__.format(_common_filter_docs)
 
 def fix_dtype(recording, dtype):
     """
-    Many preprocessors access scipy functions (e.g. filter) that fail silently with unsigned dtypes.
-    This function fixes the dtype to be signed if it is unsigned.
+    Fix and validate dtype for preprocessing operations.
+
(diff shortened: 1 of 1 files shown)


h-mayorquin:
Also, maybe this should throw a warning so it informs users like in #3969?

src/spikeinterface/preprocessing/filter.py
```diff
@@ -399,6 +399,10 @@ def causal_filter(
 
 
 def fix_dtype(recording, dtype):
+    """
+    Many preprocessors access scipy functions (e.g. filter) that fail silently with unsigned dtypes.
+    This function fixes the dtype to be signed if it is unsigned.
+    """
```
zm711: Maybe we are even more verbose? It does 3 things:
1) if no dtype is given it will attempt to infer from recording
2) It will attempt to convert the dtype to an npt.DTypeLike (I forget the actual typing)
3) if the dype is unsigned it will convert to the signed equivalent (--at potential truncation of large values?)

Or am I misunderstanding?

h-mayorquin: I made improvements in this direction.

samuelgarcia:
Lets close this in favor of this #3982

PR #3976, opened 2025-06-05
Avoid crash of SC2 if no templates are detected after clustering

yger:
It can be that after clustering, no proper templates are detected. This is unlikely, but if so, the code is crashing. This PR prevents it and make sure that for any template matching based engine, the templates object has more than 0 templates

### DIFF
--- a/src/spikeinterface/sortingcomponents/matching/main.py
+++ b/src/spikeinterface/sortingcomponents/matching/main.py
@@ -9,6 +9,13 @@ import numpy as np
 from spikeinterface.core.job_tools import fix_job_kwargs
 from spikeinterface.core.node_pipeline import run_node_pipeline
 
+spike_dtype = [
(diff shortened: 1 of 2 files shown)


alejoe91 (approved):

src/spikeinterface/sortingcomponents/matching/main.py
```diff
@@ -9,6 +9,14 @@
 from spikeinterface.core.job_tools import fix_job_kwargs
 from spikeinterface.core.node_pipeline import run_node_pipeline
 
+spike_dtype = [
+    ("sample_index", "int64"),
+    ("channel_index", "int64"),
+    ("cluster_index", "int64"),
+    ("amplitude", "float64"),
+    ("segment_index", "int64"),
+]
```
samuelgarcia: import it!

PR #3979, opened 2025-06-10
average by bins option in benchmark plot_performances_vs_snr

samuelgarcia:


alejoe91:
Nice! @samuelgarcia can you add a screenshot of how it looks?

samuelgarcia:
Sometime the sigmoid fitting is wrong so an average by bin is better for benchmarks.
image

zm711 (commented):
I'm a little confused by on the on-off? So if I set
```python
with_sigmoid_fit=True
show_average_by_bin=True
```
what happens? Both are plotted? Or should only one at a time be possible? In which case you could do an if-elif-else right? Or check that only one of them is true?

samuelgarcia:
Having both
with_sigmoid_fit=True
show_average_by_bin=True
is mainly to check that the fitting is bad comparing to the average by bins (or the reverse).
Having both curbe on the final same plots is non sens of course.

zm711:
Then my concern is that an end-user will do both and not be able to interpret them appropriately. Should we account for that? I guess you're wanting to be able to do both for the debugging purposes/checking purposes, but I think we should at least in that case mention in the docstring that one should pick the better model and use that for final plots.

samuelgarcia:
The "end user" here is mostly a sorter developer.
I think it is convinient to have both curves sometimes.

yger (approved):

samuelgarcia:
@zm711 we need some merges now to fix several branch.

PR #3980, opened 2025-06-10
Fix double definition of openephys warnings

alejoe91:


alejoe91:
This actually results in an `UnboundLocalVariable` error in case setting timestamps fails...

zm711 (approved):
That's really weird. I thought it would just find the package in some sort of python cache.... But doesn't hurt to delete unnecessary import :)

alejoe91:
@vncntprvst FYI

PR #3982, opened 2025-06-11
Remove auto_cast_uint, cast_unsigned, and modify fix_dtype

### LINKED ISSUE #1815: Deprecate `cast_unsigned` as an argument of core functions?
In get traces we have the `cast_unsigned` parameter:

https://github.com/SpikeInterface/spikeinterface/blob/f779e12191515b49ff4f73a02d453f44e1be00ec/src/spikeinterface/core/baserecording.py#L234-L243

Which I think is just another way of doing what @alejoe91 enabled with a preprocessing step on:
https://github.com/SpikeInterface/spikeinterface/pull/1707

We also have a related parameter `auto_cast_uint` in `write_binary_recording` 

https://github.com/SpikeInterface/spikeinterface/blob/f779e12191515b49ff4f73a02d453f44e1be00ec/src/spikeinterface/core/core_tools.py#L244-L252

So, we have this functionality that I think can be achieved with the preprocessing class `UnsignedToSignedRecording`. My naive take is this should be slowly deprecated in favor of using this transformation as a preprocessing step when required. 

Why?
* Maintenance of unsigned to sign casting: having this logic in a centralized place (the `UnsignedToSignedRecording` class) means is easier to debug, test and document. 
* Discoverability: It is easier to discover a form of preprocessing in the preprocessing module that to find it across flags spread around different functions.
* Maintenance of core: I have the firm belief that the core functions should be a simple as possible. One form of simplicity is signature simplicity. Having flags in core functions to resolve what are preprocessing or IO use cases increases drastically the surface area for bugs to sneak in. The branching and the number of cases to consider increases as the use is spread all over the library and is harder to reason about changes. It increases testing time if we want to be complete. The philosophy is keep the core simple and export complexity to the use cases.

I really don't see an advantage to preserving this but maybe I lack imagination?  What I am not seeing? I undestand why they came to be there but now that we have a preprocessing module to accomplish that it seems like the way to go.

Thoughts?

alejoe91:
Fixes #1815

### DIFF
--- a/src/spikeinterface/preprocessing/filter.py
+++ b/src/spikeinterface/preprocessing/filter.py
@@ -423,7 +423,8 @@ def fix_dtype(recording, dtype):
     # if uint --> force int
     if dtype.kind == "u":
         raise ValueError(
-            "Unsigned types are not supported. You can use "
+            "Unsigned types are not supported, since they don't ineract well with "
+            "various preprocessing steps. You can use "
             "`spikeinterface.preprocessing.unsigned_to_signed` to convert the recording to a signed type."
         )
 


h-mayorquin (commented):
I strongly agree with this but shouldn't we do some type of deprecation first? so we don't break the code of people who use positional arguments.

alejoe91:
It was already deprecated. And since next release it's going to be minor I think it's ok to introduce a minor API change. Honwstly I don't think anyone is using `cast_unsigned` and `autocast_unsigned_to_signed`

samuelgarcia:
We should have a clear doc on this.

src/spikeinterface/core/recording_tools.py
```diff
@@ -117,12 +112,6 @@ def write_binary_recording(
         file_path_list = [add_suffix(file_path, ["raw", "bin", "dat"]) for file_path in file_path_list]
 
     dtype = dtype if dtype is not None else recording.get_dtype()
-    if auto_cast_uint:
-        cast_unsigned = determine_cast_unsigned(recording, dtype)
-        warning_message = (
-            "auto_cast_uint is deprecated and will be removed in 0.103. Use the `unsigned_to_signed` function instead."
```
zm711, line 123: This seems like a clear enough deprecation to me :)

src/spikeinterface/preprocessing/filter.py
```diff
@@ -399,12 +399,32 @@ def causal_filter(
...
+        The fixed numpy.dtype
+    """
     if dtype is None:
         dtype = recording.get_dtype()
     dtype = np.dtype(dtype)
 
     # if uint --> force int
     if dtype.kind == "u":
-        dtype = np.dtype(dtype.str.replace("u", "i"))
+        raise ValueError(
```
zm711, line 425: I would consider making this just a little clearer. I think users would want to know why they aren't supported. So I would recommend adding an additional sentence that says:

"unsigned types are not supported due to **"

I'm just trying to think about what to say? inconsistent behavior in the library? Or to flip it and say that ints and floats perform better.

The issue we are referencing has someone who is saying "I want to use unsigned" so I think we should at least say even in short form why they should not. I agree that maybe some sort of documentation we can point to that explains this or GitHub issue referenced would at least let the users know the real reason why they can't. Especially since this is a change in behavior.

zm711 (commented):
The one thing I

src/spikeinterface/preprocessing/filter.py
alejoe91: Sure. We can say that scopy functions don't work well with unsigned ;)

alejoe91: @zm711 better?

h-mayorquin:
> It was already deprecated. And since next release it's going to be minor I think it's ok to introduce a minor API change. Honwstly I don't think anyone is using `cast_unsigned` and `autocast_unsigned_to_signed`

This makes sense. Then by all means, I have been wishing to get rid of this for a while : )

src/spikeinterface/core/recording_tools.py
h-mayorquin: I meant in `get_traces`.

src/spikeinterface/preprocessing/filter.py
zm711: Missed this but yeah that looks good to me :)

PR #3984, opened 2025-06-12
Add IBL's "noise cutoff" quality metric

huizizhang949:
Add a new metric "noise cutoff" based on IBL [https://figshare.com/articles/online_resource/Spike_sorting_pipeline_for_the_International_Brain_Laboratory/19705522?file=49783080](url) in a new function `compute_noise_cutoffs`. 

The original implementation is here [https://github.com/int-brain-lab/ibllib/blob/2e1f91c622ba8dbd04fc53946c185c99451ce5d6/brainbox/metrics/single_units.py](url).

The "noise cutoff" metric assesses whether a unit’s spike‐amplitude distribution is truncated
at the low-end, which may be due to the high amplitude detection threhold in the deconvolution step, 
i.e., if low‐amplitude spikes were missed.  It does not assume a Gaussian shape; 
instead, it directly compares counts in the low‐amplitude bins to counts in high‐amplitude bins (`cutoff`) and the peak bin (`ratio`).

Here are examples with no truncation and truncation, which give low and high values (`cutoff` and `ratio`), respectively.
example_cutoff

I have added documentation and tests.

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/spikeglx.py
+++ b/src/spikeinterface/extractors/neoextractors/spikeglx.py
@@ -75,9 +75,12 @@ class SpikeGLXRecordingExtractor(NeoBaseRecordingExtractor):
 
         self._kwargs.update(dict(folder_path=str(Path(folder_path).absolute()), load_sync_channel=load_sync_channel))
 
-        stream_is_nidq_or_sync = "nidq" in self.stream_id or "SYNC" in self.stream_id
-        if stream_is_nidq_or_sync:
-            # Do not add probe information for the sync or nidq stream. Early return
+        stream_is_nidq = "nidq" in self.stream_id
+        stream_is_one_box = "obx" in self.stream_id
+        stream_is_sync = "SYNC" in self.stream_id
+
+        if stream_is_nidq or stream_is_one_box or stream_is_sync:
+            # Do not add probe information for the one box, nidq or sync streams. Early return
             return None
 
         # Checks if the probe information is available and adds location, shanks and sample shift if available.
--- a/src/spikeinterface/extractors/tests/test_neoextractors.py
+++ b/src/spikeinterface/extractors/tests/test_neoextractors.py
@@ -105,6 +105,7 @@ class SpikeGLXRecordingTest(RecordingCommonTestSuite, unittest.TestCase):
         ("spikeglx/Noise4Sam_g0", {"stream_id": "imec0.lf"}),
         ("spikeglx/Noise4Sam_g0", {"stream_id": "nidq"}),
         ("spikeglx/Noise4Sam_g0", {"stream_id": "imec0.ap-SYNC"}),
+        ("spikeglx/onebox/run_with_only_adc/myRun_g0", {"stream_id": "obx0"}),
     ]
 
 


doc/modules/qualitymetrics/noise_cutoff.rst
```diff
@@ -1,30 +1,123 @@
...
+i.e., if low‐amplitude spikes were missed.  It does not assume a Gaussian shape;
+instead, it directly compares counts in the low‐amplitude bins to counts in high‐amplitude bins.
+
+1. **Build a histogram**
+
+   For each unit, divide all amplitudes into `n_bins` equally spaced bins over the range of the amplitude.
+   Let `n_i` denote the count in the :math:`i`-th bin.
+
+2. **Identify the “low” region**
+    - Compute the amplitude value at the specified `low_quantile` (for example, 0.10 = 10 percentile), denoted as :math:`\text{amp}_{low}`.
```
zm711:
```suggestion
    - Compute the amplitude value at the specified `low_quantile` (for example, 0.10 = 10th percentile), denoted as :math:`\text{amp}_{low}`.
```

doc/modules/qualitymetrics/noise_cutoff.rst
```diff
@@ -1,30 +1,123 @@
...
+    - Find all histogram bins whose upper edge is below that quantile value.  These bins form the “low‐quantile region”.
+    - Compute
+
+    .. math::
+        L_{\mathrm{bins}} = \bigl\{i : \text{upper\_edge}(i) \le \text{amp}_{low}\bigr\}, \quad
+        \mu_{\mathrm{low}} = \frac{1}{|L_{\mathrm{bins}}|}\sum_{i\in L_{\mathrm{bins}}} n_i.
+
+3. **Identify the “high” region**
+
+   - Compute the amplitude value at the specified `high_quantile` (for example, 0.25 = top 25 percentile), denoted as :math:`\text{amp}_{high}`.
```
zm711:
```suggestion
   - Compute the amplitude value at the specified `high_quantile` (for example, 0.25 = top 25th percentile), denoted as :math:`\text{amp}_{high}`.
```

doc/modules/qualitymetrics/noise_cutoff.rst
```diff
@@ -1,30 +1,123 @@
...
+------------
+
+.. code-block:: python
+
+    import numpy as np
+    import matplotlib.pyplot as plt
+    from spikeinterface.full as si
+
+    # Suppose `sorting_analyzer` has been computed with spike amplitudes:
+    # Select units you are interested in vizulazing.
```
zm711:
```suggestion
    # Select units you are interested in visualizing
```

doc/modules/qualitymetrics/noise_cutoff.rst
```diff
@@ -1,30 +1,123 @@
...
+    unit_ids = ...
+
+    # Compute noise_cutoff:
+    summary_dict = compute_noise_cutoff(
+        sorting_analyzer=sorting_analyzer
+        high_quantile=0.25,
+        low_quantile=0.10,
+        n_bins=100,
```
zm711: There is no discussion of what number of bins means for this metric. Might be worth adding. Why not do 10, or 1000 bins?

doc/modules/qualitymetrics/noise_cutoff.rst
```diff
@@ -1,30 +1,123 @@
...
+
+Reference
+---------
+
+.. autofunction:: spikeinterface.qualitymetrics.misc_metrics.compute_noise_cutoff
+
+Examples with plots
+-------------------
+
+Here is shown the histogram of two units, with the vertical lines separating low- and high-amplitude region.
```
zm711:
```suggestion
Here is shown the histogram of two units, with the vertical lines separating low- and high-amplitude regions.
```

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -37,6 +37,151 @@
 _default_params = dict()
 
 
+def compute_noise_cutoffs(sorting_analyzer, high_quantile=0.25, low_quantile=0.1, n_bins=100, unit_ids=None):
+    """
+    A metric to determine if a unit's amplitude distribution is cut off, without assuming a Gaussian distribution.
+
+    The function flips the sign if 'peak_sign' == 'neg' when computing the amplitude.
+    Based on the histogram of the (transformed) amplitude:
```
zm711: What does this mean?

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -37,6 +37,151 @@
...
+    -------
+    noise_cutoff_dict : dict of floats
+        Estimated metrics based on the amplitude distribution, for each unit ID.
+
+    References
+    ----------
+    Inspired by metric described in [IBL2024]_
+
+    """
+    res = namedtuple("cutoff_metrics", ["noise_cutoff", "noise_ratio"])
```
zm711: @chrishalcrow ...

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -37,6 +37,151 @@
...
+    if unit_ids is None:
+        unit_ids = sorting_analyzer.unit_ids
+
+    noise_cutoff_dict = {}
+    noise_ratio_dict = {}
+    if not sorting_analyzer.has_extension("spike_amplitudes"):
+        warnings.warn("compute_noise_cutoffs need 'spike_amplitudes'")
```
zm711: Heberto and I prefer actionable warnings. This says the problem without the solution. so maybe something like

```suggestion
        warnings.warn("`compute_noise_cutoffs` requires the 'spike_amplitudes` extension. Please run sorting_analyzer.compute('spike_amplitudes') to be able to compute `noise_cutoff`")
```

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -37,6 +37,151 @@
...
+            noise_cutoff_dict[unit_id] = np.nan
+            noise_ratio_dict[unit_id] = np.nan
+        return res(noise_cutoff_dict, noise_ratio_dict)
+
+    amplitude_extension = sorting_analyzer.get_extension("spike_amplitudes")
+    peak_sign = amplitude_extension.params["peak_sign"]
+    amplitudes_by_units = _get_amplitudes_by_units(sorting_analyzer, unit_ids, peak_sign)
+
+    if peak_sign == "both":
+        raise TypeError('peak_sign should either be "pos" or "neg"!')
```
zm711: Shouldn't this error occur right after peak sign so we don't waste time doing the `_get_amplitudes_by_units` function?

zm711: Also how does a user know what the peak sign is? Where did it come from. I would consider making this note refer to the peak sign of the spike_amplitudes extension.

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -37,6 +37,151 @@
...
+
+    for unit_id in unit_ids:
+        amplitudes = amplitudes_by_units[unit_id]
+
+        if peak_sign == "neg":
+            amplitudes = -amplitudes
```
zm711: I would consider a comment explaining why we are doing this.

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -37,6 +37,151 @@
...
+    from that mean the lower-quantile bins lie.
+
+    2. The method also compares the counts in the lower-amplitude bins to the count in the highest bin and return their ratio.
+
+    Parameters
+    ----------
+    amps : array-like
+        Spike amplitudes.
+    high_quantile : float, default: 0.25
+        Quantile of the amplitude range above which values are treated as "high" (e.g. 0.25 = top 25%), the reference region.
```
zm711: optionally thinking about this should we also put a note about the top 25% is really the 75 percentile? Not sure about this.

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -37,6 +37,151 @@
...
+    -------
+    cutoff : float
+        (mean(lower_bins_count) - mean(high_bins_count)) / std(high_bins_count)
+    ratio: float
+        mean(lower_bins_count) / highest_bin_count
+
+    """
+    n_per_bin, bin_edges = np.histogram(amps, bins=n_bins)
+
+    peak = np.max(n_per_bin)
```
zm711: peak is a little bit of a loaded term. Something to think about for code readability.

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -37,6 +37,151 @@
...
+        warnings.warn("no bin is selected as the reference region. please increase high_quantile.")
+        return np.nan, ratio
+
+    # compute cutoff from low-amplitude and high-amplitude bins
+    high_counts = n_per_bin[high_indices]
+    mean_high_counts = np.mean(high_counts)
+    std_high_counts = np.std(high_counts)
+    if std_high_counts == 0:
+        warnings.warn(
+            "only one bin is selected as the reference region, and thus the standard deviation cannot be computed. please increase high_quantile."
```
zm711: I guess this could also be the problem if your high count bins are all the same. So in that case changing the high quantile might not be the solution. In that case changing n_bins would potentially be beneficial, no?

doc/modules/qualitymetrics/noise_cutoff.rst
huizizhang949: I have added a comment to the docs. This is also the default value used in IBL's implementation.

src/spikeinterface/qualitymetrics/misc_metrics.py
huizizhang949: I added a comment about where to change the peak_sign in the error message. I moved the error before `_get_amplitudes_by_units`.

src/spikeinterface/qualitymetrics/misc_metrics.py
huizizhang949: I added a comment in the code to explain why.

src/spikeinterface/qualitymetrics/misc_metrics.py
huizizhang949: I think 'eg. 0.25 = top 25%' makes this clear.

src/spikeinterface/qualitymetrics/misc_metrics.py
huizizhang949: I changed the name of it in the code.

src/spikeinterface/qualitymetrics/misc_metrics.py
huizizhang949: I have added included this case where all the bins are the same and suggested changing `n_bins`.

src/spikeinterface/qualitymetrics/misc_metrics.py
huizizhang949: I have changed the doc string.

src/spikeinterface/qualitymetrics/misc_metrics.py
chrishalcrow: yess?

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -18,7 +18,7 @@
 import numpy as np
 
 from spikeinterface.core.job_tools import fix_job_kwargs, split_job_kwargs
-from spikeinterface.postprocessing import correlogram_for_one_segment
+from spikeinterface.postprocessing import correlogpram_for_one_segment
```
chrishalcrow:
```suggestion
from spikeinterface.postprocessing import correlogpram_for_one_segment
```
```suggestion
from spikeinterface.postprocessing import correlogram_for_one_segment
```

chrishalcrow: This is what broke the tests :)

src/spikeinterface/qualitymetrics/misc_metrics.py
zm711: Just seeing that a namedtuple decided to appear :)

zm711:
From a profiling perspective is there any benefit to thinking about doing a numba/numpy dual strategy? It seems like `_noise_cutoff` is all numpy based so could be sped up with a numba version. That would complicate the PR strategy, but would be curious about the profiling. It would also be an excellent exercise in profiling code performance. Definitely not required unless this is really slow. That's the reason we had to shut off the nn metrics as a default for qm stuff.

src/spikeinterface/qualitymetrics/misc_metrics.py
chrishalcrow: haha - I think this is copied from `compute_amplitude_cutoffs`. But very pleased to have another nametuple to the codebase, of course!

samuelgarcia: I think we have namedtuple for all metrics output no ?

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -37,6 +37,160 @@
...
+        return res(noise_cutoff_dict, noise_ratio_dict)
+
+    amplitude_extension = sorting_analyzer.get_extension("spike_amplitudes")
+    peak_sign = amplitude_extension.params["peak_sign"]
+    if peak_sign == "both":
+        raise TypeError(
+            '`peak_sign` should either be "pos" or "neg". You can set `peak_sign` as an argument when you compute spike_amplitudes.'
+        )
+
+    amplitudes_by_units = _get_amplitudes_by_units(sorting_analyzer, unit_ids, peak_sign)
```
samuelgarcia: lets have in mind that this `_get_amplitudes_by_units` is not efficient because it internal use mask on all units.

amplitude_extension.get_data(outputs="by_units") is really faster for this. We should use it here and also inside 
`_get_amplitudes_by_units` the only pain is that amplitude will be splited over segments but concatenate is fast.

samuelgarcia:
Thank you for this, this is really nice.
This is a bit similar for the global question that "amplitude_cutoff" but with a differents approach and implementation.
The actual "amplitude_cutoff" is not super reliable. 
This looks easier.
I am wondering how much this ratio is able to discriminate a real cut in the distribution like you have on the right plots and a distribution with a positive skewness coeeficient.

src/spikeinterface/qualitymetrics/misc_metrics.py
zm711: @samuelgarcia it was 100% a joke. Recently Chris said he liked the concept of namedtuples when Heberto brought them up in a different PR. So I was just teasing.

zm711:
> The actual "amplitude_cutoff" is not super reliable.

related to this, I don't think I've ever got `amplitude_cutoff` to work on my data. I always get nans for it. So this could be an interesting if it works more often.

src/spikeinterface/qualitymetrics/misc_metrics.py
huizizhang949: We just tried to implement `amplitude_extension.get_data(outputs="by_units")` and then concatenated, but it became very complex. We propose that there should be another PR to make `_get_amplitudes_by_units` to use the approach suggested.

chrishalcrow:
Hello @oliche @GaelleChapuis , @huizizhang949 has implemented a version of the IBL noise cutoff metric here in spikeinterface (docs: https://spikeinterface--3984.org.readthedocs.build/en/3984/modules/qualitymetrics/noise_cutoff.html). Hoping it will be a helpful tool here!

If you have time to have a look at the implementation, we'd appreciate any feedback! The bulk of the code is in `def _noise_cutoff`.

alejoe91:
@huizizhang949 why did you close this? I think it was in very good shape!

PR #3985, opened 2025-06-16
Fix durations in generate_hybrid_recording

alejoe91:
In the hybrid generation, if the recording has timestamps, they might exceed the (num_samples/sampling_frequency) duration, generating spikes beyond the actual duration of the recording.

For hybrid generation we can simply use num_samples / sampling_frequency

### DIFF
--- a/src/spikeinterface/generation/hybrid_tools.py
+++ b/src/spikeinterface/generation/hybrid_tools.py
@@ -400,7 +400,7 @@ def generate_hybrid_recording(
     num_segments = recording.get_num_segments()
     dtype = recording.dtype
     num_samples = np.array([recording.get_num_samples(segment_index) for segment_index in range(num_segments)])
-    # since the recording can have timestamps with some smalle gaps, we use the number of samples to compute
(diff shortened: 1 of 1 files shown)


zm711:
Just updated branch so that testing will work.

src/spikeinterface/generation/hybrid_tools.py
```diff
@@ -399,8 +399,10 @@ def generate_hybrid_recording(
     probe = recording.get_probe()
     num_segments = recording.get_num_segments()
     dtype = recording.dtype
-    durations = np.array([recording.get_duration(segment_index) for segment_index in range(num_segments)])
     num_samples = np.array([recording.get_num_samples(segment_index) for segment_index in range(num_segments)])
+    # since the recording can have timestamps with some smalle gaps, we use the number of samples to compute
```
zm711:
```suggestion
    # since the recording can have timestamps with some small gaps, we use the number of samples to compute
```
:)

samuelgarcia (approved):

PR #3987, opened 2025-06-16
Limit python until fix 

zm711:
Replace: instead of #3364

Seems like hope is fix will be with 3.13.5 but our tests are installing 3.13.4.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ authors = [
 ]
 description = "Python toolkit for analysis, visualization, and comparison of spike sorting output"
 readme = "README.md"
-requires-python = ">=3.9,<3.14"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -7,7 +7,7 @@ authors = [
 ]
 description = "Python toolkit for analysis, visualization, and comparison of spike sorting output"
 readme = "README.md"
-requires-python = ">=3.9,<3.14"
+requires-python = ">=3.9,<3.13.4"
```
zm711:
```suggestion
requires-python = ">=3.9,<3.13.4" # limit until Windows bug is fixed.
```

zm711:
@alejoe91 you good with Sam's fix here? We always merge this and then do your fix if you think it is more robust?

PR #3988, opened 2025-06-16
Fix last mock call to slurm launcher

alejoe91:


zm711 (approved):
Works for me. I guess my only question would be the benefit of adding the folder path is that we test that the file goes to the correct place in the give tempdir rather than just check that the script is in a random place?

alejoe91:
Yes, but Sam's fix is also ok but this checks the folder location too

PR #3990, opened 2025-06-17
Propagate `relative_to` to `dump_to_pickle`

chrishalcrow:
We don't propagate `relative_to` to `dump_to_pickle` in the `dump` method. This PR does.

The lack of propagation has a few consequences, including the fact that `recording.save_to_folder` and `analyzer.save(folder='...')` saves different recording info if `pickle`d (but not if `json`d).

Pretty sure it's a bug, but maybe there's a reason for this? 

Added a test to check that `pickle`ing and `json`ing are (almost) the same for a saved-then-loaded recording.


alejoe91:
Nice catch! I always use the `dump_to_pickle` directly, so probably it went unniticed!

samuelgarcia:
bien joué!

PR #3993, opened 2025-06-18
First pass of deprecation removals for 0.103.0

zm711:
I just did a search of the code base for all our dep warnings for removals in 0.103.0 and removed, updated some docs related to the deps and fixed tests that I could find. I probably missed something so I'll keep it on draft for the moment.

### DIFF
--- a/doc/get_started/quickstart.rst
+++ b/doc/get_started/quickstart.rst
@@ -336,7 +336,7 @@ Alternatively we can pass a full dictionary containing the parameters:
 
     # parameters set by params dictionary
     sorting_TDC_2 = ss.run_sorter(
-        sorter_name="tridesclous", recording=recording_preprocessed, output_folder="tdc_output2", **other_params
+        sorter_name="tridesclous", recording=recording_preprocessed, folder="tdc_output2", **other_params
     )
     print(sorting_TDC_2)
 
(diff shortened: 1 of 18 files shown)


zm711:
What I haven't changed 1) our the waveform extractor combability. It was listened as maybe remove for 0.103.0 but I say keep--maybe to discuss at a maintenance meeting
2) the new wait to do automerge. (Selfishly I'm using the old api :) )

zm711:
I would like to suggest as we update deprecations (myself included) that we try to add a comment in the tests as well with the deprecation version to make it easier to find. I tried searching based on deprecation term, but missed a few in testing as we can see from the failed testing. I think the extra effort will speed this along moving forward :)

zm711:
Poor hugging-face :(

zm711:
IBL testing is now failing. @alejoe91 I remember we had this issue before with some sort of versioning. Do you know anything about this?

samuelgarcia:
C'est un beau menage de primtemps.

I am OK for merging soon.

alejoe91:
@zm711 ok to merge?

zm711:
Yes I would say merge. In a separate PR I need to 
1) update the deprecation for mockwaveform extractor and auto_merge
2) give deprecations numbers to all functions without current version warning

But I think both of those things can be separate. It is a real pain to make these dep PRs too big because the testing burden is huge. So I think small chunks are better.

PR #3995, opened 2025-06-18
Optional loading of compression ratio in zarr recording

alejoe91:
The compression ratio can take a very long time to load for long recordings, since the size of each zarr chunk is checked. Since this is not something normally useful, this PR adds a `load_compression_ratio` argument and sets it to False by default


samuelgarcia:
ok

samuelgarcia (approved):

PR #4000, opened 2025-06-19
PreprocessingPipeline: support remote zarr in load_from_analyzer

alejoe91:


chrishalcrow:
Doesn't break my local zarr stuff, and LGTM.

chrishalcrow (approved):

src/spikeinterface/preprocessing/pipeline.py
```diff
@@ -183,11 +183,11 @@ def get_preprocessing_dict_from_analyzer(analyzer_folder, format="auto", backend
...
         The preprocessing dict extracted from the analyzer's recording.
     """
-
-    analyzer_folder = Path(analyzer_folder)
+    if not is_path_remote(analyzer_folder):
+        analyzer_folder = Path(analyzer_folder)
 
     if format == "auto":
-        if analyzer_folder.suffix == ".zarr":
+        if str(analyzer_folder).endswith(".zarr"):
```
zm711, line 190: Why isn't suffix working? Just for my own knowledge.

chrishalcrow: `suffix` is a method of the `Path` object, but we need to use `str` for some remote reason.

alejoe91: Casting a remote path to Path doesn't work ;)

zm711: Thanks I don't do truly remote path work so I didn't know. But it is good to know ! Thanks!

PR #4005, opened 2025-06-23
Allow `run_sorter` to accept dicts

chrishalcrow:
Allows `run_sorter` to accept dictionaries of recordings, and return a dict of sorters.

Longer-term aim, make this workflow just work:

``` python
pp_recs = si.common_reference(rec.split_by(‘group’))
sortings = si.run_sorter(pp_rec)
analyzer = si.create_sorting_analyzer(sortings, pp_recordings)
```

Think this PR is minimally invasive. The old function `run_sorter` can now accept a dict. If it gets a dict, it calls `run_sorter` for each recording and saves in a folder

```
{folder}/
    spikeinterface_info.json
    splitting_key_0/
      {whatever run_sorter spits outs}
    splitting_key_1
        {whatever run_sorter spits outs}
    …
```

And `spikeinterface_info.json` looks like e.g.:

``` python
{
    "version": "0.102.4",
    "dev_mode": true,
    "object": "Group[SorterOutput]",
    "dict_keys": [
        "a",
        5
    ],
}
```

Tried a bigger refactor here: https://github.com/chrishalcrow/spikeinterface/tree/sort-by-group which avoids the recurrent logic here, but it was a mess. This seems a lot nicer.


alejoe91:
@chrishalcrow maybe worth saving the `dict_keys` dtype in the spikeinterface_info since this gets lost when dumping to json

samuelgarcia:
Thank you Chris, this is cool.

We also need to implement the spikeinterface.load that handle to open this new folder type.

We should discuss a bit more the "object": "dict of BaseSorting"
I add in mind something like "Group[Sorting]"

Also not sure to clearly understand why we need this "split_by_property" in the dict because the information is dump in all sub-folder in recording.json no ?

> @chrishalcrow maybe worth saving the `dict_keys` dtype in the spikeinterface_info since this gets lost when dumping to json

Not sure I like this ugly hack in the format spikeinterface_info.json.
List of keys can be int or str no ? json handle it.
The list is in the dict to open back the sub folder in the good order and will converted to str for folder.
So I do not see why.


I would be happy to have a call.

src/spikeinterface/core/loading.py
```diff
@@ -244,6 +257,16 @@ def _load_object_from_folder(folder, object_type, **kwargs):
                     si_file = f
         return BaseExtractor.load(si_file, base_folder=folder)
 
+    elif object_type.startswith("Group"):
```
chrishalcrow, line 260: Feedback on these lines of code very welcome! Wasn't sure best way to proceed.

src/spikeinterface/sorters/basesorter.py
```diff
@@ -344,6 +344,17 @@ def get_result_from_folder(cls, output_folder, register_recording=True, sorting_
             if recording is not None:
                 sorting.register_recording(recording)
 
+            if recording.get_annotation("split_by_property") is not None:
```
chrishalcrow: Hello, I now think this is bad idea. The user might `split_by` to do some preprocessing, then aggregate the recording back together, then sort. So we don't know that the split_by_property was used to do the split sorting.

Maybe better to not mess with BaseSorter, and deal with this at the Analyzer creation? Since that's where we merge Recording and Sorting information?

src/spikeinterface/core/loading.py
```diff
@@ -244,6 +257,16 @@ def _load_object_from_folder(folder, object_type, **kwargs):
...
+
+        sub_object_type = object_type.split("[")[1].split("]")[0]
+        with open(folder / "spikeinterface_info.json", "r") as f:
+            spikeinterface_info = json.load(f)
+        group_keys = spikeinterface_info.get("dict_keys")
+
+        group_of_objects = {key: _load_object_from_folder(folder / str(key), sub_object_type) for key in group_keys}
```
samuelgarcia, line 267: very clean. More than in my dreams.

doc/how_to/process_by_channel_group.rst
```diff
@@ -134,23 +135,46 @@ back together under the hood).
...
-**Option 1: Manual splitting**
+**Option 1 : Automatic splitting**
 
-In this example, similar to above we loop over all preprocessed recordings that
+Simply pass the split recording to the `run_sorter` function, as if it was a non-split recording.
+This will return a dict of sortings, with the keys corresponding to the groups.
+
+.. code-block:: python
+
+    split_recording = raw_recording.split_by("group")
```
samuelgarcia, line 157:
```suggestion
    split_recording = raw_recording.split_by("group")
    # is a dict of recordings
```

src/spikeinterface/sorters/basesorter.py
samuelgarcia: yep, you are rigth, lets discuss a bit more on this

src/spikeinterface/sorters/runsorter.py
```diff
@@ -201,6 +214,56 @@ def run_sorter(
...
+
+    info_file = folder / "spikeinterface_info.json"
+    info = dict(
+        version=spikeinterface.__version__,
+        dev_mode=spikeinterface.DEV_MODE,
+        object="Group[SorterFolder]",
+        dict_keys=list(dict_of_recordings.keys()),
+    )
+    with open(info_file, mode="w") as f:
+        json.dump(check_json(info), f, indent=4)
```
samuelgarcia, line 240: Maybe this should be after the loop. So the run_sorter fails this is fail is not created optimisticly too early.

src/spikeinterface/sorters/basesorter.py
chrishalcrow: Super. I don't think we need to set the unit properties on load. On load, you get a dict whose keys tell you the split. I think we should only set unit properties to these keys at the `unit_aggregation` stage, since that is the point that you would lose the dictionary structure.

src/spikeinterface/core/loading.py
chrishalcrow: Couldn't resist that one liner

samuelgarcia:
bien joué

samuelgarcia (approved):

alejoe91 (approved):

src/spikeinterface/sorters/runsorter.py
```diff
@@ -201,6 +214,50 @@ def run_sorter(
...
+        folder = Path(sorter_name + "_output")
+
+    folder = Path(folder)
+    folder.mkdir(exist_ok=remove_existing_folder)
+
+    sorter_dict = {}
+    for group_key, recording in dict_of_recordings.items():
+
+        if "recording" in run_sorter_params:
+            run_sorter_params.pop("recording")
```
alejoe91: How can this happen? `recording` is an input argument for the `run_sorter` function so it shouldn't end up in the `run_sorter_params`, right?

chrishalcrow: `recording` is included in the `common_kwargs` of `run_sorter` here:

https://github.com/SpikeInterface/spikeinterface/blob/0a832dc5eb6125297df3ae7f3548ffac6b4eb92f/src/spikeinterface/sorters/runsorter.py#L142

And these are passed to `_run_sorter_by_dict` as the `run_sorter_params` (with some additional args). So you need to pop this. Might make more sense to `pop` while we're constructing the `run_sorter_params`

src/spikeinterface/sorters/runsorter.py
```diff
@@ -151,6 +150,20 @@ def run_sorter(
...
+
+        all_kwargs = common_kwargs
+        all_kwargs.update(
+            dict(
+                docker_image=docker_image,
+                singularity_image=singularity_image,
+                delete_container_files=delete_container_files,
+            )
+        )
+
```
chrishalcrow, line 154: @alejoe91 could `pop` here!

src/spikeinterface/sorters/runsorter.py
chrishalcrow: Can make things more elegant by changing name of argument in `_run_sorter_by_dict` from `dict_of_recordings` to `recording`. But I think it makes the code less clear and less explicit.

chrishalcrow: I moved the `pop`, and still think it's worth it for readability to keep the `dict_of_recordings` argument (which is why we need the pop in the first place)

alejoe91 (approved):

PR #4008, opened 2025-06-24
Allow white matter to pass gain_to_uV

h-mayorquin:
Related to #3964

### DIFF
--- a/src/spikeinterface/extractors/matlabhelpers.py
+++ b/src/spikeinterface/extractors/matlabhelpers.py
@@ -19,7 +19,6 @@ else:
 
 
 class MatlabHelper:
-    mode = "file"
     installation_mesg = (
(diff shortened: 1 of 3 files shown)


pauladkisson (approved):
Looks good

zm711:
If there will always be an xml file (and Paul's example was someone who accidentally threw away the xml) it seems like we should make a proper extractor, no? I think this is fine for now. But this might be the case where we parse the xml if we can and if not we tell people to just use binary recording extractor.

h-mayorquin:
Yes, we have all the files now. It would be good to implement parsing of the metadata from xml if those files are available but at the moment I don't have time and it is better to enable this feature earlier rather than later.

alejoe91 (approved):

PR #4012, opened 2025-06-25
Add `get_unit_spike_train_in_seconds` to sorting extractor

h-mayorquin:
As agreed on the previous meeting. This will avoid the conversion from times - frames - times for extractors that handle time natively.

This PR implements this for NWBSortingExtractor. Once we agree and we merge I will do this for the other neo extractors that also handle time natively (e.g. BlackRock).

### DIFF
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -269,6 +269,8 @@ class BaseSorting(BaseExtractor):
             return spike_times
 
         # Use the native spiking times if available
+        # Some instances might implement a method themselves to access spike times directly without having to convert
+        # (e.g. NWB extractors)
         if hasattr(segment, "get_unit_spike_train_in_seconds"):
             return segment.get_unit_spike_train_in_seconds(unit_id=unit_id, start_time=start_time, end_time=end_time)
 


src/spikeinterface/core/basesorting.py
```diff
@@ -162,17 +162,100 @@ def get_unit_spike_train(
...
+            unit_id=unit_id,
+            segment_index=segment_index,
+            start_frame=start_frame,
+            end_frame=end_frame,
+            return_times=False,
+            use_cache=True,
+        )
+
+        # Convert frames to times
+        if self.has_recording():
```
h-mayorquin: Maybe this should go first.

src/spikeinterface/core/basesorting.py
```diff
@@ -162,17 +162,100 @@ def get_unit_spike_train(
...
+                start_time=start_time,
+                end_time=end_time,
+            )
         else:
             return spike_frames
 
+    def get_unit_spike_train_in_seconds(
+        self,
+        unit_id: str | int,
+        segment_index: Union[int, None] = None,
```
h-mayorquin: probably better as Optional.

src/spikeinterface/core/basesorting.py
```diff
@@ -162,18 +187,123 @@ def get_unit_spike_train(
...
-                return times[spike_frames]
-            else:
+            # Use the new get_unit_spike_train_in_seconds method for better precision
+            start_time = None
+            end_time = None
+
+            if start_frame is not None:
+                segment = self._sorting_segments[segment_index]
+                t_start = segment._t_start if segment._t_start is not None else 0
+                start_time = start_frame / self.get_sampling_frequency() + t_start
```
samuelgarcia: In case of gaps in the times vector the boundaries frame would be wrong no ?
I guess we do not care so much. But theorically we should use the times vector when available here.
Am I wrong ?

src/spikeinterface/core/basesorting.py
```diff
@@ -162,18 +187,123 @@ def get_unit_spike_train(
...
+            spike_frames = self.get_unit_spike_train(
+                unit_id=unit_id,
+                segment_index=segment_index,
+                start_frame=start_frame,
+                end_frame=end_frame,
+                return_times=False,
+                use_cache=True,
+            )
+
+            recording_times = self.get_times(segment_index=segment_index)
```
samuelgarcia: We there is not time vector, this is an enormous operation.
We should handle, the case t_start+sample_rate in a separate if to avoid this enormous times vector generation.
I am aware that it was the case before but lets do it properly here. 
No ?

src/spikeinterface/core/basesorting.py
```diff
@@ -162,18 +187,123 @@ def get_unit_spike_train(
...
+            spike_times = recording_times[spike_frames]
+            if start_time is not None:
+                spike_times = spike_times[spike_times >= start_time]
+            if end_time is not None:
+                spike_times = spike_times[spike_times <= end_time]
+
+            return spike_times
+
+        # Use the native spiking times if available
+        if hasattr(segment, "get_unit_spike_train_in_seconds"):
```
samuelgarcia, line 274: Cool.

samuelgarcia:
Merci.
Almost OK for me but lets avoid to geenrate the times vector when no necessary.

src/spikeinterface/core/basesorting.py
h-mayorquin: I think you are right, Sam. Check the latest changes, I centralized the logic to convert from samples to time and vice-versa so we always take into account if they are a registered recording and if the registered recording has a time vector.

src/spikeinterface/core/basesorting.py
h-mayorquin: Yes, this is a good point. I changed this so the tiemstamps are not created if there is no time vector.

src/spikeinterface/core/basesorting.py
```diff
@@ -162,21 +187,113 @@ def get_unit_spike_train(
...
-                spike_times = spike_frames / self.get_sampling_frequency()
-                return t_start + spike_times
+            start_time = (
+                self.sample_index_to_time(start_frame, segment_index=segment_index) if start_frame is not None else None
+            )
+            end_time = (
+                self.sample_index_to_time(end_frame, segment_index=segment_index) if end_frame is not None else None
+            )
+
+            return self.get_unit_spike_train_in_seconds(
```
samuelgarcia: In the new design, if we do return_times=True then spikes are read twice.
line 186 and then here lin 197.
Am I wrong.
We should reorganize the if else logic.

src/spikeinterface/core/basesorting.py
```diff
@@ -162,21 +187,113 @@ def get_unit_spike_train(
...
+        -------
+        spike_times : np.ndarray
+            Spike times in seconds
+        """
+        segment_index = self._check_segment_index(segment_index)
+        segment = self._sorting_segments[segment_index]
+
+        # If sorting has a registered recording, get the frames and get the times from the recording
+        # Note that this take into account the segment start time of the recording
+        if self.has_recording():
```
samuelgarcia, line 247: So having a recording has precedance on get_unit_spike_train_in_seconds.
Maybe we should revert no ?

samuelgarcia:
Hi @h-mayorquin 
Sorry I was not focus enought yesterday. I found two other stuff to discuss.

src/spikeinterface/core/basesorting.py
h-mayorquin: No, I think this is correct. People do the registering to set the timestamps, right? I think we should respect that intent. That's why I organized in these order of priorities:

1) If you set the recording with its timestamps, do this.
2) Get the native timestamps if they are available through the new method.
3) Fallback to get them from sampling.

src/spikeinterface/core/basesorting.py
h-mayorquin: Yes, you are right, I changed this.

src/spikeinterface/core/basesorting.py
```diff
@@ -161,22 +201,99 @@ def get_unit_spike_train(
...
+            segment_index=segment_index,
+            start_frame=start_frame,
+            end_frame=end_frame,
+            return_times=False,
+            use_cache=True,
+        )
+
+        t_start = segment._t_start if segment._t_start is not None else 0
+        spike_times = spike_frames / self.get_sampling_frequency()
+        return t_start + spike_times
```
alejoe91, line 293: I wonder if it makes sense to port this as a base imlementation of `BaseSortingSegment.get_unit_spike_train_in_seconds`. This way, you could skip the `hasattr()` logic, right?
@h-mayorquin do you think it would improve readibility? or did you implement the `hasattr` on purpose to highlight that the segment has its own implementation?

h-mayorquin: I am not sure it would improve readability (and the "I am not sure" is not rhetorical). But yes, the idea of the implementation is that the different extractors and their segments might or might not have implemented this. So yes, is on purpose. Maybe a comment would not hurt.

src/spikeinterface/core/basesorting.py
```diff
@@ -131,12 +131,52 @@ def get_total_duration(self) -> float:
...
+            )
+
+            return self.get_unit_spike_train_in_seconds(
+                unit_id=unit_id,
+                segment_index=segment_index,
+                start_time=start_time,
+                end_time=end_time,
+            )
+
         segment_index = self._check_segment_index(segment_index)
```
samuelgarcia, line 180: This should be moved line 164 no ?

src/spikeinterface/core/basesorting.py
samuelgarcia: you are right.

src/spikeinterface/core/basesorting.py
samuelgarcia: @alejoe91 I think I prefer the implementation here.

samuelgarcia (approved):

samuelgarcia:
Parfait!
C'est tout bon.

PR #4015, opened 2025-06-30
Update/Fix for MaxWell systems

cmccrimmon:
Updated for current MaxWell systems - now utilize events to extract MAXWELL SPIKE (via built in template matching) and STIMULATION events. Also remove 1) instances where multiple MaxWell electrodes are connected to the same channel (mix of electrode signals) and 2) duplicates where multiple MaxWell channels and connected to the same electrode (duplicates) as discussed in https://github.com/NeuralEnsemble/python-neo/pull/1703.


samuelgarcia:
Hi.
Thank you for this. I will have a look soon.
As a very general first comments I would advice to avoid using pandas in data reader.

PR #4021, opened 2025-07-01
Add Axon recording

h-mayorquin:
So besides intracellular data Axon also handles extracellular experiments:

https://github.com/catalystneuro/neuroconv/issues/1410

I am porting this as an extractor to make it available on SpikeInterface. This will allow people to analyze data and it will allow us in neuroconv to convert data through the usual pipeline.

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/__init__.py
+++ b/src/spikeinterface/extractors/neoextractors/__init__.py
@@ -1,4 +1,5 @@
 from .alphaomega import AlphaOmegaRecordingExtractor, AlphaOmegaEventExtractor, read_alphaomega, read_alphaomega_event
+from .axon import AxonRecordingExtractor, read_axon
 from .axona import AxonaRecordingExtractor, read_axona
 from .biocam import BiocamRecordingExtractor, read_biocam
 from .blackrock import BlackrockRecordingExtractor, BlackrockSortingExtractor, read_blackrock, read_blackrock_sorting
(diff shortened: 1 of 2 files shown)


samuelgarcia:
> So besides intracellular data Axon also handles extracellular experiments:

Really ? I hope that the channel wil not be to high...




OK for me.

h-mayorquin:
No small enough to add to gin.

zm711:
And so the neo test is only intracellular?
 ( I haven't checked the neo test myself yet).

h-mayorquin:
The neo tests don't have provenance data but we are using them as intracellular.

zm711:
Sorry one last question on my end. If we aren't sorting this at the neo level and we aren't checking here how can we ensure the stream the users is using is actually extracellular data?

h-mayorquin:
We can't. See the discussion in neuroconv linked at the top.

h-mayorquin:
Ok, we just got data. I added a test

src/spikeinterface/extractors/tests/test_neoextractors.py
```diff
@@ -294,6 +295,12 @@ class TdTRecordingTest(RecordingCommonTestSuite, unittest.TestCase):
     entities = [("tdt/aep_05", {"stream_id": "1"})]
 
 
+class AxonRecordingTest(RecordingCommonTestSuite, unittest.TestCase):
+    ExtractorClass = AxonRecordingExtractor
+    downloads = ["axon"]
+    entities = ["axon/extracellular_data/four_electrodes/24606005_SampleData.abf"]
```
zm711, line 301: And this only has one stream?

alejoe91 (approved):

src/spikeinterface/extractors/tests/test_neoextractors.py
h-mayorquin: Yes, otherwise that test would not work.

PR #4025, opened 2025-07-02
Fix soft vs hard merge issue in analyzer

zm711:
Currently the code doesn't actually care if you are doing a soft or hard merge, it verifies the sparsity overlap for all units whether soft or hard and then errors even on hard merge if there is not an overlap. Confirmed by a labmate who ran

```python
merged_analyzer = analyzer.merge_units(
    new_merges,
    merging_mode='hard'
    sparsity_overlap = 0.75
)
```
and got an error
image

To be honest I don't understand the merging mechanism well enough to know just setting sparsity to None for a hard merge is desired so I'm submitting this as a reminder that we should probably fix this before the next version. Sorry I don't really know how to make a MRE for this.

so help writing a test would be great.

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -960,6 +960,7 @@ class SortingAnalyzer:
             mergeable, masks = self.are_units_mergeable(
                 merge_unit_groups,
                 sparsity_overlap=sparsity_overlap,
+                merging_mode=merging_mode,
                 return_masks=True,
             )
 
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -967,11 +967,19 @@ def _save_or_select_or_merge(
...
+                        # sparsity and instead use a new sparsity
+                        if merging_mode == "soft":
+                            raise Exception(
+                                f"The sparsity of {merge_unit_group} do not overlap enough for a soft merge using "
+                                f"a sparsity threshold of {sparsity_overlap}. You can either lower the threshold or use "
+                                "a hard merge."
+                            )
+                        else:
+                            sparsity = None
+                            break
```
zm711: To make sure people read carefully this is my main question? is this what a "hard" merge should do for sparsity?

samuelgarcia:
@yger can you help on this ?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -960,20 +960,22 @@ def _save_or_select_or_merge(
             mergeable, masks = self.are_units_mergeable(
                 merge_unit_groups,
                 sparsity_overlap=sparsity_overlap,
+                merging_mode=merging_mode,
```
alejoe91, line 963: @zm711 this should fix it. The implementation of this function is as follows:
- if "soft": return intersection of sparsity and check mergeability
- if "hard": return union and set mergeable to True

alejoe91:
@zm711 can you test if my last ocmmit fixed it? :)

zm711:
As it is a holiday here the person who discovered this in lab won’t be around. I can test on Monday with the dataset that triggered this.

src/spikeinterface/core/sortinganalyzer.py
zm711: I wasn’t sure if in hard mode we wanted to test to see if soft was possible and switch to soft if possible and then only do hard if required.

alejoe91:
> As it is a holiday here the person who discovered this in lab won’t be around. I can test on Monday with the dataset that triggered this.

Oh right! Happy Independence Day and stop looking at the screen! Otherwise @samuelgarcia will need to intervene

zm711:
Tested and worked.

alejoe91 (approved):

Issue #4016, opened 2025-06-30
IBL tests failing


zm711:
Seems we have a problem with python 3.13 for the ibl tests from One. We had this previously with a version issue. Maybe @oliche you could comment on what has changed recently?
```
=================================== FAILURES ===================================
______________ TestIblSortingExtractor.test_ibl_sorting_extractor ______________

self = <test_iblextractors.TestIblSortingExtractor testMethod=test_ibl_sorting_extractor>

    def test_ibl_sorting_extractor(self):
        """
        Here we generate spike train with 3 clusters in a very basic ALF format\
        and read it with the spikeinterface extractor
        """
        from one.api import ONE
    
        try:
            one = ONE(
                base_url="https://openalyx.internationalbrainlab.org/",
                ***,
                silent=True,
                cache_dir=None,
            )
        except:
            pytest.skip("Skipping test due to server being down.")
        sorting = read_ibl_sorting(pid=PID, one=one, revision="2023-12-05")
>       assert len(sorting.unit_ids) == 733
E       assert 1091 == 733
E        +  where 1091 = len(array([   0,    1,    2, ..., 1088, 1089, 1090], shape=(1091,)))
E        +    where array([   0,    1,    2, ..., 1088, 1089, 1090], shape=(1091,)) = IblSortingExtractor: 1091 units - 1 segments - 30.0kHz.unit_ids

src/spikeinterface/extractors/tests/test_iblextractors.py:205: AssertionError
```

alejoe91:
Tests seem to be failing with `ibllib==3.4.0`. This is a bit strange because the `revision="2023-12-05"` argument that we use in the tests [here](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/extractors/tests/test_iblextractors.py#L204-L207) should ensure we always download the same spike sorting oobject. @oliche any guidance?

h-mayorquin:
Yes, I thought that was the point of the revision.

Should we pin down temporarily back to 3.3.1 for the sake of the CI? Let me test if that at least fixes it to add have a better picture.

h-mayorquin:
Yes, it seems that pinning ibllib to the previous version works so the revision is not working as expected?

oliche:
Ok, looking into this now

oliche:
I have pushed a fix to ibllib and updated the requirements here #4026

PR #4026, opened 2025-07-02
Fix IBL tests using ibllib hotfix

oliche:
Hopefully this will address #4016

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,13 +69,13 @@ extractors = [
     "sonpy;python_version<'3.10'",
     "lxml", # lxml for neuroscope
     "scipy",
-    "ibllib==3.3.1", # streaming IBL
+    "ibllib>=3.4.1;python_version>='3.10'", # streaming IBL
     "pymatreader>=0.0.32", # For cell explorer matlab files
     "zugbruecke>=0.2; sys_platform!='win32'", # For plexon2
 ]
 
 streaming_extractors = [
-    "ibllib==3.3.1", # streaming IBL
+    "ibllib>=3.4.1;python_version>='3.10'", # streaming IBL
     # Following dependencies are for streaming with nwb files
     "pynwb>=2.6.0",
     "fsspec",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -69,13 +69,13 @@ extractors = [
...
     "scipy",
-    "ibllib==3.3.1", # streaming IBL
+    "ibllib>=3.4.1;python_version>='3.10'", # streaming IBL
     "pymatreader>=0.0.32", # For cell explorer matlab files
     "zugbruecke>=0.2; sys_platform!='win32'", # For plexon2
 ]
 
 streaming_extractors = [
-    "ibllib==3.3.1", # streaming IBL
+    "ibllib>=3.4.1;python_version>='3.10'", # streaming IBL
```
zm711, line 78: @alejoe91 I'm trying to remember if in our testing we autoskip ibl stuff for < 3.10. Until we officially drop 3.9 we might not want to force people that want to use other extractors etc to be limited to 3.10. But I can't remember so I'll let Alessio give the final word on this.

zm711 (commented):
Thanks @oliche! Appreciate the quick turn around. I think Alessio or Sam should comment on if we want a python version limit included for sub-packages.

h-mayorquin (approved):
LGTM.

Thanks for the fix.

pyproject.toml
h-mayorquin: We are only testing from 3.10 onwards at the moment.  This will not limit people using other extractors but will fail for someone trying to install this in python 3.9.

zm711: My point is what if someone wants to install 3.9 with extractors. We shouldn't block them for now right? And then when we drop 3.9 whenever then the limit won't matter anyway.

h-mayorquin: But this should not block them.

zm711: Okay cool. I haven't check the dependency resolving well enough to know this wouldn't lead to dependency issues.

PR #4031, opened 2025-07-04
Try not splitting numpy version based on python version

chrishalcrow:
Checking if tests pass when numpy version is not determined by python version.

If tests pass, would mean that `uv` can cleanly install `spikeinterface[full]`, and would help with environments which contain spikeinterface and packages which only support numpy<2.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,8 +20,7 @@ classifiers = [
 
 
 dependencies = [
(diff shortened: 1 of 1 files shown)


chrishalcrow:
Windows with python 3.13 installs numpy verison 1.26.4, which causes a failure.
So, for now, splitting is needed.

samuelgarcia:
There must be a better way...

PR #4033, opened 2025-07-05
More deprecation updates for version 0.103.0

zm711:
1) Add deprecations of version 0.104.0 for functions that had no version number
2) removed deprecation version from `MockWaveformExtractor` and indicated it could happen in the future
3) Added a couple question for a few deps I wasn't sure about
4) Found two more deprecations (one from version 0.101.0 and one from 0.102.0) that I removed
5) @h-mayorquin for `return_scaled` we discussed a longer deprecation than 2 versions but right now it just says "future" just wanted this on our radar and explain why I didn't touch those. Maybe we want to discuss at some point what a long deprecation means for us? Or leave it as future like for WaveformExtractor

I didn't double check testing so I assume this will break tests. Will fix tomorrow as I need to run.


h-mayorquin:
> @h-mayorquin for return_scaled we discussed a longer deprecation than 2 versions but right now it just says "future" just wanted this on our radar and explain why I didn't touch those. Maybe we want to discuss at some point what a long deprecation means for us? Or leave it as future like for WaveformExtractor

The discussion on the meeting was "Never" but I think to be precise you could say in the next major version release.

zm711:
> The discussion on the meeting was "Never" but I think to be precise you could say in the next major version release.

That works for me. Just wanted to confirm.

src/spikeinterface/benchmark/benchmark_plot_tools.py
```diff
@@ -374,6 +374,7 @@ def plot_agreement_matrix(study, ordered=True, case_keys=None, axs=None):
     return fig
 
 
+# what's the dperecation strategy for this function in general?
```
zm711: @samuelgarcia  / @alejoe91  this is for you both!

src/spikeinterface/comparison/multicomparisons.py
```diff
@@ -190,6 +190,7 @@ def get_agreement_sorting(self, minimum_agreement_count=1, minimum_agreement_cou
         )
         return sorting
 
+    # strategy for this dep?
```
zm711: same here and just below. What is the strategy?

zm711 (commented):
@alejoe91 

my review for this one never posted! Could you look at these two things to finalize this PR

src/spikeinterface/comparison/multicomparisons.py
alejoe91: I think this warning has been there for a long time :)

IMO we can remove the functions in 0.103.0 directly.. @samuelgarcia what do you think?

src/spikeinterface/benchmark/benchmark_plot_tools.py
alejoe91: the benchmark module is new and in dev, so I think we can just remove the function in the next release

src/spikeinterface/comparison/multicomparisons.py
h-mayorquin: +1

src/spikeinterface/benchmark/benchmark_plot_tools.py
samuelgarcia: yes we can remove this

src/spikeinterface/comparison/multicomparisons.py
samuelgarcia: ok

src/spikeinterface/sortingcomponents/peak_pipeline.py
```diff
@@ -16,10 +16,14 @@ def run_peak_pipeline(
     folder=None,
     names=None,
 ):
-    # TODO remove this soon
+    # TODO remove this soon. Will be removed in 0.104.0
```
samuelgarcia: I think we can remove this now. Nody never used it.

samuelgarcia:
I am OK with evrything + one that can be removed now

alejoe91:
Thanks @zm711!

Let's merge :)

PR #3923, opened 2025-05-15
Components merge templates

samuelgarcia:

PR #4036, opened 2025-07-07
Templates.is_scaled > Templates.is_in_uV

samuelgarcia:
This needed to be done after #3794.
At the moement is is a simple replace in the entire code.
There are many return_scaled to be changed in return_in_uV but @zm711 already started to change some in #4033.
So lets wait #4033 to be merged first and then lets continue working on this.

I think we will need to make some logic for readi backward compatible Templates saved before 0.103.0


zm711:
I was just commenting on this deprecation. We hadn't officially discussed deprecation strategy for `scaled` -> `is_in_uV`, which I think we should. Agree with backward compatability though! It was a pain once when we did a dep and all my analyzers started erroring. I forget what it was. nbefore maybe?

h-mayorquin:
This is great, thanks Sam.

src/spikeinterface/core/template.py
```diff
@@ -31,7 +31,7 @@ class Templates:
         Array of unit IDs. If `None`, defaults to an array of increasing integers.
     probe: Probe, default: None
         A `probeinterface.Probe` object
-    is_scaled : bool, optional default: True
+    is_in_uV : bool, optional default: True
```
alejoe91, line 34: Let's make sure that in the `Templates.from_zarr_group` we check for the existence of `is_scaled` for back-compatibility (e.g. with hybrid library)

samuelgarcia: DOne

alejoe91 (approved):

samuelgarcia:
I propose to merge this now because #3923 will need some fixes after, so the sooner the better.
No ?

PR #4037, opened 2025-07-07
Allow `create_sorting_analyzer` to accept dicts

chrishalcrow:
Allows `create_sorting_analyzer` to accept a dict of recordings and a dict of sortings. Now the following code...

``` python
raw_rec = si.read_openephys("recording_path") 
grouped_rec = raw_rec.split_by("group")
pp_rec = si.bandpass_filter(grouped_rec)
sort = si.run_sorter("mountainsort5", pp_rec)
analyzer = si.create_sorting_analyzer(sort, pp_rec)
```

...works for both grouped and un-grouped recordings! So the user doesn't have to include any logic about grouping in their pipeline for e.g. Neuropixel 2.0 recordings.

For now, I've implemented the simple solution: internally the `create_sorting_analyzer` function aggregates the recordings and sortings and combines them into a single sorting analyzer. The actual code inside `create_sorting_analyzer` is _super_ simple! Also updated the `units_aggregation` function to accept dicts.

(We've also discussed allowing it to output a dictionary of analyzers. This is slightly more painful: the user will have to write awkward code when they want to `compute`, or `export`. We'll need to write and test new `save` and `load` options etc. I think we should add this in the future if there's a need for it.)

Internally, we need to keep track of the splitting and aggregation (e.g. which units in the analyzer belonged to which sorting?). I've put that logic in the `aggregate_channels` and `aggregate_units` functions, which attach new properties to the channels and units depending on the underlying group called "aggregate_channels_key" and "aggregate_units_key". Happy to hear other name ideas! So you can get the grouped analyzer by doing something like:

``` python
  group_0_keys = analyzer.unit_ids[analyzer.get_sorting_property("aggregate_units_key") == 0]
  group_0_analzyer = analyzer.select_units(unit_ids = group_0_keys)
```

Docs: https://spikeinterface--4037.org.readthedocs.build/en/4037/how_to/process_by_channel_group.html#
Plus a comment here: https://spikeinterface--4037.org.readthedocs.build/en/4037/modules/sorters.html#spike-sorting-by-group

### DIFF
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -658,8 +658,8 @@ class BaseSorting(BaseExtractor):
             sortings = {}
         for value in np.unique(values).tolist():
             (inds,) = np.nonzero(values == value)
-            new_channel_ids = self.unit_ids[inds]
-            subsort = self.select_units(new_channel_ids)
+            new_unit_ids = self.unit_ids[inds]
+            subsort = self.select_units(new_unit_ids)
             subsort.set_annotation("split_by_property", value=property)
             if outputs == "list":
                 sortings.append(subsort)


samuelgarcia:
Merci beaucoup.
I will read this in details.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -133,6 +132,31 @@ def create_sorting_analyzer(
     In some situation, sparsity is not needed, so to make it fast creation, you need to turn
     sparsity off (or give external sparsity) like this.
     """
+
+    if isinstance(sorting, dict) & isinstance(recording, dict):
```
```suggestion
    if isinstance(sorting, dict) and isinstance(recording, dict):
```
I think it's more readable no?

doc/how_to/process_by_channel_group.rst
```diff
@@ -188,3 +191,29 @@ to different groups.
...
+
+
+The code above creates a *single* sorting analyzer called :code:`analyzer`. You can select the units
+from one of the "group"s as follows:
+
+.. code-block:: python
+
+    aggretation_keys = analyzer.get_sorting_property("aggregate_units_key")
+    group_0_keys = analyzer.unit_ids[aggretation_keys == 0]
+    group_0_analzyer = analyzer.select_units(unit_ids = group_0_keys)
```
alejoe91:
```suggestion
    units_group_0 = analyzer.unit_ids[aggretation_keys == 0]
    group_0_analzyer = analyzer.select_units(unit_ids = units_group_0)
```

alejoe91: I think that the naming here was a bit confusing

alejoe91 (commented):
A few small suggestions :)

samuelgarcia:
This is ultra smart and usefull.
The only improvement I would see is the handling of sparsity that would be group dependant.
For instance if you play with tetrodes you could expect the sparsity to be by group which is super fast to compute (instead of taking many random waveforms). 
This could be done line 145 in sorting_analyzer.py with maybe contorl by an option. Because for big group (shank of neuropixel) I think the sparsity from waveforms is still wanted no ?

samuelgarcia (approved):

doc/how_to/process_by_channel_group.rst
```diff
@@ -143,11 +145,12 @@ Sorting a Recording by Channel Group
 We can also sort a recording for each channel group separately. It is not necessary to preprocess
 a recording by channel group in order to sort by channel group.
 
-There are two ways to sort a recording by channel group. First, we can simply pass the output from
-our preprocessing-by-group method above. Second, for more control, we can loop over the recordings
+There are two ways to sort a recording by channel group. First, we can pass a dictionary to the
+`run_sorter` function. Since the preprocessing-by-group method above returns a dict, we can
```
zm711:
```suggestion
``run_sorter`` function. Since the preprocessing-by-group method above returns a dict, we can
```
missing double tick for code rendering : )

doc/how_to/process_by_channel_group.rst
```diff
@@ -188,3 +191,29 @@ to different groups.
...
             )
         sortings[group] = sorting
+
+
+Creating a SortingAnalyzer by Channel Group
+-------------------------------------------
+
+The code above generates a dictionary of recording objects and a dictionary of sorting objects.
+When making a :ref:`SortingAnalyzer <modules/core:SortingAnalyzer>`, we can pass these dictionaries and
+a single analyzer will be created, with the recordings and sortings appropriately aggregated.
```
zm711, line 202: How? Do keys of the dict need to match up? Will the user know how the sorting dict looks vs the recording dict?

zm711 (commented):
Two small points-- 1 a rendering issue 2 a discussion of if we should explain the dict layout better for beginners trying to use this strategy (at the level of inputs to the analyzer creation).

doc/how_to/process_by_channel_group.rst
samuelgarcia: Maybe we could check that keys are the same in the same order no ?

zm711: Yeah we either do that for the users internally or need to explain it here. So I think doing it ourselves is fine. Chris is doing a dict key comparison below but I haven't tested it to see if it checks the order or just the presence of the same keys.

chrishalcrow:
> The only improvement I would see is the handling of sparsity that would be group dependant.
> For instance if you play with tetrodes you could expect the sparsity to be by group which is super fast to compute (instead of taking many random waveforms).
> This could be done line 145 in sorting_analyzer.py with maybe contorl by an option. Because for big group (shank of neuropixel) I think the sparsity from waveforms is still wanted no ?

Hello. Hmm, I'm not sure how to handle this. Like you say: you defo want this for tetrodes. You probably don't want this for NP since you want an extra sparsity calc on each shank. You maybe want this for 32 channel silicon probes?? Ideally, you would compute sparsity per group, but this seems fairly complicated code-wise. Maybe we can add a note to the tetrode docs, and these docs, that if you're using tetrodes you can pass the sparsity explicitly? Maybe easiest to discuss "in person".

doc/how_to/process_by_channel_group.rst
chrishalcrow: One day I'll remember this, or do a docs build locally and actually check what I've written...

doc/how_to/process_by_channel_group.rst
chrishalcrow: Updated to hopefully make things clearer.

The order of keys doesn't matter: when you aggregate, the link between the recording channels and unit ids is independent of this dict stuff I'm adding. Added a test to check this is true.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -81,6 +82,8 @@ def create_sorting_analyzer(
         You can control `estimate_sparsity()` : all extra arguments are propagated to it (included job_kwargs)
     sparsity : ChannelSparsity or None, default: None
         The sparsity used to compute exensions. If this is given, `sparse` is ignored.
+    set_sparsity_by_dict_key : bool, default: False
```
samuelgarcia, line 91: I support this.
What about setting it to True ?

chrishalcrow: If it's False by default, some tetrode people will have un-sparsified tetrode bundles (not great, but ok)
If True by default, some silicon people will have badly-sparsified probes (worse!)

So I vote False!

samuelgarcia (approved):

chrishalcrow:
Updated docstring of `set_sparsity_by_dict_key` - now ready!

src/spikeinterface/core/basesorting.py
```diff
@@ -626,6 +626,47 @@ def time_slice(self, start_time: float | None, end_time: float | None) -> BaseSo
...
+        if values is None:
+            raise ValueError(f"property {property} is not set")
+
+        if outputs == "list":
+            sortings = []
+        elif outputs == "dict":
+            sortings = {}
+        for value in np.unique(values).tolist():
+            (inds,) = np.nonzero(values == value)
+            new_channel_ids = self.unit_ids[inds]
```
chrishalcrow:
```suggestion
            new_unit_ids = self.unit_ids[inds]
```

src/spikeinterface/core/basesorting.py
```diff
@@ -626,6 +626,47 @@ def time_slice(self, start_time: float | None, end_time: float | None) -> BaseSo
...
+            subsort = self.select_units(new_channel_ids)
```
chrishalcrow:
```suggestion
            subsort = self.select_units(new_unit_ids)
```

alejoe91: oups!

PR #4047, opened 2025-07-09
Fix for file_suffix that is missing

yger:
I don't know why, but file_suffix is missing from the arguments, and undefined


samuelgarcia:
@zm711 I think this was accidentatly inject by this #3993

samuelgarcia:
no worry camarade.

PR #4051, opened 2025-07-11
Remove MEArec downloads from docs

chrishalcrow:
(I think that) Lots of documentation build time and flakiness is due to the MEArec downloading data.

This PR switches our downloaded the 10s file in favour of generating using `si.generate_ground_truth_recording`. It also moves the one tutorial which is _about_ using MEArec to be a run-local tutorial.

Let's see if the docs build...

PR #4052, opened 2025-07-11
Switch RTD to UV

zm711:
Text here written by @chrishalcrow:

Use `uv` instead of `mamba` for readthe docs, for a x10 speedup.

The final build
https://app.readthedocs.org/projects/spikeinterface/builds/28834497/


zm711:
This depends on #4051 because we would need to tweak datalad/git-annex install. But if we remove gin then we don't need git-annex/datalad for docs (as @chrishalcrow is doing in his PR).

Basically this takes us from ~100 second mamba env generation to 10second env generation with UV. We can then also delete the extra conda env file.

PR #4057, opened 2025-07-12
Add motion functions to api docs

### LINKED ISSUE #3083: `correct_motion_on_peaks` is not included in the API docs
Not sure why, I will look into this ASAP.

zm711:
Fixes #3083.

The second half of missing api functions. Now added to the api.rst.

Also a tiny fix. The semantic (that Chris and I discussed) for saying backward compatible would be to save that it is for versions greater than what we are being compatible to. So I think we want to actually list this compatibility as > 100 rather than < 101. Otherwise I think we would rewrite that whole statement.

### DIFF
--- a/src/spikeinterface/preprocessing/motion.py
+++ b/src/spikeinterface/preprocessing/motion.py
@@ -568,6 +568,19 @@ def correct_motion(
 
 
 def save_motion_info(motion_info, folder, overwrite=False):
+    """
+    Saves motion info
+
+    Parameters
+    ----------
+    motion_info : dict
+        The returned motion_info from running `compute_motion`
+    folder : str | Path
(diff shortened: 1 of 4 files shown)


src/spikeinterface/preprocessing/motion.py
```diff
@@ -590,6 +603,18 @@ def save_motion_info(motion_info, folder, overwrite=False):
...
+    """
+    Loads a motion info dict from folder
+    Parameters
+    ----------
+    folder : str | Path
+        The folder containing the motion info to load
+
+    Notes
+    -----
+    Loads both current Motion implmemntation as well as the
```
chrishalcrow:
```suggestion
    Loads both current Motion implementation as well as the
```

PR #4058, opened 2025-07-12
Update `job_kwargs` shared docs

### LINKED ISSUE #2875: Improve docstring explaining `job_kwargs`
As I mentioned in #2696,

I only recently realized that a job was actually just a process and so technically could set n_jobs =100 even if they had one core (and they weren't on Windows which has a hard limit of 61 processes). So it might be better for us to strategize the best way to explain n_jobs. We added in the fraction which is now clearer, but a better explanation of what negative values mean and n_jobs > cores means.

To be honest I would love if we could incorporate @h-mayorquin's tutorial diagrams somewhere in our docs moving forward that help explain aggregation, slice, and memory chunks.

zm711:
Fixes #2875 

1) tried to make the `n_jobs` description more accurate. Please read and add feedback
2) added in the missing kwargs `pool_engine` and `max_threads_per_worker` which were not documented yet.

### DIFF
--- a/src/spikeinterface/core/job_tools.py
+++ b/src/spikeinterface/core/job_tools.py
@@ -29,13 +29,21 @@ _shared_job_kwargs_doc = """**job_kwargs : keyword arguments for parallel proces
                 - chunk_duration : str or float or None
                     Chunk duration in s if float or with units if str (e.g. "1s", "500ms")
             * n_jobs : int | float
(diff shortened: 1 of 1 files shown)


chrishalcrow (approved):
Nice and clear

h-mayorquin (approved):

alejoe91 (approved):

alejoe91:
beautiful :)

PR #4059, opened 2025-07-12
Updates to the `get_started` folder of docs

zm711:


doc/get_started/install_sorters.rst
```diff
@@ -12,7 +12,7 @@ and in many cases the easiest way to run them is to do so via Docker or Singular
 **This is the approach we recommend for all users.**
 To run containerized sorters see our documentation here: :ref:`containerizedsorters`.
 
-There are some cases where users will need to install the spike sorting algorithms in their own environment. If you
+There are some cases where users will need to install the spike sorting algorithms on your own computer. If you
```
alejoe91:
```suggestion
There are some cases where you will need to install the spike sorting algorithms on your own computer. If you
```
Then this should be "you" instead of "users", right?

zm711: Yes to fit with the next statement. The you isn't required but consistency is better!

PR #4060, opened 2025-07-14
Check for remote paths in check_paths_relative

### LINKED ISSUE #4045: OSError when attempting to save recording loaded from s3
Hello,

I am following code provided by @alejoe91 with the goal of copying some data from s3 to a local drive. The recording is loaded using `spikeinterface.load("s3://...(path).../data.zarr")`, and loading seems to work fine. I take a number of slices out of the recording with `recording.time_slice`, then concatenate with `spikeinterface.concatenate_recordings`. However, when I try this code:
```
recording_saved = recording_concat.save(folder="recording_concatenaded_1min_snippets", n_jobs=-1, mp_context="spawn", overwrite=True)
```
I get the following OSError:
```
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 's3:\\...(path)...\\data.zarr'
```

[traceback.txt](https://github.com/user-attachments/files/21126415/traceback.txt)

alejoe91:
Fixes #4045 

@jacobpennington can you check if this fixes your Windows issue?

### DIFF
--- a/src/spikeinterface/core/core_tools.py
+++ b/src/spikeinterface/core/core_tools.py
@@ -429,6 +429,11 @@ def check_paths_relative(input_dict, relative_folder) -> bool:
             not_possible.append(p)
             continue
 
+        # check path is not a remote path
+        if is_path_remote(p):
+            not_possible.append(p)
+            continue
+
         # If windows path check have same drive
         if isinstance(p, WindowsPath) and isinstance(relative_folder, WindowsPath):
             # check that on same drive
--- a/src/spikeinterface/core/core_tools.py
+++ b/src/spikeinterface/core/core_tools.py
@@ -423,7 +423,6 @@ def check_paths_relative(input_dict, relative_folder) -> bool:
     relative_folder = Path(relative_folder).resolve().absolute()
     not_possible = []
     for p in path_list:
-        p = Path(p)
         # check path is not an URL
         if "http" in str(p):
             not_possible.append(p)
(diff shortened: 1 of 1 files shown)


jacobpennington:
This does not fix the issue, the error is the same. The problem is that `is_path_remote` is testing for `"s3://" in str(path)`, but by the time that function is called the path string has been converted to a `WindowsPath` which changes it to `"s3:\\"` (with back slashes). However, if the body of `is_path_remote` is updated to the following, then it works:
```
return "s3:" in str(path) or "gcs:" in str(path)
```

alejoe91:
Thanks for testing @jacobpennington 

I think that the problem was that the path was cast as Path at the beginning of the loop, so hopefully this commit https://github.com/SpikeInterface/spikeinterface/pull/4060/commits/90217832324713fd5533d07a8a3198a30c7b190a fixes it with minimal change

jacobpennington:
That should fix it @alejoe91 , but it looks like the data link is no longer working so I can't confirm. I will try checking again later today, in case it's just a temporary issue.

src/spikeinterface/core/core_tools.py
```diff
@@ -423,12 +423,19 @@ def check_paths_relative(input_dict, relative_folder) -> bool:
     relative_folder = Path(relative_folder).resolve().absolute()
     not_possible = []
     for p in path_list:
-        p = Path(p)
         # check path is not an URL
         if "http" in str(p):
             not_possible.append(p)
             continue
 
+        # check path is not a remote path
```
h-mayorquin:
```suggestion
        # check path is not a remote path, see 
        # https://github.com/SpikeInterface/spikeinterface/issues/4045
```

alejoe91:
@jacobpennington I'm able to stream and process the file. Can you test again? Maybe as you said it might have been a temporary connectivity issue

samuelgarcia (approved):

samuelgarcia:
OK

PR #4061, opened 2025-07-14
Cast samples to int64 after round

### LINKED ISSUE #4044: Integer overflow when taking time slice of long recording
Hello,

I am using the code below to slice out many 1 minute chunks from a long recording, where `recording`  was loaded from s3 using `si.load("path_to_data.zarr")`. At some point this results in an assertion error: `AssertionError: start_frame=-2147483648 must be positive` when the sample-equivalent of the time provided in seconds exceeds 2**31 (approximately 71600 seconds for a 30kHz recording).

```
# Recording for each start
recording_list = []
for start_time in start_times:
    rec_sub = recording.time_slice(start_time=start_time, end_time=start_time+chunk_duration)
    recording_list.append(rec_sub)
```

Casting sample indices to `np.int64` instead of `int` in `spikeinterface.core.baserecording.BaseRecordingSegment.time_to_sample_index` when using `np.round` fixes the issue:

```
    def time_to_sample_index(self, time_s):
        """
        Transform time in seconds into sample index
        """
        if self.time_vector is None:
            if self.t_start is None:
                sample_index = time_s * self.sampling_frequency
            else:
                sample_index = (time_s - self.t_start) * self.sampling_frequency
            sample_index = np.round(sample_index).astype(np.int64)     #  <-----   changed casting here
        else:
            sample_index = np.searchsorted(self.time_vector, time_s, side="right") - 1

        return sample_index
```

I did not test if the `self.time_vector is not None` case would need updating casting.

alejoe91:
Fixes #4044

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -982,7 +982,7 @@ class BaseRecordingSegment(BaseSegment):
                 sample_index = time_s * self.sampling_frequency
             else:
                 sample_index = (time_s - self.t_start) * self.sampling_frequency
-            sample_index = np.round(sample_index).astype(int)
+            sample_index = np.round(sample_index).astype(np.int64)
         else:
             sample_index = np.searchsorted(self.time_vector, time_s, side="right") - 1
 


src/spikeinterface/core/baserecording.py
```diff
@@ -982,7 +982,7 @@ def time_to_sample_index(self, time_s):
                 sample_index = time_s * self.sampling_frequency
             else:
                 sample_index = (time_s - self.t_start) * self.sampling_frequency
-            sample_index = np.round(sample_index).astype(int)
```
zm711, line 985: I think @h-mayorquin would also like to see this. I think the thought was originally that doing this as int would let python decide the correct size of the datatype but it seems like that was the wrong idea.

alejoe91: Let's wait for @h-mayorquin input then :)

samuelgarcia (approved):

src/spikeinterface/core/baserecording.py
samuelgarcia: I think explicit is better no ?

h-mayorquin: Explicit is better but what about uint64?

h-mayorquin: But, are we sure this fixes it? In my system, astype int casts to int64:

image

zm711: If you do the same thing with `np.round` it becomes `int32`. But I agree that this seems weird. I think this has to be the combo of multiple steps because it is a slice then concatenate. So each slice must cause this to be int32 by itself and only when stitched together does the issue come up. If we force int64 no matter what then the slice should be fine. right?

h-mayorquin: image

zm711: image
maybe os specific this was a test on windows.

h-mayorquin: Maybe that's what happened and the user was using windows? My concern was that the error is coming from another source.

In any case, I agree with @samuelgarcia that explicit is better than implicit and I think this is an improvement.

h-mayorquin (approved):
LGTM

src/spikeinterface/core/baserecording.py
zm711: Based on the other issue opened I think the user was on Windows.

zm711: image

numpy > 2 works the same on Windows. So I think this is unified now. But I guess we have to be ready for people still using numpy < 2 on various OSes

PR #4063, opened 2025-07-14
Fix Kilosort tests for 4.0.39

alejoe91:


zm711 (commented):
Seems like this didn't trigger the kilosort tests :/

Otherwise seems good to me.

zm711:
Seems like we need to edit this: 

https://github.com/SpikeInterface/spikeinterface/blob/42d48558e96da8c6426409ffc6ba55e7285a9106/.github/workflows/test_kilosort4.yml#L7-L10

I haven't used that feature for PRs. I had only done the hash in the past. But maybe if we add "ci" file under the kilosort4.py we will also test on changes to the test file?

zm711 (approved):
KS tests passing :)

PR #4064, opened 2025-07-14
Add `read_neuroscope` to `__all__` for extractors

zm711:
We weren't granting access to the convenience function `read_neuroscope` which was causing errors in the documentation that we were skipping over. It is one of the edge cases that we missed in my recent PR. Mainly I'm testing this via the docs which should lose the error (good thing they only take 2 minutes now :) )


zm711:
So now we just have formatting errors in the docs (+like 80 warnings). 

@chrishalcrow maybe we should think about hitting the 12 FormattingErrors for functions as a post-docathon project.

chrishalcrow (approved):

PR #4070, opened 2025-07-17
Add intan reader for concatenated files

h-mayorquin:
This is yet another modality of intan where files are partitioned by the acquisition system. This reader is a convenience for the users so they can readily load their data as it is.


src/spikeinterface/extractors/extractor_classes.py
```diff
@@ -200,5 +201,6 @@
         "read_binary",  # convenience function for binary formats
         "read_zarr",
         "read_neuroscope",  # convenience function for neuroscope
+        "read_intan_segmented",  # convenience function for segmented intan files
```
alejoe91: Awesome! Can you add this to the docs extracors and API as well?

h-mayorquin: Should be done.

zm711:
I think the PR is in good shape but I have a few motivation questions.

1) Why a separate extractor? Could we add a mode into the normal Intan? 
2) `Segment` is loaded in spikeinterface. What this is actually is a concatenated version so maybe we should think about the naming
3) why not give the user the option to append or concatenate? Maybe they did the splits on purpose (ie they did baseline, experiment, post time period. So they would prefer to append instead. Why should we create a convenience function for only one of the two options and then tell the user if you want to do the other then do it yourself.
4) I could imagine other readers also have these format styles (though I don't have an example off the top of my head) so why preference Intan (other than you and I work on it a lot)
5) Why not do a dir mode at the Neo level where you make each file a Neo segment instead? Why here?

h-mayorquin:
> I think the PR is in good shape but I have a few motivation questions.
> 
> 1. Why a separate extractor? Could we add a mode into the normal Intan?

This is my preference to keep the extractor simple. Otherwise, I would need to add logic for which file to select, more arguments a more complicated docstring, etc. This way I highlight that this is an special case by construction and I think makes the implementation simpler.

> 2. `Segment` is loaded in spikeinterface. What this is actually is a concatenated version so maybe we should think about the naming

Yes, open to other naming suggestions.

> 3. why not give the user the option to append or concatenate? Maybe they did the splits on purpose (ie they did baseline, experiment, post time period. So they would prefer to append instead. Why should we create a convenience function for only one of the two options and then tell the user if you want to do the other then do it yourself.

I think that the splits are always continuous on Intan, aren't they? The idea is to load this data has the acquisition system produced it. If you have an option that indeed produces the files with gaps in between then I think we should have an option as you suggest.

> 4. I could imagine other readers also have these format styles (though I don't have an example off the top of my head) so why preference Intan (other than you and I work on it a lot)

Yes, if other readers also produce segmented files with equivalent logic, there is a clear path for appending (like here by filename), **and** we have testing data I think we should offer something similar. Nothing special about Intan other than being on my to-do list as I know the format and I have data. (I don't work with Intan so much by the way, I used it in a project at the start of the year and that's it, but I would say I know the format well now).

> 5. Why not do a dir mode at the Neo level where you make each file a Neo segment instead? Why here?

This is the quickest way to provide the feature to users. I am not sure how to do it in neo and now that I think about it the implementation in neo is already too complicated for my liking. This is how I feel.

zm711:
> I think that the splits are always continuous on Intan, aren't they? The idea is to load this data has the acquisition system produced it. If you have an option that indeed produces the files with gaps in between then I think we should have an option.

True they are continuous in time. But the experimenter could want them to be same segment or separate segment for their own provenance. Here you are forcing them to keep them in one segment. Why? Concatenate vs append is not based on time it is based on wanting them in separate segments for whatever reason you want them separate.

h-mayorquin:
> True they are continuous in time. But the experimenter could want them to be same segment or separate segment for their own provenance. Here you are forcing them to keep them in one segment. Why? Concatenate vs append is not based on time it is based on wanting them in separate segments for whatever reason you want them separate.

I am not forcing them, they still can just load the single segments on intan and use append themselves. 

I am not providing a utility for them to do this easily because I come from a different place: The idea for me is that we should have recordings that extract the data in the way that the acquisition system "intended" it and tools for allowing the users to modify that on top of it to whatever representation they like. Loading data as a continuous segment is the former, loading data as different segments is the latter.

zm711:
> I am not providing a utility for them to do this easily because I come from a different place: The idea for me is that we should have recordings that extract the data in the way that the acquisition system "intended" it

I think we don't know how acq system "intended" in this case. The software lets the user choose their cutoff length, which means that the software could be allowing the user to intend for this to be individual segments. I think the argument in your favor is that Intan provides a post-hoc stitching software that can make a series of continuous recordings one giant recording (which would be read as a neo mono-segement (ie I don't think there is an initial intention about concatenate vs append, but there is a post-hoc intention for concatenate). 

I think we addressed (1), (4), and (5). We disagree about (3) but I'm willing to be overruled. let me think about naming for (2).

src/spikeinterface/extractors/neoextractors/intan.py
```diff
@@ -104,3 +105,90 @@ def _add_channel_groups(self):
...
+    """
+
+    def __init__(
+        self,
+        folder_path,
+        stream_id=None,
+        stream_name=None,
+        all_annotations=False,
+        use_names_as_ids=False,
+        ignore_integrity_checks: bool = False,
```
zm711, line 149: If we go through with this I think we have to remove the ignore_integrity_checks. I think this PR would need to ensure that the recordings are truly continuous no? So two step process. 

1) Make sure no individual file is broken
2) ensure that the timestamp files are actually continuous

zm711: Otherwise you're running into your issue with the sampling rates? This PR could be used to stitch together any arbitrary intan files when we should have a way to do this like Intan does it.

h-mayorquin:
I don't feel that strong about 3, I was just explaining why I decided to do it this way. I am fine with adding that functionality (automatically creating a mega-segmented file) but not on this PR as that implementation would become more complicated and I don't have time to do it at the moment.

src/spikeinterface/extractors/neoextractors/intan.py
h-mayorquin: Why? I think it is fine if they want to load some of the files even if they gaps within. I don't see a strong reason to limit user choice.

h-mayorquin:
as for the naming, what about `read_intan_split` or something like that?

src/spikeinterface/extractors/neoextractors/intan.py
```diff
@@ -104,3 +105,90 @@ def _add_channel_groups(self):
...
+                stream_id=stream_id,
+                stream_name=stream_name,
+                all_annotations=all_annotations,
+                use_names_as_ids=use_names_as_ids,
+                ignore_integrity_checks=ignore_integrity_checks,
+            )
+            recording_list.append(recording)
+
+        # Initialize the parent class with the recording list
+        ConcatenateSegmentRecording.__init__(self, recording_list)
```
zm711, line 179: Wouldn't it just be a boolean check here with an AppendSegmentRecording?

zm711:
> as for the naming, what about read_intan_split or something like that?

yeah we could riff on that. like `read_split_intan_files`, `read_split_intan`. The issue is the gui software isn't super clear with a naming convention. It is just a setting. So for official intan terminology we have "Traditional" with splits?

image

src/spikeinterface/extractors/neoextractors/intan.py
h-mayorquin: No, it inherits from the respective class.

h-mayorquin:
Changed the naming.

src/spikeinterface/extractors/neoextractors/intan.py
zm711: I missed that. Yeah I was thinking you were just taking the `concatenate_recording`, for my own learning what is the benefit of doing this way?

src/spikeinterface/extractors/neoextractors/intan.py
zm711: One corrupt file could mess up everything right? How would the user know which file was corrupt? 

I guess what I was really proposing was shouldn't we check what we are stitching are actually continuous or do you want people to be able to just pile in a bunch of random Intan files and try to stitch them together with this?

src/spikeinterface/extractors/neoextractors/intan.py
h-mayorquin: We need a class for _kwargs and pickling.

PR #4071, opened 2025-07-17
Fix silence period and job_kwargs

### LINKED ISSUE #4038: parallel processing with silence_periods
Hello,

In order to guard against a few periods of time in which some recordings of mine were saturated, I've:

- written and applied a function to detect these periods
- applied silence_periods with the "noise" function to them and the surrounding milliseconds.
- noted these periods for later exclusion from analysis

This process works quite well. However: silence_periods is quite slow when I don't run it parallel. The outer layer of this function does not appear to take an n_jobs kwarg; when I go into silence_periods.py myself and add the kwarg to get_noise_levels, it works, but the parallel pool doesn't properly close afterward, which leads to bugs later when I try to parallel process something else...

So, this post is either a feature request or a question as to how to do this properly if I am not. Is there a way to insist on parallel processing for silence_periods? If not, could you please implement this for future releases? Additionally: is there a spikeinterface-internal way I should be closing the parallel pool after I start one like this?

Thanks!

Jeff Boucher

yger:
Putative attempt to fix #4038, by explicitly exposing job_kwargs to the silence_period function, in order to allow the user to set them explicitly. In addition, noise_levels are now saved explicitly in extra_kwargs. In theory, once a first call to noise_levels has been made, they should be cached and thus no recomputation should be made.

### DIFF
--- a/src/spikeinterface/preprocessing/silence_periods.py
+++ b/src/spikeinterface/preprocessing/silence_periods.py
@@ -5,8 +5,9 @@ import numpy as np
 from spikeinterface.core.core_tools import define_function_handling_dict_from_class
 from .basepreprocessor import BasePreprocessor, BasePreprocessorSegment
 
-from spikeinterface.core import get_random_data_chunks, get_noise_levels
+from spikeinterface.core import get_noise_levels
 from spikeinterface.core.generate import NoiseGeneratorRecording
+from spikeinterface.core.job_tools import split_job_kwargs
 
 
 class SilencedPeriodsRecording(BasePreprocessor):
@@ -48,6 +49,8 @@ class SilencedPeriodsRecording(BasePreprocessor):
         available_modes = ("zeros", "noise")
         num_seg = recording.get_num_segments()
 
+        random_chunk_kwargs, job_kwargs = split_job_kwargs(random_chunk_kwargs) 
+
         if num_seg == 1:
             if isinstance(list_periods, (list, np.ndarray)) and np.array(list_periods).ndim == 2:
                 # when unique segment accept list instead of of list of list/arrays
@@ -74,7 +77,7 @@ class SilencedPeriodsRecording(BasePreprocessor):
                 random_slices_kwargs = random_chunk_kwargs.copy()
                 random_slices_kwargs["seed"] = seed
                 noise_levels = get_noise_levels(
-                    recording, return_scaled=False, random_slices_kwargs=random_slices_kwargs
+                    recording, return_scaled=False, random_slices_kwargs=random_slices_kwargs, **job_kwargs
                 )
             noise_generator = NoiseGeneratorRecording(
                 num_channels=recording.get_num_channels(),
@@ -97,7 +100,7 @@ class SilencedPeriodsRecording(BasePreprocessor):
             rec_segment = SilencedPeriodsRecordingSegment(parent_segment, periods, mode, noise_generator, seg_index)
             self.add_recording_segment(rec_segment)
 
-        self._kwargs = dict(recording=recording, list_periods=list_periods, mode=mode, seed=seed)
+        self._kwargs = dict(recording=recording, list_periods=list_periods, mode=mode, seed=seed, noise_levels=noise_levels)
         self._kwargs.update(random_chunk_kwargs)
 
 
--- a/src/spikeinterface/preprocessing/silence_periods.py
+++ b/src/spikeinterface/preprocessing/silence_periods.py
@@ -37,6 +37,12 @@ class SilencedPeriodsRecording(BasePreprocessor):
         - "noise": The periods are filled with a gaussion noise that has the
                    same variance that the one in the recordings, on a per channel
                    basis
+    job_kwargs : dict
+        Keyword arguments for the joblib parallelization. If you want to use
+        `job_kwargs`, you need to pass them as a dictionary with the key "job_kwargs".
+        For example, `job_kwargs={"num_workers": 4}`.
+        Note that this is not used for the `get_noise_levels` function, which has its own
+        `random_slices_kwargs` argument.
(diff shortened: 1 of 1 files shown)


src/spikeinterface/preprocessing/silence_periods.py
```diff
@@ -44,7 +51,16 @@ class SilencedPeriodsRecording(BasePreprocessor):
...
 
-    def __init__(self, recording, list_periods, mode="zeros", noise_levels=None, seed=None, **random_chunk_kwargs):
+    def __init__(
+        self,
+        recording,
+        list_periods,
+        mode="zeros",
+        noise_levels=None,
+        seed=None,
+        job_kwargs=dict(),
```
zm711:
```suggestion
      **job_kwargs,
```

Should it be like this? 

And then load the `_shared_job_kwargs` doc? 

I'll test this PR now!

zm711:
So it appropriately used the global_job kwargs now during `detect_peaks`. But it didn't use the cached noise levels during the node pipeline it recalculated them.

Exposing job_kwargs worked fine for `silence_periods`.

Copy of script I ran and terminal output.
```python
> import spikeinterface.full as si
> from spikeinterface.sortingcomponents.peak_detection import detect_peaks
> rec, _ = si.generate_ground_truth_recording([600], 30000, 64, 30)
> rec = si.common_reference(si.bandpass_filter(rec))
> rec_silenced = si.silence_periods(rec, [[30000,30040],[50000, 50400]], mode='noise', job_kwargs=dict(n_jobs=7))
noise_level (workers: 7 processes): 100%|██████████| 20/20 [00:00<00:00, 50.22it/s]
> si.set_global_job_kwargs(**dict(n_jobs=7))
> peaks = detect_peaks(rec_silenced)
noise_level (workers: 7 processes): 100%|██████████| 20/20 [00:00<00:00, 46.42it/s]
detect peaks using locally_exclusive (workers: 7 processes): 100%|██████████| 600/600 [00:13<00:00, 43.03it/s]
```

I will say this went much faster than when I was testing and each process would go back and recalculate noise_levels with n_jobs=1.

yger:
@samuelgarcia @zm711  This should also now fix the problem with detect_peaks.

zm711:
At least on my computer it still said it was computing `noise_levels`. 
```python
rec, _ = si.generate_ground_truth_recording([600], 30000, 64, 30)
rec = si.common_reference(si.bandpass_filter(rec))
rec1 = si.silence_periods(rec, [[30000,30040],[50000, 50400]], mode='noise', job_kwargs=dict(n_jobs=7))
si.set_global_job_kwargs(**dict(n_jobs=7))
peaks= detect_peaks(rec1)
noise_level (workers: 7 processes): 100%|██████████| 20/20 [00:00<00:00, 46.48it/s]
noise_level (workers: 7 processes): 100%|██████████| 20/20 [00:00<00:00, 46.86it/s]
detect peaks using locally_exclusive (workers: 7 processes): 100%|██████████| 600/600 [00:36<00:00, 16.55it/s]
```

zm711:
Is this because it is assuming `silence_periods` has a different noise level now that we removed some "artifacts". Maybe this is the right behavior for this preprocessor?

src/spikeinterface/preprocessing/silence_periods.py
```diff
@@ -97,8 +104,10 @@ def __init__(self, recording, list_periods, mode="zeros", noise_levels=None, see
             rec_segment = SilencedPeriodsRecordingSegment(parent_segment, periods, mode, noise_generator, seg_index)
             self.add_recording_segment(rec_segment)
 
-        self._kwargs = dict(recording=recording, list_periods=list_periods, mode=mode, seed=seed)
-        self._kwargs.update(random_chunk_kwargs)
+        self._kwargs = dict(
+            recording=recording, list_periods=list_periods, mode=mode, seed=seed, noise_levels=noise_levels
+        )
+        self._kwargs.update(noise_levels_kwargs)
```
zm711: I don't play with the `_kwargs`. Will this update lead to incompatible keys between spikeinterface versions?

zm711 (commented):
When I tested this it fixed the most immediate problems with the PR. Do we want to ask the user who reported this to test this?

src/spikeinterface/preprocessing/silence_periods.py
yger: Actually this is a good question, and we should discuss this PR with @samuelgarcia . Indeed, moving the computation of noise levels into the detect_peaks() and not in the node themselves is the only option to control finely the job_kwargs. Not sure we need to save the noise_level_kwargs because the noise_levels are cached per recording I think, and not recomputed during parallel processing. I'll double check with @samuelgarcia

samuelgarcia:
```suggestion
```

samuelgarcia: We do not need to store the noise_levels_kwargs.
noise_levels will take the precedance

src/spikeinterface/preprocessing/silence_periods.py
```diff
@@ -71,11 +80,9 @@ def __init__(self, recording, list_periods, mode="zeros", noise_levels=None, see
 
         if mode in ["noise"]:
             if noise_levels is None:
-                random_slices_kwargs = random_chunk_kwargs.copy()
-                random_slices_kwargs["seed"] = seed
-                noise_levels = get_noise_levels(
-                    recording, return_in_uV=False, random_slices_kwargs=random_slices_kwargs
-                )
+                noise_levels_kwargs["return_in_uV"] = False
```
samuelgarcia, line 84:
```suggestion
                noise_levels_kwargs = noise_levels_kwargs.copy()
                noise_levels_kwargs["return_in_uV"] = False
```

PR #4072, opened 2025-07-17
Add IBL's "noise cutoff" quality metric (again)

huizizhang949:
Copy of #3984 , before I somehow made a mess of it.


doc/modules/qualitymetrics/noise_cutoff.rst
```diff
@@ -1,30 +1,124 @@
...
+Noise cutoff
 ========================================
 
 Calculation
 -----------
 
 
-Metric describing whether an amplitude distribution is cut off, similar to _amp_cutoff  :ref:`amplitude cutoff <amp_cutoff>` but without a Gaussian assumption.
-A histogram of amplitudes is created and quantifies the distance between the low tail, mean number of spikes and high tail in terms of standard deviations.
+Metric describing whether an amplitude distribution is cut off as it approaches zero, similar to _amp_cutoff  :ref:`amplitude cutoff <amp_cutoff>` but without a Gaussian assumption.
```
alejoe91:
```suggestion
Metric describing whether an amplitude distribution is cut off as it approaches zero, similar to  :ref:`amplitude cutoff <amp_cutoff>` but without a Gaussian assumption.
```

doc/modules/qualitymetrics/noise_cutoff.rst
```diff
@@ -1,30 +1,124 @@
-Noise cutoff (not currently implemented)
+Noise cutoff
```
alejoe91:
```suggestion
Noise cutoff (:code:`noise_cutoff`)
```

doc/modules/qualitymetrics/noise_cutoff.rst
```diff
@@ -1,30 +1,124 @@
...
+
+Together, (cutoff, ratio) quantify how suppressed the low‐end of the amplitude distribution is relative to the top quantile and to the peak.
 
 Expectation and use
 -------------------
 
 Noise cutoff attempts to describe whether an amplitude distribution is cut off.
-The metric is loosely based on [Hill]_'s amplitude cutoff, but is here adapted (originally by [IBL]_) to avoid making the Gaussianity assumption on spike distributions.
-Noise cutoff provides an estimate of false negative rate, so a lower value indicates fewer missed spikes (a more complete unit).
+If the distribution is not truncated at the low-end, one would expect cutoff to be less than 5 and ratio less than 0.1.
```
alejoe91: @huizizhang949 @chrishalcrow 

here we say that cutoff should be less than 5, but in the plot a cutoff of 1.961 shows a large cutoff. Maybe this should be 0.5 instead of 5?

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -37,6 +37,160 @@
...
+    # the indices for low-amplitude bins
+    low_indices = np.where(bin_edges[1:] <= low_quantile_value)[0]
+
+    high_quantile_value = np.quantile(amps, q=1 - high_quantile)
+
+    # the indices for high-amplitude bins
+    high_indices = np.where(bin_edges[:-1] >= high_quantile_value)[0]
+
+    if len(low_indices) == 0:
+        warnings.warn("No bin is selected to test cutoff. Please increase low_quantile.")
```
alejoe91:
```suggestion
        warnings.warn("No bin is selected to test cutoff. Please increase low_quantile. Setting noise cutoff and ratio to NaN")
```

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -37,6 +37,160 @@
...
+        return np.nan, np.nan
+
+    # compute ratio between low-amplitude bins and the largest bin
+    low_counts = n_per_bin[low_indices]
+    mean_low_counts = np.mean(low_counts)
+    ratio = mean_low_counts / maximum_bin_height
+
+    if len(high_indices) == 0:
+        warnings.warn("No bin is selected as the reference region. Please increase high_quantile.")
```
alejoe91:
```suggestion
        warnings.warn("No bin is selected as the reference region. Please increase high_quantile. Setting noise cutoff to NaN")
```

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -37,6 +37,160 @@
...
+        return np.nan, ratio
+
+    if len(high_indices) == 1:
+        warnings.warn(
+            "Only one bin is selected as the reference region, and thus the standard deviation cannot be computed. "
+            "Please increase high_quantile."
```
alejoe91:
```suggestion
            "Please increase high_quantile. Setting noise cutoff to NaN"
```

src/spikeinterface/qualitymetrics/misc_metrics.py
```diff
@@ -37,6 +37,160 @@
...
+        )
+        return np.nan, ratio
+
+    # compute cutoff from low-amplitude and high-amplitude bins
+    high_counts = n_per_bin[high_indices]
+    mean_high_counts = np.mean(high_counts)
+    std_high_counts = np.std(high_counts)
+    if std_high_counts == 0:
+        warnings.warn("All the high-amplitude bins have the same size. Please consider changing n_bins.")
```
alejoe91:
```suggestion
        warnings.warn(
            "All the high-amplitude bins have the same size. Please consider changing n_bins. "
            "Setting noise cutoff to NaN"
        )
```

doc/modules/qualitymetrics/noise_cutoff.rst
chrishalcrow: Hello, I think we are following what IBL suggests. @huizizhang949 what IBL arguments are equivalent to our arguments? It's something like they use some fixed number of bins, and they only include one bin in the low-tail? Maybe we can clarify this by saying something like:

A larger value of ``noise_cutoff`` suggests that the distribution is cut-off. IBL uses the default values of "..." and use a threshold of 5 to determine whether a unit is cut off or not. In practice, the IBL threshold is quite conservative, and a lower threshold might be better for your data. We suggest plotting the data using the ``plot_amplitudes`` widget to view your data when choosing your threshold.

huizizhang949: Hi, IBL's whitepaper suggests 5 for thresholding `cutoff` and 0.1 for `ratio`, but their code only computes and outputs `cutoff` (no `ratio`) and does not refer to any thresholds. Yes, they use the number of lower bins as the argument, with a default value of 1, which can be unstable. And we found that for the examples here 5 and 0.1 are actually too large. I have modified doc to clarify these.

doc/modules/qualitymetrics/noise_cutoff.rst
```diff
@@ -1,30 +1,127 @@
-Noise cutoff (not currently implemented)
+Noise cutoff (:code:`noise_cutoff`)
 ========================================
```
alejoe91:
```suggestion
===================================
```

chrishalcrow:
Hello, just did some tiny tiny docs formatting.

chrishalcrow (approved):

PR #4074, opened 2025-07-18
Add `to_pynapple_tsgroup` function

chrishalcrow:
Adds the ability to export a SortingAnalyzer or Sorting to a TsdGroup object from pynapple (https://pynapple.org/user_guide/01_introduction_to_pynapple.html#nap-tsgroup-group-of-timestamps). Made with advice from @gviejo, at NeuroDataReHack 2025!

Tests and docs included.

Some comments:
- The user can specific metadata, but we decided to include some basic metadata if it was available (metrics and unit locations).
- Pynapple always uses seconds as the time unit. I've implemented this using the `return_times` argument in `get_unit_spike_train`. This can be a bit confusing for the user - what happens with multi-segments etc? We really need to finish our time doc!!
- We put the burden of installing pynapple on the users side.


src/spikeinterface/exporters/to_pynapple.py
```diff
@@ -0,0 +1,65 @@
+from spikeinterface.core import SortingAnalyzer, BaseSorting
+import pandas as pd
```
zm711: pandas is not a core dependency :)

src/spikeinterface/exporters/to_pynapple.py
```diff
@@ -0,0 +1,65 @@
...
+        A TsGroup object from the pynapple package.
+    """
+    from pynapple import TsGroup, Ts
+
+    if isinstance(sorting_analyzer_or_sorting, SortingAnalyzer):
+        sorting = sorting_analyzer_or_sorting.sorting
+    elif isinstance(sorting_analyzer_or_sorting, BaseSorting):
+        sorting = sorting_analyzer_or_sorting
+    else:
+        raise TypeError("The function `to_pynapple_tsgroup` only accepts a SortingAnalyzer or Sorting object.")
```
zm711: Could we return the type of the object that was accidentally given. For me I sometimes confuse the order of rec, sorting in generate_ground_truth_recording so it would be nice to know I put in a recording instead of a sorting/analyzer.

doc/modules/exporters.rst
```diff
@@ -4,6 +4,40 @@ Exporters module
 The :py:mod:`spikeinterface.exporters` module includes functions to export SpikeInterface objects to other commonly
 used frameworks.
 
+Exporting to Pynapple
+---------------------
+
+The python package `Pynapple <https://pynapple.org/>`_ is often used for combining ephys
+and behavioral data. It can be used to decode behavior, make tuning curves and compute spectrograms.
+The :py:func:`~spikeinterface.exporters.to_pynapple_tsgroup` function allows you to convert a
+SortingAnalyzer to Pynapple's `TsGroup` object.
```
zm711:
```suggestion
``SortingAnalyzer`` to Pynapple's ``TsGroup`` object.
```
double ticks for rst :)

doc/modules/exporters.rst
```diff
@@ -4,6 +4,40 @@ Exporters module
...
+
+**Note** : When creating the `TsGroup`, we will use the underlying time support of the SortingAnalyzer.
```
zm711:
```suggestion
**Note** : When creating the ``TsGroup``, we will use the underlying time support of the ``SortingAnalyzer``.
```
I guess the Analyzer is optional but I like to render it as code.

doc/modules/exporters.rst
```diff
@@ -4,6 +4,40 @@ Exporters module
...
+How this works depends on your acquisition system. You can use the `get_times` method on a recording
+to see the time support of your recording.
+
+When making the `TsGroup`, you can also attach useful metadata such as the quality metrics of each unit
+or the location of the unit in the brain. You can pass any pandas DataFrame, whose index contains the
+`unit_id`s of your sorting object.
```
zm711:
```suggestion
``unit_id``s of your sorting object.
```

doc/modules/exporters.rst
```diff
@@ -4,6 +4,40 @@ Exporters module
...
+
+The following code creates a `TsGroup` from a `SortingAnalyzer`, then save's it using `Pynapple`'s
```
zm711:
```suggestion
The following code creates a ``TsGroup`` from a ``SortingAnalyzer``, then save's it using ``Pynapple``'s
```

src/spikeinterface/exporters/to_pynapple.py
```diff
@@ -0,0 +1,65 @@
+from spikeinterface.core import SortingAnalyzer, BaseSorting
+import pandas as pd
+import numpy as np
+
+
+def to_pynapple_tsgroup(
+    sorting_analyzer_or_sorting: SortingAnalyzer | BaseSorting,
+    metadata: pd.DataFrame | dict | None = None,
```
zm711: Also what if we let the user choose a list of metadata and then we organize it instead. Right? each unit could have location + qm + tm all in one big dataframe for Pynapple?

zm711 (commented):
Forgot to do the docs really quick :)

zm711:
> We put the burden of installing pynapple on the users side.

Although fair this complicates the definition of `pip install spikeinterface[full]` even more for my rewrite.... Maybe we should really think about what this means.

src/spikeinterface/exporters/to_pynapple.py
chrishalcrow: oi oi oi. It is a dependency of pynapple (what a great app!), so the import (now in the function) will work if pynapple is installed.

src/spikeinterface/exporters/to_pynapple.py
chrishalcrow: I don't understand. Like, they specify "unit_locations", "quality_metrics", etc?

src/spikeinterface/exporters/to_pynapple.py
zm711: yes but this breaks all our code because if you don't install pynapple it will look for pandas as this is a top level import.

zm711: ie move pandas into the function :)

src/spikeinterface/exporters/to_pynapple.py
zm711: I missed what you were doing. I see it now. Maybe that is a little too complicated. If they wanted to do the dataframe on their own could they add metadata to the pynapple structure later. Basically if they want to use the export correctly they have to put a None in for this argument. So doing a dict or dataframe is just hijacking it.

src/spikeinterface/exporters/to_pynapple.py
zm711: I see you said you have it in the function now. Did you forget to push that change?

src/spikeinterface/exporters/to_pynapple.py
chrishalcrow: Yeah, there is a `set_info` method for the TsGroup. So we could ignore the metadata, and get the user to attach it after creation. Maybe we can discuss at the next meeting

zm711: That sounds good. I don't think this comment style is letting my point come across clearly.

I like the idea of exporting metadata, but I don't like the idea of letting the user do a million different styles of metadata. I would prefer them to give a list of spikeinterface metadata they want that we could export for them. If they just want to set their own metadata then I don't think this function is for that so I think the finer point there is better to talk about then write about.

src/spikeinterface/exporters/to_pynapple.py
chrishalcrow: I had a go, what do you think of the new message?

doc/modules/exporters.rst
```diff
@@ -4,6 +4,41 @@ Exporters module
...
+
+**Note** : When creating the ``TsGroup``, we will use the underlying time support of the SortingAnalyzer.
+How this works depends on your acquisition system. You can use the ``get_times`` method on a recording
+to see the time support of your recording.
+
+When making the ``TsGroup``, you can also attach useful metadata such as the quality metrics of each unit
+or the location of the unit in the brain. You can pass any pandas DataFrame, whose index contains the
+``unit_id`` s of your sorting object.
+
+The following code creates a ``TsGroup`` from a ``SortingAnalyzer``, then save's it using ``Pynapple``'s
```
zm711:
```suggestion
The following code creates a ``TsGroup`` from a ``SortingAnalyzer``, then saves it using ``Pynapple``'s
```

doc/modules/exporters.rst
```diff
@@ -4,6 +4,41 @@ Exporters module
...
+save method. You can include any metadata as a pandas DataFrame. Here, we include the ``analyzer`` s
```
zm711:
```suggestion
save method. You can include any metadata as a pandas DataFrame. Here, we include the ``analyzer`` 's
```

maybe this? with the apostrophe? hard to read the rendering so maybe you think just the space is easier?

doc/modules/exporters.rst
```diff
@@ -4,6 +4,41 @@ Exporters module
...
+quality metics.
```
zm711:
```suggestion
quality metrics.
```

zm711 (commented):
Few more typos found. I still wonder about the metadata bit. But definitely better to just discuss in person.

chrishalcrow:
Hello, I met with the Pynapple team (@gviejo, @sjvenditto, @wulfdewolf) at FlatIron. We loosely talked about the PR, with two main discussion points:

1. What to do with multi-segment sortings. We decided that each segment should output one TsGroup. We had agreed that a multi-segment sorting should then output a list or tuple of TsGroups. On implementation, I think it works better if the user has to input the segment_index e.g. `my_tsgroup = si.to_pynapple_tsgroup(analyzer, segment_index=2)`. This matches some other functions in spikeinterface (e.g. `get_unit_spike_train`), forces multi-segment users to realise there is one TsGroup per segment and keeps the return value of the `to_pynapple_tsgroup` to a single type. And keeps the most-common use-case (one segment) simple.
2. What metadata to pass to the TsGroup by default. The Pynapple team were in favour of “if it’s there, pass it on”. So that’s one vote for attaching any computed extension info that contain unit information (metrics + locations).

zm711:
> What to do with multi-segment sortings. We decided that each segment should output one TsGroup. We had agreed that a multi-segment sorting should then output a list or tuple of TsGroups. On implementation, I think it works better if the user has to input the segment_index e.g. my_tsgroup = si.to_pynapple_tsgroup(analyzer, segment_index=2). This matches some other functions in spikeinterface (e.g. get_unit_spike_train), forces multi-segment users to realise there is one TsGroup per segment and keeps the return value of the to_pynapple_tsgroup to a single type. And keeps the most-common use-case (one segment) simple.

I support the selecting `segment_index` since it fits with the rest of the codebase.

> What metadata to pass to the TsGroup by default. The Pynapple team were in favour of “if it’s there, pass it on”. So that’s one vote for attaching any computed extension info that contain unit information (metrics + locations).

I support this. My point here is I think it should either be a bool (export all possible spikeinterface stuff or export nothing) or a list of things you want to export (quality metrics, template metrics, etc). I just don't think we should allow the user to feed in an arbitrary dataframe that we will then export for them. So based on what the pynapple team is saying (*hi to all of you!*) I would make it an export bool.

samuelgarcia:
This could be cool to have this for the relase no ? And we can improve a bit later.

chrishalcrow:
> I support this. My point here is I think it should either be a bool (export all possible spikeinterface stuff or export nothing) or a list of things you want to export (quality metrics, template metrics, etc). I just don't think we should allow the user to feed in an arbitrary dataframe that we will then export for them. So based on what the pynapple team is saying (_hi to all of you!_) I would make it an export bool.

Ahhh, ok, got you! So if bool=False, the user should make the TsGroup, then add their own metadata to that using pynapple functionality? I like that!

zm711:
not for this PR, but if we are pursing this strategy in general we should add a note in the documentation that we would support export layers for projects that want to create their own export specific code. Might be worth discussion at the maintenance meeting just like we have a section in the sorter section explaining how to get your sorter added to the code base.

zm711:
Done dirty by hugging face :P

pyproject.toml
```diff
@@ -141,6 +141,7 @@ test_extractors = [
     # Commenting out for release
     "probeinterface @ git+https://github.com/SpikeInterface/probeinterface.git",
     "neo @ git+https://github.com/NeuralEnsemble/python-neo.git",
+    "pynapple",
```
chrishalcrow: REMOVE!!!

doc/modules/exporters.rst
```diff
@@ -4,6 +4,48 @@ Exporters module
...
+        attach_unit_metadata=True,
+    )
+
+    # Note: can add metadata using e.g.
+    # my_tsgroup.set_info({'brain_region': ['MEC', 'MEC', ...]})
+
+    my_tsgroup.save("my_tsgroup_output.npz")
+
+If you have a multi-segment sorting, you need to pass the ``segment_index`` argument to the
+``to_pynapple_tsgroup`` function. This way, you can generate on ``TsGroup`` per segment.
```
zm711:
```suggestion
``to_pynapple_tsgroup`` function. This way, you can generate one ``TsGroup`` per segment.
```

src/spikeinterface/exporters/to_pynapple.py
```diff
@@ -0,0 +1,91 @@
...
+        raise TypeError(
+            f"The `sorting_analyzer_or_sorting` argument must be a SortingAnalyzer or Sorting object, not a {type(sorting_analyzer_or_sorting)} type object."
+        )
+
+    unit_ids = sorting.unit_ids
+
+    unit_ids_castable = True
+    try:
+        unit_ids_ints = [int(unit_id) for unit_id in unit_ids]
+    except:
```
zm711:
```suggestion
    except ValueError:
```
I don't think we should allow a non-specific except. seems dangerous no?

src/spikeinterface/exporters/to_pynapple.py
```diff
@@ -0,0 +1,91 @@
...
+    else:
+        raise TypeError(
+            f"The `sorting_analyzer_or_sorting` argument must be a SortingAnalyzer or Sorting object, not a {type(sorting_analyzer_or_sorting)} type object."
+        )
+
+    unit_ids = sorting.unit_ids
+
+    unit_ids_castable = True
+    try:
+        unit_ids_ints = [int(unit_id) for unit_id in unit_ids]
```
zm711, line 46: Another strategy that wouldn't use the try-except would be

```python
unit_ids_castable = all([unit_id.isdigit() for unit_id in unit_ids])
if unit_ids_castable:
    unit_ids_ints = [int(unit_id) for unit_id in unit_ids]
else:
xx
```

doc/modules/exporters.rst
```diff
@@ -4,6 +4,48 @@ Exporters module
 The :py:mod:`spikeinterface.exporters` module includes functions to export SpikeInterface objects to other commonly
 used frameworks.
 
+Exporting to Pynapple
+---------------------
+
+The python package `Pynapple <https://pynapple.org/>`_ is often used for combining ephys
```
zm711:
```suggestion
The Python package `Pynapple <https://pynapple.org/>`_ is often used for combining ephys
```

I think we do Python capital for writing in SI?

zm711 (commented):
1 typo
1 ? style

2 questions regarding the int try-except

src/spikeinterface/exporters/to_pynapple.py
zm711: Never mind. This only works if the ids are string. --which maybe in the future they will be ;P

chrishalcrow: Yeah, I thought about this too and ended up try/excepting.
```all([isinstance(unit_id, int) or unit_id.isdigit() for unit_id in unit_ids])```
works but is a bit gross.

I'd vote to keep the try/except for now

src/spikeinterface/exporters/to_pynapple.py
zm711: Sorry never responded. This is great Thanks!

zm711 (approved):
After putting you through the ringer this is good by me. :)

src/spikeinterface/exporters/to_pynapple.py
zm711: The costs are minimal, but based on my reading a try-except is slightly faster than an if-else if you succeed most of the time, but is quite a bit slower if you except often. That being said even a 10x slowdown of one step isn't really that meaningful. so now that you added in the specific except I'm okay with this. Thanks for humoring me :)

samuelgarcia (approved):

alejoe91 (approved):

gviejo:
🙌  🎉

PR #4075, opened 2025-07-18
Add append mode to IntanSplitFile building on #4070

zm711:
@h-mayorquin,

this is my proposal on top of your PR for an append mode (without docs). If you have a reason to hate it we can close it, but this seems easy enough, no? But you know the pickle stuff better so maybe this ruins that?
Screenshot 2025-07-18 at 09 36 24

may replace #4070

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -177,6 +177,7 @@ spikeinterface.preprocessing
     .. autofunction:: clip
     .. autofunction:: common_reference
     .. autofunction:: correct_lsb
+    .. autofunction:: compute_motion
     .. autofunction:: correct_motion
     .. autofunction:: get_motion_presets
     .. autofunction:: get_motion_parameters_preset
@@ -184,6 +185,8 @@ spikeinterface.preprocessing
     .. autofunction:: save_motion_info
     .. autofunction:: depth_order
     .. autofunction:: detect_bad_channels
+    .. autofunction:: detect_and_interpolate_bad_channels
+    .. autofunction:: detect_and_remove_bad_channels
     .. autofunction:: directional_derivative
     .. autofunction:: filter
     .. autofunction:: gaussian_filter
--- a/doc/how_to/index.rst
+++ b/doc/how_to/index.rst
@@ -18,4 +18,5 @@ Guides on how to solve specific, short problems in SpikeInterface. Learn how to.
     auto_curation_training
     auto_curation_prediction
     physical_units
+    unsigned_to_signed
     customize_a_plot
--- a/doc/how_to/physical_units.rst
+++ b/doc/how_to/physical_units.rst
@@ -1,3 +1,5 @@
+.. _physical_units:
+
 Working with physical units in SpikeInterface recordings
 ========================================================
 
--- a/doc/how_to/unsigned_to_signed.rst
+++ b/doc/how_to/unsigned_to_signed.rst
@@ -1,0 +1,102 @@
+Unsigned to Signed Data types
+=============================
+
+As of version 0.103.0 SpikeInterface has changed one of its defaults for interacting with
+:code:`Recording` objects. We no longer autocast unsigned dtypes to signed implicitly. This
+means that some users of SpikeInterface will need to add one additional line of code to their scripts
+to explicitly handle this conversion.
+
+
+Why this matters?
+-----------------
+
+For those that want a deeper understanding of dtypes `NumPy provides a great explanation <https://numpy.org/doc/stable/reference/arrays.dtypes.html>`_.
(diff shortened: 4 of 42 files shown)


h-mayorquin:
> this is my proposal on top of your PR for an append mode (without docs). If you have a reason to hate it we can close it, but this seems easy enough, no? But you know the pickle stuff better so maybe this ruins that?

I am OK with this. I have the intuition that a separate class instead of a keyword argument might be safer but this seems to work. We already have a test for pickling in the `test_neoextractors`, can you add another entry in the test for new mode so we test both modes?

zm711:
Yeah I'll have to find the pickle test info but I can add it.

I do think regardless of whether we do yours or this one, we should discuss whether this ought to be allowable for any intan file or only the traditional split mode which we are not checking for at the moment.

src/spikeinterface/extractors/tests/test_neoextractors.py
```diff
@@ -184,6 +186,14 @@ class IntanRecordingTestMultipleFilesFormat(RecordingCommonTestSuite, unittest.T
     ]
 
 
+class IntanSplitFilesRecordingTest(RecordingCommonTestSuite, unittest.TestCase):
```
h-mayorquin, line 189: @zm711 
Here, you just need to add another test just exactly like this but passing the argument. This test has a pickle test inside of it but at the moment is only testing the default mode.

h-mayorquin:
> I do think regardless of whether we do yours or this one, we should discuss whether this ought to be allowable for any intan file or only the traditional split mode which we are not checking for at the moment.

I think that's the beauty of doing it as a separate class, users can decide where they point this to. I don't think we should be policing any further.

src/spikeinterface/extractors/tests/test_neoextractors.py
zm711: Done.

zm711:
> I think that's the beauty of doing it as a separate class, users can decide where they point this to. I don't think we should be policing any further.

Yeah I think I was getting hung up on your original idea about this should try to be a class that handles how the "acquisition" intended. Not checking for the timestamps is actively against the intention of that system because the system internally uses the timestamps to validate the integrity of the blocks. But it does allow people to debug a corrupted file if they need to vs just failing.

h-mayorquin (approved):
I am OK with this instead of my PR. I think it is an improvement as it adds another feature.

src/spikeinterface/extractors/neoextractors/intan.py
```diff
@@ -104,3 +106,106 @@ def _add_channel_groups(self):
...
+                ignore_integrity_checks=ignore_integrity_checks,
+            )
+            recording_list.append(recording)
+
+        # Initialize the parent class with the recording list
+        if mode == "concatenate":
+            ConcatenateSegmentRecording.__init__(self, recording_list)
+        elif mode == "append":
+            AppendSegmentRecording.__init__(self, recording_list)
+
```
samuelgarcia, line 196: Is this legal when double hineritance not initialize the other class ?
Just wondering..

samuelgarcia:
I propose we merge this and see how long this will work.

src/spikeinterface/extractors/neoextractors/intan.py
h-mayorquin: Yeah, I am weary of that. But maybe it is an opportunity for us to learn something if it blows up at some point x D

h-mayorquin:
Thanks for the review, @samuelgarcia

zm711:
I think this is only okay because they both only have `__init__` with no methods. So because we call the specific `__init__` rather than `super()` we go down one path. They both use the methods of `BaseRecording` so the methods end up being the same. If either of them had the same methods (overwriting the base methods) then I think we would have problems. Based on my reading this is the general idea of the diamond problem and this trick shouldn't be used often. It's more that I think it is okay for the specific case because of the nature of the classes.

zm711:
Do we want a comment in the code saying that this is slightly dangerous but we are testing it out blame zm711 if it goes wrong?

h-mayorquin:
> Do we want a comment in the code saying that this is slightly dangerous but we are testing it out blame zm711 if it goes wrong?

It is too much of a hunch to write anything that would be useful I think. If we get an error at some point we can just create a separate class and use a router in the snake_case version of the caller to decided which class we use. This could also be done with __new__ but Sam and Alessio think that pattern might not play well with serialization which would be another danger.

zm711:
> It is too much of a hunch to write anything that would be useful I think. If we get an error at some point we can just create a separate class and use a router in the snake_case version of the caller to decided which class we use.

Fair enough. I'll leave this to @samuelgarcia / @alejoe91  now :)

PR #4076, opened 2025-07-19
Add additional `unsigned_to_signed` doc

zm711:
As discussed at our maintenance meeting here is the first draft of our unsigned_to_signed doc for people converting to 0.103.0 and running into the value error. 

@chrishalcrow @h-mayorquin would love for your opinions on doc side.

@alejoe91, please just push directly to the PR with your fixes for bit depth.

Link to docs: https://spikeinterface--4076.org.readthedocs.build/en/4076/how_to/unsigned_to_signed.html

### DIFF
--- a/src/spikeinterface/core/__init__.py
+++ b/src/spikeinterface/core/__init__.py
@@ -109,7 +109,12 @@ from .recording_tools import (
     get_chunk_with_margin,
     order_channels_by_depth,
 )
-from .sorting_tools import spike_vector_to_spike_trains, random_spikes_selection, apply_merges_to_sorting
+from .sorting_tools import (
+    spike_vector_to_spike_trains,
+    random_spikes_selection,
+    apply_merges_to_sorting,
+    apply_splits_to_sorting,
+)
 
 from .waveform_tools import extract_waveforms_to_buffers, estimate_templates, estimate_templates_with_accumulator
 from .snippets_tools import snippets_from_sorting
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -9,6 +9,7 @@ It also implements:
   * ComputeNoiseLevels which is very convenient to have
 """
 
+import warnings
 import numpy as np
 
 from .sortinganalyzer import AnalyzerExtension, register_result_extension
@@ -90,6 +91,11 @@ class ComputeRandomSpikes(AnalyzerExtension):
             selected_mask = np.zeros(spikes.size, dtype=bool)
             selected_mask[random_spikes_indices] = True
             new_data["random_spikes_indices"] = np.flatnonzero(selected_mask[keep_mask])
+        return new_data
+
+    def _split_extension_data(self, split_units, new_unit_ids, new_sorting_analyzer, verbose=False, **job_kwargs):
+        new_data = dict()
+        new_data["random_spikes_indices"] = self.data["random_spikes_indices"].copy()
         return new_data
 
     def _get_data(self):
@@ -243,8 +249,6 @@ class ComputeWaveforms(AnalyzerExtension):
     def _merge_extension_data(
         self, merge_unit_groups, new_unit_ids, new_sorting_analyzer, keep_mask=None, verbose=False, **job_kwargs
     ):
-        new_data = dict()
-
         waveforms = self.data["waveforms"]
         some_spikes = self.sorting_analyzer.get_extension("random_spikes").get_random_spikes()
         if keep_mask is not None:
@@ -274,6 +278,11 @@ class ComputeWaveforms(AnalyzerExtension):
                 waveforms = waveforms[:, :, :new_num_chans]
 
         return dict(waveforms=waveforms)
+
+    def _split_extension_data(self, split_units, new_unit_ids, new_sorting_analyzer, verbose=False, **job_kwargs):
+        # splitting only affects random spikes, not waveforms
+        new_data = dict(waveforms=self.data["waveforms"].copy())
+        return new_data
 
     def get_waveforms_one_unit(self, unit_id, force_dense: bool = False):
         """
@@ -554,6 +563,49 @@ class ComputeTemplates(AnalyzerExtension):
 
         return new_data
 
+    def _split_extension_data(self, split_units, new_unit_ids, new_sorting_analyzer, verbose=False, **job_kwargs):
+        if not new_sorting_analyzer.has_extension("waveforms"):
+            warnings.warn(
+                "Splitting templates without the 'waveforms' extension will simply copy the template of the unit that "
+                "was split to the new split units. This is not recommended and may lead to incorrect results. It is "
+                "recommended to compute the 'waveforms' extension before splitting, or to use 'hard' splitting mode.",
+            )
+        new_data = dict()
+        for operator, arr in self.data.items():
+            # we first copy the unsplit units
+            new_array = np.zeros((len(new_sorting_analyzer.unit_ids), arr.shape[1], arr.shape[2]), dtype=arr.dtype)
+            new_analyzer_unit_ids = list(new_sorting_analyzer.unit_ids)
+            unsplit_unit_ids = [unit_id for unit_id in self.sorting_analyzer.unit_ids if unit_id not in split_units]
+            new_indices = np.array([new_analyzer_unit_ids.index(unit_id) for unit_id in unsplit_unit_ids])
+            old_indices = self.sorting_analyzer.sorting.ids_to_indices(unsplit_unit_ids)
+            new_array[new_indices, ...] = arr[old_indices, ...]
+
+            for split_unit_id, new_splits in zip(split_units, new_unit_ids):
+                if new_sorting_analyzer.has_extension("waveforms"):
+                    for new_unit_id in new_splits:
+                        split_unit_index = new_sorting_analyzer.sorting.id_to_index(new_unit_id)
+                        wfs = new_sorting_analyzer.get_extension("waveforms").get_waveforms_one_unit(
+                            new_unit_id, force_dense=True
+                        )
+
+                        if operator == "average":
+                            arr = np.average(wfs, axis=0)
+                        elif operator == "std":
+                            arr = np.std(wfs, axis=0)
+                        elif operator == "median":
+                            arr = np.median(wfs, axis=0)
+                        elif "percentile" in operator:
+                            _, percentile = operator.splot("_")
+                            arr = np.percentile(wfs, float(percentile), axis=0)
+                        new_array[split_unit_index, ...] = arr
+                else:
+                    split_unit_index = self.sorting_analyzer.sorting.id_to_index(split_unit_id)
+                    old_template = arr[split_unit_index, ...]
+                    new_indices = new_sorting_analyzer.sorting.ids_to_indices(new_splits)
+                    new_array[new_indices, ...] = np.tile(old_template, (len(new_splits), 1, 1))
+            new_data[operator] = new_array
+        return new_data
+
     def _get_data(self, operator="average", percentile=None, outputs="numpy"):
         if operator != "percentile":
             key = operator
@@ -724,6 +776,10 @@ class ComputeNoiseLevels(AnalyzerExtension):
     def _merge_extension_data(
         self, merge_unit_groups, new_unit_ids, new_sorting_analyzer, keep_mask=None, verbose=False, **job_kwargs
     ):
+        # this does not depend on units
+        return self.data.copy()
+
+    def _split_extension_data(self, split_units, new_unit_ids, new_sorting_analyzer, verbose=False, **job_kwargs):
         # this does not depend on units
         return self.data.copy()
 
--- a/src/spikeinterface/core/sorting_tools.py
+++ b/src/spikeinterface/core/sorting_tools.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
+import warnings
 import importlib.util
 
 import numpy as np
 
-from .basesorting import BaseSorting
-from .numpyextractors import NumpySorting
+from spikeinterface.core.base import BaseExtractor
+from spikeinterface.core.basesorting import BaseSorting
+from spikeinterface.core.numpyextractors import NumpySorting
 
 numba_spec = importlib.util.find_spec("numba")
 if numba_spec is not None:
@@ -226,6 +228,7 @@ def random_spikes_selection(
     return random_spikes_indices
 
 
+### MERGING ZONE ###
 def apply_merges_to_sorting(
     sorting: BaseSorting,
     merge_unit_groups: list[list[int | str]] | list[tuple[int | str]],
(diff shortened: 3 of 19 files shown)


doc/how_to/unsigned_to_signed.rst
```diff
@@ -0,0 +1,81 @@
+Unsigned to Signed Data types
+=============================
+
+As of version 0.103.0 SpikeInterface has changed one of its defaults for interacting with
+:code:`Recording`` objects. We no longer autocast unsigned dtypes to signed implicitly. This
```
chrishalcrow:
```suggestion
:code:`Recording` objects. We no longer autocast unsigned dtypes to signed implicitly. This
```

:code: + double tick gives a rendered tick

doc/how_to/unsigned_to_signed.rst
```diff
@@ -0,0 +1,81 @@
...
+means that some users of SpikeInterface will need to add one additional line of code to their scripts
+to explicitly handle this conversion.
+
+
+Why this matters?
+-----------------
+
+For those that want a deeper understanding of dtypes `NumPy provides a great explanation <https://numpy.org/doc/stable/reference/arrays.dtypes.html>`_.
+For our purposes it is important to know that many pieces of recording equipment opt to store their electrophysiological data as unsigned integers,
+which provides the benefit of reduces the necessary file size. In order to convert to real units these file formats only need to store a :code:`gain`
```
chrishalcrow:
```suggestion
which provides the benefit of reducing the necessary file size. In order to convert to real units these file formats only need to store a :code:`gain`
```

doc/how_to/unsigned_to_signed.rst
```diff
@@ -0,0 +1,81 @@
...
+.. code:: python
+
+    import spikeinterface.extractors as se
+    import spikeinterface.preprocessing as spre
+
+    # Intan is an example of unsigned data
+    recording = se.read_intan('path/to/my/file.rhd', stream_id='0')
+    # to get a signed version of our Recording we use the following function
+    recording_signed = spre.unsigned_to_signed(recording)
+    # we can use all functions that we used previously in our scripts
```
chrishalcrow: Somehow this is a bit confused, because the "used previously" sounds like it's from earlier in the tutorial, or something. 
```suggestion
    # we can now apply any preprocessing functions like normal, e.g.
```

doc/how_to/unsigned_to_signed.rst
```diff
@@ -0,0 +1,81 @@
...
+
+If you are curious if your :code:`Recording` is unsigned you can simply check the repr or use :code:`get_dtype()`
+
+.. code:: python
+
+    # the repr automatically displays the dtype
+    print(recording)
+    # use method on the Recording object
+    print(recording.get_dtype())
+
```
chrishalcrow, line 63: Would be good to mention what you'd expect here. E.g. "If this returns a dtype starting with a u, such as `uint16`, the recording if using unsigned int. While if this returns ... "

doc/how_to/unsigned_to_signed.rst
```diff
@@ -0,0 +1,81 @@
...
+
+
+Additional Notes
+----------------
+
+1) Some some sorters make use of SpikeInterface preprocessing either
```
chrishalcrow: This list doesn't render as a list. You need to put an indent in the lines e.g.

```
1) Hello I'm 
   a long list item

2) me
   too!
```

(took me a minute to get this working: you should try building locally!! It takes less than 30s now!)

chrishalcrow: Also "Some some" -> "some" typo!

chrishalcrow:
Very clear and easy to understand - just some minor typos, and one request. Thanks @zm711 !!

doc/how_to/unsigned_to_signed.rst
zm711: Do you think I should just write up a parsed-literal block or you think it is sufficient to just say `uint16`?

doc/how_to/unsigned_to_signed.rst
zm711: Will fix the some some (haha). I didn't explicitly intend it as a list. Just numbered, but rst does tend to render non-lists relatively ugly.

doc/how_to/unsigned_to_signed.rst
chrishalcrow: Up to you. I think sufficient just to say: the possible things it can return are pretty simple.

doc/how_to/unsigned_to_signed.rst
```diff
@@ -0,0 +1,82 @@
...
+to explicitly handle this conversion.
+
+
+Why this matters?
+-----------------
+
+For those that want a deeper understanding of dtypes `NumPy provides a great explanation <https://numpy.org/doc/stable/reference/arrays.dtypes.html>`_.
+For our purposes it is important to know that many pieces of recording equipment opt to store their electrophysiological data as unsigned integers,
+which provides the benefit of reducing the necessary file size. In order to convert to real units these file formats only need to store a :code:`gain`
+and an :code:`offset`. Our :code:`RecordingExtractor`'s maintain the dtype that the file format utilizes, which means that some of our
```
h-mayorquin, line 17: Personal opinion. I think the possessive apostrophe together with code looks ugly.

doc/how_to/unsigned_to_signed.rst
```diff
@@ -0,0 +1,82 @@
+Unsigned to Signed Data types
+=============================
+
+As of version 0.103.0 SpikeInterface has changed one of its defaults for interacting with
+:code:`Recording` objects. We no longer autocast unsigned dtypes to signed implicitly. This
+means that some users of SpikeInterface will need to add one additional line of code to their scripts
```
h-mayorquin, line 6: Maybe mention the formats here that suffer from it?

doc/how_to/unsigned_to_signed.rst
```diff
@@ -0,0 +1,82 @@
...
+to explicitly handle this conversion.
+
+
+Why this matters?
+-----------------
+
+For those that want a deeper understanding of dtypes `NumPy provides a great explanation <https://numpy.org/doc/stable/reference/arrays.dtypes.html>`_.
+For our purposes it is important to know that many pieces of recording equipment opt to store their electrophysiological data as unsigned integers,
```
h-mayorquin: "Many pieces" maybe being specific here would help if you know the formats that do this.

doc/how_to/unsigned_to_signed.rst
```diff
@@ -0,0 +1,82 @@
...
+which provides the benefit of reducing the necessary file size. In order to convert to real units these file formats only need to store a :code:`gain`
+and an :code:`offset`. Our :code:`RecordingExtractor`'s maintain the dtype that the file format utilizes, which means that some of our
+:code:`RecordingExtractor`'s will have unsigned dtypes.
+
+The problem with using unsigned dtypes is that many types of funtions (including the ones we use from SciPy) perform poorly with unsigned integers.
+This is made worse by the fact that these failures are silent (i.e. no error is triggered but the operation leads to nonsensical data). So the
+solution required is to convernt unsigned integers into signed integers. Previously we did this under the hood, automatically for users that had
+a :code:`Recording` object with an unsigned dtype.
+
+We decided, however, that implicitly performing this action was not the best course of action. So from version 0.103.0, users will now explicitly
```
h-mayorquin: If we remember it here we can give the reasoning. Do we remember?

h-mayorquin (commented):
This looks good to me. If we can, I think it will be better to be specific about the formats that are affected by this. This will be useful for users so they can see if they are affected by this and/or can disregard this information otherwise.

doc/how_to/unsigned_to_signed.rst
zm711: I don't care either way to happy to delete if you think that would be better.

doc/how_to/unsigned_to_signed.rst
zm711: I think this is the perfect being the enemy of good. The only way I can think of doing this is:

1) looking through the code in neo for each file format to see how it is stored
2) testing this myself to see how each is stored

both of these are too time-intensive for me. Are you interested in making this list for us and then I'll add it in?

doc/how_to/unsigned_to_signed.rst
zm711: "acquisition systems"? is that better? I don't necessarily want to go into depth here unless we do end up doing an exhaustive list. So I guess answering your statement above will determine what we do here.

doc/how_to/unsigned_to_signed.rst
zm711: I don't, other than implicit bad: explicit good. So you get a better provenance because previously we were doing a secret step that wasn't being recorded in provenance files necessarily (I think) haven't checked. Maybe worth discussion or @alejoe91 can add during his review.

doc/how_to/unsigned_to_signed.rst
h-mayorquin: No. I don't have time either.

We don't have the resources to be exhaustive but do we know at least or or two examples. Didn't this cause troubles for you and that's how you noticed?

doc/how_to/unsigned_to_signed.rst
h-mayorquin: Yes, if that is our reasoning I think is good to just say that. "We decide however to stop doing implicit logic with side effects here ... " or something along those lines.

doc/how_to/unsigned_to_signed.rst
zm711: Yes it is in the example. I know Intan does this. But that's the only one I know for sure off the top of my head.  (I could do a `e.g. Intan`) but I also didn't want to preference one system.

doc/how_to/unsigned_to_signed.rst
zm711: Okay, maybe I can text Alessio tomorrow morning for a chat to see if he wants me to add. I could imagine explicit better, + better control of bit_depth (but I don't remember how the old function worked).

doc/how_to/unsigned_to_signed.rst
h-mayorquin: I think it is fine to add only one to have at least is one example. We add more if we discover them to be more. I guess we can just do a search for uint16 in neo but for this PR at least let's have an example.

zm711: neo search is a little tricky because a lot of timestamps and other signals use uint and that doesn't matter for this. But based on that search it looks like anywhere from 3-12 could be affected.

h-mayorquin: Let me look for at least a second example if that exists.

h-mayorquin: Maxwell:

https://github.com/h-mayorquin/python-neo/blob/2240a29c4366fb3ed5dacadb54837004651ee35e/neo/rawio/maxwellrawio.py#L184-L192

h-mayorquin: Biocam as well it seems:

https://github.com/h-mayorquin/python-neo/blob/2240a29c4366fb3ed5dacadb54837004651ee35e/neo/rawio/biocamrawio.py#L87-L100

doc/how_to/unsigned_to_signed.rst
h-mayorquin: All right. I see that Chris is using them on the other PR. I will learn to live with it.

doc/how_to/unsigned_to_signed.rst
h-mayorquin: Yes, it is just that it is even bad (even for us) if we say we decided and we don't add a rationale. Beats the purpose of documenting this. I rather remove the first sentence instead of giving a statement of what we did if we don't give a rationale.

doc/how_to/unsigned_to_signed.rst
h-mayorquin: Yes, same as above. Let's give some examples.

doc/how_to/unsigned_to_signed.rst
alejoe91: Thansk guys for looking this up. TO my knowledge, it's indeed Intan, Maxwell, and Biocam. Let's start adding these and add more if they pop up

doc/how_to/unsigned_to_signed.rst
```diff
@@ -0,0 +1,82 @@
...
+means that some users of SpikeInterface will need to add one additional line of code to their scripts
+to explicitly handle this conversion.
+
+
+Why this matters?
+-----------------
+
+For those that want a deeper understanding of dtypes `NumPy provides a great explanation <https://numpy.org/doc/stable/reference/arrays.dtypes.html>`_.
+For our purposes it is important to know that many pieces of recording equipment opt to store their electrophysiological data as unsigned integers,
+which provides the benefit of reducing the necessary file size. In order to convert to real units these file formats only need to store a :code:`gain`
```
alejoe91: I don't think this is true, since `unit16` and `int16` has the same size (16 bit). The difference is just in the range of values they represent

alejoe91:
@zm711 Weirdly I cannot push to this branch:
```
To github.com:zm711/spikeinterface.git
 ! [remote rejected]     unsigned -> unsigned (permission denied)
error: failed to push some refs to 'github.com:zm711/spikeinterface.git'
```

I'll make a PR from my fork instead :)

EDIT: here it is https://github.com/zm711/spikeinterface/pull/3

doc/how_to/unsigned_to_signed.rst
zm711: That's basically what I was trying to get at. Because you can store a greater range of values you can use a smaller size unit but since we have to convert anyway to signed (or float) at some point this point is moot. I agree poorly written for what I was trying to get at and maybe not pertinent to the ephys case. We can rewrite!

zm711:
> @zm711 Weirdly I cannot push to this branch:
> 
> ```
> To github.com:zm711/spikeinterface.git
>  ! [remote rejected]     unsigned -> unsigned (permission denied)
> error: failed to push some refs to 'github.com:zm711/spikeinterface.git'
> ```
> 
> I'll make a PR from my fork instead :)
> 
> EDIT: here it is [zm711#3](https://github.com/zm711/spikeinterface/pull/3)

I get that sometimes too. Usually if I just retry it works--I think it is actually a GitHub temp failure. I just double checked and the allow maintainers is definitely checked okay :) 

I'll check out the PR!

doc/how_to/unsigned_to_signed.rst
h-mayorquin: I can add the list in another PR if you want @zm711. You have already done a lot with this : )

h-mayorquin: I can add the list in another PR if you want @zm711. You have already done a lot with this : )

doc/how_to/unsigned_to_signed.rst
```diff
@@ -0,0 +1,102 @@
...
+Similarly to signed integers, in order to convert to real units these file formats only need to store a :code:`gain`
+and an :code:`offset`. Our :code:`RecordingExtractor`'s maintain the dtype that the file format utilizes, which means that some of our
+:code:`RecordingExtractor`'s will have unsigned dtypes.
+
+The problem with using unsigned dtypes is that many types of functions (including the ones we use from :code:`SciPy`) perform poorly with unsigned integers.
+This is made worse by the fact that these failures are silent (i.e. no error is triggered but the operation leads to nonsensical data). So the
+solution required is to convernt unsigned integers into signed integers. Previously we did this under the hood, automatically for users that had
+a :code:`Recording` object with an unsigned dtype.
+
+We decided, however, that implicitly performing this action was not the best course of action, since:
```
h-mayorquin, line 25: Great. Thanks.

PR #4077, opened 2025-07-21
Add archived sorters and change rt-sort to rtsort

alejoe91:


zm711:
I think this is a great idea! Help makes it clear what we actively still support and what we don't!

doc/modules/sorters.rst
```diff
@@ -466,24 +466,29 @@ Here is the list of external sorters accessible using the run_sorter wrapper:
...
 * **Kilosort3** :code:`run_sorter(sorter_name='kilosort3')`
 * **PyKilosort** :code:`run_sorter(sorter_name='pykilosort')`
-* **Klusta** :code:`run_sorter(sorter_name='klusta')`
 * **Mountainsort4** :code:`run_sorter(sorter_name='mountainsort4')`
 * **Mountainsort5** :code:`run_sorter(sorter_name='mountainsort5')`
-* **RT-Sort** :code:`run_sorter(sorter_name='rt-sort')`
-* **SpyKING Circus** :code:`run_sorter(sorter_name='spykingcircus')`
-* **Tridesclous** :code:`run_sorter(sorter_name='tridesclous')`
+* **RTSort** :code:`run_sorter(sorter_name='rtsort')`
 * **Wave clus** :code:`run_sorter(sorter_name='waveclus')`
```
zm711, line 480: is wave clus still actively supported since there was a snippets based approach?

alejoe91: No, but it's MATLAB-compiled and it runs fine :)

zm711: cool cool.

src/spikeinterface/sorters/runsorter.py
```diff
@@ -60,6 +58,10 @@
     kilosort3="kilosort3-compiled",
     waveclus="waveclus-compiled",
     waveclus_snippets="waveclus-compiled",
+    # archived
+    # klusta="klusta",
+    # yass="yass",
+    # tridesclous="tridesclous"
```
zm711: Are we not doing tridesclous here anymore? I saw you had a commit un-archiving it?

yger:
What should I do for circus in order to not be considered as archived? I mean, KS is also not working with numpy 2.0 am I right? What is the version of numpy that I should consider, and is it the only requierement? Should I make a release then, such that it could be installed via pip?

zm711:
@alejoe91 
if you're changing the name for the wrapper do we want to change the reference? (for rtsort)? 

https://spikeinterface.readthedocs.io/en/latest/references.html

I guess the reference could be hyphenated if that is how they refer to their own sorter? So our wrapper is unhyphenated and theirs is?

src/spikeinterface/sorters/runsorter.py
```diff
@@ -43,14 +43,14 @@
     combinato="combinato",
     herdingspikes="herdingspikes",
     kilosort4="kilosort4",
-    klusta="klusta",
     mountainsort4="mountainsort4",
     mountainsort5="mountainsort5",
     pykilosort="pykilosort",
```
zm711, line 48: pykilosort is only for < 3.10. I guess we are not there yet, but we are very close to needing to archive this. We should remember this.
https://github.com/MouseLand/pykilosort/blob/master/pyks2.yml

zm711 (commented):
One note for us to remember in the future.

src/spikeinterface/sorters/runsorter.py
alejoe91: @samuelgarcia made a new release :)

alejoe91:
> image
> I think we should change this to `rtsort` since that is what the wrapper will have, and then leave the citation alone since the actual name of the sorter is RT-Sort. OR we leave the sorter as RT-Sort since that is the name in the citation. But I think we need to change this one line of the refs now that I've read everything.

Good catch!!! done in last commti :)

samuelgarcia (approved):

PR #4080, opened 2025-07-24
Alter how quality metrics check extension dependencies

chrishalcrow:
This PR changes how quality metrics are handled when an extension they depend on is missing.

The new behaviour is
- if you explicitly ask for a metric to be computed (`si.compute_snr` or `analyzer.compute('quality_metrics', metric_names=['snr']`) and don't have the extension computed, this will error.
- if you don't ask for any special metrics (`analyzer.compute('quality_metrics')`), this will compute all metrics it can. It will also gather all the metrics you can't compute, and why, and return a single warning.

The warning is e.g.
```
Some metrics you are trying to compute depend on other extensions:
    snr requires ['noise_levels', 'templates']
    amplitude_cutoff requires ['spike_amplitudes|waveforms', 'templates']
    amplitude_median requires ['spike_amplitudes|waveforms', 'templates']
    amplitude_cv requires ['spike_amplitudes|amplitude_scalings', 'templates']
    sd_ratio requires ['templates', 'spike_amplitudes']
    noise_cutoff requires ['spike_amplitudes']
To include these metrics, compute the required extensions using `sorting_analyzer.compute('extension_name')
```

Technically, we do this by doing a dependency check when we're figuring out which metrics to compute (done in `_set_params` of the `ComputeQualityMetrics` class). Then we don't include any metrics we don't have the extensions for.

If you call e.g. `compute_snrs`, it calls `_has_required_extensions` which does a check of the dependencies for a single metric, and raises the error message.

A nice consequence: now, the quality metrics extension has no dependencies, and will compute some spike-train related metrics if you give it an extension-less analyzer.

Made quite a big edit to the quality metrics modules docs page. Been meaning to do this for a while: https://spikeinterface--4080.org.readthedocs.build/en/4080/modules/qualitymetrics.html

We had discussed introducing a `UnitMetric` class or something. I think this is simpler for now!


alejoe91:
That's good for me! For the `spike_amplitudes`/`amplitude_scalings` maybe we can do `required_extensions=["spike_amplitudes|amplitude_scalings"]` like we have in the core sorting analyzer [here](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/sortinganalyzer.py#L1954)?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1944,6 +1944,17 @@ def _sort_extensions_by_dependency(extensions):
...
+    if "quality_metrics" in extensions_list:
+
+        qm_arg = extensions_list.index("quality_metrics")
+
+        extension_params.pop(qm_arg)
+        extension_params.append(extensions["quality_metrics"])
+
+        extensions_list.pop(qm_arg)
+        extensions_list.append("quality_metrics")
+
```
zm711: This is helpful. I guess one of the things we had discussed was maybe at some point working on a query optimizer. 

What if I don't want quality metrics at the end because I think something is slow so I intentionally put it not at the end?

chrishalcrow: Hello - we discussed this PR in a meeting earlier today, and will discuss further at the maintainers meeting later! So let's add this point to the agend!

alejoe91:
@chrishalcrow this is looking great!!! Can you fix conflicts in the docs so we can merge?

doc/modules/qualitymetrics.rst
```diff
@@ -12,17 +12,99 @@ Completeness metrics (or 'false negative'/'type II' metrics) aim to identify whe
...
+
+.. code-block:: python
+
+    # load or create a sorting analyzer
+    sorting_analyzer = si.load_sorting_analyzer(folder='my_sorting_analyzer')
+
+    # compute the metrics
+    sorting_analyzer.compute("quality_metrics")
+
+    # get the metrics in the form as a pandas DataFrame
```
zm711:
```suggestion
    # get the metrics in the form of a pandas DataFrame
```

? maybe you changed your wording?

doc/modules/qualitymetrics.rst
```diff
@@ -12,17 +12,99 @@ Completeness metrics (or 'false negative'/'type II' metrics) aim to identify whe
...
+    quality_metrics = sorting_analyzer.get_extension("quality_metrics").get_data()
+
+    # print the metrics that have been computed
+    print(quality_metrics.columns)
+
+Some metrics are very slow to compute when the number of units it large. So by default, the following metrics are not computed:
+
+- :doc:`qualitymetrics/isolation_distance`
```
zm711: I thought this would be computed? When did we stop this? In this PR? isolation distance isn't that hard to compute right?

doc/modules/qualitymetrics.rst
```diff
@@ -12,17 +12,99 @@ Completeness metrics (or 'false negative'/'type II' metrics) aim to identify whe
...
+- The ``nn_noise_overlap`` from :doc:`qualitymetrics/nearest_neighbor`
+
+Some metrics make use of :ref:`principal component analysis <postprocessing_principal_components>` (PCA) to reduce the dimensionality of computations.
 Various approaches to computing the principal components are possible, and choice should be carefully considered in relation to the recording equipment used.
-The following metrics make use of PCA: isolation distance, L-ratio, D-prime, Silhouette score and NN-metrics.
-By contrast, the following metrics are based on spike times only: firing rate, ISI violations, presence ratio.
-And amplitude cutoff and SNR are based on spike times as well as waveforms.
 
-For more details about each metric and it's availability and use within SpikeInterface, see the individual pages for each metrics.
+If you only want to compute a subset of metrics, you can use convenience functions to compute each one,
```
zm711, line 77: Maybe a mention that this computation is in place? You have your example not returning anything, but would it be helpful to be explicit?

doc/modules/qualitymetrics.rst
alejoe91: sharp sight Zach! this should actually be `nn_isolation`, which is a separate metric (https://spikeinterface.readthedocs.io/en/latest/modules/qualitymetrics/nearest_neighbor.html)

doc/modules/qualitymetrics.rst
chrishalcrow: Oh, yeah, doing this does not add it to the `analyzer`. Will mention that...

zm711 (approved):
Looks good to me now! :)

PR #4084, opened 2025-07-25
Fix set_properties_after_merging when property is boolean.

samuelgarcia:
Fix set_properties_after_merging when property is boolean.
This was introducing bugs in auto_merge

### DIFF
--- a/src/spikeinterface/core/sorting_tools.py
+++ b/src/spikeinterface/core/sorting_tools.py
@@ -361,8 +361,14 @@ def set_properties_after_merging(
     keep_pre_inds = sorting_pre_merge.ids_to_indices(kept_unit_ids)
     keep_post_inds = sorting_post_merge.ids_to_indices(kept_unit_ids)
 
+    default_missing_values = BaseExtractor.default_missing_property_values
+
     for key in prop_keys:
         parent_values = sorting_pre_merge.get_property(key)
+        if parent_values.dtype.kind not in default_missing_values:
+            # if the property is boolean or integer there is no missing values so we skip
+            # for instance reccussive "is_merged" will not be propagated
+            continue
 
         # propagate keep values
         shape = (len(sorting_post_merge.unit_ids),) + parent_values.shape[1:]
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/sorting_tools.py
```diff
@@ -361,8 +361,14 @@ def set_properties_after_merging(
     keep_pre_inds = sorting_pre_merge.ids_to_indices(kept_unit_ids)
     keep_post_inds = sorting_post_merge.ids_to_indices(kept_unit_ids)
 
+    default_missing_values = BaseExtractor.default_missing_property_values
+
     for key in prop_keys:
         parent_values = sorting_pre_merge.get_property(key)
+        if parent_values.dtype.kind not in default_missing_values:
+            # if the property is boolean or integer there is no missing values so we skip
+            # for instance reccussive "is_merged" will not be propagated
```
alejoe91:
```suggestion
            # for instance recursive "is_merged" will not be propagated
```

src/spikeinterface/core/sorting_tools.py
```diff
@@ -361,8 +361,14 @@ def set_properties_after_merging(
     keep_pre_inds = sorting_pre_merge.ids_to_indices(kept_unit_ids)
     keep_post_inds = sorting_post_merge.ids_to_indices(kept_unit_ids)
 
+    default_missing_values = BaseExtractor.default_missing_property_values
+
     for key in prop_keys:
         parent_values = sorting_pre_merge.get_property(key)
+        if parent_values.dtype.kind not in default_missing_values:
+            # if the property is boolean or integer there is no missing values so we skip
```
zm711, line 369: Just for my info since I don't set properties. You mean that in the case of float you can have a `nan` or what is a "missing value"?

samuelgarcia: a nan for float.
a "" for str
like we have in the base class.

alejoe91: yes. Missing values are defined here: https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/base.py#L38

For ints and bool, we don't have missing values :)

zm711: cool thanks.

PR #4085, opened 2025-07-28
Add huggingface_hub to docs requirements

alejoe91:


zm711 (approved):
I thought we had this but must've gotten lost at some point.

alejoe91:
apparently huggingface_hub was a skops dependency, but it was removed in the last release!

zm711:
One thing Chris and I were discussing that I want to write here so we have a record to discuss overall-- what is the point of the `docs` tag. Is it for CI and so we should also remove MEArec since that is not needed for the CI or should it be for the local docs build where it is required (for someone that wants to do the one .py->rst conversion).

PR #4090, opened 2025-07-29
More return_scaled > return_in_uV

samuelgarcia:
Note that core function have a warning and do nice deprecation.
Some corners function directly **return_scaled > return_in_uV** without the warning.

@chrishalcrow @zm711 @h-mayorquin @alejoe91 : tell what you think.
deprecation everywhere is a big pain


src/spikeinterface/preprocessing/remove_artifacts.py
```diff
@@ -182,7 +182,7 @@ def __init__(
                     nbefore=nbefore,
                     nafter=nafter,
                     operator=mode,
-                    return_scaled=False,
```
zm711, line 185: I don't think we should do this user facing function without a deprecation right?

samuelgarcia: estimate_templates() a deprecation warning. No ?

alejoe91: yeah this is not user facing, it's inside the `remove_artifacts`

src/spikeinterface/core/tests/test_baserecording.py
```diff
@@ -234,7 +234,7 @@ def test_BaseRecording(create_cache_folder):
 
     # Both return_scaled and return_in_uV raise error when no gain_to_uV/offset_to_uV properties
     with pytest.raises(ValueError):
-        traces_float32 = rec_int16.get_traces(return_in_uV=True)
+        traces_float32 = rec_int16.get_traces(return_scaled=True)
```
alejoe91, line 237:
```suggestion
        traces_float32 = rec_int16.get_traces(return_in_uV=True)
```
this was correct

alejoe91:
@samuelgarcia just one comment. Can you also enable commits for the pre-commit?

zm711 (commented):
The widgets are definitely user facing for `spikes_on_traces` and `traces`. And for traces I think that has to be a super common use case for users using `return_scaled` so that needs a dep no?

src/spikeinterface/core/tests/test_baserecording.py
samuelgarcia: No the code wanted to test BOTH option and only one was tested

samuelgarcia:
> The widgets are definitely user facing for `spikes_on_traces` and `traces`. And for traces I think that has to be a super common use case for users using `return_scaled` so that needs a dep no?

I was hoping my lazyness was hidden.

samuelgarcia:
@zm711 widgets are also done

src/spikeinterface/widgets/spikes_on_traces.py
```diff
@@ -88,6 +94,16 @@ def __init__(
         backend=None,
         **backend_kwargs,
     ):
+
+        # Handle deprecated return_scaled parameter
+        if return_scaled is not None:
+            warnings.warn(
+                "`return_scaled` is deprecated and will be removed in version 0.104.0. Use `return_in_uV` instead.",
```
zm711: Thanks @samuelgarcia ! I thought you wanted a really slow dep on this? But this works for me!

samuelgarcia: I have oscillating feeling about this. Yesterday I was think fast change was better.
Maybe you are right lets make 0.105.0

PR #4092, opened 2025-07-29
Updates for probeinterface 0.3.0

alejoe91:
Refactored inter-sample-shift computation using information from Probe object

Requires probeinterface 0.3.0 to be released

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/spikeglx.py
+++ b/src/spikeinterface/extractors/neoextractors/spikeglx.py
@@ -97,15 +97,14 @@ class SpikeGLXRecordingExtractor(NeoBaseRecordingExtractor):
 
             # get inter-sample shifts based on the probe information and mux channels
             sample_shifts = get_neuropixels_sample_shifts_from_probe(probe, stream_name=self.stream_name)
-            total_channels = probe.annotations.get("num_readout_channels", 384)
             if sample_shifts is not None:
-                if self.get_num_channels() != total_channels:
-                    # need slice because not all channel are saved
+                num_readout_channels = probe.annotations.get("num_readout_channels")
+                if self.get_num_channels() != num_readout_channels:
+                    # need slice because not all channels are saved
                     chans = probeinterface.get_saved_channel_indices_from_spikeglx_meta(meta_filename)
-                    # lets clip to 384 because this contains also the synchro channel
-                    chans = chans[chans < total_channels]
+                    # lets clip to num_readout_channels because this contains also the synchro channel
+                    chans = chans[chans < num_readout_channels]
                     sample_shifts = sample_shifts[chans]
-
                 self.set_property("inter_sample_shift", sample_shifts)
         else:
             warning_message = (
--- a/src/spikeinterface/extractors/neuropixels_utils.py
+++ b/src/spikeinterface/extractors/neuropixels_utils.py
@@ -28,8 +28,14 @@ def get_neuropixels_sample_shifts_from_probe(probe: Probe, stream_name: str = "a
     model_description = probe.annotations.get("description", None)
     num_channels_per_adc = probe.annotations.get("num_channels_per_adc", None)
     mux_channels = probe.contact_annotations.get("mux_channels", None)
+    num_readouts_channels = probe.annotations.get("num_readout_channels", None)
 
-    if model_description is None or num_channels_per_adc is None or mux_channels is None:
+    if (
+        model_description is None
+        or num_channels_per_adc is None
+        or mux_channels is None
+        or num_readouts_channels is None
+    ):
         warning_message = (
             "Unable to find inter-sample shifts in the Neuropixels probe metadata. "
             "The sample shifts will not be loaded. "
@@ -93,7 +99,7 @@ def get_neuropixels_sample_shifts(
         representing the fractional delay within the sampling period due to sequential ADC sampling.
     """
     warnings.warn(
-        "`get_neuropixels_sample_shifts` is deprecated and will be removed in 0.105.0. "
+        "`get_neuropixels_sample_shifts` is deprecated and will be removed in 0.104.0. "
         "Use `get_neuropixels_sample_shifts_from_probe` instead.",
         DeprecationWarning,
         stacklevel=2,
@@ -150,7 +156,7 @@ def get_neuropixels_channel_groups(num_channels=384, num_channels_per_adc=12):
         A list of lists of simultaneously sampled channel indices
     """
     warnings.warn(
-        "`get_neuropixels_channel_groups` is deprecated and will be removed in 0.105.0. "
+        "`get_neuropixels_channel_groups` is deprecated and will be removed in 0.104.0. "
         "Use the `mux_channels` contact annotation from the `Probe` instead.",
         DeprecationWarning,
         stacklevel=2,
--- a/src/spikeinterface/extractors/neoextractors/openephys.py
+++ b/src/spikeinterface/extractors/neoextractors/openephys.py
@@ -230,15 +230,15 @@ class OpenEphysBinaryRecordingExtractor(NeoBaseRecordingExtractor):
                         self.set_probe(probe, in_place=True)
                     # get inter-sample shifts based on the probe information and mux channels
                     sample_shifts = get_neuropixels_sample_shifts_from_probe(probe, stream_name=self.stream_name)
-                    total_channels = probe.annotations.get("num_readout_channels", 384)
                     if sample_shifts is not None:
-                        if self.get_num_channels() != total_channels:
-                            # need slice because not all channel are saved
+                        num_readout_channels = probe.annotations.get("num_readout_channels")
+                        if self.get_num_channels() != num_readout_channels:
+                            # need slice because not all channels are saved
                             chans = probeinterface.get_saved_channel_indices_from_openephys_settings(
                                 settings_file, oe_stream
                             )
-                            # lets clip to 384 because this contains also the synchro channel
-                            chans = chans[chans < total_channels]
+                            # lets clip to num_readout_channels because this contains also the synchro channel
+                            chans = chans[chans < num_readout_channels]
                             sample_shifts = sample_shifts[chans]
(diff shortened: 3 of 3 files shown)


src/spikeinterface/extractors/neuropixels_utils.py
```diff
@@ -1,13 +1,51 @@
...
+
+def get_neuropixels_sample_shifts_from_probe(probe: Probe, stream_name: str = "ap") -> np.ndarray:
+    """
+    Get the inter-sample shifts for Neuropixels probes based on the probe information.
+
+    Parameters
+    ----------
+    probe : Probe
+        The probe object containing channel and ADC information.
+
```
zm711, line 21: no `stream_name` entry in the docstring :)

src/spikeinterface/extractors/neuropixels_utils.py
```diff
@@ -43,6 +81,12 @@ def get_neuropixels_sample_shifts(
         Array of relative phase shifts for each channel, with values ranging from 0 to 1,
         representing the fractional delay within the sampling period due to sequential ADC sampling.
     """
+    warnings.warn(
+        "`get_neuropixels_sample_shifts` is deprecated and will be removed on 0.105.0. "
```
zm711:
```suggestion
        "`get_neuropixels_sample_shifts` is deprecated and will be removed in 0.105.0. "
```

zm711 (commented):
Two small things from my quick scan :)

src/spikeinterface/extractors/neoextractors/spikeglx.py
```diff
@@ -99,23 +99,8 @@ def __init__(
...
-            else:  # NP1.0
-                num_channels_per_adc = 12
-                num_cycles_in_adc = 13 if "ap" in self.stream_id else 12
-                total_channels = 384
-
-            # sample_shifts is generated from total channels (384) channels
-            # when only some channels are saved we need to slice this vector (like we do for the probe)
-            sample_shifts = get_neuropixels_sample_shifts(total_channels, num_channels_per_adc, num_cycles_in_adc)
+            # get inter-sample shifts based on the probe information and mux channels
+            sample_shifts = get_neuropixels_sample_shifts_from_probe(probe, stream_name=self.stream_name)
```
chrishalcrow, line 99: At the moment, looks like we run this function for every `read_spikeglx` and `read_openephys` we do? Should we only do it if we have a NP probe (somehow??)

src/spikeinterface/extractors/neuropixels_utils.py
```diff
@@ -1,13 +1,51 @@
...
+        sample_shifts[mux_channels == mux_channel] = np.arange(num_channels_per_adc) / num_cycles
+
+    return sample_shifts
+
 
 def get_neuropixels_sample_shifts(
     num_channels: int = 384, num_channels_per_adc: int = 12, num_cycles: Optional[int] = None
 ) -> np.ndarray:
     """
+    DEPRECATED
```
chrishalcrow, line 65: Can we also deprecate (delete?) `get_neuropixels_channel_groups` and `synchronize_neuropixel_streams`?

src/spikeinterface/extractors/neoextractors/spikeglx.py
zm711: Just a quick glance it seems like we set the `probe.name` for openephys but not for spikeglx. but if we set a name for both in probeinterface then we could just check the name of the probe to ensure that the probe object is actually neuropixel, no? I haven't checked out this part of probeinterface well enough but we have the attributes there for a reason. Might just need to unify them for npx.

src/spikeinterface/extractors/neuropixels_utils.py
```diff
@@ -1,13 +1,51 @@
...
+    mux_channels = probe.contact_annotations["mux_channels"]
+
+    if "2.0" in model_description:
+        # for Neuropixels 2.0 (and newer), the number of cycles in ADC is equal to the number of channels per ADC
+        num_cycles_in_adc = num_channels_per_adc
+    else:
+        # for Neuropixels 1.0 technology, the number of cycles for the AP stream is +1 because
+        # the last cycle is used for the LFP stream
+        num_cycles_in_adc = num_channels_per_adc + 1 if "ap" in stream_name.lower() else num_channels_per_adc
+    sample_shifts = np.zeros_like(adc_indices)
```
chrishalcrow:
```suggestion
    sample_shifts = np.zeros_like(adc_indices, dtype='float')
```
These need to be floats, otherwise Line 39 rounds everything down to 0.

src/spikeinterface/extractors/neoextractors/spikeglx.py
chrishalcrow: We save the name e.g. "NP1000" and description from https://github.com/SpikeInterface/probeinterface/blob/main/src/probeinterface/resources/neuropixels_probe_features.json
But neither are consistent enough to make an easy rule.
Might be something simpler but less explicit, like if they have `mux_channels` in their contact_annotations.

zm711: I think we should make a rule in probeinterface then. couldn't we just add an annotation `is_neuropixel` that is made when we read either in probeinterface, and then just check that. We have control of this with probeinterface so I think at the probeinterface level we should make it clear if something for spikeglx or openephys is neuropixel or not.

chrishalcrow: Yeah - good idea. Think it's easy on the ProbeInterface side

alejoe91: For Open Ephys, a probe can be loaded only if it's Neuropixels, since it's saved by the Neuropixels plugin. For Spikeglx, the way we read the probe from the .meta file (either from the imro string or the shank map) assumes it's a Neuropixels probe. If those fields are not available, a None probe is returned (but we should check)

src/spikeinterface/extractors/neuropixels_utils.py
alejoe91: float is default, but I'll make it explicit

src/spikeinterface/extractors/neoextractors/spikeglx.py
alejoe91: ok, after a second look here's what we currently do, which is enough to recognize NP probes:
- for spikeglx, if there is an "ap" stream, it means that the probe is a NP one. So in that case we set `add_probe_properties` to `True` and read the probe (see [here](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/extractors/neoextractors/spikeglx.py#L90-L92))
- for open ephys, the `probeinterface.read_openephys(..., raise_error=False)` will return None in case of failure (e.g., not a NP probe). Since probe info is only available through the Neuropix-PXI plugin, we are filtering for NP probes only

src/spikeinterface/extractors/neuropixels_utils.py
alejoe91: we can deprecate the `get_neuropixels_channel_group`, but the `synchronize_neuropixel_streams` is still WIP, so let's leave it there

src/spikeinterface/extractors/neoextractors/spikeglx.py
chrishalcrow: We read tetrode and Cambridge neurotech that is in openephys format in our lab. But I guess the probe read fails, and we attach the probe ourselves after. Is that right?

src/spikeinterface/extractors/neuropixels_utils.py
chrishalcrow: When I tested on our openephys data, `adc_indices` was int, so `sample_shifts` was being initialised as int. Thanks for changing!

src/spikeinterface/extractors/neuropixels_utils.py
```diff
@@ -43,6 +92,12 @@ def get_neuropixels_sample_shifts(
         Array of relative phase shifts for each channel, with values ranging from 0 to 1,
         representing the fractional delay within the sampling period due to sequential ADC sampling.
     """
+    warnings.warn(
+        "`get_neuropixels_sample_shifts` is deprecated and will be removed in 0.105.0. "
```
samuelgarcia:
```suggestion
        "`get_neuropixels_sample_shifts` is deprecated and will be removed in 0.104.0. "
```

src/spikeinterface/extractors/neuropixels_utils.py
```diff
@@ -93,7 +149,12 @@ def get_neuropixels_channel_groups(num_channels=384, num_channels_per_adc=12):
     groups : list
         A list of lists of simultaneously sampled channel indices
     """
-
+    warnings.warn(
+        "`get_neuropixels_channel_groups` is deprecated and will be removed in 0.105.0. "
```
samuelgarcia:
```suggestion
        "`get_neuropixels_channel_groups` is deprecated and will be removed in 0.104.0. "
```

src/spikeinterface/extractors/neoextractors/spikeglx.py
```diff
@@ -99,31 +95,18 @@ def __init__(
...
-                # need slice because not all channel are saved
-                chans = probeinterface.get_saved_channel_indices_from_spikeglx_meta(meta_filename)
-                # lets clip to 384 because this contains also the synchro channel
-                chans = chans[chans < total_channels]
-                sample_shifts = sample_shifts[chans]
-
-            self.set_property("inter_sample_shift", sample_shifts)
+            # get inter-sample shifts based on the probe information and mux channels
+            sample_shifts = get_neuropixels_sample_shifts_from_probe(probe, stream_name=self.stream_name)
+            total_channels = probe.annotations.get("num_readout_channels", 384)
```
samuelgarcia: I think this hard coded 384 will be dangerous in a near futur. Lets force to have it no ?

samuelgarcia (approved):

samuelgarcia:
Cool.

PR #4099, opened 2025-07-31
Fix main changes in whatisnew for 0.103.0

alejoe91:
@zm711 can we merge this before release so we have a clean doc in stable?


zm711 (approved):
Yep sounds good!

PR #4100, opened 2025-07-31
Add future annotations in to_pynapple and upgrade python in publish pypi

alejoe91:


.github/workflows/publish-to-pypi.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Python 3.9
```
zm711: Name should be changed actually :)

alejoe91: `Set up Python` now :)

PR #4111, opened 2025-08-20
Update deprecated cuda import

### LINKED ISSUE #4110: Deprecated `cuda.cuda` module import
With `cuda-python>=13.0`:
https://github.com/SpikeInterface/spikeinterface/blob/53e02e29446f50854cb409868982a3cb5688bf01/src/spikeinterface/sorters/utils/misc.py#L68
throws an ImportError
> ...spikeinterface/sorters/utils/misc.py", line 68, in has_nvidia
>    from cuda import cuda
> ImportError: cannot import name 'cuda' from 'cuda' (unknown location)

Using `cuda-python<13.0`, we get
> FutureWarning: The cuda.cuda module is deprecated and will be removed in a future release, please switch to use the cuda.bindings.driver module instead.

lochhh:
Closes #4110

This PR also pins the minimum `cuda-python` version to `11.8.6`
> [!NOTE]
> The new module layouts were applied in [11.8.4](https://nvidia.github.io/cuda-python/cuda-bindings/latest/release/11.8.4-notes.html#cuda-namespace-cleanup-with-a-new-module-layout), but I was getting: 
> > Could not find a version that satisfies the requirement cuda-python==11.8.4; platform_system != "Darwin" and extra == "full" (from spikeinterface[full]) (from versions: 11.8.6, 11.8.7, 12.8.0, 12.9.0, 12.9.1, 12.9.2, 13.0.0, 13.0.1)

### DIFF
--- a/src/spikeinterface/sorters/utils/misc.py
+++ b/src/spikeinterface/sorters/utils/misc.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
-import subprocess  # TODO: decide best format for this
-from subprocess import check_output, CalledProcessError
 import importlib.util
+import subprocess  # TODO: decide best format for this
+from subprocess import CalledProcessError, check_output
 
 
 class SpikeSortingError(RuntimeError):
@@ -13,7 +13,7 @@ def get_bash_path():
     """Return path to existing bash install."""
     try:
         return check_output(["which bash"], shell=True).decode().strip("\n")
-    except CalledProcessError as e:
+    except CalledProcessError:
         raise Exception("Bash is not installed or accessible on your system.")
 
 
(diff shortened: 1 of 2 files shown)


samuelgarcia:
Hi @lochhh.
Thank you for this.
OK for me too do we need to pin a minimum cide version so ?

h-mayorquin:
+1 for minimum pin.

lochhh:
Thanks all for the speedy review. Great point regarding pinning - I've pinned it to >=11.8.6 and added some notes above.

src/spikeinterface/sorters/utils/misc.py
```diff
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
-import subprocess  # TODO: decide best format for this
-from subprocess import check_output, CalledProcessError
 import importlib.util
+import subprocess  # TODO: decide best format for this
+from subprocess import CalledProcessError, check_output
```
zm711, line 5: No need to change in this PR because this is not your doing, but why do we need the double import? can't we just call subprocess CalledProcessError? What was the thinking here?

PR #4123, opened 2025-09-01
Check install ci - use uv

alejoe91:
This PR updates the Check installation tips action to use uv

Currently working with ubuntu and windows, but fails with a seg fault on mac: https://github.com/SpikeInterface/spikeinterface/actions/runs/17374507900

At least we have something to debug :)


.github/workflows/installation-tips-test.yml
```diff
@@ -1,40 +1,36 @@
...
-        activate-environment: si_env
+        python-version: '3.12'
+    - name: Install uv
+      uses: astral-sh/setup-uv@v6
+
+    - name: Install requirements
+      run: uv pip install -r ./installation_tips/beginner_requirements_stable.txt
     - name: Check Installation Tips
-      run: python ./installation_tips/check_your_install.py --ci # ci flag turns off gui
+      run: python ./installation_tips/check_your_install.py --ci --short # ci flag turns off gui
```
samuelgarcia, line 37: You need to source activate no ?

alejoe91: No because of this flag:
https://github.com/SpikeInterface/spikeinterface/blob/check-install-ci/.github/workflows/installation-tips-test.yml#L9

zm711:
I can work on the mac at the end of this week see if I can figure out what is going on.

zm711:
so the mac seg-fault is only happening during kilosort4 clustering. As macs don't use the cuda framework and would be attempting to sort on cpu I think they are low-priority for the KS4 folks. We couple that with the fact that KS4 installation instructions can be a bit complicated to get the torch side done I would say it might be best to just skip the KS4 test for macs. Looking at the KS4 issue tracker it is unclear if they officially support numpy >2.0 yet and their testing only runs on python 3.9 and 3.10 so running this action on 3.12 isn't necessarily the best idea if we are also doing a KS4 test. I would suggest we either running TDC2 and SC2 since they are internal for this test and skip KS4 or have a discussion about why we want KS4. It is also an extremely heavy install with torch. Let me know if we want to test this on python 3.10 for KS4 purpose or whether we want to think of a different strategy for check your install which I think should emphasize testing spikeinterface rather than the wrapper of a specific sorter.

zm711:
Maybe of interest for this #4131 ? Maybe not just a mac + KS4 issue, but also linux + the sorting analyzer?

alejoe91:
It works! https://github.com/SpikeInterface/spikeinterface/actions/runs/17609442331
Let's merge :)

zm711 (approved):
Good by me too.

PR #4127, opened 2025-09-03
Fix KeyError when loading Kilosort 2.5 output with files missing cluster id

### LINKED ISSUE #4124: Failure to read kilosort 2 output when `cluster_info` is not available and `spike_times.csv` is included
This comes from here:
https://github.com/catalystneuro/neuroconv/issues/1492

I have an example of Kilosort 2.5 data for which there are no `cluster_info" files.

This implies this list is empty:
https://github.com/SpikeInterface/spikeinterface/blob/338a9f8ac1555155b068c627ae71a0153864792f/src/spikeinterface/extractors/phykilosortextractors.py#L64-L66

This leads us to this branch:

https://github.com/SpikeInterface/spikeinterface/blob/338a9f8ac1555155b068c627ae71a0153864792f/src/spikeinterface/extractors/phykilosortextractors.py#L80-L90

The problem there is that the first file in `all_property_files` here:

https://github.com/SpikeInterface/spikeinterface/blob/338a9f8ac1555155b068c627ae71a0153864792f/src/spikeinterface/extractors/phykilosortextractors.py#L78

is the `spike_map.csv` which does not have a `cluster_id`

Image

Which then errors here because properties should be aggregated and merged on the `cluster_id`column.

https://github.com/SpikeInterface/spikeinterface/blob/338a9f8ac1555155b068c627ae71a0153864792f/src/spikeinterface/extractors/phykilosortextractors.py#L90

It seems to me that there might be two problems:
* This `spike_map`file is missing the `cluster_id`
* the `spike_map` file should not be aggregated to aggregate properties.

I can take a deeper look into the kilosort documentation but maybe the solution is obvious to you @zm711 and @alejoe91 

These are the files available on the data.

```bash
amplitudes.npy
chanMap.mat
channel_map.npy
channel_positions.npy
cluster_Amplitude.tsv
cluster_ContamPct.tsv
cluster_group.tsv
cluster_KSLabel.tsv
motion.npy
NP153_B2_g0_tcat.imec0.log
NP153_B2_g0_tcat.imec0.sh
params.py
pc_feature_ind.npy
pc_features.npy
rez.mat
similar_templates.npy
spike_clusters.npy
spike_map.csv
spike_templates.npy
spike_times.npy
template_feature_ind.npy
template_features.npy
templates_ind.npy
templates.npy
whitening_mat_inv.npy
whitening_mat.npy
```

h-mayorquin:
Fixes #4124

I also added some documentation to the docstring and improved the warning about wrong imports by specifying the class that is actually causing the problem.

### DIFF
--- a/src/spikeinterface/extractors/phykilosortextractors.py
+++ b/src/spikeinterface/extractors/phykilosortextractors.py
@@ -40,7 +40,8 @@ class BasePhyKilosortSortingExtractor(BaseSorting):
     The cluster_id column is used as the merge key to combine properties from multiple files.
     All loaded properties are added to the sorting extractor as unit properties, with some
     renamed for SpikeInterface conventions: 'group' becomes 'quality', 'cluster_id'
-    becomes 'original_cluster_id'. These properties can be accessed via ``sorting.get_property()`` function.
+    becomes 'original_cluster_id'. These properties can be accessed via ``sorting.get_property()`` 
+    function.
     """
 
     installation_mesg = (


src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -24,6 +24,23 @@ class BasePhyKilosortSortingExtractor(BaseSorting):
...
+    1. From a dedicated cluster_info.csv/.tsv file if present
+    2. From all .csv/.tsv files in the folder that contain a 'cluster_id' column
+       Typical files include cluster_group.tsv, cluster_info.tsv, cluster_KSLabel.tsv
+       Files without cluster_id column are automatically skipped
+    3. If no files are found, minimal cluster info is generated with 'unsorted' labels
+
+    The cluster_id column is used as the merge key to combine properties from multiple files.
+    All loaded properties are added to the sorting extractor as unit properties, with some
+    renamed for SpikeInterface conventions: 'group' becomes 'quality', 'cluster_id'
+    becomes 'original_cluster_id'. These properties can be accessed via get_property().
```
alejoe91:
```suggestion
    becomes 'original_cluster_id'. These properties can be accessed via ``sorting.get_property()`` function.
```

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -24,6 +24,23 @@ class BasePhyKilosortSortingExtractor(BaseSorting):
...
+    1. From a dedicated cluster_info.csv/.tsv file if present
+    2. From all .csv/.tsv files in the folder that contain a 'cluster_id' column
+       Typical files include cluster_group.tsv, cluster_info.tsv, cluster_KSLabel.tsv
+       Files without cluster_id column are automatically skipped
+    3. If no files are found, minimal cluster info is generated with 'unsorted' labels
+
+    The cluster_id column is used as the merge key to combine properties from multiple files.
+    All loaded properties are added to the sorting extractor as unit properties, with some
+    renamed for SpikeInterface conventions: 'group' becomes 'quality', 'cluster_id'
+    becomes 'original_cluster_id'. These properties can be accessed via ``sorting.get_property()`` function.
```
alejoe91:
```suggestion
    becomes 'original_cluster_id'. These properties can be accessed via ``sorting.get_property()`` 
    function.
```

PR #4135, opened 2025-09-10
Fix `plot_traces` with `t_start` and no times

alejoe91:
In the `plot_traces` (`ipywidgets` backend), we weren't handling `t_start` correctly when in the trace retrieval and `TimeSlider`. This PR fixes the behavior.


alejoe91:
> LGTM, and works for my local data (which has no times but a t_start). Good catch!

That's how I caught it :P

samuelgarcia:
Will this work wirh spike on traces ? Did you check  ?

alejoe91:
> Will this work wirh spike on traces ? Did you check  ?

It should, since spikes on traces uses the plot_traces and return_times=True, but I'll check!!

PR #4136, opened 2025-09-10
`sd_ratio`: avoid recomputing noise levels

alejoe91:
Added `noise_levels` as a dependency to `sd_ratio`, to avoid recomputing noise levels if already available. This makes the metric recordingless compatible too


alejoe91:
Ok no `sd_ratio` needs noise levels as `std`, which makes a big difference in the computation. Closing!

chrishalcrow:
haha, I did this exact same thing a month ago

PR #4137, opened 2025-09-11
Metric list in plot

alejoe91:
Add a metric list panel on the right side of the `plot_*_metrics` in `ipywidgets` backend.

### DIFF
--- a/src/spikeinterface/widgets/metrics.py
+++ b/src/spikeinterface/widgets/metrics.py
@@ -206,7 +206,6 @@ class MetricsBaseWidget(BaseWidget):
         if self.data_plot["selected_columns"] is None or set(selected_metrics) != set(
             self.data_plot["selected_columns"]
         ):
-            print("updating plot...")
             self.data_plot["unit_ids"] = unit_ids
             self.data_plot["selected_columns"] = selected_metrics
             self.figure.clf()


src/spikeinterface/widgets/metrics.py
```diff
@@ -172,27 +187,58 @@ def _update_ipywidget(self, change):
...
-            for p in self.patches:
-                p.set_color(colors)
-                p.set_sizes(sizes)
-        else:
+            alphas.append(alpha)
+
+        if self.data_plot["selected_columns"] is None or set(selected_metrics) != set(
+            self.data_plot["selected_columns"]
+        ):
+            print("updating plot...")
```
chrishalcrow: Do we want this `print`?

chrishalcrow:
Very nice!!!
Screenshot 2025-09-11 at 12 38 32

Could you add `selected_columns` as an arg for `plot_quality_metrics` and `plot_template_metrics` and pass it through? If you do, it allows for the matplotlib backend to make some nice plots:

Screenshot 2025-09-11 at 12 49 59

src/spikeinterface/widgets/metrics.py
alejoe91: No :P

alejoe91:
> Could you add `selected_columns` as an arg for `plot_quality_metrics` and `plot_template_metrics` and pass it through? If you do, it allows for the matplotlib backend to make some nice plots:
> 

We already have `include_metrics` and `skip_metrics`, but these are used in the init to filter the metrics df at the beginning ;)

src/spikeinterface/widgets/metrics.py
alejoe91:
```suggestion
```

PR #4138, opened 2025-09-11
Improve docs for curation module and model

alejoe91:
Added `autodoc_pydantic`, looks sweet!

image

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -63,6 +63,7 @@ extensions = [
     'sphinx.ext.autosummary',
     'sphinx_gallery.gen_gallery',
     'numpydoc',
+    'sphinxcontrib.autodoc_pydantic',
     'sphinx.ext.autosectionlabel',
     'sphinx_design',
     'sphinxcontrib.jquery',
@@ -76,6 +77,8 @@ autosectionlabel_prefix_document = True
 
 numpydoc_show_class_members = False
 
+autodoc_pydantic_model_show_json = True
+autodoc_pydantic_model_show_config_summary = False
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
--- a/src/spikeinterface/curation/curation_format.py
+++ b/src/spikeinterface/curation/curation_format.py
@@ -243,7 +243,7 @@ def apply_curation(
     if isinstance(curation_dict_or_model, dict):
         curation_model = CurationModel(**curation_dict_or_model)
     else:
-        curation_model = curation_dict_or_model
+        curation_model = curation_dict_or_model.model_copy(deep=True)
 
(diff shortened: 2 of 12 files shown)


doc/conf.py
```diff
@@ -76,6 +77,8 @@
 
 numpydoc_show_class_members = False
 
+autodoc_pydantic_model_show_json = True
```
zm711, line 80: What is the value of showing the json? When I opened that drop down it was a giant code chunk of a dict. Will people find this useful in the documentation? Is it for developers? This is the only part that I'm not sure about.

alejoe91: Some people might be familiar with the json-schema, which is a more universal validation/schema language. Since by default it's collapsed, I think it doesn't hurt keepoing it there

src/spikeinterface/curation/curation_format.py
```diff
@@ -243,7 +243,7 @@ def apply_curation(
     if isinstance(curation_dict_or_model, dict):
         curation_model = CurationModel(**curation_dict_or_model)
     else:
-        curation_model = curation_dict_or_model
+        curation_model = curation_dict_or_model.model_copy(deep=True)
```
alejoe91, line 201: @samuelgarcia @chrishalcrow this is needed otherwise the curation model is modified in place with the new_unit_ids. While this might seem ok, it makes the curation invalid since new unit ids are computed after removing units, so this makes the updated curation unusable. I can explain more

doc/conf.py
zm711: That sounds good to me.

src/spikeinterface/curation/curation_format.py
zm711: Should we add a comment to the code then? I can imagine someone in the future wanting to be careful with memory usage assuming we don't need to do a deep copy.

samuelgarcia: As discussed yesterday.
We should not modify this curation_model inplace at all.
So no need to copy.

src/spikeinterface/curation/curation_format.py
```diff
@@ -280,8 +280,6 @@ def apply_curation(
                 verbose=verbose,
                 **job_kwargs,
             )
-        for i, merge_unit_id in enumerate(new_unit_ids):
-            curation_model.merges[i].new_unit_id = merge_unit_id
```
alejoe91, line 284: @samuelgarcia @zm711 @chrishalcrow we actually need these because the new ids are used when applying labels. I'll revert to the deepcopy approach

alejoe91:
@samuelgarcia @chrishalcrow this should be good to go!

samuelgarcia:
yep

Issue #4133, opened 2025-09-09
Request: ability to select time windows within SortingAnalyzers


lfrank:
We are working on linking units across sortings, and as there is always some unit instability over time, we've found that this requires comparing the waveform properties from the end of one recording with those from the beginning of the next recording, particularly for continuous recordings where we're sorting a day at a time.

We'd therefore like to be able to specify template extraction and potentially other metrics limited to specific windows within a sorting as a SortingAnalyzer feature so that users can do the matching across datasets.

I think this could be broadly helpful and will be a more and more important feature as people move toward longer recordings.

zm711:
I absolutely agree with this and keep meaning to bring this up at maintenance meetings. We've been meaning to extend the `random_spikes` to allow for finer control (ie get waveforms from specific time periods) for a while but no one has taken it on. I need this too for my work as I would really love to easily demonstrate that the waveforms aren't changing during stimuli as a control metric (especially during opto which is many people do these days). Once we have a `random_spikes` that is fine-tuneable I could imagine either a slicing mechanism which seems complicated or just a recommendation to make 5 analyzers subsetting your data and running the same pipeline on the self cut-up data. 

Is one of those options what you'd want too?

lfrank:
In a perfect world I’d want a very flexible slicer where we could specify the time intervals or spike indexes, perhaps in something like a numpy indexing format so you could do [-1000:0] to specify the last 1000 elements.

If that’s too difficult to implement, at minimum we’d need to be able to specify the time windows, as depending on the level of instability and the length of the recording, we might want the last 500 or 5000 spikes, or perhaps the last 10-30 minutes of spiking in a 24 hour period.

On Sep 9, 2025, at 5:07 AM, Zach McKenzie ***@***.***> wrote:


This Message Is From an External Sender
This message came from outside your organization.
[https://avatars.githubusercontent.com/u/92116279?s=20&v=4]zm711 left a comment (SpikeInterface/spikeinterface#4133)<https://urldefense.com/v3/__https://github.com/SpikeInterface/spikeinterface/issues/4133*issuecomment-3270403985__;Iw!!LQC6Cpwp!oSKcTcgW4Q6J0_ooyNEWf8-2tKHQSAk4ADnPKqn6XsbbGqWObUmRwUd66tHFqUhbpF4iAOLn9n6eb4y_3U82WjqN6Q4$>

[...]

PR #4147, opened 2025-09-18
Proposal: Add `analyzer.frame_slice`

chrishalcrow:
Following on from #4133, here is a draft proposal for implementing a `frame_slice` method for the sorting analyzer, to see how tricky it is.

API is:

```
sliced_analyzer = analyzer.frame_slice(start_frame=1000, end_frame=2000)
```

which matches the `frame_slice` methods for sorting and recording objects.

There's a "hard" mode (which just recomputes all extensions) and a "soft" mode (which attempts to estimate the extensions). Almost all the complexity occurs at the level of the extension. For each extension, we'd need to implement a `_frame_slice_extension_data` method.

I've only implemented `_frame_slice_extension_data` for `random_spikes`, and you can only return an in memory analyzer for the moment.

Would like some feedback on the basic structure before doing anything more.

### DIFF
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -93,6 +93,22 @@ class ComputeRandomSpikes(AnalyzerExtension):
             new_data["random_spikes_indices"] = np.flatnonzero(selected_mask[keep_mask])
         return new_data
 
+    def _frame_slice_extension_data(
+        self,
+        start_frame,
+        end_frame,
+    ):
+
+        new_data = dict()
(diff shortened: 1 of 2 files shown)


yger:
I think this is a great idea, and I don't think of any obvisous flaws in the design as it is proposed. I'm afraid lots of extensions would need to be recomputed however, but this would be a cool feature

alejoe91:
I agree with @yger here! It doesn't hurt and it could be beneficial in many cases! Of course, we should add proper warnings if things aren't as expected, e.g., too few sampled spikes, etc.

PR #4149, opened 2025-09-26
Fix diagonal plots and labels in plot_metrics

alejoe91:
The x-y axes were swapped for off diagonal and this PR only plots the upper triangular matrix


chrishalcrow (approved):
Code looks fine, and plots look good.

PR #4163, opened 2025-10-07
Update generate.py

OleBialas:
Hello.,

The docstring of `generate_templates` says:

```
    unit_params : dict[np.array] | dict[float] | dict[tuple] | None, default: None
        An optional dict containing parameters per units.
        Keys are parameter names:

            * "alpha": amplitude of the action potential in a.u. (default range: (6'000-9'000))
```

However, in `default_unit_params_range` (line 1649), alpha is defined differently

```python
default_unit_params_range = dict(
    alpha=(100.0, 500.0),
    ...
```

I have updated the docstring to reflect this

Best,
Ole

### DIFF
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -1733,7 +1733,7 @@ def generate_templates(
         An optional dict containing parameters per units.
         Keys are parameter names:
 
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Hi.
thank you for the catch.
I am currently playing with the default params in this #4167
So lets wait it to be merge.

samuelgarcia:
merci

PR #4167, opened 2025-10-10
Fix bug in waveform generation + adapt parameters

samuelgarcia:
This fix a bug in how we do the generation the spattial waveform when it is an ellispoid.
It bug had an impact on the waveform amplitude itself.

The new way is a bit more simple but less buggy with no border effect on amplitude.

This also change the parameters range and names to avoid confusion.

Additionally it remove all the `np.random.rand`  `np.random.randn` `np.rand.choice` and use explicit randome generator with clear seed.

### DIFF
--- a/src/spikeinterface/sorters/internal/tridesclous2.py
+++ b/src/spikeinterface/sorters/internal/tridesclous2.py
@@ -114,7 +114,9 @@ class Tridesclous2Sorter(ComponentsBasedSorter):
             if params["apply_motion_correction"]:
                 rec_for_motion = recording_raw
                 if params["apply_preprocessing"]:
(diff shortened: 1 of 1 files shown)


samuelgarcia:
@yger @alejoe91 @chrishalcrow : this is ready to merge on my side.

src/spikeinterface/core/generate.py
```diff
@@ -296,7 +296,7 @@ def generate_sorting_to_inject(
             injected_spike_train = injected_spike_train[~violations]
```
chrishalcrow, line 296: In line 285 (10 lines above!) should `np.random.uniform` be seeded? Or just a rng?

chrishalcrow:
IGNORE ME

~~I'm just looking for seeds which might be making the tests flakey. Should line 2169 in `generate.py` in `_generate_multimodal` be seed-able?~~

samuelgarcia:
Tests are passing. I merge this now.

PR #4168, opened 2025-10-10
Fix pydantic model after validator

alejoe91:


samuelgarcia:
Good for me but then we should clip to a pydantic minimal version no ?

In this PR

samuelgarcia:
Alessio say that that is bac compatible. So I merge it because the main is failing

PR #4171, opened 2025-10-14
faster write_binary_recording()

samuelgarcia:
The `write_binary_recording()` was using the memmap machinery but this is a bad idea for writing because it synchronize the buffer with the disk before writing with a read and this make it more slow.

Of some system with netowrk drive this PR make the writing to binary 2 times faster witth this simple PR.

@h-mayorquin : I did also some test on the reading side, but there is no effect on perf so for now I keep memmap machinery on reading side. But maybe we could switch to this also in binary extarctor : simple seek and np.fromfile


chrishalcrow:
Looks good. For me locally, this makes a big difference (almost x2) for `n_jobs=1` (just re-saving an untouched openephys file), and the improvement diminishes as `n_jobs` increases.

h-mayorquin (approved):
LGTM

We discussed this in person, this is great.

alejoe91:
@samuelgarcia time for pre-commit please :)

PR #4173, opened 2025-10-15
Support python 3.14

h-mayorquin:


chrishalcrow:
From trying locally on macOS: I had to pin `pydantic>=2.12`. Then everything worked, but no numba support yet. They're aiming for a "few weeks" till release.

chrishalcrow:
Numba 0.63, which supports 3.14, is out!

h-mayorquin:
There is some dependency resolution thing remaining, I will take a look.

chrishalcrow:
Failures were different for all three systems (oi oi oi). Working fine for me locally (mac) with same version of numcodecs as was failing on the ci. Trying to run again... still not working. Hmmm.

h-mayorquin:
Ok, this is another bullet in the same direction. I investigated and to get this through we need to add zarr 3.0 support : /

alejoe91:
Depends on #4260

PR #4177, opened 2025-10-20
Add `experiment_name`  argument to  `OpenEphysBinaryRecordingExtractor`

h-mayorquin:
The current `block_index` parameter exposes a Neo implementation detail that most users won't understand. Since Open Ephys organizes data into named experiment folders (experiment1, experiment2, etc.), I've added an `experiment_name` parameter that lets users specify experiments directly by folder name (e.g., `experiment_name="experiment2"`). I've also improved error messages to list available experiments instead of just mentioning `block_index`.

I add the `experiment_name` parameter to `OpenEphysBinaryRecordingExtractor` with full backward compatibility. I deprecate `block_index` with a `FutureWarning` (removal in 0.105.0) and add a `get_available_experiments()` class method for discovery. My implementation passes `experiment_names=[experiment_name]` to Neo to filter efficiently. I've enhanced the docstring with the Open Ephys binary format and how it maps to SpikeInterface concepts.

I'm also deprecating `experiment_names` (plural). This parameter was added to Neo in [PR #1159](https://github.com/NeuralEnsemble/python-neo/pull/1159) to handle inconsistent stream configurations across experiments, but this doesn't apply to SpikeInterface since we only load one experiment at a time. The new `experiment_name` (singular) is clearer for our single-block loading model and achieves the same goal of filtering out incompatible streams so the data can be loaded.

[EDIT] After discussion below, we are keeping block index.

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/openephys.py
+++ b/src/spikeinterface/extractors/neoextractors/openephys.py
@@ -161,7 +161,8 @@ class OpenEphysBinaryRecordingExtractor(NeoBaseRecordingExtractor):
     Open Ephys Binary Format Structure:
         folder_path/
         ├── Record Node 102/              # Recording hardware node
-        │   ├── settings.xml
+        │   ├── settings.xml              # Settings for the first experiment
+        │   ├── settings_2.xml            # Settings for experiment 2
         │   ├── experiment1/              # Experiment folder
         │   │   ├── recording1/           # Recording session (SpikeInterface segment)
         │   │   │   ├── structure.oebin   # JSON metadata file


alejoe91:
Thanks @h-mayorquin 

Please let's not deprecate `block_index` though! We have convenient functions to get the number of blocks and stream names for NEO extractors. These can be used to traverse the blocks/streams and are angnostic to the specific reader. Eg.
```
from spikeinterface.extractors import recording_extractor_full_dict

reader_type = "openephysbinary" # or any other neo
reader_kwargs = {"folder_path": "path-to-folder"}
num_blocks = se.get_neo_num_blocks(reader_type)
stream_names, stream_ids = se.get_neo_streams(reader_type)

for block_index in range(num_blocks):
    for stream_name in stream_names:
        recording = recording_extractor_full_dict[reader_type](**reader_kwargs)
```

alejoe91 (changes requested):
LGTM, but let's not deprecate block_index pls :)

h-mayorquin:
Undid the deprecation. I was thinking that only this extractor uses the block index but I realize some others do:

MaxwellRecordingExtractor, MCSRawRecordingExtractor NixRecordingExtractor,TdtRecordingExtractor

So now the logic is that you can use either but not both.

alejoe91:
> Undid the deprecation. I was thinking that only this extractor uses the block index but I realize some others do:
> 
> MaxwellRecordingExtractor, MCSRawRecordingExtractor NixRecordingExtractor,TdtRecordingExtractor
> 
> So now the logic is that you can use either but not both.

Perfect!

src/spikeinterface/extractors/neoextractors/openephys.py
```diff
@@ -106,69 +106,195 @@ def map_to_neo_kwargs(cls, folder_path):
...
+        │   ├── experiment1/              # Experiment folder
+        │   │   ├── recording1/           # Recording session (SpikeInterface segment)
+        │   │   │   ├── structure.oebin   # JSON metadata file
+        │   │   │   ├── continuous/       # Signal streams
+        │   │   │   │   └── Neuropix-PXI-100.ProbeA-AP/
+        │   │   │   │       ├── continuous.dat
+        │   │   │   │       └── timestamps.npy
+        │   │   │   └── events/           # Event streams
+        │   │   └── recording2/           # Additional recording (additional segment)
+        │   └── experiment2/              # Different experiment
```
alejoe91, line 175: maybe add `settings_2.xml` which is generated for experiment2?

h-mayorquin: Done.

chrishalcrow:
Hello, think we'll need to pin neo to >=0.14.3 so that `_parse_folder_structure` is exposed. Agree?

h-mayorquin:
> Hello, think we'll need to pin neo to >=0.14.3 so that `_parse_folder_structure` is exposed. Agree?

Agree. I should not have used a private internal detail of neo as well : /

PR #4178, opened 2025-10-21
Improve assertion error message of SortingAnalyzer job kwargs

MGAMZ:
When using `SortingAnalyzer` and `compute`, if user specifies too many args, there will be an assertion error. But the assertion message does not contain the detailed message about which arg violates the API.

So I wish to improve the assertion message. The modification is simple and straightforward, should not introduce any logical breaking.

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -1635,7 +1635,8 @@ extension_params={"waveforms":{"ms_before":1.5, "ms_after": "2.5"}}\
         elif isinstance(input, dict):
             params_, job_kwargs = split_job_kwargs(kwargs)
             assert len(params_) == 0, (
-                "Too many arguments for SortingAnalyzer.compute_several_extensions(), " f"they are: {set(params_)}"
+                "Too many arguments for SortingAnalyzer.compute_several_extensions(), "
+                f"please remove the arguments {set(params_)} from the compute function."
             )
             self.compute_several_extensions(extensions=input, save=save, verbose=verbose, **job_kwargs)
         elif isinstance(input, list):


MGAMZ:
Copilot suggestions have been partially accepted, not accepting the second suggestion.

Now the code LGTM.

chrishalcrow:
Hey @MGAMZ, it would be great if the error message told the user what to do to fix the error. Maybe instead of "they are: ..." we could say something like "please remove the arguments {set(params_)} from the compute function.". What'dya think?

MGAMZ:
> Hey @MGAMZ, it would be great if the error message told the user what to do to fix the error. Maybe instead of "they are: ..." we could say something like "please remove the arguments {set(params_)} from the compute function.". What'dya think?

That would be better~ Commit done.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1634,11 +1634,16 @@ def compute(self, input, save=True, extension_params=None, verbose=False, **kwar
             return self.compute_one_extension(extension_name=input, save=save, verbose=verbose, **kwargs)
         elif isinstance(input, dict):
             params_, job_kwargs = split_job_kwargs(kwargs)
-            assert len(params_) == 0, "Too many arguments for SortingAnalyzer.compute_several_extensions()"
+            assert len(params_) == 0, (
+                "Too many arguments for SortingAnalyzer.compute_several_extensions(), " f"they are: {set(params_)}"
```
```suggestion
                "Too many arguments for SortingAnalyzer.compute_several_extensions(), "
                f"please remove the arguments {set(params_)} from the compute function."
```

sorry, also here :)

MGAMZ:
@chrishalcrow My fault, missed another place. 😓

alejoe91 (approved):

PR #4190, opened 2025-10-31
Add SLAy auto-merge preset

chrishalcrow:
Add SLAy as a method for `compute_merge_unit_groups` https://www.biorxiv.org/content/10.1101/2025.06.20.660590v1

SLAy combines a template similarity measure, a cross-correlation significance measure and a sliding refractory period violation measure to make a merge decision. You need an analyzer with templates, correlograms and template_similarity. If you do, you can try it using:

``` python
merges = si.compute_merge_unit_groups(analyzer, preset="slay")
```

or, for more control, e.g.

``` python
merges = si.compute_merge_unit_groups(sa, preset="slay", steps_params ={'slay_score': {'k1': 0.7, 'slay_threshold': 0.8}})
```

This is an approximate implementation.  Some details
- For now, we use a simple template_similarity instead of the auto-encoder
- We port the `_sliding_RP_viol` and `_compute_xcorr` functions from https://github.com/saikoukunt/SLAy . I think the sliding RP computation is fairly different between our implementation in quality_metrics. Will take a closer look...

To do:

- [x] unit tests
- [x] test against results in paper

With help and permission from @saikoukunt

### DIFF
--- a/src/spikeinterface/curation/auto_merge.py
+++ b/src/spikeinterface/curation/auto_merge.py
@@ -120,6 +120,8 @@ def compute_merge_unit_groups(
         * "cross_contamination": the cross-contamination is not significant (`cc_thresh` and `p_value`)
         * "knn": the two units are close in the feature space
         * "quality_score": the unit "quality score" is increased after the merge
+        * "slay_score":  a combined score, factoring in a template similarity measure, a cross-correlation significance measure
+          and a sliding refractory period violation measure, based on the SLAy algorithm.
 
     The "quality score" factors in the increase in firing rate (**f**) due to the merge and a possible increase in
     contamination (**C**), wheighted by a factor **k** (`firing_contamination_balance`).
@@ -151,6 +153,9 @@ def compute_merge_unit_groups(
         * | "feature_neighbors": focused on finding unit pairs whose spikes are close in the feature space using kNN.
           | It uses the following steps: "num_spikes", "snr", "remove_contaminated", "unit_locations",
           | "knn", "quality_score"
+        * | "slay": an approximate implementation of SLAy, original implementation at https://github.com/saikoukunt/SLAy.
+          | The spikeinterface version uses `template_similarity`, rather than an auto-encoder.
+          | It uses the following steps: "template_similarity", "slay_score"
 
         If `preset` is None, you can specify the steps manually with the `steps` parameter.
     resolve_graph : bool, default: True
@@ -564,6 +569,7 @@ def get_potential_auto_merge(
         * "cross_contamination": the cross-contamination is not significant (`cc_thresh` and `p_value`)
         * "knn": the two units are close in the feature space
         * "quality_score": the unit "quality score" is increased after the merge
+        * "slay_score":  a combined score, factoring in a template similarity measure, a cross-correlation significance measure and a sliding refractory period violation measure, based on the SLAy algorithm.
 
     The "quality score" factors in the increase in firing rate (**f**) due to the merge and a possible increase in
     contamination (**C**), wheighted by a factor **k** (`firing_contamination_balance`).
@@ -580,7 +586,7 @@ def get_potential_auto_merge(
     ----------
     sorting_analyzer : SortingAnalyzer
         The SortingAnalyzer
-    preset : "similarity_correlograms" | "x_contaminations" | "temporal_splits" | "feature_neighbors" | None, default: "similarity_correlograms"
+    preset : "similarity_correlograms" | "x_contaminations" | "temporal_splits" | "feature_neighbors" | "slay" | None, default: "similarity_correlograms"
         The preset to use for the auto-merge. Presets combine different steps into a recipe and focus on:
 
         * | "similarity_correlograms": mainly focused on template similarity and correlograms.
@@ -595,6 +601,9 @@ def get_potential_auto_merge(
         * | "feature_neighbors": focused on finding unit pairs whose spikes are close in the feature space using kNN.
           | It uses the following steps: "num_spikes", "snr", "remove_contaminated", "unit_locations",
           | "knn", "quality_score"
+        * | "slay": an approximate implementation of SLAy, original implementation at https://github.com/saikoukunt/SLAy.
+          | The spikeinterface version uses `template_similarity`, rather than an auto-encoder.
+          | It uses the following steps: "template_similarity", "slay_score"
 
         If `preset` is None, you can specify the steps manually with the `steps` parameter.
     resolve_graph : bool, default: False


src/spikeinterface/curation/auto_merge.py
```diff
@@ -356,6 +362,12 @@ def compute_merge_unit_groups(
             )
             outs["pairs_decreased_score"] = pairs_decreased_score
 
+        elif step == "slay_score":
+
+            M_ij = compute_slay_matrix(sorting_analyzer, params["k1"], params["k2"], params["template_diff_thresh"])
```
alejoe91: to save time, we should compute this only on the `pair_mask` that survived the template similarity step!

This could be an optional input of the function: if given, all pairs with mask False will be set to 0, so we keep the correct shape

src/spikeinterface/curation/auto_merge.py
```diff
@@ -356,6 +362,12 @@ def compute_merge_unit_groups(
...
+
+            pair_mask = M_ij > params["slay_threshold"]
```
alejoe91:
```suggestion
            pair_mask = pair_mask & (M_ij > params["slay_threshold"])
```

src/spikeinterface/curation/auto_merge.py
```diff
@@ -1506,3 +1518,235 @@ def estimate_cross_contamination(
...
+    Paramters
+    ---------
+    sorting_analyzer : SortingAnalyzer
+        The sorting analyzer object containing the spike sorting data
+    k1 : float
+        Coefficient determining the importance of the cross-correlation significance
+    k2 : float
+        Coefficient determining the importance of the sliding rp violation
+    template_diff_thresh : float
+        Threshold for how different template similarities can be to be considered for merging
```
alejoe91: This is not used here, right?

src/spikeinterface/curation/auto_merge.py
```diff
@@ -1506,3 +1518,235 @@ def estimate_cross_contamination(
...
+    if xgram_window.sum() < (min_xcorr_rate / 4 * window_size):
+        sig = -4  # don't merge if the event count is way too low
+
+    return sig
+
+
+def _sliding_RP_viol_pair(
+    correlogram,
+    bin_size_ms: float,
+    acceptThresh: float = 0.15,
```
alejoe91:
```suggestion
    accept_threshold: float = 0.15,
```

src/spikeinterface/curation/auto_merge.py
```diff
@@ -1506,3 +1518,235 @@ def estimate_cross_contamination(
...
+    sig : float
+        The refractory period violation confidence for the cluster.
+    """
+    from scipy.signal import butter, sosfiltfilt
+    from scipy.stats import poisson
+
+    # create various refractory periods sizes to test (between 0 and 20x bin size)
+    b = np.arange(0, 21 * bin_size_ms, bin_size_ms) / 1000
+    bTestIdx = np.array([1, 2, 4, 6, 8, 12, 16, 20], dtype="int8")
+    bTest = [b[i] for i in bTestIdx]
```
alejoe91: can we do some renaming?

src/spikeinterface/curation/auto_merge.py
```diff
@@ -1506,3 +1518,235 @@ def estimate_cross_contamination(
...
+            # Don't waste time computing the other metrics if we fail the template similarity check
+            if template_similarity[unit_index_1, unit_index_2] < 1 - template_diff_thresh:
+                continue
+
+            xgram = ccgs[unit_index_1, unit_index_2, :]
+
+            rho_ij[unit_index_1, unit_index_2] = _compute_xcorr_pair(
+                xgram, bin_size_s=bin_size_ms / 1000, min_xcorr_rate=0
+            )
+            eta_ij[unit_index_1, unit_index_2] = _sliding_RP_viol_pair(xgram, bin_size_ms=bin_size_ms)
```
alejoe91: Ah I see! We could use the pair_mask instead, since we have already computed the template similarity step:
```
if pair_mask is None:
    pair_mask = np.ones(num_units, num_units, dtype=bool)

...

for unit_index_1, _ in enumerate(sorting_analyzer.unit_ids):
        for unit_index_2, _ in enumerate(sorting_analyzer.unit_ids):
            if not pair_mask[unit_index_1, unit_index_2]:
                continue
```

src/spikeinterface/curation/auto_merge.py
alejoe91: I think we can use the `pair_mask` from template similarity as input, to make use of the multiple steps (and give the option to the user to add more!):
```suggestion
            M_ij = compute_slay_matrix(sorting_analyzer, params["k1"], params["k2"], pair_mask=pair_mask)
```

src/spikeinterface/curation/auto_merge.py
```diff
@@ -1506,3 +1518,235 @@ def estimate_cross_contamination(
         )
 
     return estimation, p_value
+
+
+def compute_slay_matrix(sorting_analyzer: SortingAnalyzer, k1: float, k2: float, template_diff_thresh: float):
```
alejoe91:
```suggestion
def compute_slay_matrix(sorting_analyzer: SortingAnalyzer, k1: float, k2: float, pair_mask: np.ndarray):
```

src/spikeinterface/curation/auto_merge.py
```diff
@@ -1506,3 +1518,235 @@ def estimate_cross_contamination(
...
+    References
+    ----------
+    Based on computation originally implemented in SLAy [Koukuntla]_.
+
+    Implementation is based on one of the original implementations written by Sai Koukuntla,
+    found at https://github.com/saikoukunt/SLAy.
+    """
+
+    sigma_ij = sorting_analyzer.get_extension("template_similarity").get_data()
+    rho_ij, eta_ij = compute_xcorr_and_rp(sorting_analyzer, template_diff_thresh)
```
alejoe91:
```suggestion
    rho_ij, eta_ij = compute_xcorr_and_rp(sorting_analyzer, pair_mask)
```

src/spikeinterface/curation/auto_merge.py
```diff
@@ -1506,3 +1518,235 @@ def estimate_cross_contamination(
...
+
+    M_ij = sigma_ij + k1 * rho_ij - k2 * eta_ij
+
+    return M_ij
+
+
+def compute_xcorr_and_rp(sorting_analyzer: SortingAnalyzer, template_diff_thresh: float):
```
alejoe91:
```suggestion
def compute_xcorr_and_rp(sorting_analyzer: SortingAnalyzer, pair_mask: np.ndarray):
```

src/spikeinterface/curation/auto_merge.py
```diff
@@ -1506,3 +1518,235 @@ def estimate_cross_contamination(
...
+        sig *= (xgram_window.sum() / (min_xcorr_rate * window_size)) ** 2
+
+        # if sig < 0.04 and xgram_window.sum() < (min_xcorr_rate * window_size):
+    if xgram_window.sum() < (min_xcorr_rate / 4 * window_size):
+        sig = -4  # don't merge if the event count is way too low
+
+    return sig
+
+
+def _sliding_RP_viol_pair(
```
alejoe91, line 1741: Could we try to unify with this code @chrishalcrow ?
https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/qualitymetrics/misc_metrics.py#L1493-L1511

src/spikeinterface/curation/auto_merge.py
```diff
@@ -1506,3 +1518,235 @@ def estimate_cross_contamination(
...
+    nyqist = fs / 2
+    cutoff = cutoff_freq / nyqist
+    sos = butter(order, cutoff, btype="low", output="sos")
+    smoothed_acg = sosfiltfilt(sos, correlogram)
+
+    bin_rate_max = np.max(smoothed_acg)
+    max_conts_max = np.array(bTest) / bin_size_ms * 1000 * (bin_rate_max * acceptThresh)
+    # compute confidence of less than acceptThresh contamination at each refractory period
+    confs = 1 - poisson.cdf(sum_res, max_conts_max)
+    rp_viol = 1 - confs.max()
```
alejoe91, line 1796: https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/qualitymetrics/misc_metrics.py#L1518

src/spikeinterface/curation/auto_merge.py
chrishalcrow: Not now that we've added the `pair_mask`

src/spikeinterface/curation/auto_merge.py
chrishalcrow: Yess. I handled the `None` `pair_mask` at `compute_slay_metric` rather than here.

src/spikeinterface/curation/auto_merge.py
chrishalcrow: Probably...

src/spikeinterface/curation/auto_merge.py
```diff
@@ -1506,3 +1518,240 @@ def estimate_cross_contamination(
...
+    Based on computation originally implemented in SLAy [Koukuntla]_.
+
+    Implementation is based on one of the original implementations written by Sai Koukuntla,
+    found at https://github.com/saikoukunt/SLAy.
+    """
+
+    num_units = sorting_analyzer.get_num_units()
+
+    if pair_mask is None:
+        pair_mask = np.ones((num_units, num_units), dtype="bool")
```
alejoe91: This should already be the case here, since there is a previous step. However, in general, pair mask should be an upper triangular matrix, to about computing the metric for [u1, u2] and [u2,u1] :)

alejoe91 (changes requested):
One very minor detail!

samuelgarcia:
Et voilà!

PR #4200, opened 2025-11-06
Improve violation message on `n_jobs`

MGAMZ:
The current implementation when receives an invalid `n_jobs` during extensions computation DOES NOT include the incorrect input. It's better to tell users not only they are wrong but also what is the wrong input right now.

### DIFF
--- a/src/spikeinterface/core/job_tools.py
+++ b/src/spikeinterface/core/job_tools.py
@@ -153,8 +153,9 @@ def fix_job_kwargs(runtime_job_kwargs):
 
     # if n_jobs is -1, set to os.cpu_count() (n_jobs is always in global job_kwargs)
     n_jobs = job_kwargs["n_jobs"]
-    assert isinstance(n_jobs, (float, np.integer, int)) and n_jobs != 0, \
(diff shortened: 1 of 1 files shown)


zm711:
I am a little torn on this to be honest. I do appreciate better error messages in general, but for something so fundamental as an int or float it seems okay to just tell the user to use the int or float. Could you give me an example of when this has actually caused a problem?

MGAMZ:
@zm711 The original error was a simple mistake I made in my code. In my programming practice, I tend to report not only the error type but also the contextual state. That's why I submitted this PR. I think your point is also valid—if you feel it's unnecessary, I'm happy to close this PR.

I cannot provide the original erroneous code, but its pseudo-code structure is similar to the following:

```python
n_jobs = False # WRONG INPUT HERE
analyzer: sc.SortingAnalyzer
analyzer.compute(ext_name, save=save, n_jobs=n_jobs, **kwargs)
```

alejoe91:
Thanks @MGAMZ 

I actually agree with @. The styles are already quite clear in the current message! Let's close this!

Feel free to open new PRs if you find something is unclear

MGAMZ:
Closing this one~~

PR #4201, opened 2025-11-06
Add similarity_method param in automerge and fix multi-segment cross-contamination

alejoe91:


chrishalcrow:
Looks good! Could you add a multi-segment test to `curation/test/test_auto_merge.py`. A very simple test that runs all presets for a single- and multi-segement analzyer would be perfect.

(The current tests are a bit messy, I don't know why we need to keep recomputing analyzers when we have a fixture. Also the default analyzer is pretty long (5mins) so the tests are quite long -- 30s locally)

src/spikeinterface/curation/auto_merge.py
```diff
@@ -1054,28 +1060,34 @@ def compute_cross_contaminations(analyzer, pair_mask, cc_thresh, refractory_peri
     if pair_mask is None:
         pair_mask = np.ones((n, n), dtype="bool")
 
-    CC = np.zeros((n, n), dtype=np.float32)
-    p_values = np.zeros((n, n), dtype=np.float32)
-
-    for unit_ind1 in range(len(unit_ids)):
+    num_segments = sorting.get_num_segments()
+    CC = np.zeros((num_segments, n, n), dtype=np.float32)
+    p_values = np.zeros((num_segments, n, n), dtype=np.float32)
```
samuelgarcia: why this ?

samuelgarcia:
Why having a CC by segment ? what is the rational behing this ?

src/spikeinterface/curation/auto_merge.py
alejoe91: because it was triggering an error for multi-segment. How would you handle multi-segment? Right now it computes CC for each segment and then return the mean.

alejoe91: @samuelgarcia what do you think?

src/spikeinterface/curation/tests/common.py
```diff
@@ -3,13 +3,26 @@
 import pytest
 
 from spikeinterface.core import generate_ground_truth_recording, create_sorting_analyzer
+from spikeinterface.core.generate import inject_some_split_units
 
 job_kwargs = dict(n_jobs=-1)
 
+extensions = [
+    "random_spikes",
```
chrishalcrow, line 12:
```suggestion
    "noise_levels",
    "random_spikes",
```

should fix tests

src/spikeinterface/curation/auto_merge.py
```diff
@@ -1054,28 +1060,34 @@ def compute_cross_contaminations(analyzer, pair_mask, cc_thresh, refractory_peri
...
 
-        unit_id1 = unit_ids[unit_ind1]
-        spike_train1 = np.array(sorting.get_unit_spike_train(unit_id1))
+    for segment_index in range(num_segments):
```
alejoe91: instead of doing this, `concatenate_sortings`

PR #4202, opened 2025-11-06
Add Kilosort output to SortingAnalyzer helper function

chrishalcrow:
This PR adds a converter that takes kilosort output and makes an analyzer from it. Ideal for using the gui on your sorter output.

Unsure how to test this - any good ideas?

To do:
- [x] Figure out tests
- [x] Decide on default `compute_extras`
- [x] Test sparsity when some channels missing
- [x] Check no extra metadata leaks due to recording=generated
- [x] ~Implement Kilosort version guesser~

### DIFF
--- a/.github/scripts/test_kilosort4_ci.py
+++ b/.github/scripts/test_kilosort4_ci.py
@@ -117,6 +117,13 @@ if parse(kilosort.__version__) >= parse("4.0.37"):
     PARAMS_TO_TEST_DICT.update({"max_cluster_subset": 20})
     PARAMETERS_NOT_AFFECTING_RESULTS.append("max_cluster_subset")
 
+if parse(kilosort.__version__) >= parse("4.1.2"):
+    PARAMS_TO_TEST_DICT.update({"batch_downsampling": 2})
+    PARAMETERS_NOT_AFFECTING_RESULTS.append("batch_downsampling")
+
+    PARAMS_TO_TEST_DICT.update({"cluster_init_seed": 2})
+    PARAMETERS_NOT_AFFECTING_RESULTS.append("cluster_init_seed")
+
 
 PARAMS_TO_TEST = list(PARAMS_TO_TEST_DICT.keys())
 
@@ -329,6 +336,8 @@ class TestKilosort4Long:
             "scale",
             "file_object",
         ]
+        if parse(kilosort.__version__) >= parse("4.1.2"):
+            expected_arguments += ["batch_downsampling"]
 
         self._check_arguments(BinaryFiltered, expected_arguments)
 
@@ -352,6 +361,12 @@ class TestKilosort4Long:
         """
         recording, paths = recording_and_paths
         param_key = parameter
+
+        # Non-default batch_downsampling fails for short recordings, as there aren't
+        # enough batches. Since we test on a 5s recording, we skip it.
+        if param_key == "batch_downsampling":
+            return
+
         param_value = PARAMS_TO_TEST_DICT[param_key]
 
         # Setup parameters for KS4 and run it natively
--- a/.github/workflows/all-tests.yml
+++ b/.github/workflows/all-tests.yml
@@ -33,6 +33,17 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Free space
+        if: runner.os == 'Linux'
+        run: |
+          sudo rm -rf /usr/local/lib/android
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /opt/ghc
+          sudo rm -rf /usr/local/graalvm
+          sudo rm -rf /usr/local/share/powershell
+          sudo rm -rf /usr/local/share/chromium
+          df -h
+
       - name: Get changed files
         id: changed-files
         uses: tj-actions/changed-files@v46.0.1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
 -   repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 25.9.0
+    rev: 25.12.0
     hooks:
     -   id: black
         files: ^src/
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -397,7 +397,7 @@ spikeinterface.generation
 Core
 ~~~~
 
-.. currentmodule:: spikeinterface.generation
+.. automodule:: spikeinterface.generation
 
     .. autofunction:: generate_recording
     .. autofunction:: generate_sorting
@@ -418,7 +418,8 @@ Core
 Drift
 ~~~~~
 
-.. currentmodule:: spikeinterface.generation
+.. automodule:: spikeinterface.generation
+    :no-index:
 
     .. autofunction:: generate_drifting_recording
     .. autofunction:: generate_displacement_vector
@@ -432,7 +433,8 @@ Drift
 Hybrid
 ~~~~~~
 
-.. currentmodule:: spikeinterface.generation
+.. automodule:: spikeinterface.generation
+    :no-index:
 
     .. autofunction:: generate_hybrid_recording
     .. autofunction:: estimate_templates_from_recording
@@ -448,7 +450,8 @@ Hybrid
 Noise
 ~~~~~
 
-.. currentmodule:: spikeinterface.generation
+.. automodule:: spikeinterface.generation
+    :no-index:
 
     .. autofunction:: generate_noise
 
--- a/doc/get_started/installation.rst
+++ b/doc/get_started/installation.rst
@@ -65,6 +65,32 @@ as :code:`spikeinterface` strongly relies on these packages to interface with va
 It is also sometimes useful to have local copies of :code:`neo` and :code:`probeinterface` to make changes to the code. To achieve this, repeat the first set of commands,
 replacing :code:`https://github.com/SpikeInterface/spikeinterface.git` with the appropriate repository in the first code block of this section.
 
+Testing a PR
+------------
+
+If you have submitted an issue and another contributor has attempted to fix your issue by creating a pull request (PR), you
+might want to test the PR. To do this, first install spikeinterface "from source" in editable mode by following the
+instructions in the above Section.
+
+Now enter the :code:`spikeinterface` folder in your terminal
+
+.. code-block:: bash
+
+   cd path/to/spikeinterface
+
+
+There are a few ways to checkout a PR. The simplest is to first install the GitHub command line interface (https://cli.github.com/).
+After installing this, you can run :code:`gh pr checkout PR_NUMBER`. Every pull request has an ID, which can be found on the
+github Pull Request page (https://github.com/SpikeInterface/spikeinterface/pulls). So if you're interested in PR 4200 you should run:
+
+.. code-block:: bash
+
+   gh pr checkout 4200
+
+
+If successful, the returned message should end "Switched to branch 'blah blah blah'". You're now using :code:`spikeinterface`
+including the changes made in PR 4200.
+
 For beginners
 -------------
 
--- a/doc/how_to/analyze_neuropixels.rst
+++ b/doc/how_to/analyze_neuropixels.rst
@@ -1138,3 +1138,12 @@ And push the results to sortingview webased viewer
 .. code:: python
 
    si.plot_sorting_summary(analyzer_clean, backend='sortingview')
+
+
+
+Additional resources
+--------------------
+
+For additional resources about Neuropixels, you can take a look to the `awesome_neuropixel`_ github repo.
+
+.. _awesome_neuropixel: https://github.com/Julie-Fabre/awesome_neuropixels
--- a/doc/how_to/build_pipeline_with_dicts.rst
+++ b/doc/how_to/build_pipeline_with_dicts.rst
@@ -42,7 +42,7 @@ an example:
 
     # Usually, you would read in your raw recording
     rec, _ = si.generate_ground_truth_recording(num_channels=4, durations=[60], seed=0)
-    preprocessed_rec = si.apply_pipeline(rec, my_protocol['preprocessing'])
+    preprocessed_rec = si.apply_preprocessing_pipeline(rec, my_protocol['preprocessing'])
     sorting = si.run_sorter(recording=preprocessed_rec, **my_protocol['sorting'])
     analyzer = si.create_sorting_analyzer(recording=preprocessed_rec, sorting=sorting)
     analyzer.compute(my_protocol['postprocessing'])
--- a/doc/modules/curation.rst
+++ b/doc/modules/curation.rst
@@ -310,18 +310,17 @@ a ``BaseSorting`` or ``SortingAnalyzer`` object using the :py:func:`~spikeinterf
 
 .. code-block:: python
 
-    from spikeinterface.curation import apply_curation
+    from spikeinterface.curation import load_curation, apply_curation
 
     # load the curation JSON file
-    curation_json = "path/to/curation.json"
-    with open(curation_json, 'r') as f:
-        curation_dict = json.load(f)
+    curation_filepath = "path/to/curation.json"
+    curation = load_curation(curation_filepath)
 
     # apply the curation to the sorting output
-    clean_sorting = apply_curation(sorting, curation_dict=curation_dict)
+    clean_sorting = apply_curation(sorting, curation_dict_or_model=curation)
 
     # apply the curation to the sorting analyzer
-    clean_sorting_analyzer = apply_curation(sorting_analyzer, curation_dict=curation_dict)
+    clean_sorting_analyzer = apply_curation(sorting_analyzer, curation_dict_or_model=curation)
 
 
 Using the ``SpikeInterface GUI``
--- a/doc/releases/0.103.2.rst
+++ b/doc/releases/0.103.2.rst
@@ -1,0 +1,73 @@
+.. _release0.103.2:
+
+SpikeInterface 0.103.2 release notes
+------------------------------------
+
+December 3rd 2025
+
+Minor release with bug fixes
+
+core:
+
+* Fix probable sharedmem (#4239)
+* Fix probable shared mem no clean corretcly (#4234)
+* Implementing sparsity for ComputeTemplates (#4212)
+* Warn about non-mergeable units instead of raising an error (#4219)
+
+extractors:
+
+* Support for using event names as IDs in PL2 files (#4049)
+
+preprocessing:
+
+* Add channel_filter parameter for bad channel detection (#4228)
+
+sorters:
+
+* Warn about non-binary compatible recordings using KS4 (#4236)
+* Deal with batch_downsampling in kilosort4.1.2 (#4206)
+
+postprocessing:
+
+* Speedup template_similarity (#4211)
+
+qualitymetrics:
+
+
+curation:
+
+* Add `load_curation` function (#4220)
+* Add similarity_method param in automerge and fix multi-segment cross-contamination (#4201)
+
+widgets:
+
+* Protect unit summary and unit locations plots against NaNs (#4128)
+
+sortingcomponents:
+
+* patch for sc2: propagate real ms_before/after to template extension of final analyzer (#4215)
+* Speedup template_similarity (#4211)
+* Improvement of TDC2, SC2 and ISOSPLIT (#4182)
+
+documentation:
+
+* Add additional resources section to Analyze Neuropixels Datasets tuto (#4232)
+* Fix `apply_preprocessing_pipeline` name in docs (#4208)
+* Tiny doc fixes (#4203)
+* Add testing a PR section to docs (#4199)
+
+continuous integration:
+
+* Fix no disk space left ci (#4229)
+
+Contributors:
+
+* @MGAMZ
+* @OookTheLibrarian
+* @ShijiMi-Soup
+* @alejoe91
+* @chrishalcrow
+* @h-mayorquin
+* @samuelgarcia
+* @tayheau
+* @yger
--- a/doc/whatisnew.rst
+++ b/doc/whatisnew.rst
@@ -8,6 +8,7 @@ Release notes
 .. toctree::
   :maxdepth: 1
 
+  releases/0.103.2.rst
   releases/0.103.1.rst
   releases/0.103.0.rst
   releases/0.102.3.rst
@@ -49,6 +50,12 @@ Release notes
   releases/0.10.0.rst
   releases/0.9.9.rst
   releases/0.9.1.rst
+
+
+Version 0.103.2
+===============
+
+* Minor release with bug fixes
 
 
 Version 0.103.1
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,13 +1,13 @@
 [project]
 name = "spikeinterface"
-version = "0.103.2"
+version = "0.103.3"
 authors = [
   { name="Alessio Buccino", email="alessiop.buccino@gmail.com" },
   { name="Samuel Garcia", email="sam.garcia.die@gmail.com" },
 ]
 description = "Python toolkit for analysis, visualization, and comparison of spike sorting output"
 readme = "README.md"
-requires-python = ">=3.9,<3.14"
+requires-python = ">=3.9"
 classifiers = [
     "Programming Language :: Python :: 3 :: Only",
     "License :: OSI Approved :: MIT License",
--- a/src/spikeinterface/benchmark/benchmark_base.py
+++ b/src/spikeinterface/benchmark/benchmark_base.py
@@ -431,7 +431,7 @@ class BenchmarkStudy:
             if verbose:
                 print("### Compute result", key, "###")
             benchmark = self.benchmarks[key]
-            assert benchmark is not None
+            assert benchmark is not None, f"Benchmkark for key {key} has not been run yet!"
             benchmark.compute_result(**result_params)
             benchmark.save_result(self.folder / "results" / self.key_to_str(key))
 
--- a/src/spikeinterface/core/__init__.py
+++ b/src/spikeinterface/core/__init__.py
@@ -111,6 +111,7 @@ from .recording_tools import (
 )
 from .sorting_tools import (
     spike_vector_to_spike_trains,
+    spike_vector_to_indices,
     random_spikes_selection,
     apply_merges_to_sorting,
     apply_splits_to_sorting,
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -11,12 +11,14 @@ It also implements:
 
 import warnings
 import numpy as np
+from collections import namedtuple
 
-from .sortinganalyzer import AnalyzerExtension, register_result_extension
+from .sortinganalyzer import SortingAnalyzer, AnalyzerExtension, register_result_extension
 from .waveform_tools import extract_waveforms_to_single_buffer, estimate_templates_with_accumulator
 from .recording_tools import get_noise_levels
 from .template import Templates
 from .sorting_tools import random_spikes_selection
+from .job_tools import fix_job_kwargs, split_job_kwargs
 
 
 class ComputeRandomSpikes(AnalyzerExtension):
@@ -437,6 +439,7 @@ class ComputeTemplates(AnalyzerExtension):
             return_in_uV = self.sorting_analyzer.return_in_uV
 
             return_std = "std" in self.params["operators"]
+            sparsity_mask = None if self.sparsity is None else self.sparsity.mask
             output = estimate_templates_with_accumulator(
                 recording,
                 some_spikes,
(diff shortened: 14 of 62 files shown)


src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,177 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
 
 read_phy = define_function_from_class(source_class=PhySortingExtractor, name="read_phy")
 read_kilosort = define_function_from_class(source_class=KiloSortSortingExtractor, name="read_kilosort")
+
+
+def kilosort_output_to_analyzer(folder_path, compute_extras=False, unwhiten=True) -> SortingAnalyzer:
+    """
+    Load kilosort output into a SortingAnalyzer.
+    Output from kilosort version 4.1 and above are supported.
```
JoeZiminski: Is there any way to check the version from the output? If not we should ask them to add on the KS repo as this would be a useful general addition. But maybe we could check that the `kilosortX.log` is not `< 4`? (IIRC that the logs are formatted in this way)

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,177 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+    -------
+    sorting_analyzer : SortingAnalyzer
+        A SortingAnalyzer object.
+    """
+
+    phy_path = Path(folder_path)
+
+    sorting = read_phy(phy_path)
+    sampling_frequency = sorting.sampling_frequency
+    duration = sorting._sorting_segments[0]._all_spikes[-1] / sampling_frequency + 1
```
JoeZiminski, line 346: is the `+1` for padding?

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,177 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+    if (phy_path / "probe.prb").is_file():
+        probegroup = read_prb(phy_path / "probe.prb")
+        probe = probegroup.probes[0]
+    elif (phy_path / "channel_positions.npy").is_file():
+        probe = Probe(si_units="um")
+        channel_positions = np.load(phy_path / "channel_positions.npy")
+        probe.set_contacts(channel_positions)
+        probe.set_device_channel_indices(range(probe.get_contact_count()))
+    else:
+        AssertionError("Cannot read probe layout from folder {phy_path}.")
```
JoeZiminski:
```suggestion
        AssertionError(f"Cannot read probe layout from folder {phy_path}.")
```

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,177 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+    recording, _ = generate_ground_truth_recording(
+        probe=probe, sampling_frequency=sampling_frequency, durations=[duration]
+    )
+
+    sparsity = _make_sparsity_from_templates(sorting, recording, phy_path)
+
+    sorting_analyzer = create_sorting_analyzer(sorting, recording, sparse=True, sparsity=sparsity)
+
+    # first compute random spikes. These do nothing, but are needed for si-gui to run
+    sorting_analyzer.compute("random_spikes")
```
JoeZiminski, line 375: Just to confirm (I had a quick look at ComputeRandomSpikes), this will compute indices of valid spikes to subsample and store them on the analyzer. It's just that no waveforms are handled here, so it has no effect. If someone adds a recording and computes waveforms, it will correctly subsample I guess from the spike times?

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,177 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+    """
+    Load kilosort output into a SortingAnalyzer.
+    Output from kilosort version 4.1 and above are supported.
+
+    Parameters
+    ----------
+    folder_path : str or Path
+        Path to the output Phy folder (containing the params.py).
+    compute_extras : bool, default: False
+        Compute the extra extensions: unit_locations, correlograms, template_similarity, isi_histograms, template_metrics, quality_metrics.
```
JoeZiminski: Could the docstring also include what you get by default? (I guess spike times, templates, amplitudes, locations).

JoeZiminski: I guess the only thing missing is metrics computed from waveforms? I guess following use of this function, the user can run "compute_waveforms" after attaching a recording object?

I always thought it would be cool to load the kilosort waveforms directly in SI, but this doesn't [seem to be possible for older versions](https://github.com/MouseLand/Kilosort/issues/818). For kilosort4 (outside the scope of this PR, just a discussion) do you think we could ask Jacob to write a public kilosort function that returns the waveforms in 3d array? I guess on the kilosort4 side they are still stored as a first few components in PCA space (?). It would be cool to have an option to load the original KS4 waveforms too in this function. I always wondered exactly how they compare to those generated in spikeinterface when drift-correction is run on the kilosort side (and spikeinterface waveforms are cut-out from the preprocessed recording pre-drift correction).

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,177 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+
+    phy_path = Path(folder_path)
+
+    sorting = read_phy(phy_path)
+    sampling_frequency = sorting.sampling_frequency
+    duration = sorting._sorting_segments[0]._all_spikes[-1] / sampling_frequency + 1
+
+    if (phy_path / "probe.prb").is_file():
+        probegroup = read_prb(phy_path / "probe.prb")
+        probe = probegroup.probes[0]
```
JoeZiminski, line 352: I guess there can only ever be one probe saved here? is it worth checking there is one only probe in probegroup?

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,177 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+
+
+def _make_amplitudes(sa, phy_path: Path):
+
+    amplitudes_extension = ComputeSpikeAmplitudes(sa)
+
+    amps_np = np.load(phy_path / "amplitudes.npy")
+
+    amplitudes_extension.data = {}
+    amplitudes_extension.data["amplitudes"] = amps_np
```
JoeZiminski: could this be `amplitudes_extension.data = {"amplitudes": amps_np}`? doesn't really matter either way of course

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,177 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+
+    params = {"peak_sign": "neg"}
```
JoeZiminski: same here one liner and then all three assignments are in the same form

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,177 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+                "template_metrics": {"include_multi_channel_metrics": True},
+                "quality_metrics": {},
+            }
+        )
+
+    sorting_analyzer._recording = None
+    return sorting_analyzer
+
+
+def _make_amplitudes(sa, phy_path: Path):
```
JoeZiminski: Could these functions include breif docstring? maybe `sa` can be expanded to `sorting_analyzer` as above

JoeZiminski:
```suggestion
def _make_amplitudes(sa, phy_path):
```

I ❤️ typing, but  I guess everything should be typed or nothing should be typed

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,177 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+
+    locations_extension = ComputeSpikeLocations(sa)
+
+    locs_np = np.load(phy_path / "spike_positions.npy")
+
+    num_dims = len(locs_np[0])
+    column_names = ["x", "y", "z"][:num_dims]
+    dtype = [(name, locs_np.dtype) for name in column_names]
+
+    structured_array = np.array(np.zeros(len(locs_np)), dtype=dtype)
```
JoeZiminski:
```suggestion
    structured_array = np.zeros(len(locs_np), dtype=dtype)
```

I think this is equivalent?

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,177 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+    for a, column_name in enumerate(column_names):
```
JoeZiminski: could `a` be `idx`?

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,177 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+        structured_array[column_name] = locs_np[:, a]
+
+    locations_extension.data = {}
+    locations_extension.data["spike_locations"] = structured_array
+
+    params = {}
+    locations_extension.params = params
+
+    locations_extension.run_info = {"run_completed": True}
```
JoeZiminski, line 415: same on dict assignment form, but ofc this is completely arbitrary and can be ignored

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,177 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+    mask = np.sum(np.abs(templates), axis=1) != 0
+    return ChannelSparsity(mask, unit_ids=unit_ids, channel_ids=channel_ids)
+
+
+def _make_templates(sa, phy_path, mask, sampling_frequency, unwhiten=True):
+
+    template_extension = ComputeTemplates(sa)
+
+    whitened_templates = np.load(phy_path / "templates.npy")
+    wh_inv = np.load(phy_path / "whitening_mat_inv.npy")
```
JoeZiminski: I think this is only reliable on KS4 (KS2.5-3 the [inverse whitening matrix is scaled identity](https://github.com/MouseLand/Kilosort/issues/504)). It is clear from the docstring that this is only for KS4 but I'm sure people will try on other versions.  Checking the log file is probably the easiest way to catch this, but another possible safety net could be to check that the `wh_inv` is not all zeros off diagonal.

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,177 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+    new_templates = _compute_unwhitened_templates(whitened_templates, wh_inv, mask) if unwhiten else whitened_templates
+
+    template_extension.data = {"average": new_templates}
+
+    ops_path = phy_path / "ops.npy"
+    if ops_path.is_file():
```
JoeZiminski, line 448: Maybe a quick comment to explain what this is doing. I always wondered what `nt` stood for in KS, so `nt` is the total number of samples for the waveform and `nt0min` is the number of samples from the start to the peak? (from which samples after can be computed)?

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,177 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+
+
+def _compute_unwhitened_templates(whitened_templates, wh_inv, mask):
+
+    template_shape = np.shape(whitened_templates)
+    new_templates = np.zeros(template_shape)
+
+    sparsity_channel_ids = [np.arange(template_shape[-1])[unit_sparsity] for unit_sparsity in mask]
+
+    for a, unit_sparsity in enumerate(sparsity_channel_ids):
```
JoeZiminski: I'm a bit confused here, could the variables `a`, `b`, `c` be renamed to describe the dim they are iterative over? If I read correctly the unwhitening is applied channel-by-channel, is there scope to apply it in a single matrix multiplication per template? I guess the channel masking makes this tricky, maybe the masked channel are not contagious (e.g. a dead channel of all zeros in the middle of the template?)

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,177 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+    wh_inv = np.load(phy_path / "whitening_mat_inv.npy")
+    new_templates = _compute_unwhitened_templates(whitened_templates, wh_inv, mask) if unwhiten else whitened_templates
+
+    template_extension.data = {"average": new_templates}
+
+    ops_path = phy_path / "ops.npy"
+    if ops_path.is_file():
+        ops = np.load(ops_path, allow_pickle=True)
+
+        samples_before = ops.item(0).get("nt0min")
```
JoeZiminski: Here I would index the dictionary directly `(0)["nt0min"]` if for some reason `nt0min` is not in the `dict` it will show an access error. Otherwise `samples_before` will be `None` which will throw a more confusing error below.  Even if it should always be in the `dict`, I don't think there is a benefit of using `get` unless explicitly checking if the result is `None`

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: Yes, that's right. They make no impact on any of the extensions we create from the output. And now they are there if someone computes e.g. waveforms in the future.

 I'm being a bit lazy here - I only want them so that si-gui won't complain. But really, we should make changes to si-gui so that it can open without having random_spikes computed...

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: Yeah, I wanted to discuss this function argument. @samuelgarcia suggested that you could pass a list of what you'd like to compute instead of having this bool option. Alternatively, we could not compute any extras, and put that responsibility on the user. What'dya think?

chrishalcrow: Almost all the metrics depend on waveform features like amplitudes and locations, rather than the waveforms themselves. I don't think we miss out on anything (except the pca metrics, because the pca's are impossible to load, I think...)

Does kilosort save waveforms? My output doesn't seem to have any - just their features such as locations, amps and pcas.

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: Probably a good idea to check there's only one, and error if more. Cause if there is more, I don't trust the rest of the code!

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: I was getting errors when there wasn't a `+1`. I have a feeling it's this bug where kilosort sometimes outputs a sorting that is slightly longer than the input recording. Usually we deal with this using `remove_excess_spikes` but we don't have access to the recording here, so it's tricky...

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: Ok, so we could try to check the version number, then if it's less than 4, we only allow whitened templates?

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: On another project, I got in a mess of circular imports and eventually decided to only do typing on standard library functions. But yeah, I understand your point!

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: Have asked on KiloSort

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: I tried to make it clearer using better variable names - ok?

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: Hmmm good question actually. I'll have to try out on some data with missing channels.

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: If you've run directly from kilosort, you'll have the version in `kilosort4.log`. So we could check there, and if it's not there, we have a guess...

JoeZiminski: Great that sounds good, if this tool is only for `kilosort4` then just checking for the existing of that log file should do (unless it's extended to other versions)

src/spikeinterface/extractors/phykilosortextractors.py
JoeZiminski: Yes that is tricky, I don't know enough about the practicalities of this, I guess it basically has no affect? the duration just needs to be longer than the max spike time? Maybe just a quick comment here to explain the reason for the `+1`?

src/spikeinterface/extractors/phykilosortextractors.py
JoeZiminski: I was thinking about that when reveiwing, at the moment its okay for a quick default but immediately raises the question of what arguments are used / how to pass arguments. A list of dict of lists / arguments could work but I would proably vote for removing this and having this step dedicated only for to the loading /conversion, for better modularity

JoeZiminski: That's cool, okay great and makes sense. Would it be possible for the user to attach a recording and `compute_waveforms` afterwards if they wanted PCs?

The old versions did used to under `pc_features.npy` (see [here](https://phy.readthedocs.io/en/latest/sorting_user_guide/)), it needs to be indexed with `pc_feature_ind.npy`, I forget the exact process but it is possible to reconstruct them from the compressed spikes (stored as scores on the first few PCs). However this was the issue with recosntructing across versions as the procedure is different across KS versions. I'm not 100% sure about KS4, I'm actually running it on some data tomorrow so will get back!

src/spikeinterface/extractors/phykilosortextractors.py
JoeZiminski: I think this would work, I am not 100% confident without checking across versions, I will try and sorted and send some older datasets tomorrow! I think indeed `unwhitened` would not be possible for ks2.5 and ks3

src/spikeinterface/extractors/phykilosortextractors.py
JoeZiminski: great cheers that reads well to me

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: Hello, the log files only appeared at `v4.0.33`. Thinking of other ways to check...

chrishalcrow: Do you know of any files which KS2/2.5 defo _don't_ have in their output, that KS4 does?

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: Yeah, I want to remove it and add to docs. When a user reads this, it'll be a good moment to learn about how the analyzer extensions work too :)

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: Hello, I've added a `_guess_kilosort_version` function, to isolate this logic. Let's compare some outputs and see if we can make it reasonable.

chrishalcrow:
> *[EDIT] I think my only worry on that was if adding the mock `recording` object stores some metadata on the analyzer under the hood that is not erased when `._recording` is set to `None`. I took a quick look at the code and it doesn't seem like it does now. But maybe in future this could cause a problem, but I think there is no workaround 🤔 maybe an assert somewhere they key metadata fields are empty?

It does save some metadata related to the probe, and some basic info like sampling frequency in `rec_attributes`, but this is required for the analyzer to work at all. So: we do want to keep that, and we don't know what to protect against in the future. So I'm not sure how to protect against future problems. I'd say it's probably fine.

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: Hello, tested it on some data and all works fine. Lucky!

chrishalcrow:
Hey @JoeZiminski . Just realised that the amplitudes saved by Kilosort are the "Per-spike amplitudes, computed as the L2 norm of the PC features for each spike." https://kilosort.readthedocs.io/en/latest/export_files.html

So I don't think it's fair to call them `spike_amplitudes` as we do in spikeinterface? Thoughts?

alejoe91:
> Hey @JoeZiminski . Just realised that the amplitudes saved by Kilosort are the "Per-spike amplitudes, computed as the L2 norm of the PC features for each spike." https://kilosort.readthedocs.io/en/latest/export_files.html
> 
> So I don't think it's fair to call them `spike_amplitudes` as we do in spikeinterface? Thoughts?

Agree!

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,195 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+    sparsity_channel_ids = [np.arange(template_shape[-1])[unit_sparsity] for unit_sparsity in mask]
+
+    for unit_index, channel_indices in enumerate(sparsity_channel_ids):
+        for channel_index_1 in channel_indices:
+            for channel_index_2 in channel_indices:
+                # templates have dimension unit_index x sample_index x channel_index
+                # to undo whitening, we need do matrix multiplication on the channel_index
+                new_templates[unit_index, :, channel_index_1] += (
+                    wh_inv[channel_index_1, channel_index_2] * whitened_templates[unit_index, :, channel_index_2]
+                )
```
alejoe91: I think this should work:
```
new_templates = whitened_templates @ wh_inv
```
no?

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,195 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
 
 read_phy = define_function_from_class(source_class=PhySortingExtractor, name="read_phy")
 read_kilosort = define_function_from_class(source_class=KiloSortSortingExtractor, name="read_kilosort")
+
+
+def kilosort_output_to_analyzer(folder_path, compute_extras=False, unwhiten=True) -> SortingAnalyzer:
```
alejoe91: Not sure I like the name...maybe `read_kilosort_as_analyzer` or `load_analyzer_from_kilosort`? 

Should `kilosort` be `phy` instead? Also, we should add it to the extractors `__init__.py`

chrishalcrow: I think the file naming conventions are particular to kilosort, which is why I've not gone for "phy".  I can't really figure it out though - can't find a proper specification for phy input/output.

I like `load_analyzer_from_kilosort`!
Yes, thanks - will add.

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: This is a 3-tensor to a 3-tensor, so I wanna be clear which indices are being summed over and which are being left alone. I'd be ok with "@" if it was a standard 2-tensor (matrix) multiplication.

alejoe91: I see, makes sense

src/spikeinterface/extractors/phykilosortextractors.py
samuelgarcia: I prefer read_kilosort_as_analyzer rather than load_analyzer_from_kilosort.
But do not care sor much.

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,195 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+        channel_positions = np.load(phy_path / "channel_positions.npy")
+        probe.set_contacts(channel_positions)
+        probe.set_device_channel_indices(range(probe.get_contact_count()))
+    else:
+        AssertionError(f"Cannot read probe layout from folder {phy_path}.")
+
+    # to make the initial analyzer, we'll use a fake recording and set it to None later
+    recording, _ = generate_ground_truth_recording(
+        probe=probe, sampling_frequency=sampling_frequency, durations=[duration]
+    )
```
samuelgarcia, line 368: This is strange but I guess this is the easiest way. Maybe you should put num_units=1 to avoid warning message in case of there is not solution for unit_location and maybe we should seed also...

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,195 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+    Returns
+    -------
+    version_number : tuple
+        Version number in the form (major, minor, patch)
+    """
+
+    kilosort_log_file = Path(kilosort_path / "kilosort4.log")
+
+    if kilosort_log_file.is_file():
+        return (4, 0, 33)
```
samuelgarcia: I guess you want to parse the file here no ?
Also we check that the kilosort folder was lauch by spikeinterface and use the spikeinterface log to gete the version no ?

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,195 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+def _make_amplitudes(sorting_analyzer, kilosort_output_path):
+    """Constructs a `spike_amplitudes` extension from the amplitudes numpy array
+    in `kilosort_output_path`, and attaches the extension to the `sorting_analyzer`."""
+
+    amplitudes_extension = ComputeSpikeAmplitudes(sorting_analyzer)
+
+    amps_np = np.load(kilosort_output_path / "amplitudes.npy")
+
+    amplitudes_extension.data = {"amplitudes": amps_np}
+    amplitudes_extension.params = {"peak_sign": "neg"}
```
samuelgarcia: I guess kilosort has "both"

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,195 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+def _make_sparsity_from_templates(sorting, recording, kilosort_output_path):
+    """Constructs the `ChannelSparsity` of from kilosort output, by seeing if the
+    templates output is zero or not on all channels."""
+
+    templates = np.load(kilosort_output_path / "templates.npy")
+
+    unit_ids = sorting.unit_ids
+    channel_ids = recording.channel_ids
+
+    # The raw templates have dense dimensions (num chan)x(num units)
```
samuelgarcia: I guess that the shape is 3 dim no ? something like (num_units, num_chans, num_samples) no ?
Could you put the real dim in comments so the next line is easy to guess.

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -302,3 +315,195 @@ def __init__(self, folder_path: Path | str, keep_good_only: bool = False, remove
...
+def _compute_unwhitened_templates(whitened_templates, wh_inv, mask):
+    """Constructs unwhitened templates from whitened_templates, by
+    applying an inverse whitening matrix."""
+
+    template_shape = np.shape(whitened_templates)
+    new_templates = np.zeros(template_shape)
+
+    sparsity_channel_ids = [np.arange(template_shape[-1])[unit_sparsity] for unit_sparsity in mask]
+
+    for unit_index, channel_indices in enumerate(sparsity_channel_ids):
```
samuelgarcia: Is there a way to use np.dot instead ?

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: Yes, you can attached a recording and compute anything you'd like!

There's a PCA issue: spikeinterface assumes the number of channels used to compute PCs is fixed, while kilosort allows for variable channels per spike. So will require a bit of a change in si to get this to work => not gonna happen in this PR!

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: No, dot it too magical (you can't specify which indicies to sum other). We could use einsum, but I've been told it's a bit tricky for non-math/physics people??

chrishalcrow: `new_templates = np.einsum('ij,klj->kli', wh_inv, whitened_templates)`

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: Yeah, the idea is that we use whatever we can to figure out the kilosort version.

However, it seems like everything works fine for all versions (i.e. the probes, templates and spike locations are saved in the same way for all versions). So in the end we don't need to change behavior depending on version.

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: ignore me

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: Seems like everything works fine for all versions (i.e. the probes, templates and spike locations are saved in the same way for all versions). So in the end we don't need to change behavior depending on version. We would have to if we got pcas working, but that's not gonna happen in this PR.

chrishalcrow:
I'm happy with this now! In the end, after testing a load of output from Joe and Sai, we don't need the version checker. And we skip spike locations if we need to.

samuelgarcia (approved):

PR #4211, opened 2025-11-12
Speedup template_similarity

yger:
While restructuring the components, we introduced a slowdown in template_similarity, by mistake. I don't think this causes bug, but rather a global problem with speed since we are computing distances for all templates, instead of only those that are intersecting.

### DIFF
--- a/src/spikeinterface/sortingcomponents/clustering/tests/test_isosplit_isocut.py
+++ b/src/spikeinterface/sortingcomponents/clustering/tests/test_isosplit_isocut.py
@@ -151,7 +151,6 @@ def test_isosplit():
 
     data = data.astype("float32")
     labels = isosplit(data, isocut_threshold=2.0, n_init=40, seed=2205)
-    assert np.unique(labels).size == 3
 
     # the beauty is that it discovers the number of clusters automatically, at least for this this seed :)
     # assert np.unique(labels).size == 3


samuelgarcia:
> In addition, in this PR, I make use of the new sparsity mask in the extension, while computing Templates

I think we should not do this in the same PR. The internal sparse representaion of template will have impact at many places, like merging or plotting. Lets do this more carfully.

yger:
You are right regaring the sparsity masks. This has been moved in #4212

yger:
@samuelgarcia this one is needed for runtimes of both SC2, TDC2, and very large in vitro recordings

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -110,16 +110,16 @@ def _merge_extension_data(
...
-                for unit_ind2, unit_id2 in enumerate(all_new_unit_ids):
-                    if unit_id2 not in new_unit_ids:
-                        old_ind2 = self.sorting_analyzer.sorting.id_to_index(unit_id2)
-                        s = self.data["similarity"][old_ind1, old_ind2]
-                        similarity[unit_ind1, unit_ind2] = s
-                        similarity[unit_ind1, unit_ind2] = s
+        for old_ind1, unit_ind1 in zip(old_units_inds, sub_units_inds):
+            s = self.data["similarity"][old_ind1, old_units_inds]
+            similarity[unit_ind1, sub_units_inds] = s
+            similarity[sub_units_inds, unit_ind1] = s
```
samuelgarcia, line 122: Cool. Smater

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -319,9 +319,14 @@ def _compute_similarity_matrix_numba(
...
                     )  # shape (other_num_templates, num_channels)
                 elif support == "union":
+                    connected_mask = np.logical_and(sparsity_mask[i, :], other_sparsity_mask)
+                    not_connected_mask = np.sum(connected_mask, axis=1) == 0
                     local_mask = np.logical_or(
                         sparsity_mask[i, :], other_sparsity_mask
                     )  # shape (other_num_templates, num_channels)
+                    for local_i in np.flatnonzero(not_connected_mask):
+                        local_mask[local_i] = False
+
```
samuelgarcia, line 329: Of course!

samuelgarcia:
Merci pour le bug!

samuelgarcia:
I merge this, this is important.

PR #4194, opened 2025-11-03
Add sparsity_mask option in estimate_templates_with_accumulator()

samuelgarcia:
Add sparsity_mask option in estimate_templates_with_accumulator()

PR #4212, opened 2025-11-13
Implementing sparsity for ComputeTemplates

yger:
This implements the use of sparsity to compute templates in the Analyzer. Currently, analzers can handle sparsity mask, and while they are used to compute waveforms, they are not used to estimate templates. This can lead to large memory usage with numerous threads. This PR makes sure the analyzer, if sparse, uses the sparsity masks while estimating templates ONLY in the run() method. For compatibility with downstream extensions, the internal storage is kept dense*

See #4194

### DIFF
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -461,13 +461,7 @@ class ComputeTemplates(AnalyzerExtension):
             if self.sparsity is not None:
                 # make average and std dense again
                 for k, arr in data.items():
-                    dense_arr = np.zeros(
-                        (arr.shape[0], arr.shape[1], self.sorting_analyzer.get_num_channels()),
-                        dtype=arr.dtype,
-                    )
-                    for unit_index, unit_id in enumerate(self.sorting_analyzer.unit_ids):
-                        chan_inds = self.sparsity.unit_id_to_channel_indices[unit_id]
-                        dense_arr[unit_index][:, chan_inds] = arr[unit_index, :, : chan_inds.size]
+                    dense_arr = self.sparsity.densify_templates(arr)
                     data[k] = dense_arr
             self.data.update(data)
 
--- a/src/spikeinterface/core/sparsity.py
+++ b/src/spikeinterface/core/sparsity.py
@@ -247,6 +247,18 @@ class ChannelSparsity:
 
         return sparse_templates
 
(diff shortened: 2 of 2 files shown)


yger:
Kudo for the code rewriting... All tests are failing :-)

samuelgarcia:
@alejoe91 @chrishalcrow : this will help a lot ith very high channel count (in vitro mea).
Have a look.

samuelgarcia:
@alejoe91 we need to squash

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -445,17 +446,30 @@ def _run(self, verbose=False, **job_kwargs):
...
+            if self.sparsity is not None:
+                # make average and std dense again
+                for k, arr in data.items():
+                    dense_arr = np.zeros(
+                        (arr.shape[0], arr.shape[1], self.sorting_analyzer.get_num_channels()),
+                        dtype=arr.dtype,
+                    )
+                    for unit_index, unit_id in enumerate(self.sorting_analyzer.unit_ids):
+                        chan_inds = self.sparsity.unit_id_to_channel_indices[unit_id]
+                        dense_arr[unit_index][:, chan_inds] = arr[unit_index, :, : chan_inds.size]
```
alejoe91: We have the exact same logic in [`ChannelSparsity.densify_waveforms`](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/sparsity.py#L183): 

My proposal is to add a `densify_templates` function to the `ChannelSparsity` class (it already has a `sparsify_templates`):
```
def densify_templates(self, templates_array: np.ndarray) -> np.ndarray:
    assert templates_array.shape[0] == self.num_units

    densified_shape = (self.num_units, templates_array.shape[1], self.num_channels)
    dense_templates = np.zeros(shape=densified_shape, dtype=templates_array.dtype)
    for unit_index, unit_id in enumerate(self.unit_ids):
        sparse_template = templates_array[unit_index, ...]
        dense_template = self.densify_waveforms(waveforms=sparse_template[np.newaxis, :, :], unit_id=unit_id)
        dense_templates[unit_index, :, :] = dense_template

    return dense_templates
```
Then the logic in the ComputeTemplates extension could simply become:
```
if self.sparsity is not None:
    # make average and std dense again
    for k, arr in data.items():
        dense_arr = self.sparsity.densify_templates(arr)
        data[k] = dense_arr
```
What do you think?

chrishalcrow (approved):

PR #4215, opened 2025-11-14
patch for sc2: propagate real ms_before/after to template extension of final analyzer

yger:
Internally, SC2 creates an analyzer on the fly using precomputed templates and thus "hack" the way templates are computed within the analyzer. To be reused properly later by the GUI, one needs an "operator" field, now added


yger:
This is a minor one, @samuelgarcia

samuelgarcia (approved):

PR #4216, opened 2025-11-14
Implement `ChunkExecutor`

alejoe91:
This PR makes a `ChunkExecutor` which is agnostic to the recording and applies to any chunkable `BaseExtractor` (which defines `get_sample_size(), get_num_samples(), get_num_segments()`

This is because we are working on a new "spikeinterface" for calcium imaging, and this refactor makes it suuuuper straightforward to use the same parallelization machinery for `Imaging` objects (which inherit from the `BaseExtractor`)

## Update 29/11/25

Following up on trying to make SI tools as general as possible for its calcium imaging "clone" ([photon-mosaic](https://github.com/photon-mosaic/photon-mosaic/pull/1)) I made some more dramatic refactoring:

- implemented a `ChunkableMixin` abstract mixin that implements all method needed by the `ChunkExecutor`
  - several `BaseRecording` specific methods (e.g., `write_binary`/`write_memory` and more) are now in the `chunkable_tools` and applicable to any `ChunkableMixin` object
  - the `ChunkableMixin` has a general `get_data` function, so it can be used by many derived classes (e.g., `get_traces` for recordings, `get_videos/series` for imaging)
  - another useful function to centralize things is the `get_shape(segment_index`), which returns the buffer shape by segment
- refactored "segments": I that we didn't have any proper structure for segments: recording objects had `_recording_segments` and sorting objects `_sorting_segments`. Each class then implemented its own `add_*` function. Since segments are a `BaseExtractor` concept, I centralized the definition (`self.segments: List[BaseSegment] = []`), added a `self.segments` property, and a centralized `add_segment` function.

@samuelgarcia @chrishalcrow @yger @h-mayorquin @JoeZiminski @zm711 

Let me know what you think!!!


src/spikeinterface/core/job_tools.py
```diff
@@ -590,6 +541,127 @@ def run(self, recording_slices=None):
...
+        )
+
+    def run(self, recording_slices=None):
+        """
+        Runs the defined jobs.
+        """
+        return super().run(slices=recording_slices)
+
+    def get_chunk_memory(self):
+        return self.chunk_size * self.recording.get_dtype().itemsize * self.recording.get_num_channels()
```
alejoe91: For an `Imaging` object, this will be num pixels instead of num channels ;)

alejoe91: I think it would be even better if we added a `get_sample_size()` method

src/spikeinterface/core/baserecording.py
```diff
@@ -184,6 +184,20 @@ def add_recording_segment(self, recording_segment):
         self._recording_segments.append(recording_segment)
         recording_segment.set_parent_extractor(self)
 
+    def get_sample_size(self):
```
h-mayorquin:
```suggestion
    def get_sample_size_in_bytes(self):
```

I think that some people might mistake the method name for shape or number of samples if we don't specify it this way.

alejoe91: great point!

src/spikeinterface/core/recording_tools.py
```diff
@@ -465,7 +465,7 @@ def write_to_h5_dataset_format(
     return save_path
 
 
-def get_random_recording_slices(
+def get_random_slices(
```
samuelgarcia: This is too neutral no ? Lets find some thing like
get_random_slices_along_time

samuelgarcia:
Lets discuss this in detail. This is lots some semantic changes.
I do not like too much the "extractor" evrywhere.

src/spikeinterface/core/recording_tools.py
alejoe91: get_random_time_slices?

alejoe91:
> Lets discuss this in detail. This is lots some semantic changes. I do not like too much the "extractor" evrywhere.

What about a `Chunkable` mixin that implements:
- `get_num_segments()`
- `get_sample_size_in_bytes()`
- `get_num_samples()`
?
Then `extractor` could become `chunkable`

src/spikeinterface/core/baserecording.py
alejoe91: done

src/spikeinterface/core/recording_tools.py
alejoe91: now renamed `get_random_sample_slices`

src/spikeinterface/core/basesorting.py
```diff
@@ -76,16 +75,9 @@ def get_unit_ids(self) -> list:
     def get_num_units(self) -> int:
         return len(self.get_unit_ids())
 
-    def add_sorting_segment(self, sorting_segment):
-        self._sorting_segments.append(sorting_segment)
```
alejoe91, line 80: keep it and use `super().add_segment()`

alejoe91:
- Factor out `segment` into a separate PR
- Keep this PR only for chunkable / refactor of chunk executor and node pipeline

alejoe91:
Closing in favor of https://github.com/SpikeInterface/spikeinterface/pull/4462 and https://github.com/alejoe91/spikeinterface/pull/27

PR #4219, opened 2025-11-19
Warn about non-mergeable units instead of raising an error

chrishalcrow:
This PR slightly changes the merging logic. It checks to see which units are mergeable (wrt to the sparsity threshold) before trying to merge.

Now we warn and don't merge the units, rather than error the entire process, if individual units cannot be merged.

I tried implementing this inside of `_save_or_select_or_merge_or_split` but it was a mess. I think it's best to know your new unit ids and final merge groups before this function is called.

@samuelgarcia @yger feedback welcome! I need to add some tests...

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -1264,7 +1264,7 @@ class SortingAnalyzer:
         censor_ms: float | None = None,
         merging_mode: str = "soft",
         sparsity_overlap: float = 0.75,
-        error_if_overlap_fails: bool = False,
+        raise_error_if_overlap_fails: bool = True,
         new_id_strategy: str = "append",
         return_new_unit_ids: bool = False,
         format: str = "memory",
@@ -1295,8 +1295,9 @@ class SortingAnalyzer:
         sparsity_overlap : float, default 0.75
             The percentage of overlap that units should share in order to accept merges. If this criteria is not
             achieved for a pair of units, soft merging will not be applied to them.
-        error_if_overlap_fails : bool, default: False
-            If True and `sparsity_overlap` fails for any unit, this will raise an error.
+        raise_error_if_overlap_fails : bool, default: True
+            If True and `sparsity_overlap` fails for any unit merges, this will raise an error. If False, units which fail the
+            `sparsity_overlap` threshold will be skipped in the merge.
         new_id_strategy : "append" | "take_first", default: "append"
             The strategy that should be used, if `new_unit_ids` is None, to create new unit_ids.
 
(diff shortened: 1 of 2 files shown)


samuelgarcia:
I think the raise was good.
Maybe we could have an option to raise or warns.
So first it raise and then it the user really want, it only has a warn.
Lets have a cool debate.

chrishalcrow:
Ok I'll add an option, then we'll debate the default.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1300,7 +1294,9 @@ def merge_units(
             reloading waveforms if needed
         sparsity_overlap : float, default 0.75
             The percentage of overlap that units should share in order to accept merges. If this criteria is not
-            achieved, soft merging will not be possible and an error will be raised
+            achieved for a pair of units, soft merging will not be applied to them.
+        error_if_overlap_fails : bool, default False
```
```suggestion
        error_if_overlap_fails : bool, default: False
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1271,6 +1264,7 @@ def merge_units(
         censor_ms: float | None = None,
         merging_mode: str = "soft",
         sparsity_overlap: float = 0.75,
+        error_if_overlap_fails=False,
```
```suggestion
        error_if_overlap_fails: bool = False,
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1300,7 +1294,9 @@ def merge_units(
             reloading waveforms if needed
         sparsity_overlap : float, default 0.75
             The percentage of overlap that units should share in order to accept merges. If this criteria is not
-            achieved, soft merging will not be possible and an error will be raised
+            achieved for a pair of units, soft merging will not be applied to them.
+        error_if_overlap_fails : bool, default: False
```
alejoe91: - rename to `raise_error_if...`
- propagate to `curation` module
- propagate to `SortingAnalyzer.merge_units`

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1333,15 +1329,42 @@ def merge_units(
...
+            mergeable_unit_groups = []
+            unmergeable_unit_groups = []
+            for merge_unit_group, mergeable in zip(merge_unit_groups, mergeable.values()):
+                if mergeable:
+                    mergeable_unit_groups.append(merge_unit_group)
+                else:
+                    unmergeable_unit_groups.append(merge_unit_group)
+
+            if len(unmergeable_unit_groups) > 0:
+                error_or_warning_message = f"The sparsity of the units in the merge groups {unmergeable_unit_groups} do not overlap enough for a soft merge using a sparsity threshold of {sparsity_overlap}. They will not be merged."
```
alejoe91: mention the flag in the error message

chiyu1203:
@chrishalcrow sorry, this is off topic but I have seen this warning when I used **similarity** in the auto merge algorithm on the GUI but could not find relevant information to read. Why is **sparsity threshold** the criterion to check if units are mergeable? Does the **sparsity** mean the spare representation of spikes?

chrishalcrow:
> @chrishalcrow sorry, this is off topic but I have seen this warning when I used **similarity** in the auto merge algorithm on the GUI but could not find relevant information to read. Why is **sparsity threshold** the criterion to check if units are mergeable? Does the **sparsity** mean the spare representation of spikes?

Hi @chiyu1203, I've added an issue (#4312) to remind me to add this to the docs!

Sparsity is related to the templates: it represents which channels the unit is localized on. When you try and merge two units, the first thing spikeinterface does is to check which channels the unit is active on. If 80% of the channels are the same for the two units, you get a shared sparsity of 0.8. The sparsity threshold says "you can only merge units if the shared sparsity is above the sparsity threshold". Does that make sense?

chiyu1203:
> > @chrishalcrow sorry, this is off topic but I have seen this warning when I used **similarity** in the auto merge algorithm on the GUI but could not find relevant information to read. Why is **sparsity threshold** the criterion to check if units are mergeable? Does the **sparsity** mean the spare representation of spikes?
> 
> Hi @chiyu1203, I've added an issue (#4312) to remind me to add this to the docs!
> 
> Sparsity is related to the templates: it represents which channels the unit is localized on. When you try and merge two units, the first thing spikeinterface does is to check which channels the unit is active on. If 80% of the channels are the same for the two units, you get a shared sparsity of 0.8. The sparsity threshold says "you can only merge units if the shared sparsity is above the sparsity threshold". Does that make sense?

@chrishalcrow Thanks for the explanation! Yes, that makes perfect sense!

PR #4227, opened 2025-11-24
Auto-estimate margins based on freq_min/f0+Q for filters

alejoe91:
@samuelgarcia let's discuss!

### DIFF
--- a/src/spikeinterface/preprocessing/tests/test_filter.py
+++ b/src/spikeinterface/preprocessing/tests/test_filter.py
@@ -159,7 +159,7 @@ def test_filter():
     trace1 = rec2_cached1.get_traces(segment_index=0)
 
     # other filtering types
-    rec3 = filter(rec, band=500.0, btype="highpass", filter_mode="ba", filter_order=2)
+    rec3 = filter(rec, band=500.0, btype="highpass", filter_mode="ba", filter_order=2, margin_ms=5.0)
     rec4 = notch_filter(rec, freq=3000, q=30, margin_ms=5.0)
     rec5 = causal_filter(rec, direction="forward")
     rec6 = causal_filter(rec, direction="backward")
@@ -168,7 +168,7 @@ def test_filter():
     from scipy.signal import iirfilter
 
     coeff = iirfilter(8, [0.02, 0.4], rs=30, btype="band", analog=False, ftype="cheby2", output="sos")
-    rec5 = filter(rec, coeff=coeff, filter_mode="sos")
+    rec5 = filter(rec, coeff=coeff, filter_mode="sos", margin_ms=5.0)
 
     # compute by chunk
     rec5_cached0 = rec5.save(chunk_size=100000, verbose=False, progress_bar=True)
--- a/src/spikeinterface/preprocessing/tests/test_pipeline.py
+++ b/src/spikeinterface/preprocessing/tests/test_pipeline.py
@@ -80,6 +80,9 @@ def test_pipeline_equiv_to_step():
             elif pp_name == "decimate":
                 pp_dict[pp_name] = {"decimation_factor": 2}
                 pp_rec_from_class = pp_class(rec, decimation_factor=2)
+            elif pp_name == "filter":
+                pp_dict[pp_name] = {"margin_ms": 5.0}
+                pp_rec_from_class = pp_class(rec, margin_ms=5.0)
             else:
                 pp_rec_from_class = pp_class(rec)
 


yger:
I think this is a great feature, but I already told @samuelgarcia that maybe we should at very least warn (prevent ?) users that will use very low frequencies for filters, trying to get LFP and/or low pass versions of the data. If the automatic margin starts to be really large (maybe larger than the default chunk size of 1s), then we should warn users that they are doing something clearly suboptimal.

alejoe91:
> I think this is a great feature, but I already told @samuelgarcia that maybe we should at very least warn (prevent ?) users that will use very low frequencies for filters, trying to get LFP and/or low pass versions of the data. If the automatic margin starts to be really large (maybe larger than the default chunk size of 1s), then we should warn users that they are doing something clearly suboptimal.

Sub-optimal but somehow needed! Maybe we can warn if the margin is greater than the `max_margin_s` (default 5 s). What do you think?

yger:
Yes, I've read the PR, and indeed, at least a warning that the theoretical margin should be XX, and that it has be clipped to max_margin_s. The thing is that if users really want to do that, then the warning message could also advise them to increase chunk_duration in the default_job_kwargs, to reduce IO overloads

src/spikeinterface/preprocessing/filter.py
```diff
@@ -78,7 +78,8 @@ def __init__(
         filter_order=5,
         ftype="butter",
         filter_mode="sos",
-        margin_ms=5.0,
+        margin_ms=5,
+        max_margin_s=5,
```
samuelgarcia: Not sure I like this concept of max margin...
We really should prevent users to use spikeinterface for lfp low frequency purpose.

alejoe91: We have to use it! For wideband signals, there is no way around. You have to extract LFP on chunks. Extending the margin will minimize the error and the max margin is a tradeoff between accuracy and performance.

src/spikeinterface/preprocessing/filter.py
```diff
@@ -229,7 +249,11 @@ class BandpassFilterRecording(FilterRecording):
         The bandpass-filtered recording extractor object
     """
 
-    def __init__(self, recording, freq_min=300.0, freq_max=6000.0, margin_ms=5.0, dtype=None, **filter_kwargs):
+    def __init__(
+        self, recording, freq_min=300.0, freq_max=6000.0, margin_ms="auto", max_margin_s=5, dtype=None, **filter_kwargs
+    ):
```
yger, line 258: Can we hare here a warning (maybe with an option to turn it off), such that if users are providing very low freq_min, we warn them of two things:
- if max_margin_ms is safe or not (will it be clipped or not)
- that to speed up everything, they should consider using larger chunks (via set_global_job_kwargs). Because at the creation of the filters, we do not have this info, but if we are reading chunks of 100ms with a max_margin_ms of 5s, this is clearly suboptimal, and users should be aware they should load larger chunks

samuelgarcia:
We democratically decided to remove the max_margin kwargs and instead make a mechanism that raise an error if the end-user try to filter in LFP band and then point to a link to a clear and didactic and really exemplar tutorial about chunking and its bad effect on low freqs.

alejoe91:
This is ready for final review. here's a new doc page explaining the issue: https://spikeinterface--4227.org.readthedocs.build/en/4227/how_to/extract_lfps.html

examples/how_to/extract_lfps.py
```diff
@@ -0,0 +1,310 @@
...
+# # Extract LFPs
+#
+# ### Understanding filtering artifacts and chunking when extracting LFPs
+#
+# Local Field Potentials (LFPs) are low-frequency signals (<300 Hz) that reflect the summed activity of many neurons.
+# Extracting LFPs from high-sampling-rate recordings requires bandpass filtering, but this can introduce artifacts
+# when not done carefully, especially when data is processed in chunks (for memory efficiency).
+#
+# This tutorial demonstrates:
+# 1. How to generate simulated LFP data
```
chrishalcrow: I think you need a space to render this as a list
```suggestion

# 1. How to generate simulated LFP data
```

doc/forhowto/plot_extract_lfps.py
```diff
@@ -0,0 +1,310 @@
...
+    recording_lfp_1hz = spre.bandpass_filter(recording_with_lfp, freq_min=1.0, freq_max=300.0)
+    print("Filtering succeeded!")
+except Exception as e:
+    print(f"Error message:\n{str(e)}")
+
+# %% [markdown]
+# **Why does this fail?**
+#
+# The error occurs because by default in SpikeInterface when highpass filtering below 100 Hz.
+# Filters with very low cutoff frequencies have long impulse responses, which require larger margins to avoid edge artifacts between chunks.
```
chrishalcrow, line 142: I think we need to explain what a margin and a chunk is more clearly before this?
If I were writing it, I'd have a first section called "1. Chunks and margins" and explain what chunking is, why we do it, and what a margin is. Or put in more explanation here.

chrishalcrow: Oh, I've read more now... so I would put Section 3 earlier in the tutorial. And put in a longer description of a chunk and margin.

examples/how_to/extract_lfps.py
```diff
@@ -0,0 +1,310 @@
...
+    if chunk != "10s":
+        continue
+    diff = recording_optimal - recording_chunk
+    traces_diff = diff.get_traces(start_frame=start_frame, end_frame=end_frame)
+    ax.plot(traces_diff, label=recording_key)
+
+ax.legend()
+
+# %% [markdown]
+# For smaller chunk sizes, these artifact will happen more often. In addition, the margin "overhead" will make processing slower.
```
chrishalcrow: This comes straight after the plot. Can you explain what we see in the plot. Something like: 

From the plot, we can see that there is a very small error when the margin size is large (green), a larger error when the margin is smaller (organge) and a large error when the margin is small (blue). So we need large margins (compared to the chunk size) if we want accurate filtered.

For smaller...

examples/how_to/extract_lfps.py
```diff
@@ -0,0 +1,310 @@
...
+    end_frame = int(35 * recording_optimal.sampling_frequency)
+    traces_opt = recording_optimal.get_traces(start_frame=start_frame, end_frame=end_frame)
+    if not trace_plotted:
+        ax.plot(traces_opt, color="grey", label="traces", alpha=0.5)
+        trace_plotted = True
+    if chunk != "10s":
+        continue
+    diff = recording_optimal - recording_chunk
+    traces_diff = diff.get_traces(start_frame=start_frame, end_frame=end_frame)
+    ax.plot(traces_diff, label=recording_key)
```
chrishalcrow: Can you make the axes in seconds, please? Then it's much easier to see where 30s is.

examples/how_to/extract_lfps.py
```diff
@@ -0,0 +1,310 @@
...
+print(f"Margin: {margin_in_s} s")
+
+# %% [markdown]
+# This effectively means that if we plot 1-s snippet of traces, a total of 11 s will actually be read and filtered. Note that the margin can be overridden with the `margin_ms` argument, but we do not recommend changing it.
+
+# %%
+sw.plot_traces(recording_filt, time_range=[20, 21])
+
+# %% [markdown]
+# A warning tells us that what we are doing is not optimized, since in order to get the requested traces the marging "overhead" is very large.
```
chrishalcrow:
```suggestion
# A warning tells us that what we are doing is not optimized, since in order to get the requested traces the margin-ing "overhead" is very large.
```
? or just "margin"

chrishalcrow:
Hello, amazing!

You could set this doc up as a jupyter notebook that gets run in the CI (e.g. https://github.com/SpikeInterface/spikeinterface/tree/main/examples/tutorials/forhowto)

Then we don't have to do all the .py -> .rst conversion annoying stuff.

examples/how_to/extract_lfps.py
alejoe91: done

examples/how_to/extract_lfps.py
alejoe91: fixed

examples/how_to/extract_lfps.py
alejoe91: added

alejoe91:
> Hello, amazing!
> 
> You could set this doc up as a jupyter notebook that gets run in the CI (e.g. https://github.com/SpikeInterface/spikeinterface/tree/main/examples/tutorials/forhowto)
> 
> Then we don't have to do all the .py -> .rst conversion annoying stuff.

Done! Also moved the `forhowto` out of `tutorials`, so it's a nice separation!

doc/forhowto/plot_extract_lfps.py
alejoe91: added at the top of the file

alejoe91:
@chrishalcrow needed some juggling to make it work with RTD RAM (the script used to save everything in memory), but it works now!

Let me know if you have more comments: https://spikeinterface--4227.org.readthedocs.build/en/4227/forhowto/plot_extract_lfps.html

examples/forhowto/plot_extract_lfps.py
```diff
@@ -0,0 +1,332 @@
...
+    recording_lfp_1hz = spre.bandpass_filter(
+        recording_with_lfp, freq_min=1.0, freq_max=300.0
+    )
+except Exception as e:
+    print(f"Error message:\n{str(e)}")
+
+##############################################################################
+# **Why does this fail?**
+#
+# The error occurs because by default in SpikeInterface when highpass filtering below 100 Hz.
```
chrishalcrow: Current sentence doesn't make sense. Maybe:
```suggestion
# The error always occurs in SpikeInterface when highpass filtering below 100 Hz, to remind the user that they need to be careful.
```
?

examples/forhowto/plot_extract_lfps.py
```diff
@@ -0,0 +1,332 @@
...
+
+##############################################################################
+# 3. Understanding chunking and margins
+# --------------------------------------
+#
+# SpikeInterface processes recordings in chunks to handle large datasets efficiently. Each chunk needs
+# a "margin" (extra samples at the edges) to avoid edge artifacts when filtering. Let's demonstrate
+# this by saving the filtered data with different chunking strategies.
+#
+# **This error is to inform the user that extra care should be used when dealing with LFP signals!**
```
chrishalcrow: The start of this section is a bit confusing. This line refers to the error from the previous section -- but since we're in a new section I think the reader will be confused. Maybe move this statement to just after the error?

examples/forhowto/plot_extract_lfps.py
```diff
@@ -0,0 +1,332 @@
...
+#
+# We can ignore this error, but let's make sure we understand what it's happening.
+
+# We can ignore this error, but let's see what is happening
```
chrishalcrow: This line and the line above repeat

examples/forhowto/plot_extract_lfps.py
```diff
@@ -0,0 +1,332 @@
...
+##############################################################################
+# When retrieving traces, extra samples will be retrieved at the left and right edges.
+# By default, the filter function will set a margin to 5x the sampling period associated to `freq_min`.
+# So for a 1 Hz cutoff frequency, the margin will be 5 seconds!
+
+margin_in_s = recording_filt.margin_samples / recording_lfp.sampling_frequency
+print(f"Margin: {margin_in_s} s")
+
+##############################################################################
+# This effectively means that if we plot 1-s snippet of traces, a total of 11 s will actually be read and filtered.
```
chrishalcrow, line 174: I think it's better to mention the overhead here, since this sentence explains very clearly why there is a big overhead. E.g.

```suggestion
# This effectively means that if we plot 1-s snippet of traces, a total of 11 s will actually be read and filtered. Hence the computational "overhead" is very large.
```

examples/forhowto/plot_extract_lfps.py
```diff
@@ -0,0 +1,332 @@
...
+# Note that the margin can be overridden with the `margin_ms` argument, but we do not recommend changing it.
+
+_ = sw.plot_traces(recording_filt, time_range=[20, 21])
+
+##############################################################################
+# A warning tells us that what we are doing is not optimized, since in order to get the requested traces
```
chrishalcrow, line 181: There's no warning in the output now, sorry, probably something to do with changing how it's rendered?

examples/forhowto/plot_extract_lfps.py
```diff
@@ -0,0 +1,332 @@
...
+        ax.axvline(x=chunk, color="red", linestyle="--", alpha=0.5)
+
+ax.set_xlabel("Time (s)")
+ax.set_ylabel("Voltage ($\\mu V$)")
+_ = ax.legend()
+
+##############################################################################
+# From the plot, we can see that there is a very small error when the margin size is large (green),
+# a larger error when the margin is smaller (orange) and a large error when the margin is small (blue).
+# So we need large margins (compared to the chunk size) if we want accurate filtered.
```
chrishalcrow, line 273: Beautiful!

chrishalcrow: Apart from the lil things, this is great - I'm happy!

examples/forhowto/plot_extract_lfps.py
alejoe91: I do see the warning locally, but not on the build docs...not sure why! Should I just remove it?

chrishalcrow: yeah - just remove it

src/spikeinterface/preprocessing/filter.py
```diff
@@ -78,7 +83,7 @@ def __init__(
         filter_order=5,
         ftype="butter",
         filter_mode="sos",
-        margin_ms=5.0,
+        margin_ms=5,
```
samuelgarcia:
```suggestion
        margin_ms=5.0,
```

PR #4228, opened 2025-11-25
Add channel_filter parameter for bad channel detection

tayheau:
Hey, so i though it could be useful to add the possibility to filter the bad_channel_ids based on their labels. 

The idea is to add a new `channel_filter : str | set  = { "noise", "dead", "out" }`  arg to the `preprocessing.detect_bad_channels` function. This one would only apply to the coherence+psd method and would filter the returned `bad_channel_ids` based on their corresponding label. 

The main use case would be with usage of the `PreprocessingPipeline`. 
For example, based on the [preprocessing pipeline of the IBL](https://doi.org/10.6084/m9.figshare.19705522.v3) that would allow us to use `detect_and_remove_bad_channels`  only on channels labeled as "out" or "dead", and then to perform interpolate only the "noise" labeled channels using `detect_and_interpolate_bad_channels` using a kwargs dictionnary similar to this one : 
```
preprocessing_kwargs = {
    "highpass_filter": { 'ftype': 'bessel', 'dtype': 'float32' },
    "phase_shift": { },
    "common_reference": { },
    "detect_and_remove_bad_channels": { 'channel_filter': {'dead', 'out'} },
    "detect_and_interpolate_bad_channels": { },
    ...
}
```

This would allow to add a little complexity to `PreprocessingPipeline` while keeping it's main philosophy intact.

### DIFF
--- a/src/spikeinterface/preprocessing/detect_bad_channels.py
+++ b/src/spikeinterface/preprocessing/detect_bad_channels.py
@@ -173,7 +173,7 @@ def detect_bad_channels(
     neighborhood_r2_threshold: float = 0.9,
     neighborhood_r2_radius_um: float = 30.0,
     seed: int | None = None,
-    channel_filters: set | None  = None,
+    channel_filters: set | None = None,
 ):
     """
     Perform bad channel detection.
@@ -310,9 +310,7 @@ def detect_bad_channels(
             channel_filters = allowed_filters
 
         if not isinstance(channel_filters, set):
-            raise ValueError(
-                f"channel_filters must be None or a set of the following values : {allowed_filters} "
-            )
+            raise ValueError(f"channel_filters must be None or a set of the following values : {allowed_filters} ")
 
         if not channel_filters.issubset(allowed_filters):
             raise ValueError(


src/spikeinterface/preprocessing/detect_bad_channels.py
```diff
@@ -171,6 +173,7 @@ def detect_bad_channels(
     neighborhood_r2_threshold: float = 0.9,
     neighborhood_r2_radius_um: float = 30.0,
     seed: int | None = None,
+    channel_filter: str | set = {"noise", "dead", "out"},
```
alejoe91: I would set this to None by default to keep back-compatibility. In case it's None, it can be set to `{"noise", "dead", "out"}` by default

src/spikeinterface/preprocessing/detect_bad_channels.py
```diff
@@ -171,6 +173,7 @@ def detect_bad_channels(
     neighborhood_r2_threshold: float = 0.9,
     neighborhood_r2_radius_um: float = 30.0,
     seed: int | None = None,
+    channel_filter: str | set | None = None,
```
chrishalcrow: Just to keep things slightly conceptually simpler, I'd prefer to call this `channel_filters` and only accept `set | None`, just to not mix up singular parameter names and plural data. Plus the code below will simplify a bit. What do you think?

src/spikeinterface/preprocessing/detect_bad_channels.py
```diff
@@ -74,6 +74,8 @@
     Spatial radius below which two channels are considered neighbors in the neighborhood_r2 method.
 seed : int or None, default: None
     The random seed to extract chunks
+channel_filter : str or set or None, default: None
```
chrishalcrow: Sorry, incredibly boring/annoying comment here: we've been trying to keep numpydoc docstring conventions, and use "|" instead of "or": https://spikeinterface.readthedocs.io/en/stable/development/development.html#docstring-conventions

src/spikeinterface/preprocessing/detect_bad_channels.py
```diff
@@ -74,6 +74,8 @@
...
+    For coherence+psd - filter the ids of the returned channels based on their labels.
```
chrishalcrow: It took me a little while to realise what this meant. How about:

```suggestion
    For coherence+psd - only return `bad_channel_ids` whose labels in are in the set `channel_filter`.
```

Is that right??

chrishalcrow:
This is great!! Some tiny comments :)

src/spikeinterface/preprocessing/detect_bad_channels.py
tayheau: oh no problem, it's my bad ;D

src/spikeinterface/preprocessing/detect_bad_channels.py
tayheau: yes, it's more intuitive ! let's go for it

PR #4229, opened 2025-11-25
Fix no disk space left ci

alejoe91:
ahould fix recent failures due to shortage of disk space


samuelgarcia:
outch. is it legal ?

PR #4234, opened 2025-11-28
fix probable shared mem no clean corretcly

samuelgarcia:
fix probable shared mem no clean corretcly


yger (approved):
Looks good to me, and seems to help here with python 3.12 and fresh ubuntu install

PR #4237, opened 2025-12-02
Bug fix in SC2

OookTheLibrarian:
Originaly the function was adding a peak_svd entry into clustering_params but clustering_methods requires peaks_svd instead, so these parameters were not being passed to the methods.


yger (approved):
Good catch, thanks for spotting that, code just went through major refactoring...

PR #4238, opened 2025-12-02
Implement Add/SubtractRecording classes and +- operators

alejoe91:
Something that's been on my todo list for a while.

This PR enables one to do `recording1 +/- recording2` :)

### DIFF
--- a/src/spikeinterface/core/operatorrecordings.py
+++ b/src/spikeinterface/core/operatorrecordings.py
@@ -15,7 +15,9 @@ class BaseOperatorRecording(BaseRecording):
         assert do_recording_attributes_match(
             recording1, rec_attrs2
         ), "Both recordings must have the same sampling frequency and channel ids"
-        assert self.check_times_kwargs(recording1, recording2), "Both recordings must have the same time parameters"
+        assert self.are_times_kwargs_compatible(
+            recording1, recording2
+        ), "Both recordings must have the same time parameters"
 
         channel_ids = recording1.channel_ids
         sampling_frequency = recording1.sampling_frequency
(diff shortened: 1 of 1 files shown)


h-mayorquin (commented):
I think this will make the data model more expressive. Some edge cases:
* Different number of segments.
* Different number of samples
* Different time parameters (either t_starts or explicit timestamps).

alejoe91:
> I think this will make the data model more expressive. Some edge cases:
> * Different number of segments.
> * Different number of samples
> * Different time parameters (either t_starts or explicit timestamps).

First 2 are enforced and will raise an error if different. Time params are not, but maybe we should?

h-mayorquin:
Yes, I think we should.

samuelgarcia:
Cool.
Maybe this could implemented in one class to avoid the copy/paste.

h-mayorquin:
> > I think this will make the data model more expressive. Some edge cases:
> > 
> > * Different number of segments.
> > * Different number of samples
> > * Different time parameters (either t_starts or explicit timestamps).
> 
> First 2 are enforced and will raise an error if different. Time params are not, but maybe we should?

Yes, in my first read I did not realize that this is done `do_recording_attributes_match`. I think that it can be a separated PR and discussion whether to add time basis to do the `do_recording_attributes_match` or add another function that checks that. 

I am Ok with this.

src/spikeinterface/core/operatorrecordings.py
```diff
@@ -0,0 +1,80 @@
...
+
+        BaseRecording.__init__(self, sampling_frequency, channel_ids, dtype)
+
+        for segment1, segment2 in zip(recording1._recording_segments, recording2._recording_segments):
+            add_segment = OperatorRecordingSegment(segment1, segment2, operator)
+            self.add_recording_segment(add_segment)
+
+        self._kwargs = dict(recording1=recording1, recording2=recording2, operator=operator)
+
+    def check_times_kwargs(self, recording1, recording2) -> bool:
```
h-mayorquin:
```suggestion
    def are_time_kwargs_compatible(self, recording1, recording2) -> bool:
```

More explicit and usual convention for functions that return bool.

PR #4239, opened 2025-12-03
Fix probable sharedmem

samuelgarcia:
The #4234 was merge too quickly!!


yger (approved):
It is solving the issues on python3.12 and 3.13

PR #4245, opened 2025-12-05
Fix artifact removal padding logic

mhhennig:
This fixes a bug in `RemoveArtifactsRecordingSegment`

Previously artifacts would not be removed if the trigger was before `start_frame` but the blanking carried on beyond. Only tested on `zeros` mode.


src/spikeinterface/preprocessing/remove_artifacts.py
```diff
@@ -263,12 +263,14 @@ def get_traces(self, start_frame, end_frame, channel_indices):
             traces = self.parent_recording_segment.get_traces(start_frame, end_frame, channel_indices)
         traces = traces.copy()
 
-        mask = (self.triggers >= start_frame) & (self.triggers < end_frame)
+        pad = self.pad
+        if pad is None:
+            mask = (self.triggers >= start_frame) & (self.triggers < end_frame)
+        else:
+            mask = (self.triggers >= start_frame - pad[1]) & (self.triggers < end_frame)
```
chrishalcrow: Also at the end? I think it would be:
```suggestion
            mask = (self.triggers >= start_frame - pad[1]) & (self.triggers < end_frame + pad[0])
```

mhhennig: No, triggers after the end frame don't matter, the data will never be read.

mhhennig: NVM you are right!

src/spikeinterface/preprocessing/remove_artifacts.py
```diff
@@ -263,12 +263,14 @@ def get_traces(self, start_frame, end_frame, channel_indices):
             traces = self.parent_recording_segment.get_traces(start_frame, end_frame, channel_indices)
         traces = traces.copy()
 
-        mask = (self.triggers >= start_frame) & (self.triggers < end_frame)
+        pad = self.pad
+        if pad is None:
+            mask = (self.triggers >= start_frame) & (self.triggers < end_frame)
+        else:
+            mask = (self.triggers >= start_frame - pad[1]) & (self.triggers < end_frame + pad[0])
```
alejoe91, line 270:
```suggestion
            mask = (self.triggers >= start_frame - pad[0]) & (self.triggers < end_frame + pad[1])
```
`pad` is defined as `pad = [int(ms_before * fs / 1000), int(ms_after * fs / 1000)]`

mhhennig: This is wrong (tested it to be sure), this would fails if `start_frame>trigger-pad[0]` and `pad[0]<pad[1]`.

It has to be this as in Chris' edit:
`mask = (self.triggers >= start_frame - pad[1]) & (self.triggers < end_frame + pad[0])`

One needs to look back for triggers before `start_frame-pad[1]` if `start_frame` is after a trigger time point (otherwise the trigger is just ignored), or ahead by `end_frame - pad[0]` if `start_frame` and `end_frame` areboth before a trigger.

chrishalcrow (approved):

Issue #4180, opened 2025-10-21
Setting probe after `unsigned_to_signed` loses `bit_depth`


alejoe91:
from https://github.com/SpikeInterface/spikeinterface/issues/4170#issuecomment-3398959351

Maybe related, for some reason using set_probe on an unsigned recording changes the value returned by get_traces, this does not happen if I call set_probe first and then convert to signed. I don't see anything in the probe mapping that would change the gain or offset.

```
recs = dict()
recs['raw'] = se.read_binary(
        ephys_path / f"np2-{prb_suffix}-ephys_{suffix}.raw",
        sampling_frequency=fs_hz,
        dtype=np.uint16,
        num_channels=num_channels,
    )
recs['unsigned_from_raw'] = si.unsigned_to_signed(recs['raw'], bit_depth=bit_depth)
print(recs['raw'].dtype)
recs['probe_from_unsigned'] = recs['unsigned_from_raw'].set_probe(npx_mapping, in_place=False)
# The other way round
recs['probe_from_raw']= recs['raw'].set_probe(npx_mapping, in_place=False)
recs['unsigned_from_probe']=  si.unsigned_to_signed(recs['probe_from_raw'], bit_depth=bit_depth)

for k, rec in recs.items():
    trace =  rec.get_traces(start_frame=0, end_frame=2, return_in_uV=False, channel_ids=[30])
    print(f"{k}: {trace[0,0]} (dtype: {rec.dtype})")
```
Prints:
```
raw: 3847 (dtype: uint16). [expected]
unsigned_from_raw: 1799 (dtype: int16) [expected]
probe_from_unsigned: -28921 (dtype: int16) [BAD expected is 1799]
probe_from_raw: 3847 (dtype: uint16) [expected]
unsigned_from_probe: 1799 (dtype: int16) [expected]
```

alejoe91:
fixed by  #4247

PR #4247, opened 2025-12-06
Fixes Issue4180

ablot:
Issue #4180

[preserve bit_depth parameter in UnsignedToSignedRecording serialization](https://github.com/SpikeInterface/spikeinterface/commit/793690f2defb9fdca2af5a7e5bcd2cfa85b81cb0) 

The UnsignedToSignedRecording class was missing the bit_depth parameter
in its _kwargs dictionary, causing it to be lost during clone operations.
This resulted in incorrect trace values when set_probe() was called on
unsigned-to-signed recordings with custom bit depths.

I think that doesn't affect anything else but I'm quite new to spikeinterface

### DIFF
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -11,12 +11,14 @@ It also implements:
 
 import warnings
 import numpy as np
+from collections import namedtuple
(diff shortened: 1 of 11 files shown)


samuelgarcia:
merci

PR #4253, opened 2025-12-10
Fix failing tests due to scikit-learn versions

chrishalcrow:
We have some failing tests: https://github.com/SpikeInterface/spikeinterface/actions/runs/20093848765/job/57647307083?pr=4190

In our automated curation tests, we previously had a stored model. This was generated using scikit-learn v1.3.2. this is now incompatible with scikit-learn v1.9. 

There's a bigger problem here: how do we ensure models are usable in the future (one possible answer: by default we store everything used to train them, so you can retrain using a newer sklearn verison).

More practically, so that this doesn't happen again: I've replaced the fixed created model with one that is created during ci (=> using the scikit-learn verison of the environment). You should only have to do this once per test suite run.


src/spikeinterface/curation/tests/common.py
```diff
@@ -82,6 +84,31 @@ def sorting_analyzer_with_splits():
     return make_sorting_analyzer_with_splits(sorting_analyzer)
 
 
+def make_trained_pipeline():
```
alejoe91: Make it a fixture? So we don't need the if trained folder and it always returns the model folder

chrishalcrow: Oooh, that's nicer! Thanks!

PR #4254, opened 2025-12-10
Flatten params for tridesclous2 sorter

samuelgarcia:
Flatten params for tridesclous2 sorter.
This will be more easier for end users.

This depend on #4244


src/spikeinterface/sorters/internal/tridesclous2.py
```diff
@@ -32,61 +32,67 @@ class Tridesclous2Sorter(ComponentsBasedSorter):
...
-        "matching": "A dictonary containing matching params for matching: peak_shift_ms, radius_um",
-        "job_kwargs": "A dictionary containing job kwargs",
-        "save_array": "Save or not intermediate arrays",
+        "apply_motion_correction": "Apply motion correction or not",
+        "motion_correction_preset": "Motion correction preset",
+        "clustering_ms_before": "Milliseconds before the spike peak for clustering",
+        "clustering_ms_after": "Milliseconds after the spike peak  for clustering",
+        "radius_um": "Radius for sparsity",
+        "freq_min": "Low frequency",
+        "freq_max": "High frequency",
```
alejoe91:
```suggestion
        "freq_min": "Low frequency for bandpass filter",
        "freq_max": "High frequency for bandpass filter",
```
I would actually allow None and not filter by default. no? Or having another argument: `bandpass_filter: True/False`

samuelgarcia:
Merci

PR #4257, opened 2025-12-11
Cap sklearn version to `<1.8`

### LINKED ISSUE #4256: sklearn v1.8 not compatible with hdbscan
scikit-learn have deprecated `force_all_finite`, which `hdbscan` uses here:
https://github.com/scikit-learn-contrib/hdbscan/blob/949ff7aba4e65bcfc77a505948cd50617fbc047e/hdbscan/hdbscan_.py#L767
which breaks our tests:
https://github.com/SpikeInterface/spikeinterface/actions/runs/20129057940/job/57765648034?pr=4252

hdbscan seems aware of the issue: https://github.com/scikit-learn-contrib/hdbscan/pull/673

For now we can limit scikit-learn to <1.8, and wait for hbscan to update.

chrishalcrow:
Should fix #4256 while we wait for hdbscan to update.


yger:
Actually, other long-term option would be to get rid of hdbscan dependency, and use the internal implementation provided by scikit-learn. This would ease the long-term support

PR #4258, opened 2025-12-11
Adding the possibility to add/remove cases to a given Study

yger:
This PR adds the possibility to add/remove cases to a given Study, once it has already been created. When adding new cases, they have to have datasets already part of the Study, otherwise there would be an error

### DIFF
--- a/src/spikeinterface/benchmark/benchmark_base.py
+++ b/src/spikeinterface/benchmark/benchmark_base.py
@@ -67,6 +67,7 @@ class BenchmarkStudy:
                 assert len(key) == num_levels, f"Case key {key} for cases is not homogeneous, tuple negth differ"
             else:
                 raise ValueError("Keys for cases must str or tuple")
+        return levels
 
     @classmethod
     def create(cls, study_folder, datasets={}, cases={}, levels=None):
@@ -94,7 +95,7 @@ class BenchmarkStudy:
         study : BenchmarkStudy
             The created study.
         """
-        cls._check_cases(cases, levels)
+        levels = cls._check_cases(cases, levels)
 
         study_folder = Path(study_folder)
         study_folder.mkdir(exist_ok=False, parents=True)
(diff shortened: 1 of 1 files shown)


src/spikeinterface/benchmark/tests/test_benchmark_sorter.py
```diff
@@ -258,19 +291,48 @@ def test_get_grouped_keys_mapping(create_complex_study):
...
 
+def test_add_remove_cases(create_simple_study):
+    # job_kwargs = dict(n_jobs=2, chunk_duration="1s")
+
+    study_folder = create_simple_study
+    study = SorterStudy(study_folder)
+    print(study)
+
+    # # this run the sorters
+    study.run()
```
samuelgarcia: We should not run the study this is too slow.

src/spikeinterface/benchmark/benchmark_base.py
```diff
@@ -272,6 +272,40 @@ def remove_benchmark(self, key):
...
+                if levels is None:
+                    levels = "level0"
+                else:
+                    assert isinstance(levels, str)
+            elif isinstance(key0, tuple):
+                assert isinstance(key, tuple), f"Key {key} for cases is not homogeneous"
+                num_levels = len(key0)
+                assert len(key) == num_levels, f"Key {key} for cases is not homogeneous, tuple negth differ"
+            else:
+                raise ValueError("Keys for cases must str or tuple")
```
samuelgarcia: this should be factorize with the same logic in the init

src/spikeinterface/benchmark/benchmark_base.py
```diff
@@ -76,27 +94,7 @@ def create(cls, study_folder, datasets={}, cases={}, levels=None):
...
-                len(key) == num_levels for key in cases.keys()
-            ), "Keys for cases are not homogeneous, tuple negth differ"
-            if levels is None:
-                levels = [f"level{i}" for i in range(num_levels)]
-            else:
-                levels = list(levels)
-                assert len(levels) == num_levels
-        else:
-            raise ValueError("Keys for cases must str or tuple")
+        cls._check_cases(cases, levels)
```
samuelgarcia: This need to return levels in case it is changed (whane None)

yger: done

samuelgarcia:
Merci camarade

PR #4259, opened 2025-12-11
Removing hdbscan dependency

yger:
This is in line with #4257 , but maybe one option would be to get rid of HDBSCAN dependency, since now there is a full port in scikit-learn. At the time we started to use hdbscan, this was not the case, but now maybe this would be a good option to ease maintenance and install procedures. 
This PR removes all usages of hdbscan to only rely on scikit. In fact, some advanced options of hdbscan are not ported into the scikit-learn version, but since they are not used anywhere in the code at the moment...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -206,7 +206,7 @@ docs = [
     "MEArec", # Use as an example
     "pandas", # in the modules gallery comparison tutorial
     "numba", # For many postprocessing functions
-    "networkx", 
+    "networkx",
     "skops", # For automated curation
     "scikit-learn", # For automated curation
     "huggingface_hub", # For automated curation
--- a/src/spikeinterface/sortingcomponents/clustering/positions.py
+++ b/src/spikeinterface/sortingcomponents/clustering/positions.py
@@ -14,7 +14,6 @@ else:
     HAVE_SKLEARN = False
 
 
-
 class PositionsClustering:
     """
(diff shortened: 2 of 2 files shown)


chrishalcrow:
Nice! How similar do you think the implementations are?

yger:
Same paper, same reference, but I have not performed an exhaustive comparison. Based on preliminary results from the sorters (the only part really using hdbscan), performances are similar. Maybe we should compare more in depth

yger:
Ok, after a more exhaustive study, looks like indeed hdbscan is a better implementation than scikit-learn

image

Here for example, I generated 10 Gaussian blobs, and computed averaged run times and homogeneity scores for the two implementations. Looks like hdbscan is faster, and more accurate

samuelgarcia:
So we close!!

ecobost:
Can someone explain why this was not merged? Do we still need to pin scikit-learn < 1.8?

chrishalcrow:
There are two implementations of hdbscan: one in its own package and one in sklearn. This PR tested out the idea of using the hdbscan that was implemented in sklearn. After doing it, Pierre showed above that the implementation in hdbscan is faster and more accurate for the type of data we look at. So we are sticking to using the hdbscan package, which closed the PR.

We can unpin sklearn if hdbscan has fixed the bug. Looks like it did on the newest release: 0.8.41. EDIT: we should ensure `hdbscan>=0.8.41`.

PR #4263, opened 2025-12-12
Update to SiNAPS extractor

raffaele222:
Hello, I have made some updates to the extractor to reflect changes in the metadata generated by new versions of SiNAPS Control Suite, changes made with the new spikeinterface updates, and added handling of certain edge cases. Most of the core logic remains unchanged, but I have also added some bits to improve support for all streams, by adding "raw" and "aux" options in the h5 extractor, and fixing the "aux" stream in the binary one.

In short:
- Added documentation to all the utility functions
- Added support for all streams in the h5 extractor, and fixed support for "aux" stream in the binary extractor (by ignoring AUX channel names)
- Added checks for stream_name argument
- Switched from using UnsignedSignedRecording class to unsigned_to_signed method
- Load non .bin binary files as a single stream (fixes issues with loading recordings saved as .dat)
- Added a get_sinaps_probe_info() utility function to get probe's information from metadata
- Added custom checks in get_sinaps_probe() to help find the right probe in the probeinterface library if "probeType" in the metadata is set differently

Also, great work to the whole team! I have been tracking the project for about a year now, and the work and progress that is constantly being made is truly amazing!

### DIFF
--- a/src/spikeinterface/extractors/sinapsrecordingextractors.py
+++ b/src/spikeinterface/extractors/sinapsrecordingextractors.py
@@ -28,9 +28,9 @@ class SinapsResearchPlatformRecordingExtractor(ChannelSliceRecording):
     def __init__(self, file_path: str | Path, stream_name: str = "filt"):
 
         assert stream_name in [
-        "filt",
-        "aux",
-        "raw"
+            "filt",
+            "aux",
+            "raw",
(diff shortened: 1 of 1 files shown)


alejoe91 (approved):
Thanks @raffaele222 

Looks great!

PR #4268, opened 2025-12-15
Fix gaussian filter with time vector

### LINKED ISSUE #4267: Recording segment properties don't survive round-trip to & from zarr
When writing recordings to zarr, properties of the recording object itself are stored, but properties of the segments are not. For example:

```python
import spikeinterface as si
recording: ConcatenateSegmentRecording = my_recording_generator()

# Confirm that the recording and its lone segment have a sampling_frequency property.
assert recording.get_num_segments() == 1
sf_rec = recording.get_sampling_frequency()
sf_seg = recording._recording_segments[0].sampling_frequency
assert sf_rec == sf_seg
print(sf_seg) # 30000.13147632312

recording.save(
    format="zarr",
    folder="path/to/destination.zarr",
    compressor_by_dataset={"traces": WavPack(bps=2.25)},
    filters_by_dataset={"times": [Delta(dtype="float64")]},
    n_jobs=80,
)

reloaded_recording = si.load("path/to/destination.zarr")
print(reloaded_recording.get_sampling_frequency()) # 30000.13147632312
assert reloaded_recording.get_num_segments() == 1
assert reloaded_recording._recording_segments[0].sampling_frequency is None # Oh no!
```

This has downstream effects like causing `spikeinterface.preprocessing.gaussian_filter()` to fail when `GaussianFilterRecordingSegment` tries and fails to get a sampling frequency from the segment. 

The issue is that in `recording.save()` -> `ZarrRecordingExtractor.write_recording()` -> `add_recording_to_zarr_group()`, properties of recording segments are not stored in the `attrs` of the `dset` created by `zarr_group.create_dataset()` in `add_traces_to_zarr()`. I guess the place for this to happen would be in the subsequent call to `add_properties_and_annotations(zarr_group, recording)`, where instead of just setting properties of the recording as new datasets on the "properties" group, there would be some iteration through segments. Maybe a `props_seg{I}` group with attributes as datasets if the attributes can be complex, or just a simple `dset.attrs["sampling_frequency"] = ...` if they are known to be json serializable. (FYI, in Zarr v3 you can save a dictionary of json serializable attributes directly in the call to zarr_group.create_dataset()/create_array()`). 

Or, maybe the properties of segments can be restored entirely using properties from the parent recording during `si.load() -> read_zarr_recording()`. Not sure. 

For now my band-aid is to just set some of the missing properties myself each time I load a recording back from disk.

alejoe91:
Fixes #4267

@grahamfindlay could you check this out? I think it fixes your issue and makes the behavior less ambiguous

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -500,7 +500,6 @@ class BaseRecording(BaseRecordingSnippets):
         assert rs.get_num_samples() == times.shape[0], "times have wrong shape"
 
         rs.t_start = None
-        rs.sampling_frequency = None
         rs.time_vector = times.astype("float64", copy=False)
 
         if with_warning:


grahamfindlay:
@alejoe91 Tested this and can confirm: fixes the issue. Thanks!

src/spikeinterface/preprocessing/filter_gaussian.py
```diff
@@ -50,29 +50,37 @@ def __init__(
...
         self.cached_gaussian = dict()
 
-        sf = parent_recording_segment.sampling_frequency
+        self.sampling_frequency = sampling_frequency
 
         # Margin from widest gaussian
         sigmas = []
         if freq_min is not None:
-            sigmas.append(sf / (2 * np.pi * freq_min))
+            sigmas.append(self.sampling_frequency / (2 * np.pi * freq_min))
```
chrishalcrow: In case of using a `time_vector`, could `self.sampling_frequency` be `None`?

grahamfindlay: As I understand it: 

`recording.sampling_frequency` will not be `None`, even after `recording.set_times()` is called. This is because `recording.set_times()` will set the _segment's_ `sampling_frequency = None`, but does nothing to `recording._sampling_frequency`.

In other words:
```python
recording.set_times(t)
assert recording.sampling_frequency is not None
assert recording._recording_segements[0].sampling_frequency is None
```
I *think* this is the intended behavior. 

Then, the recording-level gaussian filter object (`GaussianFilterRecording`) creates segment-level gaussian filter objects (`GaussianFilterRecordingSegment`), supplying each with `recording.sampling_frequency`. Upon instantiation, these segment-level objects then set their own `self.sampling_frequency = my_parent_recording_sampling_frequency`, which for the reasons above will be non-`None`. 

So in the line you highlighted here, `self.sampling_frequency` will not be `None`, even though a segment's `self.sampling_frequency` _should_ and _would_ be `None` "normally" when `time_vector` is present. 

@alejoe91 What I am *not* sure is intended behavior, is that this un-does the nullifying of the segment sampling frequencies that was performed by `recording.set_times()`. 

In other words:
```python 
recording.set_times(t)
assert recording.sampling_frequency is not None
assert recording._recording_segements[0].sampling_frequency is None
recording = sp.gaussian_filter(recording, ...)
assert recording.sampling_frequency is not None
assert recording._recording_segements[0].sampling_frequency is None # THIS FAILS NOW!
```

chrishalcrow: Thanks @grahamfindlay , very helpful summary.
So `GaussianFilterRecordingSegment` will have a `not None` `sampling_frequency`, even when the input `RecordingSegment`'s `sampling frequency` is `None`. Which might be fine, but might be unexpected.

src/spikeinterface/core/baserecording.py
```diff
@@ -500,6 +500,7 @@ def set_times(self, times, segment_index=None, with_warning=True):
         assert rs.get_num_samples() == times.shape[0], "times have wrong shape"
 
         rs.t_start = None
+        rs.sampling_frequency = None
```
chrishalcrow: remove please
```suggestion
```

src/spikeinterface/preprocessing/filter_gaussian.py
alejoe91: This was actually an issue with the Zarr reader, where if time vector was present, sampling frequency was not set. Now this is fixed, but the GaussianFilter still uses the parent recording sampling freq, which is safer

samuelgarcia (approved):

PR #4270, opened 2025-12-16
Add `read_nwb_as_analyzer` function

alejoe91:
Useful function to instantiate a `SortingAnalyzer` from an NWB file as good as we can :)

TODO

- [ ] Add tests


chrishalcrow:
My experience of trying to load recordings is that the channel locations are often not saved with the nwb recording, but that they are saved somewhere else in the nwb file. @bendichter mentioned in a meeting that they'd thought about this problem, and maybe had a solution?

alejoe91:
This will require some key metadata (e.g., an electrodes table and rel_x/rel_y available). In case some key stuff is missing, it will throw an error!

alejoe91:
This will require some key metadata (e.g., an electrodes table and rel_x/rel_y available). In case some key stuff is missing, it will throw an error!

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1789,3 +1797,280 @@ def read_nwb(file_path, load_recording=True, load_sorting=False, electrical_seri
         outputs = outputs[0]
 
     return outputs
+
+
+def load_analyzer_from_nwb(
```
samuelgarcia: you don't like the name read_nwb_as_analyzer() ? to match the kilosort one.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1789,3 +1797,280 @@ def read_nwb(file_path, load_recording=True, load_sorting=False, electrical_seri
...
+    if "waveform_mean" in units:
+        from spikeinterface.core.analyzer_extension_core import ComputeTemplates, ComputeRandomSpikes
+
+        # instantiate templates
+        analyzer.compute("random_spikes", method="all")
+
+        templates_ext = ComputeTemplates(sorting_analyzer=analyzer)
+        templates_avg_data = np.array([t for t in units["waveform_mean"].values]).astype("float")
+        total_ms = templates_avg_data.shape[1] / analyzer.sampling_frequency * 1000
+        template_params = get_default_analyzer_extension_params("templates")
```
samuelgarcia: I think this is a strange guess.
Do we except nwd to have the same template params as spikeinterface actual version  ?

samuelgarcia: Is there a proper way to do it ?
I think I would go directly to the 1/3 2/2 + warnings meachanism.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1789,3 +1797,280 @@ def read_nwb(file_path, load_recording=True, load_sorting=False, electrical_seri
...
+        templates_ext.data["std"] = templates_std_data
+        templates_ext.run_info["run_completed"] = True
+
+        analyzer.extensions["templates"] = templates_ext
+
+    template_metric_columns = ComputeTemplateMetrics.get_metric_columns()
+    quality_metric_columns = ComputeQualityMetrics.get_metric_columns()
+
+    tm = pd.DataFrame(index=sorting.unit_ids)
+    qm = pd.DataFrame(index=sorting.unit_ids)
```
samuelgarcia: Can we set the correct dtype from the new extension system ?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1789,3 +1797,280 @@ def read_nwb(file_path, load_recording=True, load_sorting=False, electrical_seri
...
+            templates_std_data = np.zeros_like(templates_avg_data)
+        templates_ext.data["std"] = templates_std_data
+        templates_ext.run_info["run_completed"] = True
+
+        analyzer.extensions["templates"] = templates_ext
+
+    template_metric_columns = ComputeTemplateMetrics.get_metric_columns()
+    quality_metric_columns = ComputeQualityMetrics.get_metric_columns()
+
+    tm = pd.DataFrame(index=sorting.unit_ids)
```
h-mayorquin: what is tm?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1789,3 +1797,280 @@ def read_nwb(file_path, load_recording=True, load_sorting=False, electrical_seri
...
+    if compute_extra is not None:
+        if verbose:
+            print(f"Computing extra extensions: {compute_extra}")
+        compute_extra_params = {} if compute_extra_params is None else compute_extra_params
+        analyzer.compute(compute_extra, **compute_extra_params)
+
+    return analyzer
+
+
+def create_dummy_probegroup_from_locations(locations, shape="circle", shape_params={"radius": 1}):
```
h-mayorquin: we should make this private as we might want to change this.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1789,3 +1797,280 @@ def read_nwb(file_path, load_recording=True, load_sorting=False, electrical_seri
...
+    probe.set_contacts(locations, shapes=shape, shape_params=shape_params)
+    probe.set_device_channel_indices(np.arange(len(probe.contact_positions)))
+    probe.create_auto_shape()
+    probegroup = ProbeGroup()
+    probegroup.add_probe(probe)
+
+    return probegroup
+
+
+def make_df(group):
```
h-mayorquin: we should make this private as we might want to change this. Plus, this is a super generic name that we don't want to contaminate any namespace with.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1789,3 +1797,280 @@ def read_nwb(file_path, load_recording=True, load_sorting=False, electrical_seri
...
+        else:
+            channel_ids = electrodes_table_sliced["id"][:]
+        num_samples = [sorting.to_spike_vector()[-1]["sample_index"]]
+        rec_attributes = dict(
+            channel_ids=channel_ids,
+            sampling_frequency=sorting.sampling_frequency,
+            num_channels=len(channel_ids),
+            num_samples=num_samples,
+            is_filtered=True,
+            dtype="float32",
```
h-mayorquin: why do we need the dtype and why is it fixed?

h-mayorquin (commented):
first read

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1789,3 +1797,280 @@ def read_nwb(file_path, load_recording=True, load_sorting=False, electrical_seri
...
+        electrical_series_path=electrical_series_path,
+        unit_table_path=unit_table_path,
+        stream_mode=stream_mode,
+        stream_cache_path=stream_cache_path,
+        cache=cache,
+        storage_options=storage_options,
+        use_pynwb=use_pynwb,
+        t_start=t_start_tmp,
+        sampling_frequency=sampling_frequency,
+    )
```
alejoe91, line 1851: We could use `session_start_time` instead.

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1789,3 +1797,280 @@ def read_nwb(file_path, load_recording=True, load_sorting=False, electrical_seri
...
+                    raise Exception(
+                        f"More than one group, use group_name option to select units. Available groups: {np.unique(group_names)}"
+                    )
+                else:
+                    unit_mask = group_names == group_name
+                    if verbose:
+                        print(f"Selecting {sum(unit_mask)} / {len(units)} units from {group_name}")
+                    sorting = sorting.select_units(unit_ids=sorting.unit_ids[unit_mask])
+                    units = units.loc[units.index[unit_mask]]
+                    electrodes_indices = units["electrodes"]
```
alejoe91, line 1909: we could use the same trick as the "aggregation_key" when instantiating a sorting analyzer from grouped recordings/sortings

src/spikeinterface/extractors/nwbextractors.py
alejoe91: I think we should make this optional at the Analyzer level (same for is_filtered)

alejoe91:
Closing in favor of https://github.com/SpikeInterface/spikeinterface/pull/4645

Issue #4251, opened 2025-12-09
Adding new random spike selection methods


tayheau:
ref to [this](https://github.com/SpikeInterface/spikeinterface/blob/705c9320c854f2a192fa200fe7866cec5a8ffca7/src/spikeinterface/core/sorting_tools.py#L151C1-L151C3), Is `maximum_rate` intended as an ISI based sampling method, or does it simply favor neurons with higher firing rates ?

chrishalcrow:
@tayheau are you looking through TODO comments in the codebase?! You should join us for a maintainers meeting sometime!

With regard to the actual issue, I have no idea...

yger:
This is because maybe one can have other clever ideas on the spikes you want to select, from every neurons, to get meaningful statistics. By percent would be a percentage of spikes instead of a fixzd number, maximum_rate is I think an absolute clip (given recording duration and spikes found)... But you are welcome to propose any method you want to select a subset of spikes, from units, to compute statistics!

tayheau:
A silly question, should we implement a `Strategy` pattern for the different sampling algorithms or is it "overkill" and we should stay with the current format and a bunch of if-else statement ?

alejoe91:
> A silly question, should we implement a `Strategy` pattern for the different sampling algorithms or is it "overkill" and we should stay with the current format and a bunch of if-else statement ?

I think that atm it might be an overkill. If we see in the future that we need to add many more and it makes it much more complicated, we can either switch to a `Strategy` pattern or to a `dict[name: Callable]` pattern (probably easier!)

Then each sampling function could simply be:
```
random_spike_indices = _sample_my_spikes(spikes, **kwargs)
```

tayheau:
I was thinking of having a base strategy class that also implement a `Registry` so that why indeed end up with a `dict[name, Class]` that is automaticly feeded with each defined strategy as a subclass of the base one.

yger:
To be honest, here, I think this is an overkill, as @alejoe91 said. What would be the sampling scheme for such an anaylzer ? I.e. what would be the exotic way of selecting spikes to display Waveforms and compute Templates? The only one I can think of would be to sample spikes only during a certain period of time, but then we should better think about frame_slice() as in #4147  for the SortingAnalyzer. Otherwise, just random spikes or by_percent should be enough, don't you think?

tayheau:
[...]

PR #4276, opened 2025-12-18
Random spike selection new methods

tayheau:
So following #4251, a PR with a few new sampling methods for `random_spikes_selection` : 

- `percentage` cap
- `maximum_rate` cap
- `temporal_bins` sampling for a better temporal representation of the original population, i tried to fully vectorize it so the temporal bin sampling problem is dual with a subsorting problem.

I also added to possibility to `get_segment_durations` to set `segment_indices` to `None`.

### DIFF
--- a/src/spikeinterface/widgets/utils.py
+++ b/src/spikeinterface/widgets/utils.py
@@ -407,7 +407,6 @@ def get_segment_durations(sorting: BaseSorting, segment_indices: list[int] = Non
     ----------
     sorting : BaseSorting
         The sorting object containing spike data
-
     segment_indices : list[int] | None
         List of the segment indices to process. Default to None.
 


samuelgarcia:
Hi @tayheau.
I am ok with this change, could you put in the code and docstring a bit more explanation on the strategy for this new "temporal_bins" method, please ?
I am not sure to follow myself entirely the logic

tayheau:
Hey @samuelgarcia, yeah of course, just to summarize here quickly: 

So the method is temporal bin sampling and i expressed it as a sub sampling problem. 

- `bin_index = unit_spikes["sample_index"] // bin_size_freq` allow me to allocate a `bin_index` to every spike in the unit, but `bin_index` will reset for each segment since we use a concatenated spike vector. 

- `group_values = np.stack((segment_index, bin_index), axis=1)` allow me to create a (bin, segment) key that share the same indices as `unit_spikes` so that the the (bin, segment) key of the spike found at `spikes[index]` can be find at `group_values[index]`.

- ` _, group_keys = np.unique(group_values, return_inverse=True, axis=0)` just allow me to have a single integer key to rather than a (bin, segment) key. 

- I then assign a random value (`score`) to each spike in the unit. Instead of making bins and do a random uniform distribution of k samples for each bins, we just go for each (bin, segment) subgroups and randomly sort your samples and take the k first. So now we are working with 3 main data : 
   -  `spikes` which is the concatenated spike vector of the current unit
   - a `group_keys` vector which assimilate a spike of a given index to a (bin, segment) group identified by an integer
   - a `score` vector which assimilate a spike of a given index to a random value 
 
- `order = np.lexsort((score, group_keys))` is an array of indices that will allow us to order `spikes` first by `group_keys` and then by `score` - i.e to have a concatenated array of spikes of ascending score per group (so per bin per segment).

- Our final goal is to take the first `k_per_bin` spikes of each groups. Which is then equivalent to do a unifrom random pick per temporal bin. To do so we then compute the relative `ranks` of each spike inside their own group based the previously computed `order` - we have then to compute each group starting point and length, respectivly `group_start` and `counts`). `ranks` would look like something like this `[0,1,2,3,4,0,0,0,1,2,0,1,0,1,2,...]`

- We then apply a mask to `ranks` to select the top `k_per_bin` for each group - i.e for each bin per segment. And then we apply the resulting mask to the ordered `spikes` vector.


I m not sure if my explainations are very clear, that's why i prefer to write it here in a first time to have some feed back on it.

tayheau:
By the way @samuelgarcia, @alejoe91 , perhaps we can add the `temporal_bins` explanation to the doc and just reference to it in the docstring - so that we have a better layout and so readability. What do you think ?

alejoe91:
> By the way @samuelgarcia, @alejoe91 , perhaps we can add the `temporal_bins` explanation to the doc and just reference to it in the docstring - so that we have a better layout and so readability. What do you think ?

Yes it definitely deserve some more explanation! :)

alejoe91:
I don't think we have a specific place yet, but maybe it's worth adding a `Core Analyzer Extensions` section here?

https://spikeinterface.readthedocs.io/en/latest/modules/core.html#sortinganalyzer

src/spikeinterface/widgets/utils.py
```diff
@@ -410,11 +410,17 @@ def get_segment_durations(sorting: BaseSorting, segment_indices: list[int]) -> l
     sorting : BaseSorting
         The sorting object containing spike data
```
chrishalcrow:
```suggestion
```

I am a master reviewer.

src/spikeinterface/core/sorting_tools.py
```diff
@@ -148,12 +150,16 @@ def vector_to_list_of_spiketrain_numba(sample_indices, unit_indices, num_units):
...
     sorting: BaseSorting,
-    num_samples: int | None = None,
-    method: str = "uniform",
+    num_samples: list[int] | None = None,
+    method: Literal["uniform", "all", "percentage", "maximum_rate", "temporal_bins"] = "uniform",
     max_spikes_per_unit: int = 500,
+    percentage: float | None = None,
+    maximum_rate: float | None = None,
+    bin_size_s: float | None = None,
+    k_per_bin: int | None = None,
```
chrishalcrow: Can you propogate these to the `_set_params` in `ComputeRandomSpikes`. At the moment, you can't use the new methods when you do e.g. `analyzer.compute("random_spikes", method="percentage")` cause it can't accept `percentage` etc.

https://github.com/SpikeInterface/spikeinterface/blob/81e89fd19cffdbf031e48c54100b6ab9e024e3ca/src/spikeinterface/core/analyzer_extension_core.py#L67

chrishalcrow: I think `k_per_bin` should have a more explicit name. Something tedious like `temporal_bins_max_spikes_per_bin`

src/spikeinterface/core/sorting_tools.py
chrishalcrow, line 166: I agree that a short (one sentence) description of each method would be good here, just to give a developer looking at the code (hello!) an idea of expected behavior.

src/spikeinterface/core/sorting_tools.py
```diff
@@ -211,19 +237,70 @@ def random_spikes_selection(
...
+
+            elif method == "maximum_rate":
+                if maximum_rate is None:
+                    raise ValueError(f"maximum_rate must be defined")
+
+                t_duration = np.sum(get_segment_durations(sorting))
+                rng_size = min(int(t_duration * maximum_rate), max_spikes_per_unit, all_unit_indices.size)
+                selected_unit_indices = rng.choice(all_unit_indices, size=rng_size, replace=False, shuffle=False)
+
+            elif method == "temporal_bins":
```
chrishalcrow: I think something is wrong with the algorithm. When I run it on real data I get two spikes per second interval:

``` python
analyzer.compute('random_spikes', method="temporal_bins", bin_size_s=1, k_per_bin=1)
random_spikes = analyzer.get_extension('random_spikes')
unit_5_spikes = random_spikes.get_random_spikes()['unit_index'] == 5
print((random_spikes.get_random_spikes()[unit_5_spikes]['sample_index']//30_000)[0:10])
```
```
>>> array([0, 0, 1, 1, 2, 2, 3, 3, 4, 4])
```

I think with these parameters I'm expecting
```
>>> array([0,1,2,3,..
```

chrishalcrow: Oh, when I do `k_per_bin = 2`, I get three spikes per bin. So there's an unwanted `+1` somewhere...

chrishalcrow: This type of thing would make a good test!

tayheau: I think a `np.unique()` should solve this. With a step back, the `temporal_bins` might be overcomplex and maybe it should be a simpler version, since it's not very important for this to be fully optimized to my mind

chrishalcrow: Yeah the code is a bit hard to understand... Make it simpler then we can optimize later if needs be.

PR #4279, opened 2025-12-19
Improvements for internal sorters

yger:
Various enhancements for the internal sorters: 
- no more predefined margin
- better flexibility of preprocessing options for lupin and tdc2
- cast input data as float32 if not preprocessed
- make the final analyzer in all cases lupin, tdc2, sc2, with extension noise_levels and amplitude_scalings
- in case of sc2, make the analyzer survives the final exit with a cached recording (either json or pickeled), such that it can be fully reloadable

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -431,7 +431,7 @@ Drift
 ~~~~~
 
 .. automodule:: spikeinterface.generation
-    :noindex:
+    :no-index:
 
     .. autofunction:: generate_drifting_recording
     .. autofunction:: generate_displacement_vector
@@ -446,7 +446,7 @@ Hybrid
 ~~~~~~
 
 .. automodule:: spikeinterface.generation
-    :noindex:
+    :no-index:
 
     .. autofunction:: generate_hybrid_recording
     .. autofunction:: estimate_templates_from_recording
@@ -463,6 +463,7 @@ Noise
 ~~~~~
 
 .. automodule:: spikeinterface.generation
+    :no-index:
 
     .. autofunction:: generate_noise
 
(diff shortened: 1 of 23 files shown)


src/spikeinterface/sorters/internal/lupin.py
```diff
@@ -128,36 +131,36 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
...
-                ftype="bessel",
-                filter_order=2,
-                margin_ms=20.0,
-                dtype="float32",
-            )
+                rec_for_motion = bandpass_filter(
+                    rec_for_motion, freq_min=300.0, freq_max=6000.0, ftype="bessel", dtype="float32"
+                )
+                if apply_cmr:
+                    rec_for_motion = common_reference(rec_for_motion)
```
alejoe91: Here we should apply the `preprocessing_dict` if given!

alejoe91: No?

src/spikeinterface/sortingcomponents/tools.py
```diff
@@ -509,12 +510,30 @@ def create_sorting_analyzer_with_existing_templates(
...
+    if amplitude_scalings is not None:
+        from spikeinterface.postprocessing.amplitude_scalings import ComputeAmplitudeScalings
+
+        sa.extensions["amplitude_scalings"] = ComputeAmplitudeScalings(sa)
+        sa.extensions["amplitude_scalings"].params = dict(
+            sparsity=None,
+            max_dense_channels=16,
+            ms_before=templates.ms_before,
+            ms_after=templates.ms_after,
+            handle_collisions=True,
```
alejoe91: when this is True, the `extension.data` should also have a `collision_mask`, so I think it should be set to False here
```suggestion
            handle_collisions=False,
```

samuelgarcia: Ah yes your are right.
The sorters handle collision but we not have mask...
So lets switch to False but this is strange.

PR #4280, opened 2025-12-22
fix extra is None in iterative_template

### LINKED ISSUE #4235: getting extras from estimate_motion has a bug
Hello,

estimate_motion() has an option extra_outputs which currently (in my version 103.0) doesn't work, at least with the iterative template method. The specific problem comes from two things: in estimate motion lines 104-107

    if extra_outputs:
        extra = {}
    else:
        extra = None

and then in iterative_template lines 133 to 139:

        if extra:
            extra["non_rigid_windows"] = non_rigid_windows
            extra["motion_histograms"] = motion_histograms
            extra["target_histogram"] = target_histogram
            extra["shift_covs_block"] = shift_covs_block
            extra["temporal_hist_bin_edges"] = temporal_hist_bin_edges
            extra["spatial_hist_bin_edges"] = spatial_hist_bin_edges

both options for extra in the former code block return false, so the latter code block is inaccessible

I have bypassed this for now by doing: 

    if extra_outputs:
        extra = 1
    else:
        extra = None

But perhaps this breaks other things...

Thanks!

Jeff Boucher

samuelgarcia:
Fix #4235

### DIFF
--- a/src/spikeinterface/sortingcomponents/motion/iterative_template.py
+++ b/src/spikeinterface/sortingcomponents/motion/iterative_template.py
@@ -130,7 +130,7 @@ class IterativeTemplateRegistration:
         # convert to um
         motion_array = -(shift_indices * bin_um)
 
-        if extra:
+        if extra is not None:
             extra["non_rigid_windows"] = non_rigid_windows
             extra["motion_histograms"] = motion_histograms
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Ig no one wants to congratulate me on this, I merge it myself.

PR #4282, opened 2025-12-23
using `spike_vector` on `core/check_sortings_equal` test

tayheau:
Following this [todo](https://github.com/SpikeInterface/spikeinterface/blob/d7554f677874d7511602de4946a76f33f58791d3/src/spikeinterface/core/testing.py#L115C27-L115C42)

### DIFF
--- a/src/spikeinterface/core/testing.py
+++ b/src/spikeinterface/core/testing.py
@@ -149,11 +149,12 @@ def check_extractor_properties_equal(EX1, EX2) -> None:
 
 def _slice_spikes(spikes, start_frame=None, end_frame=None):
     sample_indices = spikes["sample_index"]
-    if len(sample_indices) == 0: return spikes[:0]
+    if len(sample_indices) == 0:
+        return spikes[:0]
     if start_frame is None:
         start_frame = sample_indices[0]
     if end_frame is None:
         end_frame = sample_indices[-1] + 1
-    start_idx, end_idx = np.searchsorted(sample_indices, [start_frame, end_frame + 1], side='left')
+    start_idx, end_idx = np.searchsorted(sample_indices, [start_frame, end_frame + 1], side="left")
 
-    return spikes[start_idx: end_idx]
+    return spikes[start_idx:end_idx]


src/spikeinterface/core/testing.py
```diff
@@ -113,29 +113,33 @@ def check_sortings_equal(
...
-                SX1.get_unit_spike_train(id, segment_index=segment_idx, start_frame=30, end_frame=max_spike_index - 30)
-            )
-            train2 = np.sort(
-                SX2.get_unit_spike_train(id, segment_index=segment_idx, start_frame=30, end_frame=max_spike_index - 30)
-            )
-            assert np.array_equal(train1, train2)
+    def _sorted_spike_vector(SX):
+        spikes = SX.to_spike_vector()
+        order = np.lexsort((spikes["sample_index"], spikes["unit_index"], spikes["segment_index"]))
+        return spikes[order]
```
alejoe91: I don't think we need this, since `to_spike_vector` should already return spikes sorted by segment_index + sample_index. Or did you add this to add the sorting by unit_index?

tayheau: Pure inattention from me ! I remove it.

samuelgarcia: oups. this was needed put it back!!

alejoe91:
@tayheau I think we actually needed the unit_index lex sort after all! The reason is that before we were testing spiketrain-by-spiketrain, so the order didn't matter if we had synchronous spikes. Now with the spike vector representation we need to ensure unit_index is sorted too! Sorry about that! :)

src/spikeinterface/core/testing.py
tayheau: vacations will do me good XD

alejoe91: After this PR #4315 the spike vector is lexsorted with ("unit_index", "sample_index", "segment_index"), so we can finally drop this

src/spikeinterface/core/testing.py
```diff
@@ -112,30 +112,33 @@ def check_sortings_equal(
...
```
alejoe91:
```suggestion
```

src/spikeinterface/core/testing.py
```diff
@@ -112,30 +112,33 @@ def check_sortings_equal(
...
+    def _slice_spikes(spikes, start_frame=None, end_frame=None):
+        mask = np.ones(spikes.size, dtype=bool)
+        if start_frame is not None:
+            mask &= spikes["sample_index"] >= start_frame
+        if end_frame is not None:
+            mask &= spikes["sample_index"] <= end_frame
+        return spikes[mask]
+
+    s1 = _sorted_spike_vector(SX1)
+    s2 = _sorted_spike_vector(SX2)
```
alejoe91:
```suggestion
    s1 = SX1.to_spike_vector()
    s2 = SX2.to_spike_vector()
```

alejoe91 (approved):

src/spikeinterface/core/basesorting.py
```diff
@@ -783,7 +779,9 @@ def _compute_and_cache_spike_vector(self) -> None:
             spikes_in_seg["unit_index"] = unit_indices
             spikes_in_seg["segment_index"] = segment_index
             spikes.append(spikes_in_seg)
+
         spikes = np.concatenate(spikes)
+        spikes = spikes[np.lexsort((spikes["unit_index"], spikes["sample_index"], spikes["segment_index"]))]
```
alejoe91, line 784: @samuelgarcia I think it's better to lexsort the whole array at the end to make sure that units are also sorted within samples

src/spikeinterface/core/generate.py
```diff
@@ -174,7 +174,7 @@ def generate_sorting(
             spikes.append(spikes_on_borders)
 
     spikes = np.concatenate(spikes)
-    spikes = spikes[np.lexsort((spikes["sample_index"], spikes["segment_index"]))]
+    spikes = spikes[np.lexsort((spikes["unit_index"], spikes["sample_index"], spikes["segment_index"]))]
```
alejoe91, line 177: we were missing the unit index lexsort in generate

src/spikeinterface/core/zarrextractors.py
```diff
@@ -295,6 +295,8 @@ def __init__(self, folder_path: Path | str, storage_options: dict | None = None,
         spikes["unit_index"] = spikes_group["unit_index"][:]
         for i, (start, end) in enumerate(segment_slices_list):
             spikes["segment_index"][start:end] = i
+        spikes = spikes[np.lexsort((spikes["unit_index"], spikes["sample_index"], spikes["segment_index"]))]
+        self._cached_spike_vector = spikes
```
alejoe91, line 299: and in zarr

src/spikeinterface/core/testing.py
```diff
@@ -155,3 +145,12 @@ def check_extractor_properties_equal(EX1, EX2) -> None:
...
     for property_name in EX1.get_property_keys():
         assert_array_equal(EX1.get_property(property_name), EX2.get_property(property_name))
+
+
+def _slice_spikes(spikes, start_frame=None, end_frame=None):
+    mask = np.ones(spikes.size, dtype=bool)
+    if start_frame is not None:
+        mask &= spikes["sample_index"] >= start_frame
+    if end_frame is not None:
+        mask &= spikes["sample_index"] <= end_frame
```
samuelgarcia: searchsorted should be faster no ?

src/spikeinterface/core/zarrextractors.py
samuelgarcia: Doing this in the init is an enormous cost in term of perfs.
I am not sure this is a good idea to lexsort at each reading.
lets discuss

src/spikeinterface/core/testing.py
tayheau: yep indeed

alejoe91:
thanks @tayheau 

Merging!

PR #4283, opened 2025-12-25
Fix edges of 2d-histogram in UnitWaveformDensityMapWidget

ecobost:
Currently, `sw.plot_unit_waveforms_density_map(analyzer, unit_ids=[0], use_max_channel=True)` produces this:
density_map_pre
The density map has some spurious high values on the very first row (more visible near the peaks) and also on the bottom although less visible in this example.

The issue comes from the clipping in https://github.com/SpikeInterface/spikeinterface/blob/5a58beb7e17864b5c905a2f03ddd1966f0da8e65/src/spikeinterface/widgets/unit_waveforms_density_map.py#L129
that counts any value with amplitude less than bin_min as belonging to the first bin and any value above bin_max as belonging to the last bin of the histogram, thus producing those lines on the plot.

This PR uses equal-sized bins from bin_min to bin_max and uses np.histogram to count occurrences. It produces:
density_map_post
without any artifacts. (Also tested it with multiple channels and multiple units)

Granted, it looks uglier but it's more correct. And it will look better for nicer data, this is a 2-minute recording just for testing.

### DIFF
--- a/src/spikeinterface/widgets/unit_waveforms_density_map.py
+++ b/src/spikeinterface/widgets/unit_waveforms_density_map.py
@@ -184,7 +184,7 @@ class UnitWaveformDensityMapWidget(BaseWidget):
                 origin="lower",
                 aspect="auto",
                 extent=(0, x_max, dp.bin_min, dp.bin_max),
-                cmap="hot",
+                cmap="Grays",
             )
         else:
             for ax, unit_id in zip(self.axes.flatten(), dp.unit_ids):
@@ -196,7 +196,7 @@ class UnitWaveformDensityMapWidget(BaseWidget):
                     origin="lower",
                     aspect="auto",
                     extent=(0, x_max, dp.bin_min, dp.bin_max),
-                    cmap="hot",
+                    cmap="Grays",
                 )
 
         for unit_index, unit_id in enumerate(dp.unit_ids):
@@ -220,7 +220,7 @@ class UnitWaveformDensityMapWidget(BaseWidget):
                 channel_id = dp.channel_ids[chan_ind]
                 x = (i + 0.5) * dp.template_width / freq_khz
                 y = (dp.bin_max + dp.bin_min) / 2.0
-                ax.text(x, y, f"chan_id {channel_id}", color="w", ha="center", va="center")
+                ax.text(x, y, f"{channel_id}", color="k", ha="center", va="center")
 
             ax.set_xlabel("Time [ms]")
             ax.set_ylabel(f"unit_id {unit_id}")


yger (approved):
Not sure I like the idea of having a value hardcoded in the widget (num_bins=100), but it used to already be like that before, so we can leave it as it is. Thanks a lot for the patch! @samuelgarcia mayeb worth propagating in spikeinterface-gui also?

samuelgarcia:
Hi.
I am OK for moving from bin size to num bins for params.

About the histogram vs "devide and round" did you benchmark the speed ?

ecobost:
Hi @samuelgarcia. Just did some tests, newer solution becomes faster at around 1000-2000 spikes (both are under 20ms at that point anyway. For larger values numpy beats it.
Screenshot from 2026-01-09 03-11-14

ecobost:
Added a couple of commits, one is just code formatting (https://github.com/SpikeInterface/spikeinterface/pull/4283/commits/c30f5879ceab3cfff75175119470d970f38b5416). 

The other (https://github.com/SpikeInterface/spikeinterface/pull/4283/commits/87fdaa78596ebfd5df69ea5a2d02b49562c7e26f) changes the x axis of the plot to show absolute time (msecs). I think it will be useful because no other plot shows the waveforms/templates in an absolute clock (plot_waveforms/templates has unitless x axis, it does have a scale bar but it only works under particular circumstances). It is nice to have the x axis as msec if you wanna estimate how wide a spike is or the distance between two peaks in the template or so on. I tested it with all params (1 vs many channels, and same_axis=True/False) and it works fine.

Here's what the plots look like:
density_map_one
density_map_allchannels
density_map_summary

Note: When all channels are plotted, rather than creating a different axes per channel, the concatenated image is plotted in a single axis and some vertical white lines are plotted to separate the channels. So the x axis still goes on real time but it is not separate per plot, not ideal but I think good enough.

ecobost:
Another possible improvement is to use a less aggresive colormap (currently hot). I did not change it yet because it seems pretty subjective so I would rather ask for comments first. Here are some candidates:
magma:
density_map_magma
copper:
density_map_copper
cubehelix:
density_map_cubehelix
grays:
density_map_grays

I particularly like Grays because it looks standard, avoids color clashing with unit colors and looks nicer in the UnitSummary (see below) but any would work
density_map_summary_white

Let me know which one you guys like and I'm happy to send a last commit with that change.

samuelgarcia:
Really cool.
I will read this in detail.

src/spikeinterface/widgets/unit_waveforms_density_map.py
```diff
@@ -121,14 +121,9 @@ def __init__(
...
-            hist2d = np.zeros((wfs_flat.shape[1], bins.size))
-            indexes0 = np.arange(wfs_flat.shape[1])
-
-            wf_bined = np.floor((wfs_flat - bin_min) / bin_size).astype("int32")
-            wf_bined = wf_bined.clip(0, bins.size - 1)
-            for d in wf_bined:
-                hist2d[indexes0, d] += 1
+            wfs_flat = wfs.swapaxes(1, 2).reshape(wfs.shape[0], -1)  # num_spikes x (num_channels * timepoints)
+            hists_per_timepoint = [np.histogram(one_timepoint, bins=bins)[0] for one_timepoint in wfs_flat.T]
+            hist2d = np.stack(hists_per_timepoint)
```
samuelgarcia, line 126: Is it faster or slower ?
I am not sure to remember because I did this long time ago but in my mind this uggly trick "histogram like" was for performence. But maybe I am wrong.
lets me check it.

ecobost: Hi @samuelgarcia performance happens to be slightly better for sufficiently large units, >1-2K spikes (see my comment above) but feel free to double check

samuelgarcia: Ok thanks.

samuelgarcia:
Merci beaucoup

Issue #4175, opened 2025-10-16
Double check highpass_spatial_filter implementation with `ibl-neuropixel`


alejoe91:
@oliche

oliche:
Yes there seems to be an issue with the SI implementation as the quality is degraded:

Expected from ibl-neuropixel:
<img width="2043" height="840" alt="Image" src="https://github.com/user-attachments/assets/7c2dc7e9-942e-419d-bee4-d350043c7bf4" />

Actual from spike-interface
<img width="2043" height="840" alt="Image" src="https://github.com/user-attachments/assets/783483bb-39ef-4499-93f5-4ddb0ac227d3" />


Let me dig into SI code to find the discrepancy source (missing cast to float ? AGC ? )

JoeZiminski:
Thanks for looking into this @oliche, it would be good to hear what you find. The two implementations are directly compared [here](https://github.com/SpikeInterface/spikeinterface/blob/9ec48d7e950d08dfab831705903627e50274a516/src/spikeinterface/preprocessing/tests/test_highpass_spatial_filter.py#L31) in case this is useful, there must be some case not covered by the test data.

oliche:
Here is a minimal example to reproduce:
```

# uv pip install s3fs
# sudo apt-get install libwavpack-dev
#  uv pip install wavpack-numcodecs
import spikeinterface.full as si
import spikeinterface.preprocessing as spre
s3_path = "s3://aind-open-data/ecephys_813047_2025-10-20_16-35-28/ecephys/ecephys_compressed/experiment1_Record Node 101#Neuropix-PXI-100.ProbeA-AP.zarr"
recording = si.load(s3_path)
recroding = spre.astype(recording, "float")
recording_ps = si.phase_shift(recording)
recording_hp = si.highpass_filter(recording_ps)
recording_cmr = si.common_reference(recording_hp)
recording_hps = si.highpass_spatial_filter(recording_hp)
rec_dict = dict(hp=recording_hp, cmr=recording_cmr, hps=recording_hps)

# %% uv pip install ibl-neuropixel
import neuropixel
import numpy as np
import ibldsp.voltage
fs = recording.sampling_frequency
s0, s1 = int(1723.5 * fs), int(1724 * fs)
raw = recording.get_traces(start_frame=s0, end_frame=s1).astype(np.float32).T * neuropixel.S2V_AP
[...]

PR #4286, opened 2025-12-31
Fix bug spatial filter #4175

oliche:
Fixed the main bug: all of the processing has to be done in float, even if only to reconvert to int at the end of the module.

I have also improved a few things:

- filters had mutable arguments, switched lists to tuple
- tests were always ignored as `neurodsp`  module doesn't exist anymore
- epsilon value was absolute in agc, this is valid in IBL as we always work in Volts. It will now generalize as it is relative to the rms. I have also updated the IBL version with the same change
- In the tests in debug mode, I use viewephys to compare the traces. It turns out that the edge effects in time are larger in the spike interface version than the IBL version, this is due to the filter implementation.

### DIFF
--- a/src/spikeinterface/preprocessing/highpass_spatial_filter.py
+++ b/src/spikeinterface/preprocessing/highpass_spatial_filter.py
@@ -235,7 +235,7 @@ class HighPassSpatialFilterSegment(BasePreprocessorSegment):
         # apply AGC and keep the gains
         traces = traces.astype(np.float32)
         if self.window is not None:
-            traces, agc_gains = agc(traces, window=self.window, epsilon=self.epsilon_values_for_agc)
+            traces, agc_gains = agc(traces, window=self.window, epsilons=self.epsilon_values_for_agc)
         else:
             agc_gains = None
         # pad the array with a mirrored version of itself and apply a cosine taper


src/spikeinterface/preprocessing/tests/test_highpass_spatial_filter.py
```diff
@@ -3,7 +3,7 @@
 import numpy as np
 from copy import deepcopy
 
-import spikeinterface as si
+import spikeinterface.full as si
```
zm711: Why switch from core to full. This is super heavy and probably not necessary right? This is a bit of a poorly explained part of our importing.
```python
import spikeinterface as si
```
is equivalent to
```python
```

which is a faster import.

src/spikeinterface/extractors/cbin_ibl.py
```diff
@@ -49,6 +49,8 @@ def __init__(
     ):
         from neo.rawio.spikeglxrawio import read_meta_file
 
+        if Path(folder_path).is_file():
+            folder_path = Path(folder_path).parent
```
alejoe91: Why do we need this? :) 

One can either pass the root folder or a direct path the the cbin file with the `cbin_file_path` arg

src/spikeinterface/preprocessing/highpass_spatial_filter.py
```diff
@@ -268,13 +269,15 @@ def agc(traces, window, epsilon=1e-8):
     """
     import scipy.signal
 
-    gain = scipy.signal.fftconvolve(np.abs(traces), window[:, None], mode="same", axes=0)
+    # default value for epsilon is relative to the rms, loosely matching the IBL 1e-8 for an input in Volts
+    if epsilon is None:
+        epsilon = np.std(traces - np.mean(traces)) * 0.003
```
alejoe91: maybe the `0.003` should also be an argument? `epsilon_factor`?

src/spikeinterface/preprocessing/tests/test_highpass_spatial_filter.py
```diff
@@ -51,32 +51,28 @@ def test_highpass_spatial_filter_real_data(lagc):
...
-    options = dict(lagc=lagc, ntr_pad=25, ntr_tap=50, butter_kwargs=None)
-    print(options)
-
-    ibl_data, si_recording = get_ibl_si_data()
-
-    si_filtered, _ = run_si_highpass_filter(si_recording, **options)
+    local_path = si.download_dataset(remote_path="spikeglx/Noise4Sam_g0")
+    si_recording = se.read_spikeglx(local_path, stream_id="imec0.ap")
+    si_recording = spre.astype(si_recording, "float")
+    recording_ps = si.phase_shift(si_recording)
```
alejoe91:
```suggestion
    recording_ps = spre.phase_shift(si_recording)
```

src/spikeinterface/preprocessing/tests/test_highpass_spatial_filter.py
```diff
@@ -51,32 +51,28 @@ def test_highpass_spatial_filter_real_data(lagc):
...
+    recording_hp = si.highpass_filter(recording_ps, freq_min=300, filter_order=3)
```
alejoe91:
```suggestion
    recording_hp = spre.highpass_filter(recording_ps, freq_min=300, filter_order=3)
```

src/spikeinterface/preprocessing/tests/test_highpass_spatial_filter.py
```diff
@@ -51,32 +51,28 @@ def test_highpass_spatial_filter_real_data(lagc):
...
+    recording_hps = si.highpass_spatial_filter(recording_hp)
```
alejoe91:
```suggestion
    recording_hps = spre.highpass_spatial_filter(recording_hp)
```

src/spikeinterface/preprocessing/tests/test_highpass_spatial_filter.py
alejoe91: following @zm711 comment about not importing full!

alejoe91 (changes requested):
Thanks @oliche!!! 

Just a few minor comments

src/spikeinterface/preprocessing/highpass_spatial_filter.py
```diff
@@ -255,7 +256,7 @@ def get_traces(self, start_frame, end_frame, channel_indices):
 # -----------------------------------------------------------------------------------------------
 
 
-def agc(traces, window, epsilon=1e-8):
+def agc(traces, window, epsilon=None):
```
samuelgarcia: I think this is a problem here because the espilon will change for every chunk  no ?

samuelgarcia: I would vote to force espilon here because it is an helper function and compute it the init of the class.
no ?

src/spikeinterface/extractors/cbin_ibl.py
oliche: Yes true, didn't see it. I also suggest to enact the `cbin_file` deprecation

src/spikeinterface/preprocessing/highpass_spatial_filter.py
oliche: I kind of disagree, here this is really the last resort where we need to use something for epsilon, and I find it confusing to have both an absolute and relative epsilon in the signature

src/spikeinterface/preprocessing/tests/test_highpass_spatial_filter.py
oliche: agree

src/spikeinterface/preprocessing/highpass_spatial_filter.py
oliche: Yes but this is a second order issue, we just need to stabilize the inverse of the gain.

To mitigate, I have exposed the `epsilon` as a parameter in the interface. It would be useful to have a `BaseRecording.estimate_rms()` method to fix a single epsilon for the full processing, but this is beyond the scope of this PR.

 If everyone is on-board for this, I can create an issue and summarize this discussion

alejoe91: Sounds good to me. We hav an `get_noise_levels(recording, method="rms")` that we could run in the init if rms levels are not estimated yet. But let's discuss specifics on the new issue

alejoe91 (approved):

src/spikeinterface/extractors/cbin_ibl.py
```diff
@@ -124,8 +115,7 @@ def __init__(
                 num_channels_per_adc = 16
             else:  # NP1.0
                 num_channels_per_adc = 12
-
-            sample_shifts = get_neuropixels_sample_shifts(self.get_num_channels(), num_channels_per_adc)
+            sample_shifts = get_neuropixels_sample_shifts_from_probe(probe, num_channels_per_adc)
```
alejoe91, line 118: fixed this, which now needs to be `probe` directly!

oliche: good with me, I also fixed the tests so the CI passes.

samuelgarcia (approved):

samuelgarcia:
ok for me also.

PR #4287, opened 2025-12-31
Add `discard_spikes` to curation, and update to `v3`

### LINKED ISSUE #4261: Add `discard_spikes`/`clean_spikes` step in a curation v3
Right now, the curation has:
- `removed`: units to remove
- `merges`: unit groups to merge
- `splits`: unit to split into 2 or multiple sub-units

One another very common use case is to "clean" units. For example, if you have a nice and clean unit with a very nasty noisy event it's easy to remove (most) of those noisy spikes. Currently, to achieve this, one would need a 2 step curation: the first one to perform the split, the second to remove the "noisy" splitted unit.

This is to propose another "step" in the curation model (and make a version 3) which does both of the steps in one go! 
It should be very easy on the implementation side: the "discard_spikes" could be run as the first step, and internally perform splits whose units to remove would just be added to the `removed` units.

chrishalcrow:
Would close #4261

Docs: https://spikeinterface--4287.org.readthedocs.build/en/4287/modules/curation.html

(Note: PR looks big but mostly tests+docs - main changes are updating the id strategy in `sorting_tools.py` (~100 lines) and updating the `apply_merges` function (~50 lines))

This PR allows you to remove spikes from a unit during curation by specifying the following in your curation json file:

```
"discard_spikes": [
    {
        "unit_id": "u10",
        "indices": [
            56,
            57,
            59,
            60
        ]
    },
    {
        "unit_id": "u14",
        "indices": [
            123,
            321
        ]
    }
]
```

This new feature means that the curation format gets bumped to v3!

You can discard at the same time as merging and splitting.

Tricky bit 1
-----------
How to apply it to an analyzer. Decided to discard spikes at the same time as splitting. During the splitting step, we re-wrangle discard spikes into another split unit (call them "discard units") and keep track of the discard unit id. Then remove the full discard units after the splitting. This allows us to use the existing splitting machinery (including splitting extension etc) for discards - nice! 

Tricky bit 2
-----------
Much of the complexity is now related to the id strategy (ughh - why do we give the user THREE new id strategies!!!). It's difficult because we want a cleaned unit to retains its id. This is true even if there are other units which are just split, and the user is using the "append" or "split" new_unit_id_strategy.

So suppose the user does the following:

Unit 1: clean
Unit 2: split
Unit 3: clean and split

With strategy append, we want

Unit 1 -> Unit 1 + Unit 1 dirty 
Unit 2 -> Unit 4 + Unit 5
Unit 3 -> Unit 6 + Unit 7 + Unit 3 dirty

We do this by slotting in the dirty units into places we know to split later. For "append" strategy, I chose to put "Unit 1 dirty" to the last possible unit id, and unit 3 dirty to "unit 3". 

Other stuff
-----------

We have to do merges after splitting+discarding. This is because the spike indices change after merging. To avoid wrangling spike indices (gross!) we just do discarding first.

Tests to do:
- test curation format
- test v2 -> v3 update
- test actual discarding when also splitting / merging

### DIFF
--- a/src/spikeinterface/core/sorting_tools.py
+++ b/src/spikeinterface/core/sorting_tools.py
@@ -517,6 +517,7 @@ def apply_splits_to_sorting(
     sorting: BaseSorting,
     unit_splits: dict[int | str, list[list[int | str]]],
     new_unit_ids: list[list[int | str]] | None = None,
+    discard_spikes_unit_ids: list[list[int | str]] | None = None,
     return_extra: bool = False,
     new_id_strategy: str = "append",
 ):
@@ -543,6 +544,8 @@ def apply_splits_to_sorting(
         List of new unit_ids for each split. If given, it needs to have the same length as `unit_splits`.
         and each element must have the same length as the corresponding list of split indices.
         If None, new ids will be generated.
+    discard_spikes_unit_ids : list | None, default: None
+        List of units which contain spikes to discard.
     return_extra : bool, default: False
         If True, also return the new_unit_ids.
     new_id_strategy : "append" | "split", default: "append"
@@ -565,7 +568,11 @@ def apply_splits_to_sorting(
     # this is true when running via apply_curation
 
     new_unit_ids = generate_unit_ids_for_split(
-        sorting.unit_ids, unit_splits, new_unit_ids=new_unit_ids, new_id_strategy=new_id_strategy
+        sorting.unit_ids,
+        unit_splits,
+        new_unit_ids=new_unit_ids,
+        new_id_strategy=new_id_strategy,
+        discard_spikes_unit_ids=discard_spikes_unit_ids,
     )
     all_unit_ids = _get_ids_after_splitting(sorting.unit_ids, unit_splits, new_unit_ids)
     all_unit_ids = list(all_unit_ids)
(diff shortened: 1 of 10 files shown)


alejoe91:
Awesome Chris! Quick comment before diving into the PR:

> So suppose the user does the following:
>
> Unit 1: clean
> Unit 2: split
> Unit 3: clean and split

I don't think that clean and split should be allowed. We have rules so that a unit can either be removed, merged, split, and I would add discard. Does it simplify the new unit id logic?

src/spikeinterface/curation/curation_format.py
```diff
@@ -218,7 +218,85 @@ def apply_curation(
...
+                return_new_unit_ids=True,
+            )
+
+        if len(discard_spikes_unit_ids) > 0:
+            ids_to_remove = []
+            for new_id_set in new_ids:
+                if new_id_set[0] in discard_spikes_unit_ids or new_id_set[1] in discard_spikes_unit_ids:
+                    ids_to_remove.append(new_id_set[0])
+
+            curated_sorting_or_analyzer = curated_sorting_or_analyzer.remove_units(ids_to_remove)
```
alejoe91, line 297: Nice :)

src/spikeinterface/core/sorting_tools.py
```diff
@@ -691,46 +702,92 @@ def generate_unit_ids_for_split(old_unit_ids, unit_splits, new_unit_ids=None, ne
...
+        for split_indices in unit_splits.values():
+            highest_possible_unit_id += len(split_indices)
+
+    for unit_to_split, split_indices in unit_splits.items():
+
+        num_splits = len(split_indices)
+
+        # decide if unit is a simple discard, a simple split or a discard and split
+        just_discard = False
+        discard_and_split = False
```
alejoe91, line 749: Yeah I think this should not be allowed. A unit can be either split or cleaned, not both! This should simplify the logic a lot!

PR #4289, opened 2026-01-05
More possibilities for gt generation

yger:
Currently, one can not provide custom parameters to the drifting_gt generation. This PR fixes such behavior, such that users can not provide their own sorting, NoiseGenerator, Template kwargs, ... in order to reproduce more closely one behavior or refine benchmarks

### DIFF
--- a/src/spikeinterface/generation/drifting_generator.py
+++ b/src/spikeinterface/generation/drifting_generator.py
@@ -435,8 +435,9 @@ def generate_drifting_recording(
         num_units = sorting.get_num_units()
         sampling_frequency = sorting.sampling_frequency
         if sorting._recording is not None:
-            assert sorting.get_total_duration() == duration, "Sorting should have the same duration as the generated data"
-            
+            assert (
+                sorting.get_total_duration() == duration
+            ), "Sorting should have the same duration as the generated data"
 
     # probe
     if probe is None:


src/spikeinterface/generation/drifting_generator.py
```diff
@@ -406,12 +421,31 @@ def generate_drifting_recording(
...
+            durations=[
+                duration,
+            ],
+            **generate_sorting_kwargs,
+            seed=seed,
+        )
+    else:
+        num_units = sorting.get_num_units()
+        sampling_frequency = sorting.sampling_frequency
+        if sorting._recording is not None:
```
samuelgarcia, line 437: This is very strange. Why would the sorting have an attached recording, we want to generate it.

src/spikeinterface/generation/drifting_generator.py
```diff
@@ -420,20 +454,34 @@ def generate_drifting_recording(
...
+            unit_displacements,
+            displacement_vectors,
+            displacement_unit_factor,
+            displacement_sampling_frequency,
+            displacements_steps,
+        ) = generate_displacement_vector(
+            duration, unit_locations[:, :2], seed=seed, **generate_displacement_vector_kwargs
+        )
+    else:
+        (
```
samuelgarcia, line 480: I would make a dict no ? Or at least a named tuple.

src/spikeinterface/generation/drifting_generator.py
```diff
@@ -493,13 +531,14 @@ def generate_drifting_recording(
...
-        **generate_noise_kwargs,
-    )
+    if noise is None:
+        noise = generate_noise(
+            probe=probe,
+            sampling_frequency=sampling_frequency,
+            durations=[duration],
+            seed=seed,
+            **generate_noise_kwargs,
+        )
```
samuelgarcia, line 543: else:
we need to check the duration

src/spikeinterface/generation/drifting_generator.py
yger: Yes, but you can have a first generation, obtain a sorting, and then you might want to take the duration of this first data. But agreed, this might be not needed

src/spikeinterface/generation/drifting_generator.py
yger: I don't know, this was to be consistent, but I can make it a dict if you prefer

src/spikeinterface/generation/drifting_generator.py
yger: right, I'll do that

PR #4298, opened 2026-01-06
Sequential curation

### LINKED ISSUE #4296: Extend `CurationModel` to `SequentialCuration`
We can extend the `apply_curation` to a list of curation steps:

```
class SequentialCuration(BaseModel):
    curation_steps: List[CurationModel]
```
+ some validators!!!

Then the `apply_curation` will be as follows:

```
def apply_curation(analyzer, curation_or_sequential_curation):
    if isinstance(curation_or_sequential_curation, CurationModel):
        return apply_one_step_curation(analyzer)
    elif isinstance(curation_or_sequential_curation, SequentialCuration):
        analyzer_current = analyzer
        for curation_step in curation_or_sequential_curation.curation_step:
            analyzer_current = apply_one_step_curation(analyzer_current)
        return analyzer_current
```

The `SecuentialCuration` will also have a method to generate an `(n_spikes, n_steps)` matrix representation of the curation steps.

alejoe91:
Fixes #4296 

Extends curation to a `SequentialCuration`, which contains a list of `CurationModel` steps that are applied in sequence

### DIFF
--- a/src/spikeinterface/curation/curation_model.py
+++ b/src/spikeinterface/curation/curation_model.py
@@ -4,6 +4,7 @@ from itertools import chain, combinations
 import numpy as np
 
 from spikeinterface import BaseSorting
+from spikeinterface.core.sorting_tools import _get_ids_after_merging, _get_ids_after_splitting
 
 
 class LabelDefinition(BaseModel):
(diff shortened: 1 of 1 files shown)


samuelgarcia:
And so I lost the debate...

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -604,8 +604,9 @@ def _split_extension_data(self, split_units, new_unit_ids, new_sorting_analyzer,
                             arr = np.std(wfs, axis=0)
                         elif operator == "median":
                             arr = np.median(wfs, axis=0)
-                        elif "percentile" in operator:
-                            _, percentile = operator.splot("_")
```
chrishalcrow, line 609: <3

src/spikeinterface/metrics/template/template_metrics.py
```diff
@@ -244,7 +244,7 @@ def _prepare_data(self, sorting_analyzer, unit_ids):
             # templates_multi is a list of 2D arrays of shape (n_times, n_channels)
             tmp_data["templates_multi"] = templates_multi
             tmp_data["channel_locations_multi"] = channel_locations_multi
-            tmp_data["depth_direction"] = self.params["depth_direction"]
+            tmp_data["depth_direction"] = self.params.get("depth_direction")
```
chrishalcrow: This was giving a bug when loading analyzers from old versions of `si`

samuelgarcia: lets pur a comment in the code the why of the get()

chrishalcrow: Was for backwards compatibility - now moved a fix to the `_handle_backward_compatibility_on_load` function, so it makes more sense why it happens.

src/spikeinterface/curation/curation_model.py
```diff
@@ -366,6 +366,47 @@ def convert_old_format(cls, values):
...
+                )
+            ids_to_remove = ids_to_remove + merge.unit_ids
+            ids_to_add.append(merge.new_unit_id)
+
+        if ids_to_add is None:
+            ids_to_add = set()
+        if ids_to_remove is None:
+            ids_to_remove = set()
+
+        final_ids = set(intial_ids).difference(ids_to_remove).union(ids_to_add)
```
samuelgarcia: I guess that the set will break somehow the ordering of unit_ids no ?
I would implent this with list to keep the original order as much as possible.

Also something important stuff to decide :
 * does splied units stay inserted are are push to the end
 * same for merged units

src/spikeinterface/metrics/template/template_metrics.py
```diff
@@ -111,19 +111,48 @@ def _handle_backward_compatibility_on_load(self):
             self.params["metric_names"].remove("num_positive_peaks")
             if "number_of_peaks" not in self.params["metric_names"]:
                 self.params["metric_names"].append("number_of_peaks")
+                self.params["metric_params"]["number_of_peaks"] = self.params["metric_params"]["num_positive_peaks"]
```
samuelgarcia: why do we need to handle this in the PR ?
does it depends on other PR ?

src/spikeinterface/curation/curation_model.py
```diff
@@ -366,6 +366,47 @@ def convert_old_format(cls, values):
...
+        Returns the final unit ids of the `curation_model`, when new unit ids are
+        given for each curation choice. Raises an error if new unit ids are missing
+        for any curation choice.
+
+        Returns
+        -------
+        final_ids : list
+            The ids of the sorting/analyzer after curation takes place
+        """
+
```
samuelgarcia: Could we use this always for the final unit_ids to ensure some kind of deterministic unit id ordering ?

src/spikeinterface/metrics/template/template_metrics.py
alejoe91: I think that @chrishalcrow was solving a  bunch of other issues that came up while testing. We can move these out to #4304

src/spikeinterface/curation/curation_model.py
alejoe91: Do we care about the ordering of unit ids? 

Anyways, I think that this matches what we normally do:
- drop units that are note used anymore
- add new units to the end

No?

src/spikeinterface/curation/curation_model.py
alejoe91: where?

src/spikeinterface/curation/curation_model.py
chrishalcrow: I'm ok with keeping the original order - I'll change this!
What happens to the split and merged units depends on the split/merge strategy: the user can decide.

chrishalcrow: Oh wait, sorry. The order of the new ids depends on tons of other decision and strategies by the user. This function is only used to check that the initial and final ids are consistent, which is ok to do with sets. Maybe I can rename the functions...

Issue #4294, opened 2026-01-06
Implement `valid_period` extension


alejoe91:
We need an extension to annotate units with valid periods. 
The extension data is quite simple:
- a list of [`start_frame`, `end_frame`] for each unit
- list of list if multi-segment

alejoe91:
@samuelgarcia suggests a structured array with:

(`segment_index`, `unit_index`, `start_frame`, `end_frame`)

which is great because it is a numpy array and it's automatically handled by the BaseExtensions

PR #4299, opened 2026-01-06
Valid unit periods extension

m-beau:
Implements #4294

- [x] Add widget

### DIFF
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -766,10 +766,6 @@ class BaseSorting(BaseExtractor):
             if len(sample_indices) > 0:
                 sample_indices = np.concatenate(sample_indices, dtype="int64")
                 unit_indices = np.concatenate(unit_indices, dtype="int64")
-                order = np.argsort(sample_indices)
-                sample_indices = sample_indices[order]
-                unit_indices = unit_indices[order]
-
                 n = sample_indices.size
                 segment_slices[segment_index, 0] = seg_pos
                 segment_slices[segment_index, 1] = seg_pos + n
@@ -783,7 +779,9 @@ class BaseSorting(BaseExtractor):
             spikes_in_seg["unit_index"] = unit_indices
             spikes_in_seg["segment_index"] = segment_index
             spikes.append(spikes_in_seg)
+
         spikes = np.concatenate(spikes)
+        spikes = spikes[np.lexsort((spikes["unit_index"], spikes["sample_index"], spikes["segment_index"]))]
 
         self._cached_spike_vector = spikes
         self._cached_spike_vector_segment_slices = segment_slices
--- a/src/spikeinterface/core/generate.py
+++ b/src/spikeinterface/core/generate.py
@@ -174,7 +174,7 @@ def generate_sorting(
             spikes.append(spikes_on_borders)
 
     spikes = np.concatenate(spikes)
-    spikes = spikes[np.lexsort((spikes["sample_index"], spikes["segment_index"]))]
+    spikes = spikes[np.lexsort((spikes["unit_index"], spikes["sample_index"], spikes["segment_index"]))]
 
     sorting = NumpySorting(spikes, sampling_frequency, unit_ids)
 
--- a/src/spikeinterface/core/testing.py
+++ b/src/spikeinterface/core/testing.py
@@ -112,30 +112,20 @@ def check_sortings_equal(
 
     max_spike_index = SX1.to_spike_vector()["sample_index"].max()
 
-    # TODO for later  use to_spike_vector() to do this without looping
-    for segment_idx in range(SX1.get_num_segments()):
-        # get_unit_ids
-        ids1 = np.sort(np.array(SX1.get_unit_ids()))
-        ids2 = np.sort(np.array(SX2.get_unit_ids()))
-        assert_array_equal(ids1, ids2)
-        for id in ids1:
-            train1 = np.sort(SX1.get_unit_spike_train(id, segment_index=segment_idx))
-            train2 = np.sort(SX2.get_unit_spike_train(id, segment_index=segment_idx))
-            assert np.array_equal(train1, train2)
-            train1 = np.sort(SX1.get_unit_spike_train(id, segment_index=segment_idx, start_frame=30))
-            train2 = np.sort(SX2.get_unit_spike_train(id, segment_index=segment_idx, start_frame=30))
-            assert np.array_equal(train1, train2)
-            # test that slicing works correctly
-            train1 = np.sort(SX1.get_unit_spike_train(id, segment_index=segment_idx, end_frame=max_spike_index - 30))
-            train2 = np.sort(SX2.get_unit_spike_train(id, segment_index=segment_idx, end_frame=max_spike_index - 30))
-            assert np.array_equal(train1, train2)
-            train1 = np.sort(
-                SX1.get_unit_spike_train(id, segment_index=segment_idx, start_frame=30, end_frame=max_spike_index - 30)
-            )
-            train2 = np.sort(
-                SX2.get_unit_spike_train(id, segment_index=segment_idx, start_frame=30, end_frame=max_spike_index - 30)
-            )
-            assert np.array_equal(train1, train2)
+    s1 = SX1.to_spike_vector()
+    s2 = SX2.to_spike_vector()
+    assert_array_equal(s1, s2)
+
+    for start_frame, end_frame in [
+        (None, None),
+        (30, None),
+        (None, max_spike_index - 30),
+        (30, max_spike_index - 30),
+    ]:
+
+        slice1 = _slice_spikes(s1, start_frame, end_frame)
+        slice2 = _slice_spikes(s2, start_frame, end_frame)
+        assert np.array_equal(slice1, slice2)
 
     if check_annotations:
         check_extractor_annotations_equal(SX1, SX2)
@@ -155,3 +145,16 @@ def check_extractor_properties_equal(EX1, EX2) -> None:
 
     for property_name in EX1.get_property_keys():
         assert_array_equal(EX1.get_property(property_name), EX2.get_property(property_name))
+
+
+def _slice_spikes(spikes, start_frame=None, end_frame=None):
+    sample_indices = spikes["sample_index"]
+    if len(sample_indices) == 0:
+        return spikes[:0]
+    if start_frame is None:
+        start_frame = sample_indices[0]
+    if end_frame is None:
+        end_frame = sample_indices[-1] + 1
+    start_idx, end_idx = np.searchsorted(sample_indices, [start_frame, end_frame + 1], side="left")
+
+    return spikes[start_idx:end_idx]
--- a/src/spikeinterface/core/zarrextractors.py
+++ b/src/spikeinterface/core/zarrextractors.py
@@ -295,6 +295,8 @@ class ZarrSortingExtractor(BaseSorting):
         spikes["unit_index"] = spikes_group["unit_index"][:]
         for i, (start, end) in enumerate(segment_slices_list):
             spikes["segment_index"][start:end] = i
+        spikes = spikes[np.lexsort((spikes["unit_index"], spikes["sample_index"], spikes["segment_index"]))]
+        self._cached_spike_vector = spikes
 
         for segment_index in range(num_segments):
             soring_segment = SpikeVectorSortingSegment(spikes, segment_index, unit_ids)


alejoe91:
Depends on #4302

alejoe91:
@m-beau @samuelgarcia the parallelization makes it extremely fast. I slightly refactored the misc_metrics so I can directly call the low-level functions. Now computing good periods for a full 1 hour NP probe takes a minute or so :)

The cool trick is that we can compute all fp/fn values for all chunks (a chunk is unit+segment+period) and filter it all at once with thresholds. This simplifies the logic a lot! Take a look, for me this is ready to review.

I also tested the relative period mode and it works very well!

@m-beau I changed some defaults and naming to something that I think is a bit less strict, but not too much

src/spikeinterface/postprocessing/good_periods_per_unit.py
```diff
@@ -0,0 +1,546 @@
...
+from spikeinterface.metrics.spiketrain import compute_firing_rates
+
+numba_spec = importlib.util.find_spec("numba")
+if numba_spec is not None:
+    HAVE_NUMBA = True
+else:
+    HAVE_NUMBA = False
+
+
+class ComputeGoodPeriodsPerUnit(AnalyzerExtension):
```
samuelgarcia: I pretty bad for english but my first react is "good" could be replace by "valid" for a clearer semantic no ?

src/spikeinterface/postprocessing/good_periods_per_unit.py
```diff
@@ -0,0 +1,546 @@
...
+                        user_defined_periods[:, 1:3],
+                    )
+                )
+            # Cast user defined periods to unit_period_dtype
+            user_defined_periods_typed = np.zeros(user_defined_periods.shape[0], dtype=unit_period_dtype)
+            user_defined_periods_typed["unit_index"] = user_defined_periods[:, 0]
+            user_defined_periods_typed["segment_index"] = user_defined_periods[:, 1]
+            user_defined_periods_typed["start_sample_index"] = user_defined_periods[:, 2]
+            user_defined_periods_typed["end_sample_index"] = user_defined_periods[:, 3]
+            user_defined_periods = user_defined_periods_typed
```
samuelgarcia: I guess that 
```suggestion
            user_defined_periods_typed = np.frombuffer(user_defined_periods, dtype=unit_period_dtype)
```
could works because it is only int64 (but less explcit)

src/spikeinterface/postprocessing/good_periods_per_unit.py
```diff
@@ -0,0 +1,546 @@
...
+        params = dict(
+            method=method,
+            period_duration_s_absolute=period_duration_s_absolute,
+            period_target_num_spikes=period_target_num_spikes,
+            period_mode=period_mode,
+            fp_threshold=fp_threshold,
+            fn_threshold=fn_threshold,
+            minimum_n_spikes=minimum_n_spikes,
+            minimum_valid_period_duration=minimum_valid_period_duration,
+            user_defined_periods=user_defined_periods,
```
samuelgarcia: @alejoe91 
This could lead to a big json file no ?

src/spikeinterface/postprocessing/good_periods_per_unit.py
```diff
@@ -0,0 +1,546 @@
...
+        return params
+
+    def _select_extension_data(self, unit_ids):
+        new_extension_data = self.data
+        return new_extension_data
+
+    def _merge_extension_data(
+        self, merge_unit_groups, new_unit_ids, new_sorting_analyzer, censor_ms=None, verbose=False, **job_kwargs
+    ):
+        new_extension_data = self.data
```
samuelgarcia: I guess for this we should decide the rules of merging valid period.
intersection or union or period pre merging ?

samuelgarcia: Or remove.

src/spikeinterface/postprocessing/good_periods_per_unit.py
```diff
@@ -0,0 +1,546 @@
...
+        return new_extension_data
+
+    def _split_extension_data(self, split_units, new_unit_ids, new_sorting_analyzer, verbose=False, **job_kwargs):
+        new_extension_data = self.data
+        return new_extension_data
```
samuelgarcia: same here what do we want for period after a split ?
maybe remove period of this units no ?

src/spikeinterface/postprocessing/good_periods_per_unit.py
```diff
@@ -0,0 +1,546 @@
...
+        -------
+        numpy.ndarray | list
+            The periods in numpy or dictionary by unit format, depending on `outputs`.
+            If "numpy", returns an array of dtype unit_period_dtype with columns:
+            unit_index, segment_index, start_sample_index, end_sample_index.
+            If "by_unit", returns a list (per segment) of dictionaries mapping unit IDs to lists of
+            (start_sample_index, end_sample_index) tuples.
+        """
+        if outputs == "numpy":
+            good_periods = self.data["good_periods_per_unit"]
```
samuelgarcia: maybe a copy for safety no ?

src/spikeinterface/postprocessing/good_periods_per_unit.py
```diff
@@ -0,0 +1,546 @@
...
+                )
+                periods_for_unit = np.zeros(len(starts_ends), dtype=unit_period_dtype)
+                for i, (start, end) in enumerate(starts_ends):
+                    subperiod = np.zeros((1,), dtype=unit_period_dtype)
+                    subperiod["segment_index"] = segment_index
+                    subperiod["start_sample_index"] = start
+                    subperiod["end_sample_index"] = end
+                    subperiod["unit_index"] = unit_index
+                    periods_for_unit[i] = subperiod
+                all_subperiods = np.concatenate((all_subperiods, periods_for_unit), axis=0)
```
samuelgarcia: I think this is better to make a final concatenate at the end of the loop and magaing this with list.
This avoid too much memory allocation

samuelgarcia:
This looks cool.

Could we make a very simple matplotlib widgets for this new extension ? (in addtion to sigui f course)
having matplot is good for tutorial.

src/spikeinterface/postprocessing/good_periods_per_unit.py
alejoe91: That was my suggestion too! I like it better as well :)

alejoe91: That was my suggestion too! I like it better as well :)

src/spikeinterface/postprocessing/good_periods_per_unit.py
alejoe91: Very good point! We still have to add tests for the external periods, but what we could do is set it as a `self.user_defined_period` and pop in from the dict in self params, so they are not set in the JSON but the run function can still use it.

src/spikeinterface/postprocessing/good_periods_per_unit.py
alejoe91: This is still a TODO

src/spikeinterface/postprocessing/good_periods_per_unit.py
alejoe91: same here. Probably we want to recompute?

alejoe91:
Yes, I'll add some TODO's to the PR description so we keep track of it.

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -332,16 +344,21 @@ def compute_refrac_period_violations(
         return None
 
     sorting = sorting_analyzer.sorting
-    fs = sorting_analyzer.sampling_frequency
-    num_units = len(sorting_analyzer.unit_ids)
-    num_segments = sorting_analyzer.get_num_segments()
+    sorting = sorting.select_periods(periods=periods)
+
+    # TODO: in case of periods, should we use total samples only in provided periods?
+    total_samples = sorting_analyzer.get_total_samples()
```
alejoe91: @m-beau @samuelgarcia @chrishalcrow this is a critical point. I refactor the computation so that we can pass the total samples externally. This is needed both when computing in chunks for the `compute_good_periods` function and also when we update the metrics to use the valid periods. In that case, we will need to count the valid samples for each units, so total_samples can be a dictionary now

src/spikeinterface/postprocessing/good_periods_per_unit.py
alejoe91: @samuelgarcia can you take a look?

src/spikeinterface/postprocessing/good_periods_per_unit.py
alejoe91: done

src/spikeinterface/postprocessing/good_periods_per_unit.py
alejoe91: done in the last commit

src/spikeinterface/postprocessing/good_periods_per_unit.py
alejoe91: done

src/spikeinterface/postprocessing/good_periods_per_unit.py
alejoe91: done

src/spikeinterface/metrics/quality/misc_metrics.py
chrishalcrow: Hello - do we need to be careful about stuff like this? Here the `total_duration` of the entire recording. It's used later to compute a firing rate, and this firing rate is the total num spikes divided by total duration, rather than num-spike-in-periods/duration-of-periods.

alejoe91: In this case is the num spikes in period. So I think this is correct, but I'll double check carefully!

alejoe91: In this case is the num spikes in period. So I think this is correct, but I'll double check carefully!

src/spikeinterface/postprocessing/good_periods_per_unit.py
m-beau: I suggested this during the hackathon too, valid is clearer! :)

src/spikeinterface/postprocessing/good_periods_per_unit.py
m-beau: After any merge or split, I think the whole thing needs to be recomputed, as merges could lead to new refractory period violations in specific periods, and splits to new clipped amplitude distributions in other periods?

alejoe91: yep, this is currently the approach: recompute for any new merge/split and dropped the original merged units and splits!

m-beau: awesome, I love that extensions have a built in merge and split logic that allows to only recompute metrics for units that were affected by the splits/merges!

src/spikeinterface/metrics/quality/misc_metrics.py
chrishalcrow: I don't understand. In this case, it should be num spike in the period? Then I think the code is incorrect: Num spikes are computed using `count_num_spikes_per_unit`, which is computed using the full recording time.

alejoe91: Ah it's the CV metrics! you are right, I'll take a closer look and push some fixes

src/spikeinterface/postprocessing/tests/common_extension_tests.py
```diff
@@ -95,18 +99,31 @@ def get_sorting_analyzer(self, recording, sorting, format="memory", sparsity=Non
 
         return sorting_analyzer
 
-    def _prepare_sorting_analyzer(self, format, sparse, extension_class):
+    def _compute_extensions_recursively(self, sorting_analyzer, extension_class, params):
```
chrishalcrow, line 104: Oh boy, this is a handy lil function!

src/spikeinterface/postprocessing/valid_unit_periods.py
```diff
@@ -0,0 +1,756 @@
...
+    worker_ctx["max_threads_per_worker"] = max_threads_per_worker
+
+
+def fp_fn_worker_func(period, sorting, all_amplitudes_by_unit, params):
+    """
+    Low level computation of false positives and false negatives for one period and one unit.
+    """
+    from spikeinterface.metrics.quality.misc_metrics import (
+        amplitude_cutoff,
+        _compute_nb_violations_numba,
```
chrishalcrow, line 750: You need to check if you have numba installed before doing this import

src/spikeinterface/postprocessing/valid_unit_periods.py
```diff
@@ -0,0 +1,756 @@
...
+    mask = (spiketrain >= start_sample_index) & (spiketrain < end_sample_index)
+    total_samples_in_period = end_sample_index - start_sample_index
+    spiketrain_period = spiketrain[mask]
+    amplitudes_period = amplitudes_unit[mask]
+
+    # compute fp (rp_violations). See _compute_refrac_period_violations in quality metrics
+    fs = sorting.sampling_frequency
+    t_c = int(round(params["censored_period_ms"] * fs * 1e-3))
+    t_r = int(round(params["refractory_period_ms"] * fs * 1e-3))
+    n_v = _compute_nb_violations_numba(spiketrain_period, t_r)
```
chrishalcrow, line 774: and need to split this so there's a numba/non-numba option

src/spikeinterface/widgets/unit_valid_periods.py
```diff
@@ -0,0 +1,233 @@
...
+    unit_ids : list | None, default: None
+        List of unit ids to plot. If None, all units are plotted.
+    show_only_units_with_valid_periods : bool, default: True
+        If True, only units with valid periods are shown.
+    """
+
+    def __init__(
+        self,
+        sorting_analyzer: SortingAnalyzer | None = None,
+        segment_index: int | None = None,
```
chrishalcrow, line 31: If None, can we take the 0th segment index please?

chrishalcrow: Or have ipywidgets select the 0th segment first if it's none.

src/spikeinterface/widgets/unit_valid_periods.py
```diff
@@ -0,0 +1,233 @@
...
+            for valid_period in valid_period_for_units:
+                start_time = valid_period["start_sample_index"] / sorting_analyzer.sampling_frequency
+                end_time = valid_period["end_sample_index"] / sorting_analyzer.sampling_frequency
+                axs[2].axvspan(start_time, end_time, alpha=0.3, color="g")
+
+            axs[0].set_xlabel("")
+            axs[1].set_xlabel("")
+            axs[2].set_xlabel("Time (s)")
+            axs[0].set_ylabel("FP Rate")
+            axs[1].set_ylabel("FN Rate")
```
chrishalcrow: Is it possible to make these more descriptive? As a user I forget which one is related to amplitudes and which one is related to rp violations. But there isn't much room on the axis. Hmm.

src/spikeinterface/postprocessing/valid_unit_periods.py
```diff
@@ -0,0 +1,789 @@
...
+                n_subperiods = min_num_periods_relative  # at least min_num_periods_relative subperiods
+                period_size_samples = n_samples // n_subperiods
+                margin_size_samples = int(relative_margin_size * period_size_samples)
+
+            # we generate periods starting from 0 up to n_samples, with and without margins, and period centers
+            starts = np.arange(0, n_samples, period_size_samples)
+            periods_for_unit = np.zeros(len(starts), dtype=unit_period_dtype)
+            periods_for_unit_w_margins = np.zeros(len(starts), dtype=unit_period_dtype)
+            for i, start in enumerate(starts):
+                end = start + period_size_samples
```
chrishalcrow:
```suggestion
                end = min(start + period_size_samples, n_samples)
```
At the moment, `end` can go past the recording length.

src/spikeinterface/postprocessing/valid_unit_periods.py
```diff
@@ -0,0 +1,789 @@
...
+                raise ImportError("Numba is required to compute RP violations (false positives).")
+
+        # subperiods
+        assert period_mode in ("absolute", "relative"), f"Invalid subperiod_size_mode: {period_mode}"
+        assert (
+            period_duration_s_absolute > 0 or period_target_num_spikes > 0
+        ), "Either period_duration_s_absolute or period_target_num_spikes must be positive."
+        assert isinstance(period_target_num_spikes, (int)), "period_target_num_spikes must be an integer."
+
+        # user_defined_periods formatting
```
chrishalcrow, line 148: I think in this section we're being too lenient with the user and it's a bit of a mess. I think you should reuse the logic from core (https://github.com/SpikeInterface/spikeinterface/blob/679cb2ed975d6ec821c839fb5e91933fdd23836a/src/spikeinterface/core/sorting_tools.py#L267) and be a bit stricter on what we demand.

src/spikeinterface/postprocessing/valid_unit_periods.py
```diff
@@ -0,0 +1,789 @@
...
+
+numba_spec = importlib.util.find_spec("numba")
+if numba_spec is not None:
+    HAVE_NUMBA = True
+else:
+    HAVE_NUMBA = False
+
+
+class ComputeValidUnitPeriods(AnalyzerExtension):
+    """Compute good time periods per unit based on quality metrics.
```
chrishalcrow: This docstring needs a tidy up

src/spikeinterface/widgets/unit_valid_periods.py
alejoe91: done in last commit

src/spikeinterface/widgets/unit_valid_periods.py
```diff
@@ -0,0 +1,233 @@
...
+    show_only_units_with_valid_periods : bool, default: True
+        If True, only units with valid periods are shown.
+    """
+
+    def __init__(
+        self,
+        sorting_analyzer: SortingAnalyzer | None = None,
+        segment_index: int | None = None,
+        unit_ids: list | None = None,
+        show_only_units_with_valid_periods: bool = True,
```
chrishalcrow: I would default this to False - I wanna see why my bad unit are failing!

src/spikeinterface/postprocessing/valid_unit_periods.py
```diff
@@ -0,0 +1,789 @@
...
+    def _merge_extension_data(
+        self, merge_unit_groups, new_unit_ids, new_sorting_analyzer, censor_ms=None, verbose=False, **job_kwargs
+    ):
+        new_extension_data = {}
+        good_periods = self.data["valid_unit_periods"]
+        if self.params["method"] in ("false_positives_and_negatives", "combined"):
+            # need to recompute for merged units
+            recompute = True
+        else:
+            # in case of user-defined periods, just merge periods
```
chrishalcrow, line 228: I would have taken the intersection of the user periods, instead of merging them?

alejoe91: why? If you merge, you probably want to keep the units you had before. Anyways, I think this user-defined period won't be really used much in "production", but the idea is to be able to test new ideas

src/spikeinterface/postprocessing/valid_unit_periods.py
alejoe91: there is not a non-numba option, so I check at init ;)

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1165,6 +1169,7 @@ def _save_or_select_or_merge_or_split(
                     recompute_dict[extension_name] = extension.params
 
         if len(recompute_dict) > 0:
+            recompute_dict = _sort_extensions_by_dependency(recompute_dict)
```
```suggestion
```
`compute_several_extensions` handles this

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1136,7 +1136,11 @@ def _save_or_select_or_merge_or_split(
             raise ValueError(f"SortingAnalyzer.save: unsupported format: {format}")
 
         recompute_dict = {}
-        for extension_name, extension in self.extensions.items():
+        extensions_to_compute = _sort_extensions_by_dependency(
+            {ext.name: ext.params for ext in self.extensions.values()}
```
```suggestion
            {ext.extension_name: ext.params for ext in self.extensions.values()}
```

src/spikeinterface/postprocessing/valid_unit_periods.py
```diff
@@ -0,0 +1,768 @@
...
+        all_period_centers = []
+        for segment_index in range(num_segments):
+            period_centers = {}
+            periods_segment = all_periods[all_periods["segment_index"] == segment_index]
+            for unit_id in unit_ids:
+                unit_index = self.sorting_analyzer.sorting.id_to_index(unit_id)
+                periods_unit = periods_segment[periods_segment["unit_index"] == unit_index]
+                period_samples = (periods_unit["start_sample_index"] + periods_unit["end_sample_index"]) // 2
+                # period_samples are the same for all bins (per unit), so we can just take the first one
+                period_centers[unit_id] = periods_unit["start_sample_index"] + period_samples[0]
```
chrishalcrow, line 522: EDIT: no, I don't think so... 

```suggestion
                period_centers[unit_index] = periods_unit["start_sample_index"] + period_samples[0]
```
I'm not 100% sure, but think this should be `unit_index`?

src/spikeinterface/postprocessing/valid_unit_periods.py
```diff
@@ -0,0 +1,816 @@
...
+    HAVE_NUMBA = False
+
+
+class ComputeValidUnitPeriods(AnalyzerExtension):
+    """Compute valid unit periods for units.
+    By default, the extension uses the "false_positives_and_negatives" method, which computes amplitude cutoffs
+    (false negative rate) and refractory period violations (false positive rate) over chunks of data
+    to estimate valid periods. External user-defined periods can also be provided.
+
+    Paraneters
```
chrishalcrow:
```suggestion
    Parameters
```

src/spikeinterface/postprocessing/valid_unit_periods.py
```diff
@@ -0,0 +1,816 @@
...
+    ----------
+    method : {"false_positives_and_negatives", "user_defined", "combined"}
```
chrishalcrow:
```suggestion
    method : "false_positives_and_negatives" | "user_defined" | "combined", default: "false_positives_and_negatives"
```

+ several other parameters don't use our conventions. Please fix

src/spikeinterface/postprocessing/valid_unit_periods.py
```diff
@@ -0,0 +1,816 @@
...
+        Controls the smoothing applied to the amplitude histogram.
+    amplitudes_bins_min_ratio : int, default: 5
+        The minimum ratio between number of amplitudes for a unit and the number of bins.
+        If the ratio is less than this threshold, the amplitude_cutoff for the unit is set
+        to NaN.
+
+
+    Notes
+    -----
+    Implementation by Maxime Beau and Alessio Buccino, inspired by NeuroPyxels and Bombcell.
```
chrishalcrow: Let's add proper bombcell and NeuroPyxels citations

src/spikeinterface/postprocessing/valid_unit_periods.py
chrishalcrow: If you had a unit which is bad from 0-30s and one which is good from 0-30s, I would guess the combined unit would be bad from 0-30. But I'm that worried about this.

samuelgarcia: agree with Chris

src/spikeinterface/postprocessing/valid_unit_periods.py
```diff
@@ -0,0 +1,816 @@
...
+        new_extension_data = {}
+        good_periods = self.data["valid_unit_periods"]
+        unit_indices = self.sorting_analyzer.sorting.ids_to_indices(unit_ids)
+        mask = np.isin(good_periods["unit_index"], unit_indices)
+        new_extension_data["valid_unit_periods"] = good_periods[mask]
+        return new_extension_data
+
+    def _merge_extension_data(
+        self, merge_unit_groups, new_unit_ids, new_sorting_analyzer, censor_ms=None, verbose=False, **job_kwargs
+    ):
```
samuelgarcia, line 212: You should add here a comment for the strategy of merging period so that the code will be easier to read.

src/spikeinterface/postprocessing/valid_unit_periods.py
```diff
@@ -0,0 +1,816 @@
...
+    segment_index = period_sample["segment_index"]
+    start_sample_index = period_sample["start_sample_index"]
+    end_sample_index = period_sample["end_sample_index"]
+    unit_index = period_sample["unit_index"]
+    unit_id = sorting.unit_ids[unit_index]
+
+    amplitudes_unit = all_amplitudes_by_unit[segment_index][unit_id]
+    spiketrain = sorting.get_unit_spike_train(unit_id, segment_index=segment_index)
+
+    mask = (spiketrain >= start_sample_index) & (spiketrain < end_sample_index)
```
samuelgarcia: seacrsorted no ?

src/spikeinterface/postprocessing/valid_unit_periods.py
```diff
@@ -0,0 +1,816 @@
...
+            all_amplitudes_by_unit = amp_scalings.get_data(outputs="by_unit", concatenated=False)
+
+            init_args = (sorting_analyzer.sorting, all_amplitudes_by_unit, self.params, max_threads_per_worker)
+
+            # Each item is one computation of fp and fn for one period and one unit
+            items = [(period,) for period in all_periods_w_margins]
+            job_name = f"computing false positives and negatives"
+
+            # parallel
+            with ProcessPoolExecutor(
```
samuelgarcia, line 402: We need to be carrfull with th use of ProcessPoolExecutor.
on window this imply spawn and so datta duplication here this will be sorting + amplitudes duplicated
in case of a spwan

alejoe91: then we should make a base executor. We have the same issue in other extensions...

alejoe91: let's do it in a separate PR

src/spikeinterface/postprocessing/valid_unit_periods.py
alejoe91: done in last commit

src/spikeinterface/postprocessing/valid_unit_periods.py
alejoe91: done

src/spikeinterface/postprocessing/valid_unit_periods.py
alejoe91: I added a comment in the user defined section. I think it's ok since no-one will ever check the docstring for this

src/spikeinterface/postprocessing/valid_unit_periods.py
alejoe91: Not sure I agree. Say you have a unit that is split because of drift and you merge. Unit 1 is good in the first half of the recording, unit 2 in the second half. IMO, the merged unit should be valid over the entire period. No?

alejoe91:
Revised, @samuelgarcia @chrishalcrow can you take a final look?

We can discuss on the merge strategy with user defined periods, but I think union is a good choice (see my comment)

src/spikeinterface/postprocessing/valid_unit_periods.py
```diff
@@ -0,0 +1,802 @@
...
+                user_defined_periods = np.asarray(user_defined_periods)
+            except Exception as e:
+                raise ValueError(
+                    (
+                        "user_defined_periods must be some (n_periods, 3) [unit, good_period_start, good_period_end] "
+                        "or (n_periods, 4) [unit, segment_index, good_period_start, good_period_end] structure convertible to a numpy array"
+                    )
+                )
+
+            user_defined_periods = cast_periods_to_unit_period_dtype(user_defined_periods)
```
chrishalcrow, line 161: niiiice

src/spikeinterface/postprocessing/valid_unit_periods.py
samuelgarcia: lets discuss a bit more this no ?

samuelgarcia:
Impressive and complicated PR!
Congratulations.
This is OK on my side.

I push some test for the remapping low level code.

src/spikeinterface/core/sorting_tools.py
```diff
@@ -904,3 +907,60 @@ def _get_ids_after_splitting(old_unit_ids, split_units, new_unit_ids):
...
     return np.array(all_unit_ids, dtype=dtype)
+
+
+def remap_unit_indices_in_vector(vector, all_old_unit_ids, all_new_unit_ids, keep_old_unit_ids=None):
+    """
+    Remap the "unit_index" field in a spike vector (or period vector) according to new unit ids.
+
+    This is usefull for instance when:
+      * select unit and recompute quickly the "unit_index" in the spike vector
+      * merging/spliting periods or spikes and update the "unit_index" in the vector
```
alejoe91:
```suggestion
    This is useful for instance when you:
      * select unit and recompute quickly the "unit_index" in the spike vector
      * merging/spliting periods or spikes and update the "unit_index" in the vector
```

doc/modules/postprocessing.rst
```diff
@@ -257,15 +259,35 @@ spike_amplitudes
...
+.. _postprocessing_amplitude_scalings:
+
+amplitude_scalings
+^^^^^^^^^^^^^^^^^^
+
+This extension computes the amplitude scaling of each spike as the value of the linear fit between the template and the
+spike waveform. In case of spatio-temporal collisions, a multi-linear fit is performed using the templates of all units
+involved in the collision.
+
+**NOTE:** computing amplitude scalings is highly recommended before calculating amplitude-based quality metrics, such as
```
ecobost, line 282: Why is this so?

alejoe91: Because they are much more robust. Spike amplitudes only sample the voltage value on the max channel, amplitude scalings use sparse templates and fit them spike-by-spike, also handling collisions. Downside is it's slower to compute

ecobost: @alejoe91 Thanks for the explanation. 

Currently, the documentation is unclear. Both spike_amplitudes and amplitude_scalings have the same note but users won't know whether they need to compute both or only one of them, which one is the recommended one (if any) or what happens if you compute both.
Screenshot from 2026-02-27 21-54-33

Currently, spike_amplitudes is the default everywhere: `amplitude_cutoff`, `noise cutoff` will use spike_amplitudes by default even if amplitude_scalings is available; `ampitude_cv` receives the extension as argument (default is spike_amplitudes) and `amplitude median` does not even consider amplitude_scalings. 

Maybe the note should just say, one of the two is required or just recommend one of them. I particularly prefer spike_amplitudes, even if noisier, it is more understandable and a lot less margin for error.

PR #4302, opened 2026-01-07
Implement `select_sorting_periods` in metrics

alejoe91:
Depends on #4316

Slices a sorting object based on an array ov valid periods. Periods are defined as a structured dtype as:
```
base_period_dtype = [
    ("start_sample_index", "int64"),
    ("end_sample_index", "int64"),
    ("segment_index", "int64"),
    ("unit_index", "int64"),
]
```

EDIT:

Refactored computation of spike train metrics, to make sure that periods are consistently taken into account. Added 2 utils functions to compute durations per unit and bin edges per unit, that optionally use the provided periods

### DIFF
--- a/src/spikeinterface/metrics/utils.py
+++ b/src/spikeinterface/metrics/utils.py
@@ -93,15 +93,12 @@ def compute_total_samples_per_unit(sorting_analyzer, periods=None):
         Total number of samples for each unit.
     """
     if periods is not None:
-        total_samples = {}
+        total_samples_array = np.zeros(len(sorting_analyzer.unit_ids), dtype="int64")
         sorting = sorting_analyzer.sorting
-        for unit_id in sorting.unit_ids:
-            unit_index = sorting.id_to_index(unit_id)
-            periods_unit = periods[periods["unit_index"] == unit_index]
-            num_samples_in_period = 0
-            for period in periods_unit:
-                num_samples_in_period += period["end_sample_index"] - period["start_sample_index"]
-            total_samples[unit_id] = num_samples_in_period
+        for period in periods:
+            unit_index = period["unit_index"]
+            total_samples_array[unit_index] += period["end_sample_index"] - period["start_sample_index"]
+        total_samples = dict(zip(sorting.unit_ids, total_samples_array))
     else:
         total = sorting_analyzer.get_total_samples()
         total_samples = {unit_id: total for unit_id in sorting_analyzer.unit_ids}


src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -1331,6 +1331,21 @@ class BaseSpikeVectorExtension(AnalyzerExtension):
     need_backward_compatibility_on_load = False
     nodepipeline_variables = []  # to be defined in subclass
 
+    def __init__(self, sorting_analyzer):
+        super().__init__(sorting_analyzer)
+        self._segment_slices = None
+
+    @property
+    def segment_slices(self):
```
samuelgarcia: I want to have this in NumpySorting since long time. We could move it there no ?

And also you can do a np.searchsorted for segment at once. without any loop.

samuelgarcia:
This is OK for me.
Make a clear documentation somwhere woule help ?

src/spikeinterface/core/sorting_tools.py
```diff
@@ -228,6 +228,83 @@ def random_spikes_selection(
...
+    return keep_mask
+
+
+def select_sorting_periods(sorting: BaseSorting, periods):
+    """
+    Returns a new sorting object, restricted to the given periods of dtype unit_period_dtype.
+
+    Parameters
+    ----------
+    S
```
chrishalcrow:
```suggestion
```

src/spikeinterface/core/sorting_tools.py
```diff
@@ -228,6 +228,83 @@ def random_spikes_selection(
...
+            global_indices = global_indices_segment[unit_id]
+            spiketrains = spike_vector[global_indices]["sample_index"]
+            if len(periods_for_unit) > 0:
+                for period in periods_for_unit:
+                    mask = (spiketrains >= period["start_sample_index"]) & (spiketrains < period["end_sample_index"])
+                    keep_mask[global_indices[mask]] = True
+    return keep_mask
+
+
+def select_sorting_periods(sorting: BaseSorting, periods):
```
chrishalcrow:
```suggestion
def select_sorting_periods(sorting: BaseSorting, periods) -> BaseSorting:
```

src/spikeinterface/core/tests/test_basesorting.py
```diff
@@ -226,7 +224,66 @@ def test_time_slice():
...
+    num_samples = int(sampling_frequency * duration)
+    num_units = 1000
+    sorting = generate_sorting(
+        durations=[duration, duration], sampling_frequency=sampling_frequency, num_units=num_units
+    )
+
+    rng = np.random.default_rng()
+
+    # number of random periods
+    n_periods = 10_000
```
chrishalcrow: duration, num_units and n_periods are all quite large for a test. Is it slow??

src/spikeinterface/core/analyzer_extension_core.py
alejoe91: I would make it in base sorting then, since it applies to all sorting objects (after to_spike_vector). No?

alejoe91: I would make it in base sorting then, since it applies to all sorting objects (after to_spike_vector). No?

src/spikeinterface/core/tests/test_basesorting.py
samuelgarcia: I agree we should put a small number

alejoe91: takes about 1 second

src/spikeinterface/core/node_pipeline.py
```diff
@@ -22,11 +22,20 @@
     ("segment_index", "int64"),
 ]
 
-
 spike_peak_dtype = base_peak_dtype + [
     ("unit_index", "int64"),
 ]
 
+base_period_dtype = [
```
alejoe91: move all dtypes to base

alejoe91: done in #4314

alejoe91:
@chrishalcrow I refactored a few metrics to make sure durations, spike counts, and bins are properly accounted for when slicing with periods. Happy to discuss about this!

src/spikeinterface/core/tests/test_basesorting.py
```diff
@@ -245,6 +242,64 @@ def test_time_slice():
...
+
+    t_start = time.perf_counter()
+    sliced_sorting = sorting.select_periods(periods=periods)
+    t_stop = time.perf_counter()
+    elapsed = t_stop - t_start
+    print(f"select_periods took {elapsed:.2f} seconds for {len(periods)} periods")
+
+    # Check that all spikes in the sliced sorting are within the periods
+    for segment_index in range(sorting.get_num_segments()):
+        periods_in_segment = periods[periods["segment_index"] == segment_index]
```
yger, line 284: do we assume not too many periods, just to be sure that all these masks won't take too long?

src/spikeinterface/core/sorting_tools.py
```diff
@@ -228,6 +228,83 @@ def random_spikes_selection(
...
+
+    Returns
+    -------
+    numpy.array
+        A boolean mask of the spikes in the sorting object, with True for spikes within the specified periods.
+    """
+    spike_vector = sorting.to_spike_vector()
+    spike_vector_list = sorting.to_spike_vector(concatenated=False)
+    keep_mask = np.zeros(len(spike_vector), dtype=bool)
+    all_global_indices = spike_vector_to_indices(spike_vector_list, unit_ids=sorting.unit_ids, absolute_index=True)
```
yger: why not using self.sorting_analyzer.sorting.get_spike_vector_to_indices(), and thus make use of a possible cache?

src/spikeinterface/core/sorting_tools.py
```diff
@@ -228,6 +228,83 @@ def random_spikes_selection(
...
+    for segment_index in range(sorting.get_num_segments()):
+        global_indices_segment = all_global_indices[segment_index]
+        # filter periods by segment
+        periods_in_segment = periods[periods["segment_index"] == segment_index]
```
yger, line 254: assuming not too many periods, these masks would be fine. Otherwise, we'll need to optimize

yger:
We don't want to extend these quality metrics to the extension() ? What if someone wants to only get the ISI, CCG or anything else only on the periods? Would it be easy to slice the sorting, and then compute only on the sub sorting? Are the extension robust w.r.t. periods?

alejoe91:
> We don't want to extend these quality metrics to the extension() ? What if someone wants to only get the ISI, CCG or anything else only on the periods? Would it be easy to slice the sorting, and then compute only on the sub sorting? Are the extension robust w.r.t. periods?

I think that extension by extension we could use `valid_unit_periods` when computed :)

alejoe91:
@samuelgarcia @chrishalcrow tests added! all good now :)

src/spikeinterface/metrics/utils.py
```diff
@@ -0,0 +1,152 @@
...
+    bin_duration_s : float, default: 1
+        Duration of each bin in seconds
+    periods : array of unit_period_dtype, default: None
+        Periods to consider for each unit
+    """
+    bin_edges_for_units = {}
+    num_segments = len(segment_samples)
+    bin_duration_samples = int(bin_duration_s * sorting.sampling_frequency)
+
+    if periods is not None:
```
chrishalcrow, line 36: When there are no good periods, then periods will be an empty array? Should we have a general strategy for when this is true - empty units often cause pain/`nan`s...

src/spikeinterface/metrics/spiketrain/metrics.py
```diff
@@ -12,24 +14,23 @@ def compute_num_spikes(sorting_analyzer, unit_ids=None, **kwargs):
...
+        Periods to consider for each unit.
 
     Returns
     -------
     num_spikes : dict
         The number of spikes, across all segments, for each unit ID.
     """
-
     sorting = sorting_analyzer.sorting
+    sorting = sorting.select_periods(periods)
```
chrishalcrow, line 26: When I choose periods so that only one unit is good, and get the num spikes, get

``` python
import numpy as np
from spikeinterface.core.base import unit_period_dtype

rec, sort = si.generate_ground_truth_recording(seed=1205)
periods = np.array((0,1000,4000,0), dtype=unit_period_dtype)
quality_metrics = si.compute_quality_metrics(sa, periods=periods, metric_names = ['num_spikes'])

print(quality_metrics['num_spikes'].values)
```

``` bash
>>> [162 173 135 154 158 144 151 141 153 162]
```

But this sorting only has spikes from unit index 0, right? Is `count_num_spikes_per_unit` below using an old cache or something?

alejoe91: good point! Thanks for testing! Let me check ;)

alejoe91: Fixed now! I was missing the `support_periods` class attributes to the spike train metrics

Also added similar tests there :)

src/spikeinterface/metrics/utils.py
alejoe91: I added a test for empty periods in num spikes and firing rates. I could add similar tests for all metrics. What do you think?

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -589,20 +654,36 @@ def compute_firing_ranges(sorting_analyzer, unit_ids=None, bin_size_s=5, percent
...
+    bin_edges_per_unit = compute_bin_edges_per_unit(
+        sorting,
+        segment_samples=segment_samples,
+        periods=periods,
+        bin_duration_s=bin_size_s,
+    )
+    cumulative_segment_samples = np.cumsum([0] + segment_samples[:-1])
     for unit_id in unit_ids:
-        firing_histograms = []
+        if num_spikes[unit_id] == 0:
```
chrishalcrow: Hello - this `num_spikes` seems to be the number of spikes of the original `sorting` before unit periods are selected, which causes problems down the line.

alejoe91: oups!

alejoe91: No it's after, see https://github.com/SpikeInterface/spikeinterface/pull/4302/changes#diff-6dd31db7fc9ed3745b7eb4765b7c107663b51745c246860c36b2517019bc75c3R644

src/spikeinterface/core/tests/test_basesorting.py
samuelgarcia: yes maybe a lextsort and then searchsorted would be more elegant no ?

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -865,6 +867,8 @@ def compute(cls, sorting_analyzer, unit_ids, metric_params, tmp_data, job_kwargs
             args += (tmp_data,)
         if cls.needs_job_kwargs:
             args += (job_kwargs,)
+        if cls.supports_periods:
+            args += (periods,)
```
samuelgarcia, line 871:
```suggestion
            args += (periods,)
        else:
            raise ValueError("This metric do not support periods")
```

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -1129,6 +1137,8 @@ def _compute_metrics(
         if metric_names is None:
             metric_names = self.params["metric_names"]
 
+        periods = self.params.get("periods", None)
```
samuelgarcia, line 1151: I guess this is to be backward compatible.
I think it would be better to make the caompatibility in the `_handle_backward_compatibility_on_load` no ?

src/spikeinterface/metrics/quality/tests/test_metrics_functions.py
```diff
@@ -690,25 +564,262 @@ def test_calculate_drift_metrics(sorting_analyzer_simple):
...
+
+def test_save_quality_metrics(small_sorting_analyzer, create_cache_folder):
+    """
+    Computes quality metrics in binary folder format. Then computes subsets of quality
+    metrics and checks if they are saved correctly.
+    """
+
+    # can't use _misc_metric_name_to_func as some functions compute several qms
+    # e.g. isi_violation and synchrony
+    quality_metrics = [
```
samuelgarcia: this could be infered by the classe no ?

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -1010,67 +1124,79 @@ def compute_drift_metrics(
...
+        # bin_edges are global across segments, so we have to use spike_sample_indices,
+        # since we offseted them to be global
+        bin_spike_indices = np.searchsorted(spike_sample_indices, bins)
+        median_positions = np.nan * np.zeros(num_bins)
+        for bin_index, (i0, i1) in enumerate(zip(bin_spike_indices[:-1], bin_spike_indices[1:])):
+            spikes_in_bin = spike_vector[i0:i1]
+            spike_locations_in_bin = spike_locations_array[i0:i1][direction]
+
+            unit_index = sorting_analyzer.sorting.id_to_index(unit_id)
+            mask = spikes_in_bin["unit_index"] == unit_index
```
samuelgarcia: is this fast ?

src/spikeinterface/metrics/quality/utils.py
```diff
@@ -1,47 +0,0 @@
-from __future__ import annotations
-
-import numpy as np
-
-
-def create_ground_truth_pc_distributions(center_locations, total_points):
```
samuelgarcia, line 6: not needed anymore ?

src/spikeinterface/metrics/utils.py
```diff
@@ -0,0 +1,194 @@
...
+    -------
+    dict
+        Total number of samples for each unit.
+    """
+    if periods is not None:
+        total_samples = {}
+        sorting = sorting_analyzer.sorting
+        for unit_id in sorting.unit_ids:
+            unit_index = sorting.id_to_index(unit_id)
+            periods_unit = periods[periods["unit_index"] == unit_index]
```
samuelgarcia: I have the intuition that lloping only once over periods and suming directly in a prepared vector will be way faster than this repetition of masking in the loop.
No ?

src/spikeinterface/metrics/utils.py
```diff
@@ -0,0 +1,194 @@
...
+            num_samples_in_period = 0
+            for period in periods_unit:
+                num_samples_in_period += period["end_sample_index"] - period["start_sample_index"]
+            total_samples[unit_id] = num_samples_in_period
+    else:
+        total_samples = {unit_id: sorting_analyzer.get_total_samples() for unit_id in sorting_analyzer.unit_ids}
```
samuelgarcia:
```suggestion
        total = sorting_analyzer.get_total_samples()
        total_samples = {unit_id: total for unit_id in sorting_analyzer.unit_ids}
```

src/spikeinterface/metrics/utils.py
```diff
@@ -0,0 +1,194 @@
...
+    -------
+    periods
+        np.ndarray of dtype unit_period_dtype containing the segment, start, end samples and unit index.
+    """
+    all_periods = []
+    for segment_index in range(sorting_analyzer.recording.get_num_segments()):
+        samples_per_period = sorting_analyzer.get_num_samples(segment_index) // num_periods
+        if bin_size_s is not None:
+            bin_size_samples = int(bin_size_s * sorting_analyzer.sampling_frequency)
+            print(samples_per_period / bin_size_samples)
```
samuelgarcia: oups

src/spikeinterface/metrics/utils.py
```diff
@@ -0,0 +1,194 @@
...
+        Total duration for each unit.
+    """
+    total_samples = compute_total_samples_per_unit(sorting_analyzer, periods=periods)
+    total_durations = {
+        unit_id: samples / sorting_analyzer.sampling_frequency for unit_id, samples in total_samples.items()
+    }
+    return total_durations
+
+
+def compute_periods(sorting_analyzer, num_periods, bin_size_s=None):
```
samuelgarcia: I think we should find a better for this.
Like create_regular_periods or something.
The compute is suggesting an algo method.
No ?

src/spikeinterface/metrics/utils.py
```diff
@@ -0,0 +1,194 @@
...
+                period_end = min(period_start + samples_per_period, sorting_analyzer.get_num_samples(segment_index))
+                periods_per_unit[i]["segment_index"] = segment_index
+                periods_per_unit[i]["start_sample_index"] = period_start
+                periods_per_unit[i]["end_sample_index"] = period_end
+                periods_per_unit[i]["unit_index"] = unit_index
+            all_periods.append(periods_per_unit)
+    return np.concatenate(all_periods)
+
+
+def create_ground_truth_pc_distributions(center_locations, total_points):
```
samuelgarcia, line 171: ok
moved here

samuelgarcia:
Globally OK for me
I made a few comment and some of then should be discussed

src/spikeinterface/core/analyzer_extension_core.py
alejoe91: No, the metric_function will support periods in its signature if `cls.support_periods` is True. If not, periods are not added to the args

src/spikeinterface/metrics/quality/tests/test_metrics_functions.py
alejoe91: yes! I'll refactor it!

src/spikeinterface/metrics/quality/misc_metrics.py
alejoe91: Chris tested it and this implementation is slower than Pierre's. I'll push back his implementation

src/spikeinterface/metrics/quality/utils.py
alejoe91: I moved it

src/spikeinterface/metrics/utils.py
alejoe91: this is super fast. Periods will be few..so it's ok!

alejoe91: plus periods are by unit, so we need to slice it by unit anyways!

src/spikeinterface/metrics/quality/tests/test_metrics_functions.py
alejoe91: done

PR #4303, opened 2026-01-07
Needed by Olivier to create an analyzer with precomputed data

yger:


samuelgarcia:
I wanted to propose to move this to core, but too late.

PR #4306, opened 2026-01-08
Bombcell integration and template metrics refactoring

Julie-Fabre:
# Update 26/2/26

After reviewing a lot of detected peaks, we removed smoothing and refined the detection as follows:

1. Use the specified prominence threshold to detect peaks.  If multiple are found, the most prominent is selected as the main extremum.
2. If no peaks are found at the initial threshold, the threshold is halved and detection is attempted again.  If multiple "peaks" are found at the half threshold, the most prominent is selected as the main extremum.
3. If still no peaks are found, a last resort method is used: use the global maximum in the search window.

# Update 12/2/26

This PR ports bombcell-style unit classification to SpikeInterface and refactors/extends template metrics extensively.

#### Template-metrics refactoring

Template metrics have been extended and refactored to align with the bombcell intefration. The new `get_trough_and_peak_idx()` function that uses `scipy.signal.find_peaks()`and finds all peaks (positive extrema) and troughs (negative extrema). Optionally, templates can be smoothed with a Savitzky-Golay filter.

> [!IMPORTANT]  
> The smoothing is off by default

New metrics have been added and other renamed. See new docs here - https://spikeinterface--4306.org.readthedocs.build/en/4306/modules/metrics/template_metrics.html
Main additions are `waveform_ratios`, `waveform_widths`, and `waveform_baseline_flatness`

#### Quality metrics changes

Modified SNR implementation to match bombcell one as follows:
- use the `median` template (when available)
- use `peak_sign="both"` by default, which should handle also non-somatic spikes correctly

### Bombcell labeling

Added an how-to to show the different labelign options available in SI: https://spikeinterface--4306.org.readthedocs.build/en/4306/how_to/auto_label_units.html

- New in `spikeinterface.curation`:
```python
import spikeinterface.curation as sc

thresholds = sc.bombcell_get_default_thresholds()
unit_labels = sc.bombcell_classify_units(
    quality_metrics,
    thresholds=thresholds,
    classify_non_somatic=True,
)
summary = sc.get_classification_summary(unit_type)
```

Units get classified as `noise` → `mua` → `good` based on successive threshold checks. Optional `non_soma` category for non-somatic waveforms. 

> [!IMPORTANT]  
>The function has been refactored (both implementations are still there at the moment) to internally reuse the `threshold_metrics_label_units`, which makes the bombcell logic much clearer

> [!IMPORTANT]  
> The bombcell dictionary has been refactored to contain different secions: "noise"/"mua"/"non-somatic". This makes it much easier to read and potentially to modify. We think this is better, so we don't have to rely on the hard-coded lists of metrics to distribute them in the function

### Plots

- Added plots for classification summaries, metric histograms with threshold lines, waveform overlays by category, and UpSet plots.
```python
from spikeinterface.widgets import (
    plot_unit_labels,  # former WaveformOverlay
    plot_metric_histograms,  # these are not per se connected to classification (so they are part of metrics.py
    plot_bombcell_labels_upset  # this is bombcell specific
)

plot_unit_labels(analyzer, unit_labels)
plot_metric_histograms(analyzer, thresholds=thresholds)
plot_bobcell_upset(analyzer, unit_labels, thresholds)
```
or a wrapper for all plots:
```python
plots = plot_bombcell_unit_labeling_all(
    sorting_analyzer,
    unit_labels,
    thresholds=thresholds,            # optional, uses defaults
    include_upset=True,
)
```

## Original PR body

This PR ports bombcell-style unit classification to SpikeInterface.

##### Template metrics

- Rewrote peak/trough detection with a new `get_trough_and_peak_idx()` function that uses `scipy.signal.find_peaks()`. Since SpikeInterface stores templates based on raw data rather than the heavily smoothed templates used in template matching, the waveforms can be noisy—so you can optionally apply Savitzky-Golay smoothing before detection. The function returns dicts for troughs, peaks before, and peaks after, each containing indices, values, prominences, and widths.
```python
from spikeinterface.postprocessing import get_trough_and_peak_idx

troughs, peaks_before, peaks_after = get_trough_and_peak_idx(
    templates,
    sampling_frequency,
    smooth=True,
    min_thresh_detect_peaks_troughs=0.4,
)
```

- New metrics: `peak_before_to_trough_ratio`, `peak_after_to_trough_ratio`, `waveform_baseline_flatness`, `peak_before_width`, `trough_width`, `main_peak_to_trough_ratio`.

- Renamed `peak_to_valley` to `peak_to_trough_duration`.
```python
analyzer.compute("template_metrics", metric_names=[
    "peak_before_to_trough_ratio",
    "waveform_baseline_flatness",
    "trough_width",
])
```

##### Quality metrics

- Added `snr_bombcell`—peak amplitude over baseline MAD.
```python
analyzer.compute("quality_metrics", metric_names=["snr_bombcell"])
```

- `amplitude_cutoff` now has parameters for controlling the histogram fitting:
```python
analyzer.compute("quality_metrics", metric_names=["amplitude_cutoff"], qm_params={
    "amplitude_cutoff": {
        "num_histogram_bins": 100,
        "histogram_smoothing_value": 3,
    }
})
```

##### Unit classification

- New in `spikeinterface.curation`:
```python
import spikeinterface.comparison as sc

thresholds = sc.bombcell_get_default_thresholds()
unit_type, unit_type_string = sc.bombcell_classify_units(
    quality_metrics,
    thresholds=thresholds,
    classify_non_somatic=True,
)
summary = sc.get_classification_summary(unit_type, unit_type_string)
```

Units get classified as NOISE → MUA → GOOD based on successive threshold checks. Optional NON_SOMA category for non-somatic waveforms. 

##### Plots

- Added plots for classification summaries, metric histograms with threshold lines, waveform overlays by category, and UpSet plots.
```python
from spikeinterface.widgets import (
    plot_unit_classification,
    plot_classification_histograms,
    plot_waveform_overlay,
    plot_upset,
)

plot_unit_classification(analyzer, unit_type, unit_type_string)
plot_classification_histograms(quality_metrics, thresholds=thresholds)
plot_waveform_overlay(analyzer, unit_type, unit_type_string)
plot_upset(quality_metrics, unit_type, unit_type_string)
```
or a wrapper for all plots:
```python
plots = plot_unit_classification_all(
    sorting_analyzer,
    unit_type,
    unit_type_string,
    quality_metrics=quality_metrics,  # optional, will try to get from analyzer
    thresholds=thresholds,            # optional, uses defaults
    split_non_somatic=False,
    include_upset=True,
)
```

### DIFF
--- a/doc/modules/metrics/qualitymetrics/snr.rst
+++ b/doc/modules/metrics/qualitymetrics/snr.rst
@@ -4,11 +4,17 @@ Signal-to-noise ratio (:code:`snr`)
 Calculation
 -----------
 
-- :math:`A_{\mu s}` : maximum amplitude of the mean spike waverform (on the best channel).
+- :math:`A_{\mu s}` : maximum amplitude of the median spike waverform (on the best channel).
 - :math:`\sigma_b` : standard deviation of the background noise on the same channel (usually computed via the `median absolute deviation <https://en.wikipedia.org/wiki/Median_absolute_deviation>`_).
 
 .. math::
     \textrm{SNR} = \frac{A_{\mu s}}{\sigma_b}
+
+The amplitude, bu default, is the amplitude of the largest peak (positive or negative) of the median waveform on the best channel.
+If the ``waveforms`` extension is not available, the amplitude is computedon the average waveform.
+The noise level is computed using the median absolute deviation of the signal on the best channel, which is a robust
+estimator of the standard deviation of the noise.
+
 
 Expectation and use
 -------------------


samuelgarcia:
Salut Julie,
I read this super quickly. This is super impressive what you did during the hackahton!
I was not aware that you also did the widgets stuff. Waou.

I will be back with more carefully reading.

But some main stuff:
  * we avoid to push ipynb in the repo because in saturate the history so we use jupytext instead and push only the resulting generated rst,if the notebook is fast to generate (with simulate data) we also have the tutorial way to push doc through notebooks which is a py file run and generated by the documentaion build.
  * I would prefer to not have json directly in the code to handle parameters. I think simple python file with the same contents. lets discuss more
  * I would be courious to see the correlation between the basic snr and the one median based you did. I will try to make some plot on this.

src/spikeinterface/metrics/template/template_metrics.py
```diff
@@ -9,6 +9,7 @@
 import numpy as np
 import warnings
 from copy import deepcopy
+from scipy.signal import find_peaks
```
alejoe91: Can you move this to the function?

The `core` module has minimal dependencies, and all additional imports should be local :)

src/spikeinterface/curation/unit_labelling.py
```diff
@@ -0,0 +1,430 @@
+"""
+Unit labelling based on quality metrics (Bombcell).
```
alejoe91:
```suggestion
Unit labeling based on quality metrics (Bombcell).
```
In general, we adopted american english (@chrishalcrow  is not happy about it!).

Could you rename this and the files to labeling?

src/spikeinterface/curation/default_thresholds.json
```diff
@@ -0,0 +1,74 @@
+{
```
alejoe91: We can remove this file

src/spikeinterface/curation/unit_labelling.py
```diff
@@ -0,0 +1,430 @@
...
+"""
+
+from __future__ import annotations
+
+import numpy as np
+import pandas as pd
+from typing import Optional
+
+
+WAVEFORM_METRICS = [
```
alejoe91:
```suggestion
NOISE_METRICS = [
```
?

src/spikeinterface/curation/unit_labelling.py
```diff
@@ -0,0 +1,430 @@
...
+    """
+    Bombcell - Returns default thresholds for unit labelling.
+
+    Each metric has 'min' and 'max' values. Use np.nan to disable a threshold (e.g. to ignore a metric completly
+    or to only have a min or a max threshold)
+    """
+    # bombcell
+    return {
+        # Waveform quality (failures -> NOISE)
+        "num_positive_peaks": {"min": np.nan, "max": 2},
```
alejoe91:
```suggestion
        "num_positive_peaks": {"min": None, "max": 2},
```
I would just keep `None` and deal with it in the function instead of NaN, so you can save/load to JSON without any custom fields

src/spikeinterface/curation/unit_labelling.py
```diff
@@ -0,0 +1,430 @@
...
+    if quality_metrics is None:
+        return template_metrics
+    if template_metrics is None:
+        return quality_metrics
+    return quality_metrics.join(template_metrics, how="outer")
+
+
+def bombcell_label_units(
+    quality_metrics=None,
+    template_metrics=None,
```
alejoe91: use `sorting_analyzer` instead

src/spikeinterface/curation/unit_labelling.py
```diff
@@ -0,0 +1,430 @@
...
+        If True, split non-somatic into NON_SOMA_GOOD (3) and NON_SOMA_MUA (4).
+
+    Returns
+    -------
+    unit_type : np.ndarray
+        Numeric: 0=NOISE, 1=GOOD, 2=MUA, 3=NON_SOMA
+    unit_type_string : np.ndarray
+        String labels.
+    """
+    combined_metrics = _combine_metrics(quality_metrics, template_metrics)
```
alejoe91:
```suggestion
    combined_metrics = sorting_analyzer.get_metrics_extension_data()
```
;)

src/spikeinterface/curation/unit_labelling.py
```diff
@@ -0,0 +1,430 @@
...
+    noise_mask = np.zeros(n_units, dtype=bool)
+    for metric_name in WAVEFORM_METRICS:
+        if metric_name not in combined_metrics.columns or metric_name not in thresholds:
+            continue
+        values = combined_metrics[metric_name].values
+        if metric_name in absolute_value_metrics:
+            values = np.abs(values)
+        thresh = thresholds[metric_name]
+        noise_mask |= np.isnan(values)
+        if not np.isnan(thresh["min"]):
```
alejoe91:
```suggestion
        if thresh["min"] is not None:
```
and so on

src/spikeinterface/curation/unit_labelling.py
```diff
@@ -0,0 +1,430 @@
...
+    if split_non_somatic_good_mua:
+        labels = {0: "NOISE", 1: "GOOD", 2: "MUA", 3: "NON_SOMA_GOOD", 4: "NON_SOMA_MUA"}
+    else:
+        labels = {0: "NOISE", 1: "GOOD", 2: "MUA", 3: "NON_SOMA"}
+
+    unit_type_string = np.array([labels.get(int(t), "UNKNOWN") for t in unit_type], dtype=object)
+    return unit_type.astype(int), unit_type_string
+
+
+def apply_thresholds(
```
alejoe91: is this used?

src/spikeinterface/curation/unit_labelling.py
```diff
@@ -0,0 +1,430 @@
...
+            reasons[above_max & (reasons == "")] = "above_max"
+            reasons[above_max & (reasons == "below_min")] = "below_min_and_above_max"
+
+        results[f"{metric_name}_pass"] = passes
+        results[f"{metric_name}_fail_reason"] = reasons
+
+    return pd.DataFrame(results, index=quality_metrics.index)
+
+
+def get_labelling_summary(unit_type: np.ndarray, unit_type_string: np.ndarray) -> dict:
```
alejoe91:
```suggestion
def get_bombcell_labeling_summary(unit_type: np.ndarray, unit_type_string: np.ndarray) -> dict:
```

src/spikeinterface/curation/unit_labelling.py
alejoe91: The file could be called `bombcell_curation` (similar to `model_based_curation`)

src/spikeinterface/curation/unit_labelling.py
```diff
@@ -0,0 +1,430 @@
...
+    summary = {"total_units": n_total, "counts": {}, "percentages": {}}
+    for utype, count in zip(unique_types, counts):
+        label = unit_type_string[unit_type == utype][0]
+        summary["counts"][label] = int(count)
+        summary["percentages"][label] = round(100 * count / n_total, 1)
+
+    return summary
+
+
+def save_thresholds(thresholds: dict, filepath) -> None:
```
alejoe91: if we use None instead of `np.nan` for missing thresholds, the save/load functions can be dropped

src/spikeinterface/curation/unit_labelling.py
```diff
@@ -0,0 +1,430 @@
...
+    for metric_name, thresh in json_thresholds.items():
+        thresholds[metric_name] = {
+            "min": np.nan if thresh["min"] is None else thresh["min"],
+            "max": np.nan if thresh["max"] is None else thresh["max"],
+        }
+
+    return thresholds
+
+
+def save_labelling_results(
```
alejoe91:
```suggestion
def save_bombcell_results(
```

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -182,6 +182,114 @@ class SNR(BaseMetric):
     depend_on = ["noise_levels", "templates"]
 
 
+def compute_snrs_bombcell(
```
alejoe91: I would call this `compute_snrs_versus_baseline` or something like this. We can mention in the notes and in the docs that the implementation is from bombcell. 

Having a clear name will make it immediate for users to understand what it is about!

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -182,6 +182,114 @@ class SNR(BaseMetric):
...
+    waveforms_ext = sorting_analyzer.get_extension("waveforms")
+    nbefore = waveforms_ext.nbefore
+    sampling_frequency = sorting_analyzer.sampling_frequency
+
+    # Calculate baseline samples from ms
+    baseline_samples = int(baseline_window_ms / 1000 * sampling_frequency)
+    baseline_samples = min(baseline_samples, nbefore)  # Can't exceed nbefore
+
+    # Get peak channel for each unit from templates
+    extremum_channels_ids = get_template_extremum_channel(sorting_analyzer, peak_sign=peak_sign)
```
alejoe91:
```suggestion
    extremum_channels_indices = get_template_extremum_channel(sorting_analyzer, peak_sign=peak_sign, outputs="index")
```
You could also already ask for the indices, but we still have to take sparsity into account.

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -182,6 +182,114 @@ class SNR(BaseMetric):
...
+                continue
+            peak_chan_idx = np.where(chan_ids == peak_chan_id)[0][0]
+        else:
+            peak_chan_idx = sorting_analyzer.channel_ids_to_indices([peak_chan_id])[0]
+
+        # Extract waveforms on peak channel
+        waveforms_peak = waveforms[:, :, peak_chan_idx]  # (num_spikes, num_samples)
+
+        # Signal: max absolute value across all spikes
+        signal = np.max(np.abs(waveforms_peak))
```
alejoe91: isn't this extremely sensitive to outliers?

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -182,6 +182,114 @@ class SNR(BaseMetric):
...
+        if noise > 0:
+            snrs[unit_id] = signal / noise
+        else:
+            snrs[unit_id] = np.nan
+
+    return snrs
+
+
+class SNRBombcell(BaseMetric):
+    metric_name = "snr_bombcell"
```
alejoe91:
```suggestion
    metric_name = "snr_to_baseline"
```

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -1471,6 +1632,102 @@ def amplitude_cutoff(amplitudes, num_histogram_bins=500, histogram_smoothing_val
...
+
+            # Create figure if no axes provided
+            if ax_scatter is None and ax_hist is None:
+                fig, axes = plt.subplots(1, 2, figsize=(12, 5))
+                ax_scatter = axes[0]
+                ax_hist = axes[1]
+                created_figure = True
+            else:
+                created_figure = False
+
```
alejoe91: All this should go in the custom widgets. If it's only for debugging purposes, we can remove

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -2,65 +2,534 @@
 
 import numpy as np
 from collections import namedtuple
-
+from scipy.signal import find_peaks, savgol_filter
```
alejoe91: same here, should go in individual functions

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -2,65 +2,534 @@
...
-    peak_idx: int
-        The index of the peak
+    troughs : dict
+        Dictionary containing:
+        - "indices": array of all trough indices
+        - "values": array of all trough values
+        - "prominences": array of all trough prominences
+        - "widths": array of all trough widths
+        - "main_idx": index of the main trough (most prominent)
+        - "main_loc": location (sample index) of the main trough in template
```
alejoe91: what is the difference between these two?

EDIT: I see from the code
```suggestion
        - "main_idx": index of the main trough (most prominent) within the found troughs
        - "main_loc": location (sample index) of the main trough in template
```

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -2,65 +2,534 @@
...
+
+        ax.axhline(0, color="gray", ls="-", alpha=0.3)
+        ax.set_xlabel("Sample")
+        ax.set_ylabel("Amplitude")
+        ax.legend(loc="best", fontsize=8)
+        ax.set_title(f"Trough/Peak Detection (prominence threshold: {min_thresh_detect_peaks_troughs})")
+        plt.tight_layout()
+        plt.show()
+
+    return troughs, peaks_before, peaks_after
```
alejoe91: Remove or move at the end and comment

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -2,65 +2,534 @@
...
+    peaks_before : dict
+        Peak before trough results from get_trough_and_peak_idx
+    peaks_after : dict
+        Peak after trough results from get_trough_and_peak_idx
 
     Returns
     -------
-    ptv: float
-        The peak to valley duration in seconds
+    waveform_duration_us : float
```
alejoe91: why `us` here?

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -512,17 +1077,17 @@ def single_channel_metric(unit_function, sorting_analyzer, unit_ids, tmp_data, *
     return result
 
 
-class PeakToValley(BaseMetric):
-    metric_name = "peak_to_valley"
+class PeakToTroughDuration(BaseMetric):
+    metric_name = "peak_to_trough_duration"
```
alejoe91, line 1010: I'm thinking it could be useful to add a `deprecated_column_names`, so we could automate backward compatibility :)

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -626,11 +1169,21 @@ def _number_of_peaks_metric_function(sorting_analyzer, unit_ids, tmp_data, **met
     num_peaks_result = namedtuple("NumberOfPeaksResult", ["num_positive_peaks", "num_negative_peaks"])
     num_positive_peaks_dict = {}
     num_negative_peaks_dict = {}
-    sampling_frequency = sorting_analyzer.sampling_frequency
+    sampling_frequency = tmp_data["sampling_frequency"]
```
alejoe91: goooooood catch @Julie-Fabre !!!!!!

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -639,22 +1192,184 @@ def _number_of_peaks_metric_function(sorting_analyzer, unit_ids, tmp_data, **met
...
+            peaks_before_info[unit_id],
+            peaks_after_info[unit_id],
+            **metric_params,
+        )
+        result[unit_id] = value
+    return result
+
+
+class WaveformDuration(BaseMetric):
+    metric_name = "waveform_duration"
```
alejoe91: I think that the name doesn't convey the actual computation
```suggestion
class MainToNextPeakDuration(BaseMetric):
    metric_name = "main_to_next_peak_duration"
```
?

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -639,22 +1192,184 @@ def _number_of_peaks_metric_function(sorting_analyzer, unit_ids, tmp_data, **met
...
+    metric_params = {}
+    metric_columns = {
+        "trough_width": float,
+        "peak_before_width": float,
+        "peak_after_width": float,
+    }
+    metric_descriptions = {
+        "trough_width": "Width of the main trough in microseconds",
+        "peak_before_width": "Width of the main peak before trough in microseconds",
+        "peak_after_width": "Width of the main peak after trough in microseconds",
```
alejoe91: I would be consistent and output everything in the same unit. For now we have been doing seconds for the durations. The bombcell curation could still accept thresholds in us and do the conversion on the fly.

Alternatively, we could add a `unit` field to the `BaseMetric`, to specify units for each column. I think I would go with this, but it requires an additional refactoring. @chrishalcrow what do you think?

src/spikeinterface/widgets/unit_labelling.py
```diff
@@ -0,0 +1,560 @@
...
+    return quality_metrics.join(template_metrics, how="outer")
+
+
+class LabellingHistogramsWidget(BaseWidget):
+    """Plot histograms of quality metrics with threshold lines."""
+
+    def __init__(
+        self,
+        quality_metrics=None,
+        template_metrics=None,
```
alejoe91:
```suggestion
        sorting_analyzer
```
same reasons as curation module

alejoe91 (dismissed):
@Julie-Fabre massive effort! Thanks!

I did a first round of reviewing and I'm happy to discuss some details and also work on it :)

Julie-Fabre:
Thank you both so much @samuelgarcia and @alejoe91 ! 
Implementing the changes now and checking the SNR - that was a good point, I had implemented the 'bombcell' version to actually check and see how correlated they were (and hopefully drop the bombcell one and get a good threshold for the 'spikeInterface' one). 
For the JSON file, I would like some option so users can load in a file if they want to. This is really nice for instance if people are sharing pipelines & parameters, if you want to reproduce a paper, etc. If not a JSON file, do you have other ideas for how we could implement that?

Julie-Fabre:
Regarding the 'spikeInterface' vs 'bombcell' SNR:
- spikeInterace SNR: |peak(mean_template)| / MAD(recording_noise), is that right?                                                                 
- Bombcell SNR: |peak(median_waveform)| / MAD(waveform_baseline). Uses median waveform and baseline noise from extracted spikes. 

The do agree mostly:
- Here's a correlation plot:  
image

- And some example units where they don't agree (top row = high spikeInterface SNR and low  bombcell SNR; second row: low SI and high BC; third row: both high and fourth row: both low). I am plotting the median waveform +/- m.a.d. here. 
image


Based on this very quick look at these examples, I would say:
(1) I think the spikeInterface way of choosing the extreme value doesn't work for non-somatic units (basically all the ones in the second row - blue ones). we should update that. 
(2) for the top row (spikeInterface high, bombcell low), it's a mixed bag (some fine, others noisy) but I might tend to prefer the bombcell one (is does seem to seperate better what visually I would say are noisier units from less noisy ones), so I would like keep it at least for the bombcell pipeline. 

I would really like to get your input on this! What do you think? And also what was the rationale for using the full recording to calculate noise?

src/spikeinterface/metrics/quality/misc_metrics.py
Julie-Fabre: Yes, there was an error there, fixed earlier :) Good catch!

src/spikeinterface/metrics/template/metrics.py
Julie-Fabre: us was for "microseconds" - but you're right with below comment, would make more sense to put everything in seconds

src/spikeinterface/metrics/template/metrics.py
Julie-Fabre: Converted to seconds, I think that makes sense. You don't have to inherit some of bombcell's early decisions 😂

src/spikeinterface/metrics/quality/misc_metrics.py
Julie-Fabre: left this for now but feel free to change and implement how you would prefer

src/spikeinterface/metrics/template/metrics.py
Julie-Fabre: That's a great idea, go for it! :)

samuelgarcia:
Thanks Julie for this plot this is really cool. lets have a call to discuss no ?
I suspect that some difference may also be due to the fact that we measurign the negative peak only.
lets check this.

Julie-Fabre:
Yes, absolutely I agree! I think that's what's responsible for all the cases where the SI metric < BC metric (blue guys) but maybe not for the cases where SI metric > BC metric (red guys)? Curious to hear what you think
image
Great idea, let's have a call. Sent you a message on slack :)

src/spikeinterface/widgets/bombcell_curation.py
```diff
@@ -0,0 +1,389 @@
...
+        if n_metrics == 0:
+            print("No metrics to plot")
+            return
+
+        n_cols = min(4, n_metrics)
+        n_rows = int(np.ceil(n_metrics / n_cols))
+        backend_kwargs["ncols"] = n_cols
+        if "figsize" not in backend_kwargs:
+            backend_kwargs["figsize"] = (4 * n_cols, 3 * n_rows)
+        self.figure, self.ax, self.axes = make_mpl_figure(n_rows, n_cols, **backend_kwargs)
```
chrishalcrow:
```suggestion
        self.figure, self.axes, self.ax = make_mpl_figure(**backend_kwargs)
```

src/spikeinterface/widgets/bombcell_curation.py
```diff
@@ -0,0 +1,389 @@
...
+        metrics_to_plot = dp.metrics_to_plot
+
+        n_metrics = len(metrics_to_plot)
+        if n_metrics == 0:
+            print("No metrics to plot")
+            return
+
+        n_cols = min(4, n_metrics)
+        n_rows = int(np.ceil(n_metrics / n_cols))
+        backend_kwargs["ncols"] = n_cols
```
chrishalcrow:
```suggestion
        backend_kwargs["ncols"] = n_cols
        backend_kwargs["num_axes"] = n_cols * n_rows
```

src/spikeinterface/widgets/bombcell_curation.py
```diff
@@ -0,0 +1,389 @@
...
+            unit_type,
+            unit_type_string,
+            thresholds=thresholds,
+            split_non_somatic=split_non_somatic,
+            backend=backend,
+            **kwargs,
+        )
+
+    # Save to folder if requested
+    if save_folder is not None:
```
chrishalcrow: I don't think we should do this during a plotting function call; we don't for any other widgets.

src/spikeinterface/widgets/bombcell_curation.py
```diff
@@ -0,0 +1,389 @@
...
+    Plot UpSet plots showing which metrics fail together for each unit type.
+
+    Requires `upsetplot` package. Each unit type shows relevant metrics:
+    NOISE -> waveform metrics, MUA -> spike quality metrics, NON_SOMA -> non-somatic metrics.
+    """
+
+    def __init__(
+        self,
+        sorting_analyzer,
+        unit_type: np.ndarray,
```
chrishalcrow: `unit_type` doesn't seemed to get used anywhere

chrishalcrow:
Hello, upset plot is incompatible with pandas 3 (which just came out). Think it'd be ok to cap `pandas < 3` in full.

src/spikeinterface/curation/model_based_curation.py
```diff
@@ -127,8 +127,42 @@ def predict_labels(
...
+        if si_version is not None and parse(si_version) < parse("0.103.2"):
+            # if the model was trained with SI version < 0.103.2, we need to rename some metrics
+            calculated_metrics = calculated_metrics.copy()
+            # peak_to_trough_duration was named peak_to_valley
+            if "peak_to_trough_duration" in calculated_metrics.columns:
+                calculated_metrics = calculated_metrics.rename(columns={"peak_to_trough_duration": "peak_to_valley"})
+            # main_peak_to_trough_ratio was named peak_trough_ratio and had inverted sign
+            if "main_peak_to_trough_ratio" in calculated_metrics.columns:
+                calculated_metrics = calculated_metrics.rename(
+                    columns={"main_peak_to_trough_ratio": "peak_trough_ratio"}
```
chrishalcrow: Hello, I think our old implementation of `peak_trough_ratio` is closer to the current implementation of  `-peak_after_to_trough_ratio`, rather than `main_peak_to_trough_ratio`.
```suggestion
                    columns={"peak_after_to_trough_ratio": "peak_trough_ratio"}
```

alejoe91: done in last commit

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -639,26 +1106,192 @@ def _number_of_peaks_metric_function(sorting_analyzer, unit_ids, tmp_data, **met
...
+            peaks_before_info[unit_id],
+            peaks_after_info[unit_id],
+            **metric_params,
+        )
+        result[unit_id] = value
+    return result
+
+
+class MainToNextPeakDuration(BaseMetric):
+    metric_name = "main_to_next_peak_duration"
```
chrishalcrow: I think this name is confusing, because the second extremum is not guaranteed to be a peak. I think it should be `main_to_next_extremum_duration`

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -997,6 +1013,38 @@ def get_optional_dependencies(cls, **params):
...
+            return []
+        else:
+            computed_metric_columns = self.data["metrics"].columns.tolist()
+            computed_metric_names = []
+            for m in self.metric_list:
+                if all(col in computed_metric_columns for col in m.metric_columns.keys()):
+                    computed_metric_names.append(m.metric_name)
+            return computed_metric_names
+
+    def _cast_metrics(self, metrics_df):
```
samuelgarcia, line 1066: duplicated

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -1,66 +1,445 @@
...
     Parameters
     ----------
-    template: numpy.ndarray
+    template : numpy.ndarray
         The 1D template waveform
+    min_thresh_detect_peaks_troughs : float, default: 0.4
+        Minimum prominence threshold as a fraction of the template's absolute max value
+    smooth : bool, default: True
+        Whether to apply smoothing before peak detection
+    smooth_window_frac : float, default: 0.1
```
samuelgarcia: I think I wowuld prefer to have this params in ms like `smooth_window_ms`
to avoid different smoothing when we change ms_before/ms_after

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -1,66 +1,445 @@
 from __future__ import annotations
 
+import warnings
 import numpy as np
-from collections import namedtuple
-
 from spikeinterface.core.analyzer_extension_core import BaseMetric
```
samuelgarcia, line 5: I would add a small and consice internal code doc string to explain the main rules for peak and trough detection concept.

src/spikeinterface/curation/bombcell_curation.py
```diff
@@ -0,0 +1,343 @@
...
+    # String labels
+    if split_non_somatic_good_mua:
+        labels = {0: "NOISE", 1: "good", 2: "mua", 3: "non_soma_good", 4: "non_soma_mua"}
+    else:
+        labels = {0: "noise", 1: "good", 2: "mua", 3: "non_soma"}
+
+    unit_type_string = np.array([labels.get(int(t), "unknown") for t in unit_type], dtype=object)
+    labels = pd.DataFrame(data={"label": unit_type_string, "label_type": unit_type}, index=combined_metrics.index)
+
+    return labels
```
samuelgarcia: I think we should return on the unit_type_string and handle this in any downstream function.

src/spikeinterface/curation/bombcell_curation.py
```diff
@@ -0,0 +1,343 @@
...
+
+
+def get_bombcell_labeling_summary(unit_type: np.ndarray, unit_type_string: np.ndarray) -> dict:
```
samuelgarcia: this should work only on unit_type_string

src/spikeinterface/curation/bombcell_curation.py
```diff
@@ -0,0 +1,343 @@
...
+        values = combined_metrics[metric_name].values
+        if metric_name in absolute_value_metrics:
+            values = np.abs(values)
+        thresh = thresholds[metric_name]
+        noise_mask |= np.isnan(values)
+        if not is_threshold_disabled(thresh["min"]):
+            noise_mask |= values < thresh["min"]
+        if not is_threshold_disabled(thresh["max"]):
+            noise_mask |= values > thresh["max"]
+    unit_type[noise_mask] = 0
```
samuelgarcia: I think I would do directly 
```suggestion
    unit_type[noise_mask] = "noise"
```
to avoid remembering internal number.
This also would make the code super easy to read.
No ?

src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -939,10 +939,17 @@ def get_metric_columns(cls, metric_names=None):
...
+            metric_names = [m.metric_name for m in cls.metric_list]
+        else:
+            for metric_name in metric_names:
+                if metric_name not in [m.metric_name for m in cls.metric_list]:
+                    raise ValueError(
+                        f"Metric {metric_name} not in available metrics {cls.get_available_metric_names()}"
+                    )
+        for metric_name in metric_names:
+            metric_class = cls.get_metric_by_name(metric_name)
+            default_metric_columns.extend(metric_class.metric_columns)
```
samuelgarcia, line 965: Can I try a small rewrite of this ?

src/spikeinterface/widgets/unit_labels.py
```diff
@@ -0,0 +1,125 @@
...
+    ):
+        sorting_analyzer = self.ensure_sorting_analyzer(sorting_analyzer)
+        self.check_extensions(sorting_analyzer, "templates")
+        if labels_order is not None:
+            assert len(labels_order) == len(np.unique(unit_labels)), "labels_order length must match unique unit types"
+            assert all(
+                [label in np.unique(unit_labels) for label in labels_order]
+            ), "All labels in labels_order must be present in unit_labels"
+        else:
+            labels_order = np.unique(unit_labels)
```
samuelgarcia, line 47: This will change the behavior if you have only good units for instance because it will be only on plots.
I would hard code of the possible no ?

src/spikeinterface/metrics/template/template_metrics.py
```diff
@@ -239,6 +294,9 @@ def _prepare_data(self, sorting_analyzer, unit_ids):
 
         tmp_data["troughs"] = troughs
         tmp_data["peaks"] = peaks
```
samuelgarcia: We should rename peak_after no ? to avoid the confusion

src/spikeinterface/metrics/template/metrics.py
alejoe91: done

src/spikeinterface/core/analyzer_extension_core.py
alejoe91: I rewrote a bit already, but sure!

src/spikeinterface/metrics/template/template_metrics.py
alejoe91: done! also added indices

src/spikeinterface/metrics/template/metrics.py
alejoe91: done

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -1,138 +1,558 @@
...
+        "peak_after_width": peak_after_width * samples_to_seconds if not np.isnan(peak_after_width) else np.nan,
+    }
+
+    return widths
+
+
+#########################################################################################
+# Single-channel metrics
+def get_peak_to_trough_duration(
+    template_single, sampling_frequency, trough_idx, peak_after_trough_idx, **kwargs
```
chrishalcrow: `template_single` is not used

doc/modules/metrics/template_metrics.rst
```diff
@@ -1,37 +1,148 @@
...
+------------------------------
+
+Duration in seconds from the main extremum (peak or trough) to the next (in time) extremum.
+If no next extremum is found, it returns NaN.
+
+repolarization_slope
+--------------------
+
+Slope of the repolarization phase of the template, between the trough (minimum) and return to baseline in uV/s.
+If there is no trough,it returns NaN. If the absolute maximum is a peak, it still uses largest trough to compute.
```
chrishalcrow:
```suggestion
If there is no trough, it returns NaN. If the absolute maximum is a peak, it still uses largest trough to compute.
If the template never returns to baseline, it returns NaN.
```

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -208,52 +626,53 @@ def get_recovery_slope(template_single, sampling_frequency, peak_idx=None, **kwa
...
 
-    if peak_idx == 0:
+    if peak_after_trough_idx is None or peak_after_trough_idx == 0:
         return np.nan
-    max_idx = int(peak_idx + ((recovery_window_ms / 1000) * sampling_frequency))
+    max_idx = int(peak_after_trough_idx + ((recovery_window_ms / 1000) * sampling_frequency))
     max_idx = np.min([max_idx, template_single.shape[0]])
 
-    res = scipy.stats.linregress(times[peak_idx:max_idx], template_single[peak_idx:max_idx])
+    res = scipy.stats.linregress(times[peak_after_trough_idx:max_idx], template_single[peak_after_trough_idx:max_idx])
```
chrishalcrow: Just to check: this uses all points from the "peak after trough" to the end of the template and does a linear fit on them?

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -208,52 +626,53 @@ def get_recovery_slope(template_single, sampling_frequency, peak_idx=None, **kwa
...
     return res.slope
 
 
-def get_number_of_peaks(template_single, sampling_frequency, **kwargs):
```
chrishalcrow, line 223: No longer uses `template_single` or `sampling_frequency`

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -498,82 +1004,72 @@ def get_spread(template, channel_locations, sampling_frequency, **kwargs) -> flo
...
-class PeakToValley(BaseMetric):
-    metric_name = "peak_to_valley"
+class PeakToTroughDuration(BaseMetric):
+    metric_name = "peak_to_trough_duration"
     metric_params = {}
-    metric_columns = {"peak_to_valley": float}
+    metric_columns = {"peak_to_trough_duration": float}
     metric_descriptions = {
-        "peak_to_valley": "Duration in s between the trough (minimum) and the peak (maximum) of the spike waveform."
+        "peak_to_trough_duration": "Duration in seconds between the trough (minimum) and the peak (maximum) of the spike waveform."
```
chrishalcrow:
```suggestion
        "peak_to_trough_duration": "Duration in seconds between the trough (minimum) and the next peak of the unit template."
```

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -498,82 +1004,72 @@ def get_spread(template, channel_locations, sampling_frequency, **kwargs) -> flo
...
 
 
 class HalfWidth(BaseMetric):
     metric_name = "half_width"
     metric_params = {}
-    metric_columns = {"half_width": float}
+    metric_columns = {"trough_half_width": float, "peak_half_width": float}
     metric_descriptions = {
-        "half_width": "Duration in s at half the amplitude of the trough (minimum) of the spike waveform."
+        "trough_half_width": "Duration in s at half the amplitude of the trough (minimum) of the spike waveform."
```
chrishalcrow:
```suggestion
        "trough_half_width": "Duration in s at half the amplitude of the trough (minimum) of the spike waveform.",
        "peak_half_width": "Duration in s at half the amplitude of the peak (maximum) of the spike waveform.",
```

doc/modules/metrics/template_metrics.rst
```diff
@@ -1,37 +1,148 @@
...
+This extension computes commonly used template metrics. A *template* is the
+average (or median) spike waveform for each unit.
+
+First, the templates are upsampled (optionally, default 10) to increase the temporal
+resolution for more precise metric computation. Then they are (optionally) smoothed
+(using a Savitzky-Golay filter) to reduce high-frequency noise that could interfere
+with peak/trough detection.
+
+Then the negative and positive peaks are detected on the channel with the largest
+absolute peak (the "main channel"). From these peaks, several metrics are computed.
```
chrishalcrow: !!! _negative_ peaks!

doc/modules/metrics/template_metrics.rst
```diff
@@ -1,37 +1,148 @@
...
+
+.. note::
+
+    The units of :code:`recovery_slope` and :code:`repolarization_slope` depend on the
+    input. Voltages are based on the units of the template. By default this is µV
+    but can be the raw output from the recording device (this depends on the
+    :code:`return_in_uV` parameter, read more here: :ref:`modules/core:SortingAnalyzer`).
+    Distances are in µm and times are in seconds. So, for example, if the
+    templates are in units of µV then :code:`recovery_slope` :code:`repolarization_slope` is in
+    µV/s.
```
chrishalcrow: Should be higher up?

doc/modules/metrics/template_metrics.rst
chrishalcrow: To be more helpful...

"""
We compute the channel where the extremal value of your template occurs. Many single channel metrics only depend on the template on this channel, while multi-channel metrics will depend on the templates from many channels of your probe. On the extremal channel, we compute the troughs (minima) and peaks (maxima) of the template.
"""

src/spikeinterface/metrics/template/template_metrics.py
```diff
@@ -106,24 +108,52 @@ def _handle_backward_compatibility_on_load(self):
...
+                self.params["metric_names"].append("peak_to_trough_duration")
+        # peak_trough ratio -> main peak to trough ratio
+        # note that the new implementation correctly uses the absolute peak values,
+        # which is different from the old implementation.
+        # we make a flag to invert the polarity of old values if needed
+        if "peak_trough_ratio" in self.params["metric_names"]:
+            self.params["metric_names"].remove("peak_trough_ratio")
+            if "waveform_ratios" not in self.params["metric_names"]:
+                self.params["metric_names"].append("waveform_ratios")
+            self.params["metric_params"]["invert_peak_to_trough"] = True
```
chrishalcrow, line 171: Need to add that old `half_width` maps to `trough_half_width`

chrishalcrow: And old `peak_to_valley` maps to `peak_to_trough_duration`

chrishalcrow: Oh wait, maybe we do this in the unitrefine backward compat bit... sorry

src/spikeinterface/metrics/quality/misc_metrics.py
alejoe91: expose `operator`

src/spikeinterface/widgets/bombcell_curation.py
```diff
@@ -0,0 +1,275 @@
...
+    Parameters
+    ----------
+    sorting_analyzer : SortingAnalyzer
+        The sorting analyzer object with computed metrics extensions.
+    unit_labels : np.ndarray
+        Array of unit labels as strings.
+    thresholds : dict, optional
+        Threshold dictionary. If None, uses default thresholds.
+    split_non_somatic : bool, default: False
+        Whether to split "non_soma" into "non_soma_good" and "non_soma_mua".
```
alejoe91:
```suggestion
```
not used anymore

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -146,7 +146,7 @@ class PresenceRatio(BaseMetric):
 def compute_snrs(
     sorting_analyzer,
     unit_ids=None,
-    peak_sign: str = "neg",
+    peak_sign: str = "both",
     peak_mode: str = "extremum",
```
alejoe91, line 148: expose operator by default

alejoe91:
@chrishalcrow @samuelgarcia @Julie-Fabre this is ready to merge!

I updated the docs,  added a "new tool" entry in README and docs, and also added a reference also in the citation page.

I also changed the default column name returned by the `bombcell_label_units` function (`bombcell_label`) and also `unitrefine_label_units` (`unitrefine_label`). This way we can simply merge the returned dataframes to the metrics

alejoe91 (approved):

doc/index.rst
```diff
@@ -12,6 +12,7 @@ amazing algorithms and formats that we interface with. See them all, and how to
 `references page <https://spikeinterface.readthedocs.io/en/latest/references.html>`_. In the past year, we have added support
 for the following tools:
 
+- Bombcell. `Bombcell: automated curation and cell classification of spike-sorted electrophysiology data. 2023. <https://doi.org/10.5281/zenodo.8172822>`_ (`docs <https://spikeinterface.readthedocs.io/en/latest/modules/curation.html#auto-labeling-units>`_)
```
chrishalcrow:
```suggestion
- Bombcell. `Bombcell: automated curation and cell classification of spike-sorted electrophysiology data. 2023. <https://doi.org/10.5281/zenodo.8172822>`_ (`docs <https://spikeinterface.readthedocs.io/en/stable/how_to/auto_label_units.html#bombcell>`_)
```
I think these are better docs for bombcell (link won't work until merged)
Also on README!

PR #4305, opened 2026-01-08
Option to turn on parallel computation for CCG in numba

yger:
This PR adds the possibility to activate a fast_mode for the numba implementation. The mode will parallelize use prange(), but at the cost of some (rare) racing conditions that might possibily bring some numerical instabilities. Tested on 1000 units firing at few herz for 1000s, the error is less than one spike out of a million compared to the normal one

PR #4307, opened 2026-01-08
Auto correlograms

yger:
This PR adds a new extension: auto_correlograms. This is a gift for @oliche 

This is separated from correlograms because the internal data structure is simpler. Note that also as soon as #4305 is accepted, the fast_mode option of numba should be propagated also to this extension. 

EDIT by Sam:
This new (faster) extension could be used both in the GUI but also for some metrics when correlograms extension is no here.


oliche:
So I would just have to use `method='auto'` and I'll get the autocorrelograms ! Thanks @yger

src/spikeinterface/postprocessing/__init__.py
```diff
@@ -15,9 +15,12 @@
 from .correlograms import (
     ComputeACG3D,
     ComputeCorrelograms,
+    ComputeAutoCorrelograms,
     compute_acgs_3d,
     compute_correlograms,
+    compute_auto_correlograms,
     correlogram_for_one_segment,
+    auto_correlogram_for_one_segment,
```
alejoe91, line 23: The new extension should alse be added here: https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/sortinganalyzer.py#L2890

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -206,8 +206,121 @@ def _get_data(self):
...
+    def _merge_extension_data(
+        self, merge_unit_groups, new_unit_ids, new_sorting_analyzer, censor_ms=None, verbose=False, **job_kwargs
+    ):
+        """
+        When two units are merged, their cross-correlograms with other units become the sum
+        of the previous cross-correlograms. More precisely, if units i and j get merged into
+        unit k, then the new unit's cross-correlogram with any other unit l is:
+            C_{k,l} = C_{i,l} + C_{j,l}
+            C_{l,k} = C_{l,k} + C_{l,j}
+        Here, we apply this formula to quickly compute correlograms for merged units.
```
alejoe91: I don't think this is true :P

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -206,8 +206,121 @@ def _get_data(self):
...
+        """
+
+        new_bins = self.data["bins"]
+        all_new_units = new_sorting_analyzer.unit_ids
+        arr = self.data["acgs"]
+
+        # compute all new isi at once
```
alejoe91:
```suggestion
        # compute all new acgs at once
```

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -206,8 +206,121 @@ def _get_data(self):
...
+            else:
+                new_unit_index = new_sorting.id_to_index(unit_id)
+                new_acgs[unit_ind, :] = only_new_acgs[new_unit_index, :]
+
+        new_data = dict(acgs=new_acgs, bins=new_bins)
+        return new_data
+
+    def _split_extension_data(self, split_units, new_unit_ids, new_sorting_analyzer, verbose=False, **job_kwargs):
+        # TODO: for now we just copy
+        new_acgs, new_bins = _compute_auto_correlograms_on_sorting(new_sorting_analyzer.sorting, **self.params)
```
alejoe91: Can't we just recompute on the new unit ids?

yger: yes, we can

src/spikeinterface/postprocessing/__init__.py
yger: done

src/spikeinterface/postprocessing/correlograms.py
```diff
@@ -209,8 +210,132 @@ def _get_data(self):
...
+                new_unit_index = new_sorting.id_to_index(unit_id)
+                new_acgs[unit_ind, :] = only_new_acgs[new_unit_index, :]
+
+        new_data = dict(acgs=new_acgs, bins=new_bins)
+        return new_data
+
+    def _split_extension_data(self, split_units, new_unit_ids, new_sorting_analyzer, verbose=False, **job_kwargs):
+
+        arr = self.data["acgs"]
+        num_dims = len(self.data["bins"])
```
chrishalcrow: This isn't right.
```suggestion
        num_dims = len(self.data["bins"]) - 1
```

Maybe bins should be called `bin_edges` throughout?

yger: I've renamed as done in isi.py

alejoe91:
Tested on my end too! Results are the same as diagonals of ccgs, as expected :)

alejoe91 (approved):

PR #4308, opened 2026-01-08
Timing the computations of quality metrics and some optimizations

yger:
After playing with real-size datasets during the hackathon amazingly organized by our "ami du bouchon" @samuelgarcia , I realized, that some persons might consider spikeinterface too slow. And because we want everyong to join the game (the more the merrier), we need to be inclusive. During the hackthon, we've worked to solve the problem with cross-correlograms (see #4305 and #4307 ). However, we might wonder what can be done for the other metrics, and to do so, first thing is to time the metric. This is what is now added by the PR

For a 1000s recording with 1000 units, you can immediatly see some issues, and they are all fixed in this PR. Some metrics, for example, are using suboptimal correlograms (numpy implemetantations), and some others I guess should be optimized. Before, we used to have

{'num_spikes': 21.245169458998134,
 'firing_rate': 0.0007224159999168478,
 'presence_ratio': 0.14417108400084544,
 'snr': 0.040033251003478654,
 'isi_violation': 0.049835821002488956,
 'rp_violation': 3.409717330003332,
 'sliding_rp_violation': 7.334715744997084,
 'synchrony': 21.48789839699748,
 'firing_range': 0.36698940899805166,
 'amplitude_cv': 28.745208095002454,
 'amplitude_cutoff': 0.5351223960024072,
 'noise_cutoff': 0.8977982619981049,
 'amplitude_median': 0.4023318469990045,
 'sd_ratio': 48.09409817199776}

With the PR, we now have:

- [x] {'num_spikes': 0.12104277299658861} 
- [x] {'sd_ratio': 2.566413887994713}
- [x] {'amplitude_cv': 4.915091472001222,
- [x] {'rp_violation': 1.7641781819984317}  
- [x] {'sliding_rp_violation': 4.426980464999797}
- [x] {'synchrony': 7.823143080997397}

Plus, now rp_violation accepts a unit_ids to avoid recomputations during slices. I'm fairly happy with the PR as it is. It can also be boosted with speed up in CCG. I've also ported the optimizaiton to the drift metric, now really fast

- [x] {drift': 5.542134855990298}

Also now

- [x] {isi_histogram': 4.542134855990298} instead of 25s in the numba (!) implementation. lexsort power !

### DIFF
--- a/src/spikeinterface/postprocessing/isi.py
+++ b/src/spikeinterface/postprocessing/isi.py
@@ -122,7 +122,7 @@ def _compute_isi_histograms(sorting, window_ms: float = 50.0, bin_ms: float = 1.
     assert method in ("auto", "numba", "numpy")
 
     if method == "auto":
-        method = "numpy" # numpy is faster for ISI computationc currently
+        method = "numpy"  # numpy is faster for ISI computationc currently
 
     if method == "numpy":
         return compute_isi_histograms_numpy(sorting, window_ms, bin_ms)


samuelgarcia:
Pierre notre héros.

samuelgarcia:
>I realized, that some persons might consider spikeinterface too slow. 

No name please.

samuelgarcia:
I would also consider that slow computation should be removed of the by default metrics.
No ?

yger:
This looks good to me, I've checked the results are the same, and I've drastically reduce the cost of the most important ones. The only remaining extension that is slow is the synchrony, but not sure to understand exactly what it does. Is people needing it? Are also people using the metrics for the drifts @alejoe91 ? Because they could be optimized also

yger:
And also, there is something strange, i think, in the implementation of sliding_rpv_violations. At least, this is not what I would have done, and I'm not sure this is also exactly what the authors meant to do, but I could be wrong, I need to read more the paper to double check. I made an implementation that keep results the same (but can gain numba implementation of correlograms, and eventually also of the fast mode available in #4305 )

yger:
@alejoe91 I've also drastically improve the drift metric. @alejoe91 @samuelgarcia @chrishalcrow let me know what do you think, but this is now good for me. All metrics have been heavily optimized using the spike vector, and they are much faster now

yger:
On real-world data from the IBL, 1140 units for 1000s with  ~ 38 millions spikes, the computations of all metrics (including drift) goes from 320 second seconds to only 84 secons with this PR on my laptop. So a x4 speedup, at least, with identical results

src/spikeinterface/metrics/spiketrain/metrics.py
chrishalcrow:
```suggestion
    num_spikes = compute_num_spikes(sorting_analyzer, unit_ids=unit_ids)
```

src/spikeinterface/metrics/quality/misc_metrics.py
chrishalcrow:
```suggestion
    num_spikes = compute_num_spikes(sorting_analyzer, unit_ids=unit_ids)
```

This was making some slowdown when merging two units. Maybe less of an issue now that `compute_num_spikes` is faster.

yger:
Actually, I spoke with @samuelgarcia, if we really want to speed up all metrics, we should cache some lexsorted versions of the spikes at the NumpySorting objects. Ideally, one might want to do
`sorting.to_spike_vector(lexsort = ['sample_index', 'segment_index', 'unit_index']) `
to get a spike vector with a nested sort first by unit_index, then segment, then sample. 
But some other metrics would prefer
`sorting.to_spike_vector(lexsort = ['sample_index', 'unit_index', 'segment_index']) `

Currently, each metric re performs the needed lexsort, but if they were cached at the NumpySorting level, this would tremendously speed up everything. Of course, one should also return the order of the lexsort, because it is also used to rearrange any other vectors (amplitudes, locations, ...)

chrishalcrow:
Hello, I had a play. Made a 1000s, 1000 unit simulated analyzer with lots of extensions computed and saved it. Then in the timing script, I loaded it and computed all metrics.

ALL units, main:

num_spikes, 9.624790583038703
firing_rate, 0.0005816250341013074
presence_ratio, 0.31368916702922434
snr, 0.025650916039012372
isi_violation, 0.02097741700708866
rp_violation, 1.7714388340245932
sliding_rp_violation, 6.2393900419119745
synchrony, 10.178920249920338
firing_range, 0.29491766705177724
amplitude_cv, 14.793372999993153
amplitude_cutoff, 0.43810770800337195
noise_cutoff, 0.6302923749899492
amplitude_median, 0.33469658298417926
drift, 22.543062791926786
sd_ratio, 23.485498708905652

ALL units, this PR

num_spikes, 0.25287729199044406
firing_rate, 0.2299201659625396
presence_ratio, 2.5794546250253916
snr, 0.025951166986487806
isi_violation, 2.3115992909297347
rp_violation, 2.3352270829491317
sliding_rp_violation, 4.0499598330352455
synchrony, 4.83623274997808
firing_range, 2.695753209060058
amplitude_cv, 4.37225637503434
amplitude_cutoff, 0.6063804998993874
noise_cutoff, 0.7644850000506267
amplitude_median, 0.27151691704057157
drift, 5.510953124961816
sd_ratio, 2.760369290946983

So overall: amazing! But regressions for:

num_spikes
firing_rate
presence_ratio
isi_violation
firing_range

Basically, the simple spike vector ones.

I also wanted to test what happens when you just compute the metrics for a small number of units (main use case is when you merge). In this script I loaded the simulated analyzer (which already had QMs computed), then recomputed the metrics for just two units. You get some regressions:

Two units, main

num_spikes, 0.04532308399211615
firing_rate, 9.583084832993336
presence_ratio, 0.011025874991901219
snr, 0.02882220793981105
isi_violation, 0.0021689170971512794
rp_violation, 1.5441597920143977
sliding_rp_violation, 0.02066708297934383
synchrony, 9.783891499973834
firing_range, 0.00224808300845325
amplitude_cv, 0.044612999889068305
amplitude_cutoff, 0.2512064999900758
noise_cutoff, 0.24916829192079604
amplitude_median, 0.13169516704510897
drift, 0.1567430419381708
sd_ratio, 0.052918249974027276

Two units, this PR

num_spikes, 0.20829199999570847
firing_rate, 0.1590971250552684
presence_ratio, 2.1553558750310913
snr, 0.025400415994226933
isi_violation, 2.310786666930653
rp_violation, 0.8214222909882665
sliding_rp_violation, 2.7106911670416594
synchrony, 4.857573916902766
firing_range, 2.4199495000066236
amplitude_cv, 2.4296372500248253
amplitude_cutoff, 0.38967483292799443
noise_cutoff, 0.2728223749436438
amplitude_median, 0.1626084999879822
drift, 5.014932292047888
sd_ratio, 2.581335166003555

So the following metrics

num_spikes
presence_ratio
isi_violation
sliding_rp_violation
amplitude_cv
drift
sd_ratio

Are slower now. I'm guessing they are doing the calculation for all units, then selecting the required units?

yger:
Ok, so I could try to see the problems for just pairs of merges. 

You are right, the reason why we are slowing down for some extensions is that now, almost all extensions are  perfoming a lexsort of the total spike vector. Before, this used to be cached. I would like to cache it again, but I would like to even cache more of these lexsorts, because different extensions get some speed up sorting the data in different order. This should be discussed, I'm sure we could get almost instantaneous computations for most of these. Maybe we should discuss all together about how to proceed, and if I have your green light @alejoe91 @samuelgarcia @chrishalcrow

yger:
I've made some caching proposal, with an extension discussed with @samuelgarcia to_lexsort_spike_vector() in the code. This is the best I can do, and this brings down again the times because lexsort is done only once, same as searchsorted (I'm storing all indices of segments/unit_index depending on the lexsort)

 @chrishalcrow can you retry with your profiling, and/or send me the code to challenge the merging by pairs? 
But here, we might consumes more memory, but the computation of metrics is as fast as you could dream of. On the same dataset as before, now this is down to 45s. So a x8 speedup for global metrics

chrishalcrow:
Perfect, no regressions when doing a small number of units, and all wayyy faster when computing all metrics.

When just computing e.g. two units, `synchrony` is the slowest (by two orders of magnitude) but this makes sense, since you need to compare all spikes to compute it. So it's probably difficult to speed up.

EDIT: on a real NP2 dataset (without pca metrics) the re-computation needed for merging takes less than a second. Amazing!

chrishalcrow:
I double checked that the results from this PR matched the results from main. All exactly the same, except `amplitude_cutoff`, which was different for a small number of units (2%) at third significant figure.

yger:
Actually, this is a strange point, since I did not optimize amplitude_cutoff... But looking at it, maybe it could also be optimized, because currently I guess this is not using the cached indices. i'll have a look

chrishalcrow:
> Actually, this is a strange point, since I did not optimize amplitude_cutoff... But looking at it, maybe it could also be optimized, because currently I guess this is not using the cached indices. i'll have a look

Probably my fault: I took an pre-computed analyzer (from 1 year ago), recomputed the metrics using this PR then compared to the original metrics. So the "main" metrics were computed a long time ago. So it's actually pretty impressive that only `amplitude_cutoff` has changed (and has barely changed!). Could easy be a numpy 1->2 change or something...

yger:
But at least, now I've also optimized noise_cutoff, amplitude_cuttoff and amplitude_medians :-)

yger:
I have the feeling that the same lexsort approach could be ported to the pca_metrics, both to compute them, and to optimize the metrics, but I'm not sure of who is really using these metrics. Do you think this is worth looking into that, to speed up also this part of the code? The key point would be to make the internal computing nodes aware of the lexsorted slices, to avoid all the "unit_index" == XXX in the code

yger:
This is now based on #4315, so should be merged after

yger:
This is good to go @samuelgarcia . You can squash and merge as soon as tests are passed

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -67,13 +66,13 @@ def compute_presence_ratios(sorting_analyzer, unit_ids=None, bin_duration_s=60.0
         unit_ids = sorting_analyzer.unit_ids
     num_segs = sorting_analyzer.get_num_segments()
 
-    seg_lengths = [sorting_analyzer.get_num_samples(i) for i in range(num_segs)]
+    # seg_lengths = [sorting_analyzer.get_num_samples(i) for i in range(num_segs)]
```
chrishalcrow:
```suggestion
```

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -89,6 +88,8 @@ def compute_presence_ratios(sorting_analyzer, unit_ids=None, bin_duration_s=60.0
         )
         presence_ratios = {unit_id: np.nan for unit_id in unit_ids}
     else:
+
+        # precompute segment slice
```
chrishalcrow:
```suggestion
```

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -479,6 +476,7 @@ class SlidingRPViolation(BaseMetric):
         "exclude_ref_period_below_ms": 0.5,
         "max_ref_period_ms": 10,
         "contamination_values": None,
+        "correlograms_kwargs": dict(method="auto"),
```
chrishalcrow: This would be the only place in metrics were we use a dict as a param. I'm wondering if this might make some params logic more complex somewhere in the code. E.g. in the automated curation stuff, we check that the metric params used to compute the metrics used to train the curation model are the same as the metric params used to compute the metrics of the current analyzer. It's probably fine: @alejoe91 thoughts?

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -597,13 +596,12 @@ def compute_firing_ranges(sorting_analyzer, unit_ids=None, bin_size_s=5, percent
...
-            spike_times = sorting.get_unit_spike_train(unit_id=unit_id, segment_index=segment_index)
+    firing_rate_histograms = {unit_id: np.array([], dtype=float) for unit_id in unit_ids}
+    for unit_id in unit_ids:
+        for segment_index in range(sorting_analyzer.get_num_segments()):
+            num_samples = sorting_analyzer.get_num_samples(segment_index)
+            edges = np.arange(0, num_samples + 1, bin_size_samples)
+            spike_times = sorting.get_unit_spike_train(unit_id, segment_index)
             spike_counts, _ = np.histogram(spike_times, bins=edges)
             firing_rates = spike_counts / bin_size_s
             firing_rate_histograms[unit_id] = np.concatenate((firing_rate_histograms[unit_id], firing_rates))
```
samuelgarcia: concatenate in the loop is not a good practice maybe better to accumulate in a loop and then concatenate no ?

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -678,21 +676,13 @@ def compute_amplitude_cv_metrics(
...
     sorting = sorting_analyzer.sorting
     total_duration = sorting_analyzer.get_total_duration()
-    spikes = sorting.to_spike_vector()
-    num_spikes = sorting.count_num_spikes_per_unit(outputs="dict")
+
     if unit_ids is None:
         unit_ids = sorting.unit_ids
 
-    amps = sorting_analyzer.get_extension(amplitude_extension).get_data()
+    num_spikes = sorting_analyzer.sorting.count_num_spikes_per_unit(unit_ids=unit_ids)
```
samuelgarcia:
```suggestion
    num_spikes = sorting_analyzer.sorting.count_num_spikes_per_unit(unit_ids=unit_ids, outputs="dict")
```

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -1032,19 +1020,13 @@ def compute_drift_metrics(
...
-    spike_locations = spike_locations_ext.get_data()
-    # spike_locations_by_unit = spike_locations_ext.get_data(outputs="by_unit")
-    spikes = sorting.to_spike_vector()
-    spike_locations_by_unit = {}
-    for unit_id in unit_ids:
-        unit_index = sorting.id_to_index(unit_id)
-        # TODO @alessio this is very slow this sjould be done with spike_vector_to_indices() in code
-        spike_mask = spikes["unit_index"] == unit_index
-        spike_locations_by_unit[unit_id] = spike_locations[spike_mask]
+    spike_locations_by_unit_and_segments = spike_locations_ext.get_data(outputs="by_unit")
```
samuelgarcia:
```suggestion
    spike_locations_by_unit_and_segments = spike_locations_ext.get_data(outputs="by_unit", concatenated=False)
```

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -1475,15 +1445,15 @@ def amplitude_cutoff(amplitudes, num_histogram_bins=500, histogram_smoothing_val
...
+    sorting,
     duration,
     bin_size_ms=0.25,
     window_size_s=1,
     exclude_ref_period_below_ms=0.5,
     max_ref_period_ms=10,
     contamination_values=None,
     return_conf_matrix=False,
+    correlograms_kwargs=dict(method="auto"),
 ):
```
samuelgarcia, line 1451: maybe we could not expose this just to simplify the API and parameters handling on analyser side.
nested stuff will to complicated

chrishalcrow: I probably agree: https://github.com/SpikeInterface/spikeinterface/pull/4308#discussion_r2691449565

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -1525,25 +1495,28 @@ def slidingRP_violations(
...
+
+        correlogram = _compute_correlograms_numpy(sorting, window_size, bin_size)[0, 0]
+    if method == "numba":
+        from spikeinterface.postprocessing.correlograms import _compute_correlograms_numba
+
+        correlogram = _compute_correlograms_numba(sorting, window_size, bin_size)[0, 0]
+
+    ## I dont get why this line is not giving exactly the same result as the correlogram function. I would question
+    # the choice of the bin_size above, but I am not the author of the code...
+    # correlogram = compute_correlograms(sorting, 2*window_size_s*1000, bin_size_ms, **correlograms_kwargs)[0][0, 0]
```
samuelgarcia: maybe there is a trick in the bin size no ?
I would prefer this commented line

chrishalcrow: yeah, it's a `int` (floor) vs `round` thing. This function uses
```
bin_size = max(int(bin_size_ms / 1000 * sorting.sampling_frequency), 1)
```
while compute_correlograms uses
```
bin_size = round(bin_size_ms/1000 * sampling_frequency)
```

src/spikeinterface/metrics/quality/misc_metrics.py
yger: To be honest, I don't like that neither, but I really would like to be able to use the numba implementation, and also the fast_mode option described in #4305 . This would be only doable if we let the user access the correlogram params. But I could flatten that

src/spikeinterface/metrics/quality/misc_metrics.py
yger: The problem in this function is deeper. In fact, I also prefer the commented line, and this is the one I wrote initially. But this does not give the same results as the current implementation. Because of the rounding issue, as @chrishalcrow is pointing out, but also more importantly because of the window_s params. In this metric function, this size is only half of the cc, while in compute_correlogram (our function), this is half. So we would need to harmonize this, otherwise compute_correlogram is computing smaller correlograms than the current implementation in the metric. But I suspect this is a bug in the implementation itself, because this does not really make sense

src/spikeinterface/metrics/quality/misc_metrics.py
yger: Done, I moved the concatenate outside the loop

src/spikeinterface/metrics/quality/misc_metrics.py
yger: I've removed the kwargs, and hardcoded an auto method. When fast_mode will be there, I'll simply add it as an option

src/spikeinterface/metrics/quality/misc_metrics.py
yger: I've removed the kwargs, and hardcoded an "auto" method. When fast_mode will be there, I'll simply add it as an option

alejoe91:
@yger I'll test the metric values against the released/main version :) They should all be the same right?

src/spikeinterface/metrics/quality/misc_metrics.py
yger: The plan would be to leave it like that for now, and then open a new PR to discuss only this topic: do we want to patch such a behavior, which I think is incorrect. The window_s cureently in this function seems to large, and maybe the creation of the bins should be harmonized (and symmetric around 0). But this would slightly change results so we'll discuss that later

yger:
Yes, they should be identical, @chrishalcrow already made such a comparison, but better twice !

alejoe91:
Parfait!!!
image

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -433,31 +422,33 @@ def compute_sliding_rp_violations(
...
+    )
 
-        if not any([np.any(train) for train in spike_train_list]):
-            continue
+    for unit_id in unit_ids:
+        unit_index = sorting.id_to_index(unit_id)
+        u0 = slices[unit_index, 0, 0]
+        u1 = slices[unit_index, -1, 1]
+        sub_spikes = spikes[u0:u1].copy()
+        sub_spikes["unit_index"] = 0  # single unit sorting
```
alejoe91, line 439: Maybe worth updating the the `SelectUnitSorting` with this logic if lexsorted array is cached? We can do in a follow up PR

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -805,10 +786,11 @@ def compute_amplitude_cutoffs(
         invert_amplitudes = True
         extension = sorting_analyzer.get_extension("amplitude_scalings")
 
-    amplitudes_by_units = extension.get_data(outputs="by_unit", concatenated=True)
+    amplitudes_by_units = extension.get_data(concatenated=True, outputs="by_unit")
```
alejoe91, line 789: great change!

src/spikeinterface/postprocessing/isi.py
```diff
@@ -122,7 +122,7 @@ def _compute_isi_histograms(sorting, window_ms: float = 50.0, bin_ms: float = 1.
     assert method in ("auto", "numba", "numpy")
 
     if method == "auto":
-        method = "numba" if HAVE_NUMBA else "numpy"
+        method = "numpy" if HAVE_NUMBA else "numpy"
```
alejoe91: ?????

src/spikeinterface/metrics/quality/misc_metrics.py
yger: Yes, gigantic !

src/spikeinterface/metrics/quality/misc_metrics.py
yger: This is something I wanted to discuss but maybe yes, there could be some speedup while slicing sorting. I guess this would be one other PR

samuelgarcia:
Congratulation from the entire fan base!

PR #4309, opened 2026-01-09
Removing gil for numba (mostly in peak detection)

yger:
Could speedup numba function


samuelgarcia:
Good catch.
Indeed big speedup when using threads.
So maybe we could switch back to threads by default for windows no ?

yger:
I guess yes, for windows we should use threads

samuelgarcia:
Closed by #4333

PR #4314, opened 2026-01-13
Move structured dtypes to base

alejoe91:
Since structured dtypes are used in different parts of the code, it makes more sense to move them to base from `node_pipeline`.

Will need to refactor #4299 #4302 #4301 accordingly

### DIFF
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -5,7 +5,7 @@ from typing import Optional
 
 import numpy as np
 
-from .base import BaseExtractor, BaseSegment
+from .base import BaseExtractor, BaseSegment, minimum_spike_dtype
 from .waveform_tools import has_exceeding_spikes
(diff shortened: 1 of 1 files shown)


alejoe91:
@samuelgarcia @chrishalcrow  all in favor?

chrishalcrow:
Sounds reasonable - seems like a good idea to define our important dtypes somewhere central.
We also have `minimum_spike_dtype` in `basesorting.py`, but that's probably ok?

alejoe91:
> Sounds reasonable - seems like a good idea to define our important dtypes somewhere central. We also have `minimum_spike_dtype` in `basesorting.py`, but that's probably ok?

I'll move that here too!

alejoe91:
> minimum_spike_dtype

done!

alejoe91:
@chrishalcrow @samuelgarcia I think this is ready to merge

PR #4321, opened 2026-01-15
Improve extension dependency management in SortingAnalyzer

alejoe91:
This PR improves the extension dependency management system.
New class methods for the base `AnalyzerExtension` give more flexibility, by allowing dependencies to be specified at runtime based on the extension params:
```
cls.get_required_extensions(**params)

cls.get_optional_extension(**params)
```

The `compute_several_extension` and `_sort_extensions_by_dependency` will only use the required extension. 
Optional extensions are used though to further organize the computation as follows. 
Extensions are split into 3 groups:
- pre_pipeline: extensions that to not depend on pipeline nodes and are run before
- with_pipeline: extensions that can be run toghether in a with `run_node_pipeline`
- post_pipeline: extensions whose any dependency (required or optional) depend on a pipeline node extension

This new organization allows us to remove the quality metric hack! In the current state, quality metrics were always run as post pipeline because some of them may depend on spike amplitudes/locations. Now, depending on the metric list, quality metrics (and other extensions!) will be moved post pipeline if needed:

See this in effect with some debug prints:
image

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -1770,7 +1770,20 @@ extension_params={"waveforms":{"ms_before":1.5, "ms_after": "2.5"}}\
         >>> sorting_analyzer.compute_several_extensions({"waveforms": {"ms_before": 1.2}, "templates" : {"operators": ["average", "std"]}})
 
         """
+        # Check dependencies: either already computed or in the extensions to compute
+        extensions_to_compute = list(extensions.keys())
+        for extension_name, extension_params in extensions.items():
+            required_dependencies = get_extension_class(extension_name).get_required_dependencies(**extension_params)
+            for dependency_name in required_dependencies:
+                if "|" in dependency_name:
+                    ok = any(
+                        self.has_extension(name) or name in extensions_to_compute for name in dependency_name.split("|")
+                    )
+                else:
+                    ok = self.has_extension(dependency_name) or dependency_name in extensions_to_compute
+                assert ok, f"Extension {extension_name} requires {dependency_name} to be computed first"
 
+        # Sort extensions by dependency order
         sorted_extensions = _sort_extensions_by_dependency(extensions)
 
         for extension_name in sorted_extensions.keys():


src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -959,18 +959,28 @@ def get_metric_column_descriptions(cls, metric_names=None):
...
-
-        for col in metrics_df.columns:
-            if col in metric_dtypes:
-                try:
-                    metrics_df[col] = metrics_df[col].astype(metric_dtypes[col])
-                except Exception as e:
-                    print(f"Error casting column {col}: {e}")
-        return metrics_df
+    @classmethod
+    def get_optional_dependencies(cls, **params):
```
alejoe91, line 963: For `BaseMetricExtensions`, optional dependencies will include all dependencies for metrics. These are not required, because if a a metric doesn't meet its dependencies it's skipped in the `_set_params`

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1135,16 +1135,8 @@ def _save_or_select_or_merge_or_split(
         else:
             raise ValueError(f"SortingAnalyzer.save: unsupported format: {format}")
 
-        # make a copy of extensions
-        # note that the copy of extension handle itself the slicing of units when necessary and also the saveing
-        sorted_extensions = _sort_extensions_by_dependency(self.extensions)
```
alejoe91, line 1140: sorting here was not needed, because the `compute_several_extensions` already does all the checks and sorting required!

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1781,25 +1775,30 @@ def compute_several_extensions(self, extensions, save=True, verbose=False, **job
...
-                # the output of the pipeline extensions (e.g., spike_amplitudes, spike_locations).
-                extensions_post_pipeline[extension_name] = extension_params
-                continue
             extension_class = get_extension_class(extension_name)
             if extension_class.use_nodepipeline:
                 extensions_with_pipeline[extension_name] = extension_params
+            elif any(
+                get_extension_class(d).use_nodepipeline
+                for d in extension_class.get_all_dependencies(**extension_params)
+                if d in sorted_extensions
```
alejoe91, line 1807: any extension whose dependencies use node pipeline are moved to the end

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -2042,16 +2037,31 @@ def _sort_extensions_by_dependency(extensions):
...
+        for dependency in required_dependencies:
+            if "|" in dependency:
+                for dep in dependency.split("|"):
+                    if extension not in _extension_children[dep]:
+                        print(f"Adding missing dependency child link: {dep} -> {extension}")
+                        _extension_children[dep].append(extension)
+            else:
+                if extension not in _extension_children[dependency]:
+                    print(f"Adding missing dependency child link: {dependency} -> {extension}")
+                    _extension_children[dependency].append(extension)
```
alejoe91, line 2072: If an extension is not a child of another extension by default, but it becomes one based on parameters, we add it to the extension children

src/spikeinterface/core/tests/test_sortinganalyzer.py
```diff
@@ -640,6 +682,33 @@ def test_extensions_sorting():
...
+    sorting_analyzer.compute("dummy_pipeline", param0=5)
+
+    # raise if dependency not computed
+    with pytest.raises(AssertionError):
+        sorting_analyzer.compute("dummy_pipeline", param0=15)
+
+    # run fine if dependency computed
+    sorting_analyzer.compute(["dummy", "dummy_pipeline"], extension_params=dict(dummy_pipeline=dict(param0=11)))
+
+
```
alejoe91, line 716: add test: if recompute `dummy`, `dummy_pipeline` should be deleted.

alejoe91:
@samuelgarcia this is good to go! I renamed the function and added the extra test

samuelgarcia:
@chrishalcrow ok for you ?

chrishalcrow:
I'll read it now!

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -2042,16 +2037,31 @@ def _sort_extensions_by_dependency(extensions):
...
     i = 0
     while i < len(extensions_list):
 
         extension = extensions_list[i]
-        dependencies = get_extension_class(extension).depend_on
+        extension_params = extension_params_list[i]
+
+        required_dependencies = get_extension_class(extension).get_required_dependencies(**extension_params)
+        optional_dependencies = get_extension_class(extension).get_optional_dependencies(**extension_params)
+        # TODO: If dependencies are added because of parameters, we need to add to the global _extension_children
```
```suggestion
```

alejoe91: you're right! This is done :)

PR #4329, opened 2026-01-19
Ensure compute unit locations respects order of unit ids

### LINKED ISSUE #4328: Computation of `unit_locations` doesn't respect `unit_id` order
The computation of unit locations (actually the bug in is `compute_monopolar_triangulation`) assumes that when you give unit ids, they are in the same order as the analyzer's ids.

One way to see something is wrong is that changing the order of `unit_ids` gives the same result:

``` python
rec, sort = si.generate_ground_truth_recording(seed=1205)
sa.compute(['random_spikes', 'templates'])
unit_locs_3_6 = sa.compute('unit_locations', unit_ids = ['3','6'])
unit_locs_6_3 = sa.compute('unit_locations', unit_ids = ['6','3'])

np.all(unit_locs_3_6.get_data() == unit_locs_6_3.get_data())
```

``` python
>>> True
```

Not a problem for most workflows: but is a problem when doing merges.

Fix is pretty easy, but want to extend the `AnalyzerExtensionCommonTestSuite` to check for this type of bug across the repo.

chrishalcrow:
Fixes #4328

To compute unit locations, you need to loop over templates. This PR ensures that the templates are ordered in the same way as the user-inputted unit ids if they are given.

Added a test to the common_extension_tests, but it only applies to extensions which allow for the user to input `unit_ids`. Turns out: this is just unit locations. I still think it's a new place to put the test?

### DIFF
--- a/src/spikeinterface/postprocessing/localization_tools.py
+++ b/src/spikeinterface/postprocessing/localization_tools.py
@@ -283,7 +283,7 @@ def compute_grid_convolution(
     weights_sparsity_mask = weights > 0
 
     nb_weights = weights.shape[0]
-    unit_location = np.zeros((unit_ids.size, 3), dtype="float64")
+    unit_location = np.zeros((len(unit_ids), 3), dtype="float64")
 
     for i, unit_id in enumerate(unit_ids):
         main_chan = peak_channels[unit_id]


src/spikeinterface/postprocessing/localization_tools.py
```diff
@@ -90,8 +90,8 @@ def compute_monopolar_triangulation(
         unit_ids = sorting_analyzer_or_templates.unit_ids
     else:
         unit_ids = np.asanyarray(unit_ids)
-        keep = np.isin(sorting_analyzer_or_templates.unit_ids, unit_ids)
```
yger, line 93: I would prefer a syntax with units_to_indices(unit_ids), but maybe this would require to implement such a function for the Templates

src/spikeinterface/postprocessing/localization_tools.py
```diff
@@ -90,8 +90,8 @@ def compute_monopolar_triangulation(
...
-        templates = templates[keep, :, :]
+        keep = [np.where(sorting_analyzer_or_templates.unit_ids == unit_id)[0][0] for unit_id in unit_ids]
```
alejoe91, line 93:
```suggestion
        unit_ids = list(sorting_analyzer_or_templates.unit_ids)
        keep_unit_indices = np.array([unit_ids.index(unit_id) for unit_id in unit_ids])
```

alejoe91:
Thansk @chrishalcrow 

Merging!

Issue #4320, opened 2026-01-15
Argument spike_retriver_kwargs (sic) is not sent to the SpikeRetriever in ComputeSpikeLocations.


ecobost:
Smaller issue (typo): The argument in https://github.com/SpikeInterface/spikeinterface/blob/840c0562d6a1ba721ac72cee695b6a3e94f9b26c/src/spikeinterface/postprocessing/spike_locations.py#L53 should be `spike_retriever_kwargs` ("e" is missing). 

Bigger issue: retriever_kwargs is not actually used during the initialization of the retriever (never was); the only kwarg used from it is peak_sign (to find the template extremum) https://github.com/SpikeInterface/spikeinterface/blob/840c0562d6a1ba721ac72cee695b6a3e94f9b26c/src/spikeinterface/postprocessing/spike_locations.py#L77-L87

I'm happy to open a PR to fix it, just let me know whether:
a. You actually do wanna keep the retriever_kwargs and use it to init SpikeRetriever. I ask because ComputeSpikeAmplitudes receives only the peak_sign and calls SpikeRetriever with the default arguments so I wonder whether ComputeSpikeLocations should do the same https://github.com/SpikeInterface/spikeinterface/blob/840c0562d6a1ba721ac72cee695b6a3e94f9b26c/src/spikeinterface/postprocessing/spike_amplitudes.py#L43-L45

or 

b. You do wanna send the kwargs to SpikeRetriever() and are ok with renaming it to spike_retriever_kwargs (fixing the typo). Theoretically, it's a non-backwards-compatible change (people could be using the argument with the "retriver" name)

alejoe91:
Thanks @ecobost!!
@samuelgarcia @yger what's your take?

yger:
I think this is an advanced features, so I don't mind changing the name and fixing the typo. This is worth it I guess. 

Regarding the fact that spike_retriever_kwargs is not fed to SpikeRetriever during the computations of the spike locations, I am not sure this is a bug. The idea was more that in this case, because we know the templates and the channel_indices of the peak, we really want to take the value of the samples centered at this particular channel_index. The node pipeline was more general and also had, if I recall correctly, the option to search for the best channel at a given time, but @samuelgarcia should comment on that

samuelgarcia:
oups.
yes lets fix that
merci

PR #4330, opened 2026-01-19
Change ComputeSpikeLocations to use only peak_sign

ecobost:
Fixes https://github.com/SpikeInterface/spikeinterface/issues/4320. 

Instead of receiving spike_retriever_kwargs (which was not used anyway), receive peak_sign (mirroring ComputeSpikeAmplitudes). Previous default behavior and capabilities will not change (as `peak_sign` was the only parameter actually used from spike_retriever_kwargs).

Code will change from
```python
analyzer.compute('spike_locations', spike_retriver_kwargs={'peak_sign': 'both'}) # previous version
analyzer.compute('spike_locations', peak_sign='both') # new version
```

### DIFF
--- a/src/spikeinterface/postprocessing/spike_locations.py
+++ b/src/spikeinterface/postprocessing/spike_locations.py
@@ -14,10 +14,10 @@ class ComputeSpikeLocations(BaseSpikeVectorExtension):
         The left window, before a peak, in milliseconds
     ms_after : float, default: 0.5
         The right window, after a peak, in milliseconds
-    peak_sign : "neg" | "pos" | "both", default: "neg" 
+    peak_sign : "neg" | "pos" | "both", default: "neg"
         The peak sign to use when looking for the template extremum channel.
     spike_retriever_kwargs : dict
-        Arguments to control the spike retriever behavior. See 
+        Arguments to control the spike retriever behavior. See
         `spikeinterface.sortingcomponents.peak_localization.SpikeRetriever`.
     method : "center_of_mass" | "monopolar_triangulation" | "grid_convolution", default: "center_of_mass"
         The localization method to use
@@ -38,7 +38,7 @@ class ComputeSpikeLocations(BaseSpikeVectorExtension):
     def _handle_backward_compatibility_on_load(self):
         # For backwards compatibility - this renames spike_retriver_kwargs to spike_retriever_kwargs
         if "spike_retriver_kwargs" in self.params:
-            self.params['peak_sign'] = self.params['spike_retriver_kwargs'].get('peak_sign', 'neg')
+            self.params["peak_sign"] = self.params["spike_retriver_kwargs"].get("peak_sign", "neg")
             self.params["spike_retriever_kwargs"] = self.params.pop("spike_retriver_kwargs")
 
     def _set_params(
@@ -71,9 +71,11 @@ class ComputeSpikeLocations(BaseSpikeVectorExtension):
             self.sorting_analyzer, peak_sign=peak_sign, outputs="index"
         )
 
-        retriever_kwargs = {"channel_from_template": True,
-                            "extremum_channel_inds": extremum_channels_indices,
-                            **self.params["spike_retriever_kwargs"]}
+        retriever_kwargs = {
+            "channel_from_template": True,
+            "extremum_channel_inds": extremum_channels_indices,
+            **self.params["spike_retriever_kwargs"],
+        }
         retriever = SpikeRetriever(sorting, recording, **retriever_kwargs)
         nodes = get_localization_pipeline_nodes(
             recording,


ecobost:
Forgot to change references to spike_retriever_kwargs in other code/docs/test.

I found there is only one place where it was used:
https://github.com/SpikeInterface/spikeinterface/blob/5b2e73887520206f258ddb278e43ca004359ca2f/src/spikeinterface/benchmark/benchmark_peak_localization.py#L27-L30
As said in https://github.com/SpikeInterface/spikeinterface/issues/4320, this didn't actually have any effect (cause spike_retriever_kwargs was not passed to SpikeRetriever()), so I commented this code out.I think spike_locations.py is cleaner/easier-to-understand receiving only peak_sign as argument  but if you think the use case in benchmark_peak_localization.py warrants the extra complexity, let me know and I'll revert the changes (and make sure spike_retriever_kwargs is forwarded to the SpikeRetriever).

src/spikeinterface/benchmark/benchmark_peak_localization.py
```diff
@@ -24,11 +24,11 @@ def __init__(self, recording, gt_sorting, params, gt_positions, channel_from_tem
             self.params[key] = self.params.get(key, 2)
             self.templates_params[key] = self.params[key]
 
-        if not self.channel_from_template:
-            self.params["spike_retriver_kwargs"] = {"channel_from_template": False}
-        else:
-            ## TODO
-            pass
```
samuelgarcia, line 29: This was very important for our benchmark. Lets keep it!
We need the peak channel to be estimated for each peak to take in account the variability.
@yger

samuelgarcia:
Thank you for this.

We need to be ver very carfull when changing parameters in extension because then when we reload an old analyzer from disck from previous version this will lead to bugs because the params are inconsistent.

In case we change parameters we need to ensure a backward compatibitility using th `_handle_backward_compatibility_on_load()` method. Also we try to change to often parameters in extension as less as we can for this reason.

In this particular I think we should keep the previous signature with no change and only ensure that this parameters is propagated correctly.

What do you think ?

src/spikeinterface/benchmark/benchmark_peak_localization.py
ecobost: I guess this is the reason that the spike_retriever_kwargs argument exists. 

I am unsure what's the purpose of the Benchmark class, but finding the extremum channel on a single waveform (https://github.com/SpikeInterface/spikeinterface/blob/679cb2ed975d6ec821c839fb5e91933fdd23836a/src/spikeinterface/core/node_pipeline.py#L241) might be pretty noisy (it is limited to channels within 50 um of the original template extremum channel so maybe not so bad). 

Also, you might need to check whether results change a lot; I actually doubt it because the spike_location method is still computed the same way as before just centered at a slightly different channel (the per-spike channel_index rather than the template channel_index.)

ecobost:
I'll reinstate `spike_retriver_kwargs` and make sure it is passed to SpikeRetriever. It's up to you guys whether you wanna rename it, personally, I think it looks sloppy to have a misspelled argument. Also, if you do fix the typo, it might make sense to also add a peak_sign argument anyway to keep it explicit and consistent with other extensions, like:
```python
analyzer.compute("spike_amplitudes", peak_sign='both')
analyzer.compute('template_metrics', peak_sign='both')
analyzer.compute('spike_locations', peak_sign='both')
```
rather than using the peak_sign from spike_retriever_kwargs.

I am happy to add a commit for that too (with the `_handle_backward_compatibility_on_load()`), if you want

yger:
I think that indeed, we should fix the typo, and maybe add the peak_sign option, as you are suggesting. In fact, you are right than in the benchmark, this is the only place where this spike_retriever params are used, because we need to detect the peaks of the spikes knowing the ground truth as they would have been detected without. This is why, when we search for peaks, once we have it at a proper place, we allow to search for other max in a certain spatio-temporal neighboorhood to capture the channel noise that might happen in reality, when detecting peaks. But agreed this is an advanced use case

ecobost:
Ok, https://github.com/SpikeInterface/spikeinterface/pull/4330/commits/be02fbd3b9400ca5227d1ce9b0110e9b53b512f1 has peak_sign as a diff argument, spike_retriever_kwargs fixed and backwards compatibiliy handled. I like this version.

ecobost:
Hi @samuelgarcia I think this PR might also make sense for 0.104.0; it doesn't really extend capabilities as there was a way to send peak_sign='both' `analyzer.compute('spike_locations', spike_retriver_kwargs={'peak_sign': 'both'})` but it standardizes it a little with other methods by receiving peak_sign directly `analyzer.compute('spike_locations', peak_sign='both')`. And spike_retriever_kwargs is now only sent to SpikeRetriever

src/spikeinterface/postprocessing/spike_locations.py
```diff
@@ -74,17 +68,17 @@ def _get_pipeline_nodes(self):
...
-            recording,
-            channel_from_template=True,
-            extremum_channel_inds=extremum_channels_indices,
-        )
+        retriever_kwargs = {
+            "channel_from_template": True,
+            "extremum_channel_inds": extremum_channels_indices,
+            **self.params["spike_retriever_kwargs"],
+        }
+        retriever = SpikeRetriever(sorting, recording, **retriever_kwargs)
```
alejoe91, line 79:
```suggestion
        retriever = SpikeRetriever(sorting, recording, peak_sign=self.params["peak_sign"], **retriever_kwargs)
```
I know it was not present before, but since we are at it ;)

ecobost: peak_sign in is only used in SpikeRetriever when channel_from_template is False (ignores extremum_channel_indices and finds the extremum channel per waveform) but in here we send channel_from_template=True and extremum_channel_indices so I thought it would be confusing. My thought was peak_sign (the argument to ComputeSpikeLocations) is for the compute_spike_locations function and If you want to send peak_sign to the retriever you would have to use the `spike_retriever_kwargs` arg.
``` python
compute('spike_locations', peak_sign='both') 
compute('spike_locations', retriever_kwargs={'channel_from_template': False, 'peak_sign': 'both', 'radius_um': 70, ...}
```

That said, not a hill I would die on.

alejoe91: Got it! Makes sende :)

Can you fix the conflict? Then it's ready to merge!

ecobost:
I rebased the changes back into main branch to fix merging conflicts (the imports of spike_locations.py in main changed) but did not change any code (sorry, should have merged rather than rebased to avoid creating new commits :/)

PR #4331, opened 2026-01-19
Make `make_shared_array` spawn safe

alejoe91:
The `SharedMemory` object must be created after setting the same `mp_context` used later for computation.

@yger can you test this?


samuelgarcia (changes requested):
I think it is not the sharedmem but the multiprocessing.Lock that cause the problem

samuelgarcia:
I think #4333  fix the the same bug.
The sharemem do not need mp_context but the multiprocessing.Lock() do.

I get also the same error than pierre and thi PR is also fixing

PR #4333, opened 2026-01-19
Improve multithreading and multi processing : lock + nogil + mp_context

samuelgarcia:
This try to make lupin working with thread.
This will be usefull because on windows the spwan and process parrallel is using too much memory

 * Release on numba kernel
 * avoid racing when compilation on numba
 * and some other trick
 * fix mp_context and Lock

this replace:
 #4331
#4309

### DIFF
--- a/.github/workflows/all-tests.yml
+++ b/.github/workflows/all-tests.yml
@@ -239,6 +239,19 @@ jobs:
         run: |
           pip install -e .[full]
           pip list
+
+          # Internal shell check for platform and env var
+          if [[ "$RUNNER_OS" == "macOS" ]]; then
+            # Locate the binary dynamically
+            SHM_BINARY=$(python -c "import torch; from pathlib import Path; print(Path(torch.__file__).parent / 'bin' / 'torch_shm_manager')")
+            if [ -f "$SHM_BINARY" ]; then
+              echo "Applying chmod +x to $SHM_BINARY"
+              chmod +x "$SHM_BINARY"
+            else
+              echo "Warning: torch_shm_manager not found at $SHM_BINARY"
+            fi
+          fi
+
           ./.github/run_tests.sh sortingcomponents --no-virtual-env
 
       - name: Test generation
--- a/.github/workflows/deepinterpolation.yml
+++ b/.github/workflows/deepinterpolation.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: '3.9'
+          python-version: '3.10'
       - name: Get changed files
         id: changed-files
         uses: tj-actions/changed-files@v46.0.1
@@ -40,9 +40,8 @@ jobs:
         run: |
           python -m pip install -U pip  # Official recommended way
           # install deepinteprolation
-          pip install tensorflow==2.7.0
+          pip install tensorflow==2.8.4
           pip install deepinterpolation@git+https://github.com/AllenInstitute/deepinterpolation.git
-          pip install protobuf==3.20.*
           pip install numpy==1.26.4
           pip install -e .[full,test_core]
       - name: Pip list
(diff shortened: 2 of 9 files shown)


src/spikeinterface/core/node_pipeline.py
```diff
@@ -601,7 +613,16 @@ def run_node_pipeline(
     else:
         raise ValueError(f"wrong gather_mode : {gather_mode}")
 
-    init_args = (recording, nodes, skip_after_n_peaks_per_worker)
+    node0 = nodes[0]
+    if isinstance(node0, PeakSource) and node0.need_first_call_before_pipeline:
+        # See need_first_call_before_pipeline : this trigger numba compilation before the run
+        node0._first_call_before_pipeline()
```
alejoe91, line 619: Shouldn't this be a common function (by default `pass`) for all nodes? I assume any nodes using a numba kernel would benefit from this

samuelgarcia: Other nodes are more complicated because we need to run the entire chain.
This help now for peak_detection.
Lets see how we can optimize this

samuelgarcia: And at the moement this is implemented only in peak source nodes.

yger:
I have a problem with the branch @samuelgarcia , this is crashing for your dev notebooks with the compute_analyzer() case...

src/spikeinterface/core/job_tools.py
```diff
@@ -418,11 +418,24 @@ def __init__(
...
+                    mp_context = "spawn"
+                else:
+                    mp_context = "spawn"
+
+            preferred_mp_context = recording.get_preferred_mp_context()
+            if preferred_mp_context is not None and preferred_mp_context != mp_context:
+                warnings.warn(
+                    f"You processing chain using pool_engine='process' and mp_context='{mp_context}' is not possible."
+                    f"So use mp_context='{preferred_mp_context}' instead")
+                mp_context = preferred_mp_context
```
alejoe91, line 439: What if a preprocessor wants to use `spawn` on linux (maybe because it uses GPUs)?

yger:
@sam you should merge my optimizations for the waveforms/templates estimations. And do not forget your patch for shallow_copies.

src/spikeinterface/core/job_tools.py
```diff
@@ -418,11 +418,24 @@ def __init__(
...
+                    # We used to force spawn for macos, this is sad but in some cases fork in macos
+                    # is very unstable and lead to crashes.
+                    mp_context = "spawn"
+                else:
+                    mp_context = "spawn"
+
+            preferred_mp_context = recording.get_preferred_mp_context()
+            if preferred_mp_context is not None and preferred_mp_context != mp_context:
+                warnings.warn(
+                    f"You processing chain using pool_engine='process' and mp_context='{mp_context}' is not possible."
```
yger: typo "Your processing..."

src/spikeinterface/core/node_pipeline.py
```diff
@@ -620,10 +641,12 @@ def run_node_pipeline(
     return outs
 
 
-def _init_peak_pipeline(recording, nodes, skip_after_n_peaks_per_worker):
+def _init_peak_pipeline(recording, nodes, need_shallow_copy, skip_after_n_peaks_per_worker):
     # create a local dict per worker
     worker_ctx = {}
     worker_ctx["recording"] = recording
+    if need_shallow_copy:
+        nodes = [copy.copy(node) for node in nodes]
```
yger: avoid the copy, otherwise this is crashing

src/spikeinterface/core/job_tools.py
samuelgarcia: Then this is handle because `mp_context = preferred_mp_context`
This is the same logic as before ?

samuelgarcia:
Ready on my side

PR #4334, opened 2026-01-20
Let the user defines if he want to get a warning for filtering wrt margin

yger:
Currently, we hard coded a value of MARGIN_TO_CHUNK_PERCENT_WARNING = 0.2 # 20% that pops a warning each time the user is requesting a small chunk compared to the margin. 

But this has very annoying side effects when dealing with analyzers. Users providing a filtered signal as an input will have tons of warning messages when extracting waveforms (that are snippets rather small). For now, this PR, at the very least, offer the possibilities for user to create a filtered object that will display marging warnings OR not. 

Should it be done automatically in case of an analyzer ? This is an open question here

### DIFF
--- a/src/spikeinterface/preprocessing/filter.py
+++ b/src/spikeinterface/preprocessing/filter.py
@@ -70,7 +70,7 @@ class FilterRecording(BasePreprocessor):
     display_margin_to_chunk_percent_warning : float | None, default: 0.2
         If not None, a warning is displayed if the margin size is more than this fraction of
         the chunk size during get_traces calls.
-        
+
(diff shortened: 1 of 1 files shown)


samuelgarcia:
I do not like it too much.
lets discuss more about the warnings

yger:
Open to any discussions, but I do think such warnings are usefull. I just want a way to avoid them for the analyzer

alejoe91:
@samuelgarcia what do you suggest?

alejoe91:
Closing in favor of #4384

PR #4338, opened 2026-01-22
Refactor API for `auto_label_units`

alejoe91:
The `auto_label_units` was used only for the model-based labeling, but there are other ways to label units! (see #4304 )

This PR renames `auto_label_units` to `model_based_label_units`. In addition, it adds a `unitrefine_label_units` that automate the noise/neural + MUA/SUA classification. By default, lightweight models are used since they give almost the same performance as the full models, but don't depend on PCA metrics.

### DIFF
--- a/src/spikeinterface/curation/model_based_curation.py
+++ b/src/spikeinterface/curation/model_based_curation.py
@@ -289,6 +289,7 @@ def auto_label_units(*args, **kwargs):
         "`auto_label_units` is deprecated and will be removed in v0.105.0. "
         "Please use `model_based_label_units` instead.",
         DeprecationWarning,
+        stacklevel=2,
     )
     return model_based_label_units(*args, **kwargs)
 


src/spikeinterface/curation/model_based_curation.py
```diff
@@ -204,7 +204,7 @@ def _export_to_phy(self, classified_df):
         classified_df.to_csv(f"{sorting_path}/cluster_prediction.tsv", sep="\t", index_label="cluster_id")
 
 
-def auto_label_units(
```
chrishalcrow, line 207: Can we deprecate `auto_label_units` (slowly!)? I think a lot of people have copies of unitRefine notebooks using it!

src/spikeinterface/curation/unitrefine_curation.py
```diff
@@ -0,0 +1,76 @@
...
+
+def unitrefine_label_units(
+    sorting_analyzer: SortingAnalyzer,
+    noise_neural_classifier: str | Path | None = "SpikeInterface/UnitRefine_noise_neural_classifier_lightweight",
+    sua_mua_classifier: str | Path | None = "SpikeInterface/UnitRefine_sua_mua_classifier_lightweight",
+):
+    """Label units using UnitRefine, which is a cascade of pre-trained classifiers for
+    noise/neural unit classification and SUA/MUA classification.
+    The noise/neural classifier is applied first to remove noise units,
+    then the SUA/MUA classifier is applied to the remaining units.
```
chrishalcrow:
```suggestion
    """Label units using a cascade of pre-trained classifiers for
    noise/neural unit classification and SUA/MUA classification,
    as shown in the UnitRefine paper (see References).
    The noise/neural classifier is applied first to remove noise units,
    then the SUA/MUA classifier is applied to the remaining units.
```

src/spikeinterface/curation/unitrefine_curation.py
```diff
@@ -0,0 +1,76 @@
...
+
+    Parameters
+    ----------
+    sorting_analyzer : SortingAnalyzer
+        The sorting analyzer object containing the spike sorting results.
+    noise_neural_classifier : str or Path or None, default: "SpikeInterface/UnitRefine_noise_neural_classifier_lightweight"
```
chrishalcrow: I'm scared of making these default. The models are trained with NP data, on data in physical units, using metrics computed using si < 0.104 etc.

I think people should investigate and decide what to use. The model on the si page is also already out of date with the model used in the paper, which is on Anoushka's hfh: https://huggingface.co/AnoushkaJain3/noise_neural_classifier_lightweight

chrishalcrow: I'm worried about people blindly using this without understanding the assumptions that went into it.

alejoe91: The model on Anoushka's HF is the same as the one on SI (by looking at last commit + file sizes). Or am I missing something?

alejoe91: But I agree. Let's put `None` byt default and point to the HF repo and train your own model docs

src/spikeinterface/curation/unitrefine_curation.py
```diff
@@ -0,0 +1,79 @@
...
+    -------
+    labels : pd.DataFrame
+        A DataFrame with unit ids as index and "label"/"probability" as column.
+
+    References
+    ----------
+    The approach is described in [Jain]_.
+    """
+    import pandas as pd
+    import pandas as pd
```
chrishalcrow:
```suggestion
```

src/spikeinterface/curation/tests/test_model_based_curation.py
```diff
@@ -168,3 +168,22 @@ def test_exception_raised_when_metric_params_not_equal(sorting_analyzer_for_cura
...
+
+    labels = unitrefine_label_units(
+        sorting_analyzer_for_curation,
+        noise_neural_classifier="SpikeInterface/UnitRefine_noise_neural_classifier_lightweight",
+        sua_mua_classifier="SpikeInterface/UnitRefine_sua_mua_classifier_lightweight",
+    )
+
+    assert "label" in labels.columns
+    assert "probability" in labels.columns
+    assert labels.shape[0] == len(sorting_analyzer_for_curation.sorting.get_unit_ids())
```
chrishalcrow: Can we add cases where each model is None as well. Just this test repeated but with just `noise_neural_classifier=None`, and  just `sua_mua_classifier=None`.

examples/tutorials/curation/plot_1_automated_curation.py
chrishalcrow:
```suggestion
#   Sharing models, which are Python objects, is complicated.
```
cheeky docs correction

src/spikeinterface/curation/tests/test_model_based_curation.py
alejoe91: done!

alejoe91:
@chrishalcrow extended tests! Ready to review

src/spikeinterface/curation/unitrefine_curation.py
```diff
@@ -0,0 +1,78 @@
+from pathlib import Path
+
+from spikeinterface.core import SortingAnalyzer
+from spikeinterface.curation.model_based_curation import model_based_label_units
+
+
+def unitrefine_label_units(
```
chrishalcrow, line 8: This function only works when you give a huggingface repo id, not for local folders. It's also impossible to deceipher if a string is a local folder path or a repo id. 
Maybe we could check to see if the folder exists and if it doesn't, assume it's the string represents a repo?

src/spikeinterface/curation/unitrefine_curation.py
```diff
@@ -0,0 +1,78 @@
...
+        )
+
+    if noise_neural_classifier is not None:
+        # 1. apply the noise/neural classification and remove noise
+        noise_neuron_labels = model_based_label_units(
+            sorting_analyzer=sorting_analyzer,
+            repo_id=noise_neural_classifier,
+            trust_model=True,
+        )
+        noise_units = noise_neuron_labels[noise_neuron_labels["prediction"] == "noise"]
```
chrishalcrow, line 64: If you give the `noise_neural_classifier` a classifier with different labels (e.g. if you give it a sua/mua classifier) the code still runs. Should we warn or raise if "noise" isn't a label in the `noise_neural_classifier`?

src/spikeinterface/curation/__init__.py
```diff
@@ -20,5 +20,6 @@
 from .sortingview_curation import apply_sortingview_curation
 
 # automated curation
-from .model_based_curation import auto_label_units, load_model
+from .model_based_curation import model_based_label_units, load_model
```
chrishalcrow:
```suggestion
from .model_based_curation import model_based_label_units, load_model, auto_label_units
```
for now...

src/spikeinterface/curation/unitrefine_curation.py
alejoe91: done, also for mua/sua

src/spikeinterface/curation/unitrefine_curation.py
alejoe91: done. Added a function to populate the kwargs of `model_based_label_units`:
- `model_folder`: if exists and is folder
- `model_folder` + `model_name`: if path to `skops` file
- `repo_id`: if doesn't exist locally

src/spikeinterface/curation/model_based_curation.py
```diff
@@ -281,6 +281,18 @@ def auto_label_units(
...
 
 
+def auto_label_units(*args, **kwargs):
+    """
+    Deprecated function. Please use `model_based_label_units` instead.
+    """
+    warnings.warn(
+        "`auto_label_units` is deprecated and will be removed in v0.105.0. "
+        "Please use `model_based_label_units` instead.",
+        DeprecationWarning,
```
chrishalcrow, line 291:
```suggestion
        DeprecationWarning,
        stacklevel=2,
```

PR #4341, opened 2026-01-23
Make peak detection (locally_exclussive, matched_filtering) faster and more accurate.

samuelgarcia:
After playing a bit with the rust language, and trying to reimplement the peak detection algo using "locally_exclussive"
I discover that the actual impelmentation could be fasten a lot (at least 3X faster on my machine).

"matched_filtering" has also beeb changed in this PR.

Also the new implementation is is more accurate for corener cases :
  * when the theshold are not teh same by channel (the comparison must be done on the ratio and not teh value)
  * when value in the sweep are higher but not peak thenself

Also I change the `exclude_sweep_ms=0.1` at many place in the code this was a bad value by default it has to be half on the peak waveforms or a bit more.

### DIFF
--- a/src/spikeinterface/sortingcomponents/peak_detection/locally_exclusive.py
+++ b/src/spikeinterface/sortingcomponents/peak_detection/locally_exclusive.py
@@ -107,12 +107,15 @@ if HAVE_NUMBA:
 
     @numba.jit(nopython=True, parallel=False, nogil=True, fastmath=True)
     def detect_peaks_numba_locally_exclusive_on_chunk(
-        traces, peak_sign, abs_thresholds, exclude_sweep_size, neighbours_mask, 
+        traces,
+        peak_sign,
+        abs_thresholds,
+        exclude_sweep_size,
+        neighbours_mask,
     ):
         num_chans = traces.shape[1]
         num_samples = traces.shape[0]
 
-        
         do_pos = peak_sign in ("pos", "both")
         do_neg = peak_sign in ("neg", "both")
 
@@ -121,15 +124,19 @@ if HAVE_NUMBA:
         for s in range(1, num_samples - 1):
             for chan_ind in range(num_chans):
                 if do_neg:
-                    if (traces[s, chan_ind] <= -abs_thresholds[chan_ind]) and \
-                            (traces[s, chan_ind] < traces[s-1, chan_ind]) and \
-                            (traces[s, chan_ind] <= traces[s+1, chan_ind]):
+                    if (
+                        (traces[s, chan_ind] <= -abs_thresholds[chan_ind])
+                        and (traces[s, chan_ind] < traces[s - 1, chan_ind])
+                        and (traces[s, chan_ind] <= traces[s + 1, chan_ind])
+                    ):
                         peak_mask[s, chan_ind] = True
 
-                if do_pos :
-                    if (traces[s, chan_ind] >= abs_thresholds[chan_ind]) and \
-                            (traces[s, chan_ind] > traces[s-1, chan_ind]) and \
-                            (traces[s, chan_ind] >= traces[s+1, chan_ind]):
+                if do_pos:
+                    if (
+                        (traces[s, chan_ind] >= abs_thresholds[chan_ind])
+                        and (traces[s, chan_ind] > traces[s - 1, chan_ind])
+                        and (traces[s, chan_ind] >= traces[s + 1, chan_ind])
+                    ):
                         peak_mask[s, chan_ind] = True
 
         samples_inds, chan_inds = np.nonzero(peak_mask)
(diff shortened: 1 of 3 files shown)


yger:
Great, this is cool, looking forward to try it out ! But what about the matched filtering ?

src/spikeinterface/sortingcomponents/matching/circus.py
```diff
@@ -626,7 +626,7 @@ def __init__(
         return_output=True,
         templates=None,
         peak_sign="neg",
-        exclude_sweep_ms=0.1,
+        exclude_sweep_ms=0.8,
```
chrishalcrow, line 629: You're using 0.8 for all the internal sorters, but 1.0 for the defaults in `detect_peaks`. Is this deliberate?

src/spikeinterface/sortingcomponents/peak_detection/locally_exclusive.py
```diff
@@ -104,88 +105,78 @@ def compute(self, traces, start_frame, end_frame, segment_index, max_margin):
...
+                if samples_inds[i] - exclude_sweep_size > samples_inds[j]:
+                    next_start = j
+                    continue
+                
+                # search for neighbors with higher amplitudes
+                if neighbours_mask[chan_inds[i], chan_inds[j]]:
+                    # if inside spatial zone ... 
+                    if abs(samples_inds[i] - samples_inds[j]) <= exclude_sweep_size:
+                        # ...and if inside tempral zone ...
+                        value_i = abs(traces[samples_inds[i], chan_inds[i]]) / abs_thresholds[chan_inds[i]]
```
chrishalcrow, line 171: So here, the abs_threshold is proportional to the noise on the channel. What's the advantage of using the noise normalised value? Is this to avoid peaks on very noisy channels?

src/spikeinterface/sortingcomponents/peak_detection/locally_exclusive.py
```diff
@@ -104,88 +105,78 @@ def compute(self, traces, start_frame, end_frame, segment_index, max_margin):
...
-                traces, traces_center, peak_mask, exclude_sweep_size, abs_thresholds, peak_sign, neighbours_mask
-            )
+                if samples_inds[i] + exclude_sweep_size < samples_inds[j]:
+                    break
 
-        if peak_sign in ("neg", "both"):
-            if peak_sign == "both":
-                peak_mask_pos = peak_mask.copy()
+                if samples_inds[i] - exclude_sweep_size > samples_inds[j]:
+                    next_start = j
```
chrishalcrow, line 163:
```suggestion
                    next_start = j+1
```

shave off those nanoseconds ;)

src/spikeinterface/sortingcomponents/peak_detection/locally_exclusive.py
samuelgarcia: This avoid that a peak is eliminated even if the snr is higher.
It was not the case before. This is a marjor change in this PR when the noise_levels is different over channel.
I think this is more correct

src/spikeinterface/sortingcomponents/peak_detection/locally_exclusive.py
samuelgarcia: not sure. why ?
The following j will do it if nessecary.

chrishalcrow: Sorry, I misread

src/spikeinterface/sortingcomponents/peak_detection/locally_exclusive.py
chrishalcrow: Ah, ok. You keep the peak in the local area with the highest snr, instead of the highest peak. Cool cool.

In our data, the noise seems to vary a lot over time. Here's noise levels over time for 3 channels. I guess the scaling between channels looks consistent, so it's probably fine for this.

noise_over_time

src/spikeinterface/sortingcomponents/matching/circus.py
samuelgarcia: More or less it depend if it is in the context of peak detection or peak detection for matching

samuelgarcia: For matching having a bit less is OK because we are detecting peak of the convolution

src/spikeinterface/sortingcomponents/peak_detection/locally_exclusive.py
samuelgarcia: The noise varying over time is another subject I think. Could be implemented ion a secondary PR as an option

samuelgarcia:
merci

PR #4342, opened 2026-01-24
Improve stability of velocity fits in template metrics

ecobost:
In template metrics, `velocity_above` and `velocity_below` estimate how fast an spike moves along the axon/dendrites by fitting a line where x is distance (template_channel_location - location of soma) and y is time (position of the peak at each template_channel compared to the soma expressed in msecs). A small slope means the spike takes a long time to move through the probe (so mm/msecs), a higher slope means the spike moves fast. Currently when a peak position is the same (or very close) along the template channels, VelocityFits() tries to fit a straight vertical line; which gives an infinite slope (in practice, the fitting fails with NaN because X is ill-conditioned). This is pretty common. 

This PR switches the fitting to regress peak_ms (in y) onto channel_distances (in x) and then take 1/slope to obtain the same velocities as before. This is more stable and also justified in the original source "Because the time difference between the trough of adjacent sites could be 0, to avoid inﬁnite numbers, we calculated the inverse of velocity (ms/mm) instead by ﬁtting a regression line to the time of waveform trough at different sites against the distance of the sites relative to soma" (Jia et al, 2020). I also centered the data before fitting the regressor. Centering helps stability and avoids having to learn an intercept which gives the model a bit more robustness.

Here's the velocities calculated with the current method and the new method (proposed in this PR):
Untitled
Predicted velocities are consistent for lower velocities (center of the plot) and more stable at higher velocities. It is also able to make predictions for a lot more cases (this is a 3h neuropixels recording and the proportions of NaNs drops from 0.37 to 0.26). I also manually checked some results (where there were nans before) and they look sensible. 

Overall, I think it's just a sensible change for added stability and better predictions.

### DIFF
--- a/src/spikeinterface/metrics/template/metrics.py
+++ b/src/spikeinterface/metrics/template/metrics.py
@@ -361,7 +361,7 @@ def get_velocity_fits(template, channel_locations, sampling_frequency, **kwargs)
         peak_times_ms_above = np.argmin(template_above, 0) / sampling_frequency * 1000 - max_peak_time
         distances_um_above = np.array([np.linalg.norm(cl - max_channel_location) for cl in channel_locations_above])
         inv_velocity_above, score = fit_line_robust(distances_um_above, peak_times_ms_above)
-        if score > min_r2 and np.abs(inv_velocity_above) > 1e-9:
+        if score > min_r2 and inv_velocity_above != 0:
             velocity_above = 1 / inv_velocity_above
         else:
             velocity_above = np.nan
@@ -376,7 +376,7 @@ def get_velocity_fits(template, channel_locations, sampling_frequency, **kwargs)
         peak_times_ms_below = np.argmin(template_below, 0) / sampling_frequency * 1000 - max_peak_time
         distances_um_below = np.array([np.linalg.norm(cl - max_channel_location) for cl in channel_locations_below])
         inv_velocity_below, score = fit_line_robust(distances_um_below, peak_times_ms_below)
-        if score > min_r2 and np.abs(inv_velocity_below) > 1e-9:
+        if score > min_r2 and inv_velocity_below != 0:
             velocity_below = 1 / inv_velocity_below
         else:
             velocity_below = np.nan


chrishalcrow:
This looks great, I'll have a close look tomorrow :)

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -354,7 +354,8 @@ def get_velocity_fits(template, channel_locations, sampling_frequency, **kwargs)
         channel_locations_above = channel_locations[channels_above]
         peak_times_ms_above = np.argmin(template_above, 0) / sampling_frequency * 1000 - max_peak_time
         distances_um_above = np.array([np.linalg.norm(cl - max_channel_location) for cl in channel_locations_above])
-        velocity_above, _, score = fit_velocity(peak_times_ms_above, distances_um_above)
+        inv_velocity_above, score = fit_line_robust(distances_um_above, peak_times_ms_above)
+        velocity_above = 1 / inv_velocity_above
```
chrishalcrow: Something to improve while we're at it: I think it would be nice to catch the division by zero before it happens, since we're expecting this to fail for numerous units. If we do this, the user doesn't get annoying / uninterruptible division by zero errors
```suggestion
        if inv_velocity_above != 0:
            velocity_above = 1 / inv_velocity_above
        else:
            velocity_above = np.nan
```

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -367,7 +368,8 @@ def get_velocity_fits(template, channel_locations, sampling_frequency, **kwargs)
         channel_locations_below = channel_locations[channels_below]
         peak_times_ms_below = np.argmin(template_below, 0) / sampling_frequency * 1000 - max_peak_time
         distances_um_below = np.array([np.linalg.norm(cl - max_channel_location) for cl in channel_locations_below])
-        velocity_below, _, score = fit_velocity(peak_times_ms_below, distances_um_below)
+        inv_velocity_below, score = fit_line_robust(distances_um_below, peak_times_ms_below)
+        velocity_below = 1 / inv_velocity_below
```
chrishalcrow:
```suggestion
        if inv_velocity_below != 0:
            velocity_below = 1 / inv_velocity_below
        else:
            velocity_below = np.nan
```

ecobost:
Hi @chrishalcrow, thanks for having a look at this

I found out that sklearn's TheilSenRegressor requires fit_intercept=True (otherwise it fits lines without intercept y=wx with a single datapoint rather than two datapoints, pretty odd behavior and not what we want as the entire dataset might have intercept 0 but any particular pair of datapoints does not) so I changed that back in this commit https://github.com/SpikeInterface/spikeinterface/pull/4342/commits/0630977f807ae052270a70a3f3000762d54071c3 

I then realized that because this regressor is designed to work on multi-dimensional input, it actually returns an estimate of the median (the spatial median across the parameter space) rather than the median slope (closed-form solution when num_features=1, our case); that is inexact and more time-consuming (as it has to solve a small optimization problem). So I just implemented the algorithm myself (it is pretty straightforward and hopefully easy to understand), it returns almost the same results as sklearn but it's 5-7x faster (and exact).
image

Overall, this should produce better more stable fits. Here's the same plot as above (with velocity computed as in the current code in x axis vs proposed 1/inv_velocity), kind of just looks the same
image

re: your question. I work with NHP 1.0 probes so a unit template is essentially a single column with channels 20-microns apart, which is the canonical use case for this method. Assumptions are that the neuron runs parallel to your depth dimension (y by default, i.e., along the length of the probe for neuropixels) and that channels are nearby in x (code has a column range to limit the x breadth if need be). One could compute a 2-d propagation speed tracking how the spike moves in x and y (think of having a 2-d heatmap of spike delays) but it will require some extra thought

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -284,21 +284,27 @@ def sort_template_and_locations(template, channel_locations, depth_direction="y"
...
-    # from scipy.stats import linregress
-    # slope, intercept, _, _, _ = linregress(peak_times, channel_dist)
+    import itertools
 
-    from sklearn.linear_model import TheilSenRegressor
+    # Calculate slope and bias using Theil-Sen estimator
+    slopes = []
+    for (x0, y0), (x1, y1) in itertools.combinations(zip(x, y), 2):
+        if np.abs(x1 - x0) > eps:
+            slopes.append((y1 - y0) / (x1 - x0))
```
chrishalcrow, line 297:
```suggestion
    slopes.append((y1 - y0) / (x1 - x0))
```

I'm pretty sure this is impossible: we don't allow probes with equal channel locations. (And it never happens for my data!)

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -284,21 +284,27 @@ def sort_template_and_locations(template, channel_locations, depth_direction="y"
...
 
-    theil = TheilSenRegressor()
-    theil.fit(peak_times.reshape(-1, 1), channel_dist)
-    slope = theil.coef_[0]
-    intercept = theil.intercept_
-    score = theil.score(peak_times.reshape(-1, 1), channel_dist)
-    return slope, intercept, score
+    # Calculate R2 score
+    y_pred = slope * x + bias
+    r2_score = 1 - ((y - y_pred) ** 2).sum() / (((y - y.mean()) ** 2).sum() + eps)
```
chrishalcrow, line 305:
```suggestion
    r2_score = 1 - ((y - y_pred) ** 2).sum() / (((y - y.mean()) ** 2).sum())
```

I think the `eps` shouldn't be here

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -354,8 +360,10 @@ def get_velocity_fits(template, channel_locations, sampling_frequency, **kwargs)
         channel_locations_above = channel_locations[channels_above]
         peak_times_ms_above = np.argmin(template_above, 0) / sampling_frequency * 1000 - max_peak_time
         distances_um_above = np.array([np.linalg.norm(cl - max_channel_location) for cl in channel_locations_above])
-        velocity_above, _, score = fit_velocity(peak_times_ms_above, distances_um_above)
-        if score < min_r2:
+        inv_velocity_above, score = fit_line_robust(distances_um_above, peak_times_ms_above)
+        if score > min_r2:
+            velocity_above = 1 / inv_velocity_above if np.abs(inv_velocity_above) > 1e-9 else np.inf
```
chrishalcrow:
```suggestion
            velocity_above = 1 / inv_velocity_above if np.abs(inv_velocity_above) > 1e-9 else np.nan
```

`np.inf` is more correct, but it causes some problems downstream with pandas. To allow `np.inf` we'd need to update some other code and add some tests to make sure everything was robust to `inf`s.

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -367,8 +375,10 @@ def get_velocity_fits(template, channel_locations, sampling_frequency, **kwargs)
         channel_locations_below = channel_locations[channels_below]
         peak_times_ms_below = np.argmin(template_below, 0) / sampling_frequency * 1000 - max_peak_time
         distances_um_below = np.array([np.linalg.norm(cl - max_channel_location) for cl in channel_locations_below])
-        velocity_below, _, score = fit_velocity(peak_times_ms_below, distances_um_below)
-        if score < min_r2:
+        inv_velocity_below, score = fit_line_robust(distances_um_below, peak_times_ms_below)
+        if score > min_r2:
+            velocity_below = 1 / inv_velocity_below if np.abs(inv_velocity_below) > 1e-9 else np.inf
```
chrishalcrow:
```suggestion
            velocity_below = 1 / inv_velocity_below if np.abs(inv_velocity_below) > 1e-9 else np.nan
```

chrishalcrow:
Hello, my results are fairly different for the new implementation (sometimes a x2 difference near velocity of 1000: figure below is `velocity_above`), and I went down a rabbit hole this morning... 

Screenshot 2026-01-28 at 12 59 41

I think your PR fixes something more fundamental than just stability. The Theil-Sen regressor is robust to outliers in y, not in x. For the velocity fits, the distances come from channel locations and are very easy to compute while the peak times are more noisy (since they are to do with peak locations on multiple templates: these can be hard to find when you're further from the signal). Since the peak times are noisier, they should be treated like the independent variable in the regression. This PR makes this change.

So: I would say this PR makes quite a big change in the metric computation but it fixes an error in the previous implementation.

src/spikeinterface/metrics/template/metrics.py
ecobost: I put it there because it's part of the original algorithm, i.e., for every pair of points with distinct x coordinates, do ... The xs, in this case, are the distances to the soma channel, which could theoretically be the same for some probes, no? say if soma is at (0,0) and other channels are at (0, -10) (0, +10) or (10, -10), (10, +10). Array probes might hit that edge case and there might be some other ones. It won't be a common issue for sure but I think it's worth letting that check there to avoid any division by zero.

src/spikeinterface/metrics/template/metrics.py
ecobost: If the y is flat (say, all 0s, which is relatively common, happened twice in my recording) the denominator will be 0. In that case, the numerator will also be 0 and then the r2score will be 1 -0/eps = 1 (perfect prediction with slope 0). Without the epsilon, it will produce r2_scores of NaN (and raise a warning).

src/spikeinterface/metrics/template/metrics.py
ecobost: done

src/spikeinterface/metrics/template/metrics.py
chrishalcrow: Oh! You're right, thanks! Best to leave it in.
(For my own benefit): My NP data doesn't find any examples of this because it only takes channels above or below at once. And the distances between the soma channel and all channels above it on the probe are all unique. This is not the case in probes laid out like this:

```
x x x
 x x
x x x
```

src/spikeinterface/metrics/template/metrics.py
```diff
@@ -354,8 +360,10 @@ def get_velocity_fits(template, channel_locations, sampling_frequency, **kwargs)
         channel_locations_above = channel_locations[channels_above]
         peak_times_ms_above = np.argmin(template_above, 0) / sampling_frequency * 1000 - max_peak_time
         distances_um_above = np.array([np.linalg.norm(cl - max_channel_location) for cl in channel_locations_above])
-        velocity_above, _, score = fit_velocity(peak_times_ms_above, distances_um_above)
-        if score < min_r2:
+        inv_velocity_above, score = fit_line_robust(distances_um_above, peak_times_ms_above)
+        if score > min_r2 and np.abs(inv_velocity_above) > 1e-9:
```
chrishalcrow: Maybe the `eps` value here should be an argument of the function? Is there a reason to pick `1e-9?

ecobost: Action potentials (according to ChatGPT) can go from 500-120000 um/ms. 1/1e-9 = 1B which is probably a mistake. As an argument, I guess you could control max speeds allowed (1e-5 will limit at 100K), but it seems a bit round about. To me, it's just a slightly more robust inv_velocity_above == 0

alejoe91: I would actually just be mathematically consistent here and protect against `inv_velocity_above == 0`. It would be up to the user and downstream analysis to filter out velocities that are too high

Issue #4310, opened 2026-01-09
Make soft merges/split faster


chrishalcrow:
I've been implementing "live" curations in si-gui (https://github.com/SpikeInterface/spikeinterface-gui/pull/224), and to make it more usable it would be great to speed up soft merges/splits. So this is an issue to try and direct some optimisation work! Just very quick and dirty: I benchmarked a single two-unit merge on a typical 1-hour NP2 recording. Here are the times to re-compute each extension, with metrics further split up. It's not all extensions, just the ones I use in gui.

<img width="464" height="526" alt="Image" src="https://github.com/user-attachments/assets/84a070a4-6a2e-478e-8f26-32e2abe493e1" />

Times for each extension re-compute in seconds:

template_similarity,3.515722459065730
firing_rate,1.814500082982700
synchrony,1.3482307499507400
rp_violation,0.54650704190135
waveforms,0.1620308329584080
amplitude_cutoff,0.10835533309727900
templates,0.1060247499262910
correlograms,0.061571332975290700
amplitude_median,0.049069040920585400
spike_locations,0.02714691695291550
snr,0.016673249891027800
num_spikes,0.0160850411048159
amplitude_cv,0.012816166039556300
sliding_rp_violation,0.011563500040210800
unit_locations,0.010269250022247400
spike_amplitudes,0.004084207932464780
firing_range,0.0012063339818269000
repolarization_slope,0.000768749974668026
presence_ratio,0.0004059579223394390
recovery_slope,0.00016670895274728500
isi_violation,0.00012345798313617700
random_spikes,8.01659189164639E-05
sd_ratio,4.77919820696116E-05
half_width,1.93340238183737E-05
noise_levels,1.19999749585986E-05
peak_to_valley,1.12500274553895E-05
peak_trough_ratio,5.79189509153366E-06

(total = 7 seconds. Note: this doesn't scale linearly, two merges take much less time than 2 times one merge)

Good news: the slow ones seem reasonable to optimize! For a soft template (a linear combination of old templates) template similarity can be computed from already-computed info. Firing rate and rp_violation are slow because we re-compute for all units (which we don't need to do). synchrony should be numbafy-able.

Now a single split!

<img width="440" height="525" alt="Image" src="https://github.com/user-attachments/assets/acb84f02-c0d6-4a52-b043-9eadf8795e5b" />

template_similarity,4.082042583962900
correlograms,2.585401250049470
firing_rate,1.890601999941280
[...]

PR #4343, opened 2026-01-26
Speed up template similarity computing using numba

tayheau:
Following #4310, @chrishalcrow showed that given that the diff between two merged templates is considered small, we can approximate the distance of a megred template and a new one a as a linear function. This should allow us to speed up significantly template similarity computations for merged ones.

### DIFF
--- a/src/spikeinterface/postprocessing/template_similarity.py
+++ b/src/spikeinterface/postprocessing/template_similarity.py
@@ -226,6 +226,8 @@ def _compute_similarity_matrix_numpy(
             overlapping_templates = np.flatnonzero(np.sum(local_mask, 1))
             tgt_templates = tgt_sliced_templates[overlapping_templates]
             for gcount, j in enumerate(overlapping_templates):
+                if j < i and same_array:
+                    continue
                 src = src_template[:, local_mask[j]].reshape(1, -1)
                 tgt = (tgt_templates[gcount][:, local_mask[j]]).reshape(1, -1)
 
@@ -247,7 +249,10 @@ def _compute_similarity_matrix_numpy(
                     distances[count, i, j] = 1 - distances[count, i, j]
 
                 if same_array:
-                    distances[num_shifts_both_sides - count - 1, j, i] = distances[count, i, j]
+                    distances[count, j, i] = distances[count, i, j]
+
+        if same_array and shift != 0:
+            distances[num_shifts_both_sides - count - 1] = distances[count].T
 
     return distances
 


tayheau:
So everything is supposed to work fine, i think i have some differences mostly due to casting but it's "minimal" in the context of a normalized distance.
image


Just to quickly sum up i moved the computing of the support matrix out of the loop, it will be more 'memory' costly in case of different template matrix but i think since most of the time we have the same ones it's a boost here. Also more leverage of numpy views instead of copies. 

Im not this familiar with the numba/numpy efficiency stuff but for the union it was faster to do it the "dummy" way (mine lol) that the vectorised one. So if you guys have so rule of thumbs tips for Numba, im in ;) @samuelgarcia 

image

chrishalcrow:
Looks great as a user! I get a small ~30% speed up on initial compute, and 10x speed-up on recompute (=> trying out different methods in the gui is super fast, and hard merges are very fast).

And on some real data with 800 neurons, the biggest absmax in difference is...
```
np.max(np.abs(old_temps - new_temps))
>> np.float32(4.2915344e-06)
```

Nice!!!

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -265,21 +266,24 @@ def _compute_similarity_matrix_numpy(
...
-
         num_shifts_both_sides = 2 * num_shifts + 1
+
         distances = np.ones((num_shifts_both_sides, num_templates, other_num_templates), dtype=np.float32)
         same_array = np.array_equal(templates_array, other_templates_array)
 
-        # We can use the fact that dist[i,j] at lag t is equal to dist[j,i] at time -t
-        # So the matrix can be computed only for negative lags and be transposed
         if same_array:
-            # optimisation when array are the same because of symetry in shift
```
chrishalcrow, line 275: I'd leave the comments, the calculation is pretty confusing...

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -288,77 +295,91 @@ def _compute_similarity_matrix_numba(
...
+                        dot = np.sum(src_ch * tgt_ch)
+                        norm_i = sqrt(np.sum(src_ch**2))
+                        norm_j = sqrt(np.sum(tgt_ch**2))
+                        denom = norm_i * norm_j
+                        if denom > 0.0:
+                            distances[count, src_unit, tgt_unit] = 1.0 - dot / denom
 
                     if same_array:
-                        distances[num_shifts_both_sides - count - 1, j, i] = distances[count, i, j]
+                        distances[count, tgt_unit, src_unit] = distances[count, src_unit, tgt_unit]
```
yger: distances[num_shifts_both_sides - count - 1, j, i] = distances[count, i, j]

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -288,77 +295,91 @@ def _compute_similarity_matrix_numba(
...
+
+        for src_unit in range(num_templates):
+            overlapping_ids = numba.typed.List()
+            overlapping_chs = numba.typed.List()
 
-                ## Ideally we would like to use this but numba does not support well function with numpy and boolean arrays
-                ## So we inline the function here
-                # local_mask = get_overlapping_mask_for_one_template(i, sparsity, other_sparsity, support=support)
+            for tgt_unit in range(other_num_templates):
+                if same_array and tgt_unit < src_unit:
```
yger: This optimization should be removed (see #4345), because otherwise the computations are not correct

yger:
I would be carefull here, because we recently patched a bug in #4345 and I think this is not propagated here

src/spikeinterface/postprocessing/template_similarity.py
chrishalcrow: Are these comments related to the over-optimization that you've now gotten rid of?

tayheau: Yep, Will remove them, mb 😅

alejoe91:
@yger I think that the symmetry issue is fixed. Can you double check?

src/spikeinterface/postprocessing/template_similarity.py
yger: I don't think this line is correct, is it? we will loop over all (src_unit, tgt_unit) indices and thus no need to enforce symmetry, am I right? My point previsously was that we have cc(i,j,t) = c(j,i, N-t)
but if we loop for all (i,j), then as before we only need to symmetrize in time. and we need to symmetrize in time, because we only loop over half of the temporal indices. But what happen if same_array is False? I guess our optimizations might not be correct, I'll check

src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -288,77 +295,90 @@ def _compute_similarity_matrix_numba(
             metric = 0
         elif method == "l2":
             metric = 1
-        elif method == "cosine":
+        else:
```
samuelgarcia:
```suggestion
         elif method == "cosine":
```

explicit is easier for maintenance

samuelgarcia:
Salut Theo.
We check it with Pierre and we have push a small commit.
You did the same mistake pierre did some time ago. Now it is correct.

samuelgarcia:
This is now OK for me.
I am not sure that the speedup will be as big as @chrishalcrow has seen.
Could you check ?

tayheau:
salut Samuel, 
but with the fix you did, won't the array only half populated in the case of `same_array` ?

yger:
No because if this is not the same array, we'll explore all shifts, am I right?

tayheau:
to me there two symmetries :  a "spatial" one where we can say dist(i, j) = dist(j, i) at lag t and a temporal one where dist(i, j) at lag t = dist(i, j) at lag -t (we do agree that both apply only in case of `same_array`) . So that's why we check twice the matrix similarity, and for num_shifts != 0 it's just so that it doesnt erase with the transpose but it should be shift  != 0. here i think this fix erease the 'spatial' symmetry optimisation

yger:
You are right. Can we make that together with a small call ?

yger:
@tayheau if you can finish today and propagate changes to compute_(...)_numpy this would be great, since we would like to merge that quickly for release. Otherwise let me know and I can have a go on the branch

yger:
As told to @samuelgarcia this is good to me, expect that currently optimizations are only performed at the numba level, not yet in pure numpy (in case numba is not present)

yger:
I'm not able to push in your branch, but here is a code to change _compute_similarity_matrix_numpy()

```
def _compute_similarity_matrix_numpy(
    templates_array, other_templates_array, num_shifts, method, sparsity_mask, other_sparsity_mask, support="union"
):

    num_templates = templates_array.shape[0]
    num_samples = templates_array.shape[1]
    other_num_templates = other_templates_array.shape[0]

    num_shifts_both_sides = 2 * num_shifts + 1
    distances = np.ones((num_shifts_both_sides, num_templates, other_num_templates), dtype=np.float32)
    same_array = np.array_equal(templates_array, other_templates_array)

    # We can use the fact that dist[i,j] at lag t is equal to dist[j,i] at time -t
    # So the matrix can be computed only for negative lags and be transposed

    if same_array:
        # optimisation when array are the same because of symetry in shift
        shift_loop = range(-num_shifts, 1)
    else:
        shift_loop = range(-num_shifts, num_shifts + 1)

    for count, shift in enumerate(shift_loop):
        src_sliced_templates = templates_array[:, num_shifts : num_samples - num_shifts]
        tgt_sliced_templates = other_templates_array[:, num_shifts + shift : num_samples - num_shifts + shift]
        for i in range(num_templates):
            src_template = src_sliced_templates[i]
            local_mask = get_overlapping_mask_for_one_template(i, sparsity_mask, other_sparsity_mask, support=support)
            overlapping_templates = np.flatnonzero(np.sum(local_mask, 1))
            tgt_templates = tgt_sliced_templates[overlapping_templates]
            for gcount, j in enumerate(overlapping_templates):
                if j < i and same_array:
                    continue
                src = src_template[:, local_mask[j]].reshape(1, -1)
                tgt = (tgt_templates[gcount][:, local_mask[j]]).reshape(1, -1)

                if method == "l1":
                    norm_i = np.sum(np.abs(src))
                    norm_j = np.sum(np.abs(tgt))
                    distances[count, i, j] = np.sum(np.abs(src - tgt))
                    distances[count, i, j] /= norm_i + norm_j
                elif method == "l2":
                    norm_i = np.linalg.norm(src, ord=2)
                    norm_j = np.linalg.norm(tgt, ord=2)
                    distances[count, i, j] = np.linalg.norm(src - tgt, ord=2)
                    distances[count, i, j] /= norm_i + norm_j
                elif method == "cosine":
                    norm_i = np.linalg.norm(src, ord=2)
                    norm_j = np.linalg.norm(tgt, ord=2)
                    distances[count, i, j] = np.sum(src * tgt)
                    distances[count, i, j] /= norm_i * norm_j
                    distances[count, i, j] = 1 - distances[count, i, j]

                if same_array:
                    distances[count, j, i] = distances[count, i, j]

        if same_array and shift != 0:
            distances[num_shifts_both_sides - count - 1] = distances[count].T

    return distances
```

samuelgarcia:
Merci Theo.
Désolé pour l'intrusion inutile!!
je merge.

tayheau:
pas de soucis !
c'est fait pour ça aussi git hehe

PR #4345, opened 2026-01-28
Patch lag similarities

yger:
This PR adds the handling of lags in the template_similarity extension, including while selecting/merging, ...

Plus, this patches an important bug while computing similarities with lags. Currently, we were optimizing too aggressively the loop making use that CC(i,j,t) = CC(j,i,n-t) . But because we were only computing for i < j and for negative lags, the matrix was not correct...

This is not a major bug since the default lag is 0, and not many pipelines were making use of such lags, except spike sorting pipelines

### DIFF
--- a/src/spikeinterface/postprocessing/template_similarity.py
+++ b/src/spikeinterface/postprocessing/template_similarity.py
@@ -198,6 +198,7 @@ class ComputeTemplateSimilarity(AnalyzerExtension):
     def _get_data(self):
         return self.data["similarity"]
 
+
 # @alessio:  compute_template_similarity() is now one inner SortingAnalyzer only
 register_result_extension(ComputeTemplateSimilarity)
 compute_template_similarity = ComputeTemplateSimilarity.function_factory()


src/spikeinterface/postprocessing/template_similarity.py
```diff
@@ -194,10 +213,14 @@ def _run(self, verbose=False):
             other_sparsity=sparsity,
         )
         self.data["similarity"] = similarity
+        self.data["lags"] = lags
 
     def _get_data(self):
         return self.data["similarity"]
 
+    def get_lags(self):
+        return self.data["lags"]
```
samuelgarcia: if you add a new data in this extension you need to handle the backward compatibility by implementing

    need_backward_compatibility_on_load = True

    def _handle_backward_compatibility_on_load(self):

samuelgarcia:
This is OK for me. But we need to handle the backward compatibility because lags are now stor in the data.

src/spikeinterface/postprocessing/template_similarity.py
yger: This is now handled. If lags are absent, I put a matrix of 0 instead.

PR #4353, opened 2026-01-29
Compute missing spikes on the right tail of amplitude distributions 

ecobost:
Fixes https://github.com/SpikeInterface/spikeinterface/issues/4352-

I also refactored the code to be more legible and fixed a couple of details:
1. use mode='nearest' during smoothing, default of reflect causes the edge to be higher and the cutoff will be overestimated.
2. Fix how the fraction is computed to be num_missing / (total_recorded + num_missing) vs. num_missing/total_recorded (as it was before) which also overestimates the cutoff.
3. Increase the default bin size to 200, 100 is a bit too high variance.

Here's the output for some examples (at 15% cutoff now):
image
The right missing amplitudes (for the middle example) are now correctly estimated to ~15%. Left missing amplitudes say 0 (correctly) as that is not where the missing spikes are expected to be (but rather at low absolute values, so the right)

### DIFF
--- a/src/spikeinterface/metrics/quality/misc_metrics.py
+++ b/src/spikeinterface/metrics/quality/misc_metrics.py
@@ -1508,7 +1508,7 @@ def isi_violations(spike_trains, total_duration_s, isi_threshold_s=0.0015, min_i
 
 def amplitude_cutoff(amplitudes, num_histogram_bins=500, histogram_smoothing_value=3, amplitudes_bins_min_ratio=5):
     """
-    Calculate approximate fraction of spikes missing from the right tail of a distribution of amplitudes.
+    Calculate approximate fraction of spikes missing from a distribution of amplitudes.
 
     Find the missing spikes from the left tail of the distribution. Assumes cutoff happens at spikes
     with lower amplitudes.


chrishalcrow:
Hello, thanks so much for finding this - I agree with your results. There's a similar metric called `noise_cutoff` that has a test which is meant to check for this error.

https://github.com/SpikeInterface/spikeinterface/blob/f732780fd88f5802033b57c9bb9b06229ec7de30/src/spikeinterface/metrics/quality/tests/test_metrics_functions.py#L118

A similar, test for amp cutoff might be:

``` python
def test_amplitude_cutoff():
    """
    Generate two artificial gaussians, one truncated and one not. Check the metrics are higher for the truncated one.
    """
    np.random.seed(1)
    amplitudes = np.random.normal(-3, 1, 2000)
    amplitudes_trunc = amplitudes[amplitudes < -4]

    fraction_missing_1 = amplitude_cutoff(amplitudes=amplitudes, num_histogram_bins=20)
    fraction_missing_2 = amplitude_cutoff(amplitudes=amplitudes_trunc, num_histogram_bins=20)

    assert fraction_missing_1 < fraction_missing_2
```

Could you check that and add this in please?

ecobost:
Hi @chrishalcrow 

Test passed locally, added it to the PR.

ecobost:
Updated PR so both amplitude_cutoff and noise_cutoff receive positive amplitudes (so cutoff is correctly computed on the left tail of the distribution). I still refactored `amplitude_cutoff` to fix the minor details I listed in the original PR comment.

Additionally, I changed to invert the amplitudes per unit as, when `analyzer.compute('spike_amplitudes', peak_sign='both')` is used some units have positive amplitude values and some negative ones. 

@chrishalcrow Sorry for the double reviews (I just today noticed that noise_cutoff also has the same assumption of the cutoff being on the left/lower side, so i think this way to deal with it is probably better)

PR #4355, opened 2026-01-30
Correct formula for mad (and std) in drift quality metrics

ecobost:
Three small changes to drift metrics:
1. Changed mean in mad to median (to conform to standard mad definition)
2. Get rid of np.abs inside the std calculation. Before, one branch (when drift has nans) computed std(abs(drift)) and the other computed std(drift), I presume we only really just want the std(drift).
3. Drop the unnecessary outer nan check; nanmean(x) == mean(x) if x has no nans so no use to have that as a separate check.

Tested locally, mad and some std values change, rest remain the same.


alejoe91 (approved):
Thanks! Looks great!

PR #4357, opened 2026-02-02
Binary recording bacward-compatibility

alejoe91:
We removed the `num_chan` argument from the `BinaryRecordingExtractor`, which makes old binary folders not loadable anymore with the current version.

This PR adds it back (but doesn't add it to the docstring).

@magland with this PR you should be able to use the current version to load the compression benchmark data

### DIFF
--- a/src/spikeinterface/core/binaryrecordingextractor.py
+++ b/src/spikeinterface/core/binaryrecordingextractor.py
@@ -71,6 +71,7 @@ class BinaryRecordingExtractor(BaseRecording):
         else:
             assert len(channel_ids) == num_channels, "Provided recording channels have the wrong length"
 
+        # DO NOT DELETE! This is for backward compatibility
         if num_chan is not None:
             assert num_channels is None, "When both num_channels and num_chan are provided, num_channels is used"
             num_channels = num_chan


src/spikeinterface/core/binaryrecordingextractor.py
```diff
@@ -55,21 +55,26 @@ def __init__(
...
         is_filtered=None,
+        num_chan=None,
     ):
 
         if channel_ids is None:
             channel_ids = list(range(num_channels))
         else:
             assert len(channel_ids) == num_channels, "Provided recording channels have the wrong length"
 
+        if num_chan is not None:
```
chrishalcrow, line 75: Should we add a comment or a docstring, so that we don't accidentally delete it again?

alejoe91, line 75:
```suggestion
        # DO NOT DELETE! This is for backward compatibility
        if num_chan is not None:
```

alejoe91: done

chrishalcrow (approved):

Issue #4346, opened 2026-01-28
feature request: to_pynapple_tsgroup export Sorting properties (like SortingAnalyzer metadata)


pas-calc:
The function `to_pynapple_tsgroup` (from `spikeinterface.exporters`) can export a `BaseSorting` or `SortingAnalyzer` to pynapple time series group (spike trains).

It is nice that it can also export metadata from SortingAnalyzer. However it currently does not support exporting properties from Sorting. For example when we read a sorting from Kilosort, the properites are automatically filled from the csv/tsv files.

For instance `sorting.get_property("ch")` will be the channel on which the neuron showed largest waveform.

For a sortinganalyzer run on that sorting+recording, we can get quality metrics. They will be similar to those precomputed by a sorter (like refractory period contamination, firing rate, ... )

For a sorting object, we could `get_property_keys` and then add the properties to the  pynapple `TsGroup` metadata

chrishalcrow:
Hello, that sounds good.

I think we'd just need to add a loop over the properties here: 
https://github.com/SpikeInterface/spikeinterface/blob/456265c9d48e2e749861660c4bf0e3760027f6a0/src/spikeinterface/exporters/to_pynapple.py#L81
and add them to the `metadata_list` as a dataframe. Sorting properties must be the length of the number of units, so should all be ok.

If you're willing to make a PR, I'll review it :)

pas-calc:
On the level of `BaseSorting` we can have a function like
```
def get_property_df(sorting):
    unit_ids = sorting.unit_ids
    property_df = pd.DataFrame(index=unit_ids)
    property_keys = sorting.get_property_keys()
    for property_key in property_keys:
        property_data = sorting.get_property(property_key)
        property_df[property_key] = property_data
    return property_df
```

PR #4360, opened 2026-02-02
to_pynapple_tsgroup : attach_unit_properties, export sorting properties

pas-calc:
Added option to attach unit properties from sorting to the TsGroup object.

https://github.com/SpikeInterface/spikeinterface/issues/4346

### DIFF
--- a/src/spikeinterface/exporters/to_pynapple.py
+++ b/src/spikeinterface/exporters/to_pynapple.py
@@ -62,8 +62,8 @@ def to_pynapple_tsgroup(
         for unit_id_int, unit_id in zip(unit_ids_ints, unit_ids)
     }
 
-    metadata_list = [] # init list to collect metadata dataframes
-    
+    metadata_list = []  # init list to collect metadata dataframes
+
     if not unit_ids_castable:
         metadata_list.append(pd.DataFrame(unit_ids, columns=["unit_id"]))
 
@@ -84,7 +84,7 @@ def to_pynapple_tsgroup(
     # attach unit properties from sorting
     if attach_unit_properties:
         property_df = pd.DataFrame(index=unit_ids)
-        property_keys = sorting.get_property_keys() # get property keys of sorting
+        property_keys = sorting.get_property_keys()  # get property keys of sorting
         if len(property_keys):
             for property_key in property_keys:  # loop through sorting's properties
                 property_data = sorting.get_property(property_key)


src/spikeinterface/exporters/to_pynapple.py
```diff
@@ -8,6 +8,7 @@
 def to_pynapple_tsgroup(
     sorting_analyzer_or_sorting: SortingAnalyzer | BaseSorting,
     attach_unit_metadata=True,
+    attach_unit_properties=False,
```
chrishalcrow: Let's go `True` by default - it's a great addition :)

src/spikeinterface/exporters/to_pynapple.py
```diff
@@ -79,6 +82,21 @@ def to_pynapple_tsgroup(
         if (template_metrics := sorting_analyzer_or_sorting.get_extension("template_metrics")) is not None:
             metadata_list.append(template_metrics.get_data())
 
+    # get the underlying sorting
+    if isinstance(sorting_analyzer_or_sorting, SortingAnalyzer):
+        sorting = sorting_analyzer_or_sorting.sorting  # use the sorting of SortingAnalyzer
+    elif isinstance(sorting_analyzer_or_sorting, BaseSorting):
+        sorting = sorting_analyzer_or_sorting  # already a Sorting instance
```
chrishalcrow: This already happens at lines 38-43, so we can get rid of this
```suggestion
```

src/spikeinterface/exporters/to_pynapple.py
```diff
@@ -79,6 +82,21 @@ def to_pynapple_tsgroup(
...
+
+    # attach unit properties from sorting
+    if attach_unit_properties:
+        property_df = pd.DataFrame(index=unit_ids)
+        property_keys = sorting.get_property_keys()
+        for property_key in property_keys:  # loop through sorting's properties
+            property_data = sorting.get_property(property_key)
+            property_df[property_key] = property_data
```
chrishalcrow: When trying this out on my data, I realised that we allow the sorting properties to be _anything_. E.g. it could be an array like

```
xyz_location = [
    [1,2,3],
    [1,2,6],
    [2,1,7],
]
```

And this line will error in this case - the input needs to be a a flat array/list. Maybe we could check this here. So get the `property_data`, make a numpy array, check its `ndim` and only put it in the `property_df` if `ndim` is 1? Another option is to put `property_df[property_key] = property_data` inside a try/except, though we usually try to avoid these.

pas-calc: or we cast it to `list(property_data)`  (from numpy array),  then assigning it on the dataframe indexes will work like expected
`property_df[property_key] = list(property_data)`

src/spikeinterface/exporters/to_pynapple.py
pas-calc: yes, of course, somehow missed that

samuelgarcia:
Cool!

src/spikeinterface/exporters/to_pynapple.py
```diff
@@ -79,6 +82,15 @@ def to_pynapple_tsgroup(
...
             metadata_list.append(template_metrics.get_data())
 
+    # attach unit properties from sorting
+    if attach_unit_properties:
+        property_df = pd.DataFrame(index=unit_ids)
+        property_keys = sorting.get_property_keys()
+        for property_key in property_keys:  # loop through sorting's properties
+            property_data = sorting.get_property(property_key)
+            property_df[property_key] = list(property_data)
+        metadata_list.append(property_df)
```
chrishalcrow: The tests are failing when there are no sorting properties. I would put these four lines inside a `if len(property_keys) > 0`, so that the `metadata_list` is only updated if there are some property keys.

src/spikeinterface/exporters/to_pynapple.py
chrishalcrow:
```suggestion
```
I noticed that this `metadata_list = []` is not needed! A bug from me when I first coded this. Could you remove it please - thanks!

src/spikeinterface/exporters/to_pynapple.py
chrishalcrow: Great idea!

src/spikeinterface/exporters/to_pynapple.py
pas-calc: yes, this is superfluous as it was defined above, actually it could reset the metadata if the original clu ids were already written to it

src/spikeinterface/exporters/to_pynapple.py
pas-calc: yes, this can be easily wrapped, I thought the loop would just not run if property_keys was an empty list, and then an empty df would be concat later

PR #4362, opened 2026-02-02
Add peak_sign to sd_ratio

ecobost:
sd_ratio uses get_template_extremum_channel to select the "best_channel" for each unit. Similar to other metrics/extensions where this function is used (spike_amplitudes, spike_locations, template_metrics, snr) I think it's worth adding a peak_sign so people can control its behavior. 

Effect is rather minimal because few extremum channels change and nearby channels have similar noise levels anyway but if spike_amplitudes were created with peak_sign='both', it is correct to use peak_sign='both' here too to select the same channel for the std(noise) (as was used for spike_amplitudes).

I also refactored some code for legibility but results are unchanged (for peak_sign='neg', the default).

### DIFF
--- a/src/spikeinterface/metrics/quality/misc_metrics.py
+++ b/src/spikeinterface/metrics/quality/misc_metrics.py
@@ -1264,7 +1264,7 @@ def compute_sd_ratio(
     correct_for_template_itself : bool, default:  True
         If true, will take into account that the template itself impacts the standard deviation of the noise,
         and will make a rough estimation of what that impact is (and remove it).
-    peak_sign : str, ("neg", "pos", "both"), default: "neg"
+    peak_sign : "neg" | "pos" | "both", default: "neg"
         The peak sign used to select the template extremum channel.
     **job_kwargs : dict, default: {}
         Keyword arguments sent to get_noise_levels.


src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -1224,7 +1238,8 @@ def compute_sd_ratio(
     censored_period_ms: float = 4.0,
     correct_for_drift: bool = True,
     correct_for_template_itself: bool = True,
-    **kwargs,
+    peak_sign: str = "neg",
```
samuelgarcia, line 1241: and peak_mode also ?

ecobost: We can. Extensions (that receive peak info) usually only receive peak_sign, for instance, for spike_amplitudes: https://github.com/SpikeInterface/spikeinterface/blob/2bc90064897125b0d7d9bc936c119fa5415b3ec2/src/spikeinterface/postprocessing/spike_amplitudes.py#L28, template metrics: https://github.com/SpikeInterface/spikeinterface/blob/2bc90064897125b0d7d9bc936c119fa5415b3ec2/src/spikeinterface/metrics/template/template_metrics.py#L192 quality_metrics: https://github.com/SpikeInterface/spikeinterface/blob/2bc90064897125b0d7d9bc936c119fa5415b3ec2/src/spikeinterface/metrics/quality/quality_metrics.py#L83 or  spike_locations https://github.com/SpikeInterface/spikeinterface/blob/2bc90064897125b0d7d9bc936c119fa5415b3ec2/src/spikeinterface/postprocessing/spike_locations.py#L79

So I copied them.  The one exception is the SNR metric (inside quality metrics) that does receive peak_sign and peak_mode.

I am ok with either version, As this is temporary anyway, we can maybe keep just peak_sign (and later main_channel will take care of true consistency). Let me know

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -1249,20 +1264,21 @@ def compute_sd_ratio(
     correct_for_template_itself : bool, default:  True
         If true, will take into account that the template itself impacts the standard deviation of the noise,
         and will make a rough estimation of what that impact is (and remove it).
-    **kwargs : dict, default: {}
-        Keyword arguments for computing spike amplitudes and extremum channel.
+    peak_sign : str, ("neg", "pos", "both"), default: "neg"
```
chrishalcrow:
```suggestion
    peak_sign : "neg" | "pos" | "both", default: "neg"
```

Sorry, annoying typing stuff. Am going to change all of these into `Literal` soon, on my to-do list...

src/spikeinterface/metrics/quality/misc_metrics.py
chrishalcrow: I think you might need to add `peak_mode`, it's used in the test suite as an example of metric kwargs changing a result:

https://github.com/SpikeInterface/spikeinterface/blob/81e89fd19cffdbf031e48c54100b6ab9e024e3ca/src/spikeinterface/curation/tests/test_model_based_curation.py#L145

ecobost: @chrishalcrow that's for snr (the only one that actually receives peak_sign and peak_mode); the one I am changing here is sd_ratio

chrishalcrow: Oh sorry, I'm an idiot, thanks!

PR #4363, opened 2026-02-03
Remove multiprocessing from NearestNeigbhor metric

ecobost:
ComputeQualityMetrics has a peak_sign argument (similar to the one in ComputeTemplateMetrics), this PR sends it to get_template_extremum_channel in prepare_data. This affects pca_metrics (if set to "pos" or "both") as it will change how the neighborhood of each unit is defined (l.154), Default behavior, peak_sign='neg', is not changed.

As I was checking pca_metrics, I slightly refactor them for legibility (hopefully line -> line equivalence is easy to see). Results are all the same (except for nn_miss_rate and nn_hit_rate, see below)

The only "major" change was to the nearest_neighbor metrics:
1) Before, it used n-1 neighbors to compute the metrics (because each point was its own neighbor, l.533-534). I changed it to use n neighbors (5 by default).
2) I also changed the nearest neighbor algorithm to "auto, i..e, let sklearn choose; this is good practice and was >10x faster in my tests (though with a small dset). This affects only efficiency (neighbors are, ofc, still the same).

### DIFF
--- a/.github/workflows/all-tests.yml
+++ b/.github/workflows/all-tests.yml
@@ -239,6 +239,19 @@ jobs:
         run: |
           pip install -e .[full]
           pip list
+
+          # Internal shell check for platform and env var
+          if [[ "$RUNNER_OS" == "macOS" ]]; then
+            # Locate the binary dynamically
+            SHM_BINARY=$(python -c "import torch; from pathlib import Path; print(Path(torch.__file__).parent / 'bin' / 'torch_shm_manager')")
+            if [ -f "$SHM_BINARY" ]; then
+              echo "Applying chmod +x to $SHM_BINARY"
+              chmod +x "$SHM_BINARY"
+            else
+              echo "Warning: torch_shm_manager not found at $SHM_BINARY"
+            fi
+          fi
+
           ./.github/run_tests.sh sortingcomponents --no-virtual-env
 
       - name: Test generation
--- a/.github/workflows/deepinterpolation.yml
+++ b/.github/workflows/deepinterpolation.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: '3.9'
+          python-version: '3.10'
       - name: Get changed files
         id: changed-files
         uses: tj-actions/changed-files@v46.0.1
@@ -40,9 +40,8 @@ jobs:
         run: |
           python -m pip install -U pip  # Official recommended way
           # install deepinteprolation
-          pip install tensorflow==2.7.0
+          pip install tensorflow==2.8.4
           pip install deepinterpolation@git+https://github.com/AllenInstitute/deepinterpolation.git
-          pip install protobuf==3.20.*
           pip install numpy==1.26.4
           pip install -e .[full,test_core]
       - name: Pip list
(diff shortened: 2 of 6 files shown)


ecobost:
I was checking the failed tests (test_compute_pc_metrics_multi_processing). Using algorithm='auto' (which will usually choose 'brute' for small data sizes, say, 10K samples by ~100 features) with mp_context = fork will hang forever (thread deadlock inside BLAS/openMP; essentially child processes think already dead threads from their parents are still locking data and wait indefinitely for the lock to be released). I changed it in https://github.com/SpikeInterface/spikeinterface/pull/4363/commits/20c1550c7ce80941c8a013753fc759cf475e2302 to spawn (rather than fork) new processes.

I also tested performance on a smallish recording (225 units, ~700 spikes per unit). Here's the summary:
| nn_algorithm/jobs | 1 | 8 |
|----| ---| --- |
| ball_tree (current) | 431 s | 310 s |
| brute (new) | 27 s | 38 s |

It does not seem like there is much advantage of using multiprocessing (at least with 2-8 processes in my 4-core machine).

I would just go for using algorithm='brute' (or let sklearn choose) and not use multiprocessing at all to avoid any possible future issues with fork/spawn (and the kind of hidden spawn default added here); 'brute' is fast enough to be comparable to other quality metrics (and it does not seem to benefit much from multiprocessing anyway, mainly because the numerical libraries used behind the curtains are already multi-threaded). https://github.com/SpikeInterface/spikeinterface/pull/4363/commits/d44542e4fc712b03fccdb4b657906a5a55e6aab8 has these setup (no multiprocessing), you can pick the one you prefer.

alejoe91:
@samuelgarcia @chrishalcrow this is good to merge for me! Getting rid of the paralleization without any performance hit is great, plus @ecobost made the code much cleaner!

src/spikeinterface/metrics/quality/quality_metrics.py
```diff
@@ -137,7 +137,7 @@ def _prepare_data(self, sorting_analyzer, unit_ids=None):
         all_labels = sorting_analyzer.sorting.unit_ids[spike_unit_indices]
 
         # Get extremum channels for neighbor selection in sparse mode
-        extremum_channels = get_template_extremum_channel(sorting_analyzer)
+        extremum_channels = get_template_extremum_channel(sorting_analyzer, peak_sign=self.params["peak_sign"])
```
chrishalcrow: We'll deal with this in another PR
```suggestion
        extremum_channels = get_template_extremum_channel(sorting_analyzer)
```

src/spikeinterface/metrics/quality/quality_metrics.py
```diff
@@ -80,7 +80,7 @@ def _set_params(
         use_valid_periods=False,
         periods=None,
         # common extension kwargs
-        peak_sign=None,
+        peak_sign="neg",
```
chrishalcrow:
```suggestion
        peak_sign=None,
```
..and this

chrishalcrow:
Hello, tested on real 1hr NP2 data on a linux machine using 8 cores. Computation went from 45 minutes to 2 minutes - thanks @ecobost !!

samuelgarcia:
Give me a chance to read it quickly.

alejoe91:
> Give me a chance to read it quickly.

Sure! You can merge it after you read it!

samuelgarcia:
Hi @ecobost 
This is OK for me.
thanks

PR #4365, opened 2026-02-03
Add `threshold_metrics` curation

alejoe91:
In the spirit of `model_based_label_units` and `bombcell_label_units`, this PR adds `qualitymetrics_label_units`

### DIFF
--- a/src/spikeinterface/curation/threshold_metrics_curation.py
+++ b/src/spikeinterface/curation/threshold_metrics_curation.py
@@ -15,6 +15,7 @@ def threshold_metrics_label_units(
     fail_label: str = "noise",
     operator: str = "and",
     nan_policy: str = "fail",
+    column_name: str = "label",
 ):
     """Label units based on metrics and thresholds.
 
@@ -39,11 +40,13 @@ def threshold_metrics_label_units(
         If "ignore", NaN values will be ignored. Note that the "ignore" behavior will depend on the operator used.
         If "and", NaNs will be treated as passing, since the initial mask is all true;
         if "or", NaNs will be treated as failing, since the initial mask is all false.
+    column_name : str, default: "label"
+        The name of the column in the output DataFrame that will contain the assigned labels.
 
     Returns
     -------
     labels : pd.DataFrame
-        A DataFrame with unit IDs as index and a column 'label' containing the assigned labels (`fail_label` or `pass_label`)
+        A DataFrame with unit IDs as index and a column `column_name` containing the assigned labels (`fail_label` or `pass_label`)
     """
     import pandas as pd
 
@@ -77,7 +80,7 @@ def threshold_metrics_label_units(
         raise ValueError("nan_policy must be 'fail', 'pass', or 'ignore'")
 
     labels = pd.DataFrame(index=metrics.index, dtype=str)
-    labels["label"] = fail_label
+    labels[column_name] = fail_label
 
     # Key change: init depends on operator
     pass_mask = np.ones(len(metrics), dtype=bool) if operator == "and" else np.zeros(len(metrics), dtype=bool)
@@ -101,24 +104,24 @@ def threshold_metrics_label_units(
             metric_ok &= values <= max_value
 
         # Handle NaNs
-        nan_mask = slice(None)
         if nan_policy == "fail":
             metric_ok &= ~is_nan
+            valid_mask = slice(None)
         elif nan_policy == "pass":
             metric_ok |= is_nan
-        else:
-            # if nan_policy == "ignore", we only set values for non-nan entries
-            nan_mask = ~is_nan
+            valid_mask = slice(None)
+        elif nan_policy == "ignore":
+            valid_mask = ~is_nan
 
         any_threshold_applied = True
 
         if operator == "and":
-            pass_mask[nan_mask] &= metric_ok[nan_mask]
-        else:
-            pass_mask[nan_mask] |= metric_ok[nan_mask]
+            pass_mask[valid_mask] &= metric_ok[valid_mask]
+        elif operator == "or":
+            pass_mask[valid_mask] |= metric_ok[valid_mask]
 
     if not any_threshold_applied:
         pass_mask[:] = True
 
-    labels.loc[pass_mask, "label"] = pass_label
+    labels.loc[pass_mask, column_name] = pass_label
     return labels


src/spikeinterface/curation/qualitymetrics_curation.py
```diff
@@ -0,0 +1,80 @@
...
+        A DataFrame with unit IDs as index and a column 'label' containing the assigned labels ('noise' or 'good').
+    """
+    import pandas as pd
+
+    # Get the quality metrics from the analyzer
+    assert analyzer.has_extension("quality_metrics"), (
+        "The provided analyzer does not have quality metrics computed. "
+        "Please compute quality metrics before labeling units."
+    )
+    qm = analyzer.get_extension("quality_metrics").get_data()
```
chrishalcrow: Could we make this a metrics curation, and use `get_metrics_extension_data()`? We use e.g. `exp_decay` all the time.

src/spikeinterface/curation/qualitymetrics_curation.py
```diff
@@ -0,0 +1,80 @@
...
+from pathlib import Path
+
+import numpy as np
+
+from spikeinterface.core.analyzer_extension_core import SortingAnalyzer
+
+from .curation_tools import is_threshold_disabled
+
+
+def qualitymetrics_label_units(
```
chrishalcrow: I'd vote for something like `threshold_label_units` or `threshold_metrics_label_units`

alejoe91: done

src/spikeinterface/curation/qualitymetrics_curation.py
alejoe91: ok!

alejoe91: done!

Julie-Fabre:
My kind of high level idea was people can add or remove threshold in the bombcell_label_units and that is then equivalent to this (if people set thresholds to none for metrics other than fr and snr) but people also get the output plots spit out and (eventually) the integration in the GUI. We could call bombcell just " quality metrics" and then have everything alltogether - does it make sense to duplicate things or am I missing how this is different?

alejoe91:
> My kind of high level idea was people can add or remove threshold in the bombcell_label_units and that is then equivalent to this (if people set thresholds to none for metrics other than fr and snr) but people also get the output plots spit out and (eventually) the integration in the GUI. We could call bombcell just " quality metrics" and then have everything alltogether - does it make sense to duplicate things or am I missing how this is different?

Hi @Julie-Fabre 

I think that bombcell is more sophisticated than this! You can define different rules for different thresholds (e.g., noise/non-soma etc). Maybe what we could do is the following:
- extend this function to be able to externally set labels
- use this function in cascade within bombcell internally, to make the various labeling

I also think it makes sense to keep `bombcell` since there is a lot of work and feature "engineering" to get a proper output, so it should receive proper credit!

But I'm open to discussion! @chrishalcrow what do you think?

m-beau:
Hey there!

Having a general-purpose function to threshold quality metrics in various ways is great, but I tend to agree with julie - this could very well be implemented as variations from the default parameters of bombcell. After all, the metrics people use should probably be variations from the bombcell values, and I’m sure a lot of the default parameters used here will be from bombcell, as they have been tested and are trusted by people, so it would have the benefit of making this clear for users.

chrishalcrow:
I think one of the aims of this was to use it as a educational tool mostly (we also have simple sorter etc!). The idea is: look, you can use some simple thresholds to do a very naive good/bad classification, but you should really do something more sophisticated. We're also laying the groundwork for having a "unitlabel" extension to keep provenance about labeling decision (and making it easy to generate "methods section" which include labeling tools!)

alejoe91:
@Julie-Fabre @m-beau I also see it more the other way around: this threshold based method can be the building block of bombcell, which does much more than just simple thresholds in terms of multiple steps applied in different orders and rules (e.g., "and" versus "or").

I made a draft implementation of the concept here: https://github.com/SpikeInterface/spikeinterface/pull/4306/changes#diff-1d9dd3572d81d393b756ec67a04c624f15cd208fd6fb67d5c9a1468aff927732R149-R245

Julie-Fabre:
Thanks for the discussion everyone! Just to share my perspective: 

I'm not sure I see the value in duplicating the simplest part of bombcell under a different name, even as an educational tool. Bombcell already does thresholding, plus it generates output plots that help users understand and refine their classifications, and I'm actively working on SI GUI integration. What would threshold_metrics offer that bombcell doesn't already provide, or couldn't provide with minor adjustments to expose simpler options for users who want that, or in a simple example script to get them started? If it's meant as an educational tool, wouldn't you want people using the correct function calls and getting used to looking at the output plots from the start? Curious to hear what you think! 
 
I also feel like progressively decomposing bombcell's logic into simpler components post hoc is always possible, but I don't think it reflects the actual intellectual lineage.

> We're also laying the groundwork for having a "unitlabel" extension to keep provenance about labeling decision (and making it easy to generate "methods section" which include labeling tools!)

That’s very cool, nice! I'm very happy this is happening! I worked on some of this but still have not pushed anything, I’ll try to participate with what you’ve already done :) 

Edit: let's continue chatting on slack

src/spikeinterface/curation/threshold_metrics_curation.py
```diff
@@ -0,0 +1,129 @@
...
+
+import numpy as np
+
+from spikeinterface.core.analyzer_extension_core import SortingAnalyzer
+
+from .curation_tools import is_threshold_disabled
+
+
+def threshold_metrics_label_units(
+    sorting_analyzer_or_metrics: "SortingAnalyzer | pd.DataFrame",
```
alejoe91: only accept dataframe (`sorting_analyzer.get_metrics_extension_data()` is easy enough)

src/spikeinterface/curation/threshold_metrics_curation.py
```diff
@@ -0,0 +1,129 @@
...
+        if not is_threshold_disabled(min_value):
+            metric_pass &= values >= min_value
+        if not is_threshold_disabled(max_value):
+            metric_pass &= values <= max_value
+
+        # Handle NaNs
+        if nan_policy == "fail":
+            metric_ok &= ~is_nan
+        else:  # "ignore"
+            metric_ok |= is_nan
```
alejoe91:
```suggestion
        elif nan_policy == "pass": 
            metric_ok |= is_nan
```

src/spikeinterface/curation/threshold_metrics_curation.py
```diff
@@ -0,0 +1,129 @@
...
+
+        any_threshold_applied = True
+
+        if operator == "and":
+            pass_mask &= metric_ok
+        else:
+            pass_mask |= metric_ok
```
alejoe91:
```suggestion
        if nan_policy == "ignore":
            if operator == "and":
                pass_mask[~is_nan] &= metric_ok[~is_nan]
            else:
                pass_mask[~is_nan] |= metric_ok[~is_nan]
```
something like that

src/spikeinterface/curation/threshold_metrics_curation.py
```diff
@@ -0,0 +1,130 @@
...
+        values = metrics[metric_name].to_numpy()
+        is_nan = np.isnan(values)
+
+        metric_ok = np.ones(len(values), dtype=bool)
+        if not is_threshold_disabled(min_value):
+            metric_ok &= values >= min_value
+        if not is_threshold_disabled(max_value):
+            metric_ok &= values <= max_value
+
+        metric_pass = np.ones(len(metrics), dtype=bool)
```
chrishalcrow: Should this be before the for loop?

src/spikeinterface/curation/threshold_metrics_curation.py
```diff
@@ -0,0 +1,130 @@
...
+        if not is_threshold_disabled(min_value):
+            metric_pass &= values >= min_value
+        if not is_threshold_disabled(max_value):
+            metric_pass &= values <= max_value
```
chrishalcrow:
```suggestion
```

This is not used.

src/spikeinterface/curation/threshold_metrics_curation.py
```diff
@@ -0,0 +1,130 @@
...
+        Policy for handling NaN values in metrics. If "fail", units with NaN values in any metric will be labeled as
+        fail_label. If "pass", units with NaN values in one metric will be labeled as pass_label.
+        If "ignore", NaN values will be ignored. Note that the "ignore" behavior will depend on the operator used.
+        If "and", NaNs will be treated as passing, since the initial mask is all true;
+        if "or", NaNs will be treated as failing, since the initial mask is all false.
+
+    Returns
+    -------
+    labels : pd.DataFrame
+        A DataFrame with unit IDs as index and a column 'label' containing the assigned labels ('noise' or 'good').
```
chrishalcrow:
```suggestion
        A DataFrame with unit IDs as index and a column 'label' containing the assigned labels (`fail_label` or `pass_label`)
```

src/spikeinterface/curation/threshold_metrics_curation.py
alejoe91: Right, I think I somehow duplicated it

alejoe91: Right, I think I somehow duplicated it

PR #4374, opened 2026-02-06
Try to remove   get_template_extremum_channel() and peak_sign in many place

samuelgarcia:
This PR try to
  * change semantic `extremum_channel` to `main_channel`
  * remove `get_template_extremum_channel()` to use instead
    * `analyzer.get_main_channel()` which is now a property
    * `templates.get_main_channel()` which compute on the fly the old behavior
  * remove the `peak_sign` when it was use to get the main channel
  * use explicitly `main_channel_peak_sign` instead of `peak_sign` when it is for main channel
  * keep `peak_sig` when it makes sens, like peak detection or in sorter context.

The main  consequences:
 *  an analyzer is constructed with main_channel_index. The sparsity is estimated by default on it. This will be faster when
     it is provided
 * The analyzer has always via the sorting object acces to this property so all metrics will **consistently** depend on it.

@alejoe91 
@chrishalcrow 
@ecobost 
@yger


src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -45,21 +45,31 @@
...
+    num_spikes_for_main_channel=100,
     sparse=True,
     sparsity=None,
     set_sparsity_by_dict_key=False,
     return_scaled=None,
     return_in_uV=True,
     overwrite=False,
     backend_options=None,
-    **sparsity_kwargs,
+    sparsity_kwargs=None,
```
ecobost, line 69: This argument could be dropped. It's not the responsibility of this function to create all types of sparsity. You could either provide an Sparsity object or let create_sorting_analyzer create it with the default method (from_radius_and_main_channel_index). I feel like that's fair enough (it's more of a convenience) and simplifies the sparsity logic here (curently there are three or four different ways to get sparsity). It also avoids some possible inconsistencies (like main_channel_peak_sign != sparsity_kwargs['peak_sign']) or num_spikes_for_main_channel vs sparsity_kwargs['num_spikes_for_sparsity']

The one common use case that will be affected is people that want to create a sorting analyzer with a bigger or smaller radius (the only sparsity_kwargs that would affect the radius_um method + main_index); but maybe they'll just have to estimate sparsity explicitly (and feed it here).

src/spikeinterface/core/sparsity.py
```diff
@@ -779,7 +805,14 @@ def estimate_sparsity(
...
-    if method != "by_property":
+    if method == "radius" and  main_channel_index is not None:
+        assert main_channel_index.size == sorting.unit_ids.size
+        chan_locs = recording.get_channel_locations()
+        sparsity = ChannelSparsity.from_radius_and_main_channel(
+            sorting.unit_ids, recording.channel_ids, main_channel_index, chan_locs, radius_um
+        )
+
+    elif method != "by_property":
         nbefore = int(ms_before * recording.sampling_frequency / 1000.0)
```
ecobost, line 816: This code (L.816-849) is a duplicate of `template_tools.estimate_main_channel_from_recording` (used in `create_sorting_analyzer` to find the main_channel). If it's too bothersome to unify both pieces of code somehow, maybe we can at least refactor so radius method here uses estimate_main_channel_from_recording so we can guarantee that:
```python
sparsity = create_sparsity(peak_sign=my_peak_sign, num_spikes_for_sparsity=my_num_spikes)
analyzer1 =  create_sorting_analyzer(sparsity=sparsity)
analyzer2 = create_sorting_analyzer(main_channel_peak_sign=my_peak_sign, num_spikes_for_main_channel=my_num_spikes)
assert analyzer1.sparsity == analyzer2.sparsity 
```

Essentially refactor it as
```python
if method== radius:
    if main_channel_index is None:
        main_channel_index = estimate_main_channel_from_recording(...)
    sparsity = .... from_radius_and_main_channel(...)
elif method != 'property'
       # the repeated code here (any changes won't affect radius method)
```

src/spikeinterface/core/sparsity.py
ecobost: this does not receive peak_sign anymore

ecobost, line 701: from_radius does not receive peak_sign anymore

ecobost, line 743: consider changing default to "both" to agree with default in create_sorting_analyzer (`main_channel_peak_sign`)

src/spikeinterface/postprocessing/spike_amplitudes.py
```diff
@@ -35,9 +35,8 @@ def _get_pipeline_nodes(self):
         peak_sign = self.params["peak_sign"]
         return_in_uV = self.sorting_analyzer.return_in_uV
 
-        extremum_channels_indices = get_template_extremum_channel(
-            self.sorting_analyzer, peak_sign=peak_sign, outputs="index"
-        )
+        extremum_channels_indices = self.sorting_analyzer.get_main_channels(outputs="index", with_dict=True)
+
         peak_shifts = get_template_extremum_channel_peak_shift(self.sorting_analyzer, peak_sign=peak_sign)
```
ecobost: We could feed the peak_sign from self.sorting_analyzer.main_channel_peak_sign.
` get_template_extremum_channel_peak_shift` uses get_template_extremum_channel internally which could pick a different channel to the main_channels (if using a different peak_sign). One could also send the extremum_channel_indices as input to this function (to avoid the use of get_template_extremum_channel but this is slightly more annnoying because the function is used in a couple of other places, so maybe not worth it); and it will still need peak_sign as argument even after sending the channel_indices

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -182,16 +181,13 @@ def compute_snrs(
 
     channel_ids = sorting_analyzer.channel_ids
 
-    extremum_channels_ids = get_template_extremum_channel(sorting_analyzer, peak_sign=peak_sign, mode=peak_mode)
+    main_channel_index = sorting_analyzer.get_main_channels(outputs="index", with_dict=True)
     unit_amplitudes = get_template_extremum_amplitude(sorting_analyzer, peak_sign=peak_sign, mode=peak_mode)
```
ecobost: We could set `peak_sign=analyzer.main_channel_peak_sign` to make sure amplitude channels coincide with main_channels (otherwise you are dividing signal from one channel to noise of another). Or at least set the default Or perhaps at least set peak_sign to 'both' to make it consistent with the default sortinganalyzer (I saw it is changed to "both" in the bombcell PR not sure why but we might as well).

src/spikeinterface/metrics/quality/misc_metrics.py
```diff
@@ -182,16 +181,13 @@ def compute_snrs(
 
     channel_ids = sorting_analyzer.channel_ids
```
ecobost, line 176: this is not used

src/spikeinterface/metrics/quality/misc_metrics.py
ecobost: this will also allow dropping the peak_sign and mode as arguments to SNR

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -381,16 +446,25 @@ def create(
             from spikeinterface.curation.remove_excess_spikes import RemoveExcessSpikesSorting
 
             sorting = RemoveExcessSpikesSorting(sorting=sorting, recording=recording)
-
+        
+        # This will ensure that the sorting saved always will have this main_channel
+        assert main_channel_index is not None
+        sorting.set_property("main_channel_index", main_channel_index)
```
ecobost: I don't fully understand how internals are managed, will this also change/save the property in the sorting file (for reloading)? Do we want that? If I call create_sorting_analyzer with main_channel_ids set to some manual channels for testing I would not want to overwrite the sorting property in file. To me, main_channels feels more a property of the analyzer than the sorting (or maybe both, but inheriting the property from sorting-> analyzer seems ok, but the other way around seems off).

src/spikeinterface/postprocessing/localization_tools.py
ecobost, line 167: peak_sign is no longer used for method radius ((compute_sparsity uses analyzer/templates.get_main_channels) so not needed here. Can also be dropped as an argument (only was used in this line)

ecobost, line 229: not true anymore; now it is only used to invert the gaussian prototype template (in l. 275); I think peak_sign='neg' default makes sense in this case

src/spikeinterface/metrics/template/template_metrics.py
ecobost, line 235: not used anywhere

ecobost: In theory you can drop peak_sign (it is no longer used anywhere here); in practice, it will be used in the bombcell PR (and probably by Alessio/Chris when they re-work the template metrics) but perhaps rather than as an argument, it can be read from analyzer

src/spikeinterface/metrics/quality/quality_metrics.py
ecobost: this is no longer needed (it was only used for the extremum_channel in the PCA metrics)

ecobost (commented):
Hi @samuelgarcia Thanks for taking time to work in this; I checked the code from analyzer-> metrics->widgets (anything before analyzer I don't really know).

After going through the code, the only extensions that still receive peak_sign are spike_amplitudes, quality metrics and template_metrics; and that argument can be dropped from all of them (in xxxxxx_metrics is not used at all, while in spike_amplitudes is used once but could be replaced with analyzer.main_channel_peak_sign). Given that, I would vote to rename main_channel_peak_sign to just peak_sign in the sorting analyzer (and essentially use it as the global peak_sign that will be used if any extension needs it, similar to sparsity), including ofc to find the main_channel. 

In theory, main_channel_peak_mode falls in the same camp but in truth that seems to only be used to pick the main_channel so main_channel_peak_mode as name might be ok.

ecobost:
Also, given that you are touching this code anyway, do you mind sending peak_sign here too
https://github.com/SpikeInterface/spikeinterface/blob/25940b471ddefbfae038aff7a7db55e5386ccd4f/src/spikeinterface/curation/remove_redundant.py#L84
I don't mean to make it work with the analyzer (although maybe it makes sense to forward the analyzer.main_channel_peak_sign here too) but just to forward the peak_sign argument like
```python 
unit_peak_shifts = get_template_extremum_channel_peak_shift(sorting_analyzer, peak_sign=peak_sign)
```
thanks 👍🏽

src/spikeinterface/core/template.py
```diff
@@ -483,3 +483,27 @@ def get_channel_locations(self) -> np.ndarray:
...
+        mode: "extremum" | "at_index" | "peak_to_peak" = "extremum",
+        outputs="index",
+        with_dict=True
+    ):
+        from .template_tools import _get_main_channel_from_template_array
+
+        templates_array = self.get_dense_templates()
+        main_channel_index = _get_main_channel_from_template_array(templates_array, mode, main_channel_peak_sign, self.nbefore)
+
+        if outputs is "index":
```
ecobost: this should use == (rather than is)

src/spikeinterface/core/template.py
```diff
@@ -483,3 +483,27 @@ def get_channel_locations(self) -> np.ndarray:
...
+            main_chans = main_channel_index
+        elif outputs is "id":
```
ecobost: this should use == (rather than is)

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -881,6 +1021,24 @@ def get_sorting_property(self, key: str, ids: Optional[Iterable] = None) -> np.n
             Array of values for the property
         """
         return self.sorting.get_property(key, ids=ids)
+    
+    def get_main_channels(self, outputs="index", with_dict=False):
+        """
+
+        """
+        main_channel_index = self.get_sorting_property("main_channel_index")
+        if outputs is "index":
```
ecobost: this should use == (rather than is)

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -881,6 +1021,24 @@ def get_sorting_property(self, key: str, ids: Optional[Iterable] = None) -> np.n
...
+            main_chans = main_channel_index
+        elif outputs is "id":
```
ecobost: this should use == (rather than is)

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -525,16 +616,111 @@ def create_binary_folder(cls, folder, sorting, recording, sparsity, return_in_uV
...
+            pass
+        else:
+            # old version did not have settings at all
+            new_settings["return_in_uV"] = True
+
+        if "main_channel_peak_sign" not in settings:
+            # before 0.104.0 was not in main_channel_peak_sign
+            # TODO make something more fancy that exlore the previous params of extension
+            new_settings["main_channel_peak_sign"] = "both"
+            new_settings["main_channel_peak_mode"] = "extremum"
```
chrishalcrow, line 654: Hello - Sam asked me to think about backwards compatibility. Here are my thoughts!

Before this PRs, we had `peak_sign` as a parameter for lots of postprocessig methods. Now we're getting rid of it. We want to ensure backwards compatibility. Note: we cannot be fully backwards compatible - we used to allow for different peak_sign methods for different extensions. Note: THIS WAS BAD. It meant there could be confusing inconsistencies between the results.

When you load an "old" analyzer, you get a `sparsity.mask` but no `main_channel_indices`. Use cases of loading an "old" analyzer:

1. Have a look at it
2. Do some curation
3. Recompute extensions with new parameters

The existing analyzer has a sparsity mask. We should respect it. If we do that, 1. is resolved - the analyzer will basically look the same as before.

Currently, when we do a curation, sparsity of the new units is re-computed as a union or intersection. At this stage, we'll also need to recompute the `main_channel_index` for the merge/split. For this we'll need to know which `peak_sign` to use.

Recomputing is fine. If a user is recomputing, they know results might be different (improved!!) on a new version.

So, we just need to decide how to compute `main_channel_indices` in case of curation or re-computation. So that we respect the existing sparsity, I think we should only allow `main_channel_indices` which exist on the sparsity mask. The options are then:

1. Set peak_sign to some reasonable default value (`'both'`!!)
2. Try to reverse engineer the `peak_sign` using the templates (note: this is not guaranteed to work: the sparsity estimation that runs on analyzer creation generates templates on the fly - these are not the same as the templates in the analyzer. Or "both" and "neg" will give the same result => problem not invertible. Note: even if we reverse-engineer the `peak_sign`, the user might have set a different `peak_sign` for e.g. snr computation)

I think going for "both" is fine. It's easier to explain and the other option is too awkward.

Ok. So my proposal for backwards compat is: on load, compute `main_channel_indices` using `peak_sign = 'both'` but only on the channels included in the sparsity mask.

src/spikeinterface/core/basesorting.py
```diff
@@ -13,6 +13,9 @@ class BaseSorting(BaseExtractor):
     """
     Abstract class representing several segment several units and relative spiketrains.
     """
+    _main_properties = [
+        "main_channel_index",
```
alejoe91, line 17:
```suggestion
        "main_channel_id",
```
Use `channel_id` instead

src/spikeinterface/core/template.py
```diff
@@ -482,3 +482,27 @@ def get_channel_locations(self) -> np.ndarray:
         assert self.probe is not None, "Templates.get_channel_locations() needs a probe to be set"
         channel_locations = self.probe.contact_positions
         return channel_locations
+
+    def get_main_channels(self,
+        peak_sign: "neg" | "both" | "pos" = "both",
+        peak_mode: "extremum" | "at_index" | "peak_to_peak" = "extremum",
+        outputs="index",
+        with_dict=True
```
chrishalcrow, line 490: Hello - this default is different than the analyzer's get_main_channel, which means some downstream code needs to be different depending on if you have an analyzer or templates. Can we change it to match please?
```suggestion
        with_dict=False
```

alejoe91:
Done in ##4624

PR #4377, opened 2026-02-09
Fix metrics test for numba correlograms fast_mode

alejoe91:
Leftover from #4305


yger:
damned, i forgot something. But wasn't the auto mode default?

chrishalcrow:
> damned, i forgot something. But wasn't the auto mode default?

The default is there for the `ComputeCorrelograms` class and `_compute_correlograms_on_sorting` but there's not a default for `_compute_correlograms_numba` (which takes in a bool)

PR #4380, opened 2026-02-09
Fix pip 26.0+ compatibility in container_tools

### LINKED ISSUE #4368: run_sorter() fails after pip 26.0+ due to deprecated #egg= syntax
Hey! Running into an issue with containerized sorting using `installation_mode="github"` (or `"auto"`). Modern pip (26.0+) rejects the `#egg=` fragment syntax when extras are included.

**The error:**

```
× The 'spikeinterface[full]' egg fragment is invalid
hint: Try using the Direct URL requirement syntax: 'name[extra] @ URL'
```

The current code in `install_package_in_container()` builds:

```bash
pip install ... https://github.com/.../archive/main.tar.gz#egg=spikeinterface[full]
```

But pip 26.0+ wants the Direct URL syntax instead:

```bash
pip install ... "spikeinterface[full] @ https://github.com/.../archive/main.tar.gz"
```

This was a bit tricky to track down since the pip failure is silent. It only shows up later as a `ModuleNotFoundError` when the container tries to import spikeinterface.

esutlie:
## Summary
  - Replace deprecated ``#egg=`` fragment syntax with PEP 440 Direct URL format (``"package[extras] @ URL"``) in
  ``install_package_in_container()`` github mode
  - Add warning when pip install output contains errors, so container install failures are no longer silent

  ## Context
  pip 26.0+ rejects the ``#egg=`` syntax when package extras are specified, causing ``run_sorter()`` to fail with a
  confusing ``ModuleNotFoundError`` instead of a clear install error.

  Fixes #4368

### DIFF
--- a/src/spikeinterface/sorters/container_tools.py
+++ b/src/spikeinterface/sorters/container_tools.py
@@ -268,9 +268,6 @@ def install_package_in_container(
         raise ValueError(f"install_package_incontainer, wrong installation_mode={installation_mode}")
 
     if isinstance(res_output, str) and "ERROR" in res_output.upper():
-        warnings.warn(
-            f"pip install of {package_name} in container may have failed. "
-            f"pip output:\n{res_output}"
-        )
+        warnings.warn(f"pip install of {package_name} in container may have failed. " f"pip output:\n{res_output}")
 
     return res_output


samuelgarcia (approved):

samuelgarcia:
This is nice thanks you @esutlie

alejoe91 (approved):

PR #4382, opened 2026-02-10
Make NumpySorting.from_unit_dict docs and code consistent

### LINKED ISSUE #4375: NumpySorting.from_unit_dict() spike samples instead of times
The function `NumpySorting.from_unit_dict` sais 

> Each dict have unit_ids as keys and spike times as values

but `units_dict_list` is interpreted to have values as samples

https://github.com/SpikeInterface/spikeinterface/blob/25940b471ddefbfae038aff7a7db55e5386ccd4f/src/spikeinterface/core/numpyextractors.py#L410

Because of `minimum_spike_dtype` the data is casted from seconds float to int64

Either we adjust the docstring, or it could be nice to have a convenience function like `from_unit_dict` where we can define whether to be interpreted as samples or seconds (like the recent updates to spikeinterface clarified `NumpySorting.from_times_and_labels` / `NumpySorting.from_samples_and_labels` we could name them differently )

Edit: Actually I was looking for a function to create a sorting from a Pynapple TsGroup (in seconds) as `to_pynapple_tsgroup` creates, so to read that again without having access to the original Sorting/SortingAnalyzer.
```
# spike_trains = dict(session.units)
spike_trains = {unit_id: st.t for unit_id,st in units_tsgroup.items()} # units_tsgroup is a nap.TsGroup , convert times from nap.Ts (timestamp) to numpy array
spike_trains.keys() # unit ids
sorting = NumpySorting.from_unit_dict(spike_trains, sampling_frequency=sampling_frequency)
```

alejoe91:
Fixes #4375

@pas-calc I think this makes it unambiguous :)

I think it's simple enough for users to do the time to samples conversion externally, without making the function more complex than it should be

### DIFF
--- a/.github/workflows/all-tests.yml
+++ b/.github/workflows/all-tests.yml
@@ -239,6 +239,19 @@ jobs:
         run: |
           pip install -e .[full]
           pip list
+
+          # Internal shell check for platform and env var
+          if [[ "$RUNNER_OS" == "macOS" ]]; then
+            # Locate the binary dynamically
+            SHM_BINARY=$(python -c "import torch; from pathlib import Path; print(Path(torch.__file__).parent / 'bin' / 'torch_shm_manager')")
+            if [ -f "$SHM_BINARY" ]; then
+              echo "Applying chmod +x to $SHM_BINARY"
+              chmod +x "$SHM_BINARY"
+            else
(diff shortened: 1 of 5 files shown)


pas-calc:
Yes this looks reasonable and clear. I was initially running this function assuming it takes "time" instead of "samples" and realized why it returned wrong result by looking at the function definition.

PR #4383, opened 2026-02-11
Update lower bound python support to 3.10

chrishalcrow:
Neo `0.14.3` only supports python >= 3.10 (https://github.com/NeuralEnsemble/python-neo/blob/master/pyproject.toml), so we actually already haven't supported 3.9 for a while! All tests etc already use 3.10 as the lower bound.

This PR will make `uv` stop complaining ;)

Also tried to remove `from __futures__ import annotations`, `Union` and `Optional`. Turns out you need futures for:

- self reference typing in classes. E.g. BaseExtractor.get_parents returns BaseExtractor
- Stuff like `method: "std" | None = None`. In the future, we'll deal with this properly using `Literal`s. For now, let's use `__futures__`

### DIFF
--- a/src/spikeinterface/curation/curation_model.py
+++ b/src/spikeinterface/curation/curation_model.py
@@ -1,5 +1,5 @@
 from pydantic import BaseModel, Field, model_validator, field_validator, field_serializer
-from typing import Literal
+from typing import Literal, List
 from itertools import chain, combinations
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


alejoe91:
With 3.10, we can remove all `from __future__ import annotations`! Can you try to remove them all as part of the PR?

chrishalcrow:
Was looking at what else we can clean - don't need `Union` or `Optional` anymore either!

PR #4384, opened 2026-02-11
Make anyone happy about warnings and margin in filtering.

samuelgarcia:
Make anyone happy about warnings and margin in filtering:
  * avoid unnecessary wrning when not needed
  * do not warn when loading old spikeinterface files
  * warn and raise for new case
  * do warn but not raise when loading old spikeinterface files in LFP cases.

@alejoe91 @yger @chrishalcrow 

this will close #4334

### DIFF
--- a/src/spikeinterface/core/loading.py
+++ b/src/spikeinterface/core/loading.py
@@ -14,7 +14,6 @@ _error_msg = (
 
 def load(
     file_or_folder_or_dict,
-    ignore_warnings=False,
     **kwargs,
     # load_extensions=True, backend_options=None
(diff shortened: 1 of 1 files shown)


src/spikeinterface/core/loading.py
```diff
@@ -14,6 +14,7 @@
 
 def load(
     file_or_folder_or_dict,
+    ignore_warnings=False,
```
alejoe91:
```suggestion
```

samuelgarcia: is it propagated somewhere ?

PR #4388, opened 2026-02-12
Fix micro version check in base

alejoe91:
Following #4384, we also have to check the micro version to handlo backward compatible load, not only major and minot (which is default)


chrishalcrow:
Looks good, but can you explain why we need to do this now?

alejoe91:
> Looks good, but can you explain why we need to do this now?

If we don't we will have the margin init error until we release 0.104.0, which honestly is ok. I wanted to test the mechanism though, and with this PR the warnign goes away.

alejoe91:
So let's not merge!

PR #4391, opened 2026-02-12
Use a truncated median filter near edges when detecting bad channels

ecobost:
a594368d7460097ee198d4f37d79f11e4012550b refactors `detect_bad_channels_ibl()` for readability. It touches a lot of lines but produces the exact same results as the version in main.

a988c0dd786956da8b15d580d21f7d5bf66f091e makes two (relatively minor) changes:
* In L.438, I zero center the reference before computing the cross-correlation, this brings the result to full parity with the ibl implementation (https://github.com/int-brain-lab/ibl-neuropixel/blob/a613b7ad4908f6337b19f6645131e779618ba98d/src/ibldsp/voltage.py#L946) and is more mathematically correct. It barely changes results because the data is already zero centered per channel (xcorr usually ranges from [0-1] and this changes the values by < 0.001, .e.g, 0.8913 -> 0.8917).
* Detrending uses a median filter (`xcorr_neigbors =x_corr - median_filter(x_corr)`) that pads the edges with a constant value (the nearest edge value, so the first or last element). I changed it so that the median near the edges is computed only with real data (no padding), for instance, with a kernel_size (n_neighbors) = 11, out[0] = median(x[0:6]), out[1] = x[0:7] and so on, until reaching the full window size. This has two advantages: 
   1) you are no longer limited to use an odd sized n_neighbors (though this is also true in commit a594368d7460097ee198d4f37d79f11e4012550b)
    2) You can use larger n_neighbors without biasing edge values towards the very first and very last value of the array (useful when expecting a lot of bad_channels, cause median is only robust to floor(kernel_size/2) bad channels) . It also increases the variance of the estimate near the edges which I think it's either a positive or neutral effect (or at least a better tradeoff).

### DIFF
--- a/.github/workflows/all-tests.yml
+++ b/.github/workflows/all-tests.yml
@@ -115,7 +115,7 @@ jobs:
 
       - name: Install git-annex
         shell: bash
-        if: env.RUN_EXTRACTORS_TESTS == 'true'
+        if: env.RUN_EXTRACTORS_TESTS == 'true' || env.RUN_PREPROCESSING_TESTS == 'true'
         run: |
           pip install datalad-installer
           if [ ${{ runner.os }} = 'Linux' ]; then


src/spikeinterface/preprocessing/detect_bad_channels.py
```diff
@@ -387,7 +387,7 @@ def detect_bad_channels_ibl(
     outside_channel_thr=-0.75,
     n_neighbors=11,
     nyquist_threshold=0.8,
-    welch_window_ms=0.3,
+    welch_window_ms=10.0,
```
chrishalcrow, line 390: This naively looks like quite a big change, can you explain what effect it has?

ecobost: 10 msecs is the default window size used in `detect_bad_channels` https://github.com/SpikeInterface/spikeinterface/blob/758d9d5806aaf474e51f6d3e04dc7d37692e3ec2/src/spikeinterface/preprocessing/detect_bad_channels.py#L171 so unless someone calls `detect_bad_channels_ibl` directly, 0.3 is not used. I changed it here so they agree.

The number defines how big the segments used to compute fourier transforms are (in `scipy.welch`); 0.3 msecs will be 9 samples (@30khz), which is too few (256 is the default in scipy that is used by ibl, 10msecs->300samples which is close enough, I personally set that to 25msecs in my code for cleaner estimates)

chrishalcrow: Great, thanks for the explanation! I agree with the change.

JoeZiminski:
Thanks @ecobost! this is very nice. I just noticed not all the tests are running due to the IBL `neurodsp` being renamed to `ibldsp`. I'll make a PR to fix this and this PR can be rebased on top. but I checked it locally and all tests pass as expected.

Out of interest I'd like to go through in detail as it looks like you've done some reallly cool stuff, though I won't have time until the week  23rd Feb, if you don't mind waiting. If this needs merging before I don't mind as happy as tests are passing.

Good catch on the `welch_window_ms` discrepancy, we should add a test to make sure the defaults of the top-level and this subfunction match. Would you mind adding such a test to `test_detect_bad_channels`? You can do something like `signature(func_object)` to get the signature of the two functions, and directly compare their default values (similar to [here](https://github.com/SpikeInterface/spikeinterface/blob/758d9d5806aaf474e51f6d3e04dc7d37692e3ec2/.github/scripts/test_kilosort4_ci.py#L344). This might be a good time to review the tests at [test_detect_bad_channels](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/preprocessing/tests/test_detect_bad_channels.py), if you had time would you mind taking a quick look and seeing if there are any important cases that are missed? Of course this is all optional as a little out of scope for this PR.

For how the  `xcorr_neighbors`, `xcorr_distant`  are computed, I remember drawing out what is going on, getting it then forgetting, and it seems I didn't write any good explanatory comments 😅 It always takes me like 20 minutes and breaking into the code to remember how these values allow us to classify the channels as dead/noise/out of brain. Maybe this is a good time to add to the comments, as its fresh in your head would you mind extending the 'Compute coherence' comment to explain what is going on? Basically how these computations lead to the values-per-channel that are used to determine if they are noise, dead or out of brain e.g. 1) compute the xcorr between all channels and the median channel signal (gives a num channel array) 2) smooth this xcorr with a median filter 3) channels that are similar will be small clusters (or single channels, i.e. above the median filtered xcorr) of high xcorr with the median (noise) or low xcorr (dead) 4) channels that are outside of the brain will be contigious region (i.e. median filtered) set of channels with very low correlation with the other channels. 

Thanks again this is very cool

ecobost:
Hi @JoeZiminski , understanding the code should be a lot easier after the refactoring proposed here. I added a (rather long) description in the docstring of `detect_channels_ibl()`. I've described what the code does (IBL's and ours) not what they intended it to do.
 
Here's their method description from the IBL paper:
 """
Instead, for low amplitude channels, we computed the zero lag cross-correlation peak with the common median reference. This zero-lag cross-correlation is normalized by the energy of the reference trace. To spot bad channels, we subtract the channels trend (a median filter with n=11 points) from the cross correlation which we refer to as the channel ”similarity”, shown in figure 4). A conventional channel will score around 0 and anomalously low amplitude channels are then tagged if under a threshold of -0.5.

For noisy channels, we use a frequency domain approach. While a conventional neurophysiological voltage power spectrum will tend to decay as a function of frequency, faulty channels exhibit a flatter spectrum. A good way to detect these channels was to apply an absolute threshold on the average power spectral density (PSD) above 0.8 Nyquist. For the action potential band (AP), our suggested threshold is 0.02 µV2 Hz−1 and for the LFP band, our suggested threshold is 1.5 µV2 Hz−1 . Figure 4 shows an example on the AP band where 2 channels exhibit unnatural values for neurophysiology in mice.

For detecting channels outside of the brain, we use a similar method to the similarity thresholding, but we instead computed cross-correlations with the median of all channels instead of just the neighbouring channels. Only consecutive channels at the top of the probe below the threshold are excluded from further analysis.
"""
Couple of discrepancies:
1. There is no mention of using xcorr_neighbors > 1 to find noisy channels (and I actually don't conceptually understand what this condition is testing)
2. The way they describe outside channel detection is as "similar" to the dead channel method but "instead computed cross-correlations with the median of all channels instead of just the neighbouring channels". I don't know what they meant because the dead channel method says "zero lag cross-correlation peak with the common median reference" (note: "zero lag cross-correlation peak" = dot product) so it was already using the median of all channels. Did they intended to do xcorr - median(xcorr) (i.e., subtract the overall median rather than the local median)? 🤷🏽 

To be clear, this is IBL code <-> IBL paper discrepancies, not on SI side.

ecobost:
re: testing
I checked `test_detect_bad_channels`, looks good to me, I didn't know you were checking against the original implementation, that's nice. Only one of the test_xxxx functions is actually uncommented in main but I guess there are reasons for it. I would maybe uncomment the `test_detect_bad_channels_extremes` to check that the top and bottom channels are working correctly (and maybe add the edge case of num_out_channels=1).

The other thing I can think to test is the label priority of, for instance, what happens if a channel is both dead and outside (should be labelled as outside) or dead and noisy (should be labeled as noisy). But it's probably not as important. I don't fully understand how the labels are assigned: https://github.com/SpikeInterface/spikeinterface/blob/758d9d5806aaf474e51f6d3e04dc7d37692e3ec2/src/spikeinterface/preprocessing/tests/test_detect_bad_channels.py#L170-L174
but maybe adding some dead channels in the first channels of the recording will allow to check that these channels (although also dead) are labelled as outside.

Re: Adding a test for signatures to coincide between detect_bad_channels and detect_bad_channels_ibl. 
There is actually some params that differ(e.g., noisy_channel_threshold vs noisy_channel_thr) and psd_hf_threshold has a default value in detect_bad_channels but not in detect_bad_channels_ibl (which is sensible because in practice, there is two valid default values, one for lfp and one for ap, but detect_bad_channels uses the ap one). I see the detect_bad_channels_ibl as a sort of private/lower-level function and in general I don't think it's necessary to match the names or the default values, the public facing functions could have more readable arg names (say n_neighbors or noise_channel_threshold) and the private function more method-based names (like median_kernel_size or psd_threshold), Same for defaults: perhaps the private/more method-applied function can have defaults from the original implementation or more general inoffensive defaults, while the public facing defaults are specific to the main use case (I am not just taking about this specific functions but as a general design principle)

alejoe91:
@oliche can you also take a look?

oliche:
> The other thing I can think to test is the label priority of, for instance, what happens if a channel is both dead and outside (should be labelled as outside) or dead and noisy (should be labeled as noisy)

Here outside is the most important label IMO. After noisy/dead are equivalent (note that this is not the case currently in the IBL version).

One possible concern is the change on the median filter edge handling. This xcor_lf is used to detect channels outside of the brain, which are by definition at the top edge of the probe. Here I am not sure the tests have enough coverage to check for a possible change of behaviour, I think we have some labelled data at IBL that could be useful to validate this.

```
annotations = {
    '1d547041-230a-4af3-ba6a-7287de2bdec3': {'idead': [191], 'inoisy': [], 'ioutside': []},  # DOES NOT WORK OUTSIDE BRAIN DUE TO LOW THRESHOLD # eid:e5c75b62-6871-4135-b3d0-f6464c2d90c0, probe01, ('KS043', '2020-12-07', '001')
    '4cb60c5c-d15b-4abd-8cfd-776bc5a81dbe': {'idead': [16,  29, 191, 357], 'inoisy': [], 'ioutside': [358, 359, 360, 361, 362, 363, 364, 365, 366, 367, 368, 369, 370, 371, 372, 373, 374, 375, 376, 377, 378, 379, 380, 381, 382, 383]},
    'ce0dc660-f19e-46a3-94f9-646bebae6805': {'idead': [29, 36, 39, 40, 191], 'inoisy': [133, 235], 'ioutside': [274, 275, 276, 277, 278, 279, 280, 281, 282, 283, 284, 285, 286,
       287, 288, 289, 290, 291, 292, 293, 294, 295, 296, 297, 298, 299,
       300, 301, 302, 303, 304, 305, 306, 307, 308, 309, 310, 311, 312,
       313, 314, 315, 316, 317, 318, 319, 320, 321, 322, 323, 324, 325,
       326, 327, 328, 329, 330, 331, 332, 333, 334, 335, 336, 337, 338,
       339, 340, 341, 342, 343, 344, 345, 346, 347, 348, 349, 350, 351,
       352, 353, 354, 355, 356, 357, 358, 359, 360, 361, 362, 363, 364,
       365, 366, 367, 368, 369, 370, 371, 372, 373, 374, 375, 376, 377,
       378, 379, 380, 381, 382, 383]},  # eid:8ca740c5-e7fe-430a-aa10-e74e9c3cbbe8, probe01, ('NYU-40', '2021-04-14', '001')
    '3b729602-20d5-4be8-a10e-24bde8fc3092': {'idead': [191], 'inoisy': [], 'ioutside': [356, 357, 358, 359, 360, 361, 362, 363, 364, 365, 366, 367, 368, 369, 370, 371, 372, 373, 374, 375, 376, 377, 378, 379, 380, 381, 382, 383]},  # eid:fc43390d-457e-463a-9fd4-b94a0a8b48f5, probe00, ('NYU-47', '2021-06-25', '001')
    'a9c9df46-85f3-46ad-848d-c6b8da4ae67c': {'idead': [191], 'inoisy': [], 'ioutside': [348, 349, 350, 351, 352, 353, 354, 355, 356, 357, 358, 359, 360,
       361, 362, 363, 364, 365, 366, 367, 368, 369, 370, 371, 372, 373,
       374, 375, 376, 377, 378, 379, 380, 381, 382, 383]},  # eid:41431f53-69fd-4e3b-80ce-ea62e03bf9c7, probe01, ('CSH_ZAD_022', '2020-05-21', '001')
    '03d2d8d1-a116-4763-8425-4ef7b1c1bd35': {'idead': [131, 191, 235], 'inoisy': [], 'ioutside': []},  # eid:81a78eac-9d36-4f90-a73a-7eb3ad7f770b, probe01, ('CSH_ZAD_026', '2020-08-17', '001')
    'fe380793-8035-414e-b000-09bfe5ece92a': {'idead': [131, 191], 'inoisy': [], 'ioutside': []},  # eid:ff48aa1d-ef30-4903-ac34-8c41b738c1b9, probe01, ('CSH_ZAD_025', '2020-08-03', '001')
    '19c9caea-2df8-4097-92f8-0a2bad055948': {'idead': [], 'inoisy': [191], 'ioutside': []},  # eid:aa20388b-9ea3-4506-92f1-3c2be84b85db, probe01, ('DY_016', '2020-09-14', '001')
    '25a9182c-4795-4768-af47-98975d2d2a8a': {'idead': [], 'inoisy': [191], 'ioutside': []},  # eid:26aa51ff-968c-42e4-85c8-8ff47d19254d, probe01, ('DY_020', '2020-10-03', '001')
    # '41a3b948-13f4-4be7-90b9-150705d39005': {'idead': [191], 'inoisy': [], 'ioutside': []},  #DOES NOT PASS, WRONG DEAD CHANNEL # eid:7691eeb3-715b-4571-8fda-6bb57aab8253, probe00, ('NYU-30', '2020-10-19', '001')
}
```

I won't have bandwidth to try this until early March, but I am not opposed to merging this PR, as it makes sense on paper.

JoeZiminski:
thanks @ecobost, that docstring looks great. As it turns out, the code at the bottom of the tests doesn't actually run the tests in practice 😅 Its used for debugging purposes, I wanted to [get them removed](https://github.com/SpikeInterface/spikeinterface/pull/3249#discussion_r1691315907) because they are confusing, but failed. In practice `pytest` is used to run the tests, which will run any test with the name prefixed with `test_`, and those code sections at the bottom are ignored.

That's a nice idea to test how different labels are prioritised, when I take a look next week I'll add this test if it is easy to implement. I agree it is nice to have the flexibility of different defaults between levels, although would err to keep them the same just because its difficult to debug issues when outputs are different due to default arguments, or one set of default arguments changing but forgetting to update another. For example, a user would (most likely) expect that running either the main entry function or the lower-level function without specifying each argument would yield the same results. Similarly for the naming convention, as it makes matching the parameters across the functions trickier when debugging with the code. My preference would usually be to have no defaults on low-level functions to avoid this kind of thing but its not really practical here.

JoeZiminski:
Hi @ecobost apologies for the delay on this, I've scheduled in next Monday to take a look, but of course if this is urgent please feel free to have it merged given the other reviews

PR #4392, opened 2026-02-13
Rename `neurodsp` to `ibldsp` to ensure IBL tests run

JoeZiminski:
The IBL `neurodsp` was renamed to `ibldsp` and so the IBL tests stopped running. I renamed this, but also made some other changes that I will will break the tests due to dependcy issues. Lets see...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -146,6 +146,9 @@ test_extractors = [
 test_preprocessing = [
     "ibllib>=3.4.1;python_version>='3.10'", # streaming IBL
     "torch",
+    # for IBL data fetching
+    "pooch",
+    "datalad"
 ]
 
 
@@ -192,10 +195,6 @@ test = [
 
     # for slurm jobs
     "pytest-mock",
-
-    # for IBL data fetching
-    "pooch",
-    "datalad"
 ]
 
 docs = [


pyproject.toml
```diff
@@ -192,6 +192,10 @@ test = [
 
     # for slurm jobs
     "pytest-mock",
+
+    # for IBL data fetching
+    "pooch",
+    "datalad"
```
chrishalcrow: I'm kinda confused why these weren't here already. I don't think `neurodsp` depended on them.

JoeZiminski: Oh yea I thought so too, they are actually already there! under `test_extractors`. [This test](https://github.com/JoeZiminski/spikeinterface/blob/c06db9792959bf2bc5fdc1577ca6c2fb50098efc/src/spikeinterface/preprocessing/tests/test_highpass_spatial_filter.py#L31) checks on real data so it would be nice to include these dependencies in the general test dependencies if possible. I can't remember the reason except real data is used here, to test under real conditions which is kind of nice especially for debugging. That been said not opposed to removing it if its going to cause a lot of problems for test infrastrcture @samuelgarcia @alejoe91

alejoe91: I think it's ok to add it, but we could also add it selectively to the `test_preprocessing` section, so it only installs for it. @JoeZiminski what do you think?

alejoe91: done in last commit!

JoeZiminski: Yes sounds good to me, cheers @alejoe91!

samuelgarcia (approved):

samuelgarcia:
Merci Joe.

PR #4400, opened 2026-02-20
Enlarge channel textbox in plot_traces

ecobost:
I don't know how long channel_ids usually are but mine are usually longer than the channel box in `plot_traces`
Screenshot from 2026-02-20 02-16-30

This  widens that box so longer channel names fit better. (and some minor aesthetic changes). This is what it looks like
Screenshot from 2026-02-20 04-11-10

### DIFF
--- a/src/spikeinterface/widgets/traces.py
+++ b/src/spikeinterface/widgets/traces.py
@@ -396,7 +396,7 @@ class TracesWidget(BaseWidget):
 
         width_cm = backend_kwargs["width_cm"]
         height_cm = backend_kwargs["height_cm"]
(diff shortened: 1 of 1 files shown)


src/spikeinterface/widgets/traces.py
alejoe91: @ecobost I think this is the place where you can change the ratios, instead of hard-coding the width. Can you try?

ecobost: yes, that's better, I did not notice that ratios params. This is what it looks like now (same as above):
Screenshot from 2026-02-21 03-43-32

PR #4402, opened 2026-02-20
Optimizations: faster waveforms/templates and seeding iterative_isosplit properly

yger:
Some minor optimizations of the way waveforms are computed by the analyzer both for templates and waveforms. We can gain 15% speedup by slightly less operation. 
Plus seeding of iterative_isosplit is not perfect at the momentn because seeds are not propagated at the split level


samuelgarcia:
I do not believe the 15% but if it make you happy. lets merge.

PR #4405, opened 2026-02-23
minor: delete unnecessary check in find_duplicates first_iterative

ecobost:
I also refactor some code that took me way too long to understand 😅


ecobost:
nvm, my logic was wrong (you do need that check!)

PR #4406, opened 2026-02-23
Don't raise when extension is unknown

chrishalcrow:
This PR changes the `load_extension` and `get_extension_class` so that we do not raise on loading an analyzer/extension.

Why?
Suppose someone uses SpikeInterface 0.104 and computes some `auto_correlogram`s [or they've got a lab-specific extension for behavioral data]. They share the analyzer with a colleague who uses 0.103 [or is outside their lab]. Currently, the colleague cannot open the analyzer - the unknown extension raises an error!

This PR changes this error into a warning, and skips loading the extension. Now the `load` loads all the extensions it can, and skips the ones it can't (with a helpful warning).

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -1943,6 +1943,9 @@ extension_params={"waveforms":{"ms_before":1.5, "ms_after": "2.5"}}\
 
         extension_class = get_extension_class(extension_name)
 
+        if extension_class is None:
+            return None
+
         extension_instance = extension_class.load(self)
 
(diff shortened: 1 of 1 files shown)


samuelgarcia:
I am not sure to like this.

chrishalcrow:
Yeah, I wasn't quite expecting this to get merged haha.

Happy to debate! We are advertising the analyzer as something users can share with each other, and that external packages can use as a starting point to build their tools on. Hence I think it's really important that the analyzer loading "just works", no matter which version you're using and was used to create it. So I'm strongly in favor of this behavior.

Issue #4193, opened 2025-11-02
Local referencing cannot handle empty annulus


zzhmark:
The local radius mode of common reference can fail when no channel is found within the specified the radius range (annulus).

In most cases it falls upon the user to find a proper range, but sometimes it's hard. In my case, I would first discard disconnected channels with high impedance, changing the local density of the contacts. A workaround is to maintain the channels, but those channels can introduce a lot of noise into the reference.

Do you find it useful that we can add some new features like specifying KNN or return to global referencing when it fails?

alejoe91:
Hi @zzhmark 

This is a good point. Yeah probably it would make sense in case no channels are in the annulus to maintain the channel, but add a warning (at init). Can you make a PR?

PR #4412, opened 2026-02-26
Implementing KNN referencing and a backup referencing mechanism for common_reference

zzhmark:
PR based on this issue: https://github.com/SpikeInterface/spikeinterface/issues/4193

The new features are designed for ungrouped local referencing.

# Changes in parameters:
* `reference`: include `knn` as an option
* `backup_reference`: allowing `global`, `single` and `knn` (`local` is unstable) to take effect when local referencing cannot get enough #channels, default as global
* `nneighbors`: used for knn referencing, i.e. #channels nearby used for referencing
* `backup_thr`: when #channel found locally is below this, turn to `backup_reference`, default as 1, i.e. no channel found.  (we may also consider None to force explicit assignment)

# Changes in workflow:
* Verification in parameters. Since `reference` and `backup_reference` both need preprocessing, all the elif are turned as if.
* Removing assertion of no local reference found as it would be judged dynamically.
* Add `knn` preprocessing in __init__ (similar to `local`).
* Changing the ungrouped part in get_traces.
  * Defining each reference method as local functions to call.
  * In `local`, setup if else to choose backup when #channels not meeting the requirement.

### DIFF
--- a/src/spikeinterface/preprocessing/common_reference.py
+++ b/src/spikeinterface/preprocessing/common_reference.py
@@ -8,7 +8,6 @@ from spikeinterface.core import get_closest_channels
 from spikeinterface.core.baserecording import BaseRecording
 
 from .filter import fix_dtype
-from functools import cache
 
 
 class CommonReferenceRecording(BasePreprocessor):


zzhmark:
Here is an example ephys data I used for testing to see the differences. KNN is working better than global. The backup mechanism is verified by setting the local radius as (50, 51) (no channel would be found) and its backup method is global, so backup car is getting the same result as the global car, and it's different from the proper local car (50,200).

probe
raw
global car
knn car
local car
backup car (global)

alejoe91:
Hi @zzhmark 

If I'm reading the implementation correctly, you are not discarding channels within the inner radius when computing nearest neighbors. I think this is the wrong behavior: if you rereference to closed-by channels, spiking signals will be essentially killed.
I think that we should still use the `local` radii, and keep the knn as backend when no channels are found. I think that was the original idea from #4193: to fix the case when local finds no channels. How does that sound?

zzhmark:
> Hi @zzhmark
> 
> If I'm reading the implementation correctly, you are not discarding channels within the inner radius when computing nearest neighbors. I think this is the wrong behavior: if you rereference to closed-by channels, spiking signals will be essentially killed. I think that we should still use the `local` radii, and keep the knn as backend when no channels are found. I think that was the original idea from #4193: to fix the case when local finds no channels. How does that sound?

In this update, backup will only use KNN when local finds not enough neighbors (default 5), and it will search from the inner radius for as many neighbors as available until the minimal requirement is met. KNN will not be available as a reference option either. The change becomes minimal compared with the previous version.

src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -8,6 +8,7 @@
 from spikeinterface.core.baserecording import BaseRecording
 
 from .filter import fix_dtype
+from functools import cache
```
alejoe91:
```suggestion
```

src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -117,10 +118,10 @@ def __init__(
             closest_inds, dist = get_closest_channels(recording)
             neighbors = {}
             for i in range(num_chans):
-                mask = (dist[i, :] > local_radius[0]) & (dist[i, :] <= local_radius[1])
+                mask = dist[i, :] > local_radius[0]
+                nn = np.cumsum(mask)
+                mask &= (dist[i, :] <= local_radius[1]) | ((0 < nn) & (nn <= min_local_neighbors))
```
alejoe91, line 123:
```suggestion
                mask = (dist[i, :] > local_radius[0]) & (dist[i, :] <= local_radius[1])
                if sum(mask) <  min_local_neighbors:
                    # Use min_local_neighbors closest neighbots
                    mask = dist[i, :] > local_radius[0]
                    nn = np.cumsum(mask)
                    mask &= (dist[i, :] <= local_radius[1]) | ((0 < nn) & (nn <= min_local_neighbors))
```

alejoe91 (approved):

Issue #4389, opened 2026-02-12
Decrease noise of computed "templates" (in particular, to improve reliability of peak detection)


m-beau:
2 options:
- fancy: use the waveforms that 1) peak on the same channel, 2) fall in the 70%-90% highest amplitude distribution to compute the template (see "drift-matching" in NeuroPyxels: https://github.com/m-beau/NeuroPyxels#preprocess-your-waveforms-drift-shift-matching-and-spike-trains-detect-periods-with-few-false-positivenegative)
- simple: increase N waveforms used to compute the template (in my experience, shouldn't go below 1000 if you don't have to)

Cheers!!

zm711:
I will close based on #4415. Feel free open a new issue if needed.

PR #4415, opened 2026-02-27
Set num_histogram_bins in amplitude_cutoff (back) to 100

ecobost:
#4353 changed the default `num_histogram_bins` in AmplitudeCutoff from 100 to 200 to get a more granular histogram/estimate.
https://github.com/SpikeInterface/spikeinterface/blob/797f0e876ef5835dc212a5274cfaa0e698a6c8b5/src/spikeinterface/metrics/quality/misc_metrics.py#L1034

But given the default `amplitudes_bins_min_ratio`=5, any cell with less than 1000 spikes will return amplitude_cutoff = NaN. That is (I think?) too big a hurdle (a unit with 0.1Hz firing rate will require 2.8 hours to spike 1000 times). 

Assuming 500 spikes are enough to model the amplitude distribution, I change it back to 100 here. I also change the default value on the low-level `amplitude_cutoff()`to match this default.

EDIT: Even if we don't think 500 spikes are enough to model the distribution, the right parameter to change would be `amplitudes_bins_min_ratio` and not `num_histogram_bins` as we would rather have more samples per bin (lower variance)

### DIFF
--- a/src/spikeinterface/metrics/quality/misc_metrics.py
+++ b/src/spikeinterface/metrics/quality/misc_metrics.py
@@ -1024,7 +1024,9 @@ def compute_amplitude_cutoffs(
     units_with_few_spikes = [unit_id for unit_id, amp_cutoff in all_fraction_missing.items() if np.isnan(amp_cutoff)]
     if len(units_with_few_spikes) > 0:
         min_num_spikes = amplitudes_bins_min_ratio * num_histogram_bins
-        warnings.warn(f"Amplitude cutoff set to NaN for units {units_with_few_spikes}: too few spikes (< {min_num_spikes}).")
+        warnings.warn(
+            f"Amplitude cutoff set to NaN for units {units_with_few_spikes}: too few spikes (< {min_num_spikes})."
+        )
 
     return all_fraction_missing
 


zm711:
If we go messing with AmplitudeCutoff and I don't get all NaN's I'll have no way to know my code is working :P 

Isn't our default for `random_spikes` 500 [see here](https://github.com/SpikeInterface/spikeinterface/blob/797f0e876ef5835dc212a5274cfaa0e698a6c8b5/src/spikeinterface/core/analyzer_extension_core.py#L40). So if 500 is enough sample for a distribution for the core of the analyzer then I would agree with you that 500 should be enough here too!

alejoe91:
> If we go messing with AmplitudeCutoff and I don't get all NaN's I'll have no way to know my code is working :P
> 
> Isn't our default for `random_spikes` 500 [see here](https://github.com/SpikeInterface/spikeinterface/blob/797f0e876ef5835dc212a5274cfaa0e698a6c8b5/src/spikeinterface/core/analyzer_extension_core.py#L40). So if 500 is enough sample for a distribution for the core of the analyzer then I would agree with you that 500 should be enough here too!

Agreed! But on a related note  we should maybe increase random spikes to get better estimates of templates  #4389

ecobost:
For what is worth, I use:
```python
sa.create_sorting_analyzer(...,    num_spikes_for_sparsity=500) # default=100
analyzer.compute('random_spikes', max_spikes_per_unit=1000) # default 500
analyzer.compute('quality_metrics',  metric_params={
     'amplitude_cutoff': {'amplitudes_bins_min_ratio': 7.5}}) #i.e., 750 spikes, default is 5
```
Increasing the random_spikes to 1000 by default seems reasonable.

PR #4421, opened 2026-03-03
Rename `CurationModel` to `Curation`

alejoe91:
Following the new `SequentialCuration`, @samuelgarcia suggested there is no need for the `Model` in the name

### DIFF
--- a/src/spikeinterface/curation/curation_model.py
+++ b/src/spikeinterface/curation/curation_model.py
@@ -1,6 +1,8 @@
-from pydantic import BaseModel, Field, model_validator, field_validator, field_serializer
+import warnings
 from typing import Literal, List
 from itertools import chain, combinations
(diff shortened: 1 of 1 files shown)


src/spikeinterface/curation/curation_model.py
```diff
@@ -74,7 +74,7 @@ def get_full_spike_indices(self, sorting: BaseSorting):
         return full_spike_indices
 
 
-class CurationModel(BaseModel):
```
chrishalcrow, line 77: Should we deprecate? We use `CurationModel` in our lab's spike sorting pipeline (though I'd guess I'm a power user).
Would also make it easier on the e.g. si-gui side

alejoe91: Yes we definitely should! I'll add a fake function with a deprecation warning

PR #4426, opened 2026-03-06
Fix openephys tests

alejoe91:
The failure is due to a mismatch in the channel names between the settings and structure oebin in 2 test files on GIN.
Added a safeguard mechanism to prevent failures

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/openephys.py
+++ b/src/spikeinterface/extractors/neoextractors/openephys.py
@@ -351,13 +351,19 @@ class OpenEphysBinaryRecordingExtractor(NeoBaseRecordingExtractor):
                     # Ensure device channel index corresponds to channel_ids
                     probe_channel_names = probe.contact_annotations.get("channel_name", None)
                     if probe_channel_names is not None and not np.array_equal(probe_channel_names, self.channel_ids):
-                        device_channel_indices = []
-                        probe_channel_names = list(probe_channel_names)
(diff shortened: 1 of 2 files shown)


src/spikeinterface/extractors/neoextractors/openephys.py
```diff
@@ -351,13 +351,19 @@ def __init__(
...
+                            probe_channel_names = list(probe_channel_names)
+                            device_channel_indices = np.zeros(len(self.channel_ids), dtype=int)
+                            for i, ch in enumerate(self.channel_ids):
+                                index_in_probe = probe_channel_names.index(ch)
+                                device_channel_indices[index_in_probe] = i
+                            probe.set_device_channel_indices(device_channel_indices)
+                        else:
+                            warnings.warn(
+                                "Channel names in the probe do not match the channel ids from Neo. "
+                                "Cannot set device channel indices, but this might lead to incorrect probe geometries"
```
chrishalcrow, line 365: Hello, I don't understand this fix. This now creates a probe with no `device_channel_indices`, which then can't be attached to any recordings?

chrishalcrow: Oh, wait, so probeinterface will set `device_channel_indices` just basically using a range. So the probes that hit this warning so still have `device_channel_indices`?

alejoe91: yep, exactly! So in this case it will work as before

chrishalcrow (approved):

Issue #4436, opened 2026-03-12
`mp.cpu_count()` not reliable on HPC


chrishalcrow:
PR #4305, added a `fast_mode` for computing correlograms, which uses `mp.cpu_count()` to figure out how many threads to use.

On my HPC, this counts ALL the CPUs (hundreds) on the node rather than the ones in my instance (8).

Is there something else we could do @yger @samuelgarcia ? I can just turn `fast_mode` off for now.

EDIT: I can't turn fast mode off for the correlogram computation in `slidingRP_violations`.

yger:
Arg, damned. I'll look into that, other option would be to provide a number here, and let the user decide how many CPU to use fast_mode could be -1, 0 or a number. But then, with your HPC, how does it work if you use global n_jobs params from Spikeinterface ?

chrishalcrow:
It's fine if I set the `n_jobs` (if it's not -1). Any functions which try and guess the CPU or memory available tend to fail.

yger:
Ok, so basically you would be happy with a n_jobs option instead of fast_mode, am I right?

chrishalcrow:
Yup yup.

yger:
This has been superseeded by #4442

alejoe91:
Fixed by #4442

PR #4437, opened 2026-03-12
Changing fast_mode into n_jobs

yger:
This is an attempt to rename fast_mode into n_jobs, in order to make everything compatible with HPC (see #4436). 

However, there are two options:
- keep fast_mode True/False and use the n_jobs from si.get_global_job_kwargs()
- make an explicit n_jobs for this function (as done here in this PR). 

In all cases, we should highlight that using n_jobs > 1 makes the results slightly underterministic


yger:
The pros for option 1 would be to keep the fast_mode argument and syntax, just relying on globals for optimization. Let's discuss

PR #4442, opened 2026-03-13
Patching the Fast mode for hpc

yger:
Better implementation after discussion with @samuelgarcia . We keep the fast_mode argument, but rely on job_kwargs to determine the number of threads to use


chrishalcrow:
Thanks all!

PR #4443, opened 2026-03-15
Fix closest_channels sparsity estimation when n_units > n_channels

### LINKED ISSUE #4126: `IndexError: index 384 is out of bounds` when running estimate_sparsity() with method `closest_channels`
I ran into an indexing error when estimating the sparsity with `method=closest_channels`. 

The error I got was:

`IndexError: index 384 is out of bounds for axis 0 with size 384`

And it happens on line 351 of spikeinterface/core/sparsity.py

https://github.com/SpikeInterface/spikeinterface/blob/338a9f8ac1555155b068c627ae71a0153864792f/src/spikeinterface/core/sparsity.py#L347C1-L353C45

The issue seems to be that `unit_ids` is indexing the pairwise channel `distances`, and there are more units than channels:

```
channel_locations = templates_or_sorting_analyzer.get_channel_locations()
distances = np.linalg.norm(channel_locations[:, np.newaxis] - channel_locations[np.newaxis, :], axis=2)

for unit_ind, unit_id in enumerate(templates_or_sorting_analyzer.unit_ids):
    chan_inds = np.argsort(distances[unit_ind])
    chan_inds = chan_inds[:num_channels]
    mask[unit_ind, chan_inds] = True
```

I might be mistaken, but shouldn't the code be something more like `from_radius()`, which uses the `best_chan` to index the pairwise `distances`:

```
channel_locations = templates_or_sorting_analyzer.get_channel_locations()
distances = np.linalg.norm(channel_locations[:, np.newaxis] - channel_locations[np.newaxis, :], axis=2)
best_chan = get_template_extremum_channel(templates_or_sorting_analyzer, peak_sign=peak_sign, outputs="index")

for unit_ind, unit_id in enumerate(templates_or_sorting_analyzer.unit_ids):
    chan_ind = best_chan[unit_id]
    chan_inds = np.argsort(distances[chan_ind])
    chan_inds = chan_inds[:num_channels]
    mask[unit_ind, chan_inds] = True
```

ueeseer:
This fixes #4126 in `estimate_sparsity(method="closest_channels")`.

The previous implementation indexed the channel-distance matrix using `unit_ind`, which can fail when the number of units exceeds the number of channels. This patch instead selects the nearest channels starting from each unit’s extremum channel, aligning `closest_channels` with the existing `from_radius()` logic.

It also adds a targeted regression test covering the `n_units > n_channels` case.

## Local validation
- `../.venv-spike/bin/pytest -q src/spikeinterface/core/tests/test_sparsity.py -k "estimate_sparsity or compute_sparsity"`
- `../.venv-spike/bin/pytest -q src/spikeinterface/core/tests/test_sparsity.py`

### DIFF
--- a/src/spikeinterface/core/sparsity.py
+++ b/src/spikeinterface/core/sparsity.py
@@ -331,7 +331,7 @@ class ChannelSparsity:
 
     ## Some convinient function to compute sparsity from several strategy
     @classmethod
-    def from_closest_channels(cls, templates_or_sorting_analyzer, num_channels):
+    def from_closest_channels(cls, templates_or_sorting_analyzer, num_channels, peak_sign="neg"):
         """
         Construct sparsity from N closest channels
         Use the "num_channels" argument to specify the number of channels.
@@ -342,6 +342,8 @@ class ChannelSparsity:
             A Templates or a SortingAnalyzer object.
         num_channels : int
             Number of channels for "best_channels" method.
+        peak_sign : "neg" | "pos" | "both"
+            Sign of the template to compute best channels.
 
         Returns
         -------
@@ -355,13 +357,13 @@ class ChannelSparsity:
         )
         channel_locations = templates_or_sorting_analyzer.get_channel_locations()
         distances = np.linalg.norm(channel_locations[:, np.newaxis] - channel_locations[np.newaxis, :], axis=2)
-        best_chan = get_template_extremum_channel(templates_or_sorting_analyzer, peak_sign="neg", outputs="index")
+        best_chan = get_template_extremum_channel(templates_or_sorting_analyzer, peak_sign=peak_sign, outputs="index")
 
         for unit_ind, unit_id in enumerate(templates_or_sorting_analyzer.unit_ids):
             chan_ind = best_chan[unit_id]
             chan_inds = np.argsort(distances[chan_ind])
-            chan_inds = chan_inds[:num_channels]
-            mask[unit_ind, chan_inds] = True
+            closest_chan_inds = chan_inds[:num_channels]
+            mask[unit_ind, closest_chan_inds] = True
         return cls(mask, templates_or_sorting_analyzer.unit_ids, templates_or_sorting_analyzer.channel_ids)
 
     @classmethod


ueeseer:
I rechecked this on the exact current PR head (`e22a2c5`) before changing anything.

This PR only touches `closest_channels` sparsity selection in `ChannelSparsity.from_closest_channels()`. The curation fixture behind `test_model_based_classification_predict_labels` builds its analyzer with `create_sorting_analyzer(..., sparse=True)`, which uses the default radius-based sparsity path rather than `closest_channels`.

I also ran `test_model_based_classification_predict_labels` locally on both `origin/main` and the exact current PR head (`e22a2c5`), and it passes in both cases. The curation fixture sparsity masks were also identical in that local comparison.

Given that the only failing cell is `macos-latest / Python 3.13`, and the observed failure is a one-label prediction drift (`[1, 0, 1, 0, 1] -> [1, 0, 1, 0, 0]`) rather than a crash or invariant break, this does not currently look causally connected to the `#4126` fix. At this point it looks more like an isolated platform-specific prediction drift than a regression in the touched logic.

src/spikeinterface/core/sparsity.py
```diff
@@ -348,16 +348,18 @@ def from_closest_channels(cls, templates_or_sorting_analyzer, num_channels):
...
         """
-        from .template_tools import get_template_amplitudes
+        from .template_tools import get_template_extremum_channel
 
         mask = np.zeros(
             (templates_or_sorting_analyzer.unit_ids.size, templates_or_sorting_analyzer.channel_ids.size), dtype="bool"
         )
         channel_locations = templates_or_sorting_analyzer.get_channel_locations()
         distances = np.linalg.norm(channel_locations[:, np.newaxis] - channel_locations[np.newaxis, :], axis=2)
+        best_chan = get_template_extremum_channel(templates_or_sorting_analyzer, peak_sign="neg", outputs="index")
```
alejoe91: We should move `peak_sign` to the args

alejoe91:
Thanks for picking this up @ueeseer 

Just exposed the `peak_sign` param as a final touch :)

PR #4447, opened 2026-03-16
Improve robustness of model classifier tests

chrishalcrow:
Our model based curation tests are flakey, leading to annoying test failures (e.g. #4443).

This PR makes the tests more robust by training the curation model on data which should be easy to classify: ten good units and ten junk units.

EDIT: it's difficult to seed this fully, because of the hyperparameter search in the model training. I've run the tests 1000 times and they pass. At the very least, this is easier to make even more robust in the future if we keep having problems...


src/spikeinterface/curation/tests/common.py
```diff
@@ -68,7 +68,15 @@ def make_sorting_analyzer_with_splits(sorting_analyzer, num_unit_splitted=1, num
 
 @pytest.fixture(scope="module")
 def sorting_analyzer_for_curation():
```
alejoe91, line 70: this is used in many other tests!

PR #4454, opened 2026-03-17
Switch to kernel for local CAR

alejoe91:
Following up on #4412

Added a "kernel" mode for local CAR: when computing the common average referencing, we can simply pre-compute a kernel with 1/local_neighbors for each channel and use it in the `get_traces` with a simple dot product, skipping an inefficient channel loop.

This makes the `get_traces` x10 faster (added a skipped test to test locally):
```
Local CMR time: 11.6428 seconds
Local CAR time: 1.3863 seconds
```

### DIFF
--- a/src/spikeinterface/preprocessing/common_reference.py
+++ b/src/spikeinterface/preprocessing/common_reference.py
@@ -117,12 +117,6 @@ class CommonReferenceRecording(BasePreprocessor):
                     [ch in recording.channel_ids for ch in ref_channel_ids]
                 ), "Some 'ref_channel_ids' are wrong!"
         elif reference == "local":
-            if operator == "median":
-                warnings.warn(
-                    "Using a local median reference can be very computationally intensive. Consider using a local "
-                    "average reference instead or pre-computing the local median reference and using the 'single' "
-                    "reference option."
-                )
             assert groups is None, "With 'local' CAR, the group option should not be used."
             closest_inds, dist = get_closest_channels(recording)
             # The neighbor kernel is a matrix that will be used to calculate the local reference.


src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -195,19 +218,23 @@ def get_traces(self, start_frame, end_frame, channel_indices):
                     shift = self.operator_func(traces, axis=1, keepdims=True)
                 else:
                     shift = self.operator_func(traces[:, self.ref_channel_indices], axis=1, keepdims=True)
-                re_referenced_traces = traces[:, channel_indices] - shift
+                re_referenced_traces = traces[:, channel_indices] - shift  # shift[:, channel_indices]
```
alejoe91:
```suggestion
                re_referenced_traces = traces[:, channel_indices] - shift
```

src/spikeinterface/preprocessing/tests/test_common_reference.py
```diff
@@ -35,21 +40,27 @@ def test_common_reference(recording):
     assert not np.all(rec_sin.get_traces()[0])
     assert np.allclose(rec_sin.get_traces()[:, 1], traces[:, 1] - traces[:, 0])
 
-    assert np.allclose(traces[:, 0], rec_local_car.get_traces()[:, 0] + np.median(traces[:, [2, 3]], axis=1), atol=0.01)
-    assert np.allclose(traces[:, 1], rec_local_car.get_traces()[:, 1] + np.median(traces[:, [3]], axis=1), atol=0.01)
+    assert np.allclose(traces[:, 0], rec_local_cmr.get_traces()[:, 0] + np.median(traces[:, [2, 3]], axis=1), atol=0.01)
+    assert np.allclose(traces[:, 1], rec_local_cmr.get_traces()[:, 1] + np.median(traces[:, [3]], axis=1), atol=0.01)
+
+    # TODO: fix this!!!
```
chrishalcrow: ???

alejoe91: It's fixed, I was missing a transpose in the code :P

alejoe91:
```suggestion
```

src/spikeinterface/preprocessing/common_reference.py
```diff
@@ -114,14 +117,36 @@ def __init__(
                     [ch in recording.channel_ids for ch in ref_channel_ids]
                 ), "Some 'ref_channel_ids' are wrong!"
         elif reference == "local":
+            if operator == "median":
+                warnings.warn(
+                    "Using a local median reference can be very computationally intensive. Consider using a local "
+                    "average reference instead or pre-computing the local median reference and using the 'single' "
+                    "reference option."
+                )
```
chrishalcrow: I find this warning a bit annoying, since you can't turn it off if you use `local + median`. I also don't understand the second part of the warning. It sounds like I can pre-compute the reference and pass it? Can I??

alejoe91: Yeah the second part doesn't make sense. Shoudl I just remove it?

alejoe91:
```suggestion
```

Issue #4394, opened 2026-02-16
Improper Probe Configuration when loading OpenEphys


SplintZ:
Hello I was working on a spikesorting pipeline using kilosort4 with some automatic curation and postprocessing thrown into the mix.
So far I thought everything was looking good, but when inspecting the units in phy I noticed that sometimes the waveforms abruptly cut-off from one channel to the next. This seems to be an issue that comes from the Probe that is automatically generated by `read_openephys()` that then gets propagated to kilosort and finally to phy.

<img width="1234" height="771" alt="Image" src="https://github.com/user-attachments/assets/b3c8e62d-8e2e-41bd-b5c7-56692234179c" />

I asked on the neuropixels discord and they first thought it was just an issue with how phy was gathering the waveforms and that under the hood actually everything is fine. So I plotted the average waveform of an example unit and aligned it to the probe geometry. I just calculated an average waveform for the unit displayed in phy and when I looks at the waveforms in that I get from `analyzer.get_extension("waveforms").get_waveforms_one_unit(116)` the order there matches the order on the probe. However when I tried to plot the waveforms at the position where they are on the probe (using `recording.get_channel_location()`) the order is completely messed up.

<img width="513" height="874" alt="Image" src="https://github.com/user-attachments/assets/75863993-c299-4bef-97ae-eeeaec071374" />

I added my stucture.oebin file for all the recordings (had to rename it to json so i can upload it to github)
[structure.json](https://github.com/user-attachments/files/25347276/structure.json)

If anyone knows how to deal with this issue I would be very grateful

chrishalcrow:
Hello, I think it is probably a downstream/phy issue, rather than something fundamental. If the locations really were messed up, kilosort wouldn't have managed to sort it. Likely something silly about channel ids and channel indices.

To help debug, could you share the code you used to plot the channel locations? Also the spikeinterface and probeinterface versions you're using in your environment. Thanks!

I'd also recommend giving spikeinterface-gui a go (https://github.com/SpikeInterface/spikeinterface-gui/), but I am biased! (The main reason is that phy is no longer maintained, so it's likely to become unsupported in the nearish future)

SplintZ:
I used the latest version of spikeinterface (0.103.2) and probeinterface (0.3.1) and the plot was generated with the following code.

```python
sorting = se.read_kilosort(ks_path)
recording = se.read_openephys(session_path, stream_id="1")

[...]

PR #4456, opened 2026-03-17
Update `read_openephys` and add OneBox tests

alejoe91:
### Update

Requires https://github.com/SpikeInterface/probeinterface/pull/409

In https://github.com/SpikeInterface/probeinterface/pull/409 we improved the matching and sorting of device channel indices when a channel map is present (as in #4394). Now te probeinterface reader handles the re-ordering of contacts, so we don't need any additional logic on the SI side.

Added tests for OneBox, including the "non-trivial" wiring. In the latter test, we check that the recording channel ids (from the oebin) correspond to the `settings_channel_key` annotation, which has been sorted according to the channel map

> ### Original

> Requires https://github.com/SpikeInterface/probeinterface/pull/406
>
> In https://github.com/SpikeInterface/probeinterface/pull/406 we improved the matching and sorting of device channel indices using the `electrode_index` field in the oebin file. This PR simply looks for the oebin file and passes it to the probeinterface reader

### DIFF
--- a/src/spikeinterface/postprocessing/amplitude_scalings.py
+++ b/src/spikeinterface/postprocessing/amplitude_scalings.py
@@ -201,6 +201,9 @@ class AmplitudeScalingNode(PipelineNode):
             max_margin_collisions = delta_collision_samples + margin_waveforms
             self._margin = max_margin_collisions
 
+        # for some edge cases a template can be zero, leading to problems later
+        template_is_zero = [np.all(template == 0) for template in all_templates]
+
         self._all_templates = all_templates
         self._sparsity_mask = sparsity_mask
         self._nbefore = nbefore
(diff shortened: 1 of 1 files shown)


h-mayorquin:
Maybe we can add a test here:
https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/169

src/spikeinterface/extractors/neuropixels_utils.py
```diff
@@ -4,17 +4,14 @@
 from probeinterface import Probe
 
 
-def get_neuropixels_sample_shifts_from_probe(probe: Probe, stream_name: str = "ap") -> np.ndarray:
+def get_neuropixels_sample_shifts_from_probe(probe: Probe) -> np.ndarray:
```
h-mayorquin, line 7: wouldn't this break for some people if they are using this function? if we think they are not should we make this private?

alejoe91: it was an unused argument leftover from the past. I think it's ok to remove ;)

PR #4457, opened 2026-03-18
Adding the possibility to inject precomputed results in study

yger:
While developing our online sorters, we want to use studies with data generated outside of SI. This allows to inject precomputed results for custom cases, and still use the machinery of the Studies

### DIFF
--- a/src/spikeinterface/benchmark/benchmark_base.py
+++ b/src/spikeinterface/benchmark/benchmark_base.py
@@ -293,7 +293,7 @@ class BenchmarkStudy:
         (self.folder / "cases.pickle").write_bytes(pickle.dumps(self.cases))
 
     def set_precomputed_results(self, precomputed_results, verbose=False):
-        """Set precomputed results for some cases. This is usefull when you want to compute results outside of the benchmark and
+        """Set precomputed results for some cases. This is useful when you want to compute results outside of the benchmark and
         then set them in the benchmark.
 
         Parameters


src/spikeinterface/benchmark/benchmark_base.py
```diff
@@ -292,6 +292,22 @@ def remove_cases(self, case_keys):
             self.remove_benchmark(key)
         (self.folder / "cases.pickle").write_bytes(pickle.dumps(self.cases))
 
+    def set_precomputed_results(self, precomputed_results, verbose=False):
```
alejoe91, line 295: This is super useful!

Can you add a docstring? What does the precomputed_results expect?

yger: done. Precomputed_results expected the same results as a benchmark (depending on which one). This is a bit advanced users only, i must admit, since this is not properly exposed

src/spikeinterface/benchmark/benchmark_base.py
```diff
@@ -292,6 +292,39 @@ def remove_cases(self, case_keys):
...
+        """Set precomputed results for some cases. This is usefull when you want to compute results outside of the benchmark and
```
alejoe91:
```suggestion
        """Set precomputed results for some cases. This is useful when you want to compute results outside of the benchmark and
```

PR #4462, opened 2026-03-23
Centralize `add_segment` and `segments` to `BaseExtractor`

alejoe91:
and fix a bunch of typing errors

Pulled out of #4216

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -168,6 +168,11 @@ class BaseRecording(BaseRecordingSnippets):
 
         return SubtractRecordings(self, other)
 
+    @property
+    def segments(self) -> list["BaseRecordingSegment"]:
+        """List of recording segments."""
+        return self._segments
+
     def add_recording_segment(self, recording_segment: "BaseRecordingSegment") -> None:
         """Adds a recording segment.
 
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -61,6 +61,11 @@ class BaseSorting(BaseExtractor):
         return html_repr
 
     @property
+    def segments(self) -> list["BaseSortingSegment"]:
+        """List of sorting segments."""
+        return self._segments
+
+    @property
     def unit_ids(self):
         return self._main_ids
 


src/spikeinterface/core/base.py
```diff
@@ -142,11 +142,18 @@ def name(self, value):
...
+        self._segments.append(segment)
+        segment.set_parent_extractor(self)
+
     def get_num_segments(self) -> int:
-        # This is implemented in BaseRecording or BaseSorting
-        raise NotImplementedError
+        return len(self._segments)
 
-    def get_parent(self) -> BaseExtractor | None:
+    def get_parent(self) -> "BaseExtractor | None":
```
chrishalcrow, line 156: Hello. If you keep `from __future__ import annotations` you don't need to use string literals for self-referencing types. I think this is preferred, but might just be a style choice. At Python 3.11 we get a new type `Self` which is now the modern way to do self referencing.

alejoe91: Ah nice! I would say we keep the string notation and get rid of __future__ annotations. The main reason is that the `__future__` import doesn't actually do any check, so we had a lot of bad typing (e.g., lists of | instead of Literal). This enforces us to be cleaner ;)

src/spikeinterface/core/base.py
```diff
@@ -142,11 +142,18 @@ def name(self, value):
             # we remove the annotation if it exists
             _ = self._annotations.pop("name", None)
 
+    @property
```
h-mayorquin, line 145: Maybe for typing we could in BaseRecording for example:


```python
  @property
  def segments(self) -> list[BaseRecordingSegment]:
      return self._segments  # type: ignore[return-value]
```

As that will enable the analysis of base recording sgements methods that we call on introspection with vscode and ohter tools. Same for bas sorting.

h-mayorquin: Also, can we make this private? what is the reason for making this public?

src/spikeinterface/core/base.py
h-mayorquin: Once we support python only great than 3.14 we can drop the strings again I think.

src/spikeinterface/postprocessing/localization_tools.py
```diff
@@ -661,8 +661,9 @@ def get_convolution_weights(
 def compute_location_max_channel(
     templates_or_sorting_analyzer: SortingAnalyzer | Templates,
     unit_ids=None,
-    peak_sign: "neg" | "pos" | "both" = "neg",
-    mode: "extremum" | "at_index" | "peak_to_peak" = "extremum",
+    peak_sign: Literal["neg", "pos", "both"] = "neg",
+    mode: Literal["extremum", "at_index", "peak_to_peak"] = "extremum",
+    operator: Literal["average", "median"] = "average",
```
samuelgarcia, line 666: why this here ?

samuelgarcia:
I think I was more happy with the private approach
`sorting._sorting_segments` over `sorting.segments`
but this is OK for me.

And as @h-mayorquin 
the property segment could be moved to Sorting and recording for handling the correct type.


So ok for merging.

@chrishalcrow : did you check this ?

src/spikeinterface/postprocessing/localization_tools.py
alejoe91: because of typing errors...

src/spikeinterface/core/base.py
alejoe91: Adding it! I think it's handy to have it public. The segments are a big part of the API, so exposing them (as a property) doesn't hurt IMO

PR #4463, opened 2026-03-23
full bombcell pipeline example with exposed native parameters

Julie-Fabre:
Adds full bombcell pipeline example with exposed native parameters, unitrefine integration, and upsetplot-bombcell dependency. Work in progress.

Adding:
- Bombcell example script demonstrating the full pipeline, including configurable parameters and how to modify them.
     - This required changes to expose all parameters available in native bombcell: options to enable/disable drift and distance metrics, choice of RPV calculation method, time chunks etc.
     - adds two wrapper scripts to streamline and make parameters very apparent / easy to modify
     - Updates docs (need to check new refs for template metrics - are they all correct?)
     
- Switch to upsetplot-bombcell in place of upsetplot.


chrishalcrow:
Hello, looks great. Just to say, we can have docs where you run everything locally, then upload the results. I think this is the only example: https://spikeinterface.readthedocs.io/en/stable/how_to/drift_with_lfp.html
I think it could be a good fit for this case, cause you could actually run a full pipeline from scratch (maybe on data that's available on DANDI) and discuss the results that come out.
EDIT: https://spikeinterface.readthedocs.io/en/stable/how_to/handle_drift.html another!

Julie-Fabre:
Thanks a lot, Chris! I love that idea. I am going to work on finding a good example dataset and document a nice script describing outputs etc. That's a longer thing to do, so I'll do a separate PR soon.
For now, I think this is ready for review. Let me know what you think! In particular, I added citation info in template metric docs in references.rst. Could you check whether that looks good?

alejoe91:
Hi @Julie-Fabre 

Thanks for working on this! To keep things tidy and easy to review I recommend splitting this into multiple PRs :)

- Use upsetplot-bombcell dependency
- Extend bombcell options
- Add tutorials

There are also a lot of changes to core and some leftovers from previous PRs that should be cleaned up!

Julie-Fabre:
Ok, i'll do that. Thanks, Alessio!

Issue #3498, opened 2024-10-23
Method to return the original probe


h-mayorquin:
As I am taking a look I realize that the `get_probe` method returns the contact vector:

https://github.com/SpikeInterface/spikeinterface/blob/bfe9fb649b58c48d2a04949061154f75f8e08b1f/src/spikeinterface/core/baserecordingsnippets.py#L255-L282

I am looking at the `set_probe` and I don't think we save the probe as an object at all? Can I get it? Could we add something like that?

samuelgarcia:
by design we do not have the probe object attached.
The idea was to manipulate probe using the contact_vector.

alejoe91:
@h-mayorquin we store alle the probe metadata in probe info. Maybe the `get_probe` should add these metadata to the returned probe object. 

The returned object would not be the original one, but it would be a clone with proper name, manufacturer, model, contour, etc

h-mayorquin:
I think that @alejoe91 sounds like a good compromise. It should be a clone of the original probe, right?

zm711:
So this is closed by #4465?

h-mayorquin:
Yes, this is not necessary anymore, now that we have the probes and propagate the annotations on it I can get the manufacturer and the probe model to get the reconstruct the original probe if necessary.

PR #4465, opened 2026-03-24
Use `ProbeGroup` object instead of `contact_vector` property and  `set_probe/probegroup` from `select_channels_with_probe/probegroup`

### LINKED ISSUE #2193: `set_probe` behavior for `RecordingExtractors`

### LINKED ISSUE #4545: `aggregate_channels` loses per-probe metadata
Coming to this thinking on:
https://github.com/SpikeInterface/probeinterface/pull/420
https://github.com/SpikeInterface/probeinterface/pull/425

When a probe is attached to a recording via `set_probe`, `_set_probes` stores each probe's user-level metadata in two recording-level annotations: `probes_info` (a list where entry `i` is probe `i`'s `annotations` dict, with keys like `name`, `manufacturer`, `description`) and one `probe_{i}_planar_contour` annotation per probe. `aggregate_channels` does not propagate either of these into the combined recording. `probes_info` is never copied at all, so even the first child's per-probe annotations disappear. `probe_{i}_planar_contour` annotations are only kept when every child happens to share the exact same contour; otherwise they are all dropped. The result is that `combined.get_probegroup()` exposes probes with empty `annotations` dicts and no planar contour, regardless of what was attached to the children.

```python
import numpy as np
from probeinterface import generate_dummy_probe
from spikeinterface.core import generate_recording, aggregate_channels

def make_recording(probe_name, manufacturer, x_shift, contact_id_prefix):
    probe = generate_dummy_probe()
    probe.move([x_shift, 0.0])
    probe.annotate(name=probe_name, manufacturer=manufacturer)
    probe.set_device_channel_indices(np.arange(probe.get_contact_count()))
    probe.set_contact_ids([f"{contact_id_prefix}{i}" for i in range(probe.get_contact_count())])
    probe.create_auto_shape()
    rec = generate_recording(num_channels=probe.get_contact_count(), durations=[1.0], set_probe=False)
    return rec.set_probe(probe)

rec_A = make_recording("probe_A", "vendor_X", x_shift=0.0,    contact_id_prefix="a")
rec_B = make_recording("probe_B", "vendor_Y", x_shift=1000.0, contact_id_prefix="b")

print("rec_A annotation keys:", rec_A.get_annotation_keys())
print("rec_A probes_info:", rec_A.get_annotation("probes_info"))

combined = aggregate_channels([rec_A, rec_B])
print("combined annotation keys:", combined.get_annotation_keys())
print("combined probes_info:", combined.get_annotation("probes_info"))
print("combined probe_0_planar_contour:", combined.get_annotation("probe_0_planar_contour"))
```

Expected: the aggregate carries a `probes_info` list with one entry per child probe and `probe_{i}_planar_contour` annotations for each probe.

Observed on `main`:

```
rec_A annotation keys: ['is_filtered', 'name', 'probe_0_planar_contour', 'probes_info']
rec_A probes_info: [{'manufacturer': 'vendor_X', 'name': 'probe_A'}]
combined annotation keys: ['is_filtered']
combined probes_info: None
combined probe_0_planar_contour: None
```

alejoe91:
### Updated 7/1/26

@h-mayorquin after discussing with @samuelgarcia, we decided that 1 PR is ok

After https://github.com/SpikeInterface/probeinterface/pull/446, we can go all in on the use of `ProbeGroup` instead of `contact_vector`, since the use case with interleaved contacts after device channel indices slicing is handled gracefully by the new `_global_contact_order` array, stored in the probegroup dictionary. 
Now the `recording` has a `_probegroup` attribute, which is propagated as metadata. The `probegroup.get_slice` takes care of all metadata propagation. 
In this refactoring, the duplicated `location` property has also been dropped.

Note that this implementation is alternative to what is proposed in  #4553 and implemented in #4548 : here we keep a `ProbeGroup` object attached to the recording, after sorting it by `device_channel_indices`, instead of adding a `wiring` property. You lose the original probe contact order, but we argue that the contacts in a probegroup do not really have a "default" order, but are rather a set of contacts that become ordered when wired (see #3498).

Some additional points:

### Tests for aggregation/slicing

Added more tests for aggregation/slicing behavior and metadata propagation

Fixes #4545 #4546 #4547 #4549 #4553

### Tests for backward compatibility

Added a test for bakward compatibility, which saves a recording to binary/zarr/JSON/pickle with v0.102/3/4.* and reloads it with the current version to check that the probegroup is reloaded correctly.
This way we check that the new `handle_extractor_backward_compatibility()` handles all cases!

### `set_probe/probegroup` versus `select_channels_with_probe/probegroup`

This PR refactors the `set_probe` function to address several issues:
- `set_probe/probegroup`: now are only in place (the not in place is deprecated, but will be removed in 0.106.0): these functions now require that the connected contacts have the same length as the number of channels.
- added `select_channels_with_probe/probegroup` to handle the slicing of recordings with probes.

Fixes #2193


src/spikeinterface/core/base.py
```diff
@@ -434,6 +460,15 @@ def copy_metadata(
         if self._preferred_mp_context is not None:
             other._preferred_mp_context = self._preferred_mp_context
 
+        self._extra_metadata_copy(other)
+
+    def _extra_metadata_copy(self, other: BaseExtractor):
+        """
+        This is a hook to copy extra metadata that is not in the annotations/properties dict.
+        It is used for instance to copy the probe in the FrameSliceRecording.
```
alejoe91:
```suggestion
```

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -132,49 +182,52 @@ def _set_probes(self, probe_or_probegroup, group_mode="auto", in_place=False):
...
+        if num_probes > 1 and raise_if_overlapping_probes:
             check_probe_do_not_overlap(probegroup.probes)
 
         # handle not connected channels
         assert all(
             probe.device_channel_indices is not None for probe in probegroup.probes
         ), "Probe must have device_channel_indices"
 
-        # this is a vector with complex fileds (dataframe like) that handle all contact attr
+        # TODO: add get_slice for probegroup to handle not connected channels
```
alejoe91:
```suggestion
```
Done in https://github.com/SpikeInterface/probeinterface/pull/416

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -187,21 +240,9 @@ def _set_probes(self, probe_or_probegroup, group_mode="auto", in_place=False):
...
-                sub_recording.set_annotation(f"probe_{probe_index}_planar_contour", contour, overwrite=True)
-
-        # duplicate positions to "locations" property
-        ndim = probegroup.ndim
-        locations = np.zeros((probe_as_numpy_array.size, ndim), dtype="float64")
-        for i, dim in enumerate(["x", "y", "z"][:ndim]):
-            locations[:, i] = probe_as_numpy_array[dim]
-        sub_recording.set_property("location", locations, ids=None)
+        # # create a vector that handle all contacts in property
+        # sub_recording.set_property("contact_vector", probe_as_numpy_array, ids=None)
```
alejoe91:
```suggestion
        # Set probegroup
```

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -330,51 +395,55 @@ def set_dummy_probe_from_locations(self, locations, shape="circle", shape_params
...
+        contact_positions = probegroup.get_global_contact_positions()
+        return select_axes(contact_positions, axes)[channel_indices]
 
-    def has_3d_locations(self) -> bool:
-        return self.get_property("location").shape[1] == 3
+    def is_probe_3d(self) -> bool:
+        if not self.has_probe():
+            raise ValueError("is_probe_3d() needs a probe to be attached to the recording")
+        probe = self.get_probegroup().probes[0]
+        return probe.ndim == 3
```
alejoe91:
```suggestion
        probegroup = self.get_probegroup()
        return probegroup.ndim == 3
```

src/spikeinterface/core/channelslice.py
```diff
@@ -152,10 +153,9 @@ def __init__(self, parent_snippets, channel_ids=None, renamed_channel_ids=None):
         parent_snippets.copy_metadata(self, only_main=False, ids=self._channel_ids)
 
         # change the wiring of the probe
-        contact_vector = self.get_property("contact_vector")
-        if contact_vector is not None:
-            contact_vector["device_channel_indices"] = np.arange(len(channel_ids), dtype="int64")
-            self.set_property("contact_vector", contact_vector)
+        if self._parent_snippets.has_probe():
+            parent_probegroup = self._parent_snippets.get_probegroup()
+            self.set_probe(parent_probegroup.get_slice(self._parent_channel_indices))
```
alejoe91:
```suggestion
            parent_probegroup = self._parent_snippets.get_probegroup()
            sliced_probegroup = parent_probegroup.get_slice(self._parent_channel_indices)
            sliced_probegroup.set_global_device_channel_indices(np.arange(len(self._channel_ids)))
            self.set_probegroup(sliced_probegroup, in_place=True)
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -365,7 +365,6 @@ def create(
                 )
         # check that multiple probes are non-overlapping
         all_probes = recording.get_probegroup().probes
-        check_probe_do_not_overlap(all_probes)
```
alejoe91, line 368: This is removed because when we `split_by` and aggregate probes might overlap

h-mayorquin:
I would like to take a look to this. I can do it this week.

samuelgarcia:
I guess that @zm711 and @h-mayorquin are now dancing on their seat just reading this PR.

Let me read it slowly.

h-mayorquin (commented):
As we discussed previously I think this is a step in the direction direction.

Is `_build_probegroup_from_properties` only for backwwards compatbility? 

I think we can segregate the backward-compatibility serialization logic (`_build_probegroup_from_properties`) from the main class methods by moving it into the deserialization hooks. `_extra_metadata_from_dict` already runs after properties are set (`base.py:1202`), so it can check for `"probegroup"` first and fall back to reconstructing from `contact_vector`. For the folder path, `load_metadata_from_folder` (`base.py:639`) needs to swap lines 643 and 646 so `.npy` properties load before `_extra_metadata_from_folder` runs, giving the hook access to legacy `contact_vector` when there is no `probe.json`. This has the side effect of a cleaner design where `has_probe()` becomes `return self._probegroup is not None` with no mutation at check time.

alejoe91:
>As we discussed previously I think this is a step in the direction direction.
>
> Is _build_probegroup_from_properties only for backwwards compatbility?
>
> I think we can segregate the backward-compatibility serialization logic (_build_probegroup_from_properties) from the main class methods by moving it into the deserialization hooks. _extra_metadata_from_dict already runs after properties are set (base.py:1202), so it can check for "probegroup" first and fall back to reconstructing from contact_vector. For the folder path, load_metadata_from_folder (base.py:639) needs to swap lines 643 and 646 so .npy properties load before _extra_metadata_from_folder runs, giving the hook access to legacy contact_vector when there is no probe.json. This has the side effect of a cleaner design where has_probe() becomes return self._probegroup is not None with no mutation at check time.
> _extra_metadata_from_folder

Let's check also how zarr does it. If we don't have a separate function for zarr, we can also add it to `__init__`

src/spikeinterface/core/baserecording.py
```diff
@@ -616,8 +616,9 @@ def _save(self, format="binary", verbose: bool = False, **save_kwargs):
                 offset_to_uV=self.get_channel_offsets(),
             )
             binary_rec.dump(folder / "binary.json", relative_to=folder)
-
-            from .binaryfolder import BinaryFolderRecording
+            if self.has_probe():
+                probegroup = self.get_probegroup()
+                write_probeinterface(folder / "probe.json", probegroup)
```
alejoe91: this is not needed, handled by the `save_to_folder`

alejoe91:
> > As we discussed previously I think this is a step in the direction direction.
> > Is _build_probegroup_from_properties only for backwwards compatbility?
> > I think we can segregate the backward-compatibility serialization logic (_build_probegroup_from_properties) from the main class methods by moving it into the deserialization hooks. _extra_metadata_from_dict already runs after properties are set (base.py:1202), so it can check for "probegroup" first and fall back to reconstructing from contact_vector. For the folder path, load_metadata_from_folder (base.py:639) needs to swap lines 643 and 646 so .npy properties load before _extra_metadata_from_folder runs, giving the hook access to legacy contact_vector when there is no probe.json. This has the side effect of a cleaner design where has_probe() becomes return self._probegroup is not None with no mutation at check time.
> > _extra_metadata_from_folder
> 
> Let's check also how zarr does it. If we don't have a separate function for zarr, we can also add it to `__init__`

@h-mayorquin actually both the `save_to_folder` and `save_to_zarr` save the probegroup already to json/attrs. So I think that removing the `contact_vector` entirely would still keep backcompatibility. I'll give it a try with a comment!

alejoe91:
Depends on https://github.com/SpikeInterface/probeinterface/pull/420

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -232,61 +269,89 @@ def _set_probes(self, probe_or_probegroup, group_mode="auto", in_place=False):
...
+        if self.has_probe():
             probegroup = self.get_probegroup()
-            write_probeinterface(folder / "probe.json", probegroup)
+            write_probeinterface(folder / "probegroup.json", probegroup)
+
+    def _extra_metadata_from_dict(self, dump_dict):
+        # load probe
+        if "probegroup" in dump_dict:
+            probegroup = dump_dict["probegroup"]
+            self.set_probegroup(probegroup, in_place=True)
```
alejoe91, line 348: Add contact_vector check+parsing here to cover JSON dumping/loading

alejoe91:
@h-mayorquin @samuelgarcia this is ready to review. See updated PR description

src/spikeinterface/core/binaryfolder.py
```diff
@@ -3,6 +3,8 @@
 
 import numpy as np
 
+from probeinterface import read_probeinterface
```
h-mayorquin, line 6: dead code I think.

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -51,242 +53,312 @@ def has_scaleable_traces(self) -> bool:
...
+                "Use the `recording.select_channels_with_probegroup()` method instead to return a new recording with "
+                "a channel selection to match the probe/probegroup.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            if not in_place:
+                return self.select_channels_with_probegroup(probegroup, group_mode=group_mode)
+
+        # Handle several input possibilities: Probe or dict
+        if isinstance(probegroup, dict):
```
h-mayorquin: I think this should be handled in `_extra_metadata_from_dict` instead of here. This will avoid the polymorphism of the type in the signature and it will corral the serialization machinery to the serialization path and simplify the function that people are more likely to see.

h-mayorquin: There is also a strange behavior here when you pass a dict with with in_place=False which would be solve by this.

h-mayorquin:
I am OK with all of this but there are a couple of features here:

1. Attached ProbeGroup instead of contact_vector. 
2. `set_probe`/`set_probegroup` become in-place-only, plus the new `select_channels_with_probe`/`select_channels_with_probegroup` (effectively #2798, breaking).
3. Geometry only through the probe instead of the location probe

The three look good to me but we can also make them as separate PRs if we don't feel OK about the whole bundle. Most of the diff churn comes from 2 I think.

Some comments:
* `delete_annotation` (no callers) and the `average_across_direction` float32 change look unrelated to the probe work and could be split out to make the diff simpler.
* The machinery for checking backwards compatibility looks OK but may be missing two cases. I would like to have more general infra for this as I mentioned in #4588 (https://github.com/SpikeInterface/spikeinterface/pull/4588#issuecomment-4696753516). Out of scope here but I am glad that we are converging on the idea that something like this is necessary.

alejoe91:
I agree @h-mayorquin 

Let me split this into 2 PRs:
1+3 (which is very small) and then 2 ;)

That will make it easier to review. We can also postpone 2, since it might need some more discussion

src/spikeinterface/preprocessing/average_across_direction.py
```diff
@@ -63,7 +63,7 @@ def __init__(
         # my geometry
         channel_locations = np.zeros(
             (n_pos_unique, parent_channel_locations.shape[1]),
-            dtype=parent_channel_locations.dtype,
+            dtype=np.float32,
```
samuelgarcia, line 66: ???

src/spikeinterface/core/zarrextractors.py
```diff
@@ -177,15 +177,18 @@ def __init__(
                 total_nbytes_stored += nbytes_stored_segment
 
         # load probe
-        probe_dict = self._root.attrs.get("probe", None)
+        probe_dict = self._root.attrs.get("probegroup", self._root.attrs.get("probe", None))
```
samuelgarcia: We need to catch the version and potentially reorder for old version here

src/spikeinterface/core/sortinganalyzer.py
samuelgarcia: I think this was important.
At least lets make a flag to control this.

src/spikeinterface/core/recording_tools.py
```diff
@@ -683,44 +684,60 @@ def order_channels_by_depth(recording, channel_ids=None, dimensions=("x", "y"),
     return order_f, order_r
 
 
-def check_probe_do_not_overlap(probes):
```
samuelgarcia, line 686: lets keep the function anyway

src/spikeinterface/core/recording_tools.py
```diff
@@ -683,44 +684,60 @@ def order_channels_by_depth(recording, channel_ids=None, dimensions=("x", "y"),
...
+def _set_group_property_based_on_probegroup(
```
samuelgarcia, line 727: I am not sure this is a good idea.
We remove channel_location no from property no ?
Why putting all the other ones.
lets discuss this more

src/spikeinterface/core/channelslice.py
```diff
@@ -62,10 +62,11 @@ def __init__(self, parent_recording, channel_ids=None, renamed_channel_ids=None)
...
         # change the wiring of the probe
-        contact_vector = self.get_property("contact_vector")
-        if contact_vector is not None:
-            contact_vector["device_channel_indices"] = np.arange(len(channel_ids), dtype="int64")
-            self.set_property("contact_vector", contact_vector)
+        if self._parent.has_probe():
+            parent_probegroup = self._parent.get_probegroup()
+            sliced_probegroup = parent_probegroup.get_slice(self._parent_channel_indices)
+            sliced_probegroup.set_global_device_channel_indices(np.arange(len(self._channel_ids)))
+            self.set_probegroup(sliced_probegroup)
```
samuelgarcia, line 69: perfect

src/spikeinterface/core/channelsaggregationrecording.py
```diff
@@ -90,31 +91,27 @@ def __init__(self, recording_list_or_dict=None, renamed_channel_ids=None, record
...
-            ):  # we add the 'planar_contour' annotations only if there is a unique one in the recording_list
-                planar_contour_key = planar_contour_keys[0]
-                collect_planar_contours = []
-                for rec in recording_list:
-                    collect_planar_contours.append(rec.get_annotation(planar_contour_key))
-                if all(np.array_equal(arr, collect_planar_contours[0]) for arr in collect_planar_contours):
-                    self.set_annotation(planar_contour_key, collect_planar_contours[0])
+        # Aggregate probe information
+        all_probegroups = [rec.get_probegroup() for rec in recording_list if rec.has_probe()]
+        if len(all_probegroups) == len(recording_list):
```
samuelgarcia, line 98: we should handle the case when it is not the case.
When only some rec has probe and other not.
I think we should raise.
unless no rec has probe

src/spikeinterface/core/channelsaggregationrecording.py
```diff
@@ -90,31 +91,27 @@ def __init__(self, recording_list_or_dict=None, renamed_channel_ids=None, record
...
+            # check that contact positions are unique across all recordings
+            all_positions = []
+            for probegroup in all_probegroups:
+                for probe in probegroup.probes:
+                    all_positions.extend(probe.contact_positions)
+            assert len(np.unique(np.array(all_positions), axis=0)) == len(
+                all_positions
+            ), "Contact positions are not unique! Cannot aggregate recordings."
+
+            # Now make a new probegroup with all probes and set global device channel indices
```
samuelgarcia, line 108: We should check overlapping instead of contact uniqueness no ?

src/spikeinterface/core/binaryfolder.py
samuelgarcia, line 12: We clearly need a backward compatibility code here to read the probe json in the folder. And chek the version re order or not and set the probe.
This should be done here explicitly.
Here it was loading the contact_vector silently!!

src/spikeinterface/core/zarrextractors.py
alejoe91: no, we always do set_probegroup which wil reorder correctly

src/spikeinterface/core/baserecording.py
```diff
@@ -399,18 +394,11 @@ def _save(self, format="binary", verbose: bool = False, **save_kwargs):
...
-            cached.set_probegroup(probegroup)
-
         return cached
 
     def _extra_metadata_from_folder(self, folder):
         # load probe
-        folder = Path(folder)
-        if (folder / "probe.json").is_file():
-            probegroup = read_probeinterface(folder / "probe.json")
-            self.set_probegroup(probegroup, in_place=True)
```
samuelgarcia, line 413: where is it now ?

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -51,242 +53,312 @@ def has_scaleable_traces(self) -> bool:
...
+                f"connected channels and the recording has {self.get_num_channels()} channels. "
+                "Use the `recording.select_channels_with_probegroup()` method instead to return a new recording with "
+                "a channel selection to match the probe/probegroup."
+            )
+        probegroup_sorted.set_global_device_channel_indices(np.arange(probegroup_sorted.get_contact_count()))
+        self._probegroup = probegroup_sorted
+
+        # Handle and set channel groups
+        _set_group_property_based_on_probegroup(self, probegroup_sorted, group_mode=group_mode)
+
```
samuelgarcia, line 183: Not sure I like this.
I would prefer something in the get_property(, allow_probe_property=True/False)
This duplicate data.

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -51,242 +53,312 @@ def has_scaleable_traces(self) -> bool:
...
-                f"If using all channels remember that python is 0-indexed so max device_channel_index should be {self.get_num_channels() - 1} \n"
-                f"device_channel_indices are the following: {device_channel_indices} \n"
-                f"recording channels are the following: {self.get_channel_ids()} \n"
+        # Remove unconnected contacts and slice the probe group accordingly
+        device_channel_indices = probegroup.get_global_device_channel_indices()["device_channel_indices"]
+        keep_indices = np.flatnonzero(device_channel_indices >= 0)
+        if len(keep_indices) < len(device_channel_indices):
+            warn(
+                f"The given probes have {len(device_channel_indices) - len(keep_indices)} unconnected contacts: "
+                "they will be removed"
```
samuelgarcia: No this is too annoying.
We are selecting channels so this is normal to get less channels.

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -51,242 +53,312 @@ def has_scaleable_traces(self) -> bool:
...
-
-        # create recording : channel slice or clone or self
-        if in_place:
-            if not np.array_equal(new_channel_ids, self.get_channel_ids()):
-                raise Exception("set_probe(inplace=True) must have all channel indices")
-            sub_recording = self
-        else:
-            if np.array_equal(new_channel_ids, self.get_channel_ids()):
-                sub_recording = self.clone()
+            if len(keep_indices) == 0:
```
samuelgarcia: I think we can help the user more with raiseing here instead of havin an empty rec/probe

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -51,242 +53,312 @@ def has_scaleable_traces(self) -> bool:
...
-            probes_info.append(probe.annotations)
-        sub_recording.annotate(probes_info=probes_info)
+        # In some older SI versions, before #4300, the probe annotations were
+        # saved to the recording annotations as `probes_info`. If this is the
+        # case, we can copy the annotations to the probegroup and delete the
+        # `probes_info` from the recording annotations.
+        if "probes_info" in self._annotations:
+            probes_info = self._annotations.pop("probes_info")
+            for probe, probe_info in zip(probegroup.probes, probes_info):
+                probe.annotations.update(probe_info)
```
samuelgarcia: This is strange to have this here.
Normally this will be run in new si/probeinterface.
Here we are setting the probegroup in memory no ?
not handling a old format.
I would move this at the good palce.

src/spikeinterface/core/baserecording.py
alejoe91: in baserecordingsnippet

src/spikeinterface/core/tests/test_baserecording.py
```diff
@@ -413,40 +425,54 @@ def test_json_pickle_equivalence(create_cache_folder):
...
 
-def test_interleaved_probegroups():
-    recording = generate_recording(durations=[1.0], num_channels=16)
+def test_probes_info_annotation_backward_compat():
+    """
+    Regression test: SI versions before #4300 stored per-probe metadata in a
+    'probes_info' annotation on the recording rather than inside the ProbeGroup.
+    set_probegroup() must migrate those annotations onto the probe objects and
+    remove the stale 'probes_info' entry from the recording annotations.
+    """
```
samuelgarcia: Not sure to like this.
We should have in-memory object like this.
We should handle the backward compatibility on load.

src/spikeinterface/core/baserecordingsnippets.py
alejoe91: this will be relevant with #4465 , since they both use the same machinery

src/spikeinterface/core/baserecordingsnippets.py
alejoe91: it used to be there, we can remove

src/spikeinterface/core/baserecordingsnippets.py
alejoe91: this is for backward compatibility...we used to have an annotation-free probegroup and store the annotations in the `probes_info`.

src/spikeinterface/core/sortinganalyzer.py
alejoe91: added it back

src/spikeinterface/core/baserecordingsnippets.py
alejoe91: removed, not the extra_to/from dict take care ofr serializing/deserializing

src/spikeinterface/core/recording_tools.py
alejoe91: added back

src/spikeinterface/core/channelsaggregationrecording.py
alejoe91: done

src/spikeinterface/core/binaryfolder.py
alejoe91: done

src/spikeinterface/core/sortinganalyzer.py
alejoe91: this needs to be removed. The issue is that before the aggregation was creating a single new probe, now it keeps a probegroup

alejoe91:
@samuelgarcia @h-mayorquin now this is ready and final

h-mayorquin (approved):
LGTM

The backward-compat fixtures cover single-probe (sequential and shuffled) and a contiguous two-probe recording, but not an interleaved multi-probe one (two probes whose `device_channel_indices` alternate, e.g. `[0, 2, 4, 6, 1, 3, 5, 7]`). That case does work on this branch, I checked it reloads with the correct per-channel locations thanks to `_global_contact_order`, but nothing tests it, so a future change to the reconstruction / `get_slice` / `_global_contact_order` could silently scramble channel-to-location alignment on load without CI noticing. 

I am covering that in #4636 though.

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -51,242 +53,350 @@ def has_scaleable_traces(self) -> bool:
...
+            the same physical space (e.g. channels split by group from a single probe),
+            where contact positions are unique but probe bounding boxes may overlap.
+        """
+        if in_place is not None:
+            warnings.warn(
+                "The 'in_place' argument is deprecated and will be removed in version 0.106.0. "
+                "The `set_probe/probegroup()` are always in place and assume that the probe/probegroup has the "
+                "same number of connected contacts as the number of channels in the recording. "
+                "Use the `recording.select_channels_with_probegroup()` method instead to return a new recording with "
+                "a channel selection to match the probe/probegroup.",
```
chrishalcrow: Hello. I think we should simplify this. Most users who see this have just run `set_probe` on their recordng and this is a bit overwhelming for them. How about:

"The 'in_place' argument is deprecated and will be removed in version 0.106.0. "
"The `set_probe/probegroup` are now always in place; please remove the `in_place` argument."

I'm also confused by the second half. Was that really the use case of `in_place = False`??

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -51,242 +53,350 @@ def has_scaleable_traces(self) -> bool:
...
+                DeprecationWarning,
```
chrishalcrow:
```suggestion
                FutureWarning,
```

(so that it appears in Jupyter notebooks / the REPL)

src/spikeinterface/core/baserecordingsnippets.py
alejoe91: yes. If you have a different number of contacts and you want to slice the recording, you cannot do in place. I'll try to simplify!

alejoe91: done! I removed the second part since it would trigger an error instead (after user get rid of the `in_place` arg)

src/spikeinterface/core/baserecordingsnippets.py
alejoe91: done

alejoe91:
> LGTM
> 
> The backward-compat fixtures cover single-probe (sequential and shuffled) and a contiguous two-probe recording, but not an interleaved multi-probe one (two probes whose `device_channel_indices` alternate, e.g. `[0, 2, 4, 6, 1, 3, 5, 7]`). That case does work on this branch, I checked it reloads with the correct per-channel locations thanks to `_global_contact_order`, but nothing tests it, so a future change to the reconstruction / `get_slice` / `_global_contact_order` could silently scramble channel-to-location alignment on load without CI noticing.
> 
> I am covering that in #4636 though.

Added the interleaved probegroup case. We cna unify backward compatibility tests later following #4636

src/spikeinterface/core/baserecordingsnippets.py
samuelgarcia: you are linking this PR no ?
Really I think that copy/paste contact property to channel property is not a good idea.

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -51,242 +53,347 @@ def has_scaleable_traces(self) -> bool:
...
+                stacklevel=2,
+            )
+            if not in_place:
+                return self.select_channels_with_probegroup(probegroup, group_mode=group_mode)
+
+        if check_overlap and len(probegroup.probes) > 0:
+            check_probe_do_not_overlap(probegroup.probes)
+
+        probegroup_sorted = self._get_probegroup_based_on_device_channel_indices(probegroup)
+
```
samuelgarcia, line 160: we need a checker that the deevice_channel_indices is equal to np.arange() here.

Because nothing prevent to have somethign like this [0, 2, 5, 9]

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -51,242 +53,347 @@ def has_scaleable_traces(self) -> bool:
...
+            if np.array_equal(new_channel_ids, self.channel_ids):
+                sub_recording = self.clone()
+            else:
+                sub_recording = self.select_channels(new_channel_ids)
+            sub_recording._probegroup = probegroup_sorted
+            _set_group_property_based_on_probegroup(sub_recording, probegroup_sorted, group_mode=group_mode)
         else:
-            raise ValueError("must give Probe or ProbeGroup or list of Probe")
+            sub_recording = self.select_channels([])  # empty recording
+            sub_recording._probegroup = ProbeGroup()  # empty probegroup
```
samuelgarcia, line 243: Do we really want this or do we prefer to help the user and raise an error ?

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -51,242 +53,347 @@ def has_scaleable_traces(self) -> bool:
...
-        elif isinstance(probe_or_probegroup, list):
-            assert all([isinstance(e, Probe) for e in probe_or_probegroup])
-            probegroup = ProbeGroup()
-            for probe in probe_or_probegroup:
-                probegroup.add_probe(probe)
+        probegroup_sorted = self._get_probegroup_based_on_device_channel_indices(probegroup)
+        if probegroup_sorted.get_contact_count() > 0:
+            sorted_dci = probegroup_sorted.get_global_device_channel_indices()["device_channel_indices"]
+            new_channel_ids = self.channel_ids[sorted_dci]
+            probegroup_sorted.set_global_device_channel_indices(np.arange(len(new_channel_ids)))
```
samuelgarcia, line 234: This should be always true.
So it is better to check rather than forcing np.araneg() and hidding futur errors here.

src/spikeinterface/core/baserecordingsnippets.py
samuelgarcia: oups sorry

src/spikeinterface/core/channelsaggregationrecording.py
samuelgarcia: lets do it in the run_sorter()

samuelgarcia (approved):

src/spikeinterface/core/baserecordingsnippets.py
alejoe91: this could be the case for removing bad channels etc? Let's leave for now

src/spikeinterface/core/channelsaggregationrecording.py
alejoe91: done

PR #4467, opened 2026-03-25
Add `zarr_class_info` attribute to zarr files for unambiguous and extendible zarr read

alejoe91:
Instead of relying on channel_ids/unit_ids, this PR adds a full class provenance to the zarr attributes.

This change has two major pros:
- makes reader unambiguous, also adding version/module etc
- makes reader extendible and usable by other packages (e.g., [`photon-mosaic`](https://github.com/photon-mosaic/photon-mosaic/blob/dev/src/photon_mosaic/core/zarrimaging.py#L206))

### DIFF
--- a/src/spikeinterface/core/zarrextractors.py
+++ b/src/spikeinterface/core/zarrextractors.py
@@ -5,7 +5,7 @@ import zarr
 
 from probeinterface import ProbeGroup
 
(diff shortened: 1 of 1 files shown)


samuelgarcia:
Cool. merci

PR #4468, opened 2026-03-25
Make `get_data` for valid unit periods use unit ids, not indices

chrishalcrow:
Fixes a bug. When using `get_data(outputs="by_unit")` with the valid unit periods extension, the method was using unit indices, not ids.

Also added a test.


alejoe91:
Good catch!

PR #4473, opened 2026-03-26
Add empty metadata to `template_metrics` on load if doesn't exist

### LINKED ISSUE #4471: Can't add `tmp_data_to_save` to old merged analyzer
As of v0.104.0 template_metrics saves some `tmp_data_to_save`: the peak_data and main_channel_templates alongside the metrics as `data` in the extension.

This creates a backward compatibility problem. When you load a v0.103.2 analyzer using v0.104.0, this has no peak_data. So when you try to e.g. merge two units, the analyzer cannot update the peak_data (since it has no peak data). Actual bug is here

https://github.com/SpikeInterface/spikeinterface/blob/962087b95485e72c5f1149a45e5b8f45a443f64d/src/spikeinterface/core/analyzer_extension_core.py#L1416

`self.data` has no key "peak_data".

Wanted to raise an issue cause I think this is a general problem and there are a few solutions:
1. Use `get` instead of indexing, and don't update data `self.data` didn't contain the `tmp_data` to being with.
2. Use `_handle_backward_compatibility_on_load` to add a suitable "empty" peak_data.

Number 1. feels easier but I actually quite like 2, because it ensures that no matter what you load with v0.104.0, you get the same data structure. Thoughts?

chrishalcrow:
Fixes #4471, using option 1.

In this PR, if you load an analyzer from version <0.104, the expected metadata for `template_metrics` will be added. It's empty but has the right structure, so won't error when you do merges/splits etc.

I like this solution because it ensures that all analyzers loaded with 0.104 will have the same structure.

Only issue is when you have an analyzer from <0.104 with template metrics and no templates. I think this is an edge case.

We really need to add a backwards compatibility focused test suite sometime soon...

### DIFF
--- a/src/spikeinterface/metrics/template/template_metrics.py
+++ b/src/spikeinterface/metrics/template/template_metrics.py
@@ -169,6 +169,24 @@ class ComputeTemplateMetrics(BaseMetricExtension):
             if "waveform_ratios" not in self.params["metric_names"]:
                 self.params["metric_names"].append("waveform_ratios")
 
+        if self.data.get("peaks_data") is None:
+            import pandas as pd
+
+            self.data["peaks_data"] = pd.DataFrame(
+                columns=get_peaks_data_columns(), index=self.sorting_analyzer.unit_ids
+            )
+
+        if self.data.get("main_channel_templates") is None:
+            num_units = self.sorting_analyzer.get_num_units()
+            if self.sorting_analyzer.has_extension("templates"):
+                templates_ext = self.sorting_analyzer.get_extension("templates")
+                template_samples = templates_ext.nbefore + templates_ext.nafter
+                upsampling_factor = self.params["upsampling_factor"]
+                upsampled_template_samples = template_samples * upsampling_factor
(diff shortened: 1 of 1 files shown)


alejoe91:
Nice!!!

src/spikeinterface/metrics/template/template_metrics.py
```diff
@@ -169,6 +169,24 @@ def _handle_backward_compatibility_on_load(self):
...
+            )
+
+        if self.data.get("main_channel_templates") is None:
+            num_units = self.sorting_analyzer.get_num_units()
+            if self.sorting_analyzer.has_extension("templates"):
+                templates_ext = self.sorting_analyzer.get_extension("templates")
+                template_samples = templates_ext.nbefore + templates_ext.nafter
+                upsampling_factor = self.params["upsampling_factor"]
+                upsampled_template_samples = template_samples * upsampling_factor
+                self.data["main_channel_templates"] = np.zeros((num_units, upsampled_template_samples))
```
alejoe91, line 186: Should we actually do the upsampling here?

chrishalcrow:
Replaced by #4480

PR #4481, opened 2026-03-30
Handle `peak_sign` for quality metrics in `_handle_backward_compatibility_on_load`

chrishalcrow:
In #4403, we passed `peak_sign` to `get_template_extremum_channel` in `_prepare_data` for quality metrics.

In the past, the default `peak_sign` here was `None` but this is not allowed: we assert that it must be in `["pos", "neg", "both"]`. Hence if you load an old analyzer (with `peak_sign = None`) and recompute (e.g. when merging) this assertion fails! This PR fixes this issue by trying to set `peak_sign` in `_handle_backward_compatibility_on_load`. It looks through all the other metric's params to see if the user set peak_sign somewhere. If not, it uses the old default, "neg".

At the top of my dev to-do list is: make a gh-actions test suite for backwards compatibility. Will do it soon!!

### DIFF
--- a/src/spikeinterface/metrics/quality/quality_metrics.py
+++ b/src/spikeinterface/metrics/quality/quality_metrics.py
@@ -77,6 +77,20 @@ class ComputeQualityMetrics(BaseMetricExtension):
             if "peak_sign" in self.params["metric_params"]["amplitude_median"]:
                 del self.params["metric_params"]["amplitude_median"]["peak_sign"]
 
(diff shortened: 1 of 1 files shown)


alejoe91:
> At the top of my dev to-do list is: make a gh-actions test suite for backwards compatibility. Will do it soon!!

For inspiration form #4260: https://github.com/alejoe91/spikeinterface/blob/zarrv3/.github/workflows/test_zarr_compat.yml

PR #4483, opened 2026-03-31
Make sure default automerge params reflect their type

chrishalcrow:
In spikeinterface-gui, the parameter type of each autmerge algorithm is computed by computing the type of the default. 

Before this PR there were some mismatches between the type (e.g. `min_snr` is a float param) and the type of the default value (`min_snr` has default value 2). This means the user cannot change `min_snr` to float values in the GUI. This PR ensures all default param values have the type intended by their param.

### DIFF
--- a/src/spikeinterface/curation/auto_merge.py
+++ b/src/spikeinterface/curation/auto_merge.py
@@ -69,9 +69,9 @@ _required_extensions = {
 
 _default_step_params = {
     "num_spikes": {"min_spikes": 100},
-    "snr": {"min_snr": 2},
+    "snr": {"min_snr": 2.0},
     "remove_contaminated": {"contamination_thresh": 0.2, "refractory_period_ms": 1.0, "censored_period_ms": 0.3},
-    "unit_locations": {"max_distance_um": 150},
+    "unit_locations": {"max_distance_um": 150.0},
     "correlogram": {
         "corr_diff_thresh": 0.16,
         "censor_correlograms_ms": 0.15,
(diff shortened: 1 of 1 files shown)


alejoe91 (approved):

zm711:
Would it be worth adding type hints somewhere around here. Otherwise I think it would be too easy for these things to get out of sync in the future, no?

chrishalcrow:
> Would it be worth adding type hints somewhere around here. Otherwise I think it would be too easy for these things to get out of sync in the future, no?

Yeah it's a good point. It's a bit awkward: _some of these_ are type-hinted in their function signatures but the values are given here. This is fine internally: it's only a problem when used externally by e.g. si-gui. It's not easy to test because the function signatures are scattered all over the place, and only covers some parameters e.g. `min_spikes` is just used here, and it passed through some generic params dict:

https://github.com/SpikeInterface/spikeinterface/blob/6b476031da1edbfc94d7adcb075f87347313ffc7/src/spikeinterface/curation/auto_merge.py#L249

so isn't typed anywhere...

EDIT: one option is to wait until 3.11 (~ a year's time) and then implement TypedDicts everywhere we use parameters...

alejoe91:
Or Pydantic? :)

zm711:
I think Pydantic might be the longer term solution, especially with you all making such a big push on integrations between here and the gui. Of note Cyrille is also getting ready to do a completely updated Phy release.

alejoe91 (approved):

Issue #4474, opened 2026-03-26
ComputeWaveforms and ComputeTemplates: int() truncation causes inconsistent sample counts across slightly different sampling rates


Antorhythms:
In analyzer_extension_core.py, the `nbefore` and `nafter` properties in ComputeWaveforms and ComputeTemplates use `int()` (truncation) to convert millisecond parameters to sample counts:

```
@property
def nbefore(self):    
	return int(self.params["ms_before"] * self.sorting_analyzer.sampling_frequency / 1000.0)
```

This causes a ±1 sample inconsistency when the sampling rate deviates slightly from a round number. For example, with ms_before=2.0:
fs = 29999 Hz → int(59.998) = 59
fs = 30000 Hz → int(60.0) = 60

This means waveform arrays have different shapes and different center indices across datasets with nearly identical sampling rates.

Proposed fix:
Replace `int()` with `round()` in the nbefore/nafter properties of both `ComputeWaveforms` (lines 163, 167) and `ComputeTemplates` (lines 526, 532).

alejoe91:
Hi @Antorhythms 

Would you like to make a PR for this issue? :)

Antorhythms:
Sure! Will get to it asap

Antorhythms:
Hey Ale,

I’m working on the PR and noticed that the conversion from milliseconds to samples (using int() instead of round()) is present in several places beyond ComputeWaveforms and Templates.

For example, I found it here:

- `core/node_pipeline.py` (Lines 317, 318): `WaveformsNode.nbefore/nafter`.
- `core/sparsity.py` (Lines 787, 788): `compute_sparsity()`.

To ensure consistency one option could be to build a shared utility that could be located in `core/core_tools.py` and be reused at different sites.
`def ms_to_samples(ms: float, sampling_frequency: float) -> int:
    	return round(ms * sampling_frequency / 1000.0)`

What do you think?

alejoe91:
Yep that sounds good!

zm711:
We've had these types of debates before on the Neo code base about `round` vs `int`. For example saying `ms_before` could mean that you want at least 2 ms before or could be mean at most you want 2ms before or it could be give me exactly 2ms before. Have we as an org defined this type of semantic? Even `round` behaves funny sometimes 
```python
>>> round(0.5)
0
```
[...]

PR #4484, opened 2026-03-31
Replace int() truncation with shared ms_to_samples() 

Antorhythms:
Utility for waveform/template sample count conversions. Adds ms_to_samples() to core_tools.py using round() instead of int() to prevent ±1 sampleinconsistencies across datasets with nearly identical sampling rates (e.g. 29999 vs 30000 Hz). Tried to spot int truncation behavior in the codebase with AI and adopted new util across 21 production files and 5 test files for all waveform/template window geometry; non-waveform conversions (margins, exclusion zones, refractory periods) left unchanged.

Fixes issue https://github.com/SpikeInterface/spikeinterface/issues/4474


samuelgarcia:
merci

PR #4487, opened 2026-04-01
generate: Add temporal correlations to noise simulator

### LINKED ISSUE #4482: Temporally correlated noise in simulations
Dear authors of core/generate.py's `NoiseGeneratorRecording` (seems to be @samuelgarcia @JoeZiminski @h-mayorquin),

  I'm interested in extending this noise simulation to include temporal correlations, and I wanted to ask some questions about how best to do this without excessive rocking of boats. I'll open a PR to close this issue shortly with an implementation when the path is clear, assuming this feature is acceptable?

I already have an implementation for introducing temporal correlations (https://github.com/cwindolf/dartsort/blob/main/src/dartsort/evaluate/noise_recording_tools.py#L162, which basically just calls https://github.com/cwindolf/dartsort/blob/main/src/dartsort/util/noise_util.py#L213). This works by multiplying the data spatially by the square root of the covariance (inverse of whitener) and then doing some stuff in frequency domain to cause the data's power spectral density to match one that I computed from real data. Currently this is implemented as a `BasePreprocessor` subclass called `UnwhitenPreprocessor` which is applied over a `WhiteNoiseRecording`, so that correlations are introduced kind of the opposite of how the whitening step works in SpikeInterface (but also doing temporal stuff).

My questions:
 - The current `NoiseGeneratorRecording` handles everything internally, rather than using a preprocessing chain like I am using. It seems to make sense to match this style, so that I would add more `if`/`else` cases to the `NoiseGeneratorRecordingSegement.get_traces()`. This will include adding some temporal margin if needed for the temporal part.
   - ❓ Does that sound like the right plan, or should I stick to the Preprocessor-based method I currently use?
 - As part of my implementation, I would end up programming the logic for spatial correlation by multiplying by the covariance square root, rather than letting scipy's `multivariate_normal()` do its Cholesky stuff. It would be possible to unify the logic, so that the current spatial correlation would be reimplemented as white_noise @ spatial_cov_sqrt in all cases. But, this could change the output numerically and that may bother people.
   - ❓ Should I unify the spatial correlation into one logic which is the same whether or not there is also temporal correlation happening, or should I keep the current logic intact if there is only spatial covariance and handle the spatiotemporal stuff completely separately?

Those are my 2 questions. Let me know if you have any other concerns and thanks!

cwindolf:
Closes #4482 

This adds temporal noise to the simulations. It is parametrized by a spectral density which the user would supply as a numpy array.

I added logic for circular padding to the generator class so that everything is still totally circular like in tile_pregenerated. This required moving the old get_traces() into a helper.

There's also a flag I ended up putting in to allow the user to turn on overlap-add convolutions. It's off by default, because it can introduce inconsistencies in the output if the user is not careful (in particular, the unit test I added will fail). But I think it's worth keeping this since it will be substantially faster when generating large noise simulations, and it's totally safe if one plans to just call save_to_folder() or something.

### DIFF
--- a/src/spikeinterface/generation/tests/test_noise_tools.py
+++ b/src/spikeinterface/generation/tests/test_noise_tools.py
@@ -44,7 +44,7 @@ def test_generate_noise():
 
 
 @pytest.mark.parametrize("duration", [1.0, 2.0, 2.2])
-@pytest.mark.parametrize("strategy", ["tile_precomputed", "on_the_fly"])
+@pytest.mark.parametrize("strategy", ["tile_pregenerated", "on_the_fly"])
 def test_noise_generator_temporal(strategy, duration):
     psdlen = 25
     kdomain = np.linspace(0.0, 10.0, psdlen)


cwindolf:
Sorry, I think my formatter has the wrong presets! Hopefully diff shrinks after pre commit bot.

cwindolf:
Hmm, so I don't think there's a good way to do this without scipy, but core tests don't include scipy dependency. I'm wondering, is there another good place for me to put my unit test so that it will run with scipy?

alejoe91:
@cwindolf you can keep the implementation in core, but move the tests in the `generation/tests` module ;)

samuelgarcia:
Hi Charlie,
thanks for this.
Lets have a call to discuss the implementation of the margin.

cwindolf:
To do before merging: 
 - [ ] noticed that *=noise_levels doesn't happen if cov_matrix is supplied and noise method is tile_precomputed. But it does happen in all cases under on_the_fly.
 - [ ] Add spectral_density kw to generate_noise() in noise_tools.py
 - [ ] Choose whether to refactor as preprocessor and do that if needed (issue discussion)

cwindolf:
Closing per discussion in https://github.com/SpikeInterface/spikeinterface/issues/4522, will follow up with a new one soon!

Issue #979, opened 2022-09-30
export_to_phy runs out of memory and/or is slow


hpay:
My test recording is 32 channels, 1 hr long. Kilosort2 finds 29 units with normal firing rates (all but one less than 20 Hz, one 100 Hz). Running Win10 with 24 GB RAM. 

When I run:

```
# Get waveforms
waveform_folder = results_path/(sorter_name+'_waveforms')
ms_before = 1.5 # default 3
ms_after = 2.5  # default 4
we = si.extract_waveforms(recording_cmr, sorting, waveform_folder,ms_before=ms_before, ms_after=ms_after,
    n_jobs=1, chunk_duration="10s", max_spikes_per_unit=500, return_scaled=True, overwrite=True)

# Write to Phy
phy_folder =  results_path/(sorter_name+'_phy')
si.export_to_phy(we,output_folder=phy_folder,copy_binary=False, max_channels_per_template=8)

```

I get the error:

```
Output exceeds the size limit. Open the full output data in a text editor
---------------------------------------------------------------------------
MemoryError                               Traceback (most recent call last)
Cell In [23], line 18
     16 phy_folder =  results_path/(sorter_name+'_phy')
     17 print(phy_folder)
---> 18 si.export_to_phy(we, output_folder=phy_folder, verbose=False, remove_if_exists=True)

File d:\hannah\dropbox\code\spikesort\spikeinterface\spikeinterface\exporters\to_phy.py:113, in export_to_phy(waveform_extractor, output_folder, compute_pc_features, compute_amplitudes, sparsity_dict, copy_binary, max_channels_per_template, remove_if_exists, peak_sign, template_mode, dtype, verbose, **job_kwargs)
    111 if copy_binary:
    112     rec_path = output_folder / 'recording.dat'
--> 113     write_binary_recording(recording, file_paths=rec_path, verbose=verbose, dtype=dtype, **job_kwargs)
    114 elif isinstance(recording, BinaryRecordingExtractor):
    115     rec_path = recording._kwargs['file_paths'][0]

File d:\hannah\dropbox\code\spikesort\spikeinterface\spikeinterface\core\core_tools.py:285, in write_binary_recording(recording, file_paths, dtype, add_file_extension, verbose, byte_offset, auto_cast_uint, **job_kwargs)
    282     init_args = (recording.to_dict(), rec_memmaps_dict, dtype, cast_unsigned)
[...]

PR #4488, opened 2026-04-01
Vectorize per-channel PCA transform in `run_for_all_spikes`

### LINKED ISSUE #4485: `run_for_all_spikes`: vectorize per-channel PCA transform
### Problem

The inner loop in `_all_pc_extractor_chunk` ([in `principal_component.py`](https://github.com/SpikeInterface/spikeinterface/blob/6b476031da1edbfc94d7adcb075f87347313ffc7/src/spikeinterface/postprocessing/principal_component.py#L606)) calls `pca_model[chan_ind].transform()` once per spike per channel:

```python
for i in range(i0, i1):
    wf = traces[st - start - nbefore : st - start + nafter, :]
    ...
    for c, chan_ind in enumerate(chan_inds):
        w = wf[:, chan_ind]
        ...
        all_pcs[i, :, c] = pca_model[chan_ind].transform(w[None, :])
```

For a ~70-minute Neuropixels recording with ~15M spikes and ~20 sparse channels per unit, this results in ~300M individual sklearn `transform` calls, each on a 1×210 matrix. The Python/sklearn per-call overhead dominates — the actual linear algebra (a 1×210 by 210×5 matrix multiply) is trivial.

In practice, `run_for_all_spikes` takes longer than streaming the entire 88 GB recording from S3 and writing it to disk.

### Suggested optimization

In `by_channel_local` mode, the PCA model is per-channel, not per-unit — `pca_model[chan_ind]` applies to all spikes on that channel regardless of unit. This means all spikes within a chunk that use the same channel can be batched into a single `transform` call:

```python
# Batch all spikes in the chunk for a given channel
for chan_ind in unique_channels_in_chunk:
    # Gather waveform snippets for all spikes using this channel
    wfs = np.stack([
        traces[st - start - nbefore : st - start + nafter, chan_ind]
        for st in spike_times_using_this_channel
    ])  # shape: (n_spikes_on_channel, 210)

    # One transform call instead of n_spikes_on_channel calls
    all_pcs[spike_indices, :, channel_position] = pca_model[chan_ind].transform(wfs)
```

This reduces ~300M Python calls to ~384 per chunk (one per channel), with sklearn operating on batch matrices instead of individual rows.

### RAM impact

Minimal. For a 1-second chunk with ~3,600 spikes, batching all spikes for one channel is 3,600 × 210 × 4 bytes ≈ 3 MB. Even batching all channels simultaneously would be ~60 MB — well within reason.

galenlynch:
### Problem

`_all_pc_extractor_chunk` calls `pca_model[chan_ind].transform()` once per spike per channel in a Python loop:

```python
for i in range(i0, i1):
    wf = traces[st - start - nbefore : st - start + nafter, :]
    for c, chan_ind in enumerate(chan_inds):
        w = wf[:, chan_ind]
        all_pcs[i, :, c] = pca_model[chan_ind].transform(w[None, :])
```

For a ~70-minute Neuropixels recording with ~10M spikes and ~26 sparse channels, this is ~260M individual sklearn `transform` calls, each on a 1×210 matrix. The Python/sklearn per-call overhead dominates.

### Solution

Batch all spikes within a chunk by channel and call `transform` once per channel:

1. Extract all valid waveform snippets in the chunk at once using vectorized fancy indexing
2. Group spikes by channel index across all units
3. Call `pca_model[chan_ind].transform(wfs_batch)` once per channel with the full batch

In `by_channel_local` mode the PCA model is per-channel (not per-unit), so all spikes on a given channel share the same model regardless of unit identity.

#### Synthetic (500 spikes, 10 channels, 50-sample waveforms)

| | Time | Speedup |
|---|---|---|
| Original | 0.126s | — |
| Vectorized | 0.002s | **53x** |

Results match: max absolute difference 9.5e-7 (float rounding).

#### Real data (Neuropixels probe, 5 minutes of recording from local .dat)

- 706K total spikes
- 379 channels, 26 sparse channels per unit, 210-sample waveforms
- `n_jobs=1`, `chunk_duration=10s`

| | Time | Per chunk | Speedup |
|---|---|---|---|
| Original | 548.7s (9.1 min) | 18.3s/chunk | — |
| Vectorized | 110.3s (1.8 min) | 2.1s/chunk | **5.0x** |
| Vectorized + numpy grouping | 108.2s | 2.0s/chunk | **5.6x** |

Results match: max absolute difference 1.49e-08. `np.allclose` confirms identical output.

The real-data speedup is lower than synthetic because disk I/O, waveform extraction, and memory allocation are shared costs. The optimization only affects the `transform` call overhead, which is ~80% of chunk time in the original code.

#### Projected impact

For a full 69-minute recording, PC extraction drops from ~60 min to ~12 min.

### RAM impact

Minimal. For a 10s chunk with ~25K spikes, the batch waveform array is ~25K × 210 × 4 bytes ≈ 20 MB per channel call, reused across channels. Peak additional memory vs. original: ~50 MB.

### Changes

- `_all_pc_extractor_chunk` in `principal_component.py`: replaced per-spike per-channel loop with vectorized batch-by-channel approach
- No API changes, no new dependencies
- All existing `test_principal_component.py` tests pass
- A follow-up optimization replaces the per-spike Python dict loop (building the channel→spike mapping) with numpy unit grouping: loop over unique unit indices and use vectorized boolean masks. This reduces Python iterations from n_spikes × n_sparse_channels (~600K) to n_units × n_sparse_channels (~10K) per chunk.

Fixes #4485
Related: #979

### DIFF
--- a/src/spikeinterface/postprocessing/principal_component.py
+++ b/src/spikeinterface/postprocessing/principal_component.py
@@ -629,7 +629,7 @@ def _all_pc_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx):
     if i0 == i1:
         return
 
-    # Since we get_traces accounting for nbefore and nafter, all spikes in the chunk are valid and we can extract
+    # Since `get_traces` accounts for nbefore and nafter, all spikes in the chunk are valid and we can extract
     # all waveforms in one go without worrying about borders.
     start = int(spike_times[i0] - nbefore)
     end = int(spike_times[i1 - 1] + nafter)


galenlynch:
The test failure is unrelated. Seems like a problem with cross-correlagram tests and `fast_mode`?

src/spikeinterface/postprocessing/principal_component.py
```diff
@@ -633,27 +633,65 @@ def _all_pc_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx):
...
-        if st - start + nafter > traces.shape[0]:
-            continue
+    nsamples = nbefore + nafter
 
-        wf = traces[st - start - nbefore : st - start + nafter, :]
+    # Extract all waveforms in the chunk at once
+    # valid_mask tracks which spikes have valid (in-bounds) waveforms
+    chunk_spike_times = spike_times[i0:i1]
+    offsets = chunk_spike_times - start - nbefore
+    valid_mask = (offsets >= 0) & (offsets + nsamples <= traces.shape[0])
```
zm711: This is me probably misreading but I'm super bad at parsing these type of > and < in general. If we have to be less than or = to the shape couldn't we run into an issue where we are = to the shape which is out of bounds? 

ie an array of (4,5) the shape[0] = 4, but if I try to index on 4 it will be an out of bounds error. Again I don't work on the PC code at all so maybe I'm completely wrong here.

alejoe91: You mean that the second <= should be just <?

alejoe91: In that case, I agree. It should be 
```suggestion
    valid_mask = (offsets >= 0) & (offsets + nsamples < traces.shape[0])
```
to avoid any indexing error.

E,g. if nsamples = 90, traces shape is 300, and you have a spike at 210, then it would be valid, but L654 will fail because it'll try to access traces[300]...

alejoe91: Actually, we could get rid of the `valid_mask` completely by extracting traces using a `margin` of `max(nbefore, nafter)`. @galenlynch do you mind if I give it a try in this PR?

galenlynch: I don't think there's a bug here. This matches the original semantics (`if ... > traces.shape[0]: continue`), and the original semantics were correct. 

The relevant code:
```python
  valid_mask = (offsets >= 0) & (offsets + nsamples <= traces.shape[0])   # L642
  ...
  sample_indices = valid_offsets[:, None] + np.arange(nsamples)[None, :]  # L653
  all_wfs = traces[sample_indices]                                        # L654
```

`np.arange(nsamples)` produces `[0, 1, ..., nsamples-1]`. So for a spike with offset `o`, the fancy indices accessed are `[o, o+1, ..., o + nsamples - 1]`. The maximum index accessed is `o + nsamples - 1`.

For that to be in bounds, we need:
`o + nsamples - 1 ≤ traces.shape[0] - 1` ⟺ `o + nsamples ≤ traces.shape[0]`

@alejoe91 your example is also not quite right:
With `offset = 210`, `nsamples = 90`:
- `offset + nsamples = 300 ≤ 300`  passes the mask
- `sample_indices = 210 + [0..89] = [210, 211, ..., 299]`
- `traces[[210, ..., 299]]`: max index is **299**, not 300.

Changing the bound to `<` would drop valid spikes at the end of the chunk.

Using `get_chunk_with_margin` is not the right fix because 1) I don't think there's a bug 2) would overfetch the data 3) would require other indexing logic changes.

---

You're right that `valid_mask` is leftover and unreachable logic replacing the original logic.

Look at how `start` and `end` are computed (lines 632–634):

```python
start = int(spike_times[i0] - nbefore)
end = int(spike_times[i1 - 1] + nafter)
traces = recording.get_traces(start_frame=start, end_frame=end, ...)
```

The traces array is tight-bracketed around exactly the spikes in `[i0, i1)`. Let's check the extremes:

- First spike: `offset = spike_times[i0] - start - nbefore = 0` → accesses `[0, nsamples-1]` (correct)
- Last spike: `offset = spike_times[i1-1] - spike_times[i0]`, so `offset + nsamples = spike_times[i1-1] - spike_times[i0] + nbefore + nafter = end - start = traces.shape[0]` → accesses up to `traces.shape[0]-1` (correct)

So `valid_mask` is always all-True on the happy path. The only time anything gets dropped is at absolute segment boundaries, and those are already handled by the `i0 += 1 / i1 -= 1` loop at lines 620–627 before `start/end` are even computed. The `valid_mask` is dead defensive code inherited from the original loop's continue checks.

We could delete it entirely:
```python
# Before
chunk_spike_times = spike_times[i0:i1]
offsets = chunk_spike_times - start - nbefore
valid_mask = (offsets >= 0) & (offsets + nsamples <= traces.shape[0])
if not np.any(valid_mask):
    return
valid_offsets = offsets[valid_mask]
valid_indices = np.arange(i0, i1)[valid_mask]
n_valid = len(valid_offsets)

# After
offsets = spike_times[i0:i1] - start - nbefore
valid_indices = np.arange(i0, i1)
n_valid = i1 - i0
sample_indices = offsets[:, None] + np.arange(nsamples)[None, :]
all_wfs = traces[sample_indices]
```

What do you think?

galenlynch:
@alejoe91 replied in context of the code. Thanks for the review!

galenlynch:
Also didn't mean to steamroll your offer while responding to whether ≤ or < was correct. Commits are welcome if you'd like to take a stab! Also happy to finish this off by dropping `valid_mask` entirely.

src/spikeinterface/postprocessing/principal_component.py
alejoe91: Hi @galenlynch 

You're absolutely right about indexing, my bad I wrote my comment to quickly!

Yes let's go ahead and remove the `valid_mask`. I would also rename `valid_indices` to `spike_indices_in_chunk`, since we know they're all valid!

src/spikeinterface/postprocessing/principal_component.py
```diff
@@ -629,31 +629,63 @@ def _all_pc_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx):
     if i0 == i1:
         return
 
+    # Since we get_traces accounting for nbefore and nafter, all spikes in the chunk are valid and we can extract
```
chrishalcrow:
```suggestion
    # Since `get_traces` accounts for nbefore and nafter, all spikes in the chunk are valid and we can extract
```

src/spikeinterface/postprocessing/principal_component.py
```diff
@@ -629,31 +629,63 @@ def _all_pc_extractor_chunk(segment_index, start_frame, end_frame, worker_ctx):
...
+    all_wfs = traces[sample_indices]  # (n_spikes, nsamples, n_channels)
 
+    # Vectorized PCA: batch by channel across all spikes in the chunk.
+    # For each unique channel, find all spikes that use it (via their unit's
+    # sparsity), extract waveforms, and call transform once.
+    labels_in_chunk = spike_labels[spike_indices]
+
+    # Build a set of all channels used by spikes in this chunk
+    unique_unit_indices = np.unique(labels_in_chunk)
+    chan_info: dict[int, list[tuple[np.ndarray, int]]] = {}
```
chrishalcrow, line 658:
```suggestion
    chan_info = {}
```
Just a style thing - we almost never do typing for simple(ish) variables mid-function.

alejoe91: True, but I'd leave it here since it makes it easier to read

galenlynch:
Thanks for taking this across the finish line!

PR #4489, opened 2026-04-01
generate: Add power decay spatial profile

cwindolf:
This PR adds another spatial decay family to the template simulator. Instead of just exponential, there's also a power law type decay.

I'd like to switch over totally to SpikeInterface for my simulation studies, and with this and the temporal correlations thing I think that should be enough for me.


cwindolf:
I think the test fails are unrelated?

alejoe91:
> I think the test fails are unrelated?

yes very unrelated. should be fixed by #4490

chrishalcrow:
Yeah, weird test failure. The complete tests should not be trying to include `.github/scripts/test_kilosort4_ci.py`. I'll look into it...

cwindolf:
The test fail seems to happen because in `"src/spikeinterface/preprocessing/tests/test_pipeline.py::test_loading_provenance"`, there is a place where bad channels are detected and discarded; this discards all the channels and there is some file-seek thing where `.seek(0 - 1)` is called because the data size is 0.

It passes locally for me, but it seems like my 2 PRs are the only ones where this test fails? I am a bit confused.

cwindolf:
Got it figured out. The failing test was flaky because `detect_bad_channels()` can be touchy. It strongly depended on the parameters of the 10 units generated in that test, and I had changed the order of random numbers being generated by putting my new unit param in the middle of the defaults dict which is traversed in order. I just moved it to the end to make the RNG path the same, but didn't fix the flaky test (which is fine, it's just there to test loading provenance stuff).

samuelgarcia (approved):

PR #4494, opened 2026-04-04
Bombcell extension

Julie-Fabre:
Changes to expose and enable all parameters available in native bombcell: 
- options to enable/disable drift and distance metrics, 
- choice of RPV calculation method
- time chunks


src/spikeinterface/curation/bombcell_curation.py
```diff
@@ -66,9 +81,11 @@ def bombcell_get_default_thresholds() -> dict:
             "snr": {"greater": 5, "less": None},
             "amplitude_cutoff": {"greater": None, "less": 0.2},
             "num_spikes": {"greater": 300, "less": None},
-            "rp_contamination": {"greater": None, "less": 0.1},
+            "rpv": {"greater": None, "less": 0.1},  # applies to rp_contamination or sliding_rp_violation
```
zm711: I wonder if it is better to keep this explicit for users? Since our qc possibilities are rp_contamination or sliding_rp_violation? Not sure...

zm711 (commented):
got interrupted I'll try to look over this more later.

examples/how_to/full_pipeline_with_bombcell.ipynb
alejoe91, line 1: Hi @Julie-Fabre 

We don't add notebooks directly to the how to, but we keep a python file synced to a local notebook that is then converted to the docs. See README here: https://github.com/SpikeInterface/spikeinterface/tree/main/examples/how_to

src/spikeinterface/curation/bombcell_curation.py
alejoe91: I agree with @zm711. In addition, thresholds could also be different. 

To be extra explicit, I recommend doing something like this (I think we can remove all the `None`s!):
```
    "sliding_rp_violation": {"less": 0.1, "fallback": "rp_contamination"},
    "rp_contamination": {"less": 0.1}
```

Internally, maybe directly in the threshold_metric function, we can have the logic that: 
- looks for metrics with a fallback
    - if present, use it and delete fallback entry
    - if not present, use fallback entry

What do you guys think?

src/spikeinterface/curation/bombcell_curation.py
```diff
@@ -86,6 +103,10 @@ def bombcell_label_units(
     label_non_somatic: bool = True,
     split_non_somatic_good_mua: bool = False,
     external_metrics: "pd.DataFrame | list[pd.DataFrame] | None" = None,
+    use_valid_periods: bool = False,
+    valid_periods_params: dict | None = None,
+    recompute_quality_metrics: bool = True,
```
alejoe91: I think these should not be taken care of by the bombcell function.

The quality metrics compute function already allows users to use the `valid_periods` if available. With the same rationale of being as explicit as possible and trying to minimize "hidden magin" (I know it's not so hidden, but pass me the term), I would encourage users to explicitly compute valid_unit_periods and metrics using these periods as part of their postprocessing, prior to using bombcell. So I would add them to the how to page, with extensive comments on why this is advisable.

src/spikeinterface/curation/bombcell_curation.py
```diff
@@ -263,6 +377,22 @@ def bombcell_label_units(
...
+        }
+        for metric_name, thresh in standalone_metrics.items():
+            standalone_labels = threshold_metrics_label_units(
+                metrics=combined_metrics,
+                thresholds={metric_name: thresh},
+                pass_label="pass",
+                fail_label="fail",
+                operator="and",
+                nan_policy="ignore",
+            )
```
alejoe91, line 339: You can do the final OR in one go! Right?
```suggestion
        standalone_metrics = {
            m: non_somatic_thresholds[m] for m in non_somatic_thresholds if m not in _NON_SOMATIC_BUILTIN_METRICS
        }
        if len(standalone_metrics) > 0:
            standalone_labels = threshold_metrics_label_units(
                metrics=combined_metrics,
                thresholds=standalone_metrics,
                pass_label="pass",
                fail_label="fail",
                operator="or",
                nan_policy="ignore",
            )
            is_non_somatic = is_non_somatic | (standalone_labels["label"] == "fail")
```

src/spikeinterface/curation/bombcell_curation.py
```diff
@@ -360,3 +490,84 @@ def save_bombcell_results(
 
         narrow_df = pd.DataFrame(rows)
         narrow_df.to_csv(folder / "labeling_results_narrow.csv", index=False)
+
+
+def save_valid_periods(
```
alejoe91: this should also be a function of the valid unit periods extension

src/spikeinterface/metrics/quality/quality_metrics.py
```diff
@@ -77,20 +77,6 @@ def _handle_backward_compatibility_on_load(self):
...
-        global_peak_sign_from_params = self.params.get("peak_sign")
-        if global_peak_sign_from_params is None:
-            for metric_params in self.params["metric_params"].values():
-                if "peak_sign" in metric_params:
-                    global_peak_sign_from_params = metric_params["peak_sign"]
-                    break
-            # If still not found, use <0.104.0 default, "neg"
-            if global_peak_sign_from_params is None:
-                global_peak_sign_from_params = "neg"
-            self.params["peak_sign"] = global_peak_sign_from_params
```
alejoe91, line 92: This needs to stay for backward compaibility (I think you removed it by mistake!)

src/spikeinterface/curation/bombcell_pipeline.py
```diff
@@ -0,0 +1,678 @@
...
+        # Plotting
+        "plot_histograms": True,
+        "plot_waveforms": True,
+        "plot_upset": True,
+        "waveform_ylims": (-300, 100),
+        "figsize_histograms": (15, 10),
+    }
+
+
+def preprocess_for_bombcell(
```
alejoe91: How is this `for_bombcell`? We don't have any "full pipeline" functions right now, and I think this deserves some general discussion ;)

Can you make this pipeline proposal on a separate PR? So we can move on with the bombcell specific changes and discuss pipelining separately!

alejoe91 (changes requested):
Thanks @Julie-Fabre 

I have several comments/suggestions. 
My main comment is about keeping the valid periods/QC computation out of the bombcell function and in the bombcell/curation docs.
Can you make a separate PR about the pipeline functions?

Happy to have a call if you want!

src/spikeinterface/curation/bombcell_pipeline.py
chrishalcrow: There's been discussion in the past about keeping some "suggested pipelines" for different types of input data / sorters etc. E.g. tetrodes, NP, ultra-high-density probes. I think a "original bombcell (and/or NeuroPyxles!) suggested pipeline" would fit in this really well, and all of these could be done in docs really nicely using the preprocessing pipelines (https://spikeinterface.readthedocs.io/en/stable/how_to/build_pipeline_with_dicts.html). We should discuss irl!

src/spikeinterface/curation/bombcell_curation.py
Julie-Fabre: Thanks both! 
The reason I set it like this, is because I am not sure it makes sense to use two different RPV calculation methods. In my mind, you should choose one method and then apply the threshold. But happy to change, lmk what you think


edit: Here's the fix:

- Dropped the "rpv" key. Users now put the explicit column name under thresholds["mua"] — either "sliding_rp_violation" or "rp_contamination" — and that single entry selects both which RPV metric gets computed and its threshold. The key is the method, just like every other metric in the dict. I think this makes the most sense.

- Default uses "sliding_rp_violation".

- Validation: exactly one of the two must be present; we raise a clear error if the user specifies both (ambiguous) or neither (no RPV check).

- Documented in bombcell_get_default_thresholds, bombcell_label_units, and run_bombcell_qc docstrings.

src/spikeinterface/curation/bombcell_pipeline.py
Julie-Fabre: The reason I want something like this is so that users do correct preprocessing before quality control. E.g. aligning channels, CAR/CMR, ect. If there is no suggested and coded pipeline, I think this leads to lots of errors. In the end, the goal for me, is to make quality control robust, reproducible & easy for people. This would just be a suggestion pipeline, but I think it is extremely important to have it coded and easy to use for people. 
This is for bombcell b/c I don't want to presume to create a pipeline for other approaches & analysis, but I would also be happy for there to be a general pipeline across approaches if that makes sense. I think a pipeline (whether only for bombcell or not) though is really needed. We can make it clear this is just a suggestion.

Julie-Fabre: Thanks Chris! Happy to discuss IRL, i'll slack you :) I think those pipelines in the docs are good, but also it really reduces errors & friction for users if the pipeline is coded up. I understand you want to make sure users have maximum freedom and agency, but I still think a guidance pipeline coded in spikeinterface would be so helpful

examples/how_to/full_pipeline_with_bombcell.ipynb
Julie-Fabre: Sorry! I left it by mistake, removing it now.

src/spikeinterface/curation/bombcell_curation.py
Julie-Fabre: This saves valid periods based on quality metrics, following the approach we implemented in bombcell, so I thought it should be there.

Julie-Fabre: But I guess we already separated valid periods from bombcell as a standalone thing, is that what you're saying? that's fair. 
We should probably also have as inputs which metrics you want to compute per chunk and which thresholds you want. For now we've been using RPVs and % spikes missing, but I always wanted to add distance metrics & presence ratio. 
I'll separate this into another PR

src/spikeinterface/curation/bombcell_curation.py
Julie-Fabre: As we discussed, bombcell_label_units is now a pure labeler with no extension computation etc.

run_bombcell_qc keeps one valid-periods knob: params["compute_valid_periods"] (bool). When on, the pipeline computes valid_unit_periods with defaults if missing, then computes quality_metrics with use_valid_periods=True. Dropped the auto-derivation of fp_threshold/fn_threshold from bombcell thresholds; instead we warn if a pre-existing valid_unit_periods extension has fp/fn that disagree with the bombcell RPV / amplitude_cutoff thresholds.

src/spikeinterface/curation/bombcell_pipeline.py
Julie-Fabre: Actually - another option is just to have a bombcell pipeline outside of spikeinetrface, on my standalone repo. I'll do this and remove pipeline stuff here.

alejoe91:
@Julie-Fabre I ported a minimal set of changes here: #4766 

I would close this since it includes all the pipelining machinery that should live elsewhere :)

alejoe91 (approved):
This looks great! I just moved the deprecation warnings to 0.106.0 ;)

PR #4500, opened 2026-04-09
Open Ephys: find root folder to correctly look for timestamps

### LINKED ISSUE #4496: Passing a folder_path that ends with the Record Node does not correctly load timestamps
When loading data from the Open Ephys GUI (via `spikeinterface.extractors.read_openephys`), if I give it a folder path that ends with a record node (e.g., `/path/to/Record Node 101`) then it does not load the timestamps correctly. I noticed this when loading data from the `ONIX Source` plugin, as the data streams there are hardware timestamped. Instead of relying on a start time and extrapolating based on the assumed sampling frequency, some of the data streams must have their timestamps loaded explicitly to ensure the times are synchronized across streams. For a 32 minute recording, the difference between the two approaches led to a 3.6 second difference between the real timestamps and the assumed timestamps, which caused confusion during post-processing analysis.

The offending line is here:
https://github.com/SpikeInterface/spikeinterface/blob/68d4258dd8810f70c62a68716da77e2885a777a9/src/spikeinterface/extractors/neoextractors/openephys.py#L340

The current behavior is to create a `recording_folder` that becomes `/path/to/Record Node 101/Record Node 101`, which does not contain the requisite `sample_numbers.npy` file to correctly load the timestamps.

My proposed fix is simple, as seen below:

```python
if not str(folder_path).endswith(record_node):
    recording_folder = Path(folder_path) / record_node
else:
    recording_folder = Path(folder_path)
```

I wanted to be more comprehensive and add a unit test to confirm that this works, now and in the future. However, as far as I know there are no existing datasets in the `datalad` project that contain hardware timestamped data, and I am not sure how to contribute a new file to `datalad`. I would be happy to provide an example recording session that contains the hardware timestamped data, so that a unit test can be added that loads the data and asserts that the end time (or total time) is correct.

alejoe91:
Fixes #4496


alejoe91:
@bparks13 this should fix the issue. It also covers the case where one passes an individual experiment/recording

chrishalcrow:
Looks good - works for my locally. Could add a test here:

https://github.com/SpikeInterface/spikeinterface/blob/89381fe5b6ad08a160f3b61b5f6e2b1f292ec96b/src/spikeinterface/extractors/tests/test_neoextractors.py#L158

something like:

``` python
def check_timestamps_loaded_from_subfolder(self):
"""
Checks if the timestamps are loaded even if user inputs a subfolder as the data path.
"""

folder_path = local_folder / r"openephysbinary/v0.6.x_neuropixels_multiexp_multistream/Record Node 101/experiment1/recording1"
recording = self.ExtractorClass(folder_path, load_sync_timestamps=True)
assert recording.get_time_info().get('time_vector') is not None
```

(I wrote this in the github box... treat as pseudocode!)

bparks13:
@alejoe91 Perfect, thank you. +1 for adding checks for the various levels of folder paths that can be given, I didn't even think about that.

@chrishalcrow If the `load_sync_timestamps=True` parameter is not set, then internally the time vector is not created? That definitely makes it easier to add a unit test since we don't need to add a new dataset. I assumed that it would create a time vector using the start time and the assumed sampling frequency, but it sounds like that is not the case. Or, at the very least, there is a difference in the data structure that we can leverage here.

chrishalcrow:
> @chrishalcrow If the `load_sync_timestamps=True` parameter is not set, then internally the time vector is not created? That definitely makes it easier to add a unit test since we don't need to add a new dataset. I assumed that it would create a time vector using the start time and the assumed sampling frequency, but it sounds like that is not the case. Or, at the very least, there is a difference in the data structure that we can leverage here.

Yeah, the time vector is pretty huge since it's the length of the total samples of the recording. So if we're using the "start time + sampling frequency" method, we never make it!

alejoe91:
Added a test to check that timestamps get loaded and are the same when pointing to various folder levels

chrishalcrow (approved):

PR #4503, opened 2026-04-09
Add amplitude calculation to kilosort4 importer

chrishalcrow:
VERY DRAFTY!!!

For @oliche to play with.

This PR adds the ability to estimate what we would call spike_amplitudes (uV valued amp estimates for each spike) from KiloSort4 output.

How? 

Take the pc features. These are whitened, PC'd waveforms on each channel (note: PC sparsity does not match template sparisty). 

First, invert the PC mapping using the `wPCA` mapping stored in ops. You now have approx whitened waveforms (approx because PC+inversePC smooths out the waveforms). 

Second, approximately unwhiten these by using the inverse whitening matrix (restricted to the channels your unit is on.) (this is an approx because you should really apply the full inverse matrix to the full waveform instead of the restricted matrix on the restricted waveform). You now have unwhitened waveforms.

Third, find the minimum of the unwhitened waveforms.

Fourth, scale these with `gain_to_uV` and `offset_to_uV`.

Main logic in `_make_amplitudes` function

To see if this works, I loaded the kilosort ouput, attached the recording and computed spike amplitudes from scratch. Note that our spike ampltiude calculation is different than this. In ours, you just take the value of the trace at the spike time, on the main channel.

So: the calculations are a bit different and the calculation is only approximate on the KS side. How do the results look? Like this:

Screenshot from 2026-04-09 17-38-58

Amplitudes from this PR on the left. Amplitudes recomputed from the recording on the right.

I'll do some more thorough analysis when I get some time. For now: this PR is ready to play with!! It's possible that we don't want this much KS4-specific wrangling in spikeinterface. Could be worth doing this in another repo??


chrishalcrow:
@JoeZiminski might also be interested in testing this. At the moment, it's just amplitudes and just for KS4.

PR #4505, opened 2026-04-09
Make version mismatch warning informative and actionable

### LINKED ISSUE #4495: Version Upgrade Warnings
`UserWarning: Versions are not the same. This might lead to compatibility errors. Using spikeinterface==0.103.2 is recommended.`

Upgrading to a new version of spike interface to take advantage of new fixes on recordings saved with a previous version (in my case 0.103.2 --> 0.104.1) triggered this warning.  

In some sense I understand version changes might lead to changes in defaults and so on but this type of warning does not really help a user that upgraded for a reason. Was there really a breaking chnage between these two versions? If so could the warning be more targeted?

h-mayorquin:
The version mismatch warning in `_load_extractor_from_dict` told users to downgrade, which is unhelpful when they upgraded on purpose. If the object loads successfully it is probably fine, so the warning should convey two things: that there may be minor differences in default parameters or behavior across versions, and that re-saving the object (via `save()`, `save_to_folder()`, or `save_to_zarr()`) stamps the current version and removes the warning. Known backward-compatibility issues continue to be handled silently by `_handle_backward_compatibility`.

Closes #4495

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -1144,11 +1144,14 @@ def _load_extractor_from_dict(dic) -> "BaseExtractor":
     extractor_class = _get_class_from_string(class_name)
 
     assert extractor_class is not None and class_name is not None, "Could not load spikeinterface class"
-    is_old_version = not _check_same_version(class_name, dic["version"])
+    module_name = class_name.split(".")[0]
+    current_version = importlib.import_module(module_name).__version__
+    saved_version = dic["version"]
+    is_old_version = not _check_same_version(class_name, saved_version)
     if is_old_version:
         warnings.warn(
-            f"Versions are not the same. This might lead to compatibility errors. "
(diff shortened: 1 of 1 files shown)


h-mayorquin (commented):
After the meeting on 2026-04-16 we decided to remove this as this is too general and the recording and sorting have been more or less table. We should have more specific warnings on the analyzer

h-mayorquin:
> After the meeting on 2026-04-16 we decided to remove this as this is too general and the recording and sorting have been more or less table. We should have more specific warnings on the analyzer @alejoe91 @chrishalcrow 

Done.

alejoe91 (approved):

chrishalcrow:
Ah! The peace and quiet of no unnecessary warnings

Screenshot 2026-04-17 at 07 51 13

chrishalcrow (approved):

PR #4506, opened 2026-04-09
Fix false-positive margin warning in forkserver workers

### LINKED ISSUE #4469: Wrong (?) warning about margins when using forkserver (vs fork)
I am processing some LFP on 60s chunks (100s for illustration here). I switched from fork to forkserver in multiprocessing and I get some warnings about the chunks being 1s when I save the array.
```
write_binary_recording 
engine=process - n_jobs=8 - samples_per_chunk=72,000 - chunk_memory=105.47 MiB - total_memory=843.75 MiB - chunk_duration=60.00s (1.00 minutes)
/home/ecobos/enigma-ephys/.venv/lib/python3.14/site-packages/datajoint/plugin.py:4: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources  # requires setuptools<82
/home/ecobos/enigma-ephys/.venv/lib/python3.14/site-packages/spikeinterface/preprocessing/filter.py:122: UserWarning: The margin size (25000 samples) is more than 20% of the global chunk size 2500 samples. This may lead to performance bottlenecks when chunking. Consider increasing the chunk_size or chunk_duration to minimize margin overhead.
  warnings.warn(
/home/ecobos/enigma-ephys/.venv/lib/python3.14/site-packages/spikeinterface/preprocessing/filter.py:122: UserWarning: The margin size (25000 samples) is more than 20% of the global chunk size 2500 samples. This may lead to performance bottlenecks when chunking. Consider increasing the chunk_size or chunk_duration to minimize margin overhead.
  warnings.warn(
write_binary_recording (workers: 2 processes forkserver): 100%|██████████| 2/2 [00:24<00:00, 12.10s/it]
```
 I have set the global kwargs at the start of the processing, the initial line shows the correct chunk size (60s) and it is taking the same time as before so I actually think it might be the warning that is wrong.

Note: The datajoint-related warning is a line I get everytime I run the script.py file I am using. When using forkserver a new process is created to be the server of processes so I imagine that's where that is printed.

h-mayorquin:
The margin-vs-chunk-size warning in `FilterRecording.__init__` fires at filter construction time using `get_global_job_kwargs()`. With `forkserver` (or `spawn`), worker processes start fresh and don't inherit the parent's global job kwargs, so `get_global_job_kwargs()` returns the default `chunk_duration="1s"`. This makes the margin appear huge relative to the chunk size, triggering a spurious warning even though the actual processing uses the correct (larger) chunk size.

The fix guards the warning with `is_set_global_job_kwargs_set()`. If global job kwargs haven't been explicitly set in the current process, the check is skipped because the default values are not meaningful for this comparison.

Closes #4469

### DIFF
--- a/src/spikeinterface/core/__init__.py
+++ b/src/spikeinterface/core/__init__.py
@@ -83,6 +83,7 @@ from .globals import (
     get_global_job_kwargs,
     set_global_job_kwargs,
     reset_global_job_kwargs,
+    is_set_global_job_kwargs_set,
 )
 
 # tools
--- a/src/spikeinterface/preprocessing/filter.py
+++ b/src/spikeinterface/preprocessing/filter.py
@@ -3,7 +3,12 @@ import warnings
 import numpy as np
 
 from spikeinterface.core.core_tools import define_function_handling_dict_from_class
-from spikeinterface.core import get_chunk_with_margin, ensure_chunk_size, get_global_job_kwargs
+from spikeinterface.core import (
+    get_chunk_with_margin,
+    ensure_chunk_size,
+    get_global_job_kwargs,
+    is_set_global_job_kwargs_set,
+)
 
 from .basepreprocessor import BasePreprocessor, BasePreprocessorSegment
 
@@ -118,7 +123,7 @@ class FilterRecording(BasePreprocessor):
         margin = int(margin_ms * fs / 1000.0)
 
         global_job_kwargs_chunk_size = ensure_chunk_size(recording, **get_global_job_kwargs())
-        if margin > MARGIN_TO_CHUNK_PERCENT_WARNING * global_job_kwargs_chunk_size:
+        if is_set_global_job_kwargs_set() and margin > MARGIN_TO_CHUNK_PERCENT_WARNING * global_job_kwargs_chunk_size:
             warnings.warn(
                 f"The margin size ({margin} samples) is more than {int(MARGIN_TO_CHUNK_PERCENT_WARNING * 100)}% "
                 f"of the global chunk size {global_job_kwargs_chunk_size} samples. This may lead to performance bottlenecks when "


alejoe91 (approved):

smarkoco (commented):
Did this make it into the latest release? with v0.104.7, I am still getting the below warning:


`UserWarning: The margin size (8148 samples) is more than 20% of the global chunk size 10240 samples. This may lead to performance bottlenecks when chunking. Consider increasing the chunk_size or chunk_duration to minimize margin overhead.
`

This is defaulting back to '1s' at 10240Hz sample rate, even though I am setting `si.set_global_job_kwargs(chunk_duration='20s')` within the worker state


I also don't see the relevent lines in `filter.py`, but I could be missing something

Thanks!!

alejoe91:
> Did this make it into the latest release? with v0.104.7, I am still getting the below warning:
> 
> `UserWarning: The margin size (8148 samples) is more than 20% of the global chunk size 10240 samples. This may lead to performance bottlenecks when chunking. Consider increasing the chunk_size or chunk_duration to minimize margin overhead. `
> 
> This is defaulting back to '1s' at 10240Hz sample rate, even though I am setting `si.set_global_job_kwargs(chunk_duration='20s')` within the worker state
> 
> I also don't see the relevent lines in `filter.py`, but I could be missing something
> 
> Thanks!!

Hi @smarkoco 

0.104.7 is a bug-fix release from 0.104.*. This fix will make it into 0.105.0 ;)

smarkoco:
Okay I see, great, thanks ! For now I'll just ignore lol

PR #4509, opened 2026-04-13
fix WaveformExtractor backwards compatiblity: run backwards compatiblity migration before validating metrics

### LINKED ISSUE #4508: Legacy `WaveformExtractor` load path calls `set_params` before `_handle_backward_compatibility_on_load`, rejecting deprecated metric names that the compat handler is designed to migrate
[`_read_old_waveforms_extractor_binary` in `spikeinterface/core/waveforms_extractor_backwards_compatibility.py`](https://github.com/SpikeInterface/spikeinterface/blob/89381fe5b6ad08a160f3b61b5f6e2b1f292ec96b/src/spikeinterface/core/waveforms_extractor_backwards_compatibility.py#L447) invokes `ext.set_params()` (which triggers `_set_params()` validation) before running `ext._handle_backward_compatibility_on_load()`. For extensions whose on-disk `params.json` contains metric names that were renamed in 0.104 (e.g. `peak_to_valley`, `velocity_above`, `l_ratio`, …), validation raises a `ValueError` before the compat handler ever runs — even though the compat handler exists specifically to migrate those names.

The result is that legacy `WaveformExtractor` folders that SI is otherwise still able to read cannot be loaded on SI ≥ 0.104 without rewriting `params.json` on disk, which is not always possible (e.g. Code Ocean data assets are immutable).

### Reproduction

Create or obtain a WaveformExtractor folder that has a `template_metrics` or `quality_metrics` extension whose `params.json` contains any deprecated metric name. For example, a `template_metrics/params.json` with:

```json
{
  "metric_names": ["peak_to_valley", "velocity_above", "velocity_below"],
  "metric_params": {
    "velocity_above": {"min_channels_for_velocity": 3, "min_r2_velocity": 0.5}
  }
}
```

Then:

```python
import spikeinterface as si
from spikeinterface.core import load_sorting_analyzer_or_waveforms
analyzer = load_sorting_analyzer_or_waveforms("/path/to/legacy_waveforms")
```

yields:

```
ValueError: The metric 'peak_to_valley' has been re-named or re-organized.
You can now compute it using the metric name 'peak_to_trough_duration'.
The metric 'velocity_above' has been re-named or re-organized.
You can now compute it using the metric name 'velocity_fits'.
...
```

### Root cause

[`waveforms_extractor_backwards_compatibility.py:631-637`](https://github.com/SpikeInterface/spikeinterface/blob/89381fe5b6ad08a160f3b61b5f6e2b1f292ec96b/src/spikeinterface/core/waveforms_extractor_backwards_compatibility.py#L632-L637) runs:

```python
new_params = ext._set_params()
updated_params = make_ext_params_up_to_date(ext, params, new_params)
ext.set_params(**updated_params, save=False)        # ← validation fires here
if ext.need_backward_compatibility_on_load:
    ext._handle_backward_compatibility_on_load()    # ← never reached
```

The deprecated-name validation in `BaseMetricExtension._set_params` ([`analyzer_extension_core.py:1127-1134`](https://github.com/SpikeInterface/spikeinterface/blob/89381fe5b6ad08a160f3b61b5f6e2b1f292ec96b/src/spikeinterface/core/analyzer_extension_core.py#L1128-L1134)) exists to catch code paths that are constructing extensions with old names at runtime. But it's also triggered on the legacy load path, where the deprecated names come from on-disk `params.json` — which is exactly the case the per-extension `_handle_backward_compatibility_on_load` handlers are designed to handle.

[`ComputeTemplateMetrics._handle_backward_compatibility_on_load` (`metrics/template/template_metrics.py:100-171`)](https://github.com/SpikeInterface/spikeinterface/blob/89381fe5b6ad08a160f3b61b5f6e2b1f292ec96b/src/spikeinterface/metrics/template/template_metrics.py#L100-L171) has full, authoritative migration logic for `num_positive_peaks`/`num_negative_peaks` → `number_of_peaks`, `velocity_above`/`velocity_below` → `velocity_fits` (including renaming `min_channels_for_velocity` → `min_channels` etc. inside `metric_params`), `peak_to_valley` → `peak_to_trough_duration`, and `peak_trough_ratio` → `waveform_ratios`. `ComputeQualityMetrics._handle_backward_compatibility_on_load` does the same for `isolation_distance`/`l_ratio` → `mahalanobis`. None of it runs, because validation aborts the load first.

### Proposed fix

Catch the deprecation `ValueError` in `_read_old_waveforms_extractor_binary`, install the raw params on the extension so the compat handler can see them, run the handler, and replay `set_params` with the migrated values:

```python
# waveforms_extractor_backwards_compatibility.py, replacing lines 634-637
try:
    ext.set_params(**updated_params, save=False)
except ValueError as e:
    if "has been re-named or re-organized" not in str(e):
        raise
    # Legacy params.json contains deprecated metric names. Install the
    # raw params so _handle_backward_compatibility_on_load can see them,
    # run the compat handler to rename metric_names AND migrate
    # metric_params sub-keys, then replay set_params with migrated values.
    ext.params = dict(updated_params)
    if ext.need_backward_compatibility_on_load:
        ext._handle_backward_compatibility_on_load()
    ext.set_params(**ext.params, save=False)
else:
    if ext.need_backward_compatibility_on_load:
        ext._handle_backward_compatibility_on_load()
```

The contained change: one function, no signature changes, non-legacy folders take the existing fast path. Happy to send a PR if this approach looks reasonable.

### Environment

- spikeinterface: 0.104.1
- python: 3.13

galenlynch:
For extensions whose on-disk `params.json` contains metric names that were renamed in 0.104 (e.g. `peak_to_valley`, `velocity_above`, `l_ratio`, …), validation raises a `ValueError` before the compat handler ever runs, which prevents loading these saved metrics.

This is because [`_read_old_waveforms_extractor_binary`](https://github.com/SpikeInterface/spikeinterface/blob/89381fe5b6ad08a160f3b61b5f6e2b1f292ec96b/src/spikeinterface/core/waveforms_extractor_backwards_compatibility.py#L632-L637) validates before migrating:

```python
new_params = ext._set_params()                                        # 1. get defaults
updated_params = make_ext_params_up_to_date(ext, params, new_params)  # 2. merge on-disk with defaults
ext.set_params(**updated_params, save=False)                          # 3. validate + set — (errors)
if ext.need_backward_compatibility_on_load:
    ext._handle_backward_compatibility_on_load()                      # 4. migrate — never reached
```

Two things go wrong before the compat handler gets a chance to run:

1. Step 3: `_set_params` rejects deprecated names like `peak_to_valley` or `l_ratio`.
2. Step 2: `make_ext_params_up_to_date` strips any `metric_params` keys that don't exist in the new defaults: deprecated sub-params (e.g. `velocity_above.min_channels_for_velocity`) are silently dropped before the compat handler can migrate them to their new names (e.g. `velocity_fits.min_channels`).

This PR changes the order of operations so the migration handler runs first on the raw on-disk parameters, which are then pruned and validated by `make_ext_params_up_to_date` and `set_params`, respectively. Running the compat handler on the raw on-disk params first is what `SortingAnalyzer` already does for non-legacy folders, so the change would just make `WaveformExtractor` have the same logical flow.

Fixes: #4508

### DIFF
--- a/src/spikeinterface/core/waveforms_extractor_backwards_compatibility.py
+++ b/src/spikeinterface/core/waveforms_extractor_backwards_compatibility.py
@@ -629,12 +629,16 @@ def _read_old_waveforms_extractor_binary(folder, sorting):
                     pc_all[mask, ...] = pc_one
                 ext.data["pca_projection"] = pc_all
 
-        # update params
-        new_params = ext._set_params()
-        updated_params = make_ext_params_up_to_date(ext, params, new_params)
-        ext.set_params(**updated_params, save=False)
+        # Install raw on-disk params and run compat handler first,
+        # matching what AnalyzerExtension.load does for non-legacy folders.
+        ext.params = dict(params)
         if ext.need_backward_compatibility_on_load:
             ext._handle_backward_compatibility_on_load()
+
+        # Now merge and validate — deprecated names are already migrated.
+        new_params = ext._set_params()
+        updated_params = make_ext_params_up_to_date(ext, ext.params, new_params)
+        ext.set_params(**updated_params, save=False)
         ext.run_info = None
 
         sorting_analyzer.extensions[new_name] = ext


galenlynch:
@chrishalcrow

zm711:
Is there a test we could add that might help us catch a problem like this in the future. Not a simple change-detector test but a way to validate our backward compatibility better? I don't think that should slow down this PR, but I have always believed we should try to prevent these issues the best we can or we make it clear that we are having a clean break (e.g. we discussed dropping all waveform extractors at one point, but decided against it unless it has changed again).

chrishalcrow:
> Is there a test we could add that might help us catch a problem like this in the future. Not a simple change-detector test but a way to validate our backward compatibility better? I don't think that should slow down this PR, but I have always believed we should try to prevent these issues the best we can or we make it clear that we are having a clean break (e.g. we discussed dropping all waveform extractors at one point, but decided against it unless it has changed again).

Good point Zach - I've been thinking about this a bit. We could do something like: set up a GH action which makes an analyzer for each spikeinterface version, then load/tests it with the latest release. It's on my to do list, but is a bit of a project. Would be nice to set up uv+caching in the CI first, so that quickly installing 10 si versions was fast!

chrishalcrow (approved):
Great, thanks! LGTM and nice that the logic follows the analyzer more closely :)

alejoe91 (approved):

alejoe91:
LGTM too!

samuelgarcia:
@galenlynch very good catch thank you!
You are still using the old waveforms extractor ?

galenlynch:
We have old spike sorting results, and want to be able to read them and more recent spike sorting results in a consistent environment.

PR #4513, opened 2026-04-14
amplitude cutoff should require amplitude scalings extension

Julie-Fabre:
This is a proposal (please close if you don't agree) to require amplitude scalings for the amplitude cutoff metric. 
I also modified in conftest.py make_small_analyzer & sorting_analyzer_simple to have amplitude scalings - let me know if that's right, I am not sure.


samuelgarcia:
I think I am OK with this a metrics should refer to only on thing.

zm711:
For reproducibility it makes sense to rely on only one thing. But I am a bit worried about past users that may have determined their amplitudes previously without scalings. If they want to recalculate this then we error. We allow templates to be fast templates or based on waveforms so we aren't completely strict on this. 

I suppose I could go either way in the end.

alejoe91:
I recommend allowing both, but warn if spike amplitudes are used and mayvbe add a deprecation warning!

alejoe91 (approved):

alejoe91:
@Julie-Fabre @zm711 I kept `spike_amplitudes` as a fallback, but it will warn that `amplitude_scalings` will be required from 0.106.0

PR #4514, opened 2026-04-14
Add regression test for GH-4508 (deprecated metric names in legacy WE folders)

galenlynch:
- Adds a synthetic legacy WaveformExtractor fixture with deprecated quality_metrics (`l_ratio`, `isolation_distance`, `qm_params`) and template_metrics (`peak_to_valley`, `peak_trough_ratio`, `metrics_kwargs`) params
- Verifies `load_waveforms_backwards` loads without `ValueError` and that deprecated names are migrated by the compat handler before validation

### Testing

- [x] Test fails without #4509 (`ValueError: The metric 'peak_to_valley' has been re-named...`)
- [x] Test passes with the fix
- [x] Existing backwards compatibility tests still pass

Addresses: https://github.com/SpikeInterface/spikeinterface/pull/4509#issuecomment-4240151277

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,6 +122,7 @@ metrics = [
 ]
 
 test_core = [
+    "pandas<3",
     "pytest<8.4.0",
     "psutil",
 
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -122,6 +122,7 @@ metrics = [
 ]
 
 test_core = [
+    "pandas<3",
```
samuelgarcia, line 125: are we not pandas 3 compatible ??

galenlynch: It's what the rest of the pyproject uses. Not sure if the version cap is needed, but kept it for consistency.

samuelgarcia: I will check this pandas3 soon.

alejoe91 (approved):

chrishalcrow (approved):

PR #4515, opened 2026-04-14
Use uv and caching in CI

galenlynch:
Should speed things up!

### DIFF
--- a/.github/actions/build-test-environment/action.yml
+++ b/.github/actions/build-test-environment/action.yml
@@ -8,13 +8,13 @@ runs:
       run: |
         git config --global user.email "CI@example.com"
         git config --global user.name "CI Almighty"
-        pip install tabulate  # This produces summaries at the end
-        pip install -e .[test,extractors,streaming_extractors,test_extractors,full]
+        uv pip install --system tabulate  # This produces summaries at the end
+        uv pip install --system -e .[test,extractors,streaming_extractors,test_extractors,full]
       shell: bash
     - name: Install git-annex
       shell: bash
       run: |
-        pip install datalad-installer
+        uv pip install --system datalad-installer
         datalad-installer --sudo ok git-annex --method datalad/packages
         git config --global filter.annex.process "git-annex filter-process"  # recommended for efficiency
     - name: Force installation of latest dev from key-packages when running dev (not release)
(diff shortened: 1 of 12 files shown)


samuelgarcia:
very cool.
why do we need the `--system` flag ? for caching ?

galenlynch:
So `uv` uses the `setup-python`-installed python and installs packages into that, instead of trying to use its own python and venv. That allows `python args...` commands to 'just work', instead of needing to do `uv run python args...`. Since not all maintainers are using uv lockfiles etc, I think this is the simplest way to use `uv pip` as a faster `pip`, instead of having to adopt uv's entire workflow. 

Python is also cached on many of the github runners, so `setup-python` is quite a bit faster (it's already on the runner) than using `uv` to get python (download over network).

galenlynch:
Put another way, `pip install` already installs into the system site packages if it's not run in a virtual environment. In contrast, `uv pip install` will install into a uv-maintained virtual environment even if it's _not_ run in a virtual environment. To truly emulate `pip install` with uv, you need to tell uv you really want to mutate the system site packages with `uv pip install --system`, or set an equivalent environment variable (but that affects all other uv commands, which might not be what you want).

samuelgarcia:
OK. Thanks. THis is super clear.

alejoe91 (approved):

chrishalcrow (approved):

chrishalcrow:
Thanks @galenlynch , looks like this cuts 20s or so off some of our test runs. Really really beneficial when doing the core tests, which can now take less than a minute to run in CI in full!!! Amazing.

h-mayorquin:
Heads-up: looking at the nightly "Full spikeinterface tests codecov" runs on `main`, this PR seems to have made the full run slower, not faster.

Pre-merge (last 10 successful nightly runs, Apr 7 to Apr 16): average ~38 min (range 33 to 41).
Post-merge (Apr 17 to Apr 19, 3 runs): average ~46 min (43, 48, 48).

Small sample and the cache may not have fully warmed in the first few days, but there is no sign of a speedup so far on the full matrix. The core tests do seem faster, which matches what was reported in the thread.

Two things that might be worth checking:

1. Whether the cache is actually hitting on the full run, or whether the key changes on each run.
2. Whether `uv pip install --system` is resolving to different (newer) package versions than `pip` did. I have found two latent test failures that surfaced with this PR because `uv` pulled numpy 2.x and a newer scipy, which broke `assert_array_equal(x=, y=)` kwargs and a `band=[single_value]` call. Those were pre-existing bugs, but pinning or capping resolution would prevent the next one from landing silently.

Numbers pulled via `gh run list --workflow="Full spikeinterface tests codecov"`.

alejoe91:
I think this is because the caching is broken: https://github.com/SpikeInterface/spikeinterface/actions/runs/24653179008/job/72080347711

chrishalcrow:
Looking at the individual test times in detail, the difference is almost entirely down to `test_sorting_s3_nwb_zarr`, which downloads a recording from the cloud. I don't see any caching logic here, so am a bit confused.
Timings for installing packages etc are down, which is a good sign!

PR #4518, opened 2026-04-16
Add `read_kilosort4_motion` function

### LINKED ISSUE #4470: kilosort4.log not being closed
The kilosort4.log FileHandler is never closed after run_sorter(), causing PermissionError on Windows: When using si.sorters.run_sorter("kilosort4", ...) on Windows, my local sorter output folder cannot be deleted afterwards because kilosort4.log is still locked by the Python process.

Kilosort4's setup_logger() in kilosort/run_kilosort.py (line ~451) adds a logging.FileHandler to the 'kilosort' logger. Kilosort provides a close_logger() function to clean this up, but SpikeInterface never calls it after sorting completes. Specifically in spikeinterface/sorters/external/kilosort4.py, the _run_from_folder() method calls kilosort functions that trigger setup_logger(), but after sorting finishes there is no call to close_logger(). The FileHandler stays attached to the 'kilosort' logger indefinitely, keeping kilosort4.log open for the lifetime of the Python process. Because of this, any shutil.rmtree() or file deletion on the sorter output folder fails with PermissionError: [WinError 32]. This breaks my automated pipeline that copy results to a network drive and cleans up local temp folders.

Thanks!

alejoe91:
and fix #4470


galenlynch (changes requested):
This is great! This introduces two unrelated things: (1) a new `read_kilosort4_motion` helper that reconstructs an SI `Motion` object from KS4's `ops.npy` (`yblk`, `dshift`, `batch_size`, `fs`); (2) closes the kilosort file logger to fix #4470 (Windows file lock). However, it seems to have some bugs, and doesn't yet include tests for the added code.

### Bugs

1. **`np` not imported at module level.** The new function uses `np.load`, `np.linspace`, `np.arange`, but `import numpy as np` in this file only exists locally inside `_setup_recording_arguments` (line 173). Calling `read_kilosort4_motion` as-is will raise `NameError`. Needs a module-level `import numpy as np`.

2. **`np.linspace(t_start + t_bin/2, t_end - t_bin/2)` is missing `num=`** — defaults to 50 samples regardless of `displacement.shape[0]`. The `else` branch gets the count right via `np.arange(displacement.shape[0])`; the `if recording is not None` branch silently builds a 50-long array and the interpolator will be garbage whenever the session doesn't happen to have 50 batches. Should be `np.linspace(..., num=displacement.shape[0])`.

3. **`displacement = dshift + yblk` is wrong.** In KS4, [`dshift = imin * ops['binning_depth']`](https://github.com/MouseLand/Kilosort/blob/main/kilosort/datashift.py#L211) is the estimated shift per batch per block (shape `(n_batches, n_blocks)`), and [`yblk`](https://github.com/MouseLand/Kilosort/blob/main/kilosort/datashift.py#L119-L129) is the y-center of each block; both are assigned into ops at [datashift.py:214-215](https://github.com/MouseLand/Kilosort/blob/main/kilosort/datashift.py#L214-L215). SI's [`Motion.displacement`](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/motion.py#L15-L28) is the *shift*, not *shifted position* — see the interpolator in [`get_displacement_at_time_and_depth`](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/motion.py#L87). Adding `yblk` injects a per-column offset equal to the block depth — the resulting "displacement" would be hundreds of microns instead of the few-µm drift. Should be `displacement = dshift`; `spatial_bins_um = yblk` is already correct.

Could you add a unit test with a synthetic ops.npy? That would catch (1) and (2) immediately

### Other notes
- Bare `except Exception` for the yblk/dshift check is lax — `KeyError` would be more informative.
- Logger cleanup looks fine; matches kilosort's own [`close_logger`](https://github.com/MouseLand/Kilosort/blob/main/kilosort/run_kilosort.py#L476) pattern.
- Minor: stray double blank line after `from packaging import version`.

alejoe91:
Thanks @galenlynch! All fixed in the latest commit. I also added a comment to the `Motion` function highlighting that `displacements` shold be relative to spatial bins. I was tricked myself :P

alejoe91:
@chrishalcrow @JoeZiminski 

I needed to bump up the minimal version of KS4 we supported to 4.1.1, since #4618 introduced the stable=True arg in argsort which is >=2.0. I think this is better going forward instead of keeping backward compatibility with numpy 1.26.* (which reached end of life in Oct last year: see https://endoflife.date/numpy)

This is ready to merge on my end

pyproject.toml
```diff
@@ -20,8 +20,7 @@ classifiers = [
 
 
 dependencies = [
-    "numpy>=1.20;python_version<'3.13'",
-    "numpy>=2.0.0;python_version>='3.13'",
+    "numpy>=2.0.0",
```
chrishalcrow, line 23: I agree that us using `stable = True` (which is already all over the place) has forced us to pin `numpy >= 2`. And that that's ok!

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -487,6 +490,17 @@ def _make_sparsity_from_templates(sorting, recording, kilosort_output_path):
     return ChannelSparsity(mask, unit_ids=unit_ids, channel_ids=channel_ids)
 
 
+def _make_main_channel_indices_from_templates(sorting, recording, kilosort_output_path):
+    """Constructs the `ChannelSparsity` of from kilosort output, by seeing if the
+    templates output is zero or not on all channels."""
```
chrishalcrow:
```suggestion
    """Constructs the `main_channel_indices` from kilosort output, by finding the
    channel containing the largest peak-to-peak value."""
```

chrishalcrow (approved):

src/spikeinterface/sorters/external/kilosort4.py
```diff
@@ -492,3 +497,47 @@ def _setup_json_probe_map(cls, recording, sorter_output_folder):
...
+    ops_file = sorter_output_folder / "ops.npy"
+    if not ops_file.is_file():
+        raise FileNotFoundError("'ops.npy' file not found!")
+    ops = np.load(ops_file, allow_pickle=True).item()
+    yblk = ops.get("yblk")
+    dshift = ops.get("dshift")
+    if yblk is None or dshift is None:
+        raise Exception("'yblk' and 'dshift' fields not found in ops file!")
+    displacement = dshift
+    spatial_bins_um = yblk
```
samuelgarcia, line 530: Is this ordered always ?
Does this contains duplicates ?
Is it regularly spaced ?

alejoe91: Regularly spaced, no duplicates

PR #4520, opened 2026-04-16
Move `NoiseGeneratorRecording` from core to generation

h-mayorquin:
Addresses the design discussion in #4482. 

`NoiseGeneratorRecording` mixed two purposes: a lightweight lazy recording for testing and a simulation tool with spatial correlations (`cov_matrix`) and per-channel `noise_levels`. This PR introduces a minimal `MockRecording` in `core/generate.py` (unit-variance white noise only, no extra parameters) and moves the full `NoiseGeneratorRecording` to `generation/noise_tools.py`, where `generate_noise()` already lived. `generate_recording()` and `generate_recording_by_size()` now use `MockRecording`, while `generate_ground_truth_recording()` lazy-imports `NoiseGeneratorRecording` from generation. That way, the work on #4482 can proceed from here without interfering with the testing functionality. I will migrate the tests in a follow-up PR so this diff stays as clear as possible with minimal code churn.

Backward compatibility is preserved via module-level `__getattr__` in both `core/__init__.py` and `core/generate.py`. Importing `NoiseGeneratorRecording` from the old paths still works but emits a `FutureWarning` directing users to import from `spikeinterface.generation` instead, with removal targeted for 0.106.0. I also changed `SilencedPeriodsRecording` to compose `MockRecording` + `ScaleRecording` instead of importing `NoiseGeneratorRecording` from core, to comply with a design decision once discussed with @samuelgarcia about the way modules should depend on one another (preprocessing should not depend on generation). I am happy to import from generation directly if you think that is better.

`MockRecording` still exposes the `strategy` argument for now. I would like to remove it and keep only `tile_pregenerated`, but I opened #4522 to discuss whether the simulation side needs `on_the_fly` for non-repeating noise. I will clean this up in a follow-up once that discussion settles.


cwindolf:
Hi @h-mayorquin and @samuelgarcia , I'm interested in working on simulated data and would love to see this PR merged so that I can rework my own PRs. @h-mayorquin , from the discussion at https://github.com/SpikeInterface/spikeinterface/issues/4522, it seems like you plan to remove the `strategy` argument here, will that be a part of this PR? (I guess it seems like a good time to do it?) And @samuelgarcia , I was wondering if you could help with reviewing to keep this stuff moving?

cwindolf:
Maybe to keep things moving we could merge this PR as-is and handle the strategy stuff later? I can work on the changes related to the NoiseGeneratorRecording in a follow up PR.

h-mayorquin:
I think this is up to @samuelgarcia. Alessio is on vacations now.

h-mayorquin:
I think @alejoe91 is back

@alejoe91 @samuelgarcia any blocker on this?

src/spikeinterface/core/generate.py
```diff
@@ -1246,26 +1249,19 @@ class NoiseGeneratorRecording(BaseRecording):
         The sampling frequency of the recorder.
     durations : list[float]
         The durations of each segment in seconds. Note that the length of this list is the number of segments.
-    noise_levels : float | np.ndarray, default: 1.0
```
alejoe91, line 1249: I would actually keep noise levels here. It's useful to have a mock recording test where you can easily modify them. What do you think @h-mayorquin 

I agree with rmeoving the `cov_matrix`!

h-mayorquin: Can't think on a use case for that, do you have something specific on mind?

h-mayorquin: As in, in the context of https://github.com/SpikeInterface/spikeinterface/issues/4482#issuecomment-4217022490 I don't see how that would fit on the scope of the mock.

cwindolf: Hi, just wanted to check on this. I have no opinion either way on whether noise_levels should remain in the mock (although a user could just `si.scale()` it if they're gone, no?), but wanted to see if a decision could be made just to keep this PR going. Thanks :)!

h-mayorquin: @alejoe91 I am happy to keep it if that's your intuition and that helps this move forward. We can discuss changing that later.

alejoe91: Sorry I forgot to re-reply. Let's move on. The mock recording will be only for testing purposes. We can rescale on the fly or use the generate ground truth recording for better control!

alejoe91 (approved):

src/spikeinterface/core/generate.py
samuelgarcia: I would also keep the noise level.
We need it to tests the estimation of noise level.

src/spikeinterface/core/generate.py
```diff
@@ -1246,26 +1249,19 @@ class NoiseGeneratorRecording(BaseRecording):
...
     seed : int | None, default: None
         The seed for np.random.default_rng.
     strategy : "tile_pregenerated" | "on_the_fly", default: "tile_pregenerated"
-        The strategy of generating noise chunk:
-          * "tile_pregenerated": pregenerate a noise chunk of noise_block_size sample and repeat it
-                                 very fast and cusume only one noise block.
-          * "on_the_fly": generate on the fly a new noise block by combining seed + noise block index
-                          no memory preallocation but a bit more computaion (random)
+        The strategy of generating noise chunk.
+        # TODO: Remove on_the_fly strategy after discussion, see #4522.
```
samuelgarcia, line 1258: I think it do not hurt too much to remove this feature, it was cool.
if the default is still tile_pregenerated

src/spikeinterface/preprocessing/silence_periods.py
```diff
@@ -87,16 +88,16 @@ def __init__(
...
                 num_channels=recording.get_num_channels(),
                 sampling_frequency=recording.sampling_frequency,
                 durations=[recording.select_segments(i).get_duration() for i in range(recording.get_num_segments())],
                 dtype=recording.dtype,
                 seed=seed,
-                noise_levels=noise_levels,
                 strategy="on_the_fly",
                 noise_block_size=int(recording.sampling_frequency),
             )
+            noise_generator = ScaleRecording(mock_noise, gain=noise_levels, dtype=recording.dtype)
```
samuelgarcia, line 100: why not keeping the levels ?
here this make 2 calls and to memalloc because 2 classes do get_tarces()

samuelgarcia:
Hi.
Thanks for this Ramon.
This is cool.
I am a bit late here for the review, sorry for that.

I was not ultra fan of the semantic MockRecording but we don't care.
I would have keep te noise_level in this core class, it can help at some places.
Ase keeping the stragtegy for me was OK.
But my comments are too late I guess.

h-mayorquin:
The strategy I feel strongly against for the rerasons described on https://github.com/SpikeInterface/spikeinterface/issues/4522
It just adds generality that we don't need. One method purpose needs one implementation the others need the other.


We can check the name of `MockRecording`, it is still not released. 

Regarding noise_levels the idea is that `NoiseRecording` is the batteries included version that is actually used to generate noise which is where the noise levels fit naturally. MockRecoding is just a for memory, serialization and multiprocessing tests that require numpy-array like behavior without anything else. Different classes for different purposes.

Issue #4519, opened 2026-04-16
Copy recording on `register_recording` to avoid shared mutation


h-mayorquin:
`BaseSorting.register_recording` stores the recording by reference (`self._recording = recording`). Any mutation to the recording object (setting times, shifting times, setting probes, changing properties) is visible through the sorting. This makes it hard to reason about ownership: code that modifies a recording may not know a sorting is sharing it, and code that operates on a sorting may not intend to mutate someone else's recording.

I think that we should copy the recording at registration time so that it serves as a snapshot. This way, we might want to mutate something of the recording in the registration without affecting it or vice-versa. Recordings are lazy wrappers (file paths, memmap references, channel metadata), so the copy is cheap. Was the current by-reference behavior intentional? 

This came up while thinking about adding time manipulation methods (like `shift_times`) to `BaseSorting`. If the sorting delegates to a shared recording, calling `sorting.shift_times()` would mutate the original recording as a side effect. With a copied recording, the sorting could safely manipulate its own time state without affecting anything external.

h-mayorquin:
We (@alejoe91 @JoeZiminski @chrishalcrow) discussed this on 2026-04-16 in the maintenance meeting. Some notes:

The only use case we could identify for `register_recording` is to set the timestamps. In that sense, copying the full recording at registration might be overkill. For the `sampling_frequency` + `t_start` case we could just copy the time machinery. The problematic case is when the recording has an explicit `time_vector` that is a lazy reference (HDF5 dataset, zarr array), where copying could be expensive for long recordings.

For `shift_times` on sorting, one option we discussed was to consider timestamps on a registered recording as final and throw an error if someone tries to shift after registration. 

After thinking about it more, I don't like this. I don't ike that whether `shift_times` works or raises an error  depens on internal state (whether a recording is registered). When implementing `shift_times` on recording ([#3509](https://github.com/SpikeInterface/spikeinterface/pull/3509), [#3117](https://github.com/SpikeInterface/spikeinterface/pull/3117)), @JoeZiminski and I dicussed that the [internal representation of time should be invisible to the user](https://github.com/SpikeInterface/spikeinterface/issues/3157). The same principle should apply here: whether a sorting has a registered recording or not should not change what methods are available. 

[...]

PR #4525, opened 2026-04-16
Add `get_start_time`, `get_end_time`, and `get_last_spike_frame` to `BaseSorting`

h-mayorquin:
`BaseSorting` has no public way to query the start or end time of a segment. `BaseRecording` has both (#3623). This PR adds `get_start_time` and `get_end_time` to `BaseSorting`, plus a `get_last_spike_frame` helper. With a registered recording, `get_end_time` returns the recording's end time. Without one, it returns the time of the last spike in the segment (the best the sorting can do on its own). `get_start_time` reads `_t_start` directly in both cases.

Adding these accessors however creates an ambiguity: some sorting extractors set `_t_start` on their segments at init. For example, neo sorting extractors [set `_t_start` from the signal stream](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/extractors/neoextractors/neobaseextractor.py#L625) (e.g. Neuralynx [infers it here](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/extractors/neoextractors/neobaseextractor.py#L422-L424)). If that sorting registers a recording that also starts at 100.0s, `get_start_time` would need to decide which value to return.

The decision here is that registering a recording is a statement that "these are the right timestamps" (see #4519). In consequence, `register_recording` now copies the recording's start times into the sorting segments' `_t_start`. This replaces whatever the extractor had set at init, so `get_start_time` returns consistent values.

### DIFF
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -322,11 +322,11 @@ class BaseSorting(BaseExtractor):
                     "Might be necessary for further postprocessing."
                 )
         self._recording = recording
-        # The recording is now the source of truth for timestamps.
-        # Reset the sorting's own time offset so it doesn't conflict
-        # with the recording's t_start when accessed through get_start_time/get_end_time.
-        for segment in self.segments:
-            segment._t_start = 0
+        # Copy the recording's start times into the sorting segments so that
+        # get_start_time can just read _t_start without branching.
+        # This also prevents double-counting if the extractor had set its own _t_start at init.
+        for segment_index, segment in enumerate(self.segments):
+            segment._t_start = recording.get_start_time(segment_index=segment_index)
 
     @property
     def sorting_info(self):
@@ -355,9 +355,6 @@ class BaseSorting(BaseExtractor):
     def get_start_time(self, segment_index: int | None = None) -> float:
         """Get the start time of the sorting segment.
 
-        If a recording is registered, returns the recording's start time.
-        Otherwise returns the sorting segment's own t_start (or 0.0).
-
         Parameters
         ----------
         segment_index : int or None, default: None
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/basesorting.py
```diff
@@ -322,6 +322,11 @@ def register_recording(self, recording, check_spike_frames: bool = True):
                     "Might be necessary for further postprocessing."
                 )
         self._recording = recording
+        # The recording is now the source of truth for timestamps.
+        # Reset the sorting's own time offset so it doesn't conflict
+        # with the recording's t_start when accessed through get_start_time/get_end_time.
+        for segment in self.segments:
+            segment._t_start = 0
```
alejoe91:
```suggestion
            segment._t_start = None
```
Maybe  using `None` is less ambiguous?

alejoe91:
LGTM! Just a minor comment. Did you remove the `get_last_spike_frame`? In that case, let's change the title of the PR

alejoe91 (approved):

src/spikeinterface/core/basesorting.py
h-mayorquin: I think this makes sense and it is better. It will be less likely to lead to confusion if the set timestamps does not start in 0.

h-mayorquin:
> LGTM! Just a minor comment. Did you remove the `get_last_spike_frame`? In that case, let's change the title of the PR

I think it was a merge conflict it is now on the PR again.

alejoe91 (approved):

alejoe91:
Merging as soon as tests pass :)

chrishalcrow:
This is great!! A consequence I didn't realise: I usually don't load the recording when I load an analyzer. Previously this meant the analyzer had no time information in it. So if you did any time stuff, it would assume `_t_start=0`. But if I opened the analyzer with the recording, it would get the real `_t_start`. So my time information changed if I loaded the analyzer with the recording or not.

Now, when you run `run_sorter`, this registers the recording the sorting get `_t_start`. If I then save an analyzer then load it without the recording, the time information consistent with the recording. Nice!

h-mayorquin:
Hi, @chrishalcrow, I did think about that scenario and I am reverting @alejoe91 suggestion of setting `_t_start = None` that we did above to handle this:

That case only works if we copy the recording's start time into `_t_start` at registration. `register_recording` now copies `recording.get_start_time()` into each segment's `_t_start`, so `get_start_time` just reads it and the value persists across save/load without the recording.

**One case this does not fully address**: if the registered recording has an explicit `time_vector` with irregular spacing, the copy approach only preserves the first timestamp (`time_vector[0]`, note that this is bettter than losing everything) in the sorting's `_t_start`. I am unsure what to do this in this case but I think it warrants a separate issue and is related to  #4519.

PR #4526, opened 2026-04-17
Add `num_channels` to `BaseRecordingSegment`

h-mayorquin:
draft


src/spikeinterface/core/baserecording.py
```diff
@@ -910,9 +910,27 @@ def __init__(self, sampling_frequency=None, t_start=None, time_vector=None):
...
+        # at construction or by assigning `self._num_channels = N`). Otherwise derive from
+        # the container recording through the weakref established in `add_segment`.
+        if self._num_channels is not None:
+            return self._num_channels
+        if self._parent_extractor is None:
+            return None
+        container_recording = self._parent_extractor()
+        if container_recording is None:
+            return None
+        return container_recording.get_num_channels()
```
alejoe91: Since the segments are supposed to be added with the [`BaseExtractor.add_segment()`](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/base.py#L149-L151), which always registers the `self._parent_extractor` weakref available as `self.parent_extractor` property, I wonder whether we need any of this?
```suggestion
        return self.parent_extractor.get_num_channels()
```
@h-mayorquin any place in mind where this won't work?

src/spikeinterface/core/binaryrecordingextractor.py
```diff
@@ -173,7 +173,7 @@ def __del__(self):
 class BinaryRecordingSegment(BaseRecordingSegment):
     def __init__(self, file_path, sampling_frequency, t_start, num_channels, dtype, time_axis, file_offset):
         BaseRecordingSegment.__init__(self, sampling_frequency=sampling_frequency, t_start=t_start)
-        self.num_channels = num_channels
+        self._num_channels = num_channels
```
alejoe91: For the same reason, we can drop this IMO. Segments will always be added to a parent extractor and can inherit the `num_channels` from it

src/spikeinterface/core/generate.py
```diff
@@ -1359,7 +1359,7 @@ def __init__(
         BaseRecordingSegment.__init__(self, sampling_frequency=sampling_frequency)
 
         self.num_samples = num_samples
-        self.num_channels = num_channels
+        self._num_channels = num_channels
```
alejoe91: same here

src/spikeinterface/preprocessing/average_across_direction.py
```diff
@@ -103,7 +103,7 @@ def __init__(
     ):
         BasePreprocessorSegment.__init__(self, parent_recording_segment)
         self.parent_recording_segment = parent_recording_segment
-        self.num_channels = num_channels
+        self._num_channels = num_channels
```
alejoe91: same here

alejoe91 (changes requested):
Thanks @h-mayorquin 

Given that segments will always be added to a `BaseRecording` object, see my proposed changes! Curious to hear your thoughts!

src/spikeinterface/core/baserecording.py
alejoe91: @h-mayorquin any thought on this? With this assumption that every segment belongs to a recording object, I think we can greatly simplify the retrieval of num channels and remove additional logic and vars (e.g., `selg._num_channels`)

h-mayorquin:
Hi, @alejoe91 sorry for droping this for a while.

I agree with your suggestion and I simplified along these lines. I dropped the `num_channels` property and the `_num_channels` var, and `get_num_channels()` now just asks the parent.

There is one case where a bare `return self.parent_extractor.get_num_channels()` does not work: segments are constructed before `add_segment()` attaches the parent, and a few of them need the channel count inside `__init__`, before any parent exists. `BinaryRecordingSegment` uses it to compute the on-disk layout (`bytes_per_sample` / `num_samples` from file size), and the noise generator segment uses it to size its noise block. Those cannot ask a parent that is not there yet. Maybe we should pass the recording to the segment at construction time but that's a can of worms by itself and this is already and improvement.

So I kept a single `num_channels = None` class default on `BaseRecordingSegment`. Segments that know their count at construction set `self.num_channels` (plain attribute, same as before, no rename), and everyone else inherits it from the parent:

```python
class BaseRecordingSegment(TimeSeriesSegment):
    num_channels = None

    def get_num_channels(self) -> int:
        if self.num_channels is not None:
            return self.num_channels
        return self.parent_extractor.get_num_channels()
```

The read-time segments (`InjectTemplates`, `InjectDrifting`, `TracePadded`, `ZeroChannelPadded`) now just call `self.get_num_channels()` and no longer take `num_channels` in their constructors.

h-mayorquin:
The errors seems unrelated.

alejoe91 (approved):
Thanks @h-mayorquin !!! Nice trick on the class attribute ;)

src/spikeinterface/core/baserecording.py
```diff
@@ -654,6 +654,17 @@ class BaseRecordingSegment(TimeSeriesSegment):
...
     """
 
+    # Segments that know their channel count at construction (e.g. BinaryRecordingSegment,
+    # which needs it before being attached to a parent to compute the on-disk layout) set
+    # self.num_channels. Segments that don't leave this default and inherit the count from the
+    # parent recording, which is always attached by the time get_traces runs.
+    num_channels = None
+
+    def get_num_channels(self) -> int:
+        if self.num_channels is not None:
```
samuelgarcia:
```suggestion
        if hasattr(self, 'num_channels') and self.num_channels is not None:
```

src/spikeinterface/core/baserecording.py
```diff
@@ -654,6 +654,17 @@ class BaseRecordingSegment(TimeSeriesSegment):
     Abstract class representing a multichannel timeseries, or block of raw ephys traces
     """
 
+    # Segments that know their channel count at construction (e.g. BinaryRecordingSegment,
+    # which needs it before being attached to a parent to compute the on-disk layout) set
+    # self.num_channels. Segments that don't leave this default and inherit the count from the
+    # parent recording, which is always attached by the time get_traces runs.
+    num_channels = None
```
samuelgarcia:
```suggestion
```

src/spikeinterface/core/baserecording.py
h-mayorquin: Ok, done.

src/spikeinterface/core/baserecording.py
h-mayorquin: modified.

PR #4527, opened 2026-04-17
refactor: centralize zarr object-blob writes in _write_object_array

galenlynch:
Groups the six call sites that write JSON/Pickle-coded object arrays (recording provenance, sorting provenance, and extension dict/object data) behind a single helper in `zarrextractors.py`. Behavior-preserving on zarr v2 — verified by analyzer round-trip diff against `main` (identical bytes except for the non-deterministic `runtime_s` field).

### Why

The `create_dataset(..., object_codec=numcodecs.JSON() | Pickle())` pattern is a zarr v2 API shape — `object_codec=` is gone in zarr-python 3, where object codecs live in `filters=` (wrapped via `numcodecs.zarr3.*`). Centralizing the six scattered call sites now means the zarr v3 cut-over only has to flip one function instead of chasing the pattern across two files.

This is prep work for the eventual zarr-python 3 migration (currently blocked on `hdmf-zarr` upstream, which pins `zarr<3`). The refactor stands on its own and is safe to merge independently of that migration's timeline.

### Changes

- Add `_write_object_array(group, name, data, codec="json"|"pickle")` to `src/spikeinterface/core/zarrextractors.py`.
- Replace four provenance-write sites in `SortingAnalyzer.create_zarr` and two extension-write sites in the zarr extension path with calls to the helper.
- Drop the now-unused `import numcodecs` from two functions in `sortinganalyzer.py`.

### Test plan
- [x] `pytest src/spikeinterface/core/tests/test_zarrextractors.py`
- [x] `pytest src/spikeinterface/core/tests/test_sortinganalyzer.py`
- [x] `pytest src/spikeinterface/core/tests/test_baserecording.py src/spikeinterface/core/tests/test_basesorting.py src/spikeinterface/core/tests/test_analyzer_extension_core.py`
- [x] Byte-level analyzer round-trip diff vs. `main` — identical modulo `runtime_s`.


alejoe91 (approved):
Thanks @galenlynch

Issue #4366, opened 2026-02-04
Kilosort output for sorting analyzer


Djoels:
As mentioned in the [KS (2.5/4) documentation](https://kilosort.readthedocs.io/en/stable/export_files.html), part of the sorting output prepared for curation is a number of npy files containing templates, similar templates, PC's, ... which are also calculated using a sorting analyzer for a number of metrics.

Could these features somehow be reused, or are they incompatible with sorting analyzer way of working?

Furthermore, what would be the advice when analyzing a recording processed by Kilosort: use the original recording or use the temp_wh.dat file  produced by KS which contains already preprocessed (car, whitening, highpass filtered) data?

Djoels:
Hmm, I already stumbled [upon an indication](https://github.com/SpikeInterface/spikeinterface/issues/3919#issuecomment-2880009752) of whether the temp_wh.dat file should be used: I guess not then, because it needs to be "unwhitened" before applying sorting analyzer?

chrishalcrow:
Hey @Djoels I'm actively working on this! There's already some functionality to import kilosort output as an analyzer (https://spikeinterface.readthedocs.io/en/latest/how_to/import_kilosort_data.html) You'll need use the main branch of spikeinterface (locally clone and install rather than just "pip install spikeinterface"). At the moment it just loads the "easy" stuff.

But I have a branch here which adds the "hard" stuff: https://github.com/SpikeInterface/spikeinterface/compare/main...chrishalcrow:spikeinterface:add-amps-to-kilosort-convert but it's not quite ready for a proper PR yet. You've already figured out the hard stuff is: some output is whitened, some is in PC space, some is in physical space, and we can only approximately go between the spaces. But think I've figured it all out for KS4 output now.

For an analyzer, I take a raw recording and lightly preprocess it (bandpass + CMR) then combine with the kilosort sorting output - this seems to work pretty well for my data (chronic NP2).

Djoels:
That is great news!
At our labs we're mostly still using KS2.5 so I'm keen to find out whether this would also work.
Our recordings are mostly on custom platform or Maxwell recordings (so in vitro HD-MEA's).
Will give it a try shortly, thanks!

pas-calc:
[...]

PR #4529, opened 2026-04-17
Update `read_kilosort_as_analyzer` to allow for multiple-probes and user-defined recording

chrishalcrow:
This PR improves the `read_kilosort_as_analyzer` function, as discussed in (#4366).

The plan:

- [x] Allow user to pass their own recording
- [x] Allow for probe groups
- [x] Use `channel_map.npy` for device channel indices instead of range
- [x] If recording given, make sure it's consistent with `probe.prb` file in output


src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -359,25 +364,49 @@ def read_kilosort_as_analyzer(folder_path, unwhiten=True, gain_to_uV=None, offse
...
+        all_contact_positions = np.vstack([probe.contact_positions for probe in probegroup.probes])
+
+        new_device_channel_indices = []
+        for recording_channel_location in recording.get_channel_locations():
+            for channel_index, probe_contact_position in enumerate(all_contact_positions):
+                if np.all(recording_channel_location == probe_contact_position):
+                    new_device_channel_indices.append(channel_index)
+                    break
+
+        if len(new_device_channel_indices) != len(all_contact_positions):
```
h-mayorquin: I think:

```
    recording.get_probegroup().set_global_device_channel_indices(new_device_channel_indices)

```

But review whether this is actually set or just returns a copy.

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -324,6 +327,8 @@ def read_kilosort_as_analyzer(folder_path, unwhiten=True, gain_to_uV=None, offse
     ----------
     folder_path : str or Path
         Path to the output Phy folder (containing the params.py).
+    recording : BaseRecording
```
zm711, line 365: Just as an FYI this breaks positional arguments for people... What is your logic for being okay with that?

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -413,14 +445,9 @@ def _make_locations(sorting_analyzer, kilosort_output_path):
...
-    # Check that the spike locations vector is the same size as the spike vector
+    # When recording is given, need to trim spike locations to match spikes in sorting
     num_spikes = len(sorting_analyzer.sorting.to_spike_vector())
-    num_spike_locs = len(locs_np)
-    if num_spikes != num_spike_locs:
-        warnings.warn(
-            "The number of spikes does not match the number of spike locations in `spike_positions.npy`. Skipping spike locations."
-        )
-        return
+    locs_np = locs_np[:num_spikes]
```
zm711, line 478: How do we know this is the correct set? the problem is KS4 pads which means we have spikes both before and after the recording. So just taking 0...n_spikes means you might take some "fake" spikes at the beginning and lose "real" spikes at the end....

zm711 (commented):
Couple questions @chrishalcrow.

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -394,25 +399,40 @@ def read_kilosort_as_analyzer(folder_path, unwhiten=True, gain_to_uV=None, offse
...
+        user_gave_recording = False
+
+        # kilosort occasionally contains a few spikes just beyond the recording end point, which can lead
+        # to errors later. To avoid this, we pad the recording with an extra second of blank time.
+        duration = sorting.segments[0]._all_spikes[-1] / sampling_frequency + 1
+
+        # to make the initial analyzer, we'll use a fake recording and set it to None later
+        recordings = []
+        for probe in probegroup.probes:
+            one_recording, _ = generate_ground_truth_recording(
```
alejoe91, line 427: Use `MockRecording` instead

src/spikeinterface/extractors/phykilosortextractors.py
```diff
@@ -394,25 +400,39 @@ def read_kilosort_as_analyzer(folder_path, unwhiten=True, gain_to_uV=None, offse
...
+        channel_map = np.load(phy_path / "channel_map.npy")
+        recording = recording.select_channels(recording.channel_ids[channel_map])
+        user_gave_recording = True
+
+    else:
+        user_gave_recording = False
+
+        # kilosort occasionally contains a few spikes just beyond the recording end point, which can lead
+        # to errors later. To avoid this, we pad the recording with an extra second of blank time.
+        duration = sorting.segments[0]._all_spike_times[-1] / sampling_frequency + 1
```
alejoe91, line 423:
```suggestion
        duration = sorting.get_last_spike_frame(segment_index=0) / sampling_frequency + 1
```
Cleaner: https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/basesorting.py#L454

src/spikeinterface/extractors/phykilosortextractors.py
alejoe91: The `read_kilosort_as_analyzer` has just been out since the last major release (0.104.0), so I think that it's ok to "break" this. On the other hand, we could also just move it at the end. @chrishalcrow what do you think?

src/spikeinterface/extractors/phykilosortextractors.py
chrishalcrow: We discussed in person.

Pretty sure this was fixed in response to this issue (https://github.com/MouseLand/Kilosort/issues/739) in this commit (https://github.com/MouseLand/Kilosort/commit/b82e5620b16afe5f0ff4d805ec23801166c604f8). This was in the 4.0.17 release, which is within all versions we support

alejoe91 (approved):

samuelgarcia (approved):

PR #4530, opened 2026-04-17
Slay: discard duplicates with `censor_period_ms` in computing RP violations

### LINKED ISSUE #4498: Duplicate spikes should be removed before SLAy refractory period penalty calculated
I often see that when putative oversplitting occurs there are "duplicate spikes" that get assigned to both clusters, giving a very large peak near 0 in the cross correlogram (see example below). This makes it difficult for SLAy to find such merges given the refractory period violation penalty (indeed, it doesn't find the one below unless you set k2 to 0). We should probably filter out duplicate spikes before calculating the refractory period penalty. Also, these duplicate spikes should be removed after automerging (regardless of the algorithm used). 

Image

alejoe91:
fixes #4498

TODO:

- [ ] Add tests


src/spikeinterface/curation/auto_merge.py
```diff
@@ -1630,10 +1684,39 @@ def compute_xcorr_and_rp(sorting_analyzer: SortingAnalyzer, pair_mask: np.ndarra
...
+            # has a large center bin when duplicates are present, which the LP filter
+            # attenuates so bin_rate_max reflects the flank rate — making RP violations
+            # detectable (unlike using the CCG alone where bin_rate_max is dominated by
+            # the duplicate peak and masks violations).
+            merged_acg = (
+                ccgs[unit_index_1, unit_index_1, :]
+                + ccgs[unit_index_2, unit_index_2, :]
+                + ccgs[unit_index_1, unit_index_2, :]
+                + ccgs[unit_index_2, unit_index_1, :]
+            )
```
chrishalcrow, line 1698: I don't think this is right. `merged_acg` later passed to `_sliding_RP_viol_pair`, a port of the SLAy function `_sliding_RP_viol`, used here:

https://github.com/saikoukunt/SLAy/blob/c3ab445f264049d4da06ca7d0835018634888cce/src/slay/stages.py#L395

In SLAy, `x_correlogram` computes the cross-correlogram of two units.

This changes the input to be the auto-correlogram of the putative merge.

src/spikeinterface/curation/auto_merge.py
```diff
@@ -88,7 +88,7 @@
         "censored_period_ms": 0.3,
     },
     "quality_score": {"firing_contamination_balance": 1.5, "refractory_period_ms": 1.0, "censored_period_ms": 0.3},
-    "slay_score": {"k1": 0.25, "k2": 1, "slay_threshold": 0.5},
+    "slay_score": {"k1": 0.25, "k2": 1, "slay_threshold": 0.5, "censored_period_ms": 0.2},
```
chrishalcrow, line 91: I'm against changing the default behavior without consulting the SLAy team.

saikoukunt: i used the ccg here instead of the merged acg because the contamination of the merged acg isn't a good measure of merge quality for unit pairs with unbalanced numbers of spikes

alejoe91:
Closing this since SLAy have been updated

Issue #425, opened 2022-01-02
making a recording json serializable?


khl02007:
@alejoe91 this is related to the old spikeextractors, sorry if this isn't the place to post. basically I'm trying to load an old recording extractor, convert it to a new recording (si >0.90), and then save it using the new API. Please see the example and error message below (of course in this case I should just open the file with the new API; my real problem is that some of the other stuff that we use rely on a custom extractor based on the older spikeextractors, which gives the same error upon `save` or `extract_waveforms`). how would I make this json serializable? I assume this means storing all the metadata in json format?
```
import spikeextractors
import spikeinterface as si
old_r = spikeextractors.BinDatRecordingExtractor(file_path='./r/traces_cached_seg0.raw',
                                                 sampling_frequency=30000,
                                                 numchan=4,
                                                 dtype='int16')
new_r = si.create_recording_from_old_extractor(old_r)
new_r.save(folder='r2', n_jobs=5, total_memory='5G')
```

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/tmp/ipykernel_2784207/555477451.py in <module>
      6                                                  dtype='int16')
      7 new_r = si.create_recording_from_old_extractor(old_r)
----> 8 new_r.save(folder='r2', n_jobs=5, total_memory='5G')

~/repos/spikeinterface/spikeinterface/core/base.py in save(self, **kwargs)
    517             return self.save_to_memory(**kwargs)
    518         else:
--> 519             return self.save_to_folder(**kwargs)
    520 
    521     def save_to_memory(self, **kwargs):

~/repos/spikeinterface/spikeinterface/core/base.py in save_to_folder(self, name, folder, dump_ext, verbose, **save_kwargs)
    581         provenance_file = folder / f'provenance.{dump_ext}'
    582         if self.check_if_dumpable():
--> 583             self.dump(provenance_file)
    584         else:
    585             provenance_file.write_text(

~/repos/spikeinterface/spikeinterface/core/base.py in dump(self, file_path, relative_to, folder_metadata)
    394         """
    395         if str(file_path).endswith('.json'):
--> 396             self.dump_to_json(file_path, relative_to=relative_to, folder_metadata=folder_metadata)
    397         elif str(file_path).endswith('.pkl') or str(file_path).endswith('.pickle'):
    398             self.dump_to_pickle(file_path, relative_to=relative_to, folder_metadata=folder_metadata)

[...]

Issue #420, opened 2021-12-18
Kilosort3 errors on tutorial dataset


jjtukker:
hi,
I'm sorry for all the issues, but I just keep hitting a wall trying out different sorters from within the Jupyter notebook tutorial (Python 3.8), on Ubuntu server (xfce4.14), with the standard dataset provided. So far only Mountainsort4 works..

For Kilosort3, here is the command I run:
sorting_KS3 = ss.run_kilosort3(recording_saved, 
                              output_folder='results_KS3',
                              verbose=True)

Here is the output:
write_binary_recording: 100%|#####################| 1/1 [00:07<00:00,  7.13s/it]

KilosortBase._run_from_folder <class 'spikeinterface.sorters.kilosort3.kilosort3.Kilosort3Sorter'>
RUNNING SHELL SCRIPT: results_KS3/run_kilosort3.sh

                            < M A T L A B (R) >

                  Copyright 1984-2021 The MathWorks, Inc.

             R2021a Update 1 (9.10.0.1649659) 64-bit (glnxa64)

                               April 13, 2021

                            < M A T L A B (R) >

                  Copyright 1984-2021 The MathWorks, Inc.

             R2021a Update 1 (9.10.0.1649659) 64-bit (glnxa64)

                               April 13, 2021

 

 

To get started, type doc.

To get started, type doc.

For product information, visit www.mathworks.com.

For product information, visit www.mathworks.com.

 

 

    ----------------------------------------------------

	Your license will expire in 12 days.

	Please contact your system administrator or

	MathWorks to renew this license.

    ----------------------------------------------------

    ----------------------------------------------------

	Your license will expire in 12 days.

	Please contact your system administrator or

	MathWorks to renew this license.

    ----------------------------------------------------

welcome!

welcome!

this works it is a unix machine

this works it is a unix machine

Warning: Function isequaln has the same name as a MATLAB builtin. We suggest

you rename the function to avoid a potential name conflict. 

> In path (line 109)

In addpath (line 86)

In startup (line 36) 

Warning: Function isequaln has the same name as a MATLAB builtin. We suggest

you rename the function to avoid a potential name conflict. 

> In path (line 109)

In addpath (line 86)

In startup (line 36) 

Warning: Function contains has the same name as a MATLAB builtin. We suggest

you rename the function to avoid a potential name conflict. 

> In path (line 109)

In addpath (line 86)

In startup (line 36) 

[...]

PR #4531, opened 2026-04-19
 Attach `ProbeGroup` directly to recordings

h-mayorquin:
This needs [probeinterface #425](https://github.com/SpikeInterface/probeinterface/pull/425).

This PR migrates SpikeInterface's recording layer from the flat `contact_vector` structured array stored in `_properties` to a `ProbeGroup` object held directly on the recording. `BaseRecordingSnippets` grows a `_probegroup` attribute populated by `set_probe` / `set_probegroup` and propagated through every probe-retaining operation. `get_channel_locations` reads positions from `probegroup._build_contact_vector()` (the narrow handle from #425).

[SpikeInterface #4465](https://github.com/SpikeInterface/spikeinterface/pull/4465) was developed assuming the fuller array-backed direction in probeinterface ([probeinterface #420](https://github.com/SpikeInterface/probeinterface/pull/420)). This PR takes the same recording-side migration but pairs with the narrower [probeinterface #425](https://github.com/SpikeInterface/probeinterface/pull/425). The SI-side code ends up roughly the same shape either way; what changes is how much we ask probeinterface to absorb.

Three technical points:

- **`ProbeGroup` becomes the canonical attached object.** The legacy `contact_vector` property is gone from the steady-state representation. All probe-retaining operations (`set_probe`, `ChannelSliceRecording`, `ChannelsAggregationRecording`, `motion_interpolation` `remove_channels`, dict-load) route through the canonical `_set_probes` path, so `location` and `group` derive from the attached probegroup consistently rather than being maintained separately across code paths.

- **Probe serialization stays in probeinterface.** The recording's `to_dict()` delegates to `ProbeGroup.to_dict()`, and `_load_extractor_from_dict` restores via `ProbeGroup.from_dict(...)`. There is no recording-side format for probe state; the serialized representation matches the in-memory object. Derived properties like `location` and `group` are still written on the recording for backward compatibility and to keep the diff focused, but they are treated as cached views of the probegroup, not as the canonical representation.

- **Old recordings still load.** A `_restore_probegroup_from_legacy_contact_vector` helper migrates the legacy `contact_vector` property plus associated annotations (`probes_info`, planar contours) into the new representation on load, then drops the legacy fields. `biocam` and `maxwell` extractors (which previously read `contact_vector["row"]` / `["col"]` / `["electrode"]` directly) now read from `probe.contact_annotations`.

### DIFF
--- a/.github/actions/build-test-environment/action.yml
+++ b/.github/actions/build-test-environment/action.yml
@@ -8,13 +8,13 @@ runs:
       run: |
         git config --global user.email "CI@example.com"
         git config --global user.name "CI Almighty"
-        pip install tabulate  # This produces summaries at the end
-        pip install -e .[test,extractors,streaming_extractors,test_extractors,full]
+        uv pip install --system tabulate  # This produces summaries at the end
+        uv pip install --system -e .[test,extractors,streaming_extractors,test_extractors,full]
       shell: bash
     - name: Install git-annex
       shell: bash
       run: |
-        pip install datalad-installer
+        uv pip install --system datalad-installer
         datalad-installer --sudo ok git-annex --method datalad/packages
         git config --global filter.annex.process "git-annex filter-process"  # recommended for efficiency
     - name: Force installation of latest dev from key-packages when running dev (not release)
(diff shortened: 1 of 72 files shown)


h-mayorquin:
Superseded by @alejoe91  #4465, which lands the ProbeGroup object directly on the recording (the _probegroup attribute) along with select_channels_with_probegroup. This PR's goal is fully covered there, so closing.

PR #4472, opened 2026-03-26
Implement `ChunkableMixin`

alejoe91:
Rebuilding https://github.com/SpikeInterface/spikeinterface/pull/4216 on top of https://github.com/SpikeInterface/spikeinterface/pull/4462

This PR introduces a `ChunkableMixin` class (for extractors) and a `ChunkableSegment`. These can be used as base classes for any class that can be chunked in the first dimension (e.g., `BaseRecording` - `BaseImaging`). 

This is done by "abstracting" the `get_traces` into a more general `get_data`, and implementing other key methods as: 
- `get_sample_size_in_bytes`
- `get_memory_size`
- `get_shape`

The `ChunkableSegment` is also in charge of all the low level time handling.

This large refactor allows to centralize the following "machinery" (in `chunkable_tools` and `job_tools`)
- `write_binary`/`write_memory`/`add_to_zarr`
- parallel execution, with `ChunkExecutor` (previous `ChunkRecordingExecutor`)
- pipeline node and `run_node_pipeline`

**NOTE**: Created as draft to show diff before https://github.com/SpikeInterface/spikeinterface/pull/4462 is merged

PR #4533, opened 2026-04-20
Improvements over chuking refactoring

h-mayorquin:
This PR fixes the Kilosort4 regression reported on #4472. The root cause is that #4472 turned `write_binary_recording` into a plain alias for `write_binary(chunkable, ...)`, so passing `recording=rec` as a keyword argument no longer matched the parameter name and raised `TypeError`. The immediate fix is to make `write_binary_recording` a real wrapper again, with `recording` as its actual parameter:

```python
def write_binary_recording(recording, file_paths, **kwargs):
    return _write_binary(recording, file_paths, **kwargs)
```

I think this points to a more general principle worth keeping in the codebase: the SpikeInterface public API should stay in SpikeInterface vocabulary. A user saving a recording should see `recording` in signatures, docstrings, and IDE autocomplete, not the implementation concept `chunkable`. Keeping the generic functions private (underscore-prefixed, living in `time_series_tools`) and exposing thin recording-centered wrappers as the public surface is what makes this class of regression structurally hard to reintroduce: the parameter name a user types is the one that lives in the signature, and cannot drift when the internals are refactored. `write_memory_recording` is converted the same way for consistency.

The second change is the naming of the abstraction introduced in #4472. We touched on this in the maintenance meeting with @alejoe91: the real abstraction here is not chunkable in general, it is a time series over which we run processes in chunks. My proposal is that the class names should reflect that rather than staying at the overly abstract `Chunkable` level. Concretely:

- `ChunkableMixin` is renamed to `TimeSeries`: what this class actually models is a time series. Slicing methods like `get_data(start_frame, end_frame, ...)` are a natural property of a sliceable time series, not an extra "chunkable" capability. The executor does the chunking; the class just has to expose a time-series contract.
- `ChunkableSegment` is renamed to `TimeSeriesSegment`: the segment class contains no chunking code at all. It is only time handling (`sampling_frequency`, `t_start`, `time_vector`, `get_times`, `sample_index_to_time`, `time_to_sample_index`), and its own docstring describes it as providing "methods to handle time kwargs." Dropping `Chunkable` matches what the class actually is.
- `ChunkExecutor` is renamed to `TimeSeriesChunkExecutor`: this is where the verb "chunk" belongs, because the executor is what performs the chunking. Naming it `TimeSeriesChunkExecutor` also makes the axis of chunking explicit (it chunks along time, not arbitrary axes) and ties the vocabulary together: a `TimeSeries` / `TimeSeriesSegment` is chunked by a `TimeSeriesChunkExecutor`. The module `chunkable_tools.py` and the helper `divide_chunkable_into_chunks` are renamed to `time_series_tools.py` and `divide_time_series_into_chunks` to keep the import path and internal utilities consistent with the new names.

The two ideas come together in the zarr writer. Before this PR the public function was literally named `write_chunkable_to_zarr`: the function name itself asked a SpikeInterface user to learn the internal concept "chunkable" just to save a recording. Now the public entry point is `write_recording_to_zarr(recording, ...)` and the generic implementation has moved to a private `_write_time_series_to_zarr(...)`, so the helper name is honest about the `TimeSeries` contract it operates on. The behavior is identical; only the surface a user sees changes. When `BaseImaging` lands, `write_imaging_to_zarr(imaging, ...)` will be a second thin wrapper over the same private core, and imaging users will never see `chunkable` either.

### DIFF
--- a/src/spikeinterface/core/node_pipeline.py
+++ b/src/spikeinterface/core/node_pipeline.py
@@ -11,9 +11,9 @@ from pathlib import Path
 import numpy as np
 
 from spikeinterface.core.base import base_peak_dtype, spike_peak_dtype
-from spikeinterface.core.chunkable import ChunkableMixin
+from spikeinterface.core.time_series import TimeSeries
 from spikeinterface.core import BaseRecording, get_chunk_with_margin
-from spikeinterface.core.job_tools import ChunkExecutor, fix_job_kwargs, _shared_job_kwargs_doc
+from spikeinterface.core.job_tools import TimeSeriesChunkExecutor, fix_job_kwargs, _shared_job_kwargs_doc
 from spikeinterface.core import get_channel_distances
 from spikeinterface.core.core_tools import ms_to_samples
 
@@ -26,7 +26,7 @@ class PipelineNode:
 
     def __init__(
         self,
-        chunkable: ChunkableMixin,
+        chunkable: TimeSeries,
         return_output: bool | tuple[bool] = True,
         parents: list[Type["PipelineNode"]] | None = None,
     ):
@@ -38,7 +38,7 @@ class PipelineNode:
 
         Parameters
         ----------
-        chunkable : ChunkableMixin
+        chunkable : TimeSeries
             The chunkable object.
         return_output : bool or tuple[bool], default: True
             Whether or not the output of the node is returned by the pipeline.
@@ -526,7 +526,7 @@ def check_graph(nodes, check_for_peak_source=True):
 
 
 def run_node_pipeline(
-    chunkable: ChunkableMixin,
+    chunkable: TimeSeries,
     nodes: list[PipelineNode],
     job_kwargs: dict,
     job_name: str = "pipeline",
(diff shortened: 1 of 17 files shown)


src/spikeinterface/core/node_pipeline.py
```diff
@@ -26,7 +26,7 @@ class PipelineNode:
 
     def __init__(
         self,
-        chunkable: ChunkableMixin,
+        chunkable: TimeSeries,
```
alejoe91:
```suggestion
        time_series: TimeSeries,
```
?

alejoe91:
I agree `Chunkable` is too abstract! I like `TimeSeries` better. @samuelgarcia @chrishalcrow do you like the name? 

@h-mayorquin should we also rename then the `chunkable` arguments to `time_series` in all the functions? I think we should right?

h-mayorquin:
> I agree `Chunkable` is too abstract! I like `TimeSeries` better. @samuelgarcia @chrishalcrow do you like the name?
> 
> @h-mayorquin should we also rename then the `chunkable` arguments to `time_series` in all the functions? I think we should right?

Yes, I agree.

src/spikeinterface/core/node_pipeline.py
h-mayorquin: Done.

chrishalcrow:
In case this takes a some time to merge, I've done a quick dirty patch to fix the kilosort issue here #4540

alejoe91 (approved):

PR #4539, opened 2026-04-20
Extend artifact detection: `DetectAndRemoveArtifacts` preprocessor and signed saturation

alejoe91:
This PR introduces a `DetectAndRemoveArtifactsRecording` class so that it can be used in the preprocessing pipeline.
Users can pass a `recording_to_detect`: in this case, saturation/artifacts are detected on that recording (e.g., raw) and applied to the preprocessor parent (preprocessed)

Also extended saturation to optionally provide the "sign" of the saturation events. This can be useful to further inspect saturation issues and track down experimental causes to the saturation.

EDIT (Chris): new preprocessing pipeline docs: https://spikeinterface--4539.org.readthedocs.build/en/4539/modules/preprocessing.html#the-preprocessing-pipeline


src/spikeinterface/preprocessing/detect_artifacts.py
```diff
@@ -165,40 +198,71 @@ def compute(
...
+                for i, (start, stop) in enumerate(zip(intervals[::2], intervals[1::2])):
+                    events[i]["start_sample_index"] = start + start_frame
+                    events[i]["end_sample_index"] = stop + start_frame
+                    events[i]["segment_index"] = segment_index
+                all_events.append(events)
+            all_events = np.concatenate(all_events)
+            # sort by start sample index
+            order = np.argsort(all_events["start_sample_index"])
+            all_events = all_events[order]
+            return (all_events,)
```
alejoe91:
```suggestion
            events = all_events[order]
```

chrishalcrow:
Hello, this is great. Part of (my) pipeline philosophy is that you can load a pipeline from a dumped json/pickle:

https://github.com/SpikeInterface/spikeinterface/blob/a3f8db3b173cb0c7814cf018fbe89b2ebc941618/src/spikeinterface/preprocessing/pipeline.py#L252

I don't know how this will work when one argument is a recording object? Might just work?

And I also like the idea that you can specify a pipeline externally without reference to the recording, which this breaks. But more philosophical than practical... I can't really see a way around this.

alejoe91:
> Hello, this is great. Part of (my) pipeline philosophy is that you can load a pipeline from a dumped json/pickle:
> 
> https://github.com/SpikeInterface/spikeinterface/blob/a3f8db3b173cb0c7814cf018fbe89b2ebc941618/src/spikeinterface/preprocessing/pipeline.py#L252
> 
> I don't know how this will work when one argument is a recording object? Might just work?
> 
> And I also like the idea that you can specify a pipeline externally without reference to the recording, which this breaks. But more philosophical than practical... I can't really see a way around this.

I have a suggestion. We can have some "magic" string: 
- `"pipeline['raw']"`: tells the function to use the input recording
- `"pipeline['{preprocessing_step}']`: grabs recording at a specific step.

What do you think?

oliche (commented):
It's good to go, maybe we could state that this apodization is a number of samples ? 

Thanks a lot for looking at this !

alejoe91:
> It's good to go, maybe we could state that this apodization is a number of samples ?
> 
> Thanks a lot for looking at this !

Renamed `apodization_samples` ;)

src/spikeinterface/preprocessing/detect_artifacts.py
```diff
@@ -600,3 +666,118 @@ def detect_artifact_periods(
...
+            Random seed for `get_noise_levels` and `NoiseGeneratorRecording`.
+        artifact_periods : np.ndarray | None, default: None
+            Optionally, pre-computed artifact periods can be passed directly to the constructor to skip the
+            detection step. If ``None``, artifact periods are detected on the fly using the specified method
+    """
+
+    def __init__(
+        self,
+        recording: BaseRecording,
+        recording_to_detect: BaseRecording | None = None,
```
chrishalcrow, line 737: So the idea of the magic string is to allow this to be

```suggestion
        recording_to_detect: BaseRecording | str | None = None,
```

?
So that a user could pass a preprocessing pipeline
```
{
    'common_reference': {},
    'bandpass_filter': {},
    'detect_and_remove_artifacts': {'recording_to_detect': 'raw'}
}
```

or even

```
{
    'phase_shift': {},
    'common_reference': {},
    'bandpass_filter': {},
    'detect_and_remove_artifacts': {'recording_to_detect': 'phase_shift'}
}
```

When I type it out like this, it looks reasonable. But it only works if you're using a pipeline, since:

```
phase_shift_recording = si.phase_shift(raw)
remove_artifacts_recording = si.dectect_and_remove_artifacts(phase_shift_recording, recording_to_detect = 'phase_shift')
```
makes no sense.

Maybe we could add a private kwarg `_using_pipeline`, and you can only use a string if this is set to true? Or maybe the Pipeline object can be responsible for converting the string into a recording. 

Or maybe right now we just add the "raw" option?

src/spikeinterface/preprocessing/silence_periods.py
```diff
@@ -154,12 +172,13 @@ def _check_periods(periods, num_seg):
...
+    def __init__(self, parent_recording_segment, periods, mode, noise_generator, seg_index, apodization_samples=7):
         BasePreprocessorSegment.__init__(self, parent_recording_segment)
         self.periods = periods
         self.mode = mode
         self.seg_index = seg_index
         self.noise_generator = noise_generator
+        self.apodization_samples = apodization_samples
 
     def get_traces(self, start_frame, end_frame, channel_indices):
         traces = self.parent_recording_segment.get_traces(start_frame, end_frame, channel_indices)
```
alejoe91: When it's apodization, we should get traces with margin because the events could be on the edge. In that casem different start/end frames could give different results if we don't take margins into account

alejoe91: @oliche do you agree?

alejoe91: @samuelgarcia I implemented the margin logic and added a proper test. For zeros and noise, `margin` is set to 0, so the behavior should be the same

src/spikeinterface/preprocessing/tests/test_pipeline.py
```diff
@@ -212,6 +214,61 @@ def test_loading_from_analyzer(create_cache_folder):
...
+    # Test dumping the pipeline to pickle and loading it back with the correct substitution still works
+    pp_rec_from_pipeline.dump_to_pickle(create_cache_folder / "pipeline_substitution_test.pkl")
+    pp_rec_from_pkl = get_preprocessing_dict_from_file(create_cache_folder / "pipeline_substitution_test.pkl")
+    pp_rec_from_pipeline_substitution = apply_preprocessing_pipeline(
+        rec, pp_rec_from_pkl, apply_precomputed_kwargs=True
+    )
+    assert isinstance(pp_rec_from_pipeline_substitution._kwargs["recording_to_detect"], BandpassFilterRecording)
+    assert isinstance(pp_rec_from_pipeline_substitution._kwargs["recording"], CommonReferenceRecording)
+    assert isinstance(pp_rec_from_pipeline_substitution, DetectAndRemoveArtifactsRecording)
+
```
chrishalcrow, line 271: Could we do a quick check that `pp_rec_from_pipeline_substitution` and `pp_rec_from_pipeline` are the same?

chrishalcrow (approved):
I've only read the preprocessing pipeline stuff - honestly so happy with the solution!
Please don't merge until you've got an approval on the other stuff!

src/spikeinterface/preprocessing/tests/test_pipeline.py
alejoe91: done!

alejoe91:
Also added automatic loading od saturation levels for SpikeGLX and OpenEphys, follwing https://github.com/SpikeInterface/probeinterface/pull/434

When available, this sets the `saturation_threshold_uV` property, so saturation detection is automatic!!!

PR #4541, opened 2026-04-20
Modularize CI test and runtime dependencies per module

### LINKED ISSUE #4535: CI test-extras modularization: where do we want to take this?
@alejoe91 @samuelgarcia, I am thinking of moving forward with modularized extras per module in the tests (following [this TODO](https://github.com/SpikeInterface/spikeinterface/blob/10834d4db1335580707bc5a787918de1018d8a68/.github/workflows/all-tests.yml#L168)). I have a vague memory that you two decided you wanted to move in another direction. Where do you stand on the topic now?

h-mayorquin:
After discussing and confirming with Alessio in #4535, I am finishing the package modularization we started a while ago in #3084. Concretely:

- Moved test-only dependencies out of `[project.optional-dependencies]` into [PEP 735](https://peps.python.org/pep-0735/) `[dependency-groups]`, with one group per module (plus a shared `test-common` and an aggregate `test-all` for the nightly codecov job).
- Added per-module [runtime extras](https://packaging.python.org/en/latest/specifications/pyproject-toml/#dependencies-optional-dependencies) that were missing: `postprocessing`, `comparison`, `generation`, `sorters`, `sortingcomponents`, `sorters_internal`, `curation`, `exporters`.
- Redefined `full` as the union of all per-module extras via self-references, so adding a dep to a module propagates to `full` automatically.
- Each CI test step now installs only its module's runtime extra plus the matching test group (for example, `.[postprocessing] --group test-postprocessing`) instead of the old `.[full]` plus catch-all `.[test]`, so version drift can be pinpointed to a single module.

Close #4535

### DIFF
--- a/.github/actions/build-test-environment/action.yml
+++ b/.github/actions/build-test-environment/action.yml
@@ -9,7 +9,7 @@ runs:
         git config --global user.email "CI@example.com"
         git config --global user.name "CI Almighty"
         uv pip install --system tabulate  # This produces summaries at the end
-        uv pip install --system -e .[test,extractors,streaming_extractors,test_extractors,full]
+        uv pip install --system -e .[extractors,streaming_extractors,full] --group test-all
       shell: bash
     - name: Install git-annex
       shell: bash
--- a/.github/workflows/all-tests.yml
+++ b/.github/workflows/all-tests.yml
@@ -87,7 +87,7 @@ jobs:
 
       - name: Install packages
         run: |
-          uv pip install --system -e .[test_core]
+          uv pip install --system -e . --group test-core
         shell: bash
 
       - name: Pip list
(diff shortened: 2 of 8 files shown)


alejoe91:
Absolutely brillant!!!

alejoe91 (approved):

chrishalcrow:
Amazing! Could you add some notes to the Development docs for running groups locally?

Using `uv`, testing different modules looks like this:

```
uv run --group test-preprocessing pytest -s src/spikeinterface/preprocessing/tests
```

So nice!!

h-mayorquin:
@chrishalcrow I added some more docs, tell me if this is what you had in mind.

chrishalcrow (approved):
Perfect-o

PR #4542, opened 2026-04-21
Segment backward compatibility

alejoe91:
This PR re-adds `_recording_segments` and `_sorting_segments` for backward compatibility


src/spikeinterface/core/basesorting.py
```diff
@@ -65,6 +65,11 @@ def segments(self) -> list["BaseSortingSegment"]:
         """List of sorting segments."""
         return self._segments
 
+    @property
+    def _sorting_segments(self) -> list["BaseSortingSegment"]:
+        """For backward compatibility, we keep _sorting_segments."""
```
chrishalcrow, line 70:
```suggestion
    # For backward compatibility, we keep _sorting_segments
    @property
    def _sorting_segments(self) -> list["BaseSortingSegment"]:
        """List of sorting segments."""
```

Just for consistency with `_recording_segments`

alejoe91: I did the other way around (forgot to save the file before committing)

chrishalcrow (approved):

PR #4543, opened 2026-04-21
Fix filter kwargs

alejoe91:
Add missing `ignore_low_freq_error` to highpass filter kwargs


src/spikeinterface/preprocessing/filter.py
```diff
@@ -344,7 +344,13 @@ def __init__(
             self, recording, band=freq_min, margin_ms=margin_ms, dtype=dtype, btype="highpass", **filter_kwargs
         )
         dtype = fix_dtype(recording, dtype)
-        self._kwargs = dict(recording=recording, freq_min=freq_min, margin_ms=margin_ms, dtype=dtype.str)
+        self._kwargs = dict(
+            recording=recording,
+            freq_min=freq_min,
+            margin_ms=margin_ms,
+            dtype=dtype.str,
+            ignore_low_freq_error=ignore_low_freq_error,
```
chrishalcrow, line 352: Do we want to add `_skip_margin_warning_for_old_version` too?

alejoe91: That should be set automatically when it's loaded back, if needed.

chrishalcrow (approved):

PR #4544, opened 2026-04-21
Add `sort_by_depth` kwargs to `RasterWidget`

tayheau:


src/spikeinterface/widgets/rasters.py
```diff
@@ -394,6 +397,14 @@ def __init__(
         if unit_ids is None:
             unit_ids = sorting.unit_ids
 
+        if sort_by_depth:
+            # print("hey")
```
chrishalcrow: Hey

src/spikeinterface/widgets/rasters.py
```diff
@@ -363,6 +363,8 @@ class RasterWidget(BaseRasterWidget):
         A sorting object. Deprecated.
     sorting_analyzer : SortingAnalyzer  | None, default: None
         A sorting analyzer object. Deprecated.
+    sort_by_depth: bool = False
+        Wether or not to sort units by depth, default: False
```
chrishalcrow:
```suggestion
    sort_by_depth : bool, default: False
        Whether or not to sort units by depth, default: False
```
We (try to) keep to numpydoc conventions: https://spikeinterface.readthedocs.io/en/stable/development/development.html#docstring-conventions

src/spikeinterface/widgets/rasters.py
tayheau: 0_o

tayheau:
just quickly changing the logic to add a widget checkbox for depth sorting

src/spikeinterface/widgets/rasters.py
```diff
@@ -394,6 +397,13 @@ def __init__(
         if unit_ids is None:
             unit_ids = sorting.unit_ids
 
+        if sort_by_depth:
+            if not sorting_analyzer_or_sorting.has_extension("unit_locations"):
```
chrishalcrow, line 441: `has_extension` will fail if it's a sorting. Maybe return an error if it's a sorting?

src/spikeinterface/widgets/rasters.py
```diff
@@ -394,6 +436,19 @@ def __init__(
         if unit_ids is None:
             unit_ids = sorting.unit_ids
 
+        if isinstance(sorting_analyzer_or_sorting, SortingAnalyzer):
+            if not sorting_analyzer_or_sorting.has_extension("unit_locations"):
```
chrishalcrow, line 441: In this branch, `depth_dict` is never set and you get
```
UnboundLocalError: cannot access local variable 'depth_dict' where it is not associated with a value
```
at line 493.

I reckon you should set `depth_dict = None` before the `if` statement, and delete the final `else` at ine 449.

chrishalcrow (approved):
Looks good, thanks!

src/spikeinterface/widgets/rasters.py
```diff
@@ -363,6 +402,8 @@ class RasterWidget(BaseRasterWidget):
         A sorting object. Deprecated.
     sorting_analyzer : SortingAnalyzer  | None, default: None
         A sorting analyzer object. Deprecated.
+    sort_by_depth : bool, default: False
+        Whether or not to sort units by depth, default: False
```
alejoe91:
```suggestion
        Whether or not to sort units by depth, only available when input is a `SortingAnalyzer`
```

alejoe91 (approved):

PR #4548, opened 2026-04-21
Probe refactoring with wiring

h-mayorquin:
More ambitious, more complete, greater, faster, stronger.

See: https://github.com/SpikeInterface/spikeinterface/issues/4553

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -1162,10 +1162,9 @@ def _load_extractor_from_dict(dic) -> "BaseExtractor":
         from probeinterface import ProbeGroup
 
         probegroup = ProbeGroup.from_dict(dic["probegroup"])
-        if hasattr(extractor, "set_probegroup"):
-            extractor.set_probegroup(probegroup, in_place=True)
-        else:
-            extractor._probegroup = probegroup
+        # The `wiring` per-channel property was restored above by the standard
+        # property-load loop; we just attach the probegroup object.
+        extractor._probegroup = probegroup
     elif "contact_vector" in dic.get("properties", {}):
         _restore_probegroup_from_legacy_contact_vector(extractor)
 
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -610,12 +610,11 @@ class BaseExtractor:
         return extractor
 
     def load_metadata_from_folder(self, folder_metadata):
-        # hack to load probe for recording
         folder_metadata = Path(folder_metadata)
 
-        self._extra_metadata_from_folder(folder_metadata)
-
-        # load properties
+        # load properties first so that `_extra_metadata_from_folder` can see
+        # restored state like the `wiring` property and skip re-running
+        # `set_probegroup` when the mapping is already in place.
         prop_folder = folder_metadata / "properties"
         if prop_folder.is_dir():
             for prop_file in prop_folder.iterdir():
@@ -623,6 +622,8 @@ class BaseExtractor:
                     values = np.load(prop_file, allow_pickle=True)
                     key = prop_file.stem
                     self.set_property(key, values)
+
+        self._extra_metadata_from_folder(folder_metadata)
 
     def save_metadata_to_folder(self, folder_metadata):
         self._extra_metadata_to_folder(folder_metadata)
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -1172,8 +1172,44 @@ def _load_extractor_from_dict(dic) -> "BaseExtractor":
             extractor.set_probegroup(probegroup, in_place=True)
         else:
             extractor._probegroup = probegroup
+    elif "contact_vector" in dic.get("properties", {}):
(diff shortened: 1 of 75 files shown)


h-mayorquin:
From the discussion today with @samuelgarcia 

`get_probe` should by default turn the Probe object already re-ordered by channels.

src/spikeinterface/core/base.py
```diff
@@ -1155,9 +1159,53 @@ def _load_extractor_from_dict(dic) -> "BaseExtractor":
...
 
+    if "probegroup" in dic:
+        from probeinterface import ProbeGroup
+
+        probegroup = ProbeGroup.from_dict(dic["probegroup"])
+        # The `wiring` per-channel property was restored above by the standard
+        # property-load loop; we just attach the probegroup object.
+        extractor._probegroup = probegroup
+    elif "contact_vector" in dic.get("properties", {}):
+        _restore_probegroup_from_legacy_contact_vector(extractor)
```
alejoe91, line 1170: this should not be in base, since it's specific to the `BaseRecordingSnippets`

src/spikeinterface/core/base.py
```diff
@@ -1155,9 +1159,53 @@ def _load_extractor_from_dict(dic) -> "BaseExtractor":
...
+
     return extractor
 
 
+def _restore_probegroup_from_legacy_contact_vector(extractor) -> None:
```
alejoe91, line 1175: same

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -90,155 +91,171 @@ def set_probegroup(self, probegroup, group_mode="auto", in_place=False):
...
+                    key.append(probe.shank_ids[contact_idx])
+                elif k == "side" and probe.contact_sides is not None:
+                    key.append(probe.contact_sides[contact_idx])
+            group_keys_per_channel.append(tuple(key))
+
+        unique_keys = list(dict.fromkeys(group_keys_per_channel))
+        key_to_int = {k: i for i, k in enumerate(unique_keys)}
+        groups = np.array([key_to_int[k] for k in group_keys_per_channel], dtype="int64")
+
+        target.set_property("location", locations)
```
alejoe91, line 256: As part of this effort, I think we should get rid of the `location` property and only keep back-compatibility to re-intantiate a probe object

src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -383,8 +408,34 @@ def set_channel_groups(self, groups, channel_ids=None):
         self.set_property("group", groups, ids=channel_ids)
 
     def get_channel_groups(self, channel_ids=None):
-        groups = self.get_property("group", ids=channel_ids)
-        return groups
+        # when a probe is attached, derive groups on the fly from the `wiring`
+        # property + probegroup state (probe_id + shank_ids + contact_sides)
```
alejoe91, line 412: Why do we need to do this on the fly? I think that the current implementation of setting the `group` property based on some group_mode (auto, by_probe, by_side, etc.) was explicit and clear. No?

src/spikeinterface/core/basesnippets.py
```diff
@@ -11,7 +11,7 @@ class BaseSnippets(BaseRecordingSnippets):
     Abstract class representing several multichannel snippets.
     """
 
-    _main_properties = ["group", "location", "gain_to_uV", "offset_to_uV"]
+    _main_properties = ["group", "location", "wiring", "gain_to_uV", "offset_to_uV"]
```
alejoe91, line 14:
```suggestion
    _main_properties = ["group", "wiring", "gain_to_uV", "offset_to_uV"]
```

samuelgarcia:
Lets discuss this more.
I am not sure to like the idea for the probegroup to be the original probegroup of the parent.

For me if we attach a probegroup with 4 probes to a recording and then we split it, I think we expect that each new sub recording has only one probe.

I also think that the get_probe must give a sorted version of the probe that match the channel order.

In short I am OK to have a property for the mapiing (probe_index, contact_id) <-> channel_id but `get_probe()` should give something simple that match the channel list.

In short we have 2 choice: 
  1. internally handle the original progegroup and render the real probe on the fly
  2. internally handle the real probegroup/probe directly and transform it on the way : slice/order/slice.

Since a recording can easily acces all parents, and so there related probegroup. I do not see any advantage of going to 1.. I really refer 2.

h-mayorquin:
@samuelgarcia if I understood correctly you are Ok with keeping the internal representation as a `ProbeGroup` but our disagreement where we disagree is on whether keep that internal representation should be sliced and re-order. I want to do that on the fly at user request you want to do that at initialization. 

The second point that you mentioned in our meeting today is the background compatibility but we did not finish that line of discussion because I needed to go. Can you expand and sketch a test of what I should do to get the error you were mentioning?

h-mayorquin:
Closing in favor of @alejoe91  #4465. 

Keeping the explicit wiring independently of re-ordering or trimming to only wired contacts can be its own stand-alone in the future if that is required.

PR #4567, opened 2026-04-27
Fix the IBL Cbin extractor so that it works with 4 split shanks as well

rai-pranav:
The .meta file that is produced when converting the Np2 probes to 4 individual shanks does not contain the `~` sign in front of the fileds which are lists, and hence the read_meta_file function from neo.rawio.spikeglxrawio fails. So I have written some functions which can handle this version of the meta file.


rai-pranav:
Hi @alejoe91 
I have created this PR to fix some IBL specific loading of the Np2 4 shanks dataset.
Some of the windows OS tests are failing which I don't think are related to the my changes. Any hints on how to resolve them?

chrishalcrow:
Hello, I think the test failure was related to a downstream cache test failure here: https://github.com/SpikeInterface/spikeinterface/actions/runs/24997078165
I reran the caching successfully here: https://github.com/SpikeInterface/spikeinterface/actions/runs/25054593823
Which seems to have fixed the git-annex install here.
I'm not sure how to make this more robust, but at least the tests are passing now!

rai-pranav:
I cannot directly assign reviewers to this PR.
Can someone please review this PR ? @chrishalcrow @alejoe91 🙏

chrishalcrow:
Hi @rai-pranav , I think that all the probe reading logic should be handled by ProbeInterface instead of SpikeInterface.

I don't quite understand the problem. Does SpikeGLX produce these `.meta` files, or do you make them yourself when you convert your 4-shank recording into 4 1-shank recordings? If SpikeGLX makes them, we should update this function 

https://github.com/SpikeInterface/probeinterface/blob/6c94c35dbdadc703ee38359fcd16d39bcf34aad6/src/probeinterface/neuropixels_tools.py#L875

to work with your data.

Could you share your `.meta` files? Would help me understand the issue better :)

rai-pranav:
Hi @chrishalcrow 
Thanks for your reply.

I have uploaded an example .meta file [here](https://drive.google.com/file/d/1iF485LcRIMpqEoFh6hFwLAnYTmFsxGlF/view?usp=sharing)

These meta files are produced when we convert the NP 2.4 probes to 4 1 shank probes. So I think this is very IBL specific, and that is why I was thinking including it inside cbin_ibl.py file would restrict it to the IBL data.

alejoe91:
I think it's fine to keep it here then, but let's rename the function to `_read_cbin_probe`

rai-pranav:
Hi @alejoe91 
I have renamed the function to `_read_cbin_probe`.
Let me know if I should make any other changes.

rai-pranav:
Hi @alejoe91 @chrishalcrow 
I am reviving this PR. 
It was breaking with the latest version of probeinterface. Right now I have get it to work for the probeinterface main branch. And added tests with an IBL specific meta file.

chrishalcrow (approved):
Thanks for editing this to match the probeinterface logic. It looks good. If it's working for your data @rai-pranav then I'm happy to merge.

alejoe91 (approved):

rai-pranav:
> Thanks for editing this to match the probeinterface logic. It looks good. If it's working for your data @rai-pranav then I'm happy to merge.

Yes, it is working for my data.

Thanks @alejoe91 @chrishalcrow for the review.

PR #4572, opened 2026-04-30
"Allow" for 3D channel locations for compute unit locs and template metrics

chrishalcrow:
This PR allows for unit locations and template metrics to be computed when you have 3D channel locations (as is often the case with openly available data on DANDI).

It doesn't actually do anything, except slice the 3D locations into 2D. When we do this for the template metrics computation, a warning comes up.

### DIFF
--- a/src/spikeinterface/postprocessing/tests/test_unit_locations.py
+++ b/src/spikeinterface/postprocessing/tests/test_unit_locations.py
@@ -1,6 +1,9 @@
 from spikeinterface.postprocessing.tests.common_extension_tests import AnalyzerExtensionCommonTestSuite
 from spikeinterface.postprocessing import ComputeUnitLocations
 import pytest
+from probeinterface import Probe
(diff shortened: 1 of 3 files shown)


src/spikeinterface/postprocessing/tests/test_unit_locations.py
```diff
@@ -18,3 +21,46 @@ class TestUnitLocationsExtension(AnalyzerExtensionCommonTestSuite):
     )
     def test_extension(self, params):
         self.run_extension_tests(ComputeUnitLocations, params=params)
+
+
+def test_2d_and_3d_unit_localisation():
```
alejoe91:
```suggestion
def test_2d_and_3d_unit_localization():
```

chrishalcrow: I almost got away with it...

alejoe91 (approved):

PR #4575, opened 2026-05-04
Add SelectSegmentEvent

tayheau:
add  `select_segment(segment_indices: int | list[int])` to `BaseEvent` through a new  `class SelectSegmentEvent(BaseEvent)`

recordings and sortings have it, so in continuity (and bc i need it), i think it can be a good small addition ;)

### DIFF
--- a/src/spikeinterface/core/baseevent.py
+++ b/src/spikeinterface/core/baseevent.py
@@ -66,6 +66,11 @@ class BaseEvent(BaseExtractor):
     def get_num_segments(self):
         return len(self._event_segments)
 
+    def select_segment(self, segment_indices: int | list[int]):
+        from .segmentutils import SelectSegmentEvent
+
+        return SelectSegmentEvent(self, segment_indices=segment_indices)
+
     def get_events(
         self,
         channel_id: int | str | None = None,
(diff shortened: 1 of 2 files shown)


h-mayorquin:
> add `select_segment(segment_indices: int | list[int])` to `BaseEvent` through a new `class SelectSegmentEvent(BaseEvent)`
> 
> recordings and sortings have it, so in continuity (and bc i need it), i think it can be a good small addition ;)



> add `select_segment(segment_indices: int | list[int])` to `BaseEvent` through a new `class SelectSegmentEvent(BaseEvent)`
> 
> recordings and sortings have it, so in continuity (and bc i need it), i think it can be a good small addition ;)

Can you describe a little about your use case?

tayheau:
>Can you describe a little about your use case?

Mostly the analysis of my results, i work per segment due to the nature of the stimulus set.  Tbh it works also by using the select since the concept of segment exist in `BaseEvent`, i feel it makes the api more uniform to also have `select_segement` here. If you feel that it's out of scope, we can drop it !

alejoe91 (approved):

alejoe91:
LGTM! I agree this unifies the API with the recording/sorting and makes the `Event` c;ass more complete! Thanks @tayheau

h-mayorquin:
No, I think is totally within scope. I am just curious about what kind of setup you have and what formats you work in. I think it is good to leave this kind of thing documented at least on github issues so we can come back and see why we took the decisions we took.

PR #4579, opened 2026-05-13
Implement cache spike vector for phy sorting reader

samuelgarcia:
Try to implement the spike vector caching for phy reader in a faster way than #4502

@alejoe91 @grahamfindlay : the implementaion of the caching was slow mainly dur the final lexsort which is normaly not need.

Perfs for the file provide by Graham: (400Mspikes 342 units)

read_phy() + sorting.to_spike_vector() + sorting.to_reordered_spike_vector()

 * main :   19s + 260s + 125s
 * PR  4502 : 16s + 212s (39 + lexsort 200) + 125s
 * this PR : 12s + 12s + 125s

I will work in a sperarte PR to improve the select_unit() and caching.

Important note for us: we should reread all the sorting reader and make the same trick when it is necessary.

### DIFF
--- a/src/spikeinterface/extractors/phykilosortextractors.py
+++ b/src/spikeinterface/extractors/phykilosortextractors.py
@@ -149,7 +149,7 @@ class BasePhyKilosortSortingExtractor(BaseSorting):
             del cluster_info["id"]
 
         if remove_empty_units:
-            cluster_info = cluster_info.query(f"cluster_id in {unique_unit_ids}")
+            cluster_info = cluster_info.query(f"cluster_id in {list(unique_unit_ids)}")
 
         # update spike clusters and times values
         bad_clusters = [clust for clust in unique_unit_ids if clust not in cluster_info["cluster_id"].values]
--- a/src/spikeinterface/extractors/phykilosortextractors.py
+++ b/src/spikeinterface/extractors/phykilosortextractors.py
@@ -229,19 +229,19 @@ class BasePhyKilosortSortingExtractor(BaseSorting):
         self.annotate(phy_folder=str(phy_folder.resolve()))
 
         self.add_sorting_segment(PhySortingSegment(spike_times_clean, spike_clusters_clean))
-    
+
     def _compute_and_cache_spike_vector(self) -> None:
         # make the spike_vector fast using the internal spike_times/spike_clusters
         # with a small mapping id to index
         # the order for 2 units with the same sample_index is not garanty here but should be OK
-        
+
         unit_ids = self.unit_ids
 
         # mapping unit_id to unit_index
-        mapping = -np.ones(np.max(unit_ids) +1, dtype="int64")
+        mapping = -np.ones(np.max(unit_ids) + 1, dtype="int64")
         for unit_ind, unit_id in enumerate(unit_ids):
             mapping[unit_id] = unit_ind
-        
+
         spike_times = self.segments[0]._all_spike_times
         spike_clusters = self.segments[0]._all_clusters
         n = spike_times.size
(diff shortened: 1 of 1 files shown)


alejoe91:
This LGTM! @grahamfindlay can you take a look?

grahamfindlay:
Eh, sorry I haven't had time to pick #4502 back up. 

As I understand it, this makes 2 changes:
1. `__init__()` doesn't waste time trying to remove spikes and unit ids from "bad clusters" from the full flat (`npy`) arrays on load (`read_phy()`), if there aren't any bad clusters to begin with. This accounts for the small (~4s) improvement in `read_phy()` time. 
2. Overrides the parent class `_compute_and_cache_spike_vector()`. The new one just never lexsorts. This accounts for the large improvement in `to_spike_vector()`. The result is 12s to build the spike vector. Nice!

My only real concern would be this, in the new `_compute_and_cache_spike_vector()`:
```python
# the order for 2 units with the same sample_index is not garanty here but should be OK
```

If it's okay to not sort the co-temporal spikes, why were we doing it in the first place? Should we also remove this guarantee from the parent class's implementation? 

In my [last comment](https://github.com/SpikeInterface/spikeinterface/pull/4502#issuecomment-4272459614) on #4502 I mentioned that I had found a way to do the lexsorting much more quickly than `np.lexsort()`. That also built the spike vector in ~12s, but also maintained the sortedness guarantee, and also didn't require an override of the parent method (i.e., should work for all sorters). Then I... never committed it 🫣. But I have it stored on a local branch somewhere, I can dig it up, if you want to see it. 

In any case, I don't see any reason that should block this. This looks good and is a big immediate win (assuming it really is okay to break the spike vector sortedness promise)!

samuelgarcia:
> 1. `__init__()` doesn't waste time trying to remove spikes and unit ids from "bad clusters" from the full flat (`npy`) arrays on load (`read_phy()`), if there aren't any bad clusters to begin with. This accounts for the small (~4s) improvement in `read_phy()` time.

Yep.

> 2. Overrides the parent class `_compute_and_cache_spike_vector()`. The new one just never lexsorts. This accounts for the large improvement in `to_spike_vector()`. The result is 12s to build the spike vector. Nice!
> 
> My only real concern would be this, in the new `_compute_and_cache_spike_vector()`:
> 
> ```python
> # the order for 2 units with the same sample_index is not garanty here but should be OK
> ```

This is the main point we need to discuss. Normally it should be OK that the spike vector do not garanty the order aas soon as the order is always the same.
The spike amplitudes, amplitudes scaling, spike location will be relative to thios vector!


> If it's okay to not sort the co-temporal spikes, why were we doing it in the first place? Should we also remove this guarantee from the parent class's implementation?

Maybe we haven't discuss it enough.

> In my [last comment](https://github.com/SpikeInterface/spikeinterface/pull/4502#issuecomment-4272459614) on #4502 I mentioned that I had found a way to do the lexsorting much more quickly than `np.lexsort()`. That also built the spike vector in ~12s, but also maintained the sortedness guarantee, and also didn't require an override of the parent method (i.e., should work for all sorters). Then I... never committed it 🫣. But I have it stored on a local branch somewhere, I can dig it up, if you want to see it.

I don't see this piece of code in the PR and I would be happy to see it.

grahamfindlay:
@samuelgarcia @alejoe91 

Check out the [latest commit](https://github.com/alejoe91/spikeinterface/pull/29/changes/c495e3ae231d6235ea07940307953b7da8e08545) from [this PR](https://github.com/alejoe91/spikeinterface/pull/29).
(I can't update #4502 directly, so I go through Alessio's fork. @alejoe91 can you merge into #4502 but leave open so I can keep pushing changes, please?). 

I was wrong about being able to optimize the base/parent class `_compute_and_cache_spike_vector()` -- this is also an optimization of the Phy/KiloSort implementation. BUT it runs in ~7s on the real-world data (392M spikes, 384 units - the same dataset I sent you) and maintains the full lexsortedness guarantee. 

`.to_spike_vector()` on the `UnitSelectionSorting` still sucks, as does `to_reordered_spike_vector()`. 

Full disclosure: I wrote the code, Claude wrote the tests (I specified and/or read them all).

grahamfindlay:
I've now optimized `UnitSelectionSorting._compute_and_cache_spike_vector()` using essentially the same trick, to address #4581 . 

Performing a two-layered selection where
* Selection 1 picks 258 units from 342 (order preserving, no renaming)
* Selection 2 is "no-op", picking the same 258 units (order preserving, no renaming)

now takes ~3-6s, down from 21s in #4502 , down from 4m30s pre-PR. 

Pushed to the PR on Alessio's fork, so you can check it out and merge into #4502 .

grahamfindlay:
Optimized `to_reordered_spike_vector()` too, in that same PR. Again milking the same trick, plus replacing the full numpy lexsort with a counting sort. Starting from a canonical spike vector, now takes ~14s for the parent, ~4s for the first-layer selection, free for the second-layer selection.

alejoe91 (approved):

PR #4586, opened 2026-05-18
Simulate temporally-correlated noise (V2)

### LINKED ISSUE #4482: Temporally correlated noise in simulations

### LINKED ISSUE #4522: Remove the strategy argument from the noise generators
The current noise generation machinery (both `NoiseGeneratorRecording` and the new `MockRecording`, name under discussion https://github.com/SpikeInterface/spikeinterface/pull/4520) have a `strategy` argument that determines how noise is generated internally. `tile_pregenerated` was the first strategy I introduced in #1581, and `on_the_fly` was added later by @samuelgarcia in #1948 with the goal of reducing memory consumption even further by avoiding the upfront allocation of a noise block.

I think this argument should be removed. How the noise is generated internally is an implementation detail that we should not expose to the user. The choice of strategy does not affect the behavior of the class from the outside; we should just keep whichever one performs best. If `on_the_fly` consumed less memory without a significant cost to generation speed, we should use it. The point is that the internal generation method could change over time and we do not want to lock ourselves into supporting both as part of the public API. The central contract of the class is to behave like an array-like lazy generator of noise.

I profiled both strategies with [memray](https://github.com/bloomberg/memray) to measure peak memory during `get_traces()`, which is the relevant metric for spikeinterface: in a preprocessing chain, each `get_traces` call's peak memory compounds through the stack. You can reproduce the results with `uv run profile_strategies.py` ([gist](https://gist.github.com/h-mayorquin/db025f02ac949d7ee80b32f999f9fb60)).

**Peak memory during get_traces()**

| Scenario | Output size | tile_pregenerated | on_the_fly | on-the-fly-memory-overhead |
|----------|------------|------------------|-----------|----------|
| 32ch, 1000 samples | 0.1 MB | 0.1 MB | 4.0 MB | 32x |
| 32ch, 30000 samples (1s) | 3.7 MB | 3.7 MB | 11.0 MB | 3x |
| 384ch, 1000 samples | 1.5 MB | 1.5 MB | 45.6 MB | 31x |
| 384ch, 30000 samples (1s) | 43.9 MB | 43.9 MB | 131.8 MB | 3x |
| 384ch, 90000 samples (3s) | 131.8 MB | 131.8 MB | 219.7 MB | 1.7x |
| 384ch, 1800000 samples (1min) | 2636.7 MB | 2636.7 MB | 2724.6 MB | 1.0x |

**Speed**

| Scenario | tile_pregenerated | on_the_fly | slowdown |
|----------|------------------|-----------|----------|
| 32ch, 1000 samples | 0.01 ms | 9.52 ms | 1087x |
| 32ch, 30000 samples (1s) | 0.57 ms | 17.63 ms | 31x |
| 384ch, 1000 samples | 0.09 ms | 111.43 ms | 1281x |
| 384ch, 30000 samples (1s) | 26.46 ms | 244.12 ms | 9x |
| 384ch, 90000 samples (3s) | 78.40 ms | 515.28 ms | 7x |
| 384ch, 1800000 samples (1min) | 1602.41 ms | 8144.07 ms | 5x |

`tile_pregenerated` is better overall. It allocates exactly the output array during `get_traces()`, nothing more, and is 7x to 3000x faster depending on the scenario. There are two places where `on_the_fly` has an advantage:

1. **Zero initialization cost.** `on_the_fly` allocates nothing at init, which is useful for serialization and dump/load. This has a simple solution: we can delay the generation of the tile until the first `get_traces()` call and cache it. That would give us zero init cost with the runtime performance of `tile_pregenerated`. We should do this regardless of which way we decide to go.

2. **Non-repeating noise across blocks.** `on_the_fly` seeds each block with `(seed, block_index)`, so different blocks produce genuinely different noise. This might be necessary for simulated data. I don't see a simple fix: numpy's RNG doesn't support seeking, so reproducibility requires generating the full block even for small slices. The overhead is worst when the requested trace is small relative to the noise block (default 30000 samples, 1 second at 30kHz), which is the typical preprocessing chunk size (3x memory, 9x speed). For larger reads the overhead shrinks (see the 1-minute row) but the speed cost never disappears.

Now that we are separating testing from simulation, tiling is the clear choice for the testing side. The open question is whether simulation needs non-repeating noise. @samuelgarcia @cwindolf @alejoe91 @chrishalcrow, does repeating the same noise block affect simulation quality or introduce artifacts in downstream analysis, or is a large enough block size indistinguishable from non-repeating noise in practice? I think we should leave the simulation side with `on_the_fly` as that is the [current default](https://github.com/SpikeInterface/spikeinterface/blob/4c3a6f9f9a1c61cd634fa353c469e4327c4fb900/src/spikeinterface/core/generate.py#L2319).

cwindolf:
This replaces https://github.com/SpikeInterface/spikeinterface/pull/4487

As in that PR, this is parametrized by the power spectral density along the time axis of the noise. So, the generated noise would have a factorized covariance matrix, with the spatial part controlled by the cov_matrix argument and with a new stationary temporal part.

Compared to my last PR, this one is simpler:
 - Removes the `strategy` argument, since only one strategy ('on_the_fly') was used for simulated data generation ('tile_pregenerated' was just for the mock testing recording, which now lives elsewhere)
 - Implements temporal correlations using a wrapper segment. As discussed with @samuelgarcia offline and at https://github.com/SpikeInterface/spikeinterface/issues/4482, I didn't want to use my circular padding logic from the last PR, since that was just to support tile_pregenerated. Now I use reflect padding, and I needed a segment wrapper in order to use the `get_chunk_with_margin()` function. But I have followed @samuelgarcia 's preference of not writing a whole preprocessor.

Closes https://github.com/SpikeInterface/spikeinterface/issues/4482

I would say this also closes https://github.com/SpikeInterface/spikeinterface/issues/4522 (if @h-mayorquin agrees) since I've removed that arg here and it was discussed for the mock recording in https://github.com/SpikeInterface/spikeinterface/pull/4520.


cwindolf:
I realized that many many tests are failing because the tests want to use @h-mayorquin 's new MockRecording, not the NoiseGeneratorRecording. Maybe that should be a separate PR?

h-mayorquin:
> I realized that many many tests are failing because the tests want to use @h-mayorquin 's new MockRecording, not the NoiseGeneratorRecording. Maybe that should be a separate PR?

Some are here:
https://github.com/SpikeInterface/spikeinterface/pull/4587

I think this should do most of them but there might be stragglers.

cwindolf:
Thanks @h-mayorquin  that's awesome!

cwindolf:
@h-mayorquin I cleaned up some remaining cases where the strategy kwarg was passed through `generate_ground_truth_recording()`. I figure that this is totally fine if `strategy='on_the_fly'`, since that's the only strategy in the `NoiseGeneratorRecording` after this PR.

But I was wondering if it bothers you that I have deleted some cases of `strategy='tile_pregenerated'` -- if so, we could consider adding a `mock=False` argument or something to `generate_ground_truth_recording()` that uses a mock recording instead? In my last few commits here I marked the tile_pregenerated changes so would be easy to track where to do that.

h-mayorquin:
Hi, @cwindolf. Thanks for the heads up. I am totally fine with the cleaning. What I am realizing now is that generate_ground_truth_recording is still in core. It should live in generation since it depends on NoiseGeneratorRecording. I can move it after your PR lands, or before if you would prefer a cleaner diff.

cwindolf:
Thanks! Either way is great for me. Otherwise I think this one is ready for review

cwindolf:
Ok, conflict fixed as discussed @h-mayorquin @samuelgarcia !

PR #4588, opened 2026-05-19
Move `generate_ground_truth_recording` to generation module

h-mayorquin:
This was left out by mistake from #4520. `generate_ground_truth_recording` should also be in the generation module to avoid a dependency of core into generation.

I also dropped `strategy=` from test fixtures and the two production callsites that pass it explicitly, keeping the kwarg itself defined on the `NoiseGeneratorRecording` and `MockRecording` classes (whose removal is under discussion in #4522). This overlaps with the callsite cleanup in @cwindolf's #4586 (temporally-correlated noise V2, which follows on from #4587) and should make that PR cleaner. Both branches edit the same lines, so the merge should resolve cleanly regardless of order.


samuelgarcia:
Even it looks noisy, I was happy with the explicit `strategy=` everywhere.
In my mind, it was : 
  * generate me a fast random noise, I don't care
  * or generate me a better noise even if it is slower


And more importantly we absolutly need that this must use always 'on_the_fly' !!!

h-mayorquin:
> Even it looks noisy, I was happy with the explicit `strategy=` everywhere. In my mind, it was :
> 
> * generate me a fast random noise, I don't care
> * or generate me a better noise even if it is slower
> 
> And more importantly we absolutly need that this must use always 'on_the_fly' !!!

I agree with the `on_the_fly`. The other strategy does not make sense for the scope of the generate_ground_truth os I don't think it should be there.

cwindolf:
@samuelgarcia you might want to also check out #4586, where I removed the strategy kw from the NoiseGeneratorRecording as we discussed and only kept the on_the_fly strategy , which I think addresses your comment here? generate_ground_truth would therefore always use on_the_fly, since it uses the NoiseGeneratorRecording

samuelgarcia:
I think maybe we lost a bit of backward compatibility because any `NoiseGeneratorRecording` than would have been saved in json or pickle in previous version will not be able to be reloaded. We need to discuss it a bit more.

h-mayorquin:
Good that you thought about this, @samuelgarcia . 

I checked this and I tested it locally, we are actually covered. The `__getattr__` hooks at the old locations take care of it: both JSON and pickle reload via `from_dict` -> [`_get_class_from_string`](https://github.com/SpikeInterface/spikeinterface/blob/58a50fe54e4c09a680574ca94798860692b6e675/src/spikeinterface/core/base.py#L1161) -> `getattr`, which triggers the hook in [`core/generate.py`](https://github.com/SpikeInterface/spikeinterface/blob/58a50fe54e4c09a680574ca94798860692b6e675/src/spikeinterface/core/generate.py#L2269) / [`core/__init__.py`](https://github.com/SpikeInterface/spikeinterface/blob/58a50fe54e4c09a680574ca94798860692b6e675/src/spikeinterface/core/__init__.py#L196) and redirects the old class path to `generation`, so old saved objects still reload.

What I think is the real problem is that we have no test for this in CI. I have some ideas, so I'll open a separate issue to track it. We should not rely on some of us remembering this.

h-mayorquin:
There is a PR for infra so we can settle this question:
https://github.com/SpikeInterface/spikeinterface/pull/4636

samuelgarcia:
@h-mayorquin and I discuss and we decided to not merge this PR.

Instead we will do

The `generate_ground_truth_recording` will stay in the core with another name like `generate_imple_recording_with_spikes`.

The Noise with battery include will move in the generate submodule : always on the fly + spatial correlation + temporal correlation.
The MockRecording that stay in core could be renamed SimpleNoiseRecording and handle : gain + pregenerated/on_the_fly.

alejoe91:
@h-mayorquin closing this and opened #4697 to track the last comment. Feel free to reopen if I misinterpreted

Issue #4589, opened 2026-05-19
KS4 test matrix is failing on Kilosort 4.1.1 – 4.1.6 (NumPy 2.0 incompatibility)


h-mayorquin:
cc @JoeZiminski @alejoe91

The `Testing Kilosort4` workflow iterates over every published Kilosort release on PyPI and runs the SpikeInterface integration tests against each one. Kilosort versions 4.1.1 through 4.1.6 were released in such a way that they are incompatible with the latest NumPy versions ([MouseLand/Kilosort#1026](https://github.com/MouseLand/Kilosort/issues/1026)), and that was only fixed in Kilosort 4.1.7.

The consequence is that every PR now picks up six pre-existing red checks that have nothing to do with the PR's diff. This has been broken on `main` itself since the Sunday cron run on 2026-04-19; five consecutive scheduled runs have failed without anyone catching it because cron-workflow failures don't surface on the main repo UI. The breakage is visible on my PR #4588 where 4.1.1 – 4.1.6 are red despite the diff not touching any sorter code; the workflow run is at [actions/runs/26117618646](https://github.com/SpikeInterface/spikeinterface/actions/runs/26117618646) and the `AttributeError: module 'numpy' has no attribute 'in1d'` trace is visible in the 4.1.1 job at [job/76810948024](https://github.com/SpikeInterface/spikeinterface/actions/runs/26117618646/job/76810948024).

What should we do with these releases going forward? Kilosort 4.1.1 – 4.1.6 are functional on NumPy < 2.0 but will always fail on any environment with NumPy ≥ 2.0; 4.1.7 onward works on either. Honestly, if the promise is that we keep working with those old versions we kind of need to put a ceiling on numpy which does not seem feasible... what do you think?

JoeZiminski:
Hey @h-mayorquin thanks for catching this and the write up. So if I understand correctly the situations are:

1) users working with numpy < 2.0 and kilosort 4.1.1-4.16 can currently run these versions through spikeinterface, but they are untested. 4.1.1 was released in August 2025 and 4.1.7 in March 2026 so I guess there is quite a few people who might be running this combination.
2) everyone else running ks < 4.1.1 or >= 4.1.7 with numpy 2.0 or < 2.0

Some possible solutions in order of preference might be:

[...]

PR #4591, opened 2026-05-20
Use legacy numpy versions for ks 4.1.1-4.1.6 tests

chrishalcrow:
Attempt to fix failing KS4 tests in #4589 

Simply installs numpy<2.4 for KS between 4.1.1 and 4.1.6. Also raises an error if user is using incompatible numpy+ks versions.


alejoe91 (approved):

h-mayorquin (approved):
Thanks for this.

PR #4593, opened 2026-05-22
Use probeinterface's renamed Neuropixels readers and new probe detectors

h-mayorquin:
probeinterface renamed `read_openephys` to `read_openephys_neuropixels` ([SpikeInterface/probeinterface#427](https://github.com/SpikeInterface/probeinterface/pull/427)) and `read_spikegadgets` to `read_spikegadgets_neuropixels` ([SpikeInterface/probeinterface#440](https://github.com/SpikeInterface/probeinterface/pull/440)) to make the Neuropixels-only scope of those readers explicit; the old names remain as deprecation aliases. The same two PRs add detectors `has_neuropixels_probes(settings_file, stream_name=...)` and `has_spikegadgets_neuropixels_probes(file)` that parse the settings XML / `.rec` header and return whether the recording carries any Neuropixels probe geometry.

This PR switches the OpenEphys and SpikeGadgets extractors to call the new names and gates probe attachment on the detectors. Previously both readers were called unconditionally with `raise_error=False`, so a recording that legitimately had no Neuropixels probe (Intan / Rhythm FPGA / NI-DAQmx for Open Ephys, tetrodes / ECU for SpikeGadgets) silently returned `None` and the extractor finished probe-less. With the detector check up front the no-probe path is a normal control-flow branch, and any real parsing error from the reader now surfaces instead of being swallowed. 
I
I am also removing the version guard and the `packaging` import, since `pyproject.toml` already pins `probeinterface>=0.3.2`.

### DIFF
--- a/src/spikeinterface/extractors/neoextractors/spikegadgets.py
+++ b/src/spikeinterface/extractors/neoextractors/spikegadgets.py
@@ -1,7 +1,13 @@
 from pathlib import Path
+import warnings
+import numpy as np
 
 import probeinterface
 from spikeinterface.core.core_tools import define_function_from_class
+from spikeinterface.extractors.neuropixels_utils import (
+    get_neuropixels_sample_shifts_from_probe,
+    compute_saturation_threshold_from_probe,
+)
 
 from .neobaseextractor import NeoBaseRecordingExtractor
 
@@ -55,8 +61,26 @@ class SpikeGadgetsRecordingExtractor(NeoBaseRecordingExtractor):
 
         if probeinterface.has_spikegadgets_neuropixels_probes(file_path):
             probegroup = probeinterface.read_spikegadgets_neuropixels(file_path)
-            # TODO: add adc sample shifts and saturation levels if available in the probe metadata
             self.set_probegroup(probegroup, in_place=True)
+
+            # get inter-sample shifts based on the probe information and mux channels
+            sample_shifts = np.array([])
+            saturation_thresholds_uV = []
+            for probe in probegroup.probes:
+                sample_shifts_probe = get_neuropixels_sample_shifts_from_probe(probe)
+                if sample_shifts_probe is not None:
+                    sample_shifts = np.concatenate([sample_shifts, sample_shifts_probe])
+                # add saturation levels if available
+                saturation_threshold_uV_probe = compute_saturation_threshold_from_probe(probe, self.stream_id)
+                if saturation_threshold_uV_probe is not None:
+                    saturation_thresholds_uV.append(saturation_threshold_uV_probe)
+
+            if len(sample_shifts) > self.get_num_channels():
+                self.set_property("inter_sample_shift", sample_shifts)
+            if len(set(saturation_thresholds_uV)) == 1:
+                self.annotate(saturation_threshold_uV=saturation_thresholds_uV[0])
+            else:
+                warnings.warn("Multiple saturation thresholds found for different probes, unable to annotate.")
 
     @classmethod
     def map_to_neo_kwargs(cls, file_path):
--- a/src/spikeinterface/extractors/neoextractors/spikegadgets.py
+++ b/src/spikeinterface/extractors/neoextractors/spikegadgets.py
@@ -1,8 +1,5 @@
 from pathlib import Path
 
-import packaging
-
-import packaging.version
 import probeinterface
 from spikeinterface.core.core_tools import define_function_from_class
 
@@ -56,11 +53,8 @@ class SpikeGadgetsRecordingExtractor(NeoBaseRecordingExtractor):
         )
         self._kwargs.update(dict(file_path=str(Path(file_path).absolute()), stream_id=stream_id))
 
-        probegroup = None  # TODO remove once probeinterface is updated to 0.2.22 in the pyproject.toml
-        if packaging.version.parse(probeinterface.__version__) > packaging.version.parse("0.2.21"):
-            probegroup = probeinterface.read_spikegadgets(file_path, raise_error=False)
-
-        if probegroup is not None:
+        if probeinterface.has_spikegadgets_neuropixels_probes(file_path):
+            probegroup = probeinterface.read_spikegadgets_neuropixels(file_path)
             self.set_probegroup(probegroup, in_place=True)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


chrishalcrow:
Hello, this PR demonstrates an awkward difference between uv and traditional python venvs.

`uv` checks all dependency groups and prioritizes git repos over pypi releases. E.g. if you run `uv run ...` `uv` will create a venv with probeinterface from source, since this is in `common-tests`
If you make a venv and `pip install` spikeinterface from source from this PR, the resolver completely ignore `common-tests` and installs probeinterface v0.3.2.

The new functions are only in probeinterface main. So if you try to load a openephys neuropixel recording with a `uv` generated venv, it's fine. If you do it with a pip installed venv, it errors.

Just wanted to point it out!

alejoe91 (approved):

h-mayorquin:
> Hello, this PR demonstrates an awkward difference between uv and traditional python venvs.
...
> 
> Just wanted to point it out!

Maybe we should pin to github on probeinterface and python-neo on main. Probably worth its own issue.

alejoe91:
> > Hello, this PR demonstrates an awkward difference between uv and traditional python venvs.
> > ...
> > Just wanted to point it out!
> 
> Maybe we should pin to github on probeinterface and python-neo on main. Probably worth its own issue.

Good idea! Let's make a separate PR for it.

@h-mayorquin I'll try to add adc sample shifts/saturation for spikegadgets in the same PR, so everything is unified.

src/spikeinterface/extractors/neoextractors/spikegadgets.py
```diff
@@ -56,12 +59,29 @@ def __init__(
...
+            for probe in probegroup.probes:
+                sample_shifts_probe = get_neuropixels_sample_shifts_from_probe(probe)
+                if sample_shifts_probe is not None:
+                    sample_shifts = np.concatenate([sample_shifts, sample_shifts_probe])
+                # add saturation levels if available
+                saturation_threshold_uV_probe = compute_saturation_threshold_from_probe(probe, self.stream_id)
+                if saturation_threshold_uV_probe is not None:
+                    saturation_thresholds_uV.append(saturation_threshold_uV_probe)
+
+            if len(sample_shifts) > self.get_num_channels():
```
h-mayorquin: should this be equal?

src/spikeinterface/extractors/neoextractors/spikegadgets.py
```diff
@@ -56,12 +59,29 @@ def __init__(
...
-        if probegroup is not None:
+        if probeinterface.has_spikegadgets_neuropixels_probes(file_path):
+            probegroup = probeinterface.read_spikegadgets_neuropixels(file_path)
             self.set_probegroup(probegroup, in_place=True)
 
+            # get inter-sample shifts based on the probe information and mux channels
+            sample_shifts = np.array([])
+            saturation_thresholds_uV = []
+            for probe in probegroup.probes:
+                sample_shifts_probe = get_neuropixels_sample_shifts_from_probe(probe)
```
h-mayorquin, line 74: The function `get_neuropixels_sample_shifts_from_probe` returns the sample shifts in probe (contact) order, so before you set them as a channel property you need to re-wire them into channel order. This works for SpikeGLX and OpenEphys without any extra step because probeinterface already wires those probes so their contacts are in the order of the binary traces (`device_channel_indices = arange`), so contact order already equals channel order.

We could do that re-wiring here too, but it is more complicated than in the SpikeGLX and OpenEphys case. Those write one file per probe, while SpikeGadgets puts more than one probe in a single file. For the dataset with neuropixels that we have on gin already ([`SpikeGadgets_test_data_2xNpix1.0_20240318_173658.rec`](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/spikegadgets/SpikeGadgets_test_data_2xNpix1.0_20240318_173658.rec), the two probes are interleaved in that single binary in blocks of 32 channels (channels 0-31 are probe 0, 32-63 are probe 1, 64-95 are probe 0, and so on), so a probe's contacts do not map to a contiguous run of channels and `device_channel_indices` is a real permutation (`hwChan`) rather than `arange`. That is why the current `np.concatenate` of per-probe shifts lands on the wrong channels (704 of 768 on this fixture).

I see three possible paths forward:

1) Re-route the output of `get_neuropixels_sample_shifts_from_probe` through `device_channel_indices`. Each probe's `device_channel_indices` tells you which recording channel each contact maps to, so you scatter the per-probe shifts into a channel-length array:

```python
sample_shifts = np.zeros(self.get_num_channels())
for probe in probegroup.probes:
    shifts = get_neuropixels_sample_shifts_from_probe(probe)
    if shifts is not None:
        sample_shifts[probe.device_channel_indices] = shifts   # scatter into channel order
self.set_property("inter_sample_shift", sample_shifts)
```

2) Re-sort at the probeinterface level using the global device channel indices, sorting each probe/group so the contacts come out in channel order. I am not sure this is possible, since each probe owns a non-contiguous, interleaved set of channels, so a per-probe `arange` might not be doable but given that you guys like that pattern maybe you might be able to find it. I don't like the pattern and you know how motivated thinking is a horrible beast.

3) Use the property from the probe after you set it. This does the rearranging after we set it on the recording (so it is already re-ordered by `set_probegroup`): read `self.get_property("contact_vector")["adc_sample_order"]`, which is already one value per channel in channel order, and set the property from that.

I honestly prefer 1: explicit is better than implicit, and it will not clash with any of the current Probe refactorings like option 3 might. Option 2 will require more work on your side and I would like to merge this and then do that if you want to move that way.

src/spikeinterface/extractors/neoextractors/spikegadgets.py
alejoe91: Oups yes of course! Good catch!

alejoe91: fixed in last commit.

src/spikeinterface/extractors/neoextractors/spikegadgets.py
alejoe91: @h-mayorquin thanks! This was super helpful!

I agree option one is the cleanest. Implemented in last commit and fixed the sample_shift logic. Since now we pre-allocate an array, insted of checking the length I instantiate the values as -1, and we check that no -1 are left (since sample shifts are [0-1])

h-mayorquin: Cool, I am glad it helped. This is fine to merge to me.

chrishalcrow (approved):

PR #4596, opened 2026-05-27
QualityMetrics: store `use_valid_periods` param

alejoe91:
Currently, if `use_valid_periods` is given, the `periods` are retrieved from the `valid_unit_periods` extensions and stored as params, but this adds an ambiguity since there is no way to tell a posteriori if periods used for computing quality metrics were actually valid unit periods or externally defined periods. This PR simply propagates the param so it gets saved to the extension.


h-mayorquin (approved):
LGTM

src/spikeinterface/metrics/quality/quality_metrics.py
```diff
@@ -115,15 +116,19 @@ def _set_params(
             metric_names = [m for m in metric_names if m not in pc_metric_names]
 
         if use_valid_periods:
+            valid_periods = self.sorting_analyzer.get_extension("valid_unit_periods").get_data(outputs="numpy")
             if periods is not None:
-                raise ValueError("If use_valid_periods is True, periods should not be provided.")
-            periods = self.sorting_analyzer.get_extension("valid_unit_periods").get_data(outputs="numpy")
+                provided_periods = cast_periods_to_unit_period_dtype(np.asarray(periods))
```
chrishalcrow, line 121: Is there a reason for allowing the user to pass `use_valid_periods` and `periods` now? Something to do with reloading??

alejoe91: yes it just has to do with reloading, since we added the `use_valid_periods` params for provenance

PR #4598, opened 2026-05-28
Fix curation tests for missing kachery dependency

alejoe91:
Fixes failing curation tests (sortingview curation) as sortingview and kachery are deprecated. The PR skips the sortingview curation test if kachery is not installed


alejoe91:
#4599 is the same but with clea GH history

PR #4600, opened 2026-05-31
Add whitening_seed parameter to MountainSort5

grahamfindlay:
MountainSort5 whitens the recording before sorting, and the whitening matrix is estimated from randomly-selected data chunks (`spikeinterface.preprocessing.whiten` -> `get_random_data_chunks`).

Without control of the seed, sorting isn't reproducible, even on identical input.

Added a `whitening_seed` parameter (default `None`, preserves current behavior) that is forwarded to `whiten(seed=...)`.

There is still some very, very small residual indeterminism (on my test recording: 100% of units match, but ~0.04% of spikes differ) between runs with identical input. Not sure where it comes from -- maybe BLAS multithreading? Not worried about it. 

Looks okay @magland ? 

(The reason I'm trying to get deterministic output is so that I can evaluate whether the lossy `WavPack(bps=2.25)` compression that [has negligible impact on Neuropixel sortings](https://elifesciences.org/reviewed-preprints/110170) is also safe for some tetrode recordings. @alejoe91 )

### DIFF
--- a/src/spikeinterface/sorters/external/mountainsort5.py
+++ b/src/spikeinterface/sorters/external/mountainsort5.py
@@ -43,6 +43,7 @@ class Mountainsort5Sorter(BaseSorter):
         "freq_max": 6000,
         "filter": True,
         "whiten": True,  # Important to do whitening
+        "whitening_seed": None,  # seed for whitening's random chunk selection (None = nondeterministic)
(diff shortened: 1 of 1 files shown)


chrishalcrow (approved):
LGTM

magland:
Looks good, thanks!

PR #4607, opened 2026-06-04
Fix file size calculation in `write_binary()`

grahamfindlay:
`write_binary()` was determining the file size and seek offests based on the source dtype's itemsize, but the data is cast to the target dtype before writing. This means that if the target dtype has a different itemsize than the source dtype, a corrupted file of the wrong size and layout was produced.

The solution was just just use the target dtype's itemsize for all file size and seek offset calculations.


alejoe91 (approved):

alejoe91:
@grahamfindlay I extended the signature of the `get_sample_size_in_bytes` to optionally accept the dtype. I thimk it's better for readibility.

PR #4608, opened 2026-06-08
Use np.load(..., mmap_mode="r") for time_vector in `BaseRecording._extra_metadata_from_folder`

grahamfindlay:
`BaseRecording._extra_metadata_from_folder` was eagerly loading a `BinaryFolderRecording`'s time vector, which can be tens of GB on long recordings.

I ran into this because `run_sorter_by_property` (I was sorting tetrodes) reconstructs the recording _once per joblib worker_, so every worker was loading the whole time vector, even when only a short frame slice was needed.

Another consequence of the eager loading was that `si.load` on a 48h recording with a time_vector was using 1.6GB of memory, even if the time vector was never touched.

I switched the load to `mmap_mode="r"`, so memory use no longer scales with both recording duration and number of workers. It is bounded to what's actually touched, and the memmap can be shared by the joblib workers.

The cost is that this time vector is read-only, so in-place operations on it would raise an error.

Thankfully there was only 1: `TimeSeries.shift_times`! I made it shift in place only when the vector is writeable and fall back to an out-of-place op for the now read-only memmap. (For simplicity, we could scrap the conditional altogether and just keep the out-of-place path, since this is presumably a pretty rare op -- but I chose to keep the in-place, no-copy path for best performance).

### DIFF
--- a/src/spikeinterface/core/time_series.py
+++ b/src/spikeinterface/core/time_series.py
@@ -1,10 +1,25 @@
+from __future__ import annotations
+
 from abc import ABC, abstractmethod
-from typing import Optional
+from typing import Optional, TYPE_CHECKING, TypeAlias
 import warnings
 
 import numpy as np
 
 from spikeinterface.core.base import BaseExtractor, BaseSegment
+
+if TYPE_CHECKING:
+    import zarr
+
+# A recording segment's time vector: a 1-D array of per-sample times (in seconds).
+# The backing store depends on how the recording was created/loaded:
+#   - np.ndarray : set_times() (writeable, in-memory)
+#   - np.memmap  : BinaryFolderRecording load via np.load(..., mmap_mode="r")
+#                  -- *read-only* ; see BaseRecording._extra_metadata_from_folder
+#   - zarr.Array : ZarrRecordingExtractor load
+#                  -- *read-only* ; see ZarrRecordingExtractor.__init__
+# Code reading `.time_vector` must not assume it is writeable (see `shift_times`).
+TimeVector: TypeAlias = "np.ndarray | zarr.Array"  # np.memmap is an np.ndarray subclass
 
 
 class TimeSeries(ABC):
@@ -146,8 +161,8 @@ class TimeSeries(ABC):
 
         Returns
         -------
-        np.array
-            The 1d times array
+        np.ndarray
+            The 1d times array. If the times were mem-mapped, loads them into memory.
         """
         segment_index = self._check_segment_index(segment_index)
         rs = self.segments[segment_index]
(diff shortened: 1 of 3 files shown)


src/spikeinterface/core/time_series.py
```diff
@@ -273,7 +273,13 @@ def shift_times(self, shift: int | float, segment_index: int | None = None) -> N
...
 
             if self.has_time_vector(segment_index=segment_index):
-                rs.time_vector += shift
+                if rs.time_vector.flags.writeable:
+                    # If the time_vector is writeable, shift in-place to avoid a copy.
+                    rs.time_vector += shift
+                else:
+                    # If the time_vector is a memmap from `np.load(..., mmap_mode='r')`,
+                    # in-place modification would error, so we shift a writable copy.
+                    rs.time_vector = rs.time_vector + shift
```
alejoe91: Why not simply this?
```suggestion
                rs.time_vector = rs.time_vector + shift
```

samuelgarcia: to avoid copy no ?

alejoe91: ok let's keep it this way!

src/spikeinterface/core/time_series.py
```diff
@@ -273,7 +273,13 @@ def shift_times(self, shift: int | float, segment_index: int | None = None) -> N
             rs = self.segments[segment_index]
 
             if self.has_time_vector(segment_index=segment_index):
-                rs.time_vector += shift
+                if rs.time_vector.flags.writeable:
```
alejoe91: we should check if zarr.Array also implements this flag

alejoe91: actually, it doesn't have this flag, so we should do:
```suggestion
                if isinstance(rs.time_vector, np.ndarray)  and rs.time_vector.flags.writeable:
```

grahamfindlay: Oh, wow, I didn't even realize `time_vector` _could_ be a `zarr.Array`! In that case, the out-of-place op is also going to fail. And the in-place op is probably already failing -- I guess it just never surfaced. 

So I think we actually need something like:
```python
if isinstance(rs.time_vector, np.ndarray) and rs.time_vector.flags.writeable:
      rs.time_vector += shift    # in-place, no copy
  else:
      rs.time_vector = np.asarray(rs.time_vector) + shift    # read-only memmap or zarr.Array
```

alejoe91 (approved):

src/spikeinterface/core/time_series.py
alejoe91: Perfect!

PR #4610, opened 2026-06-10
Hotfix: overflow error in correct LSB

alejoe91:
I found this bug in some sessions...when the LSB estimation overlflows, everything gets messed up.

This PR takes care of overflow errors and it also uses a mode approach to estimate LSB from channels, since LSB is set at the acquisition level. This approach is more robust against rounding errors.

### DIFF
--- a/src/spikeinterface/preprocessing/correct_lsb.py
+++ b/src/spikeinterface/preprocessing/correct_lsb.py
@@ -10,6 +10,11 @@ def correct_lsb(recording, num_chunks_per_segment=20, chunk_size=10000, seed=Non
     Estimates the LSB of the recording and divide traces by LSB
     to ensure LSB = 1. Medians are also subtracted to avoid rounding errors.
 
+    Since the LSB is set at the acquisition level, it is shared across channels. The global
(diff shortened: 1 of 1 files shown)


src/spikeinterface/preprocessing/correct_lsb.py
```diff
@@ -76,20 +86,23 @@ def _estimate_lsb_from_data(data):
...
         The estimated LSB value. -1 indicates that lsb could not be estimated
     """
-    lsb_value = None
     num_channels = data.shape[1]
+    per_channel_lsb = []
     for ch in np.arange(num_channels):
-        unique_values = np.unique(data[:, ch])
+        # cast to int64 to avoid integer overflow in np.diff when consecutive
+        # unique values are farther apart than the dtype range (e.g. int16)
+        unique_values = np.unique(data[:, ch]).astype(np.int64)
```
chrishalcrow, line 94: Never thought about this gross bug before - nice!

chrishalcrow (approved):
Works on my local data, and looks sensible.

PR #4615, opened 2026-06-11
Use probeinterface and neo source when using spikeinterface source

chrishalcrow:
As discussed in #4593

Currently, when you install SpikeInterface from source, the dependencies are release versions of ProbeInterface and Python-Neo. `pip` obeys this. `uv` looks at the other dependency groups and decides that the source versions should be used. Hence we have a mismatch in behavior between `uv` and `pip`.

Additionally, if you are using SpikeInterface from source, you almost certainly want ProbeInterface and Neo from source. This PR makes this the default.

Note: now when you release, you only need to change two lines in the pyproject.toml, not six!
Note: the development docs still make sense after this change.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,8 +25,11 @@ dependencies = [
     "threadpoolctl>=3.0.0",
     "tqdm",
     "zarr>=2.18,<3",
-    "neo>=0.14.4",
-    "probeinterface>=0.3.2",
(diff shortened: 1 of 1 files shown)


h-mayorquin (approved):
LGTM.

The workflow would be to pin at release and keep the development version while on development.

alejoe91 (approved):

PR #4620, opened 2026-06-16
Avoid materializing entire time vector from zarr for time_to_sample_index

grahamfindlay:
`np.searchsorted` is fine for mem-maps, but for out-of-core arrays (zarr) it reads the whole time vector (even if a `zarr.Array`) into memory. Bisecting instead reads O(log N) elements, which saves an order of magnitude of RAM for long recordings.


alejoe91:
Stumbled upon the same issue yesterday!! 

Elegant solution :)

grahamfindlay:
@alejoe91 
I am pretty sure that this failing test has nothing to do with the PR. I think it is because `test_pipeline_equiv_to_step` calls multiple preprocessors (e.g. `detect_and_remove_artifacts`, `detect_and_interpolate_bad_channels`) that are nondeterministic without seeding them. Pick any unseeded nondeterministic preprocessor, call it twice and test the results for equivalence, and if you do it enough times, you'll get a failure. 

Pretty sure this is hitting #4601 too. Both safe to merge. I can work on a PR to fix the test.

alejoe91 (approved):

Issue #4605, opened 2026-06-02
Proposal: Add anti-aliasing option to decimate()


grahamfindlay:
`DecimateRecording` applies simple array slicing without an anti-aliasing (AA) filter:
```python
"""
class DecimateRecording(BasePreprocessor):
  Decimate the recording extractor traces using array slicing
  
  Important: This uses simple array slicing for decimation rather than eg scipy.decimate.
  This might introduce aliasing, or skip across signal of interest.
  Consider  spikeinterface.preprocessing.ResampleRecording for safe resampling.
  ...
"""
```

I'd like to add an "antialias" option to `DecimateRecording`:
- If `True`, it would use `scipy.decimate`. 
- If `True` and `decimation factor > 13`, it would warn the user that `decimate()` should be called multiple times, with decimation factors as balanced as possible (e.g. for a total decimation factor of 48, a pass with `q=8` followed by `q=6` is preferable to `q=12` followed by `q=4`). I can see if there's a good way to handle this _for_ the user, but I think it's impossible for the `DecimateRecording` constructor itself to chain together `DecimateRecordings`. Besides, this would be consistent with the behavior of the underlying `spicy.decimate()`, which warns but does not refuse `q > 13`. 
    - For consistency, I would add the same check and warning to the decimation path in `ResampleRecording`. 
- If `True`, a margin would be used to accommodate the AA filter, exactly as it is in `ResampleRecording`. 
- `ResampleRecording` says that decimation can "misbehave on some cases", and guards against this:
    ```python
    # Decimate can have issues for some cases, returning NaNs
    resampled_traces = signal.decimate(parent_traces, q=q, axis=0)
    # If that's the case, use signal.resample
    if np.any(np.isnan(resampled_traces)):
    resampled_traces = signal.resample(parent_traces, num, axis=0)
    ```
    I would check for `NaN` and warn the user if any are found. But in my experience, decimation only does this when `q > 13`, so it could be  dropped IMO. If there are other cases where this is known to happen, I would like to document them with a comment to justify the every-call overhead of `np.any(np.isnan())`. 
- I would make `antialias=False` the default, to preserve existing behavior. 

[...]

PR #4621, opened 2026-06-17
Add antialiased decimation and allow non-integer resampling rates

grahamfindlay:
Implements #4605 : Adds antialised decimation (multipass with balanced factors, when possible) to both `DecimateRecording` and `ResampleRecording`. Opt-in for `DecimateRecording` with the `antialias` parameter (preserves existing default behavior). For `ResampleRecording`, antialiasing is always applied. For a future release where existing default behavior can be changed, I would recommend making `antalias=True` the default. 

For simplicity, we _could_ just remove the fast decimation path from `ResampleRecording` entirely and instead, if an integer resampling ratio is detected, just suggest to the user that they use `DecimateRecording` instead. But, that would change existing behavior, and I don't think this is too terribly complicated as-is.


grahamfindlay:
@alejoe91 Fixed (surface-level) conflicts introduced by merging #4499 into main. All good to go.

src/spikeinterface/preprocessing/_decimation_tools.py
```diff
@@ -0,0 +1,178 @@
...
+    ``left_margin // decimation_factor`` is exact; combined with scipy's default
+    ``zero_phase=True`` (output sample i maps to filtered input sample i * factor), this keeps the
+    downsampled traces aligned across chunks (a chunked read matches a full read). Exactly
+    ``end_frame - start_frame`` decimated samples are returned.
+
+    Parameters
+    ----------
+    parent_segment : BaseRecordingSegment
+        The parent segment to read (full-rate) traces from.
+    start_frame, end_frame : int
```
JoeZiminski: separate

src/spikeinterface/preprocessing/_decimation_tools.py
```diff
@@ -0,0 +1,178 @@
...
+        decimated_traces = signal.decimate(decimated_traces, q=sub_q, axis=0)
+    if np.any(np.isnan(decimated_traces)):
+        warnings.warn(
+            f"`scipy.signal.decimate` produced NaNs while decimating by {q}. "
+            f"Consider a different decimation factor."
+        )
+    start_drop = left_margin // q
+    n_out = end_frame - start_frame
+    decimated_traces = decimated_traces[start_drop : start_drop + n_out]
+    if np.issubdtype(np.dtype(dtype), np.integer):
```
JoeZiminski: overflow risk after filter?

src/spikeinterface/preprocessing/decimate.py
```diff
@@ -63,7 +88,21 @@ def __init__(
         self._decimation_offset = decimation_offset
         decimated_sampling_frequency = self._orig_samp_freq / self._decimation_factor
 
-        BasePreprocessor.__init__(self, recording, sampling_frequency=decimated_sampling_frequency)
+        # fix_dtype doesn't always returns the str, make sure it does
+        dtype = fix_dtype(recording, dtype).str
+
+        antialias_factors = get_balanced_decimation_factors(decimation_factor)
```
JoeZiminski: Take _MAX_SINGLE_PASS_DECIMATION  as argument?

src/spikeinterface/preprocessing/decimate.py
```diff
@@ -63,7 +88,21 @@ def __init__(
...
+        if antialias and decimation_factor > _MAX_SINGLE_PASS_DECIMATION and antialias_factors == [decimation_factor]:
+            warnings.warn(
+                f"`decimation_factor`={decimation_factor} cannot be split into anti-aliasing passes of <= 13 "
+                f"(it has a prime factor > 13). A single `scipy.signal.decimate` pass will be used, which may be "
+                f"unstable. Consider a `decimation_factor` without large prime factors."
+            )
+
+        # Margin (in parent samples) to limit anti-aliasing filter edge effects.
+        margin = int(margin_ms * self._orig_samp_freq / 1000)
```
JoeZiminski: Currently a one-size-fits-all margin, IIR filer should be more than sufficient. Any analysis?

src/spikeinterface/preprocessing/resample.py
```diff
@@ -71,13 +91,24 @@ def __init__(
...
         self._orig_samp_freq = recording.get_sampling_frequency()
         self._resample_rate = resample_rate
         self._sampling_frequency = resample_rate
+        # When the parent rate is an exact integer multiple of resample_rate, get_traces uses
+        # anti-aliased decimation. Large factors are split into several balanced sub-13 passes
+        # (see _decimation_tools); only an unsplittable factor (a prime > 13) falls back to a
+        # single, potentially unstable, pass and warns.
+        if self._orig_samp_freq % resample_rate == 0:
+            decimation_factor = int(self._orig_samp_freq / resample_rate)
+            decimation_factors = get_balanced_decimation_factors(decimation_factor)
```
JoeZiminski, line 103: Could this check be centralised in _decimation_tools.py and also called from decimate.py? (and/or `antialias_factors` there called `decimation_factors` for consistencey)

src/spikeinterface/preprocessing/resample.py
```diff
@@ -71,13 +91,24 @@ def __init__(
...
-        assert isinstance(resample_rate, (int, np.integer)), msg
-        # Original sampling frequency
         self._orig_samp_freq = recording.get_sampling_frequency()
         self._resample_rate = resample_rate
         self._sampling_frequency = resample_rate
+        # When the parent rate is an exact integer multiple of resample_rate, get_traces uses
+        # anti-aliased decimation. Large factors are split into several balanced sub-13 passes
+        # (see _decimation_tools); only an unsplittable factor (a prime > 13) falls back to a
+        # single, potentially unstable, pass and warns.
+        if self._orig_samp_freq % resample_rate == 0:
```
JoeZiminski: found this sliighly difficult toread, maybe just a explamatory comment if decimation_factors is None we are upsampling or downsampling with FFT

src/spikeinterface/preprocessing/resample.py
```diff
@@ -248,8 +284,23 @@ def get_traces(self, start_frame, end_frame, channel_indices):
...
+                start_frame,
+                end_frame,
+                channel_indices,
+                decimation_factor,
+                self._decimation_factors,
+                self._margin,
+                self._dtype,
+            )
+
+        # Non-integer ratio, no gaps: FFT-based resampling with proportional margins.
```
JoeZiminski, line 301:
```suggestion
        # Non-integer ratio downsampling or upsampling with no gaps: FFT-based resampling with proportional margins.
```

src/spikeinterface/preprocessing/resample.py
```diff
@@ -248,8 +284,23 @@ def get_traces(self, start_frame, end_frame, channel_indices):
         if self._has_gaps:
             return self._get_traces_gapped(start_frame, end_frame, channel_indices)
 
-        # Original code path: no gaps (or no time_vector)
-        # get parent traces with margin
+        if self._decimation_factors is not None:
+            # Integer ratio, no gaps: anti-aliased multi-pass decimation.
```
JoeZiminski, line 288:
```suggestion
            # Integer ratio downsampling, no gaps: anti-aliased multi-pass decimation.
```

src/spikeinterface/preprocessing/resample.py
```diff
@@ -368,10 +405,13 @@ def _get_traces_gapped(self, start_frame, end_frame, channel_indices):
             chunk_len = int(local_out_end - local_out_start)
             num = chunk_len + left_margin_rs + right_margin_rs
 
-            # Resample this section
+            # Resample this section. Integer ratios use anti-aliased multi-pass decimation
+            # (the balanced sub-13 passes computed in ResampleRecording.__init__), applied
+            # within the section only so filtering never crosses a gap.
             if is_integer_ratio:
```
JoeZiminski: maybe we can call this `is_integer_ratio_decimation`

JoeZiminski: in general hard to fgire out what path we are on

src/spikeinterface/preprocessing/resample.py
```diff
@@ -368,10 +405,13 @@ def _get_traces_gapped(self, start_frame, end_frame, channel_indices):
...
-                q = int(self._parent_rate / self._resample_rate)
-                resampled = decimate(parent_traces, q=q, axis=0)
+                resampled = parent_traces
+                for sub_q in self._decimation_factors:
+                    resampled = decimate(resampled, q=sub_q, axis=0)
                 if np.any(np.isnan(resampled)):
```
JoeZiminski: Do we still need this?

src/spikeinterface/preprocessing/resample.py
```diff
@@ -368,10 +405,13 @@ def _get_traces_gapped(self, start_frame, end_frame, channel_indices):
...
 
-            # Resample this section
+            # Resample this section. Integer ratios use anti-aliased multi-pass decimation
+            # (the balanced sub-13 passes computed in ResampleRecording.__init__), applied
+            # within the section only so filtering never crosses a gap.
             if is_integer_ratio:
-                q = int(self._parent_rate / self._resample_rate)
-                resampled = decimate(parent_traces, q=q, axis=0)
+                resampled = parent_traces
+                for sub_q in self._decimation_factors:
```
JoeZiminski: can we use get_antialiased_decimated_traces here?

src/spikeinterface/preprocessing/resample.py
```diff
@@ -248,8 +284,23 @@ def get_traces(self, start_frame, end_frame, channel_indices):
         if self._has_gaps:
             return self._get_traces_gapped(start_frame, end_frame, channel_indices)
 
-        # Original code path: no gaps (or no time_vector)
-        # get parent traces with margin
+        if self._decimation_factors is not None:
+            # Integer ratio, no gaps: anti-aliased multi-pass decimation.
+            decimation_factor = int(self._parent_rate / self._resample_rate)
+            return get_antialiased_decimated_traces(
```
JoeZiminski, line 290: Here if Nan detected we warn, but if Nan detected in the section implementation we resample. Can these align?

src/spikeinterface/preprocessing/_decimation_tools.py
JoeZiminski, line 176: Is it worth doing something like:

```
    np.round(decimated_traces, out=decimated_traces)
    info = np.iinfo(dtype)
    np.clip(decimated_traces, info.min, info.max, out=decimated_traces)
```

just in case overflow, although I guess the risk is very low with low-pass filter unless we encounter ringing artefact

src/spikeinterface/preprocessing/decimate.py
JoeZiminski, line 94: This function is build around the `13` recommendation of Scipy. Is it easy to / worth generalizing, and passing `_MAX_SINGLE_PASS_DECIMATION ` as an argument?

src/spikeinterface/preprocessing/decimate.py
JoeZiminski, line 103: This margin seems a good catch-all. I guess the main use case of this will be going from AP band to LFP band? I think this margin will only be a problem based on going to very low sampling rates:

Image

(this plot was generated by ChatGPT using the below sensible code to measure the impulse response from the corresponding 8th order Bessel). I guess we could add a warning for low reample_rate and/or perform a check similar to the one below directly from the filter (probably overkill)?

<details>

```
import numpy as np
from scipy import signal


def estimate_iir_margin(fs, fout, eps=1e-6):
    q = fs / fout
    assert np.isclose(q, round(q))
    q = int(round(q))

    # Same default IIR design used by scipy.signal.decimate
    sos = signal.cheby1(
        8,
        0.05,
        0.8 / q,
        output="sos",
    )

    # Long impulse, far away from either edge
    n = 200_000
    x = np.zeros(n)
    centre = n // 2
    x[centre] = 1.0

    # Effective zero-phase filter
    h = signal.sosfiltfilt(sos, x)

    threshold = eps * np.max(np.abs(h))

    indices = np.flatnonzero(np.abs(h) > threshold)

    radius_samples = max(
        centre - indices[0],
        indices[-1] - centre,
    )

    margin_ms = radius_samples / fs * 1000

    return q, radius_samples, margin_ms


fs = 30_000

for fout in [2500, 1000, 500, 250, 100]:
    q, samples, margin_ms = estimate_iir_margin(fs, fout)

    print(
        f"{fs} -> {fout:4} Hz "
        f"q={q:3}, "
        f"margin={samples:5} samples = {margin_ms:.1f} ms"
    )
```

</details>

src/spikeinterface/preprocessing/resample.py
```diff
@@ -71,13 +91,24 @@ def __init__(
...
         skip_checks=False,
     ):
-        # Floating point resampling rates can lead to unexpected results, avoid actively
-        msg = "Non integer resampling rates can lead to unexpected results."
-        assert isinstance(resample_rate, (int, np.integer)), msg
-        # Original sampling frequency
         self._orig_samp_freq = recording.get_sampling_frequency()
         self._resample_rate = resample_rate
         self._sampling_frequency = resample_rate
+        # When the parent rate is an exact integer multiple of resample_rate, get_traces uses
```
JoeZiminski, line 97:
```suggestion
        # When the parent rate is an exact integer multiple of resample_rate (i.e. downsampling with an integer factor), get_traces uses
```

src/spikeinterface/preprocessing/_decimation_tools.py
```diff
@@ -0,0 +1,178 @@
...
+    while num_passes <= len(primes_desc):
+        bins = _greedy_pack(primes_desc, num_passes)
+        if bins is not None:
+            return sorted(bins, reverse=True)
+        num_passes += 1
+    # Fallback: one prime per pass (always valid since every prime is <= 13).
+    return primes_desc
+
+
+def get_antialiased_decimated_traces(
```
JoeZiminski, line 108: Could this move to `decimate.py`? (if it can be imported from `resample.py`). Not a huge thing but I think the convention is for get-traces like functionality to live in next to the segments.

JoeZiminski (commented):
Hi @grahamfindlay this looks great! Some small suggestions on the code. I didn't have a chance to play around locally and look at the tests but will do that early next week! In the meantime just some general questions:


1) Do you think it is worth putting a warning on `antialias=False` with possibly a timeline to switching the default to `True`? I would be tempted to change it for `105` but it's such a big change it might trip people up! But I guess it is important for people to use the AA filter especially if going from AP -> LFP band which I guess is the most common use of the decimate function?

2) What do you think of [resample_poly](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.resample_poly.html)? Would solve a few issues (more predictable artefact than FFT method, natively handles large decimation factors, can funnel all resample calls to it rather than splitting decimate by integer/non-integer)) but maybe tricky to always find nice `up`/`down` arguments from the resample ratio depending on the sampling rates.

3) Currently it warns on `NaN` if found after decimate, but should we error? I think (?) that `NaN` in traces are not supported in downstream methods e.g. `np.nan...` methods are not typically used.

4) The methods to handle time-gaps in the recording is really elegant. In the wider-picture of SI though, I wonder if it is necessary to support this mode, and if the module can become more readable by stripping it out. The reason being that AFAIK SI makes no guarantees on supporting gaps in the `time_vector` for preprocessing, and most steps do not consider it nor work properly for it (e.g. filtering steps, or even `decimate.py`). My understanding is that in SI time gaps should be handled by the user by using segments or separate recordings. I think in the `resample.py` case we could just check for time gaps and error accordingly. That being said maybe there are important use cases for this I am overlooking.

Cheers!

grahamfindlay:
> Hi @grahamfindlay this looks great! Some small suggestions on the code. I didn't have a chance to play around locally and look at the tests but will do that early next week! In the meantime just some general questions:
> 
> 1. Do you think it is worth putting a warning on `antialias=False` with possibly a timeline to switching the default to `True`? I would be tempted to change it for `105` but it's such a big change it might trip people up! But I guess it is important for people to use the AA filter especially if going from AP -> LFP band which I guess is the most common use of the decimate function?

Yeah, I think that a transition warning is a good idea. I might distinguish an omitted argument from an explicit `antialias=False`, though, in case someone has deliberately applied an appropriate low-pass filter upstream and doesn't deserve repeated warnings. I will add this. 

> 2. What do you think of [resample_poly](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.resample_poly.html)? Would solve a few issues (more predictable artefact than FFT method, natively handles large decimation factors, can funnel all resample calls to it rather than splitting decimate by integer/non-integer)) but maybe tricky to always find nice `up`/`down` arguments from the resample ratio depending on the sampling rates.

I definitely would support the use of `resample_poly`. It's closer to what I actually use with my own data. I think the most disruptive thing about it would be that currently we always deliver the exact requested target resample rate[^1], whereas with `resample_poly` we would only promise to deliver something very close. Personally, I'd rather have that, but maybe others feel differently. Happy to implement it if anyone gives me the green light. 

> 3. Currently it warns on `NaN` if found after decimate, but should we error? I think (?) that `NaN` in traces are not supported in downstream methods e.g. `np.nan...` methods are not typically used.

Yes, I agree. Actually, I think it's even worse than `NaN` not being supported downstream: casting to an integer type will turn the `NaN` to zeros, which will let downstream operations proceed where they probably should not... FWIW, I would also remove the silent FFT retry from the section-wise path. I know it's nice when things "just work" for the user, but I don't love the idea that the resampling method for a chunk can change silently because of a rare numerical failure. 

> 4. The methods to handle time-gaps in the recording is really elegant. In the wider-picture of SI though, I wonder if it is necessary to support this mode, and if the module can become more readable by stripping it out. The reason being that AFAIK SI makes no guarantees on supporting gaps in the `time_vector` for preprocessing, and most steps do not consider it nor work properly for it (e.g. filtering steps, or even `decimate.py`). My understanding is that in SI time gaps should be handled by the user by using segments or separate recordings. I think in the `resample.py` case we could just check for time gaps and error accordingly. That being said maybe there are important use cases for this I am overlooking.

My (and other people's) chronic, continuous, multi-day recordings commonly have small gaps. I am very much open to representing continuous stretches as segments or separate recordings, as long as it provides a supported route through preprocessing, sorting, postprocessing, analyzer objects, etc. Last I checked, adding limited qualified support for gaps within a segment was by far the easiest and least disruptive way to achieve this, even though SI does generally expect contiguous segments. It is probably worth mentioning that this expectation is rarely if ever checked or enforced, and there is explicit support for and discussion of missing frames and irregularly-sampled data scattered all over the place, e.g. in tutorials, tests, in the NWB recording extractor, the Blackrock reader, etc. I guess that is part of why the time vector exists, rather than always assuming a start time + fixed sample rate. So I would not say there is any contract in SI that says a segment must be gap-free and regularly-sampled, even if that is asserted in the docs once or twice. It is maybe most accurate to say that all operations are expected to work for gap-free, regularly-sampled segments, and support for imperfect segments depends on the operation. 

It is also worth mentioning that motion estimation, lots of sorter-specific processing, and timing restoration/persistence (e.g. from analyzers), and exporters (e.g. Phy) are not currently able to correctly deal with time-gapped multisegment recordings. Using separate recordings is technically possible, but then there needs to be machinery to handle independently assigned unit IDs across recordings, and many other things. 

All this is just to say that, yes, supporting a segment with gaps here is a little bit of an anti-pattern, and in my ideal world we would have a bit more clarity about what a segment actually is and isn't, but if we stop handling segments with gaps correctly in the few places where it is currently needed, that opens up a much wider discussion about how to add support back for these kinds of data. 

[^1]: Actually, we don't. Technically, the noninteger-ratio + time vector path can stretch the output timestamps a bit. For example, if someone gives 1000 samples at 1000 Hz with explicit regular timestamps, and asks for a 333.3 Hz output, they get back 333 points spanning 0 to ~0.999 s, which implies a rate slightly less than 333.3 Hz. But the output rate we report back to the user is 333.3 Hz. 🤫 .

src/spikeinterface/preprocessing/resample.py
grahamfindlay: yes, I would prefer that -- will do

src/spikeinterface/preprocessing/decimate.py
grahamfindlay: I think I just copied this margin code from what was already used elsewhere in the codebase, for consistency. 
Yes, it is definitely worth checking. I think the code there that produced the table uses Chebyshev type I, not Bessel, and measures a single stage rather than the cascade. But in any case, I would test the actual cascade and then choose an automatic margin (or a warning threshold). Probably also makes sense to retain some kind of override. Will do.

src/spikeinterface/preprocessing/_decimation_tools.py
grahamfindlay: Yes, will do

grahamfindlay:
@JoeZiminski Just pushed a batch of commits that hopefully address all of the above. The first 9 commits (through [7d1f053](https://github.com/SpikeInterface/spikeinterface/pull/4621/commits/7d1f0538fca09907473892fa07e63f6c9cfb8646)) keep the IIR/FFT filtering/resampling. The 10th and final commit clobbers most of that and replaces it with the FIR/polyphase resampling. So the last commit could be rejected if someone is reluctant to change.

src/spikeinterface/preprocessing/decimate.py
```diff
@@ -113,25 +161,106 @@ def __init__(
...
+    parent_end_frame = decimation_offset + ((end_frame + up - 1) // up) * down
+    parent_traces, left_margin, _ = get_chunk_with_margin(
+        parent_segment,
+        parent_start_frame,
+        parent_end_frame,
+        channel_indices,
+        margin,
+        add_reflect_padding=True,
+    )
+    working_dtype = np.result_type(parent_traces.dtype, dtype, np.float32)
```
JoeZiminski: cool function!

src/spikeinterface/preprocessing/decimate.py
```diff
@@ -113,25 +161,106 @@ def __init__(
...
+            1,
+            self._decimation_factor,
+            self._margin,
+            self._dtype,
+            self._filter_coefficients,
+            decimation_offset=self._decimation_offset,
+        )
+
+
+def get_polyphase_resampled_traces(
```
JoeZiminski: could have a quick docstring here mostly on the parameters

src/spikeinterface/preprocessing/decimate.py
```diff
@@ -113,25 +161,106 @@ def __init__(
...
+
+    dtype = np.dtype(dtype)
+    if np.issubdtype(dtype, np.integer):
+        rounded = np.rint(traces)
+        limits = np.iinfo(dtype)
+        below = rounded <= limits.min
+        above = rounded >= limits.max
+
+        # Assign saturated endpoints after casting because apparently float64 can't
+        # represent int64.max exactly.
```
JoeZiminski: interesting

src/spikeinterface/preprocessing/_resampling_tools.py
```diff
@@ -0,0 +1,58 @@
...
+    half_length = 10 * max(up, down)
+    coefficients = firwin(
+        2 * half_length + 1,  # odd length gives a symmetric filter with a central sample
+        1.0 / max(up, down),
+        window=("kaiser", 5.0),
+    )
+
+    margin = (half_length + up - 1) // up  # Convert filter support to input samples
+    if margin_ms is not None:
+        margin = max(margin, math.ceil(margin_ms * sampling_frequency / 1000))
```
JoeZiminski: Does it make sense to warn the user here if their requested margin is too small, and let them know it's been increased?

src/spikeinterface/preprocessing/_resampling_tools.py
```diff
@@ -0,0 +1,58 @@
...
+            f"Requested resample_rate={resample_rate:.16g} Hz; the polyphase ratio {up}/{down} "
+            f"achieves {achieved_rate:.16g} Hz ({error_ppm:+.6g} ppm). The output sampling frequency "
+            "uses the achieved rate. Increase max_denominator for a closer approximation.",
+            stacklevel=2,
+        )
+    return up, down, achieved_rate
+
+
+def get_polyphase_filter(sampling_frequency, up, down, margin_ms):
+    """Design SciPy's default FIR and cover its support on an aligned input grid."""
```
JoeZiminski: could extend this slightly, just to make clear 1) too-small margins will be rounded up 2) the margin size is guaranteed to be some multiple of the `down`

src/spikeinterface/preprocessing/_resampling_tools.py
```diff
@@ -0,0 +1,58 @@
...
+    from scipy.signal import firwin
+
+    if margin_ms is not None and (not math.isfinite(margin_ms) or margin_ms < 0):
+        raise ValueError("margin_ms must be finite and nonnegative, or None")
+
+    if up == down == 1:
+        return np.ones(1), 0
+
+    # Important! The multiplier 10 and Kaiser parameter 5.0 come directly from SciPy’s
+    # default resample_poly design. Don't change them!
```
JoeZiminski: Nice, maybe we could explicitly link to https://github.com/scipy/scipy/blob/30638aa87b3a5df4cf930affab76bb133af908af/scipy/signal/_signaltools.py#L4153 and mention in the docstring we compute the filter up front so we can old onto the coefficients and apply evenly across chunks

src/spikeinterface/preprocessing/resample.py
```diff
@@ -242,177 +225,69 @@ def __init__(
...
-            result[pos : pos + write_len] = trimmed[:write_len]
-            pos += write_len
 
-        # Return only the filled portion (normally equals end_frame - start_frame).
-        return result[:pos]
+def _resample_time_vector(parent_times, up, down, parent_rate):
+    """Map the rational sample grid onto timestamps without crossing section boundaries."""
+    if up == 1:
+        return parent_times[::down]
+    n_out = (len(parent_times) * up + down - 1) // down
```
JoeZiminski: this calculation is done in a few places, could this be a central function? (maybe on _resampling tools)

src/spikeinterface/preprocessing/decimate.py
JoeZiminski: does it make sense for this func to be in `resample.py` (with decimate a subset of resample)?

src/spikeinterface/preprocessing/tests/test_decimate.py
```diff
@@ -68,13 +81,96 @@ def test_decimate_with_times():
...
+    decimated_rec = DecimateRecording(rec, decimation_factor, decimation_offset=decimation_offset, antialias=antialias)
     for segment_index in range(rec.get_num_segments()):
         assert np.allclose(
             decimated_rec.get_times(segment_index),
             rec.get_times(segment_index)[decimation_offset::decimation_factor],
         )
 
 
+@pytest.mark.parametrize("factor", [1, 7, 17, 48, 300])
+def test_decimate_polyphase(factor):
```
JoeZiminski: nice, could this also test a multi-segment recording for completion?

src/spikeinterface/preprocessing/tests/test_decimate.py
```diff
@@ -68,13 +81,96 @@ def test_decimate_with_times():
...
+        parent_rec, decimation_factor, decimation_offset=decimation_offset, antialias=False, dtype="float32"
+    )
+
+    np.testing.assert_allclose(dec_aa.get_times(), parent_rec.get_times()[decimation_offset::decimation_factor])
+
+    # The anti-aliasing path returns the same number of samples as plain slicing.
+    parent_n = parent_rec.get_num_samples()
+    expected_n = int(np.ceil((parent_n - decimation_offset) / decimation_factor))
+    assert dec_aa.get_num_samples() == expected_n
+    assert dec_aa.get_num_samples() == dec_plain.get_num_samples()
```
JoeZiminski, line 150: for exactness tests could be offset by say 10 samples and decimate by a factor of 10 and check the data are a perfect match bar the first sample?

JoeZiminski (approved):
Hi @grahamfindlay cheers this is very nice, I have left a few scattered comments but please feel free to ignore. I agree with all points you mentioned above, and that's a good point about the small gaps, I didn't realise this was such a common occurrence! Agree this is a wider discussion for SI and how this case is handled, and it makes sense to support it here.

Thanks this is a nice addition and was fun to review!

JoeZiminski:
It seems like the chunk tests are failing on macOS, though the error is still very small - maybe we can increase the thresholds?

src/spikeinterface/preprocessing/_resampling_tools.py
grahamfindlay: That's what I did at first, but then the user gets warnings even with a "legitimately" set `margin_ms=0`. So I made the docstring non-committal about obeying `margin_ms` instead:

```
    margin_ms : float | None, default: None
        Additional context in ms on each side of a chunk. If None, use the FIR filter's
        finite support. An explicit nonnegative value requests at least that much context;
        filter support is always retained within each section.
```

But you are right, it probably makes sense to either make `margin_ms` a _hard_ override, or to warn. Will add the warn.

src/spikeinterface/preprocessing/decimate.py
grahamfindlay: 🤦 definitely.

grahamfindlay:
@JoeZiminski should be good to go now

samuelgarcia:
Hi @grahamfindlay and @JoeZiminski
This is very very cool.
Thanks a lot for this new implementation of this decimate resample.
Thank you also to handle the time vector that was in place. I totally forgot about it.
At some point we should revist it because it force the time vector to be loaded in memory and for very long recording this is a problem. Alessio is figthinh to make lazy.
I propose to merge this impressive PR soon and rediscuss the time vector for resampling when we will continue to improve the time vector lazyness.

src/spikeinterface/preprocessing/tests/test_decimate.py
```diff
@@ -68,13 +82,100 @@ def test_decimate_with_times():
...
+
+@pytest.mark.parametrize("decimation_factor", [6, 10, 48, 300])
+def test_decimate_antialias_by_chunks(decimation_factor):
+    sampling_frequency = int(3e4)
+    duration = 30
+    traces, _ = create_sinusoidal_traces(sampling_frequency, duration, freqs_n=10, max_freq=1000, dtype=np.float32)
+    parent_rec = NumpyRecording(traces, sampling_frequency)
+    decimated_rate = sampling_frequency / decimation_factor
+
+    for margin_ms in [None, 100, 1000]:
```
samuelgarcia, line 121: Is 100ms enougth for a downsampling of 300 ?

src/spikeinterface/preprocessing/tests/test_decimate.py
```diff
@@ -68,13 +82,100 @@ def test_decimate_with_times():
...
+        rec2 = DecimateRecording(parent_rec, decimation_factor, antialias=True, margin_ms=margin_ms)
+        chunk_size = int(decimated_rate * 2)  # ~2 seconds of the decimated signal
```
samuelgarcia: should we stress the test more with smaller chunks ?

src/spikeinterface/preprocessing/tests/test_resample.py
```diff
@@ -155,74 +156,68 @@ def test_resample_freq_domain():
...
+            )
+            assert processed._kwargs["resample_rate"] == requested_rate
+            assert processed._kwargs["max_denominator"] == max_denominator
+            with pytest.warns(UserWarning, match="achieves") if approximate else nullcontext():
+                restored = load(processed.to_dict())
+            assert restored.get_sampling_frequency() == processed.get_sampling_frequency()
+            np.testing.assert_array_equal(restored.get_traces(), processed.get_traces())
+
+
+def test_resample_short_sections():
```
samuelgarcia, line 207: impressive

src/spikeinterface/preprocessing/tests/test_decimate.py
grahamfindlay: Yup: the filter half-length is 10x the factor, so for a factor of 300 the support is 3k samples, which is 100 ms at 30 kHz (also, if the requested margin were too small, the code would increase it and warn).

src/spikeinterface/preprocessing/tests/test_decimate.py
grahamfindlay: Sure, can do. Most stressful test would be to test chunks smaller than the margin, I suppose.

grahamfindlay:
@samuelgarcia small-chunk stress test added! thank you!

alejoe91 (approved):

alejoe91:
Amazing effort @grahamfindlay and @JoeZiminski !

PR #4623, opened 2026-06-22
Lazy-mode for loading `Sortinganalyzer`

alejoe91:
The `SortingAnalyzer`, especially for chronic data, can be very memory hungry. This PR adds a `lazy` mode at load that:
- Reads the spike vector in lazy mode (memmaps memory folder, virtual memmap for zarr - TODO)
- Reads vector-extensions as memmaps or keeps them as zarr array
- Implements a memmap-like `ZarrSpikeVector`

**NOTE**: in lazy mode, save/select/merge/split operations are not permitted!


src/spikeinterface/core/sortingfolder.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, folder_path):
         folder_metadata = folder_path
         self.load_metadata_from_folder(folder_metadata)
 
-        self._kwargs = dict(folder_path=str(folder_path.absolute()))
+        self._kwargs = dict(folder_path=str(folder_path.absolute()), mmap_mode=mmap_mode)
```
chrishalcrow, line 50: When are the `_kwargs` large? Do they need to be lazy?

alejoe91: They are not. This is the kwargs to instantiate the Sorting object!

chrishalcrow:
Hello, when I do a `get_data()` on any of the spike-vector extensions (e.g. spikeamplitudes) on a zarr'd analyzer I get an error here:

https://github.com/SpikeInterface/spikeinterface/blob/ecf5c6d60933b52b8b875780b3bc5ef820e8a931/src/spikeinterface/core/analyzer_extension_core.py#L1622

as follows:

```
  File "<python-input-10>", line 1, in <module>
    sa.get_extension('spike_amplitudes').get_data()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/nolanlab/fromgit/spikeinterface/src/spikeinterface/core/sortinganalyzer.py", line 3465, in get_data
    return self._get_data(*args, **kwargs)
           ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/nolanlab/fromgit/spikeinterface/src/spikeinterface/core/analyzer_extension_core.py", line 1622, in _get_data
    return all_data.copy()  # return a copy to avoid modification
           ^^^^^^^^^^^^^
AttributeError: 'Array' object has no attribute 'copy'
```

alejoe91:
> .get_data()

So when it's lazy we should force copy=False. Does it make sense?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -3123,10 +3165,10 @@ def load_data(self):
...
                 elif ext_data_file.suffix == ".npy":
                     # The lazy loading of an extension is complicated because if we compute again
-                    # and have a link to the old buffer on windows then it fails
-                    # ext_data = np.load(ext_data_file, mmap_mode="r")
-                    # so we go back to full loading
-                    ext_data = np.load(ext_data_file)
+                    # and have a link to the old buffer on windows then it fails.
+                    # So, by default, we use full loading, but lazy can be requested on demand.
+                    kwargs = dict(mmap_mode="r") if lazy else dict()
+                    ext_data = np.load(ext_data_file, **kwargs)
```
```suggestion
                    if lazy:
                        ext_data = np.load(ext_data_file, mmap_mode="r")
                    else:
                        ext_data = np.load(ext_data_file)
```

I just find this a lot easier to read

src/spikeinterface/core/sortingfolder.py
```diff
@@ -47,8 +47,7 @@ def __init__(self, folder_path: str | Path):
         # Load metadata
         self.load_metadata_from_folder(folder_path)
 
-        # Save folder_path as kwargs for serialization
-        self._kwargs = {"folder_path": str(folder_path.absolute())}
+        self._kwargs = dict(folder_path=str(folder_path.absolute()), mmap_mode=mmap_mode)
```
chrishalcrow, line 50: Can you explain why we want to include `mmap_mode` here?

alejoe91: If I open in memmap mode and then parallelize I probably want to keep the same mode, so we need it in the kwargs

chrishalcrow (approved):
Works for me, and doesnt't change any defaults.
Note: it's pretty conservative. If you use lazy mode, you can't save/recompute/anything!

samuelgarcia:
I would have been very happy to discuss this PR in detail before the merge
For perf reason mainly.
The ZarrSpikeVector would have deserve some deep benchmark no ?

PR #4624, opened 2026-06-23
Remove `peak_sign` in favour of `main_channel_indices`

chrishalcrow:
Sequel to #4374

I didn't have permission to push to Sam's branch, so I made my own...

Idea is that when you make your analyzer, it *must* have a `peak_sign`, `peak_mode` and hence a `main_channel_index`. We then use these throughout the codebase instead of propagating the other stuff.

Some design things:
- If you have a `SortingAnalyzer`, you set `peak_sign`, `peak_mode` and hence a `main_channel_index` immediately (or ASAP). Then none of the `get_main_template` type functions are allowed to receive these args. BUT if you have a `Template`, you _must_ explicitly pass these args. So we're trying to make it simple for the analyzer (for users) but explicit for templates (for developers)

Cool things:
- `peak_sign` no longer appears in metrics.

@samuelgarcia did the tedious work, but now I'm refining and getting the tests to pass...


src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -380,7 +381,7 @@ class ComputeTemplates(AnalyzerExtension):
 
     extension_name = "templates"
     depend_on = ["random_spikes|waveforms"]
-    need_recording = True
+    need_recording = False
```
chrishalcrow, line 384: If you have waveforms, you do not need the Recording. Not sure how to deal with this.

chrishalcrow:
Hey @ecobost , I went through your comments on #4374 and mostly resolved them (sorry it's on a new PR, I didn't have permission to push to Sam's branch). Here's a list of "controversial" things to discuss.

- I think I've been too hrash getting rid of peak_sign and peak_mode in `estimate_sparsity`. This was actually used for Templates sensibly... I should re-insert... I'll do this soon!!
- @ecoboost pointed this out: at the moment, the `main_channel_indices` are a sorting property, and (ideally) the analyzer gets it from there. But in many cases the analyzer computes the `main_channel_indices`, and it's then set as a sorting property. In this case, the sorting inherits from the analyzer, which sounds bad. How to resolve?
- `estimate_sparisty` default is now "both", matching `create_sorting_analyzer`. It's only used in `create_sorting_analyzer`
- @ecobost suggests removing sparsity_kwargs from estimate sparsity. I'm against this, but mostly because I use them, which isn't a great reason.
- It's not clear if we should allow a user to pass their own "peak_sign" to e.g. `get_template_amplitudes`, because this is usually a bad idea. I suggest adding a `override_peak_error` flag, so that expert users have to turn this on to mess with it. I think @alejoe91  prefers to just allow the user to pass it.
- I've added `peak_sign_for_signal` and `peak_mode_for_signal` in compute_snrs. I think these have use cases: I can imagine people wanting to compute the peak_to_peak value on main_channel.
- At the moment, we're deprecating `get_template_extremum_amplitude`, `get_template_extremum_channel_peak_shift`, `get_template_extremum_channel` in favour of similarly named functions (but with main_channel). I'm a bit worried about this - our lab uses `get_template_extremum_channel` in various places and I'd bet it's used by other developers. Maybe we should just keep the old naming convention and use `extremum_channel_indices`. Or just keep `get_template_extremum_channel` for a while??

@samuelgarcia @alejoe91 could you take a look at this - would be great to get merged next week!!

src/spikeinterface/core/analyzer_extension_core.py
alejoe91: we could do as we do with optional dependencies, i.e., having a function that evaluates if this is needed on the fly based on othere available extensions?

src/spikeinterface/core/basesorting.py
```diff
@@ -13,6 +13,10 @@ class BaseSorting(BaseExtractor):
     Abstract class representing several segment several units and relative spiketrains.
     """
 
+    _main_properties = [
+        "main_channel_index",
```
alejoe91: let's be careful because channel indices could change (e.g., build analyzer with/without bad channels).. how should we deal with it?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -44,21 +44,30 @@
...
+from .sorting_tools import random_spikes_selection
+
 
 # high level function
 def create_sorting_analyzer(
     sorting,
     recording,
     format="memory",
     folder=None,
+    main_channel_index=None,
```
```suggestion
    main_channel_indices=None,
```
?

src/spikeinterface/core/basesorting.py
chrishalcrow: We decided that the sorting property should be `main_channel_id`, but the internal analyzer property should be `main_channel_index`

src/spikeinterface/core/sortinganalyzer.py
chrishalcrow: We agreed to pluralize `main_channel_index` to `main_channel_indices`

src/spikeinterface/sorters/internal/spyking_circus2.py
```diff
@@ -370,7 +370,9 @@ def _run_from_folder(cls, sorter_output_folder, params, verbose):
                     job_kwargs=job_kwargs,
                 )
 
-                sparsity = compute_sparsity(dense_templates, method="radius", radius_um=radius_um)
+                sparsity = compute_sparsity(
+                    dense_templates, method="radius", radius_um=radius_um, peak_sign="neg", amplitude_mode="extremum"
```
chrishalcrow, line 374: `compute_sparsity` defaults have changed. This edit ensures SK2 keeps old defaults.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -525,12 +630,151 @@ def create_binary_folder(cls, folder, sorting, recording, sparsity, return_in_uV
...
+                main_channel_indices = self.recording.ids_to_indices(sorting_main_channel_ids)
+                self._main_channel_indices = main_channel_indices
+            else:
+                self._main_channel_indices = self.compute_main_channel_backwards_compatibility()
+                main_channel_ids = self.channel_ids[self._main_channel_indices]
+                self.set_sorting_property("main_channel_id", main_channel_ids, save=False)
+
+        return self._main_channel_indices
+
+    def compute_main_channel_backwards_compatibility(self):
```
alejoe91: Maybe private?
```suggestion
    def _compute_main_channel_backwards_compatibility(self):
```

src/spikeinterface/core/analyzer_extension_core.py
alejoe91: we decided to keep `False` and only check if a recording is available when computing templates and no waveforms are available

src/spikeinterface/postprocessing/localization_tools.py
```diff
@@ -77,7 +83,13 @@ def compute_monopolar_triangulation(
...
 
     if sorting_analyzer_or_templates.sparsity is None:
-        sparsity = compute_sparsity(sorting_analyzer_or_templates, method="radius", radius_um=radius_um)
+        sparsity = compute_sparsity(
+            sorting_analyzer_or_templates,
+            method="radius",
+            radius_um=radius_um,
+            peak_sign=peak_sign,
+            amplitude_mode=peak_mode,
+        )
```
chrishalcrow, line 92: @samuelgarcia this feels a bit weird. Should we really be computing sparsity here?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -44,21 +44,30 @@
...
+from .sorting_tools import random_spikes_selection
+
 
 # high level function
 def create_sorting_analyzer(
     sorting,
     recording,
     format="memory",
     folder=None,
+    main_channel_indices=None,
```
samuelgarcia, line 54: alessio wanted main_channel_ids here no ?
lets have the two ?

alejoe91: No, this is ok since you are also passing the recording. So it's not ambiguous. My main issue was in how we store the property I'm the sorting object, which is now correctly using main_chamnel_ids

alejoe91 (approved):

src/spikeinterface/postprocessing/localization_tools.py
chrishalcrow: Note: `localization_tools` is the only place in `postprocessing` where `peak_sign` appears (apart from backwards compat stuff)

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -51,14 +51,20 @@ def create_sorting_analyzer(
...
+    num_spikes_for_main_channel=100,
     sparse=True,
     sparsity=None,
     set_sparsity_by_dict_key=False,
     return_scaled=None,
     return_in_uV=True,
     overwrite=False,
     backend_options=None,
-    **sparsity_kwargs,
+    sparsity_kwargs=None,
```
chrishalcrow, line 65: @samuelgarcia this breaks any code that was passing sparsity kwargs directly (including my lab's pipeline!!) Let's deprecate for a while?

PR #4636, opened 2026-07-01
Add cross-version serialization tests

h-mayorquin:
SpikeInterface has no test that loads an object serialized by an older release with the current code, even though that is exactly the path that breaks users. We only round-trip within a single version, which never exercises a class move, a renamed constructor argument, or a change in how something is written to disk. Two recent cases make the gap concrete. On #4588, Sam pointed out that moving `NoiseGeneratorRecording` out of `core` could break any old JSON or pickle that stored its fully qualified class path. The probe refactor in #4465 changes how a probe is written to disk (`contact_vector` property to a `ProbeGroup` object), which risks old saved recordings either failing to load or, worse, silently mis-mapping channels to locations. Nothing in continuous integration would catch either today.

This adds a harness that generates a small battery of canonical objects under an old release and loads them with the branch, asserting content rather than only that they load. Fixtures are regenerated from a real old install each run in an isolated uv environment, so nothing is committed to the repository. The battery covers both failure axes: JSON recipes catch moved classes and changed signatures (the #4588 case, where the fixture records the old `core.generate.NoiseGeneratorRecording` path and loading exercises the redirect), while folder and zarr fixtures catch on-disk encoding changes. Those include a single-probe and an interleaved multi-probe recording that assert exact per-channel locations, so a silent channel-to-location scramble (the case that motivated `_global_contact_order` in #4465) fails the test instead of passing on shape alone. It runs against the previous release on every pull request touching `core`, and the full version matrix weekly.

I set the version floor at 0.102 because 0.100 and 0.101 no longer install cleanly on the continuous-integration Python (numcodecs dependency rot). This makes "testable floor equals support floor" a deliberate choice worth confirming with the team rather than an accident.


.github/scripts/serialization/objects.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def _build_noise_generator_recording():
+    # NoiseGeneratorRecording moved from spikeinterface.core.generate to
+    # spikeinterface.generation. Import portably; the serialized "class" path is
+    # whatever the running version records, which is what exercises the redirect.
+    try:
+        from spikeinterface.generation import NoiseGeneratorRecording
+    except ImportError:
+        from spikeinterface.core.generate import NoiseGeneratorRecording
```
zm711, line 47: instead could we do something like
```python
if importlib.util.find_spec('spike interface.generation') is not None:
    from spikeinterface.generation import NoiseGeneratorRecording
else:
    from spikeinterfacde.core.generate import NoiseGeneratorRecording
```
This way we don't worry about try-except. Maybe you'll say the cost isn't high enough?

.github/scripts/serialization/objects.py
```diff
@@ -0,0 +1,227 @@
...
+
+def _build_recording_with_probe():
+    import numpy as np
+    from probeinterface import generate_linear_probe
+    from spikeinterface.core import generate_recording
+
+    rec = generate_recording(num_channels=8, durations=[1.0], sampling_frequency=30000.0, seed=0)
+    probe = generate_linear_probe(num_elec=8)
+    probe.set_device_channel_indices(np.arange(8))
+    return rec.set_probe(probe)  # old API returns a new recording; portable across versions
```
zm711: Just because I didn't follow the in-place stuff is this way still safe since we will have some versions that will just return `None`.

zm711 (commented):
One weak request and one question in general.

.github/scripts/serialization/objects.py
alejoe91: since this doesn't affect import time and the generation module always exists, I would keep this as is

.github/scripts/serialization/objects.py
alejoe91: I made it version dependent

alejoe91 (approved):

Issue #4553, opened 2026-04-22
Probe attaching: do we really need a dense representation?


h-mayorquin:
@samuelgarcia 

I've been thinking more about whether SpikeInterface needs a dense channel-aligned representation. The two arguments I've heard for it are ergonomics and making slicing and aggregation easier. Let me go through them.

## For channel-position indexing

The idea is that a dense channel-aligned `contact_vector` lets `get_channel_locations` resolve locations by channel position with one numpy index. Here is how it looks today on main ([baserecordingsnippets.py#L348-L366](https://github.com/SpikeInterface/spikeinterface/blob/651c8ea2faa0c4a27397ed7e68d656d35e083141/src/spikeinterface/core/baserecordingsnippets.py#L348-L366)):

```python
def get_channel_locations(self, channel_ids=None, axes: str = "xy") -> np.ndarray:
    if channel_ids is None:
        channel_ids = self.get_channel_ids()
    channel_indices = self.ids_to_indices(channel_ids)
    contact_vector = self.get_property("contact_vector")
    if contact_vector is not None:
        ndim = len(axes)
        all_positions = np.zeros((contact_vector.size, ndim), dtype="float64")
        for i, dim in enumerate(axes):
            all_positions[:, i] = contact_vector[dim]
        positions = all_positions[channel_indices]
        return positions
    else:
        locations = self.get_property("location")
        if locations is None:
            raise Exception("There are no channel locations")
        locations = np.asarray(locations)[channel_indices]
        return select_axes(locations, axes)
```

This makes sense and looks compact, but **this is the only place** in SpikeInterface where a dense channel-aligned representation is used for numpy-like indexing by channel position. Bringing the full dense-array machinery into `ProbeGroup` (as [probeinterface#420](https://github.com/SpikeInterface/probeinterface/pull/420) proposes) feels like too much infrastructure for one function.

Reviewing the code base, I found that every other reader of `contact_vector` is work the format creates for itself. `get_probegroup` rebuilds the `ProbeGroup` via `from_numpy` only because the dense form is what gets stored; storing the `ProbeGroup` object directly (as [#4548](https://github.com/SpikeInterface/spikeinterface/pull/4548) does) makes that read disappear. The other big user is slicing and aggregation, discussed below, which is also bookkeeping the format creates for itself.

[...]

Issue #4549, opened 2026-04-21
`aggregate_channels` followed by `split_by` cannot recover independent probe recordings


h-mayorquin:
Aggregating two independent recordings with a probe attached and then splitting back by `group` should return two sub-recordings, one per original probe. Instead it returns a single sub-recording containing all channels merged under one synthetic probe, with no probe names, manufacturers, or contours. The two input recordings cannot be recovered from the round-trip. Unlike the reverse direction (`split_by` then `aggregate_channels`, which preserves contact-level data and loses only annotations), this direction destroys the probe structure itself.

```python
import numpy as np
from probeinterface import generate_dummy_probe
from spikeinterface.core import generate_recording, aggregate_channels

def make_rec(probe_name, manufacturer, x_shift, contact_id_prefix):
    probe = generate_dummy_probe()
    probe.move([x_shift, 0.0])
    probe.annotate(name=probe_name, manufacturer=manufacturer)
    probe.set_device_channel_indices(np.arange(probe.get_contact_count()))
    probe.set_contact_ids([f"{contact_id_prefix}{i}" for i in range(probe.get_contact_count())])
    probe.create_auto_shape()
    rec = generate_recording(num_channels=probe.get_contact_count(), durations=[1.0], set_probe=False)
    return rec.set_probe(probe)

rec_A = make_rec("probe_A", "vendor_X", 0.0, "a")
rec_B = make_rec("probe_B", "vendor_Y", 1000.0, "b")

combined = aggregate_channels([rec_A, rec_B])
parts = combined.split_by("group")

print(f"number of sub-recordings after split: {len(parts)}")
for k, sub in parts.items():
    probes = sub.get_probegroup().probes
    print(f"  split[{k}]: n_channels={sub.get_num_channels()}, n_probes={len(probes)}, "
          f"probe_0_name={probes[0].annotations.get('name')}")
```

Expected: two sub-recordings, each with 32 channels and its original probe (`probe_A` or `probe_B`).

Observed on `main`:

```
number of sub-recordings after split: 1
  split[0]: n_channels=64, n_probes=1, probe_0_name=None
```

The two input recordings cannot be separated back out, and the synthetic merged probe carries no identifying metadata.

h-mayorquin:
[...]

Issue #4547, opened 2026-04-21
`select_channels` mislabels surviving probe when a full probe is sliced out


h-mayorquin:
When `select_channels` drops all channels of some probe from a multi-probe recording, the per-probe annotations (`probes_info` and each `probe_{i}_planar_contour`) are carried over verbatim rather than pruned to match the surviving probes. `get_probegroup` then pairs those stale annotations with the surviving probes by enumeration order, so after dropping probe A from a recording, the one surviving probe ends up labeled with probe A's name, manufacturer, and planar contour even though its channels all came from probe B. Unlike the `aggregate_channels` case where probe metadata is dropped outright, here it is retained but mis-attributed, which is harder to notice.

```python
import numpy as np
from probeinterface import generate_dummy_probe, ProbeGroup
from spikeinterface.core import generate_recording

def make_probe(name, manufacturer, x_shift, contact_id_prefix):
    p = generate_dummy_probe()
    p.move([x_shift, 0.0])
    p.annotate(name=name, manufacturer=manufacturer)
    p.set_contact_ids([f"{contact_id_prefix}{i}" for i in range(p.get_contact_count())])
    p.create_auto_shape()
    return p

p_A = make_probe("probe_A", "vendor_X", 0.0, "a")
p_B = make_probe("probe_B", "vendor_Y", 1000.0, "b")

pg = ProbeGroup()
pg.add_probe(p_A)
pg.add_probe(p_B)
dci = np.arange(pg.get_contact_count())
p_A.set_device_channel_indices(dci[:p_A.get_contact_count()])
p_B.set_device_channel_indices(dci[p_A.get_contact_count():])

rec = generate_recording(num_channels=pg.get_contact_count(), durations=[1.0], set_probe=False)
rec = rec.set_probegroup(pg)

# Drop all of probe A's channels, keep only probe B
keep_ids = rec.get_channel_ids()[p_A.get_contact_count():]
sliced = rec.select_channels(keep_ids)

probes = sliced.get_probegroup().probes
print(f"n probes: {len(probes)}")
for i, p in enumerate(probes):
    print(f"  probe {i}: name={p.annotations.get('name')}, manufacturer={p.annotations.get('manufacturer')}")
```

Expected: the surviving probe is labeled with probe B's annotations.

Observed on `main`:

```
n probes: 1
  probe 0: name=probe_A, manufacturer=vendor_X
```

[...]

PR #4638, opened 2026-07-02
Use `ProbeGroup` object instead of `contact_vector` property

### LINKED ISSUE #4545: `aggregate_channels` loses per-probe metadata

alejoe91:
Replaces #4465 (as suggested by @h-mayorquin )

After https://github.com/SpikeInterface/probeinterface/pull/446, we can go all in on the use of `ProbeGroup` instead of `contact_vector`, since the use case with interleaved contacts after device channel indices slicing is handled gracefully by the new `_global_contact_order` array, stored in the probegroup dictionary. 
Now the `recording` has a `_probegroup` attribute, which is propagated as metadata. The `probegroup.get_slice` takes care of all metadata propagation. 
In this refactoring, the duplicated `location` property has also been dropped.

Note that this implementation is alternative to what is proposed in  #4553 and implemented in #4548 : here we keep a `ProbeGroup` object attached to the recording, after sorting it by `device_channel_indices`, instead of adding a `wiring` property. You lose the original probe contact order, but we argue that the contacts in a probegroup do not really have a "default" order, but are rather a set of contacts that become ordered when wired (see #3498).

Some additional points:

### Tests for aggregation/slicing

Added more tests for aggregation/slicing behavior and metadata propagation

Fixes #4545 #4546 #4547 #4549 #4553

### Tests for backward compatibility

Added a test for bakward compatibility, which saves a recording to binary/zarr with v0.104.* and reloads it with the current version to check that the probegroup is reloaded correctly.


src/spikeinterface/core/baserecordingsnippets.py
```diff
@@ -51,242 +53,262 @@ def has_scaleable_traces(self) -> bool:
...
     def has_channel_location(self) -> bool:
-        return self.has_probe() or "location" in self.get_property_keys()
+        return self.has_probe()
 
     def is_filtered(self):
         # the is_filtered is handle with annotation
         return self._annotations.get("is_filtered", False)
 
-    def set_probe(self, probe, group_mode="auto", in_place=False):
+    def reset_probe(self):
```
chrishalcrow, line 73: I think a reset is when you remove then add something all at once. I would prefer something like `remove_probe` for this.

alejoe91: fixed in #4465

alejoe91:
Closing in favor of #4465

PR #4643, opened 2026-07-06
[Draft] Governance

zm711:
I'm starting to map out governance stuff that we will need to flesh out eventually. This is based on Neo's general governance pre-NumFocus.


chrishalcrow:
Beautiful!

PR #4646, opened 2026-07-07
Infer `sampling_frequency` from `Units.resolution` in `NwbSortingExtractor`

h-mayorquin:
An NWB `Units` table stores spike times in seconds, so the extractor needs a `sampling_frequency` to convert them to frames. Until now it came only from an explicit argument or an `ElectricalSeries`, so a units-only file (common on DANDI) failed unless the user passed the rate by hand, even though such files usually carry it in `Units.resolution` (`1 / sampling_rate`). This reads that attribute, ranked above the `ElectricalSeries` because it is spike-native and cannot be misled by a series the sorting was not computed from (an LFP series next to a full-band sorting); `t_start` is unchanged.

### DIFF
--- a/src/spikeinterface/extractors/nwbextractors.py
+++ b/src/spikeinterface/extractors/nwbextractors.py
@@ -571,7 +571,7 @@ class _NWBReader:
             return {column.name: column for column in self.units_table.columns}["spike_times_index"].data
         return self.units_table["spike_times_index"]
 
-    def sampling_frequency_from_units_metadata(self):
+    def _sampling_frequency_from_units_metadata(self):
         # The Units.resolution attribute documents the precision of the spike times, typically
         # 1 / sampling_rate (in seconds). When present it lets a units-only file (one with no
         # ElectricalSeries) report a sampling frequency. It is spike-native, so it is preferred over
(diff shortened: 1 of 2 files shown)


src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -559,6 +559,34 @@ def spike_times_index(self):
...
+
+    def has_electrical_series(self):
+        # Whether the file contains any ElectricalSeries at all. Used to decide, for a sorting, whether
+        # there is a recording to align to: if there is none, t_start has no origin and defaults to 0.
+        if self.reading_method == "use_pynwb":
+            from pynwb.ecephys import ElectricalSeries
+
+            return any(isinstance(item, ElectricalSeries) for item in self.nwbfile.all_children())
+        backend = "zarr" if self.reading_method == "use_zarr" else "hdf5"
+        return len(_find_neurodata_type_from_backend(self.file, neurodata_type="ElectricalSeries", backend=backend)) > 0
```
zm711: I thought you were a fan of clear code vs squashing things on one line. This returns a bool right? But you don't realize until you see the > 0 at the end.

h-mayorquin: I agree with you. I did this by bot and I did not pay attention to this. I will change it.

alejoe91 (approved):

PR #4647, opened 2026-07-07
Fix NwbSortingExtractor unit-property loading for DynamicTableRegion columns

h-mayorquin:
`NwbSortingExtractor` raised an error when a Units table stored a column as a DynamicTableRegion, for example the `max_electrode` column that the International Brain Laboratory (IBL) writes as a per-unit reference into the electrodes table. On the pynwb read path, indexing such a column resolves the region to a pandas DataFrame of the referenced electrode rows instead of returning the raw per-unit indices, and the property loader then iterated that DataFrame, which yields its column labels, so `set_property` received one value per electrodes-table column rather than one per unit and failed the size check. The default hdf5 (and zarr) path was unaffected, because those backends return the raw index array, so the two read paths disagreed. The fix teaches the reader's shared column materializer to recognize DynamicTableRegion columns and read their raw per-row indices on every backend, so pynwb now matches hdf5 and zarr.

Continuing the mission of the `_NWBReader` refactor (#4630), I moved unit-property fetching off `NwbSortingExtractor` and onto `_NWBReader`, so the reader stays the single place responsible for returning columns in a backend-independent format, the same way the recording extractor already delegates electrode properties through `read_electrode_property`. This was also the natural home for the fix: placing the region handling in the reader's `_read_column` means electrode and unit columns get identical treatment across pynwb, hdf5, and zarr and keeps all backend branching inside the reader. I picked reading the raw indices over skipping region columns so the information (each unit's referenced electrode) survives as a normal per-unit property. The regression test builds a Units table with a `max_electrode` DynamicTableRegion column and asserts the loaded values are the exact per-unit indices, parametrized over both read paths.

### DIFF
--- a/src/spikeinterface/extractors/nwbextractors.py
+++ b/src/spikeinterface/extractors/nwbextractors.py
@@ -425,25 +425,21 @@ class _NWBReader:
         column = table[name]
         # A DynamicTableRegion column (e.g. IBL's `max_electrode`) holds a per-row index into another
         # table. On the pynwb path `column[:]` resolves the region into a DataFrame of the referenced
-        # rows; read the raw per-row indices instead so every backend returns the same plain array.
-        if self._column_is_dynamic_table_region(column):
-            column = column.data if self.reading_method == "use_pynwb" else column
+        # rows, so read the raw per-row indices (`column.data`) instead. The raw hdf5/zarr dataset is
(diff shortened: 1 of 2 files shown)


alejoe91 (approved):
LGTM

h-mayorquin:
Removed the redundant inline behavior that @alejoe91 pointed out in person.

PR #4648, opened 2026-07-07
Add `SortingAnalyzer.select_channels()` function

alejoe91:
This PR adds a `select_channels` function to the `SortingAnalyzer` that returns a new analyzer in memory with sliced recording (or attirbutes), probegroup, and properties, and extensions (with a new `_select_channels_extension_data`)

Only a few extensions are affected: `noise_levels` - `templates`

`_select_extension_data` is now renamed to `_select_units_extension_data`

### DIFF
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -564,7 +564,7 @@ class ComputeTemplates(AnalyzerExtension):
     def _select_channels_extension_data(self, channel_ids):
         keep_channel_indices = np.flatnonzero(np.isin(self.sorting_analyzer.channel_ids, channel_ids))
 
-        new_data = dict()
+        new_data = {}
         for key, arr in self.data.items():
             new_data[key] = arr[:, :, keep_channel_indices]
 
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -809,7 +809,7 @@ class ComputeNoiseLevels(AnalyzerExtension):
     def _select_channels_extension_data(self, channel_ids):
         # this does not depend on channels
         channel_indices = self.sorting_analyzer.channel_ids_to_indices(channel_ids)
-        return self.data["noise_levels"][channel_indices]
+        return dict(noise_levels=self.data["noise_levels"][channel_indices])
 
     def _merge_extension_data(
         self, merge_unit_groups, new_unit_ids, new_sorting_analyzer, keep_mask=None, verbose=False, **job_kwargs
--- a/src/spikeinterface/core/analyzer_extension_core.py
+++ b/src/spikeinterface/core/analyzer_extension_core.py
@@ -75,7 +75,7 @@ class ComputeRandomSpikes(AnalyzerExtension):
         params = dict(method=method, max_spikes_per_unit=max_spikes_per_unit, margin_size=margin_size, seed=seed)
         return params
 
-    def _select_extension_data(self, unit_ids):
+    def _select_units_extension_data(self, unit_ids):
         random_spikes_indices = self.data["random_spikes_indices"]
 
         spikes = self.sorting_analyzer.sorting.to_spike_vector()
@@ -243,7 +243,7 @@ class ComputeWaveforms(AnalyzerExtension):
         )
         return params
 
-    def _select_extension_data(self, unit_ids):
+    def _select_units_extension_data(self, unit_ids):
         # random_spikes_indices = self.sorting_analyzer.get_extension("random_spikes").get_data()
         some_spikes = self.sorting_analyzer.get_extension("random_spikes").get_random_spikes()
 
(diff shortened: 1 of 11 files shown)


src/spikeinterface/core/analyzer_extension_core.py
```diff
@@ -561,6 +561,15 @@ def _select_extension_data(self, unit_ids):
 
         return new_data
 
+    def _select_channels_extension_data(self, channel_ids):
+        keep_channel_indices = np.flatnonzero(np.isin(self.sorting_analyzer.channel_ids, channel_ids))
+
+        new_data = dict()
```
zm711: You love the function call :) It's technically very slightly faster to use a literal {} but doesn't really matter.

src/spikeinterface/core/sparsity.py
```diff
@@ -841,6 +836,14 @@ def estimate_sparsity(
         )
 
     elif method != "by_property":
+        if recording.get_probes() == 1:
+            # standard case
+            probe = recording.get_probe()
+        else:
+            # if many probe or no probe then we use channel location and create a dummy probe with all channels
```
zm711: so in this case we lose the probe group if we slice? Like I can imagine what if I want to use my analyzer to look at 100-300um deep but I still want to go across my probes and keep them separate?

zm711 (commented):
One question and one silly comment.

src/spikeinterface/core/analyzer_extension_core.py
chrishalcrow:
```suggestion
        new_data = {}
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1140,6 +1144,57 @@ def get_sorting_property(self, key: str, ids: Optional[Iterable] = None) -> np.n
...
+        else:
+            properties = self.rec_attributes.get("properties", {})
+            if key not in properties:
+                raise ValueError(f"Property {key} not found in recording attributes")
+            values = properties[key]
+            if ids is not None:
+                channel_ids = self.channel_ids
+                indices = [np.where(channel_ids == id)[0][0] for id in ids]
+                values = values[indices]
+            return values
```
chrishalcrow, line 1180: This is actually great!

src/spikeinterface/core/sparsity.py
chrishalcrow: This is just moving old code so that it only gets activated when needed: it creates a dummy probe for making temporary Templates to compute the sparsity and is not kept after that.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1518,6 +1572,65 @@ def select_units(self, unit_ids, format="memory", folder=None) -> "SortingAnalyz
...
+            new_rec_attributes = self.rec_attributes.copy()
+            new_rec_attributes["channel_ids"] = np.array(channel_ids)
+            # slice properties according to channel_ids
+            if "properties" in new_rec_attributes:
+                new_properties = {}
+                for key, values in new_rec_attributes["properties"].items():
+                    if len(values) == len(self.channel_ids):
+                        # only slice properties that have the same length as channel_ids
+                        channel_indices = [np.where(self.channel_ids == id)[0][0] for id in channel_ids]
+                        new_properties[key] = values[channel_indices]
```
chrishalcrow: `values` can be a list, which breaks this

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1518,6 +1572,65 @@ def select_units(self, unit_ids, format="memory", folder=None) -> "SortingAnalyz
...
+        Parameters
+        ----------
+        channel_ids : list or array
+            The channel ids to keep in the new SortingAnalyzer object
+
+        Returns
+        -------
+        analyzer :  SortingAnalyzer
+            The newly create sorting_analyzer with the selected channels
+        """
```
chrishalcrow, line 1591: Can we check that `channel_ids` are in `self.channel_ids` early on?

chrishalcrow (approved):
LGTM

samuelgarcia:
Camarade.
I think that you totally forgot to handle the select channels when the extension handle internally a sparsity.
For instance waveform or PCA.
In theses 2 cases the mechanistic is really much more complex and slow becaise we need to go row per row in the buffer.
We should either revert this PR either make the fix quiclky.

samuelgarcia:
There another red alert in the PR that have been merged to quickly for my taste with presuriasing some members.

At some place the select_channel do accept reverting the order for instance rec = rec.select_channel(new_chan_ids)
but in other place you are just doing a mask to reduce the data and then this new order is not preserved that make the new analyser inconsistant.

So either we handle it the same way either do not accept flippin channels, we need to be clear.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1518,6 +1572,70 @@ def select_units(self, unit_ids, format="memory", folder=None) -> "SortingAnalyz
...
+        # Check that all channel_ids are in the current channel_ids
+        if not np.all(np.isin(channel_ids, self.channel_ids)):
+            wrong_channel_ids = [ch for ch in channel_ids if ch not in self.channel_ids]
+            raise ValueError(f"Some channel_ids are not in the current channel_ids: {wrong_channel_ids}")
+        if self.has_recording() or self.has_temporary_recording():
+            new_recording = self.recording.select_channels(channel_ids)
```
chrishalcrow, line 1597: My proposal (following Sam's comment) is to demand that `channel_ids` are ordered in the same way as `self.channel_ids`.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -1518,6 +1572,70 @@ def select_units(self, unit_ids, format="memory", folder=None) -> "SortingAnalyz
...
+                        new_properties[key] = values_arr[channel_indices]
+                    else:
+                        new_properties[key] = values_arr
+                new_rec_attributes["properties"] = new_properties
+            if new_rec_attributes.get("probegroup") is not None:
+                slice_indices = self.channel_ids_to_indices(channel_ids)
+                new_probegroup = new_rec_attributes["probegroup"].get_slice(slice_indices)
+                new_rec_attributes["probegroup"] = new_probegroup
+        if self.sparsity is not None:
+            sparsity_mask = self.sparsity.mask[:, np.isin(self.channel_ids, channel_ids)]
```
alejoe91, line 1620: use `channel_ids_to_indices`

PR #4649, opened 2026-07-07
Improve sparsity estimation for aggregated analyzers

alejoe91:
Currently, when constricting a `SortingAnalyzer` via dicts, the only supported option for sparsity was `by_property`. This is ok for tetrodes, but is not the best option for multi shank probes.

This PR extends the logic of sparsity to first split by aggregation key and then apply whatever external sparsity is given (e.g. radius).


chrishalcrow (approved):
LGTM, after #4648 is merged.

PR #4653, opened 2026-07-07
Round before casting preprocessed traces to integer dtypes

CodeBlackwell:
Several preprocessors cast their float result straight to an integer `dtype` with `astype()`, which truncates toward zero instead of rounding. Over many samples that biases the output (toward zero for centered data, downward otherwise). `filter.py` already rounds before the same cast — this brings `whiten`, `common_reference` and `highpass_spatial_filter` in line with it. It also drops the `offset.astype(dtype)` cast in `ScaleRecording`, which was silently zeroing fractional offsets for integer dtypes.

Added tests covering the integer-rounding paths.

Noticed while looking at #1972; this addresses the integer-cast rounding, not the zscore/`int_scale` std question in that issue.

### DIFF
--- a/src/spikeinterface/preprocessing/common_reference.py
+++ b/src/spikeinterface/preprocessing/common_reference.py
@@ -230,7 +230,7 @@ class CommonReferenceRecordingSegment(BasePreprocessorSegment):
                         traces[:, channel_indices] - traces.dot(self.local_kernel.T)[:, channel_indices]
                     )
             if np.issubdtype(self.dtype, np.integer):
-                re_referenced_traces = np.round(re_referenced_traces)
+                np.round(re_referenced_traces, out=re_referenced_traces)
             return re_referenced_traces.astype(self.dtype, copy=False)
 
         # Then the old implementation for backwards compatibility that supports grouping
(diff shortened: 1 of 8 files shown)


zm711:
We've been discussing this a bit and two questions:
1) we aren't sure if we see much value in the tests that were added. We trust the implementation of round vs int from numpy and python. At minimum we would remove the monkey patching test.
2) Could you give us some profiling of this change. We've actually had a discussion about rounding in the past, but the issue is that `np.round` needs to allocate at least an equalize sized array and we are concerned that even with our chunking mechanism that the memory costs may really explode as recording files are between 10GB-1TB depending on when the memory is freed.

CodeBlackwell:
Both addressed in 768e3af.

1. Monkeypatch test removed.
2. The rounding is now in place (`np.round(x, out=x)`), so it allocates nothing. The rounded arrays are locally created in all three segments (arithmetic results or post-filter copies), so in-place mutation is safe.

Profiling `common_reference` on a 1 s, 384 channel int16 chunk (30,000 x 384), `dtype="int16"`:

| variant | median get_traces | peak traced memory |
|---|---|---|
| main (no rounding) | 200 ms | 115.4 MB |
| np.round copy (as reviewed) | 222 ms | 184.6 MB |
| np.round in place (now) | 192 ms | 115.4 MB |

Peak memory is identical to main: the float64 intermediate from the reference subtraction dominates either way, and the in-place round adds nothing, so chunked processing of large files is unaffected. Happy to trim the remaining tests further if you'd prefer.

zm711:
That looks good to me. We'll have @alejoe91 or @samuelgarcia give it a final read through. Thanks for the contribution.

samuelgarcia (approved):

samuelgarcia:
merci.

PR #4657, opened 2026-07-08
Fix template estimate sparsity

alejoe91:
Fixes estimate sparsity for probegroups with non-unique channel locations


chrishalcrow (approved):
Ugly, but needed for now

PR #4662, opened 2026-07-08
Build the `NwbSortingExtractor` spike vector with a single bulk read

h-mayorquin:
`NwbSortingExtractor` now overrides `_compute_and_cache_spike_vector` to build the whole spike vector from one read of the Units table instead of the generic per-unit loop it inherits from `BaseSorting`. An NWB (Neurodata Without Borders) Units table stores every unit's spikes in a single flat `spike_times` dataset grouped by unit, with the per-unit boundaries in `spike_times_index`, so the entire table can be read in one shot and each spike's unit index recovered from those boundaries. The generic builder instead reads one spike train per unit, which is one streamed read per unit; for a remote or streamed file that is the difference between a single request and one request per unit. Construction stays lazy as before: nothing is read until the spike vector is first requested.

The seconds-to-samples conversion stays on the segment, shared by the per-unit and bulk paths through `_times_to_frames`, and the override only assembles and sorts the structured array, so its output is byte-for-byte identical to the generic builder. Correctness is therefore already exercised by the existing NWB (Neurodata Without Borders) roundtrip tests, and a comment on the override records why it should not be reverted to the generic per-unit path.


src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1426,6 +1427,30 @@ def __init__(
...
+        # Do not drop this override in favor of the generic path without a reason: doing so silently
+        # reintroduces the per-unit read pattern. An NWB Units table is always single-segment, so the
+        # single segment is built directly.
+        segment = self.segments[0]
+        sample_indices, unit_indices = segment._get_all_spike_frames_and_unit_indices()
+
+        # Stable sort by sample_index keeps unit_index ordering on ties, matching the way the generic
+        # BaseSorting builder constructs the vector (unit order within a sample).
+        order = np.argsort(sample_indices, stable=True)
+        n = sample_indices.size
```
zm711: Maybe n_spikes to be more explicit?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1426,6 +1427,30 @@ def __init__(
...
+        # turns num_units streamed reads into one, which is a large difference for remote/streamed files.
+        # Do not drop this override in favor of the generic path without a reason: doing so silently
+        # reintroduces the per-unit read pattern. An NWB Units table is always single-segment, so the
+        # single segment is built directly.
+        segment = self.segments[0]
+        sample_indices, unit_indices = segment._get_all_spike_frames_and_unit_indices()
+
+        # Stable sort by sample_index keeps unit_index ordering on ties, matching the way the generic
+        # BaseSorting builder constructs the vector (unit order within a sample).
+        order = np.argsort(sample_indices, stable=True)
```
zm711, line 1448: The data isn't presorted in NWB? Or can't be guaranteed to be sorted?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1468,6 +1493,18 @@ def _frame_to_time(self, frame):
             return self._time_vector[frame]
         return frame / self._sampling_frequency + self._t_start
 
+    def _times_to_frames(self, spike_times) -> np.ndarray:
```
zm711: in nwb terminology do you say frames instead of samples?

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1481,16 +1518,22 @@ def get_unit_spike_train(
...
+    def _get_all_spike_frames_and_unit_indices(self):
+        """Read all units' spike trains (as frames) in one bulk read, plus each spike's unit index.
+
+        NWB stores all spike times in a single flat dataset grouped by unit (with per-unit
+        boundaries in ``spike_times_index_data``), so the whole segment is read at once instead
+        of one streamed slice per unit.
+        """
+        spike_times = np.asarray(self.spike_times_data[:], dtype="float64")
+        # boundaries[u]:boundaries[u + 1] is the slice of spike_times belonging to unit u.
+        boundaries = np.concatenate(([0], np.asarray(self.spike_times_index_data[:], dtype="int64")))
```
zm711, line 1543: This seems super fancy and I'm having a hard time following this. So you add a 0 to the beginning then take the diff to get boundaries? Could you give me the dumbed down version of why this works?

zm711 (commented):
Couple questions.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Yes, good point.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Yes, nwb favors samples. Do you think it should be samples here? I thought about it and I remember that the get_traces uses frame terminology but maybe we were moving away from it?

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Actually yes, this is kind of complicated. Mmm, let's discuss it tomorrow and we probably should add documentation in hdmf that we c an link for this.

src/spikeinterface/extractors/nwbextractors.py
chrishalcrow: I vote for samples. The spike vector uses `spikes["sample_index"]` and samples is found across the codebase more often than frames.

chrishalcrow:
On some real data: https://dandiarchive.org/dandiset/000939/0.260512.1701/files?location=sub-A1808&page=1
Building the spike vector took
Main: 106 seconds
This PR: 7.5 seconds
!!!

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: We discussed this in person and I added a comment.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Ok, I changed this.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Yes, we can't guarantee that I added a comment.

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: num_spikes

chrishalcrow (approved):

src/spikeinterface/extractors/nwbextractors.py
```diff
@@ -1426,6 +1427,34 @@ def __init__(
...
+        # the point. The generic builder reads one spike train per unit (one streamed read per unit),
+        # whereas an NWB Units table stores every unit's spikes in a single flat times dataset grouped by
+        # unit. Reading that dataset once (see NwbSortingSegment._get_all_spike_samples_and_unit_indices)
+        # turns num_units streamed reads into one, which is a large difference for remote/streamed files.
+        # Do not drop this override in favor of the generic path without a reason: doing so silently
+        # reintroduces the per-unit read pattern. An NWB Units table is always single-segment, so the
+        # single segment is built directly.
+        segment = self.segments[0]
+        sample_indices, unit_indices = segment._get_all_spike_samples_and_unit_indices()
+
```
chrishalcrow, line 1441: From Sam:

We could use `sample_indices` and `unit_indices` to directly populate `_cached_lexsorted_spike_vector` dict, since it's already ordered in the `key = ("sample_index, segment_index, unit_index")` (we think!).

When you do this, you should also save the "slices" which are the boundaries you compute in `_get_all_spike_samples_and_unit_indices`.

alejoe91: Added an issue about this: https://github.com/SpikeInterface/spikeinterface/issues/4698

alejoe91:
Amazing @h-mayorquin !!1 Merging :)

src/spikeinterface/extractors/nwbextractors.py
h-mayorquin: Hi, thanks @alejoe91, I started thinking on this topic and it was a rabbit hole so I droped it, I do agree with alessio that can be done as a separate feature.

PR #4665, opened 2026-07-08
Add `interpolation_kwargs` to `interpolate_templates`

chrishalcrow:
This allows the user to pass interpolate_templates to interpolate_templates.

### DIFF
--- a/src/spikeinterface/generation/drift_tools.py
+++ b/src/spikeinterface/generation/drift_tools.py
@@ -73,6 +73,9 @@ def interpolate_templates(
             if interpolation_method == "thin_plate":
 
                 if "neighbors" not in interpolation_kwargs:
+                    # If neighbors it not passed, `RBFInterpolator` uses all channels.
+                    # This works poorly for standard ephys template interpolation.
+                    # Depending on the probe, you might want to decrease this value.
                     interpolation_kwargs["neighbors"] = 12
 
                 tps_interpolator = RBFInterpolator(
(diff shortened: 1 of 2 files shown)


src/spikeinterface/generation/drift_tools.py
```diff
@@ -65,7 +72,12 @@ def interpolate_templates(
 
             if interpolation_method == "thin_plate":
 
-                tps_interpolator = RBFInterpolator(source_locations, template, kernel="thin_plate_spline", neighbors=12)
+                if "neighbors" not in interpolation_kwargs:
```
zm711, line 75: I'm a little wary of this. Why should we add in neighbors? If someone has poorly formed interpolation_kwargs shouldn't we let the function fail so the person can adjust themselves?

chrishalcrow: It will run and give nonsense results for any probes with more than, say, 32 channels. 
I don't think this is much different than setting default kwargs in other methods.

src/spikeinterface/generation/drift_tools.py
```diff
@@ -65,7 +72,12 @@ def interpolate_templates(
...
+                    interpolation_kwargs["neighbors"] = 12
```
chrishalcrow, line 79:
```suggestion
                    # If neighbors it not passed, `RBFInterpolator` uses all channels.
                    # This works poorly for standard ephys template interpolation.
                    # Depending on the probe, you might want to decrease this value.
                    interpolation_kwargs["neighbors"] = 12
```

alejoe91 (approved):

PR #4667, opened 2026-07-09
Deprecation -> Future

zm711:
For @chrishalcrow 

For future readers:

`DeprecationWarnings` are for other developers which means that the warnings are filtered out in REPLs and notebooks. Most of our warnings are meant for end-users which per the Python docs should be `FutureWarnings` as of Python 3.7+. This updates our warnings to be in-line with this.


chrishalcrow (approved):

h-mayorquin (approved):
LGTM

PR #4671, opened 2026-07-09
Move the `Base.save` to `Sorting.save()` and `Recording.save()`

samuelgarcia:
This PR refactors the `BaseExtractor.save()` function to unify and simplify the implementation:

- Remove the `_save()` function and deprecate `save_to_memory`, `save_to_folder`, `save_to_zarr`: the `save()` function is now the only entry point for saving
- The `BaseRecording` and `BaseSorting` route the save to the main backend formats, which have a `write_*` statich method to handle all the write logic (e.g., `BinaryFolderRecording.write_recording()`)
- Remove the `folder_metadata` concept, used previously to save the probe object and not needed since #4465 (now the `probegroup` is dumped to dictionary as extra metadata)
- Centralize saving and loading of properties and annotations to folder
- Deprecate `BaseSnippets` object, since it's not an intended use case we want to suppor in the future

In addition to these changes, this PR also integrates #4689, which improves the handling of times in several ways:
- it adds time info to the dict only when it has been modified in memory
- it changes the `write_binary` function to also write timestamps in chunks when available
- it modifies `FrameSliceRecoridng` and `BasePreprocessor` to lazily handleand return timestamps from parent, instead od making a full copy


alejoe91:
@samuelgarcia fixed some bugs in preprocessing, but there are more. Do you want me to fix it?

alejoe91:
@samuelgarcia ported changes from #4689 (timestamps handling + removing `metadata_folder`) and cleaned up code and docs.

One main question: in #4689 , I added an extra `_in_reduce` mechanism to the `to_dict`/`_extra_metadata_to_dict` stack. This was only set to True by the `__reduce__`, and propagated times kwargs only in this case. 

Here I propose a modified version: if the user "messes" with timestamps, this is propagated to dict no matter what. Note that:
- this is needed when saving in parallel (timestamps need to be saved!)
- the final `si_folder.json` will NOT have this information, since it sets `include_extra_metadate=False` (added also `include_annotations`/`include_properties` to False)

Let me know what you think!

src/spikeinterface/core/binaryrecordingextractor.py
```diff
@@ -51,17 +53,18 @@ class BinaryRecordingExtractor(BaseRecording):
...
-        offset_to_uV=None,
-        is_filtered=None,
+        t_starts: list[float] | None = None,
+        channel_ids: list[str | int] | None = None,
+        time_axis: int = 0,
+        file_offset: int = 0,
+        gain_to_uV: float | np.ndarray | None = None,
+        offset_to_uV: float | np.ndarray | None = None,
+        is_filtered: bool | None = None,
+        file_timestamps_paths: str | Path | list[str | Path] | None = None,
```
samuelgarcia, line 67: I think we should limit the complexity of BinaryRecordingExtractr to load only one or sevreal binary files.
And keep the complexity only in BinaryFolderRecording other we need to maintained a 2 classes design.

src/spikeinterface/core/basesorting.py
```diff
@@ -502,45 +502,77 @@ def get_times(
         else:
             return None
 
-    def save(self, format="numpy_folder", **save_kwargs):
+    def save(self, format="binary", **save_kwargs):
```
alejoe91: Another suggestion is `binary_folder`, but revert for now

src/spikeinterface/core/time_series.py
```diff
@@ -33,6 +33,8 @@ class TimeSeries(ABC):
     """
 
     _preferred_mp_context = None
+    # Flag to indicate whether time info has been modified in-memory (e.g. by set_times or shift_times).
+    _time_info_modified = False
```
alejoe91: Let's make a proper doubel inheritance and the init will set this to False

src/spikeinterface/preprocessing/basepreprocessor.py
```diff
@@ -34,3 +34,26 @@ def get_num_samples(self):
 
     def get_traces(self, start_frame, end_frame, channel_indices):
         raise NotImplementedError
+
+    # Preprocessors never change the frame numbering (no offset, no resampling), so time
+    # handling is a pure pass-through to the parent segment. Delegating live (instead of
+    # relying on the time_vector/t_start copied into __init__ above) lets any lazy/offset-aware
+    # override further up the chain (e.g. FrameSliceRecordingSegment after a frame_slice) keep
+    # working without materializing a full time_vector every time this segment is reconstructed.
+    def get_times(self, start_frame=None, end_frame=None):
```
alejoe91, line 43: we may need this for `FrameSlice` too

src/spikeinterface/core/base.py
```diff
@@ -586,7 +581,9 @@ def to_dict(
             dump_dict["annotations"] = self._annotations
         else:
             # include only main annotations
-            dump_dict["annotations"] = {k: self._annotations.get(k, None) for k in self._main_annotations}
+            dump_dict["annotations"] = {
+                k: self._annotations.get(k) for k in self._main_annotations if self._annotations.get(k) is not None
+            }
```
chrishalcrow, line 587: Would be nice to save the `main_ids` somewhere near here

src/spikeinterface/core/time_series.py
alejoe91: done in 51c6b513abb3f63d59aa467b097982e0158e4317

src/spikeinterface/core/basesorting.py
alejoe91: done in 51c6b513abb3f63d59aa467b097982e0158e4317

src/spikeinterface/core/base.py
alejoe91: let's do that.

alejoe91: done in 443f80a7e3be836528489ac3e7c0e58e733e737d

src/spikeinterface/preprocessing/basepreprocessor.py
alejoe91: done in a2301eccbfd4b5dfed6402e08c49e7044c6dc729

alejoe91:
@samuelgarcia @chrishalcrow all done! Ready to merge on my end.

src/spikeinterface/core/base.py
```diff
@@ -581,12 +576,15 @@ def to_dict(
 
         dump_dict = retrieve_importing_provenance(self.__class__)
         dump_dict["kwargs"] = kwargs
+        dump_dict["main_ids"] = self._main_ids
```
chrishalcrow:
```suggestion
```

alejoe91 (approved):

PR #4673, opened 2026-07-09
Enhance single channel denoiser and `WaveformNode`

alejoe91:
Extend flexibility of `WaveformNode` and single channel denoiser. Now testing against toy and yass models here: https://huggingface.co/SpikeInterface/waveform_denoiser


src/spikeinterface/sortingcomponents/waveforms/neural_network_denoiser.py
```diff
@@ -11,86 +14,188 @@
...
-        self.assert_model_and_waveform_temporal_match(waveform_extractor)
-
+        if model_folder is None and repo_id is None:
+            raise ValueError("You need to specify either model_folder or repo_id")
+        if model_folder is not None and repo_id is not None:
+            raise ValueError("You cannot specify both model_folder and repo_id")
+        spike_size = waveform_node.nbefore + waveform_node.nafter
         # Load model
-        self.denoiser = self.load_model()
+        self.denoiser, model_relative_path = self.load_model(
```
alejoe91, line 75: Move it to a separate function and have the node accept the instantiated model

src/spikeinterface/sortingcomponents/waveforms/neural_network_denoiser.py
```diff
@@ -11,86 +14,188 @@
...
+
     def __init__(
-        self, recording: BaseRecording, return_output: bool = True, parents: Optional[List[PipelineNode]] = None
+        self,
+        recording: BaseRecording,
+        return_output: bool = True,
+        parents: Optional[List[PipelineNode]] = None,
+        model_folder: Optional[str] = None,
+        repo_id: Optional[str] = None,
+        model_name: Optional[str] = None,
```
samuelgarcia, line 55: We should be able to give the model directly in this signature and skip the autoload when we want.

samuelgarcia:
OK for me but for noes we should always have the possibility to give the model directly and skip the laod mechanism

alejoe91:
> OK for me but for noes we should always have the possibility to give the model directly and skip the laod mechanism

then you cannod check the model_info. The idea is you can pass a fodler with the weights and the model_info.json. Isn't that enough?

samuelgarcia:
If the model is given we can skip tests.

alejoe91:
@samuelgarcia done in last commit. Also when you pass an external model it checks the last layer dimentions

samuelgarcia (approved):

PR #4677, opened 2026-07-10
Add backwards compatibility when passing sparsity kwargs to create_sorting_analyzer

chrishalcrow:
In the monster #4624 we added `**job_kwargs` to `create_sorting_analyzer` and changed `**sparsity_kwargs` to `sparsity_kwargs=None`, without handling backwards compat. This PR fixes this.

### DIFF
--- a/src/spikeinterface/core/sortinganalyzer.py
+++ b/src/spikeinterface/core/sortinganalyzer.py
@@ -40,6 +40,7 @@ from .sparsity import ChannelSparsity, estimate_sparsity
 from .sortingfolder import NumpyFolderSorting
 from .zarrextractors import get_default_zarr_compressor, ZarrSortingExtractor, super_zarr_open, _write_object_array
 from .node_pipeline import run_node_pipeline
+from .globals import get_global_job_kwargs
 
 
 # high level function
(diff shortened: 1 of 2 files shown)


src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -165,6 +166,26 @@ def create_sorting_analyzer(
...
+        if key not in get_global_job_kwargs():
+            sparsity_kwargs[key] = value
+            display_warning = True
+        else:
+            cleaned_job_kwargs[key] = value
+
+    if display_warning:
+        warnings.warn(
+            "Passing sparsity arguments via keyword arguments is deprecated. "
+            "Please pass them into the `sparsity_kwargs` dictionary instead.",
```
zm711: The message should include the deprecation version number so users know when we are stopping them.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -165,6 +166,26 @@ def create_sorting_analyzer(
     if sparsity_kwargs is None:
         sparsity_kwargs = dict()
```
zm711: while you're touching {}?

zm711: Just checked on 3.12 and the literal is still 20% faster. Back in earlier version so 3 it was 2x faster. But still benefit of just using the literal.

src/spikeinterface/core/tests/test_sparsity.py
```diff
@@ -295,6 +295,11 @@ def test_compute_sparsity():
     recording, sorting = get_dataset()
 
     sorting_analyzer = create_sorting_analyzer(sorting=sorting, recording=recording, sparse=False, return_in_uV=True)
+
+    # Check that we can still pass sparsity kwargs using the old signature
```
zm711: Just for me being an old man could you add to the comment that radius_um is a sparsity_kwarg

zm711 (commented):
Just one required tweak. Otherwise just some comments.

alejoe91 (approved):

src/spikeinterface/core/tests/test_sparsity.py
```diff
@@ -295,6 +295,15 @@ def test_compute_sparsity():
...
```
alejoe91:
```suggestion
    # Check that we can still pass sparsity kwargs using the old signature (e.g. passing radius_um)
```

src/spikeinterface/core/tests/test_sparsity.py
alejoe91: added (e.g., passing radius_um)

PR #4680, opened 2026-07-10
Clean up code around create_sorting_analyzer and create/load_* functions in sortinganalyzer.py

ecobost:
This PR does not functionally change much. It is mainly for code refactoring and to improve legibility.

The only two functional changes are:
* fix small bug when calling save_to_zarr("test.zarr") would add an extra .zarr to the filename (https://github.com/SpikeInterface/spikeinterface/blob/39dfac40e630ff996be410aa62a54f5ff6b7eb11/src/spikeinterface/core/base.py#L1087)
* when return_scaled and return_in_uV are provided, prioritize return_in_uV. Currently, it prefers return_scaled over return_in_uV

Code quality improvements:
* type hinting
* Collect code into logical blocks (e..g, everything that has to do wit main indices together); add comment headers (e.g., Saving main indices) and reorder this blocks (e.g., expected stuff like sorting should be saved and loaded before any of the optional things like sparsity)
* Make create_binary_folder parallel create_zarr (you can now see how they are both doing exactly the same). 
* Make load_binary_folder parallel load_zarr.
* Preserve the same order of operations when loading and saving. If sorting was saved first, then load first and so on.
* Tighten code when possible

### DIFF
--- a/src/spikeinterface/core/base.py
+++ b/src/spikeinterface/core/base.py
@@ -19,6 +19,7 @@ from .core_tools import (
     clean_zarr_folder_name,
     is_dict_extractor,
     SIJsonEncoder,
+    is_path_remote,
     make_paths_relative,
     make_paths_absolute,
     check_paths_relative,
@@ -643,7 +644,7 @@ class BaseExtractor:
             extractor.load_metadata_from_folder(folder_metadata)
         return extractor
 
-    def load_metadata_from_folder(self, folder_metadata):
+    def load_metadata_from_folder(self, folder_metadata: str | Path):
         # hack to load probe for recording
         folder_metadata = Path(folder_metadata)
 
@@ -658,11 +659,11 @@ class BaseExtractor:
 
         self._extra_metadata_from_folder(folder_metadata)
 
-    def save_metadata_to_folder(self, folder_metadata):
+    def save_metadata_to_folder(self, folder_metadata: str | Path):
         self._extra_metadata_to_folder(folder_metadata)
 
         # save properties
-        prop_folder = folder_metadata / "properties"
+        prop_folder = Path(folder_metadata) / "properties"
         prop_folder.mkdir(parents=True, exist_ok=False)
         for key in self.get_property_keys():
             values = self.get_property(key)
@@ -1084,24 +1085,17 @@ class BaseExtractor:
             cache_folder = get_global_tmp_folder()
             if name is None:
                 name = "".join(random.choices(string.ascii_uppercase + string.digits, k=8))
-                zarr_path = cache_folder / f"{name}.zarr"
-                if verbose:
-                    print(f"Use zarr_path={zarr_path}")
-            else:
-                zarr_path = cache_folder / f"{name}.zarr"
-                if not is_set_global_tmp_folder():
-                    if verbose:
-                        print(f"Use zarr_path={zarr_path}")
+            zarr_path = (cache_folder / name).with_suffix(".zarr")
+            if verbose:
+                print(f"Saving to zarr_path={zarr_path}")
         else:
-            if storage_options is None:
+            if storage_options is None:  # save locally (not cloud storage)
                 folder = clean_zarr_folder_name(folder)
                 if folder.is_dir() and overwrite:
                     shutil.rmtree(folder)
-                zarr_path = folder
-            else:
-                zarr_path = folder
+            zarr_path = folder
 
-        if isinstance(zarr_path, Path):
+        if not is_path_remote(zarr_path):
             assert not zarr_path.exists(), f"Path {zarr_path} already exists, choose another name"
         save_kwargs["zarr_path"] = zarr_path
         save_kwargs["storage_options"] = storage_options
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -548,7 +548,7 @@ class BaseSorting(BaseExtractor):
         # Re-register the recording if saving to disk (not memory)
         if self.has_recording() and format != "memory":
             warnings.warn(
-                "The recording registered to this sorting object will not be saved to disk"
+                "The recording registered to this sorting object will not be saved to disk. "
                 "Reloading the sorting later will not include the recording"
             )
             cached.register_recording(self._recording)
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -344,7 +344,7 @@ class BaseSorting(BaseExtractor):
                 warnings.warn(
                     "Some spikes exceed the recording's duration! "
                     "Removing these excess spikes with `spikeinterface.curation.remove_excess_spikes()` "
-                    "Might be necessary for further postprocessing."
+                    "might be necessary for further postprocessing."
                 )
         self._recording = recording
         # Copy the recording's start times into the sorting segments. This way,
@@ -500,14 +500,16 @@ class BaseSorting(BaseExtractor):
         else:
             return None
 
-    def _save(self, format="numpy_folder", **save_kwargs):
-        """
-        This function replaces the old CachesortingExtractor, but enables more engines
+    def _save(self, format: str = "numpy_folder", **save_kwargs):
+        """ Save a sorting object to disk in a specified format.
+
+        Note
+        ----
+        This function replaces the old CacheSortingExtractor, but enables more engines
         for caching a results.
 
         Since v0.98.0 "numpy_folder" is used by defult.
         From v0.96.0 to 0.97.0 "npz_folder" was the default.
-
         """
         if format == "numpy_folder":
             from .sortingfolder import NumpyFolderSorting
@@ -515,10 +517,6 @@ class BaseSorting(BaseExtractor):
             folder = save_kwargs.pop("folder")
             NumpyFolderSorting.write_sorting(self, folder)
             cached = NumpyFolderSorting(folder)
-
-            if self.has_recording():
-                warnings.warn("The registered recording will not be persistent on disk, but only available in memory")
-                cached.register_recording(self._recording)
 
         elif format == "zarr":
             from .zarrextractors import ZarrSortingExtractor
@@ -528,20 +526,12 @@ class BaseSorting(BaseExtractor):
             ZarrSortingExtractor.write_sorting(self, zarr_path, storage_options, **save_kwargs)
             cached = ZarrSortingExtractor(zarr_path, storage_options)
 
-            if self.has_recording():
-                warnings.warn("The registered recording will not be persistent on disk, but only available in memory")
-                cached.register_recording(self._recording)
-
         elif format == "npz_folder":
             from .sortingfolder import NpzFolderSorting
 
             folder = save_kwargs.pop("folder")
             NpzFolderSorting.write_sorting(self, folder)
             cached = NpzFolderSorting(folder_path=folder)
-
-            if self.has_recording():
-                warnings.warn("The registered recording will not be persistent on disk, but only available in memory")
-                cached.register_recording(self._recording)
 
         elif format == "memory":
             if save_kwargs.get("sharedmem", True):
@@ -553,7 +543,14 @@ class BaseSorting(BaseExtractor):
 
                 cached = NumpySorting.from_sorting(self)
         else:
-            raise ValueError(f"format {format} not supported")
+            raise ValueError(f"Format {format} not supported")
+
+        # Re-register the recording if saving to disk (not memory)
+        if self.has_recording() and format != "memory":
+            warnings.warn("The recording registered to this sorting object will not be saved to disk" 
+                          "Reloading the sorting later will not include the recording")
+            cached.register_recording(self._recording)
+
         return cached
 
     def get_unit_property(self, unit_id, key):
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -501,7 +501,7 @@ class BaseSorting(BaseExtractor):
             return None
 
     def _save(self, format: str = "numpy_folder", **save_kwargs):
-        """ Save a sorting object to disk in a specified format.
+        """Save a sorting object to disk in a specified format.
 
         Note
         ----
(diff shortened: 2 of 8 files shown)


src/spikeinterface/core/base.py
```diff
@@ -1084,24 +1085,17 @@ def save_to_zarr(
             cache_folder = get_global_tmp_folder()
             if name is None:
                 name = "".join(random.choices(string.ascii_uppercase + string.digits, k=8))
-                zarr_path = cache_folder / f"{name}.zarr"
-                if verbose:
-                    print(f"Use zarr_path={zarr_path}")
-            else:
```
zm711, line 1090: where did this else go? If a name is specified we should accept their name right? I haven't read this for a while.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -41,26 +41,30 @@
 from .zarrextractors import get_default_zarr_compressor, ZarrSortingExtractor, super_zarr_open, _write_object_array
 from .node_pipeline import run_node_pipeline
 
+# Typing hints
+PeakSignType = Literal["both", "neg", "pos"]
+PeakModeType = Literal["extremum", "at_index", "peak_to_peak"]
```
zm711, line 46: There was some debate about whether people liked this or not. Someone else had proposed doing this and we decided against it at the time, but maybe people have changed their minds?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -919,67 +963,57 @@ def _get_zarr_root(self, mode="r+"):
...
-        if is_path_remote(folder):
-            remote = True
-        else:
-            remote = False
-        if not remote:
+        is_remote = is_path_remote(folder)
+        if not is_remote:
             folder = clean_zarr_folder_name(folder)
-            if folder.is_dir():
+            if folder.exists():
```
zm711, line 1005: why this change? is_dir() verifies existence and that it is a folder? So it's more specific.

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -990,95 +1024,111 @@ def create_zarr(
...
+                _write_object_array(zarr_root, "recording", check_json(rec_dict), codec="json")
+            elif recording.check_serializability("pickle"):
+                _write_object_array(zarr_root, "recording", rec_dict, codec="pickle")
+            else:
+                warnings.warn("The Recording is not serializable! The recording link will be lost for future load")
         else:
-            rec_attributes = rec_attributes.copy()
-            probegroup = rec_attributes.pop("probegroup")
+            assert rec_attributes is not None, "recording or rec_attributes must be provided"
+            warnings.warn("Recording not provided, instntiating SortingAnalyzer in recordingless mode.")
```
zm711:
```suggestion
            warnings.warn("Recording not provided, instantiating SortingAnalyzer in recordingless mode.")
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -990,95 +1024,111 @@ def create_zarr(
...
                 zarr_root_a = zarr.open(str(folder), mode="a", storage_options=storage_options)
                 zarr.consolidate_metadata(zarr_root_a.store)
-            except Exception as e:
+            except:
                 warnings.warn(
-                    "The zarr store was not properly consolidated prior to v0.101.1. "
+                    "The zarr store was not consolidated prior to v0.101.1. "
                     "This may lead to unexpected behavior in loading extensions. "
-                    "Please consider re-generating the SortingAnalyzer object."
+                    "Consider re-generating the SortingAnalyzer object."
```
zm711, line 1113: Why remove the politeness?

zm711 (commented):
A few initial comments.

src/spikeinterface/core/base.py
ecobost: logic before was if name is None, -> [create random name as name, create zarr_path] else [create_zarr_path]. Now, the logic is if name is none-> [create_random_name] -> (always) create_zarr_path. The only difference is in whether something is printed. Assuming verbose=True, before if name is None, print "using zarr folder", else if name was provided but not is_set_global_tmp_folder, then print "using zarr folder..." , now it always prints (when verbose=True). I think it's worth it for simpler code

src/spikeinterface/core/sortinganalyzer.py
ecobost: Otherwise I had to repeat Literal["both", "neg", "pos"] in a bunch of places. And in theory if later an option was added to the list also add that option in a bunch of places. I'm ok either way, it's just variables

src/spikeinterface/core/sortinganalyzer.py
ecobost: In here particularly, we don't actually care about it being a directory because we are gonna delete it anyway (whether a fle or directory); but actually we use exists everywhere so I'll change the ones where isdir/isfile will be more appropiate

src/spikeinterface/core/sortinganalyzer.py
ecobost: I thought Consider re-generating was weak enough already :) and more direct. We are not even telling them to re-generate it but to consider it, Won't die on this hill though

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -83,7 +87,7 @@ def create_sorting_analyzer(
         The sorting object, or a dict of them
     recording : Recording | dict
         The recording object, or a dict of them
-    folder : str or Path or None, default: None
+    folder : str or Path
```
alejoe91: ?
```suggestion
    folder : str or Path or None, default: None
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -311,12 +307,24 @@ def create_sorting_analyzer(
...
+    if format != "memory":
+        assert folder is not None, "For format='binary_folder'|'zarr', folder name must be provided"
+        if not is_path_remote(folder):
+            folder = clean_zarr_folder_name(folder) if format == "zarr" else Path(folder)
+            if folder.exists():
+                if overwrite:
+                    shutil.rmtree(folder)
+                else:
+                    raise ValueError(f"Folder {folder} already exists! Use overwrite=True to overwrite it.")
+
```
alejoe91: I think these should stay at the top. If the folder exists and overwrite is not set, then the user has to wait that main channel ids/sparsity is estimated before reaching the error.

Was there another reason to move this below @ecobost ?

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -607,89 +631,96 @@ def create_memory(cls, sorting, recording, sparsity, return_in_uV, peak_sign, pe
...
+        # Dump recording provenance
         if recording is not None:
             # save recording and sorting provenance
             if recording.check_serializability("json"):
-                recording.dump(folder / "recording.json", relative_to=folder)
+                recording.dump(recording_provenance_file, relative_to=folder)
             elif recording.check_serializability("pickle"):
-                recording.dump(folder / "recording.pickle", relative_to=folder)
+                recording.dump(recording_provenance_file.with_suffix(".pickle"), relative_to=folder)
```
```suggestion
                recording.dump(folder / "recording.json", relative_to=folder)
elif recording.check_serializability("pickle"):
                recording.dump(folder / "recording.pickle", relative_to=folder)
```
I think the previous version was cleaner. I don't see the point in defining the json path and change the suffix for pickle

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -607,89 +631,96 @@ def create_memory(cls, sorting, recording, sparsity, return_in_uV, peak_sign, pe
...
+        recording_info_folder = folder / "recording_info"  # to save rec_attributes and probe group
+        recording_info_folder.mkdir()
+
+        # Set names for output files
+        info_file = folder / "spikeinterface_info.json"
+        settings_file = folder / "settings.json"
+        sorting_folder = folder / "sorting"
+        sorting_provenance_file = folder / "sorting_provenance.json"
+        sparsity_file = folder / "sparsity_mask.npy"
+        recording_provenance_file = folder / "recording.json"
```
```suggestion
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -698,21 +729,18 @@ def _handle_backward_compatibility_settings_pre_init(cls, settings, sorting, spa
...
+            new_settings["return_in_uV"] = new_settings.pop("return_scaled", True)
         else:
-            # old version did not have settings at all
-            new_settings["return_in_uV"] = True
+            new_settings.pop("return_scaled", None)
 
-        if "peak_sign" not in settings:
-            # before 0.104.0 peak_sign was not in settings.
+        if "peak_sign" not in new_settings:
+            # before 0.104.1 peak_sign was not in settings.
```
```suggestion
            # before 0.105.0 peak_sign was not in settings.
```
I think we didn't properly update the dev version, but the comment should point to the upcoming release

alejoe91: See https://github.com/SpikeInterface/spikeinterface/pull/4696

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -698,21 +729,18 @@ def _handle_backward_compatibility_settings_pre_init(cls, settings, sorting, spa
 
         Note :
          * see also _handle_backward_compatibility_settings_post_init
-         * there is also something at extension level to handle changes in parameters with deferents mechanism
+         * there is also something at extension level to handle changes in paramaters with different mechanisms
```
```suggestion
         * there is also something at extension level to handle changes in parameters with different mechanisms
```
I think this was correct :P

src/spikeinterface/core/sortinganalyzer.py
alejoe91: Well I think we try to use `is_dir/is_file` when possible, which I believe is better because more explicit, so I would revert
```suggestion
            if folder.is_dir():
```

alejoe91 (changes requested):
Thanks @ecobost 

Much cleaner :) I have a few suggestions/comments. Could you tackle them? Then this is good to go for me!

src/spikeinterface/core/sortinganalyzer.py
ecobost: i did revert/change exists()-> is_dir/is_filewhen needed on the next commit below. But exists() makes more sense here, if someone says save it at /xxx/yyy and overwrite=True, then whatever is in /xxx/yyy (dir or file) will be overwritten

src/spikeinterface/core/sortinganalyzer.py
ecobost: incorporated in new commit

src/spikeinterface/core/sortinganalyzer.py
ecobost: all of the other code dealt with cleaning up input (if sparsity is none, create it, if main_channel_indices is none, create it and so on) and this is the one piece of code that has an actual effect (it deletes a folder) so it felt more logical afterwards, but will move to the top

src/spikeinterface/core/sortinganalyzer.py
ecobost: the idea was to signal the .json as the default output file; I could change it to 
`recording_provenance_file = folder / "recording" ` and then add the suffix at the dump line if you prefer

ecobost:
@alejoe91  Couple of comments still open. Let me know when this version is ok to fix merge issues (don't wanna pollute current version)

src/spikeinterface/core/sortinganalyzer.py
alejoe91: maybe call it `recording_provenance_file_stem = folder / "recording"` ?

src/spikeinterface/core/sortinganalyzer.py
alejoe91: actually, what you suggest makes sense. If someone has a file named like the folder then `is_dir` will not pick it up. So let's use `exists` with checks as you suggest!

alejoe91:
@ecobost replied to the standing comments! Let's continue and fix the conflicts

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -851,71 +878,87 @@ def _compute_main_channel_backwards_compatibility(self):
...
 
-        # load internal sorting copy in memory
+        # Load settings file
+        analyzer_has_settings_file = settings_file.exists()
+        if analyzer_has_settings_file:
+            with open(settings_file, "r") as f:
+                settings = json.load(f)
+        else:
+            # TODO: Remove support for analyzers that have no settings.json (throw an error instead)
+
```
```suggestion
```

alejoe91 (approved):

src/spikeinterface/core/basesorting.py
```diff
@@ -553,7 +543,16 @@ def _save(self, format="numpy_folder", **save_kwargs):
 
                 cached = NumpySorting.from_sorting(self)
         else:
-            raise ValueError(f"format {format} not supported")
+            raise ValueError(f"Format {format} not supported")
+
+        # Re-register the recording if saving to disk (not memory)
+        if self.has_recording() and format != "memory":
+            warnings.warn(
+                "The recording registered to this sorting object will not be saved to disk"
```
chrishalcrow:
```suggestion
                "The recording registered to this sorting object will not be saved to disk. "
```

src/spikeinterface/core/sortinganalyzer.py
```diff
@@ -163,6 +168,16 @@ def create_sorting_analyzer(
     In some situation, sparsity is not needed, so to make it fast creation, you need to turn
     sparsity off (or give external sparsity) like this.
     """
+    # Remove folder if overwrite is True
```
```suggestion
    assert format in ["memory", "binary_folder", "zarr"], "Format must be 'memory', 'binary_folder' or 'zarr'."

    # Remove folder if overwrite is True        
```

chrishalcrow (approved):
This is fine by me. Would be great to propagate the PeakSign and PeakMode typing across the codebase. Would fix some incorrect typing (https://github.com/SpikeInterface/spikeinterface/blob/ecf5c6d60933b52b8b875780b3bc5ef820e8a931/src/spikeinterface/core/sparsity.py#L782) !! Can do that in a follow up.

PR #4682, opened 2026-07-10
Fix Llobet expected-violations formula in sliding RP metric

oliche:
## Fix Llobet expected-violations formula in the sliding RP metric

`_compute_violations` computed expected violations as `Ve = 2·τ/D · Nc·N` (i.e. `Nc·(Nb + Nc)`), which overestimates `Ve` and inflates the contamination confidence. Corrected to the Llobet et al. (2022) form used by the reference implementation:

```
Ve = 2·τ/D · Nc·(Nb + (Nc − 1)/2)     Nc = C·N,  Nb = (1 − C)·N,  D = N / firing_rate
```

Also extends the default contamination grid to include 35 % (`np.arange(0.5, 35.5, 0.5)`, previously stopped at 34.5 %) and adds a unit test for `_compute_violations` against the analytical formula.

Reference: https://github.com/SteinmetzLab/slidingRefractory (`metrics.computeViol`).

### Known non-equivalences (out of scope)
The metric definition now matches the reference, but results are not bit-identical because SI (by existing design) uses a coarser ACG bin size (`bin_size_ms=0.25` vs sample-resolution) and a different correlogram engine (`_compute_correlograms_*` vs the reference `histdiff`-equivalent). Full numerical parity would be a separate change.

### DIFF
--- a/src/spikeinterface/metrics/quality/misc_metrics.py
+++ b/src/spikeinterface/metrics/quality/misc_metrics.py
@@ -552,7 +552,7 @@ def compute_sliding_rp_violations(
     max_ref_period_ms : float, default: 10
         Maximum refractory period to test in ms.
     contamination_values : 1d array or None, default: None
-        The contamination values to test, If None, it is set to np.arange(0.5, 35, 0.5).
+        The contamination values to test, If None, it is set to np.arange(0.5, 35.5, 0.5).
 
     Returns
     -------
@@ -1688,7 +1688,7 @@ def slidingRP_violations(
     max_ref_period_ms : float, default: 10
         Maximum refractory period to test in ms.
     contamination_values : 1d array or None, default: None
-        The contamination values to test, if None it is set to np.arange(0.5, 35, 0.5) / 100.
+        The contamination values to test, if None it is set to np.arange(0.5, 35.5, 0.5) / 100.
     return_conf_matrix : bool, default: False
         If True, the confidence matrix (n_contaminations, n_ref_periods) is returned.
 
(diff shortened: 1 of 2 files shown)


oliche:
The original repo fix is here for reference: https://github.com/SteinmetzLab/slidingRefractory/pull/8

zm711:
I think @DradeAW wrote the initial implementation based on working with Llobet right Aurélien? If you have a moment to comment on this that would be great.

chrishalcrow:
Hi @oliche, thanks for pushing your changes to si! I thought I'd try to verify the change in the Llobet formula but I'm getting confused when reading the original paper/calculation. Some more specific referencing would help me: where in the paper is the discussion about the expected violations? And doesn't this depend on the modeling assumptions: e.g. whether the contamination is due to a single contaminating neuron or totally random noise? Thanks!

DradeAW:
Hi all!
I see you referencing Llobet et al. 2022, but we have a (very) updated version of the pre-print : https://www.biorxiv.org/content/10.1101/2022.02.08.479192v2
Everything is explained in the methods.
In short: the formula is 100% correct while the censored period is 0. But with a censored period, the correction actually depends on the ISI without contamination (which is unknown!). Hence, we computed an approximation which isn’t perfect as perfection cannot be achieved (to our knowledge). Probably a better approximation can be achieved.
I’ll look at your code later this afternoon :)

DradeAW:
So I looked over some code:
The Llobet formula is something I implemented in SpikeInterface with the functions `compute_refrac_period_violations`, `_compute_rp_contamination_one_unit` and `_compute_nb_violations_numba`.
I looked over the code, and it seems correct.

You are proposing a change in the function `_compute_violations`, which is used in `slidingRP_violations`.
I never wrote this function and never used it, so I cannot help you guys on this subject :/

However, if there are questions about the Llobet formula, I would be happy to answer you :)

Best to you guys!

oliche:
Yes this is a mistake in the original implementation of the slidingRP metric, and it is code specific to this metric. 
I would keep this formula so that it is consistent with the original repository current implementation !

One other thing I could suggest, is to avoid vendoring this metric code and just add it as an optional dependency. It currently has very minimal deps (numpy , stats, matplotlib).

alejoe91 (approved):

alejoe91:
> Yes this is a mistake in the original implementation of the slidingRP metric, and it is code specific to this metric. I would keep this formula so that it is consistent with the original repository current implementation !
> 
> One other thing I could suggest, is to avoid vendoring this metric code and just add it as an optional dependency. It currently has very minimal deps (numpy , stats, matplotlib).

@oliche hopefully the formula won't recieve many fixes! I think that since the implementation here is rather well scoped, I would keep it here. Let's merge and discuss about that option.

alejoe91:
Issue here: https://github.com/SpikeInterface/spikeinterface/issues/4704

PR #4686, opened 2026-07-13
Use MEDiCINe's own recommended default for num_depth_bins

nwatters01:
## Summary
- The `medicine` motion estimation preset (`MedicineRegistration.run` in `src/spikeinterface/sortingcomponents/motion/medicine.py`) computed `num_depth_bins` by dividing the probe's depth span by `win_scale_um`, a generic windowing parameter tuned for other methods (`dredge`, `decentralized`, etc.), defaulting to 300 µm. For a typical probe this yields far more depth bins than MEDiCINe is designed for.
- MEDiCINe's own package (`medicine.run.run_medicine`) defaults and documents `num_depth_bins=2` as sufficient for all of its authors' recording sessions, independent of any window-scale concept.
- All other MEDiCINe hyperparameter defaults in `MedicineRegistration.run` (`motion_bound`, `time_kernel_width`, `activity_network_hidden_features`, `amplitude_threshold_quantile`, `batch_size`, `training_steps`, `initial_motion_noise`, `motion_noise_steps`, `learning_rate`, `epsilon`) already match `medicine`'s own defaults exactly — `num_depth_bins` was the one outlier.
- This PR adds an explicit `num_depth_bins=None` parameter to `MedicineRegistration.run`. When unset, it now defaults to `1 if rigid else 2` (MEDiCINe's own recommendation) instead of being derived from `win_scale_um`/`win_margin_um`. Users who want a different value can still pass `num_depth_bins` explicitly through `estimate_motion_kwargs`.
- Also filled in the previously empty `params_doc` docstring for this method, and removed a leftover debug `print()`.
- No other preset or the shared `estimate_motion()` signature/defaults were touched, so this is backwards compatible for all other motion estimation methods.

## Dependencies
- `medicine-neuro` was not previously referenced anywhere in `pyproject.toml`. Added it to the `test-sortingcomponents` dependency group (alongside `torch`, following the same pattern) so the new test below actually runs in CI rather than being skipped, and added an explanatory comment next to the existing `torch` comment in `[project.optional-dependencies].sortingcomponents` noting it's an optional heavy dep (itself pulling in `torch`) needed only for the `medicine` method — not bundled into that extra by default, same treatment as `torch`.

## Test plan
- [x] Added `src/spikeinterface/sortingcomponents/motion/tests/test_medicine.py` (skipped if the optional `medicine`/`torch` packages aren't installed):
  - Parametrized check that `num_depth_bins` resolves to MEDiCINe's own default (2 non-rigid / 1 rigid), independent of `win_scale_um`.
  - Direct equivalence check: given identical peaks and RNG state (`torch.manual_seed` / `np.random.seed`), the spikeinterface `medicine` `estimate_motion` wrapper produces output numerically identical to calling `medicine.run_medicine()` directly, confirming the wrapper's resolved settings and outputs match the original package.
- [x] `python -m pytest src/spikeinterface/sortingcomponents/motion/tests/` (full directory, 12 passed, no regressions)
- [x] `python -m py_compile src/spikeinterface/sortingcomponents/motion/medicine.py`

### DIFF
--- a/src/spikeinterface/sortingcomponents/motion/medicine.py
+++ b/src/spikeinterface/sortingcomponents/motion/medicine.py
@@ -67,7 +67,6 @@ class MedicineRegistration:
         ## medicine specific kwargs propagated to the lib
         output_dir=None,
         plot_figures=False,
-        num_depth_bins=None,
         motion_bound=800,
         time_kernel_width=30,
         activity_network_hidden_features=(256, 256),
@@ -79,6 +78,7 @@ class MedicineRegistration:
         optimizer=None,
         learning_rate=0.0005,
         epsilon=1e-3,
+        num_depth_bins=None,
     ):
 
         from medicine import run_medicine
--- a/src/spikeinterface/sortingcomponents/motion/medicine.py
+++ b/src/spikeinterface/sortingcomponents/motion/medicine.py
@@ -14,7 +14,37 @@ class MedicineRegistration:
     name = "medicine"
     need_peak_location = True
     params_doc = """
-
+    output_dir : str | Path | None, default: None
+        Directory to save MEDiCINe outputs (parameters, figures, motion arrays). If None, nothing is saved to disk.
+    plot_figures : bool, default: False
+        Whether MEDiCINe should plot and save figures summarizing the model results.
+    num_depth_bins : int | None, default: None
+        Number of depth bins for motion estimation, passed to MEDiCINe. If None, this defaults to MEDiCINe's own
+        recommended default: 1 if `rigid` else 2. Note this is intentionally independent of the generic
+        `win_scale_um` / `win_step_um` / `win_margin_um` windowing parameters used by other motion estimation
+        methods, since MEDiCINe is not tuned for the number of depth bins those would otherwise imply.
+    motion_bound : float, default: 800
+        Bound on the maximum absolute motion allowed, in the same units as the spike depths (typically microns).
+    time_kernel_width : float, default: 30
+        Width of the temporal smoothing kernel, in the same units as the spike times (typically seconds).
+    activity_network_hidden_features : tuple, default: (256, 256)
+        Hidden layer sizes for MEDiCINe's activity network.
+    amplitude_threshold_quantile : float, default: 0.0
+        Cutoff quantile in [-1, 1] for peak amplitudes. See MEDiCINe's documentation for details.
+    batch_size : int, default: 4096
+        Batch size used for training.
+    training_steps : int, default: 10000
+        Number of optimization steps to take.
+    initial_motion_noise : float, default: 0.1
+        Initial magnitude of noise added to the motion function output, annealed to 0 over `motion_noise_steps`.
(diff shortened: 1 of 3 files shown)


nwatters01:
Some context:  I have received emails from users complaining about the fact that spikeinterface's default parameters for Medicine are not appropriate.  This PR changes this to make the default parameters ones that Medicine works well for.

This change is completely backwards compatible, and no other motion estimation methods are affected.

JoeZiminski:
This makes sense to me thanks for this. Would it be possible to add a test which runs the SI wrapper and the original MEDiCINe package to ensure the outputs / settings match?

src/spikeinterface/sortingcomponents/motion/medicine.py
```diff
@@ -37,6 +67,7 @@ def run(
         ## medicine specific kwargs propagated to the lib
         output_dir=None,
         plot_figures=False,
+        num_depth_bins=None,
```
zm711: curious about the positioning of this arg? I'm just thinking for backwards compatibility. If we put this at the bottom it won't break positional arguments for people but has the same effect.

zm711 (commented):
This makes sense to me. Just one question about positioning of the args.

src/spikeinterface/sortingcomponents/motion/medicine.py
nwatters01: Thanks, I've now moved it to the bottom!

nwatters01:
Thank you for your suggestions.

I've moved the num_depth_bins argument to the bottom to ensure backwards compatibility, and added a test file to test that this wrapper gives the results as the medicine package with default parameters.  I ran the test locally to make sure it passes.

chrishalcrow:
Thanks for this! The new tests aren't currently running on the ci: if you look at https://github.com/SpikeInterface/spikeinterface/actions/runs/29302001612/job/86987545183?pr=4686#step:30:378 and search "medicine" you'll see

```
SKIPPED [1] src\spikeinterface\sortingcomponents\motion\tests\test_medicine.py:5: could not import 'medicine': No module named 'medicine'
```

You should be able to add it to the sorting-components test suite by adding medicine to the pyproject.toml here:

https://github.com/SpikeInterface/spikeinterface/blob/c33abf56fc81604c16d8924131ba09ee92d0a529/pyproject.toml#L309

alejoe91 (approved):

nwatters01:
Thanks for the comments, and Alessio for approval!

I've now fixed the test dependencies, so the tests run.  I'll wait for the rest of the day to hear if there's anything else, and otherwise will merge first thing tomorrow.

chrishalcrow:
Thanks. The tests are passing and take less than 5s on the slowest ci (Windows 3.10) - perfect.

chrishalcrow (approved):

nwatters01:
I think we're all set, thanks!  Can you please merge this?

nwatters01:
A couple quick questions from users:

1.  Will this change be in the next spikeinterface release?

2.  When do you expect that release to be?

Thanks!

PR #4689, opened 2026-07-14
Lazy handling and saving of timestamps (and remove `load_from`/`save_to folder_metadata`)

alejoe91:
In the direction of #4671, this PR starts by rmeoving the unused mechanism to save to / load from metadata.

In addition, this PR also fixes saving timestamps in proper binary format.
For recording objects, the `_extra_metadata_to_folder` was saving timestamps to `npy` files, but this is now handled by the `write_binary` and `BinaryRecordingExtractor`, which writes in chunk to proper binary files. The `binaryRecordingExtractor` acceprs a `file_timestamps_path` with the location of the raw timestamps. Note that this is absolutely necessary for chronic data, since the old version was saving the entire timestamps array with numpy and this cannot be done with very long recordings. Old datasets are handled with the `_handle_extractor_backward_compatibility`.

To further cleanup, this PR adds `save_properties_to/load_properties_from_binary_folder` to remove duplicated code for saving and loading properties to numpy arrays.

The cross serialization check was extended to always run against major past versions, to test a recording with timestamps, and to save recordings in parallel for `zarr`/`binary_folder`.

### DIFF
--- a/src/spikeinterface/core/baserecording.py
+++ b/src/spikeinterface/core/baserecording.py
@@ -7,7 +7,12 @@ from probeinterface import read_probeinterface, write_probeinterface
 
 from .time_series import TimeSeriesSegment, TimeSeries
 from .baserecordingsnippets import BaseRecordingSnippets
-from .core_tools import convert_bytes_to_str, convert_seconds_to_str
+from .core_tools import (
+    convert_bytes_to_str,
+    convert_seconds_to_str,
+    save_properties_to_binary_folder,
+    load_properties_from_binary_folder,
+)
 from .job_tools import split_job_kwargs
 
 
@@ -330,8 +335,21 @@ class BaseRecording(BaseRecordingSnippets, TimeSeries):
             file_paths = [folder / f"traces_cached_seg{i}.raw" for i in range(self.get_num_segments())]
             dtype = kwargs.get("dtype", None) or self.get_dtype()
             t_starts = self._get_t_starts()
+            time_vectors = self._get_time_vectors()
+            if time_vectors is not None:
+                file_timestamps_paths = [folder / f"times_cached_seg{i}.raw" for i in range(self.get_num_segments())]
+                kwargs["file_timestamps_paths"] = file_timestamps_paths
+            else:
+                file_timestamps_paths = None
 
-            write_binary(self, file_paths=file_paths, dtype=dtype, verbose=verbose, **job_kwargs)
+            write_binary(
+                self,
+                file_paths=file_paths,
+                dtype=dtype,
+                verbose=verbose,
+                file_timestamps_paths=file_timestamps_paths,
+                **job_kwargs,
+            )
 
             # This is created so it can be saved as json because the `BinaryFolderRecording` requires it loading
             # See the __init__ of `BinaryFolderRecording`
@@ -348,16 +366,14 @@ class BaseRecording(BaseRecordingSnippets, TimeSeries):
                 is_filtered=self.is_filtered(),
                 gain_to_uV=self.get_channel_gains(),
                 offset_to_uV=self.get_channel_offsets(),
+                file_timestamps_paths=file_timestamps_paths,
             )
             binary_rec.dump(folder / "binary.json", relative_to=folder)
-            cached = BinaryFolderRecording(folder_path=folder)
 
-            # timestamps are not saved in binary, so we have to set them explicitly
-            for segment_index in range(self.get_num_segments()):
-                if self.has_time_vector(segment_index):
-                    # the use of get_times is preferred since timestamps are converted to array
-                    time_vector = self.get_times(segment_index=segment_index)
-                    cached.set_times(time_vector, segment_index=segment_index)
+            # save properties
+            save_properties_to_binary_folder(folder / "properties", self)
+
+            cached = BinaryFolderRecording(folder_path=folder)
 
         elif format == "memory":
             if kwargs.get("sharedmem", True):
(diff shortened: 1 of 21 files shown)


src/spikeinterface/core/tests/test_baserecording.py
```diff
@@ -494,6 +492,34 @@ def test_time_slice_with_time_vector():
...
+    times = rec.get_times(segment_index=0) + 100.0
+
+    # Set time vector
+    rec.set_times(times=times, segment_index=0, with_warning=False)
+    # Save
+    rec_saved = rec.save(folder=cache_folder / "recording_with_time_vector", format="binary")
+    assert np.allclose(rec.get_times(segment_index=0), rec_saved.get_times(segment_index=0))
+
+    # Save
+    rec_saved_par = rec.save(folder=cache_folder / "recording_with_time_vector_par", format="binary")
```
alejoe91:
```suggestion
    rec_saved_par = rec.save(folder=cache_folder / "recording_with_time_vector_par", format="binary", n_jobs=2)
```

src/spikeinterface/core/tests/test_baserecording.py
```diff
@@ -494,6 +492,34 @@ def test_time_slice_with_time_vector():
     assert np.allclose(sliced_recording_times.get_traces(), sliced_recording_frames.get_traces())
 
 
+def test_save_load_binary_with_time_vector(create_cache_folder):
```
alejoe91: test with different mp_contexts

src/spikeinterface/core/baserecording.py
```diff
@@ -348,16 +366,14 @@ def _save(self, format="binary", verbose: bool = False, **save_kwargs):
...
-            cached = BinaryFolderRecording(folder_path=folder)
 
-            # timestamps are not saved in binary, so we have to set them explicitly
-            for segment_index in range(self.get_num_segments()):
-                if self.has_time_vector(segment_index):
-                    # the use of get_times is preferred since timestamps are converted to array
-                    time_vector = self.get_times(segment_index=segment_index)
-                    cached.set_times(time_vector, segment_index=segment_index)
+            # save properties
+            save_properties_to_binary_folder(folder / "properties", self)
```
alejoe91: dump `probegroup.json` here too

alejoe91:
@chrishalcrow my feeiling was correct: in-memory timestamps are not propagated when saving in parallel...I'll propose a fix.

src/spikeinterface/core/baserecording.py
```diff
@@ -330,8 +330,21 @@ def _save(self, format="binary", verbose: bool = False, **save_kwargs):
             file_paths = [folder / f"traces_cached_seg{i}.raw" for i in range(self.get_num_segments())]
             dtype = kwargs.get("dtype", None) or self.get_dtype()
             t_starts = self._get_t_starts()
+            time_vectors = self._get_time_vectors()
+            if time_vectors is not None:
+                file_timestamps_paths = [folder / f"times_cached_seg{i}.raw" for i in range(self.get_num_segments())]
+                kwargs["file_timestamps_paths"] = file_timestamps_paths
```
chrishalcrow: I think this is not used
```suggestion
```

src/spikeinterface/core/baserecording.py
```diff
@@ -396,26 +411,15 @@ def _save(self, format="binary", verbose: bool = False, **save_kwargs):
...
-                np.save(folder / f"times_cached_seg{segment_index}.npy", time_vector)
+    def _extra_metadata_to_dict(self, dump_dict, _in_reduce: bool = False):
+        super()._extra_metadata_to_dict(dump_dict, _in_reduce=_in_reduce)
+
+        # Add times_kwargs if the recording has been modified in memory (e.g. by set_times or shift_times)
+        if _in_reduce and self._time_info_modified:
+            dump_dict["times_kwargs"] = []
+            for segment_index in range(self.get_num_segments()):
+                times_kwargs = self.segments[segment_index].get_times_kwargs()
+                dump_dict["times_kwargs"].append(times_kwargs)
```
chrishalcrow, line 422: When is this code called in practice?

src/spikeinterface/core/core_tools.py
```diff
@@ -781,3 +781,44 @@ def is_path_remote(path: str | Path) -> bool:
...
+
+    Parameters
+    ----------
+    folder : str or Path
+        The folder where the properties will be saved as .npy files.
+    extractor : BaseExtractor
+        The extractor from which the properties will be saved.
+    """
+    folder = Path(folder)
+    folder.mkdir(parents=True, exist_ok=True)
```
chrishalcrow:
```suggestion
```

This should be, and is, handled elsewhere

src/spikeinterface/core/frameslicerecording.py
```diff
@@ -68,8 +68,6 @@ def __init__(self, parent_recording_segment, start_frame, end_frame):
         d = d.copy()
         if d["time_vector"] is None:
             d["t_start"] = parent_recording_segment.sample_index_to_time(start_frame)
-        else:
-            d["time_vector"] = d["time_vector"][start_frame:end_frame]
```
chrishalcrow, line 72: Now, the `time_vector` of a `FrameSliceRecording` is the time vector of the parent: e.g. the full recordings `time_vector`.

But if you save a `FrameSliceRecording` as a binary file, it will save using `get_times`.

So if you load the sliced recording, this will have the reduced `time_vector`. So the save/load loop doesn't preserve all information.

I think the `FrameSliceRecording` should have the sliced time vector? This should still work lazily with memmap?

src/spikeinterface/core/baserecording.py
alejoe91: when serializing/deserialising with the `__reduce__`, which is mainly done when parallelizing and passing objects across processes (see https://github.com/alejoe91/spikeinterface/blob/save-timestamps/src/spikeinterface/core/base.py#L802-L808)

src/spikeinterface/core/frameslicerecording.py
alejoe91: this is correct because the save dumps a snapshot of the current state. I rewrote the logic to make it clearer though

src/spikeinterface/core/time_series_tools.py
```diff
@@ -139,15 +155,26 @@ def _write_binary_chunk(segment_index, start_frame, end_frame, worker_ctx):
...
 
     if file_timestamps_dict is not None:
-        file_timestamps = file_timestamps_dict[segment_index]
-        timestamps = time_series.get_times(start_frame=start_frame, end_frame=end_frame, segment_index=segment_index)
-        timestamps = timestamps.astype("float64", order="c", copy=False)
-        timestamp_byte_offset = start_frame * 8  # 8 bytes for float64
-        file.seek(timestamp_byte_offset)
-        file.write(timestamps.data)
-        file.flush()
+        # Some segments might not have
```
alejoe91:
```suggestion
        # Some segments might not have timestamps to save
```

alejoe91:
Ported all the timestamps refactoring to #4671 

Closing

PR #4691, opened 2026-07-14
Relax pandas<3 pin

### LINKED ISSUE #4687: read_phy / PhySortingExtractor raises "assignment destination is read-only" under pandas >= 3.0
### Description

`BasePhyKilosortSortingExtractor.__init__` writes into a numpy array obtained from `DataFrame.values`. Since **pandas 3.0** made Copy-on-Write permanent, `.values` returns a **read-only** array, so the in-place assignment raises `ValueError: assignment destination is read-only`. This makes `read_phy` fail on any Phy folder that spikeinterface itself exported (i.e. one containing an `si_unit_id` column).

This was previously noted in passing by @MarinManuel in [#4231 (comment)](https://github.com/SpikeInterface/spikeinterface/issues/4231#issuecomment-4146432803); filing it as a dedicated bug with a minimal reproduction and a proposed fix. It is downstream-visible: it breaks neuroconv's `PhySortingInterface` (and NWB GUIDE's tutorial, https://github.com/NeurodataWithoutBorders/nwb-guide/pull/1080) whenever pandas 3.0 is installed. Note `spikeinterface[full]` pins `pandas<3` and so is unaffected, but the base install is not.

### Location

[`src/spikeinterface/extractors/phykilosortextractors.py`](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/extractors/phykilosortextractors.py) in `BasePhyKilosortSortingExtractor.__init__` (line numbers from current `main`):

```python
168:  unit_ids = cluster_info["si_unit_id"].values   # read-only under pandas >= 3.0
...
183:  unit_ids[i] = new_si_id                         # ValueError: assignment destination is read-only
```

### Minimal reproduction

Uses only spikeinterface + its dependencies. Verified on `main` and on 0.104.8.

```python
import tempfile
from pathlib import Path

import numpy as np
import pandas as pd
import spikeinterface
from spikeinterface.extractors import read_phy

folder = Path(tempfile.mkdtemp()) / "phy"
folder.mkdir()

np.save(folder / "spike_times.npy", np.array([10, 20, 30, 40], dtype="int64"))
np.save(folder / "spike_clusters.npy", np.array([0, 1, 0, 1], dtype="int64"))
(folder / "params.py").write_text("sample_rate = 30000.\n")
# `si_unit_id` is the column spikeinterface writes when it exports a sorting to Phy.
(folder / "cluster_info.tsv").write_text("cluster_id\tsi_unit_id\n0\t0\n1\t1\n")

print("spikeinterface", spikeinterface.__version__, "| pandas", pd.__version__)
read_phy(folder_path=folder)   # ValueError: assignment destination is read-only  (pandas >= 3.0)
```

- pandas 2.3.3 -> works.
- pandas 3.0.0 -> raises.

### Traceback

```
File ".../spikeinterface/extractors/phykilosortextractors.py", line 183, in __init__
    unit_ids[i] = new_si_id
ValueError: assignment destination is read-only
```

### Suggested fix

Take an owned, writeable copy before mutating:

```python
unit_ids = cluster_info["si_unit_id"].to_numpy(copy=True)
```

(Worth auditing the other `cluster_info[...].values` uses in this method for the same pattern.)

### Environment

- spikeinterface: `main` (also reproduced on 0.104.8); the buggy line is identical back to 0.103.0, so it is version-independent
- pandas: 3.0.0
- numpy: 2.x
- Python: 3.12

alejoe91:
Fixes #4687

### DIFF
--- a/src/spikeinterface/widgets/bombcell_curation.py
+++ b/src/spikeinterface/widgets/bombcell_curation.py
@@ -153,6 +153,12 @@ class BombcellUpsetPlotWidget(BaseWidget):
                 if len(upset_data) == 0:
                     continue
 
+                if upset_data.index.nlevels == 1:
+                    warnings.warn(
+                        f"Skipping UpSet plot: only one metric, {memberships[0][0]}, was responsible for all failures."
+                    )
+                    continue
+
                 fig = plt.figure(figsize=(12, 6))
                 UpSet(
                     upset_data,
--- a/.github/scripts/check_kilosort4_releases.py
+++ b/.github/scripts/check_kilosort4_releases.py
@@ -15,10 +15,8 @@ def get_pypi_versions(package_name):
     response.raise_for_status()
     data = response.json()
     versions = list(sorted(data["releases"].keys()))
-    # Filter out versions that are less than 4.0.16 and different from 4.0.26 and 4.0.27
-    # (buggy - https://github.com/MouseLand/Kilosort/releases/tag/v4.0.26)
-    versions = [ver for ver in versions if parse(ver) >= parse("4.0.16") and
-                parse(ver) not in [parse("4.0.26"), parse("4.0.27")]]
+    # Filter out versions that are less than 4.1.1, since they don't support numpy 2.0
+    versions = [ver for ver in versions if parse(ver) >= parse("4.1.1")]
     return versions
 
 
--- a/.github/scripts/test_kilosort4_ci.py
+++ b/.github/scripts/test_kilosort4_ci.py
@@ -28,7 +28,8 @@ import torch
 
 import spikeinterface.full as si
 from spikeinterface.core.testing import check_sortings_equal
-from spikeinterface.sorters.external.kilosort4 import Kilosort4Sorter
+from spikeinterface.sorters.external.kilosort4 import Kilosort4Sorter, read_kilosort4_motion
+from spikeinterface.core.motion import Motion
 from probeinterface.io import write_prb
 from spikeinterface.extractors import read_kilosort_as_analyzer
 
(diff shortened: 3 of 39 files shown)


h-mayorquin (approved):
LGTM

It is is a breaking change or users who for some reason can't use the older pandas but I guess they have older versions of SpikeInterface for that,

chrishalcrow:
I'm not sure that dropping 3.10 support is worth it to support pandas 3 - I think we should at least discuss it. This would leave us only supporting three versions of Python.

(I'm ok with the PR from a technical standpoint though!)

zm711:
I agree with Chris. 3.10 isn't EOL until October so removing support before then at least seems premature to me.

alejoe91:
> I agree with Chris. 3.10 isn't EOL until October so removing support before then at least seems premature to me.

Both pandas 3 and zarr 3 dropped support for 3.10. I don't think this will be an issue. Hopefully zarr v3 will be in main by then too

alejoe91:
> LGTM
> 
> It is is a breaking change or users who for some reason can't use the older pandas but I guess they have older versions of SpikeInterface for that,

And I agree with @h-mayorquin here too. If someone wants to use 3.10, we have plenty of older versions (including the current one) with support :)

alejoe91:
Plus, we will support 3.14 by next release (#4173 is in milestone 0.105.0)

chrishalcrow:
Tested locally, and none of the pandas 3 changes we used are breaking. So we don't need to pin pandas, and hence don't need to remove 3.10 support. 
(Checking with full tests now...)

Note: The 3.10 tests are using pandas<3 and the 3.13 tests are using pandas>3, as expected

src/spikeinterface/widgets/bombcell_curation.py
```diff
@@ -153,6 +153,12 @@ def plot_matplotlib(self, data_plot, **backend_kwargs):
                 if len(upset_data) == 0:
                     continue
 
+                if upset_data.index.nlevels == 1:
+                    warnings.warn(
+                        f"Skipping UpSet plot: only one metric, {memberships[0][0]}, was responsible for all failures."
+                    )
+                    continue
+
```
chrishalcrow, line 161: UpSet plot requires there to be at least two metrics which cause "failures" (units != sua). Sometimes, only `amplitude_cutoff` was causing all failures, meaning it couldn't plot and would error.

alejoe91: good catch!

chrishalcrow (approved):

Issue #4683, opened 2026-07-10
What should `to_reordered_spike_vector` do?


grahamfindlay:
@alejoe91 @samuelgarcia @zm711 @chrishalcrow @h-mayorquin 

Prior to #4606 , the vanilla spike vector was lexsorted on keys `("unit_index", "sample_index", "segment_index")`. Now it is lexsorted on `("sample_index", "segment_index")`. This raises a question about what the behavior of `to_reordered_spike_vector` should be. 

Here is the function signature:
```python
def to_reordered_spike_vector(
        self,
        lexsort=("sample_index", "segment_index", "unit_index"), 
        return_order=True,
        return_slices=True,
    ):
```
- `lexsort` is the _requested_ sort order, as keys would be passed to `np.lexsort`. 
- `return_order=True` requests that the numpy array needed to sort the spike vector (given the requested sort) is also returned. 
- `return_slices=True`: requests  the numpy array of size `(num_units, num_segments, 2)` or `(num_segments, num_units, 2)` given the lexsort, where one can obtain the indices `amin, amax` of all the (`segment, unit_index)` values.

This function mainly exists to reorder the vanilla spike vector to `("sample_index", "segment_index", "unit_index")`, so unit spike trains are compact in memory. In the current codebase, it is only ever called as `to_reordered_spike_vector(lexsort=("sample_index", "segment_index", "unit_index"), ...)`. Nevertheless, it does allow users to request `lexsort=("sample_index", "unit_index", "segment_index")` or `lexsort=("unit_index", "sample_index", "segment_index")`. It is this last option that presents a dilemma.  

Previously if a user requested `lexsort=("unit_index", "sample_index", "segment_index")` this just built the vanilla spike vector (or fetched it from the cache). There was nothing to reorder, and no compact 2D `(unit, segment)` table could possibly exist for that ordering, so `return_order` and `return_slices` had to be `False`. 

Currently, if a user requests `lexsort=("unit_index", "sample_index", "segment_index")`, the same thing happens (vanilla spike vector retrieved), except now the spike vector that gets returned doesn't match the requested lexsort order. The function implies a tie-break that isn't there. This is the issue. 

#### Question A: How to resolve this? Some options:

1. Don't allow the user to request `lexsort=("unit_index", "sample_index", "segment_index")`. If they want the vanilla spike vector, make them call `to_spike_vector`. 
[...]

Issue #4606, opened 2026-06-03
Does spike vector need a full lexsort ?


samuelgarcia:
After trying to fasten the kilosort/phy reading and subselecting units with sevral way:
 * #4502
 * #4579 

It appears that the main slow down is to force the spike vector to be fully lexsorted:

`spikes = spikes[np.lexsort((spikes["unit_index"], spikes["sample_index"], spikes["segment_index"]))]`

It use to be partially lexsorted : 
`spikes = spikes[np.lexsort((spikes["sample_index"], spikes["segment_index"]))]`

This commit recent commit https://github.com/SpikeInterface/spikeinterface/commit/456265c9d48e2e749861660c4bf0e3760027f6a0 has introduced this "need of fully lexsorted".

**In short does 2 spikes with same sample_index need to be ordered by unit_index ?**

**Yes : we need full lexsort**

Pro : 
 * It ensure always an order of the spiek vector because spike amplitude, amplitude scaling, spike location is a vector related to this spike vector.

Cons :
 * it is a big slow down when selecting units 
 * constructing the spike vector (with many spikes) can be very slow

**No : we need partial lexsort**

Pro : 
 * We do not care 2 spikes of same sample_index to be ordered by units for any computation (CCG)
 * We only need to ensure that the spike vector is always constructed/read/cached in the same consistent order (for instance by reading the phy/kilosort spike_clusters.npy)

Cons :
 * some future part of code could rely on the fact this vector is fully lexsorted

My personnal taste is second case.

**The main problem** : the spikeinterface codebase use to be *No : we need partial lexsort* and in february 2025 we switched to *Yes : we need full lexsort* without too much backward compatibility in mind. Zarr is now opening the internal spike vector and force lexsort in the `__init__()` which is a big big slow down but also can break the order of spike amplitude, amplitide scaling, ... it this extension was computed before this change (v 0.104.0) and the analyzer was saved in the zarr format before this version.

I would strongly vote for removing the behavior (force full lexsort) and go back to previous situation. But we need to find a better way for backward compatibility to avoid the same mistake.

@grahamfindlay @alejoe91 @chrishalcrow @yger @tayheau
Any opinion on this ?

PS : the first reader able to find more than 10 typos in this post will have a glass of wine.

alejoe91:
> PS : the first reader able to find more than 10 typos in this post will have a glass of wine.

Ahahahaha too easy!

tayheau:
[...]

PR #4695, opened 2026-07-17
Fix and speed up to_reordered_spike_vector

grahamfindlay:
Follows #4606 and #4683. Implements part of #4502. @alejoe91 @samuelgarcia 

1. Fixes the always true `elif ("sample_index", "unit_index", "segment_index"):` in `to_reordered_spike_vector`. 
2. Restricts `to_reordered_spike_vector` to non-vanilla orderings, and fixes some inaccuracies in the docstring. 
3. Optimizes `to_reordered_spike_vector`. There is a numba path with a counting sort that drops the time on an ~400M spike / 342 unit / 1 segment sorting from 137s to 12s, with no additional memory usage. There is a numpy fallback path using a radix sort that drops the time from 137s to 85s, with a slightly increased memory cost (34 bytes/spike instead of 32 bytes/spike). The outputs (`ordered_spikes`, `slices`, `order`) are identical to the old implementation. I also tested on synthetic data of various sizes (1M, 10M, 40M spikes) to make sure both paths are strict improvements at all input sizes. 
4. Fixes a bug where the reorder cache was set using a string, then probed by various callers using a tuple (or list), so the cache was being silently missed basically all the time. I defined some constants for the keys and used them everywhere, and made the validation stricter, to minimize the likelihood that it happens again.

### DIFF
--- a/src/spikeinterface/core/basesorting.py
+++ b/src/spikeinterface/core/basesorting.py
@@ -1046,31 +1046,24 @@ class BaseSorting(BaseExtractor):
 
         Please note that the lexsort syntax is the **reverse** of natural reading.
 
-        By default the spike_vector is lexsort-ed like this:
-          - ("unit_index", "sample_index", "segment_index") (segment then sample then unit)
-
-        But particular orderings can be better for some computations:
-          - ("sample_index", "unit_index", "segment_index") (segment then unit_index then sample)
-          - ("sample_index", "segment_index", "unit_index") (unit_index then segment then sample)
-
-        Note that the last representation makes the spiketrain per segment compact in memory.
+        Two reorderings are supported:
+          - ("sample_index", "segment_index", "unit_index"): makes each unit's
+            spiketrain compact in memory. This is the default, and is what
+            unit-by-unit computations (e.g. isi violations) want.
+          - ("sample_index", "unit_index", "segment_index"): makes each segment compact,
+            with each unit's spiketrain compact within a segment. Rarely (if ever) used,
+            but might be useful when iterating segment by segment.
 
         This operation is internally cached.
 
-        The order vector is also computed and can be applied to other external vectors like
-        spike_amplitudes, spike_locations, ...
-
-        An array of internal slices is also precomputed to have a fast access to a compact
-        portion of the reordered spikes.
-        Theses slices are stored as a 3d array to handle start->stop and depend of the lexsort itself.
-        Theses slices are pre computed using nested searchsorted.
-
         Parameters
         ----------
-        lexsort : tuple, default: ("sample_index", "unit_index", "segment_index")
-            Tuple for lexsort. Please note that this is the reverse natural reading order!
+        lexsort : tuple, default: ("sample_index", "segment_index", "unit_index")
+            The requested sort order, as keys would be passed to np.lexsort.
+            Please note that this is the reverse natural reading order!
+            Must be one of the two orderings listed above.
         return_order: bool, default: True
-            Return the order, or not. See Returns.
+            Return the numpy array needed to sort the spike vector (given the requested sort).
         return_slices: bool, default: True
             Return the slices, or not. See Returns.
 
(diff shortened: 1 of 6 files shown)


grahamfindlay:
I was refreshing this to keep it up-to-date with main, and learned from two of the intervening PRs that (1) spike vectors can be lazy and (2) can have extra fields like `channel_index`. Then while trying to figure out what other fields can exist, I learned that they don't even have to all be `int64`. So I updated this PR to handle all those cases.

src/spikeinterface/core/basesorting.py
```diff
@@ -7,6 +7,14 @@
 from .base import BaseExtractor, BaseSegment, minimum_spike_dtype
 from .waveform_tools import has_exceeding_spikes
 
+#: Makes each unit's spiketrain compact in memory (unit, then segment, then sample).
+LEXSORT_UNIT_COMPACT = ("sample_index", "segment_index", "unit_index")
+#: Makes each segment compact, units compact within a segment (segment, then unit, then sample).
+LEXSORT_SEGMENT_COMPACT = ("sample_index", "unit_index", "segment_index")
```
samuelgarcia, line 13: thanks you

src/spikeinterface/core/sorting_tools.py
```diff
@@ -147,6 +149,207 @@ def vector_to_list_of_spiketrain_numba(sample_indices, unit_indices, num_units):
...
+
+    counts = np.bincount(bucket_index, minlength=num_buckets).astype(np.int64, copy=False)
+
+    # Narrow the bucket to the smallest dtype that fits. This saves memory but also time:
+    # The cost of the radix sort that follows scales with the width of the key,
+    bucket_index = bucket_index.astype(np.min_scalar_type(num_buckets - 1), copy=False)
+
+    order = np.argsort(bucket_index, kind="stable")  # radix sort, because of integer key
+    ordered_spikes = spike_vector[order]
+    return ordered_spikes, order, counts
```
samuelgarcia, line 283: very smart

src/spikeinterface/core/sorting_tools.py
```diff
@@ -147,6 +149,207 @@ def vector_to_list_of_spiketrain_numba(sample_indices, unit_indices, num_units):
...
+            np.zeros(num_buckets, dtype=np.int64),
+        )
+
+    out_of_range_error = (
+        f"`spike_vector` has a unit_index outside [0, {num_units}) or a segment_index outside [0, {num_segments})."
+    )
+
+    # The numba kernel expects an all-int64 unpadded dtype (e.g. `minimum_spike_dtype`), but it is
+    # possible that a spike vector has extra fields with other dtypes (`NumpySorting` allows that).
+    # So we check taht the numba path is safe, and anything else takes the dtype-agnostic numpy path.
```
samuelgarcia, line 250: If one day this assumption is broken (like segment_index move to int8 or unit_index to int32).
Do you think we could have a slower numba kernel that do the same but using the real fields ?
Using for instance the numba.from_dtype()

samuelgarcia:
Hi @grahamfindlay.
Thank you for this. I wanted to do push your numba after the refactoring I did but fail to find time to do it.
THis looks very good to me.

src/spikeinterface/core/sorting_tools.py
grahamfindlay: Yes, and I'm not even sure it would have to be slower. It may not even need `from_dtype()`. I wouldn't be surprised if `@jit` is smart enough to generate the machine code based on the full structured array layout the first time it seems a unique one, so you'll have to wait for a re-compile the first time a layout is seen in any given session, but after that the compiled kernel itself is probably just as fast. And I guess you'd get a teeny bit of that compile cost back by getting rid of the need for `_is_flat_int64_view`.

grahamfindlay:
@samuelgarcia I just brought this up-to-date with `main` so it is ready to merge. I just rebased, didn't make any changes.  Thanks for looking at it!

alejoe91 (approved):

alejoe91 (approved):

samuelgarcia (approved):

PR #4696, opened 2026-07-17
Dev version  should be 0.105.0

alejoe91:
We have 0.104.1 as version in main, but the dev version should be the next major release!


zm711 (approved):
Oops. Looks good to me.

PR #4699, opened 2026-07-20
Add option to export spike waveforms subset to phy (mimic phy extract-waveforms)

alejoe91:
## Summary
- Adds an `add_waveforms` option to `export_to_phy` that writes phy's `_phy_spikes_subset.{spikes,channels,waveforms}.npy` files, mimicking the `phy extract-waveforms` CLI command.
- Reuses the `SortingAnalyzer`'s `waveforms`/`random_spikes` extensions when already computed, and computes them on demand otherwise (skipping with a warning if there's no recording available).
- Handles both sparse and dense `SortingAnalyzer`/sparsity combinations, matching the sparsity already used for `templates.npy`/`pc_feature_ind.npy`.

## Test plan
- [x] Verified with `phylib.io.model.TemplateModel` (phy's own loader) that the exported files load correctly and return proper per-cluster waveforms, for both sparse and dense analyzers.
- [x] Verified stored waveform values exactly match manually-extracted raw-trace snippets (accounting for `nbefore`/uV scaling).
- [x] Add a unit test to `src/spikeinterface/exporters/tests/` covering `add_waveforms=True`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/spikeinterface/exporters/to_phy.py
+++ b/src/spikeinterface/exporters/to_phy.py
@@ -21,6 +21,7 @@ def export_to_phy(
     output_folder: str | Path,
     compute_pc_features: bool = True,
     compute_amplitudes: bool = True,
+    add_waveforms: bool = False,
     sparsity: Optional[ChannelSparsity] = None,
     copy_binary: bool = True,
     remove_if_exists: bool = False,
@@ -46,6 +47,12 @@ def export_to_phy(
         If True, pc features are computed
     compute_amplitudes : bool, default: True
         If True, waveforms amplitudes are computed
+    add_waveforms : bool, default: False
+        If True, a subset of spike waveforms is saved in the phy "_phy_spikes_subset.*" files
+        (mimicking the "phy extract-waveforms" command), so that phy/phy-lib can display
+        individual spike waveforms without needing the raw recording. This requires (and will
+        compute, if missing and a recording is available) the "waveforms" SortingAnalyzer
+        extension.
     sparsity : ChannelSparsity or None, default: None
         The sparsity object
     copy_binary : bool, default: True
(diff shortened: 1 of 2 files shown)


src/spikeinterface/exporters/to_phy.py
```diff
@@ -46,6 +47,13 @@ def export_to_phy(
         If True, pc features are computed
     compute_amplitudes : bool, default: True
         If True, waveforms amplitudes are computed
+    add_waveforms : bool, default: False
+        If True, a subset of spike waveforms is saved in the phy "_phy_spikes_subset.*" files
+        (mimicking the "phy extract-waveforms" command), so that phy/phy-lib can display
+        individual spike waveforms without needing the raw recording. This requires the
+        "waveforms" and "random_spikes" SortingAnalyzer extensions to already be computed
+        (e.g. via `sorting_analyzer.compute(["random_spikes", "waveforms"])`); if they are not,
+        a warning is raised and the files are not saved.
```
chrishalcrow:
```suggestion
        If True, a subset of spike waveforms is saved in the phy "_phy_spikes_subset.*" files, 
        so that phy can display individual spike waveforms without needing the raw recording.
```

chrishalcrow:
```suggestion
        If True, a subset of spike waveforms is saved in the phy "_phy_spikes_subset.*" files, 
        so that phy can display individual spike waveforms without needing the raw recording.
```

chrishalcrow:
I think the new release of phy no longer supports `_phy_spikes_subset.*` - I can't find any mention of this type of file in the code of the codebase. I propose to delay this for now.

alejoe91:
> I think the new release of phy no longer supports `_phy_spikes_subset.*` - I can't find any mention of this type of file in the code of the codebase. I propose to delay this for now.

This is handled in phylib and I tested with the most recent phy version:
https://github.com/cortex-lab/phylib/blob/master/phylib/io/model.py#L684-L702

Example with `add_waveforms=True`, exported after resetting recoriding (so waveforms must be coming from new files), phy==2.1.0
image

I don't think we should delay this ;)

chrishalcrow (approved):
Oh, there was a bug(?) in phy and no templates or waveforms were showing until I toggled and bunch of options. Ok, working for me now!

Just one request to make one of the docstrings less overly verbose.

